PR #4, opened 2015-04-22
Experimental data store layer to aid tracking client state

ara4n:
This is an experiment in adding a data store layer to let clients actually track all the events they receive and expose the state to clients in a sensible manner.

Events are wrapped up as MatrixEvent objects, which in practice do nothing useful at all other than possibly provide a way to annotate the events in future.  This feels deeply wrong, currently.

### DIFF
--- a/lib/matrix.js
+++ b/lib/matrix.js
@@ -256,6 +256,8 @@ var init = function(exports){
         //   local echo
         //   disambiguating display names in a room
         //   event dup suppression? - apparently we should still be doing so
+        //   tracking current display name / avatar per-message
+        //   pagination
         
         /*
          * Helper method for retrieving the name of a room suitable for display in the UI
@@ -283,7 +285,7 @@ var init = function(exports){
                 var userId = this.credentials.userId;
                 var members = this.store.getStateEvents(roomId, 'm.room.member')
                     .filter(function(event) {
-                        event.event.user_id !== userId;
+                        return event.event.user_id !== userId;
                     });
                 
                 if (members.length == 0) {
--- a/lib/matrix.js
+++ b/lib/matrix.js
@@ -110,13 +110,13 @@ var init = function(exports){
         
     MatrixInMemoryStore.prototype = {
         /*
-         * Add an array of one or more raw state events into the store, overwriting
+         * Add an array of one or more state MatrixEvents into the store, overwriting
          * any existing state with the same {room, type, stateKey} tuple.
          */
-        setRawStateEvents: function(stateEvents) {
+        setStateEvents: function(stateEvents) {
             // we store stateEvents indexed by room, event type and state key.
             for (var i = 0; i < stateEvents.length; i++) {
-                var event = stateEvents[i];
+                var event = stateEvents[i].event;
                 var roomId = event.room_id;
                 if (this.rooms[roomId] === undefined) {
                     this.rooms[roomId] = {};
@@ -127,7 +127,7 @@ var init = function(exports){
                 if (this.rooms[roomId].state[event.type] === undefined) {
                     this.rooms[roomId].state[event.type] = {};
                 }
-                this.rooms[roomId].state[event.type][event.state_key] = new MatrixEvent(event);
+                this.rooms[roomId].state[event.type][event.state_key] = stateEvents[i];
             }
         },
        
@@ -183,14 +183,14 @@ var init = function(exports){
         },
         
         /*
-         * Adds a list of arbitrary raw events into the store.
+         * Adds a list of arbitrary MatrixEvents into the store.
          * If the event is a state event, it is also updates state.
          */
-        setRawEvents: function(events) {
+        setEvents: function(events) {
             for (var i = 0; i < events.length; i++) {
-                var event = events[i];
+                var event = events[i].event;
                 if (event.type === "m.presence") {
-                    this.setRawPresenceEvents([event]);
+                    this.setPresenceEvents([events[i]]);
                     continue;
                 }
                 var roomId = event.room_id;
@@ -201,9 +201,9 @@ var init = function(exports){
                     this.rooms[roomId].timeline = [];
                 }
                 if (event.state_key !== undefined) {
-                    this.setRawStateEvents([event]);
+                    this.setStateEvents([events[i]]);
                 }
-                this.rooms[roomId].timeline.push(new MatrixEvent(event));
+                this.rooms[roomId].timeline.push(events[i]);
             }
         },
         
@@ -215,10 +215,10 @@ var init = function(exports){
             return this.room[roomId].timeline;
         },
         
-        setRawPresenceEvents: function(presenceEvents) {
+        setPresenceEvents: function(presenceEvents) {
             for (var i = 0; i < presenceEvents.length; i++) {
-                var event = presenceEvents[i];
-                this.presence[event.user_id] = new MatrixEvent(event);
+                var matrixEvent = presenceEvents[i];
+                this.presence[matrixEvent.event.user_id] = matrixEvent;
             }
         },
         
@@ -258,6 +258,7 @@ var init = function(exports){
         //   event dup suppression? - apparently we should still be doing so
         //   tracking current display name / avatar per-message
         //   pagination
+        //   reconnection - DONE
         
         /*
          * Helper method for retrieving the name of a room suitable for display in the UI
@@ -349,6 +350,10 @@ var init = function(exports){
                 if (err) {
                     console.error("err %s", JSON.stringify(err));
                     callback(err);
+                    // retry every few seconds
+                    setTimeout(function() {
+                        self._pollForEvents(callback);
+                    }, 2000);
                 } else {
                     var events = [];
                     for (var j = 0; j < data.chunk.length; j++) {
@@ -668,10 +673,22 @@ var init = function(exports){
                 function(err, data) {
                     if (self.store) {
                         // intercept the results and put them into our store
-                        self.store.setRawPresenceEvents(data.presence);
+                        self.store.setPresenceEvents(data.presence.map(
+                            function(event) {
+                                return new MatrixEvent(event);
+                            }
+                        ));
                         for (var i = 0 ; i < data.rooms.length; i++) {
-                            self.store.setRawStateEvents(data.rooms[i].state);
-                            self.store.setRawEvents(data.rooms[i].messages.chunk);
+                            self.store.setStateEvents(data.rooms[i].state.map(
+                                function(event) {
+                                    return new MatrixEvent(event);
+                                }
+                            ));
+                            self.store.setEvents(data.rooms[i].messages.chunk.map(
+                                function(event) {
+                                    return new MatrixEvent(event);
+                                }
+                            ));
                         }
                     }
                     if (data) self.fromToken = data.end;
@@ -726,7 +743,11 @@ var init = function(exports){
             return this._doAuthedRequest(
                 function(err, data) {
                     if (self.store) {
-                        self.store.setRawEvents(data.chunk);
+                        self.store.setEvents(data.chunk.map(
+                            function(event) {
+                                return new MatrixEvent(event);
+                            }
+                        ));
                     }
                     if (data) self.fromToken = data.end;
                     callback(err, data); // continue with original callback
--- a/lib/matrix.js
+++ b/lib/matrix.js
@@ -21,6 +21,9 @@ var init = function(exports){
      * Construct a Matrix Client.
      * @param {Object} credentials The credentials for this client
      * @param {Object} config The config (if any) for this client.
+     *  Valid config params include:
+     *      noUserAgent: true // to avoid warnings whilst setting UA headers
+     *      debug: true // to use console.err() style debugging from the lib
      * @param {Object} store The data store (if any) for this client.
      */
     function MatrixClient(credentials, config, store) {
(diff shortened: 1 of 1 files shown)


lib/matrix.js
```diff
@@ -52,11 +58,351 @@ var init = function(exports){
...
+    // Basic DAOs to abstract slightly from the line protocol and let the
+    // application customise events with domain-specific info
+    // (e.g. chat-specific semantics) if it so desires.
+    
+    /*
+     * Construct a Matrix Event object
+     * @param {Object} event The raw event to be wrapped in this DAO
+     */
+    function MatrixEvent(event) {
+        this.event = event;
```
Kegsay: You should probably be setting ``{ }`` as the default here else any of the ``get`` functions will throw.

lib/matrix.js
```diff
@@ -52,11 +58,351 @@ var init = function(exports){
...
+        
+    // XXX: this is currently quite procedural - we could possibly pass back
+    // models of Rooms, Users, Events, etc instead.
+    MatrixInMemoryStore.prototype = {
+                
+        /*
+         * Add an array of one or more state MatrixEvents into the store, overwriting
+         * any existing state with the same {room, type, stateKey} tuple.
+         */
+        setStateEvents: function(stateEvents) {
```
Kegsay: You have the single versions of most everything else but not this. I'd just accept ``stateEvents`` being an ``object`` then assigning it to an ``array`` if it is. Alternatively, you could add a ``setStateEvent`` method and then just call through to this.

lib/matrix.js
```diff
@@ -52,11 +58,351 @@ var init = function(exports){
...
+        },
+       
+        /*
+         * Return a list of MatrixEvents from the store, filtered by roomid, type and state key.
+         * @param {String} roomId the Room ID whose state is to be returned
+         * @param {String} type the type of the state events to be returned (optional)
+         * @param {String} stateKey the stateKey of the state events to be returned
+         *                 (optional, requires type to be specified)
+         */
+        getStateEvents: function(roomId, type, stateKey) {
```
Kegsay: Missing ``@return`` docs.

lib/matrix.js
```diff
@@ -52,11 +58,351 @@ var init = function(exports){
...
+            else {
+                return [this.rooms[roomId].state[type][stateKey]];
+            }
+        },
+        
+        /*
+         * Return a single state MatrixEvent from the store for the given roomId
+         * and type.  stateKey is optional; if missing, is assumed to be blank.
+         */
+        getStateEvent: function(roomId, type, stateKey) {
```
Kegsay: Missing jsdocs, which is particularly crucial here because of ``stateKey``. The description isn't clear what "blank" means for ``if missing, is assumed to be blank.`` Do you mean ``undefined`` or a 0-length string `""`?

lib/matrix.js
```diff
@@ -52,11 +58,351 @@ var init = function(exports){
...
+            if (stateKey === undefined) {
+                for (var i in this.rooms[roomId].state[type]) {
+                    if (this.rooms[roomId].state[type].hasOwnProperty(i)) {
+                        return this.rooms[roomId].state[type][i];
```
Kegsay: Related to my previous comment, the behaviour here is very strange. If you miss out ``stateKey`` it will return *any* matching event regardless of the state key value; we never explicitly check for `""`.

lib/matrix.js
```diff
@@ -52,11 +58,351 @@ var init = function(exports){
...
+         * Helper method for retrieving the name of a room suitable for display in the UI
+         * TODO: in future, this should be being generated serverside.
+         */
+        getFriendlyRoomName: function(roomId) {
+            // we need a store to track the inputs for calculating room names
+            if (!this.store) return roomId;
+            
+            // check for an alias, if any. for now, assume first alias is the official one.
+            var alias;
+            var mRoomAliases = this.store.getStateEvent(roomId, 'm.room.aliases');
```
Kegsay: See comments on ``getStateEvent``. This only works through the behaviour defined there. I always forget this, but ``m.room.aliases`` *does* have a `state_key`! The value is the domain of the server which assigned the aliases.

I suspect you meant this to have `""` as the `state_key` but that would never find the `m.room.aliases` event: you need it to loop through the `state_key` values which is what it happens to be doing in `getStateEvent`. If this was your intention, it should be made much more clear.

lib/matrix.js
```diff
@@ -52,11 +58,351 @@ var init = function(exports){
...
+         * and then start polling the eventStream for new events.
+         */
+        startClient: function(callback, historyLen) {
+            historyLen = historyLen || 12;
+            
+            var self = this;
+            if (!this.fromToken) {
+                this.initialSync(historyLen, function(err, data) {
+                    if (err) {
+                        console.error("err %s", JSON.stringify(err));
```
Kegsay: We shouldn't make too many assumptions on how this SDK should be used. People may have their own logging libraries setup which log in `callback(err)`. I don't propose removing this per se, just that error logging should be wrapped around an optional ``config.debug`` flag or similar.

lib/matrix.js
```diff
@@ -52,11 +58,351 @@ var init = function(exports){
...
+                this._pollForEvents(callback);
+            }
+        },
+        
+        _pollForEvents: function(callback) {
+            var self = this;
+            if (!this.clientRunning) return;
+            this.eventStream(this.fromToken, 30000, function(err, data) {
+                if (err) {
+                    console.error("err %s", JSON.stringify(err));
```
Kegsay: See earlier `console` comment.

lib/matrix.js
```diff
@@ -52,11 +58,351 @@ var init = function(exports){
...
+                }
+            }
+            return roomIds;
+        },
+        
+        // TODO
+        setMaxHistoryPerRoom: function(maxHistory) {},
+        
+        // TODO
+        reapOldMessages: function() {},
```
Kegsay: May be worth commenting these 2 functions out for now: we risk people assuming they are implemented otherwise, and spend hours thinking "why isn't it reaping, my heap is still huuuuge" until they check the source and think "Well *thanks*."

lib/matrix.js
```diff
@@ -52,11 +58,351 @@ var init = function(exports){
...
+            var self = this;
+            if (!this.clientRunning) return;
+            this.eventStream(this.fromToken, 30000, function(err, data) {
+                if (err) {
+                    console.error("err %s", JSON.stringify(err));
+                    callback(err);
+                    // retry every few seconds
+                    setTimeout(function() {
+                        self._pollForEvents(callback);
+                    }, 2000);
```
Kegsay: Given this SDK is pretty fundamental to a lot of clients, we should really be doing exp backoff ASAP. I would propose a high cap of something like 1 *hour* (bear in mind that IM isn't the only use case here) but crucially provide a manual nudge function which will kill the timer and do another poll so humans aren't at the mercy of backoff. This can wait for another PR given it's separate and a bit fiddly.

lib/matrix.js
```diff
@@ -358,8 +704,32 @@ var init = function(exports){
...
+                                return new MatrixEvent(event);
+                            }
+                        ));
+                        for (var i = 0 ; i < data.rooms.length; i++) {
+                            self.store.setStateEvents(data.rooms[i].state.map(
+                                function(event) {
+                                    return new MatrixEvent(event);
+                                }
+                            ));
+                            self.store.setEvents(data.rooms[i].messages.chunk.map(
```
Kegsay: As much as I love `Array.map` it is in ECMAScript 5 meaning this won't work in IE8, which some certain.. corporate environments still use. Given we can get by without it (helper functions yay), we should probably keep compatibility.

lib/matrix.js
```diff
@@ -52,11 +58,351 @@ var init = function(exports){
...
+            }
+        },
+        
+        /*
+         * Helper method for retrieving the name of a user suitable for display in the UI
+         * in the context of a room - i.e. disambiguating from any other users in the room.
+         * XXX: This could perhaps also be generated serverside, perhaps by just passing 
+         * a 'disambiguate' flag down on membership entries which have ambiguous displaynames?
+         */
+        getFriendlyDisplayName: function(userId, roomId) {
```
Kegsay: I feel the `getFriendly` and co. functions are in the right(ish) place but not separated enough from the `MatrixClient`. See main comment.

Kegsay:
Architecture wise, I think this is a good start. However, we're starting to leak IM specific functionality which are not separated with well-defined boundaries e.g. Someone doing IoT doesn't care about the `getFriendly*` functions. I feel this should be split out to another layer on top of `MatrixClient`. 

This new layer can handle other IM specific stuff like formatting up of `m.emotes`. This layer will require an arbitrary data store (to pull out display names from `m.room.member` events) but should also work without one (e.g. formatting up emotes with `* <user_id> does something` instead of `displayname`. The overall architecture would look like:
```
       IM Client
      MatrixClient
API Calls     DataStore(optional)
```
where `IM Client` is the entry point for chat applications and `MatrixClient` the entry point for IoT-style applications (anything not IM specific; but you're not depriving them of a `DataStore`).

lib/matrix.js
ara4n: If you don't specify a state_key it simply ignores the state_key and returns the first event that matches the room/type.  So this is working as intended; am fixing the jsdoc.

ara4n:
I've incorporated all of the above other than splitting getFriendly* and start/stopClient() into their own API layer.  This is because I think there's already a huge range of different stuff exposed in the flat MatrixClient namespace, and adding in a few well documented higher level helper methods doesn't feel like a violation.  Admittedly everything else is "just" wrapping HTTP APIs, but there are a whole bunch of different endpoints there at different levels of granularity, and the fact that start/stopClient provide helpers around some of those methods doesn't seem too bad.  Moreover, I think the faff of documenting and understanding needing to instantiate another client layer (and rename MatrixClient, if it's no longer a Matrix Client, but a wrapper of the client-server API) outweighs the inelegance of throwing in these helper methods "for free".

To put it more clearly, when you already have 44 methods in a flat namespace in the API (including helpers like getHttpUriForMxc()):

        createRoom: function(options, callback) {
        joinRoom: function(roomIdOrAlias, callback) {
        setRoomName: function(roomId, name, callback) {
        setRoomTopic: function(roomId, topic, callback) {
        setPowerLevel: function(roomId, userId, powerLevel, event, callback) {
        sendStateEvent: function(roomId, eventType, content, stateKey, 
        sendEvent: function(roomId, eventType, content, txnId, callback) {
        sendMessage: function(roomId, content, txnId, callback) {
        sendTextMessage: function(roomId, body, txnId, callback) {
        sendEmoteMessage: function(roomId, body, txnId, callback) {
        sendImageMessage: function(roomId, url, info, text, callback) {
        sendHtmlMessage: function(roomId, body, htmlBody, callback) {
        sendTyping: function(roomId, isTyping, timeoutMs, callback) {
        redactEvent: function(roomId, eventId, callback) {
        invite: function(roomId, userId, callback) {
        leave: function(roomId, callback) {
        ban: function(roomId, userId, reason, callback) {
        unban: function(roomId, userId, callback) {
        kick: function(roomId, userId, reason, callback) {
        getProfileInfo: function(userId, info, callback) {
        setProfileInfo: function(info, data, callback) {
        setDisplayName: function(name, callback) {
        setAvatarUrl: function(url, callback) {
        getThreePids: function(creds, bind, callback) {
        addThreePid: function(creds, bind, callback) {
        setPresence: function(presence, callback) {
        publicRooms: function(callback) {
        registerFlows: function(callback) {
        loginFlows: function(callback) {
        resolveRoomAlias: function(roomAlias, callback) {
        initialSync: function(limit, callback) {
        roomInitialSync: function(roomId, limit, callback) {
        roomState: function(roomId, callback) {
        scrollback: function(roomId, from, limit, callback) {
        eventStream: function(from, timeout, callback) {
        login: function(loginType, data, callback) {
        register: function(loginType, data, callback) {
        loginWithPassword: function(user, password, callback) {
        pushRules: function(callback) {
        addPushRule: function(scope, kind, ruleId, body, callback) {
        deletePushRule: function(scope, kind, ruleId, callback) {
        turnServer: function(callback) {
        getHttpUriForMxc: function(mxc, width, height, resizeMethod) {
        getIdenticonUri: function(identiconString, width, height) {
        getContentUri: function() {

...what's another 4 to provide some higher level stuff?

        getFriendlyRoomName: function(roomId) {
        getFriendlyDisplayName: function(userId, roomId) {
        startClient: function(callback, historyLen) {
        stopClient: function() {

lib/matrix.js
Kegsay: This feels like a bad design choice to make. Can you justify why returning the first matching event in hash order (not even chronologically!!)  is desirable? To me, this seems extremely error prone and makes no sense.

Kegsay:
The difference is that those 44 methods are all directly related to http calls (including `getWhateverUri`). You say it yourself that the other 4 provide "some high level stuff". It seems like poor separation of concern to mix them. I don't buy "well we have a load of methods already, why not a few more" as a good

Kegsay:
(fails at mobile typing)
[continued...] reason.

lib/matrix.js
ara4n: Okay, fair enough - you can always use getStateEvents()[0] i guess for the m.room.aliases use case. Changed it.

Kegsay:
LGTM, this is a good baseline to work off. There's extra features that need to be implemented (marked in this PR as ``TODO``) which will build upon this work.

PR #6, opened 2015-06-05
Architecture changes

Kegsay:
This is the first (and most damaging) stage of the proposed architecture changes on https://matrix.org/jira/browse/SYJS-5

This PR:
- splits up ``matrix.js`` into separate modules and begins to define the core object models.
- modifies how callbacks/Promises are handled: Promises are now always returned and it's up to the caller if they want to use them. They can still supply callbacks which will be invoked in addition to Promises.
- changes the concern of ``MatrixHttpApi`` to just be a generic API wrapper, rather than functional wrappers around every API call.
- alters the constructor parameters to both ``MatrixHttpApi`` and ``MatrixClient``; they now accept a generic `opts` Object with documented keys.
- adds copious amounts of JSDoc3 for generating API documentation. `jsdoc -r lib/`.

This PR should be functionally equivalent to `master`, but without integration/unit tests this is hard to guarantee. Both the node and browser examples are confirmed working though.


Kegsay:
Spoke with Matthew, okay to merge.

PR #8, opened 2015-06-11
Event emitter implementation (SYJS-7)

Kegsay:
This PR broadly implements what was discussed at https://matrix.org/jira/browse/SYJS-7

Basic tests for emitted events are included, along with JSDoc on all the events that can be emitted.


Kegsay:
Matthew says "from cursory inspection this looks fine". Been given the OK to merge.

PR #10, opened 2015-06-29
Push

dbkr:
Fetch & evaluate push rules for events via getPushActions on MatrixEvent

### DIFF
--- a/lib/models/event.js
+++ b/lib/models/event.js
@@ -146,7 +146,7 @@ module.exports.MatrixEvent.prototype = {
         return this.event.state_key !== undefined;
     },
 
-    getPushActions(client) {
+    getPushActions: function(client) {
         if (this.pushActions === undefined) {
             var pushProcessor = new PushProcessor(client);
             this.pushActions = pushProcessor.actionsForEvent(this.event);
--- a/lib/models/event.js
+++ b/lib/models/event.js
@@ -1,6 +1,7 @@
+"use strict";
+
 var PushProcessor = require('../pushprocessor');
 
-"use strict";
 /**
  * This is an internal module. See {@link MatrixEvent} and {@link RoomEvent} for
  * the public classes.
--- a/lib/models/event.js
+++ b/lib/models/event.js
@@ -148,10 +148,10 @@ module.exports.MatrixEvent.prototype = {
     },
 
     getPushActions: function(client) {
-        if (this.pushActions === undefined) {
+        if (this._pushActions === undefined) {
             var pushProcessor = new PushProcessor(client);
-            this.pushActions = pushProcessor.actionsForEvent(this.event);
+            this._pushActions = pushProcessor.actionsForEvent(this.event);
         }
-        return this.pushActions;
+        return this._pushActions;
     }
 };
--- a/lib/models/event.js
+++ b/lib/models/event.js
@@ -1,3 +1,5 @@
+var PushProcessor = require('../pushprocessor');
+
 "use strict";
 /**
  * This is an internal module. See {@link MatrixEvent} and {@link RoomEvent} for
@@ -142,5 +144,13 @@ module.exports.MatrixEvent.prototype = {
      */
     isState: function() {
         return this.event.state_key !== undefined;
+    },
+
+    getPushActions(client) {
+        if (this.pushActions === undefined) {
+            var pushProcessor = new PushProcessor(client);
+            this.pushActions = pushProcessor.actionsForEvent(this.event);
+        }
+        return this.pushActions;
     }
 };
--- a/lib/models/event.js
+++ b/lib/models/event.js
@@ -1,7 +1,5 @@
 "use strict";
 
-var PushProcessor = require('../pushprocessor');
-
 /**
  * This is an internal module. See {@link MatrixEvent} and {@link RoomEvent} for
  * the public classes.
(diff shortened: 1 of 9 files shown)


lib/models/event.js
```diff
@@ -142,5 +145,13 @@ module.exports.MatrixEvent.prototype = {
...
     isState: function() {
         return this.event.state_key !== undefined;
+    },
+
+    getPushActions: function(client) {
+        if (this.pushActions === undefined) {
+            var pushProcessor = new PushProcessor(client);
+            this.pushActions = pushProcessor.actionsForEvent(this.event);
+        }
+        return this.pushActions;
```
Kegsay: Nitpick: I assume we do not want to expose `this.pushActions`, so it would be nice to prefix with an `_` e.g. `this._pushActions` to make this clear.

dbkr: ah yes, that would be better.

lib/models/event.js
```diff
@@ -142,5 +145,13 @@ module.exports.MatrixEvent.prototype = {
      */
     isState: function() {
         return this.event.state_key !== undefined;
+    },
+
+    getPushActions: function(client) {
```
Kegsay: I was under the impression that we agreed that we should pass `client.pushRules` as an optional constructor parameter to `MatrixEvent`. On looking at the data you require for the push processor, I can see why you've done this though. This current signature feels wrong (passing it effectively the entire client instance) just to extract `client.pushRules`, `client.credentials.userId` and the `Room` instance further in. We should be explicit with our dependencies.

Given we require more things that just the push rules, this increasingly feels like we shouldn't be constructing and invoking `PushProcessor` in `MatrixEvent` at all. The reason we decided against doing this in `MatrixClient` was because it is relatively computationally expensive to do `PushProcessor.actionsForEvent`, given it will be called for *every* event. Perhaps we can pull out the `getPushActions` function to `MatrixClient`, and just invoke it directly on `MatrixEvent.getPushActions`? E.g.

```javascript
function MatrixEvent(event, pushActionsFn) {
  this.event = event;
  this._pushActionsFn = pushActionsFn;
}

MatrixEvent.prototype.getPushActions = function() {
  if (!this._pushActionsFn) {
    return null; // or whatever is suitable e.g. { }
  }
  return this._pushActionsFn(this.event);
```

And then have in `client.js`:

```javascript
// NB: Passing in 'room' because there isn't any guarantee that the store knows this room yet
//     (so client.getRoom(roomId) won't work)
function _PojoToMatrixEventMapper(client, room, plainOldJsObject) {
  var pushFn = function(event) {
    var pushProcessor = new PushProcessor(room, client.pushRules, client.credentials.userId);
    return pushProcessor.actionsForEvent(event);
  };
  return new MatrixEvent(plainOldJsObject, pushFn);
}
```

We would almost certainly want to optimise this (not creating a new `pushFn` function for every event ever; not creating a `new PushProcessor` every time if a static function would work just as well), but this is the general idea.

The benefits to this are:
 - Delegates push logic entirely out of `MatrixEvent` (similar to currently but currently you have to still supply the deps at the point of invocation rather than at construction time)
 - Makes `PushProcessor` dependencies explicit (much easier to test and get a high level overview of what is required in order to "process" correctly without having to comb through the code looking for method calls on `client`)

The drawback here is that it isn't entirely clear that `MatrixEvent.getPushActions` is just proxying through to a function `pushActionsFn` supplied in the constructor, but hey this is what JSDoc is for (and it also allows other people to inject in different ways of handling push rules if they're crazy enough).

lib/pushprocessor.js
```diff
@@ -0,0 +1,226 @@
+var micromatch = require("micromatch");
```
Kegsay: How big is the footprint on this (other dependencies that `micromatch` needs?) - `npm list`

lib/pushprocessor.js
```diff
@@ -0,0 +1,226 @@
...
+
+/**
+ * Construct a Push Processor.
+ * @constructor
+ * @param {Object} client The Matrix client object to use
+ */
+module.exports = function(client) {
```
Kegsay: This all looks deliciously functional; is there any reason why we need to construct instances of `PushProcessor`?

Kegsay:
LGTM other than inline comments and lack of tests for `PushProcessor` (should be able to cargo-cult the ones you did for Angular I think?)

lib/models/event.js
dbkr: Hmm - adding a constructor arg to MatrixEvent for something that specific doesn't seem ideal, and I'm not too keen on complicating it with closures. I'd be in favour of either just adding the creating client as an ctor arg to MatrixEvent so it can defer back to the client for doing this or other similar things, or leave the MatrixEvent model class as pure data and shift any more complex stuff like this out to MatrixClient.getPushActionsForEvent(event) - what do you think?

Kegsay: > shift any more complex stuff like this out to MatrixClient.getPushActionsForEvent(event) - what do you think?

I really like this. I am currently doing persistent storage and am having to construct `MatrixEvents` (the storage has no concept of the `MatrixClient`), so I prefer to keep that class as pure data.

dbkr:
I think this should be all your concerns addressed now.

Kegsay:
LGTM

PR #11, opened 2015-07-02
Move getAvatarUrl method to client so it can do mxc -> http uri conversion

dbkr:
Also fix some c+p bugs in client


Kegsay:
LGTM

PR #12, opened 2015-07-15
WebRTC support

Kegsay:
This PR adds in support for WebRTC in the JS SDK. This has been manually tested to work in Chrome using the example app also in this PR. There are currently no unit or integration tests for this component. This PR is based heavily on the angular SDK.

Overview:
 * A new class `MatrixCall` has been added which is the main component of this PR. This class is not directly exposed to the end user. Instances of `MatrixCall` can emit `hangup`, `error` and `replaced` events. There are also public methods to `hangup()`, `answer()`, `placeVideoCall(remoteVideoElement, localVideoElement)` and `placeVoiceCall()`.
 * A new global method `createNewMatrixCall` has been added. This is how end-users create outgoing calls. They specify the `roomId` they want to place the call in and a `MatrixClient` instance which is used to send `m.call.*` events. This method checks for WebRTC compatibility by inspecting the global `window`. This is the main way to create a `new MatrixCall()`.
 * `MatrixClient` emits a new event: `Call.incoming`. This is emitted when there is an incoming call, and it supplies a `MatrixCall` instance as an argument.

Placing a call:
```javascript
// matrixcs is the global attached to window.
var call = matrixcs.createNewMatrixCall(matrixClient, "!foo:bar");
call.on("error", function(err) {
  console.log("Error: %s", err);
});
call.placeVoiceCall();
```

Answering a call:
```javascript
matrixClient.on("Call.incoming", function(call) {
  call.on("error", function(err) {
    console.log("Error: %s", err);
  }
  call.answer();
});
```

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -15,6 +15,7 @@ var EventStatus = require("./models/event").EventStatus;
 var StubStore = require("./store/stub");
 var Room = require("./models/room");
 var User = require("./models/user");
+var webRtcCall = require("./webrtc/call");
 var utils = require("./utils");
 
 // TODO:
@@ -73,6 +74,17 @@ function MatrixClient(opts) {
     this._syncingRooms = {
         // room_id: Promise
     };
+    this.callList = {
+        // callId: MatrixCall
+    };
+
+    // try constructing a MatrixCall to see if we are running in an environment
+    // which has WebRTC. If we are, listen for and handle m.call.* events.
+    var call = webRtcCall.createNewMatrixCall(this);
+    if (call) {
+        setupCallEventHandler(this);
+    }
+
 }
 utils.inherits(MatrixClient, EventEmitter);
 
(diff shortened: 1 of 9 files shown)


lib/client.js
```diff
@@ -73,6 +74,17 @@ function MatrixClient(opts) {
     this._syncingRooms = {
         // room_id: Promise
     };
+    this.callList = {
+        // callId: MatrixCall
+    };
+
+    // try constructing a MatrixCall to see if we are running in an environment
+    // which has WebRTC. If we are, listen for and handle m.call.* events.
+    var call = webRtcCall.createNewMatrixCall(this);
```
dbkr: In angular, I ended up making this return a dummy object if the browser had no WebRTC. This makes it easier for the client to still manage some about of call flow (ie. ringing, stop ringing when other party hangs up or call is answered elsewhere).

lib/webrtc/call.js
```diff
@@ -0,0 +1,891 @@
...
+        this.peerConn.setRemoteDescription(
+            new this.webRtc.RtcSessionDescription(this.msg.offer),
+            hookCallback(self, self._onSetRemoteDescriptionSuccess),
+            hookCallback(self, self._onSetRemoteDescriptionError)
+        );
+    }
+    this.state = 'ringing';
+    this.direction = 'inbound';
+
+    // firefox and Safari's RTCPeerConnection doesn't add streams until it
```
dbkr: Not sure if this is my comment or yours but this is technically OpenWebRTC here rather than Safari: Safari has no WebRTC.

dbkr:
lgtm!

PR #13, opened 2015-07-16
Implement setPassword API

erikjohnston:


lib/client.js
```diff
@@ -807,6 +807,25 @@ MatrixClient.prototype.addThreePid = function(creds, bind, callback) {
 };
 
 /**
+ * @param {Object} auth_dict
+ * @param {string} new_password
+ * @param {module:client.callback} callback Optional.
+ * @return {module:client.Promise} Resolves: TODO
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixClient.prototype.setPassword = function(auth_dict, new_password, callback) {
```
Kegsay: Please use `camelCase` instead of `under_scores` here.

dbkr:
Looks fine. It would probably be nice, come to think of it, if we commented the API that used user-interactive auth (which this one optionally does).

Maybe fix whatever is making Jenkins angry.

Kegsay:
LGTM other than comments - `develop` was broken earlier so the failing tests are not relevant for this.

Kegsay:
LGTM

PR #14, opened 2015-07-16
Registration

dbkr:


lib/client.js
```diff
@@ -1381,6 +1385,47 @@ function _PojoToMatrixEventMapper(plainOldJsObject) {
...
+ * @param {string} email
+ * @param {string} client_secret
+ * @param {string} send_attempt
+ * @param {string} next_link Optional
+ * @param {module:client.callback} callback Optional.
+ * @return {module:client.Promise} Resolves: TODO
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixClient.prototype.requestEmailToken = function(email, client_secret,
+                                                    send_attempt, next_link, callback) {
```
Kegsay: `camelCase` please.

lib/http-api.js
```diff
@@ -352,6 +387,7 @@ var requestCallback = function(defer, userDefinedCallback, onlyData) {
     return function(err, response, body) {
         if (!err && response.statusCode >= 400) {
             err = new module.exports.MatrixError(body);
+            err.httpStatus = response.statusCode;
```
Kegsay: Good idea. Please add this property to the JSDoc at the bottom of this file.

lib/client.js
```diff
@@ -807,15 +808,6 @@ MatrixClient.prototype.publicRooms = function(callback) {
  * @return {module:client.Promise} Resolves: TODO
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
-MatrixClient.prototype.registerFlows = function(callback) {
```
Kegsay: Is there no alternative for this in v2?

Kegsay:
LGTM other than comments.

lib/client.js
dbkr: Not really - you just call register and if it fails, it's in the response, so it's just part of the register call.

Kegsay: Righto.

Kegsay:
LGTM

PR #17, opened 2015-07-21
Add support for end-to-end using olm

NegativeMjark:


lib/client.js
```diff
@@ -92,6 +141,243 @@ function MatrixClient(opts) {
...
+        } else {
+            deferred.resolve({
+                missingUsers: userWithoutDevices,
+                missingDevices: []
+            });
+        }
+        this.sessionStore.storeEndToEndRoom(roomId, config);
+        return deferred.promise;
+    } else {
+        throw new Error("Unknown algorithm: " + OLM_ALGORITHM);
```
dbkr: Presumably you mean, "Unknown algorithm: "+config.algorithm. OLM_ALGORITHM being the one we do know about...

lib/client.js
```diff
@@ -92,6 +141,243 @@ function MatrixClient(opts) {
...
+                queries[device[0]] = query;
+                query[device[1]] = "curve25519";
+            }
+            var path = "/keys/claim";
+            var content = {one_time_keys: queries};
+            var self = this;
+            this._http.authedRequestWithPrefix(
+                undefined, "POST", path, undefined, content,
+                httpApi.PREFIX_V2_ALPHA
+            ).then(function(res) {
```
dbkr: think this wants to be .done to avoid swallowing exceptions

lib/client.js
```diff
@@ -319,9 +606,205 @@ MatrixClient.prototype.sendEvent = function(roomId, eventType, content, txnId,
...
+    }
+
+    if (e2eRoomInfo.algorithm === OLM_ALGORITHM) {
+        var participantKeys = [];
+        for (var i = 0; i < e2eRoomInfo.members.length; ++i) {
+            var userId = e2eRoomInfo.members[i];
+            var devices = client.sessionStore.getEndToEndDevicesForUser(userId);
+            for (var deviceId in devices) {
+                if (devices.hasOwnProperty(deviceId)) {
+                var keys = devices[deviceId];
```
dbkr: Indenting?

lib/client.js
```diff
@@ -319,9 +606,205 @@ MatrixClient.prototype.sendEvent = function(roomId, eventType, content, txnId,
...
+            }
+        }
+        var encryptedContent = {
+            algorithm: e2eRoomInfo.algorithm,
+            sender_key: client.deviceCurve25519Key,
+            ciphertext: ciphertext
+        };
+        return encryptedContent;
+    } else {
+        throw new Error("Unknown end-to-end algorithm");
```
dbkr: Maybe add + e2eRoomInfo.algorithm for consistency

lib/client.js
```diff
@@ -319,9 +606,205 @@ MatrixClient.prototype.sendEvent = function(roomId, eventType, content, txnId,
...
+                        foundSession = true;
+                    }
+                    payloadString = session.decrypt(message.type, message.body);
+                    var pickled = session.pickle(client.accountKey);
+                    client.sessionStore.storeEndToEndSession(
+                        deviceKey, sessionId, pickled
+                    );
+                } catch (e) {
+                    // Failed to decrypt with an existing session.
+                    console.log(
```
dbkr: I would guess this should probably rethrow or possibly emit a 'bad encrypted message' event rather than logging a console message and just not doing anything.

lib/store/session/webstorage.js
```diff
@@ -0,0 +1,143 @@
...
+    }
+    catch (e) {
+        debuglog("Failed to get key %s: %s", key, e);
+        debuglog(e.stack);
+    }
+    return null;
+}
+
+function setJsonItem(store, key, val) {
+    store.setItem(key, JSON.stringify(val));
```
dbkr: Probably need to handle running out of space here (nb. Safari in privte browsing mode will always throw an exception when asked to save anything to localstorage, so I expect this will just break).

dbkr:
Couple of things outstanding, but looks generally good: merging.

PR #19, opened 2015-09-01
turn HTTP exceptions into errbacks or rejected deferreds rather than …

ara4n:
…bubbling them up and expecting the app to have try blocks eeeeeeeeeverywhere


Kegsay:
Given users can clobber and set their own `request` module, it's probably worth playing it safe here and catching sync errors which occur when their custom request function is called.

PR #23, opened 2015-10-16
Receipts

Kegsay:
This adds in support for receipts:
 - HTTP API exposed on `MatrixClient`
 - Accessor methods via `Room`

Has unit and integration tests.


lib/utils.js
```diff
@@ -228,6 +228,142 @@ module.exports.deepCopy = function(obj) {
...
+          // d. Increase k by 1.
+          k++;
+        }
+
+        // 9. return A
+        return A;
+      };
+    }
+};
+
```
dbkr: Where is this code from? Do we need to attribute it or anything?

dbkr:
LGTM assuming the polyfill stuff is okay (we should probably comment where it's from even if we aren't strictly required to).

Kegsay:
Both snippets are directly from MDN and linked at the top of each segment. I've added clarification that the link is where the origin of the code is from.

PR #24, opened 2015-10-16
Look for a canonical alias when determining the room name

Kegsay:


dbkr:
LGTM. I guess there was no convenience method for getting the aliases before? Perhaps we could add one but that needn't be part of this PR.

Kegsay:
Indeed: as the UI muxes the concepts of "room alias" with "room name", there hasn't been a need to create an additional public-facing way to just get the room alias.

PR #25, opened 2015-10-16
Add support for archived=true in initial sync

Kegsay:
Make `MatrixClient.startClient` take `opts` instead of `historyLen` in a backwards compatible way. Add `opts.includeArchivedRooms` as an option.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1857,10 +1857,15 @@ MatrixClient.prototype.isLoggedIn = function() {
  * This is an internal method.
  * @param {MatrixClient} client
  * @param {integer} historyLen
+ * @param {integer} includeArchived
  */
-function doInitialSync(client, historyLen) {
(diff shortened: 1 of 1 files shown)


lib/client.js
```diff
@@ -1857,10 +1857,15 @@ MatrixClient.prototype.isLoggedIn = function() {
  * This is an internal method.
  * @param {MatrixClient} client
  * @param {integer} historyLen
+ * @param {integer} includeArchived
  */
-function doInitialSync(client, historyLen) {
+function doInitialSync(client, historyLen, includeArchived) {
+    var qps = { limit: historyLen };
```
dbkr: historylen || 12 to retain the old default behaviour?

dbkr: Oh never mind, you moved it below.

dbkr:
LGTM

PR #26, opened 2015-10-16
Invite room state

Kegsay:
Read `invite_room_state` key from the `m.room.member` event if it exists and create fake state events to make invited rooms display the room name.

Has unit tests.


dbkr:
Less sure what I'm looking at here, but look plausible. :)

PR #27, opened 2015-10-19
Room avatars

Kegsay:
This PR introduces handling for `m.room.avatar` events.

BREAKING CHANGES:
 - `MatrixClient.getAvatarUrlForMember` has been removed and replaced with `RoomMember.getAvatarUrl`.
 - `MatrixClient.getAvatarUrlForRoom` has been removed and replaced with `Room.getAvatarUrl`.

*(Both of these changes require virtually zero work to map between the old and new style; see CHANGELOG.md)*

Scope each `getAvatarUrl` to be instance methods on the entity it
relates to (Room and RoomMember respectively). By doing this, we
can actually pull out specific state such as the `m.room.avatar`
event more easily rather than keeping it in the global cesspit
of `MatrixClient`.

This was complicated by `getHttpUriForMxc` and `getIdenticonUri`
which were attached to the HTTP API to pull out the `baseUrl` when
crafting the URL. Pull out this dependency out and explicitly pass
it in when crafting the URL. This is trivial to get from
`MatrixClient.getHomeserverUrl()`.


dbkr:
Yay, this looks much more sensible. LGTM.

PR #30, opened 2015-10-21
Change calculating display names from O(n^2) to O(n)

Kegsay:
Reduces initial sync times from ~30s to ~1s on accounts with heavily
populated rooms.

The problem was that f.e. RoomMember it would try to calculate the
display name, which involved looping each RoomMember to get their
display name to check for disambiguation. We now cache display names
to user IDs so we don't need to loop every member when disambiguating.

Updated tests.

### DIFF
--- a/lib/models/room-state.js
+++ b/lib/models/room-state.js
@@ -31,6 +31,8 @@ function RoomState(roomId) {
         // userId: RoomMember
     };
     this._updateModifiedTime();
+    this._displayNameToUserIds = {};
+    this._userIdsToDisplayNames = {};
 }
 utils.inherits(RoomState, EventEmitter);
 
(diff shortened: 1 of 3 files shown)


lib/models/room-state.js
```diff
@@ -181,10 +188,41 @@ RoomState.prototype.getLastModifiedTime = function() {
...
 
+
+function _updateDisplayNameCache(roomState, userId, displayName) {
+    var oldName = roomState._userIdsToDisplayNames[userId];
+    delete roomState._userIdsToDisplayNames[userId];
+    if (oldName) {
+        var existing = roomState._displayNameToUserIds[oldName] || [];
+        for (var i = 0; i < existing.length; i++) {
+            if (existing[i] === userId) {
+                existing.splice(i, 1);
```
ara4n: might be worth a comment spelling out the .splice() idiom for deleting the matching old entry from the array - i had to think about this a bit to see what was going on.

ara4n:
lgtm otherwise

lib/models/room-state.js
Kegsay: Done: It's a pretty common idiom so I didn't think to comment it, sorry!

PR #32, opened 2015-10-26
Retrieving profile info for invites

Kegsay:
This PR adds the ability for the SDK to fetch profile information for invitees. This is gated behind the flag `resolveInvitesToProfiles` which is set on `MatrixClient.startClient(opts)`.

If this flag is set to `true`, every incoming `m.room.member` **invite** event will be resolved. This resolution has the following steps:
 - Check for existing profile info in the store via `getUser(userId)`. If exists, use that. If not:
 - Hit `/profile/$user_id` to get their profile information. Set profile info from that.
 - Setting profile info involves munging the `m.room.member` `content` to include the `avatar_url` and `displayname`. This is preferable as it reduces the number of extraneous code paths this PR adds.

This has integration tests.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -141,6 +141,7 @@ function MatrixClient(opts) {
     this.callList = {
         // callId: MatrixCall
     };
+    this._config = {}; // see startClient()
 
     // try constructing a MatrixCall to see if we are running in an environment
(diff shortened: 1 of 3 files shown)


dbkr:
LGTM

PR #34, opened 2015-10-30
Simplify logic and layout

illicitonion:


lib/models/room.js
```diff
@@ -415,6 +415,11 @@ function setEventMetadata(event, stateContext, toStartOfTimeline) {
 function calculateRoomName(room, userId) {
     // check for an alias, if any. for now, assume first alias is the
     // official one.
+    var mRoomName = room.currentState.getStateEvents('m.room.name', '');
```
Kegsay: Nit: Please use `"` in this project. This is even done nearly everywhere in the SDK :exclamation:

Kegsay:
LGTM other than nit.

lib/models/room.js
illicitonion: I love your implication that I wrote any of this code.

Nonetheless, done.

PR #35, opened 2015-11-02
Add event to get last read receipt for a user.

dbkr:


lib/models/room.js
```diff
@@ -322,6 +322,21 @@ Room.prototype.getUsersReadUpTo = function(event) {
...
 
 /**
+ * Get the ID of the event that a given user has read up to, or null if we
+ * have received no read receipts from them.
+ * @param {String} user ID to get read receipt event ID for
+ * @return {String} ID of the latest event that the given user has read, or null.
+ * an empty list.
+ */
+Room.prototype.getEventReadUpTo = function(userId) {
+    if (this._receipts === undefined || this._receipts['m.read'] === undefined) {
```
Kegsay: `this._receipts` is **guaranteed** to be instantiated, so for clarity I would prefer if we didn't have the `undefined` check as it obfuscates the code. Keep the check for `m.read` though because that may not exist (though this can be written as `if (!this._receipts["m.read"]) ` which would also catch any `nulls`. Also, minor nit: please use `"` over `'` (this is done in about ~90% of cases in the JS SDK and will eventually be made into part of the linting checks if I can be bothered).

lib/models/room.js
```diff
@@ -322,6 +322,21 @@ Room.prototype.getUsersReadUpTo = function(event) {
 };
 
 /**
+ * Get the ID of the event that a given user has read up to, or null if we
+ * have received no read receipts from them.
+ * @param {String} user ID to get read receipt event ID for
```
Kegsay: Typo, needs to be `userId` to match the arg on `:331`.

lib/models/room.js
```diff
@@ -322,6 +322,21 @@ Room.prototype.getUsersReadUpTo = function(event) {
...
+ * @return {String} ID of the latest event that the given user has read, or null.
+ * an empty list.
+ */
+Room.prototype.getEventReadUpTo = function(userId) {
+    if (this._receipts === undefined || this._receipts['m.read'] === undefined) {
+        return null;
+    }
+
+    return this._receipts['m.read'][userId].eventId;
```
Kegsay: `this._receipts['m.read'][userId]` may be `undefined` if they have never acknowledged a receipt.

Kegsay:
LGTM other than comments.

lib/models/room.js
```diff
@@ -322,6 +322,24 @@ Room.prototype.getUsersReadUpTo = function(event) {
 };
 
 /**
+ * Get the ID of the event that a given user has read up to, or null if we
+ * have received no read receipts from them.
+ * @param {String} userId The user ID to get read receipt event ID for
+ * @return {String} ID of the latest event that the given user has read, or null.
+ * an empty list.
```
Kegsay: Stray line? Presumably this doesn't relate to the return arg.

PR #36, opened 2015-11-02
Wrap /events response processing in a try/catch

Kegsay:
Fixes https://github.com/vector-im/vector-web/issues/276

The diff looks 'orrid, it's literally just an indent block :(


dbkr:
I see what you mean about the diff. LGTM.

PR #39, opened 2015-11-03
Room tag support

ara4n:
WIP of room tag support (needs tests)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -637,7 +637,7 @@ MatrixClient.prototype.setRoomTopic = function(roomId, topic, callback) {
 
 /**
  * @param {string} roomId
- * @param {string} name of room tag to be set
+ * @param {string} tagName name of room tag to be set
  * @param {object} metadata associated with that tag to be stored
  * @param {module:client.callback} callback Optional.
  * @return {module:client.Promise} Resolves: TODO
@@ -656,7 +656,7 @@ MatrixClient.prototype.setRoomTag = function(roomId, tagName, metadata, callback
 
 /**
  * @param {string} roomId
- * @param {string} name of room tag to be removed
+ * @param {string} tagName name of room tag to be removed
  * @param {module:client.callback} callback Optional.
  * @return {module:client.Promise} Resolves: TODO
  * @return {module:http-api.MatrixError} Rejects: with an error response.
--- a/lib/client.js
+++ b/lib/client.js
@@ -2480,7 +2480,7 @@ function createNewUser(client, userId) {
 
 function createNewRoom(client, roomId) {
     var room = new Room(roomId);
-    reEmit(client, room, ["Room.name", "Room.timeline"]);
+    reEmit(client, room, ["Room.name", "Room.tags", "Room.timeline"]);
 
     // we need to also re-emit room state and room member events, so hook it up
     // to the client now. We need to add a listener for RoomState.members in
--- a/lib/client.js
+++ b/lib/client.js
@@ -609,6 +609,43 @@ MatrixClient.prototype.setRoomTopic = function(roomId, topic, callback) {
 };
 
 /**
+ * @param {string} roomId
+ * @param {string} name of room tag to be set
+ * @param {object} metadata associated with that tag to be stored
+ * @param {module:client.callback} callback Optional.
+ * @return {module:client.Promise} Resolves: TODO
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixClient.prototype.setRoomTag = function(roomId, tagName, metadata, callback) {
+    var path = utils.encodeUri("/user/$userId/rooms/$roomId/tags/$tag", {
+        $userId: this.credentials.userId,
+        $roomId: roomId,
+        $tag: tagName,
+    });
+    return this._http.authedRequestWithPrefix(
+        callback, "PUT", path, undefined, metadata, httpApi.PREFIX_V2_ALPHA
+    );
+};
+
+/**
+ * @param {string} roomId
+ * @param {string} name of room tag to be removed
+ * @param {module:client.callback} callback Optional.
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -1942,6 +1942,16 @@ function doInitialSync(client, historyLen, includeArchived) {
                     room.addReceipt(receipts[j]);
                 }
 
+                var privateUserData = data.rooms[i].private_user_data || [];
+                for (j = 0; j < privateUserData.length; j++) {
+                    var event = _PojoToMatrixEventMapper(client)(privateUserData[j]);
```
Kegsay: It took me longer than I'd like to grok this. We generally tend to use the idiom:
```javascript
utils.map(eventsWhichAreObjects, _PojoToMatrixEventMapper(client));
```

This does incur an extra loop to then do stuff with the events, but profiling hasn't shown this to be the cause of any slowness in the JS SDK.

lib/client.js
```diff
@@ -1942,6 +1942,16 @@ function doInitialSync(client, historyLen, includeArchived) {
...
+                    if (event.type === "m.tag") {
```
Kegsay: `event` at this point should be a `MatrixEvent` in which case we should be using `getType()` here.

lib/models/room.js
```diff
@@ -527,3 +557,14 @@ module.exports = Room;
  *   var newName = room.name;
  * });
  */
+
+/**
+ * Fires whenever a room's tags are updated.
+ * @event module:client~MatrixClient#"Room.tags"
+ * @param {Room} room The room whose Room.tags was updated.
+ * @example
+ * matrixClient.on("Room.tags", function(room){
```
Kegsay: Given this does have a 1:1 mapping of "we got some event and then fired this thing", it would be polite to include the `MatrixEvent` which made this event fire as the first arg for consistency with other events which can fire (mostly `RoomState` events).

lib/models/room.js
```diff
@@ -401,6 +410,27 @@ Room.prototype.addReceipt = function(event) {
...
+    // content: {
+    //    tags: {
+    //       $tagname: { $metadata: $value },
+    //       $tagname: { $metadata: $value },
+    //    }
+    // }
+
+    // XXX: we could do a deep-comparison to see if the tags have really
+    // changed - but do we want to bother?
+    this.emit("Room.tags", this);
```
Kegsay: I agree it sounds like a lot of work without any concrete gains. Devs can always do this manually if they care (though the reverse (only emit if deeply changed) would be impossible to detect this case since they need to know that a room has got tags in order to update!)

lib/models/room.js
```diff
@@ -401,6 +410,27 @@ Room.prototype.addReceipt = function(event) {
...
+
+    // XXX: do we need to deep copy here?
+    this._tags = event.content.tags;
```
Kegsay: The rest of the SDK tends to persist the *event* as a private variable which is then yanked as it needs to be accessed. This makes it clearer that the tags are being drawn from the event and not copied.

As an aside, `event` in this case is a `MatrixEvent` which means `event.content === undefined`. Use `event.getContent()` instead.

lib/models/room.js
```diff
@@ -401,6 +410,27 @@ Room.prototype.addReceipt = function(event) {
...
+    //       $tagname: { $metadata: $value },
+    //    }
+    // }
+
+    // XXX: we could do a deep-comparison to see if the tags have really
+    // changed - but do we want to bother?
+    this.emit("Room.tags", this);
+
+    // XXX: do we need to deep copy here?
+    this.tags = event.content.tags;
```
Kegsay: Node's `EventEmitter` is entirely *synchronous* so we should probably set these tags prior to calling `.emit` given we aren't passing them as an arg on the emitted event.

Also, Github has decided that my previous comment on this line was outdated for some reason unknown.

> The rest of the SDK tends to persist the *event* as a private variable which is then yanked as it needs to be accessed. This makes it clearer that the tags are being drawn from the event and not copied.
>
> As an aside, `event` in this case is a `MatrixEvent` which means `event.content === undefined`. Use `event.getContent()` instead.

Kegsay:
LGTM module comments and unit tests.

lib/client.js
ara4n: I think this is very subjective personally, and I'm not sure I agree - the added boilerplate and two loops and mix of map & iteration feels minging to me.  But i've done it anyway.

lib/client.js
ara4n: done

lib/models/room.js
ara4n: done

lib/models/room.js
ara4n: This isn't true of anything else in the Room class, as far as I can see - I implemented it like this originally and then changed it to match the rest of the class.  Unsure if I want to break symmetry for this.

the getContent() thing is fixed.

ara4n:
Everything should be done on this (modulo pushing back on storing tag data as events on the Room class), including basic UTs - if I can get a final review & merge would be appreciated.

spec/unit/room.spec.js
```diff
@@ -801,4 +801,41 @@ describe("Room", function() {
...
+                room.on("Room.tags", listener);
+
+                var tags = { "m.foo": { "order": 0.5 } };
+                var event = mkTags(roomId, tags);
+                room.addTags(event);
+                expect(listener).toHaveBeenCalledWith(event, room);
+            });
+
+            // XXX: shouldn't we try injecting actual m.tag events onto the eventstream
+            // rather than injecting via room.addTags()?
```
Kegsay: This is what integration tests are for in the `integ` directory.

Kegsay:
LGTM

ara4n:
woo, thanks :)

PR #40, opened 2015-11-05
Syncing bugs/fixes

Kegsay:
This PR adds the necessary structures for https://github.com/vector-im/vector-web/issues/197 to be implemented.

This PR fixes all the issues in SYJS-11:
 - We need to retry `/initialSync` if it fails. Currently we get wedged.
 - We need to emit an event when we recover from a `syncError` to let clients remove banners/warnings saying 'bad connection'.
 - We need to exp backoff on `/events` in the event of a failure.
 - We need to have a manual prod function to kick either `/initialSync` and/or `/events` if they are backing off.
 - We need to expose `/events?timeout=` to clients.
 - Room initial sync should be using `historyLen`.

This PR makes a breaking change:
 - `syncError` and `syncComplete` are no longer emitted, and are combined into a `sync` event with `state` arguments.

This PR has unit tests for the sync state transitions and forced retries.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -151,7 +151,7 @@ function MatrixClient(opts) {
         setupCallEventHandler(this);
         this._supportsVoip = true;
     }
-
+    this._syncState = null;
 }
 utils.inherits(MatrixClient, EventEmitter);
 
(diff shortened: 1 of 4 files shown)


dbkr:
LGTM

PR #41, opened 2015-11-05
Implicit read receipts

dbkr:
* Inject implicit read receipts into the timeline
 * Twiddle local echo a bit to make the implicit receipts match the various different stages of local echo.

### DIFF
--- a/lib/models/room.js
+++ b/lib/models/room.js
@@ -199,8 +199,8 @@ Room.prototype.addEventsToTimeline = function(events, toStartOfTimeline) {
         // Done after adding the event because otherwise the app would get a read receipt
         // pointing to an event that wasn't yet in the timeline
 
-        // This is really ugly because JS has no way to express an object literal where the
-        // name of a key comes from an expression
+        // This is really ugly because JS has no way to express an object literal
+        // where the name of a key comes from an expression
         if (events[i].sender) {
             var fakeReceipt = {content: {}};
             fakeReceipt.content[events[i].getId()] = {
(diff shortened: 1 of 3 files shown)


lib/models/room.js
```diff
@@ -194,7 +194,26 @@ Room.prototype.addEventsToTimeline = function(events, toStartOfTimeline) {
...
         }
-        this.emit("Room.timeline", events[i], this, Boolean(toStartOfTimeline));
+
+        // synthesize and inject implicit read receipts
+        // Done after adding the event because otherwise the app would get a read receipt
+        // pointing to an event that wasn't yet in the timeline
+
+        // This is really ugly because JS has no way to express an object literal
+        // where the name of a key comes from an expression
+        if (events[i].sender) {
```
Kegsay: `events[i]` is a `MatrixEvent` so you'll need to use `events[i].getSender()`.

lib/models/room.js
```diff
@@ -261,6 +280,29 @@ Room.prototype.addEvents = function(events, duplicateStrategy) {
 };
 
 /**
+ * Removes events from this room.
+ * @param {String} event_ids A list of event_ids to remove.
+ */
+Room.prototype.removeEvents = function(event_ids) {
+    // avoids defining a function in the loop, which is a lint error
+    function remove_event_with_id(timeline, id) {
```
Kegsay: Nit: `camelCase` please

lib/models/room.js
```diff
@@ -261,6 +280,29 @@ Room.prototype.addEvents = function(events, duplicateStrategy) {
...
+        // on the assumption that recents events are much
+        // more likley to be removed than older ones.
+        return utils.removeElement(timeline, function(e) {
+            return e.getId() == id;
+        }, true);
+    }
+
+    for (var i = 0; i < event_ids.length; ++i) {
+        var removed = remove_event_with_id(this.timeline, event_ids[i]);
+        if (removed !== false) {
```
Kegsay: Is this right? AFAIK `removeElement` now returns the element that was removed, which will be a `MatrixEvent`. This will always return `false`. Did you intend to `Boolean()` before returning in `remove_event_with_id`?

Kegsay:
LGTM other than this.

lib/models/room.js
dbkr: Haha, good catch, thank you.

PR #42, opened 2015-11-06
SH - CAS / Login token login

stevenhammerton:
Update the CAS login flow to allow the HS to do all the interaction with CAS and use the HS token login

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1722,9 +1722,9 @@ MatrixClient.prototype.loginWithSAML2 = function(relayState, callback) {
 };
 
 /**
- * @param {string} redirectUrl URL to which to be redirected to after
- * HS authenticates with CAS and issues login token
- * @return {string} CAS login URL
+ * @param {string} redirectUrl The URL to redirect to after the HS 
+ * authenticates with CAS.
+ * @return {string} The HS URL to hit to begin the CAS login process.
  */
 MatrixClient.prototype.getCasLoginUrl = function(redirectUrl) {
     return this._http.getUrl("/login/cas/redirect", {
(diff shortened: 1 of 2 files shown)


stevenhammerton:
Will need to go out at same time as https://github.com/matrix-org/matrix-react-sdk/pull/28 and https://github.com/vector-im/vector-web/pull/335

lib/http-api.js
```diff
@@ -296,6 +296,15 @@ module.exports.MatrixHttpApi.prototype = {
         return this._request(callback, method, fullUri, queryParams, data);
     },
 
+    redirect: function(path, queryParams, prefix) {
+        var queryString = "";
+        if (queryParams) {
+            queryString = "?" + utils.encodeParams(queryParams);
+        }
+        var fullUri = this.opts.baseUrl + prefix + path + queryString;
+        window.location.href = fullUri;
```
Kegsay: The JS-SDK can run in any JS environment, including those without a browser. Please delegate responsibility for doing the redirect to React SDK.

lib/client.js
```diff
@@ -1722,27 +1722,23 @@ MatrixClient.prototype.loginWithSAML2 = function(relayState, callback) {
 };
 
 /**
- * @param {module:client.callback} callback Optional.
- * @return {module:client.Promise} Resolves: TODO
- * @return {module:http-api.MatrixError} Rejects: with an error response.
+ * @param {string} redirectUrl URL to which to be redirect to after
+ * HS authenticates with CAS and issues login token
+ * Will redirect to homeserver to complete CAS login
```
Kegsay: Wha-? I sort of understand but the grammar is weird. Perhaps:

```
@param {string} redirectUrl The URL to redirect to after the HS authenticates with CAS.
```

Taking into account my other comment, you may want to also do:

```
@return {string} The HS URL to hit to begin the CAS login process.
```

Kegsay:
LGTM other than this.

Kegsay:
LGTM

PR #43, opened 2015-11-06
Implicit read receipts 2

dbkr:
Generate implicit read receipts in recalculateRoom so they're correct when the room is loaded and addReceipt only has to worry about adding them for events appended to the timeline which is fairly trivial logic.


dbkr:
D'oh, wrong branch.

PR #44, opened 2015-11-06
Synthesize implicit read receipts in recalculateRoom

dbkr:
Generate implicit read receipts in recalculateRoom so they're correct when the room is loaded and addReceipt only has to worry about adding them for events appended to the timeline which is fairly trivial logic.

### DIFF
--- a/lib/models/room.js
+++ b/lib/models/room.js
@@ -375,15 +375,13 @@ Room.prototype.recalculate = function(userId) {
 
         var readReceiptsForEvent = this.getReceiptsForEvent(e);
 
-        for (var receiptIt = 0; receiptIt < readReceiptsForEvent.length; ++receiptIt) {
-            var receipt = readReceiptsForEvent[receiptIt];
-            if (receipt.type !== "m.read") { continue; }
-
-            if (usersFound[receipt.userId]) { continue; }
+        readReceiptsForEvent.forEach(function(receipt) {
+            if (receipt.type !== "m.read") { return; }
+            if (usersFound[receipt.userId]) { return; }
 
             // Then this is the receipt we keep for this user
             usersFound[receipt.userId] = 1;
-        }
+        });
 
         if (e.sender && usersFound[e.sender.userId] === undefined) {
             // no receipt yet for this sender, so we synthesize one.
--- a/lib/models/room.js
+++ b/lib/models/room.js
@@ -13,7 +13,11 @@ var ContentRepo = require("../content-repo");
 function synthesizeReceipt(userId, event, receiptType) {
     // This is really ugly because JS has no way to express an object literal
     // where the name of a key comes from an expression
-    var fakeReceipt = {content: {}};
+    var fakeReceipt = {
+        content: {},
+        type: "m.receipt",
+        room_id: event.getRoomId()
+    };
     fakeReceipt.content[event.getId()] = {};
     fakeReceipt.content[event.getId()][receiptType] = {};
     fakeReceipt.content[event.getId()][receiptType][userId] = {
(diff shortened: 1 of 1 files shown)


lib/models/room.js
```diff
@@ -10,6 +10,19 @@ var MatrixEvent = require("./event").MatrixEvent;
...
+function synthesizeReceipt(userId, event, receiptType) {
+    // This is really ugly because JS has no way to express an object literal
+    // where the name of a key comes from an expression
+    var fakeReceipt = {content: {}};
+    fakeReceipt.content[event.getId()] = {};
+    fakeReceipt.content[event.getId()][receiptType] = {};
+    fakeReceipt.content[event.getId()][receiptType][userId] = {
+        ts: event
+    };
+    return new MatrixEvent(fakeReceipt);
```
Kegsay: It would probably be polite to put `type`, `user_id` and `room_id` on the `fakeReceipt` given we expect those keys to exist on receipts.

lib/models/room.js
```diff
@@ -347,6 +354,45 @@ Room.prototype.recalculate = function(userId) {
...
+        // loop through the timeline backwards looking for either an
+        // event sent by each user or a real receipt from them.
+        // Replace the read receipt for that user with whichever
+        // occurs later in the timeline (ie. first because we're going
+        // backwards).
+        var e = this.timeline[i];
+
+        var readReceiptsForEvent = this.getReceiptsForEvent(e);
+
+        for (var receiptIt = 0; receiptIt < readReceiptsForEvent.length; ++receiptIt) {
```
Kegsay: Confusing var name, I don't know what `receiptIt` is supposed to mean.

Kegsay:
LGTM: This is more efficient than I thought it would be, good job :D

lib/models/room.js
dbkr: Hmm, this is contrary to your documentation, I think? https://github.com/matrix-org/matrix-js-sdk/blob/master/lib/models/room.js#L339

lib/models/room.js
dbkr: 'receipt iterator' because I'd already used 'i' and I generally find using 'j' in such situations a recipe for accidentally referencing the wrong variable. Suggestions welcome.

lib/models/room.js
Kegsay: Aha, you're right it doesn't have `user_id`. The spec has `room_id` and `type` though. My documentation was only documenting the `content` part since that's the part that is being manipulated.

lib/models/room.js
Kegsay: If you're just looping through the entire array I would just use:
```javascript
readReceiptsForEvent.forEach(function(receipt) {
    if (receipt.type !== "m.read") { return; }
    if (usersFound[receipt.userId]) { return; }

     // Then this is the receipt we keep for this user
    usersFound[receipt.userId] = 1;
});
```

dbkr: This fails lint: "Don't make functions within a loop."

Kegsay: Ah. Alright let's just leave it as it is then.

lib/models/room.js
Kegsay: Could you please replace `var fakeReceipt = {content: {}};` with:

```javascript
var fakeReceipt = {
  content: {},
  type: "m.receipt",
  room_id: event.getRoomId()
};
```

Kegsay:
LGTM after https://github.com/matrix-org/matrix-js-sdk/pull/44#discussion_r44282407

lib/models/room.js
dbkr: Oh i see, done.

PR #45, opened 2015-11-09
Invite by 3PID endpoint

Kegsay:
In accordance with http://matrix.org/speculator/spec/head#third-party-invites

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1359,16 +1359,13 @@ MatrixClient.prototype.inviteByEmail = function(roomId, email, displayName, call
  * @param {string} address The address for the specified medium.
  * @param {string} displayName The human-readable name to show in the room for
  * this invite. MUST NOT be the address.
- * @param {string} idServer Optional. The identity server URL to use for inviting.
  * @param {module:client.callback} callback Optional.
  * @return {module:client.Promise} Resolves: TODO
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  * @throws If <code>displayName</code> contains the <code>address</code>.
  */
 MatrixClient.prototype.inviteByThreePid = function(roomId, medium, address,
-                                                   displayName, idServer, callback) {
-    if (utils.isFunction(idServer)) { callback = idServer; idServer = undefined; }
-    idServer = idServer || this.getIdentityServerUrl();
+                                                   displayName, callback) {
     if (displayName && displayName.indexOf(address) >= 0) {
         throw new Error(
             "The display name contains the address. This leaks the 3PID " +
@@ -1382,7 +1379,7 @@ MatrixClient.prototype.inviteByThreePid = function(roomId, medium, address,
     );
 
     return this._http.authedRequest(callback, "POST", path, undefined, {
-        id_server: idServer,
+        id_server: this.getIdentityServerUrl(),
         medium: medium,
         address: address,
         display_name: displayName
--- a/lib/client.js
+++ b/lib/client.js
@@ -1363,15 +1363,15 @@ MatrixClient.prototype.inviteByEmail = function(roomId, email, displayName, call
  * @param {module:client.callback} callback Optional.
  * @return {module:client.Promise} Resolves: TODO
  * @return {module:http-api.MatrixError} Rejects: with an error response.
- * @throws If displayName === address.
+ * @throws If <code>displayName</code> contains the <code>address</code>.
  */
 MatrixClient.prototype.inviteByThreePid = function(roomId, medium, address,
                                                    displayName, idServer, callback) {
     if (utils.isFunction(idServer)) { callback = idServer; idServer = undefined; }
     idServer = idServer || this.getIdentityServerUrl();
-    if (displayName === address) {
+    if (displayName && displayName.indexOf(address) >= 0) {
         throw new Error(
-            "The display name is the same as the address. This leaks the 3PID " +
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -1331,6 +1331,65 @@ MatrixClient.prototype.invite = function(roomId, userId, callback) {
...
+ * @param {module:client.callback} callback Optional.
+ * @return {module:client.Promise} Resolves: TODO
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixClient.prototype.inviteByEmail = function(roomId, email, displayName, callback) {
+    if (utils.isFunction(displayName)) {
+        callback = displayName; displayName = undefined;
+    }
+    if (!displayName) {
+        displayName = email.split("@")[0];
```
dbkr: Maybe we should just use the email address verbatim for the diaplayname here? Email usernames aren't always sensible and nothing's validates the address at this point so if we do split it, we should probably catch it not containing an '@'.

lib/client.js
```diff
@@ -1331,6 +1331,65 @@ MatrixClient.prototype.invite = function(roomId, userId, callback) {
...
+};
+
+/**
+ * Invite a user to a room based on a third-party identifier.
+ * @param {string} roomId The room to invite the user to.
+ * @param {string} medium The medium to invite the user e.g. "email".
+ * @param {string} address The address for the specified medium.
+ * @param {string} displayName The human-readable name to show in the room for
+ * this invite. MUST NOT be the address.
+ * @param {string} idServer Optional. The identity server URL to use for inviting.
```
dbkr: Under what circumstances would we use an ID server other than the one we created the client with? I thought the idiom was that you picked an HS and an IS for the MatrrixClient and used those.

lib/client.js
dbkr: Oh, we don't want to do this because it would leak the 3pid.

dbkr:
Otherwise, lgtm.

lib/client.js
Kegsay: > Maybe we should just use the email address verbatim for the diaplayname here?

We specifically do not want to use the email address for this because it will leak the email to user ID mapping as per http://matrix.org/speculator/spec/head#id43

> if we do split it, we should probably catch it not containing an '@'.

What would you prefer here? If `@` is not present, this will default to the entire string:

```javascript
> "foo".split("@")[0]
'foo'
```

lib/client.js
Kegsay: The HTTP API exposes the ability to set the IS to use; I don't want to be presumptuous that devs want to use the one set on the client (they may trust multiple ISes for example). We have a sensible default so people needn't care, and those that do care don't need to do horrible munging of `this.idBaseUrl` to set it, only to revert it after calling the method.

lib/client.js
dbkr: Oh I see, didn't realise this returned an array of one. This is probably fine then.

lib/client.js
dbkr: Well the C/S API exposes which IS to use by necessity since it needs to know what server to go & hit, just as we need to know what IS to use when we hit the IS directly. The SDK offer the ability to change the IS used on a request-by-request basis anywhere else so I don't think this really needs to be any different.

Kegsay: The SDK does not actually use `this.idBaseUrl` anywhere currently; there isn't any precedent for this. Where possible, the SDK should be as flexible as the HTTP API is IMO.

dbkr: Yes it does: it's passed into the creation of the http-api and used in idServerRequest, which is used for requestEmailToken. requestEmailToken doesn't support specifying a different ID server to the one you created the client with (just as none of the calls to the HS allow you to specify a different HS). If we want to allow this, we should do it consistently, but the current model of the js-sdk I think is to pick a pair of servers at client creation time and stick with them.

lib/client.js
```diff
@@ -1331,6 +1331,65 @@ MatrixClient.prototype.invite = function(roomId, userId, callback) {
...
+ * @param {module:client.callback} callback Optional.
+ * @return {module:client.Promise} Resolves: TODO
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ * @throws If displayName === address.
+ */
+MatrixClient.prototype.inviteByThreePid = function(roomId, medium, address,
+                                                   displayName, idServer, callback) {
+    if (utils.isFunction(idServer)) { callback = idServer; idServer = undefined; }
+    idServer = idServer || this.getIdentityServerUrl();
+    if (displayName === address) {
```
illicitonion: This should probably be a contains check not an equality check?

Kegsay: Good point, will add.

lib/client.js
Kegsay: Fair enough.

lib/client.js
```diff
@@ -1331,6 +1331,65 @@ MatrixClient.prototype.invite = function(roomId, userId, callback) {
 };
 
 /**
+ * Invite a user to a room based on their email address.
+ * @param {string} roomId The room to invite the user to.
+ * @param {string} email The email address to invite.
+ * @param {string} displayName The display name to use. If omitted, everything
```
illicitonion: displayName should probably be removed entirely, given our discussion today, no?

Kegsay:
Right, removed `display_name` entirely.

dbkr:
OK, LGTM.

PR #46, opened 2015-11-10
Guest access endpoints

Kegsay:
This adds the following functions required for guest access:
 - `MatrixClient.registerGuest()` - Flexible registration endpoint `?kind=guest`.
 - `MatrixClient.setGuestRooms(roomIds)` - Sets stuff which is given to `/events` as query params. Provided here rather than in `startClient(opts)` because rooms will be added whilst the client is running.


dbkr:
lgtm

PR #47, opened 2015-11-12
Don't return non-mxc URLs by default.

dbkr:


lib/content-repo.js
```diff
@@ -13,14 +13,23 @@ module.exports = {
...
      * "crop" or "scale".
+     * @param {Boolean} allowDirectLinks If true, return any non-mxc URLs
+     * directly. Fetching such URLs will leak information about the user to
+     * anyone they share a room with. If false, will return the emptry string
+     * for such URLs.
      * @return {string} The complete URL to the content.
      */
-    getHttpUriForMxc: function(baseUrl, mxc, width, height, resizeMethod) {
+    getHttpUriForMxc: function(baseUrl, mxc, width, height,
+                               resizeMethod, allowDirectLinks) {
```
Kegsay: Can we invert the flag here so we don't make a breaking API change?

Existing clients will have `allowDirectLinks == undefined == false` which means it will return an empty string rather than what they put in on lines `:27-32`. Inverting the flag to be `disallowDirectLinks` will keep existing behaviour, and still allow you to block any non-MXC URL.

Kegsay:
LGTM other than comments

lib/content-repo.js
Kegsay: Alternatively you can just do:
```javascript
if (allowDirectLinks === undefined) {
  allowDirectLinks = true;
}
```

dbkr: I feel quite strongly that this should be the default: the behaviour should be secure by default and clients can disable the security features if they want to. It's a breaking change, but IMO it's preferable to leave clients a bit broken but secure than keep the old, insecure behaviour.

Kegsay: Fair enough. Can you add an entry to the `CHANGELOG` please? Same format as other breaking changes (use version `XXX`).

dbkr:
Done.

Kegsay:
v0.3.0 is already out, you need to add a new version :)

dbkr:
I just realised that :p

Kegsay:
LGTM (please backtick "allowDirectLinks")

PR #49, opened 2015-12-02
Add ability to cancel file uploads

dbkr:


dbkr:
NB the test failure is not from develop and not caused by this patch

lib/http-api.js
```diff
@@ -148,16 +155,47 @@ module.exports.MatrixHttpApi.prototype = {
...
+            }
+        });
+        return upload.promise;
+    },
+
+    cancelUpload: function(promise) {
+        var uploadsKeys = Object.keys(this.uploads);
+        for (var i = 0; i < uploadsKeys.length; ++i) {
+            var upload = this.uploads[uploadsKeys[i]];
+            if (upload.promise === promise) {
```
ara4n: comparing promise objects seems like a bit of an ugly way to track which upload is being manipulated here, but I guess this is to avoid changing the upload API to return something more useful than a promise. ah well.

ara4n:
lgtm. might be nice to make it clear in the commit that it's also implementing multiple simultaneous uploads.

lib/http-api.js
dbkr: Yeah, agreed - chatted with Kegan about this and this was the nicest we came up with.

dbkr:
Well it allowed multiple simultaneous uploads before, it just didn't track the status of any of them. :)

PR #50, opened 2015-12-04
Scrollback improvements

Kegsay:
Add a 3s delay between scrollback requests if the previous scrollback request
failed.

Return the same promise if scrollback() is called multiple times whilst a
scrollback request is ongoing.


dbkr:
lftm

dbkr:
lgtm, even

PR #52, opened 2015-12-08
Support v2 filters

Kegsay:
With UTs and integration tests. This does not do `/sync` yet.


dbkr:
lgtm, although I wonder if it's worth running this past erik/mark /someone who has a better idea of how filters are intended to be used by a client, as I don't really know much (eg. is a client expected to store a filter once then remember it's there & trust nothing goes & overwrites it or is it supposed to fetch & check / re-set them now & again?)

Kegsay:
Already asked; they're immutable structures; need to remember the filter ID.

PR #53, opened 2015-12-14
Use /sync instead of /initialSync and /events

Kegsay:
The PR converts the JS SDK to use v2 `/sync` rather than v1 `/initialSync` and `/events`.

A brief overview of the changes:
 - All the syncing code has been moved from `client.js` to a new file: `sync.js`. A nice interface has not yet been written, so `sync.js` horribly gut wrenches `MatrixClient` currently.
 - `MatrixEvent` now supports v2-style events *in addition* to v1.
 - v2 transaction ID matching has been implemented (so the JS SDK does not need to do a linear scan for duplicate suppression on *every event* anymore!)
 - v2 catchup has been implemented (so if you put your computer to sleep overnight and start it back up, it won't lock up your browser!)
 - `/sync` call processing has been made more robust (read: paranoid) - we don't assume keys exists which we did in v1 and we set the sync token *before* processing events to "skip over" bad events.
 - Added a polyfill for `Array.forEach` because seriously, `utils.forEach` is yucky to type.

Tests:
 - All UTs and integration tests have been fixed.
 - Manual testing on Vector `develop` which involved:
    * Login and logout multiple times, make sure it all works.
    * Make sure BAU works (event sending, receiving)
    * Room invites and joining
    * Scrollback
    * Catchup after sleeping for a while

Whilst the diff looks huge, about half of it is just fixing the tests(!)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -621,9 +621,10 @@ MatrixClient.prototype.joinRoom = function(roomIdOrAlias, opts, callback) {
     this._http.authedRequest(undefined, "POST", path, undefined, {}).then(
     function(res) {
         var roomId = res.room_id;
-        var room = createNewRoom(self, roomId); // XXX FIXME TODO
+        var syncApi = new SyncApi(self);
+        var room = syncApi.createRoom(roomId);
         if (opts.syncRoom) {
-            return _syncRoom(self, room); // XXX FIXME TODO
+            return syncApi.syncRoom(room);
         }
         return q(room);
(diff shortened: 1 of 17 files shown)


dbkr:
I'm unsure how filters work (awaiting clarification) but other than that, looks good.

Kegsay:
Noticed a bug whilst doing more testing where there can be multiple `/sync` requests in-flight at the same time. Fixing this before merging.

PR #54, opened 2015-12-15
Do not erroneously emit Call.incoming events on startup

Kegsay:
Check to see if the call was answered or hung up in addition to having a valid
lifetime before emitting the event. Fixes vector-im/vector-web#344


dbkr:
lgtm

PR #55, opened 2015-12-17
Reset /sync backoff if setTimeout takes a long time to fire

Kegsay:
This fixes vector-web/vector-im#536

The bug here is that we were assuming `setTimeout` would fire after the
requested time. This is not true when the machine is asleep. We now timestamp
before and after `setTimeout` and reset the attempt count if we have waited
more than twice what we originally requested. This allows for some jitter which
is to be expected. This also reduces the max waiting time to 32s from 2.1min,
because 2 minutes was deemed excessive given "regular" clients would be polling
with a `timeout=30secs`.

### DIFF
--- a/lib/sync.js
+++ b/lib/sync.js
@@ -497,7 +497,8 @@ function startSyncingRetryTimer(client, attempt, fn) {
         var timeDeltaMs = timeAfterWaitingMs - timeBeforeWaitingMs;
         if (timeDeltaMs > (2 * timeToWaitMs)) {
             // we've waited more than twice what we were supposed to. Reset the
-            // attempt number back to 1;
+            // attempt number back to 1. This can happen when the comp goes to
+            // sleep whilst the timer is running.
             newAttempt = 1;
             console.warn(
                 "Sync retry timer: Tried to wait %s ms but actually waited %s ms",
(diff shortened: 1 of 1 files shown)


lib/sync.js
```diff
@@ -481,17 +481,31 @@ SyncApi.prototype._processRoomEvents = function(room, stateEventList,
...
+    var newAttempt = attempt;
+    var timeBeforeWaitingMs = Date.now();
+    var timeToWaitMs = retryTimeMsForAttempt(attempt);
     client._syncingRetry.timeoutId = setTimeout(function() {
-        fn();
-    }, retryTimeMsForAttempt(attempt));
+        var timeAfterWaitingMs = Date.now();
+        var timeDeltaMs = timeAfterWaitingMs - timeBeforeWaitingMs;
+        if (timeDeltaMs > (2 * timeToWaitMs)) {
+            // we've waited more than twice what we were supposed to. Reset the
```
ara4n: Should say why this could ever happen - which is basically that the comp went to sleep, or the clock was stepped.

The 2x factor feels a bit weird.  I'd have thought +/- 10% difference in what we expected should be reason to be suspicious and reset the retry schedule.

ara4n:
lgtm other than the proviso.

lib/sync.js
Kegsay: The 2x factor is mainly to provide stronger guarantees that this will fire infrequently, given we do warning logs. It would be annoying to see them constantly in the logs just because the DOM liked to block one iteration of the event loop for 2s * 0.1 = 200ms, which you've seen longer before e.g. https://github.com/vector-im/vector-web/issues/438

Will add a note about why this could happen.

Kegsay:
PTAL

ara4n:
lgtm

PR #56, opened 2015-12-18
Add syncLeftRooms()

Kegsay:
Implemented for https://github.com/vector-im/vector-web/issues/348

This adds `MatrixClient.syncLeftRooms()` which fetches left rooms and dumps them into the store. Everything Just Works. Multiple calls to `syncLeftRooms()` will return the same promise if the request is ongoing, or NOP if the request was successful. If the request failed, it will be retried. Manually tested these cases.

See also:
 - https://github.com/vector-im/vector-web/pull/541
 - https://github.com/matrix-org/matrix-react-sdk/pull/62


ara4n:
lgtm other than the extreme ming of the archived room filter having to also pull in all of the existing rooms you're  in.  filed that one as spec-307

PR #58, opened 2015-12-22
Propagate the unread count from sync onto rooms

dbkr:
Requires https://github.com/matrix-org/synapse/pull/456 to work sensibly (without it will just set the unread count to undefined).


Kegsay:
LGTM

PR #59, opened 2015-12-24
Enhancements to search results, and event context implementation

richvdh:
This change adds support to the JDK for processing the results of a room
search, as well as back-paginating the results.

It treats each search result as a 'context' object, which can itself be
backwards or forward-paginated.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -12,6 +12,7 @@ var q = require("q");
 var httpApi = require("./http-api");
 var MatrixEvent = require("./models/event").MatrixEvent;
 var EventStatus = require("./models/event").EventStatus;
+var SearchResult = require("./models/search-result");
 var StubStore = require("./store/stub");
 var webRtcCall = require("./webrtc/call");
 var utils = require("./utils");
@@ -1901,6 +1902,73 @@ MatrixClient.prototype.scrollback = function(room, limit, callback) {
     return defer.promise;
 };
 
+/**
+ * Take an EventContext, and back/forward-fill results.
+ *
+ * @param {module:models/event-context.EventContext} eventContext  context
+ *    object to be updated
+ * @param {Object}  opts
+ * @param {boolean} opts.backwards  true to fill backwards, false to go forwards
+ * @param {boolean} opts.limit      number of events to request
+ *
+ * @return {module:client.Promise} Resolves: updated EventContext object
+ * @return {Error} Rejects: with an error response.
+ */
+MatrixClient.prototype.paginateEventContext = function(eventContext, opts) {
+    // TODO: we should implement a backoff (as per scrollback()) to deal more
+    // nicely with HTTP errors.
+    opts = opts || {};
+    var backwards = opts.backwards || false;
+
+    var token = eventContext.getPaginateToken(backwards);
+    if (!token) {
+        // no more results.
+        return q.reject(new Error("No paginate token"));
+    }
+
+    var dir = backwards ? 'b' : 'f';
+    var pendingRequest = eventContext._paginateRequests[dir];
+
+    if (pendingRequest) {
+        // already a request in progress - return the existing promise
+        return pendingRequest;
+    }
+
+    var path = utils.encodeUri(
+        "/rooms/$roomId/messages", {$roomId: eventContext.getEvent().getRoomId()}
+    );
+    var params = {
+        from: token,
+        limit: ('limit' in opts) ? opts.limit : 30,
+        dir: dir
+    };
+
+    var self = this;
+    var promise =
+        self._http.authedRequest(undefined, "GET", path, params
+    ).then(function(res) {
+        var token = res.end;
+        if (res.chunk.length === 0) {
+            token = null;
+        } else {
+            var matrixEvents = utils.map(res.chunk, self.getEventMapper());
+            if (backwards) {
+                // eventContext expects the events in timeline order, but
+                // back-pagination returns them in reverse order.
+                matrixEvents.reverse();
+            }
+            eventContext.addEvents(matrixEvents, backwards);
+        }
+        eventContext.setPaginateToken(token, backwards);
+        return eventContext;
+    }).finally(function() {
+        eventContext._paginateRequests[dir] = null;
+    });
+    eventContext._paginateRequests[dir] = promise;
+
+    return promise;
+};
+
 // Registration/Login operations
 // =============================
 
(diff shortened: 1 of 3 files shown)


richvdh:
bugrit; let me investigate the failures.

lib/client.js
```diff
@@ -1901,6 +1902,71 @@ MatrixClient.prototype.scrollback = function(room, limit, callback) {
...
+        // already a request in progress - return the existing promise
+        return pendingRequest;
+    }
+
+    var path = utils.encodeUri(
+        "/rooms/$roomId/messages", {$roomId: eventContext.getEvent().getRoomId()}
+    );
+    var params = {
+        from: token,
+        limit: 30,
```
Kegsay: It would be worth moving this to be configurable. Might be worth having the function signature be `paginateEventContext(eventContent, opts)` which has `opts.backwards` and `opts.limit`?

lib/client.js
```diff
@@ -1901,6 +1902,71 @@ MatrixClient.prototype.scrollback = function(room, limit, callback) {
...
+                // back-pagination returns them in reverse order.
+                matrixEvents.reverse();
+            }
+            eventContext.addEvents(matrixEvents, backwards);
+        }
+        eventContext.setPaginateToken(token, backwards);
+        return eventContext;
+    }).finally(function() {
+        eventContext._paginateRequests[dir] = null;
+    });
```
Kegsay: Confusingly enough you need a `done()` here else you're silently dropping errors if there was any uncaught error thrown in the finally block. Yay promises.

```
}).done();
```

lib/client.js
```diff
@@ -1901,6 +1902,71 @@ MatrixClient.prototype.scrollback = function(room, limit, callback) {
...
+    if (promise.isPending()) {
+        eventContext._paginateRequests[dir] = promise;
+    }
```
Kegsay: I *think* this `isPending` check is redundant. All fulfilled/rejected handlers on Promises are set to go off asynchronously, so your `then()` and `finally()` blocks can't possibly have executed yet. If you were to immediately return a resolved/rejected promise (e.g. via `return q.reject("bad things");`) then this would be `false`, but I don't see why you'd ever want to do that. If there is a reason for that, please comment it as it is unusual.

lib/client.js
```diff
@@ -1901,6 +1902,71 @@ MatrixClient.prototype.scrollback = function(room, limit, callback) {
...
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixClient.prototype.paginateEventContext = function(eventContext, backwards) {
+    // TODO: we should implement a backoff (as per scrollback()) to deal more
+    // nicely with HTTP errors.
+
+    var token = eventContext.getPaginateToken(backwards);
+    if (!token) {
+        // no more results.
+        throw new Error("No paginate token");
```
Kegsay: Undocumented `throw`. I generally would prefer if we *didn't* throw regardless, and instead return a rejected promise e.g. `return q.reject(someError)` because otherwise you're forcing clients to have to have 2 ways of handling error responses from this single call (rejected promises and a try/catch) which sucks.

lib/client.js
```diff
@@ -2096,6 +2162,125 @@ MatrixClient.prototype.searchMessageText = function(opts, callback) {
...
+ *    results object to be updated
+ * @return {module:client.Promise} Resolves: updated result object
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixClient.prototype.backPaginateRoomEventsSearch = function(searchResults) {
+    // TODO: we should implement a backoff (as per scrollback()) to deal more
+    // nicely with HTTP errors.
+
+    if (!searchResults.next_batch) {
+        throw new Error("Cannot backpaginate event search any further");
```
Kegsay: See earlier comment about throwing as API contracts.

lib/client.js
```diff
@@ -2096,6 +2162,125 @@ MatrixClient.prototype.searchMessageText = function(opts, callback) {
...
+    var searchOpts = {
+        body: searchResults._query,
+        next_batch: searchResults.next_batch,
+    };
+
+    var promise = this.search(searchOpts).then(
+        this._processRoomEventsSearch.bind(this, searchResults)
+    ).finally(function() {
+        searchResults.pendingRequest = null;
+    });
```
Kegsay: Missing `.done()`

Kegsay: Oh but this one is interesting because you're then shoving the promise away, hmm :/ you'll need to do `promise.done()` stand-alone then.

lib/client.js
Kegsay: See https://github.com/matrix-org/matrix-js-sdk/pull/59#discussion_r48406100 - you need to do this on a separate line.

lib/models/search-result.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Construct a new SearchResult
+ *
+ * @param {number} rank   where this SearchResult ranks in the results
+ * @param {event-context.EventContext} event_context  the matching event and its
+ *    context
+ *
+ * @constructor
+ */
+function SearchResult(rank, event_context) {
```
Kegsay: Nit: Prefer `camelCase` where possible. The only places where we should be using `under_scores` is when we're directly manipulating JSON responses from the server which are, of course, with underscores.

lib/client.js
```diff
@@ -2096,6 +2162,125 @@ MatrixClient.prototype.searchMessageText = function(opts, callback) {
...
+    room_events.highlights.forEach(function(hl) { highlights[hl] = 1; });
+    searchResults.highlights.forEach(function(hl) { highlights[hl] = 1; });
+
+    // turn it back into a list.
+    searchResults.highlights = Object.keys(highlights);
+
+    // append the new results to our existing results
+    for (var i = 0; i < room_events.results.length; i++) {
+        var sr = SearchResult.fromJson(room_events.results[i], this.getEventMapper());
+        searchResults.results.push(sr);
```
Kegsay: I don't like how you're assigning `highlights` and `results` to a `SearchResults` which doesn't think it has those properties in the JSDoc for `SearchResults`. I don't like it because it encourages object instances to be a grab-bag of props which you're never really sure where something is assigned and have no real way to grep for it either. Please add setters and document what these properties are supposed to represent. I don't mind if you don't add explicit getter methods, provided you instead document the `@prop` so it is clear that it is public-facing.

lib/models/event-context.js
```diff
@@ -0,0 +1,104 @@
...
+ * it.
+ *
+ * It also stores pagination tokens for going backwards and forwards in the
+ * timeline.
+ *
+ * @param {MatrixEvent} our_event  the event at the centre of this context
+ *
+ * @constructor
+ */
+function EventContext(our_event) {
```
Kegsay: Nit: `camelCase` please

lib/models/event-context.js
```diff
@@ -0,0 +1,104 @@
...
+    this._timeline = [our_event];
+    this._ourEventIndex = 0;
+    this._paginateTokens = {};
```
Kegsay: Would be a good idea to put a one-liner explaining what the key/values here should be as it is not immediately clear where these tokens appear.

Kegsay:
Mostly LGTM aside from small things. Sorry if I seem to be a bit strict on some of this; the JS SDK forms the basis for all the other things so I tend to have a higher bar and more nits!

lib/client.js
richvdh: Done

lib/client.js
richvdh: removed.

done.

lib/client.js
richvdh: per #matrix, the expectation is that the caller calls done()

lib/client.js
richvdh: yup; replaced with q.reject.

done

lib/client.js
richvdh: done

lib/client.js
richvdh: as above

lib/models/search-result.js
richvdh: done

lib/client.js
richvdh: ugh, this is actually a failure of the jsdoc on _processRoomEventsSearch. It doesn't take a SearchResults (there is no such thing) but rather an anonymous object (with the fields as doucmented on searchRoomEvents).

richvdh: Fixed the jsdoc on _processRoomEventsSearch and backPaginateRoomEventsSearch

done

lib/models/event-context.js
richvdh: done

lib/models/event-context.js
richvdh: I've updated the initialiser to set default values.

lib/client.js
richvdh: ftr, I found the bit of spec which makes this redundant: https://promisesaplus.com/#point-34, and https://promisesaplus.com/#point-67.

Kegsay:
LGTM

PR #60, opened 2016-01-07
Add constructs for guest access

Kegsay:
Fixes https://github.com/vector-im/vector-web/issues/81

Specifically this adds:
 - The ability to peek on a single room. `MatrixClient.peekInRoom`
 - The ability to set guest access rules on a room. `MatrixClient.setGuestAccess`
 - Helpful getters for the access_token and user ID localpart (to submit in order to upgrade accounts). `MatrixClient.getAccessToken` and `MatrixClient.getUserIdLocalpart`

This modifies `MatrixClient.register` to accept another arg `guestAccessToken` to get upgrades working. Manually tested.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1943,13 +1943,15 @@ MatrixClient.prototype.registerGuest = function(opts, callback) {
  * Peek into a room and receive updates about the room. This only works if the
  * history visibility for the room is world_readable.
  * @param {String} roomId The room to attempt to peek into.
+ * @return {module:client.Promise} Resolves: Room object
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
 MatrixClient.prototype.peekInRoom = function(roomId) {
     if (this._peekSync) {
         this._peekSync.stopPeeking();
     }
     this._peekSync = new SyncApi(this);
-    this._peekSync.peek(roomId);
+    return this._peekSync.peek(roomId);
 };
 
 /**
--- a/lib/client.js
+++ b/lib/client.js
@@ -1269,6 +1269,10 @@ MatrixClient.prototype.sendHtmlEmote = function(roomId, body, htmlBody, callback
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
 MatrixClient.prototype.sendReceipt = function(event, receiptType, callback) {
+    if (this.isGuest()) {
+        return q({}); // guests cannot send receipts so don't bother.
+    }
+
     var path = utils.encodeUri("/rooms/$roomId/receipt/$receiptType/$eventId", {
         $roomId: event.getRoomId(),
         $receiptType: receiptType,
@@ -1332,6 +1336,10 @@ MatrixClient.prototype.getCurrentUploads = function() {
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
 MatrixClient.prototype.sendTyping = function(roomId, isTyping, timeoutMs, callback) {
+    if (this.isGuest()) {
+        return q({}); // guests cannot send typing notifications so don't bother.
+    }
+
     var path = utils.encodeUri("/rooms/$roomId/typing/$userId", {
         $roomId: roomId,
         $userId: this.credentials.userId
--- a/lib/client.js
+++ b/lib/client.js
@@ -1947,6 +1947,34 @@ MatrixClient.prototype.setGuestRooms = function(roomIds) {
 };
 
 /**
+ * Set r/w flags for guest access in a room.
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -1939,11 +1967,46 @@ MatrixClient.prototype.registerGuest = function(opts, callback) {
...
+    var writePromise = this.sendStateEvent(roomId, "m.room.guest_access", {
+        guest_access: opts.allowJoin ? "can_join" : "forbidden"
+    });
+
+    var readPromise = q();
+    if (opts.allowRead) {
+        readPromise = this.sendStateEvent(roomId, "m.room.history_visibility", {
+            history_visibility: "world_readable"
+        });
+    }
```
dbkr: Not sure I understand why we set history visibility to world readable when allowing guest access? Is it assumed that guest access without history visibility is useless, and if so, is the js sdk the correct place to make this assumption?

lib/client.js
```diff
@@ -2502,6 +2568,10 @@ function checkTurnServers(client) {
     if (!client._supportsVoip) {
         return;
     }
+    if (client.isGuest()) {
+        return; // guests can't access TURN servers
```
dbkr: This comment probably isn't really about the JS SDK but this surprises me. I have queried this OOB.

lib/sync.js
```diff
@@ -140,6 +141,100 @@ SyncApi.prototype.syncLeftRooms = function() {
...
+        return peekRoom;
+    });
+};
+
+/**
+ * Stop polling for updates in the peeked room. NOPs if there is no room being
+ * peeked.
+ */
+SyncApi.prototype.stopPeeking = function() {
+    this._peekRoomId = null;
```
dbkr: Shouldn't we call _peekPoll(null) here to stop the poll?

lib/sync.js
```diff
@@ -543,7 +639,23 @@ SyncApi.prototype._processRoomEvents = function(room, stateEventList,
...
+    if (!guestRooms) {
+        return "{}";
+    }
+    return JSON.stringify({
+        room: {
+            timeline: {
+                limit: 20
+            }
+        }
+    });
```
dbkr: If I've read this correctly, we add a limit of 20 to the sync query if we're peeking into a room (which is done via a separate query?) Why?

dbkr:
Otherwise, looking sensible.

lib/sync.js
Kegsay: This is actually the timeline limit of the joined rooms, not the peeked rooms. The number is arbitrary.

lib/client.js
Kegsay: Any actionable thing to do here?

lib/client.js
Kegsay: This function accepts two flags: `allowJoin` and `allowRead`. If `allowRead` is set, then the history visibility is set to `world_readable` (this is how you give read access to guests). If `allowJoin` is set, then the guest_access event is sent with `can_join`, else it is set to `forbidden`. Unlike `allowJoin`, there is no true boolean state for `allowRead` as the JS SDK doesn't know what to set the history visibility to if you set `allowRead: false` hence it NOPs.

dbkr: Oh yes, I see. Do we not want to allow people turn world readability off again?

dbkr: Also is there a reason for these params being in a object rather than... parameters?

lib/client.js
dbkr: Nobody's got back to me so from a code PoV this is fine, and that I think is all we care about for purposes of PR review.

lib/sync.js
dbkr: Right, that's my question: why is there a link between whether we're peeking into any rooms or not and the timeline limit of joined rooms?

lib/client.js
Kegsay: > Do we not want to allow people turn world readability off again?

We don't reasonably know what the previous history visibility was to set it back to if they decide to turn it off.

>  is there a reason for these params being in a object rather than... parameters?

Because this makes it much more extensible when inevitably we add in more params. I wish I did this on more things earlier (e.g. `MatrixClient.register()`).

lib/sync.js
Kegsay: There isn't any link between the two. `_getGuestFilter` is called on `/sync` (joined rooms) requests only.

dbkr: IRL: we just need to specify the filter inline if we're a guest because guests can't create filters.

dbkr:
OK, post comments, lgtm

PR #61, opened 2016-01-08
implement account data

ara4n:


lib/client.js
```diff
@@ -1782,7 +1819,7 @@ MatrixClient.prototype.setPresence = function(presence, callback) {
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
 MatrixClient.prototype.publicRooms = function(callback) {
-    return this._http.request(callback, "GET", "/publicRooms");
+    return this._http.authedRequest(callback, "GET", "/publicRooms");
```
Kegsay: Why have you changed this? You don't need auth to do this request.

lib/models/room.js
```diff
@@ -62,6 +62,9 @@ function synthesizeReceipt(userId, event, receiptType) {
  * this room.
  * @prop {object} tags Dict of room tags; the keys are the tag name and the values
  * are any metadata associated with the tag - e.g. { "fav" : { order: 1 } }
+ * @prop {object} accountData Dict of per-room account_data events; the keys are the
+ * event type and the values are the events.
+ * are any metadata associated with the tag - e.g. { "fav" : { order: 1 } }
```
Kegsay: Spurious extra line from earlier prop.

lib/models/room.js
```diff
@@ -750,3 +778,16 @@ module.exports = Room;
...
  */
+
+/**
+ * Fires whenever a room's account_data is updated.
+ * @event module:client~MatrixClient#"Room.accountData"
+ * @param {event} event The account_data event
+ * @param {Room} room The room whose account_data was updated.
+ * @example
+ * matrixClient.on("Room.accountData", function(event, room){
+ *   if (event.getType() === "m.room.colorscheme") {
```
Kegsay: Unless this is part of the spec, can we not use `m.*` please. `com.example.colorscheme` would be a better example.

Kegsay:
I would like a small rejig of how you're doing this before LGTMing because I feel that the current way of adding account data to rooms is needlessly complex.

At the moment, we get the `/sync` response and then slap on a `section` for every single event which indicates where in the `/sync` response it originated from. I don't like this because this tightly couples *all events* with the *structure* of the /sync response. This makes it harder to use existing code paths if we want to parse an endpoint which is not `/sync`. We then call the generic `Room.addEvents` method which then conditionally checks for `account_data` then calls through to `Room.addAccountData`.

The `/sync` response function *already knows* which events are `account_data` events (it's under the key!) so can we please just directly call `Room.addAccountData(event)` and be happy? This means we don't need to corrupt the `MatrixEvent` object with a `section` prop like we are currently. After this is done, LGTM aside from minor line comments.

lib/client.js
ara4n: futureproofing for SPEC-321

lib/models/room.js
ara4n: oops, thanks - fixed :)

ara4n:
Right - that's a very good point, thanks for the feedback.  Have rewritten to match.

One question remains, which is that the same complaint surely goes for trying to wedge ephemeral events into the timeline too - shouldn't these be directly piped into a room.addEphemeralEvents()?

@kegsay: PTAL

lib/client.js
Kegsay: Okay

lib/models/room.js
```diff
@@ -600,6 +604,30 @@ Room.prototype.addTags = function(event) {
...
+/**
+ * Update the account_data events for this room, overwriting events of the same type.
+ * @param {Array<MatrixEvent>} events an array of account_data events to add
+ */
+Room.prototype.addAccountData = function(events) {
+    for (var i = 0; i < events.length; i++) {
+        var event = events[i];
+        if (event.getType() === "m.tag") {
+            this.addTags(event);
+        }
```
Kegsay: We should probably put a `continue` in this `if` block else `m.tag` events will be added to the `accountData` dict, which is presumably unintentional.

lib/models/room.js
```diff
@@ -600,6 +604,30 @@ Room.prototype.addTags = function(event) {
...
+        this.accountData[event.getType()] = event;
+        this.emit("Room.accountData", event, this);
+    }
+};
+
+/**
+ * Access account_data event of given event type for this room
+ * @param {string} type the type of account_data event to be accessed
+ * @return {MatrixEvent} the account_data event in question
```
Kegsay: Nit: Technically this should be `{?MatrixEvent}` as it is possible for this function to return `undefined`.

Kegsay:
> shouldn't these be directly piped into a room.addEphemeralEvents()

Absolutely, they should be.

LGTM aside from a question and a JSDoc nit.

lib/models/room.js
ara4n: no, this was intentional.  they *are* account_data, even though we also provide an alternative API for getting at them via the js-sdk.

lib/models/room.js
ara4n: fixed, thanks.

PR #62, opened 2016-01-10
Room.getImplicitRoomName

ara4n:
add a Room.getImplicitRoomName so clients can know what a room would be named if it had no explicit m.room.name.

(This seems to also be pulling in matthew/accountdata - oops)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -192,6 +192,14 @@ MatrixClient.prototype.getIdentityServerUrl = function() {
 };
 
 /**
+ * Get the domain for this client's MXID
+ * @return {string} Domain of this MXID
+ */
+MatrixClient.prototype.getDomain = function() {
+    return this.credentials.userId.replace(/^.*:/, '');
+};
+
+/**
  * Check if the runtime environment supports VoIP calling.
  * @return {boolean} True if VoIP is supported.
  */
--- a/lib/client.js
+++ b/lib/client.js
@@ -196,7 +196,7 @@ MatrixClient.prototype.getIdentityServerUrl = function() {
  * @return {string} Domain of this MXID
  */
 MatrixClient.prototype.getDomain = function() {
-    return this.credentials.userId.replace(/^.*:/, '');
+    return this.credentials.userId.split(':')[0];
 };
 
 /**
--- a/lib/models/room.js
+++ b/lib/models/room.js
@@ -667,12 +667,12 @@ function setEventMetadata(event, stateContext, toStartOfTimeline) {
  * @param {Room} room The matrix room.
  * @param {string} userId The client's user ID. Used to filter room members
  * correctly.
- * @param {bool} implicit Return the implicit room name that we'd see if there
+ * @param {bool} ignoreRoomNameEvent Return the implicit room name that we'd see if there
  * was no m.room.name event.
  * @return {string} The calculated room name.
  */
-function calculateRoomName(room, userId, implicit) {
-    if (!implicit) {
+function calculateRoomName(room, userId, ignoreRoomNameEvent) {
+    if (!ignoreRoomNameEvent) {
         // check for an alias, if any. for now, assume first alias is the
         // official one.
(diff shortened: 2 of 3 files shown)


Kegsay:
This is blocked pending action on https://github.com/matrix-org/matrix-js-sdk/pull/61

lib/models/room.js
```diff
@@ -627,14 +667,18 @@ function setEventMetadata(event, stateContext, toStartOfTimeline) {
  * @param {Room} room The matrix room.
  * @param {string} userId The client's user ID. Used to filter room members
  * correctly.
+ * @param {bool} implicit Return the implicit room name that we'd see if there
+ * was no m.room.name event.
```
Kegsay: Why is this called `implicit`? All this flag does is ignore the `m.room.name` event if it exists, so can we perhaps call it something more descriptive like `ignoreRoomNameEvent`?

lib/models/room.js
```diff
@@ -185,6 +191,18 @@ Room.prototype.getAvatarUrl = function(baseUrl, width, height, resizeMethod,
  };
 
  /**
+  * Get the implicit room name (i.e. what a given user would see if the
+  * room had no m.room.name)
+  * @param {string} userId The userId from whose perspective we want
+  * to calculate the implicit name
+  * @return {string} The implicit room name
+  */
+ Room.prototype.getImplicitRoomName = function(userId) {
```
Kegsay: What are you planning on using this for? It seems to be conflating 2 ideas and has a rather vague function name, so I'm immediately skeptical. The 2 ideas being conflated are:
 - Getting the room name assuming there is no `m.room.name` event.
 - Getting the room name from the perspective of another user ID.

I don't know why you'd want to do either of these things, hence asking for more information on the intended usage.

ara4n: as per the PR: "add a Room.getImplicitRoomName so clients can know what a room would be named if it had no explicit m.room.name.".  The use case is maintaining the RoomSettings UI to show what a room would be called if you try to delete its name.  The only reason it takes a userId is because calculateRoomName does too, and the Room object doesn't appear to give you a way to get at the current js-sdk user's userId.

I thought 'implicitRoomName' - i.e. the implicit name a room would have if a user hasn't specified an explicit one was fine terminology personally.  Do you have a better suggestion?

lib/models/room.js
ara4n: fixed, thanks

ara4n:
n.b. have also thrown a getDomain() into the mix

ara4n:
@kegsay ptal

lib/client.js
```diff
@@ -192,6 +192,14 @@ MatrixClient.prototype.getIdentityServerUrl = function() {
 };
 
 /**
+ * Get the domain for this client's MXID
+ * @return {string} Domain of this MXID
+ */
+MatrixClient.prototype.getDomain = function() {
+    return this.credentials.userId.replace(/^.*:/, '');
```
Kegsay: This regex will fail for the user ID `@foo:bar:8448`.

lib/models/room.js
Kegsay: Okay, I understand now.

> I thought 'implicitRoomName' - i.e. the implicit name a room would have if a user hasn't specified an explicit one was fine terminology personally. Do you have a better suggestion?

Looking from your explanation of what you're using this for:

> to show what a room would be called if you try to delete its name

I'm tempted to just call this `Room.getDefaultName()`. People are happy with the idea of "default" values (without something explicitly set) so I'd prefer to piggyback off that rather than "implicit" which isn't used anywhere near as much and is more subjective IMO.

Kegsay:
LGTM aside from regex bug and naming. You'll need to:

```
(matthew/roomsettings2)$ git merge matthew/accountdata
```

before merging to make things happy.

lib/client.js
ara4n: well, that's embarassing. thanks!

PR #63, opened 2016-01-15
Support for non-contiguous event timelines

richvdh:
This provides optional support for fetching old events via the /context API,
and paginating backwards and forwards from them, eventually merging into the
live timeline.

To support it, events are now stored in an EventTimeline, rather than directly
in an array in the Room; the old names are maintained as references for
compatibility.

The feature has to be enabled explicitly, otherwise it would be impossible for
existing clients to back-paginate to the old events after a gappy /sync.

Still TODO here:

* An object which provides a window into the timelines to make them possible to
  use. (This is pr #64) .

* Rewrite the 'EventContext' used by the searchRoomEvents API in terms of an
  EventTimeline - it is essentially a subset. This isn't urgent, and can wait until this lump has landed.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -81,10 +81,12 @@ var OLM_ALGORITHM = "m.olm.v1.curve25519-aes-sha2";
  * to all requests with this client. Useful for application services which require
  * <code>?user_id=</code>.
  *
- * @param {boolean} [opts.timelineSupport = false] Set to true to enable improved
- * timeline support ((@link getEventTimeline}). It is disabled by default for
- * compatibility with older clients - in particular to maintain support for
- * back-paginating the live timeline after a '/sync' result with a gap.
+ * @param {boolean} [opts.timelineSupport = false] Set to true to enable
+ * improved timeline support ({@link
+ * module:client~MatrixClient#getEventTimeline getEventTimeline}). It is
+ * disabled by default for compatibility with older clients - in particular to
+ * maintain support for back-paginating the live timeline after a '/sync'
+ * result with a gap.
  */
 function MatrixClient(opts) {
     utils.checkObjectHasKeys(opts, ["baseUrl", "request"]);
@@ -2148,7 +2150,7 @@ MatrixClient.prototype.getEventTimeline = function(room, eventId) {
  *
  * @param {module:models/event-timeline~EventTimeline} eventTimeline timeline
  *    object to be updated
- * @param {Object =} opts
+ * @param {Object}   [opts]
  * @param {boolean}  [opts.backwards = false]  true to fill backwards,
  *    false to go forwards
  * @param {number}   [opts.limit = 30]         number of events to request
--- a/lib/client.js
+++ b/lib/client.js
@@ -2072,8 +2072,8 @@ MatrixClient.prototype.paginateEventContext = function(eventContext, opts) {
  * Get an EventTimeline for the given event
  *
  * <p>If the room object already has the given event in its store, the
- * corresponding timeline will be returned. Otherwise, we make a /context
- * request, and use it to construct an EventTimeline.
+ * corresponding timeline will be returned. Otherwise, a /context request is
+ * made, and used to construct an EventTimeline.
  *
  * @param {Room} room  The room to look for the event in
  * @param {string} eventId  The ID of the event to look for
@@ -2085,7 +2085,7 @@ MatrixClient.prototype.paginateEventContext = function(eventContext, opts) {
 MatrixClient.prototype.getEventTimeline = function(room, eventId) {
     // don't allow any timeline support unless it's been enabled.
     if (!this.timelineSupport) {
-        throw Error("timeline support is disabled. Set the 'timelineSupport'" +
+        throw new Error("timeline support is disabled. Set the 'timelineSupport'" +
                     " parameter to true when creating MatrixClient to enable" +
                     " it.");
     }
@@ -2108,7 +2108,7 @@ MatrixClient.prototype.getEventTimeline = function(room, eventId) {
         self._http.authedRequest(undefined, "GET", path
     ).then(function(res) {
         if (!res.event) {
-            throw Error("'event' not in '/context' result - homeserver too old?");
+            throw new Error("'event' not in '/context' result - homeserver too old?");
         }
 
         // by the time the request completes, the event might have ended up in
@@ -2127,7 +2127,7 @@ MatrixClient.prototype.getEventTimeline = function(room, eventId) {
 
         var timeline = room.getTimelineForEvent(matrixEvents[0].getId());
         if (!timeline) {
-            timeline = room.createTimeline();
+            timeline = room.addTimeline();
             timeline.initialiseState(utils.map(res.state,
                                                self.getEventMapper()));
             timeline.getState(false).paginationToken = res.end;
@@ -2155,8 +2155,8 @@ MatrixClient.prototype.getEventTimeline = function(room, eventId) {
  *    false to go forwards
  * @param {number}   [opts.limit = 30]         number of events to request
  *
- * @return {module:client.Promise} Resolves: false if there are no events and
- *    we reached the end of the timeline; else true. Rejects: Error with an error response
+ * @return {module:client.Promise} Resolves to a boolean: false if there are no
+ *    events and we reached either end of the timeline; else true.
  */
 MatrixClient.prototype.paginateEventTimeline = function(eventTimeline, opts) {
     // TODO: we should implement a backoff (as per scrollback()) to deal more
@@ -2164,10 +2164,15 @@ MatrixClient.prototype.paginateEventTimeline = function(eventTimeline, opts) {
     opts = opts || {};
     var backwards = opts.backwards || false;
 
+    var room = this.getRoom(eventTimeline.getRoomId());
+    if (!room) {
+        throw new Error("Unknown room " + eventTimeline.getRoomId());
+    }
+
     var token = eventTimeline.getPaginationToken(backwards);
     if (!token) {
-        // no more results.
-        return q.reject(new Error("No paginate token"));
+        // no token - no results.
+        return q(false);
     }
 
     var dir = backwards ? 'b' : 'f';
@@ -2193,9 +2198,6 @@ MatrixClient.prototype.paginateEventTimeline = function(eventTimeline, opts) {
         this._http.authedRequest(undefined, "GET", path, params
     ).then(function(res) {
         var token = res.end;
-        console.log("client: completed paginate; backwards=" + backwards +
-                    "; token=" + token);
-        var room = self.getRoom(eventTimeline.getRoomId());
         var matrixEvents = utils.map(res.chunk, self.getEventMapper());
         room.addEventsToTimeline(matrixEvents, backwards, eventTimeline, token);
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -2196,6 +2196,13 @@ MatrixClient.prototype.paginateEventTimeline = function(eventTimeline, opts) {
         var room = self.getRoom(eventTimeline.getRoomId());
         var matrixEvents = utils.map(res.chunk, self.getEventMapper());
         room.addEventsToTimeline(matrixEvents, backwards, eventTimeline, token);
+
+        // if we've hit the end of the timeline, we need to stop trying to
+        // paginate. We need to keep the 'forwards' token though, to make sure
+        // we can recover from gappy syncs.
+        if (backwards && res.end == res.start) {
+            eventTimeline.setPaginationToken(null, true);
+        }
         return res.end != res.start;
     }).finally(function() {
         eventTimeline._paginationRequests[dir] = null;
--- a/lib/client.js
+++ b/lib/client.js
@@ -27,6 +27,7 @@ var q = require("q");
 var httpApi = require("./http-api");
 var MatrixEvent = require("./models/event").MatrixEvent;
 var EventStatus = require("./models/event").EventStatus;
+var EventTimeline = require("./models/event-timeline");
 var SearchResult = require("./models/search-result");
 var StubStore = require("./store/stub");
 var webRtcCall = require("./webrtc/call");
@@ -2155,7 +2156,7 @@ MatrixClient.prototype.getEventTimeline = function(room, eventId) {
             timeline = room.addTimeline();
             timeline.initialiseState(utils.map(res.state,
                                                self.getEventMapper()));
-            timeline.getState(false).paginationToken = res.end;
+            timeline.getState(EventTimeline.FORWARDS).paginationToken = res.end;
         }
         room.addEventsToTimeline(matrixEvents, true, timeline, res.start);
 
@@ -2194,13 +2195,14 @@ MatrixClient.prototype.paginateEventTimeline = function(eventTimeline, opts) {
         throw new Error("Unknown room " + eventTimeline.getRoomId());
     }
 
-    var token = eventTimeline.getPaginationToken(backwards);
+    var dir = backwards ? EventTimeline.BACKWARDS : EventTimeline.FORWARDS;
+
+    var token = eventTimeline.getPaginationToken(dir);
     if (!token) {
         // no token - no results.
         return q(false);
     }
 
-    var dir = backwards ? 'b' : 'f';
     var pendingRequest = eventTimeline._paginationRequests[dir];
 
     if (pendingRequest) {
@@ -2230,7 +2232,7 @@ MatrixClient.prototype.paginateEventTimeline = function(eventTimeline, opts) {
         // paginate. We need to keep the 'forwards' token though, to make sure
         // we can recover from gappy syncs.
         if (backwards && res.end == res.start) {
-            eventTimeline.setPaginationToken(null, true);
+            eventTimeline.setPaginationToken(null, dir);
         }
         return res.end != res.start;
     }).finally(function() {
--- a/lib/client.js
+++ b/lib/client.js
@@ -81,6 +81,10 @@ var OLM_ALGORITHM = "m.olm.v1.curve25519-aes-sha2";
  * to all requests with this client. Useful for application services which require
  * <code>?user_id=</code>.
  *
+ * @param {boolean} [opts.timelineSupport = false] Set to true to enable improved
+ * timeline support ((@link getEventTimeline}). It is disabled by default for
+ * compatibility with older clients - in particular to maintain support for
+ * back-paginating the live timeline after a '/sync' result with a gap.
  */
 function MatrixClient(opts) {
     utils.checkObjectHasKeys(opts, ["baseUrl", "request"]);
@@ -172,6 +176,8 @@ function MatrixClient(opts) {
     this._peekSync = null;
     this._isGuest = false;
     this._ongoingScrollbacks = {};
+
+    this.timelineSupport = Boolean(opts.timelineSupport);
 }
 utils.inherits(MatrixClient, EventEmitter);
 
@@ -2056,6 +2062,145 @@ MatrixClient.prototype.paginateEventContext = function(eventContext, opts) {
         eventContext._paginateRequests[dir] = null;
     });
     eventContext._paginateRequests[dir] = promise;
+
+    return promise;
+};
+
+/**
+ * Get an EventTimeline for the given event
+ *
+ * <p>If the room object already has the given event in its store, the
+ * corresponding timeline will be returned. Otherwise, we make a /context
+ * request, and use it to construct an EventTimeline.
+ *
+ * @param {Room} room  The room to look for the event in
+ * @param {string} eventId  The ID of the event to look for
+ *
+ * @return {module:client.Promise} Resolves:
+ *    {@link module:models/event-timeline~EventTimeline} including the given
+ *    event
+ */
+MatrixClient.prototype.getEventTimeline = function(room, eventId) {
+    // don't allow any timeline support unless it's been enabled.
+    if (!this.timelineSupport) {
+        throw Error("timeline support is disabled. Set the 'timelineSupport'" +
+                    " parameter to true when creating MatrixClient to enable" +
+                    " it.");
+    }
+
+    if (room.getTimelineForEvent(eventId)) {
+        return q(room.getTimelineForEvent(eventId));
+    }
+
+    var path = utils.encodeUri(
+        "/rooms/$roomId/context/$eventId", {
+            $roomId: room.roomId,
+            $eventId: eventId,
+        }
+    );
+
+    // TODO: we should implement a backoff (as per scrollback()) to deal more
+    // nicely with HTTP errors.
+    var self = this;
+    var promise =
+        self._http.authedRequest(undefined, "GET", path
+    ).then(function(res) {
+        if (!res.event) {
+            throw Error("'event' not in '/context' result - homeserver too old?");
+        }
+
+        // by the time the request completes, the event might have ended up in
+        // the timeline.
+        if (room.getTimelineForEvent(eventId)) {
+            return room.getTimelineForEvent(eventId);
+        }
+
+        // we start with the last event, since that's the point at which we
+        // have known state.
+        var events = res.events_before
+            .concat([res.event])
+            .concat(res.events_after);
+        events.reverse();
+        var matrixEvents = utils.map(events, self.getEventMapper());
+
+        var timeline = room.getTimelineForEvent(matrixEvents[0].getId());
+        if (!timeline) {
+            timeline = room.createTimeline();
+            timeline.initialiseState(utils.map(res.state,
+                                               self.getEventMapper()));
+            timeline.getState(false).paginationToken = res.end;
+        }
+        room.addEventsToTimeline(matrixEvents, true, timeline, res.start);
+
+        // there is no guarantee that the event ended up in "timeline" (we
+        // might have switched to a neighbouring timeline) - so check the
+        // room's index again. On the other hand, there's no guarantee the
+        // event ended up anywhere, if it was later redacted, so we just
+        // return the timeline we first thought of.
+        return room.getTimelineForEvent(eventId) || timeline;
+    });
+    return promise;
+};
+
+
+/**
+ * Take an EventTimeline, and back/forward-fill results.
+ *
+ * @param {module:models/event-timeline~EventTimeline} eventTimeline timeline
+ *    object to be updated
+ * @param {Object =} opts
+ * @param {boolean}  [opts.backwards = false]  true to fill backwards,
+ *    false to go forwards
+ * @param {number}   [opts.limit = 30]         number of events to request
+ *
+ * @return {module:client.Promise} Resolves: false if there are no events and
+ *    we reached the end of the timeline; else true. Rejects: Error with an error response
+ */
+MatrixClient.prototype.paginateEventTimeline = function(eventTimeline, opts) {
+    // TODO: we should implement a backoff (as per scrollback()) to deal more
+    // nicely with HTTP errors.
+    opts = opts || {};
+    var backwards = opts.backwards || false;
+
+    var token = eventTimeline.getPaginationToken(backwards);
+    if (!token) {
+        // no more results.
+        return q.reject(new Error("No paginate token"));
+    }
+
+    var dir = backwards ? 'b' : 'f';
+    var pendingRequest = eventTimeline._paginationRequests[dir];
+
+    if (pendingRequest) {
+        // already a request in progress - return the existing promise
+        return pendingRequest;
+    }
+
+    var path = utils.encodeUri(
+        "/rooms/$roomId/messages", {$roomId: eventTimeline.getRoomId()}
+    );
+    var params = {
+        from: token,
+        limit: ('limit' in opts) ? opts.limit : 30,
+        dir: dir
+    };
+
+    var self = this;
+
+    var promise =
+        this._http.authedRequest(undefined, "GET", path, params
+    ).then(function(res) {
+        var token = res.end;
+        console.log("client: completed paginate; backwards=" + backwards +
+                    "; token=" + token);
+        var room = self.getRoom(eventTimeline.getRoomId());
+        var matrixEvents = utils.map(res.chunk, self.getEventMapper());
+        room.addEventsToTimeline(matrixEvents, backwards, eventTimeline, token);
+        return res.end != res.start;
+    }).finally(function() {
+        eventTimeline._paginationRequests[dir] = null;
+    });
+    eventTimeline._paginationRequests[dir] = promise;
 
     return promise;
 };
--- a/lib/models/event-timeline.js
+++ b/lib/models/event-timeline.js
@@ -1,0 +1,252 @@
+"use strict";
+
+/**
+ * @module models/event-timeline
+ */
+
+var RoomState = require("./room-state");
+
+/**
+ * Construct a new EventTimeline
+ *
+ * <p>An EventTimeline represents a contiguous sequence of events in a room.
+ *
+ * <p>As well as keeping track of the events themselves, it stores the state of
+ * the room at the beginning and end of the timeline, and pagination tokens for
+ * going backwards and forwards in the timeline.
+ *
+ * <p>In order that clients can meaningfully maintain an index into a timeline, we
+ * track a 'baseIndex'. This starts at zero, but is incremented when events are
+ * prepended to the timeline. The index of an event relative to baseIndex
+ * therefore remains constant.
+ *
+ * <p>Once a timeline joins up with its neighbour, we link them together into a
+ * doubly-linked list.
+ *
+ * @param {string} roomId    the ID of the room where this timeline came from
+ * @constructor
+ */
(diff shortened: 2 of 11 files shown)


lib/client.js
```diff
@@ -2060,6 +2066,145 @@ MatrixClient.prototype.paginateEventContext = function(eventContext, opts) {
     return promise;
 };
 
+/**
+ * Get an EventTimeline for the given event
+ *
+ * <p>If the room object already has the given event in its store, the
+ * corresponding timeline will be returned. Otherwise, we make a /context
```
Kegsay: Nit: Using 'we' for public-facing docs seems weird. I would say "Otherwise, a /context request will be made and used to construct an EventTimeline".

lib/client.js
```diff
@@ -2060,6 +2066,145 @@ MatrixClient.prototype.paginateEventContext = function(eventContext, opts) {
...
+ *
+ * @param {module:models/event-timeline~EventTimeline} eventTimeline timeline
+ *    object to be updated
+ * @param {Object =} opts
+ * @param {boolean}  [opts.backwards = false]  true to fill backwards,
+ *    false to go forwards
+ * @param {number}   [opts.limit = 30]         number of events to request
+ *
+ * @return {module:client.Promise} Resolves: false if there are no events and
+ *    we reached the end of the timeline; else true. Rejects: Error with an error response
```
Kegsay: I would clarify what you mean by "end" - either extremity presumably?

lib/client.js
```diff
@@ -2060,6 +2066,145 @@ MatrixClient.prototype.paginateEventContext = function(eventContext, opts) {
...
+    );
+
+    // TODO: we should implement a backoff (as per scrollback()) to deal more
+    // nicely with HTTP errors.
+    var self = this;
+    var promise =
+        self._http.authedRequest(undefined, "GET", path
+    ).then(function(res) {
+        if (!res.event) {
+            throw Error("'event' not in '/context' result - homeserver too old?");
```
Kegsay: Please always `new` your constructors. This will still work for `Error` objects due to defined factory functions but we're generally consistent using `new` everywhere.

lib/client.js
```diff
@@ -2060,6 +2066,145 @@ MatrixClient.prototype.paginateEventContext = function(eventContext, opts) {
...
+    var dir = backwards ? 'b' : 'f';
+    var pendingRequest = eventTimeline._paginationRequests[dir];
+
+    if (pendingRequest) {
+        // already a request in progress - return the existing promise
+        return pendingRequest;
+    }
+
+    var path = utils.encodeUri(
+        "/rooms/$roomId/messages", {$roomId: eventTimeline.getRoomId()}
```
Kegsay: `/messages` allows you to go forwards?

lib/client.js
```diff
@@ -2060,6 +2066,145 @@ MatrixClient.prototype.paginateEventContext = function(eventContext, opts) {
...
+
+    var self = this;
+
+    var promise =
+        this._http.authedRequest(undefined, "GET", path, params
+    ).then(function(res) {
+        var token = res.end;
+        console.log("client: completed paginate; backwards=" + backwards +
+                    "; token=" + token);
+        var room = self.getRoom(eventTimeline.getRoomId());
```
Kegsay: We don't do any safety checks before this to see if we do have a `Room` object for this room ID. This should be null guarded.

lib/models/event-timeline.js
```diff
@@ -0,0 +1,252 @@
...
+ * <p>As well as keeping track of the events themselves, it stores the state of
+ * the room at the beginning and end of the timeline, and pagination tokens for
+ * going backwards and forwards in the timeline.
+ *
+ * <p>In order that clients can meaningfully maintain an index into a timeline, we
+ * track a 'baseIndex'. This starts at zero, but is incremented when events are
+ * prepended to the timeline. The index of an event relative to baseIndex
+ * therefore remains constant.
+ *
+ * <p>Once a timeline joins up with its neighbour, we link them together into a
```
Kegsay: "we" references (and `:18`)

lib/models/event-timeline.js
```diff
@@ -0,0 +1,252 @@
...
+ * Initialise the start and end state with the given events
+ *
+ * <p>This can only be called before any events are added.
+ *
+ * @param {MatrixEvent[]} stateEvents list of state events to intialise the
+ * state with.
+ */
+EventTimeline.prototype.initialiseState = function(stateEvents) {
+    if (this._events.length > 0) {
+        throw new Error("Cannot initialise state after events are added");
```
Kegsay: Please document the `@throws`

lib/models/event-timeline.js
```diff
@@ -0,0 +1,252 @@
...
+    // this is used by client.js
+    this._paginationRequests = {'b': null, 'f': null};
+}
+
+/**
+ * Initialise the start and end state with the given events
+ *
+ * <p>This can only be called before any events are added.
+ *
+ * @param {MatrixEvent[]} stateEvents list of state events to intialise the
```
Kegsay: Typo: `intialise`

lib/models/event-timeline.js
```diff
@@ -0,0 +1,252 @@
...
+ * state with.
+ */
+EventTimeline.prototype.initialiseState = function(stateEvents) {
+    if (this._events.length > 0) {
+        throw new Error("Cannot initialise state after events are added");
+    }
+
+    // do we need to copy here? sync thinks we do but I can't see why
+    this._startState.setStateEvents(stateEvents);
+    this._endState.setStateEvents(stateEvents);
```
Kegsay: You need to deep copy here because otherwise these `_startState` and `_endState` will be backed by the *same* object and the *same* events, so mutations in the object/random state events will affect *both* old and current state. This can introduce subtle bugs.

lib/models/event-timeline.js
```diff
@@ -0,0 +1,252 @@
...
+ * Get the ID of the room for this timeline
+ * @return {string} room ID
+ */
+EventTimeline.prototype.getRoomId = function() {
+    return this._roomId;
+};
+
+/**
+ * Get the base index
+ *
```
Kegsay: Please explain this more. You go into some nice detail on the JSDoc constructor and that is sorely missing here.

lib/models/event-timeline.js
```diff
@@ -0,0 +1,252 @@
...
+ * @param {EventTimeline} neighbour previous/following timeline
+ *
+ * @param {boolean} before true to set the previous timeline; false to set
+ * following one.
+ */
+EventTimeline.prototype.setNeighbouringTimeline = function(neighbour, before) {
+    if (this.getNeighbouringTimeline(before)) {
+        throw new Error("timeline already has a neighbouring timeline - " +
+                        "cannot reset neighbour");
```
Kegsay: Please document the `@throws`

lib/models/room.js
```diff
@@ -56,18 +72,24 @@ function synthesizeReceipt(userId, event, receiptType) {
...
+ * @param {boolean} [opts.timelineSupport = false] Set to true to enable improved
+ * timeline support.
+ *
  * @prop {string} roomId The ID of this room.
  * @prop {string} name The human-readable display name for this room.
- * @prop {Array<MatrixEvent>} timeline The ordered list of message events for
- * this room, with the oldest event at index 0.
+ * @prop {Array<MatrixEvent>} timeline The live event timeline for this room,
+ * with the oldest event at index 0. Present for backwards compatibility -
+ * prefer getLiveTimeline().getEvents().
```
Kegsay: You can do *horrible* things to migrate people over for the record.

```javascript
function Room() { // constructor
  var self = this;
  Object.defineProperty(this, 'timeline', {
    get: function () {
      return self.getLiveTimeline().getEvents();
    }
  });
}
```

Any time people refer to `room.timeline` the `get` function will be invoked. If this provides equivalent backwards-compat, I'm happy to nuke the `timelineSupport` gate flag.

lib/models/room.js
```diff
@@ -56,18 +72,24 @@ function synthesizeReceipt(userId, event, receiptType) {
...
  * @prop {object} accountData Dict of per-room account_data events; the keys are the
  * event type and the values are the events.
  * @prop {RoomState} oldState The state of the room at the time of the oldest
- * event in the timeline.
+ * event in the live timeline. Present for backwards compatibility -
+ * prefer getLiveTimeline().getState(true).
  * @prop {RoomState} currentState The state of the room at the time of the
- * newest event in the timeline.
+ * newest event in the timeline. Present for backwards compatibility -
+ * prefer getLiveTimeline().getState(false).
```
Kegsay: Ditto https://github.com/matrix-org/matrix-js-sdk/pull/63#discussion_r49860084

lib/models/room.js
```diff
@@ -214,43 +292,195 @@ Room.prototype.getAvatarUrl = function(baseUrl, width, height, resizeMethod,
  };
 
 /**
- * Add some events to this room's timeline. Will fire "Room.timeline" for
- * each event added.
+ * Create a new timeline for this room
+ *
+ * @param {Array<MatrixEvent>} stateBeforeEvent state of the room before this event
```
Kegsay: Lying param, this doesn't exist!

lib/models/room.js
```diff
@@ -214,43 +292,195 @@ Room.prototype.getAvatarUrl = function(baseUrl, width, height, resizeMethod,
...
+ */
+Room.prototype.createTimeline = function() {
+    if (!this._timelineSupport) {
+        throw Error("timeline support is disabled. Set the 'timelineSupport'" +
+                    " parameter to true when creating MatrixClient to enable" +
+                    " it.");
+    }
+
+    var timeline = new EventTimeline(this.roomId);
+    this._timelines.push(timeline);
```
Kegsay: This side-effect is a little surprising given you use the word `create` instead of `add`. Can you clarify in the method docs or rename this method please?

lib/models/room.js
```diff
@@ -214,43 +292,195 @@ Room.prototype.getAvatarUrl = function(baseUrl, width, height, resizeMethod,
...
+ *
+ * @param {MatrixEvent} event           event to add
+ * @param {EventTimeline} timeline      timeline we think we should add to
+ * @param {boolean} toStartOfTimeline   true if we're adding to the start of
+ *    the timeline
+ * @return {?EventTimeline} the timeline with the event already in, or null if
+ *    none
+ * @private
+ */
+Room.prototype._checkExistingTimeline = function(event, timeline,
```
Kegsay: Vague name. Check *what* in the existing timeline? Given we call this function in precisely 1 place (and its internal), I would appreciate some clearer names so I can tell at a glance what this is doing. Perhaps `getTimelineForEvent(...): EventTimeline` ?

lib/models/room.js
```diff
@@ -214,43 +292,195 @@ Room.prototype.getAvatarUrl = function(baseUrl, width, height, resizeMethod,
...
+        return;
+    }
+
+    var updateToken = false;
+    for (var i = 0; i < events.length; i++) {
+        var existingTimeline = this._checkExistingTimeline(events[i], timeline,
+                                                           toStartOfTimeline);
+        if (existingTimeline) {
+            // switch to the other timeline
+            timeline = existingTimeline;
```
Kegsay: Took me ages to figure out why, when we've found an existing timeline, we're doing this. We're doing this so we can then probe the *next* event with this timeline..? I don't really get it. You're using the same variable `timeline` in so many places (input arg, clobbered with `existingTimeline`, adding events to it, setting tokens with it, etc) that I can't tell at any one point whether what you're doing makes sense or not. Please can you refrain from reusing `timeline` so much, as I find it hard to figure out exactly what that is supposed to represent. A few more descriptive vars would be great.

lib/models/room.js
```diff
@@ -214,43 +292,195 @@ Room.prototype.getAvatarUrl = function(baseUrl, width, height, resizeMethod,
...
+                 " - joining timeline " + timeline + " to " +
+                 existingTimeline);
+    timeline.setNeighbouringTimeline(existingTimeline, toStartOfTimeline);
+    existingTimeline.setNeighbouringTimeline(timeline, !toStartOfTimeline);
+    return existingTimeline;
+};
+
+/**
+ * Check for redactions, and otherwise add event to the given timeline. Assumes
+ * we have already checked we don't lnow about this event.
```
Kegsay: Typo: `lnow`

lib/models/room.js
```diff
@@ -386,32 +575,41 @@ Room.prototype.addEvents = function(events, duplicateStrategy) {
...
 
-    for (var i = 0; i < event_ids.length; ++i) {
-        var removed = removeEventWithId(this.timeline, event_ids[i]);
-        if (removed) {
-            this.emit("Room.timeline", removed, this, undefined, true);
-        }
+/**
+ * Removes a single event from this room.
+ * @param {String} eventId  The id of the event to remove
+ * @return {?MatrixEvent} the removed event, or null if none
```
Kegsay: What does `or null if none` mean? Do you mean "or null if no event was removed"? We're returning a `boolean` on `:601` though.

Kegsay:
This is a really big PR, so I'm not confident that I've been able to assess it all as thoroughly as normal. From the looks of it, the exposed public API and the underlying newly added data models LGTM. However, I'm concerned with some of the `Room` logic, specifically management of multiple timelines (adding events, joining timelines together, etc). I'm finding it hard to understand it, so I would like a bit more hand-holding on what you're doing there please. Otherwise, this is looking good.

lib/client.js
richvdh: it does.

lib/models/event-timeline.js
richvdh: Sorry, I don't get it. Why would the events change after they are added?

lib/models/room.js
richvdh: Sadly, it doesn't quite provide exact backwards-compat.

The problem is as follows, with an existing client which doesn't know about timelines:

* client does a sync which returns event A
* client goes to sleep for a while; meanwhile events B and C happen
* client does another sync, which has a gap since the previous event, and contains event D. The live timeline is therefore reset and contains only event D.
* client tries to back-paginate to see what happened before event D; it gets events A, B, and C back from the server.
* B and C are preprended to the live timeline, which is fine, but the room already knows about A, so the old timeline is joined to the new
* From the client's point of view (given it knows nothing about multiple timelines), it can't see event A (or anything which happened before it).

The solution I went with was just to make the room forget all existing events/timelines when it got a gappy sync if `timelineSupport` was unset. I guess an alternative would be to make `timeline` a property whose getter builds an array from the live timeline and any previous timelines which have been joined up. But I'd be worried about people assuming `room.timeline` was a cheap operation.

lib/client.js
richvdh: done e0ddd65

lib/client.js
richvdh: Done e0ddd65

lib/client.js
richvdh: yup, sorry.

Done. e0ddd65

lib/client.js
richvdh: Moved the room lookup out of the promise handler and added a null check.

Done e0ddd65

lib/models/event-timeline.js
richvdh: Done e0ddd65

lib/models/event-timeline.js
richvdh: Done e0ddd65

lib/models/event-timeline.js
richvdh: Done e0ddd65

lib/models/event-timeline.js
richvdh: Added some words.

Done e0ddd65

lib/models/event-timeline.js
richvdh: done e0ddd65

lib/models/room.js
richvdh: done e0ddd65

lib/models/room.js
richvdh: ok, I renamed it to `addTimeline`

Done e0ddd65

lib/models/room.js
richvdh: The fact that I can't think of a good name for it is probably a clear indication that it's poorly designed. Maybe it would be better if I just inlined it?

For now I've gone with `_checkForNewEventInExistingTimeline`, but that doesn't reflect the fact that it sometimes has the side-effect of joining the timelines.

lib/models/room.js
richvdh: This is obviously the gnarly bit of this change, and you won't be surprised to learn that I ended up fiddling with it a fair bit.

I've added a wordy comment which I hope makes it much clearer wtf is going on here. In the process, I've realised that I missed an edge case, which for now I have marked with an XXX.

In terms of the overuse of `timeline` - yes I suppose it *is* being overused, but here there is only one "timeline" of interest, and that is the timeline we are currently adding events to. I think it makes sense to reuse, and clobber, the input variable here.

lib/models/room.js
richvdh: Done e0ddd65

lib/models/room.js
richvdh: clarified, and fix the boolean typo.

richvdh:
I hope my comments at https://github.com/matrix-org/matrix-js-sdk/commit/e0ddd659226bc50ca8fe8e55654fbdd8c42269f2#diff-c1b10daf1b0e9cc493394eff6a845497R343 might help make it a bit clearer what's going on. I know it's big, though :(

lib/models/event-timeline.js
Kegsay: Various flags and calculations can be done on events after they have been received, e.g. the `forwardLooking` flag. These values will be wrong if they are backed by the same object because they will be clobber.

lib/models/room.js
Kegsay: Fair enough!

lib/models/room.js
Kegsay: If it's called in one place, I would be tempted to in-line with some comments explaining what you're doing (particularly with the side-effects). We can chop the function up later on boundaries which make more sense.

spec/integ/matrix-client-event-timeline.spec.js
```diff
@@ -0,0 +1,558 @@
...
+            }).catch(exceptFail).done(done);
+
+            httpBackend.flush().catch(exceptFail);
+        });
+
+    });
+});
+
+// make the test fail, with the given exception
+function exceptFail(error) {
```
Kegsay: I think this should this be `expectFail`?

spec/unit/event-timeline.spec.js
```diff
@@ -0,0 +1,331 @@
...
+            expect(function() { timeline.initialiseState(state); }).not.toThrow();
+            timeline.addEvent(event, false);
+            expect(function() { timeline.initialiseState(state); }).toThrow();
+        });
+    });
+
+    describe("paginationTokens", function() {
+        it("pagination tokens should start undefined", function() {
+            expect(timeline.getPaginationToken(true)).toBe(undefined);
+            expect(timeline.getPaginationToken(false)).toBe(undefined);
```
Kegsay: I'd prefer if they started as `null` unless there is a good reason for them not to be. We typically initialise "no value" vars to `null` in the rest of the code, so if you see `undefined` errors it's most likely a typo.

lib/models/event-timeline.js
```diff
@@ -0,0 +1,260 @@
...
+EventTimeline.prototype.getRoomId = function() {
+    return this._roomId;
+};
+
+/**
+ * Get the base index.
+ *
+ * <p>This is an index which is incremented when events are prepended to the
+ * timeline. An individual event therefore stays at the same index in the array
+ * relative to the base index.
```
Kegsay: Worth stressing that this means events can have -ve indexes?

lib/models/event-timeline.js
```diff
@@ -0,0 +1,260 @@
...
+};
+
+/**
+ * Get the room state at the start/end of the timeline
+ *
+ * @param {boolean} start true to get the state at the start of the timeline;
+ * false to get the state at the end of the timeline.
+ * @return {RoomState} state at the start/end of the timeline
+ */
+EventTimeline.prototype.getState = function(start) {
```
Kegsay: What I'm going to say is going to be a pain but I think it will be for the best.

I don't like the boolean switching for start/end and would prefer `getStartState()` and `getEndState()`. Why? Simply because when you're reading code which uses this (for me it was when I was reading the tests) it is not immediately obvious what `getState(true)` refers to. I then need to remember the line I'm on, hunt the method definition and check what the `boolean` means. This makes me sad. Legibility of code can be improved by using more descriptive function names IMO. I appreciate this is nitty and a PITA :(

Kegsay: Ditto:
 - `getPaginationToken` -> `getStartPaginationToken`, `getEndPaginationToken`
 - `setPaginationToken` -> `setStartPaginationToken`, `setEndPaginationToken`
 - `getNeighbouringTimeline` -> `getPrevTimeline`, `getNextTimeline`

I'd be equally happy with a way of describing ends as args, e.g. `getPaginationToken("f")` and `getPaginationToken("b")` rather than `true/false`.

Kegsay: I'm open to pushback here if this is a real PITA.

lib/sync.js
```diff
@@ -209,6 +226,12 @@ SyncApi.prototype.peek = function(roomId) {
         // roll backwards to diverge old state:
         peekRoom.addEventsToTimeline(messages.reverse(), true);
 
+        // set the pagination token. Make sure this happens after adding
+        // events to the timeline, otherwise it will get reset.
+        if (response.messages.start) {
+            peekRoom.oldState.paginationToken = response.messages.start;
```
Kegsay: We can't do this here; it HAS to be before we add the events. See https://github.com/matrix-org/matrix-js-sdk/commit/bd4d8433abe8cae32f1b5b4ab6e41970de307f9b#diff-d1172eb3f936fc93dbd0b89cdf357205R459

lib/models/room.js
```diff
@@ -119,11 +138,71 @@ function Room(roomId, opts) {
...
+    if (!this._timelineSupport) {
+        // if timeline support is disabled, forget about the old timelines
+        newTimeline = new EventTimeline(this.roomId);
+        this._timelines = [newTimeline];
+        this._eventIdToTimeline = {};
+    } else {
+        newTimeline = this.addTimeline();
+    }
+
+    if (this._liveTimeline) {
```
Kegsay: Clarity that this is expected to be `true` iff `resetLiveTimeout()` is called >1 times would be good (e.g. `//existing timeline from before`). This looks like a bug otherwise.

lib/models/room.js
```diff
@@ -119,11 +138,71 @@ function Room(roomId, opts) {
...
+    this.oldState = this._liveTimeline.getState(true);
+    this.currentState = this._liveTimeline.getState(false);
+};
+
+/**
+ * Get the timeline which contains the given event, if any
+ *
+ * @param {string} eventId  event ID to look for
+ * @return {?module:models/event-timeline~EventTimeline} timeline containing
+ * the given event, or undefined if unknown
```
Kegsay: Please return `null` rather than `undefined`. The JS SDK consistently returns `null` to represent "no value exists". `undefined` shouldn't be a part of the public API. By doing this consistently we help devs because if they get an `undefined` error they know it is most likely due to a typo and not due to lack of a null guard.

lib/models/room.js
```diff
@@ -237,43 +316,226 @@ Room.prototype.getAvatarUrl = function(baseUrl, width, height, resizeMethod,
  };
 
 /**
- * Add some events to this room's timeline. Will fire "Room.timeline" for
- * each event added.
+ * Add a new timeline to this room
+ *
+ * @return {module:models/event-timeline~EventTimeline} newly-created timeline
+ */
+Room.prototype.addTimeline = function() {
```
Kegsay: Feels weird that this doesn't accept a timeline, but I guess there is a good reason for it? When would the caller know enough to know that a new timeline is required yet not be able to make one themselves?

lib/models/room.js
```diff
@@ -237,43 +316,226 @@ Room.prototype.getAvatarUrl = function(baseUrl, width, height, resizeMethod,
...
+    // 4. We now need to append R and S to timeline2. So finally, we have:
+    //
+    //    timeline1      timeline2
+    //     [M, N] <----> [P, R, S]
+    //
+    // The important thing to note in the above is that we switched the
+    // timeline we were adding events to, after we found a neighbour.
+
+    // XXX: TODO: but timeline2 might already be joined to timeline3: we need
+    // to skip all the way to the end of the linked list!
```
Kegsay: Wonderful prose, thank you! As for the `XXX TODO`: This is presumably unavoidable that we need to follow the linked list and scan each element in each list? The comment suggests that there may be a way to avoid doing this? It might be worth stating your proposal even if you don't implement it now.

lib/models/room.js
```diff
@@ -237,43 +316,226 @@ Room.prototype.getAvatarUrl = function(baseUrl, width, height, resizeMethod,
...
+
+    var updateToken = false;
+    for (var i = 0; i < events.length; i++) {
+        var existingTimeline = this._checkForNewEventInExistingTimeline(
+            events[i], timeline, toStartOfTimeline);
+
+        if (existingTimeline) {
+            timeline = existingTimeline;
+            updateToken = false;
```
Kegsay: Based on your earlier prose, I am guessing that `timeline = existingTimeline` is switching the current list that we're inserting elements into. This is fine. I don't understand this next line though. Why would we set `updateToken = false`? It's not like we're touching that old timeline we've switched from anymore. Why wouldn't we always update the timeline with the token?

lib/models/room.js
```diff
@@ -237,43 +316,226 @@ Room.prototype.getAvatarUrl = function(baseUrl, width, height, resizeMethod,
...
+/**
+ * Check for redactions, and otherwise add event to the given timeline. Assumes
+ * we have already checked we don't know about this event.
+ *
+ * Will fire "Room.timeline" for each event added.
+ *
+ * @param {MatrixEvent} event
+ * @param {EventTimeline} timeline
+ * @param {boolean} toStartOfTimeline
+ * @param {boolean} spliceBeforeLocalEcho
```
Kegsay: A short summary of this option would be good.

lib/models/room.js
```diff
@@ -284,54 +546,12 @@ Room.prototype.addEventsToTimeline = function(events, toStartOfTimeline) {
...
-                        insertIndex = j + 1;
-                        break;
-                    }
-                }
-                this.timeline.splice(insertIndex, 0, events[i]); // insert element
-            }
-            else {
-                this.timeline.push(events[i]);
-            }
+        if (!this._eventIdToTimeline[events[i].getId()]) {
```
Kegsay: Presumably this dictionary is going to grow to be an unbounded map of all the event IDs we've ever seen. How would we go about purging this?

lib/models/room.js
```diff
@@ -237,43 +316,226 @@ Room.prototype.getAvatarUrl = function(baseUrl, width, height, resizeMethod,
...
+    if (updateToken) {
+        timeline.setPaginationToken(paginationToken, toStartOfTimeline);
+    }
+};
+
+/**
+ * Check if this event is already in a timeline, and join up the timelines if
+ * necessary
+ *
+ * @param {MatrixEvent} event           event to add
```
Kegsay: We aren't ever inserting this event into the timeline. We're using this event as an anchor point to determine which timeline should be glued with `timeline`. Perhaps `anchorEvent` would be a better name?

lib/models/room.js
```diff
@@ -237,43 +316,226 @@ Room.prototype.getAvatarUrl = function(baseUrl, width, height, resizeMethod,
...
+ * @param {EventTimeline} timeline      timeline we think we should add to
+ * @param {boolean} toStartOfTimeline   true if we're adding to the start of
```
Kegsay: This determines which end of the timeline we *may* be gluing to another timeline. It might be better to make this a bit clearer

Kegsay:
Overall, the comments have greatly improved my ability to grok this code. I am more confident than before on this now. I've left a bunch of random thought-streamy comments; please do not feel like they all need to be done, most are just trying to query bits I think feel strange or don't understand. Once I've got some answers on these, LGTM!

spec/integ/matrix-client-event-timeline.spec.js
richvdh: well, not really. `except` as in 'exception'. Suggestions for better names appreciated. This little idiom turned out to be really useful in figuring out why failures were happening, instead of just getting "it didn't finish in 5 seconds".

lib/models/room.js
richvdh: OK; have inlined it.

spec/unit/event-timeline.spec.js
richvdh: done

lib/models/event-timeline.js
richvdh: done

lib/sync.js
richvdh: OK; have hacked around this.

lib/models/room.js
richvdh: done

lib/models/room.js
richvdh: slightly unconvinced by this. the thin wrapper over the `[]` operator was somewhat deliberate, and having the same semantics as `[]` makes sense to me.

Done, nevertheless.

lib/models/room.js
richvdh: Never, but I see no reason to duplicate the `new EventTimeline` call to every caller (with the potential of getting the roomId on that new timeline wrong). Conceptually, timelines belong to rooms; the room should be responsible for making sure new ones are created correctly.

You objected to it being called `createTimeline`, and suggested renaming it...

lib/models/room.js
richvdh: I've thought about it a bit harder and decided that we're already covering all the plausible cases. I've expanded the comments further.

lib/models/room.js
richvdh: I've added some words about this, too.

lib/models/room.js
richvdh: done

lib/models/room.js
richvdh: well, this isn't a new problem. But I guess we could expire timelines which haven't been used for a while; we would remove entries from _eventIdToTimeline when we did so.

lib/models/room.js
richvdh: this has been inlined, so has gone away.

lib/models/room.js
richvdh: ditto

lib/client.js
richvdh: done

lib/models/room.js
richvdh: done

lib/models/event-timeline.js
richvdh: done.

lib/models/room.js
richvdh: done?

lib/models/room.js
richvdh: done

lib/sync.js
richvdh: done

lib/models/room.js
richvdh: done

lib/models/room.js
richvdh: done

lib/models/room.js
richvdh: done

lib/models/room.js
richvdh: (it has gone away).

done.

lib/models/event-timeline.js
richvdh: I'm keen to keep this as an argument, rather than having separate functions, as there are a few places in the higher-level logic which are agnostic as to which end they are working from; I'd prefer not to duplicate lots of code or litter it with ? : expressions.

I've changed these methods to take a string argument, and defined some constants to help pass the right values.

done.

richvdh:
@Kegsay: ptal.

spec/integ/matrix-client-event-timeline.spec.js
Kegsay: Oh I see! That does sound useful. Perhaps `failTest(error) ` would be clearer? I'm also happy for you to put that in a test util file, I would definitely use it.

lib/models/event-timeline.js
Kegsay: !goldstar 

Thank you :)

lib/models/room.js
Kegsay: I guess I'm just confused because we're both creating (`createTimeline`) and inserting a timeline (`addTimeline`).

Kegsay:
LGTM

spec/integ/matrix-client-event-timeline.spec.js
richvdh: Done; moved to test-utils, and added some comments.

PR #64, opened 2016-01-15
TimelineWindow object

richvdh:
A handy thing for tracking a window into a room timeline, for use with non-contiguous timelines (#63).

Could really do with some unit tests... sorry.

### DIFF
--- a/lib/timeline-window.js
+++ b/lib/timeline-window.js
@@ -64,6 +64,9 @@ function TimelineWindow(client, room, opts) {
     this._client = client;
     this._room = room;
 
+    // these will be TimelineIndex objects; they delineate the 'start' and
+    // 'end' of the window.
+    //
     // _start.index is inclusive; _end.index is exclusive.
     this._start = null;
     this._end = null;
@@ -93,7 +96,7 @@ TimelineWindow.prototype.load = function(initialEventId, initialWindowSize) {
                 // make sure that our window includes the event
                 for (var i = 0; i < tl.getEvents().length; i++) {
                     if (tl.getEvents()[i].getId() == initialEventId) {
-                        return [tl, i];
+                        return {timeline: tl, index: i};
                     }
                 }
                 throw new Error("getEventTimeline result didn't include requested event");
@@ -103,11 +106,11 @@ TimelineWindow.prototype.load = function(initialEventId, initialWindowSize) {
 
         // start with the most recent events
         var tl = this._room.getLiveTimeline();
-        prom = q([tl, tl.getEvents().length]);
+        prom = q({timeline: tl, index: tl.getEvents().length});
     }
 
     prom = prom.then(function(v) {
-        var tl = v[0], eventIndex = v[1];
+        var tl = v.timeline, eventIndex = v.index;
 
         var endIndex = Math.min(tl.getEvents().length,
                                 eventIndex + initialWindowSize / 2);
@@ -145,8 +148,8 @@ TimelineWindow.prototype.canPaginate = function(direction) {
         throw new Error("Invalid direction '" + direction + "'");
     }
 
-    return tl.timeline.getNeighbouringTimeline(direction) ||
-        tl.timeline.getPaginationToken(direction);
+    return Boolean(tl.timeline.getNeighbouringTimeline(direction) ||
+                   tl.timeline.getPaginationToken(direction));
 };
 
 /**
@@ -159,8 +162,11 @@ TimelineWindow.prototype.canPaginate = function(direction) {
  *    number are immediately available, then we return immediately rather than
  *    making an API call.
  *
- * @param {boolean} [makeRequest = true]  whether we should make API calls to
- *    fetch further events if we don't have any.
+ * @param {boolean} [makeRequest = true] whether we should make API calls to
+ *    fetch further events if we don't have any at all. (This has no effect if
+ *    the room already knows about additional events in the relevant direction,
+ *    even if there are fewer than 'size' of them, as we will just return those
+ *    we already know about.)
  *
  * @return {module:client.Promise} Resolves to a boolean which is true if more events
  *    were successfully retrieved.
@@ -200,8 +206,9 @@ TimelineWindow.prototype.paginate = function(direction, size, makeRequest) {
         debuglog("TimelineWindow: increased cap by " + count +
                  " (now " + this._eventCount + ")");
         // remove some events from the other end, if necessary
-        if (this._eventCount > this._windowLimit) {
-            this._unpaginate(direction != EventTimeline.BACKWARDS);
+        var excess = this._eventCount - this._windowLimit;
+        if (excess > 0) {
+            this._unpaginate(excess, direction != EventTimeline.BACKWARDS);
         }
         return q(true);
     }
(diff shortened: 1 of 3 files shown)


lib/timeline-window.js
```diff
@@ -0,0 +1,381 @@
...
+};
+
+/**
+ * Try move the index backwards, or into the neighbouring timeline
+ *
+ * @param {number} delta  number of events to retreat by
+ * @return {number} number of events successfully retreated by
+ */
+TimelineIndex.prototype.retreat = function(delta) {
+    return -this.advance(-delta);
```
Kegsay: Spurious `-`? If this is supposed to return negative numbers then please:
 - Explicitly state this in the `@return` docs (rather than "the number of events")
 - Use `* -1` instead of `-`. The reason for this is to make it more obvious in the code that an inversion is occurring - I almost missed seeing the `-` because I did not expect to see it there.

richvdh: It's supposed to return positive numbers, but `advance` returns negative numbers in the case of a negative delta.

I've replaced the unary `-` with `* -1`.

lib/timeline-window.js
```diff
@@ -0,0 +1,392 @@
...
+    } else {
+        debuglog("TimelineWindow: initialising with live timeline");
+
+        // start with the most recent events
+        var tl = this._room.getLiveTimeline();
+        prom = q([tl, tl.getEvents().length]);
+    }
+
+    prom = prom.then(function(v) {
+        var tl = v[0], eventIndex = v[1];
```
Kegsay: Better var names please. I would prefer it if you passed the timeline and and the index as an object with descriptive key names rather than an array with magical ordering:
```javascript
{
  timeline: ...
  index: 5
}
```

lib/timeline-window.js
```diff
@@ -0,0 +1,392 @@
...
+ * @constructor
+ */
+function TimelineWindow(client, room, opts) {
+    opts = opts || {};
+    this._client = client;
+    this._room = room;
+
+    // _start.index is inclusive; _end.index is exclusive.
+    this._start = null;
+    this._end = null;
```
Kegsay: In-line comments for the type (`TimelineIndex`) would be appreciated.

lib/timeline-window.js
```diff
@@ -0,0 +1,392 @@
...
+        if (tl.index > tl.minIndex()) { return true; }
+    } else if (direction == EventTimeline.FORWARDS) {
+        tl = this._end;
+        if (tl.index < tl.maxIndex()) { return true; }
+    } else {
+        throw new Error("Invalid direction '" + direction + "'");
+    }
+
+    return tl.timeline.getNeighbouringTimeline(direction) ||
+        tl.timeline.getPaginationToken(direction);
```
Kegsay: This violates your return type. Please wrap with `Boolean()`. `||` doesn't coerce to a boolean for you, but returns the first truthy object (which is useful for setting default values).

lib/timeline-window.js
```diff
@@ -0,0 +1,392 @@
...
+ *
+ * @param {string} direction   EventTimeline.BACKWARDS to extend the window
+ *    backwards (towards older events); EventTimeline.FORWARDS to go forwards.
+ *
+ * @param {number} size   number of events to try to extend by. If fewer than this
+ *    number are immediately available, then we return immediately rather than
+ *    making an API call.
+ *
+ * @param {boolean} [makeRequest = true]  whether we should make API calls to
+ *    fetch further events if we don't have any.
```
Kegsay: Some clarification on how this interacts with `size` would be great. E.g:
 - `size=7`, timeline has 5, `makeRequest=true` = ??? (the docs on `size` suggest no request will be made since "fewer than this number are immediately available", but I've explicitly set `makeRequest=true`. Which one wins?

lib/timeline-window.js
```diff
@@ -0,0 +1,392 @@
...
+ *
+ * @param {boolean} startOfTimeline if events should be removed from the start
+ *     of the timeline.
+ *
+ * @private
+ */
+TimelineWindow.prototype._unpaginate = function(startOfTimeline) {
+    var tl = startOfTimeline ? this._start : this._end;
+
+    var delta = this._eventCount - this._windowLimit;
```
Kegsay: I would prefer if we hoisted this to be the caller's responsibility because we are actually doing this check in https://github.com/matrix-org/matrix-js-sdk/pull/64/files#diff-e0497ffe12218015b7ef5f92abdc85a2R203 - This also makes the function signature more pleasing: `_unpaginate(num, startOfTimeline)`

lib/timeline-window.js
```diff
@@ -0,0 +1,392 @@
...
+        var startIndex = 0, endIndex = events.length;
+        if (timeline === this._start.timeline) {
+            startIndex = this._start.index + timeline.getBaseIndex();
+        }
+        if (timeline === this._end.timeline) {
+            endIndex = this._end.index + timeline.getBaseIndex();
+        }
+
+        for (var i = startIndex; i < endIndex; i++) {
+            result.push(events[i]);
```
Kegsay: Both these `for` loops need more comments. I can follow this, but this reeks of subtle indexing errors because we aren't doing `i=0; i<events.length; i++` when pushing these events. I need you to convince me that `startIndex` and `endIndex` are going to be sane in the context of `events`.

lib/timeline-window.js
```diff
@@ -0,0 +1,392 @@
...
+    // the indexes are relative to BaseIndex, so could well be negative.
+    this.index = index;
+}
+
+/**
+ * @return {number} the minimum possible value for the index in the current
+ *    timeline
+ */
+TimelineIndex.prototype.minIndex = function() {
+    return -this.timeline.getBaseIndex();
```
Kegsay: `* -1` to make it clearer when reading that this is being inverted please.

lib/timeline-window.js
```diff
@@ -0,0 +1,392 @@
...
+        x = Math.max(delta, this.minIndex() - this.index);
+        if (x < 0) {
+            this.index += x;
+            return x;
+        }
+    } else {
+        x = Math.min(delta, this.maxIndex() - this.index);
+        if (x > 0) {
+            this.index += x;
+            return x;
```
Kegsay: Please explain what you're doing, and I would prefer a most useful var name than `x`. I'm particularly interested in the case where `delta > 0` and yet `x < 0` which would mean it would fall-through. Intentional? Possible to happen?

lib/timeline-window.js
```diff
@@ -0,0 +1,392 @@
...
+    var x;
+    if (delta === 0) {
+        return 0;
+    } else if (delta < 0) {
+        x = Math.max(delta, this.minIndex() - this.index);
+        if (x < 0) {
+            this.index += x;
+            return x;
+        }
+    } else {
```
Kegsay: Careful! What if I call `advance()` without a `delta`? This block will execute due to `=== 0` earlier.

Kegsay:
There's a lot of delicate index juggling to get these timelines to advance/retreat correctly. I'm not confident that the code is correct. I would really like to see some noddy unit tests for the following functions which I feel are delicate:
 - `TimelineIndex.prototype.advance`
 - `TimelineWindow.prototype.getEvents`

lib/timeline-window.js
richvdh: done.

lib/timeline-window.js
richvdh: done.

lib/timeline-window.js
richvdh: done.

lib/timeline-window.js
richvdh: done. Have tried to clarify.

lib/timeline-window.js
richvdh: done.

lib/timeline-window.js
richvdh: Have restructured slightly and added some comments.

done

lib/timeline-window.js
richvdh: done

lib/timeline-window.js
richvdh: replaced `delta === 0` with `!delta`

done

lib/timeline-window.js
richvdh: well, the assumption is that `index` cannot be larger than `maxIndex` (if it is, something has gone wrong earlier), so `this.maxIndex() - this.index` cannot be negative, so `Math.min(delta, this.maxIndex() - this.index) > 0`. If `index` *does* end up being larger than `maxIndex`, then effectively we're off the end of the timeline and we should treat this the same as being at the end of the timeline anyway. So I think this is correct.

Renamed `x` to `cappedDelta` and added some comments.

done

Kegsay:
LGTM

richvdh:
Thanks. I'm going to add some unit tests as you suggest.

PR #65, opened 2016-01-15
Push rules settings

manuroe:
- Added methods to manage push rules
- Fixed sync.js that clobbered client.pushRules

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -2180,7 +2180,7 @@ MatrixClient.prototype.loginWithToken = function(token, callback) {
  * @return {module:client.Promise} Resolves: TODO
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
-MatrixClient.prototype.pushRules = function(callback) {
+MatrixClient.prototype.getPushRules = function(callback) {
     return this._http.authedRequest(callback, "GET", "/pushrules/");
 };
 
@@ -2219,6 +2219,24 @@ MatrixClient.prototype.deletePushRule = function(scope, kind, ruleId, callback) 
         $ruleId: ruleId
     });
     return this._http.authedRequest(callback, "DELETE", path);
+};
+
+/**
+ * Enable or disable a push notification rule.
+ * @param {string} scope
+ * @param {string} kind
+ * @param {string} ruleId
+ * @param {boolean} enabled
+ * @param {module:client.callback} callback Optional.
+ * @return {module:client.Promise} Resolves: result object
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixClient.prototype.setPushRuleEnabled = function(scope, kind, ruleId, enabled, callback) {
+    var path = utils.encodeUri("/pushrules/" + scope + "/$kind/$ruleId/enabled", {
+        $kind: kind,
+        $ruleId: ruleId
+    });
+    return this._http.authedRequest(callback, "PUT", path, undefined, enabled ? 'true' : 'false');
 };
 
 /**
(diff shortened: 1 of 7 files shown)


lib/client.js
```diff
@@ -2222,6 +2222,123 @@ MatrixClient.prototype.deletePushRule = function(scope, kind, ruleId, callback)
...
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixClient.prototype.setPushRuleEnabled = function(scope, kind,
+                                                     ruleId, enabled, callback) {
+    var path = utils.encodeUri("/pushrules/" + scope + "/$kind/$ruleId/enabled", {
+        $kind: kind,
+        $ruleId: ruleId
+    });
+    return this._http.authedRequest(
+        callback, "PUT", path, undefined, enabled ? 'true' : 'false'
```
dbkr: This should be a json object with 'enabled': true / false now in the updated spec.

lib/client.js
```diff
@@ -2222,6 +2222,123 @@ MatrixClient.prototype.deletePushRule = function(scope, kind, ruleId, callback)
...
+};
+
+/**
+ * Get the room-kind push rule associated with a room.
+ * @param {string} scope "global" or device-specific.
+ * @param {string} roomId the id of the room.
+ * @return {object} the rule or undefined.
+ */
+MatrixClient.prototype.getRoomPushRule = function(scope, roomId) {
+    // There can be only room-kind push rule per room
```
dbkr: Maybe we should throw an exception here or something if the pushrules haven't been fetched yet?

lib/client.js
```diff
@@ -2222,6 +2222,123 @@ MatrixClient.prototype.deletePushRule = function(scope, kind, ruleId, callback)
...
+    // Get the existing room-kind push rule if any
+    var roomPushRule = this.getRoomPushRule(scope, roomId);
+    if (roomPushRule) {
+        if (0 <= roomPushRule.actions.indexOf("dont_notify")) {
+            hasDontNotifyRule = true;
+        }
+    }
+
+    if (!mute) {
+        // Remove the rule only it is a muting rule
```
dbkr: 'if it'

lib/client.js
```diff
@@ -2222,6 +2222,123 @@ MatrixClient.prototype.deletePushRule = function(scope, kind, ruleId, callback)
...
+        if (hasDontNotifyRule) {
+            deferred = this.deletePushRule(scope, "room", roomPushRule.rule_id);
```
dbkr: I think this is actually a promise.

lib/client.js
```diff
@@ -2222,6 +2222,123 @@ MatrixClient.prototype.deletePushRule = function(scope, kind, ruleId, callback)
...
+        }
+    }
+    else {
+        if (!roomPushRule) {
+            deferred = this.addPushRule(scope, "room", roomId, {
+                actions: ["dont_notify"]
+            });
+        }
+        else if (!hasDontNotifyRule) {
+            // Remove the existing one before setting the mute push rule
```
dbkr: Is this the thing you're doing to work around syn-590? If so, can you comment it to indicate that?

lib/client.js
```diff
@@ -2222,6 +2222,123 @@ MatrixClient.prototype.deletePushRule = function(scope, kind, ruleId, callback)
...
+                deferred2.reject(err);
+            });
+
+            deferred = deferred2.promise;
+        }
+    }
+
+    if (deferred) {
+        // Update this.pushRules when the operation completes
+        var deferred3 = q.defer();
```
dbkr: I'm a bit lost with 3 different deferreds here. Maybe this could be rule_refresh_deferred or something?

dbkr:
Looking good generally though.

PR #68, opened 2016-01-20
Update for new unread count format

dbkr:


lib/sync.js
```diff
@@ -466,7 +469,7 @@ SyncApi.prototype._sync = function(syncOptions, attempt) {
                 var accountDataEvents = self._mapSyncEventsFormat(joinObj.account_data);
 
                 // we do this first so it's correct when any of the events fire
-                room.unread_notification_count = joinObj.unread_notification_count;
+                room.unread_notifications = joinObj.unread_notifications || {};
```
Kegsay: I was okay when this was just a number, but now this is representing the same structure as the /sync response. One of the things which has saved our figurative bacon in the past is the shielding the JS SDK provides against breaking API changes (which made the v2 swap that much easier).

Please can you add a setter/getter for this on the `Room` object, e.g:

```javascript
Room.setUnreadNotifications(highlightCount, notificationCount)
Room.getNotificationCount(isHighlight)
```

dbkr: Done. nb. I've included 'counts' in the name as it is the count of unread notifs, not the actual notifs. Also I've made the type of count a string param rather than a boolean as just having a boolean makes backwards compat more awkward if we add more types.

dbkr: ptal

dbkr:
Test failure inherited from develop

lib/models/room.js
```diff
@@ -119,10 +119,31 @@ function Room(roomId, opts) {
         //   data: <receipt data>
         // }]
     };
+    this._notificationCounts = {};
 }
 utils.inherits(Room, EventEmitter);
 
 /**
+ * Get one of the notification counst for this room
```
Kegsay: Typo: `counst`

lib/models/room.js
```diff
@@ -119,10 +119,31 @@ function Room(roomId, opts) {
...
+ * @return {Number} The notification count, or undefined if there is no count
+ *                  for this type.
+ */
+Room.prototype.getUnreadNotificationCount = function(type) {
+    type = type || 'total';
+    return this._notificationCounts[type];
+};
+
+/**
+ * Set one of the notification counst for this room
```
Kegsay: Typo

lib/sync.js
```diff
@@ -466,7 +469,15 @@ SyncApi.prototype._sync = function(syncOptions, attempt) {
...
```
Kegsay: Presumably this line should be removed now?

Kegsay:
LGTM aside from minor comments.

PR #69, opened 2016-01-21
Implement a keep-alive timer for /sync requests 

Kegsay:
When a `/sync` request fails, we spin up a keep-alive poll to `/_matrix/client/r0`
which `400`s. We treat any HTTP response code as a success for the purposes of
polling the server. When a successful poll is done, we shoot the current `/sync`
request in the head immediately (via a hacky `abort()` on the promise) and retry
the `/sync`.

The point of this is to fix the problem where OSX can black-hole connections when waking. This would normally hit the 110s client-side timeout for `/sync` which is way too long for a timeout. We effectively now use the keep alive request to check for connectivity now (which has a 5s client-side timeout) and then resend the `/sync` request when it succeeds.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -279,7 +279,7 @@ MatrixClient.prototype.retryImmediately = function() {
     // stop waiting
     clearTimeout(this._syncingRetry.timeoutId);
     // invoke immediately
-    this._syncingRetry.fn();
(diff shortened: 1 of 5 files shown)


ara4n:
have fed back on this in the context of https://github.com/vector-im/vector-web/issues/544

PR #71, opened 2016-01-29
Separate the actual processing of the sync response from the loop doing the requests.

dbkr:
This should be functionally identical: doing this as separate PR to make the diffs a bit more digestible.


Kegsay:
LGTM

PR #72, opened 2016-02-02
Add docs to getRoom noting how you shouldn't assume it will return so…

dbkr:
…mething because you've got a Room-like event


Kegsay:
LGTM

PR #73, opened 2016-02-03
Make sure we don't end up calling /messages in a loop if things go weird

richvdh:
If we somehow end up in a situation where calling /messages returns a load
of messages, but none of them are new, then currently we start calling
/messages again and again in a tight loop. This is bad, so fix it.


Kegsay:
LGTM

PR #74, opened 2016-02-03
Fix a bug which made the timelines get confused about local messages

richvdh:
Make sure that the timeline index is kept consistent when the id of an event
changes when we receive the remote echo of a message we sent.


Kegsay:
Other than lint warnings, LGTM.

PR #76, opened 2016-02-05
Change the sync state tracking slightly to gut-wrench client a bit less

dbkr:
Also add a stop method to stop syncing.

Should be functionally identical outside of client and sync.

### DIFF
--- a/lib/sync.js
+++ b/lib/sync.js
@@ -290,7 +290,7 @@ SyncApi.prototype._peekPoll = function(roomId, token) {
 /**
  * Returns the current state of this sync object
  * @see module:client~MatrixClient#event:"sync"
- * @return {String}
+ * @return {?String}
  */
 SyncApi.prototype.getSyncState = function() {
     return this._syncState;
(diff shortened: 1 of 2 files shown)


lib/sync.js
```diff
@@ -286,6 +288,15 @@ SyncApi.prototype._peekPoll = function(roomId, token) {
 };
 
 /**
+ * Returns the current state of this sync object
+ * @see module:client~MatrixClient#event:"sync"
+ * @return {String}
```
Kegsay: Should be `{?String}` since it can be `null`.

lib/sync.js
```diff
@@ -356,6 +378,11 @@ SyncApi.prototype._sync = function(syncOptions, attempt) {
     var self = this;
     attempt = attempt || 1;
 
+    if (!this._running) {
+        debuglog("Sync no longer running: exiting.");
+        this._updateSyncState("STOPPED");
```
Kegsay: We should document this new sync state.

Kegsay:
LGTM aside from docstring comments. Thanks for the refactor!

PR #77, opened 2016-02-05
Check old timelines when looking for echoes of sent messages

richvdh:
After /send completes, we check the room to see if the echo has already come
back via /sync. It's a bit of an edge-case, but we ought to check all
timelines, not just the live one.

Furthermore, we now have a map from eventId to timeline, so we can handle the
case where the echo has *not* yet come back more efficiently than searching
through the whole timeline.


dbkr:
lgtm

PR #79, opened 2016-02-09
Use only /versions requests for connection recovery

dbkr:
Just only the /versions keepalive logic to recover from lost internet connections.
    
     * This should fix the problem where we could end up with two concurrent syncs
       due to both retrying /sync and hitting /versions. There is now one and only
       one mechanism for connection recovery.
     * Hit /versions a little less often as I think every 2 seconds is a little
       over-aggressive. Also introduce randomness to minimize possibility of
       thundering herds.
     * Add a listener for the 'online' event in case any browser is nice enough
       to ever fire it.
     * Treat a 400 response from /versions as successful since older synapses
       will not support it.

Should fix root cause of https://github.com/vector-im/vector-web/issues/719

### DIFF
--- a/lib/sync.js
+++ b/lib/sync.js
@@ -70,6 +70,7 @@ function SyncApi(client, opts) {
     this._syncState = null;
     this._running = false;
     this._keepAliveTimer = null;
+    this._connectionReturnedDefer = null;
 }
 
 /**
@@ -322,8 +323,9 @@ SyncApi.prototype.sync = function() {
             client.pushRules = result;
             getFilter(); // Now get the filter
         }, function(err) {
-            self._onConnectionReturned = getPushRules;
-            self._startKeepAlives();
+            self._startKeepAlives().done(() => {;
+                getPushRules();
+            });
             self._updateSyncState("ERROR", { error: err });
         });
     }
@@ -338,8 +340,9 @@ SyncApi.prototype.sync = function() {
             debuglog("Using existing filter ID %s", filterId);
             self._sync({ filterId: filterId });
         }, function(err) {
-            self._onConnectionReturned = getFilter;
-            self._startKeepAlives();
+            self._startKeepAlives().done(() => {
+                getFilter();
+            });
             self._updateSyncState("ERROR", { error: err });
         });
     }
@@ -372,7 +375,7 @@ SyncApi.prototype.stop = function() {
  * @return {boolean} True if this resulted in a request being retried.
  */
 SyncApi.prototype.retryImmediately = function() {
-    if (!this._onConnectionReturned) { return false; }
+    if (!this._connectionReturnedDefer) { return false; }
     this._startKeepAlives();
     return true;
 };
@@ -389,6 +392,10 @@ SyncApi.prototype._sync = function(syncOptions) {
 
     if (!this._running) {
         debuglog("Sync no longer running: exiting.");
+        if (self._connectionReturnedDefer) {
+            self._connectionReturnedDefer.reject();
+            self._connectionReturnedDefer = null;
+        }
         this._updateSyncState("STOPPED");
     }
 
@@ -479,6 +486,10 @@ SyncApi.prototype._sync = function(syncOptions) {
     }, function(err) {
         if (!self._running) {
             debuglog("Sync no longer running: exiting");
+            if (self._connectionReturnedDefer) {
+                self._connectionReturnedDefer.reject();
+                self._connectionReturnedDefer = null;
+            }
             return;
         }
         console.error("/sync error %s", err);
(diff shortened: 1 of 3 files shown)


lib/sync.js
```diff
@@ -360,21 +357,33 @@ SyncApi.prototype.sync = function() {
  */
 SyncApi.prototype.stop = function() {
     debuglog("SyncApi.stop");
+    if (global.document) {
+        global.document.removeEventListener("online", this._onOnline.bind(this), false);
```
richvdh: does this actually work? doesn't `.bind` give a different function to that which was originally bound?

lib/sync.js
```diff
@@ -360,21 +357,33 @@ SyncApi.prototype.sync = function() {
...
 };
 
 /**
+ * Retry a backed off syncing request immediately. This should only be used when
+ * the user <b>explicitly</b> attempts to retry their lost connection.
+ * @return {boolean} True if this resulted in a request being retried.
+ */
+SyncApi.prototype.retryImmediately = function() {
+    if (!this._onConnectionReturned) { return false; }
+    this._startKeepAlives();
```
richvdh: won't this start another retry timer? so instead of retrying every 5 seconds or so, we'll be retrying twice every 5 seconds?

lib/sync.js
```diff
@@ -304,27 +305,29 @@ SyncApi.prototype.sync = function() {
...
-        attempt += 1;
-
+    function getPushRules() {
         client.getPushRules().done(function(result) {
             debuglog("Got push rules");
             client.pushRules = result;
             getFilter(); // Now get the filter
-        }, retryHandler(attempt, getPushRules));
+        }, function(err) {
+            self._onConnectionReturned = getPushRules;
```
richvdh: can't help feeling that this would be more elegantly achieved by waiting on a promise which resolves when the connection returns. ymmv though; happy for you to stick with this if you prefer.

lib/sync.js
```diff
@@ -394,7 +403,7 @@ SyncApi.prototype._sync = function(syncOptions, attempt) {
         since: syncToken || undefined // do not send 'null'
     };
 
-    if (attempt > 1) {
+    if (self._syncConnectionLost) {
```
richvdh: is this still required?

richvdh:
looks reasonable. a few questions though.

lib/sync.js
dbkr: As in rather than invoking getFilter()? I would probably agree, but didn't want to go too far into refactoring whilst changing other behaviour.

lib/sync.js
dbkr: No, _startKeepAlives starts by cancelling the timer if one currently exists.

lib/sync.js
dbkr: I believe so, since we only actually consider the connection back up once the sync comes back rather than when /versions returns. Possibly this is unnecessary but it feels more true to only emit the SYNCING event after we've actually got a new sync response.

lib/sync.js
dbkr: Ah yeah, thanks.

lib/sync.js
richvdh: as in, rather than having `_pokeKeepAlive` call `_onConnectionReturned`, have it resolve a promise.

lib/sync.js
richvdh: ack.

lib/sync.js
```diff
@@ -641,13 +633,48 @@ SyncApi.prototype._processSyncResponse = function(syncToken, data) {
...
+    }, function(err) {
+        if (err.httpStatus == 400) {
+            // treat this as a success because the server probably just doesn't
+            // support /versions: point is, we're getting a response.
+            // We wait a short time though, just in case somehow the server
+            // is in a mode where it 400s /versions responses and sync etc.
+            // responses fail, this will mean we don't hammer in a loop.
+            setTimeout(success, 2000);
+        } else {
+            setTimeout(
```
richvdh: need to set `self._keepAliveTimer` here

lib/sync.js
richvdh: ah right. If the objective is to emit the SYNCING event sooner, could make the comment say so?

lib/sync.js
dbkr: Aha! Yes, good catch!

lib/sync.js
dbkr: Oh yes - I think you're right actually. I was worried it would be too complicated to keep it so you can call _startKeepAlives() to restart the stream too, but it's actually fine.

lib/sync.js
richvdh: done

lib/sync.js
richvdh: done

lib/sync.js
richvdh: done

lib/sync.js
richvdh: done

lib/sync.js
richvdh: done

richvdh:
lgtm

PR #80, opened 2016-02-16
Keep redacted events in the timeline

richvdh:
Everything gets confused when we remove events from timelines, so keep
redacted events in there, and mark them as redacted instead.

### DIFF
--- a/lib/models/event.js
+++ b/lib/models/event.js
@@ -21,8 +21,6 @@ limitations under the License.
  * @module models/event
  */
 
-var utils = require("../utils");
-
 /**
  * Enum for event statuses.
  * @readonly
--- a/lib/models/event.js
+++ b/lib/models/event.js
@@ -204,14 +204,17 @@ module.exports.MatrixEvent.prototype = {
         }
         this.event.unsigned.redacted_because = redaction_event;
 
-        for (var key in utils.keys(this.event)) {
+        var key;
+        for (key in this.event) {
+            if (!this.event.hasOwnProperty(key)) { continue; }
             if (!_REDACT_KEEP_KEY_MAP[key]) {
                 delete this.event[key];
             }
         }
 
         var keeps = _REDACT_KEEP_CONTENT_MAP[this.getType()] || {};
-        for (var key in utils.keys(this.event.content)) {
+        for (key in this.event.content) {
+            if (!this.event.content.hasOwnProperty(key)) { continue; }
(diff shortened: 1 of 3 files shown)


richvdh:
This is part of the fix for vector-im/vector-web#861 and  vector-im/vector-web#901. react-sdk stuff to follow.

lib/models/event.js
```diff
@@ -189,5 +189,28 @@ module.exports.MatrixEvent.prototype = {
...
+     * if it were redacted before it was sent to us
+     *
+     * @param {Object} the raw event causing the redaction
+     */
+    makeRedacted: function(redaction_event) {
+        if (!this.event.unsigned) {
+            this.event.unsigned = {};
+        }
+        this.event.unsigned.redacted_because = redaction_event;
+        this.event.content = {};
```
ara4n: Is this sufficient?  http://matrix.org/docs/spec/r0.0.1/client_server.html#redactions says: 

Upon receipt of a redaction event, the server should strip off any keys **not** in the following list:
 * event_id
 * type
 * room_id
 * user_id
 * state_key
 * prev_state
 * content

The content object should also be stripped of all keys, unless it is one of one of the following event types:

 * m.room.member allows key membership
 * m.room.create allows key creator
 * m.room.join_rules allows key join_rule
 * m.room.power_levels allows keys ban, events, events_default, kick, redact, state_default, users, users_default.
 * m.room.aliases allows key aliases

I have used redactions quite a few times on things like membership change events in order to redact offensive displaynames - is this going to be sufficient?

ara4n:
lgtm other than the concern above

lib/models/event.js
richvdh: It should be sufficient for the case of redacting offensive displaynames; the displayname is in the `content` key so is removed along with everything else.

There is a problem in that `m.room.member` events are used to update a cached copy of the user's metadata, which is then copied to each event they send. We don't update the cache on redactions, nor any events they have sent - which means that any such events will still show the redacted displayname, as will the member list on the RHS. However, I don't believe this is a new issue.

If anything, we are removing too much stuff from `content`. It's probably best we follow the specified algorithm here, rather than allow people to build on functionality which isn't really correct; I'll tighten this up.

richvdh:
Matthew: PTAL

ara4n:
lgtm - thanks for the full redaction alg

PR #81, opened 2016-02-16
EventTimeline: Fix baseIndex after removing the last event

richvdh:
Removing the last event in an EventTimeline (as we might, for instance, if it
was a local echo in an empty timeline) got us into a state where the baseIndex
would increment when adding events to the end of the timeline, causing much
confusion.

This should fix https://github.com/vector-im/vector-web/issues/937.


ara4n:
seems plausible; lgtm

PR #82, opened 2016-02-19
Add a param to getEventReadUpTo to have it ignore implicit read receipts

dbkr:
Store real receipts separately to make this work.

### DIFF
--- a/lib/models/room.js
+++ b/lib/models/room.js
@@ -138,6 +138,8 @@ function Room(roomId, opts) {
         //   data: <receipt data>
         // }]
     };
+    // only receipts that came from the server, not synthesized ones
+    this._realReceipts = {};
 
     this._notificationCounts = {};
 
@@ -632,7 +634,7 @@ Room.prototype._addLiveEvents = function(events) {
         if (events[i].sender && !isLocalEcho) {
             this.addReceipt(synthesizeReceipt(
                 events[i].sender.userId, events[i], "m.read"
-            ));
+            ), true);
         }
     }
 };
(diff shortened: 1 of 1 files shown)


lib/models/room.js
```diff
@@ -904,7 +914,7 @@ Room.prototype.getReceiptsForEvent = function(event) {
  * Add a receipt event to the room.
  * @param {MatrixEvent} event The m.receipt event.
  */
-Room.prototype.addReceipt = function(event) {
+Room.prototype.addReceipt = function(event, fake) {
```
richvdh: could do with doc for `fake` param

lib/models/room.js
```diff
@@ -877,17 +879,25 @@ Room.prototype.getUsersReadUpTo = function(event) {
  * Get the ID of the event that a given user has read up to, or null if we
  * have received no read receipts from them.
  * @param {String} userId The user ID to get read receipt event ID for
+ * @param {Boolean} ignoreSynthesized If true, do return only receipts that have been
```
richvdh: s/do //

richvdh:
What about the receipt synthesized in `_addLocalEchoReceipt`?

This feels slightly overspecced given we only care about the distinction for our own read receipts, but I don't really have any concrete suggestions for doing it better.

dbkr:
I think I'm going with local echo receipts should not be returned if you ask for only real ones, although I'm not totally certain. I think we do want this mode to just give the server's version of the truth without any made-up stuff by the js sdk, since the receipt we sent might never actually get through to the server.

ptal.

richvdh:
lgtm

PR #84, opened 2016-02-23
Add support for new keypair style 3pid invites

dbkr:
(add an option to joinRoom for specifying the signing url)

Also @illicitonion

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -661,31 +661,50 @@ MatrixClient.prototype.joinRoom = function(roomIdOrAlias, opts, callback) {
     if (utils.isFunction(opts)) {
         throw new Error("Expected 'opts' object, got function.");
     }
-    opts = opts || {
-        syncRoom: true
-    };
+    opts = opts || {};
+    if (opts.syncRoom === undefined) { opts.syncRoom = true; }
 
     var room = this.getRoom(roomIdOrAlias);
     if (room && room.hasMembershipState(this.credentials.userId, "join")) {
         return q(room);
     }
-    var path = utils.encodeUri("/join/$roomid", { $roomid: roomIdOrAlias});
+
+    var sign_promise = q();
+
+    if (opts.inviteSignUrl) {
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -661,31 +661,50 @@ MatrixClient.prototype.joinRoom = function(roomIdOrAlias, opts, callback) {
...
 
     var room = this.getRoom(roomIdOrAlias);
     if (room && room.hasMembershipState(this.credentials.userId, "join")) {
         return q(room);
     }
-    var path = utils.encodeUri("/join/$roomid", { $roomid: roomIdOrAlias});
+
+    var sign_promise = q();
+
+    if (opts.inviteSignUrl) {
```
richvdh: needs docs

lib/client.js
```diff
@@ -661,31 +661,50 @@ MatrixClient.prototype.joinRoom = function(roomIdOrAlias, opts, callback) {
...
         }
-        return q(room);
-    }, function(err) {
-        _reject(callback, defer, err);
-    }).done(function(room) {
-        _resolve(callback, defer, room);
+
+        var path = utils.encodeUri("/join/$roomid", { $roomid: roomIdOrAlias});
+        self._http.authedRequest(undefined, "POST", path, undefined, data).then(
+        function(res) {
```
richvdh: rather than nesting the promise handlers and ending up with a pyramid of doom, how about:

```
sign_promise.then(function(signed_invite_object) {
    ...
    return self._http.authedRequest(undefined, "POST", path, undefined, data);
}).then(function(res) {
    var roomId = res.room_id;
    ...
    return q(room);
}).done(function(room) {
    _resolve(callback, defer, room);
}, function(err) {
    _reject(callback, defer, err);
});
```

richvdh:
looks fine otherwise

dbkr:
ptal

richvdh:
would be nice to see some tests for it, but otherwise lgtm

PR #85, opened 2016-02-24
Check filters before we reuse them

richvdh:
Make sure that we check the content of existing filters before we blindly reuse
them.

This feels a bit nasty because the chances are that we could reuse an existing
filter. But taking advantage of that would mean changing the synapse API, which I have little appetite right at the moment.

Fixes https://github.com/vector-im/vector-web/issues/988


dbkr:
lgtm

PR #86, opened 2016-02-25
js-sdk: Add a 'version.txt' file to the npm tarball

richvdh:


package.json
```diff
@@ -10,7 +10,8 @@
     "build": "jshint -c .jshint lib/ && browserify browser-index.js -o dist/browser-matrix-dev.js --ignore-missing",
     "watch": "watchify browser-index.js -o dist/browser-matrix-dev.js -v",
     "lint": "jshint -c .jshint lib spec && gjslint --unix_mode --disable 0131,0211,0200,0222,0212 --max_line_length 90 -r spec/ -r lib/",
-    "release": "npm run build && mkdir dist/$npm_package_version && uglifyjs -c -m -o dist/$npm_package_version/browser-matrix-$npm_package_version.min.js dist/browser-matrix-dev.js && cp dist/browser-matrix-dev.js dist/$npm_package_version/browser-matrix-$npm_package_version.js"
+    "release": "npm run build && mkdir dist/$npm_package_version && uglifyjs -c -m -o dist/$npm_package_version/browser-matrix-$npm_package_version.min.js dist/browser-matrix-dev.js && cp dist/browser-matrix-dev.js dist/$npm_package_version/browser-matrix-$npm_package_version.js",
+    "prepublish": "git rev-parse HEAD > version.txt"
```
dbkr: This will also have the side effect that our releases end up with a version.txt file in the root which is git hash, which could be slightly confusing since the advertised version will be vx.x.x.

dbkr:
let's try it, lgtm.

richvdh:
we've decided to s/version.txt/git-revision.txt/, but otherwise to go with this as the least bad approach.

PR #87, opened 2016-02-25
Fire a 'Room.timelineReset' event when we get a gappy sync

richvdh:
We need to reset things at the UI level when we get a gappy sync, so give the
clients something to listen for.

Also add a bunch of tests for that bit of code.


dbkr:
lgtm

PR #88, opened 2016-02-26
TimelineWindow.load: make the livetimeline case quicker

richvdh:
Avoid doing a loop round the reactor if we are just loading the live timeline.


dbkr:
lgtm

PR #89, opened 2016-02-26
Use our inviter's member event to get their display name if it exists.

dbkr:
Also rename slightly confusingly named member list variables.

### DIFF
--- a/lib/models/room.js
+++ b/lib/models/room.js
@@ -1136,13 +1136,15 @@ function calculateRoomName(room, userId, ignoreRoomNameEvent) {
 
     // TODO: Localisation
     if (myMemberEvent && myMemberEvent.content.membership == "invite") {
-        if (otherMembers.length && otherMembers[0].events.member) {
+        if (room.currentState.getMember(myMemberEvent.sender)) {
             // extract who invited us to the room
-            return "Invite from " + otherMembers[0].name;
+            return "Invite from " + room.currentState.getMember(
+                myMemberEvent.sender
+            ).name;
         } else if (allMembers[0].events.member) {
             // use the sender field from the invite event, although this only
             // gets us the mxid
-            return "Invite from " + allMembers[0].events.member.getSender();
+            return "Invite from " + myMemberEvent.sender;
         } else {
             return "Room Invite";
         }
--- a/lib/models/room.js
+++ b/lib/models/room.js
@@ -1120,52 +1120,63 @@ function calculateRoomName(room, userId, ignoreRoomNameEvent) {
     }
 
     // get members that are NOT ourselves and are actually in the room.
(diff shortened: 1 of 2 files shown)


lib/models/room.js
```diff
@@ -1120,70 +1120,81 @@ function calculateRoomName(room, userId, ignoreRoomNameEvent) {
     }
 
     // get members that are NOT ourselves and are actually in the room.
-    var members = utils.filter(room.currentState.getMembers(), function(m) {
+    var otherMembers = utils.filter(room.currentState.getMembers(), function(m) {
         return (m.userId !== userId && m.membership !== "leave");
     });
+    var allMembers = utils.filter(room.currentState.getMembers(), function(m) {
```
richvdh: use this to derive `otherMembers` and `myMemberEventArray` (since there might be a large number of `leave` members?)

lib/models/room.js
```diff
@@ -1120,70 +1120,81 @@ function calculateRoomName(room, userId, ignoreRoomNameEvent) {
...
+    );
+
     // TODO: Localisation
-    if (members.length === 0) {
-        var memberList = utils.filter(room.currentState.getMembers(), function(m) {
-            return (m.membership !== "leave");
-        });
-        if (memberList.length === 1) {
+    if (myMemberEvent && myMemberEvent.content.membership == "invite") {
+        if (otherMembers.length && otherMembers[0].events.member) {
```
richvdh: you seem to be assuming that it's necessarily the first member that invited us. That doesn't sound right.

lib/models/room.js
```diff
@@ -1120,70 +1120,81 @@ function calculateRoomName(room, userId, ignoreRoomNameEvent) {
...
+            // gets us the mxid
+            return "Invite from " + allMembers[0].events.member.getSender();
+        } else {
+            return "Room Invite";
+        }
+    }
+
+
+    if (otherMembers.length === 0) {
+        if (allMembers.length === 1) {
```
richvdh: maybe we could invert some of these conditions to avoid so much indentation.

richvdh:
I'm not entirely clear on what the desired functionality change is here. What's going on?

Also, how about some tests?

dbkr:
In principle I agree with all of this, but I think all of these comments relate to the existing code, and I don't really want to refactor it before fixing the bug.

Tests for the room name calculation are here: https://github.com/matrix-org/matrix-js-sdk/blob/develop/spec/unit/room.spec.js#L658

lib/models/room.js
richvdh: ok so I understand this better now. Could you add a comment to clarify the assumption being made?

dbkr:
Code is slightly more sensible now, and added tests for the new invite behaviour.

richvdh:
lgtm

PR #91, opened 2016-02-29
Don't reset the timeline when we join a room after peeking

richvdh:
If we've already got all the events in a limited sync, there is no need to reset
the timeline.

This fixes a bug where, if you peek into a world-readable room, and then join it, we get stuck paginating in a tight loop until an event arrives (or you change room).

I think this is probably also responsible for a number of other bugs, like https://github.com/vector-im/vector-web/issues/1045 and https://github.com/vector-im/vector-web/issues/943, though I haven't yet reproduced those fully.


dbkr:
OK, think I've got my head around this! lgtm.

PR #93, opened 2016-03-01
Set the back-pagination token before raising Room.timelineReset

richvdh:
This fixes another race condition on gappy syncs, wherein we weren't
back-paginating back from the start of the gappy sync.


richvdh:
This fixes the other half of https://github.com/vector-im/vector-web/issues/1035

PR #94, opened 2016-03-03
Add maySendStateEvent method, ported from react-sdk (but fixed).

dbkr:
Plus tests.

### DIFF
--- a/lib/models/room-state.js
+++ b/lib/models/room-state.js
@@ -234,7 +234,8 @@ RoomState.prototype.getUserIdsWithDisplayName = function(displayName) {
  *                        according to the room's state.
  */
 RoomState.prototype.maySendStateEvent = function(stateEventType, userId) {
-    if (!this.getMember(userId)) { return false; }
+    var member = this.getMember(userId);
+    if (!member || member.membership == 'leave') { return false; }
 
     var power_levels_event = this.getStateEvents('m.room.power_levels', '');
 
--- a/lib/models/room-state.js
+++ b/lib/models/room-state.js
@@ -244,7 +244,6 @@ RoomState.prototype.maySendStateEvent = function(stateEventType, userId) {
 
     var default_user_level = 0;
     var user_levels = [];
-    var current_user_level = 0;
 
     var state_default = 0;
     if (power_levels_event) {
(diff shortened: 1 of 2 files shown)


dbkr:
(Jenkins is broken)

lib/models/room-state.js
```diff
@@ -225,6 +225,52 @@ RoomState.prototype.getUserIdsWithDisplayName = function(displayName) {
...
+    var power_levels_event = this.getStateEvents('m.room.power_levels', '');
+
+    var power_levels;
+    var events_levels = {};
+
+    var default_user_level = 0;
+    var user_levels = [];
+    var current_user_level = 0;
+
+    var state_default = 0;
```
richvdh: why is this `0`, but `50` at line 262? it is odd that the required power level is `50` if there is an `m.room.power_levels` with no `state_default`, but `0` if there is no `m.room.power_levels` at all

(It is also odd that the spec doesn't clarify this)

richvdh:
It is not the easiest to follow bit of code, but looks ok to me modulo the above comment

richvdh:
A quick thought: could we make this `maySendEvent` with a boolean param for `stateEvent`, since the algorithm is identical modulo a single key lookup?

dbkr:
Could do: my plan was probably to make a separate maySendEvent function if necessary and then probably back both functions with the same one internally. Either way it's a backwards compatible change (assuming you'd add the param to the end).

lib/models/room-state.js
dbkr: Yes, so in this case I'd like your opinion on https://github.com/matrix-org/matrix-doc/pull/286 which is where I'm fixing the spec to document the default levels. I think it's a bug in synapse that the default is 0 with no power levels event, but apparently this is now baked into room history so we have to make the spec fit the bug.

richvdh:
Just trying to fend off having to support three functions in the js-sdk API forever. No strong feelings here though.

lib/models/room-state.js
```diff
@@ -225,6 +225,52 @@ RoomState.prototype.getUserIdsWithDisplayName = function(displayName) {
...
+ * Returns true if the given user ID has permission to send a state
+ * event of type `stateEventType` into this room.
+ * @param {string} type The type of state events to test
+ * @param {string} userId The user ID of the user to test permission for
+ * @return {boolean} true if the given user ID should be permitted to send
+ *                        the given type of state event into this room,
+ *                        according to the room's state.
+ */
+RoomState.prototype.maySendStateEvent = function(stateEventType, userId) {
+    if (!this.getMember(userId)) { return false; }
```
Kegsay: This only tests for presence of `m.room.member` events. This will return a `RoomMember` with `membership = leave` if they leave the room.

Kegsay: You can also help yourself by using `RoomMember.powerLevel` to get the actual power level of that member, which already takes into account `users_default` and explicit `users` entries.

PR #95, opened 2016-03-08
Emit an event when a local-echo is turned into a proper event

richvdh:
We need to trigger an update of the timeline when this happens, so raise an
event for it.

### DIFF
--- a/lib/models/room.js
+++ b/lib/models/room.js
@@ -595,18 +595,22 @@ Room.prototype._addLiveEvents = function(events) {
             // remote echo of an event we sent earlier
             var existingEvent = this._txnToEvent[events[i].getUnsigned().transaction_id];
             if (existingEvent) {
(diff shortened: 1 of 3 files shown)


lib/models/room.js
```diff
@@ -1315,3 +1319,21 @@ module.exports = Room;
  *   }
  * });
  */
+
+/**
+ * Fires when a local-echo of a transmitted event is replaced by its remote-echo.
+ *
+ * <p>When an event is first transmitted, then (assuming the /send completes
```
dbkr: Do your p tags need to be closed here?

dbkr:
otherwise lgtm

lib/models/room.js
richvdh: empirically not, and we don't seem to have been closing them in general in the js-sdk doc-comments.

PR #96, opened 2016-03-08
Work around confused timelines causing pagination loops

richvdh:
Look out for us getting stuck in a loop of using the same pagination token,
and use something else next time.

Hopefully this will fix https://github.com/vector-im/vector-web/issues/1089.


dbkr:
lgtm

PR #97, opened 2016-03-10
Raise localEchoUpdated events in more places

richvdh:
We need to know about more transitions for local-echo status changes, so raise
localEchoUpdated events for each transition.

Fixes an issue where we weren't turning failed transmissions red, because the
timeline wasn't being updated.


dbkr:
lgtm

PR #98, opened 2016-03-14
Try again if a pagination request gives us no new messages

richvdh:
This is basically a workaround for https://matrix.org/jira/browse/SYN-645: if
we knew about all of the events already, we want to try again.

Fixes the second half of https://github.com/vector-im/vector-web/issues/1014


dbkr:
lgtm

PR #99, opened 2016-03-14
Add a delay before we start polling the connectivity check endpoint

dbkr:
…to avoid tightlooping if the conn check succeeds but /sync etc fails.


richvdh:
lgtm

PR #100, opened 2016-03-15
Add Session.logged_out event

dbkr:
Session.logged_out fires whenever the current session is no longer valid and the user needs to log in again.

Also null check _syncApi before trying to stop it.

https://github.com/vector-im/vector-web/issues/414

### DIFF
--- a/lib/http-api.js
+++ b/lib/http-api.js
@@ -45,6 +45,7 @@ module.exports.PREFIX_IDENTITY_V1 = "/_matrix/identity/api/v1";
 /**
  * Construct a MatrixHttpApi.
  * @constructor
+ * @param {MatrixClient} client The matrix client instance to use.
  * @param {Object} opts The options to use for this HTTP API.
  * @param {string} opts.baseUrl Required. The base client-server URL e.g.
  * 'http://localhost:8008'.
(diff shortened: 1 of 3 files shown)


lib/http-api.js
```diff
@@ -45,6 +45,7 @@ module.exports.PREFIX_IDENTITY_V1 = "/_matrix/identity/api/v1";
 /**
  * Construct a MatrixHttpApi.
  * @constructor
+ * @param {MatrixClient} client The matrix client instance to use.
```
richvdh: can we call make this an `EventEmitter` rather than a client, so that we aren't tempted to start adding all sorts of binding between the HttpApi and MatrixClient?

richvdh:
looks ok except for the build fail and my comment above

dbkr:
ptal

lib/http-api.js
```diff
@@ -268,7 +270,19 @@ module.exports.MatrixHttpApi.prototype = {
...
-        return this.request(callback, method, path, queryParams, data, localTimeoutMs);
+        var self = this;
+        var request_promise = this.request(
+            callback, method, path, queryParams, data, localTimeoutMs
+        );
+        request_promise.catch(function(err) {
+            if (err.errcode == 'M_UNKNOWN_TOKEN') {
+                self.event_emitter.emit("Session.logged_out");
+            }
+            throw err;
```
richvdh: this will get lost... you might as well remove it

dbkr: The comment below? Or the promise catch?

richvdh: the thrown exception. line 281.

dbkr: Oh of course, yes - I see.

richvdh:
lgtm

PR #101, opened 2016-03-15
Clean up a codepath that was only used for crypto messages

richvdh:
Transmission of encrypted messages was happening somewhat differently to
normal messages. In particular, we weren't copying the 'unsigned' field when we
got the remote-echo, which meant the 'sync' code didn't correctly match up the
echo with the original.

The separate codepath was becoming a thorn in my side, so fix things up to
bring it back in line.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1112,6 +1112,7 @@ function _decryptMessage(client, event) {
                 room_id: payload.room_id,
                 user_id: event.getSender(),
                 event_id: event.getId(),
+                unsigned: event.getUnsigned(),
                 type: payload.type,
                 content: payload.content
             }, "encrypted");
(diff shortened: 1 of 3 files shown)


dbkr:
Seems reasonable although I'm failing to understand how this has impacted the event.unsigned thing for normal messages if you're only changing stuff that ran for encrypted events?

richvdh:
You mean, why did I have to change the tests?

Because the tests were relying on the bit of code which I have removed. It was only used by encrypted events when live, but the test wasn't accurately reproducing what we get from the server, so was ending up in the wrong bit of code.

dbkr:
Oh, I see - yes, that is very bizarre. lgtm then.

PR #102, opened 2016-03-15
Implement 'pendingEventList'

richvdh:
The existing 'pendingEventOrdering'=='end' semantics had been substantially
broken by the introduction of timelines and gappy syncs: after a gappy
sync, pending events would get stuck in the old timeline section. (Part of
https://github.com/vector-im/vector-web/issues/1120).

This change replaces 'pendingEventOrdering'=='end' with
'pendingEventOrdering'=='external', which instead keeps the pending events in a
completely separate list, accessed via room.getPendingEvents().

While we're in there, give MatrixClient.sendEvent its own entry point
'Room.addPendingMessage' instead of pushing it through addEventsToTimeline;
this considerably simplifies the implementation of the latter and also means
that we can contain the _txnId ming to MatrixClient.

Also add a new EventStatus 'SENT' for events which have been successfully sent
but whose remote echo we still haven't received.

This will require some change in the react-sdk; PR will follow shortly.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -144,7 +144,10 @@ function MatrixClient(opts) {
         var self = this;
         this.scheduler.setProcessFunction(function(eventToSend) {
             var room = self.getRoom(eventToSend.getRoomId());
-            _updateLocalEchoStatus(room, eventToSend, EventStatus.SENDING);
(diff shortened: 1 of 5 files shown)


richvdh:
I'm going to split this up

PR #106, opened 2016-03-16
make sure we show display names & avatars on parts, and use the right…

ara4n:
… type of content for displaynames for member events in general. fixes https://github.com/vector-im/vector-web/issues/1140 and https://github.com/vector-im/vector-web/issues/873 and a bunch more

### DIFF
--- a/lib/models/room-member.js
+++ b/lib/models/room-member.js
@@ -203,7 +203,10 @@ RoomMember.prototype.getAvatarUrl =
 };
 
 function calculateDisplayName(member, event, roomState) {
-    var displayName = event.getDirectionalContent().displayname;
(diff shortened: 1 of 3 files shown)


lib/models/room-member.js
```diff
@@ -203,7 +203,10 @@ RoomMember.prototype.getAvatarUrl =
 };
 
 function calculateDisplayName(member, event, roomState) {
-    var displayName = event.getDirectionalContent().displayname;
+    var displayName = event.getType() === "m.room.membership" ?
```
Kegsay: I think you mean `m.room.member`

Kegsay:
If it still works after changing the event type, then LGTM

PR #109, opened 2016-03-17
Log the stack when we get a sync error

richvdh:
If we have the stack for an exception in the /sync loop, we should log it.


dbkr:
lgtm (thank you, this was really annoying me)

PR #110, opened 2016-03-17
Refactor transmitted-messages code

richvdh:
This is some preparatory work for fixing up the problems with the timeline ordering of unsent messages
(https://github.com/vector-im/vector-web/issues/1120). The functional changes here should be minimal (bar an extra `Room.localEchoUpdated` when the local echo is first added to the timeline).

Give `MatrixClient.sendEvent` its own entry point `Room.addPendingMessage` instead of pushing it through `addEventsToTimeline`; this considerably simplifies the implementation of the latter and also means that we can contain the `_txnId` ming to MatrixClient.

Move the code which deals with a successful `/send` response from `MatrixClient` into `Room.updatePendingEvent`, since it involves fiddling with the innards of the Room.

Also adds a new EventStatus 'SENT' for events which have been successfully sent but whose remote echo we still haven't received.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -144,7 +144,10 @@ function MatrixClient(opts) {
         var self = this;
         this.scheduler.setProcessFunction(function(eventToSend) {
             var room = self.getRoom(eventToSend.getRoomId());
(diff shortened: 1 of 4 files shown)


dbkr:
lgtm

PR #111, opened 2016-03-17
Implement 'pendingEventList'

richvdh:
The existing 'pendingEventOrdering'=='end' semantics had been substantially
broken by the introduction of timelines and gappy syncs: after a gappy
sync, pending events would get stuck in the old timeline section. (Part of
https://github.com/vector-im/vector-web/issues/1120).

This is a breaking change; PR for matrix-react-sdk will follow.

PR #112, opened 2016-03-17
Support for cancelling pending events

richvdh:
Implement client.cancelPendingEvent which will cancel queued or not_sent events

This won't work properly until https://github.com/matrix-org/matrix-js-sdk/pull/111 is landed.


dbkr:
lgtm

PR #113, opened 2016-03-18
Add a method to the js sdk to look up 3pids on the ID server.

dbkr:
This also moves the manual json decoding for the ID server into the common ID server block, since we'll have to do it for all ID server requests, rather than doing it in requestEmailToken.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -3242,7 +3242,9 @@ MatrixClient.prototype.requestEmailToken = function(email, clientSecret,
  * @param {string} medium The medium of the threepid, eg. 'email'
  * @param {string} address The textual address of the threepid
  * @param {module:client.callback} callback Optional.
- * @return {module:client.Promise} Resolves: TODO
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -3233,12 +3233,27 @@ MatrixClient.prototype.requestEmailToken = function(email, clientSecret,
...
+    );
+};
+
+/**
+ * Looks up the public Matrix ID mapping for a given 3rd party
+ * identifier from the Identity Server
+ * @param {string} medium The medium of the threepid, eg. 'email'
+ * @param {string} address The textual address of the threepid
+ * @param {module:client.callback} callback Optional.
+ * @return {module:client.Promise} Resolves: TODO
```
richvdh: aside: y'know, just because we've been rubbish at documenting these so far doesn't mean that we can't start...

richvdh:
lgtm

PR #115, opened 2016-03-18
Include a counter in generated transaction IDs

richvdh:
Fixes a flaky test which sometimes failed due to sending two events in the same
millisecond.


dbkr:
lgtm

PR #117, opened 2016-03-22
get/setRoomVisibility API

ara4n:


ara4n:
depends on https://github.com/matrix-org/synapse/pull/657

lib/client.js
```diff
@@ -1938,6 +1938,37 @@ MatrixClient.prototype.resolveRoomAlias = function(roomAlias, callback) {
...
+ */
+MatrixClient.prototype.getRoomVisibility = function(roomId, callback) {
+    var path = utils.encodeUri("/directory/list/room/$roomId", {
+        $roomId: roomId
+    });
+    return this._http.authedRequest(callback, "GET", path);
+};
+
+/**
+ * Enable or disable a push notification rule.
```
Kegsay: ORLY?

lib/client.js
```diff
@@ -1938,6 +1938,37 @@ MatrixClient.prototype.resolveRoomAlias = function(roomAlias, callback) {
 
 /**
  * @param {string} roomId
+ * @param {module:client.callback} callback Optional.
+ * @return {module:client.Promise} Resolves: TODO
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixClient.prototype.getRoomVisibility = function(roomId, callback) {
```
Kegsay: The name of this isn't immediately clear this refers to the room directory listing. Whilst I appreciate the parity with the `/createRoom` API which sets this, the `visibility` key there does a lot more than *just* that (`join_rules` as well!). Given that, I would prefer it if you could thwack a "directory" in there somewhere to make it clear it is for the directory rather than say, a shortcut for `join_rules`. `getRoomDirectoryVisibility` perhaps?

lib/client.js
```diff
@@ -1938,6 +1938,37 @@ MatrixClient.prototype.resolveRoomAlias = function(roomAlias, callback) {
...
+    var path = utils.encodeUri("/directory/list/room/$roomId", {
+        $roomId: roomId
+    });
+    return this._http.authedRequest(callback, "GET", path);
+};
+
+/**
+ * Enable or disable a push notification rule.
+ * @param {string} roomId
+ * @param {string} visibility
```
Kegsay: `string`, really? I'd have thought a simple `boolean` would suffice. What's the third/fourth/etc options?

ara4n: left this way per erik's request to support group level visibility

ara4n:
thanks; done - @kegsay ptal

Kegsay:
LGTM

PR #119, opened 2016-03-24
Only count joined members for the member count in notifications.

dbkr:
Fixes https://github.com/vector-im/vector-web/issues/1067


richvdh:
lgtm

PR #120, opened 2016-03-29
Attempts to improve sleep-resume

richvdh:
Reduce a bunch of timeouts on /sync

* reduce the 'timeout' parameter on syncs to 15s (was 30s)
* reduce the additional delay on syncs to 5s (was 80s)
* reduce the timeout on /versions to 2.25s +/- .25s (was 15s)
* reduce the delay between /versions calls to 1.25s +/- .25s (was 7.5 +/- 2.5)
* If /versions times out, don't delay again

Also cancel timed-out requests.

### DIFF
--- a/lib/sync.js
+++ b/lib/sync.js
@@ -32,11 +32,19 @@ var EventTimeline = require("./models/event-timeline");
 
 var DEBUG = true;
 
+
+// the default value for the "timeout=" parameter on /sync requests. (This can
+// be overridden by the application's setting of opts.pollTimeout). This
+// determines the server-side timeout on incremental syncs. The client-side
+// timeout (for incremental and intial syncs) is determined by this, plus
+// BUFFER_PERIOD_MS.
+var DEFAULT_SYNC_TIMEOUT_MS = 15 * 1000;
+
 // /sync requests allow you to set a timeout= but the request may continue
 // beyond that and wedge forever, so we need to track how long we are willing
 // to keep open the connection. This constant is *ADDED* to the timeout= value
 // to determine the max time we're willing to wait.
-var BUFFER_PERIOD_MS = 5 * 1000;
+var BUFFER_PERIOD_MS = 30 * 1000;
 
 // if /sync fails, we call /versions in a loop until we get a response
 // (/versions typically returns much quicker than /sync, so this allows uys to
@@ -53,8 +61,8 @@ var BUFFER_PERIOD_MS = 5 * 1000;
 // KEEPALIVE_MIN_DELAY_MS and KEEPALIVE_MAX_MS before making another request.
 //
 var KEEPALIVE_MIN_DELAY_MS = 1000;
-var KEEPALIVE_TIMEOUT_MS = 2000;
-var KEEPALIVE_RANDOMNESS_MS = 500;
+var KEEPALIVE_TIMEOUT_MS = 4000;
+var KEEPALIVE_RANDOMNESS_MS = 2000;
 
 function getFilterName(userId, suffix) {
     // scope this on the user ID because people may login on many accounts
@@ -80,7 +88,7 @@ function SyncApi(client, opts) {
     opts = opts || {};
     opts.initialSyncLimit = opts.initialSyncLimit || 8;
     opts.resolveInvitesToProfiles = opts.resolveInvitesToProfiles || false;
-    opts.pollTimeout = opts.pollTimeout || (15 * 1000);
+    opts.pollTimeout = opts.pollTimeout || DEFAULT_SYNC_TIMEOUT_MS;
     opts.pendingEventOrdering = opts.pendingEventOrdering || "chronological";
     this.opts = opts;
     this._peekRoomId = null;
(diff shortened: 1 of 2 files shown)


lib/sync.js
```diff
@@ -36,7 +36,25 @@ var DEBUG = true;
 // beyond that and wedge forever, so we need to track how long we are willing
 // to keep open the connection. This constant is *ADDED* to the timeout= value
 // to determine the max time we're willing to wait.
-var BUFFER_PERIOD_MS = 80 * 1000;
+var BUFFER_PERIOD_MS = 5 * 1000;
```
ara4n: 5s seems a bit low here.  Is there a risk that the request an initial sync with a small (zero?) timeout - and then give only 5s for all the data to actually xfer down the pipe?

lib/sync.js
```diff
@@ -62,7 +80,7 @@ function SyncApi(client, opts) {
     opts = opts || {};
     opts.initialSyncLimit = opts.initialSyncLimit || 8;
     opts.resolveInvitesToProfiles = opts.resolveInvitesToProfiles || false;
-    opts.pollTimeout = opts.pollTimeout || (30 * 1000);
+    opts.pollTimeout = opts.pollTimeout || (15 * 1000);
```
ara4n: why no constant here?

lib/sync.js
```diff
@@ -36,7 +36,25 @@ var DEBUG = true;
...
+// factor attempts to avoid having all clients getting in sync, causing a
+// thundering herd.
+//
+// Each request to /versions is given a timeout of between KEEPALIVE_TIMEOUT_MS
+// and KEEPALIVE_TIMEOUT_MS+KEEPALIVE_RANDOMNESS_MS. If it times out, we make
+// another request immediately. If it fails, we delay between
+// KEEPALIVE_MIN_DELAY_MS and KEEPALIVE_MAX_MS before making another request.
+//
+var KEEPALIVE_MIN_DELAY_MS = 1000;
+var KEEPALIVE_TIMEOUT_MS = 2000;
```
ara4n: I wonder how easily we'll hit the 2s timeout on GPRS... perhaps 4s + rand(2s) might be a better compromise?

ara4n:
lgtm other than querying whether we may have swung too far the other way, as it'd suck to break everything when tethered to a crappy connection (e.g. tethered to a mobile on EDGE, which is an unfortunately common setup for me, at least).

I'm also a little confused as to why reducing the /versions timeout so massively is actually going to help us much for the bug in question - have we ever seen it get stuck on timing out /versions?  As opposed to the whole stuck-/sync failure mode?

lib/sync.js
richvdh: good question. added.

lib/sync.js
richvdh: initial sync gets the same timeout as incremental sync. So we'd actually be giving 20s to transfer all the data down the pipe (assuming the server responded immediately). Nevertheless this probably does need bumping somewhat. Let's try 30s as a compromise.

lib/sync.js
richvdh: I think I didn't think this through properly. Sure, let's try that.

richvdh:
> I'm also a little confused as to why reducing the /versions timeout so massively is actually going to help us much for the bug in question - have we ever seen it get stuck on timing out /versions? As opposed to the whole stuck-/sync failure mode?

Agreed, the timeout on /versions won't help us detect the fact that we've gone offline any quicker. Mostly this was an attempt to recover more quickly when we come back online; and it was an easy first step to see if it helped at all before I got into more drastic measures.

Reducing the timeout on /sync *should* help us detect offlineness quicker, but given that we can't reduce those timeouts beyond tens of seconds, this is only going to be marginally effective.

Anyway, ptal.

ara4n:
<M-matthew1> rich: on the timeout fun - are we going to be hard-cancelling /sync requests which don't return within 15s + 30s (buffer-period) timeout?
 <M-matthew1> 15s (timeout param) + 30s (buffer-period) even
 <M-matthew1> (am worried that this is going to suddenly make vector unusable on worse connections, especially for people with huge sync responses which necessarily take a while to trickle down the pipe)
 <M-richvdh> yes.
 <M-richvdh> and quite possibly.
 <M-richvdh> <shrug>
 <M-richvdh> maybe that whole PR is misguided
 <M-richvdh> as you say, it's basically addressing the wrong problem
 <M-matthew1> !calc 2MB/30s in KB/s
 -Squeaky/#matrix- Wolfram Alpha: 66.67 kB/s  (kilobytes per second) (≈ 0.53 × StarLAN speed ( 1 Mb/s ))
 <M-matthew1> so if the server takes 15s to respond (as it thrashes its SQL), it would then have to xfer the result at >60KB/s to get it down the pipe in time (for my 2MB initial sync, at least)
 <M-matthew1> humm
 <M-matthew1> sorry for flipflopping on this - the PR review just nudged me into thinking about the bad network scenario in more detail
 <M-richvdh> yeah np
 <M-richvdh> as I said, that PR was just a first stab
 <M-matthew1> plus i'm wondering if the 54s stall we saw in that HAR last night is even related to any of the existing timeouts
 <M-richvdh> not really
 <M-matthew1> my hunch would be to keep the PR in the wings, and see if it's actually needed after the time-jump thing lands

richvdh:
Some of the factoring-out-to-symbolic constants would have been good to save here, but for now let's close this in favour of https://github.com/matrix-org/matrix-js-sdk/issues/121.

PR #121, opened 2016-03-31
Implement HTTP callbacks in realtime

richvdh:
Hopefully this will improve our recovery time after a laptop is suspended. The
idea is to treat the timeouts on the http apis as being in realtime, rather
than in elapsed time while the machine is awake.

To do this, we add a layer on top of window.setTimeout. We run a callback every
second, which then checks the wallclock time and runs any pending callbacks.

(Hopefully a fix to https://github.com/vector-im/vector-web/issues/1288).

### DIFF
--- a/lib/http-api.js
+++ b/lib/http-api.js
@@ -21,6 +21,11 @@ limitations under the License.
 var q = require("q");
 var utils = require("./utils");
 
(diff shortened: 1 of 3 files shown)


dbkr:
lgtm

PR #122, opened 2016-04-03
URL previewing support

ara4n:
* Add client.getUrlPreview() method to wrap the new /preview_url endpoint, which attempts to get a preview of a URL as of a given point in time.
* Add a basic in-memory cache to avoid hammering synapse unnecessarily

This is part of a set of PRs spanning vector-web, matrix-react-sdk, matrix-js-sdk and synapse.
See also vector-im/vector-web#1343 and https://github.com/matrix-org/matrix-react-sdk/pull/260

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1450,11 +1450,15 @@ MatrixClient.prototype.getCurrentUploads = function() {
 };
 
 /**
- * Get a preview of the given URL as OpenGraph metadata.
+ * Get a preview of the given URL as of (roughly) the given point in time,
+ * described as an object with OpenGraph keys and associated values.
+ * Attributes may be synthesized where actual OG metadata is lacking.
  * Caches results to prevent hammering the server.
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -1449,6 +1450,37 @@ MatrixClient.prototype.getCurrentUploads = function() {
 };
 
 /**
+ * Get a preview of the given URL as OpenGraph metadata.
+ * Caches results to prevent hammering the server.
+ * @param {string} url The URL to get preview data for
+ * @param {Number} ts The timestamp (ms since epoch) for querying
+ * historical preview data.
```
richvdh: this is a bit vague. Not everyone should be expected to read the react-sdk source to figure out what it means. Can you clarify how it is meant to be used?

ara4n:
@richvdh PTAL

richvdh:
lgtm

PR #123, opened 2016-04-06
Fix a bug where we recreated sync filters

richvdh:
Fix the object comparison used for client filters (JSON.stringify is
non-deterministic)


dbkr:
I'd been attempting to not include an attempt at deep comparison, but agree that JSON.stringify() is not a good solution (have you seen it fail in practice?) and domain-specific filter comparison doesn't seem terribly appealing either, so this is probably the least bad option.

lgtm once the test passes

richvdh:
Yes, my client was creating a new filter every time I opened the "Historical" rooms

PR #125, opened 2016-04-12
Add methods to get (and set) pushers

dbkr:
Requires https://github.com/matrix-org/synapse/pull/716 for the methods to work

Get pusher API added in https://github.com/matrix-org/matrix-doc/pull/308/files


richvdh:
lgtm

PR #126, opened 2016-04-14
Clean up test shutdown

richvdh:
Make sure that the integration tests actually kill off all of their timers, so
that jasmine exits cleanly.

This probably also fixes https://github.com/vector-im/vector-web/issues/1365.


dbkr:
lgtm

PR #128, opened 2016-04-18
Track the absolute time that presence events are received, so that the relative lastActiveAgo value is meaningful.

ara4n:
This solves SYJS-28

### DIFF
--- a/lib/models/room.js
+++ b/lib/models/room.js
@@ -609,6 +609,7 @@ Room.prototype._addEventToTimeline = function(event, timeline, toStartOfTimeline
  * @private
  */
 Room.prototype._addLiveEvents = function(events) {
+    // var now = Date.now();
     for (var i = 0; i < events.length; i++) {
         if (events[i].getType() === "m.room.redaction") {
             var redactId = events[i].event.redacts;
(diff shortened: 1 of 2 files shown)


lib/models/room.js
```diff
@@ -695,6 +695,16 @@ Room.prototype._addLiveEvent = function(event, duplicateStrategy) {
...
+
+        // also, any live events from a user should be taken as implicit
+        // presence information: evidence that they are currently active.
+        // ...except in a world where we use 'user.currentlyActive' to reduce
+        // presence spam, this isn't very useful - we'll get a transition when
+        // they are no longer currently active anyway. so comment it out for now.
+
+        // var user = this.currentState.getMember(events[i].sender.userId);
+        // user.lastActiveAgo = 0;
+        // user.lastPresenceTs = Date.now();
```
Kegsay: If we don't plan to ever reinstate this code, please delete it.

lib/models/user.js
```diff
@@ -30,7 +30,12 @@ limitations under the License.
  * @prop {string} displayName The 'displayname' of the user if known.
  * @prop {string} avatarUrl The 'avatar_url' of the user if known.
  * @prop {string} presence The presence enum if known.
- * @prop {Number} lastActiveAgo The last time the user performed some action in ms.
+ * @prop {Number} lastActiveAgo The time elapsed in ms since the user interacted
+ *                proactively with the server, or we saw a message from the user
+ * @prop {Number} lastPresenceTs Timestamp (ms since the epoch) for when we last
+ *                received presence data for this user.  We can subtract
+ *                lastActiveAgo from this to approximate an absolute value for
+ *                when a user was last active.
```
Kegsay: It would be useful to have a method on `User` to do this calculation rather than just mentioning it in the docs.

Kegsay:
LGTM aside from comments.

ara4n:
@Kegsay PTAL

ara4n:
merging as i want to ship this and believe i addressed K's comments

PR #133, opened 2016-05-06
Add support for the openid interface

dbkr:


richvdh:
lgtm

is this endpoint specced?

dbkr:
Not yet, suspect either mark or I (probably mark) should spec it at some point.

PR #134, opened 2016-06-07
Various matrix-client refactorings and fixes

richvdh:
A bunch of work to try and clean up some of the crypto code in MatrixClient.

Best reviewed commit by commit.

### DIFF
--- a/lib/OlmDevice.js
+++ b/lib/OlmDevice.js
@@ -1,0 +1,329 @@
+/*
+Copyright 2016 OpenMarket Ltd
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+"use strict";
+
+var Olm = require("olm");
+var utils = require("./utils");
+
+/**
+ * Manages the olm cryptography functions. Each OlmDevice has a single
+ * OlmAccount and a number of OlmSessions.
+ *
+ * Accounts and sessions are kept pickled in a sessionStore.
+ *
+ * @constructor
+ * @param {Object} sessionStore A store to be used for data in end-to-end
+ *    crypto
+ *
+ * @property {string} deviceCurve25519Key   Curve25519 key for the account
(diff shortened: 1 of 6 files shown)


lib/client.js
```diff
@@ -41,10 +41,8 @@ var SCROLLBACK_DELAY_MS = 3000;
 var CRYPTO_ENABLED = false;
 
 try {
-    var Olm = require("olm");
-    if (Olm.Account && Olm.Session) {
-        CRYPTO_ENABLED = true;
-    }
+    var OlmDevice = require("./OlmDevice");
+    CRYPTO_ENABLED = true;
```
dbkr: Does not checking for the properties mean it will fail if you stub out the olm library? (eg. telling webpack to replace it with a stub)

lib/client.js
```diff
@@ -69,10 +67,18 @@ var OLM_ALGORITHM = "m.olm.v1.curve25519-aes-sha2";
...
+ *
+ * @param {Object=} opts.store The data store to use. If not specified,
  * this client will not store any HTTP responses.
  *
+ * @param {string=} opts.deviceId A unique identifier for this device, for use
+ *    in end-to-end crypto. If not specified, end-to-end crypto will be disabled.
+ *
+ * @param {Object=} opts.sessionStore A store to be used for end-to-end crypto
+ *    session data. If not specified, end-to-end crypto will be disabled.
+ *
```
dbkr: Could we also say what the client should pass? (A WebStorageSessionStore, presumably?)

lib/OlmDevice.js
```diff
@@ -0,0 +1,329 @@
...
+ * Signs a message with the ed25519 key for this account.
+ *
+ * @param {string} message  message to be signed
+ * @return {string} base64-encoded signature
+ */
+OlmDevice.prototype.sign = function(message) {
+    var account = new Olm.Account();
+    try {
+        var pickledAccount = this._sessionStore.getEndToEndAccount();
+        account.unpickle(this._pickleKey, pickledAccount);
```
dbkr: This is happening quite a lot. I guess you need the try / catch boilerplate everywhere because you need to create & free the account each time, but could the getting & unpickling be wrapped up in a function?

dbkr:
Otherwise, lgtm

lib/client.js
richvdh: Well, I suppose so. But we don't actually do that (we use the IgnorePlugin, which just turns the require line into a `throw`).

lib/client.js
richvdh: Ideally it would be some interface which WebStorageSessionStore implements, and which I can't be bothered to write. But ok, I'll make this clearer.

lib/OlmDevice.js
richvdh: Done 9cb032ec446ac748ea05cf5adcc2711b5605baa2

PR #135, opened 2016-06-08
Support for marking devices as verified

richvdh:
Changes the format of data stored for user devices in the session store (in a
backwards-compatible way), to include a 'verified' key (and get rid of a lot of
extraneous crap we don't use).

When we decrypt an event, keep the content we need to do a verification in the
MatrixEvent.

Add methods which allow us to flag a device as verified, and to verify the
sender of an event.


richvdh:
wrong branch

PR #136, opened 2016-06-08
Support for marking devices as verified

richvdh:
Changes the format of data stored for user devices in the session store (in a
backwards-compatible way), to include a 'verified' key (and get rid of a lot of
extraneous crap we don't use).

When we decrypt an event, keep the content we need to do a verification in the
MatrixEvent.

Add methods which allow us to flag a device as verified, and to verify the
sender of an event.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -130,10 +130,15 @@ function MatrixClient(opts) {
         signatures[opts.userId]["ed25519:" + this.deviceId] = signature;
         this.deviceKeys.signatures = signatures;
 
+        var deviceInfo = {
+            keys: this.deviceKeys.keys,
+            algorithms: this.deviceKeys.algorithms,
+            verified: true,
+        };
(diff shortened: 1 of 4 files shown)


lib/client.js
```diff
@@ -444,6 +542,84 @@ MatrixClient.prototype.listDeviceKeys = function(userId) {
...
+    dev.verified = true;
+    this.sessionStore.storeEndToEndDevicesForUser(userId, devices);
+};
+
+
+/**
+ * Check if the sender of an event is verified
+ *
+ * @param {MatrixEvent} event event to be checked
+ * @return {boolean} true if the sender of this event has been verified
```
dbkr: You could say, "if the device has been marked as verified using setDeviceVerified", just to make it clear the two APIs are linked?

dbkr:
lgtm

richvdh:
@dbkr: quick update to support emitting an event as discussed. ptal?

dbkr:
lgtm

PR #137, opened 2016-06-09
Catch exceptions when encrypting events

richvdh:
If an exception was thrown by the encryption process, the event would be queued, but the exception would not be handled. This meant that the event got stuck as a grey 'sending' event in the UI.

Fixing this correctly is slightly more complex than just handling the exception correctly. A naive approach would mean that the event would be shown as a red 'unsent' message, and clicking 'resend' would then send the message *in the clear*. Hence, move the encryption to _sendEvent, where it will be called again
when the event is resent.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -692,9 +692,11 @@ MatrixClient.prototype.setRoomEncryption = function(roomId, config) {
                         }
                     }
                     if (oneTimeKey) {
-                        self._olmDevice.createOutboundSession(
+                        var sid = self._olmDevice.createOutboundSession(
                             device[2], oneTimeKey
                         );
+                        console.log("Started new sessionid " + sid +
+                                    " for device " + device[2]);
                     } else {
                         missing[device[0]] = missing[device[0]] || [];
                         missing[device[0]].push([device[1]]);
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -1115,26 +1117,56 @@ MatrixClient.prototype.sendEvent = function(roomId, eventType, content, txnId,
...
+        return;
+    }
+
+    var roomId = event.getRoomId();
+
+    var e2eRoomInfo = client.sessionStore.getEndToEndRoom(roomId);
+    if (!e2eRoomInfo || !e2eRoomInfo.algorithm) {
+        // not encrypting messages in this room
+        return;
+    }
```
dbkr: Could be wise to wrap up these two 'are we encrypting this room' checks into a single function that will be the same one you can to determine whether there's a padlock or whatever displayed in the room. These two checks make me wonder how easy it would be to show a room as being encrypted and then accidentally end up without a sessionStore which here would cause the message to be sent in the clear.

dbkr:
otherwise lgtm

lib/client.js
richvdh: Yeah. There are some more changes afoot, which move the is-the-room-being-encrypted configuration out of the sessionstore and into the room state, which will make this look a bit different anyway. I think I'll leave this for now.

PR #143, opened 2016-06-21
crypto: use memberlist to derive recipient list

richvdh:
When we send an encrypted message with Olm, we need to know who to send it
to. Currently that is based on a user list passed to setRoomEncryption. We want
to be able to change the list as people join and leave the room; I also want to
not have to keep the member list in local storage.

So, use the member list at the point of enabling encryption to set up sessions,
and at the point of sending a message to encrypt the message.

Further work here includes:

 * updating the react-sdk not to bother setting the member list
 * monitoring for new users/devices in the room, and setting up new sessions
   with them

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -643,90 +643,116 @@ MatrixClient.prototype.setRoomEncryption = function(roomId, config) {
         return q.reject(new Error("End-to-End encryption disabled"));
     }
 
+    var self = this;
+
     if (config.algorithm === OLM_ALGORITHM) {
-        if (!config.members) {
-            throw new Error(
-                "Config must include a 'members' list with a list of userIds"
-            );
-        }
-        var devicesWithoutSession = [];
-        var userWithoutDevices = [];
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -643,90 +643,116 @@ MatrixClient.prototype.setRoomEncryption = function(roomId, config) {
...
+        var query = queries[device[0]] || {};
+        queries[device[0]] = query;
+        query[device[1]] = "curve25519";
+    }
+    var path = "/keys/claim";
+    var content = {one_time_keys: queries};
+    var self = this;
+    return this._http.authedRequestWithPrefix(
+        undefined, "POST", path, undefined, content,
+        httpApi.PREFIX_UNSTABLE
```
dbkr: /keys/claim is still unstable?

richvdh: We can always grep for PREFIX_UNSTABLE, so we think this would be a bit redundant.

PR #145, opened 2016-06-23
Configure encryption on m.room.encryption events

richvdh:
This is the first step in having a cross-room "enable encryption" button.

If encryption is enabled, add an event handler which will set up encryption
when we receive an m.room.encryption event (either at initial /sync, or in
subsequent syncs.)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -641,7 +641,7 @@ MatrixClient.prototype.isEventSenderVerified = function(event) {
  */
 function setupCryptoEventHandler(client) {
     client.on("event", function(event) {
-        if (event.getType() != "m.room.encryption") {
+        if (!event.isState() || event.getType() != "m.room.encryption") {
             return;
         }
         onCryptoEvent(client, event);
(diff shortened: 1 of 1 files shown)


richvdh:
When I say "first step", I mean "we already landed the second half of this change, at https://github.com/matrix-org/matrix-react-sdk/pull/321, but I forgot to send you the first half"

lib/client.js
```diff
@@ -633,19 +634,61 @@ MatrixClient.prototype.isEventSenderVerified = function(event) {
...
 
 /**
+ * Register a listener for m.room.encryption events which will enable encryption
+ * for a room.
+ *
+ * @param {MatrixClient} client
+ */
+function setupCryptoEventHandler(client) {
+    client.on("event", function(event) {
+        if (event.getType() != "m.room.encryption") {
```
dbkr: Should this check to see if it's a state event too? Otherwise could it get confused if you send a non-state m.room.encryption which enables encryption but then you connect with another client which doesn't enable crypto because the event has scrolled off the history?

richvdh: 'spose.

richvdh:
ptal

PR #147, opened 2016-07-06
Add register-specific request token endpoint

dbkr:
As per https://github.com/matrix-org/matrix-doc/pull/343


lib/client.js
```diff
@@ -2785,6 +2786,42 @@ MatrixClient.prototype.register = function(username, password,
 };
 
 /**
+ * Requests an email verification token for the purposes of registration.
+ * This API proxies the Identity Server /validate/email/requestToken API,
+ * adding registration-specific behaviour. Specifically, an account with
```
Kegsay: I assume you mean:
> Specifically, **if** an account with

lib/client.js
```diff
@@ -2785,6 +2786,42 @@ MatrixClient.prototype.register = function(username, password,
...
+ * the given email address already exists, it will either send an email
+ * to the address informing them of this or return M_THREEPID_IN_USE
+ * (which one is up to the Home Server).
+ *
+ * requestEmailToken calls the equivalent API directly on the ID server,
+ * therefore bypassing the registration-specific logic.
+ *
+ * @param {string} email
+ * @param {string} clientSecret
```
Kegsay: It would be good to describe what this should be given it isn't immediately obvious like say the `email` param is. E.g. how long should it be? Cryptographically random or is pseudo-random okay? Can I use non-ASCII? Smiley poo?

lib/client.js
```diff
@@ -2785,6 +2786,42 @@ MatrixClient.prototype.register = function(username, password,
...
+ * @param {string} sendAttempt
```
Kegsay: This should presumably be a `{Number}`. Start at 1? 0?

lib/client.js
```diff
@@ -2785,6 +2786,42 @@ MatrixClient.prototype.register = function(username, password,
...
+ * @param {string} nextLink Optional
```
Kegsay: This parameter is not documented on https://github.com/matrix-org/matrix-doc/pull/343/files - what does it do?

lib/client.js
```diff
@@ -2785,6 +2786,42 @@ MatrixClient.prototype.register = function(username, password,
...
+ * @param {module:client.callback} callback Optional.
+ * @return {module:client.Promise} Resolves: TODO
```
Kegsay: This resolves to the API response for `/register/email/requestToken`. We should probably just state that.

lib/client.js
```diff
@@ -2785,6 +2786,42 @@ MatrixClient.prototype.register = function(username, password,
...
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixClient.prototype.requestRegisterEmailToken = function(email, clientSecret,
+                                                    sendAttempt, nextLink, callback) {
+    var id_server_url = url.parse(this.idBaseUrl);
```
Kegsay: Last time I checked, `this.idBaseUrl` is an optional property on the `MatrixClient`. This will throw in this case:

```javascript
> url.parse()
TypeError: Parameter 'url' must be a string, not undefined
```

We should either handle this or document that it can throw.

lib/client.js
```diff
@@ -2785,6 +2786,42 @@ MatrixClient.prototype.register = function(username, password,
...
+
+    var params = {
+        client_secret: clientSecret,
+        email: email,
+        send_attempt: sendAttempt,
+        next_link: nextLink,
+        id_server: id_server_url.host,
```
Kegsay: We should probably do our due diligence and sanity check that `id_server_url.host` is not `null`:

```javascript
> var u = url.parse("foobar");
> u.host
null
```

lib/client.js
dbkr: I've docced all these on the original requestEmailToken

PR #149, opened 2016-07-14
Add more events to User

dbkr:
There was no way of observing changes to fields like currentlyActive, so add this and add one for lastPresenceTs that will be fired whenever we get a presence event.

This will hopefully help fix https://github.com/vector-im/vector-web/issues/1740

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Changes in \<unreleased>
 
  * Add room.getAliases() and room.getCanonicalAlias
  * Add API calls `/register/email/requestToken`, `/account/password/email/requestToken` and `/account/3pid/email/requestToken`
+ * Add more events for changes in fields on the User object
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ Changes in \<unreleased>
 
  * Add room.getAliases() and room.getCanonicalAlias
  * Add API calls `/register/email/requestToken`, `/account/password/email/requestToken` and `/account/3pid/email/requestToken`
+ * Add more events for changes in fields on the User object
```
Kegsay: State which events please.

lib/models/user.js
```diff
@@ -87,8 +87,15 @@ User.prototype.setPresenceEvent = function(event) {
     {
         eventsToFire.push("User.displayName");
     }
+    if (event.getContent().currently_active &&
```
Kegsay: If `currently_active` is a boolean, this will never fire.

Kegsay:
LGTM aside from comments.

PR #150, opened 2016-07-20
Support global account_data

ara4n:


lib/client.js
```diff
@@ -903,6 +903,37 @@ MatrixClient.prototype.getUsers = function() {
     return this.store.getUsers();
 };
 
+// User Account Data operations
+// ============================
+
+/**
+ * Set account data event for the current user.
+ * @param {string} eventType The event type
+ * @param {Object} the contents object for the event
```
Kegsay: For JSDoc to pick it up sanely,  the thing after the type has to be the variable name, hence this should be:
```
* @param {Object} contents the object for the event
```

lib/store/memory.js
```diff
@@ -244,7 +247,27 @@ module.exports.MatrixInMemoryStore.prototype = {
...
+    },
+
+    /**
+     * Store user-scoped account data events
+     * @param {Array<MatrixEvent>} events The events to store.
+     */
+    storeAccountDataEvents: function(events) {
+        var self = this;
+        events.forEach(function(event) {
+            self.accountData[event.getType()] = event;
```
Kegsay: I find this is somewhat surprising that `storeAccountDataEvents(Event[])` doesn't actually store all the events in the array if there are >1 events with the same `type`, but instead clobbers with the latest one. I assume this is intentional behaviour, so we should mention the clobbering semantics in the functions docs.

lib/store/memory.js
```diff
@@ -244,7 +247,27 @@ module.exports.MatrixInMemoryStore.prototype = {
...
+        });
+    },
+
+    /**
+     * Get account data event by event type
+     * @param {string} eventType The event type being queried
+     * @return {MatrixEvent} the user account_data event of given type
```
Kegsay: This is nullable, so `{?MatrixEvent}` please!

lib/client.js
```diff
@@ -4001,6 +4015,17 @@ module.exports.CRYPTO_ENABLED = CRYPTO_ENABLED;
  * @param {module:client~DeviceInfo} device information about the verified device
  */
 
+/**
+ * Fires whenever new user-scoped account_data is added.
+ * @event module:client~MatrixClient#"Room"
+ * @param {MatrixEvent} event The event describing the account_data just added
```
Kegsay: This is technically incorrect. The sync.js code will emit an `accountData` event *before* it is added to the store (and hence gettable from `MatrixClient.getAccountData`. If this is intentional, please update this line to state that this describes the `account_data` *about to be* added. If this is not intentional, then please tweak `sync.js:611` to emit events *after* they have been added to the store.

There isn't a right answer as to whether you should emit before or after adding to the store, it greatly depends on what the purpose is of the thing you're emitting. This only works if you emit **before** for example:
```javascript
MatrixClient.on("accountData", function(event) {
  if (event.getType() === "one.to.one.room") {
    // find the old 1:1 room
    var oldEvent = client.getAccountData("one.to.one.room");
  }
});
```

Kegsay:
Overall, **beautiful** :) just needs some clarification on docs.

ara4n:
@kegsay: PTAL

Kegsay:
LGTM

lib/client.js
```diff
@@ -4001,6 +4015,17 @@ module.exports.CRYPTO_ENABLED = CRYPTO_ENABLED;
  * @param {module:client~DeviceInfo} device information about the verified device
  */
 
+/**
+ * Fires whenever new user-scoped account_data is added.
+ * @event module:client~MatrixClient#"Room"
```
richvdh: should this be `module:client~MatrixClient#"accountData"` ?

lib/sync.js
```diff
@@ -605,7 +606,19 @@ SyncApi.prototype._processSyncResponse = function(syncToken, data) {
...
     }
 
-    // the returned json structure is abit crap, so make it into a
+    // handle non-room account_data
+    if (data.account_data && utils.isArray(data.account_data.events)) {
+        var events = data.account_data.events.map(client.getEventMapper());
+        client.store.storeAccountDataEvents(events);
+        events.forEach(
+            function(accountDataEvent) {
+                client.emit("accountData", accountDataEvent);
```
richvdh: it feels odd to me that our events to date have started with capital letters.

lib/client.js
Kegsay: Bah didn't spot that! Yes it should be.

PR #151, opened 2016-07-20
/login, /register: Add device_id and initial_device_display_name

richvdh:
To help test the forthcoming device_id support for /login and /register, add
the device_id and initial_device_display_name parameters to those calls. Allow
the app to specify the default device displayname when creating the client (as
well as the device_id).

Also, don't try initialising the Olm layer unless a userId is
provided. Currently this isn't a problem because react-sdk doesn't provide a
sessionStore when it doesn't provide a userId, but that is a bad thing to rely
on (and I am going to break it with a react-sdk PR).

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -80,8 +80,12 @@ var DeviceVerification = {
  * @param {Object=} opts.store The data store to use. If not specified,
  * this client will not store any HTTP responses.
  *
- * @param {string=} opts.deviceId A unique identifier for this device, for use
- *    in end-to-end crypto. If not specified, end-to-end crypto will be disabled.
+ * @param {string=} opts.deviceId A unique identifier for this device; used for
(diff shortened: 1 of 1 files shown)


lib/client.js
```diff
@@ -2675,9 +2683,19 @@ MatrixClient.prototype.paginateEventTimeline = function(eventTimeline, opts) {
...
 MatrixClient.prototype.login = function(loginType, data, callback) {
-    data.type = loginType;
+    var login_data = {
+        type: loginType,
+        device_id: this.deviceId,
+        initial_device_display_name: this.defaultDeviceDisplayName,
+    };
+
+    // merge data into login_data
+    for (var k in data) {
```
dbkr: I believe you can now use `for...of` to avoid having to do the `hasownProperty` mess

PR #156, opened 2016-07-29
Fix email registration

dbkr:
This would cause the request to 400 in the new vector that opens after you clicked the link in the email, as per the comment.


richvdh:
lgtm

PR #157, opened 2016-08-02
Add `logout`

dbkr:


richvdh:
1. any reason not to put this in MatrixBaseApis ?
2. the lint is sad about something

dbkr:
I was imagining this ought to go in MatrixBaseApis but the rest of the login stuff is in client so I kept it with that.

richvdh:
> I was imagining this ought to go in MatrixBaseApis but the rest of the login stuff is in client 

That's because `login` refers to `this.deviceId`, which, as we've discussed elsewhere, may be a mistake. I'd rather see `logout` go straight into `MatrixBaseApis`.

dbkr:
Ah, I see. Done.

PR #159, opened 2016-08-03
client.listDeviceKeys: Expose device display name

richvdh:
Synapse now (or will, after https://github.com/matrix-org/synapse/pull/978) returns the display_name in an e2e device query; we need to return it to the app.


NegativeMjark:
LGTM

PR #161, opened 2016-08-04
Refactor device key upload

richvdh:
Use another-json instead of awful manual json building. Sign the device keys at the point of upload, instead of having to keep the signed string in memory. Only upload device keys once (they are correctly merged with the one-time keys by synapse).

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -349,12 +349,12 @@ function _uploadDeviceKeys(client) {
         device_id: deviceId,
         keys: client._deviceKeys,
         user_id: userId,
-        signatures: {},
     };
 
+    var sig = client._olmDevice.sign(anotherjson.stringify(deviceKeys));
+    deviceKeys.signatures = {};
     deviceKeys.signatures[userId] = {};
-    deviceKeys.signatures[userId]["ed25519:" + deviceId] =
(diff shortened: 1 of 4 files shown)


lib/client.js
```diff
@@ -332,12 +331,42 @@ MatrixClient.prototype.uploadKeys = function(maxKeys, deferred) {
...
+
+    var userId = client.credentials.userId;
+    var deviceId = client.deviceId;
+
+    var deviceKeys = {
+        algorithms: client._cryptoAlgorithms,
+        device_id: deviceId,
+        keys: client._deviceKeys,
+        user_id: userId,
+        signatures: {},
```
NegativeMjark: I don't think we should include the signature key here. 
https://matrix.org/docs/spec/server_server/unstable.html#signing-json

richvdh: the ed25519 key? if we don't include it here, how will anyone get it?

NegativeMjark: The blob of JSON that you sign shouldn't contain the signature key. The signature key is added after signing as per the documentation I linked to.

richvdh:
@NegativeMjark: PTAL

NegativeMjark:
LGTM

PR #162, opened 2016-08-04
Factor crypto stuff out of MatrixClient

richvdh:
Introduce a new Crypto class which encapsulates all of the the crypto-related gubbins, replacing it with thin wrappers in MatrixClient.

I don't know if this is in any way reviewable :/


NegativeMjark:
I could believe that the new lines were copies of the deleted lines. In which case it would LGTM?

richvdh:
seems fair :)

thanks

PR #163, opened 2016-08-04
Verify e2e keys on download

richvdh:
Check the signature on downloaded e2e keys, and ignore those that don't match.


NegativeMjark:
LGTM

richvdh:
\o/

PR #164, opened 2016-08-05
Zero is a valid power level

dbkr:
So testing truthiness will lead to incorrect behaviour.

https://github.com/vector-im/vector-web/issues/1620


richvdh:
looks fine. Would benefit from a unit test?

PR #170, opened 2016-08-15
Update our push rules when they come down stream

dbkr:
Also expose a useful function from pushprocessor.

Fixes https://github.com/vector-im/vector-web/issues/1495

### DIFF
--- a/lib/pushprocessor.js
+++ b/lib/pushprocessor.js
@@ -18,28 +18,6 @@ limitations under the License.
  */
 
 /**
- * Convert a list of actions into a object with the actions as keys and their values
- * eg. [ 'notify', { set_tweak: 'sound', value: 'default' } ]
- *     becomes { notify: true, tweaks: { sound: 'default' } }
- * @param {array} actionlist The actions list
- *
- * @return {object} A object with key 'notify' (true or false) and an object of actions
- */
-function actionListToActionsObject(actionlist) {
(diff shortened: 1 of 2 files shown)


lib/pushprocessor.js
```diff
@@ -298,3 +306,11 @@ function PushProcessor(client) {
...
 /** The PushProcessor class. */
 module.exports = PushProcessor;
+
+/**
+ * See above
+ * @param {array} actionlist The actions list
+ *
+ * @return {object} See above
+ */
+module.exports.actionListToActionsObject = actionListToActionsObject;
```
richvdh: instead of this dance, why not define it directly with `PushProcessor.actionListToActionsObject = function(...) {...}` (and update the references elsewhere in the file accordingly)?

lib/sync.js
```diff
@@ -613,6 +613,9 @@ SyncApi.prototype._processSyncResponse = function(syncToken, data) {
         client.store.storeAccountDataEvents(events);
         events.forEach(
             function(accountDataEvent) {
+                if (accountDataEvent.getType() == 'm.push_rules') {
+                    client.pushRules = accountDataEvent.getContent();
```
richvdh: 1. ugh.

2. Does this make the update of `pushRules` in `MatrixClient.setRoomMutePushRule` redundant (or racy)?

PR #172, opened 2016-08-16
Factor Olm encryption/decryption out to new classes

richvdh:
- to make way for alternative encryption algorithms. We now store an encryption
object for each room, rather than referring to sessionstore on each event.

Also a little light tidying to the jsdocs.


NegativeMjark:
LGTM

PR #174, opened 2016-08-17
Megolm: don't dereference nullable object

richvdh:
It is possible for `room` to be null when passed to `MegolmEncryption.encryptMessage`; we need to avoid dereferencing it. Instead, make sure that the `EncryptionAlgorithm` knows about the roomId it is targeting, and use that.

Replace the increasingly-long argument list on the `EncryptionAlgorithm` constructor with a params list, and update `DecryptionAlgorithm` to match.

### DIFF
--- a/lib/crypto.js
+++ b/lib/crypto.js
@@ -625,7 +625,12 @@ Crypto.prototype.setRoomEncryption = function(roomId, config, roomMembers) {
     };
     this._sessionStore.storeEndToEndRoom(roomId, config);
 
-    var alg = new AlgClass(this._deviceId, this, this._olmDevice);
+    var alg = new AlgClass({
+        deviceId: this._deviceId,
+        crypto: this,
+        olmDevice: this._olmDevice,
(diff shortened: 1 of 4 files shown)


lib/crypto.js
```diff
@@ -772,7 +779,9 @@ Crypto.prototype.decryptEvent = function(event) {
     if (!AlgClass) {
         throw new algorithms.DecryptionError("Unable to decrypt " + content.algorithm);
     }
-    var alg = new AlgClass(this._deviceId, this, this._olmDevice);
+    var alg = new AlgClass({
+        olmDevice: this._olmDevice,
+    });
```
NegativeMjark: Why does this have fewer keys than above?

richvdh: The others weren't used, and were originally cargo-culted from EncryptionAlgorithm. I thought I would leave them out until I found a use for them, rather than including them in case they were useful.

NegativeMjark:
LGTM

PR #175, opened 2016-08-19
Move DeviceInfo and DeviceVerification to separate module

richvdh:
mostly because I'm fed up with hacking around the fact that DeviceVerification is hard to get at.


NegativeMjark:
LGTM

PR #176, opened 2016-08-19
Make encryption asynchronous

richvdh:
We're going to need to send out a load of messages to distribute the megolm
keys; as a first step, deal with the asynchronicity this will require.


NegativeMjark:
LGTM

PR #177, opened 2016-08-22
Factor out a function for doing olm encryption

richvdh:
Make a library file with some constants and a function to pack olm-encrypted
events (which we are going to use from megolm)

### DIFF
--- a/lib/OlmDevice.js
+++ b/lib/OlmDevice.js
@@ -294,6 +294,22 @@ OlmDevice.prototype.getSessionIdsForDevice = function(theirDeviceIdentityKey) {
     return utils.keys(sessions);
 };
 
+/**
+ * Get the right olm session id for encrypting messages to the given identity key
+ *
+ * @param {string} theirDeviceIdentityKey Curve25519 identity key for the
+ *     remote device
+ * @return {string?}  session id, or null if no established session
+ */
+OlmDevice.prototype.getSessionIdForDevice = function(theirDeviceIdentityKey) {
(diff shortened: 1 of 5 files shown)


NegativeMjark:
I'm not sure I really understand the module structure going on here. But it looks like valid javascript?

richvdh:
> I'm not sure I really understand the module structure going on here

Nor am I, to be honest. It's evolved out of everything being dumped in MatrixClient, and I'm not entirely satisfied with the results.

Currently we have:

 * Crypto - supposedly the interface which MatrixClient uses to do all of its cryptography bits. Delegates to:
 * crypto-algorithms: implementations of event encryption and decryption
 * OlmDevice - wrapper around olm.js library; nothing matrix-specific here. (Frankly it probably ought to be in olm.js).

The problem is that Crypto has ended up as a dumping ground instead, and includes a bunch of methods which only really exist so that crypto-algorithms implementations can call back to them; that means both that the MatrixClient-facing interface is obscured by the other crap, and that you end up with circularity in the modules. Hence deciding we need yet another module for stuff that is common between the olm and megolm crypto-algorithms interfaces.

Your suggestions for making this better are very welcome.

> But it looks like valid javascript?

Do you think it is correct javascript?

NegativeMjark:
Okay so ``crypto.js -> crypto-algorithms/*.js -> olmlib.js -> OlmDevice.js -> olm.js``

NegativeMjark:
I guess that's okay. LGTM.

richvdh:
ja

PR #178, opened 2016-08-22
Change the result structure for ensureOlmSessionsForUsers

richvdh:
Nothing was using the results (except the tests), and it's more useful to have
the devices we *do* have a session for than the ones we don't.


NegativeMjark:
LGTM

PR #179, opened 2016-08-22
Send out megolm keys when we start a megolm session

richvdh:
For now, pending the arrival of SPEC-138, this happens via inline messages in the room.

(Requires https://github.com/matrix-org/matrix-js-sdk/pull/178)

### DIFF
--- a/lib/crypto-algorithms/megolm.js
+++ b/lib/crypto-algorithms/megolm.js
@@ -82,7 +82,7 @@ MegolmEncryption.prototype._ensureOutboundSession = function(room) {
 
     // send the keys to each (unblocked) device in the room.
     var payload = {
-        type: "m.key",
+        type: "m.room_key",
         content: {
             algorithm: olmlib.MEGOLM_ALGORITHM,
             room_id: this._roomId,
--- a/lib/crypto-algorithms/megolm.js
+++ b/lib/crypto-algorithms/megolm.js
@@ -46,10 +46,12 @@ utils.inherits(MegolmEncryption, base.EncryptionAlgorithm);
 /**
  * @private
  *
+ * @param {module:models/room} room
+ *
  * @return {module:client.Promise} Promise which resolves when setup is
  *   complete.
  */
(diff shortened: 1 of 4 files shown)


lib/crypto-algorithms/megolm.js
```diff
@@ -80,28 +80,96 @@ MegolmEncryption.prototype._ensureOutboundSession = function() {
...
+            for (var deviceId in devices) {
+                if (!devices.hasOwnProperty(deviceId)) {
+                    continue;
+                }
+
+                var deviceInfo = devices[deviceId].device;
+                var encryptedContent = olmlib.encryptMessageForDevices(
+                    self._deviceId,
+                    self._olmDevice,
+                    [deviceInfo.getIdentityKey()],
```
NegativeMjark: Why not just send one in-band message for all the devices?

richvdh: Yeah, we will, hence the need for retries.

My main concern was around how many copies of the key we can fit in a single event; it's certainly not valid to assume that one will be sufficient. And we'll need to update it anyway to use OOB events.

So the question really becomes: how should this work when OOB events land? Should we be sending an OOB event per device, or grouping them together in some way? We could maybe group them by user, but it's not even valid to assume that all the messages for a single user will fit in a single event.

NegativeMjark: I'd assume that the OOB api will have a way to send messages in batches. If this was going to be a long term solution then I'd suggest sending the keys in batches known to be smaller than the max event size of 64K. Eh I guess it will do as a hack for now. Could you add some comments to explain why it doesn't work?

richvdh: sorry, why what doesn't work?

NegativeMjark: Both why sending them in one batch doesn't work because you hit the event size limits and why sending them individually doesn't work because you hit C+S API rate limits.

richvdh: k

NegativeMjark:
LGTM

PR #181, opened 2016-08-24
Reinstate device blocking for simple Olm

richvdh:
Commit 4cde51b3 broke device blocking such that we were encrypting for all
devices, including blocked ones. Reinstate it, and add a test.


NegativeMjark:
LGTM

PR #182, opened 2016-08-24
Put all of the megolm keys in one room message

richvdh:
Avoid hitting the rate-limiter by putting all of the megolm keys in a single
event.


NegativeMjark:
LGTM

PR #184, opened 2016-08-30
Fix exceptions when dealing with redactions

richvdh:
When we got a redaction event, we were adding the entire (circular) MatrixEvent
object for the redaction to the redacted event, which would then cause
exceptions down the line (particularly when dealing with gappy timelines).

We should only be adding the raw event.

Fixes (hopefully) https://github.com/vector-im/vector-web/issues/1389.


Kegsay:
LGTM

PR #185, opened 2016-08-31
Fix exceptions where HTMLMediaElement loads and plays race

ara4n:


lib/webrtc/call.js
```diff
@@ -74,6 +74,12 @@ MatrixCall.ERR_LOCAL_OFFER_FAILED = "local_offer_failed";
  */
 MatrixCall.ERR_NO_USER_MEDIA = "no_user_media";
 
+/**
+ * Lookup from opaque queue ID to a promise for media element operations that
+ * need to be serialised into a given queue
+ */
+MatrixCall.mediaPromises = {};
```
Kegsay: Why are you attaching this to the prototype rather than being an instance member of the `MatrixCall` like the other fields are?

lib/webrtc/call.js
```diff
@@ -145,6 +151,64 @@ MatrixCall.prototype.placeScreenSharingCall =
 };
 
 /**
+ * Play the given HTMLMediaElement, serialising the operation into a chain
+ * of promises to avoid racing access to the element
+ * @param {Element} HTMLMediaElement to play
+ * @param {string} Arbitrary ID to track the chain of promises to be used
```
Kegsay: The format for JSDoc is:
```
@param {Type} argName Some description
```
The `argName` is missing here. This is important because `npm run gendoc` makes HTML output based on the JSDoc here which is shipped to gh-pages on releases.

lib/webrtc/call.js
Kegsay: Also, we're using this as a dictionary where an external entity (the caller) can set arbitrary keys. We should make this a proper map then using `Object.create(null)` (an object without a prototype) else you can get [fun bugs](https://github.com/matrix-org/matrix-appservice-irc/issues/152) if people decide to use "hasOwnProperty" as a queue ID. It's unlikely to happen in this particular case, but given it's a direct replacement and everything else remains the same, it's trivial to do.

lib/webrtc/call.js
```diff
@@ -712,7 +784,7 @@ MatrixCall.prototype._onAddStream = function(event) {
         t.onstarted = hookCallback(self, self._onRemoteStreamTrackStarted);
     });
 
-    event.stream.onended = hookCallback(self, self._onRemoteStreamEnded);
+    event.stream.oninactive = hookCallback(self, self._onRemoteStreamEnded);
```
Kegsay: You didn't mention about changing this in the PR description, what's the deal? I'm guessing they changes `onended` to `oninactive`, and have deprecated `onended`? Should we be checking for the presence of `oninactive` to accomodate older browsers?

lib/webrtc/call.js
```diff
@@ -145,6 +151,64 @@ MatrixCall.prototype.placeScreenSharingCall =
...
+ * Play the given HTMLMediaElement, serialising the operation into a chain
+ * of promises to avoid racing access to the element
+ * @param {Element} HTMLMediaElement to play
+ * @param {string} Arbitrary ID to track the chain of promises to be used
+ */
+MatrixCall.prototype.playElement = function(element, queueId) {
+    if (MatrixCall.mediaPromises[queueId]) {
+        MatrixCall.mediaPromises[queueId] =
+            MatrixCall.mediaPromises[queueId].then(function() {
+                return element.play();
```
Kegsay: I'm worried this is creating a memory leak.

We never nuke the promises in the map, and I don't know enough about how Q (in this case) stores chained promises to know whether or not they get cleaned up after use (I suspect not). This means we may be holding onto references to `element` in the middle of a promise chain.

I don't expect you to do anything right now for this (you can heap dump a few times if you really want to find out if it is leaking, the deltas on the `HTMLMediaElement` types will clearly show it), but heads up either way.

Kegsay:
LGTM aside from comments.

Kegsay:
LGTM

ara4n:
ironically there was a horrific bug here; by discarding failed promises, the whole promise chain will break forever if a play() fails.  Turns out this is quite common, as things like unmounting the `<video/>` tag whilst its play() is waiting for media to arrive causes the play()'s promise to fail.  This is easy to do when toggling between rooms whilst waiting for calls to start, etc.

Hopefully this is fixed now and so going to merge (with a bunch of debugging in case of future problems)

PR #186, opened 2016-08-31
always play audio out of the remoteAudioElement if it exists.

ara4n:
fixes https://github.com/vector-im/vector-web/issues/1271 and https://github.com/vector-im/vector-web/issues/621


Kegsay:
LGTM

PR #187, opened 2016-09-01
Update to fixed olmlib

richvdh:
Olm 1.0.0 had broken OlmAccounts; upgrade to a more recent one.

When we try to restore the old OlmAccount, we'll get an error; detect this and
create a new OlmAccount.

Because we do this, we will see key changes on other devices; we want to
tolerate this, so we define a new algorithm tag,
m.olm.v2.curve25519-aes-sha2. If we see a device which previously didn't
support that algorithm, but now does, we tolerate a key change (but reset the
device to unverified).

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,6 @@
     "uglifyjs": "^2.4.10"
   },
   "optionalDependencies": {
-    "olm": "https://matrix.org/packages/npm/olm/olm-1.0.0.tgz"
+    "olm": "https://matrix.org/packages/npm/olm/olm-1.1.0.tgz"
   }
 }


lib/crypto-algorithms/olm.js
```diff
@@ -185,4 +185,5 @@ OlmDecryption.prototype.decryptEvent = function(event) {
     }
 };
 
-base.registerAlgorithm(olmlib.OLM_ALGORITHM, OlmEncryption, OlmDecryption);
+base.registerAlgorithm(olmlib.OLM_V2_ALGORITHM, OlmEncryption, OlmDecryption);
```
NegativeMjark: Why do you need a OLM_V2_ALGORITHM here? I don't think anything in the olm ratchet changed?

richvdh: see the PR description...

NegativeMjark: Can we just force a logout/login cycle on the clients to give them a new device-ID instead? I'm really not happy about bumping an unrelated crypto algorithm version number here.

richvdh:
So we've decided we're happy to force people with broken OlmAccounts to log back in, in which case, this is a bit easier...

Also needs a corresponding update to vector-web, but I might just land that on develop.

NegativeMjark:
LGTM

PR #188, opened 2016-09-02
Olm-related cleanups

richvdh:
A couple of small refactors which fell out of the aborted stuff for upgrading
to secure Ed25519 keys, but are useful in their own right.

The main functional change here is to calculate the "algorithms" list from
DECRYPTION_CLASSES (and hence include megolm in the list).


NegativeMjark:
LGTM

PR #189, opened 2016-09-04
Add debug methods to get the state of OlmSessions

richvdh:
I've been trying to track down issues with the OlmSessions getting out of sync
between two devices. To help with this, add a method which can be used from the
JS console to inspect the state of OlmSessions.

(Requires https://github.com/matrix-org/olm-backup/pull/15 to actually work)


NegativeMjark:
LGTM

PR #190, opened 2016-09-04
MatrixClient.getStoredDevicesForUser

richvdh:
Implement MatrixClient.getStoredDevicesForUser which uses
Crypto.getStoredDevicesForUser, which is more powerful than listDeviceKeys, and
isn't deprecated.

Also a couple of accessors for DeviceInfo.


NegativeMjark:
LGTM

PR #192, opened 2016-09-06
Fix sender verification for megolm messages

richvdh:
Turns out all we need to do is to make sure we use the Olm device table when we
look up megolm senders.


NegativeMjark:
LGTM

PR #194, opened 2016-09-07
Use to-device events for key sharing

richvdh:
Synapse now supports out-of-band messages, so use them instead of sending the
key-sharing messages in-band.


NegativeMjark:
LGTM

PR #195, opened 2016-09-07
Don't create Olm sessions proactively

richvdh:
In what I hoped would be a five-minute refactor to help clean up an annoying not-really-used codepath, but turned into a bit of a hackathon on the tests, create Olm sessions lazily in Olm rooms, just as we do in megolm rooms, which allows us to avoid having to get the member list before configuring e2e in a room.


NegativeMjark:
LGTM

PR #196, opened 2016-09-07
Introduce EventTimelineSet, filtered timelines and global notif timeline.

ara4n:
* Implements basic support for client-side filtering (ported from synapse)
* Splits the bits of Room which tracks timelines of events into a new class called EventTimelineSet.
* Rooms can now have multiple EventTimelineSets - one unfiltered one, and several filtered ones (e.g. for displaying FilePanels).
* Add the concept of a global Notification EventTimelineSet, used for backing the NotificationPanel
* Make TimelineWindow take an EventTimelineSet rather than a Room.


lib/client.js
```diff
@@ -1766,6 +1793,88 @@ MatrixClient.prototype.paginateEventTimeline = function(eventTimeline, opts) {
     return promise;
 };
 
+
+/**
+ * Take an EventTimeline, and backfill results from the notifications API.
+ * In future, the notifications API should probably be replaced by /messages
+ * with a custom filter or something - so we don't feel too bad about this being
+ * cargoculted from paginateEventTimeLine.
```
richvdh: hrm.

lib/filter-component.js
```diff
@@ -0,0 +1,141 @@
...
+        return actual_value === filter_value;
+    }
+}
+
+/**
+ * FilterComponent is a section of a Filter definition which defines the
+ * types, rooms, senders filters etc to be applied to a particular type of resource.
+ * This is all ported over from synapse's Filter object.
+ *
+ * N.B. that synapse refers to these as 'Filters', and what js-sdk refers to as
```
richvdh: So why are we calling this a FilterComponent and doubling the size of Filter, rather than fixing the inconsistency?

lib/models/event-timeline-set.js
```diff
@@ -0,0 +1,654 @@
...
+    this.room = room;
+
+    this._timelineSupport = Boolean(opts.timelineSupport);
+    this._liveTimeline = new EventTimeline(this);
+
+    // just a list - *not* ordered.
+    this._timelines = [this._liveTimeline];
+    this._eventIdToTimeline = {};
+
+    this._filter = opts.filter;
```
richvdh: you probably want to set this to `null` rather than `undefined` if opts.filter is `undefined`, if nothing else to make getFilter match its jsdoc

lib/models/event-timeline-set.js
```diff
@@ -0,0 +1,654 @@
...
+    var eventId = event.getId();
+    timeline.addEvent(event, toStartOfTimeline);
+    this._eventIdToTimeline[eventId] = timeline;
+
+    var data = {
+        timeline: timeline,
+        liveEvent: !toStartOfTimeline && timeline == this._liveTimeline,
+        timelineSet: this,
+    };
+    this.emit("Room.timeline", event, this.room,
```
richvdh: Do we want to be firing this for each timelineset? If we do, should the timelineset be included in the parameters?

lib/models/event-timeline-set.js
```diff
@@ -0,0 +1,654 @@
...
+ * matrixClient.on("Room.timeline",
+ *                 function(event, room, toStartOfTimeline, removed, data) {
+ *   if (!toStartOfTimeline && data.liveEvent) {
+ *     var messageToAppend = room.timeline.[room.timeline.length - 1];
+ *   }
+ * });
+ */
+
+/**
+ * Fires whenever the live timeline in a room is reset.
```
richvdh: I think this is out of place - it should still be in `room.js` ?

lib/models/event-timeline.js
```diff
@@ -25,16 +25,17 @@ var MatrixEvent = require("./event").MatrixEvent;
...
  */
-function EventTimeline(roomId) {
-    this._roomId = roomId;
+function EventTimeline(eventTimelineSet) {
+    this._eventTimelineSet = eventTimelineSet;
+    this._roomId = eventTimelineSet.roomId;
     this._events = [];
     this._baseIndex = 0;
-    this._startState = new RoomState(roomId);
+    this._startState = new RoomState(this._roomId);
```
richvdh: presumably `this._roomId` is going to be null for the notifications timeline. `RoomState`'s jsdoc explicitly says that the roomId is required, and a brief skim through the implementation doesn't convince me that it is lying. Have you checked what will happen here?

lib/models/room.js
```diff
@@ -254,39 +223,59 @@ Room.prototype._fixUpLegacyTimelineFields = function() {
...
-    return (res === undefined) ? null : res;
+    return this.getUnfilteredTimelineSet().getTimelineForEvent(eventId);
+};
+
+/**
+ * Add a new timeline to this room's unfiltered timeline set
+ *
+ * @return {module:models/event-timeline~EventTimeline} newly-created timeline
+ */
+Room.prototype.addTimeline = function() {
```
richvdh: I think we can vape this?

lib/models/room.js
```diff
@@ -438,245 +456,94 @@ Room.prototype.getCanonicalAlias = function() {
...
-    var lastEventWasNew = false;
-    for (var i = 0; i < events.length; i++) {
-        var event = events[i];
-        var eventId = event.getId();
+    // XXX: This is risky as our timeline
+    // may have grown huge and so take a long time to filter.
+    // see https://github.com/vector-im/vector-web/issues/2109
 
-        var existingTimeline = this._eventIdToTimeline[eventId];
+    var unfilteredLiveTimeline = this.getUnfilteredTimelineSet().getLiveTimeline();
```
richvdh: I feel like `s/this.getUnfilteredTimelineSet().getLiveTimeline()/this.getLiveTimeline()` throughout this file would make things rather more concise.

lib/models/room.js
```diff
@@ -438,245 +456,94 @@ Room.prototype.getCanonicalAlias = function() {
...
-            // RoomMember objects which are then attached to other events
-            // (in the sender and target fields). We should get those
-            // RoomMember objects to update themselves when the events that
-            // they are based on are changed.
+        for (i = 0; i < this._timelineSets.length; i++) {
+            var timelineSet = this._timelineSets[i];
+            // if we know about this event, redact its contents now.
+            var redactedEvent = timelineSet.findEventById(redactId);
+            if (redactedEvent) {
+                redactedEvent.makeRedacted(event);
```
richvdh: You're sharing the `MatrixEvent` between the timelineSets, so you probably only want to do this once (for the unfiltered timeline, presumably). Not sure that doing it multiple times is going to hurt anything, but it's misleading at best.

lib/models/room.js
```diff
@@ -1013,95 +869,17 @@ Room.prototype.removeEvents = function(event_ids) {
...
-        if (tl === timeline2) {
-            // timeline2 is before timeline1
-            return 1;
-        }
-        tl = tl.getNeighbouringTimeline(EventTimeline.BACKWARDS);
-    }
-
-    // the timelines are not contiguous.
-    return null;
+    return removedAny;
```
richvdh: this method claims to return a `MatrixEvent` but this will make it return either `undefined` or `true` :-1:

lib/models/event-timeline-set.js
```diff
@@ -0,0 +1,654 @@
...
+ * timeline to which events are being added in real-time as they are received
+ * from the /sync API. Note that you should not retain references to this
+ * timeline - even if it is the current timeline right now, it may not remain
+ * so if the server gives us a timeline gap in /sync.
+ *
+ * <p>In order that we can find events from their ids later, we also maintain a
+ * map from event_id to timeline and index.
+ *
+ * @constructor
+ * @param {?String} roomId  the roomId of this timelineSet's room, if any
```
richvdh: why do we have this?

lib/models/event-timeline-set.js
```diff
@@ -0,0 +1,654 @@
...
+ * Replaces event with ID oldEventId with one with newEventId, if oldEventId is
+ * recognised.  Otherwise, add to the live timeline.
+ *
+ * @param {MatrixEvent} localEvent     the new event to be added to the timeline
+ * @param {String} oldEventId          the ID of the original event
+ * @param {boolean} newEventId         the ID of the replacement event
+ *
+ * @fires module:client~MatrixClient#event:"Room.timeline"
+ */
+EventTimelineSet.prototype.replaceOrAddEvent = function(localEvent, oldEventId,
```
richvdh: I think this is a bad name for this method, because it makes me think that the old MatrixEvent is going to be replaced with the new one, which it really isn't. I'd be inclined to call it something like `handleRemoteEcho` instead, since it's so specialised.

lib/sync.js
```diff
@@ -655,6 +655,8 @@ SyncApi.prototype._processSyncResponse = function(syncToken, data) {
         }
     }
 
+    this.notifEvents = [];
```
richvdh: ugh.

richvdh:
With so much changing at once, it's hard to be confident that this isn't going to break things. There are also quite a lot of stylistic things which it would have been nice to fix, but I guess that's not happening :/

lib/filter-component.js
ara4n: because the inconsistency spans the spec, synapse, and js-sdk. arguably synapse should be fixed to use the Filter v. FilterComponent terminology, but i suspect erik would claim that synapse got there first. so rather than making the refactor even huger by renaming Filters here, have left it as is for now with a separate bug & todos to fix later.

lib/models/event-timeline-set.js
ara4n: Yes, it's included 2 lines above :).  And yes, it should fire every time an event is added to that particular timelineset so that the UI is kicked to update.

lib/models/event-timeline-set.js
ara4n: oops, done - thanks

lib/models/event-timeline-set.js
ara4n: done

lib/models/room.js
ara4n: hm, are you sure? i tried to keep the old API around for compatibility - e.g. this is used from room-spec.js if nowhere else...

lib/sync.js
ara4n: ugh in terms of abusing a member field for gathering the notifs? or something else? (failure to declare the field? needing a _ prefix?)

lib/models/event-timeline.js
ara4n: actually RoomState behaves fairly sanely if roomId is not specified. have updated jsdoc to reflect.

lib/models/room.js
```diff
@@ -438,245 +456,94 @@ Room.prototype.getCanonicalAlias = function() {
...
-                            "switching to " + existingTimeline);
-            } else {
-                debuglog("Event " + eventId + " already in a different " +
-                            "timeline " + existingTimeline);
-            }
-            timeline = existingTimeline;
-            continue;
-        }
+    timelineSet.getLiveTimeline().setPaginationToken(
+        unfilteredLiveTimeline.getPaginationToken(EventTimeline.BACKWARDS),
```
ara4n: done (i think)

lib/models/room.js
ara4n: hm. yup. done.

lib/models/room.js
ara4n: true. the only thing that calls it seems to be room.removeEvents, which discards the return code anyway. have fixed the jsdoc.

lib/models/event-timeline-set.js
ara4n: good point; complete thinko.

lib/models/event-timeline-set.js
ara4n: done

ara4n:
Agreed that there's a risk of breakage thanks to the invasiveness here; i definitely wish this wasn't a rush job :|

Thanks for the great review despite the indigestibility - PTAL.

lib/filter-component.js
richvdh: I will eat my hat on the day that happens.

lib/models/event-timeline-set.js
richvdh: <- idiot

lib/models/room.js
```diff
@@ -1620,3 +1359,15 @@ module.exports = Room;
  *
  * @param {EventStatus} oldStatus The previous event status.
  */
+
+/**
+ * Fires whenever the live timeline in a room is reset.
```
richvdh: :( wish this wasn't being moved pointlessly

lib/models/room.js
richvdh: that's just a unit test, isn't it?

I can't see any reason why an application should be adding a timeline to the Room; it seems like the sort of thing that would break things. I would encourage vaping it.

lib/models/room.js
```diff
@@ -477,14 +477,20 @@ Room.prototype.getOrCreateFilteredTimelineSet = function(filter) {
...
+    var unfilteredLiveTimeline = this.getLiveTimeline();
 
     unfilteredLiveTimeline.getEvents().forEach(function(event) {
         timelineSet.addLiveEvent(event);
     });
 
+    // find the earliest unfiltered timeline
+    var timeline = unfilteredLiveTimeline;
+    while (timeline.getNeighbouringTimeline(EventTimeline.BACKWARDS)) {
+        timeline = timeline.getNeighbouringTimeline(EventTimeline.BACKWARDS);
```
richvdh: you need to add the events for the timelines you're skipping back over!

lib/client.js
```diff
@@ -1793,6 +1793,38 @@ MatrixClient.prototype.paginateEventTimeline = function(eventTimeline, opts) {
...
+    // added to the timeline both from /sync and /notifications, and lots of
+    // slow and wasteful processing and pagination.  The correct solution is to
+    // extend /messages or /search or something to filter on notifications.
+
+    // use the fictitious token 'end'. in practice we would ideally give it
+    // the oldest backwards pagination token from /sync, but /sync doesn't
+    // know about /notifications, so we have no choice but to start paginating
+    // from the current point in time.  This may well overlap with historical
+    // notifs which are then inserted into the timeline by /sync responses.
+    this._notifTimelineSet.resetLiveTimeline('end', true);
```
richvdh: what's with the `flush` business?

lib/sync.js
richvdh: ugh in terms of abusing a member field, mostly. but the other changes help ameliorate this.

lib/client.js
```diff
@@ -1793,6 +1793,38 @@ MatrixClient.prototype.paginateEventTimeline = function(eventTimeline, opts) {
...
+    }
+
+    // FIXME: This thing is a total hack, and results in duplicate events being
+    // added to the timeline both from /sync and /notifications, and lots of
+    // slow and wasteful processing and pagination.  The correct solution is to
+    // extend /messages or /search or something to filter on notifications.
+
+    // use the fictitious token 'end'. in practice we would ideally give it
+    // the oldest backwards pagination token from /sync, but /sync doesn't
+    // know about /notifications, so we have no choice but to start paginating
```
richvdh: This took me quite a lot of head-scratching to figure out. I think what you mean is that `/sync` doesn't give you a global back-pagination token, so there is no plausible point to back-paginate from?

What happens with `end`? Does synapse happen to treat it as "right now"? Or is it special-cased in the JS somewhere I can't see?

lib/client.js
```diff
@@ -1793,6 +1793,38 @@ MatrixClient.prototype.paginateEventTimeline = function(eventTimeline, opts) {
...
+ */
+MatrixClient.prototype.resetNotifTimelineSet = function() {
+    if (!this._notifTimelineSet) {
+        return;
+    }
+
+    // FIXME: This thing is a total hack, and results in duplicate events being
+    // added to the timeline both from /sync and /notifications, and lots of
+    // slow and wasteful processing and pagination.  The correct solution is to
+    // extend /messages or /search or something to filter on notifications.
```
richvdh: It seems that the problem is more in lack of support in `/sync` than whether you use `/messages` or `/notifications` to back-paginate? (and I don't think `/search` would ever be useful)?

lib/sync.js
```diff
@@ -392,6 +393,16 @@ SyncApi.prototype.sync = function() {
         client.getOrCreateFilter(
             getFilterName(client.credentials.userId), filter
         ).done(function(filterId) {
+            // prepare the notif timeline for pagination before we receive
+            // sync responses.  Technically there are two small races here:
+            // 1: the reset will execute async and may race with the /sync.
```
richvdh: what? what's async in the reset?

lib/sync.js
```diff
@@ -392,6 +393,16 @@ SyncApi.prototype.sync = function() {
...
+            // 2: the reset may return before the sync, and the room may go
```
richvdh: don't understand this.

lib/sync.js
```diff
@@ -392,6 +393,16 @@ SyncApi.prototype.sync = function() {
         client.getOrCreateFilter(
             getFilterName(client.credentials.userId), filter
         ).done(function(filterId) {
+            // prepare the notif timeline for pagination before we receive
```
richvdh: I'm sad that this is happening in a function called `getFilter`, ftr.

lib/timeline-window.js
```diff
@@ -254,7 +254,12 @@ TimelineWindow.prototype.paginate = function(direction, size, makeRequest,
 
     debuglog("TimelineWindow: starting request");
     var self = this;
-    var prom = this._client.paginateEventTimeline(tl.timeline, {
+
+    var paginateTimeline = tl.timeline.getRoomId() ?
```
richvdh: FTR, I'm sad that this has to happen here. I feel like this decision should be made by `MatrixClient.paginateEventTimeline`: it would clean up the API for `MatrixClient`

lib/sync.js
```diff
@@ -742,6 +753,12 @@ SyncApi.prototype._processSyncResponse = function(syncToken, data) {
                 room.currentState.paginationToken = syncToken;
                 self._deregisterStateListeners(room);
                 room.resetLiveTimeline(joinObj.timeline.prev_batch);
+
+                // XXX: for now we have to assume any gap in any timeline is
```
richvdh: This doesn't seem like a thing that is likely to change, as long as you are populating the notification timeline from the room timelines?

lib/sync.js
```diff
@@ -72,6 +72,7 @@ function SyncApi(client, opts) {
     this._running = false;
     this._keepAliveTimer = null;
     this._connectionReturnedDefer = null;
+    this._notifEvents = []; // accumulator of sync events in the current sync response
```
richvdh: It would be easier to review this stuff if you didn't mix everything up together :/

lib/models/room.js
richvdh: looks better

lib/sync.js
```diff
@@ -72,6 +72,9 @@ function SyncApi(client, opts) {
...
+
+    reEmit(client, client.getNotifTimelineSet(), ["Room.timeline"]);
```
richvdh: Looks like this has broken the tests

lib/client.js
ara4n: flush lets you reset both the live timeline and all the others on a timelineset, as documented on the method.  useful for things like notifTimelineSet which don't support discontiguous timelines.  I guess in practice a notifTimelineSet will only have one timeline anyway, so might not be a major problem?

lib/client.js
ara4n: Yes, the understanding is right. I guess my comments suck.

'end' is a fictitious token; synapse treats the same as a missing token. there is no special casing; it's a hack to say "please back paginate from wherever we currently are".

lib/client.js
ara4n: probably.

lib/sync.js
ara4n: originally i had it trying to guess a pagination token, which is async. fixed.

lib/timeline-window.js
ara4n: ok, done.

lib/sync.js
ara4n: clarified.

lib/sync.js
ara4n: yes.

ara4n:
So, i've incorporated all of rich's review, as well as tested the hell out of this over the last few days, and caught the most egregious regressions which predictably snuck in.  There are no substantive changes since Rich's review; only incorporating PR feedback and bugfixes... so I'm going to risk merging it.

PR #197, opened 2016-09-08
Refactor event handling in Crypto

richvdh:
Move the event-handler registration from client.js into crypto.js


NegativeMjark:
LGTM

PR #198, opened 2016-09-08
Move crypto bits into a subdirectory

richvdh:
It was getting a bit sprawly; this should help keep things together.


NegativeMjark:
LGTM

PR #199, opened 2016-09-08
Share the current ratchet with new members

richvdh:
When a new member joins the room, we don't need to reset the megolm session;
instead we can just share the current state with the new user.


NegativeMjark:
LGTM

PR #202, opened 2016-09-09
Show warnings on to-device decryption fail

richvdh:
If we can't decrypt a to-device message, show a warning about it, rather than swallowing the error.


NegativeMjark:
LGTM

PR #203, opened 2016-09-12
Wrap the crypto event handlers in try/catch blocks

NegativeMjark:
Might fix vector-im/vector-web#2160

### DIFF
--- a/lib/crypto/index.js
+++ b/lib/crypto/index.js
@@ -89,32 +89,47 @@ function Crypto(baseApis, eventEmitter, sessionStore, userId, deviceId) {
 
 function _registerEventHandlers(crypto, eventEmitter) {
     eventEmitter.on("sync", function(syncState, oldState, data) {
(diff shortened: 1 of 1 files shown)


lib/crypto/index.js
```diff
@@ -89,32 +89,47 @@ function Crypto(baseApis, eventEmitter, sessionStore, userId, deviceId) {
...
 
-    eventEmitter.on(
-        "RoomMember.membership",
-        crypto._onRoomMembership.bind(crypto)
-    );
+    eventEmitter.on("RoomMember.membership", function(event, member, oldMembership) {
+        try {
+            crypto._onRoomMembership(event, member, oldMembership);
+        } catch (e) {
+             console.error("Error handling membership change:", e);
```
NegativeMjark: Should I try and include information about the RoomID and UserID or the EventID here?

dbkr: I vote let's not bother for now

PR #204, opened 2016-09-14
Send a 'm.new_device' when we get a message for an unknown group session

NegativeMjark:
This should reduce the risk of a device getting permanently stuck unable
to receive encrypted group messages.

TODO:
 - [x] Ratelimit the OH HAIs
 - [x] Set a device_id in the outbound m.room.encrypted events so that the recipients know who to spam with OH HAIs.

### DIFF
--- a/lib/crypto/index.js
+++ b/lib/crypto/index.js
@@ -828,7 +828,7 @@ Crypto.prototype.decryptEvent = function(event) {
     } else {
         // We've got a message for a session we don't have.
         // Maybe the sender forgot to tell us about the session.
-        // Remind the sender that we exists so that they might
+        // Remind the sender that we exist so that they might
         // tell us about the sender.
         if (event.getRoomId !== undefined && event.getSender !== undefined) {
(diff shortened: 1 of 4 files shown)


lib/crypto/index.js
```diff
@@ -820,7 +822,63 @@ Crypto.prototype.decryptEvent = function(event) {
...
         olmDevice: this._olmDevice,
     });
-    return alg.decryptEvent(event);
+    var r = alg.decryptEvent(event);
+    if (r.sessionExists) {
+        return r.result;
+    } else {
+        // We've got a message for a session we don't have.
+        // Maybe the sender forgot to tell us about the session.
+        // Remind the sender that we exists so that they might
```
ara4n: *exist

ara4n:
This feels a bit clunky in that the problematic message will still be dropped on the floor, even if subsequent ones are then okay.  But it's better than it ending up stuck forever.

LGTM.

PR #205, opened 2016-09-14
Update the olm library version to 1.3.0

NegativeMjark:


ara4n:
What will be the symptoms of the migration to 1.3.0?

NegativeMjark:
> What will be the symptoms of the migration to 1.3.0?

Any exist OlmGroupSessions won't unpickle. But those aren't persisted across restarts anyway. So that change won't have any affect.
Attempts to share keys with older versions will fail.
Existing sessions should continue to work.

ara4n:
lgtm modulo lint.

ara4n:
holding off on merge until cutting the next staging release (qv)

PR #206, opened 2016-09-15
Add getKeysProved and getKeysClaimed methods to MatrixEvent.

NegativeMjark:
These list the keys that sender of the event must have ownership
of and the keys of that the sender claims ownership of.

All olm and megolm messages prove ownership of a curve25519 key.
All new olm and megolm message will now claim ownership of a
ed25519 key.

This allows us to detect if an attacker claims ownership of a curve25519
key they don't own when advertising their device keys, because when we
receive an event from the original user it will have a different ed25519 key
to the attackers.

### DIFF
--- a/lib/crypto/OlmDevice.js
+++ b/lib/crypto/OlmDevice.js
@@ -514,21 +514,23 @@ OlmDevice.prototype.getOutboundGroupSessionKey = function(sessionId) {
  * store an InboundGroupSession in the session store
  *
  * @param {string} roomId
- * @param {string} senderKey
+ * @param {string} senderCurve25519Key
  * @param {string} sessionId
  * @param {Olm.InboundGroupSession} session
+ * @param {object} keysClaimed Other keys the sender claims.
  * @private
  */
 OlmDevice.prototype._saveInboundGroupSession = function(
-    roomId, senderKey, sessionId, session
+    roomId, senderCurve25519Key, sessionId, session, keysClaimed
 ) {
     var r = {
         room_id: roomId,
         session: session.pickle(this._pickleKey),
+        keysClaimed: keysClaimed,
     };
 
     this._sessionStore.storeEndToEndInboundGroupSession(
-        senderKey, sessionId, JSON.stringify(r)
+        senderKey, senderCurve25519Key, sessionId, JSON.stringify(r)
     );
 };
 
@@ -569,7 +571,12 @@ OlmDevice.prototype._getInboundGroupSession = function(
     var session = new Olm.InboundGroupSession();
     try {
         session.unpickle(this._pickleKey, r.session);
-        return {sessionExists: true, result: func(session)};
+        return {
+            sessionExists: true,
+            result: func(session),
+            keysProved: {curve25519: senderKey},
+            keysClaimed: r.keysClaimed || {},
+        };
     } finally {
         session.free();
     }
--- a/lib/crypto/OlmDevice.js
+++ b/lib/crypto/OlmDevice.js
@@ -530,7 +530,7 @@ OlmDevice.prototype._saveInboundGroupSession = function(
     };
 
     this._sessionStore.storeEndToEndInboundGroupSession(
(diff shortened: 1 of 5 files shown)


NegativeMjark:
@matrixbot retest this please

NegativeMjark:
This should allow us to fix vector-im/vector-web#2215

ara4n:
looks plausible

lib/crypto/OlmDevice.js
```diff
@@ -569,7 +571,12 @@ OlmDevice.prototype._getInboundGroupSession = function(
     var session = new Olm.InboundGroupSession();
     try {
         session.unpickle(this._pickleKey, r.session);
-        return {sessionExists: true, result: func(session)};
+        return {
```
richvdh: The jsdoc is now a lie.

lib/crypto/OlmDevice.js
```diff
@@ -514,21 +514,23 @@ OlmDevice.prototype.getOutboundGroupSessionKey = function(sessionId) {
  * store an InboundGroupSession in the session store
  *
  * @param {string} roomId
- * @param {string} senderKey
+ * @param {string} senderCurve25519Key
  * @param {string} sessionId
  * @param {Olm.InboundGroupSession} session
+ * @param {object} keysClaimed Other keys the sender claims.
```
richvdh: should I be surprised that I can't find where this is set?

richvdh: Yes, I think I should. It seems to be missing from `addInboundGroupSession`.

richvdh: and `decrypt` will lose it too?

richvdh: 1. It's a shame that this doesn't specify the format of the object. But that doesn't matter too much, because it's a private method.

2. It does't appear to be set by either `addInboundGroupSession` or `decryptGroupMessage`.

lib/crypto/algorithms/olm.js
```diff
@@ -200,7 +211,13 @@ OlmDecryption.prototype.decryptEvent = function(event) {
     // TODO: Check the sender user id matches the sender key.
     // TODO: check the room_id and fingerprint
     if (payloadString !== null) {
-        return {result: JSON.parse(payloadString), sessionExists: true};
+        var payload = JSON.parse(payloadString);
+        return {
```
richvdh: again, the jsdoc is now a lie

richvdh: (see also algorithms/base.js which has yet another copy of the jsdoc. sorry.)

lib/crypto/index.js
```diff
@@ -823,8 +823,11 @@ Crypto.prototype.decryptEvent = function(event) {
         olmDevice: this._olmDevice,
     });
     var r = alg.decryptEvent(event);
+    var payload = r.result;
+    payload.keysClaimed = r.keysClaimed;
+    payload.keysProved = r.keysProved;
     if (r.sessionExists) {
-        return r.result;
+        return payload;
```
richvdh: jsdoc needs an update

lib/models/event.js
```diff
@@ -233,12 +233,28 @@ module.exports.MatrixEvent.prototype = {
         return Boolean(this._clearEvent.type);
     },
 
+    /**
+     * The curve25519 key that sent this event
+     * @return {string}
```
richvdh: s/string/string?/

lib/models/event.js
```diff
@@ -233,12 +233,28 @@ module.exports.MatrixEvent.prototype = {
...
-        return c.sender_key;
+        return this.getKeysProved().curve25519 || null;
+    },
+
+    /**
+     * The keys that must have been owned by the sender of this encrypted event.
+     * @return {object}
+     */
+    getKeysProved: function() {
+        return this._clearEvent.keysProved || {};
```
richvdh: I'm not happy with messing about with the contents of `_clearEvent`. It's *supposed* to be *exactly* what was sent on the wire. keysProved and keysClaimed should be in different properties.

PR #207, opened 2016-09-16
Reset oneTimeKey to null on each loop iteration.

NegativeMjark:
Otherwise we will use a value from a previous iteration of the loop.


NegativeMjark:
This should stop us from sending bogus olm messages to devices that have run out of keys. vector-im/vector-web#2204

ara4n:
lgtm

PR #208, opened 2016-09-16
Check if we need to upload new one-time keys every 10 minutes

NegativeMjark:
This should fix vector-im/vector-web#2204


richvdh:
lgtm

PR #210, opened 2016-09-16
Include keysProved and keysClaimed in the local echo for events we send.

NegativeMjark:
Should fix vector-im/vector-web#2232


ara4n:
lgtm. thanks for the improved commentary :)

PR #211, opened 2016-09-16
Support /publicRooms pagination

dbkr:


ara4n:
hum. i'm wondering if we need to worry about breaking develop against pre 0.18 homeservers...

dbkr:
Actually this would be simple enough to fix: just use the GET version if opts is empty

dbkr:
done

ara4n:
lgtm

PR #212, opened 2016-09-17
Pull user device list on join

richvdh:
When a new user joins a room, make sure we download their device list if we
don't already have it.

This should fix at least one cause of
https://github.com/vector-im/vector-web/issues/2249.

### DIFF
--- a/lib/crypto/algorithms/megolm.js
+++ b/lib/crypto/algorithms/megolm.js
@@ -169,7 +169,11 @@ MegolmEncryption.prototype._prepareNewSession = function(room) {
  * @private
  *
  * @param {string} session_id
+ *
  * @param {Object<string, Object<string, boolean>|boolean>} shareMap
+ *    Map from userid to true (meaning this is a new user in the room, so all
+ *    of his devices need the keys), or a map from deviceid to true (meaning
+ *    this user has one or more new devices, which need the keys).
  *
  * @return {module:client.Promise} Promise which resolves once the key sharing
  *     message has been sent.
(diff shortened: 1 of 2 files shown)


lib/crypto/algorithms/megolm.js
```diff
@@ -169,7 +169,11 @@ MegolmEncryption.prototype._prepareNewSession = function(room) {
  * @private
  *
  * @param {string} session_id
+ *
  * @param {Object<string, Object<string, boolean>|boolean>} shareMap
+ *    Map from userid to true (meaning this is a new user in the room, so all
+ *    of his devices need the keys), or a map from deviceid to true (meaning
+ *    this user has one or more new devices, which need the keys).
```
ara4n: this comment doesn't seem to match the typing of the param? what is the string->{string, boolean} scenario?

lib/crypto/algorithms/megolm.js
```diff
@@ -319,6 +325,23 @@ MegolmEncryption.prototype.onRoomMembership = function(event, member, oldMembers
...
 /**
+ * handle a new user joining a room
+ *
+ * @param {string} userId   new member
+ */
+MegolmEncryption.prototype._onNewRoomMember = function(userId) {
+    // make sure we have a list of this user's devices. We are happy to use a
+    // cached version here: we assume that if we already have a list of the
+    // user's devices, then we already share an e2e room with them, which means
+    // that they will have announced any new devices via an m.new_device.
```
ara4n: Is there any chance that our cache will be stale here because we haven't yet processed the `m.new_device`s for those devices?  (i.e. during a catchup sync, do we always process `to_device` before the membership events?)

ara4n:
i'm going to merge so more people can test it, and as my review questions are just questions rather than anything blocking.

lib/crypto/algorithms/megolm.js
richvdh: good point; clarified in 4f22610

lib/crypto/algorithms/megolm.js
richvdh: >  during a catchup sync, do we always process to_device before the membership events

Yes we do: that's necessary to make sure we get the megolm keys before we try to decrypt timeline messages.

There is a chance of our cache being stale, but I'm hoping to avoid forcing an update here if possible.

PR #213, opened 2016-09-17
Fix sending of oh_hais on bad sessions

richvdh:
Fix a bunch of bugs in the code which tried to send an oh_hai message when we
got a message with an unknown megolm session.

### DIFF
--- a/lib/crypto/index.js
+++ b/lib/crypto/index.js
@@ -883,12 +883,15 @@ Crypto.prototype.decryptEvent = function(event) {
         //
         // XXX: this is a band-aid which masks symptoms of other bugs. It would
         // be nice to get rid of it.
-        if (event.room_id !== undefined && event.sender !== undefined &&
-            event.content.device_id !== undefined
-        ) {
-            this._sendPingToDevice(
-                event.sender, event.content.device_id, event.room_id
-            );
+        if (event.room_id !== undefined && event.sender !== undefined) {
+            var device_id = event.content.device_id;
+            if (device_id === undefined) {
+                // if the sending device didn't tell us its device_id, fall
+                // back to all devices.
+                device_id = null;
+            }
+
+            this._sendPingToDevice(event.sender, device_id, event.room_id);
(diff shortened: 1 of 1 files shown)


lib/crypto/index.js
```diff
@@ -872,13 +873,21 @@ Crypto.prototype.decryptEvent = function(event) {
...
-        if (event.getRoomId !== undefined && event.getSender !== undefined) {
+        // We've got a message for a session we don't have.  Maybe the sender
+        // forgot to tell us about the session.  Remind the sender that we
+        // exist so that they might tell us about the session on their next
+        // send.
+        //
+        // (Alternatively, it might be that we are just looking at
+        // scrollback... at least we rate-limit the m.new_device events :/)
+        //
+        // XXX: this is a band-aid which masks symptoms of other bugs. It would
```
ara4n: Is the bandaid not useful in papering over scenarios where we upgrade the protocol (e.g. the session id change from 1.2 to 1.3)?

lib/crypto/index.js
```diff
@@ -897,34 +906,40 @@ Crypto.prototype.decryptEvent = function(event) {
  * @param {string} roomId   The ID of the room we want to remind them about.
  */
 Crypto.prototype._sendPingToDevice = function(userId, deviceId, roomId) {
-    if (deviceId === undefined) {
-        deviceId = "*";
```
ara4n: why have we lost this? it feels like it's needed if deviceId is optional.

lib/crypto/index.js
richvdh: uh, mayyybee? I'm afraid I'm not *entirely* clear on the difference between 1.2 and 1.3; and in any case I don't really think this is a good way to solve that problem.

lib/crypto/index.js
richvdh: deviceId is always included now, so it just felt like cruft to me. I suppose it might as well stay in.

PR #216, opened 2016-09-22
Handle the first /sync failure differently.

erikjohnston:
A `/sync` request may spuriously fail on occasion, without the "connection" actually being lost. To avoid spurious "Connection Lost" warning messages we ignore the first `/sync` and immediately retry, and only if that fails do we enter an ERROR state.

Is this sensible? Should we emit an event to indicate we *may* have lost a connection?

### DIFF
--- a/lib/sync.js
+++ b/lib/sync.js
@@ -545,13 +545,22 @@ SyncApi.prototype._sync = function(syncOptions) {
         console.error("/sync error %s", err);
         console.error(err);
 
-        debuglog("Starting keep-alive");
-        self._syncConnectionLost = true;
-        self._startKeepAlives().done(function() {
+        if (!self._syncConnectionLost) {
+            // This is the first failure, which may be spurious. To avoid unnecessary
+            // connection error warnings we simply retry the /sync immediately. Only
+            // if *that* one fails too do we say the connection has been lost.
+            // Should we emit a state like "MAYBE_CONNETION_LOST"?
+            self._syncConnectionLost = true;
             self._sync(syncOptions);
-        });
-        self._currentSyncRequest = null;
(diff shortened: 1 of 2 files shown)


lib/sync.js
```diff
@@ -545,13 +545,22 @@ SyncApi.prototype._sync = function(syncOptions) {
         console.error("/sync error %s", err);
         console.error(err);
 
-        debuglog("Starting keep-alive");
-        self._syncConnectionLost = true;
-        self._startKeepAlives().done(function() {
+        if (!self._syncConnectionLost) {
+            // This is the first failure, which may be spurious. To avoid unnecessary
```
Kegsay: "which may be spurious" - Can you give examples on this please as per RL discussion. Namely that things like restarting servers / load balancers / etc will knife any ongoing connection, and we want to effectively ignore that case.

Kegsay:
> Is this sensible? Should we emit an event to indicate we may have lost a connection?

After much musing, I guess it's probably fine. AFAIK clients have no real reason to know about this state, so you should not emit an event to say that you might have lost the connection. If it turns out we need it in the future we can add it then.

Knee jerk reaction was no because it decouples when the `ERROR` state is triggered (no longer *always* on a `/sync` failure) but it is desirable to effectively smooth over errors (much like we do when sending messages).

LGTM

Kegsay:
LGTM!

dbkr:
I'm backing this out because this will cause the client to sit around for another 1 minute and 50 seconds waiting for a connection to time out, without displaying an error message if the sync it triggers goes into the void. I suspect this is causing https://github.com/vector-im/vector-web/issues/2425

PR #223, opened 2016-10-06
Don't emit ERROR until a keepalive poke fails

dbkr:
This accomplishes the same as
https://github.com/matrix-org/matrix-js-sdk/pull/216/files, but
without the client waiting 110 seconds for a sync request to time
out. That is, don't display an error message a soon as a sync
request fails, since we should accept that sometimes long lived
HTTP connections will go away and that's fine.

Also:
 * Use request rather than deprecated requestWithPrefix
 * http-api: The empty string may be falsy, but it's a valid prefix
 * Update the tests to mock keepalive behaviour appropriately, and change one to assert it just keeps emiting ERROR events if it keeps failing, rather than specifically if subsequent sync requests fail.

### DIFF
--- a/lib/sync.js
+++ b/lib/sync.js
@@ -856,10 +856,14 @@ SyncApi.prototype._startKeepAlives = function(delay) {
         clearTimeout(this._keepAliveTimer);
     }
     var self = this;
-    self._keepAliveTimer = setTimeout(
-        self._pokeKeepAlive.bind(self),
-        delay
-    );
+    if (delay > 0) {
+        self._keepAliveTimer = setTimeout(
+            self._pokeKeepAlive.bind(self),
+            delay
+        );
+    } else {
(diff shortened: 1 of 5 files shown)


lib/sync.js
```diff
@@ -546,12 +546,17 @@ SyncApi.prototype._sync = function(syncOptions) {
...
 
         debuglog("Starting keep-alive");
-        self._syncConnectionLost = true;
-        self._startKeepAlives().done(function() {
+        // Note that we do *not* mark the sync connection as
+        // lost yet: we only do this if a keepalive poke
+        // fails, since long lived HTTP connections will
+        // go away sometimes and we shouldn't treat this as
+        // erroneous.
+        // We sent the first keep alive immediately because
```
richvdh: sent -> send

PR #225, opened 2016-10-07
Adds MatrixClient.prototype.mxcUrlToAlternateHttp

lukebarnard1:
I needed to push this somewhere - it might not need to be merged.

This allows specification of an alternate media repo that is not the `baseurl` of the client. The IRC bridge could need this for such configurations.

See https://github.com/matrix-org/matrix-appservice-irc/pull/201

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1868,6 +1868,28 @@ MatrixClient.prototype.mxcUrlToHttp =
 };
 
 /**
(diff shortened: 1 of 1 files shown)


lukebarnard1:
Not needed, see https://github.com/matrix-org/matrix-appservice-irc/pull/201#discussion_r83614827

PR #226, opened 2016-10-08
Fix uploadContent for node.js

richvdh:
9e89e71e broke uploadContent, making it set 'json=true' on the request, so that we would try to turn raw content into JSON. It also misguidedly set a client-side timeout of 30s.

Fix that, and add some tests to check `uploadContent` works.

In mock-request: distinguish between an expectation (`ExpectedRequest`) and an actual request (`Request`). Add support for checking the headers, and the request options in general, to `Request`.


richvdh:
I've tried to address the bifurcated API in #230.

PR #228, opened 2016-10-09
Fix sync breaking when an invalid filterId is in localStorage

pik:
* if getFilter fails for a filterId, null out the localStorage id and redirect to the createFilter path. 

Currently it keeps retrying at `keepAlive` without invalidating: https://github.com/matrix-org/matrix-js-sdk/blob/master/lib/sync.js#L386 (Probably this should actually check whether it is an error that can be recovered through retry or not, rather than looping endlessly?) 

 * add spec

 * fix unit/matrix-client.spec.js http response not matching synapse
 Here is what a response looks like (note that httpStatus was missing from the mock response in the beforeEach() setup of the spec): 
```es6
{
  "errcode":"M_UNKNOWN",
  "name":"M_UNKNOWN",
  "message":"No row found",
  "data":{"errcode":"M_UNKNOWN","error":"No row found"},
  "httpStatus":404
}
```

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -2395,7 +2395,26 @@ MatrixClient.prototype.getOrCreateFilter = function(filterName, filter) {
             }
             // debuglog("Existing filter ID %s: %s; new filter: %s",
             //          filterId, JSON.stringify(oldDef), JSON.stringify(newDef));
-            return;
+            self.store.setFilterIdByName(filterName, undefined);
+            return undefined;
+        }, function(error) {
+            // Synapse currently returns the following when the filter cannot be found:
+            // {
+            //     errcode: "M_UNKNOWN",
+            //     name: "M_UNKNOWN",
+            //     message: "No row found",
+            //     data: Object, httpStatus: 404
+            // }
+            if (error.httpStatus === 404 &&
+                (error.errcode === "M_UNKNOWN" || error.errcode === "M_NOT_FOUND")) {
+                // Clear existing filterId from localStorage
+                // if it no longer exists on the server
+                self.store.setFilterIdByName(filterName, undefined);
+                // Return a undefined value for existingId further down the promise chain
+                return undefined;
+            } else {
+                throw error;
+            }
         });
     }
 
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -2396,6 +2396,16 @@ MatrixClient.prototype.getOrCreateFilter = function(filterName, filter) {
             // debuglog("Existing filter ID %s: %s; new filter: %s",
             //          filterId, JSON.stringify(oldDef), JSON.stringify(newDef));
             return;
+        }, function(error) {
+            // {errcode: "M_UNKNOWN", name: "M_UNKNOWN", message: "No row found", data: Object, httpStatus: 404}
+            if (error.httpStatus === 404 && (error.errcode === "M_UNKNOWN" || error.errcode === "M_NOT_FOUND")) {
```
pik: Updated as discussed to use `errcode` instead of `message`.

lib/client.js
```diff
@@ -2396,6 +2396,16 @@ MatrixClient.prototype.getOrCreateFilter = function(filterName, filter) {
...
+                // Clear existing filterId if the localStorage value is out of sync with the server
+                self.store.setFilterIdByName(filterName, null);
```
richvdh: I don't think this is necessary, since we're about to set it to something else below.

(if it were necessary, the 'out of sync' case would need updating too.)

lib/client.js
```diff
@@ -2396,6 +2396,16 @@ MatrixClient.prototype.getOrCreateFilter = function(filterName, filter) {
...
+                // Return a null value for existingId further down the promise chain
+                return null;
```
richvdh: should probably return undefined, for consistency with the 'out of sync' case and the 'unknown filter name' case.

(though I don't know why the out of sync case has an explicit `return` rather than just letting flow reach the end of the function.)

lib/client.js
pik: Yeah but the `createFilter` could fail and we might retry the whole thing, imho it makes sense to `null` it out if it's definitely been deleted.

richvdh:
Thanks!

PR #237, opened 2016-10-18
Consistency checks for E2E device downloads

richvdh:
Check that the user_id and device_id in device query responses match those that
we expect.

This resolves an unknown-key attack whereby Eve can re-sign Bob's keys with her
own key, thus getting Alice to send her messages which she can then forward to
Bob, making Bob think that Alice sent the messages to him.


NegativeMjark:
LGTM

PR #239, opened 2016-10-19
Check recipient and sender in Olm messages

richvdh:
Embed the sender, recipient, and recipient keys in the plaintext of Olm
messages, and check those fields on receipt.

Fixes https://github.com/vector-im/vector-web/issues/2483


NegativeMjark:
LGTM

PR #240, opened 2016-10-20
Rotate megolm sessions

richvdh:
In order to mitigate backward-secrecy concerns, make sure that we rotate the
outbound megolm session at regular intervals (every week/100 msgs by default).


NegativeMjark:
LGTM

PR #241, opened 2016-10-20
Check for duplicate message indexes for group messages

NegativeMjark:


lib/crypto/OlmDevice.js
```diff
@@ -58,6 +58,15 @@ function OlmDevice(sessionStore) {
     // we don't bother stashing outboundgroupsessions in the sessionstore -
     // instead we keep them here.
     this._outboundGroupSessionStore = {};
+
+    // Store a set of decrypted message indexes for each group session.
+    // This partially mitigates a replay attack where a MITM resends a group
+    // message into the room.
+    //
```
richvdh: please could you say what the shape of this map is? what exactly are the keys and values?

PR #243, opened 2016-10-21
Sign one-time keys, and verify their signatures

richvdh:
We have decided that signing one-time keys is the lesser of two evils;
accordingly, use a new key algorithm type (`signed_curve25519`), sign the
one-time keys that we upload to the server, and verify the signatures on those
we download.

This will mean that develop won't be able to talk to master, but hey, we're in
beta.


NegativeMjark:
LGTM

PR #264, opened 2016-10-24
Update package.json for building browser-matrix without olm

pik:
I tried checking out `v0.6.3` tag and release branch but I am unable to reproduce (?) the dist files currently bundled when running `npm run release` - so just the package.json update for now. 
### Respective file sizes

```
1.6M browser-matrix-0.6.3.js
885K browser-matrix-0.6.3.min.js
168K browser-matrix-0.6.3-lite.min.js
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,10 +7,10 @@
     "test": "istanbul cover --report cobertura --config .istanbul.yml -i \"lib/**/*.js\" jasmine-node -- spec --verbose --junitreport --captureExceptions",
     "check": "jasmine-node spec --verbose --junitreport --captureExceptions",
     "gendoc": "jsdoc -r lib -P package.json -R README.md -d .jsdoc",
-    "build": "jshint -c .jshint lib/ && browserify browser-index.js -o dist/browser-matrix-dev.js --ignore-missing",
+    "build": "jshint -c .jshint lib/ && browserify browser-index.js -o dist/browser-matrix-dev.js --ignore-missing && browserify browser-index --exclude='./lib/crypto/*' browser-index.js  -o dist/browser-matrix-dev-lite.js --ignore-missing",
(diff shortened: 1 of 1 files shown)


ara4n:
@richvdh is this still relevant? @pik - sorry for sitting on it.

pik:
It looks like `release` got removed entirely from package.json in a recent commit (6d67de06a2ca0ea87743c03a26d7dcbae6856cc5). I would like to be able to run a bundled version from the git master without olm - so for me this is still not resolved.

richvdh:
> @richvdh is this still relevant? 

well, ish. It's somewhat superceded by https://github.com/matrix-org/matrix-js-sdk/commit/6d67de06a2ca0ea87743c03a26d7dcbae6856cc5, which makes excluding olm the default. 

> @pik - sorry for sitting on it.

indeed.

richvdh:
> I would like to be able to run a bundled version from the git master without olm - so for me this is still not resolved.

could you be clearer about what you're missing here? You can run `npm run build` without olm.

pik:
> You can run npm run build without olm.

Ah I missed this.. Is it still worthwhile that build seems to include `./lib/crypto` without olm ? Or is that supposing someone wants to serve `olm` separately? 
(Wondering because the file got 20kb larger): 
```
753K browser-matrix.js
199K browser-matrix.min.js
```

richvdh:
> Ah I missed this.. Is it still worthwhile that build seems to include ./lib/crypto without olm ? Or is that supposing someone wants to serve olm separately? 

Yes, as with riot.im, the idea is that you can serve the olm.js asset separately if you happen to have it.

No doubt rather cleverer schemes are possible, but since we don't use browser-matrix.js internally, I'm somewhat reluctant to go down the rabbithole of trying to make them work.

PR #276, opened 2016-11-10
Ignore reshares of known megolm sessions

richvdh:
If we get a second key for a known megolm session, ignore it.

Fixes https://github.com/vector-im/vector-web/issues/2326, one hopes.


NegativeMjark:
LGTM. It might be nice if it merged the ratchet state into the existing ratchet, but that can wait for another day.

PR #279, opened 2016-11-12
Fix examples.

fred-wang:


ara4n:
thanks for this - @kegsay, can you check please?

ara4n:
@fred-wang - please can you PR against the develop branch rather than master?

PR #280, opened 2016-11-12
Send a STOPPED sync updated after call to stopClient

fred-wang:


ara4n:
this needs to be against the develop branch please :) (c.f. https://github.com/matrix-org/matrix-js-sdk/blob/master/CONTRIBUTING.rst)

PR #281, opened 2016-11-12
Disable DEBUG for sync.js

fred-wang:


ara4n:
this needs to be against the develop branch please

PR #284, opened 2016-11-13
Disable DEBUG for sync.js

fred-wang:


fred-wang:
cc' @ara4n @Kegsay 

These were introduced in 7cce41df2e1f7f11f5884fe774d367824497869a and 0bab00c47cfc527726056a66e98f07c72c8d2c99.

DEBUG is disabled in lib/scheduler.js, lib/timeline-window.js, lib/store/webstorage.js, lib/store/session/webstorage.js by default.

fred-wang:

Kegsay:
We still make use of the debug logging, particularly for `sync.js` which can be very difficult to reproduce so we would prefer if logging was enabled so when we do finally reproduce [issues](https://github.com/vector-im/vector-web/issues/544), we have logs.

I'm happy for you to export a `setDebugEnabled(bool)` function which flips on or off debugging if you would rather not have the `console.log` spam, but disabling logging altogether isn't really a viable solution for us. Sorry!

fred-wang:
OK, I think setDebugEnabled is fine. I would actually put the logic into the Utils function so that it is shared by all the files and you do not have to duplicate this DEBUG and debuglog functions everywhere.

BTW, I think https://addons.mozilla.org/en-US/firefox/ has some policy to reject addon that have debug logging (i.e. non-errors). So that might be a problem to use matrix-js-sdk in browser add-ons.

Kegsay:
The intention would be that the default will be `DEBUG=false` and in the projects where we want debug logging (nominally vector-web) we can add `setDebugEnabled` on the respective file. Having a single debug switch is not ideal, since we don't care about *all* debug logging, e.g. the aforementioned scheduler, web-storage and timeline-window.

PR #286, opened 2016-11-13
Send a STOPPED sync updated after call to stopClient

fred-wang:


Kegsay:
@dbkr - Since you implemented the `STOPPED` state, can you please check this?

fred-wang:
I believe this happens when one call MatrixClient::stopClient while the this._currentSyncRequest request in SyncApi::_sync is still executed. Then SyncApi::stop() is called, setting this._running to false and aborting this._currentSyncRequest. This then leads to the place where the code is added in this commit.

dbkr:
Yep, this should definitely have been there: thanks.

Sorry, but can you do the DCO signoff please? (https://github.com/matrix-org/matrix-js-sdk/blob/master/CONTRIBUTING.rst)

fred-wang:

dbkr:
Thanks!

PR #292, opened 2016-11-16
Fix bug in verifying megolm event senders

richvdh:
1a03e534bda introduced a bug which mixed up the keys_proved and the
keys_claimed. Switch them around again so that megolm messages are correctly
tied back to the sending device.


NegativeMjark:
LGTM

PR #293, opened 2016-11-16
Allows to start client with initialSyncLimit = 0

arxcode:
If no "old" timeline data is required when starting the client, it should be possible to set initialSyncLimit to 0 when calling startClient. See http://bit.ly/2f49Kbs for corresponding discussion.


arxcode:
Signed-off-by: Alexander Rudyk (ar-git@mailbox.org)

PR #294, opened 2016-11-16
distinguish unknown users from deviceless users

richvdh:
Fixes https://github.com/vector-im/vector-web/issues/2275, though more usefully it lays some groundwork for other changes


NegativeMjark:
LGTM

PR #295, opened 2016-11-16
E2E: Check devices to share keys with on each send

richvdh:
Instead of trying to maintain a list of devices we need to share with, just check all the devices for all the users on each send.

This should fix https://github.com/vector-im/vector-web/issues/2568, and generally mean we're less likely to get out of sync.


NegativeMjark:
LGTM

PR #296, opened 2016-11-16
Apply unknown-keyshare mitigations

richvdh:
Now that the mobile clients have been updated to send the right fields, enforce
their correctness on the recipient side.


NegativeMjark:
LGTM.

PR #297, opened 2016-11-17
Avoid a packetstorm of device queries on startup

richvdh:
Two main changes here:
 * when we get an m.new_device event for a device we know about, ignore it
 * Batch up the m.new_device events received during initialsync and spam out all the queries at once.

This fixes https://github.com/vector-im/vector-web/issues/2274.

It also removes EncryptionAlgorithm.onNewDevice, so will break develop unless
https://github.com/matrix-org/matrix-js-sdk/pull/295 is landed.

### DIFF
--- a/lib/crypto/index.js
+++ b/lib/crypto/index.js
@@ -54,6 +54,10 @@ function Crypto(baseApis, eventEmitter, sessionStore, userId, deviceId) {
     this._userId = userId;
     this._deviceId = deviceId;
 
+    this._initialSyncCompleted = false;
+    // userId -> deviceId -> true
+    this._pendingNewDevices = {};
+
     this._olmDevice = new OlmDevice(sessionStore);
 
     // EncryptionAlgorithm instance for each room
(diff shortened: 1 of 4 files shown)


lib/crypto/index.js
```diff
@@ -298,30 +303,79 @@ Crypto.prototype.downloadKeys = function(userIds, forceDownload) {
         return q(stored);
     }
 
-    return this._baseApis.downloadKeysForUsers(
+    var r = this._doKeyDownloadForUsers(downloadUsers);
+    var promises = [];
+    downloadUsers.map(function(u) {
```
NegativeMjark: Should this be ``map`` or should it be ``forEach``?

richvdh: `Array.forEach` doesn't exist pre-ES5, and I don't think we have a polyfill for it. Is there a reason to avoid `map` ?

NegativeMjark: Nope, `map` is fine, was just curious.

NegativeMjark:
LGTM

PR #303, opened 2016-12-08
Start a new megolm session when devices are blacklisted

richvdh:
If we have shared the session with a device which is subsequently blacklisted,
we need to start a new session for the next message.

Rather than doing this proactively (which would be subject to false-positives
and require slightly awkward tracking of who we had shared the session with),
we check the list of who we have shared the session with on each send, and
start a new session if any of them are blocked.

Fixes https://github.com/vector-im/riot-web/issues/2146.

### DIFF
--- a/lib/crypto/algorithms/megolm.js
+++ b/lib/crypto/algorithms/megolm.js
@@ -36,8 +36,6 @@ var base = require("./base");
  * @property {string} sessionId
  * @property {Number} useCount     number of times this session has been used
  * @property {Number} creationTime when the session was created (ms since the epoch)
- * @property {module:client.Promise?} sharePromise  If a share operation is in progress,
- *    a promise which resolves when it is complete.
  *
  * @property {object} sharedWithDevices
  *    devices with which we have shared the session key
@@ -47,7 +45,6 @@ function OutboundSessionInfo(sessionId) {
     this.sessionId = sessionId;
     this.useCount = 0;
     this.creationTime = new Date().getTime();
-    this.sharePromise = null;
     this.sharedWithDevices = {};
 }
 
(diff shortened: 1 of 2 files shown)


lib/crypto/algorithms/megolm.js
```diff
@@ -402,7 +430,26 @@ MegolmEncryption.prototype._getDevicesInRoom = function(room) {
     // have a list of the user's devices, then we already share an e2e room
     // with them, which means that they will have announced any new devices via
     // an m.new_device.
-    return this._crypto.downloadKeys(roomMembers, false);
+    return this._crypto.downloadKeys(roomMembers, false).then(function(devices) {
+        // remove any blocked devices
+        for (var userId in devices) {
```
dbkr: Do you need hasOwnProperty boilerplate here?

lib/crypto/algorithms/megolm.js
```diff
@@ -402,7 +430,26 @@ MegolmEncryption.prototype._getDevicesInRoom = function(room) {
...
+            if (!devices.hasOwnProperty(userId)) {
+                continue;
+            }
+
+            var userDevices = devices[userId];
+            for (var deviceId in userDevices) {
+                if (!userDevices.hasOwnProperty(deviceId)) {
```
dbkr: and here?

lib/crypto/algorithms/megolm.js
ara4n: it looks to be on the next line ;)

lib/crypto/algorithms/megolm.js
ara4n: and seems to be here too :)

PR #305, opened 2016-12-09
Update release script to do signed releases

dbkr:
if a signing ID is set in release_config.yaml

Also set the release text to the relevant changelog entry

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -154,8 +154,9 @@ if [ $dodist -eq 0 ]; then
     # it's building
     DIST_VERSION="$tag" npm run dist
 
-    `dirname $0`/scripts/changelog_head.py > latest_changes.md
     popd
+
+    cat "${builddir}/CHANGELOG.md" | `dirname $0`/scripts/changelog_head.py > "${builddir}/latest_changes.md"
     for i in "$builddir"/dist/*; do
         assets="$assets -a $i"
         if [ -n "$signing_id" ]
--- a/release.sh
+++ b/release.sh
@@ -114,15 +114,23 @@ fi
 
 set -x
 
-# Bump package.json, build the dist, and tag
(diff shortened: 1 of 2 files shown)


release.sh
```diff
@@ -145,21 +153,43 @@ if [ $dodist -eq 0 ]; then
     # We haven't tagged yet, so tell the dist script what version
     # it's building
     DIST_VERSION="$tag" npm run dist
+
+    `dirname $0`/scripts/changelog_head.py > latest_changes.md
```
richvdh: this looks like it won't work because of the `pushd "$builddir"`?

richvdh: Also, where is changelog_head.py?

dbkr: Oops, added changelog_head. And yeah, this should work since it's taking the path the script itself is executed with (which will be node_modules/matrix-js-sdk/ because of https://github.com/vector-im/riot-web/blob/master/release.sh#L12) and executing changelog_head relative to that

richvdh: except that at this point you are in the build dir, so node_modules won't be found

ara4n: are we doing anything here to ensure that the riot-web changelog references or links to the equivalent matrix-react-sdk changelog (and similarly js-sdk changelog)?  At the least, we might want to include non-versioned links to the latest_changes for those projects.  And slightly better would be links to the right version of latest_changes for the deps...

ara4n:
lgtm modulo comments, assuming it works :D

release.sh
dbkr: The release script is common to js-sdk and react-sdk though, and this would all be riot-web specific logic to insert js-sdk and react-sdk changelogs.

dbkr: Right, I believe this works now.

PR #306, opened 2016-12-12
Add setRoomDirectoryVisibilityAppService

lukebarnard1:
Add a function for setting the publicity of a room that is bridged to a 3rd party network. This change reflects the second bullet point of matrix-org/synapse#1676


Kegsay:
LGTM

PR #310, opened 2016-12-15
User presence list feature

Deadid:
There are users presence subscription methods in matrix specs, I've added methods that allows us to get user presence list, add users there and drop them from there.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1473,6 +1473,54 @@ MatrixClient.prototype.setPresence = function(opts, callback) {
 };
 
 /**
+* This is internal method
+* @param {MatrixClient} client
+* @param {string} opts
+* @param {module:client.callback} callback Optional.
+* @return {module:client.Promise} Resolves: TODO
+* @return {module:http-api.MatrixError} Rejects: with an error response.
+**/
+function _presenceList(callback, client, opts, method) {
+  var path = utils.encodeUri("/presence/list/$userId", {
+      $userId: client.credentials.userId
+  });
+  return client._http.authedRequest(callback, method, path, undefined, opts);
+}
+
+/**
+* Retrieve current user presence list.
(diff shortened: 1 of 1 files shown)


ara4n:
oops, i totally missed this - thanks for the PR! @kegsay can you take a look?

lib/client.js
```diff
@@ -1473,6 +1473,54 @@ MatrixClient.prototype.setPresence = function(opts, callback) {
 };
 
 /**
+* This is internal method
+* @param {MatrixClient} client
+* @param {string} opts
+* @param {module:client.callback} callback Optional.
```
Kegsay: Docstrings are incorrect. For internal methods, I don't mind what level of documentation this has, but they should be correct if they are included at all. Please either remove these docs or make them correct.

lib/client.js
```diff
@@ -1473,6 +1473,54 @@ MatrixClient.prototype.setPresence = function(opts, callback) {
...
+
+/**
+* Invite another user to presence list.
+* @param {module:client.callback} callback Optional.
+* @param {string} userId
+* @return {module:client.Promise} Resolves: TODO
+* @return {module:http-api.MatrixError} Rejects: with an error response.
+*/
+MatrixClient.prototype.inviteToPresenceList = function(callback, userId) {
+  let opts = {"invite" : [userId]};
```
Kegsay: I'd prefer if the API shape accepted a string array rather than a single user ID. I prefer the array form because it:
 - More accurately represents the underlying HTTP API shape
 - Allows bulk invites/drops in a single round-trip.

Same applies to `:1519`.

Kegsay:
Other than that, LGTM! Thank you!

Deadid:
Fxed

Kegsay:
LGTM - Thank you!

PR #311, opened 2016-12-21
Add `getDate` function to MatrixEvent

lukebarnard1:
This is useful because clients may often wish to construct Date instances using `getTs
()`, but `origin_server_ts` should never change and thus a Date instance can be created at construction time.

### DIFF
--- a/lib/models/event.js
+++ b/lib/models/event.js
@@ -78,6 +78,7 @@ module.exports.MatrixEvent = function MatrixEvent(
     this.status = null;
     this.forwardLooking = true;
     this._pushActions = null;
+    this._date = this.event.origin_server_ts ? new Date(this.event.origin_server_ts) : null;
 
     this._clearEvent = {};
     this._keysProved = {};
@@ -140,6 +141,15 @@ utils.extend(module.exports.MatrixEvent.prototype, {
      */
     getTs: function() {
         return this.event.origin_server_ts;
+    },
+
+    /**
+     * Get a Date instance created in the constructor with the event timestamp
+     * as input.
+     * @return {Date} The event date, e.g. <code>new Date(1433502692297)</code>
+     */
+    getDate: function() {
+        return this._date;
     },
 
     /**
--- a/lib/models/event.js
+++ b/lib/models/event.js
@@ -145,8 +145,7 @@ utils.extend(module.exports.MatrixEvent.prototype, {
     },
 
     /**
-     * Get a Date instance created in the constructor with the event timestamp
-     * as input.
+     * Get the timestamp of this event, as a Date object.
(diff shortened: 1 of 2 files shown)


lib/models/event.js
```diff
@@ -143,6 +145,15 @@ utils.extend(module.exports.MatrixEvent.prototype, {
     },
 
     /**
+     * Get a Date instance created in the constructor with the event timestamp
```
richvdh: can you reword this

"Get the timestamp of this event, as a Date object"

lukebarnard1: yeup

lukebarnard1:
@richvdh WDYT about checking that `this.event.origin_server_ts === this._date.getTime()` in `getDate`, given that `event` is technically public.

richvdh:
I think that if somebody changes the `event` from outside the js-sdk, all hell will break loose; and given the whole motivation for this is for optimisation, I'm not keen on the overhead of the extra check.

lib/models/room.js
```diff
@@ -675,6 +675,7 @@ Room.prototype._handleRemoteEcho = function(remoteEvent, localEvent) {
     // replace the event source (this will preserve the plaintext payload if
     // any, which is good, because we don't want to try decoding it again).
     localEvent.event = remoteEvent.event;
```
richvdh: suggest moving this into `handleRemoteEcho`

lib/models/room.js
```diff
@@ -675,6 +675,7 @@ Room.prototype._handleRemoteEcho = function(remoteEvent, localEvent) {
...
+    localEvent.handleRemoteEcho();
 
     // successfully sent.
     localEvent.status = null;
```
richvdh: ditto

lib/models/event.js
```diff
@@ -374,6 +384,16 @@ utils.extend(module.exports.MatrixEvent.prototype, {
      setPushActions: function(pushActions) {
         this._pushActions = pushActions;
      },
+
+     /**
+      * Replace the `event` property and recalculate any properties based on it.
+      */
+     handleRemoteEcho: function(event) {
```
richvdh: that's gonna fail lint... please document the param.

lukebarnard1: Apparently it didn't but I agree that it should

lib/models/event.js
```diff
@@ -387,6 +387,7 @@ utils.extend(module.exports.MatrixEvent.prototype, {
 
      /**
       * Replace the `event` property and recalculate any properties based on it.
+      * @param {Object} event the object to assign to the `event` property
```
richvdh: sorry, can you add `@internal` here too? then ship it.

PR #312, opened 2016-12-21
Add API to delete threepid

dbkr:
Uses https://github.com/matrix-org/synapse/pull/1714

### DIFF
--- a/lib/base-apis.js
+++ b/lib/base-apis.js
@@ -713,6 +713,24 @@ MatrixBaseApis.prototype.addThreePid = function(creds, bind, callback) {
 };
 
 /**
+ * @param {string} medium The threepid medium (eg. 'email')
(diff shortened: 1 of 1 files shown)


lib/base-apis.js
```diff
@@ -713,6 +713,24 @@ MatrixBaseApis.prototype.addThreePid = function(creds, bind, callback) {
 };
 
 /**
+ * @param {string} medium The threepid medium (eg. 'email')
+ * @param {string} address The threepid address (eg. 'bob@example.com')
```
ara4n: comment about canonicalisation or lack there-of would be nice.

ara4n:
lgtm other than doc improvement request

lib/base-apis.js
dbkr: done

PR #314, opened 2016-12-26
Edit Event Support (experimental)

pik:
This doesn't follow a particular Spec for edit events at the moment which I assume might not come until the Google Doc discussion related to: https://github.com/matrix-org/synapse/pull/1633 ? 

But at least provides some preliminary functionality which allows the matrix-react-sdk (https://github.com/matrix-org/matrix-react-sdk/pull/588) and vector-web (https://github.com/vector-im/riot-web/pull/2725) PR's to function.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -968,6 +968,26 @@ MatrixClient.prototype.sendTextMessage = function(roomId, body, txnId, callback)
 
 /**
  * @param {string} roomId
+ * @param {string} targetId
+ * @param {string} body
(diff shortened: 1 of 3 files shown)


jryans:
Thanks for making this contribution a while back. Since the code base has changed since this was opened, it no longer applies cleanly, and I don't think there's a need to keep it open in this state, as we can always find the code here again if needed. If you are still interested in pursuing this feature, please discuss with us in #riot-dev:matrix.org to find a good approach forward.

PR #316, opened 2017-01-04
release.sh: Bail early if there are uncommitted changes

Kegsay:
This catches both things that are visible on `git diff` as well as staged (added) files. Cargo-culted from http://stackoverflow.com/questions/2657935/checking-for-a-dirty-index-or-untracked-files-with-git and manually tested both scenarios.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -32,6 +32,18 @@ if [ "$ret" -eq 0 ]; then
     exit
 fi
 
+git diff-index --quiet --cached HEAD || ret=$?
+if [ "$ret" -eq 1 ]; then
+    echo "this git checkout has staged (uncommitted) changes. Refusing to release."
+    exit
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -32,6 +32,19 @@ if [ "$ret" -eq 0 ]; then
     exit
 fi
 
+ret=0
+git diff-index --quiet --cached HEAD || ret=$?
+if [ "$ret" -eq 1 ]; then
```
richvdh: how about `-ne 0` instead here, to catch other errors?

In fact, why not just

```
if ! git diff-index --quiet --cached HEAD; then
 ...
```

Kegsay: Done

richvdh:
lgtm, thanks

PR #318, opened 2017-01-11
Add Babel for ES6 support

Kegsay:
- Code now lives in `/src` : you can use ES6.
- `npm run compile` to produce `/lib` output.
- Added as a `prepublish` hook so when we `npm publish` it will compile correctly before release.
- Watch and rebuild via `npm run start`.


package.json
```diff
@@ -47,6 +48,10 @@
     "request": "^2.53.0"
   },
   "devDependencies": {
+    "babel": "^6.5.2",
+    "babel-cli": "^6.18.0",
+    "babel-polyfill": "^6.20.0",
```
richvdh: should this not be a dependency rather than a `devDependency`?

package.json
```diff
@@ -7,11 +7,12 @@
     "test": "istanbul cover --report cobertura --config .istanbul.yml -i \"lib/**/*.js\" jasmine-node -- spec --verbose --junitreport --captureExceptions",
     "check": "jasmine-node spec --verbose --junitreport --captureExceptions",
     "gendoc": "jsdoc -r lib -P package.json -R README.md -d .jsdoc",
+    "compile": "babel -d lib src",
```
richvdh: can you make this write a source map, please? (`-s` ?)

richvdh: we use `build` for this in react-sdk. Any reason to break that convention?

package.json
Kegsay: The polyfill in particular? Probably actually, yeah.

package.json
Kegsay: Done.

package.json
Kegsay: Done.

Kegsay:
> should we include src in the npm package, so people who target ES6 in their apps don't need to use the transpiled version?

Done.

> y no `start` target? think you just need to repeat the babel command with a -w.

What are you talking about? What's a "start target"?

richvdh:
> What are you talking about? What's a "start target"?

well, the react-sdk has a `start` script which starts a watch/rebuild doofer. Likewise vector-web.

Kegsay:
If it's as easy as adding `-w` then fine, I'll do it. I don't want to make a huge PR to add this though if it's a gigantic PITA.

Kegsay:
Seems to work. Done.

package.json
richvdh: half done.

Kegsay:
`build` is already in use

PR #319, opened 2017-01-11
'babel' is not a babel package anymore

dbkr:
and so is just redundant as you have babel-cli which is what you
actually want.


Kegsay:
LGTM

PR #320, opened 2017-01-11
Update build process for ES6

dbkr:
* Make npm run build run npm compile (it needs the output)
 * Switch to ESlint so we can actually use ES6 without the linter
   crying.


package.json
```diff
@@ -7,12 +7,12 @@
     "test": "istanbul cover --report cobertura --config .istanbul.yml -i \"lib/**/*.js\" jasmine-node -- spec --verbose --junitreport --captureExceptions",
     "check": "jasmine-node spec --verbose --junitreport --captureExceptions",
     "gendoc": "jsdoc -r lib -P package.json -R README.md -d .jsdoc",
-    "compile": "babel -s -d lib src",
+    "compile": "npm run lint && babel -s -d lib src",
```
richvdh: is there a reason to maintain a separate `compile` target? why not just stick it into `build` ?

.eslintrc.js
```diff
@@ -0,0 +1,15 @@
...
+        ecmaFeatures: {
+        }
+    },
+    rules: {
+        "max-len": ["warn", {
+            code: 90,
+            ignoreComments: true,
+        }],
+        curly: ["error", "multi-line"],
+    }
```
richvdh: should we enable http://eslint.org/docs/rules/require-jsdoc and http://eslint.org/docs/rules/valid-jsdoc? It would be a shame if we started slipping into bad habits with the jsdoc.

dbkr: Interestingly, turning these on produces 537 warnings, so presumably it's somewhat stricter than jshint :/

Should we turn them on and aspire to fix the warnings?

package.json
dbkr: wfm

dbkr:
ptal

package.json
```diff
@@ -7,12 +7,11 @@
...
     "check": "jasmine-node spec --verbose --junitreport --captureExceptions",
     "gendoc": "jsdoc -r lib -P package.json -R README.md -d .jsdoc",
-    "compile": "babel -s -d lib src",
     "start": "babel -s -w -d lib src",
-    "build": "jshint -c .jshint lib/ && rimraf dist && mkdir dist && browserify --exclude olm browser-index.js -o dist/browser-matrix.js --ignore-missing && uglifyjs -c -m -o dist/browser-matrix.min.js dist/browser-matrix.js",
+    "build": "npm run lint && babel -s -d lib src && rimraf dist && mkdir dist && browserify --exclude olm browser-index.js -o dist/browser-matrix.js --ignore-missing && uglifyjs -c -m -o dist/browser-matrix.min.js dist/browser-matrix.js",
     "dist": "npm run build",
     "watch": "watchify --exclude olm browser-index.js -o dist/browser-matrix-dev.js -v",
-    "lint": "jshint -c .jshint src spec && gjslint --unix_mode --disable 0131,0211,0200,0222,0212 --max_line_length 90 -r spec/ -r src/",
+    "lint": "eslint src",
```
richvdh: shouldn't we check the `spec` folder too? we used to, and `jenkins.sh` apparently does.

richvdh: can we set --max-warnings on this?

.eslintrc.js
```diff
@@ -0,0 +1,25 @@
...
+        ecmaFeatures: {
+        }
+    },
+    rules: {
+        "max-len": ["warn", {
+            code: 90,
+            ignoreComments: true,
+        }],
+        curly: ["error", "multi-line"],
+        "require-jsdoc": ["warn", {
```
richvdh: why are these warnings rather than errors?

package.json
```diff
@@ -7,12 +7,11 @@
...
     "prepublish": "npm run compile && git rev-parse HEAD > git-revision.txt"
```
richvdh: s/compile/build/

.eslintrc.js
dbkr: fixed

package.json
dbkr: fixed (and max-warnings now redundant since everything is an error)

package.json
dbkr: done

PR #321, opened 2017-01-11
Remove babel-polyfill

dbkr:
react-sdk's tests are failing because babel-polyfill is being
pulled in twice.

As per https://babeljs.io/docs/usage/polyfill/ babel-polyfill is
"intended to be used in an application rather than a library/tool".
From a library, we're limited to things that don't modify globals,
since the js env is not ours to start screwing around with.

### DIFF
--- a/browser-index.js
+++ b/browser-index.js
@@ -1,5 +1,3 @@
-require("babel-polyfill");
-
 var matrixcs = require("./lib/matrix");
 matrixcs.request(require("browser-request"));
 module.exports = matrixcs; // keep export for browserify package deps
--- a/index.js
+++ b/index.js
@@ -1,5 +1,3 @@
-require("babel-polyfill");
-
 var matrixcs = require("./lib/matrix");
 matrixcs.request(require("request"));
 module.exports = matrixcs;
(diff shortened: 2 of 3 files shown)


richvdh:
ooi, what do you think about the polyfills in https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/utils.js#L367 and at https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/utils.js#L585?

Kegsay:
> babel-polyfill is "intended to be used in an application rather than a library/tool".

That's a shame. It means we cannot use ES6 in its entirety because we'll be potentially missing ES6 methods and new types (Set, Map, and co). This really limits the value of swapping to ES6 and is going to be wonderfully error prone. 👎

Kegsay:
http://babeljs.io/docs/plugins/transform-runtime/ may be acceptable alternative, but still not fully ES6 compatible :/

> Externalise references to helpers and builtins, automatically polyfilling your code without polluting globals. This plugin is recommended in a library/tool.

dbkr:
Good point, the existing polyfills are probably bad too. We probably ought to just be defining the same things as something that doesn't pollute the global namespace / prototypes.

Basically I believe if you want to use the full set of ES6, you require that the JS environment you're run be compatible, ie. you mandate the app does the polyfills (presumably checking for what you need and barfing if you don't find it).

PR #324, opened 2017-01-13
BREAKING CHANGE: Remove WebStorageStore

Kegsay:
This will be replaced with an IndexedDB style solution. Maintaining 2 different
persistent stores is not my idea of fun.


Kegsay:
I thought about this but it will be automatically populated based off the title of this PR, which is why I used shouty-caps. I feel this is probably clear enough?

dbkr:
Possibly - we normally break it out into a separate 'breaking changes' section. Maybe this is fine though.

Kegsay:
Well let me know either way, I don't really mind which.

PR #326, opened 2017-01-14
Support for importing and exporting megolm sessions

richvdh:
Add a couple of functions to MatrixClient to import and export the megolm data.
Requires a development libolm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,6 @@
     "watchify": "^3.2.1"
   },
   "optionalDependencies": {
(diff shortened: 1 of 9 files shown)


richvdh:
(travis is failing because it's using an old libolm).

NegativeMjark:
> (travis is failing because it's using an old libolm).

Might want to fix this before landing on develop?

richvdh:
> Might want to fix this before landing on develop?

done.

PR #327, opened 2017-01-14
Remove default timeout on Sync

pik:
There is currently no backup strategy for low-bandwidth connections which cannot complete a  `/sync` within `localTimeoutMs` -- this means that slow connections will *never* load riot-web as the set of state events even for several popular rooms is [very large](https://github.com/vector-im/riot-web/issues/2825#issuecomment-270692398) - this removes the localTimeoutMs entirely from the default `/sync` calls (it can still be set via `client.opts`) as per https://github.com/vector-im/riot-web/issues/2737#issuecomment-271740918.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -32,12 +32,6 @@ let EventTimeline = require("./models/event-timeline");
 
 let DEBUG = true;
 
-// /sync requests allow you to set a timeout= but the request may continue
-// beyond that and wedge forever, so we need to track how long we are willing
-// to keep open the connection. This constant is *ADDED* to the timeout= value
-// to determine the max time we're willing to wait.
-let BUFFER_PERIOD_MS = 80 * 1000;
-
 function getFilterName(userId, suffix) {
     // scope this on the user ID because people may login on many accounts
     // and they all need to be stored!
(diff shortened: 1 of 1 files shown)


ara4n:
I wonder whether removing the client timeout means that if you lose your connectivity during a sync (e.g. by sleeping your laptop, or switching network, etc), the sync will never timeout and the client will never recover from the outage, wedging solid.

Unless the keepAlives logic is adequate to recover from that scenario...

@kegsay, wdyt?

pik:
The browser should fail with either, `net::ERR_INCOMPLETE_CHUNKED_ENCODING` `net::ERR_NETWORK_CHANGED` etc. I think? Is there a case where the browser might not set request status to `failed` ?

Kegsay:
In practice, browsers **do not** fail and just wedge entirely. Notably, Safari does this. This is why `localTimeoutMs` is A Thing. I can't accept this PR because it will break other users.

pik:
> In practice, browsers do not fail and just wedge entirely. Notably, Safari does this

Really? In my testing both Firefox and Chromium fail correctly, is there an open bug for this in Safari? 

At any-rate if this is a possibility we have to contend with - what about hooking into `onprogress` to debounce the timeout event?

Kegsay:
> what about hooking into onprogress to debounce the timeout event?

That's what I would prefer.

richvdh:
I think this got superceded by #330, and thence #392.

PR #330, opened 2017-01-16
Debounce Timeout on XMLHTTPRequest if loading

pik:
Addresses the same issue as #327 but keeps the timeout in case of the browser failing to correctly error out a stagnant request per @Kegsay's comment.

### DIFF
--- a/spec/integ/matrix-client-syncing.spec.js
+++ b/spec/integ/matrix-client-syncing.spec.js
@@ -20,7 +20,7 @@ describe("MatrixClient syncing", function() {
 
     beforeEach(function() {
         utils.beforeEach(this); // eslint-disable-line no-invalid-this
-        httpBackend = new HttpBackend();
+        httpBackend = new HttpBackend(true);
         sdk.request(httpBackend.requestFn);
         client = sdk.createClient({
             baseUrl: baseUrl,
@@ -43,12 +43,25 @@ describe("MatrixClient syncing", function() {
             presence: {},
         };
 
+        it("should start keepalives if a /sync hits a timeout error", function(done) {
+            httpBackend.when("GET", "/sync").respond(200, syncData).waitFor((30+80+1)*1000);
+            client.startClient();
+            spyOn(client._syncApi, '_startKeepAlives').andCallThrough();
+            // Work-around since
+            // Jasmine expect(<method>).toThrow() seems to break self-references in
+            // that method.
+            httpBackend.flush().done(function() {
+                expect(client._syncApi._startKeepAlives).toHaveBeenCalled();
+                done();
+            });
+        });
+
         it("should /sync after /pushrules and /filter.", function(done) {
             httpBackend.when("GET", "/sync").respond(200, syncData);
-
             client.startClient();
-
+            spyOn(client._syncApi, '_startKeepAlives').andCallThrough();
             httpBackend.flush().done(function() {
+                expect(client._syncApi._startKeepAlives).not.toHaveBeenCalled();
                 done();
             });
         });
@@ -59,7 +72,6 @@ describe("MatrixClient syncing", function() {
             httpBackend.when("GET", "/sync").check(function(req) {
                 expect(req.queryParams.since).toEqual(syncData.next_batch);
             }).respond(200, syncData);
-
             client.startClient();
 
             httpBackend.flush().done(function() {
@@ -76,7 +88,6 @@ describe("MatrixClient syncing", function() {
             },
             rooms: {
                 join: {
-
                 },
             },
         };
--- a/spec/mock-request.js
+++ b/spec/mock-request.js
@@ -1,14 +1,25 @@
 "use strict";
 const q = require("q");
+let fakeDate;
+let callbacks = require("../lib/realtime-callbacks");
+
+let tick = function tick(millis) {
+    // make sure we tick the fakedate first, otherwise nothing will happen!
+    fakeDate += millis;
+    jasmine.Clock.tick(millis);
+};
+
 
 /**
  * Construct a mock HTTP backend, heavily inspired by Angular.js.
  * @constructor
  */
-function HttpBackend() {
+function HttpBackend(useMockClock) {
     this.requests = [];
     this.expectedRequests = [];
+    this.useMockClock = useMockClock ? true : false
     const self = this;
+
     // the request function dependency that the SDK needs.
     this.requestFn = function(opts, callback) {
         const req = new Request(opts, callback);
@@ -41,11 +52,20 @@ HttpBackend.prototype = {
         const defer = q.defer();
         const self = this;
         let flushed = 0;
-        let triedWaiting = false;
+        let waitLoops = 0;
         console.log(
             "HTTP backend flushing... (path=%s  numToFlush=%s)", path, numToFlush
         );
-        const tryFlush = function() {
+        // Mock out Clock for .timeout() tests
+        if (this.useMockClock) {
+            jasmine.Clock.useMock();
+            fakeDate = Date.now();
+            callbacks.setNow(function() {
+                return fakeDate;
+            });
+        }
+
+        const tryFlush = function tryFlush() {
             // if there's more real requests and more expected requests, flush 'em.
             console.log(
                 "  trying to flush queue => reqs=%s expected=%s [%s]",
@@ -57,23 +77,27 @@ HttpBackend.prototype = {
                 flushed += 1;
                 if (numToFlush && flushed === numToFlush) {
                     console.log("  [%s] Flushed assigned amount: %s", path, numToFlush);
+                    // Restore Clock
+                    if (self.useMockClock) {
+                        callbacks.setNow();
+                    }
                     defer.resolve();
                 } else {
-                    setTimeout(tryFlush, 0);
+                    setImmediate(tryFlush);
                 }
-            } else if (flushed === 0 && !triedWaiting) {
-                // we may not have made the request yet, wait a generous amount of
-                // time before giving up.
-                setTimeout(tryFlush, 5);
-                triedWaiting = true;
+            } else if (flushed === 0 && waitLoops < 50) {
+                waitLoops += 1
+                setImmediate(tryFlush);
             } else {
                 console.log("  no more flushes. [%s]", path);
+                // Restore Clock
+                if (self.useMockClock) {
+                    callbacks.setNow();
+                }
                 defer.resolve();
             }
         };
-
-        setTimeout(tryFlush, 0);
-
+        setImmediate(tryFlush);
         return defer.promise;
     },
 
@@ -116,6 +140,9 @@ HttpBackend.prototype = {
                     matchingReq.checks[j](req);
                 }
                 testResponse = matchingReq.response;
+                if (matchingReq._waitMs) {
+                    tick(matchingReq._waitMs);
+                }
                 console.log("    responding to %s", matchingReq.path);
                 let body = testResponse.body;
                 if (Object.prototype.toString.call(body) == "[object Function]") {
@@ -203,6 +230,7 @@ ExpectedRequest.prototype = {
      * @param {Number} code The HTTP status code.
      * @param {Object|Function} data The HTTP JSON body. If this is a function,
      * it will be invoked when the JSON body is required (which should be returned).
+     * @return {Request} for chaining calls.
      */
     respond: function(code, data) {
         this.response = {
@@ -213,6 +241,7 @@ ExpectedRequest.prototype = {
             body: data,
             err: null,
         };
+        return this;
     },
 
     /**
@@ -229,6 +258,17 @@ ExpectedRequest.prototype = {
             body: null,
             err: err,
         };
+    },
+
+    /**
+     * Advance the time before processing a request by waitMs.
+     *
+     * @param {Number} waitMs number of miliseconds of timeout.
+     * @return {Request} for chaining calls.
+     */
+    waitFor: function(waitMs) {
+        this._waitMs = waitMs;
+        return this;
     },
 };
 
--- a/src/http-api.js
+++ b/src/http-api.js
@@ -627,12 +627,15 @@ module.exports.MatrixHttpApi.prototype = {
 
         const defer = q.defer();
 
-        let timeoutId;
         let timedOut = false;
         let req;
+        let loaded;
+        let timeoutId;
+        let onProgressFunc;
         const localTimeoutMs = opts.localTimeoutMs || this.opts.localTimeoutMs;
+
         if (localTimeoutMs) {
-            timeoutId = callbacks.setTimeout(function() {
+            let timedOut = function () {
                 timedOut = true;
                 if (req && req.abort) {
                     req.abort();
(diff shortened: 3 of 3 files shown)


src/realtime-callbacks.js
```diff
@@ -56,6 +56,80 @@ module.exports.setNow = function(f) {
...
+ *
+ * After a jobName has been defined it may be called without
+ * {func} or {waitMs} arguments to debounce with initial settings.
+ * New {func} or {waitMs} params will over-write ones previously
+ * set in the debouncer.
+ * If called without a jobName a random one is selected.
+ *
+ * @param {function} func   callback to be called after a delay
+ * @param {Number} waitMs  number of milliseconds to delay by
+ *
```
Kegsay: Missing docs for `jobName`.

src/realtime-callbacks.js
```diff
@@ -56,6 +56,80 @@ module.exports.setNow = function(f) {
...
+ * @return {String} a jobName to reference the debouncer.
+ */
+
+module.exports.debounce = function(jobName, func, waitMs) {
+    if (jobName === undefined) {
+        while(_debouncers[jobName = Math.random().toString(16).slice(2, 10)]);
```
Kegsay: Why the `while`? This seems rather opaque given you're just generating an ID.

src/realtime-callbacks.js
```diff
@@ -56,6 +56,80 @@ module.exports.setNow = function(f) {
...
+    if (debouncer) {
+        debouncer.stop();
+    } else {
+        _debouncers[jobName] = debouncer = new Debouncer(jobName, func, waitMs);
+    }
+    debouncer.start(func, waitMs);
+    return jobName;
+};
+/**
+ * Clear a debounce job..
```
Kegsay: Extra `.`

src/realtime-callbacks.js
```diff
@@ -56,6 +56,80 @@ module.exports.setNow = function(f) {
...
+ *
+ * @param {String} a jobName to reference the debouncer.
+ *
+ * @return {Bool} whether a debouncer with jobName was cleared.
```
Kegsay: `{boolean}` for primitive `booleans` please. I don't think JSDoc3 will like `{Bool}`.

src/realtime-callbacks.js
```diff
@@ -56,6 +56,80 @@ module.exports.setNow = function(f) {
...
+
+class Debouncer {
+
+    constructor(jobName) {
+        this.jobName = jobName;
+    }
+
+    complete() {
+        delete _debouncers[this.jobName];
+        this.func();
```
Kegsay: You're assuming that a `func` is always provided, even though you're null guarding on `start(func, waitMs)`. Which is it?

Kegsay:
Tests are also failing on `removeEventListener`.

src/realtime-callbacks.js
pik: This is to make sure it doesn't collide with an existing random ID.

Kegsay: Aaah, right I see now. This line feels a bit too clever, you're:
 - Generating the random ID.
 - Assigning it to `jobName`.
 - Keying into `_debouncers` to check if there's a collision.
 - If there is a collision, rinse and repeat else continue to the next line.

...all on one line. Please can you break this up onto multiple lines which are more self-explanatory, and even add a comment to say what you're doing e.g:

```js
// make sure the job name doesn't collide with an existing debouncer
while (!jobName || _debouncers[jobName]) {
    jobName = Math.random().toString(16).slice(2, 10);
}
```

spec/integ/matrix-client-syncing.spec.js
```diff
@@ -42,12 +42,24 @@ describe("MatrixClient syncing", function() {
...
 
+        it("should start keepalives if a /sync hits a timeout error", function(done) {
+            httpBackend.when("GET", "/sync").respond(200, syncData).timeout(61*1000);
+            client.startClient();
+            spyOn(client._syncApi, '_startKeepAlives').andCallThrough();
+            // Work-around since
+            // Jasmine expect(<method>).toThrow() seems to break self-references in
+            // that method.
+            httpBackend.flush().done(function() {
+                done();
```
Kegsay: Missing assertion on `_startKeepAlives`? To check if it has/has not been called?

spec/mock-request.js
```diff
@@ -59,21 +75,24 @@ HttpBackend.prototype = {
                     console.log("  [%s] Flushed assigned amount: %s", path, numToFlush);
                     defer.resolve();
                 } else {
-                    setTimeout(tryFlush, 0);
+                    // setImmediate(tryFlush);
+                    process.nextTick(tryFlush);
+                    // tick(0);
```
Kegsay: Presumably you're doing this to get around the fact that we're now mocking time so `setTimeout` isn't working as we expect. Can you add a comment to this effect please.

Kegsay: Oh, you do mention this further down! Wonderful! :smiley:

Kegsay: Though one wonders why `setImmediate` is used below and `process.nextTick` is used here? Any reason for it? If not, can we be consistent please.

spec/mock-request.js
```diff
@@ -227,6 +251,16 @@ ExpectedRequest.prototype = {
             err: err,
         };
     },
+
+    /**
+     * Timeout a request by waitMs.
+     *
+     * @param {Number} waitMs number of miliseconds of timeout.
+     */
+    timeout: function(waitMs) {
```
Kegsay: What this actually does is advance the amount of time this HTTP request has taken - this may or may not result in a timeout. Can you please clarify this in the docs, else it appears as if you're saying "time out the request after `waitMs`".

src/sync.js
```diff
@@ -869,6 +869,7 @@ SyncApi.prototype._processSyncResponse = function(syncToken, data) {
  * @return {promise}
  */
 SyncApi.prototype._startKeepAlives = function(delay) {
+    console.log('KEEP ALIVES');
```
Kegsay: Spurious console line.

src/realtime-callbacks.js
pik: Supposing the first debounce call is like `debounce('clientOffline', notifyClientOffline, 1000)` this would allow future `debounce` calls to be made with just `debounce('clientOffline')` without having to maintain a reference to the same function (but also to over-ride the callback of the `debounce` if necessary).

spec/mock-request.js
```diff
@@ -1,14 +1,26 @@
...
 
 /**
  * Construct a mock HTTP backend, heavily inspired by Angular.js.
  * @constructor
  */
-function HttpBackend() {
+function HttpBackend(useMockClock) {
     this.requests = [];
     this.expectedRequests = [];
+    this.useMockClock = useMockClock ? true : false
```
pik: Ideally mock-requests should always use a mock-clock but that seems to be break some internal setTimeout uses?

spec/mock-request.js
```diff
@@ -57,23 +78,27 @@ HttpBackend.prototype = {
...
-                    setTimeout(tryFlush, 0);
+                    setImmediate(tryFlush);
                 }
-            } else if (flushed === 0 && !triedWaiting) {
-                // we may not have made the request yet, wait a generous amount of
-                // time before giving up.
-                setTimeout(tryFlush, 5);
-                triedWaiting = true;
+            } else if (flushed === 0 && waitLoops < 50) {
+                waitLoops += 1
```
pik: `setTimeout` for waiting isn't compatible with mocking that out, so I ended up with something like this -- but it seems like a hack. I believe Jasmine has a waitFor (but it's removed in 2.0) it would be great to do this in a nicer way though...

src/http-api.js
```diff
@@ -661,8 +687,9 @@ module.exports.MatrixHttpApi.prototype = {
                     _matrix_opts: this.opts,
                 },
                 function(err, response, body) {
-                    if (localTimeoutMs) {
-                        callbacks.clearTimeout(timeoutId);
+                    if (timeoutContext) {
+                        timeoutContext.stop();
+                        delete req.onprogres;
```
pik: I'm not sure whether clearing `onprogress` on a request is required or not, once it has loaded or errored.

src/sync.js
```diff
@@ -36,7 +36,7 @@ const DEBUG = true;
 // beyond that and wedge forever, so we need to track how long we are willing
 // to keep open the connection. This constant is *ADDED* to the timeout= value
 // to determine the max time we're willing to wait.
-const BUFFER_PERIOD_MS = 80 * 1000;
+const BUFFER_PERIOD_MS = 30 * 1000;
```
pik: Decreased this since it is now debounced; I think 60 seconds to wait for an onProgressEvent is enough before timing out?

src/realtime-callbacks.js
Kegsay: Hmm, okay then.

src/sync.js
Kegsay: Please keep this at 80. We can tweak it further should it become a problem.

src/realtime-callbacks.js
```diff
@@ -56,6 +56,91 @@ module.exports.setNow = function(f) {
...
+ *
+ * @param {String} jobName  a jobName to reference the debouncer
+ * @param {function} func   callback to be called after a delay
+ * @param {Number} waitMs   number of milliseconds to delay by
+ *
+ * @return {String} jobName a jobName to reference the debouncer
+ */
+
+module.exports.debounce = function(jobName, func, waitMs) {
+    if (jobName === undefined || jobName === null) {
```
Kegsay: Given `jobName` is supposed to be a `string`, why not just:
```js
if (!jobName) {
    // ...
}
```

src/realtime-callbacks.js
```diff
@@ -56,6 +56,91 @@ module.exports.setNow = function(f) {
...
+    }
+
+    complete() {
+        delete _debouncers[this.jobName];
+        this.func();
+    }
+
+    start(func, waitMs) {
+        if (func) this.func = func;
+        if (!isNaN(waitMs)) this.waitMs = waitMs;
```
Kegsay: This won't do what you want it to do.
```js
> isNaN(null)
false
```

Meaning `waitMs` could be set to `null`. You really only want positive `waitMs` values, so just do:

```js
if (waitMs > 0) {
    this.waitMs = waitMs;
}
```

spec/mock-request.js
Kegsay: You absolutely need to comment this and explain why you're doing this.

spec/mock-request.js
```diff
@@ -1,14 +1,26 @@
 "use strict";
 const q = require("q");
+let clock = jasmine.Clock;
```
Kegsay: Why are you aliasing `jasmine.Clock`? It's useful to know at the callsites that you're manipulating jasmine's stuff and not our own, and the alias removes this context.

src/http-api.js
Kegsay: Well you're not spelling `onprogress` correctly...

spec/mock-request.js
pik: I guess jobName could be `0`? If there is no reasonable case for that than `!` should be fine.

src/realtime-callbacks.js
pik: I guess jobName could be `0`? If there is no reasonable case for that than `!` should be fine.

src/http-api.js
```diff
@@ -679,6 +720,10 @@ module.exports.MatrixHttpApi.prototype = {
                     handlerFn(err, response, body);
                 }
             );
+            if (timeoutContext) {
+                req.onprogress = timeoutContext.onProgressCheck;
```
Kegsay: How does this work as intended without `.bind()`ing, given you utilise `this` in `onProgressCheck`?

spec/mock-request.js
pik: I'd like very much to improve this rather than committing it to the code-base (haven't done a lot of testing in Node), are you aware of a better strategy or a reference code-base to take a peek at?

spec/mock-request.js
pik: I was following the way it was done in `realtime-callbacks.spec.js`, /agree referencing it explicitly is better.

src/http-api.js
```diff
@@ -83,6 +83,44 @@ module.exports.MatrixHttpApi = function MatrixHttpApi(event_emitter, opts) {
...
+    this.jobName = null;
+    this.timeoutFunction = timeoutFunction;
+    this.waitMs = waitMs;
+};
+
+// For testing
+module.exports.__TimeoutContext = TimeoutContext;
+
+TimeoutContext.prototype.onProgressCheck = function(progressEvent) {
+    if (this.lastLoaded || progressEvent.loaded > this.lastLoaded) {
```
Kegsay: Why are you comparing Numbers to `null` initially? Why isn't `lastLoaded` set to `0`? Also, you should comment why you are debouncing for any previous positive number. Isn't `progressEvent.loaded > this.lastLoaded` good enough? One can make the argument that if the request has wedged at say `20582` loaded bytes, and isn't changing, then you should NOT be debouncing and waiting forever if we are not making any progress.

pik: Should be `if (!this.lastLoaded ||` or rather `if (this.lastLoaded != null`. Is it safe to just set loaded to 0, the first onProgressEvent may have loaded `0` or is that not the case?

Kegsay:
I think I'm about to make myself unpopular. :disappointed: 

You only need to do a few actions to debounce requests - repeatedly refresh a `setTimeout`, and yet this PR has got a *lot* of additional bloat around such a simple concept.

This PR is adding 2 new classes: `Debouncer` and `TimeoutContext`, both of which are internal to the SDK and both of which are used in precisely 1 place: to debounce HTTP requests. In order to expose enough information to drive this in tests you're exporting additional "private" functions like `__TimeoutContext` and `__getDebouncers`. You're modifying the `HttpBackend` to add additional timeout logic which is used solely in said tests. All of this adds what I feel is needless uncertainty to this PR.

I appreciate the desire to make generic, reusuable components which may be of use elsewhere, but the fact is that they *aren't* being used elsewhere at the moment, and I feel are adding needless complexity. I wouldn't be so paranoid if it weren't the fact that this is literally touching every single HTTP request that will be made in both clients and bridges: I want to have more confidence in this PR and I hoped the tests would do that, but it isn't.

Can you please re-write this as simply as possible: there is no shame to inlining the debouncing code in the `_request` function in `HttpApi` where appropriate if it improves clarity, which for such a simple task it really will. I appreciate you've spent a decent amount of time on this by now, and I'm sorry my bar is set quite so high, but I hope you can understand where I'm coming from given how important this section of code is.

pik:
> You only need to do a few actions to debounce requests - repeatedly
refresh a setTimeout,

In order to repeatedly refresh a setTimeout two separate contexts are
required, the `timeoutId` of the previous timeout to clear and the `loaded`
number returned from the last ProgressEvent. Both of these need to be made
available to the `onprogress` function and both need to be stored in a
permanent data-structure since `loaded` and `timeoutId` will both change
per `onprogress` call. If you would prefer I can inline the data-structure
into the request creation (I think this is what you are suggesting?) but
from my estimation that would add *more* rather than *reduce* the amount of
untestable code.

In fact currently the `/sync` function is not tested anywhere directly (it
requires internal state and has no return value) --- all of the spec/integ
tests use `startClient()`: hence the initially hackish setTimeout(tryFlush,
5) in the mock-request.

Kegsay:
> from my estimation that would add *more* rather than *reduce* the amount of
untestable code

I don't care whether or not we test whether the `loaded` value is always increasing. I care about the underlying concept: if I do not respond to an HTTP request and do not call `onprogress`, do we time out after X seconds? If I do call `onprogress`, we should not time out after those same X seconds. That is the essence of what we are doing here, and seem testable to me.

pik:
Closing since https://github.com/matrix-org/matrix-js-sdk/pull/392 is merged

PR #331, opened 2017-01-18
Add prefer-const

dbkr:
Our code style says we prefer consts, so add it to the linter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "build": "babel -s -d lib src && rimraf dist && mkdir dist && browserify --exclude olm browser-index.js -o dist/browser-matrix.js --ignore-missing && uglifyjs -c -m -o dist/browser-matrix.min.js dist/browser-matrix.js",
     "dist": "npm run build",
     "watch": "watchify --exclude olm browser-index.js -o dist/browser-matrix-dev.js -v",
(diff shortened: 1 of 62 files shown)


richvdh:
pfff. 1860 warnings is a bit much to be dealing with.

1) Are we sure we want to enforce this rule? Even if the code style says so, I'm not necessarily sure I want to be forced into it every time.
2) If we are, maybe we should try a --fix rather than tolerating all those warnings?

dbkr:
I agree with it being annoying being forced into particular things, although consts can help prevent bug, so could be useful. Running --fix seems to do sensible things, so happy to add that.

package.json
```diff
@@ -11,7 +11,7 @@
     "build": "babel -s -d lib src && rimraf dist && mkdir dist && browserify --exclude olm browser-index.js -o dist/browser-matrix.js --ignore-missing && uglifyjs -c -m -o dist/browser-matrix.min.js dist/browser-matrix.js",
     "dist": "npm run build",
     "watch": "watchify --exclude olm browser-index.js -o dist/browser-matrix-dev.js -v",
-    "lint": "eslint --max-warnings 122 src spec",
+    "lint": "eslint --max-warnings 1860 src spec",
```
richvdh: still?

PR #333, opened 2017-01-20
Turn on comma-dangle for function calls

dbkr:
Our code style mandates this, but it's not the default.

Also use the babel-eslint parser because the standard one doesn't
support dangling commas on functions.

Sadly ups the warning count to 510, but it's probably for the best in the long run.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,6 +19,13 @@ module.exports = {
         }],
         curly: ["error", "multi-line"],
         "prefer-const": ["error"],
+        "comma-dangle": ["error", {
+            arrays: "always-multiline",
+            objects: "always-multiline",
+            imports: "always-multiline",
+            exports: "always-multiline",
+            functions: "always-multiline",
+        }],
 
         // loosen jsdoc requirements a little
         "require-jsdoc": ["error", {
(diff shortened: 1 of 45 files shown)


.eslintrc.js
```diff
@@ -57,5 +57,12 @@ module.exports = {
         "no-extend-native": ["warn"],
         "camelcase": ["warn"],
 
+        "comma-dangle": ["warn", {
```
richvdh: so can this be an error now?

package.json
```diff
@@ -11,7 +11,7 @@
     "build": "babel -s -d lib src && rimraf dist && mkdir dist && browserify --exclude olm browser-index.js -o dist/browser-matrix.js --ignore-missing && uglifyjs -c -m -o dist/browser-matrix.min.js dist/browser-matrix.js",
     "dist": "npm run build",
     "watch": "watchify --exclude olm browser-index.js -o dist/browser-matrix-dev.js -v",
-    "lint": "eslint --max-warnings 121 src spec",
+    "lint": "eslint --max-warnings 510 src spec",
```
richvdh: I assume this is no longer needed?

PR #335, opened 2017-01-21
Support for warning users when unknown devices show up

ara4n:
This adds a new crypto.DeviceInfo field called `known` which tracks whether the user has ever been made aware of a device's existence.

If the user tries to send a message into a room with unknown devices present, it throws a custom exception to the app with the details of the devices so the app can warn the user.  It also marks the devices as `known` given the user now knows they exist.

`known` is implemented as a flag on DeviceInfo rather than part of the verification enum as it's slightly semantically distinct from the concept of whether a device is verified or blocked (although it's questionable as to how a user would be able to verify or block a device without knowing it existed!).

The first draft of this tried to track `m.new_device` events as they came in, and warn the user if they referred to devices which the user hadn't seen before - however, this felt prone to races between correlating new-devices and room membership, downloading the actual device info, and the user speaking.  Hopefully this simpler approach doesn't fall foul of the racing concerns mentioned on the original bug.

Hopefully goes some way towards fixing vector-im/riot-web#2143

Twinned with https://github.com/matrix-org/matrix-react-sdk/pull/635.

### DIFF
--- a/docs/warning-on-unverified-devices.txt
+++ b/docs/warning-on-unverified-devices.txt
@@ -1,0 +1,31 @@
+Random notes from Matthew on the two possible approaches for warning users about unexpected
+unverified devices popping up in their rooms....
+
+Original idea...
+================
+
+Warn when an existing user adds an unknown device to a room.
+
+Warn when a user joins the room with unverified or unknown devices.
+
+Warn when you initial sync if the room has any unverified devices in it.
+ ^ this is good enough if we're doing local storage.
+ OR, better:
+Warn when you initial sync if the room has any new undefined devices since you were last there.
+ => This means persisting the rooms that devices are in, across initial syncs.
+
+
+Updated idea...
+===============
+
+Warn when the user tries to send a message:
+  - If the room has unverified devices which the user has not yet been told about in the context of this room
+    ...or in the context of this user?  currently all verification is per-user, not per-room.
+    ...this should be good enough.
+
+  - so track whether we have warned the user or not about unverified devices - blocked, unverified, verified, unverified_warned.
+    throw an error when trying to encrypt if there are pure unverified devices there
+    app will have to search for the devices which are pure unverified to warn about them - have to do this from MembersList anyway?
+      - or megolm could warn which devices are causing the problems.
+
+Why do we wait to establish outbound sessions?  It just makes a horrible pause when we first try to send a message... but could otherwise unnecessarily consume resources?
--- a/src/crypto/algorithms/megolm.js
+++ b/src/crypto/algorithms/megolm.js
@@ -388,6 +388,10 @@ MegolmEncryption.prototype._shareKeyWithDevices = function(session, devicesByUse
 MegolmEncryption.prototype.encryptMessage = function(room, eventType, content) {
     const self = this;
     return this._getDevicesInRoom(room).then(function(devicesInRoom) {
+        // check if any of these devices are not yet known to the user.
+        // if so, warn the user so they can verify or ignore.
+        self._checkForUnknownDevices(devicesInRoom);
+
         return self._ensureOutboundSession(devicesInRoom);
     }).then(function(session) {
         const payloadJson = {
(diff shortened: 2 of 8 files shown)


ara4n:
(yes, i know the tests are broken; i've left them as I'm out of time and I'm genuinely not sure what the right way is to handle the fact that i'm deliberately breaking them by legitimately firing the UnknownDeviceException)

src/crypto/algorithms/megolm.js
```diff
@@ -433,6 +472,9 @@ MegolmEncryption.prototype._getDevicesInRoom = function(room) {
     // have a list of the user's devices, then we already share an e2e room
     // with them, which means that they will have announced any new devices via
     // an m.new_device.
+    //
+    // XXX: what if the cache is stale, and the user left the room we had in common
+    // and then added new devices before joining this one? --Matthew
```
richvdh: yup, this is exactly what https://github.com/vector-im/riot-web/issues/2305#issue-178058560 is on about

src/crypto/index.js
```diff
@@ -706,10 +710,16 @@ Crypto.prototype.setDeviceVerification = function(userId, deviceId, verified, bl
         verificationStatus = DeviceVerification.UNVERIFIED;
     }
 
-    if (dev.verified === verificationStatus) {
+    let knownStatus = dev.known;
+    if (known !== null) {
```
richvdh: should probably also check for `undefined` here

richvdh:
Looks pretty sane to me. I can try and pick this up and make the tests work once I (finally) finish session import/export.

> known is implemented as a flag on DeviceInfo rather than part of the verification enum as it's slightly semantically distinct from the concept of whether a device is verified or blocked (although it's questionable as to how a user would be able to verify or block a device without knowing it existed!).

Yeah, it certainly is. I feel like it's very odd for `DeviceInfo.isVerified` to be able to return `true` for an 'unknown' device - I don't know what that even means. Feels like a bad smell to me.

ara4n:
`DeviceInfo.isVerified` returning `true` for an 'unknown' device means that we haven't told the user that the device exists yet... but it got verified anyway; presumably due to autoverification or cross-signing or similar.

My logic was that: just because known-ness and verification-ness (and blockedness) happen to be orthogonal currently doesn't mean that they should all be put in the same enum (especially one called `DeviceVerification`.  Hence splitting them up into more informative field names.  But i agree it can be argued both ways :)

ara4n:
also, thanks for picking this up; sorry for leaving it 95% done.

docs/warning-on-unverified-devices.txt
```diff
@@ -0,0 +1,31 @@
...
+  - If the room has unverified devices which the user has not yet been told about in the context of this room
+    ...or in the context of this user?  currently all verification is per-user, not per-room.
+    ...this should be good enough.
+
+  - so track whether we have warned the user or not about unverified devices - blocked, unverified, verified, unverified_warned.
+    throw an error when trying to encrypt if there are pure unverified devices there
+    app will have to search for the devices which are pure unverified to warn about them - have to do this from MembersList anyway?
+      - or megolm could warn which devices are causing the problems.
+
+Why do we wait to establish outbound sessions?  It just makes a horrible pause when we first try to send a message... but could otherwise unnecessarily consume resources?
```
richvdh: We've discussed this before. Because each megolm session consumes 1K of localstorage on every other device in the room. I don't want to add a megolm session to my storage every time someone in #megolm logs in on a new device, and it feels O(N^2)y to start to do so.

We *should* be more proactive in fetching the device list (this ties into Erik's work there) - and a lot of the work we do (and consequent pause) is in verifying the content of that device list. See https://github.com/vector-im/riot-web/issues/2157#issuecomment-247857964 for further thoughts on reducing the pause.

richvdh:
I think we can land this and https://github.com/matrix-org/matrix-react-sdk/pull/635.

richvdh:
... or rather, we can land this once we can land https://github.com/matrix-org/matrix-react-sdk/pull/635, which is waiting for a fix to https://github.com/vector-im/riot-web/issues/3018

PR #336, opened 2017-01-22
Support for blacklisting unverified devices, both per-room and globally

ara4n:
Builds on #335, which should be landed before this review makes any sense.

Twinned with https://github.com/matrix-org/matrix-react-sdk/pull/636.

Hopefully solves https://github.com/vector-im/riot-web/issues/2313

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -419,6 +419,17 @@ MatrixClient.prototype.setGlobalBlacklistUnverifiedDevices = function(value) {
 };
 
 /**
+ * @return {boolean} whether to unilaterally blacklist all
+ * unverified devices
+ */
+MatrixClient.prototype.getGlobalBlacklistUnverifiedDevices = function() {
+    if (this._crypto === null) {
+        throw new Error("End-to-end encryption disabled");
+    }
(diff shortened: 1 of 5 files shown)


richvdh:
The main thing outstanding here seems to be persistence of the settings.

richvdh:
> The main thing outstanding here seems to be persistence of the settings.

I see that https://github.com/matrix-org/matrix-react-sdk/pull/636 contains functionality to persist the user setting, but we still need to save the room setting somehow.

ara4n:
persisting roomsettings is done in https://github.com/matrix-org/matrix-react-sdk/pull/636/files#diff-7810353d2ee5218b7242e81423030433R279 no?

richvdh:
> persisting roomsettings is done in https://github.com/matrix-org/matrix-react-sdk/pull/636/files#diff-7810353d2ee5218b7242e81423030433R279 no?

ahhh, that's meant to be a per-user setting - got it

ara4n:
@richvdh now the other PR has landed, this is unblocked. is this LGTMed?

PR #345, opened 2017-01-30
Enable sourcemaps in browserified distro

richvdh:
* use sourceify to read the sourcemaps written by babel
* use {browserify, watchify} -d to write sourcemap
* use exorcist to write the sourcemap to a separate file
* add some uglifyjs incantations to read and write sourcefiles
* simples


richvdh:
Hrm, apparently sourceify doesn't work with more recent browserify. Hold on, will have another go at this.

richvdh:
better

richvdh:
yeah. I think you're supposed to use gulp once you get this far with browserify. But I didn't really fancy dragging that in, either.

PR #346, opened 2017-02-02
Factor out device list management

richvdh:
crypto/index.js was getting huge, so move the device list update management out
to a separate file.

This shouldn't have any effect on functionality.

### DIFF
--- a/src/crypto/DeviceList.js
+++ b/src/crypto/DeviceList.js
@@ -1,5 +1,5 @@
 /*
-Copyright 2017 OpenMarket Ltd
+Copyright 2017 Vector Creations Ltd
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/src/crypto/DeviceList.js
+++ b/src/crypto/DeviceList.js
@@ -1,0 +1,446 @@
+/*
+Copyright 2017 OpenMarket Ltd
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


src/crypto/DeviceList.js
```diff
@@ -0,0 +1,446 @@
+/*
+Copyright 2017 OpenMarket Ltd
```
ara4n: VCL

src/crypto/DeviceList.js
```diff
@@ -0,0 +1,446 @@
...
+     * managed to get a list of devices for this user yet.
+     */
+    getStoredDevicesForUser(userId) {
+        const devs = this._sessionStore.getEndToEndDevicesForUser(userId);
+        if (!devs) {
+            return null;
+        }
+        const res = [];
+        for (const deviceId in devs) {
+            if (devs.hasOwnProperty(deviceId)) {
```
ara4n: bit of a shame that we're doing a weird hybrid of `const` and then horrid `hasOwnProperty` crap. but this can be fixed up in a later hygiene pass.....

ara4n:
this LGTM - massive improvement, if only for gathering that stuff into one place rather than dotting it around in crypto/index.js

src/crypto/DeviceList.js
richvdh: yeah, this was deliberately an airlift of the code to another file with absolutely minimal changes.

src/crypto/DeviceList.js
richvdh: we were OpenMarket when I wrote it ;). Will fix.

PR #347, opened 2017-02-02
Rewrite the device key query logic

richvdh:
Only permit one query per user at a time.

Builds on top of https://github.com/matrix-org/matrix-js-sdk/pull/346.

### DIFF
--- a/src/crypto/DeviceList.js
+++ b/src/crypto/DeviceList.js
@@ -25,7 +25,6 @@ import q from 'q';
 
 import DeviceInfo from './deviceinfo';
 import olmlib from './olmlib';
-import utils from '../utils';
 
 /**
  * @alias module:crypto/DeviceList
@@ -36,10 +35,11 @@ export default class DeviceList {
         this._sessionStore = sessionStore;
         this._olmDevice = olmDevice;
 
+        // users with outdated device lists
         // userId -> true
         this._pendingUsersWithNewDevices = {};
 
-        // userId -> [promise, ...]
+        // userId -> promise
         this._keyDownloadsInProgressByUser = {};
     }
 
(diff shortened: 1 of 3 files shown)


src/crypto/DeviceList.js
```diff
@@ -0,0 +1,421 @@
+/*
+Copyright 2017 OpenMarket Ltd
```
ara4n: VCL

src/crypto/DeviceList.js
```diff
@@ -210,134 +195,124 @@ export default class DeviceList {
      * @param {String} userId
      */
     invalidateUserDeviceList(userId) {
+        if (typeof userId !== 'string') {
```
ara4n: comment why this guard is needed, or if it's just paranoia?

src/crypto/DeviceList.js
```diff
@@ -210,134 +195,124 @@ export default class DeviceList {
...
 
     /**
-     * Start device queries for any users who sent us an m.new_device recently
+     * Start device queries for any users with outdated device lists
+     *
+     * We tolerate multiple concurrent device queries, but only one query per
+     * user.
+     *
+     * If any users already have downloads in progress, they are ignored -
+     * they will be refreshed when the current download completes anyway.
```
ara4n: might be worth explaining why the fact this sounds racey (with the downloads completing potentially at quite different times) isn't a problem...

src/crypto/DeviceList.js
```diff
@@ -53,45 +53,30 @@ export default class DeviceList {
...
             }
-        }
+        });
 
-        if (downloadUsers.length > 0) {
-            const r = this._doKeyDownloadForUsers(downloadUsers);
-            downloadUsers.map(function(u) {
-                promises.push(r[u].catch(perUserCatch(u)));
-            });
+        if (needsFlush) {
```
ara4n: every time i've read the flush code i've been confused, as flush to me always sounds as if you're trying to write to something and you want the writes to all complete.  But here, it's actually flushing *reads* to download new device lists, aiui?  I wonder if there is better naming to be had here...

ara4n:
LGTM, modulo a few q's

src/crypto/DeviceList.js
richvdh: Mostly paranoia. Added a comment.

src/crypto/DeviceList.js
richvdh: Hrm. I've expanded the comment abit, but perhaps I'm missing the source of confusion here?

src/crypto/DeviceList.js
richvdh: Fair point. Renamed to `flushDeviceListRequests`.

PR #348, opened 2017-02-02
Use the device change notifications interface

richvdh:
Listen to the notifications from /sync to tell us when a user's devices are out
of date. We're leaving the `m.new_device` stuff in place for now, for
compatibility with old homeservers/clients.

Builds on https://github.com/matrix-org/matrix-js-sdk/pull/347.

Fixes (together with the server-side support) https://github.com/vector-im/riot-web/issues/2305.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -2638,8 +2638,6 @@ MatrixClient.prototype.startClient = function(opts) {
         };
     }
 
-    this._clientOpts = opts;
-
     if (this._crypto) {
         this._crypto.uploadKeys(5).done();
         const tenMinutes = 1000 * 60 * 10;
@@ -2657,6 +2655,13 @@ MatrixClient.prototype.startClient = function(opts) {
         console.error("Still have sync object whilst not running: stopping old one");
         this._syncApi.stop();
     }
+
+    // shallow-copy the opts dict before modifying and storing it
+    opts = Object.assign({}, opts);
+
+    opts.crypto = this._crypto;
+    this._clientOpts = opts;
+
     this._syncApi = new SyncApi(this, opts);
     this._syncApi.sync();
 };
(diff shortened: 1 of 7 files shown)


richvdh:
Should also fix vector-im/riot-web#2562

richvdh:
And https://github.com/vector-im/riot-web/issues/2564

src/client.js
```diff
@@ -3022,12 +3027,26 @@ module.exports.CRYPTO_ENABLED = CRYPTO_ENABLED;
...
  * "SYNCING", "ERROR", "STOPPED".
+ *
  * @param {?string} prevState An enum representing the previous syncing state.
  * One of "PREPARED", "SYNCING", "ERROR", "STOPPED" <b>or null</b>.
+ *
  * @param {?Object} data Data about this transition.
+ *
  * @param {MatrixError} data.err The matrix error if <code>state=ERROR</code>.
+ *
+ * @param {String} data.oldSyncToken The 'since' token passed to /sync.
```
ara4n: is this new, ooi? or just fixing missing doc?

src/crypto/DeviceList.js
```diff
@@ -318,6 +318,10 @@ export default class DeviceList {
                 this._sessionStore.storeEndToEndDevicesForUser(
                     userId, storage,
                 );
+
+                if (token) {
+                    this._sessionStore.storeEndToEndDeviceSyncToken(token);
```
ara4n: hm. ooi, is there any reason why this would differ from the general /sync token? presumably this will collide with @kegsay's incremental sync stuff?

src/crypto/index.js
```diff
@@ -48,12 +48,16 @@ const DeviceList = require('./DeviceList').default;
  * @param {string} userId The user ID for the local user
  *
  * @param {string} deviceId The identifier for this device.
+ *
+ * @param {Object} clientStore the MatrixClient data store.
```
ara4n: yay!

src/crypto/index.js
```diff
@@ -106,12 +110,6 @@ function Crypto(baseApis, eventEmitter, sessionStore, userId, deviceId) {
 function _registerEventHandlers(crypto, eventEmitter) {
     eventEmitter.on("sync", function(syncState, oldState, data) {
         try {
-            if (syncState == "PREPARED") {
-                // XXX ugh. we're assuming the eventEmitter is a MatrixClient.
```
ara4n: yay^2!

src/crypto/index.js
```diff
@@ -796,6 +795,88 @@ Crypto.prototype._sendNewDeviceEvents = function() {
...
+ * get the users we share an e2e-enabled room with
+ *
+ * @returns {Object<string>} userid->userid map (should be a Set but argh ES6)
+ */
+Crypto.prototype._getE2eRoomMembers = function() {
+    const userIds = Object.create(null);
+
+    const rooms = this._getE2eRooms();
+    for (const r of rooms) {
+        const members = r.getJoinedMembers();
```
ara4n: Do we care at all about invited users?
How does E2E work with peeking, if at all?

ara4n:
LGTM.

src/client.js
richvdh: it's new.

src/crypto/DeviceList.js
richvdh: it's slightly different - the point is to store a token representing the time when we last *successfully* updated the device lists (which may be some time before the last successful sync) - so that when we resume, we can ask for the users which have been updated since then.

When @Kegsay's stuff lands, we'll need to watch out that we still do the `/changes` dance, even though we won't be doing an initial sync - because we don't have a record of which users were out-of-date at the point of exit. (Or we could keep a localstorage/indexeddb record of which users are out of sync.)

src/crypto/index.js
richvdh: To-date, we have currently glossed over those points. Invited users don't get the keys. Peeking users certainly don't.

So yeah, we need to care about it, but there are bigger problems with them.

PR #356, opened 2017-02-08
Install source-map-support in each test

richvdh:
This makes exception traces use the source map, which is much more helpful when
debugging.


package.json
```diff
@@ -64,6 +64,7 @@
     "jasmine-node": "^1.14.5",
     "jsdoc": "^3.4.0",
     "rimraf": "^2.5.4",
+    "source-map-support": "^0.4.11",
```
richvdh: worth noting we have this already, via babel. Just making the dep explicit.

PR #358, opened 2017-02-08
Switch from jasmine to mocha + expect + lolex

richvdh:
Much of this transformation has been done automatically:
* add expect import to each file
* replace `not.to` with `toNot`
* replace `to[Not]Be{Undefined,Null}` with equivalents
* replace `jasmine.createSpy(...)` with `except.createSpy`, and`andCallFake` with `andCall`

Also:
* replace `jasmine.createSpyObj` with manual alternatives
* replace `jasmine.Clock` with `lolex`

The reasons, for the record, for the switch were:

* No way in Jasmine 1 to stop a test run on first failure. A switch to Jasmine  2 was not significantly simpler than a switch to something else.
* Under Jasmine 2, there is a `--stop-on-failure=true` option, but it just stops the test, and doesn't tell you what failed.
* I find Mocha's output much easier to read than jasmine's, thanks to the way jasmine summarises the result at the end rather than as it goes along.
* A general dissatisfaction with Jasmine's do-everything-badly approach, as opposed to letting you pick your own expectations/mocking libraries.


richvdh:
good point, though mocha has its own magicly appearing definitions.

dbkr:
Ah yes, there is a mocha env.

PR #359, opened 2017-02-08
Invalidate device lists when encryption is enabled in a room

richvdh:
... and other stories.

Most of this was cleanup work to get to a state where this change didn't break all the tests inexplicably. The commits might make some sense on their own - or I can pull it into separate PRs if you like.

### DIFF
--- a/spec/integ/matrix-client-crypto.spec.js
+++ b/spec/integ/matrix-client-crypto.spec.js
@@ -393,7 +393,9 @@ describe("MatrixClient crypto", function() {
 
     afterEach(function() {
         aliTestClient.stop();
+        aliTestClient.httpBackend.verifyNoOutstandingExpectation();
         bobTestClient.stop();
+        bobTestClient.httpBackend.verifyNoOutstandingExpectation();
     });
 
     it('Ali knows the difference between a new user and one with no devices',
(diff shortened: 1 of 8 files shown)


src/crypto/algorithms/olm.js
```diff
@@ -64,7 +64,7 @@ OlmEncryption.prototype._ensureSession = function(roomMembers) {
     }
 
     const self = this;
-    this._prepPromise = self._crypto.downloadKeys(roomMembers, true).then(function(res) {
+    this._prepPromise = self._crypto.downloadKeys(roomMembers).then(function(res) {
```
ara4n: is this for the old `m.room.encrypted m.olm.v1.curve25519-aes-sha2` messages?

richvdh: yup. that's why it hasn't been a problem in practice.

spec/integ/matrix-client-methods.spec.js
```diff
@@ -351,7 +351,6 @@ describe("MatrixClient", function() {
 
             httpBackend.when("POST", "/keys/query").check(function(req) {
                 expect(req.data).toEqual({device_keys: {
-                    '@alice:localhost': {},
```
ara4n: i'm a little surprised that we no longer see our own key here...

spec/integ/matrix-client-crypto.spec.js
```diff
@@ -684,4 +684,44 @@ describe("MatrixClient crypto", function() {
...
+                    },
+                };
+                syncData.rooms.join[roomId] = {
+                    state: {
+                        events: [
+                            testUtils.mkEvent({
+                                type: 'm.room.encryption',
+                                skey: '',
+                                content: {
+                                    algorithm: 'm.olm.v1.curve25519-aes-sha2',
```
ara4n: should I be surprised that the alg here is olm and not megolm, given it's for a room?

ara4n:
sounds plausible to me, modulo a few questions of confusion

spec/integ/matrix-client-methods.spec.js
richvdh: or rather, our own userId.

so... the reason for this is, as per the comment on bd07310, we no longer automatically queue up our own userId as needing a refresh - instead it happens once the initialsync happens. In this test, there is no initialsync, so our own userid isn't invalidated.

spec/integ/matrix-client-crypto.spec.js
richvdh: Per https://github.com/matrix-org/matrix-js-sdk/blob/0baea5c1a6fa87523f0c847ab6a889f0d45a4144/spec/integ/matrix-client-crypto.spec.js#L21, this test file doesn't really do megolm - most of the tests here pre-existed megolm. The actual algorithm doesn't make any difference for this test.

ara4n:
lgtm

PR #361, opened 2017-02-09
Clean up/improve e2e logging

richvdh:
In an attempt to make the rageshake logs a bit more useful, try to make the
logging a bit saner. Firstly, make sure we log every decryption failure, and
log it exactly once, rather than in several places. Also record when we receive
megolm keys. Also add some more explicit logging in the sync loop.

### DIFF
--- a/src/crypto/algorithms/base.js
+++ b/src/crypto/algorithms/base.js
@@ -169,7 +169,7 @@ utils.inherits(DecryptionError, Error);
  * @returns {String}
  */
 DecryptionError.prototype.toString = function() {
-    let result = this.name + '[msg: '+ this.message;
+    let result = this.name + '[msg: ' + this.message;
 
     if (this.details) {
         result += ', ' +
--- a/src/crypto/algorithms/base.js
+++ b/src/crypto/algorithms/base.js
@@ -148,14 +148,42 @@ DecryptionAlgorithm.prototype.importRoomKey = function(session) {
 /**
  * Exception thrown when decryption fails
  *
+ * @alias module:crypto/algorithms/base.DecryptionError
  * @constructor
- * @param {string} msg message describing the problem
+ * @param {string} msg user-visible message describing the problem
+ *
+ * @param {Object=} details key/value pairs reported in the logs but not shown
+ *   to the user.
+ *
  * @extends Error
  */
-module.exports.DecryptionError = function(msg) {
+const DecryptionError = function(msg, details) {
+    this.name = 'DecryptionError';
     this.message = msg;
+    this.details = details;
 };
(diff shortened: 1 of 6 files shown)


src/sync.js
```diff
@@ -360,9 +360,6 @@ SyncApi.prototype.getSyncState = function() {
  * Main entry point
  */
 SyncApi.prototype.sync = function() {
-    debuglog("SyncApi.sync: starting with sync token " +
```
lukebarnard1: Why get rid of this ooi?

src/crypto/algorithms/base.js
```diff
@@ -148,14 +148,42 @@ DecryptionAlgorithm.prototype.importRoomKey = function(session) {
...
 };
-utils.inherits(module.exports.DecryptionError, Error);
+utils.inherits(DecryptionError, Error);
+
+/** override the string used when logging
+ *
+ * @returns {String}
+ */
+DecryptionError.prototype.toString = function() {
+    let result = this.name + '[msg: '+ this.message;
```
richvdh: done. surprised the linter didn't pick this up.

src/crypto/algorithms/megolm.js
```diff
@@ -593,26 +600,27 @@ MegolmDecryption.prototype._addEventToPendingList = function(event) {
...
+    const sessionId = content.session_id;
 
     if (!content.room_id ||
-        !content.session_id ||
+        !sessionId ||
         !content.session_key ||
         !senderKey
        ) {
-        console.error("key event is missing fields");
+        console.error(`key event is missing fields`);
```
richvdh: yeah it might, but the code to do so is fiddly, and given we don't really expect to hit it, I'm not convinced it's worth it. I've separated out `senderKey` to a separate condition.

src/crypto/algorithms/olm.js
```diff
@@ -174,59 +174,51 @@ OlmDecryption.prototype.decryptEvent = function(event) {
...
     } catch (e) {
-        console.warn(
-            "Failed to decrypt Olm event (id=" +
-                event.getId() + ") from " + deviceKey +
-                ": " + e.message,
+        throw new base.DecryptionError(
+            "Bad Encrypted Message", {
+                sender: deviceKey,
+                err: e,
+            },
```
richvdh: yup, our lint rules enforce this.

src/sync.js
richvdh: It felt somewhat redundant/noisy, having added https://github.com/matrix-org/matrix-js-sdk/pull/361/files#diff-b5ce4044fcc6267228c6281fd7483b4cR511.

lukebarnard1: Fair enough!

PR #362, opened 2017-02-10
Spread out device verification work

richvdh:
Avoid a big freeze when we process the results of a device query, by splitting
the work up by user.

Fixes https://github.com/vector-im/riot-web/issues/3158

### DIFF
--- a/spec/integ/matrix-client-crypto.spec.js
+++ b/spec/integ/matrix-client-crypto.spec.js
@@ -130,7 +130,12 @@ function expectAliClaimKeys() {
         return {one_time_keys: result};
     });
 
-    return aliTestClient.httpBackend.flush("/keys/claim", 1);
+    // it can take a while to process the key query, so give it some extra
+    // time, and make sure the claim actually happens tather than ploughing on
+    // confusingly.
+    return aliTestClient.httpBackend.flush("/keys/claim", 1, 20).then((r) => {
+        expect(r).toEqual(1);
+    });
 }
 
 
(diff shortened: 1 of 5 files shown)


richvdh:
This might exacerbate some races, so I'm not inclined to merge it on a Friday afternoon. Fixing races is up next on my list anyway.

src/crypto/DeviceList.js
```diff
@@ -306,41 +306,57 @@ export default class DeviceList {
...
-                // map from deviceid -> deviceinfo for this user
-                const userStore = {};
-                const devs = this._sessionStore.getEndToEndDevicesForUser(userId);
-                if (devs) {
-                    Object.keys(devs).forEach((deviceId) => {
-                        const d = DeviceInfo.fromStorage(devs[deviceId], deviceId);
-                        userStore[deviceId] = d;
-                    });
-                }
+                prom = prom.delay(5).then(() => {
```
lukebarnard1: why the delay ooi?

lukebarnard1: Also, should this be overwriting `prom` on each iteration? Should this not use `.all` ?

richvdh: The delay is simply to try to spread out the work a bit and make the UI feel a bit snappier, by letting other things run in the event loop. It's an entirely subjective thing - it just made it feel a bit more responsive.

.all would make it run all the promises "in parallel" - or rather, in the same cycle of the event loop, which is exactly what we're trying to avoid here.

spec/integ/matrix-client-crypto.spec.js
```diff
@@ -130,7 +130,12 @@ function expectAliClaimKeys() {
         return {one_time_keys: result};
     });
 
-    return aliTestClient.httpBackend.flush("/keys/claim", 1);
+    // it can take a while to process the key query, so give it some extra
+    // time, and make sure the claim actually happens tather than ploughing on
```
richvdh: fixed, thanks

richvdh:
checking what's up with the tests here.

src/crypto/DeviceList.js
lukebarnard1: OK, fair enough

richvdh:
@lukebarnard1: are you happy with this now?

PR #367, opened 2017-02-15
Add docs to event

dbkr:
to say that sender won't always be set


src/models/event.js
```diff
@@ -61,7 +61,9 @@ module.exports.EventStatus = {
  * from changes to event JSON between Matrix versions.
  *
  * @prop {RoomMember} sender The room member who sent this event, or null e.g.
- * this is a presence event.
+ * this is a presence event. This is only guarenteed to be set for events that
```
Kegsay: > guarenteed

Kegsay:
Otherwise LGTM

PR #370, opened 2017-02-16
Poll /sync with a short timeout while catching up

richvdh:
On first connect, or after a disconnection, poll /sync with timeout=0 until
we get no to_device messages back. This will allow us to figure out whether
we have more to_device messages queued up for us on the server, which in
turn will help us fix https://github.com/vector-im/riot-web/issues/2782.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -3132,6 +3132,9 @@ module.exports.CRYPTO_ENABLED = CRYPTO_ENABLED;
  *    will become the 'since' token for the next call to /sync. Only present if
  *    <code>state=PREPARED</code> or <code>state=SYNCING</code>.
  *
+ * @param {boolean} data.catchingUp True if we are working our way through a
+ *    backlog of events after connecting. Only present if <code>state=SYNCING</code>.
+ *
  * @example
  * matrixClient.on("sync", function(state, prevState, data) {
  *   switch (state) {
--- a/src/client.js
+++ b/src/client.js
@@ -3054,7 +3054,7 @@ module.exports.CRYPTO_ENABLED = CRYPTO_ENABLED;
  * already ERROR. <i>This is the equivalent of "syncError" in the previous
  * API.</i></li>
  *
- * <li>RECONNECTING: The sync connedtion has dropped, but not (yet) in a way that
+ * <li>RECONNECTING: The sync connection has dropped, but not (yet) in a way that
  * should be considered erroneous.
  * </li>
  *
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -3038,21 +3038,26 @@ module.exports.CRYPTO_ENABLED = CRYPTO_ENABLED;
...
  * <li>ERROR : The client has had a problem syncing with the server. If this is
  * called <i>before</i> PREPARED then there was a problem performing the initial
  * sync. If this is called <i>after</i> PREPARED then there was a problem polling
  * the server for updates. This may be called multiple times even if the state is
  * already ERROR. <i>This is the equivalent of "syncError" in the previous
  * API.</i></li>
- * <li>RECONNECTING: The sync connedtion has dropped, but not in a way that should
- * be considered erroneous.
+ *
+ * <li>RECONNECTING: The sync connedtion has dropped, but not (yet) in a way that
```
Kegsay: > connedtion

src/sync.js
```diff
@@ -483,9 +484,30 @@ SyncApi.prototype._sync = function(syncOptions) {
 
     const syncToken = client.store.getSyncToken();
 
+    let pollTimeout = this.opts.pollTimeout;
+
+    if (this.getSyncState() !== 'SYNCING' || this._catchingUp) {
+        // unless we are happily syncing already, we want the server to return
+        // as quickly as possible, even if there are no events queued. This
```
Kegsay: > we want the server to return as quickly as possible, even if there are no events queued.

Why? There's no events to process.

Kegsay:
This feels like a hack. Why isn't `/sync` returning if there are `to_device` messages waiting for you?

Code-wise, LGTM but I'm questioning why this is necessary.

richvdh:
> This feels like a hack. Why isn't /sync returning if there are to_device messages waiting for you?

It is.

/sync doesn't return all to_device messages at once. (It returns up to 100, but I don't want to rely on that, and anyway that knowledge doesn't help me if there are exactly 100 to_device messages waiting.) I need a way to figure out if there are more messages on the server, or if we have reached the end of the queue.

src/client.js
richvdh: fixed, thanks

Kegsay:
Then it'd be great if we can comment that this is a hack and that ideally Synapse would return from `/sync` immediately, so if and when Synapse does do that we can remove this code.

LGTM after comment.

richvdh:
I think you're missing the point.

It does return from /sync immediately, if there are messages waiting.

richvdh:
you could argue for a flag in the /sync response saying "this is the end of the list", but frankly that seems to complicate the API and the server-side implementation and doesn't make the client-side implementation any easier?

Kegsay:
Oh I see, you don't know when you've "caught up". Okay, LGTM then.

PR #371, opened 2017-02-16
Add indent rule to eslint

dbkr:


richvdh:
sounds plausible, but it completely breaks with the existing code. It looks like it doesn't like some fairly reasonable things, so this might need more configuration.

dbkr:
Yeah, just discovered this as the jenkins build failed. A lot of it is actually our indenting being all over the place, but some of it is us being a bit undecided about what to do with function parameters.

ara4n:
do we actually have a problem with indentation anywhere? i don't see what these linting diversions are buying us...

Kegsay:
> do we actually have a problem with indentation anywhere? i don't see what these linting diversions are buying us...

I agree with @ara4n on this. I like having a standard code format, but you quickly hit diminishing returns unless the tooling can auto-fix things for you. We run everything through PRs, so any style violations should be spotted then and corrected. If it helped with correctness of code (e.g shifting from `var` to `let`/`const`) then great, I can justify that. I can't justify indent style. Just my thoughts anyway, \*gets out of your PR\*

richvdh:
dave, I'm going to close this. It doesn't look like we're going to be in a position to land it any time soon.

PR #378, opened 2017-02-21
Emit an error when UnknownDeviceError occurs during VoIP operations

lukebarnard1:


ara4n:
i think we may need this in a lot more places than just `m.call.invite` - e.g. attaachments, m.call.answer, m.call.hangup etc.  is there a more generic place? or just a matter of putting in lots of bubbles?

lukebarnard1:
The obvious generic place is 
```js
const sendEvent = function(self, eventType, content) {
    return self.client.sendEvent(self.roomId, eventType, content);
};
```

richvdh:
Looks fine. Could you document the new event type and `error` at the bottom, in a similar way to https://github.com/matrix-org/matrix-js-sdk/blob/e571271673fea843c1f69c9ca456bff86814adcf/src/client.js#L3006, so that we remember what the difference is? I'm particularly interested in recording the behaviour of MatrixCall when it encounters a `send_event_error` condition, and what your options are as an app when you get such an event.

src/client.js
```diff
@@ -3020,6 +3020,25 @@ module.exports.CRYPTO_ENABLED = CRYPTO_ENABLED;
...
+ * does a hangup when it encounters a normal `error`, which in turn could lead to an
+ * UnknownDeviceError.)
+ * <p>
+ * To deal with an UnknownDeviceError when trying to send events, the application should let
+ * users know that there are new devices in the encrypted room (into which the event was
+ * sent) and give the user the options to resend unsent events or cancel them.
+ *
+ * @event module:client~MatrixClient#"send_event_error"
+ * @param {Error} err The error caught from calling client.sendEvent in call.js.
+ * matrixClient.on("send_event_error", function(err){
```
richvdh: example needs `@example`

src/client.js
```diff
@@ -3020,6 +3020,25 @@ module.exports.CRYPTO_ENABLED = CRYPTO_ENABLED;
  */
 
 /**
+ * Fires whenever an error occurs when call.js sendEvent encounters an error.
+ * <p>
+ * This is required to allow UnknownDeviceError to propagate from client.sendEvent, whilst
```
richvdh: it's not specific to unknowndeviceerror though? it's about all event-transmission errors?

src/client.js
```diff
@@ -3020,6 +3020,25 @@ module.exports.CRYPTO_ENABLED = CRYPTO_ENABLED;
...
+ * differentiating it from being a normal `error` emission. (This is because call.js
+ * does a hangup when it encounters a normal `error`, which in turn could lead to an
+ * UnknownDeviceError.)
+ * <p>
+ * To deal with an UnknownDeviceError when trying to send events, the application should let
+ * users know that there are new devices in the encrypted room (into which the event was
+ * sent) and give the user the options to resend unsent events or cancel them.
+ *
```
richvdh: worth noting what the MatrixCall does. Does it hang up? does it carry on regardless? does it block until you resend the event?

src/client.js
```diff
@@ -3020,6 +3020,25 @@ module.exports.CRYPTO_ENABLED = CRYPTO_ENABLED;
...
+ * Fires whenever an error occurs when call.js sendEvent encounters an error.
+ * <p>
+ * This is required to allow UnknownDeviceError to propagate from client.sendEvent, whilst
+ * differentiating it from being a normal `error` emission. (This is because call.js
+ * does a hangup when it encounters a normal `error`, which in turn could lead to an
+ * UnknownDeviceError.)
+ * <p>
+ * To deal with an UnknownDeviceError when trying to send events, the application should let
+ * users know that there are new devices in the encrypted room (into which the event was
+ * sent) and give the user the options to resend unsent events or cancel them.
```
richvdh: how does the app resend/cancel unsent events?

src/client.js
```diff
@@ -3020,6 +3020,25 @@ module.exports.CRYPTO_ENABLED = CRYPTO_ENABLED;
  */
 
 /**
+ * Fires whenever an error occurs when call.js sendEvent encounters an error.
```
richvdh: `sendEvent` is an implementation detail. "Fires when the MatrixCall encounters an error when sending a Matrix event".

src/client.js
richvdh: I think this whole paragraph is unnecessary tbh.

src/client.js
```diff
@@ -3020,6 +3020,25 @@ module.exports.CRYPTO_ENABLED = CRYPTO_ENABLED;
...
+ * This is required to allow UnknownDeviceError to propagate from client.sendEvent, whilst
+ * differentiating it from being a normal `error` emission. (This is because call.js
+ * does a hangup when it encounters a normal `error`, which in turn could lead to an
+ * UnknownDeviceError.)
+ * <p>
+ * To deal with an UnknownDeviceError when trying to send events, the application should let
+ * users know that there are new devices in the encrypted room (into which the event was
+ * sent) and give the user the options to resend unsent events or cancel them.
+ *
+ * @event module:client~MatrixClient#"send_event_error"
```
richvdh: ought to be associated with the MatrixCall, not the MatrixClient.

PR #380, opened 2017-02-27
Add m.login.email.identity support to UI auth

dbkr:
Extends the interactive-auth to support m.login.email.identity
This also includes the ability to resume a UI auth session given
the appropriate information (ie. to resume the auth flow having
clicked a link in a verification email) and supports polling for
the auth session having been completed elsewhere.

### DIFF
--- a/src/interactive-auth.js
+++ b/src/interactive-auth.js
@@ -220,6 +220,8 @@ InteractiveAuth.prototype = {
      * This must be set in order to successfully poll for completion
      * of the email validation.
      * Specific to m.login.email.identity
+     *
+     * @param {string} sid The sid for the email validation session
      */
     setEmailSid: function(sid) {
         this._emailSid = sid;
--- a/src/interactive-auth.js
+++ b/src/interactive-auth.js
@@ -56,7 +56,9 @@ const MSISDN_STAGE_TYPE = "m.login.msisdn";
  *     called when the status of the UI auth changes, ie. when the state of
  *     an auth stage changes of when the auth flow moves to a new stage.
  *     The arguments are: the login type (eg m.login.password); and an object
- *     specific to the login type.
+ *     specific to the login type. These are:
+ *         m.login.email.identity:
+ *          * emailSid: string, the sid of the active email auth session
  *
  * @param {object?} opts.inputs Inputs provided by the user and used by different
  *     stages of the auto process. The inputs provided will affect what flow is chosen.
--- a/src/interactive-auth.js
+++ b/src/interactive-auth.js
@@ -1,5 +1,6 @@
 /*
 Copyright 2016 OpenMarket Ltd
+Copyright 2017 Vector Creations Ltd
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -17,8 +18,12 @@ limitations under the License.
 
 /** @module interactive-auth */
 const q = require("q");
+const url = require("url");
 
 const utils = require("./utils");
+
+const EMAIL_STAGE_TYPE = "m.login.email.identity";
+const MSISDN_STAGE_TYPE = "m.login.msisdn";
 
 /**
  * Abstracts the logic used to drive the interactive auth process.
@@ -37,6 +42,8 @@ const utils = require("./utils");
  *
  * @param {object} opts  options object
  *
+ * @param {object} opts.matrixClient A matrix client to use for the auth process
+ *
  * @param {object?} opts.authData error response from the last request. If
  *    null, a request will be made with no auth before starting.
  *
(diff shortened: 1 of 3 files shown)


src/interactive-auth.js
```diff
@@ -201,8 +339,30 @@ InteractiveAuth.prototype = {
...
+            let flowHasMsisdn = false;
+            for (const stage of flow.stages) {
+                if (stage === EMAIL_STAGE_TYPE) {
+                    flowHasEmail = true;
+                } else if (stage == MSISDN_STAGE_TYPE) {
+                    flowHasMsisdn = true;
+                }
+            }
+
+            if (flowHasEmail == haveEmail && flowHasMsisdn == haveMsisdn) {
```
richvdh: surely we don't want to require that the flow demands an email address just because we happen to have an email address? Likewise msisdn?

src/interactive-auth.js
```diff
@@ -45,17 +52,50 @@ const utils = require("./utils");
...
  *
- * @param {function(string, object?)} opts.startAuthStage
- *     called to ask the UI to start a particular auth stage. The arguments
- *     are: the login type (eg m.login.password); and (if the last request
- *     returned an error), an error object, with fields 'errcode' and 'error'.
+ * @param {function(string, object?)} opts.stateUpdated
+ *     called when the status of the UI auth changes, ie. when the state of
+ *     an auth stage changes of when the auth flow moves to a new stage.
+ *     The arguments are: the login type (eg m.login.password); and an object
+ *     specific to the login type.
```
richvdh: this isn't a terribly helpful description.

src/interactive-auth.js
```diff
@@ -168,15 +257,64 @@ InteractiveAuth.prototype = {
         if (!nextStage) {
             throw new Error("No incomplete flows from the server");
         }
+        if (nextStage == this._currentStage) {
```
richvdh: won't this swallow errors from (eg) password login?

src/interactive-auth.js
```diff
@@ -168,15 +257,64 @@ InteractiveAuth.prototype = {
...
                 errcode: this._data.errcode || "",
                 error: this._data.error || "",
-            };
+            });
+            return;
+        }
+
+        const stageStatus = {};
+        if (nextStage == EMAIL_STAGE_TYPE) {
+            stageStatus.busy = true;
```
richvdh: isn't it always 'busy' until the next stage happens?

src/interactive-auth.js
```diff
@@ -143,19 +221,30 @@ InteractiveAuth.prototype = {
...
                     // doesn't look like an interactive-auth failure. fail the whole lot.
                     throw error;
                 }
-                self._data = error.data;
+                if (errorFlows) self._data = error.data;
                 self._startNextAuthStage();
             },
-        ).catch(this._completionDeferred.reject).done();
+        );
+        if (!ignoreFailure) {
```
richvdh: why is this necessary? surely we should only be ignoring interactive-auth failures, and they are caught further up?

src/interactive-auth.js
```diff
@@ -168,15 +257,64 @@ InteractiveAuth.prototype = {
...
+        // an obvious thing to stick in a different file / function if there
+        // were more of them.
+        if (nextStage == EMAIL_STAGE_TYPE) {
+            if (this._emailSid) {
+                this.poll();
+            } else {
+                this._requestEmailToken().catch(
+                    this._completionDeferred.reject,
+                ).finally(() => {
+                    this._stateUpdatedCallback(nextStage, { busy: false });
```
richvdh: oh wait, what's happening here? why aren't we calling submitAuthDict ourselves here?

src/interactive-auth.js
dbkr: Well, the code previously worked as you suggest, but I realised that, at least in the case of registration, if the user provided, say, an email address and it ended up not being used, you would end up registering an account without an email address associated, which would be surprising and potentially leave you without a means to reset your password. I therefore went with requiring all the inputs to be used.

src/interactive-auth.js
dbkr: Not necessarily: I don't think we want the whole flow to fail because the internet connection died briefly while we were doing a poll.

src/interactive-auth.js
richvdh: Seems fair. I guess this will want to become pluggable at some point.

Could you add a couple of comments explaining your thinking here?

src/interactive-auth.js
```diff
@@ -143,19 +221,30 @@ InteractiveAuth.prototype = {
...
-                if (error.httpStatus !== 401 || !error.data || !error.data.flows) {
+                // sometimes UI auth errors don't come with flows
+                const errorFlows = error.data ? error.data.flows : null;
+                const haveFlows = Boolean(self._data.flows) || Boolean(errorFlows);
+                if (error.httpStatus !== 401 || !error.data || !haveFlows) {
                     // doesn't look like an interactive-auth failure. fail the whole lot.
                     throw error;
                 }
-                self._data = error.data;
+                if (errorFlows) self._data = error.data;
```
richvdh: why is this the right thing to do? We know we at least have a non-null `error.data` here; why do we need to check that we have `error.data.flows` too?

src/interactive-auth.js
richvdh: mmhmm. ok. can you make the doc-comment explicit in that this applies to non-UI-auth failures?

src/interactive-auth.js
dbkr: Done

src/interactive-auth.js
dbkr: Ah, maybe - it's not necessary anymore anyway

src/interactive-auth.js
dbkr: done

src/interactive-auth.js
dbkr: Yeah, we probably do want the error. I've made it keep the data in any case and use the flows we had before if they're absent.

dbkr:
Right, ptal. This has taken stuff out of the js-sdk but the react-sdk has gained more parameter passing.

src/interactive-auth.js
```diff
@@ -45,17 +52,50 @@ const utils = require("./utils");
...
  *
- * @param {function(string, object?)} opts.startAuthStage
- *     called to ask the UI to start a particular auth stage. The arguments
- *     are: the login type (eg m.login.password); and (if the last request
- *     returned an error), an error object, with fields 'errcode' and 'error'.
+ * @param {function(string, object?)} opts.stateUpdated
+ *     called when the status of the UI auth changes, ie. when the state of
+ *     an auth stage changes of when the auth flow moves to a new stage.
+ *     The arguments are: the login type (eg m.login.password); and an object
+ *     specific to the login type. These are:
```
dbkr: done

src/interactive-auth.js
```diff
@@ -45,17 +52,50 @@ const utils = require("./utils");
...
+ *     auth, the sid for the email verification session.
  *
  */
 function InteractiveAuth(opts) {
-    this._data = opts.authData;
+    this._matrixClient = opts.matrixClient;
+    this._data = opts.authData || {};
     this._requestCallback = opts.doRequest;
-    this._startAuthStageCallback = opts.startAuthStage;
+    this._stateUpdatedCallback = opts.stateUpdated;
```
dbkr: done

dbkr:
ptal!

src/interactive-auth.js
```diff
@@ -120,7 +212,17 @@ InteractiveAuth.prototype = {
...
-        this._doRequest(auth);
+        this._doRequest(auth, ignoreFailure);
+    },
+
+    /**
+     * Sets the sid for the email validation session
+     * This must be set in order to successfully poll for completion
+     * of the email validation.
+     * Specific to m.login.email.identity
+     */
```
richvdh: needs doc for the `sid` param, as per the lint fail

dbkr: oops, done

PR #390, opened 2017-03-15
Increase BUFFER_PERIOD_MS from 80s to 20mins

lukebarnard1:
Fixes https://github.com/vector-im/riot-web/issues/2737

Seems to work - I forced Chome into a heavily throttled state during /sync (1-10kb/s).
2017-03-15-135556_427x45_scrot

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -37,7 +37,9 @@ const DEBUG = true;
 // beyond that and wedge forever, so we need to track how long we are willing
 // to keep open the connection. This constant is *ADDED* to the timeout= value
 // to determine the max time we're willing to wait.
-const BUFFER_PERIOD_MS = 80 * 1000;
+//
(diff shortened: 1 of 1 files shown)


richvdh:
I'm not sure that we can just bump this to 20 minutes without adding some sort of indicator to the UI that something is going on. It looks to me as though the 20 minute figure was pulled out of the air as an arbitrary value which effectively disables the timeout; I'd like to see a better justification.

How does this look in the UI (a) on first connection, and (b) when reconnecting after an error?

lukebarnard1:
>How does this look in the UI (a) on first connection, and (b) when reconnecting after an error?

It just looks like it's loading the entire time, i.e. a spinner spins for ages.
 
We could show some sort of progress bar based on /sync. Or some indication from the status bar at the top of page like "Disconnected... Reconnecting... Connected" a la facebook messenger.

PR #391, opened 2017-03-15
Give a better error message if the HS doesn't support msisdn registeration

dbkr:
As commented, this isn't really the best way to do this, it just happens to be the only way it can fail right now, and doing it properly is a reasonable chunk of work that it doesn't necessarily make sense to be doing right now. Reckon this is OK for now?

### DIFF
--- a/src/interactive-auth.js
+++ b/src/interactive-auth.js
@@ -383,9 +383,12 @@ InteractiveAuth.prototype = {
         // in general this is not really going to be an accurate error message.
         // Ideally this would signal what inputs could be removed such that a matching
         // flow could be found.
-        throw new Error(
-            "This server does not support registration with a phone number",
-        );
+        const err = new Error("No appropriate authentication flow found");
+        err.required_stages = [];
+        if (haveEmail) err.required_stages.push(EMAIL_STAGE_TYPE);
+        if (haveMsisdn) err.required_stages.push(MSISDN_STAGE_TYPE);
+        err.available_flows = flows;
+        throw err;
     },
 
     /**
--- a/src/interactive-auth.js
+++ b/src/interactive-auth.js
@@ -339,7 +339,9 @@ InteractiveAuth.prototype = {
             // in general this is not really going to be an accurate error message.
             // Ideally this would signal what inputs could be removed such that a matching
             // flow could be found.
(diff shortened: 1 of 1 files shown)


dbkr:
Have changed to emit a better error and added commit to https://github.com/matrix-org/matrix-react-sdk/pull/750 to handle it.

src/interactive-auth.js
```diff
@@ -379,7 +379,16 @@ InteractiveAuth.prototype = {
                 return flow;
             }
         }
-        return null;
+        // XXX: This happens to be the only way this can fail right now, but
+        // in general this is not really going to be an accurate error message.
+        // Ideally this would signal what inputs could be removed such that a matching
+        // flow could be found.
+        const err = new Error("No appropriate authentication flow found");
+        err.required_stages = [];
```
richvdh: I think if we're going to expect apps to catch this, we should give it a `name`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name. `AuthenticationFlowNotFoundError` or something.

Can you also document the function(s) that will throw it?

src/interactive-auth.js
```diff
@@ -379,7 +379,16 @@ InteractiveAuth.prototype = {
                 return flow;
             }
         }
-        return null;
+        // XXX: This happens to be the only way this can fail right now, but
+        // in general this is not really going to be an accurate error message.
```
richvdh: It seems an accurate error message to me?

dbkr: oops, updated

src/interactive-auth.js
dbkr: done

src/interactive-auth.js
```diff
@@ -109,21 +109,25 @@ InteractiveAuth.prototype = {
     /**
      * begin the authentication process.
      *
-     * @return {module:client.Promise}  which resolves to the response on success,
-     * or rejects with the error on failure.
+     * @return {module:client.Promise} which resolves to the response on success,
+     * or rejects with the error on failure. Rejects with NoAuthFlowFoundError} if
```
richvdh: }

PR #392, opened 2017-03-15
Use XHR onprogress to debounce http._request timeout

lukebarnard1:
Instead of just using a timeout to reject ongoing requests, reset the timeout when progress is observed (at least when requests are done from browsers).

This is to fix https://github.com/vector-im/riot-web/issues/2737

Supersedes https://github.com/matrix-org/matrix-js-sdk/pull/390

### DIFF
--- a/src/http-api.js
+++ b/src/http-api.js
@@ -686,11 +686,17 @@ module.exports.MatrixHttpApi.prototype = {
                     handlerFn(err, response, body);
                 },
             );
-            req.onprogress = (e) => {
-                // Prevent the timeout from rejecting the deferred promise if progress is
-                // seen with the request
-                resetTimeout();
-            };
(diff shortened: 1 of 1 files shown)


src/http-api.js
```diff
@@ -679,6 +686,11 @@ module.exports.MatrixHttpApi.prototype = {
                     handlerFn(err, response, body);
                 },
             );
+            req.onprogress = (e) => {
```
richvdh: I think you should comment that this will only work in the browser (where `opts.request` is `browser-request` instead of `request`).

Indeed, I'm not sure about the morality of setting `onprogress` attributes on the `request` object. Perhaps this should be protected by `if ('onprogress' in req)` or something?

lukebarnard1: SGTM

PR #395, opened 2017-03-20
Fix leaking room state objects on limited sync responses

Kegsay:
This PR does 2 things to save memory:
 - It forces strings to be shared between `MatrixEvent` objects. By default they are NOT shared in events because we do `JSON.parse(someString)`, so we do a no-op assignment into a global `interns` map to force V8 to share strings. Strings are immutable so this has no practical effect other than a bit of space saving. We never remove events either, so the strings which are there should be net positive. On my account, this reduces the amount of memory used from 350MB to 320MB.
 - It force clears the non-live timelines on a `limited` sync response. I was pleased to see there was already a flag for this: `flush`. Tested it by spamming `#matrix-test` whilst offline and:
    * Viewing the room (and making sure it jumps down on reconnect)
    * Not viewing the room
  
    Can confirm via heap snapshots it is now reaping correctly. This should fix https://github.com/vector-im/riot-web/issues/3307#issuecomment-282895568

### DIFF
--- a/src/models/event-timeline-set.js
+++ b/src/models/event-timeline-set.js
@@ -187,7 +187,7 @@ EventTimelineSet.prototype.resetLiveTimeline = function(backPaginationToken, flu
     newTimeline.setPaginationToken(backPaginationToken, EventTimeline.BACKWARDS);
 
     this._liveTimeline = newTimeline;
-    this.emit("Room.timelineReset", this.room, this);
+    this.emit("Room.timelineReset", this.room, this, flush);
 };
 
 /**
@@ -655,4 +655,6 @@ module.exports = EventTimelineSet;
  * @event module:client~MatrixClient#"Room.timelineReset"
  * @param {Room} room The room whose live timeline was reset, if any
  * @param {EventTimelineSet} timelineSet timelineSet room whose live timeline was reset
+ * @param {boolean} resetAll The return value for canResetEntireTimeline() for this room.
+ * If true, ALL timeline sets for this room will be reset.
  */
--- a/src/models/event-timeline-set.js
+++ b/src/models/event-timeline-set.js
@@ -154,10 +154,11 @@ EventTimelineSet.prototype.replaceEventId = function(oldEventId, newEventId) {
  * @fires module:client~MatrixClient#event:"Room.timelineReset"
  */
 EventTimelineSet.prototype.resetLiveTimeline = function(backPaginationToken, flush) {
-    let newTimeline;
+    // if timeline support is disabled, forget about the old timelines
+    const resetAllTimelines = !this._timelineSupport || flush;
 
-    if (!this._timelineSupport || flush) {
-        // if timeline support is disabled, forget about the old timelines
+    let newTimeline;
+    if (resetAllTimelines) {
         newTimeline = new EventTimeline(this);
         this._timelines = [newTimeline];
         this._eventIdToTimeline = {};
@@ -187,7 +188,7 @@ EventTimelineSet.prototype.resetLiveTimeline = function(backPaginationToken, flu
     newTimeline.setPaginationToken(backPaginationToken, EventTimeline.BACKWARDS);
 
     this._liveTimeline = newTimeline;
-    this.emit("Room.timelineReset", this.room, this, flush);
+    this.emit("Room.timelineReset", this.room, this, resetAllTimelines);
 };
 
 /**
@@ -655,6 +656,5 @@ module.exports = EventTimelineSet;
  * @event module:client~MatrixClient#"Room.timelineReset"
  * @param {Room} room The room whose live timeline was reset, if any
  * @param {EventTimelineSet} timelineSet timelineSet room whose live timeline was reset
- * @param {boolean} resetAll The return value for canResetEntireTimeline() for this room.
- * If true, ALL timeline sets for this room will be reset.
+ * @param {boolean} resetAllTimelines True if all timelines were reset.
  */
--- a/src/models/event.js
+++ b/src/models/event.js
@@ -49,6 +49,8 @@ module.exports.EventStatus = {
     CANCELLED: "cancelled",
 };
 
+const interns = {};
+
 /**
  * Construct a Matrix Event object
  * @constructor
@@ -75,7 +77,31 @@ module.exports.EventStatus = {
 module.exports.MatrixEvent = function MatrixEvent(
     event,
 ) {
+    ["state_key", "type", "sender", "room_id"].forEach((prop) => {
+        if (!event[prop]) {
+            return;
+        }
+        if (!interns[event[prop]]) {
+            interns[event[prop]] = event[prop];
+        }
+        event[prop] = interns[event[prop]];
+    });
+
+    ["membership", "avatar_url", "displayname"].forEach((prop) => {
+        if (!event.content || !event.content[prop]) {
+            return;
+        }
+        if (!interns[event.content[prop]]) {
+            interns[event.content[prop]] = event.content[prop];
+        }
+        event.content[prop] = interns[event.content[prop]];
+    });
+
+    
+
     this.event = event || {};
+
+
     this.sender = null;
     this.target = null;
     this.status = null;
--- a/src/models/event.js
+++ b/src/models/event.js
@@ -77,6 +77,9 @@ const interns = {};
 module.exports.MatrixEvent = function MatrixEvent(
     event,
 ) {
+    // intern the values of matrix events to force share strings and reduce the
+    // amount of needless string duplication. This can save moderate amounts of
+    // memory (~10% on a 350MB heap).
     ["state_key", "type", "sender", "room_id"].forEach((prop) => {
         if (!event[prop]) {
             return;
@@ -97,10 +100,7 @@ module.exports.MatrixEvent = function MatrixEvent(
         event.content[prop] = interns[event.content[prop]];
     });
 
-    
-
     this.event = event || {};
-
 
     this.sender = null;
     this.target = null;
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -204,15 +204,12 @@ Room.prototype.getLiveTimeline = function() {
  * <p>This is used when /sync returns a 'limited' timeline.
  *
  * @param {string=} backPaginationToken   token for back-paginating the new timeline
+ * @param {boolean=} flush True to remove all events in all timelines. If false, only
+ * the live timeline is reset.
  */
-Room.prototype.resetLiveTimeline = function(backPaginationToken) {
+Room.prototype.resetLiveTimeline = function(backPaginationToken, flush) {
     for (let i = 0; i < this._timelineSets.length; i++) {
-        // Flush out non-live timelines. We do this to reduce the amount of memory
-        // used, as storing multiple distinct copies of room state (each of which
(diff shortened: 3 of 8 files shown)


src/models/event.js
```diff
@@ -75,7 +77,31 @@ module.exports.EventStatus = {
...
+        if (!event[prop]) {
+            return;
+        }
+        if (!interns[event[prop]]) {
+            interns[event[prop]] = event[prop];
+        }
+        event[prop] = interns[event[prop]];
+    });
+
+    ["membership", "avatar_url", "displayname"].forEach((prop) => {
```
richvdh: I wonder if we should worry about `interns` filling up with old displaynames and avatar_urls.

Kegsay: > We never remove events either, so the strings which are there should be net positive.

If we deleted `m.room.member` events then absolutely, we probably would need to start caring. We still keep all previous names/avatar URLs when the user changes their name/avatar, so frequent name changes will still be referenced somewhere in the code.

src/models/room.js
```diff
@@ -207,7 +207,12 @@ Room.prototype.getLiveTimeline = function() {
  */
 Room.prototype.resetLiveTimeline = function(backPaginationToken) {
     for (let i = 0; i < this._timelineSets.length; i++) {
-        this._timelineSets[i].resetLiveTimeline(backPaginationToken);
+        // Flush out non-live timelines. We do this to reduce the amount of memory
```
richvdh: this implies we're *only* flushing out the non-live timelines; I don't think that is correct?

src/models/room.js
```diff
@@ -207,7 +207,12 @@ Room.prototype.getLiveTimeline = function() {
...
+        // used, as storing multiple distinct copies of room state (each of which
+        // can be 10s of MBs) for each timeline is expensive. This is particularly
+        // noticeable when the client unsleeps and there are a lot of 'limited: true'
+        // responses. https://github.com/vector-im/riot-web/issues/3307#issuecomment-282895568
+        this._timelineSets[i].resetLiveTimeline(backPaginationToken, true);
```
richvdh: you're going to need to convince me that you've properly thought through the consequences of clearing `_eventIdToTimeline` while the user is still potentially viewing events in an old timeline. This feels like it's going to break as much as it fixes right now :/.

Kegsay: >  Tested it by spamming #matrix-test whilst offline and viewing the room (and making sure it jumps down on reconnect)

If you want more convincing then state what you would like.

src/models/room.js
Kegsay: Oops,  right you are

src/models/room.js
richvdh: it shouldn't jump down if you're viewing historical events.

A handful of examples: what happens when you scroll down so that the historical timeline merges with the new one? what happens if you then scroll back up? what happens if you jump to a permalink to the events that are now in a timeline which is attached to the live timeline, but are no longer in `_eventIdToTimeline` ?

Mostly though, I think that a change that deliberately leaves the datastructures in an inconsistent state is a bad idea.

Kegsay: > a change that deliberately leaves the datastructures in an inconsistent state is a bad idea.

I'm not, I'm removing all the events in the timeline then adding the ones I got from a `limited` sync. Where is the inconsistency? The event ID to timeline map is cleared when we `flush` to preserve consistency: https://github.com/matrix-org/matrix-js-sdk/blob/master/src/models/event-timeline-set.js#L163

> it shouldn't jump down if you're viewing historical events.

Why? This strictly happens on a `limited` /sync only, which restricts this to startup/load/unsleep. If we *don't* jump down, then the events needs to hang around until some indeterminate point in time when they can be reaped, and it's unclear when that should happen or how that should work, and sounds exceptionally prone to leaking the events beyond the point where you've removed them from the JS SDK object model. Your examples focus on the assumption that the events are still around, and the entire point of this PR is to make them NOT hang around.

If they don't hang around, then it's really quite simple: we drop all events from the room and only populate the most recent 20 messages. The things to check are:
 - Make sure that the other events are GCed.
 - Make sure that if you were viewing some of the older events, that the screen jumps to the bottom as they no longer exist.

Both of those cases I have manually tested and confirmed. I think we may have philosophical differences on how this should behave, so may need @ara4n's opinion on how to proceed.

richvdh:
> I'm not, I'm removing all the events in the timeline then adding the ones I got from a limited sync. Where is the inconsistency?

If the user is viewing historical events, and then forward paginates to the live timeline, the historical events will get joined onto the live timeline, hence: "what happens if you jump to a permalink to the events that are now in a timeline which is attached to the live timeline, but are no longer in _eventIdToTimeline ?"

> > it shouldn't jump down if you're viewing historical events.

> Why? This strictly happens on a limited /sync only, which restricts this to startup/load/unsleep

... or any period where more than 20 messages arrive between /sync calls. If I've carefully scrolled back to a bit of history, and a flurry of activity combined with a network glitch means that I get a limited sync, I am going to be cursing your name if my view suddenly jumps to the bottom.

richvdh:
Anyway: if the application layer is obliged to make sure it has dropped all references to historical timelines on a limited sync (which I still think is a bad idea for the UX), you need to plaster that big and bold across the APIs, rather than discovering that it appears to be ok in Riot and calling it job done.

richvdh:
I'm not quite sure where the disconnect is happening here - we are obviously still somehow at cross-purposes. So let me try to be absolutely clear.

I fired up a riot build with your branch of the js-sdk, and did the following:

* scroll back a little way in the room. Grab a permalink while you are there.
* go offline
* spam a bunch of messages into the room to trigger a limited sync
* go back online (wait for the sync to notice)
* scroll back down.
* You now have events in the timeline which are not in `_eventIdToTimeline`. (I checked.)
* Navigate to the permalink (I did this just by sticking the event ID in the URL bar)
* The console logs now say: 
```
event-timeline-set.js:400 Already have timeline for $1490131020162teLDt:fred.sw1v.org - joining timeline !RoiHwUcCARaykMwyRo:fred.sw1v.org:2017-03-21T21:26:38.955Z to !RoiHwUcCARaykMwyRo:fred.sw1v.org:2017-03-21T21:24:04.933Z
event-timeline.js:217 Uncaught Error: timeline already has a neighbouring timeline - cannot reset neighbour
    at EventTimeline.setNeighbouringTimeline (event-timeline.js:217)
    at EventTimelineSet.addEventsToTimeline (event-timeline-set.js:404)
    at client.js:1963
    at _fulfilled (q.js:834)
    at self.promiseDispatch.done (q.js:863)
    at Promise.promise.promiseDispatch (q.js:796)
    at q.js:604
    at runSingle (q.js:137)
    at flush (q.js:125)
    at onNextTick (main.js:64)
```

This process used to work fine. TBH, it still does, modulo the scary error in the console, but I think that errors like that in the console logs are unacceptable, and in this case symptomatic of other problems.

Let's discuss what to do about it when we meet in person.

Kegsay:
> scroll back down.

This is the bit where we are crossing purposes. I can't scroll down, since the instant it syncs the timeline flashes and jumps to the bottom. Other than that,  your steps sound the same as mine.

richvdh:
Weird. It used to work, and it worked for me in testing last night. 

On 22 March 2017 08:32:15 GMT+00:00, Kegsay <notifications@github.com> wrote:
>> scroll back down.
>
>This is the bit where we are crossing purposes. I can't scroll down,
>since the instant it syncs the timeline flashes and jumps to the
>bottom. Other than that,  your steps sound the same as mine. 
>
>-- 
>You are receiving this because you were assigned.
>Reply to this email directly or view it on GitHub:
>https://github.com/matrix-org/matrix-js-sdk/pull/395#issuecomment-288330684

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Kegsay:
Right, I think I know why. We do different things depending on the scroll position when the timeline resyncs. If it's at the **bottom** things work correctly and the timeline reloads, else it ignores the reset request. It sounds like we've been hitting one or the other case which is why we don't agree on the results. 

https://github.com/matrix-org/matrix-react-sdk/blob/master/src/components/structures/TimelinePanel.js#L429

richvdh:
Yes, that's what I've been trying to tell you...

Kegsay:
Conclusions from IRL:
 - The PR in its current state causes inconsistencies in the timeline because the app may have references to older parts of the timeline which will now get removed.
 - Due to this, it is ultimately not possible for the SDK to dictate when it is safe to reap events from a timeline: the SDK needs to ask the app if it is safe to reset the entire timeline for a room.
 - Proposal is to add a function to the `opts` of `MatrixClient`: `canResetEntireTimeline(roomId) => bool` which the SDK will call when it wants to reset the timeline (on a `limited` sync).
 - By default, this function always returns `false` which preserves the original behaviour.
 - The result of this function is directly passed as the `flush` flag: this means if you can reset the entire timeline for a room, then flush all non-live timelines.
 - An extra arg will be added to the `Room.timelineReset` event which fires, which indicates whether the app said it was safe to reset the entire timeline.

React SDK will implement this function and will always return `true` UNLESS:
 - The user is currently viewing the room AND
 - The timeline panel is not at the bottom.

@richvdh Does this sound accurate?

richvdh:
ja think so

Anything you can do to improve crappy jsdocs while you are in there would be very much appreciated

Kegsay:
@richvdh PTAL

src/models/event-timeline-set.js
```diff
@@ -187,7 +187,7 @@ EventTimelineSet.prototype.resetLiveTimeline = function(backPaginationToken, flu
     newTimeline.setPaginationToken(backPaginationToken, EventTimeline.BACKWARDS);
 
     this._liveTimeline = newTimeline;
-    this.emit("Room.timelineReset", this.room, this);
+    this.emit("Room.timelineReset", this.room, this, flush);
```
richvdh: any chance we can improve the name of `flush` (`resetAll` seems fine) and clarify what it means to flush the non-live timelines?

richvdh: also, should be `flush || !this._timelineSupport`. Or better yet, factor out the condition so that they don't get out of sync.

src/models/event-timeline-set.js
```diff
@@ -655,4 +655,6 @@ module.exports = EventTimelineSet;
  * @event module:client~MatrixClient#"Room.timelineReset"
  * @param {Room} room The room whose live timeline was reset, if any
  * @param {EventTimelineSet} timelineSet timelineSet room whose live timeline was reset
+ * @param {boolean} resetAll The return value for canResetEntireTimeline() for this room.
```
richvdh: this feels like a bit of a backwards description to me. At this point, as an app, I am interested in what the js-sdk did (ie, reset all the timelines - again, needs a better description), rather than why it did it.

Kegsay: Done. https://github.com/matrix-org/matrix-js-sdk/pull/395/commits/14727d75ac90b49e25b720b01e8cadbc122ff421

src/models/event-timeline-set.js
Kegsay: Done. https://github.com/matrix-org/matrix-js-sdk/pull/395/commits/14727d75ac90b49e25b720b01e8cadbc122ff421

PR #396, opened 2017-03-21
Extend 'ignoreFailure' to be 'background'

dbkr:
This allows us to also use it to decide whether or not to show
the app as busy in the UI. We pass this flag up into the
makeRequest callback so it can use it as such.


richvdh:
commit summaries could do with some work, mind: they don't mention that its about interactive auth.

PR #400, opened 2017-03-23
Fix undefined reference in http-api

dbkr:
Put the check for 'onprogress' within the check to see if req is
defined, because sometimes it isn't, apparently.


ara4n:
lgtm

PR #401, opened 2017-03-24
Wrong variable name is used.

pavankarthikboddeda:
Ran into this while trying this out.It should be `|client.on("sync")`|` instead |matrixClient.on("sync")| .
tested it, works fine :)

### DIFF
--- a/examples/voip/browserTest.js
+++ b/examples/voip/browserTest.js
@@ -44,7 +44,7 @@ window.onload = function() {
     disableButtons(true, true, true);
 };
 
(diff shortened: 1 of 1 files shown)


dbkr:
I've merged this manually as you're targeting the wrong branch (and have not signed off on your contribution, although given I think we basically arrived at this fix mutually on Riot, it's probably OK in this case).

pavankarthikboddeda:
Hey dbkr,
What is the right branch then?
and what do you mean by "have not signed off on your contribution" ?
Thanks

PR #405, opened 2017-03-28
Don't be SYNCING until updating from the server

dbkr:
Syncing should probably mean the stream is up to date and
streaming messages in real-time from the server, which is not the
case if we've only loaded the cached response. Stay PREPARED until
we actually get the latest from the server.


Kegsay:
LGTM!

PR #406, opened 2017-03-31
Change interface of IndexedDBStore: hide internals

dbkr:
Hide the IndexedDBBackend and SyncAccumulator objects and have
indexeddbstore instaniate them itself, rather than making the app
instantiate them. Pass the approipriate options through. This
gives us much more flexibility to move things around under the
hood.

### DIFF
--- a/src/store/indexeddb.js
+++ b/src/store/indexeddb.js
@@ -234,7 +234,6 @@ IndexedDBStoreBackend.prototype = {
  *
  * @constructor
  * @extends MatrixInMemoryStore
- * loaded from IndexedDB and periodically saved to IndexedDB.
  * @param {Object} opts Options optject.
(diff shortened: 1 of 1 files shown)


src/store/indexeddb.js
```diff
@@ -234,29 +234,31 @@ IndexedDBStoreBackend.prototype = {
  *
  * @constructor
  * @extends MatrixInMemoryStore
- * @param {IndexedDBStoreBackend} backend The indexed db backend instance.
- * @param {SyncAccumulator} syncAccumulator The sync accumulator which will be
- * loaded from IndexedDB and periodically saved to IndexedDB.
- * @param {Object=} opts Options for MatrixInMemoryStore.
+ * @param {Object} opts Options optject.
```
Kegsay: > optject

PR #407, opened 2017-03-31
Further reorganising of indexeddb sync code

dbkr:
* Make sync communicate with the sync accumulator via the store
 * Consequently get rid of getSyncAccumulator as it's now
   unnecessary.
 * Make the bit that gets the saved sync response async, because
   we'll need it to be when it's coming over postMessage from a
   webworker.

Based off of https://github.com/matrix-org/matrix-js-sdk/pull/406

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -60,8 +60,6 @@ function debuglog() {
  * @param {MatrixClient} client The matrix client instance to use.
  * @param {Object} opts Config options
  * @param {module:crypto=} opts.crypto Crypto manager
- * kept up-to-date. If one is supplied, the response to getJSON() will be used
- * initially.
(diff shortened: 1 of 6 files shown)


src/sync.js
```diff
@@ -60,7 +60,6 @@ function debuglog() {
  * @param {MatrixClient} client The matrix client instance to use.
  * @param {Object} opts Config options
  * @param {module:crypto=} opts.crypto Crypto manager
- * @param {SyncAccumulator=} opts.syncAccumulator An accumulator which will be
  * kept up-to-date. If one is supplied, the response to getJSON() will be used
  * initially.
```
Kegsay: Presumably these 2 lines also need to go?

PR #409, opened 2017-04-04
Move more functionality to the indexeddb backend

dbkr:
Now the backend drives the sync accumulator as well. Also moves
the backend out to a separate file.

### DIFF
--- a/src/store/indexeddb-local-backend.js
+++ b/src/store/indexeddb-local-backend.js
@@ -1,0 +1,316 @@
+/*
+Copyright 2017 Vector Creations Ltd
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


src/store/indexeddb-local-backend.js
```diff
@@ -0,0 +1,316 @@
...
+ *
+ * Construct a new Indexed Database store backend. This requires a call to
+ * <code>connect()</code> before this store can be used.
+ * @constructor
+ * @param {Object} indexedDBInterface The Indexed DB interface e.g
+ * <code>window.indexedDB</code>
+ * @param {string=} dbName Optional database name. The same name must be used
+ * to open the same database.
+ */
+const LocalIndexedDBStoreBackend = function LocalIndexedDBStoreBackend(
```
Kegsay: Why the `Local` prefix? Is there a non-local one?

Kegsay:
Otherwise probably LGTM - There's a lot of moved lines which unfortunately makes it hard to see where things have been tweaked. Superficially looks the same.

src/store/indexeddb-local-backend.js
dbkr: Not yet, but there will be.

Kegsay: ack

PR #410, opened 2017-04-04
Indicate syncState ERROR after many failed /syncs

lukebarnard1:
when a /sync leads to an error, increase a counter and use the counter to decide which state to be in when starting keepAlives. If the count is above a certain threshold (arbitrary 3 chosen here), switch from RECONNECTING to ERROR to show that the server is facing difficulties that aren't affeting is ability to return on /versions.

Attempts to fix https://github.com/vector-im/riot-web/issues/3336

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -39,6 +39,13 @@ const DEBUG = true;
 // to determine the max time we're willing to wait.
 const BUFFER_PERIOD_MS = 80 * 1000;
 
+// Number of consecutive failed /sync requests
+let failedSyncCount = 0;
+// Number of consecutive failed syncs that will lead to a syncState of ERROR as opposed
+// to RECONNECTING. This is needed to inform the client of server issues when the
+// keepAlive is successful but the server /sync fails.
+const FAILED_SYNC_ERROR_THRESHOLD = 3;
+
 function getFilterName(userId, suffix) {
     // scope this on the user ID because people may login on many accounts
     // and they all need to be stored!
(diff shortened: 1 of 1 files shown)


src/sync.js
```diff
@@ -39,6 +39,13 @@ const DEBUG = true;
 // to determine the max time we're willing to wait.
 const BUFFER_PERIOD_MS = 80 * 1000;
 
+// Number of consecutive failed /sync requests
+let failedSyncCount = 0;
```
dbkr: This should be a member variable on SyncApi to match all the others (there could be multiple SyncApi objects).

src/sync.js
```diff
@@ -610,6 +617,9 @@ SyncApi.prototype._sync = function(syncOptions) {
         client.store.save();
 
         self._sync(syncOptions);
+
+        // Reset after a successful sync
+        failedSyncCount = 0;
```
dbkr: This wants to be within the _sync function in the `done()` (somewhere around `client.store.setSyncToken(data.next_batch);`). This doesn't do what you think because _sync doesn't block, ie. this will set `failedSyncCount` to 0 right after *starting* a sync request.

src/sync.js
```diff
@@ -635,7 +648,10 @@ SyncApi.prototype._sync = function(syncOptions) {
             self._sync(syncOptions);
         });
         self._currentSyncRequest = null;
-        self._updateSyncState("RECONNECTING");
+        // Transition from RECONNECTING to ERROR after a given number of failed syncs
+        self._updateSyncState(
+            failedSyncCount > FAILED_SYNC_ERROR_THRESHOLD ? "ERROR" : "RECONNECTING",
```
dbkr: Minor, but this probably wants to be >= rather than > so the ERROR state is triggered after FAILED_SYNC_ERROR_THRESHOLD failed syncs which is probably just a bit more intuitive than after n + 1 failed syncs which is what this does.

PR #412, opened 2017-04-06
Add ability to do indexeddb sync work in webworker

dbkr:
This adds RemoteIndexedDBStoreBackend which spools up a LocalIndexedDBStoreBackend in a webworker via the provided script (if one is provided) and runs the indexeddb queries on it.

Things known to suck in this PR (and are commented in the code):
 * Sync data is pulled over to the main script and then sent back to the web worker for no reason
 * It works around the fact that sync data is modified for some reason by waiting for the data to be sent over to the webworker. There are a lot of changes to the tests to make them not dependent on the sync processing happening in a single event loop cycle.

These things don't affect much apart from making startup slower, so I'll fix them in a separate PR.

https://github.com/vector-im/riot-web/issues/3252

### DIFF
--- a/spec/test-utils.js
+++ b/spec/test-utils.js
@@ -1,8 +1,22 @@
 "use strict";
 import expect from 'expect';
+import q from 'q';
 
 const sdk = require("..");
 const MatrixEvent = sdk.MatrixEvent;
+
+/**
+ * Return a promise that is resolved when the client next emits a
+ * SYNCING event.
+ * @param {Object} client The client
+ */
+module.exports.syncPromise = function(client) {
+    const def = q.defer();
+    client.on('sync', (state) => {
+        if (state == 'SYNCING') def.resolve();
+    });
+    return def.promise;
+};
 
 /**
  * Perform common actions before each test case, e.g. printing the test case
--- a/spec/test-utils.js
+++ b/spec/test-utils.js
@@ -9,6 +9,7 @@ const MatrixEvent = sdk.MatrixEvent;
  * Return a promise that is resolved when the client next emits a
  * SYNCING event.
  * @param {Object} client The client
+ * @return {Promise} Resolves once the client has emitted a SYNCING event
  */
 module.exports.syncPromise = function(client) {
     const def = q.defer();
--- a/spec/test-utils.js
+++ b/spec/test-utils.js
@@ -13,9 +13,14 @@ const MatrixEvent = sdk.MatrixEvent;
  */
 module.exports.syncPromise = function(client) {
     const def = q.defer();
-    client.on('sync', (state) => {
-        if (state == 'SYNCING') def.resolve();
-    });
+    const cb = (state) => {
+        if (state == 'SYNCING') {
+            def.resolve();
+        } else {
+            client.once('sync', cb);
+        }
+    };
+    client.once('sync', cb);
     return def.promise;
 };
 
--- a/src/matrix.js
+++ b/src/matrix.js
@@ -61,6 +61,8 @@ module.exports.Filter = require("./filter");
 module.exports.TimelineWindow = require("./timeline-window").TimelineWindow;
 /** The {@link module:interactive-auth} class. */
 module.exports.InteractiveAuth = require("./interactive-auth");
+/** The {@link module:indexeddb-remote-worker} class. */
+module.exports.IndexedDbStoreWorker = require("./store/indexeddb-remote-worker.js");
 
 
 /**
--- a/src/matrix.js
+++ b/src/matrix.js
@@ -61,8 +61,8 @@ module.exports.Filter = require("./filter");
 module.exports.TimelineWindow = require("./timeline-window").TimelineWindow;
 /** The {@link module:interactive-auth} class. */
 module.exports.InteractiveAuth = require("./interactive-auth");
-/** The {@link module:indexeddb-remote-worker} class. */
-module.exports.IndexedDbStoreWorker = require("./store/indexeddb-remote-worker.js");
+/** The {@link module:indexeddb-store-worker~IndexedDBStoreWorker} class. */
+module.exports.IndexedDBStoreWorker = require("./store/indexeddb-store-worker.js");
 
 
 /**
--- a/src/matrix.js
+++ b/src/matrix.js
@@ -61,8 +61,6 @@ module.exports.Filter = require("./filter");
 module.exports.TimelineWindow = require("./timeline-window").TimelineWindow;
 /** The {@link module:interactive-auth} class. */
 module.exports.InteractiveAuth = require("./interactive-auth");
-/** The {@link module:indexeddb-store-worker~IndexedDBStoreWorker} class. */
-module.exports.IndexedDBStoreWorker = require("./store/indexeddb-store-worker.js");
 
 
 /**
--- a/src/store/indexeddb-local-backend.js
+++ b/src/store/indexeddb-local-backend.js
@@ -15,8 +15,6 @@ limitations under the License.
 */
 
 import q from "q";
-import User from "../models/user";
-import {MatrixEvent} from "../models/event";
 import SyncAccumulator from "../sync-accumulator";
 import utils from "../utils";
 
@@ -104,13 +102,6 @@ const LocalIndexedDBStoreBackend = function LocalIndexedDBStoreBackend(
     this._dbName = "matrix-js-sdk:" + (dbName || "default");
     this.db = null;
     this._syncAccumulator = new SyncAccumulator();
-
-    // Records the last-modified-time of each user at the last point we saved
-    // the database, such that we can derive the set if users that have been
-    // modified since we last saved.
-    this._userModifiedMap = {
-        // user_id : timestamp
-    };
 };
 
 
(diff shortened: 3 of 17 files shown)


src/store/indexeddb.js
```diff
@@ -88,24 +107,31 @@ IndexedDBStore.prototype.startup = function() {
...
+        userPresenceEvents.forEach(([userId, rawEvent]) => {
+            const u = new User(userId);
+            if (rawEvent) {
+                u.setPresenceEvent(new MatrixEvent(rawEvent));
+            }
+            this._userModifiedMap[u.userId] = u.getLastModifiedTime();
             this.storeUser(u);
         });
         this._syncTs = Date.now(); // pretend we've written so we don't rewrite
-        this.setSyncToken(syncData.nextBatch);
```
dbkr: See https://github.com/matrix-org/matrix-js-sdk/pull/412/commits/0594a8d03a939df1e379013f53a71e8b095bbcae for why this is correct

spec/test-utils.js
```diff
@@ -1,10 +1,25 @@
...
 
 /**
+ * Return a promise that is resolved when the client next emits a
+ * SYNCING event.
+ * @param {Object} client The client
+ * @return {Promise} Resolves once the client has emitted a SYNCING event
+ */
+module.exports.syncPromise = function(client) {
+    const def = q.defer();
+    client.on('sync', (state) => {
```
Kegsay: Nit: We should probably be using [once](https://nodejs.org/api/events.html#events_emitter_once_eventname_listener) to clean up that listener after it is invoked.

src/matrix.js
```diff
@@ -61,6 +61,8 @@ module.exports.Filter = require("./filter");
 module.exports.TimelineWindow = require("./timeline-window").TimelineWindow;
 /** The {@link module:interactive-auth} class. */
 module.exports.InteractiveAuth = require("./interactive-auth");
+/** The {@link module:indexeddb-remote-worker} class. */
+module.exports.IndexedDbStoreWorker = require("./store/indexeddb-remote-worker.js");
```
Kegsay: Let's be consistent on the casing please. We have `IndexedDBStore` and `IndexedDBStoreBackend`, so `IndexedDBStoreWorker` would be most preferable.

src/store/indexeddb-remote-worker.js
```diff
@@ -0,0 +1,104 @@
+/*
```
Kegsay: This file is called `src/store/indexeddb-remote-worker.js` and yet defines a single class called `IndexedDbStoreWorker`. Can we please keep them consistent and call it `indexeddb-store-worker`?

src/store/indexeddb-local-backend.js
```diff
@@ -239,42 +223,39 @@ LocalIndexedDBStoreBackend.prototype = {
     },
 
     /**
-     * Persist a list of User objects. Users with the same 'userId' will be
-     * replaced.
-     * @param {User[]} users An array of users
+     * Persist a list of [user id, presence event] they are for.
+     * Users with the same 'userId' will be replaced.
+     * Prfesence events should be the event in its raw form (not the Event
```
Kegsay: > Prfesence

src/store/indexeddb.js
```diff
@@ -64,6 +67,10 @@ const WRITE_DELAY_MS = 1000 * 60 * 5; // once every 5 minutes
  * <code>window.indexedDB</code>
  * @param {string=} opts.dbName Optional database name. The same name must be used
  * to open the same database.
+ * @param {string=} opts.workerScript Optional URL to a script to invooke a web
```
Kegsay: > invooke

src/store/indexeddb-remote-worker.js
```diff
@@ -0,0 +1,104 @@
...
+
+import q from "q";
+import LocalIndexedDBStoreBackend from "./indexeddb-local-backend.js";
+
+/**
+ * This class lives in the webworker and drives a LocalIndexedDbStoreBackend
+ * controlled by messages from the main process.
+ */
+class IndexedDbStoreWorker {
+    constructor(postMessage) {
```
Kegsay: I'm unhappy with the level of documentation on this class. We don't seem to use it at all in this PR, but we do in a riot-web PR here: https://github.com/vector-im/riot-web/pull/3578/files#diff-34344b4ed4dd336568358240bfdfb066R26

We fail to mention that:
 - `onMessage` should be hooked into `window.onmessage`.
 - `postMessage` in the constructor should be `window.postMessage`.
 - This class should be instantiated in the script which will be fed into the `IndexedDBStore`'s `workerScript`.

Please give an example of the script (basically the entirety of https://github.com/vector-im/riot-web/pull/3578/files#diff-34344b4ed4dd336568358240bfdfb066R26 probably with fully-qualified `window.` for clarity rather than assumed global variables) in the constructor for this class.

src/store/indexeddb-remote-worker.js
```diff
@@ -0,0 +1,104 @@
...
+            case 'loadSyncData':
+                prom = this.backend.loadSyncData();
+                break;
+        }
+
+        if (prom === undefined) {
+            postMessage({
+                command: 'cmd_fail',
+                seq: msg.seq,
+                // Canb't be an Error because they're not structured cloneable
```
Kegsay: > Canb't

src/store/indexeddb-remote-worker.js
```diff
@@ -0,0 +1,104 @@
...
+                );
+                prom = q();
+                break;
+            case 'connect':
+                prom = this.backend.connect();
+                break;
+            case 'clearDatabase':
+                prom = this.backend.clearDatabase().then((result) => {
+                    // This returns special classes which can't be cloned
+                    // accross to the main script, so don't try.
```
Kegsay: > accross

src/store/indexeddb-remote-worker.js
```diff
@@ -0,0 +1,104 @@
...
+                    return {};
+                });
+                break;
+            case 'getSavedSync':
+                prom = this.backend.getSavedSync(false);
+                break;
+            case 'setSyncData':
+                prom = this.backend.setSyncData(...msg.args);
```
Kegsay: Is `msg.args` always guaranteed to be an array?

src/store/indexeddb-remote-worker.js
```diff
@@ -0,0 +1,104 @@
...
+            this.postMessage.call(null, {
+                command: 'cmd_success',
+                seq: msg.seq,
+                result: ret,
+            });
+        }, (err) => {
+            this.postMessage.call(null, {
+                command: 'cmd_fail',
+                seq: msg.seq,
+                error: err,
```
Kegsay: You presumably cannot do this because as you say:

> // Can't be an Error because they're not structured cloneable

src/store/indexeddb-local-backend.js
```diff
@@ -239,42 +223,39 @@ LocalIndexedDBStoreBackend.prototype = {
...
+     * object)
+     * @param {Object[]} tuples An array of [userid, event] tuples
```
Kegsay: Why tuples, which you need to consistently remember which index has what, over an `Object` where the key names are self-documenting?

src/store/indexeddb-remote-backend.js
```diff
@@ -0,0 +1,155 @@
...
+    this._worker = new Worker(workerScript);
+    this._nextSeq = 0;
+    // The currently in-flight requests to the actual backend
+    this._inFlight = {
+        // seq: promise,
+    };
+
+    this._worker.onmessage = this._onWorkerMessage.bind(this);
+
+    this._doCmd('_setupWorker', [dbName]).done(() => {
```
Kegsay: This looks like it will race with functions on this class. Do web workers guarantee any kind of ordering to the messages we post? If they do, then this is probably fine since this command must be delivered before the function call commands, and `_setupWorker` synchronously sets up `this.backend`. This certainly warrants a comment either way.

src/store/indexeddb.js
Kegsay: SGTM!

Kegsay:
Otherwise this is looking good.

spec/test-utils.js
dbkr: Done

src/store/indexeddb-remote-worker.js
dbkr: done

src/store/indexeddb-local-backend.js
dbkr: done

src/store/indexeddb.js
dbkr: done

src/store/indexeddb-remote-worker.js
dbkr: Done, but the lack of window is intentional: it's a web worker, there is no window, the variables live in the scope the worker runs in.

src/store/indexeddb-remote-worker.js
dbkr: done

src/store/indexeddb-remote-worker.js
dbkr: done

src/store/indexeddb-remote-worker.js
dbkr: Yep

src/store/indexeddb-remote-worker.js
dbkr: done

src/store/indexeddb-local-backend.js
dbkr: Happy to change if you prefer. Tuples will be a little smaller and so go over postMessage faster. For 2 things I personally think a tuple is fine. It's an internal interface so trivial to change to objects if we want.

src/store/indexeddb-remote-backend.js
dbkr: Good point: I believe postMessage does guarantee ordering, but since we only ever connect once I've just moved it to connect so it's all nice & serialised.

src/store/indexeddb-remote-worker.js
Kegsay: Oh interesting wrt scoping, didn't know that!

src/store/indexeddb-remote-worker.js
Kegsay: Carry on then!

src/store/indexeddb-local-backend.js
Kegsay: > Tuples will be a little smaller

Source? Heapdumping an array and an object with the same data (and `userId` and `event` keys actually shows that:
```
Object form
  shallow size = 40 bytes
  retained size = 40 bytes

Array form
  shallow size = 32 bytes
  retained size = 64 bytes
```

Tested via:
```
window.ALPHABAT = {}
window.ALPHABAT.foo = ["alpha","bet"]
window.ALPHABAT.bar = {userId:"alpha",event:"bet"}
```

If `postMessage` sends over actual JSON then sure, array form will indeed be slightly smaller. I don't like how we've prematurely optimised for that at the sake of clarity, but as you say, it's an internal interface and I don't care enough to make you change this.

Kegsay:
The only outstanding thing from what I can see now is https://github.com/vector-im/riot-web/pull/3578#discussion_r110343392 - I need your view on this given it affects the comments in this PR. Other than that, LGTM!

Kegsay:
LGTM!

PR #413, opened 2017-04-06
Make indexeddb startup faster

dbkr:
Don't needlessly transfer data from the worker to the main script
only to immediately transfer it right back again.

This is based off https://github.com/matrix-org/matrix-js-sdk/pull/412

### DIFF
--- a/src/store/indexeddb-local-backend.js
+++ b/src/store/indexeddb-local-backend.js
@@ -133,13 +133,15 @@ LocalIndexedDBStoreBackend.prototype = {
             this.db.onversionchange = () => {
                 this.db.close();
             };
+
+            return this._init();
         });
     },
 
     /**
      * Having connected, load initial data from the database and prepare for use
      */
-    init: function() {
+    _init: function() {
         return q.all([
             this._loadAccountData(),
             this._loadSyncData(),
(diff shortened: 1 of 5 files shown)


src/store/indexeddb.js
```diff
@@ -111,17 +111,10 @@ IndexedDBStore.prototype.startup = function() {
     // the web worker to then immediately push it back again without
     // using it.
```
Kegsay: This comment is presumably now wrong.

src/store/indexeddb-local-backend.js
```diff
@@ -137,6 +137,24 @@ LocalIndexedDBStoreBackend.prototype = {
     },
 
     /**
+     * Having connected, load initial data from the database and prepare for use
+     */
+    init: function() {
```
Kegsay: Is there *ever* a reason why we would want to `connect()` and NOT call `init()`? Can we bundle them together? Otherwise, it's not really clear which should come first: connect then init, or init then connect?

Kegsay:
If we can, I'd like to see `init` and `connect` bundled together into one function call. If that's not possible, I'd like to see `init` renamed to something like `loadStoredData`, because it's slightly more obvious that you need to connect to the database before you can load, whereas "initing" **after** you connect seems weird and unclear.

src/store/indexeddb-local-backend.js
dbkr: Not really - I wondered about this but ended up just adding one. Have changed it.

Kegsay:
LGTM

PR #414, opened 2017-04-06
Make indexeddb save after the first sync

dbkr:
Save the sync data to indexeddb after the first catch-up

Fixes vector-im/riot-web#3527

This is based off of https://github.com/matrix-org/matrix-js-sdk/pull/413


Kegsay:
Very sensible, I should've spotted that when I wrote this. LGTM! (though blocked on the stacked PRs behind it)

PR #415, opened 2017-04-07
Fix release script

dbkr:
Publish to npm before switching to the doc branch: previously we
published from master, but since we now now longer merge
pre-releases to master, publish from the release branch (just
not the doc branch because that won't work).

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -217,6 +217,9 @@ fi
 rm "${release_text}"
 rm "${latest_changes}"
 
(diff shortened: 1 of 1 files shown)


richvdh:
(ftr this got broken by me in https://github.com/matrix-org/matrix-js-sdk/commit/cb88f53587cbb8cacf5f88b608c38cc6d9e060a7 :/ )

PR #416, opened 2017-04-07
Add feature detection for webworkers

dbkr:
Only use web worker store if we have web workers available


Kegsay:
It feels like this check is at the wrong point in the stack. It means Node can't ever use the web worker version. There are npm packages which meet the web worker API but we can't inject that either. Can't we put this in react SDK?

dbkr:
Does this tick that box?

Kegsay:
LGTM!

PR #418, opened 2017-04-10
Don't leave the gh-pages branch checked out

dbkr:
After a pre-release, check out the release branch again rather
than leaving the working copy on the gh-pages branch

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -238,6 +238,7 @@ fi
 # if it is a pre-release, leave it on the release branch for now.
 if [ $prerelease -eq 1 ]; then
     exit 0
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -238,6 +238,7 @@ fi
 # if it is a pre-release, leave it on the release branch for now.
 if [ $prerelease -eq 1 ]; then
     exit 0
+    git checkout "$rel_branch"
```
richvdh: before `exit 0`, surely?

dbkr: heh, good point

richvdh:
> lgtm otherwise

PR #419, opened 2017-04-12
Implement API for setting RM

lukebarnard1:
This is now stored on the server with similar treatment to RRs. The server will only store the specified eventId as the current read marker for a room if the event is ahead in the stream when compared to the existing RM. The exception is when the RM has never been set for this room for this user, in which case the event ID will be stored as the RM without any comparison.

This API also allows for an optional RR event ID to be sent in the same request. This is because it might be the common case for some clients to update the RM at the same time as updating the RR.

See design: https://docs.google.com/document/d/1UWqdS-e1sdwkLDUY0wA4gZyIkRp-ekjsLZ8k6g_Zvso/edit

See server-side PR: https://github.com/matrix-org/synapse/pull/2120

### DIFF
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -470,6 +470,33 @@ MatrixBaseApis.prototype.roomInitialSync = function(roomId, limit, callback) {
     );
 };
 
+/**
+ * Set a marker to represent which event the user was last reading in a room. This can
+ * be retrieved from room account data and displayed as a horizontal line in the client
+ * that is visually distinct to the position of the user's own read receipt.
+ * @param {string} roomId ID of the room that has been read
+ * @param {string} rmEventId ID of the event that has been read
+ * @param {string} rrEventId ID of the event tracked by the read receipt. This is here
+ * for convenience because the RR and the RM are commonly updated at the same time as
+ * each other. Optional.
+ * @return {module:client.Promise} Resolves: (@see module:http-api.authedRequest)
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixBaseApis.prototype.setRoomReadMarkers = function(roomId, rmEventId, rrEventId) {
+    const path = utils.encodeUri("/rooms/$roomId/read_marker", {
+        $roomId: roomId,
+    });
+
+    const content = {
+        "m.read_marker": rmEventId,
+        "m.read": rrEventId,
+    };
+
+    return this._http.authedRequest(
+        null, "POST", path, undefined, content,
+    );
+};
+
 
 // Room Directory operations
 // =========================
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -483,12 +483,12 @@ MatrixBaseApis.prototype.roomInitialSync = function(roomId, limit, callback) {
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
 MatrixBaseApis.prototype.setRoomReadMarkers = function(roomId, rmEventId, rrEventId) {
-    const path = utils.encodeUri("/rooms/$roomId/read_marker", {
+    const path = utils.encodeUri("/rooms/$roomId/read_markers", {
         $roomId: roomId,
     });
 
     const content = {
-        "m.read_marker": rmEventId,
+        "m.fully_read": rmEventId,
         "m.read": rrEventId,
     };
 
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -471,16 +471,16 @@ MatrixBaseApis.prototype.roomInitialSync = function(roomId, limit, callback) {
 };
 
 /**
- * Set a marker to represent which event the user was last reading in a room. This can
- * be retrieved from room account data and displayed as a horizontal line in the client
- * that is visually distinct to the position of the user's own read receipt.
+ * Set a marker to indicate the point in a room before which the user has read every
+ * event. This can be retrieved from room account data (the event type is `m.fully_read`)
+ * and displayed as a horizontal line in the timeline that is visually distinct to the
+ * position of the user's own read receipt.
  * @param {string} roomId ID of the room that has been read
  * @param {string} rmEventId ID of the event that has been read
  * @param {string} rrEventId ID of the event tracked by the read receipt. This is here
  * for convenience because the RR and the RM are commonly updated at the same time as
  * each other. Optional.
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -843,6 +843,43 @@ MatrixClient.prototype.setRoomAccountData = function(roomId, eventType,
...
+ * @param {string} eventId ID of the event that has been read
+ * @param {string} rrEvent the event tracked by the read receipt. This is here for
+ * convenience because the RR and the RM are commonly updated at the same time as each
+ * other. Optional.
+ * @param {module:client.callback} callback Optional.
+ * @return {module:client.Promise} Resolves: TODO
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixClient.prototype.setRoomReadMarker = function(roomId, eventId, rrEvent, callback) {
+    if (typeof rrEventId === 'function') {
```
richvdh: why is this a thing? some functions have this sort of hackery, but it's just for backwards compat. You shouldn't need it for a new func

src/client.js
```diff
@@ -843,6 +843,43 @@ MatrixClient.prototype.setRoomAccountData = function(roomId, eventType,
 };
 
 /**
+ * @param {string} roomId ID of the room that has been read
+ * @param {string} eventId ID of the event that has been read
+ * @param {string} rrEvent the event tracked by the read receipt. This is here for
+ * convenience because the RR and the RM are commonly updated at the same time as each
+ * other. Optional.
+ * @param {module:client.callback} callback Optional.
```
richvdh: we're deprecating `callback`s and not adding them for new funcs

src/client.js
```diff
@@ -843,6 +843,43 @@ MatrixClient.prototype.setRoomAccountData = function(roomId, eventType,
...
+ * @return {module:client.Promise} Resolves: TODO
```
richvdh: can you find something better than this please

src/client.js
```diff
@@ -843,6 +843,43 @@ MatrixClient.prototype.setRoomAccountData = function(roomId, eventType,
 };
 
 /**
```
richvdh: what does it DOOOOOOOOOO?

src/client.js
lukebarnard1: Would it to be terrible to do `(See @http.authedRequest)` or whatever?

src/client.js
lukebarnard1: `rrEvent` and `callback` are both optional.

src/client.js
lukebarnard1: Oh. OK

lukebarnard1:
>can you move as much of this as possible into a low-level func in base-apis.js

I'm guessing you mean 
```js
const room = this.getRoom(roomId);
if (room) {
    room._addLocalEchoReceipt(this.credentials.userId, rrEvent, "m.read");
}
```
but `getRoom` is in `client.js` so it'd feel wrong to move it.

richvdh:
No, I mean the all the rest of it.

src/base-apis.js
```diff
@@ -470,6 +470,33 @@ MatrixBaseApis.prototype.roomInitialSync = function(roomId, limit, callback) {
...
+ * Set a marker to represent which event the user was last reading in a room. This can
+ * be retrieved from room account data and displayed as a horizontal line in the client
+ * that is visually distinct to the position of the user's own read receipt.
+ * @param {string} roomId ID of the room that has been read
+ * @param {string} rmEventId ID of the event that has been read
+ * @param {string} rrEventId ID of the event tracked by the read receipt. This is here
+ * for convenience because the RR and the RM are commonly updated at the same time as
+ * each other. Optional.
+ * @return {module:client.Promise} Resolves: (@see module:http-api.authedRequest)
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
```
richvdh: it never returns a MatrixError; it always returns a Promise.

Sorry, I should have been clearer about what to put here. There are a lot of historical functions (including authedRequest) which are basically full of shit. Note also that `httpApi.opts.onlyData` is true as far as base-apis is concerned, so authedRequest always just returns the result of the api call, so linking to that isn't particularly helpful.

Basically the best thing to do here is just have one `@return` line, along the lines of:

```
@return {module:client.Promise} resolves to the result object for the API call.
```

If you have the time and energy to document the shame of that object, so much the better, but generally we expect people to go and read the spec.

src/base-apis.js
```diff
@@ -470,6 +470,33 @@ MatrixBaseApis.prototype.roomInitialSync = function(roomId, limit, callback) {
     );
 };
 
+/**
+ * Set a marker to represent which event the user was last reading in a room. This can
```
richvdh: it's not the event the user was last reading though, is it?

src/client.js
```diff
@@ -1201,6 +1201,33 @@ MatrixClient.prototype.sendReadReceipt = function(event, callback) {
...
+ * that is visually distinct to the position of the user's own read receipt.
+ * @param {string} roomId ID of the room that has been read
+ * @param {string} eventId ID of the event that has been read
+ * @param {string} rrEvent the event tracked by the read receipt. This is here for
+ * convenience because the RR and the RM are commonly updated at the same time as each
+ * other. Optional.
+ * @return {module:client.Promise} Resolves: (@see module:http-api.authedRequest)
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixClient.prototype.setRoomReadMarker = function(roomId, eventId, rrEvent) {
```
richvdh: let's not distinguish these two by the presence or absence of 's'.

How about calling the other `setRoomReadMarkersHttpRequest` or something (we have a `_sendEventHttpRequest` for distinction from `_sendEvent`)

src/client.js
```diff
@@ -1201,6 +1201,33 @@ MatrixClient.prototype.sendReadReceipt = function(event, callback) {
     return this.sendReceipt(event, "m.read", callback);
 };
 
+/**
```
richvdh: jsdocs need tweaks as above

src/interactive-auth.js
```diff
@@ -318,6 +318,13 @@ InteractiveAuth.prototype = {
         }
         this._currentStage = nextStage;
 
+        if (nextStage == 'm.login.dummy') {
```
richvdh: ?

lukebarnard1:
Now that I've changed the base of the merge, those changes to `release.sh` and `interactive-auth.js` should've disappeared

PR #423, opened 2017-04-19
webrtc/call: Make it much less likely that callIds collide locally

superdump:
Previously if two calls were constructed within 1ms they could have the
same id.


Kegsay:
LGTM

PR #424, opened 2017-04-20
WebRTC: Support recvonly for video for those without a webcam

superdump:
When making video calls, if the caller or callee did not have a microphone, the call would fail. These changes make video callers/callees always offer to receive audio and video, regardless of what they are able to send. This allows one person to send audio and video and the other to send only audio, for example. Technically it is more flexible than that, in that the peer can send anything from nothing to both audio and video, but that is the primary purpose.

There is a commit that explicitly unmutes the remote audio element when setting it. This is so that one can re-use a remote video element also for remote audio, which is useful in some cases and probably a no-op in most others. It could unmute a muted remote audio element, but that can easily be handled externally after setting the remote audio element in my opinion.

### DIFF
--- a/src/webrtc/call.js
+++ b/src/webrtc/call.js
@@ -535,7 +535,12 @@ MatrixCall.prototype._maybeGotUserMediaForInvite = function(stream) {
     const self = this;
 
     const error = stream;
-    let constraints = null;
+    const constraints = {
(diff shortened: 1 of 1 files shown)


ara4n:
@dbkr can you take a look at this one please?

PR #425, opened 2017-04-25
Rework device list tracking logic

richvdh:
Yet another attempt at fixing https://github.com/vector-im/riot-web/issues/2305.

This now implements the algorithm described at http://matrix.org/speculator/spec/HEAD/client_server/unstable.html#tracking-the-device-list-for-a-user:

* We now keep a flag to tell us which users' device lists we are tracking. That makes it much easier to figure out whether we should care about device-update notifications from /sync (thereby fixing https://github.com/vector-im/riot-web/issues/3588).

* We use the same flag to indicate whether the device list for a particular user is out of date. Previously we did this implicitly by only updating the stored sync token when the list had been updated, but that was somewhat complicated, and in any case didn't help in cases where we initiated the key download due to a user joining an encrypted room.

Also fixes https://github.com/vector-im/riot-web/issues/3310.

### DIFF
--- a/spec/TestClient.js
+++ b/spec/TestClient.js
@@ -60,6 +60,7 @@ TestClient.prototype.toString = function() {
  * @return {Promise}
  */
 TestClient.prototype.start = function() {
+    console.log(this + ': starting');
     this.httpBackend.when("GET", "/pushrules").respond(200, {});
     this.httpBackend.when("POST", "/filter").respond(200, { filter_id: "fid" });
     this.expectDeviceKeyUpload();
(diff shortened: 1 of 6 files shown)


richvdh:
(should also indirectly help with #3660, since it will mean we are less likely to fill up the localstorage for device lists we don't care about)

NegativeMjark:
There's a lot of console.log being added. I assume it's there for debugging unknown session/key errors in the wild.

richvdh:
> There's a lot of console.log being added. I assume it's there for debugging unknown session/key errors in the wild.

well, I'm not sure it's a lot - most of them are just being moved around. But yeah, it's basically to try to debug uisis.

richvdh:
Now amended by https://github.com/matrix-org/matrix-js-sdk/pull/431.

PR #427, opened 2017-04-28
ability to specify webrtc audio/video inputs for the lib to request

t3chguy:


ara4n:
it feels a bit dodgy that audioInput and videoInput are just top-level `let`s rather than on an object.  But I guess we just don't have an appropriate object currently, so happy to leave it until we clean it up and introduce one.

PR #432, opened 2017-05-10
Implement API for username availability

lukebarnard1:
Requires synapse with https://github.com/matrix-org/synapse/pull/2183 and https://github.com/matrix-org/synapse/pull/2209 

**Update**: this needs to be changed to reflect https://github.com/matrix-org/synapse/pull/2213

### DIFF
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -121,6 +121,20 @@ MatrixBaseApis.prototype.makeTxnId = function() {
 // =============================
 
 /**
+ * Check whether a username is available prior to registration. An error response
+ * indicates an invalid/unavailable username.
+ * @param {string} username The username to check the availability of.
+ * @return {module:client.Promise} Resolves: to `true`.
+ */
+MatrixBaseApis.prototype.isUsernameAvailable = function(username) {
+    return this._http.authedRequest(
+        undefined, "POST", '/register/available', null, { username: username },
+    ).then((response) => {
+        return response.available;
+    });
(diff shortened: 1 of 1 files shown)


krombel:
To be correct it requires matrix-org/synapse#2209

lukebarnard1:
thanks, @krombel

richvdh:
looks ok, but why is this a POST rather than a GET?

lukebarnard1:
@richvdh I made the mistake of thinking a `GET` can include a body, which inherently it cannot. Hence https://github.com/matrix-org/synapse/pull/2209

lukebarnard1:
>yes, but why put the user id in the body rather than a query param?

Good question. When I originally proposed this, the response was largely "small things like this don't need a proposal". Things like this would probably have been looked at if a proposal was made.

lukebarnard1:
I shall modify this PR to reflect the changes to the synapse API - it shall be a `GET` with a query parameter instead.

krombel:
What is in case we want to trigger this function where this endpoint is not implemented?
Shouldn't the function raise such an error or do we assume the developer just triggers this function when he knows that this is available (then how should he determine)

lukebarnard1:
@krombel if the endpoint is not available on the server, a `MatrixError` will be raised with the `errcode` set to `M_UNRECOGNIZED`.

krombel:
OK. Good to know. Thanks

PR #434, opened 2017-05-14
make screen sharing call FF friendly :D

t3chguy:
FF is uber nice that it lets us select the display
does not seem to allow the composite ALL displays though like chromium does by default


t3chguy:
screenshot 30

ara4n:
lgtm - thanks :)

PR #439, opened 2017-05-20
Add in a public api getStoredDevice allowing clients to get a specific device

kscz:
Expose the `crypto.getStoredDevice(userId, deviceId)` api publically

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -360,6 +360,21 @@ MatrixClient.prototype.getStoredDevicesForUser = function(userId) {
 };
 
 /**
+ * Get the stored device key for a user id and device id
+ *
+ * @param {string} userId the user to list keys for.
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -360,6 +360,21 @@ MatrixClient.prototype.getStoredDevicesForUser = function(userId) {
 };
 
 /**
+ * Get the stored device key for a user id and device id
+ *
+ * @param {string} userId the user to list keys for.
+ * @param {string} deviceId unique identifier for the device
+ *
+ * @return {module:crypto-deviceinfo} device or nothing
```
t3chguy: since its nullable it should be
`@return {?module:crypto-deviceinfo} device or null`

kscz: Whoops, okay. Will update comment!

PR #445, opened 2017-05-31
Initial framework for indexeddb-backed crypto store

richvdh:
Doesn't do anything useful yet - just demonstrates a framework for how I hope it will fit into the sdk.

The idea here is that you can configure a factory to build a crypto store whenever you call `createClient` (if you don't pass one in explicitly). The default is an in-memory store, but `browser-index.js` sets the factory to the indexeddb impl.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -156,8 +156,9 @@ function MatrixClient(opts) {
     this._notifTimelineSet = null;
 
     this._crypto = null;
+    this._cryptoStore = opts.cryptoStore;
     if (CRYPTO_ENABLED && Boolean(opts.sessionStore) &&
-            Boolean(opts.cryptoStore) &&
+            Boolean(this._cryptoStore) &&
             userId !== null && this.deviceId !== null) {
         this._crypto = new Crypto(
             this, this,
@@ -172,6 +173,25 @@ function MatrixClient(opts) {
 }
 utils.inherits(MatrixClient, EventEmitter);
 utils.extend(MatrixClient.prototype, MatrixBaseApis.prototype);
+
+/**
+ * Clear any data out of the persistent stores used by the client.
+ *
+ * @returns {Promise} Promise which resolves when the stores have been cleared.
+ */
+MatrixClient.prototype.clearStores = function() {
+    if (this._clientRunning) {
+        throw new Error("Cannot clear stores while client is running");
+    }
+
+    const promises = [];
+
+    promises.push(this.store.deleteAllData());
+    if (this._cryptoStore) {
+        promises.push(this._cryptoStore.deleteAllData());
+    }
+    return q.all(promises);
+};
 
 /**
  * Get the user-id of the logged-in user
--- a/src/client.js
+++ b/src/client.js
@@ -104,6 +104,9 @@ try {
  * disabled by default for compatibility with older clients - in particular to
  * maintain support for back-paginating the live timeline after a '/sync'
  * result with a gap.
+ *
+ * @param {module:crypto.store.base~CryptoStore} opts.cryptoStore
+ *    crypto store implementation.
  */
 function MatrixClient(opts) {
     MatrixBaseApis.call(this, opts);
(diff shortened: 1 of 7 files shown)


src/client.js
```diff
@@ -160,6 +160,7 @@ function MatrixClient(opts) {
             opts.sessionStore,
             userId, this.deviceId,
             this.store,
+            opts.cryptoStore,
```
Kegsay: Missing JSDoc in `function MatrixClient(opts)`

src/matrix.js
```diff
@@ -63,6 +63,11 @@ module.exports.TimelineWindow = require("./timeline-window").TimelineWindow;
 module.exports.InteractiveAuth = require("./interactive-auth");
 
 
+module.exports.MemoryCryptoStore =
+    require("./crypto/store/memory-crypto-store").default;
```
Kegsay: What does `.default` do? I thought the default export is the default if you just `require()` and use it?

src/matrix.js
```diff
@@ -136,6 +154,11 @@ module.exports.wrapRequest = function(wrapper) {
  * {@link module:scheduler~MatrixScheduler}.
  * @param {requestFunction} opts.request If not set, defaults to the function
  * supplied to {@link request} which defaults to the request module from NPM.
+ *
+ * @param {module:crypto.store.base~CryptoStore=} opts.cryptoStore
+ *    crypto store implementation. Calls the factory supplied to
+ *    {@link setCryptoStoreFactory if unspecified}
```
Kegsay: End brace mispositioned.

src/matrix.js
```diff
@@ -136,6 +154,11 @@ module.exports.wrapRequest = function(wrapper) {
  * {@link module:scheduler~MatrixScheduler}.
  * @param {requestFunction} opts.request If not set, defaults to the function
  * supplied to {@link request} which defaults to the request module from NPM.
+ *
+ * @param {module:crypto.store.base~CryptoStore=} opts.cryptoStore
+ *    crypto store implementation. Calls the factory supplied to
```
Kegsay: Should specify what the default is if the caller never explicitly calls `setCryptoStoreFactory`. Does it error out? In-memory?

src/crypto/store/indexeddb-crypto-store.js
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * @module
+ */
+
+const VERSION = 1;
+
+/**
+ * @implements {module:crypto/store/base~CryptoStore}
+ */
+export default class IndexedDBCryptoStore {
```
Kegsay: This doesn't get displayed correctly when you `npm run gendoc`. The class is listed as `crypto/store/indexeddb-crypto-store` and it misses the constructor.

It's likely that this is happening because you're writing this in ES6 which JSDoc parses but does so incorrectly, for example members of class instances https://github.com/jsdoc3/jsdoc/issues/1301 - this is why we still use old-style ES5 in the JS SDK.

src/matrix.js
richvdh: apparently that was true in babel 5, but in babel 6 the default export is given the special `default` name in `module.exports`. ¯\_(ツ)_/¯

src/crypto/store/indexeddb-crypto-store.js
richvdh: The class is shown as `crypto/store/indexeddb-crypto-store` because it's the default export of that module. From an external point of view, its name is not `IndexedDBCryptoStore` but rather `module:crypto/store/indexeddb-crypto-store`.

It's missing the constructor because I forgot to document it. Now fixed.

src/matrix.js
richvdh: fixed

src/matrix.js
richvdh: fixed

src/client.js
richvdh: fixed

src/matrix.js
Kegsay: ¯\\\_(ツ)\_/¯ ok then

Kegsay:
LGTM!

richvdh:
thanks. I'm going to put the store-clearing stuff in here too in a sec.

Kegsay:
LGTM

PR #446, opened 2017-06-01
Load Olm from the global rather than requiring it.

richvdh:
This means that we can avoid confusing everybody in the world about how to
webpack js-sdk apps.


richvdh:
> lgtm. Do we need to add this to breaking changes?

I guess so, ideally. Do wwe have a place to put such things?

PR #447, opened 2017-06-01
indexeddb-crypto-store: fix db deletion

richvdh:
Add an `onversionchange` listener to close the db, so that we can delete it
without blocking.


Kegsay:
Ah yes, the wonderfully named `onversionchange` callback to let you know when someone wants to delete the database. LGTM.

PR #448, opened 2017-06-01
Send m.room_key_request events when we fail to decrypt an event

richvdh:


src/crypto/OutgoingRoomKeyRequestManager.js
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * Called when the client is stopped. Stops any running background processes.
+     */
+    stop() {
+        // stop the timer on the next run
+        this._clientRunning = false;
+    }
+
+    /**
+     * Send off a room key request, if we don't already have one
```
Kegsay: > if we don't already have one

Doesn't make sense to me. Do you mean: "if we don't already have one *queued*."? How do we determine if one request is the same as another request?

src/crypto/OutgoingRoomKeyRequestManager.js
```diff
@@ -0,0 +1,208 @@
...
+     *    pending list
+     */
+    makeRoomKeyRequest(requestBody, recipients) {
+        return this._cryptoStore.getOrAddOutgoingRoomKeyRequest({
+            requestBody: requestBody,
+            recipients: recipients,
+            requestId: this._baseApis.makeTxnId(),
+            state: ROOM_KEY_REQUEST_STATES.UNSENT,
+        }).then((req) => {
+            if (!req.state) {
```
Kegsay: It's not obvious what this is meant to mean. Why do we only start a timer if there is no `state` prop?

src/crypto/OutgoingRoomKeyRequestManager.js
```diff
@@ -0,0 +1,208 @@
...
+        if (!this._clientRunning) {
+            this._sendOutgoingRoomKeyRequestsTimer = null;
+            return q();
+        }
+
+        console.log("Looking for queued outgoing room key requests");
+
+        return this._cryptoStore.getOutgoingRoomKeyRequestByState([
+            ROOM_KEY_REQUEST_STATES.UNSENT,
+        ]).then((req) => {
```
Kegsay: It seems like a weird design decision to have this return a single request rather than an array of requests. I'm guessing that there is some batching going on or something? Why return only 1 request if N match? Is there significance to which request is selected (FIFO/LIFO/arbitrary?).

src/crypto/OutgoingRoomKeyRequestManager.js
```diff
@@ -0,0 +1,208 @@
...
+    _startTimer() {
+        if (this._sendOutgoingRoomKeyRequestsTimer) {
+            return;
+        }
+
+        const startSendingOutgoingRoomKeyRequests = () => {
+            if (this._sendOutgoingRoomKeyRequestsRunning) {
+                throw new Error("RoomKeyRequestSend already in progress!");
+            }
+            this._sendOutgoingRoomKeyRequestsRunning = true;
```
Kegsay: Is this state variable strictly required? Can the presence of `_sendOutgoingRoomKeyRequestsTimer` be enough of an indicator? It's preferable to reduce the number of states your state machine can be in, so if we can get by without this and still have it be clear I would prefer it.

src/crypto/OutgoingRoomKeyRequestManager.js
```diff
@@ -0,0 +1,208 @@
...
+            contentMap[recip.userId][recip.deviceId] = message;
+        }
+
+        return this._baseApis.sendToDevice(
+            'm.room_key_request', contentMap, txnId,
+        );
+    }
+}
+
+function stringifyRequestBody(requestBody) {
```
Kegsay: If `requestBody` was an actual type you could leverage the [built-in](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/toString) `toString` method so you don't need to explicitly stringify here. Might be nice.

src/crypto/index.js
```diff
@@ -788,6 +796,23 @@ Crypto.prototype.userDeviceListChanged = function(userId) {
...
+ * Send a request for some room keys, if we have not already done so
+ *
+ * @param {module:crypto~RoomKeyRequestBody} requestBody
+ * @param {Array<{userId: string, deviceId: string}>} recipients
+ */
+Crypto.prototype.requestRoomKey = function(requestBody, recipients) {
+    this._outgoingRoomKeyRequestManager.makeRoomKeyRequest(
+        requestBody, recipients,
+    ).catch((e) => {
+        // this normally means we couldn't talk to the store */
```
Kegsay: Stray `*/`

src/crypto/OutgoingRoomKeyRequestManager.js
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * Send off a room key request, if we don't already have one
+     *
+     * @param {module:crypto~RoomKeyRequestBody} requestBody
+     * @param {Array<{userId: string, deviceId: string}>} recipients
+     *
+     * @returns {Promise} resolves when the request has been added to the
+     *    pending list
+     */
+    makeRoomKeyRequest(requestBody, recipients) {
```
Kegsay: Naming: I dislike the name `makeRoomKeyRequest` because it sounds to me like it is constructing a request rather than sending a request. The docs are clear but at call-sites I find it confusing, see [here](https://github.com/matrix-org/matrix-js-sdk/pull/448/files#diff-fa4472db829c5a59596c29a2f43e8f9cR805) for an example. Can we `s/make/send/` please?

src/crypto/OutgoingRoomKeyRequestManager.js
richvdh: Updated the comment.

src/crypto/OutgoingRoomKeyRequestManager.js
richvdh: I'm actually relying on the fact that ROOM_KEY_REQUEST_STATES.UNSENT is zero. I agree it's horribly unclear. Updated.

src/crypto/OutgoingRoomKeyRequestManager.js
richvdh: > It seems like a weird design decision to have this return a single request rather than an array of requests. I'm guessing that there is some batching going on or something? Why return only 1 request if N match?

well, I have to loop anyway, even if I pull N out of the database at once, because more might turn up while I'm sending those N. So I might as well just do one at a time and save some complexity.

> Is there significance to which request is selected (FIFO/LIFO/arbitrary?).

well, it's done by order of pkey - ie, request_id - so is more-or-less FIFO. But basically, I'm assuming that arbitrary order is good enough.

src/crypto/OutgoingRoomKeyRequestManager.js
richvdh: It is not required - it is strictly a sanity check, as per https://github.com/matrix-org/matrix-js-sdk/pull/448/files#diff-d8c620291b1726321a300d1e99bb1256R53.

I see your point that it is adding complexity. I still think it is a net improvement, but will remove it if you feel strongly.

src/crypto/OutgoingRoomKeyRequestManager.js
richvdh: yeah, but sometimes requestBodies come out of the db (as sub-objects of requests), so making it into an actual class is non-trivial.

src/crypto/index.js
richvdh: fixed, ta

src/crypto/OutgoingRoomKeyRequestManager.js
richvdh: agreed. done.

src/crypto/OutgoingRoomKeyRequestManager.js
Kegsay: I don't feel strongly.

src/crypto/OutgoingRoomKeyRequestManager.js
Kegsay: OK.

src/crypto/algorithms/megolm.js
```diff
@@ -527,6 +527,10 @@ utils.inherits(MegolmDecryption, base.DecryptionAlgorithm);
  *   problem decrypting the event
  */
 MegolmDecryption.prototype.decryptEvent = function(event) {
+    this._decryptEvent(event, true);
+};
+
+MegolmDecryption.prototype._decryptEvent = function(event, requestKeysOnFail) {
```
Kegsay: I understand this is an internal method, but I'd like there to be a bit of prose by what you mean by "fail" on `requestKeysOnFail`. I see this exclusively means if the error message is `OLM.UNKNOWN_MESSAGE_INDEX` but that's as clear as mud.

src/crypto/algorithms/megolm.js
```diff
@@ -543,6 +547,9 @@ MegolmDecryption.prototype.decryptEvent = function(event) {
     } catch (e) {
         if (e.message === 'OLM.UNKNOWN_MESSAGE_INDEX') {
             this._addEventToPendingList(event);
+            if (requestKeysOnFail) {
+                this._requestKeysForEvent(event);
+            }
```
Kegsay: We don't recover at this point, we just throw as if `requestKeysOnFail == false` so from the caller's perspective they have no idea whether they will ever be able to decrypt this message. This might be fine? I don't know how people are using `MegolmDecryption.prototype.decryptEvent` currently.

src/crypto/algorithms/megolm.js
```diff
@@ -657,7 +689,7 @@ MegolmDecryption.prototype._retryDecryption = function(senderKey, sessionId) {
 
     for (let i = 0; i < pending.length; i++) {
         try {
-            this.decryptEvent(pending[i]);
+            this.decryptEvent(pending[i], false);
```
Kegsay: Wrong method call. Should be `this._decryptEvent`. Also, it would be good to explain why you're passing `false` here, presumably because we're retrying once we've got the keys, so getting the keys again isn't going to make a difference.

src/crypto/store/base.js
```diff
@@ -9,3 +9,23 @@
...
+
+/**
+ * Represents an outgoing room key request
+ *
+ * @typedef {Object} OutgoingRoomKeyRequest
+ *
+ * @property {string} requestId    unique id for this request
+ *
+ * @property {string?} cancellationRequestId
+ *    transaction id for the cancellation, if any
```
Kegsay: Terminology: The variable states its a request ID, but the comment says it's a transaction ID. Is there a difference? If not, can we stick to one term throughout please.

src/crypto/store/base.js
```diff
@@ -9,3 +9,23 @@
  *
  * @interface CryptoStore
  */
+
+/**
+ * Represents an outgoing room key request
+ *
+ * @typedef {Object} OutgoingRoomKeyRequest
+ *
+ * @property {string} requestId    unique id for this request
```
Kegsay: The usual questions I ask whenever I see people add in IDs:
 - Is there a min length? Max length?
 - Can I use smiley poo? Spaces?

richvdh: it's an internal method. I don't think this needs documenting.

src/crypto/index.js
```diff
@@ -1126,5 +1151,13 @@ Crypto.prototype._signObject = function(obj) {
 };
 
 
+/**
+ * The parameters of a room key request. The details of the request may
+ * vary with the crypto algorithm, but the management and storage layers for
+ * outgoing requests expect it to have 'room_id' and 'session_id' properties.
```
Kegsay: > the management and storage layers for outgoing requests expect it to have 'room_id' and 'session_id' properties.

Is this just a long way of saying:
```
@typedef {Object} RoomKeyRequestBody
@prop {string} room_id
@prop {string} session_id

Other keys are determined by the precise crypto algorithm used.
```
right? There may be value in `@prop`ing up the mandatory fields.

src/crypto/store/indexeddb-crypto-store.js
```diff
@@ -110,6 +111,238 @@ export default class IndexedDBCryptoStore {
...
+                onFound: (existing) => {
+                    // this entry matches the request - return it.
+                    console.log(
+                        `already have key request outstanding for ` +
+                        `${requestBody.room_id} / ${requestBody.session_id}: ` +
+                        `not sending another`,
+                    );
+                    deferred.resolve(existing);
+                },
+                onNotFound: () => {
```
Kegsay: API shape feels weird. Is there any reason why we don't invoke a single function and give a `null` parameter if it was not found?

src/crypto/store/indexeddb-crypto-store.js
```diff
@@ -110,6 +111,238 @@ export default class IndexedDBCryptoStore {
...
+    }
+
+    /**
+     * Look for room key requests by state
+     *
+     * @param {Array<Number>} wantedStates list of acceptable states
+     *
+     * @return {Promise} resolves to the a
+     *    {@link module:crypto/store/base~OutgoingRoomKeyRequest}, or null if
+     *    there are no pending requests in those states
```
Kegsay: As per your comment on https://github.com/matrix-org/matrix-js-sdk/pull/448/files#r119830241 - I would mention here that if there are multiple matching requests, an arbitrary request is chosen.

src/crypto/store/indexeddb-crypto-store.js
```diff
@@ -110,6 +111,238 @@ export default class IndexedDBCryptoStore {
...
+     */
+    getOutgoingRoomKeyRequestByState(wantedStates) {
+        if (wantedStates.length === 0) {
+            return null;
```
Kegsay: The JSDoc clearly states this function will always return a Promise. Please wrap this up.

src/crypto/store/base.js
Kegsay: This goes over the wire as `request_id` - so either link to the docs for the wire protocol or state the protocol here please. There's no other way for me to know what is valid and what is not otherwise.

src/crypto/algorithms/megolm.js
richvdh: done

src/crypto/algorithms/megolm.js
richvdh: well, it's fine for now. Until the key turns up it is just an undecryptable event.

decryptEvent is called [here](https://github.com/matrix-org/matrix-js-sdk/blob/00a46273bc0add9a7088ca8a0a0c12deec01b345/src/crypto/index.js#L783) which in turn called [here](https://github.com/matrix-org/matrix-js-sdk/blob/00a46273bc0add9a7088ca8a0a0c12deec01b345/src/client.js#L601).

I was toying with setting a flag on the event to show that its key had been requested or something, but I didn't need it yet.

src/crypto/algorithms/megolm.js
richvdh: done, thanks.

src/crypto/store/base.js
richvdh: this one is certainly a transaction id. I can change that.

The other one (`requestId`) is both: we use it both as a request id within the message as a unique identifier for the recipient to correlate with the cancellation when it arrives, and as the transaction_id for sending the to-device message.

src/crypto/store/base.js
richvdh: https://docs.google.com/document/d/1m4gQkcnJkxNuBmb5NoFCIadIY-DyqqNAS3lloE73BlQ/edit#heading=h.2rxyh346wc9c

src/crypto/index.js
richvdh: Maybe, but I think it gives more emphasis to the presence of those fields than is entirely justified. The management layer only uses them for debug, and the storage layer just uses them as keys in the indexeddb for efficiency. Both should be perfectly happy with them being absent.

src/crypto/store/indexeddb-crypto-store.js
richvdh: good question. fixed.

src/crypto/store/indexeddb-crypto-store.js
richvdh: done

src/crypto/store/indexeddb-crypto-store.js
richvdh: done

richvdh:
@Kegsay ptal?

src/crypto/store/base.js
Kegsay: In absence of this being in the spec, please add a link to the gdoc so readers know where to go to find the wire protocol.

src/crypto/index.js
Kegsay: Okay then.

src/crypto/algorithms/megolm.js
Kegsay: OK.

src/crypto/store/base.js
Kegsay: It might be worth mentioning that on `:18` then when describing what `requestId` is then, since I didn't manage to grok that from looking at the code alone.

Kegsay:
I'd like for you to add a link to https://docs.google.com/document/d/1m4gQkcnJkxNuBmb5NoFCIadIY-DyqqNAS3lloE73BlQ/edit#heading=h.2rxyh346wc9c somewhere so people can see the wire protocol that this code is supposed to be implementing. I don't need to see the code afterwards. 

Other than that, LGTM!

src/crypto/store/base.js
richvdh: ok. Added.

src/crypto/store/base.js
richvdh: Added at the top of OutgoingKeyRequestManager

richvdh:
I think this is ready to land, but blocked on the un-reversion of https://github.com/matrix-org/matrix-react-sdk/pull/983, which is blocked on the landing of the guest access branch

richvdh:
I'm going to land this and #449 on a feature branch, until we have a fix in place for https://github.com/matrix-org/matrix-react-sdk/pull/983 and have the rest of this feature in place.

PR #449, opened 2017-06-01
Process received room key requests

richvdh:
Doesn't actually do any of the crypto magic yet.

### DIFF
--- a/src/crypto/algorithms/base.js
+++ b/src/crypto/algorithms/base.js
@@ -140,6 +140,26 @@ class DecryptionAlgorithm {
     importRoomKey(session) {
         // ignore by default
     }
+
+    /**
+     * Determine if we have the keys necessary to respond to a room key request
+     *
+     * @param {module:crypto#RoomKeyRequest} keyRequest
+     * @return {boolean} true if we have the keys and could (theoretically) share
+     *  them; else false.
+     */
+    hasKeysForKeyRequest(keyRequest) {
+        return false;
+    }
+
+    /**
+     * Send the response to a room key request
+     *
+     * @param {module:crypto#RoomKeyRequest} keyRequest
+     */
+    shareKeysWithDevice(keyRequest) {
+        throw new Error("shareKeysWithDevice not supported for this DecryptionAlgorithm");
+    }
 }
 export {DecryptionAlgorithm}; // https://github.com/jsdoc3/jsdoc/issues/1272
 
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -94,8 +94,10 @@ function Crypto(baseApis, eventEmitter, sessionStore, userId, deviceId,
 
     this._globalBlacklistUnverifiedDevices = false;
 
-    // list of IncomingRoomKeyRequests we received in the current sync.
+    // list of IncomingRoomKeyRequests/IncomingRoomKeyRequestCancellations
+    // we received in the current sync.
     this._receivedRoomKeyRequests = [];
+    this._receivedRoomKeyRequestCancellations = [];
 
     let myDevices = this._sessionStore.getEndToEndDevicesForUser(
         this._userId,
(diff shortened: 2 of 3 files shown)


src/crypto/algorithms/base.js
```diff
@@ -140,6 +140,26 @@ class DecryptionAlgorithm {
     importRoomKey(session) {
         // ignore by default
     }
+
+    /**
+     * Determine if we have the keys necessary to respond to a room key request
+     *
+     * @param {module:crypto#RoomKeyRequest} keyRequest
+     * @return {boolean} true if we have the keys and could (theoretically) share
```
Kegsay: You'll probably be querying indexeddb to get the answer right? Probably want this to be a `Promise` then?

src/crypto/index.js
```diff
@@ -93,6 +94,11 @@ function Crypto(baseApis, eventEmitter, sessionStore, userId, deviceId,
 
     this._globalBlacklistUnverifiedDevices = false;
 
+    // list of IncomingRoomKeyRequests/IncomingRoomKeyRequestCancellations
+    // we received in the current sync.
+    this._receivedRoomKeyRequests = [];
+    this._receivedRoomKeyRequestCancellations = [];
```
Kegsay: Storing them like this means we lose ordering information from `/sync`. Is this a problem?

src/crypto/index.js
```diff
@@ -1056,6 +1067,106 @@ Crypto.prototype._onNewDeviceEvent = function(event) {
...
+
+        if (!decryptor.hasKeysForKeyRequest(req)) {
+            console.log(
+                `room key request for unknown session ${roomId} / ` +
+                body.session_id,
+            );
+            continue;
+        }
+
+        req._shareCallback = () => {
```
Kegsay: I don't like that `Crypto` is clobbering an `_` variable on a different class (`IncomingRoomKeyRequest`) like this. Please either make a public function to do this, or don't pretend that this is private. Just because both classes are in the same 1300 line file is not a good enough reason to gut wrench.

src/crypto/index.js
Kegsay: E.g if I send a key request then immediately cancel it, currently we will process the key request due to how `_processReceivedRoomKeyRequests` is currently implemented.

Kegsay:
Otherwise LGTM.

src/crypto/algorithms/base.js
richvdh: well, not currently - the relevant info is in localstorage. Admittedly we'd probably like it to move to indexeddb, but that's a whole separate project

src/crypto/index.js
richvdh: well now.

AFAIK there is no guarantee in the protocol that the ordering of to-device messages is preserved anyway, so this whole thing is a steaming pile of fragility.

That said - in the case you mention, we will process the key request, and then process the cancellation. Which is considerably better than processing (and, presumably, ignoring) the cancellation and *then* the request.

src/crypto/index.js
richvdh: fair. Hopefully fixed in https://github.com/matrix-org/matrix-js-sdk/pull/449/commits/1664312c80f6e729c030ae9e959ff8b0e21fa8c6.

src/crypto/algorithms/base.js
Kegsay: If it's in local storage, won't we hit size limits again? It's pretty easy to convert this to be `q(bool)` right? I would prefer if we did an async design up-front, as it's often a pain to turn a synchronous API into an async one retrospectively.

src/crypto/index.js
Kegsay: > we will process the key request, and then process the cancellation.

Doesn't that race? We should at least mention the ordering concerns and decisions / things we've punted on for now somewhere.

src/crypto/index.js
Kegsay: WFM!

src/crypto/index.js
richvdh: yes, of course it races:

> AFAIK there is no guarantee in the protocol that the ordering of to-device messages is preserved anyway, so this whole thing is a steaming pile of fragility.

the order in which requests and cancellations order are, tbh, the least of my concerns here. I want to ship it and iterate it rather than think through the edge cases any more. This work has taken weeks already.

src/crypto/algorithms/base.js
richvdh: the relevant info is *already* in localstorage.

Kegsay: OK.

src/crypto/index.js
Kegsay: OK.

Kegsay:
LGTM

PR #450, opened 2017-06-04
Update README: Clarify how to install libolm

arxcode:
Olm is not available via the npm registry, so ``npm install olm`` doesn't work. What's worse, it has been available in the past but is now unpublished, which combined with an issue in npm leads to a confusing error message when trying to do the install.

The approach used by Riot Web (which I assume is the "right" approach) is to use the tarball notation of npm to install the latest olm tarball and (if needed) add it to the project dependencies while doing so.

This PR hopes to clarify the README in this regard.

(Also, just using ``require('olm');`` didn't work for me, I needed to explicitly assign it to the ``global.Olm`` object via ``global.Olm = require('olm');``. I have clarified this as well.)

Signed-off-by: Alexander Rudyk (ar-git@mailbox.org)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,8 @@ In a browser
 Download either the full or minified version from
 https://github.com/matrix-org/matrix-js-sdk/releases/latest and add that as a
 ``<script>`` to your page. There will be a global variable ``matrixcs``
-attached to ``window`` through which you can access the SDK.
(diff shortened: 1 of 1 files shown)


ara4n:
@arxcode thanks for this - sorry you were bitten by the somewhat not-yet-mature shift here.

PR #454, opened 2017-06-06
Implement sharing of megolm keys

richvdh:
This builds on previously-landed code (specifically #449) to implement sharing of megolm keys.

When we get a request from one of our other devices we check (via the existing hook https://github.com/matrix-org/matrix-js-sdk/blob/665ed057e364ec54f20513c489749cc1e985fbac/src/crypto/index.js#L1158) whether we have the keys for that session. If the device is verified, or the user accepts the request, we then send an m.forwarded_room_key event with the keys to the session.

### DIFF
--- a/src/crypto/algorithms/megolm.js
+++ b/src/crypto/algorithms/megolm.js
@@ -662,7 +662,9 @@ MegolmDecryption.prototype.onRoomKeyEvent = function(event) {
     this._retryDecryption(senderKey, sessionId);
 };
 
-
+/**
+ * @inheritdoc
+ */
 MegolmDecryption.prototype.hasKeysForKeyRequest = function(keyRequest) {
     const body = keyRequest.requestBody;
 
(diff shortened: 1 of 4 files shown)


src/crypto/algorithms/megolm.js
```diff
@@ -663,6 +663,89 @@ MegolmDecryption.prototype.onRoomKeyEvent = function(event) {
...
+    return this._olmDevice.hasInboundSessionKeys(
+        body.room_id,
+        body.sender_key,
+        body.session_id,
+        // TODO: ratchet index
+    );
+};
+
+
+MegolmDecryption.prototype.shareKeysWithDevice = function(keyRequest) {
```
NegativeMjark: Do these functions need documentation to explain what they are doing?

richvdh: They override the methods in the base class - see https://github.com/matrix-org/matrix-js-sdk/blob/2c54d760857116711d2780f9e787533aa6d616d2/src/crypto/algorithms/base.js#L156. I've updated the jsdoc stuff so that the jsdoc will be correctly inherited from the base class.

PR #456, opened 2017-06-06
Send a cancellation for room key requests

richvdh:
When we receive a room key, cancel any pending requests we have open for that
key.

### DIFF
--- a/src/crypto/OutgoingRoomKeyRequestManager.js
+++ b/src/crypto/OutgoingRoomKeyRequestManager.js
@@ -144,68 +144,73 @@ export default class OutgoingRoomKeyRequestManager {
                 // no request was made for this key
                 return;
             }
-
-            if (req.state === ROOM_KEY_REQUEST_STATES.CANCELLATION_PENDING) {
-                // nothing to do here
-                return;
-            }
-
-            if (req.state === ROOM_KEY_REQUEST_STATES.UNSENT) {
-                // just delete it
-
-                // FIXME: ghahah we may have attempted to send it, and
-                // not yet got a successful response. So the server
-                // may have seen it, so we still need to send a cancellation
-                // in that case :/
-
-                console.log(
-                    'deleting unnecessary room key request for ' +
-                    stringifyRequestBody(requestBody),
-                );
-                return this._cryptoStore.deleteOutgoingRoomKeyRequest(
-                    req.requestId, ROOM_KEY_REQUEST_STATES.UNSENT,
-                );
-            }
-
-            // by elimination, we are in ROOM_KEY_REQUEST_STATES.SENT
-
-            // send a cancellation.
-            return this._cryptoStore.updateOutgoingRoomKeyRequest(
-                req.requestId, ROOM_KEY_REQUEST_STATES.SENT, {
-                    state: ROOM_KEY_REQUEST_STATES.CANCELLATION_PENDING,
-                    cancellationTxnId: this._baseApis.makeTxnId(),
-                },
-            ).then((updatedReq) => {
-                if (!updatedReq) {
-                    // we raced with another tab to mark the request cancelled
-                    // - no point in sending another cancellation
-                    console.log(
-                        'Tried to cancel room key request for ' +
-                        stringifyRequestBody(requestBody) +
-                        ' but it was already cancelled in another tab',
-                    );
+            switch (req.state) {
+                case ROOM_KEY_REQUEST_STATES.CANCELLATION_PENDING:
+                    // nothing to do here
                     return;
-                }
 
-                // We don't want to wait for the timer, so we send it
-                // immediately. (We might actually end up racing with the timer,
-                // but that's ok: even if we make the request twice, we'll do it
-                // with the same transaction_id, so only one message will get
-                // sent).
-                //
-                // (We also don't want to wait for the response from the server
-                // here, as it will slow down processing of received keys if we
-                // do.)
-                this._sendOutgoingRoomKeyRequestCancellation(
-                    updatedReq,
-                ).catch((e) => {
-                    console.error(
-                        "Error sending room key request cancellation; will retry later.",
-                        e,
+                case ROOM_KEY_REQUEST_STATES.UNSENT:
+                    // just delete it
+
+                    // FIXME: ghahah we may have attempted to send it, and
+                    // not yet got a successful response. So the server
+                    // may have seen it, so we still need to send a cancellation
+                    // in that case :/
+
+                    console.log(
+                        'deleting unnecessary room key request for ' +
+                        stringifyRequestBody(requestBody),
                     );
-                    this._startTimer();
-                }).done();
-            });
+                    return this._cryptoStore.deleteOutgoingRoomKeyRequest(
+                        req.requestId, ROOM_KEY_REQUEST_STATES.UNSENT,
+                    );
+
+                case ROOM_KEY_REQUEST_STATES.SENT:
+                    // send a cancellation.
+                    return this._cryptoStore.updateOutgoingRoomKeyRequest(
+                        req.requestId, ROOM_KEY_REQUEST_STATES.SENT, {
+                            state: ROOM_KEY_REQUEST_STATES.CANCELLATION_PENDING,
+                            cancellationTxnId: this._baseApis.makeTxnId(),
+                        },
+                    ).then((updatedReq) => {
+                        if (!updatedReq) {
+                            // updateOutgoingRoomKeyRequest couldn't find the
+                            // request in state ROOM_KEY_REQUEST_STATES.SENT,
+                            // so we must have raced with another tab to mark
+                            // the request cancelled. There is no point in
+                            // sending another cancellation since the other tab
+                            // will do it.
+                            console.log(
+                                'Tried to cancel room key request for ' +
+                                stringifyRequestBody(requestBody) +
+                                ' but it was already cancelled in another tab',
+                            );
+                            return;
+                        }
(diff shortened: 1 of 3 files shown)


src/crypto/OutgoingRoomKeyRequestManager.js
```diff
@@ -41,6 +41,9 @@ const ROOM_KEY_REQUEST_STATES = {
 
     /** request sent, awaiting reply */
     SENT: 1,
+
+    /** reply received, cancellation not yet sent */
+    CANCELLATION_PENDING: 2,
```
Kegsay: > reply received, cancellation not yet sent

I don't understand what state this represents in the state machine. I understood when it was a simple sent/not sent. Is this accurate:
```
UNSENT -------> SENT
  |              ^
  |              |
  +----> CANCELLATION_PENDING
```
as in, you cannot go to this state if the request has already been sent? It might be nice to have a diagram akin to this somewhere as it helps me grok things. Does it matter that `SENT` can mean "sent cancellation" AND "sent request"?

src/crypto/OutgoingRoomKeyRequestManager.js
```diff
@@ -107,6 +110,79 @@ export default class OutgoingRoomKeyRequestManager {
         });
     }
 
+    /**
+     * Cancel room key requests, if any match the given details
```
Kegsay: Can you please define what "match" means? Deep equality checks of every field in the `requestBody`?

src/crypto/OutgoingRoomKeyRequestManager.js
```diff
@@ -107,6 +110,79 @@ export default class OutgoingRoomKeyRequestManager {
...
+            if (!req) {
+                // no request was made for this key
+                return;
+            }
+
+            if (req.state === ROOM_KEY_REQUEST_STATES.UNSENT) {
+                // just delete it
+                console.log(
+                    'deleting unnecessary room key request for ' +
+                    stringifyRequestBody(requestBody),
```
Kegsay: Trailing comma. This is sadly not valid syntax.
```
$ cat t
var t = {};
console.log(
  'fui wehfuihe' +
   JSON.stringify(t),
);

$ node t
/.../t:5
);
^
SyntaxError: Unexpected token )
    at Object.exports.runInThisContext (vm.js:76:16)
    at Module._compile (module.js:542:28)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:394:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:509:3
```

richvdh: babel transpiles this for us.

src/crypto/OutgoingRoomKeyRequestManager.js
```diff
@@ -107,6 +110,79 @@ export default class OutgoingRoomKeyRequestManager {
...
+                );
+                return this._cryptoStore.deleteOutgoingRoomKeyRequest(
+                    req.requestId, ROOM_KEY_REQUEST_STATES.UNSENT,
+                );
+            }
+
+            if (req.state !== ROOM_KEY_REQUEST_STATES.SENT) {
+                // the only other option is cancellation_pending, in which case
+                // there is nothing else to do here
+            }
```
Kegsay: I feel you may be missing something here? Surely you mean:
```js
if (req.state !== ROOM_KEY_REQUEST_STATES.SENT) {
    return;
}
```

As an aside: I dislike and would discourage the use of "handle 2 cases, only 1 left so don't bother to check" since this will break in weird ways if we end up adding more states, plus it's clearer meaning you needn't document what you mean for the last case:

```js
if (req.state === ROOM_KEY_REQUEST_STATES.CANCELLATION_PENDING) {
    return;
}
```
in a function called `cancelRoomKeyRequest` is nice and clear that we've already sent the cancellation and therefore don't need to resend.

src/crypto/OutgoingRoomKeyRequestManager.js
```diff
@@ -107,6 +110,79 @@ export default class OutgoingRoomKeyRequestManager {
...
+            // send a cancellation.
+            return this._cryptoStore.updateOutgoingRoomKeyRequest(
+                req.requestId, ROOM_KEY_REQUEST_STATES.SENT, {
+                    state: ROOM_KEY_REQUEST_STATES.CANCELLATION_PENDING,
+                    cancellationTxnId: this._baseApis.makeTxnId(),
+                },
+            ).then((updatedReq) => {
+                if (!updatedReq) {
+                    // we raced with another tab to mark the request cancelled
+                    // - no point in sending another cancellation
```
Kegsay: I don't understand how the absence of `updateReq` means we raced with another tab. Please explain this more.

src/crypto/OutgoingRoomKeyRequestManager.js
Kegsay: \o/

src/crypto/OutgoingRoomKeyRequestManager.js
richvdh: The state machine is:

```
-> UNSENT -----> SENT -----> CANCELLATION_PENDING ---->  (deleted)
       |                                                                                                     ^
      +------------------------------------------------------------------------------------+
```

I'll do some ascii art in the file.


> Does it matter that SENT can mean "sent cancellation" AND "sent request"?

it doesn't mean that.

src/crypto/OutgoingRoomKeyRequestManager.js
richvdh: yes, just the same as sendRoomKeyRequest

src/crypto/OutgoingRoomKeyRequestManager.js
richvdh: > I feel you may be missing something here?

apparently! thank you!

> As an aside: I dislike and would discourage the use of "handle 2 cases, only 1 left so don't bother to check" since this will break in weird ways if we end up adding more states

I don't really understand what you want me to do here. If I explicitly handle each case, it's *still* going to break in weird ways if we end up adding more states. I've rearranged the code a bit. I think it's clear, but you may disagree.

I've also added a FIXME because I've spotted another edge case :-( :-(

src/crypto/OutgoingRoomKeyRequestManager.js
richvdh: `updatedReq` being false means that the store didn't find the matching request in `ROOM_KEY_REQUEST_STATES.SENT`. We know it was in that state when we did the lookup at line 140, and the only way out of `SENT` is when we send the cancellation... so we must have raced.

src/crypto/OutgoingRoomKeyRequestManager.js
Kegsay: >  If I explicitly handle each case, it's still going to break in weird ways if we end up adding more states.

The point is to try to identify new states and fail as quickly as possible rather than executing whichever state was last to be handled. I was hoping for something like:
```js
switch(foo) {
  case 1:
    break;
  case 2:
    break;
  case 3:
    break;
  default:
    throw new Error("unhandled state: " + foo);
}
```

which makes is abundantly clear in logs that we forgot to implement stuff, rather than absolute silence and it sending a cancellation/sending the request/deleting the request/other things you really don't want.

src/crypto/OutgoingRoomKeyRequestManager.js
```diff
@@ -33,6 +33,24 @@ const SEND_KEY_REQUESTS_DELAY_MS = 500;
...
+ *     V                                  |
+ *    SENT                                |
+ *     |                                  |
+ *     | (cancellation requested)         |
+ *     V                                  |
+ * CANCELLATION_PENDING                   |
+ *     |                                  |
+ *     | (cancellation sent)              |
+ *     V                                  |
+ * (deleted)  <---------------------------+
```
Kegsay: This is awesome, thank you! :star:

src/crypto/OutgoingRoomKeyRequestManager.js
Kegsay: Aha, I see now. The bit I failed to understand before was:

> updatedReq being false means that the store didn't find the matching request in ROOM_KEY_REQUEST_STATES.SENT

coupled with:

> We know it was in that state

I guess my problem was that I failed to understand that:
```js
+            return this._cryptoStore.updateOutgoingRoomKeyRequest(
 +                req.requestId, ROOM_KEY_REQUEST_STATES.SENT, {
 +                    state: ROOM_KEY_REQUEST_STATES.CANCELLATION_PENDING,
 +                    cancellationTxnId: this._baseApis.makeTxnId(),
 +                },
 +            )
```
means "transition to SENT -> CANCELLATION_PENDING and return a bool if you did so". Don't know how best to improve clarity here though.

Kegsay:
Other than https://github.com/matrix-org/matrix-js-sdk/pull/456#discussion_r120594736 - LGTM

src/crypto/OutgoingRoomKeyRequestManager.js
richvdh: yup, fair enough.

src/crypto/OutgoingRoomKeyRequestManager.js
richvdh: I've tried to improve the comment.

PR #457, opened 2017-06-07
Add API for POST /user_directory/search

lukebarnard1:
This takes a JSON body of the form:
```json
{
    "term": "search term",
    "limit": 42,
}
```
where "term" is the term to match against user IDs, display names and domains and "limit" is the maximum number of results to return.

The response body looks like
```json
{
    "results ": [
        { "user_id": "@someid:mydomain.com", "display_name": "Some Name", "avatar_url": "mx://..." },
        ...
    ],
    "limited": false
}
```
where "limited" indicates whether the "limit" was used to truncate the list.

### DIFF
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -1,4 +1,4 @@
-    /*
+/*
 Copyright 2015, 2016 OpenMarket Ltd
 Copyright 2017 Vector Creations Ltd
 
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -691,8 +691,11 @@ MatrixBaseApis.prototype.setRoomDirectoryVisibilityAppService =
 MatrixBaseApis.prototype.searchUserDirectory = function(opts) {
     const body = {
         search_term: opts.term,
-        limit: opts.limit,
     };
+
+    if (opts.limit !== undefined) {
+        body.limit = opts.limit;
+    }
 
     return this._http.authedRequest(
         undefined, "POST", "/user_directory/search", undefined, body,
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -685,13 +685,13 @@ MatrixBaseApis.prototype.setRoomDirectoryVisibilityAppService =
  * Query the user directory with a term matching user IDs, display names and domains.
  * @param {object} opts options
  * @param {string} opts.term the term with which to search.
- * @param {number=} opts.limit the maximum number of results to return. Defaults to 20.
+ * @param {number=} opts.limit the maximum number of results to return. Defaults to 10.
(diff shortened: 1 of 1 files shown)


src/base-apis.js
```diff
@@ -678,6 +678,27 @@ MatrixBaseApis.prototype.setRoomDirectoryVisibilityAppService =
...
+ * Query the user directory with a term matching user IDs, display names and domains.
+ * @param {object} opts options
+ * @param {string} opts.term the term with which to search.
+ * @param {number=} opts.limit the maximum number of results to return. Defaults to 20.
+ * @return {module:client.Promise} Resolves: an array of results.
+ */
+MatrixBaseApis.prototype.searchUserDirectory = function(opts) {
+    const body = {
+        search_term: opts.term,
+        limit: opts.limit || 20,
```
richvdh: It seems funny to impose a default here. Either a limit is a thing we want in general, in which case the default should be applied on the server - or it's not, in which case we shouldn't impose a default here.

in other words: let's make this reflect the server's behaviour. If the app passes a limit, pass it in the request, else leave it up to the server.

src/base-apis.js
```diff
@@ -1,4 +1,4 @@
-/*
+    /*
```
richvdh: mistake?

src/base-apis.js
lukebarnard1: That's true, let's just pass it thru

src/base-apis.js
lukebarnard1: yup!

src/base-apis.js
```diff
@@ -678,6 +678,27 @@ MatrixBaseApis.prototype.setRoomDirectoryVisibilityAppService =
...
 };
 
+// User Directory Operations
+// =========================
+
+/**
+ * Query the user directory with a term matching user IDs, display names and domains.
+ * @param {object} opts options
+ * @param {string} opts.term the term with which to search.
+ * @param {number=} opts.limit the maximum number of results to return. Defaults to 10.
```
richvdh: think it would be clearer to say something like "the server will apply a limit if unspecified". I don't think you need to make it explicit what the limit is.

src/base-apis.js
```diff
@@ -678,6 +678,27 @@ MatrixBaseApis.prototype.setRoomDirectoryVisibilityAppService =
...
+ * @return {module:client.Promise} Resolves: an array of results.
+ */
+MatrixBaseApis.prototype.searchUserDirectory = function(opts) {
+    const body = {
+        search_term: opts.term,
+        limit: opts.limit,
```
richvdh: think you only want to add this field if it's not `undefined`.

PR #459, opened 2017-06-09
Inspect the state of the timeline to possibly resolve with less event loops in theory

lukebarnard1:
If we already have a timeline for an event, don't do an async call, just use the timeline found directly.

This is the alternative to https://github.com/matrix-org/matrix-js-sdk/pull/458 which will immediately check if a promise is resolved.

Instead, this patch will check if there is a timeline for the event synchronously and then use that to instantiate the timeline window.

### DIFF
--- a/spec/unit/timeline-window.spec.js
+++ b/spec/unit/timeline-window.spec.js
@@ -12,6 +12,22 @@ import expect from 'expect';
 const ROOM_ID = "roomId";
 const USER_ID = "userId";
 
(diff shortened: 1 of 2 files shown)


lukebarnard1:
This is a more stable/nicer fix but the result is the same. Let's consider merging once the release is out the door.

richvdh:
this doesn't seem to be happening, and has now bitrotted. let's close it for now at least.

PR #461, opened 2017-06-12
[POC] Implementation of WebSocketAPI

krombel:
All proposed functionality is implemented and working.
Also the fallback if the WebSocket endpoint is not available on server-side is implemented.

A hosted version can be found at https://msgs.tk/riotws/

This currently requires [my matrix-websocket-proxy](https://github.com/krombel/matrix-websockets-proxy) to be run.

Currently the implemented endpoint is `/_matrix/client/unstable/stream` (instead of `/_matrix/client/ws/r0` which got proposed)


dbkr:
I think the big remaining things on this PR are:
 1. What I believe @ara4n was discussing with you on #matrix-dev:matrix.org, in that we would want t some concrete evidence of the benefit of websockets, ie. a benchmark of some kind demonstrating how much more efficient it is in whatever way / ways.
 2. This duplicates a lot of fairly complex logic from the syncing code by essentially introducing a parallel implementation of the syncing logic with all the bootstrapping logic duplicated. I would expect the setup code to be shared, then deferring to either websockets or long poll code to do the sync stream.

krombel:
My current opinion is:
WebSocket is a **different** technology which can be used for transmission.
The speed is nearly the same
- Single messages currently need nearly the same amount of time to be traversed
- On flood there will be more small messages via websocket - `/sync` aggregates several messages to one bundle (as it took some time to handle the previous one and request the next `/sync`)
  So one message is earlier at the client but as handling each message needs some time it might be that the message will be shown a little bit later (while having higher CPU consumption)

I currently see only those benefits:
- The server has to handle less requests. When there is new content for a client it just pushes that and it is done
- The "timeout between `/syncs`" technically can be infinite as the websocket for itself takes care that the connection is usable
- The client only has to do something when the server pushes something to the client - not every 30s (current timeout for `/sync`)

In general I think that clients that are quite inactive - so where there approximately data every x seconds (>30) - would benefit most while users which get several `/sync`/s might get higher CPU-consumption (might be possible to enhance)

ara4n:
> The speed is nearly the same

I'm surprised by this, as I'd expect to see *some* benefit to not having to set up a new request for every batch (even with batching).  I'd also expect to see quite a noticeable bandwidth advantage too.  It really would be good to try to measure this for different real-world loads (particularly bandwidth).

jryans:
Thanks for making this contribution a while back. Since the code base has changed since this was opened, it no longer applies cleanly, and I don't think there's a need to keep it open in this state, as we can always find the code here again if needed. If you are still interested in pursuing this feature, please discuss with us in #riot-dev:matrix.org to find a good approach forward.

PR #462, opened 2017-06-12
allow setting iceTransportPolicy to relay through forceTURN option

t3chguy:
https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection#RTCIceTransportPolicy_enum


ara4n:
lgtm - thanks :)

PR #465, opened 2017-06-15
Fix npm@5 bug when doing npm install

MTRNord:
NPM@5.x droped support of prepublish for npm install. prepare replaces prepublish. See https://github.com/npm/npm/issues/10074


richvdh:
unfortunately npm v3 doesn't support `prepare`, and node 6 (which is still in LTS and widely used, including amongst the team) ships with npm v3.

In other words: currently this will break things for more people than it fixes things for.

richvdh:
Oh, I see you've already had this conversation over at https://github.com/matrix-org/matrix-react-sdk/pull/1103... let's try and resolve it there

PR #472, opened 2017-06-21
Fix issues with firefox private browsing

richvdh:
Treat errors when deleting indexeddb as non-fatal

also let the webworker be used for deleting the indexeddb without it having previously been used for syncing.

(part of the fix to https://github.com/vector-im/riot-web/issues/4340)

### DIFF
--- a/src/crypto/store/indexeddb-crypto-store.js
+++ b/src/crypto/store/indexeddb-crypto-store.js
@@ -72,9 +72,9 @@ export default class IndexedDBCryptoStore {
             };
 
             req.onblocked = () => {
-                reject(new Error(
-                    "unable to upgrade indexeddb because it is open elsewhere",
-                ));
+                console.log(
+                    `can't yet open IndexedDBCryptoStore because it is open elsewhere`,
+                );
             };
 
             req.onerror = (ev) => {
@@ -110,9 +110,9 @@ export default class IndexedDBCryptoStore {
             const req = this._indexedDB.deleteDatabase(this._dbName);
 
             req.onblocked = () => {
-                reject(new Error(
-                    "unable to delete indexeddb because it is open elsewhere",
-                ));
+                console.log(
+                    `can't yet delete IndexedDBCryptoStore because it is open elsewhere`,
+                );
             };
 
             req.onerror = (ev) => {
--- a/src/crypto/store/indexeddb-crypto-store.js
+++ b/src/crypto/store/indexeddb-crypto-store.js
@@ -116,9 +116,11 @@ export default class IndexedDBCryptoStore {
             };
 
             req.onerror = (ev) => {
-                reject(new Error(
-                    "unable to delete indexeddb: " + ev.target.error,
-                ));
+                // in firefox, with indexedDB disabled, this fails with a
+                // DOMError. We treat this as non-fatal, so that people can
+                // still use the app.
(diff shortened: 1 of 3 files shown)


src/store/indexeddb-remote-backend.js
```diff
@@ -62,7 +64,7 @@ RemoteIndexedDBStoreBackend.prototype = {
      * @return {Promise} Resolved when the database is cleared.
      */
     clearDatabase: function() {
-        return this._doCmd('clearDatabase');
+        return this._startPromise.then(() => this._doCmd('clearDatabase'));
```
dbkr: Should we make the worker not replace the backend if it already has one? otherwise this will make a new worker if you clear the db after having opened it which may cause issues because it's been opened by a different indexedb instance?

src/crypto/store/indexeddb-crypto-store.js
```diff
@@ -110,15 +110,17 @@ export default class IndexedDBCryptoStore {
...
             };
 
             req.onerror = (ev) => {
-                reject(new Error(
-                    "unable to delete indexeddb: " + ev.target.error,
-                ));
+                // in firefox, with indexedDB disabled, this fails with a
+                // DOMError. We treat this as non-fatal, so that people can
+                // still use the app.
+                console.warn(`unable to delete IndexedDBCryptoStore: ${ev.target.error}`);
```
dbkr: Would it be sensible to give up on using indexeddb entirely if we fail to clear the database? This would mitigate the possibility of ending up opening the database with spurious data from a different run.

src/store/indexeddb-remote-backend.js
richvdh: I don't think it will make a new worker?

src/crypto/store/indexeddb-crypto-store.js
richvdh: possibly, but how would we implement that, given this will happen on a different MatrixClient to subsequent access?

src/store/indexeddb-remote-backend.js
dbkr: Will it not create another backend object in https://github.com/matrix-org/matrix-js-sdk/blob/master/src/store/indexeddb-store-worker.js#L59 though?

src/crypto/store/indexeddb-crypto-store.js
dbkr: Yeah, it would have to be a global flag somewhere which is a bit grim - potentially setting the createMatrixClient worker script property to null...

src/store/indexeddb-remote-backend.js
richvdh: No, because _setupWorker is only called once, from the constructor. We're just piggy-backing on the promise created there.

src/crypto/store/indexeddb-crypto-store.js
richvdh: >  setting the createMatrixClient worker script property to null...

createMatrixClient is in a different project...

src/store/indexeddb-remote-backend.js
dbkr: Oh right, of course

src/crypto/store/indexeddb-crypto-store.js
dbkr: yeah, would have to be done in react-sdk in any case I guess, so not one for this PR

PR #473, opened 2017-06-22
Fix failure on Tor browser

richvdh:
If we don't have indexeddb at all, don't try to make an indexeddb crypto store.


richvdh:
fixes https://github.com/vector-im/riot-web/issues/4379

PR #475, opened 2017-06-22
Fall back to MemoryCryptoStore if indexeddb fails

richvdh:
If we get an error when connecting to th indexeddb, fall back to a
MemoryCryptoStore.

This takes a bit of reorganising, because we don't get the error until we try
to connect to the database.


richvdh:
let's hope so.

PR #476, opened 2017-06-26
Add 'group summary' API

dbkr:
Start of group support


src/base-apis.js
```diff
@@ -398,6 +398,11 @@ MatrixBaseApis.prototype.roomState = function(roomId, callback) {
     return this._http.authedRequest(callback, "GET", path);
 };
 
+MatrixBaseApis.prototype.getGroupSummary = function(groupId, callback) {
```
lukebarnard1: I think we've deprecated `callback` so we don't need to add it to new APIs so I think you can just pass `null` to `authedRequest` instead.

PR #482, opened 2017-07-05
Sanity-check response from /thirdparty/protocols

richvdh:
Check that /thirdparty/protocols gives us an object (rather than a string, for
instance). I saw a test explode, apparently because it gave us a string. Which
is odd, but in general we ought to be sanity-checking the things coming back
from the server.


richvdh:
flaky test is flaky

PR #484, opened 2017-07-07
Make it possible to use async/await

richvdh:
Enables the babel plugin that transpiles async/await to generator functions,
and load the regenerator runtime so that generator functions work.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,6 @@
 {
-  "presets": ["es2015"]
+    "presets": ["es2015"],
+    "plugins": [
+        "transform-async-to-generator",
+    ],
 }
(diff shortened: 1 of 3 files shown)


.babelrc
```diff
@@ -1,3 +1,6 @@
 {
-  "presets": ["es2015"]
+    "presets": ["es2015"],
+    "plugins": [
+        "transform-async-to-generator",
+    ],
```
Kegsay: I assume babel loads this as a JS object rather than a JSON file, which is why these trailing commas work?

Kegsay:
LGTM

.babelrc
richvdh: I presume that too. Seems to work, anyway.

PR #489, opened 2017-07-11
make the npm test script windows-friendly

t3chguy:


richvdh:
really? that doesn't work on windows? sigh

PR #490, opened 2017-07-12
Switch matrix-js-sdk to bluebird

richvdh:
The reasons for this are twofold.

Firstly, the use of `q` has long been a thorn in our side. Reasons to prefer bluebird include:
 * much improved performance (allegedly)
 * better stack traces (allegedly)
 * Automatic logging of uncaught exceptions, even if you forget to call `done`.
 * Support for cancellation of pending promises.

Secondly, and more pressingly, I am starting to introduce (on a different branch) use of `async` and `await`, which is much nicer to work with than managing the promises directly, but introduces the problem that (by default) `async` functions return native `Promises`, whilst historical parts of the js-sdk return `q` promises, meaning that one is never sure whether or not to call `done`. With this change, everything will return a bluebird `Promise`, which means that you *can* call `done` (so backwards compatibility is maintained), but don't have to.

It's worth noting that, having done `import Promise from 'bluebird'`, any shims generated for `async` functions by Babel's transform-async-to-generator end up as bluebird Promises.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,9 +49,9 @@
   ],
   "dependencies": {
     "another-json": "^0.2.0",
+    "bluebird": "^3.5.0",
     "browser-request": "^0.3.3",
     "content-type": "^1.0.2",
-    "q": "^1.4.1",
     "regenerator-runtime": "^0.10.5",
     "request": "^2.53.0"
   },
--- a/spec/TestClient.js
+++ b/spec/TestClient.js
@@ -24,7 +24,7 @@ import sdk from '..';
 import testUtils from './test-utils';
 import MockHttpBackend from 'matrix-mock-request';
 import expect from 'expect';
-import q from 'q';
+import Promise from 'bluebird';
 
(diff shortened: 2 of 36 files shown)


spec/npm-debug.log
```diff
@@ -0,0 +1,48 @@
...
+23 error If you do, this is most likely a problem with the matrix-js-sdk package,
+23 error not with npm itself.
+23 error Tell the author that this fails on your system:
+23 error     npm run test:build && npm run test:run
+23 error You can get information on how to open an issue for this project with:
+23 error     npm bugs matrix-js-sdk
+23 error Or if that isn't available, you can get their info via:
+23 error     npm owner ls matrix-js-sdk
+23 error There is likely additional logging output above.
+24 verbose exit [ 1, true ]
```
dbkr: You probably didn't mean to commit this

dbkr:
oh, and apart from the conflict

PR #493, opened 2017-07-12
use device_one_time_keys_count transmitted by /sync

krombel:
As of matrix-org/synapse#2237 the server sends `"device_one_time_keys_count"` as part of the `/sync`-response.

But currently this is nowhere used.
This PR changes this.

Another change: While uploading `one_time_keys` now the response will be used to measure how many keys to upload are left over. Especially useful when trying to abort uploading in tests.

Signed-off-by: Matthias Kesler <krombel@krombel.de>

### DIFF
--- a/spec/integ/matrix-client-crypto.spec.js
+++ b/spec/integ/matrix-client-crypto.spec.js
@@ -524,6 +524,41 @@ describe("MatrixClient crypto", function() {
             });
     });
 
+    it("React on device_one_time_keys send by /sync", function() {
+        // Send a response which causes a key upload
+        const syncDataEmpty = {
+            next_batch: "d",
+            device_one_time_keys_count: {
+                signed_curve25519: 0
+            },
+        };
+        const syncDataFull = {
+            next_batch: "e",
+            device_one_time_keys_count: {
+                signed_curve25519: 70
+            },
+        };
+
+        bobTestClient.httpBackend.when("GET", "/sync").respond(200, syncDataEmpty);
+        return q()
+            .then(() => bobTestClient.start())
+            .then(() => bobTestClient.flushSync())
+            .then(() => bobTestClient.awaitOneTimeKeyUpload())
+            .then((keys) => {
+                expect(Object.keys(keys).length).toEqual(5);
+                expect(Object.keys(bobTestClient.deviceKeys).length).toNotEqual(0);
+
+                bobTestClient.httpBackend.when("POST", "/keys/upload")
+                    .respond(200, (path, content) => {
+                        // This endpoint should now not be called anymore
+                        expect(2).toBe(3);
+                    });
+                bobTestClient.httpBackend.when("GET", "/sync").respond(200, syncFull);
+            })
+            .then(() => bobTestClient.httpBackend.flush('/sync', 2));
+            // As the key store assumes to be full there should be two syncs in a row
+    });
+
     it("Ali sends a message", function(done) {
         q()
             .then(() => aliTestClient.start())
--- a/spec/integ/matrix-client-crypto.spec.js
+++ b/spec/integ/matrix-client-crypto.spec.js
@@ -689,4 +689,54 @@ describe("MatrixClient crypto", function() {
                 return aliTestClient.httpBackend.flush('/keys/query', 1);
             });
     });
+
+    it("React on device_one_time_keys send by /sync", function(done) {
+        // Send a response which causes a key upload
+        const self = aliTestClient;
+        const httpBackend = self.httpBackend;
+        const syncDataEmpty = {
+            next_batch: "a",
+            device_one_time_keys_count: {
+                signed_curve25519: 0,
+            },
+        };
+
+        // enqueue expectations:
+        // * Sync with empty one_time_keys => upload keys
+
+        return Promise.resolve()
+            .then(() => {
+                console.log(self + ': starting');
+                httpBackend.when("GET", "/pushrules").respond(200, {});
+                httpBackend.when("POST", "/filter").respond(200, { filter_id: "fid" });
+                self.expectDeviceKeyUpload();
+
+                // we let the client do a very basic initial sync, which it needs before
+                // it will upload one-time keys.
+                httpBackend.when("GET", "/sync").respond(200, syncDataEmpty);
+
+                self.client.startClient({
+                    // set this so that we can get hold of failed events
+                    pendingEventOrdering: 'detached',
+                });
+
+                return httpBackend.flushAllExpected().then(() => {
+                    console.log(self + ': started');
+                });
+            })
+            .then(() => httpBackend.when("POST", "/keys/upload")
+                .respond(200, (path, content) => {
+                    expect(content.one_time_keys).toBeTruthy();
+                    expect(content.one_time_keys).toNotEqual({});
+                    console.log('received %i one-time keys',
+                                Object.keys(content.one_time_keys).length);
+                    return {};
+                }))
+            .then(() => httpBackend.flushAllExpected())
+            .then((flushed) => {
+                // 1x /keys/upload
+                expect(flushed).toEqual(1);
+                done();
+            });
+    });
 });
--- a/spec/integ/matrix-client-crypto.spec.js
+++ b/spec/integ/matrix-client-crypto.spec.js
@@ -524,43 +524,6 @@ describe("MatrixClient crypto", function() {
             });
     });
 
-    it("React on device_one_time_keys send by /sync", function() {
-        // Send a response which causes a key upload
-        const syncDataEmpty = {
-            next_batch: "d",
-            device_one_time_keys_count: {
-                signed_curve25519: 0,
-            },
-        };
-        const syncDataFull = {
-            next_batch: "e",
-            device_one_time_keys_count: {
-                signed_curve25519: 70,
-            },
-        };
-
-        bobTestClient.httpBackend.when("GET", "/sync").respond(200, syncDataEmpty);
-        return Promise.resolve()
-            .then(() => bobTestClient.start())
-            .then(() => bobTestClient.awaitOneTimeKeyUpload())
-            .then((keys) => {
-                expect(Object.keys(keys).length).toEqual(5);
-                expect(Object.keys(bobTestClient.deviceKeys).length).toNotEqual(0);
-
-                bobTestClient.httpBackend.when("POST", "/keys/upload")
-                    .respond(200, (path, content) => {
-                        // This endpoint should now not be called anymore
-                        expect(2).toEqual(3);
-                    });
-                bobTestClient.httpBackend.when("GET", "/sync").respond(200, syncDataFull);
-            })
-            .then(() => bobTestClient.httpBackend.flush('/sync', 2).then((flushed) => {
-                expect(flushed).toEqual(2);
-            }))
-            .done();
-            // As the key store assumes to be full there should be two syncs in a row
-    });
-
     it("Ali sends a message", function(done) {
         Promise.resolve()
             .then(() => aliTestClient.start())
--- a/spec/integ/matrix-client-crypto.spec.js
+++ b/spec/integ/matrix-client-crypto.spec.js
@@ -690,7 +690,7 @@ describe("MatrixClient crypto", function() {
             });
     });
 
-    it("React on device_one_time_keys send by /sync", function(done) {
+    it("React on device_one_time_keys send by /sync", function() {
         // Send a response which causes a key upload
         const self = aliTestClient;
         const httpBackend = self.httpBackend;
@@ -730,13 +730,21 @@ describe("MatrixClient crypto", function() {
                     expect(content.one_time_keys).toNotEqual({});
                     console.log('received %i one-time keys',
                                 Object.keys(content.one_time_keys).length);
-                    return {};
+                    // cancel futher calls by telling the client
+                    // we have more than we need
+                    return {
+                       device_one_time_keys_count: {
+                           signed_curve25519: 70,
+                       },
+                    };
                 }))
             .then(() => httpBackend.flushAllExpected())
             .then((flushed) => {
                 // 1x /keys/upload
                 expect(flushed).toEqual(1);
-                done();
-            });
+                // ignore all following request
+            })
+            .then(() => self.client.stopClient())
+            .then(() => httpBackend.flush());
     });
 });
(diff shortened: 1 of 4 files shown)


src/crypto/index.js
```diff
@@ -246,8 +246,12 @@ Crypto.prototype.uploadDeviceKeys = function() {
     });
 };
 
+Crypto.prototype.updateCurrentKeyCount = function(currentKeyCount) {
```
richvdh: can you give this a docstring please. What is it for, what does it do?

src/crypto/index.js
```diff
@@ -266,7 +270,10 @@ function _maybeUploadOneTimeKeys(crypto) {
         now - crypto._lastOneTimeKeyCheck < uploadPeriod
        ) {
         // we've done a key upload recently.
-        return;
+        if (currentKeyCount === undefined) {
+            // only return when currentCount (from /sync-response) is not set
```
richvdh: why? We should also back off when the count is returned from the sync response, otherwise the following could happen:

* say we start with zero keys
* we start uploading the keys
* the server starts sending us a sync response (with key_count == 0)
* we finish uploading the keys
* we receive the sync response
* key_count == 0, so we upload lots more keys!

src/crypto/index.js
```diff
@@ -246,8 +246,12 @@ Crypto.prototype.uploadDeviceKeys = function() {
     });
 };
 
+Crypto.prototype.updateCurrentKeyCount = function(currentKeyCount) {
+     _maybeUploadOneTimeKeys(this, currentKeyCount);
```
richvdh: we can't just call this here, because we don't want to start uploading one-time keys until we've caught up with the device messages - see https://github.com/krombel/matrix-js-sdk/blob/fa2982a1bf80fc5ae8994a7d9bb05faf1fcd54ba/src/crypto/index.js#L927.

src/crypto/index.js
```diff
@@ -246,8 +246,12 @@ Crypto.prototype.uploadDeviceKeys = function() {
...
+};
+
 // check if it's time to upload one-time keys, and do so if so.
-function _maybeUploadOneTimeKeys(crypto) {
+function _maybeUploadOneTimeKeys(crypto, currentKeyCount) {
```
richvdh: can you document the currentKeyCount parameter? what is the behaviour when it is passed, or not passed?

src/crypto/index.js
```diff
@@ -288,11 +295,19 @@ function _maybeUploadOneTimeKeys(crypto) {
 
     crypto._oneTimeKeyCheckInProgress = true;
     q().then(() => {
+        if (currentKeyCount !== undefined && isFinite(currentKeyCount)) {
+            // previously this parameter was not defined.
```
richvdh: the parameter is still not defined when this function is called from _onSyncCompleted.

richvdh:
Could you read http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html too please :)

src/crypto/index.js
krombel: This was done by intention. As the value would be `undefined` I can use the call in _onSyncCompleted without worrying that I break up with the old behaviour.

src/crypto/index.js
krombel: I changed to your suggested behavior. And I added a reset after all keys are uploaded. This should avoid this race.

src/crypto/index.js
```diff
@@ -288,11 +304,18 @@ function _maybeUploadOneTimeKeys(crypto) {
 
     crypto._oneTimeKeyCheckInProgress = true;
     q().then(() => {
+        if (crypto._currentKeyCount !== undefined) {
+            // this._currentKeyCount and the handling got postponed. Use this value
```
richvdh: got postponed? what does this mean?

src/crypto/index.js
```diff
@@ -266,7 +279,10 @@ function _maybeUploadOneTimeKeys(crypto) {
         now - crypto._lastOneTimeKeyCheck < uploadPeriod
        ) {
         // we've done a key upload recently.
-        return;
+        if (crypto._currentKeyCount === undefined) {
+            // this._currentKeyCount is not set so handling of it is not postponed.
```
richvdh: why not? what?

src/crypto/index.js
```diff
@@ -326,6 +347,7 @@ function _maybeUploadOneTimeKeys(crypto) {
     }).catch((e) => {
         console.error("Error uploading one-time keys", e.stack || e);
     }).finally(() => {
+        crypto._currentKeyCount = undefined;
```
richvdh: why?

src/crypto/index.js
```diff
@@ -271,6 +283,28 @@ function _maybeUploadOneTimeKeys(crypto) {
...
+    // To complicate things further then can be a delay between a device
+    // claiming a public one time key from the server and it sending us a
+    // message. We need to keep the corresponding private key locally until
+    // we receive the message.
+    // But that message might never arrive leaving us stuck with duff
+    // private keys clogging up our local storage.
+    // So we need some kind of enginering compromise to balance all of
+    // these factors.
+
+    // We then check how many keys we can store in the Account object.
```
richvdh: this comment doesn't make any sense any more, because you have removed the previous sentence ("We first find out...").

I suggest you just `s/We then //`

src/crypto/index.js
```diff
@@ -281,42 +315,29 @@ function _maybeUploadOneTimeKeys(crypto) {
 
         // Ask olm to generate new one time keys, then upload them to synapse.
         crypto._olmDevice.generateOneTimeKeys(keysThisLoop);
-        return _uploadOneTimeKeys(crypto).then(() => {
+        return _uploadOneTimeKeys(crypto).then((res) => {
+            const keyCount = res.one_time_key_counts.signed_curve25519 || 0;
```
richvdh: what's going on here? what is the purpose of || 0? why are we suddenly ignoring `numberToGenerate` but *only* in this case?

(I think I kinda see what you're doing, but wouldn't passing the current `keyCount `into `uploadLoop` instead of `numberToGenerate` be a rather clearer way of doing it?)

src/crypto/index.js
```diff
@@ -246,6 +246,18 @@ Crypto.prototype.uploadDeviceKeys = function() {
     });
 };
 
+/**
+ * Stores the current key count which will be handled postponed (in a call of
```
richvdh: 'postponed' is a verbal adjective, you can't use it as an adverb to modify 'handled' :).  s/postponed/later/.

It would be worth mentioning here where the key count is coming from: the /sync response.

src/crypto/index.js
```diff
@@ -246,6 +246,18 @@ Crypto.prototype.uploadDeviceKeys = function() {
...
+ * _onSyncCompleted)
+ *
+ * @param {Number} currentKeyCount to be stored and handled later
+ */
+Crypto.prototype.updateCurrentKeyCount = function(currentKeyCount) {
```
richvdh: Can we rename this (and update the comment) to reflect the fact that these are *one-time* keys? `updateOneTimeKeyCount` or sth

src/crypto/index.js
```diff
@@ -246,6 +246,18 @@ Crypto.prototype.uploadDeviceKeys = function() {
...
+    if (isFinite(currentKeyCount)) {
```
richvdh: I'd say it's up to the caller of the API to check that the parameter is valid. If you want to double-check it for robustness, throw an exception:

```
if (!isFinite(currentKeyCount)) {
    throw new Error("Invalid parameter to updateCurrentKeyCount");
}
```

or similar

src/crypto/index.js
```diff
@@ -281,42 +315,29 @@ function _maybeUploadOneTimeKeys(crypto) {
...
+                return uploadLoop(Math.max(keyLimit - keyCount, 0));
+            }
             return uploadLoop(numberToGenerate - keysThisLoop);
         });
     }
 
     crypto._oneTimeKeyCheckInProgress = true;
     Promise.resolve().then(() => {
+        if (crypto._currentKeyCount !== undefined) {
+            // this._currentKeyCount got set and the handling got postponed.
```
richvdh: This isn't clear, something like this would be better:

"We already have the current key count from a /sync response. Use this value instead of asking the server for the current key count."

spec/integ/matrix-client-crypto.spec.js
```diff
@@ -689,4 +689,62 @@ describe("MatrixClient crypto", function() {
                 return aliTestClient.httpBackend.flush('/keys/query', 1);
             });
     });
+
+    it("React on device_one_time_keys send by /sync", function() {
```
richvdh: The description here should explain what we are expecting to happen. "uploads an appropriate number of one-time keys given the /sync response" or something.

spec/integ/matrix-client-crypto.spec.js
```diff
@@ -689,4 +689,62 @@ describe("MatrixClient crypto", function() {
...
+                httpBackend.when("POST", "/filter").respond(200, { filter_id: "fid" });
+                self.expectDeviceKeyUpload();
+
+                // we let the client do a very basic initial sync, which it needs before
+                // it will upload one-time keys.
+                httpBackend.when("GET", "/sync").respond(200, syncDataEmpty);
+
+                self.client.startClient({
+                    // set this so that we can get hold of failed events
+                    pendingEventOrdering: 'detached',
```
richvdh: this probably isn't necessary for this test

spec/integ/matrix-client-crypto.spec.js
```diff
@@ -689,4 +689,62 @@ describe("MatrixClient crypto", function() {
...
+
+                return httpBackend.flushAllExpected().then(() => {
+                    console.log(self + ': started');
+                });
+            })
+            .then(() => httpBackend.when("POST", "/keys/upload")
+                .respond(200, (path, content) => {
+                    expect(content.one_time_keys).toBeTruthy();
+                    expect(content.one_time_keys).toNotEqual({});
+                    console.log('received %i one-time keys',
```
richvdh: shouldn't we check the number of keys uploaded here?

spec/integ/matrix-client-crypto.spec.js
```diff
@@ -689,4 +689,62 @@ describe("MatrixClient crypto", function() {
...
+                    return {
+                       device_one_time_keys_count: {
+                           signed_curve25519: 70,
+                       },
+                    };
+                }))
+            .then(() => httpBackend.flushAllExpected())
+            .then((flushed) => {
+                // 1x /keys/upload
+                expect(flushed).toEqual(1);
```
richvdh: `flushAllExpected` will fail if the number flushed doesn't match the number of expectations, so this is redundant.

spec/integ/matrix-client-crypto.spec.js
```diff
@@ -689,4 +689,62 @@ describe("MatrixClient crypto", function() {
...
+                // ignore all following request
+            })
+            .then(() => self.client.stopClient())
+            .then(() => httpBackend.flush());
```
richvdh: this is probably redundant too.

spec/integ/matrix-client-crypto.spec.js
```diff
@@ -689,4 +689,62 @@ describe("MatrixClient crypto", function() {
                 return aliTestClient.httpBackend.flush('/keys/query', 1);
             });
     });
+
+    it("React on device_one_time_keys send by /sync", function() {
+        // Send a response which causes a key upload
+        const self = aliTestClient;
```
richvdh: using `self` instead of `aliTestClient` is unclear. I suggest you just use `aliTestClient` throughout.

spec/integ/matrix-client-crypto.spec.js
krombel: I do not want to run TestClient.start() as this responds to /sync in a way which will led this test fail.
So I copied the part and modified the /sync response. So I need to run client.startClient()

Should I implement this in another way?

spec/integ/matrix-client-crypto.spec.js
krombel: We check if the one_time_keys are `NotEqual` to `{}` so we have one_time_keys as a payload.
To keep this step dynamic I think this should be sufficient

spec/integ/matrix-client-crypto.spec.js
krombel: I run into the case that the test was failing because there was another pending /sync request I did not handle. Need to test if it works now as I added the "break" because we return `70` (I had not implemented this break when I ran into this issue)

spec/integ/matrix-client-crypto.spec.js
richvdh: I mean, you could just do

```
self.client.startClient({});
```

spec/integ/matrix-client-crypto.spec.js
richvdh: well, it's a list, so even if it's empty, it will be `NotEqual` to `{}`:
```
> [] == {}
<- false
```

but still, you could check that at least one key is there, or something. Otherwise I really don't feel like this test is testing much.

krombel: added a line that checks if at least one key is present

src/crypto/index.js
```diff
@@ -271,61 +285,74 @@ function _maybeUploadOneTimeKeys(crypto) {
...
-        const keysThisLoop = Math.min(numberToGenerate, maxKeysPerCycle);
+        const keysThisLoop = Math.min(keyLimit - keyCount, maxKeysPerCycle);
 
         // Ask olm to generate new one time keys, then upload them to synapse.
         crypto._olmDevice.generateOneTimeKeys(keysThisLoop);
-        return _uploadOneTimeKeys(crypto).then(() => {
-            return uploadLoop(numberToGenerate - keysThisLoop);
+        return _uploadOneTimeKeys(crypto).then((res) => {
+            if (res.one_time_key_counts && res.one_time_key_counts.signed_curve25519) {
+                // if the response contains a more up2date value use this
```
richvdh: ugh! "up to date", please!

src/crypto/index.js
```diff
@@ -271,61 +285,74 @@ function _maybeUploadOneTimeKeys(crypto) {
...
+                // for the next loop
+                return uploadLoop(res.one_time_key_counts.signed_curve25519);
+            } else {
+                return uploadLoop(keyCount + keysThisLoop);
```
richvdh: if the response fields aren't present, doesn't that suggest something has gone horribly wrong? it might be better to give up with an error, than to carry on regardless

krombel: Yes, seems reasonable. Will change it

PR #495, opened 2017-07-13
Fix spamming of key-share-requests

richvdh:
Fixes a race in the memory-backed crypto store which meant that we would spam
out multiple key-share-requests for the same session.

(This didn't happen very often in practice, because normally we use the
indexeddb-backed store, which is race-free. Or at least, doesn't have this
race.)

### DIFF
--- a/src/crypto/store/memory-crypto-store.js
+++ b/src/crypto/store/memory-crypto-store.js
@@ -54,8 +54,10 @@ export default class MemoryCryptoStore {
     getOrAddOutgoingRoomKeyRequest(request) {
         const requestBody = request.requestBody;
 
-        // first see if we already have an entry for this request.
-        return this.getOutgoingRoomKeyRequest(requestBody).then((existing) => {
+        return Promise.try(() => {
+            // first see if we already have an entry for this request.
+            const existing = this._getOutgoingRoomKeyRequest(requestBody);
+
             if (existing) {
                 // this entry matches the request - return it.
                 console.log(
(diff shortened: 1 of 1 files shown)


richvdh:
Test fix at https://github.com/matrix-org/matrix-js-sdk/pull/497

src/crypto/store/memory-crypto-store.js
```diff
@@ -54,8 +54,10 @@ export default class MemoryCryptoStore {
     getOrAddOutgoingRoomKeyRequest(request) {
         const requestBody = request.requestBody;
 
-        // first see if we already have an entry for this request.
-        return this.getOutgoingRoomKeyRequest(requestBody).then((existing) => {
+        return Promise.try(() => {
```
lukebarnard1: I'm not seeing an obvious reason for `_getOutgoingRoomKeyRequest` to throw an exception.

lukebarnard1:
I'm not 100% what race this fixes

richvdh:
Ok, sorry, this should probably have come with more explanation.

The important change here is not the use of `Promise.try` in `getOrAddOutgoingRoomKeyRequest` (which, agreed, is neither here nor there), but rather the use of a *synchronous* version of `_getOutgoingRoomKeyRequest` instead of the *asynchronous* `getOutgoingRoomKeyRequest`.

The problem is that this was being called from code like this:

```
promises = [];
for(const event of undecryptableEvents) {
    promises.push(getOrAddOutgoingRoomKeyRequest(makeKeyRequestFor(event)));
}
return Promise.all(promises)
```

... where `makeKeyRequestFor` might return the same thing for multiple events.

So there are multiple calls to `getOrAddOutgoingRoomKeyRequest` with the same request in the same tick, each of which fired off a request to `getOutgoingRoomKeyRequest` and then returned. All of the calls to `getOutgoingRoomKeyRequest` would run, and then we would add all of the (identical) requests to the queue.

lukebarnard1:
>The important change here is ... the use of a synchronous version of _getOutgoingRoomKeyRequest instead of the asynchronous getOutgoingRoomKeyRequest.

But the only thing using the sync version is something that immediately makes it an async thing, so I'm still not sure how this would affect something calling `getOrAddOutgoingRoomKeyRequest`

richvdh:
> But the only thing using the sync version is something that immediately makes it an async thing

https://github.com/matrix-org/matrix-js-sdk/pull/495/files#diff-5ca22f5d766794cc167a563e63e6060cR59 calls it synchronously and then synchronously adds a new entry if none was found.

lukebarnard1:
Ah! OK, I get it now :smile:

Issue #496, opened 2017-07-13
Promise returned by sendMessage is never resolved if the message is cancelled


richvdh:
we should probably reject the promise with a CancelledError or sth?

PR #497, opened 2017-07-13
Fix racy 'matrixclient retrying' test

richvdh:
when a message send fails, the promise returned by `sendMessage` is
rejected. Until we switched to bluebird, the rejection was happily being
swallowed, but with bluebird, there's a better chance of the unhandled
rejection being caught by the runtime and mocha and failing the test.

### DIFF
--- a/spec/integ/matrix-client-retrying.spec.js
+++ b/spec/integ/matrix-client-retrying.spec.js
@@ -1,5 +1,7 @@
 "use strict";
 import 'source-map-support/register';
+import Promise from 'bluebird';
+
 const sdk = require("../..");
 const HttpBackend = require("matrix-mock-request");
 const utils = require("../test-utils");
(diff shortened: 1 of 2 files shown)


spec/integ/matrix-client-retrying.spec.js
```diff
@@ -52,15 +54,20 @@ describe("MatrixClient retrying", function() {
...
-        client.sendMessage(roomId, "m1").then(function(ev) {
-            expect(ev).toEqual(ev1);
-        });
-        client.sendMessage(roomId, "m2").then(function(ev) {
-            expect(ev).toEqual(ev2);
+        const p1 = client.sendMessage(roomId, "m1").then(function(ev) {
+            // we expect the first message to fail
+            throw new Error('Message 1 unexpectedly sent successfully');
+        }, (e) => {
+            // this is expected
```
lukebarnard1: But this won't happen because of  https://github.com/matrix-org/matrix-js-sdk/issues/496?

richvdh: It should do, because this message fails *before* it is cancelled. The promise is rejected when it fails.

OTOH that doesn't explain why the tests are failing. /me investigates

richvdh: well, they are different tests. I don't think they are related.

PR #502, opened 2017-07-18
Fix member events breaking on timeline reset

dbkr:
We were re-adding the listeners too late as state is inserted into
the timeline as part of resetting it causing, among other things,
newMember events to fire. These would be missed, causing those new
RoomMember objects to not get their reEmit set up.

Fixes https://github.com/vector-im/riot-web/issues/4624


richvdh:
as per rambling conversation on #riot-dev: this will end up firing `RoomState.event` events for lots of state events we've already seen... which may be relatively harmless, but doesn't sound desirable. Is there any  way to avoid this?

dbkr:
we decided to do this a different way that doesn't cause replays of all the events from the previous state

PR #503, opened 2017-07-18
Make bits of the js-sdk api asynchronous

richvdh:
We're going to need to make a bunch of these APIs asynchronous, so that we can do indexeddb stuff within them.

Add MatrixClient.initCrypto: initialising the crypto layer needs to become
asynchronous. Rather than making `sdk.createClient` asynchronous, which would
break every single app in the world, add `initCrypto`, which will only break
those attempting to do e2e (and in a way which will fall back to only
supporting unencrypted events).

Make the following return Promises:

* `MatrixClient.getStoredDevicesForUser`
* `MatrixClient.getStoredDevice`
* `MatrixClient.setDeviceVerified`
* `MatrixClient.setDeviceBlocked`
* `MatrixClient.setDeviceKnown`
* `MatrixClient.getEventSenderDeviceInfo`
* `MatrixClient.isEventSenderVerified`
* `MatrixClient.importRoomKeys`

Remove `listDeviceKeys` altogether: it's been deprecated for ages, and
since applications are going to have to be changed anyway, they might as
well use its replacement (`getStoredDevices`).

The changes required to react-sdk are in https://github.com/matrix-org/matrix-react-sdk/pull/1233.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,41 @@
+Unreleased changes
+==================
+
+BREAKING CHANGE
+---------------
+
+In order to support a move to a more scalable storage backend, we need to make
+a number of the APIs related end-to-end encryption asynchronous.
+
+This release of the JS-SDK includes the following changes which will affect
+applications which support end-to-end encryption:
+
+1. `MatrixClient` now provides a new (asynchronous) method,
+   `initCrypto`. Applications which support end-to-end encryption must call
+   this method (and wait for it to complete) before calling `startClient`, to
+   give the crypto layer a chance to initialise.
+
+2. The following APIs have been changed to return promises:
+
+   * `MatrixClient.getStoredDevicesForUser`
+   * `MatrixClient.getStoredDevice`
+   * `MatrixClient.setDeviceVerified`
+   * `MatrixClient.setDeviceBlocked`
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,41 @@
...
+   * `MatrixClient.getEventSenderDeviceInfo`
+   * `MatrixClient.isEventSenderVerified`
+   * `MatrixClient.importRoomKeys`
+
+   Applications using the results of any of the above methods will need to be
+   updated to wait for the result of the promise.
+
+
+3. `MatrixClient.listDeviceKeys` has been removed altogether. It's been
+   deprecated for some time. Applications using it should insted be changed to
```
lukebarnard1: spelling

src/client.js
```diff
@@ -324,6 +312,72 @@ MatrixClient.prototype.setNotifTimelineSet = function(notifTimelineSet) {
...
+    );
+
+    reEmit(this, crypto, [
+        "crypto.roomKeyRequest",
+        "crypto.roomKeyRequestCancellation",
+    ]);
+
+    await crypto.init();
+
+    // if crypto initialisation was sucessful, tell it to attach its event handlers.
```
lukebarnard1: successful

PR #504, opened 2017-07-19
Fix member events breaking on timeline reset, 2

dbkr:
Re-use the same RoomState from the old live timeline so we re-use
all the same member objects etc, so all the listeners stay attached

### DIFF
--- a/src/models/event-timeline-set.js
+++ b/src/models/event-timeline-set.js
@@ -162,8 +162,8 @@ EventTimelineSet.prototype.resetLiveTimeline = function(
     // special because they will have listeners attached to monitor changes to
     // the current room state, so we move this RoomState from the end of the
     // current live timeline to the end of the new one and, if necessary,
-    // replace it with the fresh one created for the new timeline. We also make
-    // a copy for the start of the new timeline.
+    // replace it with a newly created one. We also make a copy for the start
+    // of the new timeline.
 
     // if timeline support is disabled, forget about the old timelines
     const resetAllTimelines = !this._timelineSupport || !forwardPaginationToken;
--- a/src/models/event-timeline-set.js
+++ b/src/models/event-timeline-set.js
@@ -180,8 +180,6 @@ EventTimelineSet.prototype.resetLiveTimeline = function(
         // we just stole it and put it on the new live timeline
         // (If we're resetting all timelines, don't bother because the old live
         // timeline is about to be thrown away anyway.
-        this._liveTimeline._startState = new RoomState(this._roomId);
-        this._liveTimeline._endState = new RoomState(this._roomId);
         const evMap = this._liveTimeline.getState(EventTimeline.FORWARDS).events;
         const events = [];
         for (const evtype in evMap) {
@@ -195,6 +193,8 @@ EventTimelineSet.prototype.resetLiveTimeline = function(
                 events.push(evMap[evtype][stateKey]);
             }
         }
+        this._liveTimeline._startState = new RoomState(this._roomId);
+        this._liveTimeline._endState = new RoomState(this._roomId);
         this._liveTimeline.initialiseState(events);
 
         this._liveTimeline.setPaginationToken(
--- a/src/models/event-timeline-set.js
+++ b/src/models/event-timeline-set.js
@@ -188,9 +188,10 @@ EventTimelineSet.prototype.resetLiveTimeline = function(
     newTimeline.initialiseState(events);
 
     const freshEndState = newTimeline._endState;
-    // Now clobber the end state with that from the previous live timeline.
-    // It will be identical except that we'll keep using the same RoomMember
-    // objects for the 'live' set of members with any listeners still attached
+    // Now clobber the end state of the new live timeline with that from the
+    // previous live timeline. It will be identical except that we'll keep
+    // using the same RoomMember objects for the 'live' set of members with any
+    // listeners still attached
     newTimeline._endState = this._liveTimeline._endState;
 
     // If we're not resetting all timelines, we need to fix up the old live timeline
@@ -200,6 +201,8 @@ EventTimelineSet.prototype.resetLiveTimeline = function(
         // new live timeline.
         this._liveTimeline._endState = freshEndState;
 
+        // Now set the forward pagination token on the old live timeline
+        // so it can be forward-paginated.
         this._liveTimeline.setPaginationToken(
             forwardPaginationToken, EventTimeline.FORWARDS,
         );
@@ -210,6 +213,7 @@ EventTimelineSet.prototype.resetLiveTimeline = function(
     // stuck without realising that they *can* back-paginate.
     newTimeline.setPaginationToken(backPaginationToken, EventTimeline.BACKWARDS);
 
+    // Now we can swap the live timeline to the new one.
     this._liveTimeline = newTimeline;
     this.emit("Room.timelineReset", this.room, this, resetAllTimelines);
 };
--- a/src/models/event-timeline-set.js
+++ b/src/models/event-timeline-set.js
@@ -20,7 +20,6 @@ limitations under the License.
 const EventEmitter = require("events").EventEmitter;
 const utils = require("../utils");
 const EventTimeline = require("./event-timeline");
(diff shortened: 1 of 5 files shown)


src/models/event-timeline-set.js
```diff
@@ -149,13 +149,14 @@ EventTimelineSet.prototype.replaceEventId = function(oldEventId, newEventId) {
  * <p>This is used when /sync returns a 'limited' timeline.
  *
  * @param {string=} backPaginationToken   token for back-paginating the new timeline
- * @param {?bool} flush  Whether to flush the non-live timelines too.
```
Kegsay: Why are you removing this? We need this and use it to purge old timelines.

dbkr: See the next parameter

src/models/event-timeline-set.js
```diff
@@ -166,21 +167,35 @@ EventTimelineSet.prototype.resetLiveTimeline = function(backPaginationToken, flu
...
+    newTimeline._endState = this._liveTimeline._endState;
+
+    if (!resetAllTimelines) {
+        // Now regenerate the state for the previously-live timeline, because
+        // we just stole it and put it on the new live timeline
+        // (If we're resetting all timelines, don't bother because the old live
+        // timeline is about to be thrown away anyway.
+        this._liveTimeline._startState = new RoomState(this._roomId);
+        this._liveTimeline._endState = new RoomState(this._roomId);
+        const evMap = this._liveTimeline.getState(EventTimeline.FORWARDS).events;
```
Kegsay: Aren't we doing this too late? It'll be an empty map since we just clobbered the state?

dbkr: oh, yes, probably

src/models/event-timeline-set.js
```diff
@@ -166,21 +170,37 @@ EventTimelineSet.prototype.resetLiveTimeline = function(backPaginationToken, flu
...
-    const events = [];
-    for (const evtype in evMap) {
-        if (!evMap.hasOwnProperty(evtype)) {
-            continue;
-        }
-        for (const stateKey in evMap[evtype]) {
-            if (!evMap[evtype].hasOwnProperty(stateKey)) {
+    // move the state objects over from the old live timeline, then we'll
+    // keep using the same RoomMember objects for the 'live' set of members
+    newTimeline._startState = this._liveTimeline._startState;
```
richvdh: shouldn't this be `this._liveTimeline._endState`? Or rather, a new copy thereof?

src/models/event-timeline-set.js
```diff
@@ -166,21 +170,37 @@ EventTimelineSet.prototype.resetLiveTimeline = function(backPaginationToken, flu
...
+    newTimeline._endState = this._liveTimeline._endState;
+
+    if (!resetAllTimelines) {
+        // Now regenerate the state for the previously-live timeline, because
+        // we just stole it and put it on the new live timeline
+        // (If we're resetting all timelines, don't bother because the old live
+        // timeline is about to be thrown away anyway.
+        this._liveTimeline._startState = new RoomState(this._roomId);
```
richvdh: ... and this one can stay in place?

src/models/event-timeline-set.js
Kegsay: That's not backwards compatible.

src/models/event-timeline-set.js
dbkr: Good point - have done some rejigging.

src/models/event-timeline-set.js
richvdh: do we care about backwards compat here? I can't imagine any app having any valid reason for calling this method.

src/models/event-timeline-set.js
```diff
@@ -180,8 +183,29 @@ EventTimelineSet.prototype.resetLiveTimeline = function(backPaginationToken, flu
...
+    // It will be identical except that we'll keep using the same RoomMember
+    // objects for the 'live' set of members with any listeners still attached
+    newTimeline._endState = this._liveTimeline._endState;
+
+    if (!resetAllTimelines) {
+        // We just stole the old timeline's end state, so it needs a new one.
+        // Just swap them around and give it the one we just generated for the
+        // new live timeline.
+        // (If we're resetting all timelines, don't bother because the old live
+        // timeline is about to be thrown away anyway.
```
richvdh: ... but the new one isn't, and at this point it is using new RoomMember objects, unless I misunderstand?

richvdh: I do misunderstand.

src/models/room.js
```diff
@@ -205,12 +205,14 @@ Room.prototype.getLiveTimeline = function() {
  * <p>This is used when /sync returns a 'limited' timeline.
  *
  * @param {string=} backPaginationToken   token for back-paginating the new timeline
- * @param {boolean=} flush True to remove all events in all timelines. If false, only
- * the live timeline is reset.
+ * @param {string=} forwardPaginationToken token for forward-paginating the old live timeline,
+ * if absent or null, all timelines are reset.
```
richvdh: at first sight, this leads to a slightly weird-feeling API. Why does the absence of a forward-pagination token mean that we reset *all* timelines?

(I think it's plausible: it might just need some more explanation.)

src/models/event-timeline-set.js
Kegsay: Riot does for purging timelines if you're not viewing the room IIRC

dbkr:
Added some more comments - hopefully they clear things up

src/models/event-timeline-set.js
richvdh: not afaict

richvdh:
well, they kinda help, but what I really think would help would be a separate block comment at the start of the function.

> Each EventTimeline has RoomState objects tracking the state at the start and end of that timeline. We move the existing RoomState from the end of the current live timeline to the end of the new one, and make a copy for the start of it.

etc

dbkr:
added - have fleshed it out a bit with more details of the swapping voodoo which hopefully improves clarity, if not also happy to go with your version :)

src/models/event-timeline-set.js
```diff
@@ -149,13 +149,24 @@ EventTimelineSet.prototype.replaceEventId = function(oldEventId, newEventId) {
...
-EventTimelineSet.prototype.resetLiveTimeline = function(backPaginationToken, flush) {
+EventTimelineSet.prototype.resetLiveTimeline = function(
+    backPaginationToken, forwardPaginationToken,
+) {
+    // Each EventTimeline has RoomState objects tracking the state at the start
+    // and end of that timeline. The copies at the end of the live timeline are
+    // special because they will have listeners attached to monitor changes to
+    // the current room state, so we move this RoomState from the end of the
+    // current live timeline to the end of the new one and, if necessary,
+    // replace it with the fresh one created for the new timeline. We also make
```
richvdh: it's not really created for the new timeline, though? I think "replace it with a newly created one" would be clearer.

src/models/event-timeline-set.js
Kegsay: Right ignore me, we call `this.opts.canResetEntireTimeline` still, and just ternary it to set the token or not.

src/models/event-timeline-set.js
```diff
@@ -180,13 +191,37 @@ EventTimelineSet.prototype.resetLiveTimeline = function(backPaginationToken, flu
...
+    // If we're not resetting all timelines, we need to fix up the old live timeline
+    if (!resetAllTimelines) {
+        // Firstly, we just stole the old timeline's end state, so it needs a new one.
+        // Just swap them around and give it the one we just generated for the
+        // new live timeline.
+        this._liveTimeline._endState = freshEndState;
+
+        // Now set the forward pagination token on the old live timeline
+        // so it can be forward-paginated.
+        this._liveTimeline.setPaginationToken(
```
Kegsay: Nit: A helpfully named `oldTimeline` variable to complement `newTimeline` would help readability on this.

Kegsay:
Otherwise LGTM, great job :)

Kegsay:
LGTM! :shipit:

PR #505, opened 2017-07-20
Use babel transform-runtime instead of regenerator-runtime

richvdh:
Attempting to use the regnerator-runtime ourselves led to a fight with riot-web about whether `global.regeneratorRuntime` could be defined. By using the transform-runtime plugin, references to `global.regeneratorRuntime` which are created by the transform-regenerator plugin are turned into references to an imported module, which works much better.

(The full tragic tale went as follows:

- riot-web uses transform-runtime, which adds an import of `regenerator-runtime` to index.js
- `regenerator-runtime`:
   - loads `regenerator-runtime/runtime`, which defines `global.regeneratorRuntime`
   - then clears the global property and returns the regeneratorRuntime object as the exported value from the module
- later, the js-sdk tried to import `regenerator-runtime/runtime`, which then did nothing because the module had already been loaded once.

For added fun, this only manifested itself when riot-web and js-sdk shared an instance of the `regenerator-runtime` package, which happens on proper builds, but not a normal development setup.)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,13 @@
 {
     "presets": ["es2015"],
     "plugins": [
+        // this transforms async functions into generator functions, which
+        // are then made to use the regenerator module by babel's
+        // transform-regnerator plugin (which is enabled by es2015.
         "transform-async-to-generator",
+
+        // This makes sure that the regenerator runtime is available to
+        // the transpiled code.
+        "transform-runtime",
     ],
 }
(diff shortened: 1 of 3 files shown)


.babelrc
```diff
@@ -1,6 +1,13 @@
 {
     "presets": ["es2015"],
     "plugins": [
+        // this transforms async functions into generator functions, which
+        // are then made to use the regenerator module by babel's
+        // transform-regnerator plugin (which is enabled by es2015.
```
Kegsay: Missing `)`

Kegsay:
Otherwise LGTM assuming tests pass.

PR #507, opened 2017-07-20
Retrying test: wait for localEchoUpdated event

richvdh:
We need to wait for the js-sdk to have an opportunity to process the 400 from
the /send/ request before checking the event state

### DIFF
--- a/spec/integ/matrix-client-retrying.spec.js
+++ b/spec/integ/matrix-client-retrying.spec.js
@@ -93,7 +93,14 @@ describe("MatrixClient retrying", function() {
             }).toThrow();
         }).respond(400); // fail the first message
 
-        const p3 = httpBackend.flush().then(function() {
+        // wait for the localecho of ev1 to be updated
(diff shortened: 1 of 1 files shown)


spec/integ/matrix-client-retrying.spec.js
```diff
@@ -93,7 +93,14 @@ describe("MatrixClient retrying", function() {
             }).toThrow();
         }).respond(400); // fail the first message
 
-        const p3 = httpBackend.flush().then(function() {
+        // wait for the localecho of ev1 to be updated
+        const p3 = new Promise((resolve, reject) => {
+            room.on("Room.localEchoUpdated", (ev0) => {
+                if(ev0 === ev1) {
+                    resolve();
+                }
```
dbkr: does this need an else reject?

richvdh: no, the thing is that there are a couple of localEchoUpdated events and we specifically want to wait for the ev1 one.

PR #508, opened 2017-07-21
Refactor handling of crypto events for async

richvdh:
We're going to need to handle m.room.crypto events asynchronously, so restructure the way we do that.


package.json
```diff
@@ -14,7 +14,7 @@
     "build": "babel -s -d lib src && rimraf dist && mkdir dist && browserify -d browser-index.js | exorcist dist/browser-matrix.js.map > dist/browser-matrix.js && uglifyjs -c -m -o dist/browser-matrix.min.js --source-map dist/browser-matrix.min.js.map --in-source-map dist/browser-matrix.js.map dist/browser-matrix.js",
     "dist": "npm run build",
     "watch": "watchify -d browser-index.js -o 'exorcist dist/browser-matrix.js.map > dist/browser-matrix.js' -v",
-    "lint": "eslint --max-warnings 110 src spec",
+    "lint": "eslint --max-warnings 109 src spec",
```
Kegsay: ⭐️

richvdh:
It looks like we're going to have to land #509 to sort out the tests here.

Kegsay:
Surprisingly, LGTM. I have nothing of note to say on this, good stuff!

PR #509, opened 2017-07-21
Let event decryption be asynchronous

richvdh:
Once everything moves to indexeddb, it's going to require callbacks and the like, so let's make the decrypt API asynchronous in preparation.

(There are two commits here; they should make sense independently)

### DIFF
--- a/spec/integ/matrix-client-crypto.spec.js
+++ b/spec/integ/matrix-client-crypto.spec.js
@@ -274,8 +274,8 @@ function expectSendMessageRequest(httpBackend) {
         };
     });
 
-    // it can take a while to process the key query, so give it 20ms
-    return httpBackend.flush(path, 1, 20).then(() => deferred.promise);
+    // it can take a while to process the key query
+    return httpBackend.flush(path, 1).then(() => deferred.promise);
 }
 
 function aliRecvMessage() {
@@ -542,19 +542,18 @@ describe("MatrixClient crypto", function() {
             .nodeify(done);
     });
 
-    it("Bob receives a message", function(done) {
-        Promise.resolve()
+    it("Bob receives a message", function() {
+        return Promise.resolve()
             .then(() => aliTestClient.start())
             .then(() => bobTestClient.start())
             .then(() => firstSync(aliTestClient))
             .then(aliEnablesEncryption)
             .then(aliSendsFirstMessage)
-            .then(bobRecvMessage)
-            .nodeify(done);
+            .then(bobRecvMessage);
     });
 
-    it("Bob receives a message with a bogus sender", function(done) {
-        Promise.resolve()
+    it("Bob receives a message with a bogus sender", function() {
+        return Promise.resolve()
             .then(() => aliTestClient.start())
             .then(() => bobTestClient.start())
             .then(() => firstSync(aliTestClient))
@@ -594,20 +593,28 @@ describe("MatrixClient crypto", function() {
                         return;
                     }
 
-                    expect(event.getType()).toEqual("m.room.message");
-                    expect(event.getContent().msgtype).toEqual("m.bad.encrypted");
                     expect(event.isEncrypted()).toBeTruthy();
 
-                    bobTestClient.client.removeListener("event", onEvent);
-                    deferred.resolve();
+                    // it will still be encrypted at first - wait for it to be decrypted.
+                    expect(event.getType()).toEqual("m.room.encrypted");
+                    expect(event.isBeingDecrypted()).toBe(true);
+
+                    console.log(
+                        `bob waiting for event to be decrypted`,
+                    );
+                    event.once('Event.decrypted', (ev) => {
+                        expect(ev).toBe(event);
+                        expect(event.getType()).toEqual("m.room.message");
+                        expect(event.getContent().msgtype).toEqual("m.bad.encrypted");
+                        deferred.resolve();
+                    });
                 };
 
-                bobTestClient.client.on("event", onEvent);
+                bobTestClient.client.once("event", onEvent);
 
                 bobTestClient.httpBackend.flush();
                 return deferred.promise;
-            })
-            .nodeify(done);
+            });
     });
 
     it("Ali blocks Bob's device", function(done) {
--- a/src/crypto/algorithms/base.js
+++ b/src/crypto/algorithms/base.js
@@ -113,13 +113,10 @@ class DecryptionAlgorithm {
      * @method module:crypto/algorithms/base.DecryptionAlgorithm#decryptEvent
      * @abstract
      *
-     * @param {object} event raw event
+     * @param {MatrixEvent} event undecrypted event
      *
-     * @return {null} if the event referred to an unknown megolm session
-     * @return {module:crypto.DecryptionResult} decryption result
-     *
-     * @throws {module:crypto/algorithms/base.DecryptionError} if there is a
-     *   problem decrypting the event
+     * @return {Promise} resolves once we have finished decrypting. Rejects with an
+     * `algorithms.DecryptionError` if there is a problem decrypting the event.
      */
 
     /**
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -164,16 +164,7 @@ Crypto.prototype.registerEventHandlers = function(eventEmitter) {
     });
 
     eventEmitter.on("toDeviceEvent", function(event) {
-        try {
-            if (event.getType() == "m.room_key"
-                    || event.getType() == "m.forwarded_room_key") {
-                crypto._onRoomKeyEvent(event);
-            } else if (event.getType() == "m.room_key_request") {
-                crypto._onRoomKeyRequestEvent(event);
-            }
-        } catch (e) {
-            console.error("Error handling toDeviceEvent:", e);
-        }
+        crypto._onToDeviceEvent(event);
     });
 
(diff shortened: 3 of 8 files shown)


spec/integ/matrix-client-crypto.spec.js
```diff
@@ -542,19 +542,18 @@ describe("MatrixClient crypto", function() {
             .nodeify(done);
     });
 
-    it("Bob receives a message", function(done) {
-        Promise.resolve()
+    it("Bob receives a message", function() {
```
Kegsay: Removing `done` makes the test synchronous, at least in Jasmine it does. Can you link to me the docs which show that returning a promise from a test is sufficient to make the test async?

src/crypto/algorithms/base.js
```diff
@@ -113,13 +113,10 @@ class DecryptionAlgorithm {
...
      *
-     * @param {object} event raw event
+     * @param {MatrixEvent} event undecrypted event
      *
-     * @return {null} if the event referred to an unknown megolm session
-     * @return {module:crypto.DecryptionResult} decryption result
-     *
-     * @throws {module:crypto/algorithms/base.DecryptionError} if there is a
-     *   problem decrypting the event
+     * @return {Promise} resolves once we have finished decrypting. Rejects with an
```
Kegsay: Resolves to what? Can it still resolve to `null` to indicate an unknown megolm session?

src/crypto/index.js
```diff
@@ -973,6 +965,25 @@ Crypto.prototype._getE2eRooms = function() {
...
 
+
+Crypto.prototype._onToDeviceEvent = function(event) {
+    try {
+        if (event.getType() == "m.room_key"
+            || event.getType() == "m.forwarded_room_key") {
+            this._onRoomKeyEvent(event);
+        } else if (event.getType() == "m.room_key_request") {
+            this._onRoomKeyRequestEvent(event);
+        } else if (event.isBeingDecrypted()) {
```
Kegsay: It feels like the ordering is important here? `_onToDeviceEvent` is being called with `m.room.encrypted` events? And only after decryption do we know if they are `m.room_key` etc requests and that's why we recursively call this after decryption?

src/models/event.js
```diff
@@ -327,17 +347,50 @@ utils.extend(module.exports.MatrixEvent.prototype, {
...
             );
+            this._retryDecryption = true;
+            return;
+        }
+
+        this._decrypting = true;
+
+        this._doDecryption(crypto).finally(() => {
+            this._decrypting = false;
+            this._retryDecryption = false;
```
Kegsay: Why do we set this to false without inspecting for its truthiness and retrying? I see we do a check in `_doDecryption` but it's only a half retry as we don't retry even if asked to if we fail with a non-DecryptionError. That feels strange and poorly justified.

src/models/event.js
```diff
@@ -299,7 +310,16 @@ utils.extend(module.exports.MatrixEvent.prototype, {
...
+     * Check if this event is currently being decrypted.
+     *
+     * @return {boolean} True if this event is currently being decrypted, else false.
+     */
+    isBeingDecrypted: function() {
+        return this._decrypting;
+    },
+
+    /**
+     * Start the process of trying to decrypt this event.
```
Kegsay: This feels useless. Callers can kick off the process of decrypting but have no way to know when it succeeds or fails. Why aren't we returning a Promise here?

src/models/event.js
```diff
@@ -299,6 +299,58 @@ utils.extend(module.exports.MatrixEvent.prototype, {
...
+        ) {
+            // we may want to just ignore this? let's start with rejecting it.
+            throw new Error(
+                "Attempt to decrypt event which has already been encrypted",
+            );
+        }
+
+        try {
+            crypto.decryptEvent(this);
+        } catch (e) {
```
Kegsay: Is this actually ever going to do anything? The previous commit changed `decryptEvent` to be:
```js
OlmDecryption.prototype.decryptEvent = async function(event) {
```
Meaning this function will never throw, it will always return a Promise. If you `yield`ed then it might throw though, but you don't.

Kegsay:
For some reason this overview page isn't showing https://github.com/matrix-org/matrix-js-sdk/pull/509/commits/86f2c86440315bcdd016c4d617dbc166ebf5e65c#r128784338

spec/integ/matrix-client-crypto.spec.js
richvdh: https://mochajs.org/#working-with-promises

src/crypto/algorithms/base.js
richvdh: It resolves to undefined. The old return description was actually a lie.

src/crypto/index.js
richvdh: yes.

Kegsay: Then I don't understand at what point we start decrypting these events. I guess they are being started before the event is emitted?

Kegsay: Oh, it's when we map them from JS objects to MatrixEvents.

src/models/event.js
```diff
@@ -299,6 +310,100 @@ utils.extend(module.exports.MatrixEvent.prototype, {
...
+     * @param {module:crypto} crypto crypto module
+     */
+    attemptDecryption: function(crypto) {
+        if (!crypto) {
+            this._badEncryptedMessage("Encryption not enabled");
+            return;
+        }
+
+        if (!this.isEncrypted()) {
+            throw new Error("Attempt to decrypt event which isn't encrypted");
```
Kegsay: Why do we throw for unencrypted messages but set the encrypted message if there is no `crypto` module?

richvdh: the former sounds like somebody using the API incorrectly. The second happens whenever you receive an encrypted event but haven't loaded `olm.js` (because you're in Syria, or whatever)

Kegsay: I don't know why we make the distinction. In both cases the app is unusable, and it just creates confusion as to what you should be handling error-wise IMO.

src/models/event.js
Kegsay: nvm this was git commit messup.

src/models/event.js
richvdh: There being no crypto module is an expected condition. The app is not usable; the user can use it as normal in unencrypted rooms. Encrypted messages are shown as "Cannot decrypt: Encryption not enabled".

`attemptDecryption` being called on an event which is not encrypted is not an expected condition; it means something has gone wrong somewhere; the test here is just for robustness.

What would you like me to do?

src/models/event.js
richvdh: mostly because the places this is called from don't actually care about when it completes. I could make it return a promise if you want.

src/models/event.js
richvdh: _retryDecryption is set when we have new data that might help fix a DecryptionError. It's never going to fix any other sort of error, so retrying the decryption will just cause more errors.

src/models/event.js
Kegsay: Feels icky API-wise if we don't return a promise even if this is an internal function. I would instinctually be asking myself "**why** aren't we returning a promise here" if I were to hack on this code, thinking there may be a deeper meaning other than "we have never cared so we've never bothered". If you can make it return a promise and leave it up to the callers to use/discard that'd be appreciated.

src/models/event.js
Kegsay: I guess my point is more that it's unclear that when you call `event.attemptDecryption` a 2nd time, that it will sometimes maybe attempt a 2nd decryption iff we previously failed with a `DecryptionError`.

It seems backwards for the decryption function to be dictating that "no really, you can't retry decrypting this" when surely it should be up to the caller to decide this. My concern is that there will be a case whereby decryption fails without a `DecryptionError` and the "obvious" fix of adjusting internal state followed by `attemptDecryption` again will just not work because of this function's implementation. Can you allay my concerns?

src/models/event.js
Kegsay: > it means something has gone wrong somewhere

I don't agree with this. It only means that we failed to do a conditional check prior to calling this function. This makes the API cumbersome to deal with because there's so many ways you can do it "the wrong way". I would rather that there wasn't a wrong way, eg by nopping if the event is decrypted/doesn't need decryption. I don't think that failing to do this conditional check is a symptom of the entire state machine being corrupt, which I think is what you're implying (in which case failing fast is highly desirable).

I'm happy to punt this for now, but I'd like to re-visit this at some point.

Kegsay:
My remaining concerns are all to do with the internal API shape of JS SDK crypto:
- Not bothering to returning promises since we don't use happen to use them.
- Not having a clear idea what multiple calls to the same function will do.
- Throwing errors for failing to do a simple conditional check on what would otherwise be a nop.

 This isn't relevant to this PR, so I'm happy to LGTM this to avoid introducing artificial roadblocks, but I'd like to revisit some of this design at some point.

src/models/event.js
richvdh: oh, another reason, which I'll put down here for future reference:

In the situation that we already have an attempt ongoing, we effectively queue up a retry in the event that decryption fails, and return immediately. If we returned a promise here, we'd want to resolve that promise once the *retry* completes - which is a slightly complicated dance which seems over-engineered when we never use the promise.

PR #510, opened 2017-07-21
Make a load of methods in the 'Crypto' module asynchronous

richvdh:
There are 6 commits here, each making a different method asynchronous, and updating its callsites.

It depends on https://github.com/matrix-org/matrix-js-sdk/pull/508 to build.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -632,12 +632,15 @@ MatrixClient.prototype.exportRoomKeys = function() {
  * Import a list of room keys previously exported by exportRoomKeys
  *
  * @param {Object[]} keys a list of session export objects
+ *
+ * @return {module:client.Promise} a promise which resolves when the keys
+ *    have been imported
  */
-MatrixClient.prototype.importRoomKeys = async function(keys) {
+MatrixClient.prototype.importRoomKeys = function(keys) {
     if (!this._crypto) {
         throw new Error("End-to-end encryption disabled");
     }
-    this._crypto.importRoomKeys(keys);
+    return this._crypto.importRoomKeys(keys);
 };
 
 // Room ops
--- a/src/client.js
+++ b/src/client.js
@@ -590,14 +590,13 @@ MatrixClient.prototype.isEventSenderVerified = async function(event) {
  * Enable end-to-end encryption for a room.
  * @param {string} roomId The room ID to enable encryption in.
  * @param {object} config The encryption config for the room.
- * @return {Object} A promise that will resolve when encryption is setup.
+ * @return {Promise} A promise that will resolve when encryption is set up.
  */
 MatrixClient.prototype.setRoomEncryption = function(roomId, config) {
     if (!this._crypto) {
         throw new Error("End-to-End encryption disabled");
     }
-    this._crypto.setRoomEncryption(roomId, config);
-    return Promise.resolve();
+    return this._crypto.setRoomEncryption(roomId, config);
 };
 
 /**
--- a/src/client.js
+++ b/src/client.js
@@ -523,7 +523,7 @@ async function _setDeviceVerification(
     if (!client._crypto) {
         throw new Error("End-to-End encryption disabled");
     }
-    const dev = client._crypto.setDeviceVerification(
+    const dev = await client._crypto.setDeviceVerification(
         userId, deviceId, verified, blocked, known,
     );
     client.emit("deviceVerificationChanged", userId, deviceId, dev);
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -580,7 +580,7 @@ Crypto.prototype.getEventSenderDeviceInfo = function(event) {
  * @param {boolean=} inhibitDeviceQuery true to suppress device list query for
  *   users in the room (for now)
  */
-Crypto.prototype.setRoomEncryption = function(roomId, config, inhibitDeviceQuery) {
+Crypto.prototype.setRoomEncryption = async function(roomId, config, inhibitDeviceQuery) {
     // if we already have encryption in this room, we should ignore this event
     // (for now at least. maybe we should alert the user somehow?)
     const existingConfig = this._sessionStore.getEndToEndRoom(roomId);
@@ -842,7 +842,7 @@ Crypto.prototype.onCryptoEvent = async function(event) {
     try {
         // inhibit the device list refresh for now - it will happen once we've
         // finished processing the sync, in _onSyncCompleted.
-        this.setRoomEncryption(roomId, content, true);
+        await this.setRoomEncryption(roomId, content, true);
     } catch (e) {
         console.error("Error configuring encryption in room " + roomId +
                       ":", e);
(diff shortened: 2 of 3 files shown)


src/client.js
```diff
@@ -517,16 +517,17 @@ MatrixClient.prototype.setDeviceKnown = function(userId, deviceId, known) {
...
     if (!client._crypto) {
         throw new Error("End-to-End encryption disabled");
     }
-    const dev = client._crypto.setDeviceVerification(
+    return client._crypto.setDeviceVerification(
         userId, deviceId, verified, blocked, known,
-    );
-    client.emit("deviceVerificationChanged", userId, deviceId, dev);
+    ).then((dev) => {
+        client.emit("deviceVerificationChanged", userId, deviceId, dev);
```
Kegsay: Why swap from async back to promises here? If you need to wait for `client._crypto.setDeviceVerification` now why aren't you just doing:
```js
const dev = await client._crypto.setDeviceVerification(...);
client.emit("deviceVerificationChanged", userId, deviceId, dev);
```

src/client.js
```diff
@@ -590,14 +591,13 @@ MatrixClient.prototype.isEventSenderVerified = async function(event) {
  * Enable end-to-end encryption for a room.
  * @param {string} roomId The room ID to enable encryption in.
  * @param {object} config The encryption config for the room.
- * @return {Object} A promise that will resolve when encryption is setup.
+ * @return {Object} A promise that will resolve when encryption is set up.
```
Kegsay: If we return a Promise then please `{Promise}`.

src/crypto/index.js
```diff
@@ -377,22 +380,26 @@ function _uploadOneTimeKeys(crypto) {
...
     }
 
-    return crypto._baseApis.uploadKeysRequest({
-        one_time_keys: oneTimeJson,
-    }, {
-        // for now, we set the device id explicitly, as we may not be using the
-        // same one as used in login.
-        device_id: crypto._deviceId,
+    return Promise.all(promises).then(() => {
+        return crypto._baseApis.uploadKeysRequest({
```
Kegsay: Given our ability to use async/await now, I would prefer it if we converted this. E.g
```js
await Promise.all(promises);
const res = await crypto._baseApis.uploadKeysRequest({
    one_time_keys: oneTimeJson,
}, {
    // for now, we set the device id explicitly, as we may not be using the
    // same one as used in login.
    device_id: crypto._deviceId,
});
crypto._olmDevice.markKeysAsPublished();
return res;
```

src/crypto/index.js
```diff
@@ -504,7 +512,7 @@ Crypto.prototype.setDeviceVerification = function(userId, deviceId, verified,
  *
  * @return {Object.<string, {deviceIdKey: string, sessions: object[]}>}
```
Kegsay: Lying now.

src/crypto/index.js
```diff
@@ -718,17 +726,20 @@ Crypto.prototype.exportRoomKeys = function() {
...
+ * @return {module:client.Promise} a promise which resolves once the keys have been imported
  */
 Crypto.prototype.importRoomKeys = function(keys) {
-    keys.map((session) => {
-        if (!session.room_id || !session.algorithm) {
-            console.warn("ignoring session entry with missing fields", session);
-            return;
-        }
+    return Promise.map(
+        keys, (session) => {
```
Kegsay: The naming on this makes me grumpy. Mapping over an array of `keys` *obviously* means mapping over a `session`? Say what? This is either `sessions` and `session` or `keys` and `key`, not `keys` and `session`. Please choose one and stick with it throughout.

Kegsay: I'm aware you didn't actually do this in this PR.

src/crypto/index.js
```diff
@@ -1052,79 +1063,117 @@ Crypto.prototype._onRoomKeyRequestEvent = function(event) {
...
-    const requests = this._receivedRoomKeyRequests;
-    this._receivedRoomKeyRequests = [];
-    for (const req of requests) {
-        const userId = req.userId;
-        const deviceId = req.deviceId;
+Crypto.prototype._processReceivedRoomKeyRequests = async function() {
+    if (this._processingRoomKeyRequests) {
+        // we're still processing last time's requests; keep queuing new ones
+        // up for now.
+        return;
```
Kegsay: Where is this queuing happening?

src/crypto/index.js
```diff
@@ -1052,79 +1063,117 @@ Crypto.prototype._onRoomKeyRequestEvent = function(event) {
...
+        const requests = this._receivedRoomKeyRequests;
+        this._receivedRoomKeyRequests = [];
+        const cancellations = this._receivedRoomKeyRequestCancellations;
+        this._receivedRoomKeyRequestCancellations = [];
+
+        await Promise.map(
+            requests, (req) =>
+                this._processReceivedRoomKeyRequest(req),
+        );
+        await Promise.map(
```
Kegsay: So this will do ALL key requests and wait for them prior to doing ANY cancellation requests. Is this what you intended?

src/client.js
richvdh: The logic went that the async-to-generator stuff probably brought some overhead which wasn't really justified here (whereas an `async` function without any `awaits` transpiles to [bluebird.method](http://bluebirdjs.com/docs/api/promise.method.html) which is pretty lightweight).

I can see plenty of counterarguments, though. Happy to use the async flavour if you prefer.

src/crypto/index.js
richvdh: Yes. That makes sure that if we get a request and its cancellation in the same /sync result, then we process the request before the cancellation (and end up with a cancelled request), rather than the cancellation before the request (and end up with an outstanding request which should have been cancelled.)

src/client.js
Kegsay: I would prefer to use the async flavour as I think it is clearer than promises in general. If we profile it and it turns out to be a bottleneck then we can go ahead and change things IMO.

richvdh: sounds sensible. will asyncify.

src/crypto/index.js
Kegsay: This is a beautiful clear explanation, thank you. Please add this as a comment to this code, as it is unclear why we're doing it in this order otherwise.

src/crypto/index.js
richvdh: https://github.com/matrix-org/matrix-js-sdk/blob/89e649f51f7f0e2668cab0e89e9ac9814f890196/src/crypto/index.js#L1053

This whole function is about processing that queue, so I don't think it merits making this comment more explicit.

src/client.js
richvdh: done 1302d6d

src/client.js
richvdh: done 540ae8f

src/crypto/index.js
richvdh: done 540ae8f

src/crypto/index.js
richvdh: done 83d83d1

src/crypto/index.js
richvdh: done 83d83d1

richvdh:
@Kegsay PTAL

src/crypto/index.js
Kegsay: SGTM - sorry must just be me.

Kegsay:
LGTM

PR #511, opened 2017-07-21
Transform `async` functions to bluebird promises

richvdh:
Now that we use transform-runtime instead of regenerator-runtime, we need to
use the async-to-bluebird transform to make sure that `async` functions get
transformed into bluebird promises.


Kegsay:
LGTM

PR #512, opened 2017-07-21
Add more group APIs

dbkr:
Set profile for https://github.com/matrix-org/matrix-react-sdk/pull/1229 and also get users because I'll need it soon

### DIFF
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -413,8 +413,11 @@ MatrixBaseApis.prototype.getGroupSummary = function(groupId) {
 
 /**
  * @param {string} groupId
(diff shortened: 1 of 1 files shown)


src/base-apis.js
```diff
@@ -412,6 +412,30 @@ MatrixBaseApis.prototype.getGroupSummary = function(groupId) {
 };
 
 /**
+ * @param {string} groupId
+ * @param {Object} profile The group profile object, with name, avatar_url,
```
lukebarnard1: Perhaps specify the individual options in the same style as https://github.com/matrix-org/matrix-js-sdk/pull/512/files#diff-7ecbd1742c9fa46478efe534d035eb28R374

PR #513, opened 2017-07-24
Add rawDisplayName to RoomMember

lukebarnard1:
This will at first be the `userId`, but when the members membership event is set, `rawDisplayName` will be assigned to the raw `displayname` of the membership event. This deliberately avoids disambiguation so that clients can disambiguate themselves (via a tooltip or otherwise).

### DIFF
--- a/src/models/room-member.js
+++ b/src/models/room-member.js
@@ -33,7 +33,9 @@ const utils = require("../utils");
  * @prop {string} roomId The room ID for this member.
  * @prop {string} userId The user ID of this member.
  * @prop {boolean} typing True if the room member is currently typing.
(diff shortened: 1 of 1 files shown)


src/models/room-member.js
```diff
@@ -34,6 +34,7 @@ const utils = require("../utils");
  * @prop {string} userId The user ID of this member.
  * @prop {boolean} typing True if the room member is currently typing.
  * @prop {string} name The human-readable name for this room member.
```
richvdh: you might say "disambiguated using the user-id if necessary" here, just to help clarify

lukebarnard1: sgtm

PR #519, opened 2017-08-08
Make methods in crypto/algorithms async

richvdh:


spec/unit/crypto/algorithms/megolm.spec.js
```diff
@@ -119,29 +119,32 @@ describe("MegolmDecryption", function() {
...
+            return megolmDecryption.hasKeysForKeyRequest(
+                keyRequest,
+            ).then((hasKeys) => {
+                expect(hasKeys).toBe(true);
+
+                // set up some pre-conditions for the share call
+                const deviceInfo = {};
+                mockCrypto.getStoredDevice.andReturn(deviceInfo);
+
+                const awaitEnsureSessions = new Promise((res, rej) => {
```
Kegsay: You never return this from the promise chain which is `return`ed to the test framework, so this test isn't waiting for all the stuff inside here to run.

Kegsay: Ignore me, indentation on this diff threw me.

Kegsay:
LGTM

PR #520, opened 2017-08-08
Make some of DeviceList asynchronous

richvdh:
* DeviceList._updateStoredDeviceKeysForUser
 * DeviceList._processQueryResponseForUser
 * DeviceList._storeDeviceKeys

PR #521, opened 2017-08-09
Make bits of `olmlib` asynchronous

richvdh:
(This won't build until #520 lands)

### DIFF
--- a/src/crypto/algorithms/megolm.js
+++ b/src/crypto/algorithms/megolm.js
@@ -288,7 +288,7 @@ MegolmEncryption.prototype._shareKeyWithDevices = function(session, devicesByUse
     return olmlib.ensureOlmSessionsForDevices(
         this._olmDevice, this._baseApis, devicesByUser,
     ).then(function(devicemap) {
-        let haveTargets = false;
+        const promises = [];
 
         for (const userId in devicesByUser) {
             if (!devicesByUser.hasOwnProperty(userId)) {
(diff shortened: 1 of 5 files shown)


src/crypto/algorithms/megolm.js
```diff
@@ -328,31 +328,35 @@ MegolmEncryption.prototype._shareKeyWithDevices = function(session, devicesByUse
...
 
-        if (!haveTargets) {
+        if (promises.length === 0) {
+            // no devices to send to
             return Promise.resolve();
         }
 
-        // TODO: retries
-        return self._baseApis.sendToDevice("m.room.encrypted", contentMap);
+        return Promise.all(promises).then(() => {
```
Kegsay: This is a subtle change. Previously, we just set a flag to say "we have targets plz send". Now, we wait for all messages to be encrypted and only if they succeed do we send. Is this a problem if we successfully encrypt messages for 5/6 devices and then fail to send to the 5 which succeeded?

src/crypto/algorithms/olm.js
```diff
@@ -123,15 +125,17 @@ OlmEncryption.prototype.encryptMessage = function(room, eventType, content) {
...
+                        encryptedContent.ciphertext,
+                        self._userId, self._deviceId, self._olmDevice,
+                        userId, deviceInfo, payloadFields,
+                    ),
                 );
             }
         }
 
-        return encryptedContent;
+        return Promise.all(promises).return(encryptedContent);
```
Kegsay: Same deal here as https://github.com/matrix-org/matrix-js-sdk/pull/521/files#r132168405

Kegsay:
Otherwise LGTM

src/crypto/algorithms/megolm.js
richvdh: I'm failing to understand the problem you're alluding to.

Previously, if `encryptMessageForDevice` failed on the Nth attempt, we would bail out, despite having successfully encrypted (and not sent) messages for the previous N-1 devices.

Now, if encryption attempt N fails, then again, we bail out in the same way. (`encryptMessageForDevice` might succeed for N+1 onwards, but that doesn't matter any more than 0..(N-1).) So I don't see what's changing here.

From the point of view of the Olm protocol: no it doesn't matter if a message we encrypted gets lost or never gets sent: the protocol recovers from that.

src/crypto/algorithms/olm.js
richvdh: same answer :)

src/crypto/algorithms/megolm.js
Kegsay: > Previously, if encryptMessageForDevice failed on the Nth attempt, we would bail out,

This is what I didn't know. I guess we threw an error or something then?

Kegsay:
LGTM

src/crypto/algorithms/megolm.js
richvdh: I still don't understand what you're asking. If `encryptMessageForDevice` threw an exception, it would (of course) make the promise returned from `_shareKeyWithDevices` be rejected.

Ultimately that would bubble up into a "failed to send event" error in the UI.

PR #522, opened 2017-08-09
reduce sendToDevice payload

krombel:
instead of sending one huge request split them up and just do one
request for uploading one-time-keys for 20 devices.

Signed-Off by: Matthias Kesler <krombel@krombel.de>


src/crypto/algorithms/megolm.js
```diff
@@ -352,12 +363,36 @@ MegolmEncryption.prototype._shareKeyWithDevices = function(session, devicesByUse
...
+                return Promise.resolve();
+            }
+            const contentMap = contentMaps[slice];
+            return self._baseApis.sendToDevice("m.room.encrypted", contentMap)
+                .then(() => {
+                    console.log(`Uploaded megolm keys in ${self._roomId} `
+                        + `(slice ${slice}/${maxContentMapId})`);
+
+                    // store that we successfully uploaded the keys of the current slice
+                    for (const userId in contentMap) {
```
krombel: When I use your suggested line and remove the following `if [...] continue` I get the following error on `npm run lint`:
```
The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype  guard-for-in
```

richvdh:
@krombel: I still think that if you're doing a bunch of work on this code, it's a good opportunity to refactor it so that it's clearer and better encapsulated. My suggestion remains:

> I suggest building up a list of lists of (userId, deviceId) pairs for devices with valid Olm sessions as the first step after calling ensureOlmSessionsForDevices. You can then define a separate method which takes one of those lists of (userId, deviceId) pairs and is responsible for calling encryptMessageForDevice and sendToDevice, and setting sharedWithDevices.

ara4n:
@krombel - are you in any position to incorporate @richvdh's review requests?

src/crypto/algorithms/megolm.js
```diff
@@ -262,15 +270,162 @@ MegolmEncryption.prototype._prepareNewSession = async function() {
...
+    session, devicemap, devicesByUser,
+) {
+    const maxToDeviceMessagesPerRequest = 20;
+
+    // use an array where the slices of a content map gets stored
+    const contentMaps = [];
+    let maxContentMapId = 0; // start inserting in the first contentMap
+    let currentToDeviceId = 0;
+
+    const key = this._olmDevice.getOutboundGroupSessionKey(session.sessionId);
```
richvdh: suggest you pass `chain_index` as a param alongside `session` rather than calling this again

src/crypto/algorithms/megolm.js
```diff
@@ -262,15 +270,162 @@ MegolmEncryption.prototype._prepareNewSession = async function() {
...
+                // so just skip it.
+                continue;
+            }
+
+            console.log(
+                "share keys with device " + userId + ":" + deviceId,
+            );
+
+            if (currentToDeviceId > maxToDeviceMessagesPerRequest) {
+                // the first slice is filled up. Start inserting into the next slice
```
richvdh: s/first/current/

src/crypto/algorithms/megolm.js
```diff
@@ -262,15 +270,162 @@ MegolmEncryption.prototype._prepareNewSession = async function() {
...
+ * @return {array<object<userid, deviceInfo>>}
+ */
+MegolmEncryption.prototype._splitUserDeviceMap = function(
+    session, devicemap, devicesByUser,
+) {
+    const maxToDeviceMessagesPerRequest = 20;
+
+    // use an array where the slices of a content map gets stored
+    const contentMaps = [];
+    let maxContentMapId = 0; // start inserting in the first contentMap
```
richvdh: this doesn't seem like a good name. How about `currentSlice`?

src/crypto/algorithms/megolm.js
```diff
@@ -262,15 +270,162 @@ MegolmEncryption.prototype._prepareNewSession = async function() {
...
+    let currentToDeviceId = 0;
```
richvdh: this doesn't seem like a good name. How about  `entriesInCurrentSlice`?

src/crypto/algorithms/megolm.js
```diff
@@ -262,15 +270,162 @@ MegolmEncryption.prototype._prepareNewSession = async function() {
...
+                "share keys with device " + userId + ":" + deviceId,
+            );
+
+            if (currentToDeviceId > maxToDeviceMessagesPerRequest) {
+                // the first slice is filled up. Start inserting into the next slice
+                currentToDeviceId = 0;
+                maxContentMapId++;
+            }
+            if (!contentMaps[maxContentMapId]) {
+                contentMaps[maxContentMapId] = [{
```
richvdh: can you make this `contentMaps[maxContentMapId] = []` and then make the `push` unconditional, to reduce duplication?

src/crypto/algorithms/megolm.js
```diff
@@ -262,15 +270,162 @@ MegolmEncryption.prototype._prepareNewSession = async function() {
...
-    const self = this;
+MegolmEncryption.prototype._encryptAndSendKeysToDevices = function(
+    session, userDeviceMap, payload,
+) {
+    const encryptedContent = {
+        algorithm: olmlib.OLM_ALGORITHM,
+        sender_key: this._olmDevice.deviceCurve25519Key,
+        ciphertext: {},
+    };
+    const key = this._olmDevice.getOutboundGroupSessionKey(session.sessionId);
```
richvdh: again, pass `chain_index` as a param.

src/crypto/algorithms/megolm.js
```diff
@@ -283,104 +438,34 @@ MegolmEncryption.prototype._shareKeyWithDevices = function(session, devicesByUse
...
+    }
 
-        if (promises.length === 0) {
-            // no devices to send to
-            return Promise.resolve();
-        }
+    for (let i = 0; i < userDeviceMaps.length; i++) {
+        await this._encryptAndSendKeysToDevices(
+            session, userDeviceMaps[i], payload,
+        ).then(() => {
```
richvdh: The `then` and `catch` here are unnecessary. I think you want:

```
try:
    await this._encryptAndSendKeysToDevices(...);
    console.log(...);
catch e:
    console.log(...);
    throw e;
```

src/crypto/algorithms/megolm.js
```diff
@@ -283,104 +438,34 @@ MegolmEncryption.prototype._shareKeyWithDevices = function(session, devicesByUse
...
+            console.log(`Uploaded megolm keys in ${this._roomId} `
```
richvdh: "Uploaded" seems the wrong term. "Completed megolm keyshare" was more appropriate.

src/crypto/algorithms/megolm.js
```diff
@@ -262,15 +270,162 @@ MegolmEncryption.prototype._prepareNewSession = async function() {
...
+                payload,
+            ),
+        );
+    }
+
+    return Promise.all(promises).then(() => {
+        return this._baseApis.sendToDevice("m.room.encrypted", contentMap).then(() => {
+            // store that we successfully uploaded the keys of the current slice
+            for (const userId in contentMap) {
+                if (!contentMap.hasOwnProperty(userId)) {
```
richvdh: again, prefer `Object.keys()` to this

src/crypto/algorithms/megolm.js
```diff
@@ -262,15 +270,162 @@ MegolmEncryption.prototype._prepareNewSession = async function() {
...
+                    continue;
+                }
+                for (const deviceId in contentMap[userId]) {
+                    if (!contentMap[userId].hasOwnProperty(deviceId)) {
```
richvdh: ditto

src/crypto/algorithms/megolm.js
```diff
@@ -279,21 +281,17 @@ MegolmEncryption.prototype._prepareNewSession = async function() {
...
-    const key = this._olmDevice.getOutboundGroupSessionKey(session.sessionId);
+    const mapSlices = [];
+    let currentSliceId = 0; // start inserting in the first slice
+    let entriesInCurrentSlice = 0;
 
-    for (const userId in devicesByUser) {
-        if (!devicesByUser.hasOwnProperty(userId)) {
-            continue;
-        }
+    for (let j = 0, userIds = Object.keys(devicesByUser); j < userIds.length; j++) {
```
richvdh: `for (const userId of Object.keys(devicesByUser))`

Sorry, think I said `in` where I meant `of` before

src/crypto/algorithms/megolm.js
```diff
@@ -400,16 +396,12 @@ MegolmEncryption.prototype._encryptAndSendKeysToDevices = function(
...
-            for (const userId in contentMap) {
-                if (!contentMap.hasOwnProperty(userId)) {
-                    continue;
-                }
-                for (const deviceId in contentMap[userId]) {
-                    if (!contentMap[userId].hasOwnProperty(deviceId)) {
-                        continue;
-                    }
+            const userIds = Object.keys(contentMap);
+            for (let i = 0; i < userIds.length; i++) {
```
richvdh: `for (const userId of Object.keys(contentMap))`

src/crypto/algorithms/megolm.js
```diff
@@ -400,16 +396,12 @@ MegolmEncryption.prototype._encryptAndSendKeysToDevices = function(
...
+                const deviceIds = Object.keys(contentMap[userIds[i]]);
+                for (let j = 0; j < deviceIds.length; j++) {
```
richvdh: `for (const deviceId of contentMap[userId])`

PR #523, opened 2017-08-09
Make crypto.decryptMessage return decryption results

richvdh:
... instead of having it call event.setClearData.

The main advantage of this is that it fixes a race condition, wherein apps
could see `event.isDecrypting()` to be true, but in fact the event had been
decrypted (and there was no `Event.decrypted` event on its way).

We're also fixing another race, wherein if the first attempt to decrypt failed,
a call to `attemptDecryption` would race against the first call and a second
attempt to decrypt would never happen.

It also gives a cleaner interface to MatrixEvent, at the expense of making the
`megolm` unit test a bit more hoop-jumpy.

### DIFF
--- a/spec/unit/event.spec.js
+++ b/spec/unit/event.spec.js
@@ -43,6 +43,8 @@ describe("MatrixEvent", () => {
         it('should retry decryption if a retry is queued', () => {
             let callCount = 0;
 
+            let prom2;
+
             const crypto = {
                 decryptEvent: function() {
                     ++callCount;
@@ -50,12 +52,15 @@ describe("MatrixEvent", () => {
                     if (callCount == 1) {
                         // schedule a second decryption attempt while
                         // the first one is still running.
-                        encryptedEvent.attemptDecryption(crypto);
+                        prom2 = encryptedEvent.attemptDecryption(crypto);
 
                         const error = new Error("nope");
                         error.name = 'DecryptionError';
                         return Promise.reject(error);
                     } else {
+                        expect(prom2.isFulfilled()).toBe(
+                            false, 'second attemptDecryption resolved too soon');
+
                         return Promise.resolve({
                             clearEvent: {
                                 type: 'm.room.message',
@@ -68,6 +73,9 @@ describe("MatrixEvent", () => {
             return encryptedEvent.attemptDecryption(crypto).then(() => {
                 expect(callCount).toEqual(2);
                 expect(encryptedEvent.getType()).toEqual('m.room.message');
+
+                // make sure the second attemptDecryption resolves
+                return prom2;
             });
         });
     });
--- a/spec/unit/event.spec.js
+++ b/spec/unit/event.spec.js
@@ -1,0 +1,74 @@
+/*
+Copyright 2017 New Vector Ltd
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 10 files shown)


spec/unit/event.spec.js
```diff
@@ -0,0 +1,74 @@
...
+            let callCount = 0;
+
+            const crypto = {
+                decryptEvent: function() {
+                    ++callCount;
+                    console.log(`decrypt: ${callCount}`);
+                    if (callCount == 1) {
+                        // schedule a second decryption attempt while
+                        // the first one is still running.
+                        encryptedEvent.attemptDecryption(crypto);
```
Kegsay: Can you please also verify that when this `Promise` resolves it has the cleartext.

src/crypto/algorithms/megolm.js
```diff
@@ -535,18 +535,12 @@ utils.inherits(MegolmDecryption, base.DecryptionAlgorithm);
...
  */
-MegolmDecryption.prototype.decryptEvent = function(event) {
-    return this._decryptEvent(event, true);
-};
-
-
-// helper for the real decryptEvent and for _retryDecryption. If
-// requestKeysOnFail is true, we'll send an m.room_key_request when we fail
-// to decrypt the event due to missing megolm keys.
-MegolmDecryption.prototype._decryptEvent = async function(event, requestKeysOnFail) {
```
Kegsay: Why are we removing the `requestKeysOnFail` flag?

src/crypto/index.js
```diff
@@ -1295,6 +1296,25 @@ class IncomingRoomKeyRequestCancellation {
...
 
 /**
+ * The result of a (successful) call to decryptEvent.
+ *
+ * @typedef {Object} EventDecryptionResult
+ *
+ * @property {Object} clearEvent The plaintext payload for the event
+ *     (typically containing <tt>type</tt> and <tt>content</tt> fields).
+ *
+ * @property {string?} senderCurve25519Key Key owned by the sender of this event.
```
Kegsay: You need to start with a `?` not end with one. http://usejsdoc.org/tags-type.html

Kegsay:
Otherwise LGTM, shame I failed to spot this race before :/

src/crypto/algorithms/megolm.js
richvdh: https://github.com/matrix-org/matrix-js-sdk/pull/523/commits/9550bca099b32059f54314f68f887f51dda3c0a9

spec/unit/event.spec.js
richvdh: well, it won't, because we return straight away at https://github.com/matrix-org/matrix-js-sdk/pull/523/files#diff-c2b46f9b1e3be0b9100ad9ef4970a4e3R358. I'm not keen to have to maintain code to keep track of a promise which we can return there when nothing relies on the result of `attemptDecryption` except the tests.

src/crypto/index.js
richvdh: thanks. fixed.

spec/unit/event.spec.js
Kegsay: Having a function which returns different promises like this is incredibly error prone, regardless of its use solely in tests or not. If you don't want to fix the API then at the very least you should be documenting this edge case.

richvdh: So I did this "properly" in 38c9a05.

richvdh:
Hope this is close enough to the thing kegan LGTMed to be landed.

PR #524, opened 2017-08-10
Make lots of OlmDevice asynchronous

richvdh:
A whole series of commits, each making a different method of OlmDevice asynchronous.

The tests in this code suffer from the races fixed in #523, so the tests are failing until that lands.

### DIFF
--- a/spec/integ/matrix-client-crypto.spec.js
+++ b/spec/integ/matrix-client-crypto.spec.js
@@ -314,31 +314,37 @@ function recvMessage(httpBackend, client, sender, message) {
         },
     };
     httpBackend.when("GET", "/sync").respond(200, syncData);
-    const deferred = Promise.defer();
-    const onEvent = function(event) {
-        console.log(client.credentials.userId + " received event",
-                    event);
 
-        // ignore the m.room.member events
-        if (event.getType() == "m.room.member") {
-            return;
-        }
+    const eventPromise = new Promise((resolve, reject) => {
+        const onEvent = function(event) {
+            // ignore the m.room.member events
+            if (event.getType() == "m.room.member") {
+                return;
+            }
+            console.log(client.credentials.userId + " received event",
+                        event);
 
+            client.removeListener("event", onEvent);
+            resolve(event);
+        };
+        client.on("event", onEvent);
+    });
+
+    httpBackend.flush();
+
+    return eventPromise.then((event) => {
+        expect(event.isEncrypted()).toBeTruthy();
+
+        // it may still be being decrypted
+        return testUtils.awaitDecryption(event);
+    }).then((event) => {
         expect(event.getType()).toEqual("m.room.message");
         expect(event.getContent()).toEqual({
             msgtype: "m.text",
             body: "Hello, World",
         });
         expect(event.isEncrypted()).toBeTruthy();
-
-        client.removeListener("event", onEvent);
-        deferred.resolve();
-    };
-
-    client.on("event", onEvent);
-
-    httpBackend.flush();
-    return deferred.promise;
+    });
 }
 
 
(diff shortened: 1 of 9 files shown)


spec/integ/matrix-client-crypto.spec.js
```diff
@@ -314,31 +314,37 @@ function recvMessage(httpBackend, client, sender, message) {
...
+    const eventPromise = new Promise((resolve, reject) => {
+        const onEvent = function(event) {
+            // ignore the m.room.member events
+            if (event.getType() == "m.room.member") {
+                return;
+            }
+            console.log(client.credentials.userId + " received event",
+                        event);
+
+            client.removeListener("event", onEvent);
```
Kegsay: You could've also used `client.once`. https://nodejs.org/api/events.html#events_emitter_once_eventname_listener

spec/unit/crypto/algorithms/megolm.spec.js
```diff
@@ -144,7 +148,7 @@ describe("MegolmDecryption", function() {
                 megolmDecryption.shareKeysWithDevice(keyRequest);
 
                 // it's asynchronous, so we have to wait a bit
-                return awaitEnsureSessions;
+                return awaitEncryptForDevice;
```
Kegsay: Why are we waiting for something different now?

src/crypto/OlmDevice.js
```diff
@@ -112,6 +105,32 @@ function OlmDevice(sessionStore) {
...
+    try {
+        _initialise_account(this._sessionStore, this._pickleKey, account);
+        e2eKeys = JSON.parse(account.identity_keys());
+
+        this._maxOneTimeKeys = account.max_number_of_one_time_keys();
+    } finally {
+        account.free();
+    }
+
+    this.deviceCurve25519Key = e2eKeys.curve25519;
```
Kegsay: This will throw if `_initialise_account` throws or if there are no identity keys as `e2eKeys` will be `undefined`.

src/crypto/OlmDevice.js
```diff
@@ -367,10 +384,10 @@ OlmDevice.prototype.getSessionIdsForDevice = function(theirDeviceIdentityKey) {
  *
  * @param {string} theirDeviceIdentityKey Curve25519 identity key for the
  *     remote device
- * @return {string?}  session id, or null if no established session
+ * @return {Promise<string?>}  session id, or null if no established session
```
Kegsay: `?string`

src/crypto/index.js
```diff
@@ -106,6 +100,20 @@ function Crypto(baseApis, sessionStore, userId, deviceId,
     this._receivedRoomKeyRequestCancellations = [];
     // true if we are currently processing received room key requests
     this._processingRoomKeyRequests = false;
+}
+utils.inherits(Crypto, EventEmitter);
+
+/**
+ * Initialise the crypto module so that it is ready for use
```
Kegsay: You should mention that it returns a promise and that it is only initialised once said promise resolves.

spec/integ/matrix-client-crypto.spec.js
richvdh: No, because then it would be harder to ignore the `m.room.member` events

spec/unit/crypto/algorithms/megolm.spec.js
richvdh: There are two steps here:
1: ensure we have olm sessions for each of the devices (this has always been asynchronous as it involves API calls)
2: encrypt the content for each device (which used to be synchronous, but is now async)

In short, we now need to wait for step 2 to happen, whereas previously waiting for step 1 was sufficient

src/crypto/algorithms/megolm.js
```diff
@@ -706,24 +706,26 @@ MegolmDecryption.prototype.onRoomKeyEvent = function(event) {
...
-
-    // have another go at decrypting events sent with this session.
-    this._retryDecryption(senderKey, sessionId);
 };
 
 /**
  * @inheritdoc
  */
-MegolmDecryption.prototype.hasKeysForKeyRequest = async function(keyRequest) {
+MegolmDecryption.prototype.hasKeysForKeyRequest = function(keyRequest) {
```
richvdh: They all need to be asynchronous. (The actual logic was that I wanted anything that touched sessionStore to be async, because I wanted to back the sessionStore with indexeddb, so sessionStore had to get an async interface).

Of course they can be asynchronous without being `async`. In this case: `hasInboundSessionKeys` is being changed to return a Promise, so the `async` here is redundant. I agree it could have stayed.

src/crypto/index.js
richvdh: done e5565c6

src/crypto/OlmDevice.js
richvdh: done e5565c6

richvdh:
I've had enough of this stuff. I'm taking the executive decision to merge it, in kegan's absence.

PR #528, opened 2017-08-22
Groups: Sync Stream, Accept Invite & Leave

dbkr:
* API wrappers for accepting invites and leaving rooms (reject invite == leave)
 * Parse groups that come down in the sync stream and store them in the JS SDK
 * Make the sync accumulator support groups
 * Add new events for groups
 * Methods to get groups out of the js-sdk store

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -27,6 +27,7 @@ limitations under the License.
 import Promise from 'bluebird';
 const User = require("./models/user");
 const Room = require("./models/room");
+const Group = require('./models/group');
 const utils = require("./utils");
 const Filter = require("./filter");
 const EventTimeline = require("./models/event-timeline");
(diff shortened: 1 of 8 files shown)


src/sync.js
```diff
@@ -1071,6 +1111,35 @@ SyncApi.prototype._pokeKeepAlive = function() {
...
+ * @param {Object} groupsSection Groups section object, eg. response.groups.invite
+ * @param {string} sectionName Which section this is ('invite', 'join' or 'leave')
+ */
+SyncApi.prototype._processGroupSyncEntry = function(groupsSection, sectionName) {
+    // Processes entries from 'groups' section of the sync stream
+    for (const groupId of Object.keys(groupsSection)) {
+        const groupInfo = groupsSection[groupId];
+        let group = this.client.store.getGroup(groupId);
+        const isBrandNew = group === null;
+        if (group === null) {
```
krombel: Why not `if (isBrandNew) {` ?

PR #536, opened 2017-09-12
Only re-emit events from Event objects if needed

dbkr:
The only event an Event emits is 'Event.decrypted', so don't
bother to add listeners if the event isn't encrypted.


ara4n:
lgtm

ara4n:
...other than borked tests

dbkr:
js-sdk tests seem to be broken on develop too :( Not related to this though.

dbkr:
Merging this because test fails are due to jsdoc being broken on Node 8.5: https://github.com/matrix-org/matrix-js-sdk/pull/537 works around this.

PR #537, opened 2017-09-12
Temporarily pin to node 8.4

dbkr:
To fix jsdoc build as commented


dbkr:
Never mind: jsdoc 3.5.5 is out which fixes it

PR #538, opened 2017-09-13
Make re-emitting events much more memory efficient

dbkr:
The previous impl bluntly created a closure for every event type
and source emitter we set up a re-emit for. We can do much better
than this fairly easily by having one bound handler for each event
name and moving it into a class so we have one emitter per target,
since 99% of the time the target is the client object.

https://github.com/matrix-org/matrix-js-sdk/pull/537 to fix Jenkins


ara4n:
cute! lgtm

PR #539, opened 2017-09-14
Add ignore users API

turt2live:
Required for https://github.com/matrix-org/matrix-react-sdk/pull/1389

These are just convenience methods to hide developers from the underlying API.


ara4n:
this looks very plausible; i'd totally forgotten the shape of the API; i'm kinda surprised that it's just magicked into accountData, but apparently that's what happened. lgtm!

PR #543, opened 2017-09-20
Implement API to add users to group summaries

lukebarnard1:


src/base-apis.js
```diff
@@ -478,6 +478,23 @@ MatrixBaseApis.prototype.removeUserFromGroup = function(groupId, userId) {
 
 /**
  * @param {string} groupId
+ * @param {string} userId
+ * @param {string} roleId Optional.
+ * @return {module:client.Promise} Resolves: Empty object
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixBaseApis.prototype.addUserToSummary = function(groupId, userId, roleId) {
```
dbkr: I would go with `addUserToGroupSummary` probably, to make it clear it's specific to groups. otherwise lgtm

PR #544, opened 2017-09-21
Fix a race in decrypting megolm messages

richvdh:
This fixes a race wherein it was possible for us to fail to decrypt a message,
if the keys arrived immediately after our attempt to decrypt it. In that case,
a retry *should* have been scheduled, but was not.

Fixes https://github.com/vector-im/riot-web/issues/5001.

### DIFF
--- a/src/crypto/algorithms/megolm.js
+++ b/src/crypto/algorithms/megolm.js
@@ -652,7 +652,7 @@ MegolmDecryption.prototype._addEventToPendingList = function(event) {
 };
 
 /**
- * Remove an event to the list of those awaiting their session keys.
+ * Remove an event from the list of those awaiting their session keys.
  *
  * @private
  *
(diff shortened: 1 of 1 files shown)


src/crypto/algorithms/megolm.js
```diff
@@ -632,11 +646,32 @@ MegolmDecryption.prototype._addEventToPendingList = function(event) {
...
     const k = content.sender_key + "|" + content.session_id;
     if (!this._pendingEvents[k]) {
-        this._pendingEvents[k] = [];
+        this._pendingEvents[k] = new Set();
+    }
+    this._pendingEvents[k].add(event);
+};
+
+/**
+ * Remove an event to the list of those awaiting their session keys.
```
lukebarnard1: "Remove an event from"?

lukebarnard1:
(I probably could've squashed that better, sorry :grimacing: )

PR #554, opened 2017-10-04
don't try to decrypt a redacted message (fixes vector-im/riot-web#3744)

uhoreg:
set the decrypted data to an empty message

Signed-off-by: Hubert Chathi <hubert@uhoreg.ca>

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -769,6 +769,15 @@ Crypto.prototype.encryptEvent = function(event, room) {
  *  is a problem decrypting the event.
  */
 Crypto.prototype.decryptEvent = function(event) {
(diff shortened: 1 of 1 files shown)


uhoreg:
@richvdh PTAL at this one too

src/crypto/index.js
```diff
@@ -769,6 +769,15 @@ Crypto.prototype.encryptEvent = function(event, room) {
  *  is a problem decrypting the event.
  */
 Crypto.prototype.decryptEvent = function(event) {
+    if (event.isRedacted()) {
+        return {
```
richvdh: This is supposed to return a promise; suggest you wrap it with `Promise.resolve`

PR #556, opened 2017-10-14
Consider cases where the sender may not redact their own event

turt2live:
Just because they sent the event does not mean they'll be able to actually redact it.


ara4n:
I don't like the fact at all that we currently let rooms disable self-redactions with the m.room.redaction event threshold on powerlevels.  But agreed that we might as well be consistent with the behaviour for now.

Separately filed https://github.com/matrix-org/matrix-doc/issues/1071 on the underlying spec issue here.

ara4n:
lgtm - thanks. sorry for delay.

PR #557, opened 2017-10-14
Support set_presence=offline for syncing

turt2live:
see https://github.com/matrix-org/matrix-react-sdk/pull/1482


ara4n:
lgtm! very glad we finally have this in the SDK... it was implemented *years* ago in the server (and I'm astonished it still works there :S)

PR #558, opened 2017-10-16
Fix the glob-to-regex code

dbkr:
It was only making one replacement so would fail for anything with
more than ine glob special char (eg. \*foo would be fine, but \*foo\*
was broken)


lukebarnard1:
Vaguely surprised by `\\\*` (an escaped `\*`) being doubly escaped whilst `\?` is not.

PR #560, opened 2017-10-16
Implement API wrapper for GET /group/$groupId/invited_users

lukebarnard1:


lukebarnard1:
>Sort of surprising this is a separate API call

Basically pagination would be a PITA.

PR #562, opened 2017-10-23
Support room notifs

dbkr:
Specifically, add support for the sender_notification_permission
push rule condition. Also delays calculating notification events
until after they're added to the timeline, as per comment.

NB. there is a bug that causes this to be wrong in some circumstances because the power level for the sentinel member is incorrect, but this is just a general bug in the js-sdk and doesn't need to be part of this PR.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -1277,6 +1277,12 @@ SyncApi.prototype._processRoomEvents = function(room, stateEventList,
     // may make notifications appear which should have the right name.
     room.recalculate(this.client.credentials.userId);
 
+    // execute the timeline events, this will begin to diverge the current state
+    // if the timeline has any state events in it.
+    // This also needs to be done before running push rules on the events as they need
+    // to be decorated with sender etc.
+    room.addLiveEvents(timelineEventList);
+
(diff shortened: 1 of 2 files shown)


src/sync.js
```diff
@@ -1277,6 +1277,12 @@ SyncApi.prototype._processRoomEvents = function(room, stateEventList,
     // may make notifications appear which should have the right name.
     room.recalculate(this.client.credentials.userId);
 
+    // execute the timeline events, this will begin to diverge the current state
+    // if the timeline has any state events in it.
+    // This also needs to be done before running push rules on the events as they need
+    // to be decorated with sender etc.
+    room.addLiveEvents(timelineEventList);
```
lukebarnard1: I understand why this needs to happen before we notify, but I wasn't able to convince myself that moving this wouldn't have bug-free side-effects.

dbkr: I'm sufficiently convinced that it's correct to do in this order and won't have any unintended side effects: if anything we shouldn't be doing anything at all with the events before they're added to a timeline.

PR #563, opened 2017-10-23
Fix power level of sentinel members

dbkr:
When a ppower_levels event arrived, the current member objects were
being updated but the sentinel members were not, meaning that the
power levels of RoomMember objects associated with events had stale
power levels. This was causing power level based push rules to
be wrong.

### DIFF
--- a/src/models/room-state.js
+++ b/src/models/room-state.js
@@ -205,12 +205,7 @@ RoomState.prototype.setStateEvents = function(stateEvents) {
                 const newSentinel = new RoomMember(event.getRoomId(), userId);
                 newSentinel.setMembershipEvent(oldSentinel.events.member, self);
                 newSentinel.setPowerLevelEvent(event);
-                if (
-                    newSentinel.powerLevel != oldSentinel.powerLevel ||
-                    newSentinel.powerLevelNorm != oldSentinel.powerLevelNorm
-                ) {
-                    self._sentinels[userId] = newSentinel;
-                }
+                self._sentinels[userId] = newSentinel;
             }
         }
     });
--- a/src/models/room-state.js
+++ b/src/models/room-state.js
@@ -197,6 +197,21 @@ RoomState.prototype.setStateEvents = function(stateEvents) {
                 member.setPowerLevelEvent(event);
                 self.emit("RoomState.members", event, self, member);
             });
+
+            // Go through the sentinel members and see if any of them would be
+            // affected by the new power levels. If so, replace the sentinel.
+            for (const userId of Object.keys(self._sentinels)) {
+                const oldSentinel = self._sentinels[userId];
+                const newSentinel = new RoomMember(event.getRoomId(), userId);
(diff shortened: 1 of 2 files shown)


dbkr:
This is breaking because the test injects a power_levels with no `users` object which the js-sdk barfs on. It worked before because the actual room members are mocked out, but fails now because the sentinels are not. It feels like we ought to handle power_levels events with no users object sensibly.

src/models/room-state.js
```diff
@@ -197,6 +197,21 @@ RoomState.prototype.setStateEvents = function(stateEvents) {
                 member.setPowerLevelEvent(event);
                 self.emit("RoomState.members", event, self, member);
             });
+
+            // Go through the sentinel members and see if any of them would be
+            // affected by the new power levels. If so, replace the sentinel.
+            for (const userId of Object.keys(self._sentinels)) {
+                const oldSentinel = self._sentinels[userId];
+                const newSentinel = new RoomMember(event.getRoomId(), userId);
+                newSentinel.setMembershipEvent(oldSentinel.events.member, self);
```
richvdh: can we copy the old sentinel in a way that doesn't result in `RoomMember.powerLevel`s being raised for every member of the room?

src/models/room-state.js
```diff
@@ -197,6 +197,21 @@ RoomState.prototype.setStateEvents = function(stateEvents) {
...
+                newSentinel.setPowerLevelEvent(event);
+                if (
```
richvdh: I am unsure if this condition is (a) necessary (b) a good idea.

In particular, I can see a world where a power_level event arrives to change (say) the 'notify' power_level. This doesn't change the individual members' power levels, but an application might compare the sentinel member's power level against the power_level event associated with the sentinel member, and end up with the wrong result.

In short: should we make replacing the sentinel unconditional? (Presumably at the risk of ending up with even more `RoomMember` objects floating about in the heap)

src/models/room-state.js
dbkr: Since we don't emit `RoomState.newMember` here, the event won't be re-emitted by the client, won't go anywhere as no listeners have been added. Seems like this is probably the right semantics.

src/models/room-state.js
dbkr: Yeah, I was basically trying to avoid making 10,000 new sentinels whenever we put a power_levels event in HQ, but yes, it does seem a bit dicey that the sentinels will be associated with a power levels event that isn't the one actually in effect unless it specifically changed that user's power_level.

I guess we do the simpler thing for now and optimise later if needs be.

dbkr:
Updated with a fix for your comment and also a separate commit that fixes the unit tests and the fact that power levels were generally wrong in paginated history. This is in 34428a811c3554043cbe58526cc273e759e70fb4 if i's easier to review separately.

src/models/room-state.js
richvdh: oh, I see. I suppose. Comment might help.

PR #564, opened 2017-10-25
Fix notifEvent computation

dbkr:
We were previously computing notifEvents at the point where we
processed them but before the room they belong to was stored.
This was problematic because some push conditions try to get the
room and therefore failed. Since the push actions are cached, this
spurious calculation also got cached.

This moves the calculation out to a separate function that gets
called only after the room has been stored.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -231,6 +231,8 @@ SyncApi.prototype.syncLeftRooms = function() {
             room.recalculate(client.credentials.userId);
             client.store.storeRoom(room);
             client.emit("Room", room);
+
+            self._processEventForNotifs(room, timelineEvents);
         });
         return rooms;
     });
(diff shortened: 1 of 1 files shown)


src/sync.js
```diff
@@ -1282,9 +1288,20 @@ SyncApi.prototype._processRoomEvents = function(room, stateEventList,
...
+/**
+ * Takes a list of timelineEvents and adds and adds to _notifEvents
+ * as appropriate.
+ * This must be called after the room the events belong to has been stored.
+ *
+ * @param {Room} room
+ * @param {MatrixEvent[]} [timelineEventList] A list of timeline events. Lower index
+ * is earlier in time. Higher index is later.
+ */
+SyncApi.prototype._processEventForNotifs = function(room, timelineEventList) {
```
richvdh: processEvent**s**, surely

PR #568, opened 2017-11-09
[BREAKING] Change the behaviour of the unverfied devices blacklist flag

turt2live:
Previously the global flag was used as a way to completely ignore the per-room option. This commit makes the per-room and global settings be more flexible to allow users to, for example, blacklist unverified devices in all room with the exception of one or two. This is done by making the global setting a device-level default and the per-room option allowing for 3 states: true, false, and unset (use device default).

Signed-off-by: Travis Ralston <travpc@gmail.com>

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -543,11 +543,10 @@ async function _setDeviceVerification(
 
 /**
  * Set the global override for whether the client should ever send encrypted
- * messages to unverified devices.  If false, it can still be overridden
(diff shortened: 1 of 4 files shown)


ara4n:
lgtm, and has been heavily discussed in terms of the breaking change. i would be astonished if it bit anyone in practice, but we'll be sure to yell about it in the release notes when the time comes.

PR #571, opened 2017-11-10
Ignore pushrules in our cached sync response

dbkr:
As hopefully explained in the comment

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -616,7 +616,7 @@ SyncApi.prototype._sync = async function(syncOptions) {
     }
 
     try {
-        await this._processSyncResponse(syncToken, data);
+        await this._processSyncResponse(syncToken, data, isCachedResponse);
     } catch(e) {
         // log the exception with stack if we have it, else fall back
         // to the plain description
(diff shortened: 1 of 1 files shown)


src/sync.js
```diff
@@ -779,7 +782,13 @@ SyncApi.prototype._processSyncResponse = async function(syncToken, data) {
...
         events.forEach(
             function(accountDataEvent) {
-                if (accountDataEvent.getType() == 'm.push_rules') {
+                // XXX: This is awful: ignore push rules from our cached sync. We fetch the
+                // push rules before syncing so we actually have up-to-date ones. We do want
+                // to honour new push rules that come down the sync but synapse doesn't
+                // put new push rules in the sync stream when the base rules change, so
+                // if the base rules change, we do need to refresh. We therefore ignore
+                // the push rules in our cached sync response.
+                if (accountDataEvent.getType() == 'm.push_rules' && !isCachedResponse) {
```
lukebarnard1: My worry here is that the js sdk now has to worry about where the sync data is coming from (obviously).  Would the first incremental sync not update the push rules if they've changed? If so, can we instead ignore `client.pushRules` until the first incremental sync is done?

dbkr: Well, we have a convenient flag for both so either is fine with me really (`isCachedResponse` or `syncOptions.hasSyncedBefore`). The difference is essentially whether we discard the pushrules in an initial /sync. Strictly speaking it's more correct to clobber the pushrules we fetched earlier with the ones from the initial /sync because then if there was a change in the time between us fetching them and us syncing, we'll have the up to date ones.For that reason I'd probably argue for using `isCachedResponse`.

lukebarnard1: IRL we chatted and agreed that this is probably OK for now. But the server really ought to be sending new push rules if the base rules get updated.

PR #572, opened 2017-11-10
Remove unused ICE candidate retrying logic

dbkr:
This was never used because the code which I've also added a
comment to swallows the error that it would otherwise have caught.

We should retry these candidates, but the js-sdk already has
backoff based retrying in scheduler. It only queues / retries
m.room.message events currently, but we should probably use this
for ICE candidates too.

### DIFF
--- a/src/webrtc/call.js
+++ b/src/webrtc/call.js
@@ -1,5 +1,6 @@
 /*
 Copyright 2015, 2016 OpenMarket Ltd
+Copyright 2017 New Vector Ltd
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


src/webrtc/call.js
```diff
@@ -972,6 +973,11 @@ const setState = function(self, state) {
  * @return {Promise}
  */
 const sendEvent = function(self, eventType, content) {
+    // XXX: This makes this promise always succeed (ie. swallows the error)
```
richvdh: This comment is a bit hard to grok. Why is what the candidate-sending call does relevant here?

dbkr: Basically I was trying to explain why I added a catch to the send-candidate call even though it'll never get hit and that it's sort-of OK that this error gets swallowed. Any suggestions on how I can word this better?

richvdh: well, I suggest noting the fact that the send-candidate's `catch` is just for hygiene in the send-candidate code, and just losing the "but only one ... " phrase here.

richvdh: No :-p

"sending one sendEvent call"?

dbkr:
oh actually, I've just realised this broke relatively recently and I've just removed all the things that listen for this event - let's just stop it swallowing the error

PR #575, opened 2017-11-16
Supporting fixes For making UnknownDeviceDialog not pop up automatically

dbkr:
This is mostly a lot of VoIP fixes to make the flow of trying to place or answer a call when there are unknown devices in the room. We don't want to stack up VoIP events in the pending events and then fire them all out later: we should just abort the call and place a new one once the user has reviewed the new devices.

Reviewing this commit-by-commit may make more sense: the commits are fairly well contained, functionality-wise, but all in the same commit because they'd otherwise leave things a bit broken.

### DIFF
--- a/src/webrtc/call.js
+++ b/src/webrtc/call.js
@@ -26,6 +26,22 @@ const DEBUG = true;  // set true to enable console logging.
 // events: hangup, error(err), replaced(call), state(state, oldState)
 
 /**
+ * Fires whenever an error occurs when call.js encounters an issue with setting up the call.
+ * <p>
+ * The error given will have a code equal to either `MatrixCall.ERR_LOCAL_OFFER_FAILED` or
+ * `MatrixCall.ERR_NO_USER_MEDIA`. `ERR_LOCAL_OFFER_FAILED` is emitted when the local client
+ * fails to create an offer. `ERR_NO_USER_MEDIA` is emitted when the user has denied access
+ * to their audio/video hardware.
+ *
+ * @event module:webrtc/call~MatrixCall#"error"
+ * @param {Error} err The error raised by MatrixCall.
+ * @example
+ * matrixCall.on("error", function(err){
+ *   console.error(err.code, err);
+ * });
+ */
+
+/**
  * Construct a new Matrix Call.
  * @constructor
  * @param {Object} opts Config options.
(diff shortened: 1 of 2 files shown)


src/webrtc/call.js
```diff
@@ -593,6 +584,27 @@ MatrixCall.prototype._maybeGotUserMediaForInvite = function(stream) {
...
+    sendEvent(this, 'm.call.answer', this._answerContent).then(() => {
+        setState(this, 'connecting');
+        // If this isn't the first time we've tried to send the answer,
+        // we may have candidates queued up, so send them now.
+        _sendCandidateQueue(this);
+    }).catch((error) => {
+        // We've failed to answer: back to the ringing state
+        setState(this, 'ringing');
+        this.client.cancelPendingEvent(error.event);
+        this.emit(
```
ara4n: surely we should only emit this if the error is UnkDevices?

ara4n: (it's a bit unintuitive also that the fact there are unknown devices in the room gets in the way of call signalling, which the user might reasonably expect to be 1:1. it's almost an argument that we should be using toDevice messaging for the 1:1 bits of VoIP Signalling...)

src/webrtc/call.js
```diff
@@ -753,14 +767,26 @@ MatrixCall.prototype._gotLocalOffer = function(description) {
...
+            setState(self, 'invite_sent');
+            setTimeout(function() {
+                if (self.state == 'invite_sent') {
+                    self.hangup('invite_timeout');
+                }
+            }, MatrixCall.CALL_TIMEOUT_MS);
+        }).catch((error) => {
+            self.client.cancelPendingEvent(error.event);
+            terminate(self, "local", "unknown_devices", false);
+            self.emit(
```
ara4n: again, surely we should check the type of error before assuming it's an Unknown Devices?

ara4n:
looks plausible mod comments

src/webrtc/call.js
dbkr: oops, thanks!

dbkr: and yeah, although in the case of an invite you'll still want to send it to all devices. For answers it is a bit silly since it only needs to go to the device the invite came from.

src/webrtc/call.js
dbkr: fixed here too

PR #577, opened 2017-11-17
Fix the force TURN option

dbkr:
Option needed to be passed in when creating a webrtc call, but for
incoming calls the js-sdk creates the call itself, so the app never
gets a chance to set the option.

### DIFF
--- a/src/webrtc/call.js
+++ b/src/webrtc/call.js
@@ -1300,7 +1300,7 @@ module.exports.setVideoInput = function(deviceId) { videoInput = deviceId; };
  * @param {boolean} options.forceTURN whether relay through TURN should be forced.
  * @return {MatrixCall} the call or null if the browser doesn't support calling.
  */
-module.exports.createNewMatrixCall = function(client, roomId, options) {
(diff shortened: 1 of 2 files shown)


src/webrtc/call.js
```diff
@@ -1350,14 +1350,17 @@ module.exports.createNewMatrixCall = function(client, roomId, options) {
...
+
     const opts = {
         webRtc: webRtc,
         client: client,
         URL: w.URL,
         roomId: roomId,
         turnServers: client.getTurnServers(),
         // call level options
-        forceTURN: options ? options.forceTURN : false,
+        forceTURN: client._forceTURN || optionsForceTURN,
```
lukebarnard1: This feels more than deprecation because the functionality has changed.

dbkr: Functionality from this point should be the same? ie. if you pass true here, it'll only use relay candidates

lukebarnard1: Oh hang on.. client._forceTURN didn't exist before so it's fine. My bad!

PR #579, opened 2017-11-21
Move Olm account to IndexedDB

dbkr:
Migrates the Olm account storage from the session store to the crypto store, backed by indexedDB. This allows us to lock access to it in a transaction, so multiple tabs can't overwrite each other's data.

Also adds a localstorage-backed crypto store to prevent people using Firefox in ones of its modes where IndexedDB support is broken from losing all their e2e data. It still uses the in-memory store for key requests as before, but saves the Olm account to localstorage. We may want to kick the app to warn the user that they're running in a degraded mode?

Olm accounts in the old session store are migrated into the crypto store seamlessly.

### DIFF
--- a/src/crypto/OlmDevice.js
+++ b/src/crypto/OlmDevice.js
@@ -198,7 +198,7 @@ OlmDevice.prototype._getAccount = async function(func) {
                 return save(pickledAccount);
             });
         } finally {
-            if (account !== null) account.free();
+            account.free();
         }
     });
     return result;
--- a/src/crypto/OlmDevice.js
+++ b/src/crypto/OlmDevice.js
@@ -139,15 +139,15 @@ OlmDevice.prototype.init = async function() {
 
 
 async function _initialise_account(cryptoStore, pickleKey, account) {
-    const accountTxn = await cryptoStore.endToEndAccountTransaction();
-    if (accountTxn.account !== null) {
-        account.unpickle(pickleKey, accountTxn.account);
-        return;
-    }
-
-    account.create();
-    const pickled = account.pickle(pickleKey);
-    await accountTxn.save(pickled);
+    await cryptoStore.endToEndAccountTransaction((accountData, save) => {
+        if (accountData !== null) {
+            account.unpickle(pickleKey, accountData);
+        } else {
+            account.create();
+            const pickled = account.pickle(pickleKey);
+            save(pickled);
+        }
+    });
 }
 
 /**
@@ -171,23 +171,22 @@ OlmDevice.getOlmVersion = function() {
 OlmDevice.prototype._getAccount = async function(func) {
     let result;
 
-    let account = null;
-    try {
-        const accountTxn = await this._cryptoStore.endToEndAccountTransaction();
+    await this._cryptoStore.endToEndAccountTransaction((accountData, save) => {
         // Olm has a limited stack size so we must tightly control the number of
         // Olm account objects in existence at any given time: once created, it
         // must be destroyed again before we await.
-        account = new Olm.Account();
-        account.unpickle(this._pickleKey, accountTxn.account);
+        const account = new Olm.Account();
+        try {
+            account.unpickle(this._pickleKey, accountData);
 
-        result = func(account, () => {
+            result = func(account, () => {
                 const pickledAccount = account.pickle(this._pickleKey);
-                return accountTxn.save(pickledAccount);
-            }
-        );
-    } finally {
-        if (account !== null) account.free();
-    }
+                return save(pickledAccount);
+            });
+        } finally {
+            if (account !== null) account.free();
+        }
+    });
     return result;
 };
 
--- a/src/crypto/OlmDevice.js
+++ b/src/crypto/OlmDevice.js
@@ -208,18 +208,6 @@ OlmDevice.prototype._getAccount = async function(func) {
 
 
 /**
- * store our OlmAccount in the session store
- *
- * @param {OlmAccount} account
- * @private
- */
-OlmDevice.prototype._saveAccount = async function(account) {
-    const pickledAccount = account.pickle(this._pickleKey);
-    await this._cryptoStore.storeEndToEndAccount(pickledAccount);
-};
-
-
-/**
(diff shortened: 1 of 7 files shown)


src/crypto/OlmDevice.js
```diff
@@ -158,21 +159,36 @@ OlmDevice.getOlmVersion = function() {
...
+OlmDevice.prototype._getAccount = async function(func) {
+    let result;
+
+    let account = null;
     try {
-        const pickledAccount = this._sessionStore.getEndToEndAccount();
-        account.unpickle(this._pickleKey, pickledAccount);
-        return func(account);
+        const accountTxn = await this._cryptoStore.endToEndAccountTransaction();
+        // Olm has a limited stack size so we must tightly control the number of
```
richvdh: s/stack/heap/

src/crypto/OlmDevice.js
```diff
@@ -137,16 +141,29 @@ OlmDevice.prototype.init = async function() {
...
 
-function _initialise_account(sessionStore, pickleKey, account) {
-    const e2eAccount = sessionStore.getEndToEndAccount();
-    if (e2eAccount !== null) {
-        account.unpickle(pickleKey, e2eAccount);
-        return;
-    }
+async function _initialise_account(sessionStore, cryptoStore, pickleKey, account) {
+    let removeFromSessionStore = false;
+    await cryptoStore.endToEndAccountTransaction((accountData, save) => {
```
richvdh: can we s/accountData/pickledAccount/ or somesuch? I'm finding myself a bit confused.

src/crypto/OlmDevice.js
```diff
@@ -182,9 +213,9 @@ OlmDevice.prototype._getAccount = function(func) {
  * @param {OlmAccount} account
  * @private
  */
-OlmDevice.prototype._saveAccount = function(account) {
+OlmDevice.prototype._saveAccount = async function(account) {
```
richvdh: this is presumably no longer used?

src/crypto/OlmDevice.js
```diff
@@ -158,21 +175,35 @@ OlmDevice.getOlmVersion = function() {
 
 
 /**
- * extract our OlmAccount from the session store and call the given function
+ * extract our OlmAccount from the crypto store and call the given function
+ * with the account object and a 'save' function which returns a promise.
+ * The function will not be awaited upon and the save function must be
```
richvdh: this is maybe a bit misleading. The function *will* be awaited upon, by virtue of any promise it returns being returned by this `async` function - so the promise returned by `func` must resolve before the promise returned by `_getAccount` will resolve.

It would probably be clearer to say "The `account` will be freed as soon as `func` returns - even if `func` returns a promise".

src/crypto/store/indexeddb-crypto-store-backend.js
```diff
@@ -257,6 +257,38 @@ export class Backend {
         };
         return promiseifyTxn(txn);
     }
+
+    /**
+     * Load the end to end account for the logged-in user. Once the account
+     * is retrieved, the given function is executed and passed the base64
+     * encoded account string and a method for saving the account string
```
richvdh: s/base64 encoded/pickled/? And s/account string/pickle/?

Might be more consistent. just a thought.

src/crypto/store/indexeddb-crypto-store-backend.js
```diff
@@ -257,6 +257,38 @@ export class Backend {
...
+     * back to the database. This allows the account to be read and writen
+     * atomically.
+     * @param {func} func Function called with the account data and a save function
```
richvdh:
```
@param {function(string, function())} func ....
```
would be the correct annotation here I think

src/crypto/store/indexeddb-crypto-store-backend.js
```diff
@@ -257,6 +257,38 @@ export class Backend {
...
+     * @return {Promise} Resolves with the return value of the function once
```
richvdh: s/the function/func/, possibly

src/crypto/store/indexeddb-crypto-store-backend.js
```diff
@@ -257,6 +257,38 @@ export class Backend {
...
+     *     the transaction is complete (ie. once data is written back if the
+     *     save function is called.
```
richvdh: dangling (

src/crypto/OlmDevice.js
dbkr: yeah, agreed. done.

src/crypto/OlmDevice.js
dbkr: good point, it is not.

src/crypto/store/indexeddb-crypto-store-backend.js
dbkr: yeah, wfm

src/crypto/store/indexeddb-crypto-store-backend.js
dbkr: done

src/crypto/store/indexeddb-crypto-store-backend.js
dbkr: done

src/crypto/store/indexeddb-crypto-store-backend.js
dbkr: done

src/crypto/store/indexeddb-crypto-store-backend.js
```diff
@@ -260,14 +260,15 @@ export class Backend {
...
-     * is retrieved, the given function is executed and passed the base64
-     * encoded account string and a method for saving the account string
+     * is retrieved, the given function is executed and passed the pickled
+     * account string and a method for saving the pickle
      * back to the database. This allows the account to be read and writen
      * atomically.
-     * @param {func} func Function called with the account data and a save function
-     * @return {Promise} Resolves with the return value of the function once
+     * @param {function(string, function())} func Function called with the
+     *     account data and a save function
```
richvdh: pickled account

PR #582, opened 2017-11-29
Change crypto store transaction API

dbkr:
To allow multiple things to be fetched/stored in a single
transaction.

Currently it is still just the account that's actually in
indexeddb though.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "build": "babel -s -d lib src && rimraf dist && mkdir dist && browserify -d browser-index.js | exorcist dist/browser-matrix.js.map > dist/browser-matrix.js && uglifyjs -c -m -o dist/browser-matrix.min.js --source-map dist/browser-matrix.min.js.map --in-source-map dist/browser-matrix.js.map dist/browser-matrix.js",
     "dist": "npm run build",
     "watch": "watchify -d browser-index.js -o 'exorcist dist/browser-matrix.js.map > dist/browser-matrix.js' -v",
-    "lint": "eslint --max-warnings 109 src spec",
+    "lint": "eslint --max-warnings 102 src spec",
     "prepublish": "npm run clean && npm run build && git rev-parse HEAD > git-revision.txt"
   },
   "repository": {
--- a/src/crypto/OlmDevice.js
+++ b/src/crypto/OlmDevice.js
@@ -14,7 +14,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-"use strict";
+
+import IndexedDBCryptoStore from './store/indexeddb-crypto-store';
 
 /**
  * olm.js wrapper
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -15,7 +15,7 @@
     "build": "babel -s -d lib src && rimraf dist && mkdir dist && browserify -d browser-index.js | exorcist dist/browser-matrix.js.map > dist/browser-matrix.js && uglifyjs -c -m -o dist/browser-matrix.min.js --source-map dist/browser-matrix.min.js.map --in-source-map dist/browser-matrix.js.map dist/browser-matrix.js",
     "dist": "npm run build",
     "watch": "watchify -d browser-index.js -o 'exorcist dist/browser-matrix.js.map > dist/browser-matrix.js' -v",
-    "lint": "eslint --max-warnings 109 src spec",
+    "lint": "eslint --max-warnings 102 src spec",
```
richvdh: :tada:

src/crypto/OlmDevice.js
```diff
@@ -177,36 +181,30 @@ OlmDevice.getOlmVersion = function() {
...
  * extract our OlmAccount from the crypto store and call the given function
- * with the account object and a 'save' function which returns a promise.
- * The `account` will be freed as soon as `func` returns - even if func returns
- * a promise
+ * with the account object
+ * The `account` object is useable only within the callback passed to this
+ * function and will be freed as soon the callback returns. It is *not*
+ * useable for the rest of the lifetime of the transaction.
  *
+ * @param {*} txn
```
richvdh: could you document this?

src/crypto/OlmDevice.js
```diff
@@ -177,36 +181,30 @@ OlmDevice.getOlmVersion = function() {
...
+    this._cryptoStore.getAccount(txn, (pickledAccount) => {
         const account = new Olm.Account();
         try {
             account.unpickle(this._pickleKey, pickledAccount);
-
-            result = func(account, () => {
-                const pickledAccount = account.pickle(this._pickleKey);
-                save(pickledAccount);
-            });
+            func(account);
```
richvdh: should we not return the result here and at [line 194](https://github.com/matrix-org/matrix-js-sdk/pull/582/commits/f11a58e2cb2b98dff8f77e3031db18dc7d0bb359#diff-d0d27d71268ce9cfb61c9b3bc338fd79R194)?

src/crypto/OlmDevice.js
```diff
@@ -177,36 +181,30 @@ OlmDevice.getOlmVersion = function() {
...
         } finally {
             account.free();
         }
     });
-    return result;
 };
 
+OlmDevice.prototype._storeAccount = function(txn, account) {
```
richvdh: some docs please

src/crypto/OlmDevice.js
```diff
@@ -270,9 +268,13 @@ OlmDevice.prototype._getUtility = function(func) {
  * @return {Promise<string>} base64-encoded signature
  */
 OlmDevice.prototype.sign = async function(message) {
-    return await this._getAccount(function(account) {
-        return account.sign(message);
+    let result;
```
richvdh: again, can't we pass the result up through the callbacks to save us having to fiddle with a local result?

(perhaps it actually ends up being more confusing?)

dbkr: we can, but it could now be quite a few callbacks, and you could be doing more than one get in the transaction, at which point it's not clear which one the transaction promise would return, so I think I'd vote for just making you assign it to a captured variable.

src/crypto/OlmDevice.js
dbkr: Possibly, but as per other comment it could get quite hairy, so probably simpler to just never do it at all.

src/crypto/OlmDevice.js
dbkr: done

src/crypto/OlmDevice.js
dbkr: done

src/crypto/OlmDevice.js
richvdh: the transaction promise would return whatever is returned by the `func` passed to `doTxn`.

I don't think I agree with you, but I don't feel super strongly about it.

PR #589, opened 2017-12-31
Add getThirdpartyUser to base api

johbo:
I investigated options to build a new bridge for XMPP and found that this API method would be very useful for it. My assumption is that it could be used to enhance the invite dialog in a way so that it also understands protocols and external IDs, similar to the way how the room directory currently works.

Signed-off-by: Johannes Bornhold <johannes@bornhold.name>

### DIFF
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -1668,7 +1668,7 @@ MatrixBaseApis.prototype.getThirdpartyProtocols = function() {
  * Get information on how a specific place on a third party protocol
  * may be reached.
  * @param {string} protocol The protocol given in getThirdpartyProtocols()
- * @param {object} params Protocol-specific parameters, as given in th
+ * @param {object} params Protocol-specific parameters, as given in the
  *                        response to getThirdpartyProtocols()
  * @return {module:client.Promise} Resolves to the result object
  */
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -1684,6 +1684,25 @@ MatrixBaseApis.prototype.getThirdpartyLocation = function(protocol, params) {
 };
 
 /**
+ * Get information on how a specific user on a third party protocol
+ * may be reached.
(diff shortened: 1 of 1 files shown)


src/base-apis.js
```diff
@@ -1683,6 +1683,25 @@ MatrixBaseApis.prototype.getThirdpartyLocation = function(protocol, params) {
...
+ * @return {module:client.Promise} Resolves to the result object
+ */
+MatrixBaseApis.prototype.getThirdpartyUser = function(protocol, params) {
+    const path = utils.encodeUri("/thirdparty/user/$protocol", {
+        $protocol: protocol,
+    });
+
+    return this._http.authedRequestWithPrefix(
+        undefined, "GET", path, params, undefined,
+        httpApi.PREFIX_UNSTABLE,
```
dbkr: This is in the v2 API now so if we're adding it, we should probably add it using the stable endpoint.

johbo: Should the same be tweaked for `getThirdpartyProtocols` ? I assume they go together.

dbkr: Good question - since these are existing APIs, we should probably leave them using `unstable` for a while as not all servers will have updated to expose the APIs on the stable path yet.

dbkr:
Also please note you need to do the DCO signoff as per https://github.com/matrix-org/matrix-js-sdk/blob/master/CONTRIBUTING.rst

johbo:
Had the sign off already in the commit. I've added it into the PR comment as well. Does it work this way?

dbkr:
oh, I missed that, either is fine (although if it's on the commits it should technically be on all of them)

Half-Shot:
out of interest, what's holding this up?

jryans:
This appears ready to merge to me. It has been reviewed and comments were replied to. We have sign off on 1 of 2 commits and the PR as a whole, which seems like sufficient coverage.

PR #590, opened 2018-01-04
Add event type for stickers [WIP]

rxl881:


ara4n:
lgtm - sorry for the spectacular fuckup there :( serves me right for blindly porting from python and not writing any UTs for it - sorry :/

rxl881:
No worries at all. Glad to have got to the bottom of it ... I was staring at that code for a little while trying to understand how exactly it was intended to work :)

Half-Shot:
Thanks for taking the time with getting stickers going, it's been a long time coming. With that being said, I've got several issues with this.

I can't see how this differs from an image event, other than the metadata of it being a sticker which says...that I am meant to render it slightly smaller. There is also zero information on what ``info`` is meant to be. I'm assuming it's a ``ImageInfo`` type, which brings me back to point one.

Anyway, besides that there is no spec for this. Nothing merged, nothing pending. No explanation of what a sticker even is in this context. If it's my assumption, a telegram sticker then it is missing a LOT of information like which pack it is part of and which emoji it might correspond to.

It could also mean **another** type of sticker altogether that I am not aware of, in which case it is **vital** that there is a spec explaining.
This may be WIP (and I'd even go as far as to argue a WIP implementation should not be merged as a feature), but it certainly should already be clearly distinct from existing types from the get go.

ara4n:
i think all these concerns can be allayed with a spec proposal googledoc, which we'll rustle up asap. the m.room.sticker event is very much experimental atm (and probably should be im.vector.sticker or something whilst we mess with it, although that'd probably cause hysteria that we were trying to make stickers riot-specific or something daft).

To quickly answer the questions: using a differrent event type is a deliberate move towards FINALLY solving the [human representation of arbitrary events "problem"](https://docs.google.com/document/d/1cLSlMZ0tulRJM4ki-ImQ_z6DSySQgTRafO2m13RtgZo).

The difference with m.room.message-with-msgtype-image is that clients are meant to display it inline without any lightboxing or download links or metadata info like you might do for an image.  It also spells out the necessary semantics.

In terms of whether we link back to the source stickerpack or give it an emoji equivalent - that sounds like very useful stuff; suggestions welcome :) Let's go graffiti on a spec proposal as soon as there's one written up out of the convos which produced the current experiment.

PR #591, opened 2018-01-05
Optimise pushprocessor

lukebarnard1:
by not [re]creating RegExps unnecessarily.


lukebarnard1:
I am making quite a large assumption about what `rule_id` is, I may have intuited it totally incorrectly but I couldn't imagine what else it could be apart from a plain user_id or room_id according to the rule type.

lukebarnard1:
>Alternatively, put a cache in globToRegexp.

Sounds like a much better, simpler way to go about it.

PR #592, opened 2018-01-09
Set event error before emitting

dbkr:
So the event object properties are in a consistent state when we
the event is emitted.

Fixes first part of https://github.com/vector-im/riot-web/issues/5936

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -1,6 +1,6 @@
 /*
 Copyright 2015, 2016 OpenMarket Ltd
-Copyright 2017 Vector Creations Ltd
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -1,6 +1,6 @@
 /*
 Copyright 2015, 2016 OpenMarket Ltd
-Copyright 2017 Vector Creations Ltd
+Copyright 2017, 2018 Vector Creations Ltd
```
richvdh: <cough> new vector

richvdh:
(pending the tests passing)

PR #596, opened 2018-01-19
arguments true and false are actually invalid

t3chguy:
Signed-off-by: Michael Telatynski 7t3chguy@gmail.com

See https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/models/event-timeline.js#L165-L167

### DIFF
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -94,10 +94,10 @@ function synthesizeReceipt(userId, event, receiptType) {
  * event type and the values are the events.
  * @prop {RoomState} oldState The state of the room at the time of the oldest
  * event in the live timeline. Present for backwards compatibility -
- * prefer getLiveTimeline().getState("b").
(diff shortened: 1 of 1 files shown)


krombel:
Why not use `EventTimeline.BACKWARDS` and `EventTimeline.FORWARDS` as defined [here](https://github.com/matrix-org/matrix-js-sdk/blob/f5ceaffc5c/src/models/event-timeline.js#L69-L75)?
EventTimeline is actually imported so people should be able to refer to that easily.

t3chguy:
Sure, but then to specify it as an example I'd have to also mention that you have to import it, and EventTimeline is not something applications tend to explicitly import

ara4n:
we really should be using the constants here. the API is done in terms of constants, not the string literals which happen to be under the hood...

PR #597, opened 2018-01-22
Migrate room encryption store to crypto store

dbkr:
Final cryptostore migration PR!

Based on https://github.com/matrix-org/matrix-js-sdk/pull/594 so probably want to get that merge before trying to review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "bluebird": "^3.5.0",
     "browser-request": "^0.3.3",
     "content-type": "^1.0.2",
+    "olm": "https://matrix.org/packages/npm/olm/olm-2.2.2.tgz",
     "request": "^2.53.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -54,7 +54,6 @@
     "bluebird": "^3.5.0",
     "browser-request": "^0.3.3",
     "content-type": "^1.0.2",
-    "olm": "https://matrix.org/packages/npm/olm/olm-2.2.2.tgz",
     "request": "^2.53.0"
   },
   "devDependencies": {
--- a/src/client.js
+++ b/src/client.js
@@ -42,6 +42,7 @@ const MatrixBaseApis = require("./base-apis");
 const MatrixError = httpApi.MatrixError;
 
 import ReEmitter from './ReEmitter';
+import RoomList from './crypto/RoomList';
 
 const SCROLLBACK_DELAY_MS = 3000;
 let CRYPTO_ENABLED = false;
@@ -181,6 +182,11 @@ function MatrixClient(opts) {
     if (CRYPTO_ENABLED) {
         this.olmVersion = Crypto.getOlmVersion();
     }
+
+    // List of what rooms have encryption enabled: separate from crypto because
+    // we still want to know what rooms are encrypted even if crypto is disabled:
+    // we don't want to start sening unencrypted events to them.
+    this._roomList = new RoomList(this._cryptoStore);
 }
 utils.inherits(MatrixClient, EventEmitter);
 utils.extend(MatrixClient.prototype, MatrixBaseApis.prototype);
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -54,6 +54,7 @@
     "bluebird": "^3.5.0",
     "browser-request": "^0.3.3",
     "content-type": "^1.0.2",
+    "olm": "https://matrix.org/packages/npm/olm/olm-2.2.2.tgz",
```
richvdh: olm was deliberately omitted here for some reason involving making it possible to do without it. the details currently escape me, but what is the reason for adding it?

src/client.js
```diff
@@ -181,6 +182,11 @@ function MatrixClient(opts) {
     if (CRYPTO_ENABLED) {
         this.olmVersion = Crypto.getOlmVersion();
     }
+
+    // List of what rooms have encryption enabled: separate from crypto because
```
richvdh: which rooms

src/client.js
```diff
@@ -181,6 +182,11 @@ function MatrixClient(opts) {
...
+    // we still want to know what rooms are encrypted even if crypto is disabled:
```
richvdh: which

src/client.js
```diff
@@ -181,6 +182,11 @@ function MatrixClient(opts) {
...
+    // we don't want to start sening unencrypted events to them.
```
richvdh: sening

src/crypto/RoomList.js
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * @alias module:crypto/RoomList
+ */
+export default class RoomList {
+    constructor(cryptoStore, sessionStore) {
+        this._cryptoStore = cryptoStore;
+        this._sessionStore = sessionStore;
+
+        // Object of roomId -> room e2e info object
```
richvdh: could we document the shape of said object hereabouts?

src/crypto/index.js
```diff
@@ -601,21 +615,19 @@ Crypto.prototype.getEventSenderDeviceInfo = function(event) {
...
     // if we already have encryption in this room, we should ignore this event
     // (for now at least. maybe we should alert the user somehow?)
-    const existingConfig = this._sessionStore.getEndToEndRoom(roomId);
-    if (existingConfig) {
-        if (JSON.stringify(existingConfig) != JSON.stringify(config)) {
-            console.error("Ignoring m.room.encryption event which requests " +
-                          "a change of config in " + roomId);
-            return;
-        }
+    const existingConfig = await this.getRoomEncryption(roomId);
```
richvdh: given this seems to be the only place `getRoomEncryption` is used, any reason not to inline it?

src/crypto/index.js
```diff
@@ -587,6 +590,17 @@ Crypto.prototype.getEventSenderDeviceInfo = function(event) {
     return device;
 };
 
+/**
+ * Get the current end-to-end encryption config for a room
+ *
+ * @param {string} roomId The room ID to query
+ *
+ * @return {object} The current end-to-end encyption status, or null if
```
richvdh: I don't really understand why you've made this `async`, but given you have, the return type is a promise

package.json
dbkr: bah, because npm now adds any package you install to package.json even if you don't specify --save.

src/client.js
dbkr: fixed :)

src/client.js
dbkr: also fixed

src/client.js
dbkr: also done

src/crypto/RoomList.js
dbkr: have added that it's the body of the m.room.encryption object

src/crypto/index.js
dbkr: ah ok. I think the reason it's async was to match the rest of the crypto stuff which assumes it's going to be indexeddb backed. I'd also assumed that the Promise return type was implicit on an async function.

src/crypto/index.js
dbkr: mm, point

PR #598, opened 2018-01-25
Don't synthesize RR for m.room.redaction as causes the RR to go missing.

t3chguy:
Fixes https://github.com/vector-im/riot-web/issues/3511

Signed-off-by: Michael Telatynski 7t3chguy@gmail.com


ara4n:
lgtm!

PR #601, opened 2018-01-29
Allow a mode of operation without HS connection

lukebarnard1:
Instead of blocking the first sync on getting the push rules and
filter, load sync data from disc.

When the client comes online, the push rules will be acquired and
a sync cycle started. This could be immediate if the client is
already online.

This could be a breaking change for clients that get push rules
before a successful sync has been done.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -510,6 +510,9 @@ SyncApi.prototype.retryImmediately = function() {
  * @param {Object} syncOptions
  * @param {string} syncOptions.filterId
  * @param {boolean} syncOptions.hasSyncedBefore
+ * @param {Object} syncOptions.savedSync a saved sync that was persisted by a
+ *                                       store. Should have been acquired via
+ *                                       client.store.getSavedSync().
  */
 SyncApi.prototype._sync = async function(syncOptions) {
     const client = this.client;
--- a/src/sync.js
+++ b/src/sync.js
@@ -462,8 +462,8 @@ SyncApi.prototype.sync = function() {
         // no push rules for guests, no access to POST filter for guests.
         self._sync({});
     } else {
-        // Don't do an HTTP hit to /sync. Instead, load up the persisted /sync data,
-        // if there is data there.
+        // Before fetching push rules, fetching the filter and syncing, check
+        // for persisted /sync data and use that if present.
         client.store.getSavedSync().then((savedSync) => {
             if (savedSync) {
                 self._syncFromCache(savedSync);
@@ -676,9 +676,7 @@ SyncApi.prototype._sync = async function(syncOptions) {
     // keep emitting SYNCING -> SYNCING for clients who want to do bulk updates
     this._updateSyncState("SYNCING", syncEventData);
 
-    // tell databases that everything is now in a consistent state and can be
-    // saved (no point doing so if we only have the data we just got out of the
-    // store).
+    // tell databases that everything is now in a consistent state and can be saved.
     client.store.save();
 
     // Begin next sync
--- a/src/sync.js
+++ b/src/sync.js
@@ -504,7 +504,11 @@ SyncApi.prototype.retryImmediately = function() {
     this._startKeepAlives(0);
     return true;
 };
-
+/**
+ * Process a single set of cached sync data.
+ * @param {Object} savedSync a saved sync that was persisted by a store. This
+ * should have been acquired via client.store.getSavedSync().
+ */
 SyncApi.prototype._syncFromCache = async function(savedSync) {
     debuglog("sync(): not doing HTTP hit, instead returning stored /sync data");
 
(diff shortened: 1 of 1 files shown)


src/sync.js
```diff
@@ -459,7 +461,18 @@ SyncApi.prototype.sync = function() {
         // no push rules for guests, no access to POST filter for guests.
         self._sync({});
     } else {
-        getPushRules();
+        // Don't do an HTTP hit to /sync. Instead, load up the persisted /sync data,
+        // if there is data there.
+        client.store.getSavedSync().then((savedSync) => {
+            // Indicate whether we should sync after getting push rules,
```
richvdh: I don't understand this comment. Indicate how, to whom?

src/sync.js
```diff
@@ -568,16 +581,9 @@ SyncApi.prototype._sync = async function(syncOptions) {
...
-    if (!syncOptions.hasSyncedBefore) {
-        // Don't do an HTTP hit to /sync. Instead, load up the persisted /sync data,
-        // if there is data there.
-        savedSync = await client.store.getSavedSync();
-    }
-
     let isCachedResponse = false;
     let data;
-
+    const savedSync = syncOptions.savedSync;
```
richvdh: this needs documenting in the jsdoc.

src/sync.js
lukebarnard1: Oh sorry, that was from a previously removed impl.

src/sync.js
```diff
@@ -500,6 +510,9 @@ SyncApi.prototype.retryImmediately = function() {
  * @param {Object} syncOptions
  * @param {string} syncOptions.filterId
  * @param {boolean} syncOptions.hasSyncedBefore
+ * @param {Object} syncOptions.savedSync a saved sync that was persisted by a
```
richvdh: can you start the description on a new line so that you get more than 10 chars wide?

lukebarnard1:
We've decided that it'd be a good idea to split out a `_syncFromCache` or something from `_sync`

src/sync.js
```diff
@@ -646,26 +656,22 @@ SyncApi.prototype._sync = async function(syncOptions) {
...
+    // keep emitting SYNCING -> SYNCING for clients who want to do bulk updates
+    this._updateSyncState("SYNCING", syncEventData);
 
-        // tell databases that everything is now in a consistent state and can be
-        // saved (no point doing so if we only have the data we just got out of the
-        // store).
-        client.store.save();
-    }
+    // tell databases that everything is now in a consistent state and can be
+    // saved (no point doing so if we only have the data we just got out of the
```
richvdh: I think the second part of this comment is irrelevant now

src/sync.js
```diff
@@ -504,6 +504,38 @@ SyncApi.prototype.retryImmediately = function() {
     return true;
 };
 
+SyncApi.prototype._syncFromCache = async function(savedSync) {
+    debuglog("sync(): not doing HTTP hit, instead returning stored /sync data");
+
+    const oldSyncToken = this.client.store.getSyncToken();
+    const nextSyncToken = savedSync.nextBatch;
+
+    this.client.store.setSyncToken(nextSyncToken);
```
richvdh: is this really right? it looks like we're going to update the stored sync token each time we start up, even if we're offline, which can't be right?

src/sync.js
```diff
@@ -460,6 +462,13 @@ SyncApi.prototype.sync = function() {
         // no push rules for guests, no access to POST filter for guests.
         self._sync({});
     } else {
+        // Don't do an HTTP hit to /sync. Instead, load up the persisted /sync data,
```
richvdh: this comment is a bit misleading. It might be better phrased as "before we start fetching the push rules and doing /sync hits, see if we have persisted /sync data and use it if so"

src/sync.js
richvdh: this is now outdated

src/sync.js
```diff
@@ -496,11 +505,46 @@ SyncApi.prototype.retryImmediately = function() {
     return true;
 };
 
+SyncApi.prototype._syncFromCache = async function(savedSync) {
```
richvdh: jsdoc wouldn't hurt, particularly since you've [already written it](https://github.com/matrix-org/matrix-js-sdk/pull/601/files#diff-b5ce4044fcc6267228c6281fd7483b4cR545)

src/sync.js
lukebarnard1: If we're offline, it'll always be the same - the `nextBatch` of the cached sync. When we start syncing, we use that token as the starting point. 

This was happening before. Am I missing something here?

richvdh: so you're saying that `oldSyncToken == nextSyncToken` in this case, so this is a no-op?

richvdh: I agree it does look like it was happening before, but I remain confused about how it actually works

lukebarnard1: >so you're saying that oldSyncToken == nextSyncToken in this case, so this is a no-op?

PR #603, opened 2018-02-07
Fix NPE when loading sync from cache

lukebarnard1:


src/sync.js
```diff
@@ -528,13 +530,13 @@ SyncApi.prototype._syncFromCache = async function(savedSync) {
     };
 
     try {
-        await this._processSyncResponse(oldSyncToken, data, true);
+        await this._processSyncResponse(oldSyncEventData, data, true);
```
richvdh: why aren't we just using `syncEventData` here? it seems to be what the other call to `_processSyncResponse` does?

lukebarnard1: No idea, we should be using `syncEventData`.

src/sync.js
```diff
@@ -513,7 +513,11 @@ SyncApi.prototype._syncFromCache = async function(savedSync) {
     debuglog("sync(): not doing HTTP hit, instead returning stored /sync data");
 
     // No previous sync, set old token to null
-    const oldSyncToken = null;
+    const syncEventData = {
+        oldSyncToken: null,
+        nextSyncToken,
```
richvdh: nextSyncToken isn't even set here

richvdh: [I feel like you could have spotted this yourself with a little more care...]

lukebarnard1: oops, sorry

PR #604, opened 2018-02-07
Set sync token before incrementally syncing

lukebarnard1:
otherwise we risk doing a full sync by using the null token.

This must be done synchronously before we `getPushRules` and start
incrementally syncing.


src/sync.js
```diff
@@ -514,15 +518,16 @@ SyncApi.prototype._syncFromCache = async function(savedSync) {
 
     const nextSyncToken = savedSync.nextBatch;
 
+    // Set sync token for future incremental syncing
+    self.client.store.setSyncToken(nextSyncToken);
```
richvdh: not sure why this needs moving, but whatever

PR #605, opened 2018-02-07
Guard against no pushRules in getPushRuleById

lukebarnard1:
so that clients that attempt to get push rules before they have
been fetched will receive `null` instead of an NPE.


lukebarnard1:
This isn't quite adequate because we're pretending that there are no push rules when in fact we don't know yet.

For offline-mode to work legitimately, we need a client that doesn't ask questions about pushRules until it has them.

PR #609, opened 2018-02-16
Some more doc on how sentinels work

dbkr:


src/models/room-state.js
```diff
@@ -16,6 +16,24 @@ limitations under the License.
 "use strict";
 /**
  * @module models/room-state
+ *
+ * Room State represents the state of the room at a given point.
+ * It can be mutated by adding state events to it.
+ * There are two types of room member associated with a state event:
+ * normal member objects (accessed via getMember/getMembers) mutate
```
ara4n: which mutate?

dbkr:
afaik they should include avatar URLs, as well as various other things like the member's power level.

ara4n:
lgtm

PR #610, opened 2018-02-16
Create sentinel members on-demand

dbkr:
We only need sentinel members for things like the 'sender' field
of events, so we previously created sentinels for everyone in the
room, but a large number of them were never used.

Instead, create them on-demand and cache them.


ara4n:
just for my reference: the reason we can blow away the sentinels when new members (or power level) events come in is because we were replacing them anyway.  So by blowing them away, we effectively go back to lazy-creating them if/when we needed them.  In other words, this should be no more violent than before.

lgtm!

PR #611, opened 2018-02-16
De-dup code: use the initialiseState function

dbkr:
This should behave identically, but the code here appeared to be
identical to the code in initialiseState, so let's use it (it also
has an extra sanity check in there that we only init empty timelines).

Edit: I take back what I said about this being functionally identical: it isn't. It appears that previously we were applying any events in the `state` field of the join room to both the old state and the new state, which is surely wrong. Although these are normally only ever sent in a limited or initial sync (so it wouldn't matter) they are well-defined in other syncs as events that precede the timeline events, so we should process them as we would timeline events, before we process the timeline events. The tests send sync responses with state changes in `state` rather than `timeline`.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -1294,6 +1294,9 @@ SyncApi.prototype._resolveInvites = function(room) {
  */
 SyncApi.prototype._processRoomEvents = function(room, stateEventList,
                                                 timelineEventList) {
+    // We'll only get state events in stateEventList if this is a limited
+    // sync, in which case this should be a fresh timeline, and this is us
+    // initialising it.
     if (stateEventList.length > 0) {
         room.getLiveTimeline().initialiseState(stateEventList);
     }
--- a/src/sync.js
+++ b/src/sync.js
@@ -1296,9 +1296,10 @@ SyncApi.prototype._processRoomEvents = function(room, stateEventList,
                                                 timelineEventList) {
     // If there are no events in the timeline yet, initialise it with
     // the given state events
-    const timelineWasEmpty = room.getLiveTimeline().getEvents().length == 0;
+    const liveTimeline = room.getLiveTimeline();
+    const timelineWasEmpty = liveTimeline.getEvents().length == 0;
     if (timelineWasEmpty) {
-        room.getLiveTimeline().initialiseState(stateEventList);
+        liveTimeline.initialiseState(stateEventList);
     }
 
     this._resolveInvites(room);
--- a/src/sync.js
+++ b/src/sync.js
@@ -1297,7 +1297,12 @@ SyncApi.prototype._processRoomEvents = function(room, stateEventList,
     // We'll only get state events in stateEventList if this is a limited
     // sync, in which case this should be a fresh timeline, and this is us
     // initialising it.
-    if (stateEventList.length > 0) {
+
+
+    // If there are no events in the timeline yet, initialise it with
+    // the given state events
+    const timelinewasEmpty = room.getLiveTimeline().getEvents().length == 0;
+    if (timelinewasEmpty) {
         room.getLiveTimeline().initialiseState(stateEventList);
     }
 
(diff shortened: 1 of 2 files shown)


src/sync.js
```diff
@@ -1294,41 +1294,35 @@ SyncApi.prototype._resolveInvites = function(room) {
...
-    // will leak through to current state!
-    const oldStateEvents = utils.map(
-        utils.deepCopy(
-            stateEventList.map(function(mxEvent) {
-                return mxEvent.event;
-            }),
-        ), client.getEventMapper(),
-    );
-    const stateEvents = stateEventList;
+    // We'll only get state events in stateEventList if this is a limited
```
richvdh: hrm... this is an assumption based on the way synapse behaves. Indeed it seems to be in conflict with the fact that you then do something with the state events if the timeline is not empty.

Could you rephrase this comment please.

src/sync.js
```diff
@@ -1294,41 +1294,35 @@ SyncApi.prototype._resolveInvites = function(room) {
...
     room.recalculate(this.client.credentials.userId);
 
-    // execute the timeline events, this will begin to diverge the current state
+    // If the timeline wasn't empty, we process the state events here: they're
+    // defined as updates to the state before the start of the timeline, so this
+    // starts to roll the state forward.
+    if (!timelinewasEmpty) {
+        room.addLiveEvents(stateEventList || []);
+    }
+    // execute the timeline events, this will continue to diverge the current state
```
richvdh: s/,/./

src/sync.js
```diff
@@ -1294,41 +1294,35 @@ SyncApi.prototype._resolveInvites = function(room) {
...
-    // XXX: what if we've already seen (some of) the events in the timeline,
-    // and they modify some of the state set in stateEvents? In that case we'll
-    // end up with the state from stateEvents, instead of the more recent state
-    // from the timeline.
-    room.oldState.setStateEvents(oldStateEvents);
-    room.currentState.setStateEvents(stateEvents);
+
+    // If there are no events in the timeline yet, initialise it with
+    // the given state events
+    const timelinewasEmpty = room.getLiveTimeline().getEvents().length == 0;
```
richvdh: `timelineWasEmpty` ?

src/sync.js
```diff
@@ -1294,41 +1294,35 @@ SyncApi.prototype._resolveInvites = function(room) {
...
+    if (timelinewasEmpty) {
+        room.getLiveTimeline().initialiseState(stateEventList);
+    }
 
     this._resolveInvites(room);
```
richvdh: why does this (and recalculate) happen after adding the state events for an empty timeline, but before it for an existing timeline?

src/sync.js
richvdh: can we factor out the myriad calls to `room.getLiveTimeline()` into a local var?

richvdh:
also: any chance of some tests for this stuff?

src/sync.js
dbkr: tbh I think deleting it may make more sense: I think this was left in from when the code was relying on this behaviour in synapse and I failed to remove it.

src/sync.js
dbkr: done

src/sync.js
dbkr: mm, yeah - fixed

src/sync.js
dbkr: My reasoning for this was the the state events ought to be treated the same as if they'd arrived in the timeline section. That said, I opted not to chase that particular rabbit down the hole to figure out what actually happens on a room name change, etc.

dbkr:
Now with tests that fails on develop

src/sync.js
richvdh: well, if that's true, why isn't it true for when the timeline is empty?

"I didn't go down the rabbithole" is valid but an `XXX` comment saying so would be good for future reference

dbkr: good point, and now looking at it, it looks a bit broken. Have added an XXX to explain how I think it might be broken.

PR #612, opened 2018-02-20
Stop cloning events when adding to state

dbkr:
As comment hopefully explains.

On my test account:
Before: 394657 MatrixEvents, 53MB shallow size
After: 198863 MatrixEvents, 27MB shallow size

### DIFF
--- a/src/models/event-timeline.js
+++ b/src/models/event-timeline.js
@@ -94,6 +94,10 @@ EventTimeline.prototype.initialiseState = function(stateEvents) {
     // not change. Duplicating the events uses a lot of extra memory,
     // so we now no longer do it. To assert that they really do never change,
     // freeze them! Note that we can't do this for events in general:
+    // although it looks like the only things preventing us are the
+    // 'status' flag, forwardLooking (which is only set once when adding to the
+    // timeline) and possibly the sender (which seems like it should never be
+    // reset but in practice causes a lot of the tests to break).
     for (const e of stateEvents) {
         Object.freeze(e);
     }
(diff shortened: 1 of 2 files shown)


src/models/event-timeline.js
```diff
@@ -88,19 +86,19 @@ EventTimeline.prototype.initialiseState = function(stateEvents) {
...
-
-    this._startState.setStateEvents(oldStateEvents);
+    // We previously deep copied events here and used different copies in
+    // the oldState and state events: this decision seems to date back
+    // quite a way and was apparently made to fix a bug where modifications
+    // made to the start state leaked through to the end state.
+    // This really shouldn't be possible though: the events themselves should
+    // not change. Duplicating the events uses a lot of extra memory,
+    // so we now no longer do it. To assert that they really do never change,
+    // freeze them! Note that we can't do this for events in general:
```
richvdh: (why not?)

src/models/event-timeline.js
```diff
@@ -88,19 +86,19 @@ EventTimeline.prototype.initialiseState = function(stateEvents) {
...
+    for (const e of stateEvents) {
+        Object.freeze(e);
```
richvdh: do we need to do some babel foo to make this work on node?

dbkr: Remarkably no: Object.freeze has been supported in node basically forever: http://node.green/#ES2015-misc-Object-static-methods-accept-primitives-Object-freeze

src/models/event-timeline.js
dbkr: have elaborated

PR #614, opened 2018-02-21
try a hunch to fix https://github.com/vector-im/riot-web/issues/6212

ara4n:
by deleting fields from /sync responses as you emit them into events,
to avoid wasting RAM when parsing the enormous accumulated sync response
from indexeddb on launch

### DIFF
--- a/src/store/indexeddb-local-backend.js
+++ b/src/store/indexeddb-local-backend.js
@@ -361,7 +361,7 @@ LocalIndexedDBStoreBackend.prototype = {
      */
     _loadSyncData: function() {
         console.log(
-            `LocalIndexedDBStoreBackend: loaded sync data`,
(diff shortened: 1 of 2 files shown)


ara4n:
i haven't been able to test this yet as my initial sync won't complete (been waiting 10 minutes now), but if my theory at https://github.com/vector-im/riot-web/issues/6212#issuecomment-367155453 is right, it might help a bit?

dbkr:
was the IRL conclusion that this wouldn't help because they were being retained elsewhere?

ara4n:
yeah, it was. the Promises.series() stuff was holding onto a reference, although if you've killed that it might work quite nicely?

dbkr:
Assuming this is probably unnecessary now

PR #615, opened 2018-02-22
Intern legacy top-level 'membership' field

dbkr:
There is a legacy top-level 'membership' field on events. We were
interning the normal one in 'content', but not this one, so the
legacy field was still keeping a copy of the string 'join' / 'leave'
etc in memory for every member event.

This make the interning code know about this field.


ara4n:
lgtm, although it'd be nice if synapse didn't waste bandwidth by sending us the legacy fields...

dbkr:
Yeah :) I failed to find where this was in synapse, so I'll settle for gently kicking Erik until it goes away :p

PR #616, opened 2018-02-22
Fix bug where global "Never send to unverified..." is ignored

lukebarnard1:
- Emit `crypto.initComplete` for apps that need to execute code that requires crypto to have initialised
 - Fix overriding `undefined` per-room unverified devices setting

Codep https://github.com/matrix-org/matrix-react-sdk/pull/1772


lukebarnard1:
Erm actually no. I shall modify this and the related react-sdk fix

lukebarnard1:
Probably should have squashed that, but oh well.

PR #617, opened 2018-02-23
Precompute push actions for state events

dbkr:
State events get froze now, so we can't write to the push actions
cache after having done so: precompute the push actions to work
around this.


ara4n:
lgtm under the circumstances

PR #621, opened 2018-03-06
Add cancelAndRetry flag to decryptEvent

lukebarnard1:
that allows for cancellation of outgoing key requests before
attempting to decrypt, which itself may do an outgoing key
request upon failure.

This is useful for cases where other clients have ignored the
requests but the user would like to resend the key request
manually.

### DIFF
--- a/src/crypto/OutgoingRoomKeyRequestManager.js
+++ b/src/crypto/OutgoingRoomKeyRequestManager.js
@@ -135,7 +135,7 @@ export default class OutgoingRoomKeyRequestManager {
      * @param {module:crypto~RoomKeyRequestBody} requestBody
      *
      * @returns {Promise} resolves when the request has been updated in our
-     *    pending list.
+     *    pending list and we have sent the cancellation.
      */
     cancelRoomKeyRequest(requestBody) {
         return this._cryptoStore.getOutgoingRoomKeyRequest(
@@ -195,9 +195,10 @@ export default class OutgoingRoomKeyRequestManager {
                         // with the same transaction_id, so only one message will get
                         // sent).
                         //
-                        // (We also don't want to wait for the response from the server
-                        // here, as it will slow down processing of received keys if we
-                        // do.)
+                        // (We want to wait for the response from the server here, so that
+                        // we can be sure that the request has been cancelled before
+                        // resolving, despite the fact that it will slow down processing
+                        // of received keys)
                         return this._sendOutgoingRoomKeyRequestCancellation(
                             updatedReq,
                         ).catch((e) => {
--- a/src/crypto/OutgoingRoomKeyRequestManager.js
+++ b/src/crypto/OutgoingRoomKeyRequestManager.js
@@ -198,7 +198,7 @@ export default class OutgoingRoomKeyRequestManager {
                         // (We also don't want to wait for the response from the server
                         // here, as it will slow down processing of received keys if we
                         // do.)
-                        this._sendOutgoingRoomKeyRequestCancellation(
+                        return this._sendOutgoingRoomKeyRequestCancellation(
                             updatedReq,
                         ).catch((e) => {
                             console.error(
@@ -206,7 +206,7 @@ export default class OutgoingRoomKeyRequestManager {
                                 + " will retry later.", e,
                             );
                             this._startTimer();
-                        }).done();
+                        });
                     });
 
                 default:
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -856,7 +856,8 @@ Crypto.prototype.requestRoomKey = function(requestBody, recipients) {
  *
  * @param {module:crypto~RoomKeyRequestBody} requestBody
  *    parameters to match for cancellation
- * @returns {Promise}
+ * @returns {Promise} Promise which resolves when the room key request has been
+ *                    cancelled.
  */
 Crypto.prototype.cancelRoomKeyRequest = function(requestBody) {
     return this._outgoingRoomKeyRequestManager.cancelRoomKeyRequest(requestBody)
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -856,12 +856,13 @@ Crypto.prototype.requestRoomKey = function(requestBody, recipients) {
  *
  * @param {module:crypto~RoomKeyRequestBody} requestBody
  *    parameters to match for cancellation
+ * @returns {Promise}
  */
 Crypto.prototype.cancelRoomKeyRequest = function(requestBody) {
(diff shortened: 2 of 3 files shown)


src/crypto/OutgoingRoomKeyRequestManager.js
```diff
@@ -198,15 +198,15 @@ export default class OutgoingRoomKeyRequestManager {
                         // (We also don't want to wait for the response from the server
                         // here, as it will slow down processing of received keys if we
                         // do.)
-                        this._sendOutgoingRoomKeyRequestCancellation(
+                        return this._sendOutgoingRoomKeyRequestCancellation(
```
richvdh: Please don't change code without changing the corresponding comments to make them not be lies. [here](https://github.com/matrix-org/matrix-js-sdk/pull/621/files#diff-d8c620291b1726321a300d1e99bb1256R198) and [here](https://github.com/matrix-org/matrix-js-sdk/pull/621/files#diff-d8c620291b1726321a300d1e99bb1256R137)

src/crypto/index.js
```diff
@@ -856,12 +856,13 @@ Crypto.prototype.requestRoomKey = function(requestBody, recipients) {
  *
  * @param {module:crypto~RoomKeyRequestBody} requestBody
  *    parameters to match for cancellation
+ * @returns {Promise}
```
richvdh: promise which resolves when?

src/crypto/index.js
```diff
@@ -856,12 +856,13 @@ Crypto.prototype.requestRoomKey = function(requestBody, recipients) {
...
  */
 Crypto.prototype.cancelRoomKeyRequest = function(requestBody) {
-    this._outgoingRoomKeyRequestManager.cancelRoomKeyRequest(requestBody)
+    return this._outgoingRoomKeyRequestManager.cancelRoomKeyRequest(requestBody)
     .catch((e) => {
         console.warn("Error clearing pending room key requests", e);
```
richvdh: Are you sure we want to swallow the error for your usecase?

src/models/event.js
```diff
@@ -340,19 +340,34 @@ utils.extend(module.exports.MatrixEvent.prototype, {
...
+            const wireContent = this.getWireContent();
+            await crypto.cancelRoomKeyRequest({
+                algorithm: wireContent.algorithm,
+                room_id: this.getRoomId(),
+                session_id: wireContent.session_id,
+                sender_key: wireContent.sender_key,
+            });
+        }
+
+        if (!cancelAndRetry && (
```
richvdh: why do we not throw an error if `cancelAndRetry` is set?

src/models/event.js
```diff
@@ -340,19 +340,34 @@ utils.extend(module.exports.MatrixEvent.prototype, {
...
-    attemptDecryption: async function(crypto) {
+    attemptDecryption: async function(crypto, cancelAndRetry=false) {
         // start with a couple of sanity checks.
         if (!this.isEncrypted()) {
             throw new Error("Attempt to decrypt event which isn't encrypted");
         }
 
-        if (
-            this._clearEvent && this._clearEvent.content &&
+        if (cancelAndRetry) {
```
richvdh: the comment says "start with a couple of sanity checks." This isn't a sanity check. Again, please try not to make comments lie.

richvdh: what happens if we get here while the previous decryption attempt is still in progress? (ie `_decryptionPromise` is truthy)

src/crypto/index.js
lukebarnard1: If `cancelRoomKeyRequest`, the request was in an unhandled state or it didn't find a matching request in the store. If there's no matching request then there's nothing to cancel, so it's fine?

src/models/event.js
lukebarnard1: We want to avoid
```
                this._clearEvent && this._clearEvent.content &&
                this._clearEvent.content.msgtype !== "m.bad.encrypted"
```
which will be true for an event that we already failed to decrypt.

src/models/event.js
lukebarnard1: If the decryption is ongoing (attempt A) and we call `attemptDecryption` with `cancelAndRetry = true` (attempt B), then we cancel any ongoing key request and return `_decryptionPromise`. 

So either:
 - We attempted decryption (A), were successful and therefore didn't send a key request. We then cancel the key request (via B), but none was present (nop). Finally we return the `_decryptionPromise`.
 - We attempted decryption (A), failed and therefore sent a key request. We then cancel the key request (via B), and either:
     - It was in the UNSENT or SENT state, in which case we cancel it or 
     - It was in CANCELLATION_PENDING state somehow (nop) or
     - we throw, but this is swallowed.
 - We then continue to try and decrypt the event, which will fail causing a new key request to be sent.

So it's not optimal. If decryption is ongoing, we probably shouldn't cancel any requests.

ara4n: presumably this is quite a likely scenario, as an overenthusiastic user will stab the fudge button on adjacent messages (which is doubly bad if we end up sending multiple keyshare requests for the same session?)

ara4n:
@richvdh reading through this one, i very much agree that we need to be thorough on e2e review - although the choice of language comes across pretty harshly.  Perhaps s/make the comments not be lies/remember to update the comments and API docs when changing the implementation/?

richvdh:
Yeah sorry, I was in a cranky mood, if you can believe it. Keep the faith,  @lukebarnard1!

ara4n:
to be clear, i do think the feedback is legitimate though (just unconstructively expressed): luke, given how critical and fragile the e2e codebase is, please be apply maximum attention to detail.

richvdh:
So I still have some specific thoughts on the code, but I'm afraid I'm now wondering if we should consider a slightly different approach.

In particular, why are we going to so much effort to make sure that we send out cancellations before sending another key request? Why don't we just send another request (with the same txn id)? I think that will achieve the result we want, with fewer race conditions and fewer messages flying around.

Making that work might require some tweaks to the [state machine](https://github.com/matrix-org/matrix-js-sdk/blob/c7ba8c0115b079f6d24b8824b99b1311d58e1bdb/src/crypto/OutgoingRoomKeyRequestManager.js#L34) but I don't think it'll be anything too huge.

What do you think?

lukebarnard1:
It is a bit strange to send a cancellation followed immediately by another request but I think it might be worth keeping to the existing state machine instead of modifying it to do something it technically can already do. I'm anxious about having an early optimisation here.

But it could be that I'm not all too familiar with the implications of sending a new request. Would/should clients not potentially ignore messages with txn ids that they've seen before?

lukebarnard1:
Closing in favour of https://github.com/matrix-org/matrix-js-sdk/pull/624

PR #625, opened 2018-03-08
Don't do /keys/changes on incremental sync

dbkr:
Remove the call to /keys/changes when we do an incremental syn
where the old sync token doesn't match the one in the device list
store. To allow us to do this, always save the device list store
before saving the sync data, so we can safely assume the device
list store is at least as fresh as the sync token in the sync store.

Thread save functions through to allow this, add a delay parameter
so the sync can save the device list immediately and skip the wait,
and add a wantsSave() method so the sync can skip saving the device
list if the sync store isn't going to save anyway.

Fixes https://github.com/vector-im/riot-web/issues/6068

### DIFF
--- a/src/crypto/DeviceList.js
+++ b/src/crypto/DeviceList.js
@@ -146,18 +146,23 @@ export default class DeviceList {
      * The actual save will be delayed by a short amount of time to
      * aggregate multiple writes to the database.
      *
+     * @param {integer} delay Time in ms before which the save actually happens.
+     *     By default, the save is delayed for a short period in order to batch
+     *     multiple writes, but this behaviour can be disabled by passing 0.
+     *
      * @return {Promise<bool>} true if the data was saved, false if
      *     it was not (eg. because no changes were pending). The promise
      *     will only resolve once the data is saved, so may take some time
      *     to resolve.
      */
-    async saveIfDirty() {
+    async saveIfDirty(delay) {
         if (!this._dirty) return Promise.resolve(false);
+        if (delay === undefined) delay = 500;
 
         if (this._savePromise === null) {
             // Delay saves for a bit so we can aggregate multiple saves that happen
             // in quick succession (eg. when a whole room's devices are marked as known)
-            this._savePromise = Promise.delay(500).then(() => {
+            this._savePromise = Promise.delay(delay).then(() => {
                 console.log('Saving device tracking data at token ' + this._syncToken);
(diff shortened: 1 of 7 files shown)


src/crypto/index.js
```diff
@@ -811,27 +827,18 @@ Crypto.prototype.decryptEvent = function(event) {
...
-            this._deviceList.getSyncToken() + " and " + syncData.oldSyncToken,
-        );
-        const gapDeviceLists = await this._baseApis.getKeyChanges(
-            this._deviceList.getSyncToken(), syncData.oldSyncToken,
-        );
-        this._evalDeviceListChanges(gapDeviceLists);
-        this._evalDeviceListChanges(syncDeviceLists);
-    }
+    // Here, we're relying on the fact that we only ever save the sync data after
+    // sucessfully saving the device list data, so we're guarenteed that the device
```
richvdh: guaranteed

src/crypto/DeviceList.js
```diff
@@ -146,18 +146,23 @@ export default class DeviceList {
...
+     *
      * @return {Promise<bool>} true if the data was saved, false if
      *     it was not (eg. because no changes were pending). The promise
      *     will only resolve once the data is saved, so may take some time
      *     to resolve.
      */
-    async saveIfDirty() {
+    async saveIfDirty(delay) {
         if (!this._dirty) return Promise.resolve(false);
+        if (delay === undefined) delay = 500;
```
richvdh: I'm a bit concerned that, if there is already a save scheduled, it might not happen for 500ms even if delay is 0. can you reorganise this somehow so that it will happen immediately?

dbkr: I was wondering whether this was worth worrying about. Have added it (although since Promise.delay can only be left to run or canceled outright, it's all had to change to a setTimeout)

src/crypto/DeviceList.js
```diff
@@ -146,38 +152,68 @@ export default class DeviceList {
...
+        if (savePromise === null) {
+            savePromise = new Promise((resolve, reject) => {
+                this._resolveSavePromise = resolve;
+            });
+            this._savePromise = savePromise;
+        }
 
-        if (this._savePromise === null) {
+        if (this._saveTimer === null) {
             // Delay saves for a bit so we can aggregate multiple saves that happen
```
richvdh: this comment probably needs to move up now

PR #627, opened 2018-03-09
Create indexeddb worker when starting the store

dbkr:
Rather than when creating it, otherwise we could potentially end
up starting workers unnecessarily.

### DIFF
--- a/src/store/indexeddb-remote-backend.js
+++ b/src/store/indexeddb-remote-backend.js
@@ -29,11 +29,11 @@ import Promise from 'bluebird';
  * @param {Object} WorkerApi The web worker compatible interface object
  */
 const RemoteIndexedDBStoreBackend = function RemoteIndexedDBStoreBackend(
-    workerScript, dbName, WorkerApi,
+    workerScript, dbName, workerApi,
 ) {
     this._workerScript = workerScript;
     this._dbName = dbName;
-    this._workerApi = WorkerApi;
+    this._workerApi = workerApi;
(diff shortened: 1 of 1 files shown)


src/store/indexeddb-remote-backend.js
```diff
@@ -31,20 +31,16 @@ import Promise from 'bluebird';
 const RemoteIndexedDBStoreBackend = function RemoteIndexedDBStoreBackend(
     workerScript, dbName, WorkerApi,
 ) {
+    this._workerScript = workerScript;
     this._dbName = dbName;
-    this._worker = new WorkerApi(workerScript);
+    this._workerApi = WorkerApi;
```
ara4n: i don't understand why we're assigning this to a field rather than just using it direct when we need it

src/store/indexeddb-remote-backend.js
```diff
@@ -31,20 +31,16 @@ import Promise from 'bluebird';
...
         // seq: promise,
     };
-
-    this._worker.onmessage = this._onWorkerMessage.bind(this);
-
-    // tell the worker the db name.
-    this._startPromise = this._doCmd('_setupWorker', [this._dbName]).then(() => {
-        console.log("IndexedDB worker is ready");
-    });
+    this._startPromise = null;
```
ara4n: some doc on the lifetime of this might be less confusing

ara4n:
lgtm other than two comments

PR #629, opened 2018-03-14
Start first incremental sync request early

dbkr:
So it can run while we process our sync data.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -396,6 +396,13 @@ SyncApi.prototype.getSyncState = function() {
     return this._syncState;
 };
 
+SyncApi.prototype.waitForKeepaliveAndSavedSync = async function (savedSyncPromise, err) {
+    await savedSyncPromise;
+    const keepaliveProm = this._startKeepAlives();
+    this._updateSyncState("ERROR", { error: err });
+    await keepaliveProm;
+};
+
 /**
  * Main entry point
  */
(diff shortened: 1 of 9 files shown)


dbkr:
as also discussed IRL, splitting sync state out turned out to be very nontrivial and quite invasive (and didn't feel like much of an improvement). Have fixed the promise chain screwup, could look at rejigging further.

src/sync.js
```diff
@@ -419,27 +419,26 @@ SyncApi.prototype.sync = function() {
...
-                const keepaliveProm = self._startKeepAlives();
-                self._updateSyncState("ERROR", { error: err });
-                return keepaliveProm;
-            }).then(() => {
-                getPushRules();
-            });
+            await savedSyncPromise;
+            const keepaliveProm = self._startKeepAlives();
+            self._updateSyncState("ERROR", { error: err });
+            await keepaliveProm;
```
lukebarnard1: The 6 lines above should probably be factored out from here and from getPushRules

src/sync.js
```diff
@@ -449,42 +448,40 @@ SyncApi.prototype.sync = function() {
...
+
+        if (self._currentSyncRequest === null) {
+            // Send this first sync request here so we can then wait for the saved
+            // sync data to finish processing before we process the results of this one.
+            const qps = self._getSyncParams({ filterId }, savedSyncToken);
+            console.log("Sending first sync request...");
+            self._currentSyncRequest = client._http.authedRequest(
+                undefined, "GET", "/sync", qps, undefined,
+                qps.timeout + BUFFER_PERIOD_MS,
+            );
```
lukebarnard1: I think the above few lines (qps onwards) should be factored out of here and _sync.

src/sync.js
dbkr: done, although suggestions welcome on function name

src/sync.js
dbkr: done

PR #643, opened 2018-04-30
fix typo where `headers` was not being used and thus sent wrong content-type

t3chguy:
higher up this method copies `opts.headers` into `headers` and if `json && data` then sets the content-type to `application/json`
Whilst writing matrix-search I was left with js-sdk reaching out to the express server with some json payload but plain text content type


t3chguy:
It'd only break things if the server was expecting JSON with a content-type of `text/plain`
which I'm hoping no-one did :')

PR #644, opened 2018-05-03
Add getMediaLimits to client

Half-Shot:
Part of https://github.com/matrix-org/matrix-doc/pull/1189 and https://github.com/matrix-org/synapse/pull/3184

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -727,6 +727,20 @@ MatrixClient.prototype.getGroups = function() {
     return this.store.getGroups();
 };
 
+/**
+ * Get the room for the given room ID.
+ * This function will return a valid room for any room for which a Room event
+ * has been emitted. Note in particular that other events, eg. RoomState.members
+ * will be emitted for a room before this function will return the given room.
+ * @param {module:client.callback} callback Optional.
+ * @return {module:client.Promise} Resolves with an object containing the limits.
+ */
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -727,6 +727,20 @@ MatrixClient.prototype.getGroups = function() {
     return this.store.getGroups();
 };
 
+/**
+ * Get the room for the given room ID.
```
turt2live: *ahem*

src/client.js
```diff
@@ -727,6 +727,20 @@ MatrixClient.prototype.getGroups = function() {
...
+ * This function will return a valid room for any room for which a Room event
+ * has been emitted. Note in particular that other events, eg. RoomState.members
+ * will be emitted for a room before this function will return the given room.
+ * @param {module:client.callback} callback Optional.
+ * @return {module:client.Promise} Resolves with an object containing the limits.
+ */
+MatrixClient.prototype.getMediaLimits = function(callback) {
```
turt2live: This should probably just be `getMediaRepoConfiguration` or something, given the opportunity for future expansion.

t3chguy: +++1 for magic this super generic

Half-Shot:
Aaaaaanyone? :(

Issue #645, opened 2018-05-22
`updatePendingEvent` doesn't emit oldEventId to 'Room.localEchoUpdated' event.


leonlianght:
I have a situation that when I send a message using `sendMessage`, matrix client returns a matrix event with a temporary event id, the `Room.localEchoUpdated` get fired when the message changing its status from `sending` to `sent`, however, I won't be able to get the temporary event id from `Room.localEchoUpdated`, that is, there is no way to connect the changed event.

Based on the matrix-js-sdk doc, the third parameter of [MatrixClient"Room.localEchoUpdated"](http://matrix-org.github.io/matrix-js-sdk/0.9.2/module-client.html#~event:MatrixClient%2522Room.localEchoUpdated%2522) should be oldEventId. However this line of code emits the newEventId since `event.getId()` returns the newEventId instead of the old one.
https://github.com/matrix-org/matrix-js-sdk/blob/4153845346c0d8ab087e0e6e854f6f75d744af6d/src/models/room.js#L820

I am not so sure whether it returns the `event.getId()` on purpose, if not, I can create a pr and fix this issue.

Thanks in advance.

PR #646, opened 2018-05-22
emit oldEventId on "updatePendingEvent"

leonlianght:
It should fire `Room.localEchoUpdated` event and returns the oldEventId

#645 

Signed-off-by: Leon leon390422914@gmail.com

### DIFF
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -817,7 +817,7 @@ Room.prototype.updatePendingEvent = function(event, newStatus, newEventId) {
         this.removeEvent(oldEventId);
     }
 
-    this.emit("Room.localEchoUpdated", event, this, event.getId(), oldStatus);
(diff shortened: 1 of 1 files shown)


dbkr:
This looks correct and sensible. Even though it changes the API to match the docs, people could be relying on the old behaviour, so this is probably worth a breaking change in the changelog.

Also, you forgot the DCO signoff: https://github.com/matrix-org/matrix-js-sdk/blob/master/CONTRIBUTING.rst

leonlianght:
I am sorry that I've missed your reply. Here is my singoff:
Signed-off-by: Leon <leon390422914@gmail.com>

jryans:
This change was reviewed and we have sign off. Should be safe to merge. I assume the breaking change notice will be handled when preparing the changelog.

PR #647, opened 2018-05-22
Emit Session.no_consent when M_CONSENT_NOT_GIVEN received

lukebarnard1:


lukebarnard1:
>top level, "no_consent"

sure. (/me wished these things were better named)

> Needs adding to the doc

sure.

src/crypto/RoomList.js
```diff
@@ -54,6 +54,7 @@ export default class RoomList {
                     } else {
                         this._roomEncryption = result;
                     }
+                    console.info('room encryption received');
```
dbkr: hello?

lukebarnard1: whoops

PR #649, opened 2018-05-23
Add `erase` option to deactivateAccount

lukebarnard1:
For erasing messages etc. after deactivation.

### DIFF
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -338,10 +338,11 @@ MatrixBaseApis.prototype.logout = function(callback) {
  * it is up to the caller to either reset or destroy the MatrixClient after
  * this method succeeds.
  * @param {object} auth Optional. Auth data to supply for User-Interactive auth.
- * @param {module:client.callback} callback Optional.
+ * @param {boolean} erase Optional. If true, send erase=true query parameter which
+ * indicates that the account should be erased.
  * @return {module:client.Promise} Resolves: On success, the empty object
  */
-MatrixBaseApis.prototype.deactivateAccount = function(auth, callback) {
+MatrixBaseApis.prototype.deactivateAccount = function(auth, erase) {
     let body = {};
     if (auth) {
         body = {
@@ -349,7 +350,12 @@ MatrixBaseApis.prototype.deactivateAccount = function(auth, callback) {
         };
     }
     return this._http.authedRequestWithPrefix(
-        callback, "POST", '/account/deactivate', undefined, body, httpApi.PREFIX_UNSTABLE,
+        undefined,
+        "POST",
+        '/account/deactivate',
+        erase ? { erase: true } : undefined,
+        body,
+        httpApi.PREFIX_UNSTABLE,
     );
 };
 
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -350,7 +350,7 @@ MatrixBaseApis.prototype.deactivateAccount = function(auth, erase) {
         );
     }
 
-    const body = { auth, erase };
+    const body = {} ;
     if (auth) {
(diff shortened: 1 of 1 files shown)


src/base-apis.js
```diff
@@ -338,16 +338,28 @@ MatrixBaseApis.prototype.logout = function(callback) {
...
  * @param {module:client.callback} callback Optional.
  * @return {module:client.Promise} Resolves: On success, the empty object
  */
-MatrixBaseApis.prototype.deactivateAccount = function(auth, callback) {
-    let body = {};
-    if (auth) {
-        body = {
-            auth: auth,
-        };
+MatrixBaseApis.prototype.deactivateAccount = function(auth, erase, callback) {
```
richvdh: this should never need *three* parameters? if you want to pass erase, you should use promises instead of callbacks

src/base-apis.js
```diff
@@ -338,16 +338,28 @@ MatrixBaseApis.prototype.logout = function(callback) {
...
+    if (typeof(erase) === 'function') {
```
richvdh: elsewhere we have a comment along the lines of "erase used to be callback" or words to that effect

src/base-apis.js
```diff
@@ -338,16 +338,28 @@ MatrixBaseApis.prototype.logout = function(callback) {
...
+        callback = erase;
+        erase = undefined;
+    }
+    if (typeof(auth) === 'function') {
```
richvdh: why would auth be a function?

src/base-apis.js
```diff
@@ -338,16 +338,28 @@ MatrixBaseApis.prototype.logout = function(callback) {
...
+        callback = auth;
+    }
+
+    const body = {};
+    if (typeof(auth) === 'object') {
```
richvdh: what was wrong with `if (auth)` ?

src/base-apis.js
```diff
@@ -338,18 +338,29 @@ MatrixBaseApis.prototype.logout = function(callback) {
...
-MatrixBaseApis.prototype.deactivateAccount = function(auth, callback) {
-    let body = {};
+MatrixBaseApis.prototype.deactivateAccount = function(auth, erase) {
+    if (typeof(erase) === 'function') {
+        throw new Error(
+            'deactivateAccount no longer accepts a callback parameter',
+        );
+    }
+
+    const body = { auth, erase };
```
richvdh: `auth` shouldn't be here?

src/base-apis.js
```diff
@@ -338,18 +338,29 @@ MatrixBaseApis.prototype.logout = function(callback) {
...
+        body.auth = auth;
+    }
+    if (erase !== undefined) {
+        body.erase = erase;
     }
+
     return this._http.authedRequestWithPrefix(
-        callback, "POST", '/account/deactivate', undefined, body, httpApi.PREFIX_UNSTABLE,
+        undefined, "POST", '/account/deactivate', undefined, body,
+        httpApi.PREFIX_UNSTABLE,
```
richvdh: can you fix this while you are here please

src/base-apis.js
lukebarnard1: actually, neither should erase... gah my brain is melting over simple changes

src/base-apis.js
lukebarnard1: fix what?

richvdh: PREFIX_UNSTABLE

richvdh: cf https://github.com/vector-im/riot-web/issues/5325

lukebarnard1: I've moved it to R0, which I think is correct.

PR #651, opened 2018-05-27
check whether notif level is undefined, because `0` is falsey 

t3chguy:
so if you set `@room` notif=0, then js-sdk would think its actually 50 :(

### DIFF
--- a/src/models/room-state.js
+++ b/src/models/room-state.js
@@ -455,7 +455,7 @@ RoomState.prototype.mayTriggerNotifOfType = function(notifLevelKey, userId) {
         powerLevelsEvent &&
         powerLevelsEvent.getContent() &&
         powerLevelsEvent.getContent().notifications &&
-        powerLevelsEvent.getContent().notifications[notifLevelKey]
+        powerLevelsEvent.getContent().notifications[notifLevelKey] !== undefined
     ) {
         notifLevel = powerLevelsEvent.getContent().notifications[notifLevelKey];
     }
--- a/src/models/room-state.js
+++ b/src/models/room-state.js
@@ -329,7 +329,7 @@ RoomState.prototype._hasSufficientPowerLevelFor = function(action, powerLevel) {
     }
 
     let requiredLevel = 50;
-    if (powerLevels[action] !== undefined) {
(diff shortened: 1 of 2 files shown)


src/models/room-state.js
```diff
@@ -455,7 +455,7 @@ RoomState.prototype.mayTriggerNotifOfType = function(notifLevelKey, userId) {
         powerLevelsEvent &&
         powerLevelsEvent.getContent() &&
         powerLevelsEvent.getContent().notifications &&
-        powerLevelsEvent.getContent().notifications[notifLevelKey]
+        powerLevelsEvent.getContent().notifications[notifLevelKey] !== undefined
```
turt2live: and if it's `null`, `false`, `NaN`, or `[]`? Might be best to change the check to be something like this:

```
!isNaN(powerLevelsEvent.getContent().notifications[notifLevelKey]) 
&& typeof(powerLevelsEvent.getContent().notifications[notifLevelKey]) === 'number'
```

t3chguy:
I kept it consistent with the other checks but did a not NaN previously, will make another pr to change them all after this one. I guess here we're trusting the server to sanitize the values

src/models/room-state.js
t3chguy: I got impatient and switched it and the others in this file over to a utils method `isNumber` I added which does pretty much the above

src/utils.js
```diff
@@ -652,3 +652,13 @@ module.exports.inherits = function(ctor, superCtor) {
         },
     });
 };
+
+/**
+ * Returns whether the given value is a finite number without type-coercion
+ *
+ * @param {*} value the value to test
+ * @return {boolean} whether ot not value is a finite number without type-coercion
```
lukebarnard1: "ot"

lukebarnard1:
> I guess here we're trusting the server to sanitize the values

I don't think we are doing that here.

>but did a not NaN previously

I can't see evidence of us doing that before. `isFinite(NaN)` returns false so I think we're fine here.

t3chguy:
> but did a not NaN previously

as in a local change before this commit I made, then went for `!== undefined` to match the rest of the class, then refactored the class

PR #659, opened 2018-06-15
fix auth header and filename=undefined

t3chguy:
+ fix one instance where useAuthorizationHeader was ignored
+ fix filename=undefined on e2ee uploads because `name || file.name` = undefined

### DIFF
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -1071,6 +1071,10 @@ MatrixBaseApis.prototype.searchUserDirectory = function(opts) {
  * @param {string=} opts.name   Name to give the file on the server. Defaults
  *   to <tt>file.name</tt>.
  *
+ * @param {boolean=} opts.omitFilename if true will not send the filename,
+ *   e.g for encrypted file uploads where filename leaks are undesirable.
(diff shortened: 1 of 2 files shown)


src/base-apis.js
```diff
@@ -1071,6 +1071,10 @@ MatrixBaseApis.prototype.searchUserDirectory = function(opts) {
  * @param {string=} opts.name   Name to give the file on the server. Defaults
  *   to <tt>file.name</tt>.
  *
+ * @param {boolean=} opts.omitFilename if true will not send the filename,
+ *   e.g for encrypted file uploads where filename leaks are undesirable.
+ *   Defaults to false.
```
dbkr: I'm going to bang my, "positive arguments" drum again, ie. `if (includeFilename)` vs `if (!omitFilename)`

t3chguy: But then it's a breaking change unless you mean to default it to true

dbkr: yep, suggesting defaulting to true

t3chguy: changed that for you :)

PR #660, opened 2018-06-17
fix filtering

t3chguy:
### Fixes rest of https://github.com/vector-im/riot-web/issues/6250

`/messages` has a weird filter format, currently js-sdk sent the derived fields at the root and the original filter defn in `filter_json` which is unread and unspecced.
image

We want to pass the raw defn to Synapse so that any filtering it does is identical to if this filter was used by ID via the sync API.

Also check if event.content.url is a string to decide whether it `contains_url`


jryans:
Thanks for making this contribution a while back. It looks like this needs a review and a rebase to resolve some conflicts. Do you still want move forward with this?

turt2live:
The merge conflicts on this seem fairly easy to resolve (Bluebird changes), but due to inactivity and age it's probably best to close this. @t3chguy if you want to move forward on this, I suggest a new PR.

PR #663, opened 2018-07-04
Add decryption error codes to base.DecryptionError

lukebarnard1:
These should roughly follow https://github.com/matrix-org/matrix-ios-sdk/blob/9732cf593206a10d2b60cd01151a759c88c1e9a2/MatrixSDK/Crypto/Algorithms/MXDecryptionResult.h#L21-L39

### DIFF
--- a/src/crypto/algorithms/megolm.js
+++ b/src/crypto/algorithms/megolm.js
@@ -641,8 +641,9 @@ MegolmDecryption.prototype.decryptEvent = async function(event) {
         if (e.message === 'OLM.UNKNOWN_MESSAGE_INDEX') {
             this._requestKeysForEvent(event);
         }
+        // TODO: make OlmDevice throw base.DecryptionErrors too
         throw new base.DecryptionError(
-            "OLM_UNKNOWN_MESSAGE_INDEX",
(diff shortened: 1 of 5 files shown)


src/crypto/algorithms/megolm.js
```diff
@@ -639,6 +642,7 @@ MegolmDecryption.prototype.decryptEvent = async function(event) {
             this._requestKeysForEvent(event);
         }
         throw new base.DecryptionError(
+            "OLM_UNKNOWN_MESSAGE_INDEX",
```
dbkr: The error is not necessarily `OLM_UNKNOWN_MESSAGE_INDEX` here, or at least there is a conditional above that checks for it so I would assume not.

lukebarnard1: IRL, me and @lampholder discussed. We don't especially care about all of the errors thrown by `decryptGroupMessage` (all of which are plain old `Error`) so I'll just rename this to `OLM_GROUP_MESSAGE_ERROR`.

lukebarnard1: (I changed my mind, we can still emit OLM_UNKNOWN_MESSAGE_INDEX)

PR #664, opened 2018-07-04
Expose last decryption error of an event that failed to be decrypted

lukebarnard1:


lukebarnard1:
Actually, yep that works too and has the added benefit of being a more typical way of providing error state.

lukebarnard1:
New PR: https://github.com/matrix-org/matrix-js-sdk/pull/665

PR #665, opened 2018-07-04
Expose event decryption error via Event.decrypted event

lukebarnard1:


t3chguy:
needs adding to jsdoc at bottom:
https://github.com/matrix-org/matrix-js-sdk/pull/665/files#diff-c2b46f9b1e3be0b9100ad9ef4970a4e3L710

lukebarnard1:
ah, thanks @t3chguy

PR #666, opened 2018-07-04
encrypt for invited users if history visibility allows.

ara4n:
fixes https://github.com/vector-im/riot-web/issues/2713

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -1085,8 +1085,7 @@ Crypto.prototype._onRoomMembership = function(event, member, oldMembership) {
         console.log('Join event for ' + member.userId + ' in ' + roomId);
         // make sure we are tracking the deviceList for this user
         this._deviceList.startTrackingDeviceList(member.userId);
-    }
-    else if (member.membership == 'invite' &&
+    } else if (member.membership == 'invite' &&
              this._clientStore.getRoom(roomId).shouldEncryptForInvitedUsers()) {
         console.log('Invite event for ' + member.userId + ' in ' + roomId);
         this._deviceList.startTrackingDeviceList(member.userId);
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -652,7 +652,7 @@ Crypto.prototype.setRoomEncryption = async function(roomId, config, inhibitDevic
         throw new Error(`Unable to enable encryption in unknown room ${roomId}`);
     }
 
-    const members = room.getJoinedMembers();
+    const members = room.getEncryptionTargetMembers();
     members.forEach((m) => {
         this._deviceList.startTrackingDeviceList(m.userId);
     });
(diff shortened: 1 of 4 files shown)


src/crypto/index.js
```diff
@@ -1085,6 +1085,10 @@ Crypto.prototype._onRoomMembership = function(event, member, oldMembership) {
         console.log('Join event for ' + member.userId + ' in ' + roomId);
         // make sure we are tracking the deviceList for this user
         this._deviceList.startTrackingDeviceList(member.userId);
+    } else if (member.membership == 'invite' &&
+             this._clientStore.getRoom(roomId).shouldEncryptForInvitedUsers()) {
+        console.log('Invite event for ' + member.userId + ' in ' + roomId);
+        this._deviceList.startTrackingDeviceList(member.userId);
```
richvdh: Hard to check from my phone, but I worry about this causing (more) device tracking bugs, since we won't be reliably receiving device list updates from this user, but maybe we will think we are up to date, and hence not resync once the user does join the room?

dbkr: Agreed that encrypting messages to invited user could work unreliably if the server doesn't have the corresponding change, but it should be limited to invited users, right? Messages to joined user should work no worse than before?

dbkr: Actually no, I see what you mean. If the server doesn't have https://github.com/matrix-org/synapse/pull/3484 then it breaks when:

 1. Alice is in a room
 1. Bob is invited to the room. Alice starts tracking Bob's device list
 1. Bob adds a device. The server doesn't send an update poke.
 1. Bob joins. Alice's client will call `startTrackingDeviceList(bob)` but nothing will happen because Bob is already tracked.
 1. Alice sends a message, but she doesn't know about Bob's new device list.

In other words, I think this is perfectly safe in the situation where we can be sure the server will have https://github.com/matrix-org/synapse/pull/3484 . In the real world, one workaround could be to always force a re-fetch of the device list for a user when their membership transitions to `join` in any room?

ara4n: this of course assumes the synapse patch works, and is going to send enough devicelist updates for invited users correctly. i will test by seeing what happens if you add a device to an invited user before they join the room. (and then for the dumb synapse scenario, explicitly refetch the list when they join, as dave says)

ara4n: actually, it's going to be really hard for me to investigate this today, and we ideally need it for tomorrow. @dbkr: would you be able to check the concern above (and mitigate it by resyncing on /join if needed)?

dbkr: Following up, there was much discussion and the conclusion is that there's a big edge case here in that federated servers do not know what other servers are in the room at the time they've been invited. They therefore will not send device updates if a user logs in whilst their state is `invite`.

dbkr: Fixed. Having looks into forcing an update when a member joins the room, actually it's not necessary since synapse sends a device list update when the user joins the room anyway. Arguably this is synapse impl specific behaviour though.

ara4n:
Conclusion here is that this works, other than for the edge case of https://github.com/matrix-org/synapse/issues/3504.

ara4n:
...and doesn't cause any regressions; hence merging.

PR #667, opened 2018-07-10
Lazy load room members - Part I

bwindels:
Lazy loading of members, for now using the /joined_members endpoint until we have something better.
 - [x] prototype lazily loaded members with /joined_members
- [x] pull code that assumes `RoomMember.events...` into RoomMember so it can transparently take lazily loaded members into account. 
 - [x] return fake sentinels with just user id when users haven't been loaded yet, better than breaking message continuation.
 - [x] timeline reset should take into account lazily loaded members of previous timeline
 - [x] use `/members` endpoint
 - [x] fix rooms not visible on initial sync where user wasn't active recently 
- [x] write/fix unit tests

Parent issue: https://github.com/vector-im/riot-web/issues/6611

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -754,7 +754,7 @@ MatrixClient.prototype.loadRoomMembersIfNeeded = function(roomId) {
     }
     const membersPromise = this.joinedMembers(roomId);
     room.setLazilyLoadedMembers(membersPromise);
-}
+};
 
 /**
  * Retrieve all known rooms.
--- a/src/client.js
+++ b/src/client.js
@@ -752,7 +752,18 @@ MatrixClient.prototype.loadRoomMembersIfNeeded = async function(roomId) {
     if (!room || !room.membersNeedLoading()) {
         return;
     }
-    const membersPromise = this.joinedMembers(roomId);
+    const joinedMembersPromise = this.joinedMembers(roomId);
+    const membersPromise = joinedMembersPromise.then((profiles) => {
+        return Object.entries(profiles).map(([userId, profile]) => {
+            return {
+                userId: userId,
+                avatarUrl: profile.avatar_url,
+                displayName: profile.display_name,
+                membership: "join",  // as we need to support invitees as well
+                                    // in the future, already include but hardcode it
+            };
+        });
+    });
     await room.setLazilyLoadedMembers(membersPromise);
 };
 
--- a/src/client.js
+++ b/src/client.js
@@ -764,7 +764,7 @@ MatrixClient.prototype.loadRoomMembersIfNeeded = async function(roomId) {
             };
         });
     });
-    await room.setLazilyLoadedMembers(membersPromise);
+    await room.setLazyLoadedMembers(membersPromise);
 };
 
 /**
--- a/src/client.js
+++ b/src/client.js
@@ -756,10 +756,16 @@ MatrixClient.prototype.loadRoomMembersIfNeeded = async function(roomId) {
     // room state at a given point in time. The plan is to do this by
     // passing the current next_batch sync token to the endpoint we use
     // to fetch the members. For now, this is a prototype that uses
-    // the /joined_members api, which does not support this synchronization,
-    // so there is a race condition here between the current /sync call
+    // the /joined_members api, which only tells us about the joined members
+    // (not invites for example) and does not support this synchronization.
+    // So there is a race condition here between the current /sync call
     // and the /joined_members call: if the have conflicting information, which one
     // represents the most recent state?
+    //
+    // Addressing this race condition and the fact that this only tells us about
+    // joined members is a prerequisite for taking this out of the prototype stage and
+    // enabling the feature flag (feature_lazyloading) that
+    // the call to this method is behind.
     const joinedMembersPromise = this.joinedMembers(roomId);
     const membersPromise = joinedMembersPromise.then((profiles) => {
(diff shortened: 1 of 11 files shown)


src/models/room-state.js
```diff
@@ -287,7 +321,7 @@ RoomState.prototype.getLastModifiedTime = function() {
 /**
  * Get user IDs with the specified display name.
  * @param {string} displayName The display name to get user IDs from.
- * @return {string[]} An array of user IDs or an empty array.
+ * @return {string[]} An array of user IDs or an empty array.
```
t3chguy: seems like a spurious trailing whitespace was added for no reason?

bwindels:
want to do a general review if it's ready to merge, but all should be here for a first PR.
Note that the react SDK code won't call into lazy loading code if the feature is not enabled.

src/models/room.js
```diff
@@ -201,6 +204,27 @@ Room.prototype.getLiveTimeline = function() {
...
+    return this._membersNeedLoading;
+};
+
+/**
+ * Sets the lazily loaded members from the result of calling /joined_members
+ * @param {Promise} joinedMembersPromise promise with result of /joined_members endpoint
+ */
+Room.prototype.setLazilyLoadedMembers = async function(joinedMembersPromise) {
+    this._membersNeedLoading = false;
+    const members = await joinedMembersPromise;
```
krombel: You might want to catch the case that this promise fails - e.g. when something is weird with the connection or server

src/models/room.js
```diff
@@ -201,6 +204,27 @@ Room.prototype.getLiveTimeline = function() {
...
+    //wait 10 seconds
+    await new Promise((resolve) => setTimeout(resolve, 5000));
```
krombel: Why should this wait? `await joinedMembersPromise` should make sure that it only continues executing once it is (successfully) done.
Besides that "5000" != 10 seconds ;)

src/models/room.js
bwindels: Good point!

src/client.js
```diff
@@ -742,6 +742,20 @@ MatrixClient.prototype.getRoom = function(roomId) {
...
+ * in case lazy loading of memberships is in use.
+ * @param {string} roomId The room ID
+ */
+MatrixClient.prototype.loadRoomMembersIfNeeded = function(roomId) {
+    const room = this.getRoom(roomId);
+    if (!room || !room.membersNeedLoading()) {
+        return;
+    }
+    const membersPromise = this.joinedMembers(roomId);
+    room.setLazilyLoadedMembers(membersPromise);
```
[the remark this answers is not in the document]
bwindels: Because putting the flag (`_membersNeedLoading`) whether loading has started in Room felt natural as it needs to be kept per room. I agree it's a bit weird, but modifying the flag, private field on Room, from MatrixClient would also not be pretty IMO. I didn't put the whole thing in Room as the pattern seems to be to do network I/O calls in client only. Wdyt?

I could still return a promise from here though, and put the error logging in RoomView then. I'll do that.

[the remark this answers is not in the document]
bwindels: @dbkr To clarify how the flag is used to avoid a race, I commented here in the follow-up PR. Makes more sense now?:
https://github.com/matrix-org/matrix-js-sdk/pull/669/commits/6570f56538be8198e69ff4308b69f11c47511fb6

bwindels:
@dbkr I think the LL members should be set on every state of every timeline of every set (not sure about search, but don't see any reason why that's different?), as we can't know since when that LL member is valid, but you have a point the the LL members should never take precedence over members known through state, as this would indeed override historical data. I'll make changes so this is the case.

Concerning tests: yes, I'll write some :+1:

bwindels:
Working on the tests, don't merge yet.

dbkr:
Just been having a think about this. I think you're right in that we would need to set the members on every timeline in order to get members for events in previous timeline when scrolling back? However, this surely won't work if the member has since left the room since we won't get them in `/joinedMembers` anyway, so it seems like this is never going to work reliably without server support for back-paginating (this is exactly the sort of edge case a unit test would be perfect for, ie. back-paginating to an event from a member who has now left).

I think that's the only reason for setting members on anything but the forward edge of the live timeline? In general I would definitely rather avoid putting member events on historical bits of the timeline - especially when we might want to use this data to figure out who was in a room at the time a message was sent to decide whether to send e2e keys to someone.

We should also comment the corners we're cutting here, ie. I'd expect a massive XXX comment saying that `/jojnedMembers` lacks a way to get the members at a specific point and the resulting raciness.

bwindels:
@dbkr Right, the only edge case I can think off is when you'd back paginate before the LL members are loaded, but that's probably avoidable.

Adding support for lazy loading in back-pagination is both the plan for client and server before releasing this, see https://github.com/vector-im/riot-web/issues/6611 (adding this to the description here as well), so you're right we don't really need to correct any other room state but the forward, live one. I'll change it back.

Added a big XXX corners cut comment where we call /joined_members :+1:

bwindels:
@dbkr @t3chguy ready for another look

dbkr:
Yeah, this feels a lot safer. Once we have a `/members` API that can get state for any given point, it might make sense to track the `membersNeedLoading` per roomState so we can also load member lists for these historical states. This can come later though.

bwindels:
Just had a look at the /members endpoint, and as it returns state events, not just profile information, some of the changes here could probably be avoided. So I'll make the change as part of this PR.

PR #669, opened 2018-07-18
Lazy loading: store lazily loaded members in indexeddb

bwindels:
This PR adds storing the lazy loaded members in indexeddb, so we don't have to call /members in every browser session (per browser session the members are already kept in memory once you've looked at the room).

It also stores a marker object in the same IDB store to know whether lazy loading was already done for a room even though it didn't yield any members to store (like in a DM where both members where active in the last 10 messages).

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -747,7 +747,7 @@ MatrixClient.prototype._loadMembers = async function(room) {
     // were the members loaded from the server?
     let fromServer = false;
     let rawMembersEvents = await this.store.getOutOfBandMembers(roomId);
-    if (rawMembersEvents.length == 0) {
+    if (rawMembersEvents === null) {
         fromServer = true;
         const lastEventId = room.getLastEventId();
         const response = await this.members(roomId, "join", "leave", lastEventId);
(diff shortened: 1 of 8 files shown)


src/store/indexeddb-local-backend.js
```diff
@@ -187,6 +198,94 @@ LocalIndexedDBStoreBackend.prototype = {
...
+        console.log(`LL: backend about to store ${membershipEvents.length}` +
+            ` members for ${roomId}`);
+        function ignoreResult() {}
+        // run everything in a promise so anything that throws will reject
+        return new Promise((resolve) =>{
+            const tx = this.db.transaction(["oob_membership_events"], "readwrite");
+            const store = tx.objectStore("oob_membership_events");
+            const eventPuts = membershipEvents.map((e) => {
+                const putPromise = promiseifyRequest(store.put(e));
+                return putPromise.then(ignoreResult);
```
dbkr: We're discarding the result of these individually and then as a batch later - is this to be particularly memory conservative or something?

src/client.js
```diff
@@ -752,13 +768,28 @@ MatrixClient.prototype.loadRoomMembersIfNeeded = async function(roomId) {
...
+    membersPromise = membersPromise.then(({memberEvents, fromServer}) => {
+        membersNeedStoring = fromServer;
+        return memberEvents;
     });
-    await room.loadOutOfBandMembers(eventsPromise);
+    await room.loadOutOfBandMembers(membersPromise);
+    // if loadOutOfBandMembers throws, this wont be called
+    // but that's fine as we don't want to store members
+    // that caused an error.
+    if (membersNeedStoring) {
```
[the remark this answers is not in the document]
bwindels: Exactly, so we only store members that are unknown by looking at the sync data.

src/store/indexeddb-local-backend.js
bwindels: Yeah, indeed, so the IDB event to signal success (which is the result of the promise) is discarded ASAP. I can remove it if you think it's confusing.

dbkr: Might be worth a comment but, yep, that makes sense.

dbkr: thanks!

PR #673, opened 2018-07-27
Fix infinite spinner upon joining a room

bwindels:
Fixes https://github.com/vector-im/riot-web/issues/6679

The way the spinner is supposed to disappear when clicking join in the room preview bar is a bit convoluted, but as far as I could figure out it works like this:
1. `RoomView` listens for `"RoomMember.membership`
1. when this fires, it checks if its for the logged in user, if so, forces a rerender
1.  in render, we will lookup and check the membership of the logged in user, to determine if the aux panel with the spinner (if !join) and the composer should be shown (if join). Even if `state.joining` remains `true`, which it is in this case.

What went wrong is that the lookup in `render` failed because the event was emitted before the new member was added to `RoomState::members`, and not seeing the membership is now `join`. This PR addresses that.

This also explains why this problem does not occur when you were invited into a room. Your member and membership would already be added to the state.

Also, I don't think `RoomViewStore/RoomView::joining` ever gets cleared apart from viewing the room again, which doesn't help to understand all of this. But just the changes here already fixed the problem for me.

### DIFF
--- a/spec/unit/room-state.spec.js
+++ b/spec/unit/room-state.spec.js
@@ -162,6 +162,7 @@ describe("RoomState", function() {
             ];
             let emitCount = 0;
             state.on("RoomState.newMember", function(ev, st, mem) {
+                expect(state.getMember(mem.userId)).toEqual(mem);
                 expect(mem.userId).toEqual(memberEvents[emitCount].getSender());
                 expect(mem.membership).toBeFalsy();  // not defined yet
                 emitCount += 1;
--- a/src/models/room-state.js
+++ b/src/models/room-state.js
@@ -217,8 +217,13 @@ RoomState.prototype.setStateEvents = function(stateEvents) {
             }
 
             let member = self.members[userId];
+            self._joinedMemberCount = null;
+
             if (!member) {
                 member = new RoomMember(event.getRoomId(), userId);
+                // add member to members before emitting any events,
+                // as event handlers often lookup the member
+                self.members[userId] = member;
                 self.emit("RoomState.newMember", event, self, member);
             }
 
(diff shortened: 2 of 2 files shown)


bwindels:
@dbkr Right, the `RoomState.newMember` event gets listened to on `sync.js:153`, to reemit some events, amongst which `RoomMember.membership`. After emitting `RoomState.newMember` in `room-state.js:222`, the very next thing that happens is calling `RoomMember::setMembershipEvent` which will emit `RoomMember.membership`, the relevant event for this bug. Since I didn't want to add a cross-class test (e.g. test that `RoomState.getMember` returns the member when `RoomMember` emits `RoomMember.membership`) I added a test for `RoomMember.newMember`. That looks fine to me as it wouldn't make sense to fire `RoomMember.membership` before `RoomState.newMember`. Maybe this is what you found confusing?

I think it makes sense to say that when firing `RoomState.newMember`,  `RoomState` should already consistently know about the new member (e.g. be able to return it from `getMember`). Good point about the docs, I can update those to be more explicit. You are right this very slightly changes the semantics of the event, but it should not break anything.

About the integration tests, yes, indeed. This is currently blocked on being able to have multiple users (something on the roadmap) as the tests start with a blank database every time, meaning no rooms or users. For this bug to occur you need to join a room (which is best if someone else created it). I plan to create a script before the tests are run that create a number of users and rooms. Once this is done it should be easy to catch this as part of the join a room test. I though it didn't make sense to block solving this bug on that though, so I just added a unit test for now.

bwindels:
Please review the back-port for lazy loading together with this PR: #674

dbkr:
Aha, yes, so the key here is that we're setting the member before `setMembershipEvent()` happens rather than before `RoomState.newMember` is emitted. That makes more sense. Possibly adding to that comment saying that `setMembershipEvent` causes `RoomMember.membership` to be emitted would have cleared up my confusion.

PR #675, opened 2018-07-31
make sure room.tags is always a valid object to avoid crashes

bwindels:
See https://github.com/vector-im/riot-web/issues/7025


t3chguy:
(except base branch)

t3chguy:
(rebuilding)

PR #676, opened 2018-08-01
Lazy loading: Lazy load members while backpaginating

bwindels:
Based on #677 

- [x] pass LL filter to server (combining with timeline filters?)
- [x] add received state to correct timeline
- [x] add LL flag in client instead of just passing in filter? (done, but needs cleanup)
- [x]  ~~Add RoomState to EventContext~~
- [x] Fix avatars not always appearing in history
- [x] ~~Fix MessagePanel::_getReadMarkerGhostTile generating double keys~~


bwindels:
`MatrixClient::scrollback` doesn't seem to be used anywhere (actually scrolling back triggers `MatrixClient::paginateEventTimeline`)  apart from an example and specs. Leaving it in (and added support for LL state) but we could consider removing the scrollback method altogether.

src/client.js
```diff
@@ -3038,8 +3017,11 @@ MatrixClient.prototype.getTurnServers = function() {
  *
  * @param {Boolean=} opts.disablePresence True to perform syncing without automatically
  * updating presence.
+ * @param {Boolean=} opts.lazyLoadMembers True to not load all membership events during
+ * initial sync but fetch them when needed by calling `loadOutOfBandMembers`
+ * This will override the filter option at this moment.
  */
-MatrixClient.prototype.startClient = function(opts) {
+MatrixClient.prototype.startClient = async function(opts) {
```
dbkr: Hmm - in terms of backwards compat, what will happen if a client doesn't wait on this promise and starts calling API methods immediately?

src/models/room-state.js
```diff
@@ -262,6 +262,23 @@ RoomState.prototype.clone = function() {
...
 
+/**
+ * Add previously unknown state events just before
+ * prepending events to the timeline.
+ * When lazy loading members while back-paginating,
+ * the relevant room state for the timeline chunk at the end
+ * of the chunk can be set with this method.
+ * @param {MatrixEvent[]} events state events to prepend
+ */
+RoomState.prototype.prependStateEvents = function(events) {
```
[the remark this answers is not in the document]
bwindels: The idea was that it inserts the state events as if they came before the state already present, hence prepend. But you're right, setUnknownStateEvents makes more sense.

src/client.js
bwindels: Good point. `MatrixClient::_syncApi` wouldn't be set yet, which could have some side-effects:
 - `getSyncState` returning null
 - `startClient` twice could leave you with two running `SyncApi`s
 - `stopClient` wouldn't stop the sync
 - `retryImmediately` would crash

Hmm, I guess this could be a breaking change then, and hence if we want to do it carry a major version number change.

bwindels: Maybe I just keep the filter creation inside MatrixClientPeg then?

dbkr: (out of band: let's do a version bump)

PR #677, opened 2018-08-02
Lazy loading: don't assume we have our own member available

bwindels:


bwindels:
@dbkr The room constructor is called once from react-sdk (from `VectorConferenceHandler.js:87`), but that should be addressed by the PR that you already merged. https://github.com/matrix-org/matrix-react-sdk/pull/2102/files

PR #683, opened 2018-08-06
Lazy loading: fix end-to-end encryption rooms

bwindels:
- [x] don't break device key tracking
- [x] when offline, make sure you can send messages in e2e room when back online

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -672,3 +672,12 @@ module.exports.removeHiddenChars = function(str) {
     return str.normalize('NFD').replace(removeHiddenCharsRegex, '');
 };
 const removeHiddenCharsRegex = /[\u200B-\u200D\u0300-\u036f\uFEFF\s]/g;
+
+module.exports.Deferred = function Deferred() {
+    let resolve = null;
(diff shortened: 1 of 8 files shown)


src/utils.js
```diff
@@ -672,3 +672,12 @@ module.exports.removeHiddenChars = function(str) {
...
 const removeHiddenCharsRegex = /[\u200B-\u200D\u0300-\u036f\uFEFF\s]/g;
+
+module.exports.Deferred = function Deferred() {
+    let resolve = null;
+    const promise = new Promise((r) => {
+        resolve = r;
+    });
+    promise.resolve = resolve;
+    return promise;
+};
```
dbkr: I think bluebird are fairly keen to avoid to avoid you doing this, ie. it's very strongly geared towards using `new Promise()` directly, so we should probably drink their kool-aid rather than try to work against the library.

bwindels: As all our target platforms all support native promises, it would be nice to drop bluebird at some point, and just use the DOM APIs on Promise. But given the changes we discussed this morning, I don't need this anymore, so getting rid of it.

PR #684, opened 2018-08-08
e2e key backups

uhoreg:
Follow-on work to https://github.com/matrix-org/matrix-js-sdk/pull/595

### DIFF
--- a/src/crypto/olmlib.js
+++ b/src/crypto/olmlib.js
@@ -38,7 +38,7 @@ module.exports.MEGOLM_ALGORITHM = "m.megolm.v1.aes-sha2";
 /**
  * matrix algorithm tag for megolm backups
  */
-module.exports.MEGOLM_BACKUP_ALGORITHM = "m.megolm_backup.v1";
+module.exports.MEGOLM_BACKUP_ALGORITHM = "m.megolm_backup.v1.curve25519-aes-sha2";
 
 
 /**


src/client.js
```diff
@@ -703,6 +703,90 @@ MatrixClient.prototype.importRoomKeys = function(keys) {
...
+            // decrypt response.data.session_data
+            const data = response.data;
+            const key = JSON.parse(decryptionKey.decrypt(data.session_data.ephemeral, data.session_data.mac, data.session_data.ciphertext));
+            // set room_id and session_id
+            key.room_id = roomId;
+            key.session_id = sessionId;
+            keys.push(key);
+            return this.importRoomKeys(keys);
+        } else {
+            callback("aargh!");
```
ara4n: we might need to do better than aargh ;)

spec/unit/crypto/backup.spec.js
```diff
@@ -0,0 +1,218 @@
+try {
```
ara4n: should prolly put a (c) on this

uhoreg:
@dbkr here is the initial draft of the API.  I haven't really tested it to see if it actually works yet -- this is mainly to give a general idea of what the API might look like (and to give a somewhat plausible implementation).  If you have any suggestions for improving the API (especially to make it fit in better with the rest of the SDK), let me know.  As far as implementation goes, I'm guessing you'll probably want to just replace things with stubs when you're doing the UI work, until I get it all working and tested.

The general idea is that you would call `client.getKeyBackupVersion` to see if there is a usable backup.  It'll throw an exception if there isn't.  If there is, it will return an object, which you can pass to `client.enableKeyBackup`.  Once `client.enableKeyBackup` is called, then any keys that are received will be backed up.  Backups can be disabled by calling `client.disableKeyBackup`.  A client can create a new backup version by calling `client.createKeyBackupVersion`.  This will automatically enable the backup.  `client.createKeyBackupVersion` will return a promise that will resolve to the backup recovery key, which can be presented to the user.

To restore from the key backup, you would call `client.restoreKeyBackups` with the backup recovery key, room ID (optional if session ID is not given), and session ID (optional).  This should fetch the key(s) from the server and supposedly re-attempt to decrypt previous events.

You can also delete keys from the backup by calling `client.deleteKeyBackups`.

Currently missing are:

* an API for backing up a key to the server (there is `client.sendKeyBackup`, but that's mainly intended for internal use, because it has the "data" parameter, which would be fiddly for people to attempt to build themselves)
* a proper backup recovery key (need to figure out what it should look like)
* probably other things that I'm forgetting right now

src/crypto/store/indexeddb-crypto-store-backend.js
```diff
@@ -460,6 +460,71 @@ export class Backend {
...
+            }
+        });
+    }
+
+    unmarkSessionsNeedingBackup(sessions) {
+        const txn = this._db.transaction("sessions_needing_backup", "readwrite");
+        const objectStore = txn.objectStore("sessions_needing_backup");
+        return Promise.all(sessions.map((session) => {
+            return new Promise((resolve, reject) => {
+                console.log(session);
```
dbkr: stray debug logging

src/crypto/index.js
```diff
@@ -1018,26 +1030,26 @@ Crypto.prototype.backupGroupSession = function(
...
+        'readwrite', [IndexedDBCryptoStore.STORE_SESSIONS, IndexedDBCryptoStore.STORE_BACKUP], (txn) => {
+            this._cryptoStore.getAllEndToEndInboundGroupSessions(txn, (session) => {
+                if (session !== null) {
+                    this._cryptoStore.markSessionsNeedingBackup([session], txn);
+                }
+            });
+        }
+    );
+
+    this._maybeSendKeyBackup();
```
dbkr: should await on this so the `backupAllGroupSessions` promise doesn't complete until the backup is done

src/crypto/index.js
```diff
@@ -1018,26 +1030,26 @@ Crypto.prototype.backupGroupSession = function(
...
-        sessionId, sessionKey, keysClaimed,
-        exportFormat,
-    );
-    return this._baseApis.sendKeyBackup(roomId, sessionId, this.backupInfo.version, data);
+    await this._cryptoStore.markSessionsNeedingBackup([{
+        senderKey: senderKey,
+        sessionId: sessionId,
+    }]);
+
+    this._maybeSendKeyBackup();
```
dbkr: probably worth `await`ing on this too? Actually possibly not: when we call it from megolm we only want to wait for it to be saved to indexeddb, not actually backed up. worth a comment though.

src/crypto/index.js
```diff
@@ -965,51 +966,62 @@ Crypto.prototype.importRoomKeys = function(keys) {
...
+                    await this._baseApis.sendKeyBackup(undefined, undefined, this.backupInfo.version, {rooms: data});
+                    successful = true;
+                    await this._cryptoStore.unmarkSessionsNeedingBackup(sessions);
+                }
+                catch (e) {
+                    console.log("send failed", e);
+                    // FIXME: pause
+                }
+            } while (!successful);
+            // FIXME: pause between iterations?
```
dbkr: to consolidate my comments from the individual commits, we should put the pause in, or just stop & retry next time a backup is triggered which would be my inclination.

src/crypto/index.js
uhoreg: Actually, I think it does make sense to await here, and then remove the awaits from megolm.

dbkr:
lgtm once new Olm is released

src/crypto/DeviceList.js
```diff
@@ -357,13 +373,24 @@ export default class DeviceList {
...
+            // if userId is omitted, shift the other arguments, and look up the
+            // userid
+            senderKey = algorithm;
+            algorithm = userId;
+            userId = this._userByIdentityKey[senderKey];
+            if (!userId) {
+                return null;
+            }
+        }
+
```
dbkr: This is a bit magic, but given the name of the function I can't really suggest anything better. Maybe in the future we can deprecate the 3 arg usage of this function altogether.

src/crypto/index.js
```diff
@@ -843,6 +969,126 @@ Crypto.prototype.importRoomKeys = function(keys) {
...
+        (txn) => {
+            this._cryptoStore.getAllEndToEndInboundGroupSessions(txn, (session) => {
+                if (session !== null) {
+                    this._cryptoStore.markSessionsNeedingBackup([session], txn);
+                }
+            });
+        },
+    );
+
+    await this._maybeSendKeyBackup();
```
manuroe: `backupAllGroupSessions` is called on user action but `_maybeSendKeyBackup` can delay its actual execution randomly up to 10s.
In the context of this function, we would like `_maybeSendKeyBackup` to start ASAP.

src/crypto/index.js
```diff
@@ -843,6 +969,126 @@ Crypto.prototype.importRoomKeys = function(keys) {
...
+                        session.senderKey, session.sessionId, session.sessionData,
+                    );
+                    sessionData.algorithm = olmlib.MEGOLM_ALGORITHM;
+                    delete sessionData.session_id;
+                    delete sessionData.room_id;
+                    const firstKnownIndex = sessionData.first_known_index;
+                    delete sessionData.first_known_index;
+                    const encrypted = this.backupKey.encrypt(JSON.stringify(sessionData));
+
+                    const forwardedCount =
```
manuroe: There is no forwardingCurve25519KeyChain field in the dict returned by exportInboundGroupSession but forwarding_curve25519_key_chain.

The mix of forwarding_curve25519_key_chain and forwardingCurve25519KeyChain in the js code makes me terrified.

src/crypto/algorithms/megolm.js
```diff
@@ -956,6 +973,18 @@ MegolmDecryption.prototype.importRoomKey = function(session) {
         session.sender_claimed_keys,
         true,
     ).then(() => {
+        if (this._crypto.backupInfo) {
+            // don't wait for it to complete
+            this._crypto.backupGroupSession(
```
manuroe: If this method is called because we are restoring keys from the current backup session, it looks inefficient to send back up the key to the homeserver.

src/crypto/index.js
dbkr: NB. I've added a delay param in https://github.com/matrix-org/matrix-js-sdk/pull/736

src/crypto/index.js
dbkr: Good spot - turns out this was completely broken

src/crypto/algorithms/megolm.js
dbkr: Yeah, at some point we may want to flag keys that came from the backup in the first place (and what version)

PR #686, opened 2018-08-13
Lazy loading: use sync token for /members?at=... request, as synapse expects it now

bwindels:


bwindels:
@dbkr Yes, this is handled by marking the `RoomMember::supersedesOutOfBand`. When the /members request comes back it doesn't override anything with that flag set. I was thinking about possibles races with this change here before, and there seemed to be some danger of the opposite: as processing the /sync response is not synchronous due to e2e being async, it could happen that:
 1. /sync comes back
 2. we save the next sync token
 3. /members?at=synctokensavedatstep2 is requested
 4. /members comes back with "user A" and mark it as out-of-band.
 5. we asynchronously process a membership event for "user A"

"User a" could be marked as out-of-band when it really shouldn't be (which would be sub-optimal). But this also won't happen as any update to `RoomMember` through `setMembershipEvent` will clear the out-of-band flag (step 5).

PR #687, opened 2018-08-13
Lazy loading: check with server for feature support

bwindels:
as per https://github.com/matrix-org/synapse/commit/e235005be68ee52680110c5bf0a1e20d17f52fb0

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -203,6 +203,8 @@ function MatrixClient(opts) {
 
     // The pushprocessor caches useful things, so keep one and re-use it
     this._pushProcessor = new PushProcessor(this);
+
+    this._serverSupportsLazyLoading = null;
 }
 utils.inherits(MatrixClient, EventEmitter);
 utils.extend(MatrixClient.prototype, MatrixBaseApis.prototype);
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -3078,6 +3087,28 @@ MatrixClient.prototype.stopClient = function() {
     global.clearTimeout(this._checkTurnServersTimeoutID);
 };
 
+/*
+ * Query the server to see if it support members lazy loading
+ * @return {Promise<boolean>} true if server supports lazy loading
+ */
+MatrixClient.prototype.doesServerSupportLazyLoading = async function() {
+    if (this._serverSupportsLazyLoading === null) {
```
dbkr: We should probably do this somewhere common and/or do it once & cache it for whatever other features we want to detect down the line. Probably just as good to do this when the time comes though.

bwindels: I did it like this because it won't always be called from the same point. It's used from two places with the currently proposed PRs:
- in `MatrixClient::startClient` only if the `lazyLoadMembers` flag is enabled to make sure the server supports it.
- from `UserSettings` to confirm it is supported before enabling it from the labs checkbox.

So I didn't see a point in the code to always call it unconditionally from there and then use that cached value from then on.

PR #689, opened 2018-08-14
Add getVersion to Room

dbkr:
To get the version number of the room as per https://github.com/matrix-org/matrix-doc/issues/1425

### DIFF
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -175,6 +175,18 @@ function Room(roomId, opts) {
 utils.inherits(Room, EventEmitter);
 
 /**
+ * Gets the version of the room
+ * @returns {string} The version of the room, or null if it could not be determined
+ */
+Room.prototype.getVersion = function() {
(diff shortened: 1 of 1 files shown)


src/models/room.js
```diff
@@ -174,6 +174,18 @@ function Room(roomId, opts) {
...
 
+/**
+ * Gets the version of the room
+ * @returns {string} The version of the room, or null if it could not be determined
+ */
+Room.prototype.getVersion = function() {
+    const createEvent = this.currentState.getStateEvents("m.room.create", "");
+    if (!createEvent) return null;
+    const ver = createEvent.getContent()['room_version'];
+    if (ver === undefined) return '1';
```
bwindels: `room_version` is also a string, and not a number, in the json of a room create event? Odd if it's numeric ...

dbkr: It is indeed defined to be a string (so you can have a version of 'dave-thing-testing' or whatever)

PR #694, opened 2018-08-20
Api to fetch events via /room/.../event/..

Half-Shot:
Use case for this is so bridges can pull out events from replies without all the guff around timelines and contexts.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -2078,6 +2078,7 @@ MatrixClient.prototype.getEventTimeline = function(timelineSet, eventId) {
  *
  * @param {string} roomId  The ID of the room to look in
  * @param {string} eventId  The ID of the event to look for
+ * @param {module:client.callback} callback Optional.
  *
  * @return {Promise} Resolves to an object containing the event.
  */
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -2071,6 +2071,28 @@ MatrixClient.prototype.getEventTimeline = function(timelineSet, eventId) {
...
+ * Get an event in a room by its event id.
+ *
+ * This blindly calls the /rooms/$roomId/event/$eventId endpoint.
+ *
+ * @param {string} roomId  The ID of the room to look in
+ * @param {string} eventId  The ID of the event to look for
+ *
+ * @return {Promise} Resolves to an object containing the event.
+ */
+MatrixClient.prototype.fetchRoomEvent = function(roomId, eventId) {
```
bwindels: Should this go in `MatrixBaseApis` (the base class of `MatrixClient`, so access wouldn't change) rather? That's where most of the pure network calls go, and as it doesn't return a `MatrixEvent` like most of the code in `MatrixClient`, but pure JSON instead, it feels like a bitter fit.

Half-Shot: Sure thing, I was mainly emulating some of the other calls we make that directly make calls. How do I transform a pure json (that is in standard C-S event format) to a MatrixEvent?

Half-Shot: Oh, ``new MatrixEvent({eventcontents})``

Half-Shot: Err, nevermind. I've just moved the func to base-apis.js

PR #700, opened 2018-08-24
Improve setRoomEncryption guard against multiple m.room.encryption st…

bwindels:
…ate events

we were only bailing out when receiving a non JSON-identical m.room.encryption event.
When receiving an identical event, the algorithm in _roomEncryptors would be reset,
generating a new megolm session every time this happens (there is a LL synapse bug
where this happens on every sync).

As the _roomList is backed by indexeddb you might already have a config without the algorithm being present though,
so we first check for the room encryptor algorithm being present. If so, always bail out as setRoomEncryption was
already called for the given room.

If no algorithm is present, still check if the config is not being changed.
Also setup the roomlist and room encryption synchronously before awaiting
the indexeddb operation to store the room encryption config in roomlist.

Fixes https://github.com/vector-im/riot-web/issues/7223 / https://github.com/vector-im/riot-web/issues/7222

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -648,7 +648,7 @@ Crypto.prototype.setRoomEncryption = async function(roomId, config, inhibitDevic
     // to avoid races when calling this method multiple times. Hence keep a hold of the promise.
     let storeConfigPromise = null;
     if(!existingConfig) {
-        const storeConfigPromise = this._roomList.setRoomEncryption(roomId, config);
+        storeConfigPromise = this._roomList.setRoomEncryption(roomId, config);
     }
 
     const AlgClass = algorithms.ENCRYPTION_CLASSES[config.algorithm];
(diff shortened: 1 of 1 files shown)


src/crypto/index.js
```diff
@@ -619,20 +619,35 @@ Crypto.prototype.getEventSenderDeviceInfo = function(event) {
 Crypto.prototype.setRoomEncryption = async function(roomId, config, inhibitDeviceQuery) {
     // if we already have encryption in this room, we should ignore this event
     // (for now at least. maybe we should alert the user somehow?)
-    const existingConfig = this._roomList.getRoomEncryption(roomId);
-    if (existingConfig && JSON.stringify(existingConfig) != JSON.stringify(config)) {
-        console.error("Ignoring m.room.encryption event which requests " +
-                      "a change of config in " + roomId);
+    const existingAlg = this._roomEncryptors[roomId];
+    if (existingAlg) {
```
dbkr: It might be nice to keep the console error in the case that we get an encryption event that's different to what we previously had, since this is a sign something fishy is going on with crypto.

ara4n: hasn't it just moved a few lines down? :)

ara4n: oh, no it hasn't. yeah; let's keep this warning in place.

bwindels: :+1: should be ready for another look

PR #702, opened 2018-08-28
Fix: show spinner again while recovering from connection error

bwindels:
Pass through PREPARED state after error, when keepalive returns succes.
This is according to the state diagram in client.js.
This will show a spinner at the bottom of a room again
while the catchup sync is in progress,
which seems to have broken at some point.

Addresses item 6 (We no longer show a spinner when recovering after losing connectivity) of https://github.com/vector-im/riot-web/issues/7182

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ BREAKING CHANGE
 ---------------
 
  * `MatrixClient::startClient` now returns a Promise. No method should be called on the client before that promise resolves. Before this method didn't return anything.
+ * A new `CATCHUP` sync state, emitted by `MatrixClient#"sync"` and returned by `MatrixClient::getSyncState()`, when doing initial sync after the `ERROR` state. See `MatrixClient` documentation for details.
 
 Changes in [0.11.0](https://github.com/matrix-org/matrix-js-sdk/releases/tag/v0.11.0) (TDB)
 ==================================================================================================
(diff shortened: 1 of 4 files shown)


bwindels:
I wonder if we should also show the spinner while in RECONNECTING state. If you sleep your laptop and open it again 3 days later and go straight back to riot, you might not have 3 sync fails to go from RECONNECTING to ERROR. If the catchup sync then takes long (Matthew had one that took 5 minutes, I guess also because of retry logic after 80 seconds), it'll be very confusing to not have a spinner. Downside would be that if a connection drop on a sync request, we'll show a spinner briefly... maybe that's better than the catchup-sync-without-spinner confusion? Wdyt @dbkr?

bwindels:
FTR, closed https://github.com/matrix-org/matrix-react-sdk/pull/2138 in favor of this approach.

bwindels:
Now needs https://github.com/matrix-org/matrix-react-sdk/pull/2143 on react-sdk side to fix the spinner.

bwindels:
@dbkr  Yeah, good point, this might be a breaking change, but these changes would be released together with the already breaking changes for lazy loading (e.g. `MatrixClient::startClient` returning a promise now), so I think we're good, no?

dbkr:
ah, yes - I'd forgotten about that one. Might be an idea to stick a 'breaking change' heading at the top of the changelog, or note them down somewhere, otherwise I'll forget.

PR #704, opened 2018-08-29
Pass through function to discard megolm session

dbkr:
To make debugging crypto slightly faster


bwindels:
Also, I guess you'll need a rebase to include the revert done on /develop to get the tests to succeed here, if desired.

PR #706, opened 2018-08-30
Fix DM avatar

bwindels:
Ready for first review, but need to test and fix lints and stuff first, so don't merge yet.

### DIFF
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -288,8 +288,7 @@ Room.prototype.getDMInviter = function() {
 
 /**
  * Assuming this room is a DM room, tries to guess with which user.
- * @return {string} user id of the other member, if found, otherwise undefined.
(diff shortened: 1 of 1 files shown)


src/models/room.js
```diff
@@ -279,14 +279,67 @@ Room.prototype.getDMInviter = function() {
...
+        if (sender !== this.myUserId) {
+            return sender;
+        }
+    }
+};
+
+Room.prototype.getAvatarFallbackMember = function() {
+    const memberCount = this.getInvitedAndJoinedMemberCount();
+    if (memberCount <= 2) {
+        if (this._summaryHeroes.length) {
```
dbkr: is this defined if LL is turned off?

bwindels: it is not, good catch

PR #707, opened 2018-08-30
Fix: also load invited members when lazy loading members

bwindels:
Fixes https://github.com/vector-im/riot-web/issues/7243


bwindels:
This will also include "ban" members, but that's probably fine.

PR #708, opened 2018-08-31
Fix for "otherMember.getAvatarUrl is not a function"

bwindels:
https://github.com/matrix-org/riot-web-rageshakes/issues/962
was hitting `Unhandled rejection TypeError: otherMember.getAvatarUrl is not a function` because in case of summary heroes it is returning a userid and not a member.

Also adds another fallback route for getting the avatar by looking up the user.


bwindels:
Regarding the user fallback, the complete thing to do here would also be to fallback to fetching the user profile for a hero if the user and member are not available, but then the whole things becomes async. I guess this will be fine for now.

PR #711, opened 2018-08-31
Remove lazy loaded members when leaving room

bwindels:
Fixes https://github.com/vector-im/riot-web/issues/7226, but needs more testing so don't merge yet.

### DIFF
--- a/src/store/indexeddb-local-backend.js
+++ b/src/store/indexeddb-local-backend.js
@@ -71,7 +71,7 @@ function selectQuery(store, keyRange, resultMapper) {
     });
 }
 
-function promiseifyTxn(txn) {
+function txnAsPromise(txn) {
     return new Promise((resolve, reject) => {
         txn.oncomplete = function(event) {
             resolve(event);
@@ -82,7 +82,7 @@ function promiseifyTxn(txn) {
     });
 }
 
-function promiseifyRequest(req) {
+function reqAsEventPromise(req) {
     return new Promise((resolve, reject) => {
         req.onsuccess = function(event) {
             resolve(event);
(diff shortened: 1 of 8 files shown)


bwindels:
Did more testing, seems to work well.

src/store/indexeddb-local-backend.js
```diff
@@ -292,6 +303,44 @@ LocalIndexedDBStoreBackend.prototype = {
...
+            [minStateKeyProm, maxStateKeyProm]);
+
+        const writeTx = this.db.transaction(
+            ["oob_membership_events"],
+            "readwrite");
+        const writeStore = writeTx.objectStore("oob_membership_events");
+        const membersKeyRange = IDBKeyRange.bound(
+            [roomId, minStateKey],
+            [roomId, maxStateKey],
+        );
```
dbkr: I'm missing why you can't just delete with `IDBKeyRange.only(roomId)`

bwindels: The primary key of `oob_membership_events` is a composed key of `[roomId, stateKey]`, as you need those two to uniquely identify a member. I didn't try this, but I don't think `IDBKeyRange.only` supports matching only a part of a composed key.

bwindels: Actually, I'm quite certain `IDBKeyRange.only` doesn't support prefix-matches as they are called, on composed keys; I asked this question earlier on stackoverflow to make sure this approach would work across browsers, and the person awnsering said prefix-matches are on peoples wish-lists for idb.

https://stackoverflow.com/questions/52111115/safe-to-assume-same-order-in-index-as-store-with-upper-part-of-key-same/52120855#52120855

PR #716, opened 2018-09-03
Fix filepanel also filtering main timeline with LL turned on.

bwindels:
Fixes https://github.com/vector-im/riot-web/issues/7262

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -2145,6 +2145,8 @@ function(roomId, fromToken, limit, dir, timelineFilter = undefined) {
 
     let filter = null;
     if (this._clientOpts.lazyLoadMembers) {
+        // create a shallow copy of LAZY_LOADING_MESSAGES_FILTER,
+        // so the timelineFilter doesn't get written into it below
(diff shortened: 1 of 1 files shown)


ara4n:
looks plausible. would be good to get a comment about why this is needed though (as i'm failing to see why/where the timeline filter ever gets written into this object?)

bwindels:
@ara4n Here: https://github.com/matrix-org/matrix-js-sdk/pull/716/commits/3363cc4f1d97a7556373db7bab321f4581d81e66#diff-cf27c1d543e886c89cd9ac8b8aeaf05bR2154

If the method got called once with both the LL & timeline filter (anytime you open the filepanel with LL enabled), the timeline filter would modify the original LL filter object, persisting the filepanel filter on the LL filter.

ara4n:
right. lgtm then - just need to get the comment in the code to spell out where the 'bad' assign happens to avoid someone breaking it or getting confused by it in future.

bwindels:
Right, will do!

bwindels:
@ara4n ready for another look

ara4n:
lgtm!

PR #717, opened 2018-09-03
Don't rely on members to query if syncing user can post to room

bwindels:
Required by https://github.com/matrix-org/matrix-react-sdk/pull/2145

### DIFF
--- a/src/models/room-state.js
+++ b/src/models/room-state.js
@@ -667,11 +667,6 @@ RoomState.prototype.maySendStateEvent = function(stateEventType, userId) {
  *                        according to the room's state.
  */
 RoomState.prototype._maySendEventOfType = function(eventType, userId, state) {
-    const member = this.getMember(userId);
-    if (!member || member.membership == 'leave') {
-        return false;
-    }
-
     const power_levels_event = this.getStateEvents('m.room.power_levels', '');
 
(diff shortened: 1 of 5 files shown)


src/models/room-state.js
```diff
@@ -663,25 +677,29 @@ RoomState.prototype._maySendEventOfType = function(eventType, userId, state) {
 
     let state_default = 0;
     let events_default = 0;
+
     if (power_levels_event) {
         power_levels = power_levels_event.getContent();
         events_levels = power_levels.events || {};
 
-        if (utils.isNumber(power_levels.state_default)) {
+        if (Number.isFinite(power_levels.state_default)) {
```
ara4n: why the change?

ara4n:
lgtm modulo one random q :)

src/models/room-state.js
bwindels: Better to use things that are now available in the standard JS as opposed to rolling our own?

dbkr: There is another `utils.isNumber` below we should probably change for consistency

bwindels:
making changes to avoid looking up the member, don't merge yet.

bwindels:
ready for another look

PR #727, opened 2018-09-07
Only emit CATCHUP if recovering from conn error

dbkr:
Have the keepalive promise return a boolean indicating whether it
detected a connectivity failure or not. Use this to only emit
CATCHUP if there was a connectivity error, to try & suppress
the state flip-flopping back & forth between CATCHUP and ERROR
in the case where we have connectivity but the sync is returning
and error for whatever reason.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -779,8 +779,12 @@ SyncApi.prototype._onSyncError = function(err, syncOptions) {
     // erroneous. We set the state to 'reconnecting'
     // instead, so that clients can observe this state
     // if they wish.
-    this._startKeepAlives().then(() => {
-        if (this.getSyncState() === 'ERROR') {
+    this._startKeepAlives().then((connDidFail) => {
+        // Only emit CATCHUP if we detected a connectivity error: if we didn't,
+        // it's quite likely the sync will fail again for the same reason and we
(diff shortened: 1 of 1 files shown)


src/sync.js
```diff
@@ -1246,16 +1253,17 @@ SyncApi.prototype._pokeKeepAlive = function() {
             // responses fail, this will mean we don't hammer in a loop.
             self._keepAliveTimer = setTimeout(success, 2000);
         } else {
+            connDidFail = true;
```
bwindels: Anything but a 400 would trigger this, no? Is that a safe assumption (e.g. 404, 500, ...)?
Rather than doing something like this:
https://stackoverflow.com/questions/20231075/how-to-catch-networkerror-in-javascript

dbkr: True, although given we're polling the /versions endpoint, it could be assumed that it's not going to return anything other than 200 or possibly 400 in the case of a very old HS. We could also have HTTP level errors from HTTP proxies which should probably be classed as connectivity errors. Not sure though - wdyt?

bwindels: Hmm, I'd add `404` to line 1241 because `400` seems more like a synapse idiosyncrasy than anything else?

dbkr: wfm

PR #734, opened 2018-09-17
Introduce Room.myMembership event

bwindels:
As you don't always have your own member with lazy loading
of members enabled, looking at the sync response section
where a room appears is the most reliable way of determining
the syncing user's membership in a room.

Before we already used this to read the
current room membership with `room.getMyMembership()`,
but we were still using the `RoomMember.membership` event
to detect when the syncing user's membership changed.

This event will help make those checks work well with LL enabled.

### DIFF
--- a/spec/unit/room.spec.js
+++ b/spec/unit/room.spec.js
@@ -1405,6 +1405,23 @@ describe("Room", function() {
             room.updateMyMembership("invite");
             expect(room.getMyMembership()).toEqual("invite");
         });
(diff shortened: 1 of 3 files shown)


bwindels:
Ran the e2e tests locally as they fail on CI atm, ran successfully.

bwindels:
Abandoning this PR in favor of #735 so the branch names matches with `react-sdk`, where the e2e tests fail because it's not using the right code in the js-sdk side.

PR #736, opened 2018-09-18
[WIP] Support for e2e key backups

dbkr:
Adds support for backing up e2e keys to the homeserver and restoring with a secret recovery key.

https://github.com/vector-im/riot-web/issues/3661

### DIFF
--- a/spec/unit/crypto/backup.spec.js
+++ b/spec/unit/crypto/backup.spec.js
@@ -1,0 +1,480 @@
+/*
+Copyright 2018 New Vector Ltd
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+try {
+    global.Olm = require('olm');
+} catch (e) {
+    console.warn("unable to run megolm backup tests: libolm not available");
+}
+
+import expect from 'expect';
+import Promise from 'bluebird';
+
+import sdk from '../../..';
+import algorithms from '../../../lib/crypto/algorithms';
+import WebStorageSessionStore from '../../../lib/store/session/webstorage';
+import MemoryCryptoStore from '../../../lib/crypto/store/memory-crypto-store.js';
+import MockStorageApi from '../../MockStorageApi';
+import testUtils from '../../test-utils';
+
+// Crypto and OlmDevice won't import unless we have global.Olm
+let OlmDevice;
+let Crypto;
+if (global.Olm) {
+    OlmDevice = require('../../../lib/crypto/OlmDevice');
+    Crypto = require('../../../lib/crypto');
+}
+
+const Olm = global.Olm;
+
+const MatrixClient = sdk.MatrixClient;
+const MatrixEvent = sdk.MatrixEvent;
+const MegolmDecryption = algorithms.DECRYPTION_CLASSES['m.megolm.v1.aes-sha2'];
+
+const ROOM_ID = '!ROOM:ID';
+
+const SESSION_ID = 'o+21hSjP+mgEmcfdslPsQdvzWnkdt0Wyo00Kp++R8Kc';
+const ENCRYPTED_EVENT = new MatrixEvent({
+    type: 'm.room.encrypted',
+    room_id: '!ROOM:ID',
+    content: {
+        algorithm: 'm.megolm.v1.aes-sha2',
+        sender_key: 'SENDER_CURVE25519',
+        session_id: SESSION_ID,
+        ciphertext: 'AwgAEjD+VwXZ7PoGPRS/H4kwpAsMp/g+WPvJVtPEKE8fmM9IcT/N'
+            + 'CiwPb8PehecDKP0cjm1XO88k6Bw3D17aGiBHr5iBoP7oSw8CXULXAMTkBl'
+            + 'mkufRQq2+d0Giy1s4/Cg5n13jSVrSb2q7VTSv1ZHAFjUCsLSfR0gxqcQs',
+    },
+    event_id: '$event1',
+    origin_server_ts: 1507753886000,
+});
+
+const KEY_BACKUP_DATA = {
+    first_message_index: 0,
+    forwarded_count: 0,
+    is_verified: false,
+    session_data: {
+        ciphertext: '2z2M7CZ+azAiTHN1oFzZ3smAFFt+LEOYY6h3QO3XXGdw'
+            + '6YpNn/gpHDO6I/rgj1zNd4FoTmzcQgvKdU8kN20u5BWRHxaHTZ'
+            + 'Slne5RxE6vUdREsBgZePglBNyG0AogR/PVdcrv/v18Y6rLM5O9'
+            + 'SELmwbV63uV9Kuu/misMxoqbuqEdG7uujyaEKtjlQsJ5MGPQOy'
+            + 'Syw7XrnesSwF6XWRMxcPGRV0xZr3s9PI350Wve3EncjRgJ9IGF'
+            + 'ru1bcptMqfXgPZkOyGvrphHoFfoK7nY3xMEHUiaTRfRIjq8HNV'
+            + '4o8QY1qmWGnxNBQgOlL8MZlykjg3ULmQ3DtFfQPj/YYGS3jzxv'
+            + 'C+EBjaafmsg+52CTeK3Rswu72PX450BnSZ1i3If4xWAUKvjTpe'
+            + 'Ug5aDLqttOv1pITolTJDw5W/SD+b5rjEKg1CFCHGEGE9wwV3Nf'
+            + 'QHVCQL+dfpd7Or0poy4dqKMAi3g0o3Tg7edIF8d5rREmxaALPy'
+            + 'iie8PHD8mj/5Y0GLqrac4CD6+Mop7eUTzVovprjg',
+        mac: '5lxYBHQU80M',
+        ephemeral: '/Bn0A4UMFwJaDDvh0aEk1XZj3k1IfgCxgFY9P9a0b14',
+    },
+};
+
+describe("MegolmBackup", function() {
+    if (!global.Olm) {
+        console.warn('Not running megolm backup unit tests: libolm not present');
+        return;
+    }
+
+    let olmDevice;
+    let mockOlmLib;
+    let mockCrypto;
+    let mockStorage;
+    let sessionStore;
+    let cryptoStore;
+    let megolmDecryption;
+    beforeEach(async function() {
+        await Olm.init();
+        testUtils.beforeEach(this); // eslint-disable-line no-invalid-this
+
+        mockCrypto = testUtils.mock(Crypto, 'Crypto');
+        mockCrypto.backupKey = new Olm.PkEncryption();
+        mockCrypto.backupKey.set_recipient_key(
+            "hSDwCYkwp1R0i33ctD73Wg2/Og0mOBr066SpjqqbTmo",
+        );
+        mockCrypto.backupInfo = {
+            algorithm: "m.megolm_backup.v1",
+            version: 1,
+            auth_data: {
+                public_key: "hSDwCYkwp1R0i33ctD73Wg2/Og0mOBr066SpjqqbTmo",
+            },
+        };
+
+        mockStorage = new MockStorageApi();
+        sessionStore = new WebStorageSessionStore(mockStorage);
+        cryptoStore = new MemoryCryptoStore(mockStorage);
+
+        olmDevice = new OlmDevice(sessionStore, cryptoStore);
+
+        // we stub out the olm encryption bits
+        mockOlmLib = {};
+        mockOlmLib.ensureOlmSessionsForDevices = expect.createSpy();
+        mockOlmLib.encryptMessageForDevice =
+            expect.createSpy().andReturn(Promise.resolve());
+    });
+
+    describe("backup", function() {
+        let mockBaseApis;
+
+        beforeEach(function() {
+            mockBaseApis = {};
+
+            megolmDecryption = new MegolmDecryption({
+                userId: '@user:id',
+                crypto: mockCrypto,
+                olmDevice: olmDevice,
+                baseApis: mockBaseApis,
+                roomId: ROOM_ID,
+            });
+
+            megolmDecryption.olmlib = mockOlmLib;
+        });
+
+        it('automatically calls the key back up', function() {
+            const groupSession = new Olm.OutboundGroupSession();
+            groupSession.create();
+
+            // construct a fake decrypted key event via the use of a mocked
+            // 'crypto' implementation.
+            const event = new MatrixEvent({
+                type: 'm.room.encrypted',
+            });
+            const decryptedData = {
+                clearEvent: {
+                    type: 'm.room_key',
+                    content: {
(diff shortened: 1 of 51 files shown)


dbkr:
Shifted to https://github.com/matrix-org/matrix-js-sdk/pull/684 again

spec/unit/crypto/backup.spec.js
```diff
@@ -0,0 +1,490 @@
...
+                        auth_data: {
+                            public_key: "hSDwCYkwp1R0i33ctD73Wg2/Og0mOBr066SpjqqbTmo",
+                        },
+                    });
+                    let numCalls = 0;
+                    return new Promise((resolve, reject) => {
+                        client._http.authedRequest = function(
+                            callback, method, path, queryParams, data, opts,
+                        ) {
+                            expect(++numCalls <= 1);
```
ara4n: putting a pre-increment in here as a side-effect to the expect feels unnecessarily magical & terse;  can we just have a numCalls++ on the line above  or something?

spec/unit/crypto/backup.spec.js
```diff
@@ -0,0 +1,490 @@
...
+                        };
+                        client._crypto.backupGroupSession(
+                            "roomId",
+                            "F0Q2NmyJNgUVj9DGsb4ZQt3aVxhVcUQhg7+gvW0oyKI",
+                            [],
+                            groupSession.session_id(),
+                            groupSession.session_key(),
+                        );
+                    })
+                        .then(() => {
```
ara4n: this feels like it's indented one level too deep?

spec/unit/crypto/backup.spec.js
```diff
@@ -0,0 +1,490 @@
...
+                        auth_data: {
+                            public_key: "hSDwCYkwp1R0i33ctD73Wg2/Og0mOBr066SpjqqbTmo",
+                        },
+                    });
+                    let numCalls = 0;
+                    return new Promise((resolve, reject) => {
+                        client._http.authedRequest = function(
+                            callback, method, path, queryParams, data, opts,
+                        ) {
+                            expect(++numCalls <= 2);
```
ara4n: pre-increment feels icky

spec/unit/crypto/backup.spec.js
```diff
@@ -0,0 +1,490 @@
...
+                        };
+                        client._crypto.backupGroupSession(
+                            "roomId",
+                            "F0Q2NmyJNgUVj9DGsb4ZQt3aVxhVcUQhg7+gvW0oyKI",
+                            [],
+                            groupSession.session_id(),
+                            groupSession.session_key(),
+                        );
+                    })
+                        .then(() => {
```
ara4n: indentation weirdness again

spec/unit/crypto/backup.spec.js
```diff
@@ -0,0 +1,490 @@
...
+            });
+        });
+
+        it('sends backups to the server', function() {
+            const groupSession = new Olm.OutboundGroupSession();
+            groupSession.create();
+            const ibGroupSession = new Olm.InboundGroupSession();
+            ibGroupSession.create(groupSession.session_key());
+
+            const scheduler = [
```
ara4n: rather than copypasting this massive wodge of setup code across all the tests, can we factor it out?

src/client.js
```diff
@@ -737,6 +770,301 @@ MatrixClient.prototype.importRoomKeys = function(keys) {
...
+    if (this._crypto === null) {
+        throw new Error("End-to-end encryption disabled");
+    }
+
+    this._crypto.backupInfo = info;
+    if (this._crypto.backupKey) this._crypto.backupKey.free();
+    this._crypto.backupKey = new global.Olm.PkEncryption();
+    this._crypto.backupKey.set_recipient_key(info.auth_data.public_key);
+
+    this.emit('keyBackupStatus', true);
```
ara4n: I wonder if this event should be `crypto.keyBackupStatus` to match the namespace used for other crypto stuff?

src/client.js
```diff
@@ -737,6 +770,301 @@ MatrixClient.prototype.importRoomKeys = function(keys) {
...
+ * @returns {Promise<object>} Object with 'version' param indicating the version created
+ */
+MatrixClient.prototype.createKeyBackupVersion = function(info) {
+    if (this._crypto === null) {
+        throw new Error("End-to-end encryption disabled");
+    }
+
+    const data = {
+        algorithm: info.algorithm,
+        auth_data: info.auth_data, // FIXME: should this be cloned?
```
ara4n: hm, so should it? this feels like something we should be confident about

src/client.js
```diff
@@ -737,6 +770,301 @@ MatrixClient.prototype.importRoomKeys = function(keys) {
...
+    } catch(e) {
+        decryption.free();
+        throw e;
+    }
+
+    return this._http.authedRequest(
+        undefined, "GET", path.path, path.queryData,
+    ).then((res) => {
+        if (res.rooms) {
+            for (const [roomId, roomData] of Object.entries(res.rooms)) {
```
ara4n: ooh, i had no idea you could do that

src/crypto/DeviceList.js
```diff
@@ -129,6 +132,22 @@ export default class DeviceList {
                             deviceData.trackingStatus : {};
                         this._syncToken = deviceData ? deviceData.syncToken : null;
                     }
+                    this._userByIdentityKey = {};
+                    for (const user in this._devices) {
```
ara4n: can we use `of` rather than `in` and avoid evil ugly `hasOwnProperty` checks?

src/crypto/DeviceList.js
```diff
@@ -364,13 +383,24 @@ export default class DeviceList {
     /**
      * Find a device by curve25519 identity key
      *
-     * @param {string} userId     owner of the device
+     * @param {string} userId     owner of the device (optional)
```
ara4n: i'm completely blanking on why we bother letting you specify userId at all, if it's optional - given we can work it out from senderKey.  i assume there's an excellent reason,  but it should probably be spelt out.  it also feels really weird to  put the optional param at the beginning of the arg list and do this shift dance rather than having it as an optional last arg.

src/crypto/DeviceList.js
```diff
@@ -415,7 +445,23 @@ export default class DeviceList {
...
+            for (const [deviceId, dev] of Object.entries(this._devices[u])) {
+                const identityKey = dev.keys['curve25519:'+deviceId];
+
+                delete this._userByIdentityKey[identityKey];
+            }
+        }
+
         this._devices[u] = devs;
+
+        // add enw ones
```
ara4n: new

src/crypto/OlmDevice.js
```diff
@@ -81,6 +81,17 @@ function OlmDevice(sessionStore, cryptoStore) {
     this.deviceEd25519Key = null;
     this._maxOneTimeKeys = null;
 
+    // track which of our other devices (if any) have cross-signed this device
+    // XXX: this should probably have a single source of truth in the /devices
+    // API store or whatever we use to track our self-signed devices.
+    this.crossSelfSigs = [];
```
ara4n: this interloper doesn't seem to be being used?

src/crypto/OlmDevice.js
```diff
@@ -81,6 +81,17 @@ function OlmDevice(sessionStore, cryptoStore) {
...
+
+    // track whether we have already suggested to the user that they should
+    // restore their keys from backup or by cross-signing the device.
+    // We use this to avoid repeatedly emitting the suggestion event.
+    // XXX: persist this somewhere!
+    this.suggestedKeyRestore = false;
```
ara4n: this feels like it's deeply spurious stuff left over from my very original PR?

src/crypto/algorithms/megolm.js
```diff
@@ -840,6 +848,15 @@ MegolmDecryption.prototype.onRoomKeyEvent = function(event) {
 
         // have another go at decrypting events sent with this session.
         this._retryDecryption(senderKey, sessionId);
+    }).then(() => {
+        if (this._crypto.backupInfo) {
+            // don't wait for it to complete
```
ara4n: 'it' being the promise returned by backupGroupSession? should prolly spell 'it' out

src/crypto/index.js
```diff
@@ -181,6 +189,115 @@ Crypto.prototype.init = async function() {
...
+ * to it.
+ */
+Crypto.prototype._checkAndStartKeyBackup = async function() {
+    console.log("Checking key backup status...");
+    let backupInfo;
+    try {
+        backupInfo = await this._baseApis.getKeyBackupVersion();
+    } catch (e) {
+        console.log("Error checking for active key backup", e);
+        if (Number.isFinite(e.httpStatus) && e.httpStatus / 100 === 4) {
```
ara4n: when would you ever have something which is divisible by 100 to give 4... but isn't a finite number?

src/crypto/index.js
```diff
@@ -851,6 +969,135 @@ Crypto.prototype.importRoomKeys = function(keys) {
...
+    if (!this._sendingBackups) {
+        this._sendingBackups = true;
+        try {
+            if (delay === undefined) {
+                // by default, wait between 0 and 10 seconds, to avoid backup
+                // requests from different clients hitting the server all at
+                // the same time when a new key is sent
+                delay = Math.random() * 10000;
+            }
+            if (delay > 0) {
```
ara4n: the `if` seems redundant here and just confuses things (even if it makes the 1 in 10,000 scenario that delay=0 infinitessimally faster)

src/crypto/recoverykey.js
```diff
@@ -0,0 +1,67 @@
...
+    buf.set(OLM_RECOVERY_KEY_PREFIX, 0);
+    buf.set(key, OLM_RECOVERY_KEY_PREFIX.length);
+
+    let parity = 0;
+    for (let i = 0; i < buf.length - 1; ++i) {
+        parity ^= buf[i];
+    }
+    buf[buf.length - 1] = parity;
+    const base58key = bs58.encode(buf);
+
```
ara4n: weird double NL?

src/crypto/recoverykey.js
```diff
@@ -0,0 +1,67 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import bs58 from 'bs58';
+
+// picked arbitrarily but to try & avoid clashing with any bitcoin ones
+// (also base58 encoded, albeit with a lot of hashing)
```
ara4n: call me blind, but where is this hashing happening?

src/crypto/store/localStorage-crypto-store.js
```diff
@@ -167,6 +168,57 @@ export default class LocalStorageCryptoStore extends MemoryCryptoStore {
...
     }
 
+    getSessionsNeedingBackup(limit) {
+        const sessionsNeedingBackup
+              = getJsonItem(this.store, KEY_SESSIONS_NEEDING_BACKUP) || {};
+        const sessions = [];
+
+        for (const session in sessionsNeedingBackup) {
+            if (Object.prototype.hasOwnProperty.call(sessionsNeedingBackup, session)) {
+                const senderKey = session.substr(0, 43);
```
ara4n: the magic numbers here feel scary; is this really the best we can do? at least, can they be const'd?

ara4n: or factored out ,given they appear in the memory-crypto-store too?

ara4n:
generally looks excellent. all my feedback is relatively minor nitpicks.

src/client.js
uhoreg: Looking through all the places where auth_data is used, it looks safe to leave it uncloned at the moment.  The only concern would be that it might break future things, it seems pretty unlikely.

src/crypto/index.js
```diff
@@ -851,6 +969,135 @@ Crypto.prototype.importRoomKeys = function(keys) {
...
+                    );
+                    sessionData.algorithm = olmlib.MEGOLM_ALGORITHM;
+                    delete sessionData.session_id;
+                    delete sessionData.room_id;
+                    const firstKnownIndex = sessionData.first_known_index;
+                    delete sessionData.first_known_index;
+                    const encrypted = this.backupKey.encrypt(JSON.stringify(sessionData));
+
+                    const forwardedCount =
+                          (sessionData.forwardingCurve25519KeyChain || []).length;
```
uhoreg: should be `sessionData.forwarding_curve25519_key_chain` (snake-case), as [Manu pointed out](https://github.com/matrix-org/matrix-js-sdk/pull/684#discussion_r227703573)

src/client.js
dbkr: It is probably unnecessary - removing the fixme

src/crypto/DeviceList.js
dbkr: Turns out there was only one other place it was called anyway so I have just made it the 2 arg version

src/crypto/OlmDevice.js
dbkr: presumably popping in from the cross signing stuff :)

src/crypto/index.js
dbkr: I think this was essentially checking it's a number as opposed to null / undefined, although either can be divided by 100 to give something that is not 4, so arguably unnecessary.

src/crypto/index.js
```diff
@@ -851,6 +969,135 @@ Crypto.prototype.importRoomKeys = function(keys) {
...
+                } catch (err) {
+                    numFailures++;
+                    console.log("send failed", err);
+                    if (err.httpStatus === 400
+                        || err.httpStatus === 403
+                        || err.httpStatus === 401
+                        || !retry) {
+                        // retrying probably won't help much, so we should give up
+                        // FIXME: disable backups?
+                        throw err;
```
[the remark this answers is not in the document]
dbkr: good point actually - I think the best place to do this is with a catch on the call sites of backupGroupSession().

src/crypto/index.js
dbkr: ah yes, it should.

src/crypto/index.js
```diff
@@ -181,6 +189,115 @@ Crypto.prototype.init = async function() {
...
+    this._checkedForBackup = true;
+
+    const trustInfo = await this.isKeyBackupTrusted(backupInfo);
+
+    if (trustInfo.usable && !this.backupInfo) {
+        console.log("Found usable key backup: enabling key backups");
+        this._baseApis.enableKeyBackup(backupInfo);
+    } else if (!trustInfo.usable && this.backupInfo) {
+        console.log("No usable key backup: disabling key backup");
+        this._baseApis.disableKeyBackup();
```
manuroe: Why not managing the case `(trustInfo.usable && this.backupInfo`) ?

If I understand the logic right, it would be:
```
else if (trustInfo.usable && this.backupInfo) {
        console.log("Found usable key backup: use the new key backup version");
        this._baseApis.disableKeyBackup();
        this._baseApis.enableKeyBackup(backupInfo);
    }
```

manuroe: Of course, we should do nothing if the backup version we got from the hs is the same as in `this.backupInfo`.

dbkr: ah sorry - replied to this on matrix but should do so here for completeness: we probably ought to but there's nothing that would actually trigger us to look for this case since we don't get told about new backup version in the sync stream.

src/crypto/recoverykey.js
```diff
@@ -0,0 +1,66 @@
...
+        throw new Error("Incorrect parity");
+    }
+
+    for (let i = 0; i < OLM_RECOVERY_KEY_PREFIX.length; ++i) {
+        if (result[i] !== OLM_RECOVERY_KEY_PREFIX[i]) {
+            throw new Error("Incorrect prefix");
+        }
+    }
+
+    if (
```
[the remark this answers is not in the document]
uhoreg: Yup.  I don't think it matters much for JavaScript, but it would matter for lower-level languages.

PR #739, opened 2018-09-20
add new examples, to be expanded into a post

benparsons:


README.md
```diff
@@ -30,9 +30,61 @@ In Node.js
...
   });
 ```
+
 See below for how to include libolm to enable end-to-end-encryption. Please check
 [the Node.js terminal app](examples/node) for a more complex example.
 
+To start the client:
+
+```javascript
+client.startClient({initialSyncLimit: 10});
```
bwindels: With lazy-loading, this now returns a promise, so one would either need to `await client.startClient(...);` or use `then` like so: `client.startClient(...).then(() => {/* work with client here*/});`. First option is probably preferable.

README.md
```diff
@@ -30,9 +30,61 @@ In Node.js
...
+  }
+  console.log(event.event.content.body);
+});
+```
+
+By default, the `matrix-js-sdk` client uses the `MatrixInMemoryStore` to store events as they are received. Access this via `client.store`. For example to iterate through the currently stored timeline for a room:
+
+```javascript
+Object.keys(client.store.rooms).forEach((roomId) => {
+  client.store.rooms[roomId].timeline.forEach(t => {
```
bwindels: Might want to use `client.getRoom(roomId).timeline` here as it's a slightly more public and stable API I feel. Unless you specifically want to show the usage of the memory store here, that is.

PR #740, opened 2018-09-20
WIP: track gappy sync performance metrics

bwindels:
this is quite hacky but couldn't find an immediate better way of going about it, and also wanted to validate the idea first with the team before putting more effort in. So any feedback welcome on better ways of doing this.
Related to https://github.com/vector-im/riot-web/issues/7327
See https://github.com/matrix-org/matrix-react-sdk/pull/2172 of how this could be sent to Analytics...

### DIFF
--- a/src/http-api.js
+++ b/src/http-api.js
@@ -766,6 +766,10 @@ module.exports.MatrixHttpApi.prototype = {
                         }
                     }
 
+                    if (typeof opts.reportContentLength === "function") {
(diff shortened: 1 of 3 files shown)


turt2live:
(is this actually still a WIP?)

bwindels:
Sort off, it's mostly done but switched to other stuff while in London. Will pick it back up again today probably.

PR #742, opened 2018-09-25
Split npm start into an init and watch script

turt2live:
This is to better support riot-web's build process without losing the functionality supplied by `npm start`. The watch script no longer performs an initial build and thus `start:init` has been created for this purpose.

Required for https://github.com/vector-im/riot-web/pull/7355


ara4n:
lgtm

PR #746, opened 2018-09-26
Detect when lazy loading has been toggled in client.startClient

bwindels:
And throw an `InvalidStoreError`, which is then handled in the react-sdk.

This is needed to enable triggering a resync when lazy loaded is enabled by default and users encounter it for the first time.


spec/unit/matrix-client.spec.js
```diff
@@ -191,31 +194,40 @@ describe("MatrixClient", function() {
...
+                    if (state === "SYNCING") {
+                        client.removeListener("sync", syncListener);
+                        resolve();
+                    }
+                });
             });
-            client.startClient();
+            */
+            await client.startClient();
+            // await syncingPromise;
```
dbkr: what happened to this test?

PR #749, opened 2018-09-28
Extra checks to avoid release script blowing up mid-process.

bwindels:


turt2live:
@bwindels is this still needed?

bwindels:
huh, I thought we merged this already. Yeah, still makes sense for it to go in.

PR #751, opened 2018-10-02
Reject with the actual error on indexeddb error

dbkr:
Rather than the event


turt2live:
The mozilla docs seem to suggest using `txn.error` instead:

```
  transaction.onerror = function(event) {
    note.innerHTML += '<li>Transaction not opened due to error: ' + transaction.error + '</li>';
  };
```

Source: https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction/onerror

Does this give you the same error you're seeing in `.target.error`?

dbkr:
The target of the event will be the transaction, so I think these are the same thing.

PR #755, opened 2018-10-04
Default to a room version of 1 when there is no room create event

turt2live:
Fixes https://github.com/vector-im/riot-web/issues/7331

There is something to be worried about when there is no room create event, however. This should always be available, although due to cache problems or servers that don't provide the event we can't be sure of this.


turt2live:
I'm not certain, but for the room joining problem specifically: the create event appears to arrive after the room join completes, leaving a small gap where there is nothing in the cache.

PR #756, opened 2018-10-05
Propagate Indexeddb VersionError as InvalidStoreError

bwindels:


bwindels:
Making a simpler version of this, as not doing the version error anymore, see https://github.com/matrix-org/matrix-react-sdk/pull/2199#issuecomment-428099019

PR #760, opened 2018-10-11
never replace /sync'ed memberships with OOB ones

bwindels:
Fixes https://github.com/vector-im/riot-web/issues/7450


bwindels:
Still thinking why I did it like this though, and if this won't break anything else ...

PR #763, opened 2018-10-23
Logging lib. Fixes #332

### LINKED ISSUE #332: Remove verbose logging from files in src/crypto
In order to not clutter the console for people who are developing using this SDK, perhaps we should remove the most verbose logging which comes up when using crypto?

(An alternative would be to use some logging library, such as the ones recommended in [this post](https://blog.risingstack.com/node-js-logging-tutorial/))

These are the relevant lines that I could find:

`src/crypto/algorithms/megolm.js`:
 * [#L68](https://github.com/matrix-org/matrix-js-sdk/blob/9c6973a46a7ce4e3b4001f1ce0bd8b1b782ee7a6/src/crypto/algorithms/megolm.js#L68)
 * [#L96](https://github.com/matrix-org/matrix-js-sdk/blob/9c6973a46a7ce4e3b4001f1ce0bd8b1b782ee7a6/src/crypto/algorithms/megolm.js#L96)

`src/crypto/algorithms/olm.js`:
 * [#L256](https://github.com/matrix-org/matrix-js-sdk/blob/9c6973a46a7ce4e3b4001f1ce0bd8b1b782ee7a6/src/crypto/algorithms/olm.js#L256)
 * [#L311](https://github.com/matrix-org/matrix-js-sdk/blob/9c6973a46a7ce4e3b4001f1ce0bd8b1b782ee7a6/src/crypto/algorithms/olm.js#L311)

`src/crypto/olmlib.js`:
 *   [#L65](https://github.com/matrix-org/matrix-js-sdk/blob/9c6973a46a7ce4e3b4001f1ce0bd8b1b782ee7a6/src/crypto/olmlib.js#L65)
 *   [#L229](https://github.com/matrix-org/matrix-js-sdk/blob/9c6973a46a7ce4e3b4001f1ce0bd8b1b782ee7a6/src/crypto/olmlib.js#L229)

`src/crypto/index.js`:
 * [#L356](https://github.com/matrix-org/matrix-js-sdk/blob/9c6973a46a7ce4e3b4001f1ce0bd8b1b782ee7a6/src/crypto/index.js#L356)
 *  [#L390](https://github.com/matrix-org/matrix-js-sdk/blob/9c6973a46a7ce4e3b4001f1ce0bd8b1b782ee7a6/src/crypto/index.js#L390)

`src/crypto/OlmDevice.js`:
 * [#L655](https://github.com/matrix-org/matrix-js-sdk/blob/9c6973a46a7ce4e3b4001f1ce0bd8b1b782ee7a6/src/crypto/OlmDevice.js#L655)

Ryuno-Ki:
I am not sure, what the guidelines regarding merges are here. Should I squash my changes?

There are more places where `console` is used. I can take another round if this gets accepted. Or open another ticket.

Since megolm isn't installed, I am not :100:% sure, this does not break. However, I used `loglevel` in the past and pretty sure, this should work :crossed_fingers: (hoping your CI will catch it if not).

The tests were broken before I even started, though :disappointed:

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -27,10 +27,10 @@ const log = require("loglevel");
 // Part of #332 is introducing a logging library in the first place.
 const DEFAULT_NAME_SPACE = "matrix";
 const logger = log.getLogger(DEFAULT_NAME_SPACE);
-log.setLevel(log.levels.DEBUG);
+logger.setLevel(log.levels.DEBUG);
 
 /**
  * Drop-in replacement for <code>console</code> using {@link https://www.npmjs.com/package/loglevel|loglevel}.
  * Can be tailored down to specific use cases if needed.
 */
-export logger;
+export default logger;
--- a/src/logger.js
+++ b/src/logger.js
@@ -17,4 +17,4 @@ log.setLevel(log.levels.DEBUG);
  * Drop-in replacement for <code>console</code> using {@link https://www.npmjs.com/package/loglevel|loglevel}.
  * Can be tailored down to specific use cases if needed.
 */
-module.exports = logger;
+export logger;
--- a/src/logger.js
+++ b/src/logger.js
@@ -18,4 +18,4 @@ log.setDefaultLevel(log.levels.WARN);
  * Drop-in replacement for <code>console</code> using {@link https://www.npmjs.com/package/loglevel|loglevel}.
  * Can be tailored down to specific use cases if needed.
 */
-module.exports.logger = logger;
+module.exports = logger;
--- a/src/logger.js
+++ b/src/logger.js
@@ -1,4 +1,3 @@
-"use strict";
 /**
  * @module logger
  */
--- a/src/logger.js
+++ b/src/logger.js
@@ -12,7 +12,7 @@ const log = require("loglevel");
 // Part of #332 is introducing a logging library in the first place.
 const DEFAULT_NAME_SPACE = "matrix";
 const logger = log.getLogger(DEFAULT_NAME_SPACE);
-log.setLevel(log.levels.WARN);
+log.setLevel(log.levels.DEBUG);
 
 /**
  * Drop-in replacement for <code>console</code> using {@link https://www.npmjs.com/package/loglevel|loglevel}.
--- a/src/logger.js
+++ b/src/logger.js
@@ -1,3 +1,19 @@
+/*
+Copyright 2018 André Jaenisch
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 /**
  * @module logger
  */
--- a/src/logger.js
+++ b/src/logger.js
@@ -1,0 +1,21 @@
+"use strict";
+/**
+ * @module logger
+ */
+const log = require("loglevel");
+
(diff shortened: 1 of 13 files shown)


Ryuno-Ki:
Oh, and where's the AUTHORS.rs mentioned in [CONTRIBUTION.rst](https://github.com/matrix-org/matrix-js-sdk/blob/master/CONTRIBUTING.rst#attribution)? I couldn't find it in the root of the project.

Ryuno-Ki:
Okay, _some_ of those failing tests are caused by me. Fixing them.

src/logger.js
```diff
@@ -0,0 +1,21 @@
...
+
+// This is to demonstrate, that you can use any namespace you want.
+// Namespaces allow you to turn on/off the logging for specific parts of the
+// application.
+// An idea would be to control this via an environment variable (on Node.js).
+// See https://www.npmjs.com/package/debug to see how this could be implemented
+// Part of #332 is introducing a logging library in the first place.
+const DEFAULT_NAME_SPACE = "matrix";
+const logger = log.getLogger(DEFAULT_NAME_SPACE);
+log.setDefaultLevel(log.levels.WARN);
```
dbkr: This is creating and exporting the 'matrix' logger but setting the default log level on the root logger, so I don't think the `setDefaultLevel` actually does anything here.

Also we probably want to maintain the current behaviour by default (for now, at least) but setting the default level to WARN would turn them off (`log.log` being a synonym for `log.debug`). I'd probably suggest using `log.info` for the current messages and setting the default level to DEBUG?

src/logger.js
```diff
@@ -0,0 +1,21 @@
+"use strict";
```
dbkr: Need a copyright notice for a new file (© you since you wrote it).

Also we don't need `use strict` anymore.

src/logger.js
```diff
@@ -0,0 +1,21 @@
...
+// Part of #332 is introducing a logging library in the first place.
+const DEFAULT_NAME_SPACE = "matrix";
+const logger = log.getLogger(DEFAULT_NAME_SPACE);
+log.setDefaultLevel(log.levels.WARN);
+
+/**
+ * Drop-in replacement for <code>console</code> using {@link https://www.npmjs.com/package/loglevel|loglevel}.
+ * Can be tailored down to specific use cases if needed.
+*/
+module.exports = logger;
```
dbkr: Not vital, but we generally use ES6 style exports now.

src/logger.js
Ryuno-Ki: From the docs:
> * A log.setDefaultLevel(level) method.
> This sets the current log level only if one has not been persisted and can’t be loaded. This is useful when initializing scripts; if a developer or user has previously called setLevel(), this won’t alter their settings.

I'll adjust it nevertheless to reduce confusion.

src/logger.js
Ryuno-Ki: I can't guarantee, ES6 export style will work (awaiting Travis log).

It feels … awkward to mix Node.js `require` (CommonJS) with ES6 `export` though.

Ryuno-Ki: Okay, according to the [last build](https://travis-ci.org/matrix-org/matrix-js-sdk/builds/445831102) I can run `npm run build` to test locally.
Sorry for the breakage!

Ryuno-Ki:
> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

But at least it doesn't break too early …

src/logger.js
dbkr: Yep, but I think they're still independent between loggers as far as I can tell? ie. `log.setDefaultLevel()` sets a level if one hasn't been set with `log.setLevel()`, and affects `log.log()`, but will not affect `logger.setLevel()` or `logger.log()`? Anyway, if they're all using `logger` this is fine.

dbkr:
Tests are now fixed too (node 11 was breaking them). Thanks!

src/logger.js
Ryuno-Ki: Thinking about creating a linter rule for it (separate ticket).

PR #764, opened 2018-10-25
Support `request` in the browser and support supplying servers to try in joinRoom()

turt2live:
See REACT_PR

### DIFF
--- a/browser-index.js
+++ b/browser-index.js
@@ -1,5 +1,5 @@
 var matrixcs = require("./lib/matrix");
-matrixcs.request(require("browser-request"));
+matrixcs.request(require("request"));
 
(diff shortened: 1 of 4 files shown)


ara4n:
lgtm, other than paranoia about what weird regressions switching HTTP client is going to introduce.  Particularly worried about whether timeout behaviour is going to change, or unexpected headers being introduced/removed by the lib etc.  Timeouts particularly won't be caught by E2E tests or UTs.  Might be worth checking for paranoia's sake?

turt2live:
I walked around the app and didn't encounter any obvious issues. I also knocked out synapse to get a connection refused error and that worked. Also knocked out the entire VPS synapse runs on and the timeouts appear to work.

PR #765, opened 2018-10-25
Use new stop() method on matrix-mock-request

dbkr:
To finish all pending flushes between tests. This stops the unit
tests from hanging on node 11 when run in certain combinations.

Requires https://github.com/matrix-org/matrix-mock-request/pull/6
(so will need a release of matrix-mock-request before merging)

For the record, one such combination was:

```
npx mocha specbuild/integ/devicelist-integ-spec.js specbuild/integ/matrix-client-methods.spec.js specbuild/integ/matrix-client-room-timeline.spec.js
```

This would complete all the tests but hang spinning a CPU core instead of exiting. Adding `--exit` (so mocha called process.exit once the tests were all done) made it exit. Adding `specbuild/integ/megolm-integ.spec.js` would cause it to hang before completing the tests.

### DIFF
--- a/spec/TestClient.js
+++ b/spec/TestClient.js
@@ -105,6 +105,7 @@ TestClient.prototype.start = function() {
  */
 TestClient.prototype.stop = function() {
     this.client.stopClient();
+    return this.httpBackend.stop();
 };
 
 /**
--- a/spec/TestClient.js
+++ b/spec/TestClient.js
@@ -102,6 +102,7 @@ TestClient.prototype.start = function() {
 
 /**
  * stop the client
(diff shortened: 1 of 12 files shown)


spec/TestClient.js
```diff
@@ -105,6 +105,7 @@ TestClient.prototype.start = function() {
  */
```
bwindels: lint error here:
```
/home/travis/build/matrix-org/matrix-js-sdk/spec/TestClient.js
  103:1  error  Missing JSDoc @returns for function  valid-jsdoc
```

PR #767, opened 2018-10-26
loglevel should be a normal dependency

dbkr:
rather than a dev dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "src"
   ],
   "dependencies": {
+    "loglevel": "1.6.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -52,6 +52,7 @@
     "src"
   ],
   "dependencies": {
+    "loglevel": "1.6.1",
```
t3chguy: I think these are meant to be alphabetical 😝

Ryuno-Ki: You could throw a script at it, which sorts a JSON by keys in alphabetical order ;-)

PR #770, opened 2018-10-30
Manually construct query strings for browser-request instances

turt2live:
Because `request` just doesn't work for us in the browser, but `browser-request` is fine despite us having to do our own query strings.

Fixes https://github.com/vector-im/riot-web/issues/7620


turt2live:
A proper solution would be https://github.com/vector-im/riot-web/issues/7634 however that is taking way too much time.

PR #771, opened 2018-10-30
Broken example in docs. Fixes #421

### LINKED ISSUE #421: Fix generated documentation
Public methods are not visible in the documentation. eg.
https://matrix-org.github.io/matrix-js-sdk/0.7.6/module-models_event.MatrixEvent.html
https://matrix-org.github.io/matrix-js-sdk/0.7.6/models_event.js.html#line132

The [landing page for the documentation](https://matrix-org.github.io/matrix-js-sdk/0.7.6/index.html) has a link labelled "working example" which currently points to https://matrix-org.github.io/matrix-js-sdk/0.7.6/examples/browser which is a a 404 page.

Ryuno-Ki:
This is what I think is needed to repair the link to the broken working examples.

Why?

Right now, we use `npm run gendoc` in three places. Only one deals with gh-pages, so I left `travis.sh` and `jenkins.sh` untouched.

In `release.sh` we do the following: https://github.com/matrix-org/matrix-js-sdk/blob/v0.12.1/release.sh#L266

```sh
if [ -z "$skip_jsdoc" ]; then
    echo "generating jsdocs"
    npm run gendoc

    echo "copying jsdocs to gh-pages branch"
    git checkout gh-pages
    git pull
    cp -a ".jsdoc/matrix-js-sdk/$release" .
    perl -i -pe 'BEGIN {$rel=shift} $_ =~ /^<\/ul>/ && print
        "<li><a href=\"${rel}/index.html\">Version ${rel}</a></li>\n"' \
        $release index.html
    git add "$release"
    git commit --no-verify -m "Add jsdoc for $release" index.html "$release"
fi
```

So here we only copy the jsdocs folder to gh-pages. jsdocs gets generated from `src` only:
https://github.com/matrix-org/matrix-js-sdk/blob/v0.12.1/package.json#L12

So that's why I included the `examples` directory also.

However, two things I want to highlight:
1. `examples/browser/lib/matrix.js` and `examples/voip/lib/matrix.js` are symlinks on my Linux box.
2. `README.md` of all three examples are pushed to GH-Pages with this patch as well. Shall I strip them out?

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -260,9 +260,10 @@ if [ -z "$skip_jsdoc" ]; then
     echo "generating jsdocs"
     npm run gendoc
 
-    echo "copying jsdocs to gh-pages branch"
+    echo "copying jsdocs and examples to gh-pages branch"
     git checkout gh-pages
     git pull
+    cp -ar "examples/" .
     cp -a ".jsdoc/matrix-js-sdk/$release" .
     perl -i -pe 'BEGIN {$rel=shift} $_ =~ /^<\/ul>/ && print
         "<li><a href=\"${rel}/index.html\">Version ${rel}</a></li>\n"' \


jryans:
Thank you for your contribution! 😁 

> However, two things I want to highlight:
> 
>     1. `examples/browser/lib/matrix.js` and `examples/voip/lib/matrix.js` are symlinks on my Linux box.

I assume for the script to be available on the docs site, then you should also copy the `dist` directly as well. Does that seem correct?

>     2. `README.md` of all three examples are pushed to GH-Pages with this patch as well. Shall I strip them out?

I would go with whatever seems best to you.

Ryuno-Ki:
You're welcome.

> I assume for the script to be available on the docs site, then you should also copy the `dist` directly as well. Does that seem correct?

I am not sure, whether I read the bash script correctly, but is `dist` guaranteed to be present? (If I judge that correctly the `dodist` part should take care of that).

I don't have my pc in front of me at the moment. Shall I just `cp -ar` the whole dist directory?

> I would go with whatever seems best to you.

They're not doing any harm fwiw. We can remove them later on if it confuses people.

jryans:
Hmm, looks like this fell through the cracks... 😓 @Ryuno-Ki, are you still interested in continuing here?

Ryuno-Ki:
That was totally not on my radar, too!
Let me skim, what I initially did. (My gut feeling is that so much has changed, that it would be better to close this and start over).

Ryuno-Ki:
@jryans Yupp, the `release.sh` looks very different. Let's close this. I have no hard feelings.
Keep on rocking Matrix, will ya? :)

Issue #130, opened 2016-04-21
loginWithPassword gives access token error


McOmghall:
I'm using the following code to connect to the matrix.org matrix server

```
var sdk = require('matrix-js-sdk')
var client = sdk.createClient("http://matrix.org")
var login = client.loginWithPassword("user", "pass")
```

Login is then set as the following object when the request resolves:

```
    { state: 'fulfilled',
  value: 
   { access_token: '<token>',
     refresh_token: '<token>',
     home_server: 'matrix.org',
     user_id: '@myAlias:matrix.org' } }
```

I assume this means login has finished correctly, however doing 

```
client.on("sync", function (state, prevState, result) {
  console.log("On sync: State %s -> %s", prevState, state)
  console.log("On sync: Result: %j", result)
  console.log("On sync: Client: %j", Object.keys(client))
})

client.startClient()
```

Gives the following results, never going out of the ERROR state (server responds with a 403, basically), meaning the client object is not updated with the access key:

```
On sync: State null -> ERROR
On sync: Result: {"error":{"errcode":"M_MISSING_TOKEN","name":"M_MISSING_TOKEN","message":"Missing access token.","data":{"errcode":"M_MISSING_TOKEN","error":"Missing access token."},"httpStatus":403}}
On sync: Client: ["baseUrl","idBaseUrl","store","sessionStore","accountKey","deviceId","scheduler","clientRunning","credentials","_http","callList","_supportsVoip","_syncingRetry","_syncApi","_peekSync","_isGuest","_ongoingScrollbacks","_txnCtr","timelineSupport","_events","_eventsCount","_clientOpts"]
```

Tested in node v5.10.1 and npm 3.8.3 over Linux Mint and Windows 7. 

Am I missing something here?

McOmghall:
Solution at the moment goes like this
```
var sdk = require("matrix-js-sdk")
var login = sdk.createClient("https://matrix.org").loginWithPassword("user", "pass", function (err, res) {
  console.log("On login: Err %j, res %j", err, res)

  var authedClient = sdk.createClient({
      baseUrl: "https://matrix.org",
      accessToken: res.access_token,
      userId: res.user_id
  })

  authedClient.on("sync", function (state, prevState, result) {
    console.log("On sync: State %s -> %s", prevState, state)
    console.log("On sync: Result: %j", result)
    console.log("On sync: Client: %j", Object.keys(authedClient))
  })

  authedClient.startClient()
})
```

[...]

PR #778, opened 2018-11-04
Set access_token and user_id after login in with username and password.

rxl881:
Set access token and user ID after successfully logging in with username and password.

This fixes - https://github.com/matrix-org/matrix-js-sdk/issues/130.

### DIFF
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -277,7 +277,21 @@ MatrixBaseApis.prototype.loginWithPassword = function(user, password, callback) 
     return this.login("m.login.password", {
         user: user,
         password: password,
-    }, callback);
+    }, (error, response) => {
+        if (response && response.access_token) {
+            this._http.opts.accessToken = response.access_token;
+        }
+
+        if (response && response.user_id) {
+            this.credentials = {
+                userId: response.user_id,
+            };
+        }
+
+        if(callback) {
+            callback(error, response);
+        }
+    });
 };
 
 /**


rxl881:
> This is probably a major version bump though as it's technically changing behaviour in a way that could break things (if someone is relying on login not changing the state of the client). Plus it's in base-apis which really ought to be just the rest calls rather than anything higher level.

Yep, this makes sense (being a major version bump). Is there an impact on the timeline of getting this rolled out, if we do that?

It would be good to get this out quickly, if at all possible, not just for the things that I am currently working on but as the behaviour seems quite broken at the moment.

E.g. you have things like https://matrix.org/docs/guides/usage-of-the-matrix-js-sdk explaining how to log in with a username and password:

selection_012

However, this documentation doesn't explain how you can then go on to use that to continue performing actions with the auth token. If this isn't broken, it the intended behaviour really to throw away the existing client and create a new one (like the fix suggested in #130 ?).

> More specifically, would it be a good idea to set either both or neither the user ID and password?

I'm happy to do this if you think that it helps?  Would doing so help to get this out? If not, what's the best way forward? 

Cheers.

src/base-apis.js
```diff
@@ -277,7 +277,21 @@ MatrixBaseApis.prototype.loginWithPassword = function(user, password, callback)
     return this.login("m.login.password", {
```
bwindels: Should we be doing this in the `login` method instead to keep the API somewhat consistent in between these two methods?

src/base-apis.js
```diff
@@ -277,7 +277,21 @@ MatrixBaseApis.prototype.loginWithPassword = function(user, password, callback)
...
+            this._http.opts.accessToken = response.access_token;
+        }
+
+        if (response && response.user_id) {
+            this.credentials = {
+                userId: response.user_id,
+            };
+        }
+
+        if(callback) {
```
bwindels: you'll need a space after `if` to satisfy the linter

bwindels: Might look nicer with promises/async await instead of a callback here:
```
... = async function(user, password, callback) {
  const response = await this.login(..., callback);
  if (response.access_token) {
   ...
  }
  if (response.user_id) {
   ...
  }
  return response;
}
```

bwindels: Hmm, not sure what is the right thing to do here actually. The promise is resolved before the user callback in `http-api.js:848`, but that will be run as a microtask. Right after, the user callback is called (synchronously). So the user callback code will run first, so setting the access token/user id in there is probably safest to make sure usages of the sdk using callbacks rather than promises don't break.

src/base-apis.js
rxl881: Good point. As discussed, I've refactored the code to the base 'login' method.

src/base-apis.js
rxl881: Thanks for catching this (not sure why eslint didn't).

rxl881:
@dbkr, @bwindels -- Thanks very much for the feedback on this. Hopefully all of that is now addressed, can you please take another look?

Issue #440, opened 2017-05-22
Add joined_rooms and joined_members endpoints


turt2live:
Currently a spec omission in https://github.com/matrix-org/matrix-doc/issues/734 because the endpoints were added in https://github.com/matrix-org/synapse/pull/1680

It would be nice if the two new endpoints from https://github.com/matrix-org/synapse/pull/1680 were supported in the js-sdk.

t3chguy:
BUMP
no longer a spec-omission

turt2live:
This happened: https://github.com/matrix-org/matrix-js-sdk/pull/779

PR #779, opened 2018-11-04
Add function to get currently joined rooms.

rxl881:


turt2live:
This fixes half of https://github.com/matrix-org/matrix-js-sdk/issues/440 - can it be expanded to also include the other half?

rxl881:
@turt2live -- Sure. That should do the trick.

PR #793, opened 2018-11-28
Bwindels/serverpresence

bwindels:


bwindels:
this is the code that we would need to shown only the first and last disconnect/connect alert, but just showing all for now, so not needed.

bwindels:
In progress because to code is commented out for now, but should work.

t3chguy:
@bwindels is this still relevant?

PR #795, opened 2018-11-29
cross-signing

uhoreg:
- [X] check cross-signing data
- [x] invalidate caches when devices change
- [x] sign and send attestations

The js-sdk part of https://github.com/vector-im/riot-web/issues/2714


src/crypto/index.js
```diff
@@ -1192,6 +1240,217 @@ Crypto.prototype.decryptEvent = function(event) {
...
+                    devices.get(fromUserId).get(fromDeviceId)
+                        .keys["ed25519:" + fromDeviceId],
+                );
+            } catch (e) {
+                continue;
+            }
+            if (att.state === "verified") {
+                verified = true;
+            } else if (att.state === "revoked") {
+                // FIXME: need to figure out exactly how revocations work
```
dbkr: is this such that this can be merged as-is? seems pretty reasonable I think?

src/crypto/index.js
```diff
@@ -1192,6 +1240,217 @@ Crypto.prototype.decryptEvent = function(event) {
...
+            }
+        }
+        return verified;
+    }
+
+    // traverse the attestations graph for the given user, starting from the
+    // trusted devices.  Will modify the trusted device map by adding the
+    // devices that can be reached by following the edges.
+    async function traverseAttestations(edges, trustedDevices, userId, devices, value) {
+        const stack = [...trustedDevices.keys()];
```
dbkr: isn't this just `const stack = trustedDevices.keys();`?

uhoreg:
dropping this, because we're doing something different for cross-signing

PR #797, opened 2018-12-03
Add support for `fetch` and cbor

Half-Shot:
This does quite a few sweeping changes to support `fetch`, `CBOR encoding` and the `Accept` header which is part of https://github.com/matrix-org/matrix-doc/pull/1740.  **Please do not merge or review this just yet** 

Anyway, this PR seems to work perfectly fine with a few issues noted below.

Notable failings:
- `supportsCbor` is always true, and not detected. Ergo, this PR breaks support for regular JSON servers.
- I had to swap `request` for fetch because I could not get ArrayBuffer responses out of it which is important.
- `borc` is the library we are using for `cbor` and is wasmy, untested, and seems to work most of the time.

---

### DIFF
--- a/src/http-api.js
+++ b/src/http-api.js
@@ -19,6 +19,8 @@ limitations under the License.
  * @module http-api
  */
 import Promise from 'bluebird';
+import borc from "borc";
+import queryString from 'qs';
 const parseContentType = require('content-type').parse;
 
 const utils = require("./utils");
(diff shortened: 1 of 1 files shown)


bradjones1:
The related MSC hasn't gone anywhere for a few years but `fetch` support would be good to have. Might be advisable to break out the CBOR-related components to make this more atomic.

gsouquet:
It appears that native `fetch` support has been merged in Node.js quite recently, https://github.com/nodejs/node/pull/41749

3nprob:
> It appears that native `fetch` support has been merged in Node.js quite recently, [nodejs/node#41749](https://github.com/nodejs/node/pull/41749)

Noteworthy that the v18+ implementation is provided by [`undici`](https://github.com/nodejs/undici), so using that as a polyfill should ensure minimal differences between the versions.

t3chguy:
Closing this as it no longer applies cleanly

3nprob:
Is the current intention still to go with fetch APIs across the board (last update I can see on the matter) or is still still up for reconsideration?

t3chguy:
Yes, fetch given native browser and node support

PR #799, opened 2018-12-04
Support .well-known autodiscovery in the js-sdk

turt2live:
*Note to the reviewer*: This also includes a change for supporting ES6 modules. See https://github.com/matrix-org/matrix-js-sdk/pull/799/commits/7e0752b48a9a0fac3992c679e63effac6a01eca3 for details.

----

It's much more useful here than in the react-sdk as it can be reused by more applications. This is also required to make the react-sdk a little easier to manage .well-known lookups as soon it'll be doing it in several places. 

Automatic discovery is an abstract concept in the spec and could include more than .well-known in the future, so this is made to be generic enough to support future mechanisms and other resources to discover. There's also a ton of comments (more than normally needed) as people may wish to use this as a reference in their own implementation and it doesn't hurt to explain what everything is doing.

This abstraction ends up manifesting itself as returning a .well-known response-like object. Although the object we return is our own format, we mimic .well-known for maximum compatibility.

Many of the functions are air lifted from the react-sdk and modified to work within the confines of the js-sdk.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "matrix-js-sdk",
-  "version": "0.14.1",
+  "version": "0.13.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -296,6 +296,17 @@
       "integrity": "sha1-MU3QpLM2j609/NxU7eYXG4htrzw=",
       "dev": true
     },
+    "align-text": {
+      "version": "0.1.4",
+      "resolved": "https://registry.npmjs.org/align-text/-/align-text-0.1.4.tgz",
+      "integrity": "sha1-DNkKVhCT810KmSVsIrcGlDP60Rc=",
+      "dev": true,
+      "requires": {
+        "kind-of": "^3.0.2",
+        "longest": "^1.0.1",
+        "repeat-string": "^1.5.2"
+      }
+    },
     "amdefine": {
       "version": "1.0.1",
       "resolved": "https://registry.npmjs.org/amdefine/-/amdefine-1.0.1.tgz",
@@ -310,7 +321,7 @@
     },
     "ansi-escapes": {
       "version": "1.4.0",
-      "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-1.4.0.tgz",
+      "resolved": "http://registry.npmjs.org/ansi-escapes/-/ansi-escapes-1.4.0.tgz",
       "integrity": "sha1-06ioOzGapneTZisT52HHkRQiMG4=",
       "dev": true
     },
@@ -447,7 +458,7 @@
         },
         "util": {
           "version": "0.10.3",
-          "resolved": "https://registry.npmjs.org/util/-/util-0.10.3.tgz",
+          "resolved": "http://registry.npmjs.org/util/-/util-0.10.3.tgz",
           "integrity": "sha1-evsa/lCAUkZInj23/g7TeTNqwPk=",
           "dev": true,
           "requires": {
@@ -1553,6 +1564,12 @@
       "integrity": "sha1-r6uWJikQp/M8GaV3WCXGnzTjUMo=",
       "dev": true
     },
+    "camelcase": {
+      "version": "1.2.1",
+      "resolved": "https://registry.npmjs.org/camelcase/-/camelcase-1.2.1.tgz",
+      "integrity": "sha1-m7UwTS4LVmmLLHWLCKPqqdqlijk=",
+      "dev": true
+    },
     "caseless": {
       "version": "0.12.0",
       "resolved": "https://registry.npmjs.org/caseless/-/caseless-0.12.0.tgz",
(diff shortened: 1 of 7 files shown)


turt2live:
Looks like the build is very unhappy about ES6 somewhere - will look into this before requesting review.

dbkr:
So the thing with having to update uglify here is that a recursive dependency (tr46 of whatwg-url) has exported ES6 rather than ES5. It looks like we're going to have to do something about this as this is the second time this has come up recently. Doing this will mean we will, in turn, start exporting ES6 from js-sdk which will cause the same problem for any other libraries that depend on us.

We should make a conscious decision about what to do - I've filed https://github.com/matrix-org/matrix-js-sdk/issues/803 for this

src/autodiscovery.js
```diff
@@ -0,0 +1,379 @@
...
+     *   error: The actual Error, if one exists.
+     * @param {string} url The URL to fetch a JSON object from.
+     * @return {Promise<object>} Resolves to the returned state.
+     * @private
+     */
+    static async _fetchWellKnownObject(url) {
+        return new Promise(function(resolve, reject) {
+            const request = require("./matrix").getRequest();
+            if (!request) throw new Error("No request library available");
+            request(
```
bwindels: doesn't request return a Promise you could await? Then all the code below can be written in a synchronous style without calling resolve by hand below ... Either that, or remove the async keyword.

src/autodiscovery.js
```diff
@@ -0,0 +1,379 @@
...
+     * If valid, the URL will also be stripped of any trailing slashes.
+     * @param {string} url The potentially invalid URL to sanitize.
+     * @return {string|boolean} The sanitized URL or a falsey value if the URL is invalid.
+     * @private
+     */
+    static _sanitizeWellKnownUrl(url) {
+        if (!url) return false;
+
+        try {
+            const parsed = parseURL(url);
```
bwindels: I believe that whatwg url parsing is supported in all browsers we support, so you could just use the [`URL` constructor](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) here, [like so](https://url.spec.whatwg.org/#example-5434421b). That might be what the whatwg-url dependency will fall back to, but it still pulls in the dependency...

package.json
```diff
@@ -14,7 +14,7 @@
     "start:watch": "babel -s -w --skip-initial-build -d lib src",
     "start:init": "babel -s -d lib src",
     "clean": "rimraf lib dist",
-    "build": "babel -s -d lib src && rimraf dist && mkdir dist && browserify -d browser-index.js | exorcist dist/browser-matrix.js.map > dist/browser-matrix.js && uglifyjs -c -m -o dist/browser-matrix.min.js --source-map dist/browser-matrix.min.js.map --in-source-map dist/browser-matrix.js.map dist/browser-matrix.js",
+    "build": "babel -s -d lib src && rimraf dist && mkdir dist && browserify -d browser-index.js | exorcist dist/browser-matrix.js.map > dist/browser-matrix.js && uglifyjs -c -m -o dist/browser-matrix.min.js --source-map \"filename='dist/browser-matrix.min.js.map',content='dist/browser-matrix.js.map'\" dist/browser-matrix.js",
```
bwindels: You were discussing this with Dave at the standup, but I forgot what the conclusion was. Does anything need to be done here still?

bwindels: Dave comment is above here, so never mind this :)

src/autodiscovery.js
bwindels: Although not a solution, it would also allow us to ignore the export ES6 vs ES5/uglify-es problem for a bit longer.

turt2live: Node doesn't support the URL constructor though, which is why the dependency is pulled in. I'd love to use the URL constructor, but people do use the js-sdk outside of a web browser.

src/autodiscovery.js
turt2live: I'll take off the async keyword - request does not return a promise, annoyingly.

turt2live:
I'm probably going to change this such that it spits out ES5 as a bundle so we can punt the ES5/6 issue down the road. It might not be pretty though, sorry.

src/autodiscovery.js
bwindels: Seems supported since 6.x in node?
https://nodejs.org/docs/latest-v6.x/api/url.html#url_the_whatwg_url_api

src/autodiscovery.js
bwindels: :+1:

src/autodiscovery.js
turt2live: I hate it when the internet lies to me :/

I'll give it a go

turt2live:
@bwindels please take another look. I've changed the URL handling, and as a consequence removed all the ES6 stuff.

src/autodiscovery.js
```diff
@@ -18,7 +18,7 @@ limitations under the License.
 
 import Promise from 'bluebird';
 const logger = require("./logger");
-import { parseURL } from "whatwg-url";
+import { URL as NodeURL } from "url";
```
bwindels: No need to import it, URL is available on the global object in node, just like in the browser.

package-lock.json
```diff
@@ -310,7 +321,7 @@
     },
     "ansi-escapes": {
       "version": "1.4.0",
-      "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-1.4.0.tgz",
+      "resolved": "http://registry.npmjs.org/ansi-escapes/-/ansi-escapes-1.4.0.tgz",
```
bwindels: Hmm, any idea what happened here? Seems odd to switch to an unencrypted connection ...

src/autodiscovery.js
turt2live: It's not, I tried and it yelled at me.

package-lock.json
turt2live: npm keeps flipping them and I don't know why :(

bwindels: (the overall diff doesn't change the file, so not an issue)

PR #802, opened 2018-12-06
Emit an event when room notification counts change

turt2live:


turt2live:
This is actually super unreliable.

PR #805, opened 2018-12-12
Support custom status messages

turt2live:
Part of https://github.com/vector-im/riot-web/issues/1528

Status messages are short strings of text that users can set to be visible to other users. The js-sdk bits here are for setting and reading the status message for a given user.

Status messages are completely unrelated to the presence status message in that they do not expire and have a different intent. Presence status messages are intended to be read as "Online reading documentation" whereas status messages would be just "Reading documentation" (no associative presence). 

This currently uses a state event in direct chats to represent the status. It is intended that in the future there will be extensible profiles which can better communicate this information. Because this is a feature we're trying out, the event type is `im.vector.user_status` instead of something in the Matrix namespace. The API here should be abstract enough that changing to extensible profiles in the future should be easy.

----

**`im.vector.user_status` spec**:

The user status event is a state event with a state key of the user ID the status is for with a single `status` field in the content. The `status` field is a string of preferably limited length (<60 characters, not enforced) where falsey values (empty, null, undefined, not set, etc) equate to "no status" whereas a string with content represents the user's status. 

It is not intended at this time to support per-room status messages, despite using a state event. This is why this js-sdk PR happily overwrites the user's status message when it sees an `im.vector.user_status` state event. Simply seeing the state event in any room will change that user's status message everywhere.

### DIFF
--- a/src/models/user.js
+++ b/src/models/user.js
@@ -39,6 +39,9 @@ limitations under the License.
  *                when a user was last active.
  * @prop {Boolean} currentlyActive Whether we should consider lastActiveAgo to be
  *               an approximation and that the user should be seen as active 'now'
+ * @prop {string} statusMessage The status message for the user, if known. This is
+ *                different from the presenceStatusMsg in that this is not tied to
+ *                the user's presence, and should be represented differently.
  * @prop {Object} events The events describing this user.
  * @prop {MatrixEvent} events.presence The m.presence event for this user.
  */
@@ -46,6 +49,7 @@ function User(userId) {
     this.userId = userId;
     this.presence = "offline";
     this.presenceStatusMsg = null;
+    this.statusMessage = "";
     this.displayName = userId;
     this.rawDisplayName = userId;
     this.avatarUrl = null;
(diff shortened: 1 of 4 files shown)


Half-Shot:
I have reservations about including more `im.*` specific unspecced things into the js-sdk :/

turt2live:
The event itself should be fairly well abstracted so it can be replaced with extensible profiles in the future, as mentioned in the description. This is intentionally being done without a spec proposal (hence the vector namespace) to get the feature in front of people and iterate upon it while extensible profiles are being worked on.

Half-Shot:
To clarify, my worry is people will see `setStatusMessage` and believe it is the defacto way to set a status message in Matrix, while this is a im.vector brand feature. I don't really have an issue with this ending up in the spec as a feature. Anyway, we sort of discussed this oob and I expect everyone is aware of this already.

turt2live:
@Half-Shot I've made the entire API be prefixed with `_unstable_` in https://github.com/matrix-org/matrix-js-sdk/pull/805/commits/08b3dfa3b5b5d0b63272f0b80b9fdd88d0795c45 - does this alleviate your concerns? Also, if you have suggestions for making it not as ugly please let me know.

src/models/user.js
```diff
@@ -39,13 +39,17 @@ limitations under the License.
  *                when a user was last active.
  * @prop {Boolean} currentlyActive Whether we should consider lastActiveAgo to be
  *               an approximation and that the user should be seen as active 'now'
+ * @prop {string} statusMessage The status message for the user, if known. This is
```
bwindels: `_unstable_statusMessage` for now?

Half-Shot:
>@Half-Shot I've made the entire API be prefixed with _unstable_ in 08b3dfa - does this alleviate your concerns? Also, if you have suggestions for making it not as ugly please let me know.

Thank you :). I don't really know how we could make this better, than perhaps `unstableGetBlah...`

turt2live:
tbh I kinda like `_unstable_` because it makes searching for this stuff a little easier, and having slept on it it doesn't look as bad as it did last night.

PR #807, opened 2018-12-13
Redesign: SDK support for jump to timestamp

bwindels:


turt2live:
the react-sdk side was closed because it apparently won't get merged soon

PR #810, opened 2018-12-18
Conform to base-x >= 3.0.5 API

jryans:
The library `base-x` now expects an explicit `Buffer` instance, so let's change
to that to make things easier.  (Alternatively, we could pin to a previous
version, but it's a small tweak to make.)

Signed-off-by: J. Ryan Stinnett <jryans@gmail.com>


jryans:
Yes, tests pass locally and I manually tested the Riot feature that uses this code path.

Issue #803, opened 2018-12-10
Do something about ES6 exporting


dbkr:
Recently we're starting to want to pull in libraries that export, or have dependencies that export, ES6. As far as I can tell, there are no guidelines or structure whatsoever in NPM / the JS ecosystem on what language NPM modules are exported as, and now that any vaguely recent node version and all modern browsers support ES6, the answer increasingly seems to be ES6 rather than ES5.

Our options are:
 * Continue to export ES5 by babeling everything, dependencies and all (slow)
 * Continue to export ES5 by babeling only specific dependencies (major faff to keep the list of babel-ed depds up to date)
 * Switch to exporting ES6

jryans:
I would suggest switching to exporting ES6, which is well supported in recent browser and node versions.

bwindels:
(partial) list of projects that depend on js-sdk, whose build we could potentially break: https://www.npmjs.com/package/matrix-js-sdk
Thinking if it would be ok to switch to exporting ES6 if we'd release it as a major version change (0.x)...?

turt2live:
+1 to ES6 and bumping the version

dbkr:
Seems reasonable: the implication of this, of course, is that we stop babel-ing our source files (unless we need to babel down to ES6) since if we are exporting ES6 we might as well export everything in ES6.

jkasun:
Can I hop in to this?

jryans:
@dbkr has made an attempt a while back in https://github.com/matrix-org/matrix-js-sdk/pull/817. Perhaps building on that and fixing the remaining issues would be a way forward here.

stalniy:
I’d suggest to use rollup with its plugins. Also it’ll allow to bundle browser version inside `npm` package. 

Currently we need to download it from github releases which is really inconvenient. 

Another benefit from ES6 version is that during development of client web app, unused ask functionality will be shakes out

jryans:
https://github.com/matrix-org/matrix-js-sdk/pull/931 has added a tiny amount of ES 6 (via the dependency `base-x`) which will be released as 2.0.0 because of the breaking change for syntax.

This issue remains though, because we haven't yet started exporting ES 6 for the SDK's own code. When we do so, it will no longer be a breaking change, so most likely a MINOR version will be appropriate.

PR #817, opened 2019-01-07
Export ES6 from js-sdk

dbkr:
This is a somewhat experimental PR to make the build process target
ES6 rather than ES5 given ES6 is now widely supported.

https://github.com/matrix-org/matrix-js-sdk/issues/803

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "start:watch": "babel -s -w --skip-initial-build -d lib src",
     "start:init": "babel -s -d lib src",
     "clean": "rimraf lib dist",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -85,8 +83,10 @@
     "rimraf": "^2.5.4",
     "source-map-support": "^0.4.11",
     "sourceify": "^0.1.0",
-    "uglify-js": "^2.8.26",
-    "watchify": "^3.11.0"
+    "uglify-es": "^3.3.9",
+    "watchify": "^3.11.0",
+    "webpack": "^4.28.3",
```
jryans: Does this package actually use Webpack? It's added here, but it doesn't seem it's used before or after your changes.

dbkr: ah, no it doesn't but I forgot to remove it.

dbkr:
Tests are currently failing because mocha is exiting with status 1 for no reason. Sigh.

dbkr:
Fixed one error but it's still unhappy about the abuses we're committing in order to do inheritance in an ES5 world. We probably need to either find the specific thing we're doing wrong here or bite the bullet & port it to real classes.

t3chguy:
@dbkr is this still relevant?

PR #818, opened 2019-01-08
Key verification

uhoreg:
Implements short-authentication-string verification and some QR-code verification, though bi-directional QR-code verification isn't complete.  SAS verification requires https://github.com/matrix-org/olm-backup/pull/80

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -52,6 +52,9 @@ const defaultVerificationMethods = {
     [SAS.NAME]: SAS,
 };
 
+/**
+ * verification method names
+ */
 export const verificationMethods = {
     QR_CODE_SCAN: ScanQRCode.NAME,
     QR_CODE_SHOW: ShowQRCode.NAME,
@@ -1592,6 +1595,12 @@ Crypto.prototype._onRoomKeyEvent = function(event) {
     alg.onRoomKeyEvent(event);
 };
 
+/**
+ * Handle a key verification request event.
+ *
+ * @private
+ * @param {module:models/event.MatrixEvent} event verification request event
+ */
 Crypto.prototype._onKeyVerificationRequest = function(event) {
     const content = event.getContent();
     if (!("from_device" in content) || typeof content.from_device !== "string"
@@ -1602,6 +1611,7 @@ Crypto.prototype._onKeyVerificationRequest = function(event) {
         // ignore event if malformed
         return;
     }
+
     const now = Date.now();
     if (now < content.timestamp - (5 * 60 * 1000)
         || now > content.timestamp + (10 * 60 * 1000)) {
@@ -1609,6 +1619,7 @@ Crypto.prototype._onKeyVerificationRequest = function(event) {
         logger.log("received verification that is too old or from the future");
         return;
     }
+
     const sender = event.getSender();
     if (this._verificationTransactions.has(sender)) {
         if (this._verificationTransactions.get(sender).has(content.transaction_id)) {
@@ -1622,7 +1633,8 @@ Crypto.prototype._onKeyVerificationRequest = function(event) {
                 this._verificationTransactions.get(sender).get(content.transaction_id)
                     .verifier.cancel(err);
             } else {
-                this._verificationTransactions.get(sender).get(content.transaction_id).reject(err);
+                this._verificationTransactions.get(sender).get(content.transaction_id)
+                    .reject(err);
                 this.sendToDevice("m.key.verification.cancel", {
                     [sender]: {
                         [content.from_device]: err.getContent(),
@@ -1636,6 +1648,7 @@ Crypto.prototype._onKeyVerificationRequest = function(event) {
         this._verificationTransactions.set(sender, new Map());
     }
 
+    // determine what requested methods we support
     const methods = [];
     for (const method of content.methods) {
         if (typeof method !== "string") {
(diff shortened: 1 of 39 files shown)


src/crypto/verification/SAS.js
```diff
@@ -0,0 +1,248 @@
...
+            const sas = olmSAS.generate_bytes(sasInfo, 5).reduce((acc, elem) => {
+                return acc + ('0' + elem.toString(16)).slice(-2);
+            }, "");
+            const macInfo = "MATRIX_KEY_VERIFICATION_MAC"
+                  + this._baseApis.getUserId() + this._baseApis.deviceId
+                  + this.userId + this.deviceId
+                  + this.transactionId;
+            const verifySAS = new Promise((resolve, reject) => {
+                const keyId = `ed25519:${this._baseApis.deviceId}`;
+                const keyMac = olmSAS.calculate_mac(macInfo, this._baseApis.getDeviceEd25519Key());
```
dbkr: are these params the right way around?

src/crypto/index.js
```diff
@@ -1426,6 +1544,233 @@ Crypto.prototype._onRoomKeyEvent = function(event) {
...
+            this.sendToDevice(
+                "m.key.verification.cancel", {
+                    [sender]: {
+                        [deviceId]: err.getContent(),
+                    },
+                },
+            );
+            return;
+        } else {
+            /* TODO:
```
ara4n: should this be to-done?

src/crypto/index.js
```diff
@@ -68,16 +89,33 @@ const MIN_FORCE_SESSION_INTERVAL_MS = 60 * 60 * 1000;
...
     this._sessionStore = sessionStore;
     this._userId = userId;
     this._deviceId = deviceId;
     this._clientStore = clientStore;
     this._cryptoStore = cryptoStore;
     this._roomList = roomList;
+    this._verificationMethods = {};
+    if (verificationMethods) {
+        for (const method of verificationMethods) {
+            if (typeof method === "string") {
```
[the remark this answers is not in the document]
uhoreg: I added both because passing a string would be the way to use the built-in methods, but applications could define their own verification methods and use them by passing in a class.

src/crypto/verification/SAS.js
uhoreg: fixed

src/crypto/index.js
uhoreg: The `TwoPartVerification` class isn't done yet, and it may end up being more complicated than this, so not quite done yet.

src/crypto/index.js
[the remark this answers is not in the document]
dbkr: ah, I see

src/crypto/verification/Base.js
```diff
@@ -0,0 +1,171 @@
...
+    handleEvent(e) {
+        if (this._done) {
+            return;
+        } else if (e.getType() === this._expectedEvent) {
+            this._expectedEvent = undefined;
+            this._rejectEvent = undefined;
+            this._resolveEvent(e);
+        } else {
+            this._expectedEvent = undefined;
+            const exception = new Error("Unexected message");
```
dbkr: Unexpected spelling

dbkr: also would be v useful for debugging if it printed an expected/actual event type

src/crypto/index.js
```diff
@@ -1426,6 +1544,233 @@ Crypto.prototype._onRoomKeyEvent = function(event) {
...
+        || !("transaction_id" in content) || typeof content.from_device !== "string"
+        || !("methods" in content) || !(content.methods instanceof Array)
+        || !("timestamp" in content) || typeof content.timestamp !== "number") {
+        logger.warn("received invalid verification request from " + event.getSender());
+        // ignore event if malformed
+        return;
+    }
+    const now = Date.now();
+    if (now < content.timestamp - (5 * 60 * 1000)
+        || now > content.timestamp + (10 * 60 * 1000)) {
```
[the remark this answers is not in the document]
uhoreg: Possibly.  This would be a spec question.  I'll leave it as-is for now, and we can discuss it in the spec.

src/crypto/index.js
```diff
@@ -1426,6 +1544,233 @@ Crypto.prototype._onRoomKeyEvent = function(event) {
...
+                // FIXME: make sure we're in a two-part verification, and the start matches the second part
+            }
+        }
+        this._baseApis.emit("crypto.verification.start", verifier);
+    }
+};
+
+Crypto.prototype._onKeyVerificationMessage = function(event) {
+    const sender = event.getSender();
+    const transactionId = event.getContent().transaction_id;
```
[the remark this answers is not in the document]
uhoreg: This function only gets called if event.getContent().transaction_id is truthy, so I think it should be OK.

uhoreg:
@dbkr it looks like some key backup changes got mixed in when I pulled in your branch.  Are they safe to merge in?  If so, go ahead and hit the merge button.

dbkr:
Oh I see, yes, it looks like this has picked up changes from experimental. I think the fix here is to merge experimental into this branch so github fixes its diff (did you merge in the branch which was at a later point than where you forked off? Probably could have avoided it in this case by cherry picking just `6a2da51`.

uhoreg:
merging the latest experimental in seems to have fixed it, so I'm going to merge this in now...

PR #820, opened 2019-01-09
Change initial key backup to background

jryans:
Alters the APIs used for initial key backup so that the actual upload happens in
the background after all session are marked for backup.

In addition, this also emits a counts of sessions remaining to backup to drive an upload progress UI.

Needed for https://github.com/vector-im/riot-web/issues/7863.
Used in https://github.com/matrix-org/matrix-react-sdk/pull/2416.

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -988,15 +988,17 @@ Crypto.prototype.importRoomKeys = function(keys) {
 /**
  * Schedules sending all keys waiting to be sent to the backup, if not already
  * scheduled. Retries if necessary.
+ *
+ * @param {number} maxDelay Maximum delay to wait in ms. 0 means no delay.
  */
-Crypto.prototype.scheduleKeyBackupSend = async function() {
+Crypto.prototype.scheduleKeyBackupSend = async function(maxDelay = 10000) {
     if (this._sendingBackups) return;
 
     try {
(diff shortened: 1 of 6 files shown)


src/crypto/index.js
```diff
@@ -1088,7 +1093,16 @@ Crypto.prototype._backupPendingKeys = async function(limit) {
...
         {rooms: data},
     );
-    await this._cryptoStore.unmarkSessionsNeedingBackup(sessions);
+
+    remaining = await this._cryptoStore.doTxn(
+        'readwrite',
+        IndexedDBCryptoStore.STORE_BACKUP,
+        async (txn) => {
+            await this._cryptoStore.unmarkSessionsNeedingBackup(sessions, txn);
+            return this._cryptoStore.countSessionsNeedingBackup(txn);
```
dbkr: I'm not sure this will work reliably since afaik indexeddb will close the txn as soon as the callback function is done finishes synchronously, so if the promises impl doesn't resume the async function straight away, the txn will have been closed by the time it gets to `countSessionsNeedingBackup`.

jryans: Thanks for pointing this out. I have reworked things to just use separate transactions, which to be honest is also easier to read (and just fine for this situation).

PR #821, opened 2019-01-09
Update ESLint

dbkr:
* Bump version
 * Add a couple of rules to match our existing precedent
 * Fix a few genuine lint errors
 * Ignore a guard-for-in (not sure why eslint doesn't like this?)
 * Update max warnings
 * Update ECMAScript version to 2017 which I think is closer to what we actually use
 * Stop using babel-eslint as we don't need it anymore (the built in parser now supports trailing commas)


dbkr:
Yep - sorry, was doing a react-sdk PR to fix them there but took longer than I expected! https://github.com/matrix-org/matrix-react-sdk/pull/2422

PR #824, opened 2019-01-17
Set key backup guard flag to avoid duplicate uploads

jryans:
Not sure how this was missed, but anyway, guard against duplicate uploads with
the existing flag.


jryans:
It seems the flag was removed by https://github.com/matrix-org/matrix-js-sdk/commit/66cdb62a3d4f97e17589d05e21e2fbcdd743e897, so probably it was just missed when reorganizing the backup functions.

PR #826, opened 2019-01-17
Include signature info for unknown devices

dbkr:
Add a 'deviceId' property and leave 'device' undefined for unknown
devices.

https://github.com/vector-im/riot-web/issues/8142

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -258,8 +258,9 @@ Crypto.prototype.checkKeyBackup = async function() {
  * @return {object} {
  *     usable: [bool], // is the backup trusted, true iff there is a sig that is valid & from a trusted device
  *     sigs: [
- *         valid: [bool],
- *         device: [DeviceInfo],
+ *         valid: [bool || null], // true: valid, false: invalid, null: cannot attempt validation
+ *         deviceId: [string],
+ *         device: [DeviceInfo || null],
  *     ]
  * }
  */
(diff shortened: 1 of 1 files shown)


src/crypto/index.js
```diff
@@ -287,26 +287,28 @@ Crypto.prototype.isKeyBackupTrusted = async function(backupInfo) {
     }
 
     for (const keyId of Object.keys(mySigs)) {
+        const sigInfo = { deviceId: keyId.split(':')[1] }; // XXX: is this how we're supposed to get the device ID?
         const device = this._deviceList.getStoredDevice(
-            this._userId, keyId.split(':')[1], // XXX: is this how we're supposed to get the device ID?
+            this._userId, sigInfo.deviceId,
```
jryans: This seems like it extends the returned `sigs` by adding a `deviceId` property for each. Should it be added to the function docs?

src/crypto/index.js
```diff
@@ -287,26 +287,28 @@ Crypto.prototype.isKeyBackupTrusted = async function(backupInfo) {
...
+                    device.deviceId,
+                    device.getFingerprint(),
+                );
+                sigInfo.valid = true;
+            } catch (e) {
+                console.log("Bad signature from device " + device.deviceId, e);
+                sigInfo.valid = false;
+            }
+        } else {
+            sigInfo.valid = null; // Can't determine validity because we don't have the signing device
```
jryans: It might be good to document this tri-state (null / false / true) behavior in the function docs.

dbkr: Oops, good shout :)

PR #834, opened 2019-02-05
Make linting rules more consistent

dbkr:
* Put back babel-eslint for class-properties
 * Allow arrow functions without params

This makes the style more consistent with react-sdk.

NB. The line lengths are still inconsistent but it's not clear which
way to go on that yet.


dbkr:
ah, yes, we should. I'll do that now.

PR #836, opened 2019-02-07
Trust on decrypt

dbkr:
Trust backups that we've restored by saving the matching pubkey
locally.

NB. Contains technically breaking API changes to the backup restore
(takes backupInfo rather than version).

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -1093,28 +1093,28 @@ MatrixClient.prototype.isValidRecoveryKey = function(recoveryKey) {
     }
 };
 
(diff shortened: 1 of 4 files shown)


src/client.js
```diff
@@ -1093,45 +1093,55 @@ MatrixClient.prototype.isValidRecoveryKey = function(recoveryKey) {
     }
 };
 
+MatrixClient.prototype.RESTORE_BACKUP_ERROR_BAD_KEY = 'RESTORE_BACKUP_ERROR_BAD_KEY';
```
turt2live: as per the react PR, I think this should be static rather than on the instance.

PR #837, opened 2019-02-08
change hex SAS verification to decimal and emoji

uhoreg:


uhoreg:
@dbkr your changes lgtm

PR #838, opened 2019-02-12
Fix: dont check for unverified devices in left members

bwindels:


src/models/room.js
```diff
@@ -599,9 +599,14 @@ Room.prototype.hasUnverifiedDevices = async function() {
     }
     const memberIds = Object.keys(this.currentState.members);
     for (const userId of memberIds) {
-        const devices = await this._client.getStoredDevicesForUser(userId);
-        if (devices.some((device) => device.isUnverified())) {
-            return true;
+        const membership = this.currentState.members[userId].membership;
+        // check for negative values rather than positive to be
+        // extra careful to not report false positives here
```
dbkr: This isn't specific to your change, but this isn't actually a question of being extra careful: this is (currently) a tristate where a device can be verified, unverified or blocked, so it depends what you want to do in the case where the room contains blocked devices. In this case the presence of blocked devices won't affect the return of `Room. hasUnverifiedDevices()` which I think is probably the correct behaviour?

bwindels: I meant being careful with checking the membership, but actually I should probably use `getEncryptionTargetMembers` instead.

bwindels:
@dbkr good for another look

PR #839, opened 2019-02-13
Release script: commit package-lock.json

dbkr:
Commit the package-lock.json when bumping the version, otherwise
the versions get out of sync, and this is going to matter more now
that jenkins runs `npm ci` which is fussy about these things.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -154,7 +154,15 @@ echo "npm version"
 # only turn off both of these behaviours, so we have to
 # manually commit the result.
 npm version --no-git-tag-version "$release"
-git commit package.json -m "$tag"
+
+# commit package-lock.json if it exists and is versioned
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -154,7 +154,15 @@ echo "npm version"
 # only turn off both of these behaviours, so we have to
 # manually commit the result.
 npm version --no-git-tag-version "$release"
-git commit package.json -m "$tag"
+
+# commit package-lock.json if it exists and is versioned
+if [[ -f package-lock.json && -z `git st --porcelain --ignored package-lock.json` ]];
```
jryans: Does `git st` work in a default install...? (I happen to already have `git st` as a local alias, but what happens by default...?) I think it's best to use the full `git status` in a script.

dbkr: good point, no it doesn't - this is the problem with setting up aliases like this on your local box :/

PR #840, opened 2019-02-13
Add method to force re-check of key backup

dbkr:
Also detect when the key backup version changes and do the right
thing

https://github.com/vector-im/riot-web/issues/8524

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -264,7 +264,7 @@ Crypto.prototype._checkAndStartKeyBackup = async function() {
     if (this._baseApis.isGuest()) {
         console.log("Skipping key backup check since user is guest");
         this._checkedForBackup = true;
-        return;
+        return null;
     }
     let backupInfo;
     try {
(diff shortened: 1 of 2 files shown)


src/crypto/index.js
```diff
@@ -302,10 +319,16 @@ Crypto.prototype.setTrustedBackupPubKey = async function(trustedPubKey) {
...
+ * @return {Object} Object with backup info (as returned by
+ *     getKeyBackupVersion) in backupInfo and
+ *     trust information (as returned by isKeyBackupTrusted)
+ *     in trustInfo.
  */
 Crypto.prototype.checkKeyBackup = async function() {
     this._checkedForBackup = false;
-    await this._checkAndStartKeyBackup();
+    const returnInfo = await this._checkAndStartKeyBackup();
+    return returnInfo;
```
jryans: Nit: Can also do `return await`, whichever you prefer

src/crypto/index.js
```diff
@@ -275,21 +275,38 @@ Crypto.prototype._checkAndStartKeyBackup = async function() {
...
         console.log("No usable key backup: not enabling key backup");
+    } else if (trustInfo.usable && this.backupInfo) {
+        // may not be the same version: if not, we should switch
+        if (backupInfo.version !== this.backupInfo.version) {
+            console.log(
+                "On backup version " + this.backupInfo.version + " but found " +
+                "version " + backupInfo.version + ": switching.",
+            );
+            this._baseApis.disableKeyBackup();
+            this._baseApis.enableKeyBackup(backupInfo);
```
jryans: Hmm, I thought there was a similar path that would switch like this as well, but can't seem to find it now...

dbkr: I don't remember writing it, although possibly we could get away with just the `enableKeyBackup`

PR #841, opened 2019-02-14
Try again to commit package-lock.json

jryans:
This appears to handle case of versioned and modified a bit better.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -155,8 +155,8 @@ echo "npm version"
 # manually commit the result.
 npm version --no-git-tag-version "$release"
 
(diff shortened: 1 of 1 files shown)


jryans:
(Push CI is failing because I first pushed to matrix-org repo and then deleted the branch, safe to ignore.)

jryans:
> expects an expression so this is evaluating the output

Right, I was counting on that, I guess: if there is a match, we'll have a non-zero string, else an empty string.

dbkr:
oh - I was somehow assuming that said `grep -q`. as you were then. :p

PR #847, opened 2019-02-26
Fix "e is undefined" masking the original error in MegolmDecryption

turt2live:
Someone with a bit of knowledge around the state machine for encryption probably wants to review this.

### DIFF
--- a/src/crypto/algorithms/megolm.js
+++ b/src/crypto/algorithms/megolm.js
@@ -758,7 +758,7 @@ MegolmDecryption.prototype.decryptEvent = async function(event) {
     } catch (e) {
         let errorCode = "OLM_DECRYPT_GROUP_MESSAGE_ERROR";
 
-        if (e && e.message && e.message === 'OLM.UNKNOWN_MESSAGE_INDEX') {
+        if (e && e.message === 'OLM.UNKNOWN_MESSAGE_INDEX') {
             this._requestKeysForEvent(event);
 
             errorCode = 'OLM_UNKNOWN_MESSAGE_INDEX';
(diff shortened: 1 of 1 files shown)


src/crypto/algorithms/megolm.js
```diff
@@ -758,15 +758,15 @@ MegolmDecryption.prototype.decryptEvent = async function(event) {
     } catch (e) {
         let errorCode = "OLM_DECRYPT_GROUP_MESSAGE_ERROR";
 
-        if (e.message === 'OLM.UNKNOWN_MESSAGE_INDEX') {
+        if (e && e.message && e.message === 'OLM.UNKNOWN_MESSAGE_INDEX') {
```
uhoreg: I think only `e && e.message === 'OLM.UNKNOWN_MESSAGE_INDEX'` is required, since if `e.message` is falsy, then it won't equal to `OLM.UNKNOWN_MESSAGE_INDEX`.

turt2live:
Good point - if we're getting undefined then something else is horribly wrong. At least this should (hopefully) make it a bit clearer though, even if it is just reducing the noise in the logs.

bwindels:
> Why is anything throwing `undefined`? open_mouth
> 
> Looks reasonable, though I don't know how much good it will do, given that if `e` is undefined, then there isn't really an "original error" that we can get any useful information from. On the other hand, it surely can't hurt, so +1 .

Could be https://github.com/matrix-org/matrix-js-sdk/blob/master/src/crypto/store/indexeddb-crypto-store-backend.js#L697 Any reason we don't take the same approach as [in the other idb store](https://github.com/matrix-org/matrix-js-sdk/blob/master/src/store/indexeddb-local-backend.js#L86)?

uhoreg:
@bwindels good catch.  I don't know why it different, but certainly the `reject()` is incorrect.

PR #848, opened 2019-02-28
handle partially-shared sessions better

uhoreg:
- don't cancel key requests if we can't decrypt everything in the session
- overwrite the session key if we get a better version

fixes https://github.com/vector-im/riot-web/issues/8793

### DIFF
--- a/spec/unit/crypto.spec.js
+++ b/spec/unit/crypto.spec.js
@@ -224,7 +224,9 @@ describe("Crypto", function() {
                     }
                 }));
 
-                const bobDecryptor = bobClient._crypto._getRoomDecryptor(roomId, olmlib.MEGOLM_ALGORITHM);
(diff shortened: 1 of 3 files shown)


spec/unit/crypto.spec.js
```diff
@@ -119,4 +123,154 @@ describe("Crypto", function() {
...
+
+                // keyshare the session key starting at the first message, so
+                // that it can now be decrypted
+                eventPromise = awaitEvent(events[0], "Event.decrypted");
+                ksEvent = await keyshareEventForEvent(events[0], 0);
+                await bobDecryptor.onRoomKeyEvent(ksEvent);
+                await eventPromise;
+                expect(events[0].getContent().msgtype).toNotBe("m.bad.encrypted");
+                // the room key request should still be there, since we haven't
+                // decrypted everything
```
dbkr: Comment needs fixing

PR #851, opened 2019-03-05
Calculate encrypted notification counts

turt2live:
Fixes https://github.com/vector-im/riot-web/issues/7489
~~**Required by https://github.com/matrix-org/matrix-react-sdk/pull/2744**~~ Not needed.

~~I'm not entirely convinced that the changes here are the right place to shove this code, but it does work well enough to at least open a PR.~~ I've since shuffled things around.

My GIF recorder is broken, so here's a deconstructed 2 frame GIF:
image
image

### DIFF
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -1425,6 +1425,39 @@ Room.prototype.getEventReadUpTo = function(userId, ignoreSynthesized) {
 };
 
 /**
+ * Determines if the given user has read a particular event ID with the known
+ * history of the room. This is not a definitive check as it relies only on
+ * what is available to the room at the time of execution.
+ * @param {String} userId The user ID to check the read state of.
+ * @param {String} eventId The event ID to check if the user read.
+ */
+Room.prototype.hasUserReadEvent = function(userId, eventId) {
+    const readUpToId = this.getEventReadUpTo(userId, false);
+    if (readUpToId === eventId) return true;
+
+    if (this.timeline.length
+        && this.timeline[this.timeline.length - 1].getSender()
+        && this.timeline[this.timeline.length - 1].getSender() === userId) {
+        // It doesn't matter where the event is in the timeline, the user has read
+        // it because they've sent the latest event.
+        return true;
+    }
+
+    for (let i = this.timeline.length - 1; i >= 0; --i) {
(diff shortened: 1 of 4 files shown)


src/models/room.js
```diff
@@ -1424,6 +1424,40 @@ Room.prototype.getEventReadUpTo = function(userId, ignoreSynthesized) {
...
+
+    if (this.timeline.length
+        && this.timeline[this.timeline.length - 1].getSender()
+        && this.timeline[this.timeline.length - 1].getSender() === userId) {
+        // It doesn't matter where the event is in the timeline, the user has read
+        // it because they've sent the latest event.
+        return true;
+    }
+
+    for (let i = this.timeline.length - 1; i >= 0; --i) {
```
ara4n: doing a linear search of the timeline feels quite heavy here, especially if we calc this for every new received event?

turt2live:
> we should be super clear that it ignores counts of events which arrived whilst the client was offline - ie the server can’t yet calculate correct counts for us.

Fair point, I didn't mention it in the PR but I did take some time to try and figure it out. The timeline stuff is impenetrable and non-trivial, and I'm not very inclined to keep bashing my head against this. A proper implementation of some MSC would almost certainly help with this.

src/models/room.js
turt2live: It's a reverse search, and very much ripped off from `Unread.doesRoomHaveUnreadMessages()` in the react-sdk. Worst case is the room is very popular and the loop falls behind by a couple iterations, but it should never get as bad as traversing the whole timeline.

turt2live:
Actually, wait a minute: this implementation does fix the counts. We already get a count for the number of unread messages (grey badge) and this specifically manipulates the highlight count using the client-side calculation of the push rules. If the client disappears for a while and comes back, this should be able to do the math correctly and come out with a number. An earlier implementation didn't do this, which is where my last comment is based on.

ara4n:
the problem is that if you’ve been binged in a room and your client wasn’t online at the time and hasn’t backfilled far enough to see it, the count will not include it. which is okay for now, but needs to be flagged in big warning letters and bugfiled

ara4n:
lgtm - thanks!

PR #852, opened 2019-03-05
Use 'ideal' rather than 'exact' for deviceid

dbkr:
We were using 'exact' which means we fail outright if the device
we wanted isn't available. This means if a user selects a specific
device then later unplugs it, we fail to open a capture device
the next time they make a call even if there's one available.
Using 'ideal' uses the chosen device in preference, but something
else if it isn't available.

Also log the name of the exception when we fail to open a capture
device to give us more of an idea of what's gone wrong.

Should help fix vector-im/riot-web#8993

### DIFF
--- a/src/matrix.js
+++ b/src/matrix.js
@@ -88,21 +88,21 @@ module.exports.createNewMatrixCall = require("./webrtc/call").createNewMatrixCal
 
 
 /**
- * Set an audio output device to use for MatrixCalls
+ * Set a preferred audio output device to use for MatrixCalls
  * @function
  * @param {string=} deviceId the identifier for the device
  * undefined treated as unset
  */
 module.exports.setMatrixCallAudioOutput = require('./webrtc/call').setAudioOutput;
 /**
- * Set an audio input device to use for MatrixCalls
+ * Set a preferred audio input device to use for MatrixCalls
  * @function
  * @param {string=} deviceId the identifier for the device
  * undefined treated as unset
  */
 module.exports.setMatrixCallAudioInput = require('./webrtc/call').setAudioInput;
 /**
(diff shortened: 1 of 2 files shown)


dbkr:
Yeah, the device selection is still a little bit odd. It gets the current preferred device ID from the js-sdk (if any) and sets that as the 'value' of the select, then populates the select with the devices it got from the browser. If the preferred device isn't there any more, you'll therefore have a select with a value that isn't one of the options in the select, in which case browsers will just show the first as selected (ie. the default device). So that means you open settings and it says you have the default device selected when in fact you do have a preferred device.

So another thing we could do would be to add the current preferred device to the select options, although we don't currently save the name (just the ID) so we'd have to display it as 'Unknown Device' or something.

jryans:
> It gets the current preferred device ID from the js-sdk (if any) and sets that as the 'value' of the select, then populates the select with the devices it got from the browser.

It seems like the sequence is:

1. `VoiceUserSettingsTab` calls `CallMediaHandler.getAudioOutput()` etc. on will mount
2. `CallMediaHandler.getAudioOutput` in React SDK reads `SettingsStore.getValueAt(SettingLevel.DEVICE, "webrtc_audiooutput")` (which would only be storing the preferred device after this change)

> So another thing we could do would be to add the current preferred device to the select options, although we don't currently save the name (just the ID) so we'd have to display it as 'Unknown Device' or something.

That sounds like a good improvement to file, if we don't want to do it right now. It sounds like that could good be added later if we want, so I'll mark this as approved and leave the next step up to you.

dbkr:
Filed as https://github.com/vector-im/riot-web/issues/9045

PR #855, opened 2019-03-06
More key request fixes

uhoreg:
fixed a typo, and make sure that we use a separate transaction ID

### DIFF
--- a/spec/unit/crypto.spec.js
+++ b/spec/unit/crypto.spec.js
@@ -12,6 +12,7 @@ import TestClient from '../TestClient';
 import {MatrixEvent} from '../../lib/models/event';
 import Room from '../../lib/models/room';
 import olmlib from '../../lib/crypto/olmlib';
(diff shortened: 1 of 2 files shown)


spec/unit/crypto.spec.js
```diff
@@ -137,9 +138,15 @@ describe("Crypto", function() {
             )).client;
             await aliceClient.initCrypto();
             await bobClient.initCrypto();
+            // clobber the setTimeout function to run 10x faster.
+            realSetTimeout = global.setTimeout;
+            global.setTimeout = function(f, n) {
+                return realSetTimeout(f, n/10);
+            };
```
dbkr: generally the way you do this is by installing a fake clock with lolex which just runs timeouts immediately (unless you actually need the delay?)

uhoreg:
Additional fix for vector-im/riot-web#6838

PR #856, opened 2019-03-08
Switch to `yarn` for dependency management

jryans:
Most `npm` operations are replaced with `yarn`, which generally has better
behavior. However, steps like publish that write to the NPM registry are left to
`npm`, which currently handles these tasks best.

Part of https://github.com/vector-im/riot-web/issues/7756


travis.sh
```diff
@@ -2,12 +2,15 @@
...
 
-npm run lint
+yarn lint
 
 # install Olm so that we can run the crypto tests.
-npm install https://matrix.org/packages/npm/olm/olm-3.1.0-pre1.tgz
+# This will add Olm as dependency, since it's currently unlisted.
+# (`yarn` does not have an install dependency without adding mode.)
+# TODO: Should Olm be a listed dev dependency instead, so that we can have it for testing
+# and don't need to run an extra step here?
```
t3chguy: this seems reasonable to me

jryans: Thanks, it has been done: https://github.com/matrix-org/matrix-js-sdk/pull/856/commits/cc86f427d226241c7b5a6bde0346524faefca3fd

jryans:
Assuming this conversion approach looks good for JS SDK, I'll open PRs for the other two repos as well. (There's no particular need to merge them all at the same time.)

PR #859, opened 2019-03-13
Use Buildkite for CI

turt2live:
TODO: Better pipelines (see https://github.com/vector-im/riot-web/issues/9019)

Part of fixing https://github.com/vector-im/riot-web/issues/9164

**Partnered with**:
* https://github.com/matrix-org/matrix-react-sdk/pull/2788
* https://github.com/vector-im/riot-web/pull/9165

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -1,0 +1,24 @@
+steps:
+  - label: ":eslint: Lint"
+    command:
+      - "yarn install"
+      - "yarn lint"
+    plugins:
+      - docker#v3.0.1:
+          image: "node:10"
+
+  - label: ":karma: Tests"
+    command:
+      - "yarn install"
+      - "yarn test"
+    plugins:
+      - docker#v3.0.1:
+          image: "node:10"
+
+  - label: "📃 Docs"
+    command:
+      - "yarn install"
+      - "yarn gendoc"
+    plugins:
+      - docker#v3.0.1:
+          image: "node:10"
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.yaml
```diff
@@ -0,0 +1,24 @@
+steps:
+  - label: ":eslint: Lint"
+    command:
+      - "yarn install"
```
jryans: Looks like we get a modern Yarn version, so that's good. Is that controlled by the `node:10` Docker image?

.buildkite/pipeline.yaml
```diff
@@ -0,0 +1,24 @@
...
+      - "yarn lint"
+    plugins:
+      - docker#v3.0.1:
+          image: "node:10"
+
+  - label: ":karma: Tests"
+    command:
+      - "yarn install"
```
jryans: Is it really worth it to have separate steps like this? (I am assuming steps run in parallel, but please correct me if that's wrong...) I would assume `yarn install` is the longest part of the process, so having to repeat it 3 times feels a bit wasteful...

.buildkite/pipeline.yaml
turt2live: yea, iirc as of node 6 or 8 it's automatically included in the Docker images.

.buildkite/pipeline.yaml
turt2live: It sure isn't. That's what "better pipelines" is though: https://github.com/vector-im/riot-web/issues/9019

(rationale: fixing it now is not fun because we'll want to think carefully about what we actually want. The purpose of this exercise it to ditch Travis CI and migrate over to something useful, bringing whatever we have along with it).

Things are parallel.

PR #860, opened 2019-03-15
Add a concept of default push rules, using it for tombstone notifications

turt2live:
Part of vector-im/riot-web#8447
See also matrix-org/matrix-doc#1930

**Intended for https://github.com/matrix-org/matrix-react-sdk/pull/2798**

### DIFF
--- a/src/pushprocessor.js
+++ b/src/pushprocessor.js
@@ -342,15 +342,22 @@ function PushProcessor(client) {
         // Deep clone the object before we mutate it
         const ruleset = JSON.parse(JSON.stringify(clientRuleset));
 
-        if (!clientRuleset['global']) clientRuleset['global'] = {};
-        if (!clientRuleset['global']['override']) clientRuleset['global']['override'] = [];
+        if (!clientRuleset['global']) {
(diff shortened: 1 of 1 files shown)


turt2live:
Synapse implements it for new logins/users, but existing users who haven't re-logged will not have the push rule in their sync stream. 

This is more to solve the problem of synapse not telling us about the new push rule and the smaller case of people using an older homeserver which does not have the push rule defined. This way, the user still receives notifications for upgrades but doesn't need to update their homeserver or re-login.

dbkr:
right, yep - I remember this from last time we added a default rule. Back then I think we fetched the rules separately. If we can't trust what synapse is sending us down the sync stream, there's an argument for fetching them separately rather than trying to second guess what we think they probably ought to look like. That's probably more complex and almost as bad though, so maybe just put a big XXX comment here to say it's doing this because of a bug in synapse?

turt2live:
That comment is the first line of  the diff - it's less of a synapse thing and more of a "we want red notifications for tombstones, regardless of server support"

PR #861, opened 2019-03-19
Rename `MatrixInMemoryStore` to `MemoryStore`

jryans:
None of the other store classes use the `Matrix` prefix, and I find the mismatch
confusing (it leads me to think it might have a different purpose than the
others).

This change removes the prefix from the store for consistency. The old name is
left as an export for existing SDK consumers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ client.on("Room.timeline", function(event, room, toStartOfTimeline) {
 });
 ```
 
-By default, the `matrix-js-sdk` client uses the `MatrixInMemoryStore` to store events as they are received. For example to iterate through the currently stored timeline for a room:
+By default, the `matrix-js-sdk` client uses the `MemoryStore` to store events as they are received. For example to iterate through the currently stored timeline for a room:
 
 ```javascript
 Object.keys(client.store.rooms).forEach((roomId) => {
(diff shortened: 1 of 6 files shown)


t3chguy:
This is a breaking change as some things depend on this store. For example: https://github.com/matrix-org/matrix-search/blob/ef289c6e8b3293089a5a422c5664da208f864a52/js_fetcher/index.ts#L167

turt2live:
Not to mention all the bots and bridges which use it. A stub class might work, but comments which clear up the confusion might be better.

jryans:
The PR currently [preserves exporting](https://github.com/matrix-org/matrix-js-sdk/pull/861/files#diff-44f2bbaefc4af762ff42533cf1f158d4R31) the old name for compatibility, as I was expecting issues just like what you've mentioned otherwise.

Does that address these issues?

jryans:
> Can it be made more obvious somehow?

I could add "COMPATIBILITY:" or some other loud string to the comment, maybe?

turt2live:
It sure reminds me that the project has never been updated for new build practices, but that's a different issue.

I think it should at least get a deprecation notice so we can remove it eventually.

jryans:
Okay, I have added the deprecation notice as well. (Sorry for tripping everyone's breaking change alarm... 😰)

jryans:
Yes, I've tested that both the old and new exported names are working as expected.

PR #866, opened 2019-03-20
Add debugging for spurious room version warnings

turt2live:
See https://github.com/vector-im/riot-web/issues/9225


ara4n:
oh, it doesn’t belong to a room; nm

PR #871, opened 2019-03-25
Add IndexedDB existence checks

jryans:
This will be useful for future storage diagnostics as part of vector-im/riot-web#9271.

Used by https://github.com/matrix-org/matrix-react-sdk/pull/2826

### DIFF
--- a/src/indexeddb-helpers.js
+++ b/src/indexeddb-helpers.js
@@ -35,6 +35,8 @@ export function exists(indexedDB, dbName) {
         };
         req.onblocked = () => reject();
         req.onsuccess = () => {
+            const db = req.result;
+            db.close();
             if (!exists) {
                 // The DB did not exist before, but has been created as part of this
                 // existence check. Delete it now to restore previous state. Delete can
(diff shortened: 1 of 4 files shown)


src/indexeddb-helpers.js
```diff
@@ -0,0 +1,50 @@
...
+            // should only fire if the DB did not exist before at any version.
+            exists = false;
+        };
+        req.onblocked = () => reject();
+        req.onsuccess = function() {
+            if (!exists) {
+                // The DB did not exist before, but has been created as part of this
+                // existence check. Delete it now to restore previous state. Delete can
+                // actually take a while to complete in some browsers, so don't wait for
+                // it. This won't block future open calls the might issue next to properly
```
turt2live: I think I know what you're trying to say, but I don't really know enough to write up a suggestion. The confusing bit is:
> future open calls the might issue next to properly

jryans: Thanks for catching! I guess I was trying to say "the app"... Anyway, I went for "a store" in the end.

PR #873, opened 2019-03-29
Refuse splicing the live timeline into a broken position

turt2live:
Credit to Matthew for basically solving this.

Theoretically fixes spontaneous timeline corruption: https://github.com/vector-im/riot-web/issues/8593

When the live timeline ends up in a position where it can no longer be live (such as becoming the second timeline in the set, rather than the first) we end up getting neighbouring timeline errors. By refusing to splice the live timeline into such a position, we hopefully keep the live timeline in a position of still being live for when it is next used.

The running theory that leads to this fix is multiple limited syncs coming in, causing holes in the timeline. When trying to patch up the holes, the timeline set would end up splicing all over the place, leading to potentially splicing the live timeline into a broken position.

Disclaimer: I've done light testing on Riot to make sure it doesn't fail in miserable ways, but have not reproduced the case which causes the issue in the first place. The issue is a bit complex to test, hence the word "theory" being littered above.

### DIFF
--- a/src/models/event-timeline-set.js
+++ b/src/models/event-timeline-set.js
@@ -415,14 +415,20 @@ EventTimelineSet.prototype.addEventsToTimeline = function(events, toStartOfTimel
 
         if (direction === EventTimeline.BACKWARDS && existingIsLive) {
             // The live timeline should never be spliced into a non-live position.
-            console.warn("Refusing to splice live timeline as a backwards timeline");
+            console.warn(
+                "Refusing to set a preceding existingTimeLine on our " +
+                "timeline as the existingTimeLine is live",
+            );
         } else {
             timeline.setNeighbouringTimeline(existingTimeline, direction);
         }
 
         if (inverseDirection === EventTimeline.BACKWARDS && timelineIsLive) {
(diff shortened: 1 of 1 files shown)


src/models/event-timeline-set.js
```diff
@@ -408,8 +408,21 @@ EventTimelineSet.prototype.addEventsToTimeline = function(events, toStartOfTimel
         console.info("Already have timeline for " + eventId +
                      " - joining timeline " + timeline + " to " +
                      existingTimeline);
-        timeline.setNeighbouringTimeline(existingTimeline, direction);
-        existingTimeline.setNeighbouringTimeline(timeline, inverseDirection);
+
```
ara4n: the warnings are a bit cryptic here, to the point of being very confusing for the 2nd one. how about:

"Refusing to set a preceding existingTimeLine on our timeline as the existingTimeLine is live"
and
"Refusing to set our preceding timeline on a existingTimeLine as our timeline is live"
perhaps?

ara4n:
lgtm. only other change would be to log the timeline ids so we can do forensics later

PR #877, opened 2019-04-02
Refuse to link live timelines into the forwards/backwards position when either is invalid

turt2live:
See https://github.com/vector-im/riot-web/issues/8593#issuecomment-478681816

Previously (https://github.com/matrix-org/matrix-js-sdk/pull/873) we allowed half-linking timelines to each other if they satisfy the conditions, however this appears to not be helping. Instead, it seems like the timelines are getting stuck in a position where one direction is spliced but the other is broken. To avoid this case, we'll just avoid splicing in both directions when one of the directions is invalid.

### DIFF
--- a/src/models/event-timeline-set.js
+++ b/src/models/event-timeline-set.js
@@ -413,25 +413,30 @@ EventTimelineSet.prototype.addEventsToTimeline = function(events, toStartOfTimel
         const existingIsLive = existingTimeline === this._liveTimeline;
         const timelineIsLive = timeline === this._liveTimeline;
 
-        if (direction === EventTimeline.BACKWARDS && existingIsLive) {
+        const backwardsIsLive = direction === EventTimeline.BACKWARDS && existingIsLive;
+        const forwardsIsLive = direction === EventTimeline.FORWARDS && timelineIsLive;
+
+        if (backwardsIsLive || forwardsIsLive) {
             // The live timeline should never be spliced into a non-live position.
-            console.warn(
-                "Refusing to set a preceding existingTimeLine on our " +
-                "timeline as the existingTimeLine is live (" + existingTimeline + ")",
-            );
-        } else {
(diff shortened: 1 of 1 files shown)


turt2live:
It would be true only in the cases where the live timeline is being put in a position where it cannot be live (ie: A -> B -> C, where B would be the live timeline). The variables are already a bit confusing here, but `forwardIsLive` uses the inverse of `backwardsIsLive`, just like it was before. 

The problem this aims to solve is certainly vague, and it might very well indeed not fix anything. The concern is that by allowing 50% of a link to go through, we're actually making the problem worse. The theory is that by preventing the link altogether that we'll end up with a stable timeline which can be nicely spliced later on if needed.

bwindels:
Ok, so the only actual change is the continue statement so setNeighbour ... doesn't get called for either timelines. :+1:

src/models/event-timeline-set.js
```diff
@@ -413,26 +413,31 @@ EventTimelineSet.prototype.addEventsToTimeline = function(events, toStartOfTimel
...
+        if (backwardsIsLive || forwardsIsLive) {
             // The live timeline should never be spliced into a non-live position.
-            console.warn(
-                "Refusing to set our preceding timeline on a existingTimeLine " +
-                "as our timeline is live (" + timeline + ")",
-            );
-        } else {
-            existingTimeline.setNeighbouringTimeline(timeline, inverseDirection);
+            // We use independent logging to better discover the problem at a glance.
+            console.warn({backwardsIsLive, forwardsIsLive}); // debugging
```
bwindels: is this intended to stay in? Should be clear from seeing either of the below messages already?

turt2live: It's intended to stay in to make it slightly easier to troubleshoot from the logs without having to go hunting for messages in the code.

PR #879, opened 2019-04-02
Remove console.log lines from sendEvent

Half-Shot:
This has been plaguing the `matrix-appservice-bridge` project for years, as it means any events sent will go to stdout whereas most bridges run their own logging system. I'd gently push for Riot to log this inside it's own code.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -1670,8 +1670,6 @@ MatrixClient.prototype.sendEvent = function(roomId, eventType, content, txnId,
         txnId = this.makeTxnId();
     }
 
-    console.log(`sendEvent of type ${eventType} in ${roomId} with txnId ${txnId}`);
-
     // we always construct a MatrixEvent when sending because the store and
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -1846,9 +1844,6 @@ function _sendEventHttpRequest(client, event) {
     return client._http.authedRequest(
         undefined, "PUT", path, undefined, event.getWireContent(),
     ).then((res) => {
```
t3chguy: whats the point of this passthrough then handler now?

Half-Shot: Good point, dropped.

Half-Shot:
Presumably [this](https://github.com/matrix-org/matrix-js-sdk/pull/763). It's a shame the author didn't fix all of the console references.

turt2live:
Riot doesn't make use of the logging library, and by default it dumps to the console anyways. It also looks like this got addressed in the time it took to review, sorry :(

PR #883, opened 2019-04-03
Key backup logging improvements

dbkr:
To try & diagnose https://github.com/vector-im/riot-web/issues/9357

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -357,7 +357,12 @@ Crypto.prototype.isKeyBackupTrusted = async function(backupInfo) {
     const mySigs = backupInfo.auth_data.signatures[this._userId] || [];
 
     for (const keyId of Object.keys(mySigs)) {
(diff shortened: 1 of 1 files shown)


src/crypto/index.js
```diff
@@ -373,7 +378,7 @@ Crypto.prototype.isKeyBackupTrusted = async function(backupInfo) {
                 );
                 sigInfo.valid = true;
             } catch (e) {
-                logger.info("Bad signature from device " + device.deviceId, e);
+                logger.info("Bad signature from key ID " + keyId, e);
```
jryans: Would it be useful to keep logging the `device.deviceId` and then also add the key ID?

dbkr: The device ID we just get by `keyId.split(':')[1]` so probably not really

jryans: Okay, makes sense then!

PR #885, opened 2019-04-05
Refuse to set forwards pagination token on live timeline

turt2live:
Should fix the error seen in https://github.com/matrix-org/riot-web-rageshakes/issues/1389 (https://github.com/vector-im/riot-web/issues/8593)


turt2live:
will give it a try and see how the weekend goes.

Issue #881, opened 2019-04-02
Calls to /redact are POSTed when they should be PUT according to the spec


turt2live:
Synapse supports both, but we can't rely on both being present. We should be using PUT.

jkasun:
This seems like a good first issue, may i fix this?

turt2live:
Absolutely! I see you've already visited #riot-dev, which is a great first step.

PR #887, opened 2019-04-06
change event redact,  POST request to PUT request

jkasun:
This fixes the issue Calls to /redact are POSTed when they should be PUT according to the spec #881.  Now redact event is a PUT request instead of a POST request. A transaction id is also added.

Signed-off-by: Janith Kasun janithisipathana@gmail.com

----

Fixes https://github.com/matrix-org/matrix-js-sdk/issues/881 (for magic github stuff) ~ TravisR

### DIFF
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -1,3 +1,23 @@
+/**
+ * @param {string} roomId
+ * @param {string} eventId
+ * @param {string=} txnId     transaction id. One will be made up if not
+ *    supplied.
+ * @param {module:client.callback} callback Optional.
+ * @return {module:client.Promise} Resolves: TODO
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixBaseApis.prototype.redactEvent = function(
+    roomId, eventId, txnId, callback,
+) {
+    const path = utils.encodeUri("/rooms/$roomId/redact/$eventId/$tnxId", {
+        $roomId: roomId,
+        $eventId: eventId,
+        $txnId: txnId ? txnId : this.makeTxnId(),
+    });
+
+    return this._http.authedRequest(callback, "PUT", path, undefined, {});
+};
 /*
 Copyright 2015, 2016 OpenMarket Ltd
 Copyright 2017 Vector Creations Ltd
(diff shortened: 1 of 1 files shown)


src/base-apis.js
```diff
@@ -1,3 +1,23 @@
+/**
```
bwindels: Paste accident? The method is duplicated above the file header.

src/base-apis.js
```diff
@@ -891,18 +911,25 @@ MatrixBaseApis.prototype.sendStateEvent = function(roomId, eventType, content, s
...
+ * @param {string=} txnId     transaction id. One will be made up if not
+ *    supplied.
  * @param {module:client.callback} callback Optional.
  * @return {module:client.Promise} Resolves: TODO
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
-MatrixBaseApis.prototype.redactEvent = function(roomId, eventId, callback) {
-    const path = utils.encodeUri("/rooms/$roomId/redact/$eventId", {
+MatrixBaseApis.prototype.redactEvent = function(
+    roomId, eventId, txnId, callback,
```
bwindels: I suppose as it stands this would be a breaking change, as a parameter is added to a public method. Either we'll need to bump the major version and add a warning to the CHANGELOG, or you shift the callback argument to the left if the txnId is undefined, so it behaves the same as before.

I think I would prefer the latter.

src/base-apis.js
jkasun: Oh that's a mistake.

src/base-apis.js
jkasun: Yeah that seems fair to use callback if tnxId is undefined. Do I have to close this pull request and send a new one.

jkasun: Also checking if that's a instance of Funcation may help

bwindels: Just pushing more commits here is fine. I suppose the most correct way would be to check arguments.length, and if it's 3, do `callback = txnId`.

We generally don't check the type of arguments in the SDK fwiw.

jkasun:
Seems like a build error, what is happening

src/base-apis.js
```diff
@@ -891,18 +891,29 @@ MatrixBaseApis.prototype.sendStateEvent = function(roomId, eventType, content, s
 /**
  * @param {string} roomId
  * @param {string} eventId
+ * @param {string=} txnId     transaction id. One will be made up if not
```
bwindels: I'd mark this `@param {string} [txnId]` now

jkasun: oh is it since that is optional

jkasun: fixed :)

PR #890, opened 2019-04-08
Cache failed capabilities lookups for shorter amounts of time

turt2live:
This should fix https://github.com/vector-im/riot-web/issues/9225 for showing up too often/too long.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -445,9 +445,10 @@ MatrixClient.prototype.setNotifTimelineSet = function(notifTimelineSet) {
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
 MatrixClient.prototype.getCapabilities = function() {
+    const now = new Date().getTime();
+
     if (this._cachedCapabilities) {
-        const now = new Date().getTime();
-        if (now - this._cachedCapabilities.lastUpdated <= CAPABILITIES_CACHE_MS) {
+        if (now < this._cachedCapabilities.expiration) {
             console.log("Returning cached capabilities");
             return Promise.resolve(this._cachedCapabilities.capabilities);
         }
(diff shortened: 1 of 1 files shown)


src/models/room.js
```diff
@@ -38,7 +38,7 @@ import ReEmitter from '../ReEmitter';
 // to upgrade (ie: "stable"). Eventually, we should remove these when all homeservers
 // return an m.room_versions capability.
 const KNOWN_SAFE_ROOM_VERSION = '1';
-const SAFE_ROOM_VERSIONS = ['1', '2'];
+const SAFE_ROOM_VERSIONS = ['1', '2', '3'];
```
jryans: Shouldn't this change be made in a separate PR? Normally I wouldn't mind personally, but this seems like the kind of change that should be done by itself.

src/client.js
```diff
@@ -459,9 +460,16 @@ MatrixClient.prototype.getCapabilities = function() {
     ).catch(() => null).then((r) => {
```
jryans: Would it help diagnose room version issues to at least log errors here?

src/client.js
```diff
@@ -445,9 +445,10 @@ MatrixClient.prototype.setNotifTimelineSet = function(notifTimelineSet) {
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
 MatrixClient.prototype.getCapabilities = function() {
```
jryans: The current code seems to hit the server twice in quick succession during startup, perhaps because two code paths call this before we're able to receive and cache the first attempt.

I wonder if that's at all related to the issue here, perhaps if the server were rate limiting or something and sent an error response the second time...?

Anyway, it's likely something worth improving at some stage, though doesn't have to be this PR. (Just mentioning here in case it helps.)

src/models/room.js
turt2live: backed out to https://github.com/matrix-org/matrix-js-sdk/pull/893

PR #891, opened 2019-04-08
Fix highlight notifications for unencrypted rooms

turt2live:
A logic error introduced by https://github.com/matrix-org/matrix-js-sdk/pull/886 meant that all unencrypted rooms were not getting highlight notifications.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -1106,7 +1106,8 @@ SyncApi.prototype._processSyncResponse = async function(
             // server's for this case, and therefore will assume that our non-zero
             // count is accurate.
             const encrypted = client.isRoomEncrypted(room.roomId);
-            if (!encrypted || (encrypted && room.getUnreadNotificationCount('highlight') <= 0)) {
+            if (!encrypted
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
     rules: {
         // rules we've always adhered to or now do
         "max-len": ["error", {
-            code: 90,
+            code: 120,
```
turt2live: I'm tired of constantly breaking lines in weird ways in the js-sdk just to appease the length limit, and we use 120 for the react-sdk.

dbkr: I think I suggested this before, but people seemed fond of the 90 character limit.

jryans: Personally, I prefer 90 as a max since it's easier to read two side by side files that way, but that's just my opinion.

Maybe we should discuss it as a team first, and then make a separate PR for it if we want to change?

turt2live: I'll back out the change and save my protest for another day

PR #892, opened 2019-04-09
Explicitly guard store usage during sync startup

jryans:
This adds explicit `try` blocks in the spots where we interact with the store
during sync startup. This shouldn't be necessary as the store should already be
catching this and degrading as of
https://github.com/matrix-org/matrix-js-sdk/pull/884, but that doesn't seem to
have been enough for the affected user in
https://github.com/vector-im/riot-web/issues/7769, as they are seeing sync just
stop when storing without any further detail.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -499,9 +499,16 @@ SyncApi.prototype.sync = function() {
             const supported = await client.doesServerSupportLazyLoading();
             if (supported) {
                 debuglog("Creating and storing lazy load sync filter...");
-                this.opts.filter = await client.createFilter(
-                    Filter.LAZY_LOADING_SYNC_FILTER,
-                );
+                try {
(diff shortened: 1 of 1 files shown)


jryans:
> Should we be rethrowing the exception here so it bails as it would have before? Or Is it safe to catch the exception and power through? (If so we should probably at least move the `debuglog("Stored client options");` into the try block so it doesn't print that after claiming it's just failed.

Yeah, I wasn't sure which way to go with this... If these fail to store but then somehow sync data in a later step _does_ store to IDB, there would be a mismatch between the options and data, which would be bad for the next session.

Since I am still not really sure what's happening here, let's be safe and re-throw. I've changed to that, but also moved the logs anyway since that seems like the right thing for readability.

PR #894, opened 2019-04-11
Expose better autodiscovery error messages

turt2live:
Fixes https://github.com/vector-im/riot-web/issues/7925

**Please review with https://github.com/matrix-org/matrix-react-sdk/pull/2916**


turt2live:
I'll stick error codes on the beautification stage of the implementation

PR #895, opened 2019-04-15
Stop syncing when the token is invalid

turt2live:
Fixes https://github.com/vector-im/riot-web/issues/9451

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -445,6 +445,16 @@ SyncApi.prototype._wasLazyLoadingToggled = async function(lazyLoadMembers) {
     return false;
 };
 
+SyncApi.prototype._shouldAbortSync = function(error) {
+    if (err.errcode === "M_UNKNOWN_TOKEN") {
+        // The logout already happened, we just need to stop.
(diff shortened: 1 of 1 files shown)


src/sync.js
```diff
@@ -462,6 +462,16 @@ SyncApi.prototype.sync = function() {
     let savedSyncPromise = Promise.resolve();
     let savedSyncToken = null;
 
+    async function shouldAbortSync(err) {
```
bwindels: can remove `async` keyword?

src/sync.js
```diff
@@ -877,6 +889,12 @@ SyncApi.prototype._onSyncError = function(err, syncOptions) {
     console.error("/sync error %s", err);
     console.error(err);
 
+    if (err.errcode === "M_UNKNOWN_TOKEN") {
```
bwindels: avoid duplication with `if (shouldAbortSync(err)) { return; }` here?

turt2live: This is a different function, but there's nothing stopping the function from being defined in a higher scope I guess.

PR #897, opened 2019-04-15
Support being fed a .well-known config object for validation

turt2live:
Used by Riot to consume the user's provided config. This also includes a change to carry over custom keys on m.homeserver and m.identity_server which aren't intentionally controlled.

Required for https://github.com/vector-im/riot-web/issues/9290

### DIFF
--- a/src/autodiscovery.js
+++ b/src/autodiscovery.js
@@ -226,7 +226,7 @@ export class AutoDiscovery {
             logger.error("No m.homeserver key/base_url in config");
 
             clientConfig["m.homeserver"].state = AutoDiscovery.FAIL_PROMPT;
-            clientConfig["m.homeserver"].error = AutoDiscovery.ERROR_GENERIC_FAILURE;
+            clientConfig["m.homeserver"].error = AutoDiscovery.ERROR_INVALID_HS_BASE_URL;
 
             return Promise.resolve(clientConfig);
         }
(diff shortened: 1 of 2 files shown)


src/autodiscovery.js
```diff
@@ -346,14 +327,98 @@ export class AutoDiscovery {
...
         // Step 7: Copy any other keys directly into the clientConfig. This is for
         // things like custom configuration of services.
-        Object.keys(wellknown.raw)
-            .filter((k) => k !== "m.homeserver" && k !== "m.identity_server")
-            .map((k) => clientConfig[k] = wellknown.raw[k]);
+        Object.keys(wellknown)
+            .map((k) => {
+                if (k === "m.homeserver" || k === "m.identity_server") {
+                    // Only copy selected parts of the config to avoid overwriting
+                    // important information.
```
jryans: "Important information" is a bit vague here... I think you mean you don't want to overwrite the properties this logic has already calculated, right?

turt2live: Yup, was hoping it was clear in the context though. Wording suggestions?

jryans: Hmm, well, if your goal is copy the entire object minus the properties you have calculated, could you just mutate `wellknown` directly and return it, rather having this extra copying step? Is the concern that someone might place the keys `error` or `state` in the .well-known, and you want to ignore those?

Anyway, assuming the you want to keep the copying step, perhaps:

```suggestion
                    // properties computed by the validation logic above.
```

turt2live: I'd rather not mutate the object because the returned object is meant to be a "discovery config" which is backed by a discovery mechanism (in our case, .well-known). Theory is that in future we can support different mechanisms and everything just works, and that our output format happens to be very similar to .well-known. The other reason would be that the calling code may want to preserve their object for future calls, and so we shouldn't mutate it.

PR #899, opened 2019-04-16
Fix tests on develop

turt2live:
See commits for fixes

### DIFF
--- a/spec/unit/autodiscovery.spec.js
+++ b/spec/unit/autodiscovery.spec.js
@@ -94,7 +94,7 @@ describe("AutoDiscovery", function() {
                 const expected = {
                     "m.homeserver": {
                         state: "FAIL_PROMPT",
-                        error: "Invalid homeserver discovery response",
+                        error: "Failed to get autodiscovery configuration from server",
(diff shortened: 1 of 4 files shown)


spec/unit/autodiscovery.spec.js
```diff
@@ -94,7 +94,7 @@ describe("AutoDiscovery", function() {
                 const expected = {
                     "m.homeserver": {
                         state: "FAIL_PROMPT",
-                        error: "Invalid homeserver discovery response",
+                        error: "Failed to get autodiscovery configuration from server",
```
uhoreg: It seems like maybe these strings should be defined as constants in the `AutoDiscovery` class, so that the unit tests can use those rather than trying to keep all the strings in sync.  Or else the tests shouldn't care about the human-readable strings.  Maybe not worth blocking this PR, though.

uhoreg: The PR lgtm otherwise

jryans:
Merging as @turt2live recommended so that tests will pass.

PR #900, opened 2019-04-17
emit self-membership event at end of handling sync update

bwindels:
otherwise the room state isn't updated yet (and we can't for
example distinguish a leave from a kick)

this is only used for updating the UI,
seems safe to emit this event at a later point


bwindels:
need to figure out why those tests are failing

PR #907, opened 2019-05-03
yarn upgrade

dbkr:


dbkr:
Yep, yarn pulled in the broken version of base-x. Looks like depending on a specific version in package.json doesn't work with yarn because it just installs two versions to satisfy the requirements, so I don't know of a way of pinning a child dependency short of manually editing yarn.lock...

jryans:
Hmm, so is this a trap waiting for the next `yarn upgrade` which will push `base-x` to 3.0.5+ again?

Something doesn't seem right here...

dbkr:
Yep, exactly. Not sure how to fix this short of not using the module.

jryans:
It looks like we can use Yarn's [selective dependency resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/) feature for this.

1. Remove `base-x` from deps (which you already did)
2. Add the following block to `package.json`:
```
  "resolutions": {
    "bs58/base-x": "3.0.4"
  },
```
3. Run `yarn install`

dbkr:
Oh, that seems to work! Good find!

PR #910, opened 2019-05-08
Add basic read path for relations

jryans:
This adds a read path for relations (gated behind an unstable option). A few
basic client-side grouping and sorting operations are supported. Consumers are
expected to ask the `EventTimelineSet` for a relation container when desired.

See [MSC 1849](https://github.com/matrix-org/matrix-doc/blob/matthew/msc1849/proposals/1849-aggregations.md) to learn more about these events.

Part of https://github.com/vector-im/riot-web/issues/9485

### DIFF
--- a/src/models/relations.js
+++ b/src/models/relations.js
@@ -1,0 +1,149 @@
+/*
+Copyright 2019 New Vector Ltd
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * A container for relation events that supports easy access to common ways of
+ * aggregating such events. Each instance holds events that of a single relation
+ * type and event type. All of the events also relate to the same original event.
+ *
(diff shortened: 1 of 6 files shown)


src/models/relations.js
```diff
@@ -0,0 +1,149 @@
...
+    /**
+     * Get all events in this collection grouped by sender.
+     *
+     * This is currently only supported for the annotation relation type.
+     *
+     * @return {Object}
+     * An object with each relation sender as a key and the matching list of
+     * events for that sender as a value.
+     */
+    getAnnotationsBySender() {
```
bwindels: OOI, where will we need this?

src/models/room.js
```diff
@@ -92,9 +92,12 @@ function synthesizeReceipt(userId, event, receiptType) {
  * "<b>detached</b>", pending messages will appear in a separate list,
  * accessbile via {@link module:models/room#getPendingEvents}. Default:
  * "chronological".
- *
  * @param {boolean} [opts.timelineSupport = false] Set to true to enable improved
  * timeline support.
+ * @param {boolean} [opts.unstableClientRelationAggregation = false]
```
bwindels: passing this down to the `EventTimelineSet` constructor will happen in a subsequent PR?

src/models/relations.js
jryans: I am using this with reactions to keep track of the current user's own reactions, so that we can highlight what you chose, etc.

src/models/room.js
jryans: `room.js` actually passes the whole options object through to `EventTimelineSet`, which is a bit surprising, but anyway it means it already happens:

https://github.com/matrix-org/matrix-js-sdk/blob/6078bbbe2475a1b95d211b64caa588b375c371d9/src/models/room.js#L174

PR #911, opened 2019-05-10
Redaction and change events for relations

jryans:
**Reviewer:** Each commit in this PR is self-contained and has some additional context in the commit message, so you may want to review commit by commit.

This adds reactions and change event support needed for relations events in the JS SDK, which will be used for reactions and message editing features.

Fixes https://github.com/vector-im/riot-web/issues/9485
Part of https://github.com/vector-im/riot-web/issues/9572
Part of https://github.com/vector-im/riot-web/issues/9574
Used in https://github.com/matrix-org/matrix-react-sdk/pull/2954


.babelrc
```diff
@@ -1,11 +1,11 @@
 {
     "presets": ["es2015"],
     "plugins": [
+        "transform-class-properties",
         // this transforms async functions into generator functions, which
         // are then made to use the regenerator module by babel's
         // transform-regnerator plugin (which is enabled by es2015).
         "transform-async-to-bluebird",
-
```
turt2live: extreme nit: this blank line helps with legibility. Would be nice to keep it, and to insert one after line 4 for the comment block.

src/models/relations.js
```diff
@@ -22,21 +24,29 @@ limitations under the License.
...
  */
-export default class Relations {
+export default class Relations extends EventEmitter {
     /**
      * @param {String} relationType
      * The type of relation involved, such as "m.annotation", "m.reference",
      * "m.replace", etc.
      * @param {String} eventType
      * The relation event's type, such as "m.reaction", etc.
+     * @param {?Room} room
```
turt2live: isn't it usually `{Room?}` ?

src/models/relations.js
```diff
@@ -111,6 +121,49 @@ export default class Relations {
...
     }
 
+    /**
+     * For relations that are about to be redacted, remove them from aggregation
+     * data sets and emit an update event.
+     *
+     * @param {MatrixEvent} redactedEvent
+     * The original relation event that is about to be redacted.
+     */
+    _onBeforeRedaction = (redactedEvent) => {
```
turt2live: This doesn't seem to handle the case of a redaction failing to apply? (would expect the count to go down optimistically then back up due to failure to redact)

.babelrc
jryans: Okay, will update!

src/models/relations.js
jryans: I don't use JSDoc type syntax much, but that seems to the nullable type convention there: https://jsdoc.app/tags-type.html

src/models/relations.js
jryans: This is runs after the server has approved the redaction, but before we have marked the original event redacted, so we can trust it has happened here. https://github.com/vector-im/riot-web/issues/9675 will cover local echo for redactions, but the current code doesn't need to worry about that right now.

turt2live: a comment to help future explorers might be worthwhile

jryans:
> For future PRs, please separate things like [00e4236](https://github.com/matrix-org/matrix-js-sdk/commit/00e4236cd62bc414845aa9cf96f9cecd68a16a85) due to their size. It's fine for this PR because it's already done though.

Right, sorry... I try to separate when it makes sense to me. In this case, I was worried it would add lag to the review process for the feature.

> As an aside: how does this behave if the relation comes in before the event? What if the event is cut off by `limited: true` on a sync but the relation is not? (not questions to fix for this PR)

Possibly badly, will need to test. 😅 We _should_ be storing any relations as they come in, but haven't tried it yet. In the long run, we'll switch to server-side aggregation instead, which should handle those cases better.

src/models/relations.js
jryans: Comment [added](https://github.com/matrix-org/matrix-js-sdk/pull/911/commits/f411d50253ff344080641c2de0a750f389e22190).

PR #913, opened 2019-05-14
Support for replacing message through m.replace relationship.

bwindels:
Initial support for m.replace relationships. Doesn't do local echo yet, and only works if the original event is received before the replacements (e.g. like through sync). Follow-up PRs to follow.

https://github.com/vector-im/riot-web/issues/9671

Accompanying react-sdk PR: https://github.com/matrix-org/matrix-react-sdk/pull/2952

### DIFF
--- a/src/models/event-timeline-set.js
+++ b/src/models/event-timeline-set.js
@@ -244,6 +244,35 @@ EventTimelineSet.prototype.findEventById = function(eventId) {
 };
 
 /**
+ * Find an event by id which is stored in our timelines and replace it with a new event
+ *
+ * @param {string} eventId  event ID to look for
+ * @param {string} newEvent the event to replace with
+ * @return {?module:models/event~MatrixEvent} the replaced event, or undefined if unknown
+ */
+EventTimelineSet.prototype.tryReplaceEvent = function(eventId, newEvent) {
+    // we can't use this._eventIdToTimeline here to find the matching timeline
+    // because we need to match the original event id,
+    // which might have been replaced by a previous m.replace already.
+    // for now, loop through all timelines. Maybe a separate index would be justifyable.
+    let oldEvent;
+    for(const tl of this._timelines) {
+        oldEvent = tl.tryReplaceEvent(eventId, newEvent);
+        if (oldEvent) {
+            break;
+        }
+    }
+    if (oldEvent) {
+        let originalEvent;
+        if (oldEvent.isReplacement()) {
+            originalEvent = oldEvent.getReplacedEvent();
+        }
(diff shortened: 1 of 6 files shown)


bwindels:
Build seems to be failing because olm can't be fetched. Someone is looking at this ftr.

src/models/event-timeline-set.js
```diff
@@ -243,6 +243,31 @@ EventTimelineSet.prototype.findEventById = function(eventId) {
     });
 };
 
+/**
+ * Find an event by id which is stored in our timelines and replace it with a new event
```
turt2live: nit: capitalize ID

src/models/event-timeline.js
```diff
@@ -339,6 +339,27 @@ EventTimeline.prototype.addEvent = function(event, atStart) {
     }
 };
 
+
+/**
+ * Replaces an event in the timeline, if it exists.
+ *
+ * @param {string} eventId  the id of the existing event
```
turt2live: nit: capitalize ID

src/models/event-timeline.js
```diff
@@ -339,6 +339,27 @@ EventTimeline.prototype.addEvent = function(event, atStart) {
...
+EventTimeline.prototype.tryReplaceEvent = function(eventId, newEvent) {
+    const eventIndex = this._events.findIndex(event => event.getOriginalId() === eventId);
+    if (eventIndex === -1) {
+        return;
+    }
+    const oldEvent = this._events[eventIndex];
+    this._events[eventIndex] = newEvent;
+    // copy over metadata, as we don't hold state for random locations in the timeline, only for the end or start
+    // and assuming for now that replacing an event doesn't change the sender that should be shown.
+    newEvent.sender = oldEvent.sender;
```
jryans: Will this work move into an `Event.replace(...)` method, since I believe you're planning to change to be more like redaction?

bwindels: Yes, I already have that almost working, apart from react not updating the tile for some reason. I'd be tempted to still merge this as it is, as it works and I'd prefer as much testing time on this as we can get.

jryans: Right, seems okay to merge as is. I just wanted to make sure I understand the plan. 😄

src/models/event-timeline-set.js
bwindels: This method will be removed again in the next PR, so should be ok.

src/models/event-timeline.js
bwindels: This method will be removed again in the next PR, so should be ok.

PR #914, opened 2019-05-15
Message editing: mark original event as replaced instead of replacing the event object

bwindels:
this is more in line with what happens on the server-side,
and also doesn't break existing reply relations,
and is a simpler approach as well.

### DIFF
--- a/src/models/event.js
+++ b/src/models/event.js
@@ -122,7 +122,7 @@ module.exports.MatrixEvent = function MatrixEvent(
     this.error = null;
     this.forwardLooking = true;
     this._pushActions = null;
-    this._replacedEvent = null;
+    this._replacingEvent = null;
 
     this._clearEvent = {};
 
@@ -209,9 +209,6 @@ utils.extend(module.exports.MatrixEvent.prototype, {
      * @return {Number} The event timestamp, e.g. <code>1433502692297</code>
      */
     getTs: function() {
-        if (this._replacedEvent) {
-            return this._replacedEvent.getTs();
-        }
         return this.event.origin_server_ts;
     },
 
@@ -220,9 +217,6 @@ utils.extend(module.exports.MatrixEvent.prototype, {
      * @return {Date} The event date, e.g. <code>new Date(1433502692297)</code>
      */
     getDate: function() {
-        if (this._replacedEvent) {
-            return this._replacedEvent.getDate();
-        }
         return this.event.origin_server_ts ? new Date(this.event.origin_server_ts) : null;
     },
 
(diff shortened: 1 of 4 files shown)


src/models/event.js
```diff
@@ -741,37 +741,34 @@ utils.extend(module.exports.MatrixEvent.prototype, {
...
 
-    getReplacedEvent() {
-        return this._replacedEvent;
+    /**
+     * Set an event that replaces the content of this event, through an m.replace relation.
+     *
+     * @param {MatrixEvent} newEvent the event with the replacing content.
+     */
+    makeReplaced(newEvent) {
+        this._replacingEvent = newEvent;
```
jryans: Should this directly set `this.event.content`, rather than checking the replacing event during `getContent`? It depends whether you want to keep access to the unreplaced content...

src/models/room.js
```diff
@@ -1032,22 +1032,15 @@ Room.prototype._addLiveEvent = function(event, duplicateStrategy) {
...
-            if (originalEvent.isReplacement()) {
-                originalEvent = originalEvent.getReplacedEvent();
-            }
-            event.setReplacedEvent(originalEvent);
-            // report replacedEvent and not originalEvent because replaceEvent was in the timeline so far
-            this.emit("Room.replaceEvent", replacedEvent, event, this);
+            replacedEvent.makeReplaced(event);
+            this.emit("Room.replaceEvent", replacedEvent, this);
         }
-        // we don't add the event because the event type would get rendered
```
jryans: Ah interesting, so now it's okay to go in the timeline?

jryans: Ah, I guess you're hiding in React SDK now.

src/models/event.js
bwindels: felt nicer to not modify the actual event ... I don't have any pref.

bwindels: Maybe it's nicer to not be doing Object.assign in getContent that way though, will have a look.

jryans: Well, let's go with this for now, can adjust later.

bwindels: Did so in the end :+1:

src/models/event.js
```diff
@@ -747,7 +741,9 @@ utils.extend(module.exports.MatrixEvent.prototype, {
      * @return {boolean}
      */
     isRelation(relType = undefined) {
-        const content = this.getContent();
+        // must use event.content as m.relates_to is not encrypted
+        // and _clearEvent doesn't have it.
+        const content = this.event.content;
```
jryans: Locally I have added a `getRelation` helper and replaced all accesses of `m.relates_to` to use that, so I'll apply the same cleanup to these bits once they land.

src/models/event.js
```diff
@@ -759,6 +755,24 @@ utils.extend(module.exports.MatrixEvent.prototype, {
...
      */
     makeReplaced(newEvent) {
+        if (this.isRedacted()) {
+            return;
+        }
+        // ignore previous replacements
+        if (this._replacingEvent && this._replacingEvent.getTs() > newEvent.getTs()) {
+            return;
+        }
+        if (newEvent.isBeingDecrypted()) {
```
jryans: What about if the current event is still being decrypted (like when reapplying replacement from a cached sync)?

bwindels: yep, have that on my list, but as it will be similar to handling back pagination, I'm putting it off till then.

PR #916, opened 2019-05-15
Add stringify helper to summarise events when debugging

jryans:
This has been useful while working on encrypted reactions.

### DIFF
--- a/src/models/event.js
+++ b/src/models/event.js
@@ -784,6 +784,35 @@ utils.extend(module.exports.MatrixEvent.prototype, {
     replacingEventId() {
         return this._replacingEvent && this._replacingEvent.getId();
     },
+
+    /**
+     * Summarise the event as JSON for debugging. If encrypted, include both the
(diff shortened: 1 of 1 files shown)


src/models/event.js
```diff
@@ -784,6 +784,33 @@ utils.extend(module.exports.MatrixEvent.prototype, {
...
         return this._replacingEvent && this._replacingEvent.getId();
     },
+
+    /**
+     * Summarise the event as JSON for debugging. If encrypted, include both the
+     * decrypted and encrypted view of the event.
+     *
+     * @return {Object}
+     */
+    toJSON() {
```
turt2live: teeeeeeeeecccchnnically this is `toObject` because JSON is a string

jryans: Ah, I should have explained further... It has this name for use with `JSON.stringify`, which checks objects for functions named `toJSON` and will call them to customise the output if they are defined. I'll add this to the method comment.

PR #917, opened 2019-05-15
Wait until decrypt before aggregating

jryans:
This is part of getting encrypted reactions in https://github.com/vector-im/riot-web/issues/9672.

### DIFF
--- a/src/models/event.js
+++ b/src/models/event.js
@@ -735,18 +735,32 @@ utils.extend(module.exports.MatrixEvent.prototype, {
     },
 
     /**
-     * Get whether the event is a relation event, and of a given type if `relType` is passed in.
+     * Get whether the event is a relation event, and of a given type if
+     * `relType` is passed in.
      *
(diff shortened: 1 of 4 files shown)


src/models/event.js
```diff
@@ -735,20 +735,34 @@ utils.extend(module.exports.MatrixEvent.prototype, {
...
         return relation && relation.rel_type && relation.event_id &&
             ((relType && relation.rel_type === relType) || !relType);
     },
 
+    /**
+     * Get relation info for the event, if any.
+     *
+     * @return {Object}
+     */
+    getRelation() {
```
bwindels: should we use this in `ReplyThread.js` as well?

jryans: Unfortunately, we can't use it there at the moment because `isRelation` requires a `rel_type`, and current replies don't have a `rel_type`. I'll likely add a note about this in React SDK.

PR #920, opened 2019-05-16
Local echo for m.replace relations

bwindels:
Part of: https://github.com/vector-im/riot-web/issues/9671
React-SDK PR: https://github.com/matrix-org/matrix-react-sdk/pull/2982

### DIFF
--- a/src/models/event-timeline-set.js
+++ b/src/models/event-timeline-set.js
@@ -759,7 +759,7 @@ EventTimelineSet.prototype.aggregateRelations = function(event) {
         }
     }
 
-    relationsWithEventType.addEvent(event);
+    relationsWithEventType.addEvent(event, relatesToEvent);
 };
 
 /**
--- a/src/models/event-timeline-set.js
+++ b/src/models/event-timeline-set.js
@@ -747,19 +747,24 @@ EventTimelineSet.prototype.aggregateRelations = function(event) {
         relationsWithRelType = relationsForEvent[relationType] = {};
     }
     let relationsWithEventType = relationsWithRelType[eventType];
+    const relatesToEvent = this.findEventById(relatesToEventId);
     if (!relationsWithEventType) {
         relationsWithEventType = relationsWithRelType[eventType] = new Relations(
             relationType,
             eventType,
             this.room,
         );
-        const relatesToEvent = this.findEventById(relatesToEventId);
         if (relatesToEvent) {
             relatesToEvent.emit("Event.relationsCreated", relationType, eventType);
         }
     }
 
     relationsWithEventType.addEvent(event, relatesToEvent);
+
+    if (relationType === "m.replace" && relatesToEvent) {
+        const replacement = relationsWithEventType.getLastReplacement();
+        relatesToEvent.makeReplaced(replacement);
+    }
 };
 
 /**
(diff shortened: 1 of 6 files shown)


src/models/event.js
```diff
@@ -226,7 +226,12 @@ utils.extend(module.exports.MatrixEvent.prototype, {
      * @return {Object} The event content JSON, or an empty object.
      */
     getContent: function() {
-        return this._clearEvent.content || this.event.content || {};
+        if (this._replacingEvent && !this.isRedacted()) {
+            return this._replacingEvent.getContent()["m.new_content"] || {};
+            // content = Object.assign({}, content, newContent);
```
jryans: I'm assuming this gets removed later...

src/models/event.js
```diff
@@ -226,7 +226,12 @@ utils.extend(module.exports.MatrixEvent.prototype, {
      * @return {Object} The event content JSON, or an empty object.
      */
     getContent: function() {
-        return this._clearEvent.content || this.event.content || {};
+        if (this._replacingEvent && !this.isRedacted()) {
```
jryans: Why the redacted check here...?

src/models/event-timeline-set.js
```diff
@@ -759,7 +759,7 @@ EventTimelineSet.prototype.aggregateRelations = function(event) {
         }
     }
 
-    relationsWithEventType.addEvent(event);
+    relationsWithEventType.addEvent(event, relatesToEvent);
```
jryans: We should add a separate `setTargetEvent` or similar to do this separately, and also ensure we only do it one time per relations collection (once the target event is found), as `findEventById` is not very efficient.

src/models/event.js
```diff
@@ -801,6 +801,18 @@ utils.extend(module.exports.MatrixEvent.prototype, {
             this._replacingEvent = newEvent;
             this.emit("Event.replaced", this);
         }
+    },
+
+    /**
+     * Returns the  status of the event, or the replacing event in case `makeReplace` has been called.
```
jryans: Two spaces before "status", should be one

src/models/event.js
bwindels: Because we don't want to return any replaced content for a redacted event. I guess a better way to do this would be to set `this._replacingEvent` to `null` in `makeRedacted`.

src/models/event-timeline-set.js
```diff
@@ -708,6 +709,35 @@ EventTimelineSet.prototype.getRelationsForEvent = function(
     return relationsWithRelType[eventType];
 };
 
+
```
jryans: Nit: Looks like two blank lines, maybe remove one?

src/models/relations.js
```diff
@@ -316,13 +316,19 @@ export default class Relations extends EventEmitter {
...
+
     /*
-     * @param {MatrixEvent} targetEvent the event related to.
+     * @param {MatrixEvent} targetEvent the event the relations are related to.
      */
     setTargetEvent(event) {
-        this._targetEvent = event;
-        if (this.relationType === "m.replace") {
-            this._targetEvent.makeReplaced(this.getLastReplacement());
+        if (!this._targetEvent) {
```
jryans: Maybe do:

```
if (this._targetEvent) {
    return;
}
```

for less indenting? (I am a bit "early return obsessive", feel free to ignore...)

PR #924, opened 2019-05-19
Extend loglevel logging for the whole project

### LINKED ISSUE #772: Extend logging to whole project
After #763 was merged, let's extend it to the rest of the code base, so we can get rid of `console.log` statements.

jkasun:
Now logging using the loglevel is extend to the whole project instead of `console.log`
Fixes #772

Signed-off-by: Janith Kasun janithisipathana@gmail.com

### DIFF
--- a/spec/olm-loader.js
+++ b/spec/olm-loader.js
@@ -14,9 +14,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// try to load the olm library.
 import logger from '../src/logger';
 
+// try to load the olm library.
 try {
     global.Olm = require('olm');
     logger.log('loaded libolm');
--- a/spec/olm-loader.js
+++ b/spec/olm-loader.js
@@ -15,10 +15,11 @@ limitations under the License.
 */
 
 // try to load the olm library.
+import logger from '../src/logger';
 
 try {
     global.Olm = require('olm');
-    console.log('loaded libolm');
+    logger.log('loaded libolm');
 } catch (e) {
-    console.warn("unable to run crypto tests: libolm not available");
+    logger.warn("unable to run crypto tests: libolm not available");
 }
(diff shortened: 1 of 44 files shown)


spec/olm-loader.js
```diff
@@ -15,10 +15,11 @@ limitations under the License.
 */
 
 // try to load the olm library.
+import logger from '../src/logger';
```
turt2live: should be above the comment

src/models/room.js
```diff
@@ -555,9 +556,9 @@ Room.prototype.clearLoadedMembersIfNeeded = async function() {
  */
 Room.prototype._cleanupAfterLeaving = function() {
     this.clearLoadedMembersIfNeeded().catch((err) => {
-        console.error(`error after clearing loaded members from ` +
+        logger.error(`error after clearing loaded members from ` +
             `room ${this.roomId} after leaving`);
-        console.dir(err);
+        logger.dir(err);
```
turt2live: does this actually work?

src/realtime-callbacks.js
```diff
@@ -39,9 +39,10 @@ let _realCallbackKey;
 // each is an object with keys [runAt, func, params, key].
 const _callbackList = [];
 
+import logger from '../src/logger';
+
 // var debuglog = console.log.bind(console);
```
turt2live: should fix the debuglog too, and please add line 44's blank line back

jkasun:
thanks for the review @turt2live. I'll fix these issues and push another commit.

src/models/room.js
jkasun: @turt2live nice catch. Yes this is not working. What can we do about that? Maybe just a log

spec/olm-loader.js
jkasun: This is fixed

src/realtime-callbacks.js
jkasun: This is fixed

src/models/room.js
turt2live: just a log is probably fine, yea

jkasun: Done

PR #925, opened 2019-05-21
Propose Revision To Readme/Contributors

JQuilty:
Hi,

I was reviewing the contributor guidelines and readme, and noticed it used some confusing language for newcomers. Particularly the "Here be dragons". In this phrasing, it can be unclear what dragons are, especially to non-English speakers or those unfamiliar with the phrase. I would propose re-phrasing it to something like "known irregularities", and within the doc something like "irregularities that will not be fixed", "irregularities we'd like to fix", etc.


turt2live:
@JQuilty it looks like your changes weren't included in the PR - I'm closing this because the branches don't line up.

It sounds like you're looking for the "new issue" button (found here: https://github.com/matrix-org/matrix-js-sdk/issues/new ) or the edit button on the relevant documents.

PR #926, opened 2019-05-22
interactive-auth now handles requesting email tokens

dbkr:
interactive-auth now has a callback to request the email token which
it will call at the appropriate time (now at the start of the
auth process if the chosen flow contain an email auth stage).

This does make this a breaking change, although not sure this is
used outside of Riot. We could make it backwards compatible by
having an option for the new behaviour. It may not be worthwhile
though.

https://github.com/vector-im/riot-web/issues/9586

### DIFF
--- a/src/interactive-auth.js
+++ b/src/interactive-auth.js
@@ -1,6 +1,7 @@
 /*
 Copyright 2016 OpenMarket Ltd
 Copyright 2017 Vector Creations Ltd
+Copyright 2019 The Matrix.org Foundation C.I.C.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -88,6 +89,12 @@ const MSISDN_STAGE_TYPE = "m.login.msisdn";
  * @param {string?} opts.emailSid If returning from having completed m.login.email.identity
  *     auth, the sid for the email verification session.
  *
+ * @param {function?} opts.requestEmailToken A function that takes the email address (string),
+ *     clientSecret (string), attempt number (int) and sessionId (string) and calls the
+ *     relevant requestToken function and returns the promise returned by that function.
+ *     If the resulting promise rejects, the rejection will propagate through to the
+ *     attemptAuth promise.
+ *
  */
 function InteractiveAuth(opts) {
     this._matrixClient = opts.matrixClient;
@@ -95,14 +102,17 @@ function InteractiveAuth(opts) {
     this._requestCallback = opts.doRequest;
     // startAuthStage included for backwards compat
     this._stateUpdatedCallback = opts.stateUpdated || opts.startAuthStage;
-    this._completionDeferred = null;
+    this._resolveFunc = null;
+    this._rejectFunc = null;
     this._inputs = opts.inputs || {};
+    this._requestEmailTokenCallback = opts.requestEmailToken;
 
     if (opts.sessionId) this._data.session = opts.sessionId;
     this._clientSecret = opts.clientSecret || this._matrixClient.generateClientSecret();
     this._emailSid = opts.emailSid;
     if (this._emailSid === undefined) this._emailSid = null;
 
+    this._chosenFlow = null;
     this._currentStage = null;
 }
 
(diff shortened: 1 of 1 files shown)


src/interactive-auth.js
```diff
@@ -115,19 +125,17 @@ InteractiveAuth.prototype = {
      *     no suitable authentication flow can be found
      */
     attemptAuth: function() {
-        this._completionDeferred = Promise.defer();
+        return new Promise(async (resolve, reject) => {
```
bwindels: no need for async

src/interactive-auth.js
```diff
@@ -253,58 +265,66 @@ InteractiveAuth.prototype = {
...
-                self._completionDeferred.resolve(result);
-            }, function(error) {
-                // sometimes UI auth errors don't come with flows
-                const errorFlows = error.data ? error.data.flows : null;
-                const haveFlows = Boolean(self._data.flows) || Boolean(errorFlows);
-                if (error.httpStatus !== 401 || !error.data || !haveFlows) {
-                    // doesn't look like an interactive-auth failure. fail the whole lot.
-                    throw error;
+            const result = await this._requestCallback(auth, background);
+            console.log("result from request: ", result);
```
bwindels: left-over logging?

src/interactive-auth.js
```diff
@@ -253,58 +265,66 @@ InteractiveAuth.prototype = {
...
+                // if the email address is not valid (ie. already taken or not
+                // registered, depending on what the operation is).
+                try {
+                    const requestTokenResult = await this._requestEmailTokenCallback(
+                        this._inputs.emailAddress,
+                        this._clientSecret,
+                        1, // TODO: Multiple send attempts?
+                        this._data.session,
+                    );
+                    this._emailSid = requestTokenResult.sid;
```
bwindels: As it stands, do we need to resolve here? I'm unsure what will cause `attemptAuth` to resolve in this case.

src/interactive-auth.js
```diff
@@ -313,14 +333,14 @@ InteractiveAuth.prototype = {
      * @private
      * @throws {NoAuthFlowFoundError} If no suitable authentication flow can be found
      */
-    _startNextAuthStage: function() {
+    _startNextAuthStage: async function() {
```
bwindels: no need for async

src/interactive-auth.js
```diff
@@ -253,58 +265,66 @@ InteractiveAuth.prototype = {
...
+            const haveFlows = Boolean(this._data.flows) || Boolean(errorFlows);
+            if (error.httpStatus !== 401 || !error.data || !haveFlows) {
+                // doesn't look like an interactive-auth failure.
+                if (!background) {
+                    this._rejectFunc(error);
+                } else {
+                    // We ignore all failures here (even non-UI auth related ones)
+                    // since we don't want to suddenly fail if the internet connection
+                    // had a blip whilst we were polling
+                    console.log("Ignoring error from UI auth: " + error);
```
bwindels: If meaning to keep this in, maybe make the message a bit less generic/more detailed?

src/interactive-auth.js
```diff
@@ -88,21 +89,30 @@ const MSISDN_STAGE_TYPE = "m.login.msisdn";
...
     this._matrixClient = opts.matrixClient;
     this._data = opts.authData || {};
     this._requestCallback = opts.doRequest;
     // startAuthStage included for backwards compat
     this._stateUpdatedCallback = opts.stateUpdated || opts.startAuthStage;
-    this._completionDeferred = null;
+    this._resolveFunc = null;
+    this._rejectFunc = null;
     this._inputs = opts.inputs || {};
+    this._requestEmailTokenCallback = opts.requestEmailToken;
```
bwindels: The callback provided by the react-sdk PR calls straight back into the JS-SDK. Curious why you opted for not doing that straight in here?

dbkr: It's because there is a requestToken function for each endpoint which all have slightly different semantics (ie. the one for registration errors if the email address is already registered, the one for password reset errors if the email is *not* registered) and interactive-auth doesn't know what endpoint you're calling, so it doesn't know what requestToken method to use.

src/interactive-auth.js
dbkr: hopefully clarified with comments

PR #927, opened 2019-05-23
Add a bunch of debugging to .well-known IS validation

turt2live:


src/autodiscovery.js
```diff
@@ -296,6 +298,7 @@ export class AutoDiscovery {
                 wellknown["m.identity_server"]["base_url"],
             );
             if (!isUrl) {
+                logger.log("[Debug] First error: isUrl is falsey");
```
uhoreg: This seems unnecessary given that there's a `logger.error`statement right below?

src/autodiscovery.js
```diff
@@ -308,11 +311,20 @@ export class AutoDiscovery {
                 `${isUrl}/_matrix/identity/api/v1`,
             );
             if (!isResponse || !isResponse.raw || isResponse.action !== "SUCCESS") {
+                logger.log("[Debug] Second error: response from IS failed");
```
uhoreg: Also possible unnecessary.

src/autodiscovery.js
turt2live: in theory, yes. There's about 2-3 places where we say `Invalid base_url for m.identity_server` on the error channel though :(

src/autodiscovery.js
turt2live: this one is, yes. I plan on just reverting this whole commit once I don't need it anymore, and am feeling a bit lazy to fix this

PR #931, opened 2019-05-30
Fix recovery key encoding for base-x 3.0.5

jryans:
**Reviewer**: This is heading to release-v2.0.0 branch for immediate release.

This fixes our recovery key encoding to work with base-x 3.0.5, which requires the
encoding input to be in a `Buffer`. (base-x is a dependency of bs58.)

Unfortunately, base-x hasn't marked this as breaking change through a major
version increment, so semantic versioning can't help us here to ensure we get
particular version.

I have verified that the change to `Buffer` works with both old and new base-x,
so we should be okay no matter what version is actually selected.

For extra fun, base-x 3.0.5 also uses newer JS features, which `uglify-js`
doesn't support. I have migrated to `terser` instead, which is what we're using
for Riot these days (via Webpack).

Fixes https://github.com/vector-im/riot-web/issues/9901

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "start:watch": "babel -s -w --skip-initial-build -d lib src",
     "start:init": "babel -s -d lib src",
     "clean": "rimraf lib dist",
-    "build": "babel -s -d lib src && rimraf dist && mkdir dist && browserify -d browser-index.js | exorcist dist/browser-matrix.js.map > dist/browser-matrix.js && uglifyjs -c -m -o dist/browser-matrix.min.js --source-map dist/browser-matrix.min.js.map --in-source-map dist/browser-matrix.js.map dist/browser-matrix.js",
+    "build": "babel -s -d lib src && rimraf dist && mkdir dist && browserify -d browser-index.js | exorcist dist/browser-matrix.js.map > dist/browser-matrix.js && terser -c -m -o dist/browser-matrix.min.js --source-map \"content='dist/browser-matrix.js.map'\" dist/browser-matrix.js",
     "dist": "yarn build",
     "watch": "watchify -d browser-index.js -o 'exorcist dist/browser-matrix.js.map > dist/browser-matrix.js' -v",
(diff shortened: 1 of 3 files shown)


turt2live:
iirc it was previously pinned to .3 or .4 - did the package itself get re-released or something?

Also at some point we should just upgrade. This isn't the first time base-x has bit us, and seemingly won't be the last.

jryans:
Yes, we were pinning it to 3.0.4 to avoid doing the (small) upgrade work in this PR. I believe what happened though is that the pinning method we used ([Yarn's selective resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/)) don't flow through to packages that depend on the JS SDK, so Riot and React SDK would have had to also pin (as well as anyone else who wants to use this code in the JS SDK). I believe for local development with symlinks, the JS SDK's pinned version was used properly, but when we made a release at the Riot level, the pin was not applied, resulting in a difference between development and production.

For these fun reasons, we should probably avoid this style of pinning. This PR does exactly that: removes the pin and makes other changes so that we can work with the latest base-x version.

jryans:
As @turt2live mentioned, this PR at the moment does start exporting a small amount of ES6 (via base-x), so we'd need to decide whether that's okay or not. (See https://github.com/matrix-org/matrix-js-sdk/issues/803 for the last discussion about it.)

jryans:
I discussed the issues of exporting ES 6 with @ara4n IRL, and he has approved the idea, along with bumping the major version to indicate the breaking change (so to 2.0.0).

jryans:
To clarify, 2.0.0 would contain only a tiny drip of ES 6 via `base-x`. Actually exporting ES 6 for the SDK code is left as a future exercise.

jryans:
(I will change the branch name after merging this PR, as I'd rather not lose the comments by reopening...)

jryans:
I have confirmed that:

* This PR works in a browser
* A recovery key encoded with this PR works correctly for recovery in a Riot without this PR

I believe this means we're ready to merge.

PR #934, opened 2019-06-03
Refresh safe room versions when the server looks more modern than us

turt2live:
Fixes https://github.com/vector-im/riot-web/issues/9845

### DIFF
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -263,6 +263,33 @@ Room.prototype.getRecommendedVersion = async function() {
         }
     }
 
+    let result = this._checkVersionAgainstCapability(versionCap);
+    if (result.urgent && result.needsUpgrade) {
+        // Something doesn't feel right: we shouldn't need to update
+        // because the version we're on should be in the protocol's
+        // namespace. This usually means that the server was updated
+        // before the client was, making us think the newest possible
+        // room version is not stable. As a solution, we'll refresh
+        // the capability we're using to determine this.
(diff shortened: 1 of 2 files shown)


src/models/room.js
```diff
@@ -263,6 +263,33 @@ Room.prototype.getRecommendedVersion = async function() {
...
+        // before the client was, making us think the newest possible
+        // room version is not stable. As a solution, we'll refresh
+        // the capability we're using to determine this.
+        logger.warn(
+            "Refreshing room version capability because the server looks " +
+            "to be supporting a newer room version we don't know about.",
+        );
+
+        const caps = await this._client.getCapabilities(true);
+        if (!caps) {
```
bwindels: getCapabilities seems to always return an object, so this branch would never run? Should it be `if (caps["m.room_versions"]) ...` rather?

src/models/room.js
```diff
@@ -291,12 +318,12 @@ Room.prototype.getRecommendedVersion = async function() {
         } else {
             logger.warn(`Non-urgent upgrade required on ${this.roomId}`);
         }
-        return Promise.resolve(result);
+        return result;
```
bwindels: There is a Promise.resolve remaining on line 304

src/models/room.js
turt2live: Does it actually? That's embarassing. Should be checking the room version capability, yes.

PR #937, opened 2019-06-03
Local echo for redactions

bwindels:
https://github.com/vector-im/riot-web/issues/9675

### DIFF
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -896,7 +896,7 @@ MatrixBaseApis.prototype.sendStateEvent = function(roomId, eventType, content, s
  * @return {module:client.Promise} Resolves: TODO
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
-MatrixBaseApis.prototype.redactEvent = function(
+MatrixBaseApis.prototype.sendRedaction = function(
     roomId, eventId, txnId, callback,
 ) {
     if (arguments.length === 3) {
--- a/src/models/event.js
+++ b/src/models/event.js
@@ -124,6 +124,7 @@ module.exports.MatrixEvent = function MatrixEvent(
     this.forwardLooking = true;
     this._pushActions = null;
     this._replacingEvent = null;
+    this._locallyRedacted = false;
 
     this._clearEvent = {};
 
@@ -228,6 +229,9 @@ utils.extend(module.exports.MatrixEvent.prototype, {
      * @return {Object} The event content JSON, or an empty object.
      */
     getOriginalContent: function() {
+        if (this._locallyRedacted) {
+            return {};
+        }
         return this._clearEvent.content || this.event.content || {};
     },
 
@@ -239,7 +243,9 @@ utils.extend(module.exports.MatrixEvent.prototype, {
      * @return {Object} The event content JSON, or an empty object.
      */
     getContent: function() {
-        if (this._replacingEvent) {
+        if (this._locallyRedacted) {
+            return {};
+        } else if (this._replacingEvent) {
             return this._replacingEvent.getContent()["m.new_content"] || {};
         } else {
             return this.getOriginalContent();
@@ -666,6 +672,19 @@ utils.extend(module.exports.MatrixEvent.prototype, {
         return this.event.unsigned || {};
     },
 
+    unmarkLocallyRedacted: function() {
+        const value = this._locallyRedacted;
+        this._locallyRedacted = false;
+        return value;
+    },
+
+    markLocallyRedacted: function(redactionEvent) {
+        if (this._locallyRedacted) {
+            return;
+        }
+        this.emit("Event.beforeRedaction", this, redactionEvent);
+        this._locallyRedacted = true;
+    },
     /**
      * Update the content of an event in the same way it would be by the server
      * if it were redacted before it was sent to us
@@ -678,6 +697,8 @@ utils.extend(module.exports.MatrixEvent.prototype, {
         if (!redaction_event.event) {
             throw new Error("invalid redaction_event in makeRedacted");
         }
+
+        this._locallyRedacted = false;
 
         this.emit("Event.beforeRedaction", this, redaction_event);
 
@@ -721,7 +742,7 @@ utils.extend(module.exports.MatrixEvent.prototype, {
      * @return {boolean} True if this event has been redacted
      */
     isRedacted: function() {
-        return Boolean(this.getUnsigned().redacted_because);
+        return this._locallyRedacted || Boolean(this.getUnsigned().redacted_because);
     },
 
     /**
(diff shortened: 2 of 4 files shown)


bwindels:
Not ready for review yet, need to fix something with cancelling a redaction

bwindels:
Ok, ready for review now.

src/models/room.js
```diff
@@ -1141,6 +1136,25 @@ Room.prototype.addPendingEvent = function(event, txnId) {
 
     this.emit("Room.localEchoUpdated", event, this, null, null);
 };
+// live events are aggregated in the timelineset
+// but for local echo (and undoing the local echo of a redaction) we do it here.
```
jryans: I don't think I am following this line very well... Maybe something like:

```suggestion
// Since we want local echoes of relations (and to undo the local echo of redacted 
// relations), we aggregate them into the timeline set's relation storage instead of 
// keeping them detached as we do for regular messages.
```

src/models/room.js
```diff
@@ -1277,19 +1291,37 @@ Room.prototype.updatePendingEvent = function(event, newStatus, newEventId) {
...
+    }
+    const redactedEvent = this.getUnfilteredTimelineSet()
+        .findEventById(redactId);
+    if (redactedEvent) {
+        redactedEvent.unmarkLocallyRedacted();
+        // re-render after undoing redaction
+        this.emit("Room.redaction", redactionEvent, this);
+        // reapply relation now redaction failed
+        if (redactedEvent.isRelation()) {
+            this._aggregateNonLiveEvent(redactedEvent);
```
jryans: Hmm, does this re-add the relation to the relations collection, meaning there are now two copies of the relation, one that's still marked redacted and another live one? If so, I am worried that will get a bit confusing to keep track of.

Since we already have the relations collection listening for redaction, could it listen for unredaction and revive the event internally as opposed to duplicating it?

src/models/room.js
```diff
@@ -1277,19 +1291,37 @@ Room.prototype.updatePendingEvent = function(event, newStatus, newEventId) {
...
+    const redactId = redactionEvent.event.redacts;
+    if (!redactId) {
+        return;
+    }
+    const redactedEvent = this.getUnfilteredTimelineSet()
+        .findEventById(redactId);
+    if (redactedEvent) {
+        redactedEvent.unmarkLocallyRedacted();
+        // re-render after undoing redaction
+        this.emit("Room.redaction", redactionEvent, this);
```
jryans: Would a different event be better, in case listeners want to take a separate action for undoing reactions? Or is there a way to distinguish the two from the args?

src/models/event.js
```diff
@@ -721,7 +749,7 @@ utils.extend(module.exports.MatrixEvent.prototype, {
      * @return {boolean} True if this event has been redacted
      */
     isRedacted: function() {
-        return Boolean(this.getUnsigned().redacted_because);
+        return this._locallyRedacted || Boolean(this.getUnsigned().redacted_because);
```
jryans: `markLocallyRedacted` seems to also set `redacted_because`, so perhaps we don't need to check `this._locallyRedacted` here?

src/base-apis.js
```diff
@@ -896,7 +896,7 @@ MatrixBaseApis.prototype.sendStateEvent = function(roomId, eventType, content, s
  * @return {module:client.Promise} Resolves: TODO
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
-MatrixBaseApis.prototype.redactEvent = function(
+MatrixBaseApis.prototype.sendRedaction = function(
```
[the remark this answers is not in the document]
bwindels: Indeed, unused. We do this from `_sendEventHttpRequest` now.

src/models/room.js
```diff
@@ -1141,6 +1136,25 @@ Room.prototype.addPendingEvent = function(event, txnId) {
 
     this.emit("Room.localEchoUpdated", event, this, null, null);
 };
+// live events are aggregated in the timelineset
```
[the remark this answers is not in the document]
bwindels: Yeah, I went with non-live because it's also used to reapply a relation when the redaction for it is cancelled. In this case the relation might not just be a local echo anymore, but already have gone through the timelineset addliveevent path before the local echo of the redaction was applied to it. This is why I was hesitant to go with a name that has local echo in it. `_aggregateRelationEvent` seems better indeed but doesn't explain why this is not used for live events, in that one could assume this is the only path that relations get aggregated when reading it.

How about `_aggregateNonLiveRelation` with a better comment was is meant by non-live?

src/models/room.js
bwindels: At this point, the relation event shouldn't be aggregated in `Relations` anymore, as the local echo for the redaction emits `Event.beforeRedaction`, which causes `Relations::_onBeforeRedaction` to remove the relation. This is also why we need to add it again when the redaction is cancelled.

> Since we already have the relations collection listening for redaction, could it listen for unredaction and revive the event internally as opposed to duplicating it?

I considered this, but apart from this case, there is nothing interested in "unredaction" so it seemed a bit overkill to add an event for just this. The only duplication I can see here would be between aggregating a live event, and re-aggregating the relation on "unredaction" here. If we had an "unredaction" event, these paths wouldn't be unified, just these 3 lines would move to Relations... (+ the code for the unredaction event). This seems more direct to me, but the opposite would indeed be more symmetrical with the redaction event to me. Your choice :)

src/models/room.js
bwindels: Initially I added a boolean flag as the third parameter, but nothing would use it. Only the `TimelinePanel` listens for `Room.redaction" and it does a `forceUpdate`, which is also what we do now on "unredaction". Seemed a bit artificial to add an event that calls the same event handler there, or duplicates it?

src/models/room.js
bwindels: Please let me know if it's still hard to follow after the changes I made.

src/models/room.js
jryans: Well, since the JS SDK has other consumer than just the React SDK, I wonder if others would want handle them differently...? (I have no idea.) I guess I am just thinking that if I were using this SDK in my app, I might be surprised by the same event be reused for the opposite action. Up to you though, maybe we just wait and see if hear it's a concern...?

src/models/room.js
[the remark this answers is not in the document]
jryans: Thanks, I think it's clearer now with the updated comment. 😁

src/models/room.js
jryans: It's much better now, thanks! 😁

src/models/room.js
jryans: Ah okay, you're right, it does get removed from the collection on `Event.beforeRedaction`, so I think it's fine as you have it.

I am starting to worry we've got little bits of redaction code spread all over the place, which feels hard to maintain, but that's not specific to this PR, so we can think about how to improve it separately.

src/models/room.js
bwindels: That is a fair point, I'll add a `Room.redactionCancelled` event then.

PR #941, opened 2019-06-05
Encode event IDs when redacting events

turt2live:
Because v3 rooms are a thing.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -1887,8 +1887,10 @@ function _sendEventHttpRequest(client, event) {
         }
         path = utils.encodeUri(pathTemplate, pathParams);
     } else if (event.getType() === "m.room.redaction") {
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -1887,8 +1887,10 @@ function _sendEventHttpRequest(client, event) {
         }
         path = utils.encodeUri(pathTemplate, pathParams);
     } else if (event.getType() === "m.room.redaction") {
-        const pathTemplate = `/rooms/$roomId/redact/${event.event.redacts}/$txnId`;
-        path = utils.encodeUri(pathTemplate, pathParams);
+        const pathTemplate = `/rooms/$roomId/redact/$redactsEventId/$txnId`;
+        path = utils.encodeUri(pathTemplate, Object.assign({
```
turt2live: I would have loved to use `...pathParams`, but our version of JS in the sdk doesn't support it.

PR #944, opened 2019-06-07
Fix backup sig validation with multiple sigs

dbkr:
verifySignature modifies the object so we need to clone if we're
verifying more than one signature.

Fixes https://github.com/vector-im/riot-web/issues/9357

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -371,7 +371,9 @@ Crypto.prototype.isKeyBackupTrusted = async function(backupInfo) {
             try {
                 await olmlib.verifySignature(
                     this._olmDevice,
-                    backupInfo.auth_data,
+                    // verifySignature modifies the object so we need to copy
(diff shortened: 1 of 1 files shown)


src/crypto/index.js
```diff
@@ -371,7 +371,9 @@ Crypto.prototype.isKeyBackupTrusted = async function(backupInfo) {
             try {
                 await olmlib.verifySignature(
                     this._olmDevice,
-                    backupInfo.auth_data,
+                    // verifySignature modifies the object so we need to copy
```
jryans: This is fine as a workaround, but maybe `verifySignature` should also be changed to not do that...? (It seems like other might also be confused by it...)

dbkr: Yeah - I'm slightly hesitant to do it without fully understanding why it was done like that, although I suspect just for speed.

PR #950, opened 2019-06-10
Remove check for `file.stream` in HttpApi.uploadContent

### LINKED ISSUE #949: File uploads are broken on latest Firefox Nightly
Under the latest version of Firefox Nightly, upon uploading a file, the following text is always uploaded in place of the files real contents
```
function stream() {
    [native code]
}
```

This seems to be caused by [http-api.js:70](https://github.com/matrix-org/matrix-js-sdk/blob/a0456dc43041af4528e57dfd393da701a4692732/src/http-api.js#L170):
```js
const body = file.stream ? file.stream : file;
```

On other browsers, `File.stream` will be undefined; however, Firefox Nightly now adheres to the latest draft of the [File API](https://w3c.github.io/FileAPI/) which defines a [`.stream()` method](https://w3c.github.io/FileAPI/#stream-method-algo), meaning this line is swapping the content of the file for this `stream` method and later stringifying and uploading that method.

The original purpose of this line seems to be to keep compatibility with older code, which was supposedly encouraged to set `file.stream` to the stream to be uploaded. I'm not sure if this is still necessary, especially considering this is apparently undocumented behavior. Deleting this line entirely solves the issue.

If it is necessary to keep compatibility, we could just ensure the `file` isn't a `File` before checking for `file.stream`.

This may be the same issue as #908, as the description of that issue sounds similar.

dnaf:
This fixes an incompatibility with the latest draft of the W3C File API and fixes #949, which is related to vector-im/riot-web#9924 and makes progress towards closing that issue. I believe simply removing these lines is reasonable, as the functionality they implement is undocumented.

Signed-off-by: Katie Wolfe <katie@dnaf.moe>

### DIFF
--- a/src/http-api.js
+++ b/src/http-api.js
@@ -165,10 +165,6 @@ module.exports.MatrixHttpApi.prototype = {
         const contentType = opts.type || file.type || 'application/octet-stream';
         const fileName = opts.name || file.name;
 
-        // we used to recommend setting file.stream to the thing to upload on
-        // nodejs.
-        const body = file.stream ? file.stream : file;
-
         // backwards-compatibility hacks where we used to do different things
         // between browser and node.
         let rawResponse = opts.rawResponse;


jryans:
Thanks for filing an issue and working on this fix @dnaf! 😁 More people have noticed this problem, so I've decided to submit [my own PR](https://github.com/matrix-org/matrix-js-sdk/pull/952) for this so we can hopefully get this fixed very soon. I'll close this PR as I don't believe it's needed now. Again, thanks for reporting the problem!

PR #951, opened 2019-06-10
Don't overlap auth submissions with polls

dbkr:
Wait for polls to complete before submitting auth dicts, otherwise
we risk the requests overlapping and both returning a 200.

Also introduces a setBusy interface to interactive-auth to explicitly
set the busy status, since otherwise it doesn't get set until the
request actually starts.

Requires https://github.com/matrix-org/matrix-react-sdk/pull/3085

### DIFF
--- a/src/interactive-auth.js
+++ b/src/interactive-auth.js
@@ -55,6 +55,12 @@ const MSISDN_STAGE_TYPE = "m.login.msisdn";
  *     promise which resolves to the successful response or rejects with a
  *     MatrixError.
  *
+ * @param {function(bool): module:client.Promise} opts.setBusy
+ *     called whenever the interactive auth logic is busy submitting
+ *     information provided by the user. After this has been called with
+ *     true the UI should indicate that a request is in progress until it
+ *     is called again with false.
+ *
  * @param {function(string, object?)} opts.stateUpdated
  *     called when the status of the UI auth changes, ie. when the state of
  *     an auth stage changes of when the auth flow moves to a new stage.
@@ -101,6 +107,7 @@ function InteractiveAuth(opts) {
     this._matrixClient = opts.matrixClient;
     this._data = opts.authData || {};
     this._requestCallback = opts.doRequest;
+    this._setBusyCallback = opts.setBusy;
     // startAuthStage included for backwards compat
     this._stateUpdatedCallback = opts.stateUpdated || opts.startAuthStage;
     this._resolveFunc = null;
@@ -117,7 +124,9 @@ function InteractiveAuth(opts) {
     this._chosenFlow = null;
     this._currentStage = null;
 
-    this._polling = false;
+    // if we are currently trying to submit an auth dict (which includes polling)
+    // the promise the will resolve/reject when it completes
+    this._submitPromise = false;
 }
 
 InteractiveAuth.prototype = {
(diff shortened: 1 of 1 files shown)


src/interactive-auth.js
```diff
@@ -117,7 +124,9 @@ function InteractiveAuth(opts) {
     this._chosenFlow = null;
     this._currentStage = null;
 
-    this._polling = false;
+    // if we are currently trying to submit an auth dict (which includes polling)
+    // the promise the will resolve/reject when it completes
+    this._submitPromise = false;
```
jryans: Maybe `null` is more natural to indicate the state without a `Promise` value?

```suggestion
    this._submitPromise = null;
```

src/interactive-auth.js
```diff
@@ -101,6 +107,7 @@ function InteractiveAuth(opts) {
     this._matrixClient = opts.matrixClient;
     this._data = opts.authData || {};
     this._requestCallback = opts.doRequest;
+    this._setBusyCallback = opts.setBusy;
```
jryans: Maybe `busyChanged` for the name of the option key? Not sure how callbacks are typically named in JS SDK...

src/interactive-auth.js
```diff
@@ -138,7 +147,10 @@ InteractiveAuth.prototype = {
             // if we have no flows, try a request (we'll have
             // just a session ID in _data if resuming)
             if (!this._data.flows) {
-                this._doRequest(this._data);
+                this._setBusyCallback(true);
```
jryans: Shouldn't the callback be optional? If so, we need to check for the callback everywhere it's used or else set it to a no-op function.

src/interactive-auth.js
```diff
@@ -235,13 +244,37 @@ InteractiveAuth.prototype = {
...
+
         // use the sessionid from the last request.
         const auth = {
             session: this._data.session,
         };
         utils.extend(auth, authData);
 
-        await this._doRequest(auth, background);
+        try {
+            // NB. the 'background' flag is deprecated by the setBusy
```
jryans: This deprecation should be copied up to the comments on the `background` param as well.

src/interactive-auth.js
dbkr: oops - did it in one place but nowhere else

src/interactive-auth.js
```diff
@@ -235,13 +244,39 @@ InteractiveAuth.prototype = {
...
+
         // use the sessionid from the last request.
         const auth = {
             session: this._data.session,
         };
         utils.extend(auth, authData);
 
-        await this._doRequest(auth, background);
+        try {
+            // NB. the 'background' flag is deprecated by the busyChanged
```
jryans: Ah, when I mentioned documenting the deprecation, I was actually thinking of the `background` param docs on this `submitAuthDict` method. Oh but maybe this `submitAuthDict` is only called internally by other method of this object...?

src/interactive-auth.js
```diff
@@ -49,11 +49,17 @@ const MSISDN_STAGE_TYPE = "m.login.msisdn";
...
  *    null, a request will be made with no auth before starting.
  *
- * @param {function(object?, bool?): module:client.Promise} opts.doRequest
- *     called with the new auth dict to submit the request and a flag set
- *     to true if this request is a background request. Should return a
- *     promise which resolves to the successful response or rejects with a
- *     MatrixError.
+ * @param {function(object?): module:client.Promise} opts.doRequest
+ *     called with the new auth dict to submit the request. Also passes a
+ *     second deprecated arg which is a flag set to true if this request
```
jryans: It might be nice to mention what's meant to replace this deprecated thing in these comments (the `busyChanged` callback) so it's easier to follow what we now recommend here.

dbkr: Oops, good point

src/interactive-auth.js
dbkr: Ah, so, it's a little weird in that the `submitAuthDict` is public and called from externally, but never with the `background` param - also, the background param is still used at this level, just not passed out to the app.

PR #952, opened 2019-06-11
Fix content uploads for modern browsers

jryans:
Modern browsers now expose a `stream` function on the Blob and File interfaces.
This conflicts with an older style of passing data to the `uploadContent` SDK
method, which supported supplying the data to upload in the `stream` property of
an object.

Since this old style is still in active use in the Matrix JS ecosystem, we
preserve the backwards compatibility for now by checking whether `stream` is a
function.

This fix has been tested in Firefox Nightly (69), Firefox Release (67), Chrome
Canary (77), and Chrome Stable (75).

Fixes https://github.com/vector-im/riot-web/issues/9913
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/949

### DIFF
--- a/src/http-api.js
+++ b/src/http-api.js
@@ -165,9 +165,21 @@ module.exports.MatrixHttpApi.prototype = {
         const contentType = opts.type || file.type || 'application/octet-stream';
         const fileName = opts.name || file.name;
 
-        // we used to recommend setting file.stream to the thing to upload on
-        // nodejs.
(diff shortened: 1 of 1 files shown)


ser:
I think it is that important that should be moved from "develop" as soon as possible

jryans:
> I think it is that important that should be moved from "develop" as soon as possible

It was released in v2.0.1 of JS SDK on 2019-06-19 alongside Riot v1.2.2.

PR #954, opened 2019-06-13
Key verification request fixes

uhoreg:
- fix requestVerification in MatrixClient to match the function in crypto
  - reorder the arguments so that the arguments actually do what they say they
    do
  - pass through the third argument, which was accidentally omitted
- ignore verification requests from ourselves
- also fix a comment

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -734,19 +734,19 @@ async function _setDeviceVerification(
  * Request a key verification from another user.
  *
  * @param {string} userId the user to request verification with
- * @param {Array} devices array of device IDs to send requests to.  Defaults to
- *    all devices owned by the user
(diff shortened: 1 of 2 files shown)


uhoreg:
> Is anything using this that's going to be broken by the argument re-ordering?

No, because even though the arguments are reordered in the MatrixClient method, the order in which they're passed isn't (i.e. the second argument of MatrixClient#requestVerification gets passed as the second argument to crypto#requestVerification), and the order of the crypto versions doesn't change in this PR.  So in fact, someone who tried to use the currently documented order of arguments would have wound up with something that doesn't work properly, since it would have passed `devices` into `methods`.  And given that nobody has complained yet, I can be pretty certain that nobody is using this function with the second or third arguments. ;)

PR #955, opened 2019-06-14
README: clarify protocols

celeduc:
Olm was removed as a supported protocol in this commit: 3b09ab3ca142a45a9be8d2d393053042ec020104 but the sentence no longer makes sense. Changing the sentence so that it clarifies that Megolm is the only protocol supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,7 +296,7 @@ Then visit ``http://localhost:8005`` to see the API docs.
 End-to-end encryption support
 =============================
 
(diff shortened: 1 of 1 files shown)


turt2live:
Olm should have never left - I've fixed this directly here: https://github.com/matrix-org/matrix-js-sdk/commit/0121bdbb75467f0a275d706d747f32be2e65055c

Thanks for pointing it out!

PR #960, opened 2019-06-18
return 'sending' status for an event that is only locally redacted

bwindels:
Replacement PR for https://github.com/matrix-org/matrix-js-sdk/pull/959 to make branch names same as https://github.com/matrix-org/matrix-react-sdk/pull/3117, so karma tests can pass.

Related to https://github.com/vector-im/riot-web/issues/10080


bwindels:
merging as this incorporates all suggestions from #959 which was approved.

PR #961, opened 2019-06-19
Time out verification attempts after 10 minutes of inactivity

turt2live:
Fixes https://github.com/vector-im/riot-web/issues/10117
Fixes https://github.com/vector-im/riot-web/issues/9103 (I think?)

### DIFF
--- a/src/crypto/verification/Base.js
+++ b/src/crypto/verification/Base.js
@@ -22,6 +22,9 @@ limitations under the License.
 import {MatrixEvent} from '../../models/event';
 import {EventEmitter} from 'events';
 import logger from '../../logger';
+import {newTimeoutError} from "./Error";
+
+const timeoutException = new Error("Verification timed out");
 
 export default class VerificationBase extends EventEmitter {
     /**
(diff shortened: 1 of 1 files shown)


src/crypto/verification/Base.js
```diff
@@ -60,9 +63,25 @@ export default class VerificationBase extends EventEmitter {
...
         this._promise = null;
+        this._transactionTimeoutTimer = null;
+
+        // At this point, the verification request was received so start the timeout timer.
+        this._pingTransaction();
+    }
+
+    _pingTransaction() {
+        console.log("Refreshing/starting the verification transaction timeout timer");
+        clearTimeout(this._transactionTimeoutTimer);
```
dbkr: I assume the browser is fine with you passing null here, but probably worth a null check anyway.

src/crypto/verification/Base.js
```diff
@@ -60,9 +63,25 @@ export default class VerificationBase extends EventEmitter {
...
         this._parent = parent;
         this._done = false;
         this._promise = null;
+        this._transactionTimeoutTimer = null;
+
+        // At this point, the verification request was received so start the timeout timer.
+        this._pingTransaction();
+    }
+
+    _pingTransaction() {
```
dbkr: maybe `resetTimer`?

src/crypto/verification/Base.js
turt2live: It's very fine with it. The JS spec even says it should silently do nothing for invalid handles.

I'll add a null check though

PR #963, opened 2019-06-23
Use cameCase instead of underscore

huan:
To make all the variable names to be consistent


huan:
Signed-off-by: Huan LI (李卓桓) <zixia@zixia.net>

PR #964, opened 2019-06-24
Emit original errcode by default in http-api

Yoshiin:
In http-api emit the errcode by default.

Please advise if you want an other format of the emitted event.

Signed-off-by: Léo Mora l.mora@outlook.fr

### DIFF
--- a/src/http-api.js
+++ b/src/http-api.js
@@ -477,6 +477,8 @@ module.exports.MatrixHttpApi.prototype = {
                     err.message,
                     err.data.consent_uri,
                 );
+            } else {
+                self.event_emitter.emit(err.errcode);
             }
         });
 


Yoshiin:
See https://github.com/dinsic-pim/tchap-web/issues/40 for the case.
If you prefer, I can only emit the event in the case of `ORG_MATRIX_EXPIRED_ACCOUNT` is returned.

dbkr:
You do get the error from sync requests as the `error` key in the `data` parameter of the `sync` event - would that be good enough?

giomfo:
@dbkr: I understand your suggestion. The problem is this will not prevent the web-client from displaying an unexpected error dialog (like for example when we load the profile info).
The `ORG_MATRIX_EXPIRED_ACCOUNT` error is returned by all authenticated requests as soon as the account has expired.
What do you think to emit only the errors with `ORG_MATRIX_` prefix? (in order to handle all requests which may return this error)

dbkr:
Hmm - an error being an ORG_MATRIX doesn't really mean it will be one the whole app needs to care about.

How would this prevent dialogs? The js-sdk call would still throw an exception or fail in some way, so which I assume is what is causing the error dialog?

giomfo:
> an error being an ORG_MATRIX doesn't really mean it will be one the whole app needs to care about.

I agree, but there is no other way to hook this error from any requests, isn't there?

> How would this prevent dialogs?

I don't really know, my plan was to copy the `M_CONSENT_NOT_GIVEN` handling ;)

In conclusion, if there is no actual way to catch this kind of error, we will handle this unspecified error `ORG_MATRIX_EXPIRED_ACCOUNT` from the `/sync` response. Can you please point here where this error management should take place in matrix-react-sdk? Is there only one place to catch `/sync` error? 
Some specific actions must be performed on this error: clear cache, prompt the user. Where do you think @Yoshiin should handle these actions in matrix-react-sdk?

dbkr:
You'll get the error in https://github.com/matrix-org/matrix-react-sdk/blob/develop/src/components/structures/MatrixChat.js#L1298 so that would be the place

Yoshiin:
I will take the @dbkr solution. No more dispatch modification needed.

PR #969, opened 2019-06-27
Handle relations in encrypted rooms

bwindels:
Part of https://github.com/matrix-org/matrix-react-sdk/pull/3151

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -1886,6 +1886,21 @@ function _encryptEventIfNeeded(client, event, room) {
 
     return client._crypto.encryptEvent(event, room);
 }
+/**
+ * Returns the eventType that should be used taking encryption into account
+ * for a given eventType.
+ * @param {MatrixClient} client the client
+ * @param {string} roomId the room for the events `eventType` relates to
+ * @param {string} eventType the event type
+ * @return {string} the event type taking encryption into account
+ */
+function _getEncryptedIfNeededEventType(client, roomId, eventType) {
+    if (eventType === "m.reaction") {
+        return eventType;
+    }
+    const isEncrypted = client.isRoomEncrypted(roomId);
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -3991,14 +3993,25 @@ MatrixClient.prototype.getCanResetTimelineCallback = function() {
  */
 MatrixClient.prototype.relations =
 async function(roomId, eventId, relationType, eventType, opts = {}) {
+    const isEncrypted = this.isRoomEncrypted(roomId);
+    const fetchedEventType = isEncrypted ? "m.room.encrypted" : eventType;
```
jryans: Hmm, so this `fetchedEventType` will be wrong for reactions, which are now always `m.reaction` even in E2E rooms. 😓 It does seem nice to handle encryption transparently though... Maybe we should add a helper function similar to `_encryptEventIfNeeded` called something like `_getEncryptedIfNeededEventType` that returns the eventual event type (with a similar exception for reactions)?

src/client.js
```diff
@@ -3991,14 +4008,25 @@ MatrixClient.prototype.getCanResetTimelineCallback = function() {
...
         eventId,
         relationType,
-        eventType,
+        fetchedEventType,
         opts);
+
+    let events = result.chunk.map(this.getEventMapper());
+    if (isEncrypted) {
+        await Promise.all(events.map(e => {
+            return new Promise(resolve => e.once("Event.decrypted", resolve));
```
jryans: Reactions in an encrypted room won't be encrypted events themselves, so we can't use `isEncrypted` for the whole room to decide whether to await `Event.decrypted`.

Since the event mapper will invoke `attemptDecryption`, I think you only want to wait for `Event.decrypted` if `event.isBeingDecrypted()`.

PR #974, opened 2019-07-02
Handle self read receipts for fixing e2e notification counts

turt2live:
Fixes https://github.com/vector-im/riot-web/issues/9421

This also adds a context to the ReEmitter so we have access to the Room at the time of read receipt. Without this, we have to bind handlers to every encrypted room (which is tedious to maintain) or figure out which room `$something` belong to (CPU intensive).

### DIFF
--- a/src/ReEmitter.js
+++ b/src/ReEmitter.js
@@ -34,12 +34,18 @@ export default class Reemitter {
     }
 
     reEmit(source, eventNames) {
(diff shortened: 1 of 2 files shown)


turt2live:
Putting it in the room pulls it away from the existing code which handles this, and we'd have upwards of a few hundred listeners.

PR #976, opened 2019-07-04
Add default push rule to ignore reactions

jryans:
This adds a default push rule to ignore reactions as proposed in
[MSC2153](https://github.com/matrix-org/matrix-doc/pull/2153). By adding it here
in the client directly, we can try out the idea early even if it hasn't appeared
in the user's HS yet.

Part of https://github.com/vector-im/riot-web/issues/10208

### DIFF
--- a/src/pushprocessor.js
+++ b/src/pushprocessor.js
@@ -52,6 +52,22 @@ const DEFAULT_OVERRIDE_RULES = [
             },
         ],
     },
+    {
+        // For homeservers which don't support MSC2153 yet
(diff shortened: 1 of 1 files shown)


jryans:
> Does this need a react-sdk component similar to the tombstone (or are we okay with it being one of the many hidden options)?

I think it's okay for it to be hidden for the moment. If we later find out that there in fact are many people who want reactions to notify, we can expose an option for that.

jryans:
Dave has approved the MSC, so I'll proceed with landing this.

PR #977, opened 2019-07-04
Upgrade dependencies

jryans:


jryans:
The PR is the result of running `yarn upgrade`, which doesn't appear to touch `package.json`. I then tried `yarn upgrade-interactive` and `syncyarnlock`, but neither did the right thing. 😭

PR #978, opened 2019-07-04
Fix not hiding redacted edits

bwindels:
Fixes https://github.com/vector-im/riot-web/issues/10112

Preserves rel_type after redacting an edit, so we still know it wasn't a normal message

### DIFF
--- a/src/models/event.js
+++ b/src/models/event.js
@@ -713,6 +713,9 @@ utils.extend(module.exports.MatrixEvent.prototype, {
 
         this.emit("Event.beforeRedaction", this, redaction_event);
 
+        // before removing content, preserve the rel_type
+        // if this is a relation, so we can still tell the
+        // difference between a redacted edit and message
         const relation = this.getRelation();
         if (relation) {
             this._relationTypeBeforeRedaction = relation.rel_type;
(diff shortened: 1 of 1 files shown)


src/models/event.js
```diff
@@ -712,6 +713,13 @@ utils.extend(module.exports.MatrixEvent.prototype, {
 
         this.emit("Event.beforeRedaction", this, redaction_event);
 
+        // before removing content, preserve the rel_type
+        // if this is a relation, so we can still tell the
+        // difference between a redacted edit and message
+        const relation = this.getRelation();
+        if (relation) {
+            this._relationTypeBeforeRedaction = relation.rel_type;
```
turt2live: is this field persisted? Concern would be it showing correctly at the time it happened, but when switching rooms or refreshing the page the double redaction comes back.

bwindels: Don't think it needs to be. All events are kept in memory, so switching rooms shouldn't clear this.
As for refreshing the page, both the edit and its redaction should be loaded in the same order when processing the sync accumulator blob, so that should also work.

bwindels: Argh, you might be right. Just saw them popup again (fetched through /messages I presume). I was assuming redacted messages wouldn't show up through /messages, but looks like that is not the case...

bwindels:
This needs backend support to solve properly, see https://github.com/matrix-org/matrix-doc/pull/2154#discussion_r300725473

PR #981, opened 2019-07-05
Block read marker and read receipt from advancing into pending events

jryans:
This changes the methods that update the read marker and read receipts to
prevent advancing into pending events.

Part of https://github.com/vector-im/riot-web/issues/9952

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -2148,7 +2148,12 @@ MatrixClient.prototype.sendReceipt = function(event, receiptType, callback) {
  * @return {module:client.Promise} Resolves: TODO
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
-MatrixClient.prototype.sendReadReceipt = function(event, callback) {
+MatrixClient.prototype.sendReadReceipt = async function(event, callback) {
+    const eventId = event.getId();
+    const room = this.getRoom(event.getRoomId());
+    if (room && room.hasPendingEvent(eventId)) {
+        throw new Error(`Cannot set read receipt to a pending event (${eventId})`);
+    }
     return this.sendReceipt(event, "m.read", callback);
 };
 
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -2148,7 +2149,12 @@ MatrixClient.prototype.sendReceipt = function(event, receiptType, callback) {
  * @return {module:client.Promise} Resolves: TODO
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
-MatrixClient.prototype.sendReadReceipt = function(event, callback) {
+MatrixClient.prototype.sendReadReceipt = async function(event, callback) {
+    const eventId = event.getId();
+    const room = this.getRoom(event.getRoomId());
+    if (room && room.hasPendingEvent(eventId)) {
+        throw new Error(`Cannot set read receipt to a pending event (${eventId})`);
```
turt2live: should we log instead of throwing?

src/client.js
```diff
@@ -2158,20 +2164,25 @@ MatrixClient.prototype.sendReadReceipt = function(event, callback) {
...
  * other. The local echo of this receipt will be done if set. Optional.
  * @return {module:client.Promise} Resolves: the empty object, {}.
  */
-MatrixClient.prototype.setRoomReadMarkers = function(roomId, eventId, rrEvent) {
-    const rmEventId = eventId;
-    let rrEventId;
+MatrixClient.prototype.setRoomReadMarkers = async function(roomId, rmEventId, rrEvent) {
+    const room = this.getRoom(roomId);
+    if (room && room.hasPendingEvent(rmEventId)) {
+        throw new Error(`Cannot set read marker to a pending event (${rmEventId})`);
```
turt2live: likewise: log instead of throw?

src/client.js
```diff
@@ -2158,20 +2164,25 @@ MatrixClient.prototype.sendReadReceipt = function(event, callback) {
...
+    }
 
     // Add the optional RR update, do local echo like `sendReceipt`
+    let rrEventId;
     if (rrEvent) {
         rrEventId = rrEvent.getId();
-        const room = this.getRoom(roomId);
+        if (room && room.hasPendingEvent(rrEventId)) {
+            throw new Error(`Cannot set read receipt to a pending event (${rrEventId})`);
```
turt2live: log over throw?

src/models/room.js
```diff
@@ -346,13 +347,30 @@ Room.prototype.userMayUpgradeRoom = function(userId) {
 Room.prototype.getPendingEvents = function() {
     if (this._opts.pendingEventOrdering !== "detached") {
         throw new Error(
-            "Cannot call getPendingEventList with pendingEventOrdering == " +
+            "Cannot call getPendingEvents with pendingEventOrdering == " +
```
turt2live: thank you for fixing this - I spent  far too much time debugging an issue here recently only to discover the function doesn't exist >:(

src/client.js
jryans: I think it's important to throw in these cases so that the [calling code](https://github.com/matrix-org/matrix-react-sdk/pull/3184/files#diff-9dd3a0bf58495df3165ed62d2237e0eeR709) knows to revert its own state and retry again later.

turt2live: ah, fair.

PR #982, opened 2019-07-05
Add a function to flag keys for backup without scheduling a backup

turt2live:
For https://github.com/vector-im/riot-web/issues/10263 / https://github.com/matrix-org/matrix-react-sdk/pull/3187

Starting/scheduling the backup won't help us because the token would be invalid from a server perspective. Instead, we should update what needs to be done and return a count.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -1208,6 +1208,19 @@ MatrixClient.prototype.scheduleAllGroupSessionsForBackup = async function() {
     await this._crypto.scheduleAllGroupSessionsForBackup();
 };
 
+/**
+ * Marks all group sessions as needing to be backed up without scheduling
+ * them to upload in the background.
+ * @returns {Promise<int>} Resolves to the number of sessions requiring a backup.
+ */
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -1208,6 +1208,19 @@ MatrixClient.prototype.scheduleAllGroupSessionsForBackup = async function() {
     await this._crypto.scheduleAllGroupSessionsForBackup();
 };
 
+/**
+ * Marks all group sessions as needing to be backed up without scheduling
+ * them to upload in the background.
+ * @returns {Promise<int>} Resolves to the number of sessions requiring a backup.
+ */
+MatrixClient.prototype.flagAllGroupSessionsForBackup = function() {
```
bwindels: the API of this method is a bit confusing, as it seems both query and command, but I guess the reason will become clear once I see the usage for it in the next PR :)

turt2live: it is a bit confusing, but I'd argue it's no worse than `select * from x` telling you how many rows it fetched.

PR #986, opened 2019-07-09
Don't accept any locally known edits earlier than the last known server-side aggregated edit

bwindels:
This will prevent riot showing the wrong content for an edit event if we have a later SSA edit. This *does* trigger https://github.com/vector-im/riot-web/issues/10290 when redacting the last edit but that is a lesser issue IMO, and needs fixing separately.

### DIFF
--- a/src/models/event.js
+++ b/src/models/event.js
@@ -898,7 +898,7 @@ utils.extend(module.exports.MatrixEvent.prototype, {
         return this.status;
     },
 
-    _getServerAggregatedRelation(relType) {
+    getServerAggregatedRelation(relType) {
         const relations = this.getUnsigned()["m.relations"];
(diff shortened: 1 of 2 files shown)


bwindels:
hmm, it's not *super* clear-cut which is the most important issue to solve here (where solving them both [will be some work](https://github.com/vector-im/riot-web/issues/10290#issuecomment-510065807)):

**not seeing the correct content for an edited message when following a permalink**
(under the condition that the last edit happened far enough ahead of the original event that it's not paged in when viewing the permalink)

versus

**redacting the currently last edit doesn't update the content**
(under the condition that there was a gappy sync when to original event was received)

Both edge cases, so fair to merge this.

PR #996, opened 2019-07-12
Fix incorrect autodiscovery error handling

arcnmx:
`response` can be undefined if a connection error occurrs, in which case you'll get:

```
TypeError: Cannot read property 'statusCode' of undefined
```

### DIFF
--- a/src/autodiscovery.js
+++ b/src/autodiscovery.js
@@ -495,7 +495,7 @@ export class AutoDiscovery {
                     if (err || response.statusCode < 200 || response.statusCode >= 300) {
                         let action = "FAIL_PROMPT";
                         let reason = (err ? err.message : null) || "General failure";
-                        if (response.statusCode === 404) {
+                        if (!err && response.statusCode === 404) {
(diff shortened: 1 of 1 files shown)


src/autodiscovery.js
```diff
@@ -495,7 +495,7 @@ export class AutoDiscovery {
                     if (err || response.statusCode < 200 || response.statusCode >= 300) {
                         let action = "FAIL_PROMPT";
                         let reason = (err ? err.message : null) || "General failure";
-                        if (response.statusCode === 404) {
+                        if (!err && response.statusCode === 404) {
```
turt2live: We try and check for positivity in our code as a general rule:
```suggestion
                        if (response&& response.statusCode === 404) {
```

arcnmx:
> is it possible to get sign-off as per our contributing guidelines so we can merge this

Sorry not really interested, feel free to consider this to be a bug report instead?

turt2live:
Please file an issue for bug reports - we can't take PRs as a form of report.

PR #999, opened 2019-07-26
[WIP] Immutable direct chats tracking/implementation

turt2live:
See https://github.com/vector-im/riot-web/issues/10416

This brings the direct chat management down a layer so it may be accessed by bots (and bridges with enough effort).

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -18,6 +18,8 @@ limitations under the License.
 */
 "use strict";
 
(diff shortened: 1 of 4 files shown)


turt2live:
The useful bits of this have been cherrypicked already. The feature is also called Canonical DMs now, and this doesn't even implement the correct version of the proposal. Closing for cleanup, though will leave the branch around for future reference.

PR #1000, opened 2019-07-29
Remove deprecated authedRequestWithPrefix and requestWithPrefix

t3chguy:
replacing as documented with authedRequest

Fixes https://github.com/matrix-org/matrix-js-sdk/issues/991


turt2live:
We should treat it as a major change, although not breaking I think. They've been deprecated for years so it's possible albeit unlikely that someone is using it.

jryans:
> We should treat it as a major change, although not breaking I think.

Okay, so does that mean only minor version bump, like with a new feature, etc.?

turt2live:
It means that I think we should actually measure impact and make a decision off the data. Some searches in GitHub are likely to reveal places where people might be impacted.

I do suspect we can get away with a minor bump though.

turt2live:
@jryans I'm starting to think that we should just push it through a minor release (2.x.0) rather than as a breaking change. The only thing I can find still using these is the IRC bridge, which @Half-Shot can fix if it's still a thing (looks like it's a test for the bridge).

What are your thoughts? (also can we merge it so it gets out of the review column?)

jryans:
Thanks for checking the ecosystem! Minor release sounds fine to me as well then.

Yes, let's merge this. (I should have been clearer upfront that we did not need to resolve the version decision before merging...)

PR #1001, opened 2019-07-29
Tombstone bugfixes

Half-Shot:
This sets some limits on room upgrade history:
 - Tombstones which self reference the room they are sent to cause the history fetching process to cease. 
- If the array which stores historical rooms ever contains two of the same room, the history fetcher returns early.


turt2live:
@Half-Shot this needs a description, and a better title please (the titles are our changelog)

PR #1002, opened 2019-07-29
Add support for IS v2 API with authentication

jryans:
This adds support for the IS v2 API in all existing IS APIs. It also adds new support for request IS account tokens (which are new in the v2 API version).

This will fall back to v1 APIs if v2 is not supported for the moment, but this will eventually be removed once servers are updated.

Implements part of [MSC2140](https://github.com/matrix-org/matrix-doc/pull/2140)
Fixes part of https://github.com/vector-im/riot-web/issues/10408


turt2live:
> This will fall back to v1 APIs if v2 is not supported.

for outside readers: the plan is to not continue this fallback once servers are reasonably updated.

jryans:
> > This will fall back to v1 APIs if v2 is not supported.
> 
> for outside readers: the plan is to not continue this fallback once servers are reasonably updated.

Thanks, wasn't quite sure how to state that at the JS SDK level. Updated original comment to match.

src/base-apis.js
```diff
@@ -1717,22 +1739,49 @@ MatrixBaseApis.prototype.getKeyChanges = function(oldToken, newToken) {
  * @param {string} nextLink Optional If specified, the client will be redirected
  *                 to this link after validation.
  * @param {module:client.callback} callback Optional.
+ * @param {string} isAccessToken The `access_token` field of the Identity Server
```
turt2live: like the react-sdk: `accessToken` or `identityAccessToken` please. Booleans are not strings.

src/base-apis.js
```diff
@@ -1717,22 +1739,49 @@ MatrixBaseApis.prototype.getKeyChanges = function(oldToken, newToken) {
...
-        callback, "POST", "/validate/email/requestToken",
-        params, httpApi.PREFIX_IDENTITY_V1,
-    );
+
+    try {
+        const response = await this._http.idServerRequest(
+            undefined, "POST", "/validate/email/requestToken",
+            params, httpApi.PREFIX_IDENTITY_V2, isAccessToken,
+        );
+        // TODO: Fold callback into above call once v1 path below is removed
```
turt2live: we should also make sure to have a maintenance issue on riot-web to clean this fallback up at a future date. Not a blocker for this PR to be merged, but should be done sometime around the merge at latest.

src/base-apis.js
jryans: Right, sorry for the naming... All fixed! Still not sure why it sounded okay at the time... 😅

src/base-apis.js
jryans: Right, good call. I filed https://github.com/vector-im/riot-web/issues/10443 for this. I'll link to it in the code as well.

PR #1006, opened 2019-07-31
Support rewriting push rules when our internal defaults change

turt2live:
and change our internal default for tombstones

### DIFF
--- a/src/pushprocessor.js
+++ b/src/pushprocessor.js
@@ -43,6 +43,11 @@ const DEFAULT_OVERRIDE_RULES = [
                 key: "type",
                 pattern: "m.room.tombstone",
             },
+            {
+                kind: "event_match",
+                key: "state_key",
+                pattern: "",
+            }
         ],
         actions: [
             "notify",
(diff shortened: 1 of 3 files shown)


src/sync.js
```diff
@@ -1030,8 +1031,9 @@ SyncApi.prototype._processSyncResponse = async function(
                 // honour push rules that were previously cached. Base rules
                 // will be updated when we recieve push rules via getPushRules
                 // (see SyncApi.prototype.sync) before syncing over the network.
-                if (accountDataEvent.getType() == 'm.push_rules') {
-                    client.pushRules = accountDataEvent.getContent();
+                if (accountDataEvent.getType() === 'm.push_rules') {
+                    const rules = accountDataEvent.getContent();
+                    client.pushRules = PushProcessor.rewriteDefaultRules(rules);
```
bwindels: do we still need to do `applyRuleDefaults(client.pushRules)` in `PushProcessor::actionsForEvent` with this change? If not, we could get rid of `applyRuleDefaults`, which is very similar to the `rewriteDefaultRules` method introduced.

src/pushprocessor.js
```diff
@@ -43,6 +43,11 @@ const DEFAULT_OVERRIDE_RULES = [
                 key: "type",
                 pattern: "m.room.tombstone",
             },
+            {
+                kind: "event_match",
```
bwindels: this will match any non-state event? What is this used for?

src/sync.js
turt2live: They do slightly different things for different points in time. The `rewriteDefaultRules` replaces rules the server already sent to the client whereas `applyRuleDefaults` adds rules that are missing from the server. The intention is that we don't add default rules early for fear of confusing the app that the rules exist, as they are likely to try and modify them against the homeserver and not through our `PushProcessor` stuff. When they try and modify them with the homeserver, they are likely to receive errors or problems due to the name of the rule.

TLDR: we need both because they do subtly different things.

PR #1009, opened 2019-08-07
Add API for bulk lookup on the Identity Server

jryans:
This adds support for querying `/bulk_lookup` on the IS to check several 3PIDs
at the same time.

Part of https://github.com/vector-im/riot-web/issues/10159

### DIFF
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -1892,7 +1892,8 @@ MatrixBaseApis.prototype.lookupThreePid = async function(
 /**
  * Looks up the public Matrix ID mappings for multiple 3PIDs.
  *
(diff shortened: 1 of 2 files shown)


src/base-apis.js
```diff
@@ -1889,6 +1889,44 @@ MatrixBaseApis.prototype.lookupThreePid = async function(
     }
 };
 
+/**
+ * Looks up the public Matrix ID mappings for multiple 3PIDs.
+ *
+ * @param {array} query Array of arrays containing [medium, address]
```
t3chguy: the type could perhaps be more explicit like: `Array.<Array.<{string, string}>>`

jryans: Thanks, updated type.

PR #1014, opened 2019-08-12
Throw an exception if trying to do an ID server request with no ID server

dbkr:
Fixes https://github.com/vector-im/riot-web/issues/10540

### DIFF
--- a/src/http-api.js
+++ b/src/http-api.js
@@ -390,7 +390,7 @@ module.exports.MatrixHttpApi.prototype = {
         accessToken,
     ) {
         if (!this.opts.idBaseUrl) {
-            throw new Error("No ID Server Base URL set");
+            throw new Error("No Identity Server base URL set");
         }
 
         const fullUri = this.opts.idBaseUrl + prefix + path;
--- a/src/http-api.js
+++ b/src/http-api.js
@@ -389,6 +389,10 @@ module.exports.MatrixHttpApi.prototype = {
         prefix,
         accessToken,
     ) {
+        if (!this.opts.idBaseUrl) {
+            throw new Error("No ID Server Base URL set");
+        }
+
         const fullUri = this.opts.idBaseUrl + prefix + path;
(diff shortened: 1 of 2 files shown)


src/http-api.js
```diff
@@ -389,6 +389,10 @@ module.exports.MatrixHttpApi.prototype = {
         prefix,
         accessToken,
     ) {
+        if (!this.opts.idBaseUrl) {
+            throw new Error("No ID Server Base URL set");
```
jryans: Maybe let's be slightly more formal and say "identity server"?

```suggestion
            throw new Error("No Identity Server base URL set");
```

I also noticed this much more verbose error style:

https://github.com/matrix-org/matrix-js-sdk/blob/17724fc8d35f75d1a7a751b62da6db98798ef548/src/client.js#L2438-L2442

I guess it's just a historic thing we used to do for SDK errors a while ago?

src/http-api.js
```diff
@@ -389,6 +389,10 @@ module.exports.MatrixHttpApi.prototype = {
         prefix,
         accessToken,
     ) {
+        if (!this.opts.idBaseUrl) {
```
jryans: We should also guard:

* `MatrixBaseApis.prototype.registerWithIdentityServer`

Also not quite sure what this one will do for empty string:

* `MatrixClient.prototype._requestTokenFromEndpoint`

jryans:
It's probably good enough to upgrade the PR saying it fixes this issue. It's all we're likely to do about it at the moment I think...

src/http-api.js
dbkr: Mmm, that looks like it's basically pretending to be a response from the server when it isn't, so not sure about copying that one.

src/http-api.js
dbkr: url.parse will parse the empty string to null everything so the second one if fine as-is

PR #1015, opened 2019-08-13
Change ICE fallback server and make fallback opt-in

jryans:
This changes the STUN fallback server from a Google server to one hosted at
`turn.matrix.org`. This also changes the ICE server fallback to be disabled by default. The SDK consumer will receive a new event in case the homeserver has no ICE servers of its own,
and can prompt the user to agree to the fallback if desired.

Part of https://github.com/vector-im/riot-web/issues/10173
Used by https://github.com/matrix-org/matrix-react-sdk/pull/3309

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -221,6 +221,7 @@ function MatrixClient(opts) {
     this._verificationMethods = opts.verificationMethods;
 
     this._forceTURN = opts.forceTURN || false;
+    this._fallbackICEServerAllowed = false;
 
     // List of which rooms have encryption enabled: separate from crypto because
     // we still want to know which rooms are encrypted even if crypto is disabled:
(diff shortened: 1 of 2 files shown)


jryans:
> (imo it should only show up when you make a call, to avoid throwing dialogs at all the people who haven't bothered to set up a TURN server because they don't make calls)

I also agree it is unfortunate, so you may wish to shout about this... I also don't make calls or inted to make them in future... The reasoning for showing on session start is so that we have prearranged permission for incoming calls. It's not exactly clear if that's a strong enough reason to justify annoying every user on session start. 😓

turt2live:
I've posted my rant at https://github.com/vector-im/riot-web/issues/10546

PR #1017, opened 2019-08-16
Split getCapabilities into server / user

dbkr:
Supports the new user / server capabilities

User capabilities falls back to old-style plain /capabilities
getCapabilities() remains as an alias for getUserCapabilities()

Implements https://github.com/matrix-org/matrix-doc/pull/2233


dbkr:
We're not doing this anymore.

PR #1018, opened 2019-08-16
Add getIdServer() & doesServerRequireIdServerParam()

dbkr:
Remove individual cache for lazy loading and just cache the whole
versions response, then we can cache both of these flags

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -238,7 +238,7 @@ function MatrixClient(opts) {
     // The pushprocessor caches useful things, so keep one and re-use it
     this._pushProcessor = new PushProcessor(this);
 
-    this._serverSupportsLazyLoading = null;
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -4059,15 +4060,38 @@ MatrixClient.prototype.doesServerSupportLazyLoading = async function() {
                 prefix: '',
             },
         );
+    }
+    return this._serverVersionsCache;
```
turt2live: we should expire this after ~8 hours, similar to how capabilities are dealt with. Concern would be someone who runs Riot for days on end (me) and upgrades their server in the background.

dbkr: Done: https://github.com/matrix-org/matrix-js-sdk/issues/1020

PR #1019, opened 2019-08-16
Make requestToken endpoints work without ID Server

dbkr:
Hopefully with doc in appropriate functions

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -3303,7 +3303,7 @@ MatrixClient.prototype.requestRegisterMsisdnToken = function(phoneCountry, phone
  * The doesServerRequireIdServerParam() method can be used to determine if
  * the server requires the id_server parameter to be provided.
  * If an account with the given email address already exists and is
- * associated with an account other than the one the user is authed as, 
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -3428,14 +3422,16 @@ MatrixClient.prototype.requestPasswordMsisdnToken = function(phoneCountry, phone
...
  */
 MatrixClient.prototype._requestTokenFromEndpoint = function(endpoint, params) {
-    const id_server_url = url.parse(this.idBaseUrl);
-    if (id_server_url.host === null) {
-        throw new Error("Invalid ID server URL: " + this.idBaseUrl);
+    const postParams = Object.assign({}, params);
+
+    if (this.idBaseUrl) {
+        const idServerUrl = url.parse(this.idBaseUrl);
+        if (idServerUrl.host === null) {
```
turt2live: might be good to just check for falsey values in the event someone somehow creates a zero-length string as a host.
```suggestion
        if (!idServerUrl.host) {
```

PR #1021, opened 2019-08-21
Use the v2 (hashed) lookup for identity server queries

turt2live:
Fixes https://github.com/vector-im/riot-web/issues/10556
Implements [MSC2134](https://github.com/matrix-org/matrix-doc/pull/2134) with assistance from [MSC2140](https://github.com/matrix-org/matrix-doc/pull/2140) for auth.

Note: this also changes all identity server requests to use JSON as a request body. URL encoded forms were allowed in v1 but deprecated in favour of JSON. v2 APIs do not allow URL encoded forms.

### DIFF
--- a/src/http-api.js
+++ b/src/http-api.js
@@ -412,7 +412,7 @@ module.exports.MatrixHttpApi.prototype = {
         if (method == 'GET') {
             opts.qs = params;
         } else if (typeof params === "object") {
-            opts.form = params;
+            opts.json = params;
         } else if (typeof params === "string") {
             // Assume the caller has serialised the body to JSON
             opts.body = params;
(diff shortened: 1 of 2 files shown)


src/http-api.js
```diff
@@ -412,7 +412,7 @@ module.exports.MatrixHttpApi.prototype = {
         if (method == 'GET') {
             opts.qs = params;
         } else if (typeof params === "object") {
-            opts.form = params;
+            opts.json = params;
         } else if (typeof params === "string") {
             // Assume the caller has serialised the body to JSON
```
jryans: Is it possible to remove this branch and all the places where I plopped `JSON.stringify(params)` in `idServerRequest` call sites? Your approach with `opts.json` seems much more sane, so it'd be nice to just do that everywhere...

turt2live: I can do that in a different PR, I'd rather not make this diff any harder to read if I can avoid it

PR #1022, opened 2019-08-22
Send id_access_token to HS for use in proxied IS requests

jryans:
This passes along the `id_access_token` to the HS, which it will need when
speaking v2 IS APIs to the IS.

Unfortunately, some HSes seem to explode when given this new parameter, so we
only pass it along for the moment if an unstable feature `m.id_access_token` is
also set.

Part of https://github.com/vector-im/riot-web/issues/10525
Used by https://github.com/matrix-org/matrix-react-sdk/pull/3337
Defined in [MSC2140](https://github.com/matrix-org/matrix-doc/pull/2140)

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -108,6 +108,14 @@ function keyFromRecoverySession(session, decryptionKey) {
  *
  * @param {string} opts.userId The user ID for this user.
  *
+ * @param {Function} [opts.getIdentityAccessToken]
+ * Optional. A callback that returns a Promise<String> of an identity access
+ * token to supply with identity requests. If the callback is unset, no access
+ * token will be supplied.
+ * See also https://github.com/vector-im/riot-web/issues/10615 which seeks to
+ * replace the previous approach of manual access tokens params with this
+ * callback throughout the SDK.
+ *
  * @param {Object=} opts.store
  *    The data store used for sync data from the homeserver. If not specified,
  *    this client will not store any HTTP responses. The `createClient` helper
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -108,6 +108,14 @@ function keyFromRecoverySession(session, decryptionKey) {
  *
  * @param {string} opts.userId The user ID for this user.
  *
+ * @param {Function} [opts.getIdentityAccessToken]
```
turt2live: can we change the pattern here and use an `IdentityServerProvider` or similar interface which has a function to implement? I'm not really a fan of passing functions around and treating them like class functions, and would prefer the indirection of `this.identityServer.getAccessToken()`

jryans: Hmm, I mean we could, but I guess I feel like it leaves me with more questions that I wasn't sure how we should answer at this immediate juncture...

Would `IdentityServerProvider` just provide only the access token, or does it also provide the IS URL and other bits like that as well? If it's only the access token, you'd want to save to a very specific property, like `this.identityServerAccessTokenProvider.get()` or something.

Maybe this is a clash of functional vs. OO perspectives, but taking a single function feels a bit more flexible that mandating an interface with a single member function. With the current functional approach, if the caller wants to implement the function via a stateful object, they can (and React SDK does as it's less code change), or they can just pass a function (which a bot consumer might do).

With the provider style, if you just want to pass a function, I guess you'd make a wrapper object like:

```js
identityAccessTokenProvider: {
    get: () => {
        ...
    }
}

which is fine, but it seems like an extra step?

Comment (turt2live) [reply src/client.js]: It wouldn't be any more difficult than the bridge-sdk's theory of a "controller":
```
identityServerProvider: {
  getAccessToken: () => new IdentityAuthClient().getAccessToken();
}
```

Making it provide the URL might be a thing we do, but for now it's probably fine to just leave it  with the single responsibility. 

Alternatively, stick "Fn" at the end of the variable name so it's clear that it's meant to get a function and not overriding a function already available on the MatrixClient object (right now it looks like an override).

jryans: Well, I didn't really mean it would be difficult, but more like... aesthetically displeasing. Anyway, I suppose it's quite subjective. I've gone ahead with your proposal here.

PR #1024, opened 2019-08-22
Add AutoDiscovery.getRawClientConfig() for easy .well-known lookups

turt2live:
Useful in cases where you don't need to validate the homeserver information.


bwindels:
> Lgtm! Just wondering in which cases we wouldn't want to validate?

Answered my own question in the next react-sdk PR ;)

PR #1025, opened 2019-08-23
Don't convert errors to JSON if they are JSON already

turt2live:
For example, if the identity server throws a 401 on `/account`, we end up here with a JSON object. Don't convert the string `object Object` to JSON because it'll fail: just use the object.


turt2live:
About 4 levels higher in this code there's a path which adds a `json: false` to the request, and I don't want to break that because I don't know why it does that.

PR #1027, opened 2019-09-02
Add Synapse admin functions for deactivating a user

turt2live:
For https://github.com/matrix-org/matrix-react-sdk/pull/3371

Note: they're all intentionally named to have "Synapse" somewhere in the function name. These are implementation-specific endpoints that we don't want to be perceived as Matrix spec.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -3985,7 +3985,7 @@ MatrixClient.prototype.whoisSynapseUser = function(userId) {
  * Deactivates a user using Synapse's administrator API. <strong>This
  * function is implementation specific and may change as a result.</strong>
  * @param {string} userId the User ID to deactivate.
- * @return {object} the deactive response - see Synapse docs for information.
+ * @return {object} the deactivate response - see Synapse docs for information.
  */
 MatrixClient.prototype.deactivateSynapseUser = function(userId) {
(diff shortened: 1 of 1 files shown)


t3chguy:
IMO this would be nicer in an external Synapse class which gets passed a MatrixClient, e.g `new SynapseAdminClient(MatrixClient)`

turt2live:
Typically I would agree, however as it stands there's plenty in the js-sdk which should be structured like that and isn't. Not really looking to open the architecture can of worms on this one.

src/client.js
```diff
@@ -3948,6 +3948,55 @@ MatrixClient.prototype.isFallbackICEServerAllowed = function() {
...
+    return this._http.authedRequest(
+        undefined, 'GET', path, undefined, undefined, {prefix: ''},
+    );
+};
+
+/**
+ * Deactivates a user using Synapse's administrator API. <strong>This
+ * function is implementation specific and may change as a result.</strong>
+ * @param {string} userId the User ID to deactivate.
+ * @return {object} the deactive response - see Synapse docs for information.
```
jryans:
```suggestion
 * @return {object} the deactivate response - see Synapse docs for information.
```

PR #1039, opened 2019-09-19
handle dont_push push rule action

ara4n:
Part of https://github.com/matrix-org/matrix-react-sdk/pull/3461


jryans:
Looks like the server support has been abandoned, so I think we can close this for now.

PR #1044, opened 2019-09-25
Remove id_server from creds for interactive auth

jryans:
For HSes that do not require an IS, we can remove `id_server` from the auth
params.

Fixes https://github.com/vector-im/riot-web/issues/10959

### DIFF
--- a/src/interactive-auth.js
+++ b/src/interactive-auth.js
@@ -174,16 +174,19 @@ InteractiveAuth.prototype = {
             // The email can be validated out-of-band, but we need to provide the
             // creds so the HS can go & check it.
             if (this._emailSid) {
-                const idServerParsedUrl = url.parse(
-                    this._matrixClient.getIdentityServerUrl(),
-                );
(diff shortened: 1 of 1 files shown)


src/interactive-auth.js
```diff
@@ -174,16 +174,19 @@ InteractiveAuth.prototype = {
...
             // creds so the HS can go & check it.
             if (this._emailSid) {
-                const idServerParsedUrl = url.parse(
-                    this._matrixClient.getIdentityServerUrl(),
-                );
+                const creds = {
+                    sid: this._emailSid,
+                    client_secret: this._clientSecret,
+                };
+                if (await this._matrixClient.doesServerRequireIdServerParam()) {
```
bwindels: I imagine this would throw when there is no network... should this have a try/catch?

jryans: I think it should be okay... It's in an async function, and the call sites are checking for rejected promises.

PR #1049, opened 2019-10-03
Freeze all events in the store

dbkr:
Currently any event stored in the js-sdk can be modified by anything
that does event.getContent() and then, accidentally or otherwise,
modifies the object that is returned. This has scope for all sorts
of strange behaviour and is generally terrifying.

This PR freezes the events as they go into the model objects,
although the event class itself does quite a bit of manipulation
of this object (mostly in redactions) so there's quite a bit of
code making a copy, modifying it and the refreezing it.

This commit just modifies the events to be frozen and doesn't
attempt to fix up any of the places we've been modifying the
events, so this currently breaks everything.

TODO: the interactions between redactions and edits look somewhat
subtle: more may be needed here.

~~Edit: I should have created this as a draft but apparently you can't change a normal PR into a draft, so whatever.~~

---

### DIFF
--- a/src/models/event.js
+++ b/src/models/event.js
@@ -27,6 +27,37 @@ import utils from '../utils.js';
 import logger from '../../src/logger';
 
 /**
+ * return a copy of o that has been recursively frozen with
+ * Object.freeze().
+ * The object will be squashed down into a plain object with
+ * JSON.stringify/parse before freezing.
+ * Will throw if o contains cycles
+ *
+ * @param {object} o The object to be copied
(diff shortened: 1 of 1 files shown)


t3chguy:
This has caused bugs around Spaces/DMs where things were mutating the content of the `m.direct` accountData event so comparing prevEvent to mxEvent was always showing 0 changes.

t3chguy:
> Edit: I should have created this as a draft but apparently you can't change a normal PR into a draft, so whatever.

Converted now

PR #1050, opened 2019-10-08
verification in DMs

uhoreg:
Fixes vector-im/riot-web#10793

implements [MSC2241](https://github.com/matrix-org/matrix-doc/pull/2241)

also cleans up some bits of code, and removes some obsolete (will no longer be used) code

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -756,6 +756,128 @@ Crypto.prototype.setDeviceVerification = async function(
 };
 
 
+function verificationEventHandler(target, userId, roomId, eventId) {
+    return function(event) {
+        // listen for events related to this verification
+        if (event.getRoomId() !== roomId
+            || event.getSender() !== userId) {
+            return;
+        }
+        const content = event.getContent();
+        if (!content["m.relates_to"]) {
+            return;
+        }
+        const relatesTo
+              = content["m.relationship"] || content["m.relates_to"];
+        if (!relatesTo.rel_type
+            || relatesTo.rel_type !== "m.reference"
+            || !relatesTo.event_id
+            || relatesTo.event_id !== eventId) {
+            return;
+        }
+
+        // the event seems to be related to this verification, so pass it on to
+        // the verification handler
+        target.handleEvent(event);
+    };
+}
+
+Crypto.prototype.requestVerificationDM = function(userId, roomId, methods) {
+    let methodMap;
+    if (methods) {
+        methodMap = new Map();
+        for (const method of methods) {
+            if (typeof method === "string") {
+                methodMap.set(method, defaultVerificationMethods[method]);
+            } else if (method.NAME) {
+                methodMap.set(method.NAME, method);
+            }
+        }
+    } else {
+        methodMap = this._baseApis._crypto._verificationMethods;
+    }
+
+    return new Promise(async (_resolve, _reject) => {
+        const listener = (event) => {
+            // listen for events related to this verification
+            if (event.getRoomId() !== roomId
+                || event.getSender() !== userId) {
+                return;
+            }
+            const content = event.getContent();
+            const relatesTo
+                  = content["m.relationship"] || content["m.relates_to"];
+            if (!relatesTo || !relatesTo.rel_type
+                || relatesTo.rel_type !== "m.reference"
+                || !relatesTo.event_id
+                || relatesTo.event_id !== eventId) {
+                return;
+            }
+
+            // the event seems to be related to this verification
+            switch (event.getType()) {
+            case "m.key.verification.start": {
+                const verifier = new (methodMap.get(content.method))(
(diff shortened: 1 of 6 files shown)


src/crypto/index.js
```diff
@@ -756,6 +756,128 @@ Crypto.prototype.setDeviceVerification = async function(
...
+
+    return new Promise(async (_resolve, _reject) => {
+        const listener = (event) => {
+            // listen for events related to this verification
+            if (event.getRoomId() !== roomId
+                || event.getSender() !== userId) {
+                return;
+            }
+            const content = event.getContent();
+            const relatesTo
```
bwindels: Could use `MatrixEvent.getRelation` here. We stuck with `m.relates_to` for now, but if we ever change it to `m.relationship`, best to have to change it only once.

src/crypto/index.js
```diff
@@ -756,6 +756,128 @@ Crypto.prototype.setDeviceVerification = async function(
...
+                  = content["m.relationship"] || content["m.relates_to"];
+            if (!relatesTo || !relatesTo.rel_type
+                || relatesTo.rel_type !== "m.reference"
+                || !relatesTo.event_id
+                || relatesTo.event_id !== eventId) {
```
bwindels: I think this would throw if an event comes in before the verification event gets sent (and eventId is defined).
I tried this:
```
function scope() {
	const f = () => console.log("foo", foo);
	f();
	const foo = 5;
	f();
}
scope();
```
and it throws with `ReferenceError: foo is not defined` at the first call of `f()`.

You could put a `let eventId` at the top and reassign it at the bottom?

src/crypto/index.js
```diff
@@ -756,6 +756,128 @@ Crypto.prototype.setDeviceVerification = async function(
...
+                methodMap.set(method, defaultVerificationMethods[method]);
+            } else if (method.NAME) {
+                methodMap.set(method.NAME, method);
+            }
+        }
+    } else {
+        methodMap = this._baseApis._crypto._verificationMethods;
+    }
+
+    return new Promise(async (_resolve, _reject) => {
```
bwindels: If `sendEvent` inside this function (or anything after it) would throw, it would get swallowed and leave this promise hanging, as the return value from the promise executor is ignored (here the promise of the async function).

You could instead put the `sendEvent` after creating this promise (and putting it in a variable) and make the whole method async rather than the promise executor, like so:

```
const listenPromise = new Promise((_resolve, _reject) => {
...
});
const res = await this._baseApis.sendEvent(...);
return listenPromise;
```

src/crypto/index.js
```diff
@@ -756,6 +756,128 @@ Crypto.prototype.setDeviceVerification = async function(
...
+    }
+    const content = event.getContent();
+    const verifier = new Method(
+        this._baseApis, event.getSender(), content.from_device, event.getId(),
+        event.getRoomId(),
+    );
+    verifier.handler = verificationEventHandler(
+        verifier, event.getSender(), event.getRoomId(), event.getId(),
+    );
+    this._baseApis.on("event", verifier.handler);
```
bwindels: how will this handler get removed?

src/crypto/verification/Base.js
```diff
@@ -103,6 +117,21 @@ export default class VerificationBase extends EventEmitter {
...
 
+    /* send a message to the other participant, using in-roomm messages
+     */
+    _sendMessage(type, content) {
+        if (this._done) {
+            return Promise.reject(new Error("Verification is already done"));
+        }
+        // FIXME: only use one of m.relationship/m.relates_to, once MSC1849
+        // decides which one to use
+        content["m.relationship"] = content["m.relates_to"] = {
```
bwindels: I guess you're doing both here because of [this discussion](https://github.com/matrix-org/matrix-doc/pull/1849/files/fa7c338a42389bb378cd2e52c1e62654e7fb0c8d#r317411401)? I'd just stick with `"m.relates_to"` for now, as we're not doing this anywhere else, but feel free to leave it like this.

src/crypto/index.js
uhoreg: I've added a comment that explains where it gets removed

uhoreg:
Comments have been addressed.

BK tests passed: https://buildkite.com/matrix-dot-org/matrix-js-sdk/builds/608

src/crypto/index.js
```diff
@@ -797,23 +797,23 @@ Crypto.prototype.requestVerificationDM = function(userId, roomId, methods) {
         methodMap = this._baseApis._crypto._verificationMethods;
     }
 
-    return new Promise(async (_resolve, _reject) => {
+    let eventId = undefined;
+    const listenPromise = new Promise(async (_resolve, _reject) => {
```
bwindels: nit: this doesn't need to be async anymore i think?

PR #1052, opened 2019-10-09
Don't include .js in the import string.

clokep:
This removes a `.js` from one of the imports. This was causing me issues while trying to update the SDK used in Thunderbird and it seems unnecessary. Running the tests locally pass and I'm pretty sure there's no need for it since none of the other imports use it.


clokep:
Thank you for the fast merge! 👍

Issue #1004, opened 2019-07-30
[Doc] EventTimelineSet~getRelationsForEvent may return undefined


V02460:
This method might return undefined, but JSDoc says otherwise.

https://github.com/matrix-org/matrix-js-sdk/blob/9a8af05bfb6d01ddd835ee6f5104b3b7368d1273/src/models/event-timeline-set.js#L692-L695

t3chguy:
Fixed by #1054

PR #1054, opened 2019-10-15
Properly documented the function possible returns

rcsm:
Added the `nullable` possibility of return to the documentation of the function and described when that could happen, solving issue #1004. The tests made were visual by generating the docs everytime a change was made.

Signed-off-by: rcsm cascalhoeng@gmail.com

### DIFF
--- a/src/models/event-timeline-set.js
+++ b/src/models/event-timeline-set.js
@@ -689,8 +689,9 @@ EventTimelineSet.prototype.compareEventOrdering = function(eventId1, eventId2) {
  * @param {String} eventType
  * The relation event's type, such as "m.reaction", etc.
  *
- * @returns {Relations}
- * A container for relation events.
+ * @returns {?Relations}
+ * A container for relation events, or undefined if either an event argument is invalid or
+ * aggregation is disabled.
  */
 EventTimelineSet.prototype.getRelationsForEvent = function(
     eventId, relationType, eventType,


src/models/event-timeline-set.js
```diff
@@ -696,8 +696,9 @@ EventTimelineSet.prototype.compareEventOrdering = function(eventId1, eventId2) {
  * @param {String} eventType
  * The relation event's type, such as "m.reaction", etc.
  *
- * @returns {Relations}
- * A container for relation events.
+ * @returns {?Relations}
+ * A container for relation events, or undefined if either an event argument is invalid or
```
t3chguy: the jsdoc bit looks right but the reasoning here seems wrong, in those two cases this method actually throws

rcsm: Yeah, that's true, just realized it. I'll correct the return comment asap

src/models/event-timeline-set.js
```diff
@@ -695,9 +695,11 @@ EventTimelineSet.prototype.compareEventOrdering = function(eventId1, eventId2) {
  * The type of relation involved, such as "m.annotation", "m.reference", "m.replace", etc.
  * @param {String} eventType
  * The relation event's type, such as "m.reaction", etc.
+ * @throws If <code>eventId, relationType or eventType</code> are not valid.
```
t3chguy:
```suggestion
 * @throws If <code>eventId</code>, <code>relationType</code> or <code>eventType</code> are not valid.
```

PR #1061, opened 2019-10-31
Fix incorrect return value of MatrixClient.prototype.uploadKeys

beaclnd92:
It resolves nothing when I call the MatrixClient.prototype.uploadKeys function.


t3chguy:
As per contributing doc: https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.rst#sign-off

PR #1062, opened 2019-11-01
Relax identity server discovery checks to FAIL_PROMPT

jryans:
As discussed in MSC2284, this relaxes the identity server discovery to a
`FAIL_PROMPT` state so that clients can choose to warn and continue.

Part of https://github.com/vector-im/riot-web/issues/11102
Implements https://github.com/matrix-org/matrix-doc/pull/2284

### DIFF
--- a/spec/unit/autodiscovery.spec.js
+++ b/spec/unit/autodiscovery.spec.js
@@ -416,8 +416,8 @@ describe("AutoDiscovery", function() {
         ]);
     });
 
-    it("should return FAIL_ERROR when the identity server configuration is wrong " +
-        "(missing base_url)", function() {
+    it("should return SUCCESS / FAIL_PROMPT when the identity server configuration " +
+        "is wrong (missing base_url)", function() {
         httpBackend.when("GET", "/_matrix/client/versions").check((req) => {
             expect(req.opts.uri)
                 .toEqual("https://chat.example.org/_matrix/client/versions");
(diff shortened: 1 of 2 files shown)


jryans:
Still need to fix the failing tests... As part of that, I am trying find out why the BK log doesn't show the test summary.

jryans:
Okay, tests should be passing now. I'll continue to explore the test log problem separately.

turt2live:
We need to get the MSC across the line before we can merge this. Our new spec process doesn't allow us to move quickly for .well-known

jryans:
Fair enough, I'll mark the issue blocked then.

jryans:
(It would still be good to review the code, even if we can't merge it yet.)

jryans:
@turt2live What phase of the spec process do we need to wait for in this case?

turt2live:
Release with a formal version number.

turt2live:
unless we fix the process, of course.

jryans:
From https://github.com/vector-im/riot-web/issues/11102, it was decided it was safe to merge this work once we cleared FCP.

jryans:
This PR was accidentally forgotten... It doesn't appear to be required for React SDK / Riot to process a dead IS as optional, but that's more of a curiosity of how it interprets discovery results. This work is the key of implementing the MSC text as written and work be important for any other uses of JS SDK.

jryans:
It turns out this side was more important than I remembered... 😞 In particular, it's needed at least for the startup autodiscovery, so that we don't show a hard error as in https://github.com/vector-im/riot-web/issues/11508.

PR #1063, opened 2019-11-01
Remove Jenkins scripts

jryans:
We haven't used Jenkins for a while, so it seems safe to remove the scripts that
supported it.


jryans:
> we aren't using this for Buildkite anywhere?

I can't find any references to it at least across all 3 repos... BK has its own pipeline steps that reference other unrelated things, but not this stuff.

PR #1064, opened 2019-11-04
Don't double cancel verification request

bwindels:
This fixes the verifier not handling incoming `m.key.verification.cancel` events, and just treating them as unexpected and sending a `m.key.verification.cancel` event of its own with an error. These double `m.key.verification.cancel` events make it harder to detect which user actually pressed cancel.

Initial PR for verification over DM was https://github.com/matrix-org/matrix-js-sdk/pull/1050

### DIFF
--- a/src/crypto/verification/Base.js
+++ b/src/crypto/verification/Base.js
@@ -152,11 +152,11 @@ export default class VerificationBase extends EventEmitter {
             this._resetTimer();
             this._resolveEvent(e);
         } else {
-            this._expectedEvent = undefined;
             const exception = new Error(
                 "Unexpected message: expecting " + this._expectedEvent
                     + " but got " + e.getType(),
             );
+            this._expectedEvent = undefined;
             if (this._rejectEvent) {
                 const reject = this._rejectEvent;
                 this._rejectEvent = undefined;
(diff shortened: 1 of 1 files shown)


src/crypto/verification/Base.js
```diff
@@ -151,12 +151,16 @@ export default class VerificationBase extends EventEmitter {
             this._rejectEvent = undefined;
             this._resetTimer();
             this._resolveEvent(e);
+        } else if (e.getType() === "m.key.verification.cancel") {
+            const reject = this._reject;
```
uhoreg: I think that if you just do a `this.cancel(e)`, then it should do the right thing, because the `cancel` function says that it handles cancellations from the other user (if it gets passed a `MatrixEvent`).  Though we probably also want to call `this._rejectEvent` if it's defined, so it may be better to just stick this in with the `else` clause, and assign to `exception` based on whether `e.getType() === "m.key.verification.cancel"`.

Though it looks like the `else` clause isn't doing the right thing right now, because it really should be causing the client to send an `m.key.verification.cancel` with `code` set to `m.unexpected_message`.  So, I think if you do something like

```javascript
const exception = e.getType() === "m.key.verification.cancel" ? e : newUnexpectedMessageError();
```

in the `else` clause (you'll need to import `newUnexpectedMessageError` from `Error.js`), then it should fix that and the double-cancel issue.

bwindels: Hmm, I tried calling `_rejectEvent`, but that made it send an `m.key.verification.cancel` event like before. Will have another look

PR #1074, opened 2019-11-18
remove bluebird specific things

t3chguy:


t3chguy:
Bluebird is dead. Long live Promises

PR #1077, opened 2019-11-18
Fix SAS verification in encrypted DMs

bwindels:
Fixes https://github.com/vector-im/riot-web/issues/11383

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -789,13 +789,9 @@ function verificationEventHandler(target, userId, roomId, eventId) {
             || event.getSender() !== userId) {
             return;
         }
-        const content = event.getContent();
-        if (!content["m.relates_to"]) {
-            return;
-        }
-        const relatesTo
-              = content["m.relationship"] || content["m.relates_to"];
-        if (!relatesTo.rel_type
+        const relatesTo = event.getRelation();
+        if (!relatesTo
+            || !relatesTo.rel_type
             || relatesTo.rel_type !== "m.reference"
             || !relatesTo.event_id
             || relatesTo.event_id !== eventId) {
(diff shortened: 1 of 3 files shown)


bwindels:
I have broken verification over device_to apparently though, will have a look tomorrow.

src/crypto/index.js
```diff
@@ -65,6 +65,32 @@ export function isCryptoAvailable() {
...
+function listenForEvents(client, roomId, listener) {
+    let isEncrypted = false;
+    if (roomId) {
+        isEncrypted = client.isRoomEncrypted(roomId);
+    }
+
+    if (isEncrypted) {
+        client.on("Event.decrypted", listener);
+    } else {
+        client.on("event", listener);
```
uhoreg: I think we need to listen on `event` even in encrypted rooms, because a client might send an unencrypted event to it.

bwindels:
> I have broken verification over device_to apparently though, will have a look tomorrow.

Actually, this is working fine. I was trying to verify a different device than for which I had riot open :man_facepalming:

src/crypto/index.js
bwindels: Right, yes. The reason I didn't is that `handleEvent` in `VerifierBase` throws and cancels when it encounters any event type that it isn't what is is expecting, like `m.room.encrypted`. I could filter out `m.room.encrypted`, but I can imagine that, especially when the verification is in the right panel (as opposed to a dialog), it will be common to have non-verification events be sent in the room while the verification is ongoing (e.g. a message, or to setup a VoIP call).

So unless there is a reason not to, instead I'd propose to just ignore events in `VerifierBase` that are non related to verification, and then indeed also listen to `event` for encrypted rooms.

bwindels: VerifierBase only receives events with the correct relation to the start event, so messages and other events unrelated to verification shouldn't cancel verification. Made the changes to ignore `m.room.encrypted` and route unencrypted events through as well.

PR #1079, opened 2019-11-20
Add secret storage bootstrap flow

jryans:
This adds a secret storage bootstrap flow, intended to be a safe way to create both SSSS and cross-signing keys.

Part of https://github.com/vector-im/riot-web/issues/11212


src/crypto/index.js
```diff
@@ -276,6 +281,60 @@ Crypto.prototype.init = async function() {
...
+            { doInteractiveAuthFlow },
+        );
+        crossSigningKeysChanged = true;
+    }
+
+    // Check if Secure Secret Storage has a default key. If we don't have one, create the
+    // default key (which will also be signed by the cross-signing master key).
+    if (!this._secretStorage.hasKey()) {
+        logger.log("Secret storage default key not found, creating new key");
+        const newKeyId = await this.addSecretKey(
```
dbkr: Riot will want to pass a passphrase into this so we'll need to expose that somehow, and can't pass it in each time we start up so I fear we may need to split this function up into "SSSS is set up" and "no? ok, go set it up with this passphrase"

jryans: Ah right. I guess the main options would be:

* Split it up into a checking only path vs. actual setup as you describe
* Add a UI callback for requesting a passphrase

src/crypto/index.js
```diff
@@ -276,6 +281,60 @@ Crypto.prototype.init = async function() {
...
+} = {}) {
+    logger.log("Bootstrapping Secure Secret Storage");
+
+    // Create cross-signing keys if they don't exist, as we want to sign the SSSS default
+    // key with the cross-signing master key. The cross-signing master keys is also used
+    // to verify the signature on the SSSS default key when adding secrets, so we
+    // effectively need it for both reading and writing secrets.
+    let crossSigningKeysChanged = false;
+    if (!this._crossSigningInfo.getId()) {
+        logger.log("Cross-signing keys not found, creating new keys");
```
dbkr: We should also check if there are cross-signing keys for this account on the server but we haven't confirmed them and stored them locally, in which case we need to get the private keys.

src/crypto/index.js
```diff
@@ -328,39 +387,49 @@ Crypto.prototype.checkPrivateKey = function(privateKey, expectedPublicKey) {
...
+        await this._signObject(this._crossSigningInfo.keys.master);
+
+        // send keys to server first before caching locally (to ensure upload succeeds)
+        const keys = {};
+        for (const [name, key] of Object.entries(this._crossSigningInfo.keys)) {
+            keys[name + "_key"] = key;
+        }
+        await doInteractiveAuthFlow(async authDict => {
+            await this._baseApis.uploadDeviceSigningKeys(authDict, keys);
+        });
```
dbkr: We could also take a copy of the old keys (ie. probably the empty object) and set them back on the CrossSigningInfo object if this throws, so everything's back in the state it started if the user aborts the UI auth flow.

jryans: Good idea, added.

src/crypto/index.js
jryans: I have added a path to get the cross-signing private keys from secret storage if they exist.

src/crypto/CrossSigning.js
```diff
@@ -101,12 +105,92 @@ export class CrossSigningInfo extends EventEmitter {
...
+        const privateKeys = {};
+        for (const name of ["master", "self_signing", "user_signing"]) {
+            const encodedKey = await secretStorage.get(`m.cross_signing.${name}`);
+            privateKeys[name] = decodeBase64(encodedKey);
+        }
+
+        // Regenerate public keys from private keys
+        // XXX: Do we want to _also_ download public keys from the homeserver to
+        // verify they agree...?
+        // See also https://github.com/vector-im/riot-web/issues/11558
```
dbkr: Hmm - yeah, we should not be re-signing them after they're initially generated. I think you want `checkOwnCrossSigningTrust()` for this which will get the public keys from the HS (or rather, we should have already fetched them if our device list is up to date) and then try to get the matching private key from the app by calling the `getCrossSigningKey` callback. That said, looks like you're trying to move as much as possible into the js-sdk (which is probably a good thing) but that will mean we'll need to try & handle that within the js-sdk too (ie. probably call the callback first and then if that says, "no key here", look in SSSS automatically).

src/crypto/SecretStorage.js
```diff
@@ -132,6 +124,49 @@ export default class SecretStorage extends EventEmitter {
...
+        );
+        if (!keyInfoEvent) {
+            throw new Error(`Key ${keyId} does not exist in account data`);
+        }
+        const keyInfo = keyInfoEvent.getContent();
+
+        await this._crossSigningInfo.signObject(keyInfo, 'master');
+        await this._baseApis.setAccountData(
+            `m.secret_storage.key.${keyId}`, keyInfo,
+        );
```
dbkr: Something else is generating the key and saving to account data then this is overwriting it with a signed version? Should we be uploading the key to account data before its signed?

src/crypto/index.js
```diff
@@ -363,41 +487,99 @@ Crypto.prototype.checkPrivateKey = function(privateKey, expectedPublicKey) {
...
+        // XXX: Do we also need to sign the cross-signing master key with the
+        // device key as in `resetCrossSigningKeys`?
+
+        // write a copy locally so we know these are trusted keys
+        await this._cryptoStore.doTxn(
+            'readwrite', [IndexedDBCryptoStore.STORE_ACCOUNT],
+            (txn) => {
+                this._cryptoStore.storeCrossSigningKeys(txn, this._crossSigningInfo.keys);
+            },
+        );
```
dbkr: This one looks like it's also doing things that `checkOwnCrossSigningTrust()` ought to be doing (ie. the `this._cryptoStore.storeCrossSigningKeys` part).

src/crypto/olmlib.js
```diff
@@ -399,3 +400,21 @@ module.exports.pkVerify = function(obj, pubkey, userId) {
...
     }
 };
+
+/**
+ * Encode a typed array of uint8 as base64.
+ * @param {Uint8Array} uint8Array The data to encode.
+ * @return {string} The base64.
+ */
+module.exports.encodeBase64 = function(uint8Array) {
+    return Buffer.from(uint8Array).toString("base64");
```
dbkr: nb. if @uhoreg update the spec to be unpadded base64 we'll just need to trim the `=` signs off the end. Decoding could be more annoying depending on whether `Buffer` will refuse to decode strings without the correct padding.

src/crypto/CrossSigning.js
jryans: Okay, I have reworked to store more things in JS SDK and rely on `checkOwnCrossSigningTrust`.

src/crypto/index.js
jryans: Right, I have converted to that.

jryans:
Tests still need a bit of repair, but I think another review of layering would still be useful.

src/crypto/SecretStorage.js
dbkr: ah, that makes sense then.

jryans:
Looks like I was actually just missing some deps after the recent build changes... Tests now passing locally, and CI agrees as well.

PR #1083, opened 2019-11-21
Always process call events in batches

dbkr:
We had a bunch of logic in place to suppress calls if the answer
or hangup had already arrived, but we only used it on startup.
This extends this logic to happen all the time, which means we'll
also do the same suppression if a call happenned while we were
offline.


dbkr:
(assigning straight to you, travis because it's so similar to the last one which you reviewed.)

PR #1085, opened 2019-11-22
Monitor verification request over DM as well

bwindels:
And emit `crypto.verification.request` events for both verification over DM and mock it (see commit messages) for to_device verifications.

Required for https://github.com/matrix-org/matrix-react-sdk/pull/3661

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -930,11 +930,8 @@ Crypto.prototype.registerEventHandlers = function(eventEmitter) {
         crypto._onToDeviceEvent(event);
     });
 
-    // only sync timeline events, not events when backpaginating, loading, ...
     eventEmitter.on("Room.timeline", function(event) {
-        if (event.getRoomId()) {
-            crypto._onTimelineEvent(event);
-        }
+        crypto._onTimelineEvent(event);
     });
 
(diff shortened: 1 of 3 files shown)


src/crypto/index.js
```diff
@@ -65,6 +65,9 @@ export const verificationMethods = {
     SAS: SAS.NAME,
 };
 
+const VERIFICATION_REQUEST_TIMEOUT = 5 * 60 * 1000; //5m
+const VERIFICATION_REQUEST_MARGIN = 3 * 1000; //3s
```
dbkr: Could use some docs here on what this is - I'm assuming the fudge factor I was talking about so we don't flash up requests and then hide them right after.

src/crypto/index.js
```diff
@@ -919,6 +922,17 @@ Crypto.prototype.registerEventHandlers = function(eventEmitter) {
     eventEmitter.on("toDeviceEvent", function(event) {
         crypto._onToDeviceEvent(event);
     });
+
+    // only sync timeline events, not events when backpaginating, loading, ...
```
dbkr: Isn't it `event` for just events that arrive down the sync?

PR #1086, opened 2019-11-22
Fix calls in e2e rooms

dbkr:
Events will be decrypted after the sync event, so we were having
to wait until the next sync event before they got processed.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -4644,8 +4644,11 @@ function setupCallEventHandler(client) {
     // This happens quite often, eg. replaying sync from storage, catchup sync
     // after loading and after we've been offline for a bit.
     let callEventBuffer = [];
-    function onSync(state) {
-        if (state === "SYNCING") {
+    function evaluateEventBuffer() {
+        if (client.getSyncState() === "SYNCING") {
+            // don't process any events until they are all decrypted
+            if (callEventBuffer.some((e) => e.isBeingDecrypted())) return;
+
             const ignoreCallIds = {}; // Set<String>
             // inspect the buffer and mark all calls which have been answered
             // or hung up before passing them to the call event handler.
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -4670,20 +4673,31 @@ function setupCallEventHandler(client) {
...
-            }
-            return;
+        // any call events or ones that might be once they're decrypted
+        if (event.getType().indexOf("m.call.") === 0 || event.isBeingDecrypted()) {
+            // queue up for processing once all events from this sync have been
+            // processed (see above).
+            callEventBuffer.push(event);
+        }
+
+        if (event.isBeingDecrypted() || event.isDecryptionFailure()) {
```
bwindels: why do we want to listen for decryption if that has already failed? Because it might succeed later when the keys are received?

src/client.js
```diff
@@ -4644,8 +4644,11 @@ function setupCallEventHandler(client) {
     // This happens quite often, eg. replaying sync from storage, catchup sync
     // after loading and after we've been offline for a bit.
     let callEventBuffer = [];
-    function onSync(state) {
-        if (state === "SYNCING") {
+    function evaluateEventBuffer() {
+        if (client.getSyncState() === "SYNCING") {
+            // don't process any events until they are all decrypted
+            if (callEventBuffer.some((e) => e.isBeingDecrypted())) return;
```
bwindels: if for some events we never get the keys, won't calls completely stop working?

src/client.js
dbkr: Yep, this happens fairly often if the to_device vent with the keys just lags behind the message itself

src/client.js
dbkr: `decrypted` is a bit misleading and actually means, "finished trying to decrypt" so will fire when we've either decrypted it or given up trying.

dbkr:
Yeah, I've trying to avoid relying on local clocks for calls after experiencing my mac's clock being off by about 2 minutes very recently which would have just made all calls stop working.

src/client.js
bwindels: right, but some events in the buffer might never get decrypted, even though all the `m.call` events in the buffer are already decrypted. As the buffer is never emptied in this case, I'm wondering if this would ever resolve itself. Maybe I'm missing something.

bwindels: > `decrypted` is a bit misleading and actually means, "finished trying to decrypt" so will fire when we've either decrypted it or given up trying.

Oh, it also fires when giving up, ok :+1:

PR #1089, opened 2019-11-25
Fix empty string handling in push notifications

turt2live:
Fixes https://github.com/vector-im/riot-web/issues/11460

Empty strings are falsey, and the state key match for a tombstone event is an empty string. Ergo, nothing happens because all the conditions fail. 

This PR also does a small amount of drive-by cleanup.

Evidence of this working:
image

### DIFF
--- a/src/pushprocessor.js
+++ b/src/pushprocessor.js
@@ -268,7 +268,7 @@ function PushProcessor(client) {
         }
 
         const val = valueForDottedKey(cond.key, ev);
-        if (isNullOrUndefined(val) || typeof val != 'string') {
+        if (typeof val !== 'string') {
             return false;
         }
 
--- a/src/pushprocessor.js
+++ b/src/pushprocessor.js
@@ -15,7 +15,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import {escapeRegExp, globToRegexp} from "./utils";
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -714,3 +714,7 @@ module.exports.ensureNoTrailingSlash = function(url) {
 module.exports.sleep = (ms, value) => new Promise((resolve => {
     setTimeout(resolve, ms, value);
 }));
+
+module.exports.isNullOrUndefined = function(val) {
```
turt2live: rationale: I'm super lazy and don't feel like typing this OR condition all over the place.

src/pushprocessor.js
```diff
@@ -268,7 +268,7 @@ function PushProcessor(client) {
         }
 
         const val = valueForDottedKey(cond.key, ev);
-        if (!val || typeof val != 'string') {
+        if (isNullOrUndefined(val) || typeof val != 'string') {
```
jryans: Is the left side needed here? `typeof null === "object"` and `typeof undefined === "undefined"`...

Fine to keep if it's clearer as-is though. I guess a comment on what cases we mean to include vs. exclude would help.

turt2live: Probably not, but I'm terrified of changes in this area of code (given age).

turt2live:
I'm merging regardless of tests because the tests are flakey :|

PR #1093, opened 2019-11-26
Fix Olm unwedging test

turt2live:
Deep within the crypto layers we call `getId()`, and when we don't have that function the async call on the emitter fails but doesn't fail the test. This manifests as a timeout because the code path that would call the thing blew up.

Part of https://github.com/vector-im/riot-web/issues/11520


turt2live:
(other tests are still broken though)

PR #1094, opened 2019-11-27
Fix a couple SAS tests

turt2live:
See commits.

Part of https://github.com/vector-im/riot-web/issues/11520


turt2live:
(there's still one broken test that needs fixing)

PR #1102, opened 2019-12-03
Update alias handling

turt2live:
Fixes https://github.com/vector-im/riot-web/issues/11551

### DIFF
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -839,9 +839,15 @@ Room.prototype.getAliases = function() {
         for (let i = 0; i < aliasEvents.length; ++i) {
             const aliasEvent = aliasEvents[i];
             if (utils.isArray(aliasEvent.getContent().aliases)) {
-                Array.prototype.push.apply(
-                    aliasStrings, aliasEvent.getContent().aliases,
-                );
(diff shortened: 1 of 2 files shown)


src/models/room.js
```diff
@@ -839,9 +839,15 @@ Room.prototype.getAliases = function() {
...
-                );
+                const filteredAliases = aliasEvent.getContent().aliases.filter(a => {
+                    if (typeof(a) !== "string") return false;
+                    if (a[0] !== '#') return false;
+                    if (!a.endsWith(`:${aliasEvent.getStateKey()}`)) return false;
+
+                    // It's probably valid by here.
+                    return true;
+                });
+                Array.prototype.push.apply(aliasStrings, filteredAliases);
```
jryans: I am puzzled why we cannot just call `aliasStrings.push(...)`... It's definitely an array, we made it above... Anyway, not related to the change you are making, just puzzling.

turt2live: yea, this is weird. It might be an artifact of 2015/2016

turt2live:
@jryans turns out our tests are just as helpful. Please take a quick look.

PR #1104, opened 2019-12-05
ignore m.key.verification.done messages when we don't expect any more messages

uhoreg:
This should fix some test flakiness.


bwindels:
@uhoreg should this be merged?

uhoreg:
Yes, thanks for reminding me.

PR #1109, opened 2019-12-09
Refactor verification request code

bwindels:
Part of https://github.com/vector-im/riot-web/issues/11224

This extracts most of the verification request code from `crypto/index` and puts it into it's own `VerificationRequest` class with the specificities of whether to use to_device verification or verification over DM into a `Channel` class. Care has been taken to maintain the same public API from the crypto module and the request object passed as an argument of the crypto.verification.request event.

The next step will be add support for the `m.key.verification.ready` event, and to decorate verification events with a `verificationRequest` property and use this to drive toasts and verification timeline tiles in the react-sdk, removing the `KeyVerificationStateObserver`.

### DIFF
--- a/spec/unit/crypto/verification/qr_code.spec.js
+++ b/spec/unit/crypto/verification/qr_code.spec.js
@@ -39,7 +39,10 @@ describe("QR code verification", function() {
 
     describe("showing", function() {
         it("should emit an event to show a QR code", async function() {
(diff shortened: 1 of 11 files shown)


bwindels:
(cleared myself as assignee to avoid misunderstandings that this is somehow waiting for me to do anything; it's not, and fully ready for review)

PR #1106, opened 2019-12-08
Babel 7 draft

t3chguy:

PR #1012, opened 2019-08-09
[WIP] Upgrade to babel@7 and support TypeScript

turt2live:
**See https://github.com/matrix-org/matrix-react-sdk/pull/3292**

Fixes https://github.com/matrix-org/matrix-js-sdk/issues/983

PR #1112, opened 2019-12-10
Part 1 of many: Upgrade to babel@7 and TypeScript

turt2live:
This Just Works™ on its own. Part 1 for fixing our build output to be useful for debugging.

Makes https://github.com/matrix-org/matrix-js-sdk/pull/1012 redundant.
Makes https://github.com/matrix-org/matrix-js-sdk/pull/1106 redundant.
Inspiration taken from both PRs.

----

This changes our js-doc output significantly, but uses a thing that is better equipped for typescript. Obligatory disclaimer that `better-docs` is the thing I use for the js-bot-sdk.

Check out the before and after (there's plenty of room for improvement, but that's for another PR):
image
image


t3chguy:
Fails lint

PR #1114, opened 2019-12-10
[WIP] Support & fix TypeScript tests

turt2live:
**Built on https://github.com/matrix-org/matrix-js-sdk/pull/1113**
**Note: This is set to merge against `travis/sourcemaps` because it breaks all the downstream builds.**

Supporting TypeScript tests is an important milestone in transitioning our riot-web build to a single webpack because it demonstrates that the SDK can be imported from `src` and from arbitrary locations. It also makes it so we're testing our actual code and not something babel spat out, therefore making our line numbers match up in the tests.


turt2live:
Deprioritizing this in favour of fixing our sourcemaps (this part isn't urgent, but is super nice to have)

PR #1118, opened 2019-12-12
Add support for key backups using secret storage

jryans:
This add various helpers for creating and restoring key backups where the private key is stored in secret storage.

Part of https://github.com/vector-im/riot-web/issues/11209
Used by https://github.com/matrix-org/matrix-react-sdk/pull/3720

### DIFF
--- a/spec/unit/crypto/backup.spec.js
+++ b/spec/unit/crypto/backup.spec.js
@@ -336,7 +336,7 @@ describe("MegolmBackup", function() {
             });
             await client.resetCrossSigningKeys();
             let numCalls = 0;
-            await new Promise(async (resolve, reject) => {
+            await new Promise((resolve, reject) => {
                 client._http.authedRequest = function(
                     callback, method, path, queryParams, data, opts,
                 ) {
@@ -361,7 +361,7 @@ describe("MegolmBackup", function() {
                     resolve();
                     return Promise.resolve({});
                 };
-                await client.createKeyBackupVersion({
+                client.createKeyBackupVersion({
                     algorithm: "m.megolm_backup.v1",
                     auth_data: {
                         public_key: "hSDwCYkwp1R0i33ctD73Wg2/Og0mOBr066SpjqqbTmo",
--- a/src/client.js
+++ b/src/client.js
@@ -1519,7 +1519,7 @@ MatrixClient.prototype.createKeyBackupVersion = async function(info) {
     // sessions.
     await this.checkKeyBackup();
     if (!this.getKeyBackupEnabled()) {
-        throw new Error("Key backup not usable even though we just created it");
+        logger.error("Key backup not usable even though we just created it");
     }
 
     return res;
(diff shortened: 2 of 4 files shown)


src/client.js
```diff
@@ -1450,32 +1452,36 @@ MatrixClient.prototype.prepareKeyBackupVersion = async function(
...
-        },
+        auth_data: authData,
         recovery_key: encodedPrivateKey,
     };
 };
 
+/**
+ * Check whether the key backup private key is stored in secret storage.
+ */
+MatrixClient.prototype.isKeyBackupKeyStored = async function() {
```
turt2live: it shouldn't need to be async if we don't await

src/client.js
```diff
@@ -1450,32 +1452,36 @@ MatrixClient.prototype.prepareKeyBackupVersion = async function(
...
-            private_key_salt: keyInfo.passphrase.salt,
-            private_key_iterations: keyInfo.passphrase.iterations,
-        },
+        auth_data: authData,
         recovery_key: encodedPrivateKey,
     };
 };
 
+/**
+ * Check whether the key backup private key is stored in secret storage.
```
turt2live: For the jsdoc it would be good to have the return type specified explicitly.

spec/unit/crypto/backup.spec.js
```diff
@@ -361,7 +361,7 @@ describe("MegolmBackup", function() {
                     resolve();
                     return Promise.resolve({});
                 };
-                await client.createKeyBackupVersion({
+                client.createKeyBackupVersion({
```
turt2live: I'm impressed the tests still pass even though the function is in fact async

src/client.js
```diff
@@ -1624,6 +1634,15 @@ MatrixClient.prototype.restoreKeyBackupWithPassword = async function(
     );
 };
 
+MatrixClient.prototype.restoreKeyBackupWithSecretStorage = async function(
```
turt2live: Docs please

src/client.js
jryans: The inner function is async and / or returns a promise, but `return await` is considered poor form by some as it adds some unnecessary overhead... I thought it best to at least leave the async to suggest the return type of a promise, but also haven't seen a clear style choice for this in code base.

turt2live: I roughly read that as "AHHHHHHHHHHHHHHHHHHHHHHHHH" in my head and am fine with the change as-is.

spec/unit/crypto/backup.spec.js
jryans: This bit triggers the handler installed above which is wrapped in a promise... It's a bit of a monster. 😅

src/client.js
jryans: Okay, added!

src/client.js
jryans: Okay, added!

PR #1121, opened 2019-12-16
[WIP] Convert imports and exports to ES6

turt2live:
**This is against `travis/sourcemaps` because it breaks everything.**

Changes:
* Imports and exports are all ES6-style - see below for more info)
* Minification on the browser bundle (it was trivial to add - just add uglifyify)
* Moved the browser bundle into `lib/browser` to collapse the tree a bit
* Migrated the browser and regular indexes into `src/` for more logical transpiling and IDE suggestions (many IDEs don't look at ignored directories, but will if you compile the entry point)
* Tests are now pointed at `src` instead of `lib` - this is for ES6 imports/exports and future typechecking.

----

The old `module.exports` syntax doesn't work downstream in the react-sdk or riot-web layers anymore, so this brings it up to 2019 standards. 

We do not use default exports anymore because they are discouraged by the ES6 community. It's not an official recommendation, but it's certainly commonplace. 

This also changes browser-index.js and index.js to use the `src` tree instead of `lib`. This is primarily to have IDEs autocomplete the package more easily, and because we're compiling this all in the riot-web layer it doesn't really matter. We still export `lib` and `dist/browser-matrix.js` for backwards compatibility (ie: everyone who has imported from `matrix-js-sdk/lib/whatever` or uses `<script>` tags). In a future version/PR we could drop these if we feel like it.

This commit does not address the test failures that are almost certainly caused by this change. While we're here, the ancient `"use strict";` flags have been removed as well.

Some files were not as easily converted. Most were simple keyword changes though there are a couple instances where the file structure is redone to accommodate ES6 exports.


.babelrc
```diff
@@ -1,11 +1,8 @@
 {
-    "sourceMaps": "inline",
+    "sourceMaps": true,
```
turt2live: with `true` we just let babel decide what is best for us. When we need inline sourcemaps we ask for them.

.babelrc
```diff
@@ -1,11 +1,8 @@
...
     "presets": [
         ["@babel/preset-env", {
             "targets": {
-                "browsers": [
-                    "last 2 versions"
-                ],
```
turt2live: browserify takes care of the browser for us.

package.json
```diff
@@ -4,11 +4,11 @@
   "description": "Matrix Client-Server SDK for JavaScript",
   "scripts": {
     "start": "echo THIS IS FOR LEGACY PURPOSES ONLY. && babel src -w -s -d lib --verbose --extensions \".ts,.js\"",
-    "clean": "rimraf lib dist",
+    "clean": "rimraf lib",
     "build": "yarn clean && git rev-parse HEAD > git-revision.txt && yarn build:compile && yarn build:compile-browser && yarn build:types",
     "build:types": "tsc --emitDeclarationOnly",
     "build:compile": "babel src -s -d lib --verbose --extensions \".ts,.js\"",
-    "build:compile-browser": "mkdirp dist && browserify -d browser-index.js -t [ babelify --sourceMaps=inline ] > dist/browser-matrix.js",
+    "build:compile-browser": "mkdirp lib/browser && browserify -d src/browser-index.js -p [ tsify -p ./tsconfig.json ] -t [ babelify --sourceMaps=inline ] -t uglifyify > lib/browser/matrix.js",
```
turt2live: tsify is so we can load typescript from `src`. uglifify is for minification.

spec/integ/matrix-client-crypto.spec.js
```diff
@@ -24,9 +24,6 @@ limitations under the License.
  * See also `megolm.spec.js`.
  */
 
-"use strict";
-import 'source-map-support/register';
```
turt2live: We don't need these imports anymore - they actually make the test's sourcemaps wrong.

turt2live:
Now that this passes, I'm going to chop it up into more sensible PRs.

turt2live:
Replaced by:
* https://github.com/matrix-org/matrix-js-sdk/pull/1122
* https://github.com/matrix-org/matrix-js-sdk/pull/1123
* https://github.com/matrix-org/matrix-js-sdk/pull/1124
* https://github.com/matrix-org/matrix-js-sdk/pull/1125
* https://github.com/matrix-org/matrix-js-sdk/pull/1126
* https://github.com/matrix-org/matrix-js-sdk/pull/1127

PR #1127, opened 2019-12-17
Target NodeJS 10, minified browser bundle, and other publishing/package things

turt2live:
**This is going to `travis/sourcemaps` for safety.**

**This PR is built on top of https://github.com/matrix-org/matrix-js-sdk/pull/1122 - see https://github.com/matrix-org/matrix-js-sdk/compare/refs/pull/1122/head...refs/pull/1127/head for the real diff**

We release with Node 10 currently, so we should use that. The browser bundle is minfied because we want to keep it, so we might as well shave about 1mb off of it.


.babelrc
```diff
@@ -3,10 +3,7 @@
     "presets": [
         ["@babel/preset-env", {
             "targets": {
-                "browsers": [
-                    "last 2 versions"
-                ],
-                "node": 12
```
dbkr: No browser support?

package.json
```diff
@@ -7,8 +7,8 @@
     "clean": "rimraf lib dist",
     "build": "yarn clean && git rev-parse HEAD > git-revision.txt && yarn build:compile && yarn build:compile-browser && yarn build:types",
     "build:types": "tsc --emitDeclarationOnly",
-    "build:compile": "babel src -s -d lib --verbose --extensions \".ts,.js\"",
-    "build:compile-browser": "mkdirp dist && browserify -d browser-index.js -t [ babelify --sourceMaps=inline ] > dist/browser-matrix.js",
+    "build:compile": "babel -d lib --verbose --extensions \".ts,.js\" src",
+    "build:compile-browser": "mkdirp dist && browserify -d src/browser-index.js -p [ tsify -p ./tsconfig.json ] -t [ babelify --sourceMaps=inline ] -t uglifyify > dist/browser-matrix.js",
```
dbkr: `dist/browser-matrix.js` is minified now and we've ditched the non minified version?

.babelrc
turt2live: The commit details this a bit: https://github.com/matrix-org/matrix-js-sdk/pull/1127/commits/f1475cd3d79e0a2c536ad7014ebb3e15d759daf3

The tldr version is this babel config isn't used for the browser bundle - browserify is.

package.json
turt2live: *flips table*

I hate everything about browser bundles. Will fix - we should be exporting both, I just forgot :(

.babelrc
turt2live: turns out this is a partial lie, but https://github.com/matrix-org/matrix-js-sdk/pull/1127/commits/b376a7c39918c226f0b0a4a54ff3ed2e7f36c176 fixes it (see commit message)

package.json
```diff
@@ -5,10 +5,11 @@
   "scripts": {
     "start": "echo THIS IS FOR LEGACY PURPOSES ONLY. && babel src -w -s -d lib --verbose --extensions \".ts,.js\"",
     "clean": "rimraf lib dist",
-    "build": "yarn clean && git rev-parse HEAD > git-revision.txt && yarn build:compile && yarn build:compile-browser && yarn build:types",
+    "build": "yarn clean && git rev-parse HEAD > git-revision.txt && yarn build:compile && yarn build:compile-browser && yarn build:minify-browser && yarn build:types",
     "build:types": "tsc --emitDeclarationOnly",
-    "build:compile": "babel src -s -d lib --verbose --extensions \".ts,.js\"",
-    "build:compile-browser": "mkdirp dist && browserify -d browser-index.js -t [ babelify --sourceMaps=inline ] > dist/browser-matrix.js",
+    "build:compile": "babel -d lib --verbose --extensions \".ts,.js\" src",
+    "build:compile-browser": "mkdirp dist && browserify -d src/browser-index.js -p [ tsify -p ./tsconfig.json ] -t [ babelify --sourceMaps=inline --presets [ @babel/preset-env @babel/preset-typescript ] ] > dist/browser-matrix.js",
```
dbkr: Hmm - sorry, but this is still spitting out the non-minified bundle with an inline source map which is now what we did before. I guess if you're worried about the file size then arguably you ought to use the minified bundle anyway, but if it works with them separate that seems nicer since you can just load the source maps if you actually want them. We previously did this with exorcist which runs after everything and takes the inline source map and splits it out into its own file.

turt2live: I would definitely argue that people should be using the minified version if they care:
image

I know it's not the same as before, but honestly this feels like a better way to do things.

.babelrc
dbkr: I guess I'm unclear on what we are saying we now export. As of b376a7c I think it's:
 * ES2015 in the browser bundle (as we're using preset-env with an empty target list)
 * Whatever node 10 supports for the npm exports, which presumably means if you're targeting node 10+ you're fine, but if you're targeting a browser you must now transpile?

What's the intention here and can we be very explicit about it?

turt2live: yea, I'll put it in the readme. The short version is:

* Browser bundle supports whatever the modern browsers do (ES2015 I think)
* ES6 for everything else
* Consumers are meant to figure out what they need/want out of this. In the case of riot-web, we're taking ES6 and bundling it ourselves.

package.json
dbkr: That misses the point though since over half of that 5.1MB is the source map. The comparison is:
```
-rw-r--r--  1 dave  staff   1.8M 19 Dec 20:32 browser-matrix-no-source-map.js
-rw-r--r--  1 dave  staff   4.9M 19 Dec 20:06 browser-matrix.js
```

...but yes, I think the only reason for not using the minified version is if you want to debug and your debugger doesn't do source maps. Not sure whether this is a case worth worrying about or not.

Why does it feel like a better way to do things? I would argue having two js files both with their source maps separate is more consistent and gives the flexibility to have source maps  if you want them.

.babelrc
turt2live: Should be covered by https://github.com/matrix-org/matrix-js-sdk/pull/1127/commits/89173be05504614c843654852a50e064d4c71a5d

package.json
turt2live: It's a very weak argument, but having the sourcemap inline on an already large bundle gets you debugging for free - the bundle is not a direct representation of the source, so a sourcemap is still required for it to make any sense. Also something about encouraging people to do the "right" thing and not use a 2-5mb JS file as their production build.

turt2live: It's quite tempting to rip out all of this and replace it with a webpack which gives us the 4 files for essentially the same amount of dependencies :(

browserify/babelify only supports one target, so sourcemaps are inline regardless. `exorcist` does work well enough though is starting to age a bit. It's still recommended by browserify though, so I guess we'll use it.

turt2live: should be done in https://github.com/matrix-org/matrix-js-sdk/pull/1127/commits/975368de8f6f75f31529cab62903e8b3a79640e1

result is:
image

.babelrc
dbkr: OK - I remember from looking at this last time that babel has no way to explicitly ask it to output a version of ES, so presumably we are relying on 'node 10' being at least approximately equal to ES6?

turt2live: yea, and in practice it's basically just adding some filler to the top and bottom of each file, so it's basically source -> output with no major differences.

package.json
dbkr: cool, thanks

PR #1130, opened 2019-12-20
[WIP] Last remaining functional changes for riot-web's new webpack

turt2live:
Partnered with https://github.com/vector-im/riot-web/pull/11679

**This doesn't work yet.**

PR #1134, opened 2019-12-23
Use a safer import/export scheme for the ContentRepo utilities

turt2live:
**This is against `travis/sourcemaps` for safety.**

Split from https://github.com/matrix-org/matrix-js-sdk/pull/1130

~~Exporting it the way we were was causing problems for webpack way down the line, so we export it differently here to get around that. We also have to fix all the import references so we import the right thing.~~

Per the comments later on in this PR we now use more standard exports for safety reasons.

----

This PR and others in the series have their overview covered here: https://gist.github.com/turt2live/a3fc7c9712b8ef0f1f758611aa33382d

### DIFF
--- a/spec/unit/content-repo.spec.js
+++ b/spec/unit/content-repo.spec.js
@@ -1,4 +1,4 @@
-import {ContentRepo} from "../../src/content-repo";
+import {getHttpUriForMxc, getIdenticonUri} from "../../src/content-repo";
 
 describe("ContentRepo", function() {
     const baseUrl = "https://my.home.server";
@@ -7,7 +7,7 @@ describe("ContentRepo", function() {
         it("should do nothing to HTTP URLs when allowing direct links", function() {
             const httpUrl = "http://example.com/image.jpeg";
             expect(
-                ContentRepo.getHttpUriForMxc(
+                getHttpUriForMxc(
                     baseUrl, httpUrl, undefined, undefined, undefined, true,
                 ),
             ).toEqual(httpUrl);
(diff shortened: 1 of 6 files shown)


turt2live:
it is a red flag, though realistically this should be a utilities class anyways. It's more ES6-export-stuff-for-the-index oddities we haven't had to deal with until now.

dbkr:
Not sure I see why it should be a utility class: I would argue that utility classes shouldn't be a thing in a language where you can instead just have a collection of functions in a module.

Either way, if it's necessary because of the build system:
 * Are features like this that we can no longer use documented somewhere so people don't try to use them and get confusing errors?
 * Is there an anticipated time when we would be able to use the features again?

turt2live:
we should probably document them, yes. I don't completely understand why it's suddenly a problem for us, so would want to look into that first. We should be able to export functions, though it's arguable when we do that versus spinning up a class - assuming the build system working, anyways.

dbkr:
Actually having thought about this more, presumably the problem is that when converting to ES6 exports, the choice is to either `export default Class Foo` / `import Foo from 'bar'` or `export function baz()` / `import * as Foo from 'bar'`?

So maybe my questions are:
 * How did you decide which combination of those to use, ie. it's the former for ContentRepo but the latter for `index.js`?
 * Which bit of the new build process is it that requires ES6ifying all the exports? (Agree we're probably overdue for doing it but might have been another candidate for doing separately).

turt2live:
There's more options in there and not quite true for assumptions:

* `export default Class Foo` / `import Foo from "bar"`
* `export default Class Foo` / `import * as Foo from "bar"`
* `export Class Foo` / `import * as Foo from "bar"`
* `export Class Foo` / `import {Foo} from "bar"`

The problem comes when we `export * from "bar"` in the index: that only works for named exports (non-default ones). The theory says we should be able to use `export function whatever(){}` with an `import * as ContentRepo from "bar"`, and that is indeed fine, but the `export * from "bar"` in the index prevents us from importing the `ContentRepo` from the matrix-js-sdk in other layers. The imports we expect to use are `import {ContentRepo} from "matrix-js-sdk"` so we end up needing a `ContentRepo` class with the static functions.

Alternatively, we import the functions directly and stop assuming it is a class in other layers. For this case it is probably fine because the functions are well-described enough, though if we have thoughts on code using `import {getHttpUriForMxc} from "matrix-js-sdk"` and then calling it elsewhere, now would be the time to raise those thoughts.

turt2live:
this is also the investigative effort I wanted to do - it sounds like it's a code style concern and less a technical limitation.

Oh, and we don't use default exports because `export * from "bar"` doesn't pick them up (or in the case of webpack, it does but through a disgusting means which isn't equivalent)

dbkr:
Ah - thanks for the detailed explanation, that does indeed make it a lot clearer. For default exports, I assume it doesn't work with an explicit import and then export? ie. `import Foo from 'foo'; export Foo`?

turt2live:
yea, importing then exporting doesn't work because reasons beyond my comprehension.

turt2live:
@dbkr I've updated this to go back to plain `export function`, altering the imports to be more stable in the SDK. The diff should be relatively sane.

PR #1135, opened 2019-12-23
notify devices when we don't send them keys

uhoreg:
Implements part of [MSC2399](https://github.com/matrix-org/matrix-doc/pull/2399) (notifications on initial key sending, but not on keyshare requests, and does not properly handle `m.no_olm` yet -- the remaining functionality will come in a future PR).

### DIFF
--- a/src/crypto/OlmDevice.js
+++ b/src/crypto/OlmDevice.js
@@ -919,6 +919,15 @@ OlmDevice.prototype.addInboundGroupSession = async function(
     );
 };
 
+/**
+ * Record in the data store why an inbound group session was withheld.
+ *
+ * @param {string} roomId     room that the session belongs to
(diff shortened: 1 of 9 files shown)


src/crypto/OlmDevice.js
```diff
@@ -914,6 +920,60 @@ OlmDevice.prototype.addInboundGroupSession = async function(
...
+                    reason: reason,
+                },
+                txn,
+            );
+        },
+    );
+};
+
+export const WITHHELD_MESSAGES = {
+    "m.unverified": "The sender has disabled encrypting to unverified devices.",
```
turt2live: shouldn't these also be `org.matrix` for now? (annoyingly)

uhoreg: I thought about that, but since they only appear in `org.matrix.room_key.withheld` messages, I think it should be fine.

PR #1139, opened 2020-01-03
Fix creating a key backup with cross signing diabled

dbkr:
It broke if no secret key callback was supplied but a cross-signing
identity did exist (as hopefully explained in comment).

Fixes https://github.com/vector-im/riot-web/issues/11763


dbkr:
Sorry, it wasn't quite that simple

PR #1146, opened 2020-01-10
record, report, and notify about olm errors

uhoreg:
Implements the `no_olm` part of [MSC2399](https://github.com/matrix-org/matrix-doc/pull/2399), and tells the user when the olm session got wedged.

The general algorithm is that when there's a problem with the olm session (either it's wedged, or the sender tells us they can't establish a session), then we try to recover by creating a session ourselves.  Then we record the problem, timestamp, and whether we (think we) fixed the problem.  When we have a megolm event that we can't decrypt, we check if it was sent within the time that we had a problem, and if so, tell the user what the problem was, and whether we were able to fix it.

### DIFF
--- a/src/crypto/store/indexeddb-crypto-store-backend.js
+++ b/src/crypto/store/indexeddb-crypto-store-backend.js
@@ -467,6 +467,7 @@ export class Backend {
         return result;
     }
 
+    // FIXME: we should probably prune this when devices get deleted
     async filterOutNotifiedErrorDevices(devices) {
         const txn = this._db.transaction("notified_error_devices", "readwrite");
         const objectStore = txn.objectStore("notified_error_devices");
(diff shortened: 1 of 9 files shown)


src/crypto/store/indexeddb-crypto-store-backend.js
```diff
@@ -426,6 +426,74 @@ export class Backend {
...
+        const req = index.getAll(deviceKey);
+        req.onsuccess = (event) => {
+            const problems = req.result;
+            if (!problems.length) {
+                result = null;
+                return;
+            }
+            problems.sort((a, b) => {
+                return a.time - b.time;
+            });
```
dbkr: fwiw you can have multi-level indices on indexeddb now*, so you would be able to index by timestamp, although not 100% sure if you would be able to get the largest ts lower than the given ts. Hopefully this table won't be big enough that it's a problem in terms of efficiency although if we ever want the index in future then could it be useful to add it now to avoid db changes?

*they might still not be supported on Edge but we already have other tables that use them

uhoreg: I did take a look at it, but didn't see an easy way to do the query that I needed.

PR #1148, opened 2020-01-13
Send device messages for the same user in same API call.

erikjohnston:
Currently we split the device messages up to limit the number per call,
but that can end up splitting messages to a given users device over
separate API calls. This is fine, but means that the server can't e.g.
bundle them into a single EDU for remote users or sanity check that the
client is sending to the right set of devices (i.e. its device list
cache isn't wrong).

---

P.S. I have no idea what I'm doing.

### DIFF
--- a/src/crypto/algorithms/megolm.js
+++ b/src/crypto/algorithms/megolm.js
@@ -330,6 +330,9 @@ MegolmEncryption.prototype._prepareNewSession = async function() {
 };
 
 /**
+ * Splits the user device map into multiple chunks to reduce the number of
+ * devices we encrypt to per API call.
+ *
  * @private
  *
  * @param {module:crypto/algorithms/megolm.OutboundSessionInfo} session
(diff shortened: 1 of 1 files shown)


src/crypto/algorithms/megolm.js
```diff
@@ -402,6 +400,17 @@ MegolmEncryption.prototype._splitUserDeviceMap = function(
 
             entriesInCurrentSlice++;
         }
+
+        // We do this in the per-user loop as we prefer that all messages to the
+        // same user end up in the same API call to make it easier for the
+        // server (e.g. only have to send one EDU if a remote user, etc). This
+        // does mean that if a user has many devices we may go over the desired
+        // limit, but its not a hard limit so that is fine.
```
uhoreg: Going over the limit is probably also fine now that the payload size is much smaller since we don't include everyone's ciphertext any more.

erikjohnston:
I'm quite tempted to refactor those two functions so that we have one `_splitUserDeviceMap` that does the splitting (i.e. what `_splitBlockedDevices` does), and another that does the filtering. Though I don't know if this is a hot path and we have duplicated for good reasons, so leaving it for now

PR #1150, opened 2020-01-13
Add an option to ignore unverified devices

dbkr:
If this requires any more information than is on the docstring then I need to write better docs :p

Default behaviour remains unchanged.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -947,6 +947,35 @@ MatrixClient.prototype.getGlobalBlacklistUnverifiedDevices = function() {
 };
 
 /**
+ * Set whether sendMessage in a room with unknown and unverified devices
+ * should throw an error and not send them message. This has 'Global' for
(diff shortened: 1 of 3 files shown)


src/client.js
```diff
@@ -946,6 +946,35 @@ MatrixClient.prototype.getGlobalBlacklistUnverifiedDevices = function() {
     return this._crypto.getGlobalBlacklistUnverifiedDevices();
 };
 
+/**
+ * Set whether sendMessage in a room with unknown and unverified devices
+ * should throw an error and not send them message. This has 'Global' for
+ * symmertry with setGlobalBlacklistUnverifiedDevices but there is currently
```
jryans:
```suggestion
 * symmetry with setGlobalBlacklistUnverifiedDevices but there is currently
```

PR #1151, opened 2020-01-13
Fix sourcemaps by refactoring the build system

turt2live:
**Review with https://github.com/vector-im/riot-web/pull/11843**

This PR doesn't have any additional alterations and just needs a blessing to join develop.


release.sh
```diff
@@ -201,9 +201,7 @@ if [ $dodist -eq 0 ]; then
     # global cache here to ensure we get the right thing.
     yarn cache clean
     yarn install
-    # We haven't tagged yet, so tell the dist script what version
-    # it's building
-    DIST_VERSION="$tag" yarn dist
+    yarn build
```
dbkr: hmm, won't this break the riot-web release since riot-web uses the same release script and uses the dist step to make the built tarball?

turt2live: What? Really? Why is riot using the js-sdk release script to release itself?

This was removed because it's a no-op in the js-sdk.

dbkr: yep, this is a shared release script between the three projects

package.json
```diff
@@ -2,20 +2,21 @@
...
     "clean": "rimraf lib dist",
-    "build": "babel src -s -d lib --verbose --extensions \".ts,.js\" && rimraf dist && mkdir dist && browserify -d browser-index.js | exorcist dist/browser-matrix.js.map > dist/browser-matrix.js && terser -c -m -o dist/browser-matrix.min.js --source-map \"content='dist/browser-matrix.js.map'\" dist/browser-matrix.js",
-    "dist": "yarn build",
-    "watch": "watchify -d browser-index.js -o 'exorcist dist/browser-matrix.js.map > dist/browser-matrix.js' -v",
-    "lint": "eslint --max-warnings 93 src spec",
-    "prepare": "yarn clean && yarn build && git rev-parse HEAD > git-revision.txt"
+    "build": "yarn clean && git rev-parse HEAD > git-revision.txt && yarn build:compile && yarn build:compile-browser && yarn build:minify-browser && yarn build:types",
+    "build:types": "tsc --emitDeclarationOnly",
+    "build:compile": "babel -d lib --verbose --extensions \".ts,.js\" src",
+    "build:compile-browser": "mkdirp dist && browserify -d src/browser-index.js -p [ tsify -p ./tsconfig.json ] -t [ babelify --sourceMaps=inline --presets [ @babel/preset-env @babel/preset-typescript ] ] | exorcist dist/browser-matrix.js.map > dist/browser-matrix.js",
```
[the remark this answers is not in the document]
dbkr: Hmm, ok - maybe the answer is we try & kill off browserify at some point then.

release.sh
turt2live: That's a bit awkward. I guess the fastest way forward is to fix it and add a warning to the top rather than refactor the thing out to the layers.

PR #1153, opened 2020-01-15
Add API to get account data from server

jryans:
This adds an API account data getter that bypasses the local store and goes
directly to the homeserver.

Part of https://github.com/vector-im/riot-web/issues/11214
Used in https://github.com/matrix-org/matrix-react-sdk/pull/3847

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -1943,6 +1943,25 @@ MatrixClient.prototype.getAccountData = function(eventType) {
 };
 
 /**
+ * Get account data event of given type for the current user. This variant
+ * bypasses the local store and gets account data directly from the homeserver,
+ * which can be useful very early in startup before the initial sync.
+ * @param {string} eventType The event type
+ * @return {module:client.Promise} Resolves: The contents of the given account
+ * data event.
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -1942,6 +1942,25 @@ MatrixClient.prototype.getAccountData = function(eventType) {
...
+/**
+ * Get account data event of given type for the current user. This variant
+ * bypasses the local store and gets account data directly from the homeserver,
+ * which can be useful very early in startup before the initial sync.
+ * @param {string} eventType The event type
+ * @return {module:client.Promise} Resolves: The contents of the given account
+ * data event.
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixClient.prototype.getAccountDataFromServer = function(eventType) {
```
turt2live: why not a `cache=false` on `getAccountData` ooi?

jryans: Mostly because the original is sync whereas this one is async.

turt2live: ah, yes. Good call.

PR #1162, opened 2020-01-22
Remove :deviceId from /keys/upload/:deviceId as not spec-compliant

t3chguy:
Fixes https://github.com/vector-im/riot-web/issues/11945


t3chguy:
Its fine, the review requirement is mostly a formality

PR #1163, opened 2020-01-22
For dm-verification, also consider events sent by other devices of same user as "our" events

bwindels:
Fixes https://github.com/vector-im/riot-web/issues/11964

When another client send the .ready event, the observing client ended up putting it in the "their" event map, and therefore not entering observeOnly mode.

Also added unit tests.

### DIFF
--- a/spec/unit/crypto/verification/verification_request.spec.js
+++ b/spec/unit/crypto/verification/verification_request.spec.js
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Matrix.org Foundation C.I.C.
+Copyright 2020 The Matrix.org Foundation C.I.C.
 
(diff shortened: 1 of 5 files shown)


bwindels:
I have move the unit test commit to be first (instead of last) with an interactive rebase and force-pushed after that which seems to have left github CI integration a bit confused with the green check left at the "remove obsolete comment" commit.

spec/unit/crypto/verification/verification_request.spec.js
```diff
@@ -0,0 +1,177 @@
+/*
+Copyright 2019 The Matrix.org Foundation C.I.C.
```
jryans:
```suggestion
Copyright 2020 The Matrix.org Foundation C.I.C.
```

PR #1167, opened 2020-01-23
add export/import of Olm devices

cedricvanrompay:
Adds an `export` method to Olm device that exports the device “account” and “sessions”.

Adds a parameter to the asynchronous initialization of the Olm device to import data that was exported with the said export method.

Adds methods to Matrix Client to expose these methods at a high level, including the device ID and user ID in the exported data (and using them during import).

Adds Jest tests for export/import at the Olm device level, and a HTML + JS document to try the mechanism in the browser at a high level in a non-automated way.

### DIFF
--- a/examples/olm-device-export-import.html
+++ b/examples/olm-device-export-import.html
@@ -1,0 +1,185 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Test Crypto in Browser</title>
+    <script src="lib/olm.js"></script>
+    <script src="lib/matrix.js"></script>
+</head>
+<body>
+    <h1>Testing export/import of Olm devices in the browser</h1>
+    <p>
+        Inspired by <code>examples/browser</code>
+    </p>
+    <ul>
+        <li>
+            Make sure you built the current version of the Matrix JS SDK
+            (<code>yarn build</code>)
+        </li>
+        <li>
+            make sure there is
+            <code>matrix.js</code>, <code>olm.js</code> and <code>olm.wasm</code>
+            in directory <code>lib/</code>
+        </li>
+        <li>start a local Matrix homeserver (on port 8008, or change the port in the code)</li>
+        <li>Serve this HTML file (e.g. <code>python3 -m http.server</code>) and go to it through your browser</li>
+        <li>
+            in the JS console, do:
+            <pre>
+aliceMatrixClient = await newMatrixClient("alice-"+randomHex());
+await aliceMatrixClient.exportDevice();
+await aliceMatrixClient.getAccessToken();
+            </pre>
+        </li>
+        <li>
+            copy the result of <code>exportDevice</code> and <code>getAccessToken</code> somewhere
+            (<strong>not</strong> in a JS variable as it will be destroyed when you refresh the page)
+        </li>
+        <li><strong>refresh the page (F5)</strong> to make sure the client is destroyed</li>
+        <li>
+            Do the following, replacing <code>ALICE_ID</code>
+            with the user ID of Alice (you can find it in the exported data)
+            <pre>
+bobMatrixClient = await newMatrixClient("bob-"+randomHex());
+roomId = await bobMatrixClient.createEncryptedRoom([ALICE_ID]);
+await bobMatrixClient.sendTextMessage('Hi Alice!', roomId);
+            </pre>
+        </li>
+        <li>Again, <strong>refresh the page (F5)</strong>. You may want to clear your console as well.</li>
+        <li>
+            Now do the following, using the exported data and the access token you saved previously:
+            <pre>
+aliceMatrixClient = await importMatrixClient(EXPORTED_DATA, ACCESS_TOKEN);
+            </pre>
+        </li>
+        <li>You should see the message sent by Bob printed in the console.</li>
+    </ul>
+
+    <script>
+if (!Olm) {
+    console.error(
+        "global.Olm does not seem to be present."
+        + " Did you forget to add olm in the lib/ directory?"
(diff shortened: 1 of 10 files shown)


cedricvanrompay:
woops, tests fail. I forgot to re-run them after rebasing.

cedricvanrompay:
I'll fix the lint errors as well

src/client.js
```diff
@@ -98,6 +98,12 @@ function keyFromRecoverySession(session, decryptionKey) {
  *
  * @param {string} opts.userId The user ID for this user.
  *
+ * @param {Object} opts.deviceToImport Device data exported with
+ *     (TODO link to export method) that must be imported to recreate this device.
+ *     Should only be useful for deviced with end-to-end crypto enabled.
```
uhoreg:
```suggestion
 *     Should only be useful for devices with end-to-end crypto enabled.
```

src/client.js
```diff
@@ -98,6 +98,12 @@ function keyFromRecoverySession(session, decryptionKey) {
  *
  * @param {string} opts.userId The user ID for this user.
  *
+ * @param {Object} opts.deviceToImport Device data exported with
+ *     (TODO link to export method) that must be imported to recreate this device.
```
uhoreg: (The TODO should be resolved before this gets merged)

src/crypto/OlmDevice.js
```diff
@@ -132,18 +143,51 @@ OlmDevice.prototype.init = async function() {
...
-    await cryptoStore.doTxn('readwrite', [IndexedDBCryptoStore.STORE_ACCOUNT], (txn) => {
-        cryptoStore.getAccount(txn, (pickledAccount) => {
-            if (pickledAccount !== null) {
-                account.unpickle(pickleKey, pickledAccount);
-            } else {
-                account.create();
-                pickledAccount = account.pickle(pickleKey);
-                cryptoStore.storeAccount(txn, pickledAccount);
-            }
+async function _initialiseAccount(fromExportedDevice, cryptoStore, pickleKey, account) {
```
uhoreg: Given that the function is a big `if` statement conditioned on `fromExportedDevice`, it might make more sense to just leave `_initialiseAccount` as-is, and create a new function (maybe called `_initialiseAccountFormExportedDevice`) that loads the export?

src/crypto/index.js
```diff
@@ -228,12 +228,24 @@ utils.inherits(Crypto, EventEmitter);
  * Initialise the crypto module so that it is ready for use
  *
  * Returns a promise which resolves once the crypto module is ready for use.
+ *
+ * @param {Object} kwargs keyword arguments.
+ * @param {string} kwargs.exportedOlmDevice (Optional) data from exported device
+ *     that must be re-created.
  */
-Crypto.prototype.init = async function() {
+Crypto.prototype.init = async function(kwargs) {
```
uhoreg: Arguments like this are generally called `opts` in other functions.

```suggestion
Crypto.prototype.init = async function(opts) {
```

(with the change applied elsewhere in the function too, obviously)

examples/olm-device-export-import.html
```diff
@@ -0,0 +1,185 @@
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
cedricvanrompay: OK, I was hesitating between doing what you suggest and just remove the file after it was used by the PR reviewers. I'll create a new example directory then.

src/client.js
cedricvanrompay: Done.

src/client.js
cedricvanrompay: Done.

src/crypto/index.js
cedricvanrompay: Ok, done.

To me `opts` can be misleading as it sounds like “optional arguments” while some may in fact be required. Usage of `kwargs` come from Python world.

Although I agree that naming consistency is more important, usage of `kwargs` instead of `opts` should be discussed in a separate issue/PR, if at all.

src/crypto/OlmDevice.js
cedricvanrompay: Done (used name `_initialiseFromExportedDevice` because we initialise more than just the account).

examples/olm-device-export-import.html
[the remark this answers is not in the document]
cedricvanrompay: Done: `examples/crypto-browser/`

src/crypto/OlmDevice.js
```diff
@@ -115,12 +115,29 @@ export function OlmDevice(cryptoStore) {
  * found.
  *
  * Reads the device keys from the OlmAccount object.
+ *
+ * @param {object} fromExportedDevice (Optional) data from exported device
+ *     that must be re-created.
  */
-OlmDevice.prototype.init = async function() {
+OlmDevice.prototype.init = async function(fromExportedDevice) {
```
uhoreg: Since the device is exported using the pickle key, there should probably be a way to set the pickle key (when not imported from an exported key).  Maybe it makes sense to make the argument an `opts` object argument that you can use to set a pickle key or pass the exported device.  (And then pass it in from the `MatrixClient` initialization.)

cedricvanrompay: So you would suggest to start this method with a `const { pickleKey, fromExportedDevice } = opts`?

Right now `fromExportedDevice` contains the pickleKey, so I would see two different strategy:
1. the caller of the method has to split the exported data before passing them to the method (and `fromExportedDevice` does **not** contain the pickle key any more
2. the caller still passes `fromExportedDevice` unmodified and with a pickleKey, in which case this method will throw in case the pickle key is passed both through `fromExportedDevice` and as another keyword argument.

Are we on the same page and which strategy would you prefer?

uhoreg: Yes.  Strategy 2 was what I was thinking of (or perhaps ignore `opts.pickleKey` and print a warning, instead of doing a `throw`)

cedricvanrompay:
Signed off with commit b217f6aa

PR #1173, opened 2020-01-25
Fix bootstrap cleanup

dbkr:
As hopefully explained in the comment. The symptom of this was that
bootstrapping would work just fine the first time you called it
in any run of the app, but then if called a second time (eg. if you
cancelled by dismissing the password prompt) it would create keys and
upload the public parts but not store the private parts in SSSS,
leaving you with cross signing keys you don't have the private parts
of.

Also use object.assign in the save keys callback just in case we
ever reset a subset of the keys (and also because it makes it a
bit simpler to reason about what objects are where).

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -351,7 +351,7 @@ Crypto.prototype.bootstrapSecretStorage = async function({
     // key with the cross-signing master key. The cross-signing master key is also used
     // to verify the signature on the SSSS default key when adding secrets, so we
     // effectively need it for both reading and writing secrets.
-    let crossSigningPrivateKeys = {};
+    const crossSigningPrivateKeys = {};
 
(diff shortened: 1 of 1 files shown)


src/crypto/index.js
```diff
@@ -468,7 +468,15 @@ Crypto.prototype.bootstrapSecretStorage = async function({
             }
         }
     } finally {
-        this._baseApis._cryptoCallbacks = appCallbacks;
+        // Restore the original callbacks. NB. we must do this by manipulating
```
jryans: An alternate approach could be for `CrossSigning` to only reference `baseApi` and always pull the callbacks from there. (Either way is fine.)

dbkr: Yeah, I suspect this would be better, really.

PR #1175, opened 2020-01-27
Update QR code handling for new spec

turt2live:
**For https://github.com/matrix-org/matrix-react-sdk/pull/3959**

### DIFF
--- a/src/crypto/verification/QRCode.js
+++ b/src/crypto/verification/QRCode.js
@@ -59,10 +59,10 @@ export class ReciprocateQRCode extends Base {
         const targetUserId = this.startEvent.getSender();
         if (!this.userId) {
             console.log("Asking to confirm user ID");
-            await new Promise((resolve, reject) => {
+            this.userId = await new Promise((resolve, reject) => {
                 this.emit("confirm_user_id", {
                     userId: targetUserId,
-                    confirm: resolve,
+                    confirm: resolve, // takes a userId
                     cancel: () => reject(newUserMismatchError()),
                 });
             });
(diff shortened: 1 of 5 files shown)


turt2live:
nevermind, I guess this breaks everything...

src/crypto/verification/QRCode.js
```diff
@@ -21,103 +22,59 @@ limitations under the License.
...
+        if (this.startEvent.getContent()['secret'] !== this.request.encodedSharedSecret) {
+            throw newKeyMismatchError();
+        }
+
+        // If we've gotten this far, verify the user's master cross signing key
+        const xsignInfo = this._baseApis.getStoredCrossSigningInfo(this.userId);
+        if (!xsignInfo) throw new Error("Missing cross signing info");
+
+        const masterKey = xsignInfo.getId("master");
+        await this._verifyKeys(this.userId, [masterKey, masterKey]);
```
dbkr: I'm not sure I really get what's going on here - presumably trying to get it into the right format so _verifyKeys can do Object.entries() but why twice? And do we not want to also verify device keys?

turt2live: we don't know the device keys at this point anymore because of how our verification setup is done.

As far as I'm aware, the format is `[keyId, key]`, and the master key's ID is the key. This should probably be `[[masterKey, masterKey]]` though

dbkr: ok, thought it would probably be something like that

turt2live:
@dbkr can you quickly take a look at https://github.com/matrix-org/matrix-js-sdk/pull/1175/commits/5fbaa9cfa7dc206df7d92323d0024d1e71e6eff0 to make sure I didn't introduce any werid vulnerabilities?

PR #1178, opened 2020-01-27
Publish pre-releases as a separate tag on npm

jryans:
npm will install the newest version a package has published to the `latest` tag,
including pre-releases, which is not ideal since those may not be ready for
production use yet.

This uses an alternate tag (`next` is a common convention, but it can be
anything) for pre-releases so the default installs only get stable versions.

Fixes https://github.com/vector-im/riot-web/issues/12029

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -294,9 +294,14 @@ fi
 rm "${release_text}"
 rm "${latest_changes}"
 
+npm_publish_flags=''
+if [ $prerelease -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


jryans:
> Everything I've read says we should use `next`, but somewhere along the lines I ended up using `beta`... not sure what the difference is in semantic versioning world, or if it matters.

`latest` is only one with any real meaning: it's the default tag used for regular installs. All other words have no specific meaning, and consumers would have to opt-in by installing with that specific tag.

PR #1180, opened 2020-01-28
Add helper to obtain the cancellation code for a verification request

t3chguy:


src/crypto/verification/request/VerificationRequest.js
```diff
@@ -251,6 +251,14 @@ export class VerificationRequest extends EventEmitter {
         return undefined;
     }
 
+    /**
+     * the cancellation code e.g m.user which is responsible for cancelling this verification
```
turt2live:
```suggestion
     * The cancellation code e.g m.user which is responsible for cancelling this verification
```

t3chguy: You asked for it :D

PR #1182, opened 2020-01-28
Add a bunch of null guards to feature checks

turt2live:
Dendrite apparently doesn't give us a bunch of these fields


foldleft:
Is there a bug that this should be linked to?

turt2live:
nope, just something we should be doing

PR #1189, opened 2020-01-29
Fix typo

dbkr:
This would probably just cause apps to wait until the first live
sync had finished rather than the one from the store, so slowing
them down / breaking offline support.


jryans:
😭 Thanks! 😭

PR #1190, opened 2020-01-30
Add function to estimate target device for a VerificationRequest

turt2live:
**For https://github.com/matrix-org/matrix-react-sdk/pull/4001**

### DIFF
--- a/src/crypto/verification/request/VerificationRequest.js
+++ b/src/crypto/verification/request/VerificationRequest.js
@@ -272,6 +272,26 @@ export class VerificationRequest extends EventEmitter {
         return this._sharedSecret;
     }
 
+    /**
+     * Estimates which device the verification should be started with
+     * given the events sent so far in the verification. This is the
+     * same algorithm used to determine which device to send the
+     * verification to when no specific device is specified.
+     * @returns {{userId: *, deviceId: *}} The device information
+     */
+    get estimatedTargetDevice() {
+        const theirFirstEvent =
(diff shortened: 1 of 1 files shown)


src/crypto/verification/request/VerificationRequest.js
```diff
@@ -645,16 +665,7 @@ export class VerificationRequest extends EventEmitter {
...
-                this._eventsByThem.get(REQUEST_TYPE) ||
-                this._eventsByThem.get(READY_TYPE) ||
-                this._eventsByThem.get(START_TYPE);
-            const theirFirstContent = theirFirstEvent.getContent();
-            const fromDevice = theirFirstContent.from_device;
-            targetDevice = {
-                userId: this.otherUserId,
-                deviceId: fromDevice,
-            };
+            targetDevice = this.estimatedTargetDevice;
```
foldleft: shouldn't this be a function call?

dbkr: It's a getter so it's a secret function call, although tbh given it's already confused one person I'm not sure it's the best place for one. I'd have just gone with `estimateTargetDevice()`, or in fact probably just `targetDevice()` - aren't we always verifying the device that's talking to us, by definition?

turt2live: The code didn't feel very reliable at picking it up, though it can certainly be renamed to `targetDevice`.

PR #1195, opened 2020-02-04
add logging when a request is being cancelled

bwindels:
so we can more easily see (especially for to_device requests)
why something was cancelled, to debug issues like https://github.com/vector-im/riot-web/issues/12226

### DIFF
--- a/src/crypto/verification/request/VerificationRequest.js
+++ b/src/crypto/verification/request/VerificationRequest.js
@@ -557,6 +557,15 @@ export class VerificationRequest extends EventEmitter {
         } else if (this._observeOnly !== wasObserveOnly) {
             this.emit("change");
         }
+
+        // log cancellations so we can see from rageshakes why riot sometimes cancels
+        // requests on its own
+        if (type === CANCEL_TYPE) {
+            logger.log(`verification request ${this.channel.transactionId}: ` +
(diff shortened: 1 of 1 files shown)


src/crypto/verification/request/VerificationRequest.js
```diff
@@ -557,6 +557,15 @@ export class VerificationRequest extends EventEmitter {
         } else if (this._observeOnly !== wasObserveOnly) {
             this.emit("change");
         }
+
+        // log cancellations so we can see from rageshakes why riot sometimes cancels
+        // requests on its own
+        if (type === CANCEL_TYPE) {
+            logger.log(`verification request ${this.channel.transactionId}: ` +
+                `.cancel event with ${JSON.stringify(event.getContent())} ` +
```
jryans: Maybe pass this as a separate not-stringifed arg so it's easier to inspect in the console?

bwindels: logger.log doesn't support multiple arguments AFAIK...

jryans: Fine to merge either way. `console.log` definitely does, so we might want to change the logger in some other PR. Rageshakes [already stringify](https://github.com/matrix-org/matrix-react-sdk/blob/develop/src/rageshake/rageshake.js#L74-L97) where needed.

PR #1199, opened 2020-02-04
Convert utils.js -> utils.ts

foldleft:
Note: if you set `noImplicitAny` to `true` this still compiles :)


src/utils.ts
```diff
@@ -20,15 +20,15 @@ limitations under the License.
...
+import * as unhomoglyph from 'unhomoglyph';
 
 /**
  * Encode a dictionary of query parameters.
  * @param {Object} params A dict of key/values to encode e.g.
  * {"foo": "bar", "baz": "taz"}
  * @return {string} The encoded string e.g. foo=bar&baz=taz
  */
-export function encodeParams(params) {
+export function encodeParams(params: {[key: string]: string}): string {
```
t3chguy: could we use the more readable utility type here? `Record<string, string>`

src/utils.ts
```diff
@@ -48,7 +48,8 @@ export function encodeParams(params) {
  * variables with. E.g. { "$bar": "baz" }.
  * @return {string} The result of replacing all template variables e.g. '/foo/baz'.
  */
-export function encodeUri(pathTemplate, variables) {
+export function encodeUri(pathTemplate: string,
+                          variables: {[key: string]: string}): string {
```
t3chguy: same as above

src/utils.ts
```diff
@@ -114,7 +116,7 @@ export function keys(obj) {
  * @param {Object} obj The object to get the values for.
  * @return {Array<*>} The values of the object.
  */
-export function values(obj) {
+export function values<T>(obj: {[keys: string]: T}): T[] {
```
t3chguy: same as above

src/utils.ts
t3chguy: https://www.typescriptlang.org/docs/handbook/utility-types.html

foldleft:
ooh, I did not know about those! Thanks :)

turt2live:
(you also have build failures to look at)

src/@types/unhomoglyph/index.d.ts
```diff
@@ -0,0 +1,21 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+declare function unhomoglyph (s: string): string;
+
+declare module 'unhomoglyph' {
```
turt2live: We shouldn't bother with third party declarations in our projects.

src/utils.ts
```diff
@@ -617,7 +628,7 @@ export function inherits(ctor, superCtor) {
  * @param {any} SuperType The type to act as a super instance
  * @param {any} params Arguments to supply to the super type's constructor
  */
-export function polyfillSuper(thisArg, SuperType, ...params) {
+export function polyfillSuper(thisArg: any, SuperType: any, ...params: any) {
```
turt2live:
```suggestion
export function polyfillSuper(thisArg: any, SuperType: any, ...params: any[]) {
```

tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "module": "commonjs",
     "moduleResolution": "node",
     "target": "es2016",
+    "esModuleInterop": true,
```
turt2live: from the docs, we shouldn't need this?

src/utils.ts
turt2live: this is because the spread operator is an array btw

PR #1202, opened 2020-02-07
New option to fully reset Secret Storage keys in boostrapSecretStorage

foldleft:
relates to https://github.com/vector-im/riot-web/issues/12221


src/crypto/index.js
```diff
@@ -411,12 +413,13 @@ Crypto.prototype.bootstrapSecretStorage = async function({
                 );
             }
         } else {
+            if (setupNewSecretStorage) throw new Error("What the HELL");
```
turt2live: probably want to remove this debug line

foldleft: WHOOPS

foldleft: I really need some kind of pre-push hook I can use in conjunction with some magic string to yell at me for this

PR #1206, opened 2020-02-11
Remove methods argument to verification

bwindels:
Fixes https://github.com/vector-im/riot-web/issues/12319
It's easy for all the calls in react-sdk to `requestVerification(DM)` to have a different set of methods they pass along. There is also little value to it, as clients built on top of js-sdk can already select the subset of methods they can provide an UI for through the `verificationMethods` option when creating the client.

This is sort of a breaking change, as the extra argument consumers of the js-sdk might still pass will be ignored.


jryans:
> This is sort of a breaking change, as the extra argument consumers of the js-sdk might still pass will be ignored.

Right, it seems that way. I added a breaking change label to this PR.

PR #1211, opened 2020-02-14
log MAC calculation during SAS

bwindels:
To help debug https://github.com/vector-im/riot-web/issues/12226

### DIFF
--- a/src/crypto/verification/SAS.js
+++ b/src/crypto/verification/SAS.js
@@ -28,6 +28,7 @@ import {
     newUnknownMethodError,
     newUserCancelledError,
 } from './Error';
+import {logger} from '../../logger';
 
 const START_TYPE = "m.key.verification.start";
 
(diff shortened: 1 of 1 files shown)


src/crypto/verification/SAS.js
```diff
@@ -165,6 +166,15 @@ const macMethods = {
     "hmac-sha256": "calculate_mac_long_kdf",
 };
 
+function calculateMAC(olmSAS, method) {
+    return function(...args) {
+        const macFunction = olmSAS[macMethods[method]];
+        const mac = macFunction.apply(olmSAS, args);
+        logger.log("SAS calculateMAC:", method, args, mac);
```
jryans: Hmm, I think you told me the other day this doesn't take multiple args... 😜

bwindels: I tried now and it indeed works :smile:

PR #1213, opened 2020-02-14
method for checking if other party supports verification method

bwindels:
Part of https://github.com/vector-im/riot-web/issues/12368

### DIFF
--- a/src/crypto/verification/request/VerificationRequest.js
+++ b/src/crypto/verification/request/VerificationRequest.js
@@ -185,6 +185,33 @@ export class VerificationRequest extends EventEmitter {
             && this._phase !== PHASE_CANCELLED;
     }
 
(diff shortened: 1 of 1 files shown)


src/crypto/verification/request/VerificationRequest.js
```diff
@@ -185,6 +185,33 @@ export class VerificationRequest extends EventEmitter {
             && this._phase !== PHASE_CANCELLED;
     }
 
+    /** Checks whether the other party supports a given verification method.
+     *  This is useful setting up the QR code UI, as it is somewhat asymmetrical:
```
jryans:
```suggestion
     *  This is useful when setting up the QR code UI, as it is somewhat asymmetrical:
```

PR #1214, opened 2020-02-14
make logging compatible with rageshakes

bwindels:


src/logger.js
```diff
@@ -29,6 +29,25 @@ import log from "loglevel";
 // Part of #332 is introducing a logging library in the first place.
 const DEFAULT_NAMESPACE = "matrix";
 
+// because rageshakes in react-sdk hijack the console log, also at module load time,
+// initializing the logger here races with the initialization of rageshakes.
+// to avoid the issue, we override the methodFactory of loglevel that binds to the
+// console methods at initialization time by a factory that looks op the console methods
```
dbkr: up?

src/logger.js
```diff
@@ -29,6 +29,25 @@ import log from "loglevel";
...
+// initializing the logger here races with the initialization of rageshakes.
+// to avoid the issue, we override the methodFactory of loglevel that binds to the
+// console methods at initialization time by a factory that looks up the console methods
+// when logging so we always get the current value of console methods.
+log.methodFactory = function(methodName, logLevel, loggerName) {
+    return function(...args) {
+        const supportedByConsole = methodName === "error" ||
+            methodName === "warn" ||
+            methodName === "trace" ||
+            methodName === "info";
```
jryans: `debug` is also supported I think.

bwindels: Ah, when I tried `console.debug(...)` didn't seemed to output anything. But indeed seems to work :man_shrugging:

PR #1215, opened 2020-02-17
allow updating device keys

bwindels:
Fixes https://github.com/vector-im/riot-web/issues/12226

Device keys can get rotated because idb got cleared out on low disk. So far, other riots wouldn't update the rotated keys in this case to prevent MITM. This prevents verification as the other clients are stuck with an old device key and the MAC fails.

This PR changes the above to accept the new device key. This should be fine because the device becomes unverified and depending on the paranoia settings we will or will not encrypt for that device after accepting the new key.

There does seem to be a bug verifying the rotated device afterwards though, see https://github.com/vector-im/riot-web/issues/12399


bwindels:
Ok, will think of ways to avoid losing the keys in the first place then.

PR #1219, opened 2020-02-18
add method for new /aliases endpoint

bwindels:


src/base-apis.js
```diff
@@ -1119,6 +1119,19 @@ MatrixBaseApis.prototype.deleteAlias = function(alias, callback) {
     );
 };
 
+/**
+ * @param {string} roomId
+ * @param {module:client.callback} callback Optional.
+ * @return {Promise} Resolves: an object with an `aliases` property, containing an array of local aliases
+ * @return {module:http-api.MatrixError} Rejects: with an error response.
+ */
+MatrixBaseApis.prototype.getLocalAliases =
```
turt2live: This should be flagged as unstable, and use the unstable endpoint?

richvdh: cf https://github.com/matrix-org/synapse/pull/6948

bwindels:
I tested this works with rich's synapse branch btw.

PR #1221, opened 2020-02-20
Reduce secret size for new binary packing

turt2live:
See https://github.com/matrix-org/matrix-react-sdk/pull/4091


turt2live:
The docs are part of the MSC, and the secret size changed in recent commits on said MSC: https://github.com/matrix-org/matrix-doc/pull/1544/files#diff-be05e010f156555a15712a6db76abd9bR203-R206

PR #1222, opened 2020-02-20
Fix: error during tests

bwindels:
Add null check for when there is no response. Seems like the tests were intermittently failing [here](https://buildkite.com/matrix-dot-org/matrix-react-sdk/builds/4676#c8bf091a-096c-4f11-901a-83072cb80c47/302-2039) (assuming because there was on network?).


bwindels:
oops, forgot to merge this

PR #1228, opened 2020-02-24
Use symmetric encryption for SSSS

uhoreg:
fixes https://github.com/vector-im/riot-web/issues/12431

Note: I have not yet tested to see what happens when Riot gets run with this.  It may explode.  It probably should not be merged until the Riot bits are adjusted to work properly with this, and we have a migration from old SSSS.

### DIFF
--- a/src/crypto/SecretStorage.js
+++ b/src/crypto/SecretStorage.js
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Matrix.org Foundation C.I.C.
+Copyright 2019, 2020 The Matrix.org Foundation C.I.C.
 
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


src/crypto/SecretStorage.js
```diff
@@ -19,8 +19,225 @@ import {logger} from '../logger';
...
+            hash: "SHA-256",
+        },
+        hkdfkey,
+        512,
+    );
+
+    const aesKey = keybits.slice(0, 32);
+    const hmacKey = keybits.slice(32);
+
+    const aesProm = await subtleCrypto.importKey(
```
dbkr: Looks like you don't want to `await` here if you want to pass the promises to Promise.all at the end?

PR #1233, opened 2020-02-27
Perform crypto store operations directly after transaction

jryans:
At least on Safari but perhaps other browsers as well, you must perform
IndexedDB operations in the same JS task as you start the transaction. As a
concrete example, you cannot open the transaction and await some promise before
actually using it.

This fixes the crypto store to meet this requirement.

Fixes https://github.com/vector-im/riot-web/issues/12207

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -685,6 +685,9 @@ MatrixClient.prototype.initCrypto = async function() {
         throw new Error(`Cannot enable encryption: no cryptoStore provided`);
     }
 
+    logger.log("Crypto: Starting up crypto store...");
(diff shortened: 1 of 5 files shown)


jryans:
> If doTxn was able to create a transaction, _backendPromise must have been resolved already. So was it just awaiting `_backendPromise` that caused the transaction to close?

Well, `_backendPromise` is separate from the transaction, as it is always resolved after the store has started up.

Yes, just awaiting `_backendPromise` is the issue here: even if a promise is resolved, any `then` handlers are run asynchronously in a separate task, so that's enough for the transaction to close.

PR #1236, opened 2020-02-27
Use unpadded base64 for QR code secrets

turt2live:


src/crypto/olmlib.js
```diff
@@ -407,6 +407,15 @@ export function encodeBase64(uint8Array) {
     return Buffer.from(uint8Array).toString("base64");
 }
 
+/**
+ * Encode a typed array of uint8 as unpadded base64.
+ * @param {Uint8Array} uint8Array The data to encode.
+ * @return {string} The unpadded base64.
+ */
+export function encodeUnpaddedBase64(uint8Array) {
+    return encodeBase64(uint8Array).replace(/=$/g, '');
```
uhoreg:
```suggestion
    return encodeBase64(uint8Array).replace(/=+$/g, '');
```

or something, since base64 can have more than one `=` at the end

turt2live: indeed
image

PR #1238, opened 2020-02-28
Migrate SSSS to use symmetric algorithm

uhoreg:
includes the changes from https://github.com/matrix-org/matrix-js-sdk/pull/1228

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -502,7 +502,9 @@ Crypto.prototype.bootstrapSecretStorage = async function({
                 logger.log("Secret storage default key not found, using key backup key");
 
                 // FIXME: ask for recovery passphrase/key
-                const backupKey = Buffer.from("XrmITOOdBhw6yY5Bh7trb/bgp1FRdIGyCUxxMP873R0", "base64");
+                const backupKey = Buffer.from(
+                    "XrmITOOdBhw6yY5Bh7trb/bgp1FRdIGyCUxxMP873R0", "base64",
+                );
 
                 if (!newKeyId) {
                     const opts = {};
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -395,7 +395,7 @@ Crypto.prototype.bootstrapSecretStorage = async function({
     const ssssKeys = {};
 
     this._baseApis._cryptoCallbacks.getSecretStorageKey =
-        async (keys, name) => {
+        async ({keys}, name) => {
             // if we already have a key that works, return it
             for (const keyId of Object.keys(keys)) {
                 if (ssssKeys[keyId]) {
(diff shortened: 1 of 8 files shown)


uhoreg:
fixes https://github.com/vector-im/riot-web/issues/12508

This should be basically ready for review, but needs some figuring out about how it interfaces with higher levels, which probably needs some react-sdk implementation.

src/crypto/index.js
```diff
@@ -413,19 +439,42 @@ Crypto.prototype.bootstrapSecretStorage = async function({
...
+                // if we already have a backup key, use the same key as the
+                // secret storage key
                 logger.log("Secret storage default key not found, using key backup key");
-                const opts = {
-                    pubkey: keyBackupInfo.auth_data.public_key,
-                };
+                const opts = {};
+
+                // FIXME: ask for recovery passphrase/key
+                const backupKey = "foobar";
```
uhoreg: I don't think we have a callback yet for asking for the backup key, so one may need to be added.

src/crypto/index.js
```diff
@@ -388,10 +388,36 @@ Crypto.prototype.bootstrapSecretStorage = async function({
...
+                ssssKeys[keyId] = keyData;
+            }
+            return key;
+        };
+
     try {
-        const inStorage = !setupNewSecretStorage &&
-            await this._crossSigningInfo.isStoredInSecretStorage(this._secretStorage);
+        const decryptionKeys =
+              await this._crossSigningInfo.isStoredInSecretStorage(this._secretStorage);
```
foldleft: I worry that waiting on a promise might return control to the main loop, leaving the callbacks above in place

foldleft: not that I'm sure there's a better way to do this

src/crypto/index.js
```diff
@@ -413,19 +439,42 @@ Crypto.prototype.bootstrapSecretStorage = async function({
                     { authUploadDeviceSigningKeys },
                 );
             }
+        } else if (!(Object.values(decryptionKeys).some(
+            info => info.algorithm === SECRET_STORAGE_ALGORITHM_V1_AES,
+        ))) {
+            // we already have cross-signing keys, but they're encrypted using
+            // the old algorithm
+            logger.log("Switching to symmetric");
```
foldleft: this probably merits a `logger.warn`, plus more context

src/crypto/index.js
foldleft: file a task if you're leaving this unfinished

uhoreg: Yeah, this is part of the reason why this PR is marked as draft.  It'll probably need some figuring out with the react side how to do it properly.

turt2live:
Taking riot-web review off this as it's my understanding it's not completely ready yet.

PR #1240, opened 2020-02-28
start documention for verification request and friends

bwindels:
---

## ✨ Features
 * start documention for verification request and friends ([\#1240](https://github.com/matrix-org/matrix-js-sdk/pull/1240)).

### DIFF
--- a/docs/verification-request.md
+++ b/docs/verification-request.md
@@ -1,0 +1,83 @@
+# Verification requests
+
+## VerificationRequest
+
+VerificationRequest is that class responsible for transitioning between the various phases that a verification request can be in. This is both for self-verification over to-device messages, as for verification of other users over in-room messages. It also creates the verifier object (depending on the chosen method) once the .start event is received.
+
+## Channels
+
+As mentioned, VerificationRequest handles requests for both in-room as to-device events. The differences between both have been factored out into `InRoomChannel` and `ToDeviceChannel` although there are some flag properties on those with the logic checking them being in VerificationRequest or the verifier.
+
+The main differences between both kinds of requests are:
+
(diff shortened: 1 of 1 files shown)


t3chguy:
@bwindels can this land as is to be better than nothing?

docs/verification-request.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Channels
+
+As mentioned, VerificationRequest handles requests for both in-room as to-device events. The differences between both have been factored out into `InRoomChannel` and `ToDeviceChannel` although there are some flag properties on those with the logic checking them being in VerificationRequest or the verifier.
+
+The main differences between both kinds of requests are:
+
+ - in-room requests support being "historical", request that have concluded and/or were conducted more than 10 minutes ago. They should be persisted over reloads, but they should only observe the request for displaying. No events should be sent in it.
+ - in-room requests are assumed to be between two users, to-device can be between the same user on different devices.
+ - .request and .ready events are optional for `ToDeviceChannel` (proposal to remove this difference and make these events mandatory for to-device as well).
```
MadLittleMods, line 15: What is `.request` and `.ready`?

I think it's talking about `m.key.verification.ready` and `m.key.verification.request` 🤔

docs/verification-request.md
```diff
@@ -0,0 +1,83 @@
+# Verification requests
+
+## VerificationRequest
+
+VerificationRequest is that class responsible for transitioning between the various phases that a verification request can be in. This is both for self-verification over to-device messages, as for verification of other users over in-room messages. It also creates the verifier object (depending on the chosen method) once the .start event is received.
```
MadLittleMods, line 5: We should put the `VerificationRequest` class in backticks like the other classes (appears other places)

```suggestion
`VerificationRequest` is that class responsible for transitioning between the various phases that a verification request can be in. This is both for self-verification over to-device messages, as for verification of other users over in-room messages. It also creates the verifier object (depending on the chosen method) once the .start event is received.
```

PR #1242, opened 2020-03-02
Fixed up tests to match new way that crypto stores are created

foldleft:


src/crypto/store/indexeddb-crypto-store.js
```diff
@@ -314,8 +314,7 @@ export class IndexedDBCryptoStore {
      * @param {string} type A key type
      */
     async getCrossSigningPrivateKey(txn, func, type) {
-        const backend = await this._connect();
-        return backend.getCrossSigningPrivateKey(txn, func, type);
+        this._backend.getCrossSigningPrivateKey(txn, func, type);
```
jryans: Needs `return` I think?

foldleft: oop

src/crypto/store/indexeddb-crypto-store.js
```diff
@@ -314,7 +314,7 @@ export class IndexedDBCryptoStore {
      * @param {string} type A key type
      */
     async getCrossSigningPrivateKey(txn, func, type) {
-        this._backend.getCrossSigningPrivateKey(txn, func, type);
+        return this._backend.getCrossSigningPrivateKey(txn, func, type);
```
jryans: Oh, I think I misunderstood how these work...? Looks like the actual data is passed to `func`, so perhaps the `return` wasn't needed before or now...?

jryans: Also I guess this could drop `async` now...?

foldleft: mm, I suppose it could

PR #1245, opened 2020-03-04
For self-verifications, also request keys from the other device

foldleft:
Fixes: https://github.com/vector-im/riot-web/issues/12302

To view the cache:
```
let { getCrossSigningKeyCache } = mxMatrixClientPeg.get()._crypto._crossSigningInfo._cacheCallbacks;
getCrossSigningKeyCache && await Promise.all([getCrossSigningKeyCache("self_signing"), getCrossSigningKeyCache("user_signing")])
```
Screenshot 2020-03-04 at 14 58 01
Screenshot 2020-03-04 at 14 58 07

### DIFF
--- a/src/crypto/verification/Base.js
+++ b/src/crypto/verification/Base.js
@@ -24,6 +24,8 @@ import {EventEmitter} from 'events';
 import {logger} from '../../logger';
 import {DeviceInfo} from '../deviceinfo';
 import {newTimeoutError} from "./Error";
+import {CrossSigningInfo} from "../CrossSigning";
+import {decodeBase64} from "../olmlib";
 
 const timeoutException = new Error("Verification timed out");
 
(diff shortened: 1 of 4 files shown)


src/crypto/verification/Base.js
```diff
@@ -188,6 +192,63 @@ export class VerificationBase extends EventEmitter {
...
+                        const decoded = decodeBase64(result);
+                        return Uint8Array.from(decoded);
+                    } },
+                    original._cacheCallbacks,
+                );
+                crossSigning.keys = original.keys;
+
+                // XXX: get all keys out if we get one key out
+                // XXX: https://github.com/vector-im/riot-web/issues/12604
+                // XXX: then change here to reject on the timeout
```
turt2live: Please make the XXX a single line so IDEs don't think there's 3 separate issues here.

turt2live:
actually, going to request a second opinion to make sure there's not a security issue here.

foldleft:
I don't think _this part_ risks a security issue – all it does is _request_ keys, and it validates that the keys it gets back match a public key it already has.  ie all we can do is gain extra private keys that we didn't have before.  The security review might be more merited on the counterpart to this over on `matrix-react-sdk` though :)

PR #1247, opened 2020-03-05
keyFromPassphrase compability for node

unclejay80:
client.restoreKeyBackupWithPassword was not working on node because of missing implementation using node "crypto" lib.

Signed-off-by: Jürgen Wischer <github@unclejay.de>


jryans:
Since we haven't seen a reply here in a bit and we'd need to restart this work to clean up history, I'll close this PR for now. If you'd like make another attempt, please feel free to reopen or start another PR. 😄

PR #1248, opened 2020-03-06
Cleanup: remove crypto.verification.start event

bwindels:
After https://github.com/matrix-org/matrix-react-sdk/pull/4180 is merged, this will not be used by the react-sdk anymore. Other users of the js-sdk might be using it (although I am not aware of any) so this is a breaking change.


jryans:
Hmm, I do wish we had a way to search typical JS SDK consumers to check whether things like this are used...

jryans:
If we knew no one was listening for the event, I'd be tempted to ignore the breaking change, but only if we can have confidence about that.

PR #1252, opened 2020-03-09
refactor megolm encryption to improve perceived speed

uhoreg:
- allow applications to pre-send decryption keys before the message is sent
- establish new olm sessions with a shorter timeout first, and then re-try in the background with a longer timeout without blocking message sending

fixes https://github.com/vector-im/riot-web/issues/11836

### DIFF
--- a/spec/unit/crypto/algorithms/megolm.spec.js
+++ b/spec/unit/crypto/algorithms/megolm.spec.js
@@ -320,14 +320,14 @@ describe("MegolmDecryption", function() {
 
             // this should have claimed a key for alice as it's starting a new session
             expect(mockBaseApis.claimOneTimeKeys).toHaveBeenCalledWith(
-                [['@alice:home.server', 'aliceDevice']], 'signed_curve25519',
+                [['@alice:home.server', 'aliceDevice']], 'signed_curve25519', 2000,
             );
             expect(mockCrypto.downloadKeys).toHaveBeenCalledWith(
                 ['@alice:home.server'], false,
             );
             expect(mockBaseApis.sendToDevice).toHaveBeenCalled();
             expect(mockBaseApis.claimOneTimeKeys).toHaveBeenCalledWith(
-                [['@alice:home.server', 'aliceDevice']], 'signed_curve25519',
+                [['@alice:home.server', 'aliceDevice']], 'signed_curve25519', 2000,
             );
 
             mockBaseApis.claimOneTimeKeys.mockReset();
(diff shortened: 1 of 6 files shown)


spec/unit/crypto/algorithms/megolm.spec.js
```diff
@@ -540,6 +540,12 @@ describe("MegolmDecryption", function() {
             content: {},
         });
         await aliceClient._crypto.encryptEvent(event, aliceRoom);
+        await new Promise((resolve) => {
+            // encryptMessage retries senders in the background before giving
+            // up and telling them that there's no olm channel, so we need to
+            // wait a bit before checking that we got the message
+            setTimeout(resolve, 100);
```
foldleft: Do we use the `jest.useFakeTImers()` API?
https://jestjs.io/docs/en/timer-mocks

src/crypto/algorithms/megolm.js
```diff
@@ -260,49 +246,100 @@ MegolmEncryption.prototype._ensureOutboundSession = async function(
...
+                await this._shareKeyWithOlmSessions(
+                    session, key, payload, olmSessions,
+                );
+            })(),
+            (async () => {
+                // meanwhile, establish olm sessions for devices that we don't
+                // already have a session for, and share keys with them.  Use a
+                // shorter timeout when fetching one-time keys.
+                await this._shareKeyWithDevices(
+                    session, key, payload, devicesWithoutSession, errorDevices, 2000,
```
foldleft: constant?

src/crypto/algorithms/megolm.js
```diff
@@ -766,6 +762,37 @@ MegolmEncryption.prototype._notifyBlockedDevices = async function(
...
 };
 
+/**
+ * Perform any background tasks that can be done before a message is ready to
+ * send, in order to speed up sending of the message.
+ *
+ * @param {module:models/room} room the room the event is in
+ */
+MegolmEncryption.prototype.prepareToEncrypt = function(room) {
+    logger.log(`Preparing to encrypt events for ${this._roomId}`);
```
foldleft: Is this for every message sent? Could get spammy, maybe `logger.debug`?

src/crypto/algorithms/megolm.js
```diff
@@ -766,6 +762,37 @@ MegolmEncryption.prototype._notifyBlockedDevices = async function(
...
+
+    if (this.encryptionPreparation) {
+        // We're already preparing something, so don't do anything else.
+        // FIXME: check if we need to restart
```
foldleft: Ideally, file a task and link here

spec/unit/crypto/algorithms/megolm.spec.js
uhoreg: I think we have some things that use jest timer mocks, but I'm not sure how to use them in the context of waiting for a background task to complete.  Though looking at it again, I see a way of doing it without the timer at all.

src/crypto/algorithms/megolm.js
uhoreg: I'm not sure that it's worth creating a constant for. :man_shrugging:

PR #1254, opened 2020-03-10
If a key upload fails, throw an error and emit an event

foldleft:
Relates to: https://github.com/vector-im/riot-web/issues/11574

I think the only user-visible change here is that when they fail to upload a key, a cancellation event now fires and so both sides see the change as cancelled. 

Inexplicably, the device that throws considers the other to be verified until it's refreshed.  I cannot track down why this is, and it looks like the way it's meant to work is that the key is uploaded and then retrieved in the next sync.

I think this task only half-completes the matter – the other is to surface it to the user.


foldleft:
nb: feverish

not sure if genius or idiot, but now i hand the event a callback which retries and returns a promise

PR #1263, opened 2020-03-16
Fix editing of unsent messages by waiting for actual event id

t3chguy:
Fixes vector-im/riot-web#11834
Supersedes https://github.com/matrix-org/matrix-react-sdk/pull/4220

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -287,7 +287,15 @@ export function MatrixClient(opts) {
                 _updatePendingEventStatus(room, eventToSend,
                                           EventStatus.SENDING);
             }
-            return _sendEventHttpRequest(self, eventToSend);
+            const promise = _sendEventHttpRequest(self, eventToSend);
+            if (room) {
+                // ensure we update pending event before the next scheduler run so that any listeners to event id
+                // updates on the synchronous event emitter get a chance to run first.
+                promise.then(res => {
+                    room.updatePendingEvent(eventToSend, EventStatus.SENT, res.event_id);
+                });
+            }
+            return promise;
         });
     }
     this.clientRunning = false;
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -287,7 +287,15 @@ export function MatrixClient(opts) {
                 _updatePendingEventStatus(room, eventToSend,
                                           EventStatus.SENDING);
             }
-            return _sendEventHttpRequest(self, eventToSend);
+            const promise = _sendEventHttpRequest(self, eventToSend);
+            if (room) {
+                // ensure we update pending event before the next scheduler run so that any listeners to event id
+                // updates on the synchronous event emitter get a chance to run first.
+                promise.then(res => {
```
foldleft: should this be returned, so anyone waiting on the promise also waits on this? if not, probably best to leave a comment that this is deliberate

src/client.js
```diff
@@ -287,7 +287,16 @@ export function MatrixClient(opts) {
                 _updatePendingEventStatus(room, eventToSend,
                                           EventStatus.SENDING);
             }
-            return _sendEventHttpRequest(self, eventToSend);
+            let promise = _sendEventHttpRequest(self, eventToSend);
```
bwindels: nit: you could make this callback async, then you can just await here and don't need the promise reassignment below.

src/client.js
```diff
@@ -281,22 +281,19 @@ export function MatrixClient(opts) {
...
-            let promise = _sendEventHttpRequest(self, eventToSend);
+            const res = await _sendEventHttpRequest(self, eventToSend);
             if (room) {
                 // ensure we update pending event before the next scheduler run so that any listeners to event id
                 // updates on the synchronous event emitter get a chance to run first.
-                promise = promise.then(res => {
-                    room.updatePendingEvent(eventToSend, EventStatus.SENT, res.event_id);
-                    return res;
-                });
+                room.updatePendingEvent(eventToSend, EventStatus.SENT, res.event_id);
```
bwindels: before you were returning the value of this call, is that not needed anymore?

gdvine:
Nice work!

PR #1267, opened 2020-03-17
handle racing .start event during self verification

bwindels:
by comparing the device id rather than the user id, as defined in [MSC2366](https://github.com/uhoreg/matrix-doc/blob/key_verification_accept/proposals/2366-key-verification-accept.md):

> After the m.key.verification.ready event is sent, either party can send an m.key.verification.start event to begin the verification. If both parties send an m.key.verification.start event, and they both specify the same verification method, then the event sent by the user whose user ID is the smallest is used, and the other m.key.verification.start event is ignored. **In the case of a single user verifying two of their devices, the device ID is compared instead.**

Fixes https://github.com/vector-im/riot-web/issues/12764

### DIFF
--- a/src/crypto/verification/request/VerificationRequest.js
+++ b/src/crypto/verification/request/VerificationRequest.js
@@ -566,6 +566,43 @@ export class VerificationRequest extends EventEmitter {
         return newTransitions;
     }
 
+    _isWinningStartRace(newEvent) {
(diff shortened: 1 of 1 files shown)


bwindels:
tested that the race handling now works for both self-verification and user verification.

PR #1269, opened 2020-03-18
don't kill the whole device download if one device gives an error

uhoreg:
fixes https://github.com/vector-im/riot-web/issues/12761


uhoreg:
The error in this case was that the server managed to get a new MSK but not the corresponding SSK, so the SSK that it sent to the client wasn't signed by the MSK.  I don't think it will lead to UISI, though it might show the user as untrusted.

PR #1271, opened 2020-03-19
Add a store for backup keys

foldleft:
Relates to: https://github.com/vector-im/riot-web/issues/12704


foldleft:
I suppose not, I just thought semantically that `storeCrossSigningPrivateKey` didn't seem right.  Perhaps we could rename this `storeSSSSPrivateKey` and use that instead?

More broadly, I was doing it like this because the entire process looks like it has to happen differently, and I guess I figured different things look different.

foldleft:
yeah, good catch

PR #1272, opened 2020-03-19
Migration to symmetric SSSS

dbkr:
Mostly from @uhoreg

Fixes https://github.com/vector-im/riot-web/issues/12508
Fixes https://github.com/vector-im/riot-web/issues/12431


src/crypto/CrossSigning.js
```diff
@@ -141,14 +141,26 @@ export class CrossSigningInfo extends EventEmitter {
...
+     *     with, or null if it is not present or not encrypted with a trusted
+     *     key
      */
     async isStoredInSecretStorage(secretStorage) {
-        let stored = true;
-        for (const type of ["master", "self_signing", "user_signing"]) {
-            stored &= await secretStorage.isStored(`m.cross_signing.${type}`, false);
+        const stored =
+              await secretStorage.isStored("m.cross_signing.master", false) || {};
+        function intersect(s) {
```
jryans: So this is to ensure all 3 are stored using the same key...? Maybe worth commenting.

src/crypto/SecretStorage.js
```diff
@@ -19,8 +19,225 @@ import {logger} from '../logger';
...
+export const SECRET_STORAGE_ALGORITHM_V1_CURVE25519
+    = "m.secret_storage.v1.curve25519-aes-sha2";
+
+const subtleCrypto = (typeof window !== "undefined" && window.crypto) ?
+    (window.crypto.subtle || window.crypto.webkitSubtle) : null;
+
+// salt for HKDF, with 8 bytes of zeros
+const zerosalt = new Uint8Array(8);
+
+/** encrypt a string in Node.js
```
jryans: Most of our function docs (including others in this file) use an empty first line, with summary on the next line:

```
/**
 * encrypt a string in Node.js
```

...so it would be nice to match.

src/crypto/SecretStorage.js
```diff
@@ -19,8 +19,225 @@ import {logger} from '../logger';
...
+ *
+ * @param {string} data the plaintext to encrypt
+ * @param {Uint8Array} key the encryption key to use
+ * @param {string} name the name of the secret
+ */
+async function encryptNode(data, key, name) {
```
jryans: Are these encryption / decryption flows specific to secret storage? They seem fairly generic, so I'd suggest moving them to a new file for easier comprehension.

src/crypto/SecretStorage.js
```diff
@@ -207,24 +438,13 @@ export class SecretStorage extends EventEmitter {
                 throw new Error("Unknown key: " + keyId);
             }
 
-            // check signature of key info
```
jryans: Hmm, so we don't need to verify the 4S key is trusted when storing secrets now...?

src/crypto/index.js
```diff
@@ -395,10 +399,80 @@ Crypto.prototype.bootstrapSecretStorage = async function({
...
+            info => info.algorithm === SECRET_STORAGE_ALGORITHM_V1_AES,
+        ))) {
+            // we already have cross-signing keys, but they're encrypted using
+            // the old algorithm
+            logger.log("Switching to symmetric");
+            const keys = {};
+            // fetch the cross-signing private keys (needed to sign the
+            // new SSSS key)
+            this._baseApis._cryptoCallbacks.getCrossSigningKey =
+                name => crossSigningPrivateKeys[name];
```
jryans: How is this working? Only the existing branch [below](https://github.com/matrix-org/matrix-js-sdk/pull/1272/files#diff-fa4472db829c5a59596c29a2f43e8f9cR488) seems to attach a function that would write to `crossSigningPrivateKeys`...

src/crypto/index.js
```diff
@@ -424,40 +498,50 @@ Crypto.prototype.bootstrapSecretStorage = async function({
...
-                );
+                if (!newKeyId) {
+                    const opts = {};
 
-                // Add an entry for the backup key in SSSS as a 'passthrough' key
-                // (ie. the secret is the key itself).
-                this._secretStorage.storePassthrough('m.megolm_backup.v1', newKeyId);
+                    if (
+                        keyBackupInfo.auth_data.private_key_salt &&
+                            keyBackupInfo.auth_data.private_key_iterations
```
jryans: Nit: Consider reducing indent to match line above

src/crypto/SecretStorage.js
dbkr: Yeah, seems sensible - done

src/crypto/SecretStorage.js
uhoreg: Yes.  We know the key is trusted because the user entered the secret key.

src/crypto/SecretStorage.js
```diff
@@ -362,55 +558,82 @@ export class SecretStorage extends EventEmitter {
      * @param {string} name the name of the secret
      * @param {boolean} checkKey check if the secret is encrypted by a trusted key
      *
-     * @return {boolean} whether or not the secret is stored
+     * @return {object?} map of key name to key info the secret is encrypted
+     *     with, or null if it is not present or not encrypted with a trusted
+     *     key
      */
     async isStored(name, checkKey) {
```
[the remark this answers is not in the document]
uhoreg: Changed the docs

src/crypto/CrossSigning.js
uhoreg: added a comment

src/crypto/index.js
uhoreg: added some comments.  Hopefully it's clearer

PR #1276, opened 2020-03-20
Make screen sharing works on Chrome using getDisplayMedia()

mjattiot:
Screen sharing has never been working on recent versions of Chrome (fixes https://github.com/vector-im/riot-web/issues/6181).

This PR intends to use the new API getDisplayMedia() to make it works on all recent browsers : https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getDisplayMedia

Signed-off-by: Maxime Jattiot <mjattiot@opensense.fr>

### DIFF
--- a/src/webrtc/call.js
+++ b/src/webrtc/call.js
@@ -171,7 +171,6 @@ MatrixCall.prototype.placeScreenSharingCall =
     this.localVideoElement = localVideoElement;
     this.remoteVideoElement = remoteVideoElement;
     const self = this;
-
     try {
         self.screenSharingStream = await this.webRtc.getDisplayMedia({'audio': false});
         debuglog("Got screen stream, requesting audio stream...");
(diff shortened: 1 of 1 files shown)


jryans:
MDN [support info](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getDisplayMedia#Browser_compatibility) is apparently outdated, as it seems Safari 13 on desktop is supported as well.

We should test in all 3 supported browsers as part of the review to be sure it does indeed work.

mjattiot:
Well the function exists on Safari 13:
Screenshot 2020-03-27 at 09 37 42

However I tried to make it works on a console without success, nothing happens when I click:
Screenshot 2020-03-27 at 09 42 22

And it seems Riot on Safari doesn't support it anyway ...
Screenshot 2020-03-27 at 09 50 22

Still this feature is a must have for Chrome and supported by Firefox as well. So I campaign for a merge ! :)

PR #1277, opened 2020-03-21
Allow guests to fetch TURN servers

### LINKED ISSUE #1274: Guests are not allowed to use TURN servers
Synapse allows to configure that TURN servers are available for guests. The matrix-js-sdk prevents all guests from requesting the TURN server from the server (Synapse) in the `checkTurnServers` function in the `client.js` file.

Is there a way to check the servers configuration before blocking TURN servers?

awesome-michael:
Allows guests to use the servers endpoint to try and fetch TURN servers. The configuration at the server decides if they are allowed to do this and either return the TURN servers or an GUEST_ACCESS_FORBIDDEN error.

Fixes #1274

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -5130,9 +5130,6 @@ function checkTurnServers(client) {
     if (!client._supportsVoip) {
         return;
     }
-    if (client.isGuest()) {
-        return; // guests can't access TURN servers
-    }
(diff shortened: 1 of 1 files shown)


awesome-michael:
like this?

PR #1279, opened 2020-03-23
Add function for checking cross-signing is ready

dbkr:
Aggregate function that checks the various things are in place for
cross-signing to work.

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -353,6 +353,37 @@ Crypto.prototype.createRecoveryKeyFromPassphrase = async function(password) {
 };
 
 /**
+ * Checks whether cross signing:
+ * - is enabled on this account
+ * - is trusted by this device
+ * - has private keys stored in secret storage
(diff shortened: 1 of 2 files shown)


src/crypto/index.js
```diff
@@ -352,6 +352,37 @@ Crypto.prototype.createRecoveryKeyFromPassphrase = async function(password) {
...
+ * If this function returns false, bootstrapSecretStorage() can be used
+ * to fix things such that it returns true. That is to say, after
+ * bootstrapSecretStorage() completes sucessfully, this function should
+ * return true.
+ *
+ * The cross-signing API is currently UNSTABLE and may change without notice.
+ *
+ * @return {bool} True if cross-signing is ready to be used on this device
+ */
+Crypto.prototype.crossSigningReady = async function() {
```
jryans: Not sure how strict we want to be, but maybe:

```suggestion
Crypto.prototype.isCrossSigningReady = async function() {
```

PR #1285, opened 2020-03-25
Add a flag for whether cross signing signatures are trusted

dbkr:
<del>TODO: Double check things are using the appropriate `isVerified / isCrossSigningVerified` for what they actually care about as this will only apply for things that check `isVerified()`.</del>

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -1078,8 +1078,11 @@ Crypto.prototype._checkDeviceInfoTrust = function(userId, device) {
 
     const userCrossSigning = this._deviceList.getStoredCrossSigningForUser(userId);
     if (device && userCrossSigning) {
+        // The _trustCrossSignedDevices only affects trust of other people's cross-signing
+        // signatures
+        const trustCrossSigning =  this._trustCrossSignedDevices || userId === this._userId;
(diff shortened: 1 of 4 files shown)


src/crypto/index.js
```diff
@@ -319,6 +321,44 @@ Crypto.prototype.init = async function(opts) {
...
+ */
+Crypto.prototype.setCryptoTrustCrossSignedDevices = function(val) {
+    this._trustCrossSignedDevices = val;
+
+    for (const userId of this._deviceList.getKnownUserIds()) {
+        const devices = this._deviceList.getRawStoredDevicesForUser(userId);
+        for (const deviceId of Object.keys(devices)) {
+            const deviceTrust = this.checkDeviceTrust(userId, deviceId);
+            if (
+                !deviceTrust.isLocallyVerified() &&
```
bwindels: How about the device is trusted locally but not by cross-signing? Should we also emit then? Or perhaps we fallback to local trust elsewhere when cross-signing is enabled?

dbkr: Yeah, this only changes the return value of `isVerified()` which is always true if `isLocallyVerified()` is true. Have commented.

PR #1291, opened 2020-03-27
Request the key backup key & restore backup

dbkr:
After a successful verification with ourselves, request the key
backup key too and restore a key backup if we get it.

Also cache the key backup key when we cache the SSK & USK so we have
it available to share.

Fixes https://github.com/vector-im/riot-web/issues/12704

### DIFF
--- a/src/crypto/verification/Base.js
+++ b/src/crypto/verification/Base.js
@@ -1,5 +1,6 @@
 /*
 Copyright 2018 New Vector Ltd
+Copyright 2020 The Matrix.org Foundation C.I.C.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


src/crypto/verification/Base.js
```diff
@@ -237,11 +240,39 @@ export class VerificationBase extends EventEmitter {
...
+                const backupKeyPromise = new Promise(async resolve => {
+                    const cachedKey = await client._crypto.getSessionBackupPrivateKey();
+                    if (!cachedKey) {
+                        logger.info("No cached backup key found. Requesting...");
+                        const secretReq = client.requestSecret(
+                            'm.megolm_backup.v1', [this.deviceId],
+                        );
+                        const base64Key = await secretReq.promise;
+                        logger.info("Got key backup key, decoding...");
+                        const decodedKey = decodeBase64(base64Key);
```
foldleft: this could be null

PR #1292, opened 2020-03-30
Track whether we have verified a user before

jryans:
This tracks whether we have ever cross-signing verified a user before (at least
as far as the current device has ever observed). This info helps to present an
alert in case the user subsequently becomes unverified.

Part of https://github.com/vector-im/riot-web/issues/12808
Used by https://github.com/matrix-org/matrix-react-sdk/pull/4303


jryans:
> If it is supposed to be persisted somewhere, I probably just missed where that's happening.

It's maybe not so obvious at a glance... This patch updates the raw cross-signing data in the device list, which is persisted to IDB.

PR #1293, opened 2020-03-30
Receive private key for caching from the app layer

jryans:
This passes through the new secret storage key as well as the key info so that
bootstrap can use it immediately without triggering an immediate prompt.

Part of https://github.com/vector-im/riot-web/issues/12867
Used by https://github.com/matrix-org/matrix-react-sdk/pull/4308

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -368,7 +368,7 @@ Crypto.prototype.setCryptoTrustCrossSignedDevices = function(val) {
  * @param {string} password Passphrase string that can be entered by the user
  *     when restoring the backup as an alternative to entering the recovery key.
  *     Optional.
- * @returns {Promise<Array>} Array with public key metadata, encoded private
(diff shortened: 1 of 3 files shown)


src/crypto/index.js
```diff
@@ -637,12 +637,13 @@ Crypto.prototype.bootstrapSecretStorage = async function({
             } else {
                 if (!newKeyId) {
                     logger.log("Secret storage default key not found, creating new key");
-                    const keyOptions = await createSecretStorageKey();
+                    const [keyOptions, , key] = await createSecretStorageKey();
```
turt2live: We should probably document the call signature wherever this gets defined.

jryans: While working on this yesterday, I had wanted to change this to an object for easier use, but didn't want to make a breaking change... Today though, I noticed it's part of the unstable API surface for cross-signing.

I've changed this to return a object instead and updated docs to match.

PR #1294, opened 2020-03-30
Add ability to check symmetric SSSS key before we try to use it

uhoreg:
fixes (partially) https://github.com/vector-im/riot-web/issues/12897 (along with https://github.com/matrix-org/matrix-react-sdk/pull/4309)

Only remaining question is if we need to migrate users who already have an existing Symmetric SSSS.

### DIFF
--- a/src/crypto/SecretStorage.js
+++ b/src/crypto/SecretStorage.js
@@ -27,6 +27,8 @@ export const SECRET_STORAGE_ALGORITHM_V1_AES
 export const SECRET_STORAGE_ALGORITHM_V1_CURVE25519
     = "m.secret_storage.v1.curve25519-aes-sha2";
 
+const ZERO_STR = "\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0";
+
 /**
  * Implements Secure Secret Storage and Sharing (MSC1946)
  * @module crypto/SecretStorage
@@ -92,13 +94,13 @@ export class SecretStorage extends EventEmitter {
         switch (algorithm) {
         case SECRET_STORAGE_ALGORITHM_V1_AES:
         {
-            const decryption = new global.Olm.PkDecryption();
-            try {
-                if (opts.passphrase) {
-                    keyData.passphrase = opts.passphrase;
-                }
-            } finally {
-                decryption.free();
+            if (opts.passphrase) {
+                keyData.passphrase = opts.passphrase;
+            }
+            if (opts.key) {
+                const {iv, mac} = await encryptAES(ZERO_STR, opts.key, "");
+                keyData.iv = iv;
+                keyData.mac = mac;
             }
             break;
         }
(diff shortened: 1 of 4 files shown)


src/crypto/index.js
```diff
@@ -609,12 +612,14 @@ Crypto.prototype.bootstrapSecretStorage = async function({
                         };
                     }
 
+                    // use the backup key as the new ssss key
+                    ssssKeys[newKeyId] = backupKey;
+                    opts.key = backupKey;
```
jryans: Do we need to make a similar change near "Secret storage default key not found" below (line 648) which also adds a 4S key?

uhoreg: This will be handled by the change at https://github.com/matrix-org/matrix-js-sdk/pull/1294/files#diff-fa4472db829c5a59596c29a2f43e8f9cR390 because react-sdk uses that function to get the backup key.

src/crypto/SecretStorage.js
```diff
@@ -92,13 +94,13 @@ export class SecretStorage extends EventEmitter {
...
-                if (opts.passphrase) {
-                    keyData.passphrase = opts.passphrase;
-                }
-            } finally {
-                decryption.free();
+            if (opts.passphrase) {
+                keyData.passphrase = opts.passphrase;
+            }
+            if (opts.key) {
+                const {iv, mac} = await encryptAES(ZERO_STR, opts.key, "");
```
jryans: I guess this is the part that we won't have for existing keys, which leads to your potential migration question...?

Not quite sure what "style" of migration you had in mind for this one... Could we potentially do this one "silently": as in we would not toast or prompt about it, but just whenever we next happen to unlock 4S successfully and notice that this extra data is missing, we could add it in for the future...?

I am very curious what @dbkr thinks about this as well.

src/crypto/index.js
```diff
@@ -387,7 +387,7 @@ Crypto.prototype.createRecoveryKeyFromPassphrase = async function(password) {
         } else {
             keyInfo.pubkey = decryption.generate_key();
         }
-        const privateKey = decryption.get_private_key();
+        const privateKey = keyInfo.key = decryption.get_private_key();
```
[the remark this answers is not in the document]
uhoreg: So, I'm blind and I didn't notice that the private key is already returned, *and* it's already fetched in the bootstrap function, which can add it to the keyInfo itself.  So I've fixed this in the latest version.

src/crypto/aes.js
```diff
@@ -112,10 +112,16 @@ function deriveKeysNode(key, name) {
  * @param {string} data the plaintext to encrypt
  * @param {Uint8Array} key the encryption key to use
  * @param {string} name the name of the secret
+ * @param {string} ivStr the initialization vector to use
```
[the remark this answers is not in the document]
uhoreg: Yup.  Done

src/crypto/SecretStorage.js
uhoreg: Yeah, we could either do an upgrade the next time the user enters their passphrase, or we could do the whole toast-and-prompt dance.

dbkr: Nicest would probably be if we checked the key next time we did a decryption, adding this info if it succeeded. This would still presumably end up with secrets encrypted with 2 different keys if we added a secret before doing this and the user got their passphrase wrong? But also presumably this is what happens now?

PR #1298, opened 2020-04-02
Expose function to force-reset outgoing room key requests

foldleft:
Part 1 of 2; still needs a `matrix-react-sdk` counterpart.

Relates to https://github.com/vector-im/riot-web/issues/12843

### DIFF
--- a/src/crypto/store/indexeddb-crypto-store-backend.js
+++ b/src/crypto/store/indexeddb-crypto-store-backend.js
@@ -60,7 +60,6 @@ export class Backend {
         return new Promise((resolve, reject) => {
             const txn = this._db.transaction("outgoingRoomKeyRequests", "readwrite");
             txn.onerror = reject;
-            txn.onabort = (ev) => reject(ev.target.error);
 
             // first see if we already have an entry for this request.
             this._getOutgoingRoomKeyRequest(txn, requestBody, (existing) => {
@@ -218,7 +217,6 @@ export class Backend {
 
             request.onsuccess = (ev) => resolve(ev.target.result);
             request.onerror = (ev) => reject(ev.target.error);
-            request.onabort = (ev) => reject(ev.target.error);
         });
     }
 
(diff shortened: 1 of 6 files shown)


src/crypto/store/indexeddb-crypto-store-backend.js
```diff
@@ -203,6 +204,24 @@ export class Backend {
...
+    getAllOutgoingRoomKeyRequestsByState(wantedState) {
+        return new Promise((resolve, reject) => {
+            const txn = this._db.transaction("outgoingRoomKeyRequests", "readonly");
+            const store = txn.objectStore("outgoingRoomKeyRequests");
+            const index = store.index("state");
+            const request = index.getAll(wantedState);
+
+            request.onsuccess = (ev) => resolve(ev.target.result);
+            request.onerror = (ev) => reject(ev.target.error);
+            request.onabort = (ev) => reject(ev.target.error);
```
bwindels, line 221: `onabort` doesn't seem to be a thing on an `IDBRequest` according to MDN. I think an aborted txn rather uses onerror.

src/crypto/store/indexeddb-crypto-store-backend.js
```diff
@@ -60,6 +60,7 @@ export class Backend {
         return new Promise((resolve, reject) => {
             const txn = this._db.transaction("outgoingRoomKeyRequests", "readwrite");
             txn.onerror = reject;
+            txn.onabort = (ev) => reject(ev.target.error);
```
bwindels, line 63: Might be missing something, but looks like you could use `promiseifyTxn` here.

src/crypto/store/indexeddb-crypto-store-backend.js
foldleft: while I was testing I ran into a situation where it fired, but I can take this out

src/crypto/store/indexeddb-crypto-store-backend.js
foldleft: yeah okay

foldleft: nope, because we need to do another call first and need to get a handle on the `resolve` in linen 73

PR #1300, opened 2020-04-03
Add some logging around sync for stuck message debugging

turt2live:
For https://github.com/matrix-org/synapse/issues/7206

**Reviewer: if this looks okay, please merge it so the backend team can test with it ASAP.**

Intended to be removed (or possibly converted into a proper off-by-default setting) once the issue concludes.

There's a bunch of null checking through defaults here because this is a very critical part of the whole stack. We default to writing errors to the console, but do not break the app for failing to log info about the sync.

To enable, run `localStorage.setItem("mx_do_sync_logging", "true");` in the JS console

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -824,10 +824,56 @@ SyncApi.prototype._sync = async function(syncOptions) {
 
 SyncApi.prototype._doSyncRequest = function(syncOptions, syncToken) {
     const qps = this._getSyncParams(syncOptions, syncToken);
+    console.log("[sync_dbg] Doing sync with query string: ", qps);
     return this.client._http.authedRequest(
         undefined, "GET", "/sync", qps, undefined,
         qps.timeout + BUFFER_PERIOD_MS,
-    );
+    ).then(r => {
+        // TODO: Remove logging once https://github.com/matrix-org/synapse/issues/7206
+        // is fixed
+
+        try {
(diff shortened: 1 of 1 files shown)


src/sync.js
```diff
@@ -824,10 +824,55 @@ SyncApi.prototype._sync = async function(syncOptions) {
...
     return this.client._http.authedRequest(
         undefined, "GET", "/sync", qps, undefined,
         qps.timeout + BUFFER_PERIOD_MS,
-    );
+    ).then(r => {
+        // TODO: Remove logging once https://github.com/matrix-org/synapse/issues/7206
+        // is fixed
+
+        try {
+            if (localStorage && !localStorage.getItem("mx_do_sync_logging")) {
```
jryans, line 836: Huh, somehow [tests are still failing](https://buildkite.com/matrix-dot-org/matrix-js-sdk/builds/1480#1d3c676c-59ea-46c9-998c-deb5c384ad5f) here with `ReferenceError: localStorage is not defined` at least on Buildkite... 🤔

turt2live: ugh, the environment throws an error if it's so much as referenced :/

turt2live:
Looks like we're just going to have to give up on this and debug somehow else (like having devtools open to consume *all* the RAM)

PR #1304, opened 2020-04-06
Make auth argument in the register request compliant with r0.6.0

MTRNord:
This fix has a companion PR over at react-sdk. https://github.com/matrix-org/matrix-react-sdk/pull/4347

This fix is tested with synapse and works fine on matrix.org!

This fix is required for https://git.koesters.xyz/timo/matrixserver which implements the r0.6.0 spec. In that spec version it says that auth shouldn't be `auth: {}` but rather not be present. This fix should fix this. I tested it both with synapse and the linked server to check that everything still works

Signed-off-by: Marcel Radzio<mtrnord1@gmail.com>

### DIFF
--- a/src/interactive-auth.js
+++ b/src/interactive-auth.js
@@ -145,7 +145,7 @@ InteractiveAuth.prototype = {
 
             // if we have no flows, try a request (we'll have
             // just a session ID in _data if resuming)
-            if (!this._data.flows) {
+            if (this._data  && !this._data.flows) {
                 if (this._busyChangedCallback) this._busyChangedCallback(true);
                 this._doRequest(this._data).finally(() => {
                     if (this._busyChangedCallback) this._busyChangedCallback(false);
@@ -162,12 +162,12 @@ InteractiveAuth.prototype = {
      * be resolved.
      */
     poll: async function() {
-        if (!this._data.session) return;
+        if (this._data && !this._data.session) return;
         // if we currently have a request in flight, there's no point making
         // another just to check what the status is
         if (this._submitPromise) return;
 
-        let authDict = {};
+        let authDict = null;
         if (this._currentStage == EMAIL_STAGE_TYPE) {
             // The email can be validated out-of-band, but we need to provide the
             // creds so the HS can go & check it.
@@ -321,7 +321,7 @@ InteractiveAuth.prototype = {
         } catch (error) {
             // sometimes UI auth errors don't come with flows
             const errorFlows = error.data ? error.data.flows : null;
-            const haveFlows = Boolean(this._data.flows) || Boolean(errorFlows);
+            const haveFlows = Boolean(this._data && this._data.flows) || Boolean(errorFlows);
             if (error.httpStatus !== 401 || !error.data || !haveFlows) {
                 // doesn't look like an interactive-auth failure.
                 if (!background) {
@@ -364,7 +364,7 @@ InteractiveAuth.prototype = {
                         this._inputs.emailAddress,
                         this._clientSecret,
                         1, // TODO: Multiple send attempts?
-                        this._data.session,
+                        this._data ? this._data.session : null,
                     );
                     this._emailSid = requestTokenResult.sid;
                     // NB. promise is not resolved here - at some point, doRequest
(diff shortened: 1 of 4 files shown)


src/interactive-auth.js
```diff
@@ -318,14 +318,18 @@ InteractiveAuth.prototype = {
      */
     _doRequest: async function(auth, background) {
         try {
+            if (Object.keys(auth).length === 0 &&
+                auth.constructor === Object) auth = null;
+            console.warn(auth);
```
turt2live, line 323: This debugging should be removed

src/interactive-auth.js
```diff
@@ -162,14 +162,14 @@ InteractiveAuth.prototype = {
...
-        if (!this._data.session) return;
+        if (this._data && !this._data.session) return;
         // likewise don't poll if there is no auth session in progress
         if (!this._resolveFunc) return;
         // if we currently have a request in flight, there's no point making
         // another just to check what the status is
         if (this._submitPromise) return;
 
-        let authDict = {};
+        let authDict = null;
```
turt2live, line 172: This is supposed to be something by here

src/interactive-auth.js
```diff
@@ -162,14 +162,14 @@ InteractiveAuth.prototype = {
      * be resolved.
      */
     poll: async function() {
-        if (!this._data.session) return;
+        if (this._data && !this._data.session) return;
```
turt2live, line 165: we should have `this._data`

src/interactive-auth.js
```diff
@@ -318,14 +318,18 @@ InteractiveAuth.prototype = {
      */
     _doRequest: async function(auth, background) {
         try {
+            if (Object.keys(auth).length === 0 &&
+                auth.constructor === Object) auth = null;
```
turt2live, line 322: I don't think we need to modify `auth` here - we should be supplying the function with what it needs.

src/interactive-auth.js
```diff
@@ -318,14 +318,18 @@ InteractiveAuth.prototype = {
...
+            console.warn(auth);
             const result = await this._requestCallback(auth, background);
             this._resolveFunc(result);
             this._resolveFunc = null;
             this._rejectFunc = null;
         } catch (error) {
             // sometimes UI auth errors don't come with flows
             const errorFlows = error.data ? error.data.flows : null;
-            const haveFlows = Boolean(this._data.flows) || Boolean(errorFlows);
+            const haveFlows = Boolean(this._data &&
```
turt2live, line 331: we don't need this first `Boolean` cast anymore

src/interactive-auth.js
```diff
@@ -368,7 +372,7 @@ InteractiveAuth.prototype = {
                         this._inputs.emailAddress,
                         this._clientSecret,
                         1, // TODO: Multiple send attempts?
-                        this._data.session,
+                        this._data ? this._data.session : null,
```
turt2live, line 375: we should have a session by here

src/interactive-auth.js
MTRNord: Oh sorry missed that one. Will do that!

src/interactive-auth.js
MTRNord: I think you are right. This seems to be a bad workaround that I made. I should do it properly as I probably now broke things without realising.

turt2live:
@MTRNord is this ready for re-review?

MTRNord:
@turt2live No I didn't have time today to work on this. I will work on this tomorrow. (German timezone)

MTRNord:
@turt2live I think this is fixed now. The real change is made in the FallbackStage on the react-sdk side. I think this should be everything. Currently building the docker for another local test

MTRNord:
Actually no thats not the correct stage I changed. I am confused :/ I tracked it back to the InteractiveAuthEntryComponents.js in react-sdk but I am not sure which part gets triggered with the "m.login.dummy" stage

MTRNord:
@turt2live I give up. I cant get this to work anymore. Not sure where I need to change something to have that initial stage with auth: null instead auth: {} in the code.

turt2live:
the system is quite complicated, sorry :(

I'll try and take a look next week

turt2live:
I'm not going to be able to take a look, sorry :(

Leaving this in the pile of things people are welcome to try and help with.

turt2live:
right, this should be fixed and a whole lot of reruns later the CI is happy. Asking for review from someone else on the team given I've put some code into this now.

MTRNord:
Thanks for the work on this to make it actually work!

jryans:
(Merged now as this was our last release blocker, and I'd like to start the RC process.)

PR #1305, opened 2020-04-06
Avoid creating two devices on registration

dbkr:
Ignore poll requests after a UIA session has finished. Sometimes
the app calls poll() even after the registration has succeeded
which causes the API call to happen again which, in the case of
registration, creates a phantom device.

Fixes https://github.com/vector-im/riot-web/issues/13040

### DIFF
--- a/src/interactive-auth.js
+++ b/src/interactive-auth.js
@@ -163,6 +163,8 @@ InteractiveAuth.prototype = {
      */
     poll: async function() {
         if (!this._data.session) return;
+        // likewise don't poll if there is no auth session in progress
+        if (!this._resolveFunc) return;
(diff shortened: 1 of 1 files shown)


src/interactive-auth.js
```diff
@@ -318,6 +320,8 @@ InteractiveAuth.prototype = {
         try {
             const result = await this._requestCallback(auth, background);
             this._resolveFunc(result);
+            this._resolveFunc = null;
```
bwindels, line 323: Should we do the same when calling `_rejectFunc` below on line 387?

dbkr: oh yes, probably

PR #1307, opened 2020-04-07
Use the correct request body for the /keys/query endpoint.

MTRNord:
According to https://matrix.org/docs/spec/client_server/r0.6.0#post-matrix-client-r0-keys-query

The request should contain:

```
{
  "device_keys": {
    "@alice:example.com": []
  }
}
```

instead of

```
{
  "device_keys": {
    "@alice:example.com": {}
  }
}
```

which currently gets send.

Signed-off-by: Marcel Radzio<mtrnord1@gmail.com>

### DIFF
--- a/spec/TestClient.js
+++ b/spec/TestClient.js
@@ -185,7 +185,7 @@ TestClient.prototype.expectKeyQuery = function(response) {
         200, (path, content) => {
             Object.keys(response.device_keys).forEach((userId) => {
                 expect(content.device_keys[userId]).toEqual(
-                    {},
+                    [],
                     "Expected key query for " + userId + ", got " +
                     Object.keys(content.device_keys),
                 );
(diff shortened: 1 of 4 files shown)


MTRNord:
Not removing tests until it is certain that spec is correct and riot is wrong :D Sometimes the spec isn't updates fast enough

MTRNord:
People at ruma (and Timo who implements the new rust server) found this. I am only implementing their findings in riot/react-sdk/js-sdk as I have knowledge about riot stuff  :)

uhoreg:
The spec and this PR are correct.  It's an array of device IDs, or an empty array if the query is for all devices.  Synapse just does a `if not device_ids` to check if it's empty, which works with both arrays and objects, which is why synapse doesn't complain about it.

Once the unit tests are fixed, this should be fine to merge.

MTRNord:
@uhoreg @jryans all 4 tests should be fixed. Was a bit hard to find them all as the error log of the tests is not really helping (especially the last one was kinda hard to find)

PR #1308, opened 2020-04-07
Store key backup key in cache as Uint8Array

jryans:
Bootstrap was incorrectly storing the key backup key as base64 encoded, when
instead it should be a Uint8Array.

Fixes https://github.com/vector-im/riot-web/issues/13057


foldleft:
oh this is just embarrassing for me

jryans:
> oh this is just embarrassing for me

I think either @dbkr or myself added the part in bootstrap that's causing the confusion, so don't feel bad. 😅

PR #1311, opened 2020-04-09
Fix incorrect backup key format in SSSS

uhoreg:
fixes https://github.com/vector-im/riot-web/issues/12970

I'm planning on tidying up the unit tests a bit, but otherwise this should be ready for review.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -1822,7 +1822,7 @@ MatrixClient.prototype.restoreKeyBackupWithSecretStorage = async function(
     const fixedKey = fixBackupKey(storedKey);
     if (fixedKey) {
         const [keyId] = await this.getSecretStorageKey();
-        await this.storeSecret("m.megolm_backup.v1", [keyId]);
+        await this.storeSecret("m.megolm_backup.v1", fixedKey, [keyId]);
     }
 
     const privKey = decodeBase64(fixedKey || storedKey);
--- a/src/client.js
+++ b/src/client.js
@@ -42,7 +42,7 @@ import * as olmlib from "./crypto/olmlib";
 import {ReEmitter} from './ReEmitter';
 import {RoomList} from './crypto/RoomList';
 import {logger} from './logger';
-import {Crypto, isCryptoAvailable} from './crypto';
+import {Crypto, isCryptoAvailable, fixBackupKey} from './crypto';
 import {decodeRecoveryKey} from './crypto/recoverykey';
 import {keyFromAuthData} from './crypto/key_passphrase';
 import {randomString} from './randomstring';
@@ -1815,7 +1815,17 @@ MatrixClient.prototype.restoreKeyBackupWithPassword = async function(
 MatrixClient.prototype.restoreKeyBackupWithSecretStorage = async function(
     backupInfo, targetRoomId, targetSessionId, opts,
 ) {
-    const privKey = decodeBase64(await this.getSecret("m.megolm_backup.v1"));
+    const storedKey = await this.getSecret("m.megolm_backup.v1");
+
+    // ensure that the key is in the right format.  If not, fix the key and
+    // store the fixed version
+    const fixedKey = fixBackupKey(storedKey);
+    if (fixedKey) {
+        const [keyId] = await this.getSecretStorageKey();
+        await this.storeSecret("m.megolm_backup.v1", [keyId]);
+    }
+
+    const privKey = decodeBase64(fixedKey || storedKey);
     return this._restoreKeyBackup(
         privKey, targetRoomId, targetSessionId, backupInfo, opts,
     );
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -659,7 +659,8 @@ Crypto.prototype.bootstrapSecretStorage = async function({
                 const secretName = `m.cross_signing.${type}`;
                 const secret = await this.getSecret(secretName);
                 keys[type] = secret;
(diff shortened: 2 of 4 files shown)


src/crypto/index.js
```diff
@@ -501,172 +509,220 @@ Crypto.prototype.bootstrapSecretStorage = async function({
...
     try {
+        const oldSSSSKey = await this.getSecretStorageKey();
+        const [oldKeyId, oldKeyInfo] = oldSSSSKey || [null, null];
         const decryptionKeys =
               await this._crossSigningInfo.isStoredInSecretStorage(this._secretStorage);
         const inStorage = !setupNewSecretStorage && decryptionKeys;
-        if (decryptionKeys && !(Object.values(decryptionKeys).some(
+
+        if (!decryptionKeys && !keyBackupInfo) {
+            // either we don't have anything, or we've been asked to restart
```
jryans, line 572: Isn't the "asked to restart" case one where `setupNewSecretStorage === true` and thus `inStorage === false`?

src/crypto/index.js
```diff
@@ -711,8 +767,26 @@ Crypto.prototype.bootstrapSecretStorage = async function({
...
+                    "m.megolm_backup.v1", fixedBackupKey, [newKeyId || oldKeyId],
+                );
+            }
+            const decodedBackupKey = new Uint8Array(olmlib.decodeBase64(
+                fixedBackupKey || sessionBackupKey,
+            ));
+            await this.storeSessionBackupPrivateKey(decodedBackupKey);
+        }
+
+        if (setupNewKeyBackup && !keyBackupInfo) {
```
jryans, line 784: I recently moved this on develop to fix the key backup cache (https://github.com/matrix-org/matrix-js-sdk/pull/1303/commits/060549656e7f4806496daf7113c9e8144f9e868f), and it looks like you have it in the new position several blocks above, so this block can be removed...?

src/crypto/index.js
```diff
@@ -501,172 +509,220 @@ Crypto.prototype.bootstrapSecretStorage = async function({
...
+
+            // create new cross-signing keys
+            await resetCrossSigning();
+
+            // create a new SSSS key and use the backup key as the new SSSS key
+            const opts = {};
+
+            if (
+                keyBackupInfo.auth_data.private_key_salt &&
+                    keyBackupInfo.auth_data.private_key_iterations
```
jryans, line 616: Our current trend is to keep multi-line conditionals like this at the same indent on all lines, but probably we should try to automate / lint for this...

```suggestion
                keyBackupInfo.auth_data.private_key_iterations
```

src/crypto/index.js
```diff
@@ -501,172 +509,220 @@ Crypto.prototype.bootstrapSecretStorage = async function({
...
         };
 
+    // create a new SSSS key and set it as default
+    const createSSSS = async (opts, privateKey) => {
+        opts = opts || {};
+        if (privateKey) {
+            opts.key = privateKey;
+        }
+
+        const newKeyId = await this.addSecretStorageKey(
```
jryans, line 519: This shadows the other `newKeyId` in the outer scope, could we rename one of them?

src/crypto/index.js
```diff
@@ -501,172 +509,220 @@ Crypto.prototype.bootstrapSecretStorage = async function({
...
+            logger.log(
+                "Cross-signing private keys not found in secret storage, " +
+                    "creating new keys",
+            );
+
+            await resetCrossSigning();
+
+            if (oldKeyInfo && oldKeyInfo.algorithm === SECRET_STORAGE_ALGORITHM_V1_AES) {
+                // if we already have a default SSSS key; use it
+                newKeyId = oldKeyId;
```
jryans, line 583: Is this actually needed? It seems confusing to write to `newKeyId` when it's not actually new, and also all the places that read `newKeyId` seem to either have just generated a new key on the previous line or they check `newKeyId || oldKeyId`, which should cover this case.

src/client.js
```diff
@@ -1815,7 +1815,17 @@ MatrixClient.prototype.restoreKeyBackupWithPassword = async function(
...
 ) {
-    const privKey = decodeBase64(await this.getSecret("m.megolm_backup.v1"));
+    const storedKey = await this.getSecret("m.megolm_backup.v1");
+
+    // ensure that the key is in the right format.  If not, fix the key and
+    // store the fixed version
+    const fixedKey = fixBackupKey(storedKey);
+    if (fixedKey) {
+        const [keyId] = await this.getSecretStorageKey();
+        await this.storeSecret("m.megolm_backup.v1", [keyId]);
```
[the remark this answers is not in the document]
uhoreg: Yes.  I managed to fix the places in `crypto/index.js` where I made that mistake, and forgot to fix this one.

src/crypto/index.js
uhoreg: Hmm, yes.  `git stash` seems to have done something odd here.

src/crypto/index.js
uhoreg: Yes.  I think at one point, the code matched the comment...

uhoreg:
Manually testing with an actual Riot, I found one bug in the bootstrap process that prevented bootstrapping when starting with only key backup, which is now fixed in this PR.  I also have one scenario in which key backups don't seem to be working at all; I'm investigating whether it's due to this PR or something else.

uhoreg:
> I also have one scenario in which key backups don't seem to be working at all; I'm investigating whether it's due to this PR or something else.

FTR, it appears to be https://github.com/vector-im/riot-web/issues/13151

PR #1312, opened 2020-04-09
Clear cross-signing keys when detecting the keys have changed

bwindels:
Fixes https://github.com/vector-im/riot-web/issues/13056

reset-keys

### DIFF
--- a/src/crypto/CrossSigning.js
+++ b/src/crypto/CrossSigning.js
@@ -310,7 +310,9 @@ export class CrossSigningInfo extends EventEmitter {
         }
     }
 
-    /** unsets the keys, used when another session has reset the keys, to disable cross-signing */
+    /**
+     * unsets the keys, used when another session has reset the keys, to disable cross-signing
+     */
     clearKeys() {
         this.keys = {};
     }
--- a/src/crypto/CrossSigning.js
+++ b/src/crypto/CrossSigning.js
@@ -310,6 +310,11 @@ export class CrossSigningInfo extends EventEmitter {
         }
     }
 
+    /** unsets the keys, used when another session has reset the keys, to disable cross-signing */
+    clearKeys() {
+        this.keys = {};
+    }
+
     setKeys(keys) {
         const signingKeys = {};
         if (keys.master) {
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -1145,13 +1145,17 @@ Crypto.prototype._onDeviceListUserCrossSigningUpdated = async function(userId) {
             // If it's not changed, just make sure everything is up to date
             await this.checkOwnCrossSigningTrust();
         } else {
-            this.emit("crossSigning.keysChanged", {});
             // We'll now be in a state where cross-signing on the account is not trusted
             // because our locally stored cross-signing keys will not match the ones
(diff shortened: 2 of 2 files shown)


src/crypto/CrossSigning.js
```diff
@@ -310,6 +310,11 @@ export class CrossSigningInfo extends EventEmitter {
         }
     }
 
+    /** unsets the keys, used when another session has reset the keys, to disable cross-signing */
```
jryans, line 313: Maybe use multi-line doc comments to match others here?

```
/**
 * blah
 */
```

src/crypto/index.js
```diff
@@ -1145,13 +1145,17 @@ Crypto.prototype._onDeviceListUserCrossSigningUpdated = async function(userId) {
...
-            // on the server for our account. The app must call checkOwnCrossSigningTrust()
-            // to fix this.
-            // XXX: Do we need to do something to emit events saying every device has become
-            // untrusted?
+            // on the server for our account. So we clear our own stored cross-signing keys,
+            // effectively disabling cross-signing until the user gets verified by the device
+            // that reset the keys
+            this._storeTrustedSelfKeys(null);
+            // XXX this event doesn't seem to be used anywhere
+            // the DeviceListener in react-sdl does recheck though,
```
jryans, line 1155:
```suggestion
            // the DeviceListener in react-sdk does recheck though,
```

src/crypto/index.js
```diff
@@ -1145,13 +1145,17 @@ Crypto.prototype._onDeviceListUserCrossSigningUpdated = async function(userId) {
...
             // because our locally stored cross-signing keys will not match the ones
-            // on the server for our account. The app must call checkOwnCrossSigningTrust()
-            // to fix this.
-            // XXX: Do we need to do something to emit events saying every device has become
-            // untrusted?
+            // on the server for our account. So we clear our own stored cross-signing keys,
+            // effectively disabling cross-signing until the user gets verified by the device
+            // that reset the keys
+            this._storeTrustedSelfKeys(null);
+            // XXX this event doesn't seem to be used anywhere
```
jryans, line 1154: Maybe rewrite this a bit, as we're at least using it in the one place you mention on the next line...? 😄

src/crypto/index.js
```diff
@@ -1145,13 +1145,17 @@ Crypto.prototype._onDeviceListUserCrossSigningUpdated = async function(userId) {
...
+            // the DeviceListener in react-sdl does recheck though,
+            // as `crypto.devicesUpdated`` and `accountData` are
+            // emitted on the client when this case happens.
+            this.emit("crossSigning.keysChanged", {});
```
jryans, line 1158: What happens with the shield icons in room tiles, room headers, user info, etc.? Do they update or show wrong info? Perhaps they need to also watch for this event?

bwindels: So, room shields were already correctly updating (probably because the device trust event is emitted, which causes the trust to be completely re-evaluated for a room). But I agree it's not good to rely on this, so I created a react-sdk PR that also re-evaluates room trust on `crossSigning.keysChanged`. The comments should make more sense as well.

I also emit a user trust change to make sure if we're observing our own user somewhere (e.g. userinfo panel) it also updates correctly.

PR #1313, opened 2020-04-09
Fix: more informative error message when we cant find a key to decrypt with

bwindels:
Should give a more informative error message when something like https://github.com/vector-im/riot-web/issues/13105 happens.

### DIFF
--- a/src/crypto/SecretStorage.js
+++ b/src/crypto/SecretStorage.js
@@ -363,6 +363,11 @@ export class SecretStorage extends EventEmitter {
             }
         }
 
+        if (Object.keys(keys) === 0) {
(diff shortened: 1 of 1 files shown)


src/crypto/SecretStorage.js
```diff
@@ -363,6 +363,11 @@ export class SecretStorage extends EventEmitter {
             }
         }
 
+        if (Object.keys(keys) === 0) {
```
jryans, line 366:
```suggestion
        if (Object.keys(keys).length === 0) {
```

bwindels:
> Other than the possibly missing bit, seems good to have! smile

Oops, good catch! :)

PR #1320, opened 2020-04-16
Implement QR code reciprocate for self-verification with untrusted MSK

bwindels:
Fixes https://github.com/vector-im/riot-web/issues/13195

When doing self-verification with a QR code of a new web session (MSK still untrusted), we should mark our MSK as trusted.

### DIFF
--- a/src/crypto/verification/QRCode.js
+++ b/src/crypto/verification/QRCode.js
@@ -65,7 +65,7 @@ export class ReciprocateQRCode extends Base {
             this.emit("show_reciprocate_qr", this.reciprocateQREvent);
         });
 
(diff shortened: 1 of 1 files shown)


bwindels:
> This looks right but I think the comments about signing the keys (lines 68 & 90) are misleading - afaics those are the keys that we verifying with the other side (which may mean we then sign them).

It should not sign anything as Crypto:setDeviceVerification only signs other user's MSK or your own devices. My intention was just to mark the MSK as trusted, which would happen in Crypto:setDeviceVerification (called from the verifier).

I'll adjust the comment to `sign key/mark as verified`

PR #1327, opened 2020-04-17
Fix FilterComponent allowed_values check

nchamo:
Hey Matrix devs!!

I started working with the SDK a few days ago, and I found that I wasn't able to filter timeline events by type. I don't have much experience with JS so I could be wrong, but I looked at the code, and I don't think the condition could ever be `true`.

I made a change, and added a test based on some of the other tests on the repo.

If you would like me to make some changes, don't hesitate to let me know, and I'll make them. And if I got something wrong and the current behaviour is correct, please let me know and I'll close the PR 😄 

Signed-off-by: Nicolas Chamo <nicolas@chamo.com.ar>

### DIFF
--- a/src/filter-component.js
+++ b/src/filter-component.js
@@ -109,7 +109,7 @@ FilterComponent.prototype._checkFields =
 
         const allowed_values = self[name];
         if (allowed_values && allowed_values.length > 0) {
-            const anyMatch = allowed_values.map(match_func).reduce((a, b) => a || b);
+            const anyMatch = allowed_values.some(match_func);
             if (!anyMatch) {
                 return false;
             }
--- a/src/filter-component.js
+++ b/src/filter-component.js
@@ -108,8 +108,9 @@ FilterComponent.prototype._checkFields =
         }
 
         const allowed_values = self[name];
-        if (allowed_values) {
-            if (!allowed_values.map(match_func)) {
+        if (allowed_values && allowed_values.length > 0) {
+            const anyMatch = allowed_values.map(match_func).reduce((a, b) => a || b);
+            if (!anyMatch) {
                 return false;
             }
         }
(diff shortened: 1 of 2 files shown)


src/filter-component.js
```diff
@@ -108,8 +108,9 @@ FilterComponent.prototype._checkFields =
         }
 
         const allowed_values = self[name];
-        if (allowed_values) {
-            if (!allowed_values.map(match_func)) {
+        if (allowed_values && allowed_values.length > 0) {
+            const anyMatch = allowed_values.map(match_func).reduce((a, b) => a || b);
```
t3chguy, line 112:
```suggestion
            const anyMatch = allowed_values.some(match_func);
```

would be simpler unless I am missing something

nchamo: No, you are right. I didn't know something like `some` existed.

t3chguy:
Any chance you could put `Signed-off-by: Nicolas Chamo <nicolas@chamo.com.ar>` on your PR description, as it isn't on all of your commits?

Thanks

nchamo:
Sure, just added it to the description 😄

nchamo:
Hey @t3chguy !

Any chance we can get this fix into 5.3.1?

I have no problem in creating the release PR, just want to make sure I can get the fix into my project as soon as possible 😄

t3chguy:
Unlikely, even though it is minor, typically only regressions get cherry-picked into release branches. Can raise this in a meeting on Monday though.

t3chguy:
You could temporarily change your dependency to be direct from git, like 

`"matrix-js-sdk": "github:matrix-org/matrix-js-sdk#develop"`

nchamo:
Ok, I'll go with the git dependency then. Thanks for the quick response!

PR #1329, opened 2020-04-19
Reject attemptAuth promise if no auth flow found

MichaelKohler:
When there was no auth flow found, the attemptAuth promise didn't get rejected, which leads to the client not knowing that there was an error. This lead to issues like https://github.com/vector-im/riot-web/issues/13269, which should be fixed by this.

Due to my inexperience, I can't guarantee this won't break any flows in other clients though :/

Signed-off-by: Michael Kohler <me@michaelkohler.info>

Fixes https://github.com/vector-im/riot-web/issues/13269


MichaelKohler:
I have tested this with Riot on the Mozilla instance

PR #1332, opened 2020-04-20
Fix: don't do extra /filter request when enabling lazy loading of members

bwindels:
Fixes https://github.com/vector-im/riot-web/issues/13280

This had the side-effect that `room.timeline.limit` on the filter wasn't being set when using lazy loading. So since the introduction of lazy loading, we've been working with the default which is 10 IIRC. This will restore it to 20 as set in `MatrixClientPeg`.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -511,6 +511,12 @@ SyncApi.prototype.sync = function() {
         checkLazyLoadStatus(); // advance to the next stage
     }
 
+    function createDefaultFilter() {
+        const filter = new Filter(client.credentials.userId);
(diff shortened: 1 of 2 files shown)


src/sync.js
```diff
@@ -511,6 +511,12 @@ SyncApi.prototype.sync = function() {
         checkLazyLoadStatus(); // advance to the next stage
     }
 
+    function createDefaultFilter() {
```
jryans, line 514: Maybe call this `build...` to differentiate from `create` which is a network call for filters?

PR #1335, opened 2020-04-21
Convert bunch of things to TypeScript

t3chguy:


src/matrix.ts
```diff
@@ -16,8 +16,15 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import type Request from "request";
```
turt2live, line 19: I'm mildly concerned that this will cause webpack to bring in the whole request library when it shouldn't. A quick double check to see if the browser bundle is relatively sane (roughly the same size, as it doesn't exactly work right now) and that riot-web's webpack doesn't blow up would be appreciated.

t3chguy: Before:
image
After:
image

Somehow, its smaller 🤷 
image

and the build works so looks fine from my view

PR #1345, opened 2020-04-23
Retry account data upload requests

bwindels:
Related to https://github.com/vector-im/riot-web/issues/13216

This doesn't fix the issue completely, in that if the request fails more than 5 times, the problem will still occur. But should already be an improvement.

I've tested somewhat that this works.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -757,6 +757,7 @@ export async function retryNetworkOperation(maxAttempts, callback) {
         try {
             if (attempts > 0) {
                 const timeout = 1000 * Math.pow(2, attempts);
+                console.log(`network operation failed ${attempts} times, retrying in ${timeout}ms...`);
                 await new Promise(r => setTimeout(r, timeout));
             }
             return await callback();
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -21,6 +21,7 @@ limitations under the License.
  */
 
 import unhomoglyph from 'unhomoglyph';
+import {ConnectionError} from "./http-api";
 
 /**
  * Encode a dictionary of query parameters.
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -748,3 +749,32 @@ export function setCrypto(c: Object) {
...
+export async function retryNetworkOperation(maxAttempts, callback) {
+    let attempts = 0;
+    let success = false;
+    let lastConnectionError = null;
+    let result;
+    while (!success && attempts < maxAttempts) {
+        try {
+            if (attempts > 0) {
+                const timeout = 1000 * Math.pow(2, attempts);
+                await new Promise(r => setTimeout(r, timeout));
```
jryans, line 762: Should we log when retrying?

src/utils.ts
```diff
@@ -748,3 +749,32 @@ export function setCrypto(c: Object) {
 export function getCrypto(): Object {
     return crypto;
 }
+
+export async function retryNetworkOperation(maxAttempts, callback) {
```
jryans, line 753: Maybe more natural to put this in `http-api` as well, since it's meant to be used for network things...?

src/utils.ts
```diff
@@ -748,3 +749,32 @@ export function setCrypto(c: Object) {
...
+    let lastConnectionError = null;
+    let result;
+    while (!success && attempts < maxAttempts) {
+        try {
+            if (attempts > 0) {
+                const timeout = 1000 * Math.pow(2, attempts);
+                await new Promise(r => setTimeout(r, timeout));
+            }
+            result = await callback();
+            success = true;
```
jryans, line 765: Could also `return result` here perhaps, but either way is fine.

src/utils.ts
bwindels: I wasn't sure if we can do general request retrying without side-effects. I'd rather be conservative and do it just for account data to begin with.

jryans: Oh, that's fine and I agree it's risky for general requests and good to be conservative. I was just reacting to the placement in `utils.ts` when it seems fairly specific to network activity, so was just suggesting moving this function into the file `http-api.js` instead of here.

bwindels: Ah, sure :+1:

PR #1346, opened 2020-04-23
Prevent attempts to send olm messages to ourselves

dbkr:
The cause of this I've seen is us sending keyshare reqiests when
we didn't have keys but them arriving when we do, and us replying to
ourseles with keys, so exclude that explicity.

Also catch this lower down to catch any other code paths that
mistaklenly try this. Hopefully the comment explains enough as to why
it won't work.

### DIFF
--- a/src/crypto/olmlib.js
+++ b/src/crypto/olmlib.js
@@ -215,7 +215,7 @@ export async function ensureOlmSessionsForDevices(
                 // they get a message and see that the 'other side' has started a
                 // new chain when this side has an active sender chain.
                 // If you see this message being logged in the wild, we should find
-                // the thing that it trying to send Olm messages to itself and fix it.
+                // the thing that is trying to send Olm messages to itself and fix it.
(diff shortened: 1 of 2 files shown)


src/crypto/olmlib.js
```diff
@@ -207,6 +207,25 @@ export async function ensureOlmSessionsForDevices(
...
             const key = deviceInfo.getIdentityKey();
+
+            if (key === olmDevice.deviceCurve25519Key) {
+                // We should never be trying to start a session with ourself.
+                // Apart from talking to yourself being the first sign of madness,
+                // olm sessions can't do this because they get confused when
+                // they get a message and see that the 'other side' has started a
+                // new chain when this side has an active sender chain.
+                // If you see this message being logged in the wild, we should find
+                // the thing that it trying to send Olm messages to itself and fix it.
```
jryans, line 218:
```suggestion
                // the thing that is trying to send Olm messages to itself and fix it.
```

PR #1348, opened 2020-04-24
Don't send key requests until after sync processing is finished

dbkr:
Key requests wait for a short time (500ms) before being sent as
an attempt to wait for the key to arrive before sending a request
for it, but client startup takes way longer than this so this
would still result in key requests being sent for keys that we'd
fetched but were still waiting to be read out of the sync response
and put into the database.

### DIFF
--- a/src/crypto/OutgoingRoomKeyRequestManager.js
+++ b/src/crypto/OutgoingRoomKeyRequestManager.js
@@ -97,10 +97,6 @@ export class OutgoingRoomKeyRequestManager {
      */
     start() {
         this._clientRunning = true;
-
-        // set the timer going, to handle any requests which didn't get sent
-        // on the previous run of the client.
-        this._startTimer();
     }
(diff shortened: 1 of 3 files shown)


src/crypto/OutgoingRoomKeyRequestManager.js
```diff
@@ -302,7 +302,6 @@ export class OutgoingRoomKeyRequestManager {
                                 "Error sending room key request cancellation;"
                                 + " will retry later.", e,
                             );
-                            this._startTimer();
```
bwindels, line 305: Retry isn't needed anymore?

dbkr: Mmm, good point - they would be sent the next time the queue was sent which would be on the next sync, but I guess keeping it closer to the original behaviour is probably better for now.

PR #1349, opened 2020-04-24
Remove some dead code from _retryDecryption

dbkr:
I think this was attempting to remove the events from _pendingEvents
but a) it wasn't and b) it probably shouldn't be because the retry
itself will re-add them when the decryption attempt starts and remove
them if it succeeds.

Also fix what was presumably a c+p fail.


dbkr:
No particular need to go to release

PR #1351, opened 2020-04-28
Added a progressCallback for backup key loading

foldleft:
Relates to: https://github.com/vector-im/riot-web/issues/13167

If we're handed a `progressCallback`, call it during the key loading process.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -1436,15 +1436,16 @@ MatrixClient.prototype.exportRoomKeys = function() {
  * Import a list of room keys previously exported by exportRoomKeys
  *
  * @param {Object[]} keys a list of session export objects
+ * @param {Object} opts
  *
  * @return {Promise} a promise which resolves when the keys
  *    have been imported
  */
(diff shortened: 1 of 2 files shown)


src/client.js
```diff
@@ -1436,15 +1436,16 @@ MatrixClient.prototype.exportRoomKeys = function() {
  * Import a list of room keys previously exported by exportRoomKeys
  *
  * @param {Object[]} keys a list of session export objects
+ * @param {Object} opts
```
dbkr, line 1439: We've been using this style: https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/base-apis.js#L51 to document parameters passed in objects

foldleft: done

foldleft:
> Looks good otherwise - maybe in the future we can somehow make this a more generic thing so the app needs less knowledge of what strings to look for in the progress callbacks, etc (if we can make it work with i18n).

mm, but i18n exists in the react layer so i'd rather not

PR #1353, opened 2020-04-28
Remove redundant key backup setup path

jryans:
Bootstrap was creating key backup versions twice due to some code duplication we missed in
previous refactorings.

Fixes https://github.com/vector-im/riot-web/issues/13423


uhoreg:
I think the copy from line ~600 is the one that should be removed, because this should run in all cases.

jryans:
@uhoreg Okay, I've updated as you suggest, I wasn't sure which would be best.

PR #1354, opened 2020-04-28
Add crypto.willUpdateDevices event and make getStoredDevices/getStoredDevicesForUser synchronous

dbkr:
Add an event fired before devices are updated, so apps can monitor
changes in devices by observing the state before and after.

BREAKING CHANGE:
Make getStoredDevices/getStoredDevicesForUser synchronous so they
can safely be called from the event listener without racing with
the data being updated. They didn't do any async work so didn't need
to be async.

Add doc for crypto.devicesUpdated

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -737,6 +737,7 @@ MatrixClient.prototype.initCrypto = async function() {
         "crypto.roomKeyRequestCancellation",
         "crypto.warning",
         "crypto.devicesUpdated",
+        "crypto.willUpdateDevices",
         "deviceVerificationChanged",
         "userTrustStatusChanged",
         "crossSigning.keysChanged",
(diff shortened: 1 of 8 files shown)


src/client.js
```diff
@@ -840,9 +841,9 @@ MatrixClient.prototype.getStoredDevicesForUser = async function(userId) {
  * @param {string} userId the user to list keys for.
  * @param {string} deviceId unique identifier for the device
  *
- * @return {Promise<?module:crypto/deviceinfo>} device or null
+ * @return {module:crypto/deviceinfo} device or null
  */
-MatrixClient.prototype.getStoredDevice = async function(userId, deviceId) {
```
jryans, line 845: There are various places that `await` these that no longer need to do so in JS SDK... Did you want to clean those up?

dbkr: ah yes - think I've got them all now

PR #1363, opened 2020-04-30
Return null for key backup state when we haven't checked yet

jryans:
This allows distinguishing the "unknown" case from "not present".

Part of https://github.com/vector-im/riot-web/issues/13404
Used by https://github.com/matrix-org/matrix-react-sdk/pull/4534

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -1482,8 +1482,9 @@ Crypto.prototype._checkAndStartKeyBackup = async function() {
         backupInfo = await this._baseApis.getKeyBackupVersion();
     } catch (e) {
         logger.log("Error checking for active key backup", e);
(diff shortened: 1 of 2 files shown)


src/crypto/index.js
```diff
@@ -1482,8 +1482,9 @@ Crypto.prototype._checkAndStartKeyBackup = async function() {
         backupInfo = await this._baseApis.getKeyBackupVersion();
     } catch (e) {
         logger.log("Error checking for active key backup", e);
-        if (e.httpStatus / 100 === 4) {
-            // well that's told us. we won't try again.
+        if (e.httpStatus === 404) {
```
foldleft, line 1485: `e.httpStatus / 100 === 4` dear god

PR #1364, opened 2020-05-01
Emit event when a trusted self-key is stored

dbkr:
This will cause our own user trust status to change, so emit an
event to say so.

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -2027,6 +2027,8 @@ Crypto.prototype.setDeviceVerification = async function(
 
         if (!this._crossSigningInfo.getId() && userId === this._crossSigningInfo.userId) {
             this._storeTrustedSelfKeys(xsk.keys);
+            // This will cause our own user trust to change, so emit the event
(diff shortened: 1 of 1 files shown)


src/crypto/index.js
```diff
@@ -2027,6 +2027,10 @@ Crypto.prototype.setDeviceVerification = async function(
 
         if (!this._crossSigningInfo.getId() && userId === this._crossSigningInfo.userId) {
             this._storeTrustedSelfKeys(xsk.keys);
+            // This will cause our own user trust to change, so emit the event
+            this.emit(
+                "userTrustStatusChanged", this._userId, this.checkUserTrust(userId),
```
jryans, line 2032: Potentially worth filing a clean-up issue to simplify some of this event flow...? `userTrustStatusChanged` and `crossSigning.keysChanged` seem to occur in most of the same places, but not all the same ones, and have emitters duplicated in many places... Anyway, just a future thought.

dbkr: Yeah, I got somewhat confused trying to get my head around it. Filed https://github.com/vector-im/riot-web/issues/13472

PR #1370, opened 2020-05-08
Sanity checks for 4S/cross-sign bootstrapping

bwindels:
Needed for https://github.com/matrix-org/matrix-react-sdk/pull/4563

### DIFF
--- a/src/crypto/CrossSigning.js
+++ b/src/crypto/CrossSigning.js
@@ -32,6 +32,20 @@ function publicKeyFromKeyInfo(keyInfo) {
     return Object.values(keyInfo.keys)[0];
 }
 
+function validateKey(privkey, expectedPubkey) {
+    if (!privkey) return;
+    const signing = new global.Olm.PkSigning();
+    try {
+        const gotPubkey = signing.init_with_seed(privkey);
+        if (gotPubkey === expectedPubkey) {
+            return true;
+        }
+    } finally {
+        signing.free();
+    }
+    return false;
+}
+
 export class CrossSigningInfo extends EventEmitter {
     /**
      * Information about a user's cross-signing keys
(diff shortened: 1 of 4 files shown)


src/crypto/CrossSigning.js
```diff
@@ -63,6 +78,90 @@ export class CrossSigningInfo extends EventEmitter {
...
+                const privkey =
+                    await this._callbacks.getCrossSigningKey(keyType, expectedPubkey);
+                // first, make sure that the published key matches
+                // the private key stored in 4S
+                if (!validateKey(privkey, expectedPubkey)) {
+                    console.log("checkCrossSignKeysIn4SMatchPublishedKeys: " +
+                        "cross-signing public key mismatch", keyType);
+                    return false;
+                }
+                // the private key here is provided by a callback, and it doesn't feel
```
bwindels, line 98: @uhoreg wdyt, is the rest here needed?

uhoreg: 4S should be safe now that we're using a symmetric algorithm, since only someone who knows the secret key can store to it now.

bwindels: Oh, I didn't mean so much whether we can trust the 4S key here. I probably should have given some more context, the checks here are to see whether the bootstrapping finished completely or was left in a broken state due to network failures/cancelling during bootstrapping.

The first check (above this line) should be to verify that the published public key matches the private key we get back from the `_callbacks.getCrossSigningKey` callback.

The second check, below, checks that the 4S account data for the cross-signing key is what we expect it to be, e.g. matches the private key.

It's the second check I was not sure is necessary. I've since checked and indeed `_callbacks.getCrossSigningKey` reads from the 4S account data, so it is not be possible for the account data to be different than the private key returned from there.

What would make sense is to compare the private key from 4S cache to the one returned from `getCrossSigningKey`, as the cache is written immediately, before the PUT account data request finishes.

dbkr: seems like it could be useful for sanity though, especially if we get to the point where the app provides keys from OS-level secure storage and we want to make sure they're in SSSS too.

bwindels:
Not merging this as we won't do the sanity checks and rather focus on fixing it for future bootstrappers. The code will still come in handy there though.

PR #1373, opened 2020-05-12
Remove support for asymmetric 4S encryption

bwindels:
Removes support for Curve25519 encryption of SSSS secrets. Also removes signing of the SSSS key by the cross-signing master key, as that is not needed for AES encryption.

This is a **breaking change**, as it removes the `secretStorageKeyNeedsUpgrade` method on `MatrixClient`.

Part of https://github.com/vector-im/riot-web/issues/13581
Required by https://github.com/matrix-org/matrix-react-sdk/pull/4581


src/crypto/SecretStorage.js
```diff
@@ -80,38 +75,29 @@ export class SecretStorage extends EventEmitter {
...
      * @param {object} opts the options for the algorithm.  The properties used
      *     depend on the algorithm given.
      * @param {string} [keyId] the ID of the key.  If not given, a random
      *     ID will be generated.
      *
      * @return {string} the ID of the key
      */
-    async addKey(algorithm, opts, keyId) {
-        const keyData = {algorithm};
+    async addKey(opts, keyId) {
```
uhoreg, line 85: Should probably keep the `algorithm` parameter as it may be needed in the future if we add other algorithms.

src/client.js
```diff
@@ -1310,7 +1310,6 @@ wrapCryptoFuncs(MatrixClient, [
     "bootstrapSecretStorage",
     "addSecretStorageKey",
     "hasSecretStorageKey",
-    "secretStorageKeyNeedsUpgrade",
```
uhoreg, line 1313: Maybe should keep this in order to not break backwards compat, but just have it return `false`.

bwindels: I'd rather remove it and mark this as a breaking change tbh. Will do so in the PR description :+1:

bwindels:
Sorry, had to force-push because I pushed a WIP commit for the next phase by accident. The only thing I changed was bringing back the algorithm parameter.

src/client.js
jryans: About breaking changes, we've been marking 4S / cross-signing APIs in `client.js` with the text "The Secure Secret Storage API is currently UNSTABLE and may change without notice.", so I think that may be enough wiggle room to make breaking changes like this without a major version bump.

jryans: Filed https://github.com/vector-im/riot-web/issues/13651 to track properly stabilising these APIs.

PR #1378, opened 2020-05-15
make MAC check robust against unpadded vs padded base64 differences

uhoreg:


src/crypto/SecretStorage.js
```diff
@@ -167,7 +167,7 @@ export class SecretStorage extends EventEmitter {
         if (info.algorithm === SECRET_STORAGE_ALGORITHM_V1_AES) {
             if (info.mac) {
                 const {mac} = await SecretStorage._calculateKeyCheck(key, info.iv);
-                return info.mac === mac;
+                return info.mac.replace(/=+$/g, '') === mac.replace(/=+$/g, '');
```
t3chguy, line 170: is info mac one mac or multiple, one per line? just wondering about the point of the g modifier

uhoreg: Just one.  The g probably isn't needed (but shouldn't be harmful either).  It was just copied from `encodeUnpaddedBase64` in `src/crypto/olmlib.js`

PR #1379, opened 2020-05-20
Combine the two places we processed client-level default push rules

dbkr:
We did this in two places: updated existing rules when the rules
were stored and then added new ones every time we computed push
action for an event. Just do both when the rules are saved.

The two different methods appeared in https://github.com/matrix-org/matrix-js-sdk/pull/1006 and https://github.com/matrix-org/matrix-js-sdk/pull/860

### DIFF
--- a/src/pushprocessor.js
+++ b/src/pushprocessor.js
@@ -364,33 +364,6 @@ export function PushProcessor(client) {
         return actionObj;
     };
 
-    const applyRuleDefaults = function(clientRuleset) {
-        // Deep clone the object before we mutate it
-        const ruleset = JSON.parse(JSON.stringify(clientRuleset));
-
-        if (!clientRuleset['global']) {
-            clientRuleset['global'] = {};
-        }
-        if (!clientRuleset['global']['override']) {
-            clientRuleset['global']['override'] = [];
-        }
-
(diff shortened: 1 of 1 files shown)


dbkr:
Hmm - I'm not sure I understand: I still don't get why two different places - if we need to fix up the push rules each time then wouldn't we need to do all the fixups? Otherwise they'll just be half wrong.

Also, we fetch the push rules from the server on each startup (https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/sync.js#L637) to work around the fact that the server doesn't push new default rules. But, even without that, I think the rules will still get fixed up when we load them in from the sync accumulator: https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/sync.js#L1035 so I don't understand in what situation they'd be wrong. The fixups we're applying are in the code so they can't change without a restart of the app, at which point we'd do both of the above.

turt2live:
The code itself is fine to combine (and should be). The concern is that the thing in the sync loop which calls for the push rules is now more important, although on second thought it's probably fine.

Updating the comment on the constant is all that's needed, as the object isn't applied on each event.

dbkr:
Ah, I see - yes. Hopefully this is better.

PR #1385, opened 2020-05-26
Fix replying to key share requests

Sorunome:
This check was meant to filter out key sharing requests from ourselves. Accidentally, it filtered out sharing requests from anyone *but* ourselves.

Signed-off-by: Sorunome <mail@sorunome.de>


t3chguy:
Needs DCO sign-off as per Contributing doc

Sorunome:
Updated first comment with sign-off 👍

PR #1386, opened 2020-05-27
Code for compositionally faking up data

foldleft:
I'm really not sure where to put this in the codebase, if anywhere

### DIFF
--- a/src/faking.ts
+++ b/src/faking.ts
@@ -78,12 +78,12 @@ class State<T, V> {
  * Using the state monad here allows us to write random object generators.
  * We can build them up combinatorially while threading the state of
  * the RNG through stuff.
- * 
+ *
  * A state monad is basically a big, composed function that says "when you get this input
  * state, do these things and return the new state and result".  The magic is that we can
  * put them together, with helper functions that mutate the state, without ever having to
  * think about threading the state around again.
- * 
+ *
  * Normal, healthy programmers would probably just make it a global.
  */
 
--- a/src/faking.ts
+++ b/src/faking.ts
@@ -1,0 +1,252 @@
+/*
+ * Utils for generating fake data
+ *
+ * Not very complete. For now I'm injecting rooms into a live client like this:
+ * mxMatrixClientPeg.get()._syncApi._processSyncResponse({}, {
+ *   rooms: { join: Faking.randomly(Faking.Rooms(100)) }
+ * })
+ */
+
+/*
+ * Irritatingly, Javascript does not provide a seedable RNG.  A seedable RNG is very
+ * helpful when you want reproduceable results.
+ */
+interface Xorshift128pState {
+    a: bigint,
+    b: bigint,
+}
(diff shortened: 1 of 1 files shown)


src/faking.ts
```diff
@@ -0,0 +1,275 @@
+/*
```
turt2live, line 1: needs copyright

src/faking.ts
```diff
@@ -0,0 +1,275 @@
...
+
+/*
+ * Irritatingly, Javascript does not provide a seedable RNG.  A seedable RNG is very
+ * helpful when you want reproduceable results.
+ */
+interface Xorshift128pState {
+    a: bigint,
+    b: bigint,
+}
+const MAX = (BigInt(1) << BigInt(64)) - BigInt(1);
```
turt2live, line 18: Isn't this the same as Number.MAX_INT or whatever?

src/faking.ts
```diff
@@ -0,0 +1,275 @@
...
+export function xorshift128p(s: Xorshift128pState): { state: Xorshift128pState, value: number } {
+    let t: bigint = s.a;
+    t ^= (t << BigInt(23)) & MAX;
+    t ^= (t >> BigInt(17));
+    t ^= s.b ^ (s.b >> BigInt(26));
+    const state = { a: s.b, b: t };
+    const value = Number((t + s.a) & MAX) / Number(MAX);
+    return { state, value };
+}
```
turt2live, line 27: This needs comments, and probably documenting as a seedable RNG instead of some complicated math function.

src/faking.ts
```diff
@@ -0,0 +1,275 @@
...
+ * state, do these things and return the new state and result".  The magic is that we can
+ * put them together, with helper functions that mutate the state, without ever having to
+ * think about threading the state around again.
+ *
+ * Normal, healthy programmers would probably just make it a global.
+ */
+
+ /* Helpfully, `xorshit128p` already follows the { state, value } pair output format that
+  * our state monad defines, so the implemntation is trivial */
+const Random = new State<Xorshift128pState, number>((s: Xorshift128pState) => xorshift128p(s));
```
turt2live, line 92: we don't use UpperCamelCase for variable names. If they're supposed to be class representations, they'll need to be classes.

src/faking.ts
```diff
@@ -0,0 +1,275 @@
...
+const Seed = State.put({a: BigInt(1), b: BigInt(1)})
+             .flatMap(() => State.sequence((new Array(20)).fill(Random)))
+             .map(() => undefined);
+
+/* A helper to execute a state monad using our predefined seed. */
+function randomly<T, V>(s: State<undefined, V>) {
+    return Seed.flatMap(() => s).runState(undefined).value;
+}
+
+(window as unknown as any).Faking = {
```
turt2live, line 258:
```suggestion
(<any>window).Faking = {
```

jryans:
I believe the author is no longer engaged, so it makes sense to close this for now.

PR #1388, opened 2020-05-29
Fix verification request timeouts to match spec

t3chguy:
Fixes https://github.com/vector-im/riot-web/issues/12430

Implements https://matrix.org/docs/spec/client_server/latest#key-verification-framework
> The prompt for Bob to accept/reject Alice's request (or the unsupported method prompt) should be automatically dismissed 10 minutes after the timestamp field or 2 minutes after Bob's client receives the message, whichever comes first, if Bob does not interact with the prompt. The prompt should additionally be hidden if an appropriate m.key.verification.cancel


t3chguy:
I think there is still some unexpected funkiness with `observeOnly` which is tripping up the tests

PR #1394, opened 2020-06-01
Add js-sdk mechanism for polling client well-known for config

t3chguy:


src/client.js
```diff
@@ -4800,6 +4804,29 @@ MatrixClient.prototype.startClient = async function(opts) {
     this._clientOpts = opts;
     this._syncApi = new SyncApi(this, opts);
     this._syncApi.sync();
+
+    if (opts.clientWellKnownPollPeriod !== undefined) {
+        this._clientWellKnownInterval =
```
turt2live, line 4809: we should probably call this a `timerId` of sorts, as it sounds like a number (even though timer IDs are numbers...)

PR #1395, opened 2020-06-02
Batch up URL previews to prevent excessive requests

turt2live:
Cache expiration is still not implemented here.

image

(the failures are because I've failed to configure a media repo correctly)

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -3000,6 +3000,9 @@ MatrixClient.prototype.getUrlPreview = function(url, ts, callback) {
     // If there's already a request in flight (or we've handled it), return that instead.
     const cachedPreview = this.urlPreviewCache[key];
     if (cachedPreview) {
+        if (callback) {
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -2991,25 +2991,29 @@ MatrixClient.prototype.setRoomReadMarkers = async function(
...
+
     const key = ts + "_" + url;
-    const og = this.urlPreviewCache[key];
-    if (og) {
-        return Promise.resolve(og);
+
+    // If there's already a request in flight (or we've handled it), return that instead.
+    const cachedPreview = this.urlPreviewCache[key];
+    if (cachedPreview) {
+        return cachedPreview;
```
t3chguy, line 3003: this path never calls the `callback`

turt2live: it didn't before either tbh

t3chguy: fair point

PR #1398, opened 2020-06-05
Bring back backup key format migration

bwindels:
Introduced in https://github.com/matrix-org/matrix-js-sdk/pull/1311 and removed in https://github.com/matrix-org/matrix-js-sdk/pull/1375 because thought not needed anymore.

Turns out this migration only seems to run on new logins or when restoring a backup, things users (like myself) might not have done since the format bug was introduced, so this can still cause problems for users trying out cross-signing on develop.

An alternative would be to tell all users that have this problem and tried cross-signing on develop to reset their key backup and 4S.


bwindels:
Tests time out for some reason :man_shrugging:

PR #1403, opened 2020-06-12
Use existing session id for fetching flows as to not get a new session

t3chguy:
Fixes https://github.com/vector-im/riot-web/issues/13990

Regression caused by https://github.com/matrix-org/matrix-js-sdk/commit/bebeec7d84295a994036502974e76f3327ef842f

### DIFF
--- a/src/interactive-auth.js
+++ b/src/interactive-auth.js
@@ -148,8 +148,14 @@ InteractiveAuth.prototype = {
             // if we have no flows, try a request to acquire the flows
             if (!hasFlows) {
                 if (this._busyChangedCallback) this._busyChangedCallback(true);
-                // Do a fresh request as we're just acquiring flows.
-                this._doRequest(null).finally(() => {
+                // use the existing sessionid, if one is present.
+                let auth = null;
+                if (this._data.session) {
+                    auth = {
+                        session: this._data.session,
+                    };
+                }
+                this._doRequest(auth).finally(() => {
                     if (this._busyChangedCallback) this._busyChangedCallback(false);
                 });
             } else {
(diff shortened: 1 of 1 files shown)


clokep:
> hmm, I'm fairly sure we're supposed to know the flows by here. Would be good to test with a homeserver like conduit as Synapse doesn't validate the session correctly.

Would love to have more info on what the issues are with how Synapse validates the session! We've put a bunch of work into this recently, so if there are other issues it'd be nice to fix it while the info is still in someone's head!

t3chguy:
> hmm, I'm fairly sure we're supposed to know the flows by here

Not if we come in from an email verification apparently as we get the session id from the URL and <InteractiveAuth calls this method after passing that ID to the ctor but it gets thrown away by us using null

turt2live:
> Would love to have more info on what the issues are with how Synapse validates the session! We've put a bunch of work into this recently, so if there are other issues it'd be nice to fix it while the info is still in someone's head!

@clokep I believe it came down to spec compliance last I looked: Synapse happily accepted `{}` as an auth dict even when it's not supposed to, and likewise somehow manages to track different session IDs across the same request for the same user (as under some conditions riot-web loses its session ID and just gets a new one in the middle of the flow).

clokep:
> likewise somehow manages to track different session IDs across the same request for the same user

It definitely doesn't do this. There are some hacks added though for updating sessions and such through the UI Auth flow, which are all...not really speced. :(

t3chguy:
Registering on a Conduit worked but then again it didn't have many flows (no email or consent) so maybe coincidental.

PR #1404, opened 2020-06-13
Remove support for unspecced device-specific push rules

t3chguy:
Part of https://github.com/matrix-org/matrix-doc/issues/2616
Fixes https://github.com/vector-im/riot-web/issues/14000

Specifically line 328 was to blame for 14000.


turt2live:
The tests take forever on this, implying we depend on this somewhere.

PR #1405, opened 2020-06-15
Add method to get current in-flight to-device requests

dbkr:
Hopefully the jsdoc explains all.

### DIFF
--- a/src/crypto/verification/request/ToDeviceChannel.js
+++ b/src/crypto/verification/request/ToDeviceChannel.js
@@ -358,9 +358,9 @@ export class ToDeviceRequests {
     }
 
     getRequestsInProgress(userId) {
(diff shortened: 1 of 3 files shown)


src/crypto/verification/request/ToDeviceChannel.js
```diff
@@ -356,4 +356,12 @@ export class ToDeviceRequests {
             }
         }
     }
+
+    getRequestsInProgress(userId) {
+        const requests = this._requestsByUserId.get(userId);
```
jryans, line 361: Maybe name this `requestsByTxnId` like other functions do? It's not obvious from reading this one by itself that there are two level of maps flying around here.

dbkr: mm, good point

PR #1406, opened 2020-06-15
Don't trust keys megolm received from backup for verifying the sender

uhoreg:
also, if we received a key from backup, don't automatically re-upload it to the backup

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -1978,7 +1978,11 @@ MatrixClient.prototype._restoreKeyBackup = function(
             }
         }
 
-        return this.importRoomKeys(keys, { progressCallback, untrusted: true, source: "backup" });
+        return this.importRoomKeys(keys, {
+            progressCallback,
+            untrusted: true,
+            source: "backup",
+        });
     }).then(() => {
         return this._crypto.setTrustedBackupPubKey(backupPubKey);
     }).then(() => {
--- a/src/client.js
+++ b/src/client.js
@@ -1193,13 +1193,15 @@ MatrixClient.prototype.checkEventSenderTrust = async function(event) {
  * @param {module:models/event.MatrixEvent} event event to be checked
  *
  * @return {object} An object with the fields:
- *    - encrypted: whether the event is encrypted
+ *    - encrypted: whether the event is encrypted (if not encrypted, some of the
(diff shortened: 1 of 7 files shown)


src/models/event.js
```diff
@@ -689,6 +694,10 @@ utils.extend(MatrixEvent.prototype, {
         return this._forwardingCurve25519KeyChain;
     },
 
+    isUntrusted: function() {
```
jryans, line 697: At the backup session level, the general word "untrusted" is perhaps fairly clear, but at the event level, I am worried that are too many different kinds of trust that we should be a bit more specific... Also, is it possible to write this in terms of a positive word (to avoid double negatives)? Perhaps `isKeySourceTrusted`?

nadonomy:
> If I understand correctly, Design would like to also tweak shield handling as part of this, so I'll clear the review for now. Please let me know if I've misunderstood.

The expectation is that until we fix the core issue, the messages which fall foul to this will be decorated with a black shield, with hover copy of:

> End-to-end encrypted. The authenticity of the sender of this message can't be guaranteed on this device.

src/crypto/index.js
```diff
@@ -2281,6 +2286,77 @@ Crypto.prototype.getEventSenderDeviceInfo = function(event) {
...
+    const forwardingChain = event.getForwardingCurve25519KeyChain();
+    if (forwardingChain.length > 0 || event.isKeySourceUntrusted()) {
+        // we got the key this event from somewhere else
+        // TODO: check if we can trust the forwarders.
+        ret.authenticated = false;
+    } else {
+        ret.authenticated = true;
+    }
+
+
```
jryans, line 2326: Maybe remove one of these blank lines.

src/client.js
```diff
@@ -1185,6 +1185,25 @@ MatrixClient.prototype.checkEventSenderTrust = async function(event) {
     return await this._crypto.checkDeviceTrust(event.getSender(), device.deviceId);
 };
 
+/**
+ * Get information about the encryption of an event
```
jryans, line 1189: Should `checkEventSenderTrust` be removed here? The React SDK PR removes the only use of it, and it was marked unstable for cross-signing, so could be removed safely I think.

PR #1416, opened 2020-06-29
Fix ICreateClientOpts missing idBaseUrl

jryans:
Port https://github.com/matrix-org/matrix-js-sdk/pull/1413 to release


jryans:
Since there are still other bits to grab, and time is low, let's leave this on develop for now.

PR #1419, opened 2020-07-09
[BREAKING] Convert RoomState's stored state map to a real map

turt2live:
**Requires https://github.com/matrix-org/matrix-react-sdk/pull/4936**

**We probably do not want to land this until after an RC as it is very high risk.**

Though the dictionary format works fine, it's slow. Access times are around the 1ms range for rooms like HQ, which doesn't seem like much, but when compared to the Map's access time of 0.05ms it's slow.

Converting things to a map means we lose index access and have to instead use `.keys()` and `.values()`, both of which return iterables and not arrays. Even doing the `Array.from()` conversion we see times in the 0.05ms range. This is also what makes this a breaking change.

Memory-wise there does not appear to be any measurable impact for a large account.

### DIFF
--- a/src/models/room-state.js
+++ b/src/models/room-state.js
@@ -68,7 +68,7 @@ export function RoomState(roomId, oobMemberFlags = undefined) {
     this.members = {
         // userId: RoomMember
     };
-    this.events = new Map(); // { eventType: { stateKey: MatrixEvent } }
+    this.events = new Map(); // Map<eventType, Map<stateKey, MatrixEvent>>
     this.paginationToken = null;
 
     this._sentinels = {
--- a/src/models/room-state.js
+++ b/src/models/room-state.js
@@ -68,9 +68,7 @@ export function RoomState(roomId, oobMemberFlags = undefined) {
     this.members = {
         // userId: RoomMember
     };
-    this.events = {
-        // eventType: { stateKey: MatrixEvent }
-    };
+    this.events = new Map(); // { eventType: { stateKey: MatrixEvent } }
     this.paginationToken = null;
 
     this._sentinels = {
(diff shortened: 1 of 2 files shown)


turt2live:
As an interesting bit of testing: `Object.values` is relatively `O(n)` with its call time with fluctuations of +/-0.4ms on the exact same data set. Our very own `utils.values` is also `O(n)` but more stable at a +/-0.05ms variation with the same data set. Our implementation is also consistently faster at ~0.87ms for a 5.1k data set whereas `Object.values` takes ~1.1ms

jryans:
> Our very own `utils.values` is also `O(n)` but more stable at a +/-0.05ms variation with the same data set.

That's quite surprising...! 😱 Definitely would have expected `Object.values` to be optimised by JS engines more than a manual version, but perhaps they do some checks we are able to assume and skip.

jryans:
@turt2live Which part is so high risk about this...? The late-breaking change to APIs?

turt2live:
It's high risk because I can't guarantee that I've caught all the places we use the public property. A combination of being 2016-era JS and generically named means there's huge potential that a reference was missed.

src/models/room-state.js
```diff
@@ -68,9 +68,7 @@ export function RoomState(roomId, oobMemberFlags = undefined) {
     this.members = {
         // userId: RoomMember
     };
-    this.events = {
-        // eventType: { stateKey: MatrixEvent }
-    };
+    this.events = new Map(); // { eventType: { stateKey: MatrixEvent } }
```
t3chguy, line 71: Can we update this comment to mention that the outer map values are also maps and not dicts

PR #1420, opened 2020-07-15
removeHiddenChars strip combining diacritic marks

t3chguy:


src/utils.ts
```diff
@@ -662,12 +662,13 @@ export function removeHiddenChars(str: string): string {
 
 // Regex matching bunch of unicode control characters and otherwise misleading/invisible characters.
 // Includes:
+// combining diacritics which are barely visible in most fonts U+0300 - U+036F
 // various width spaces U+2000 - U+200D
 // LTR and RTL marks U+200E and U+200F
 // LTR/RTL and other directional formatting marks U+202A - U+202F
 // Combining characters U+0300 - U+036F
```
t3chguy, line 669: err I am blind

PR #1433, opened 2020-08-04
Filter out non-string display names

bwindels:


src/utils.ts
```diff
@@ -661,7 +661,9 @@ export function isNumber(value: any): boolean {
  * @return {string} a string with the hidden characters removed
  */
 export function removeHiddenChars(str: string): string {
-    return unhomoglyph(str.normalize('NFD').replace(removeHiddenCharsRegex, ''));
+    if (typeof str === "string") {
+        return unhomoglyph(str.normalize('NFD').replace(removeHiddenCharsRegex, ''));
+    }
```
t3chguy, line 666: This should probably return an empty string rather than undefined to maintain the method signature

PR #1436, opened 2020-08-04
Dehydrate and rehydrate devices

uhoreg:
depends on https://github.com/matrix-org/synapse/pull/7955

implements https://github.com/matrix-org/matrix-doc/pull/2697

relates to https://github.com/matrix-org/matrix-react-sdk/pull/5239


src/client.js
```diff
@@ -459,6 +460,115 @@ export function MatrixClient(opts) {
...
+            undefined,
+            "GET",
+            "/dehydrated_device",
+            undefined, undefined,
+            {
+                prefix: "/_matrix/client/unstable/org.matrix.msc2697.v2",
+            },
+        );
+    } catch (e) {
+        console.info("could not get dehydrated device", e);
```
jryans, line 493: JS SDK uses the `logger` object instead of direct `console` access, please change this and all others here.

(We should really add a lint rule for this...)

src/crypto/dehydration.ts
```diff
@@ -0,0 +1,251 @@
...
+                            this.key = decodeBase64(decrypted);
+                            this.keyInfo = keyInfo;
+                            const now = Date.now();
+                            const delay = Math.max(1, time + oneweek - now);
+                            this.timeoutId = global.setTimeout(
+                                this.dehydrateDevice.bind(this), delay,
+                            );
+                        }
+                    },
+                    DEHYDRATION_ALGORITHM,
```
jryans, line 72: Hmm, using the algorithm name to store and retrieve the key is a bit mysterious / unclear... It doesn't really say anything about dehydration in the algorithm. Should we use a different name for that key that clarifies its purpose?

src/crypto/dehydration.ts
```diff
@@ -0,0 +1,251 @@
...
+
+            console.log("Uploading account to server");
+            const dehydrateResult = await this.crypto._baseApis._http.authedRequest(
+                undefined,
+                "PUT",
+                "/dehydrated_device",
+                undefined,
+                {
+                    device_data: deviceData,
+                    // FIXME: initial device name?
```
jryans, line 174: I think we do need some kind of device name, or else do something special to handle it differently in the device list. It's quite perplexing to see a blank device in the device list. I'd suggest checking with Nad about this.

jryans (changes requested):
Overall, looking good, but a few questions to resolve.

src/crypto/dehydration.ts
uhoreg: Yes.  I'm not sure why I did this originally.

src/crypto/dehydration.ts
uhoreg: (Have punted this to the react-sdk layer)

jryans (approved):
Thanks, this side looks good to me! 😄

src/client.js
jryans: I opened https://github.com/matrix-org/matrix-js-sdk/pull/1489 to enforce this with lint rules, hopefully that will help us remember (as there were ~20 other uses of `console` to fix as well).

PR #1437, opened 2020-08-06
Store and request master cross-signing key

jryans:
This enables storing and requesting master cross-signing private key when verifying other devices or using your security key.

**Reviewer:** I suggest reviewing commit by commit.

Part of https://github.com/vector-im/element-web/issues/13896
Used by https://github.com/matrix-org/matrix-react-sdk/pull/5092

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -1221,6 +1221,20 @@ Crypto.prototype.checkOwnCrossSigningTrust = async function() {
     if (oldSelfSigningId !== newCrossSigning.getId("self_signing")) {
         logger.info("Got new self-signing key", newCrossSigning.getId("self_signing"));
 
+        // Try to cache the self-signing private key as a side-effect
+        let signing = null;
+        try {
+            const ret = await this._crossSigningInfo.getCrossSigningKey(
+                "self_signing", newCrossSigning.getId("self_signing"),
+            );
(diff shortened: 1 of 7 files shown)


src/crypto/index.js
```diff
@@ -1221,6 +1221,20 @@ Crypto.prototype.checkOwnCrossSigningTrust = async function() {
     if (oldSelfSigningId !== newCrossSigning.getId("self_signing")) {
         logger.info("Got new self-signing key", newCrossSigning.getId("self_signing"));
 
+        // Try to cache the self-signing private key as a side-effect
```
bwindels, line 1224: I'm wondering why we need to cache the SSK & USK PKs as part of this change? Or is this unrelated to MSK key requests?

jryans: I added after noticing that if you set up a new session via security phase, we were only retrieving the MSK and SSK. The USK was not retrieved since it's not actually used during login. By eagerly caching it here, we avoid another prompt when you go to verify a user, and also ensures we have the key stored to share with another device during verification.

jryans: The SSK portion here is actually unneeded, as it will be implicitly retrieved as part of `signDevice` just below, but it felt clearer to explicitly cache the key, so we now have one flow where all 3 keys are being cached.

PR #1442, opened 2020-08-18
Use SAS emoji data from matrix-doc (mk II)

turt2live:
Round 2 of https://github.com/matrix-org/matrix-js-sdk/pull/1440 - see that PR for details.

---


t3chguy:
This is incompatible with Rust Crypto, tackling a different way

PR #1444, opened 2020-08-19
Add method to check whether client .well-known has been fetched

jryans:
This allows clearly detecting whether we have _ever_ fetched .well-known at all.
Without this, it's hard to be sure whether the value is `undefined` because the
fetch has not been attempted yet or because an error occurred.

Part of https://github.com/vector-im/element-web/issues/14954
Used by https://github.com/matrix-org/matrix-react-sdk/pull/5130

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -357,6 +357,9 @@ export function MatrixClient(opts) {
 
     this._cachedCapabilities = null; // { capabilities: {}, lastUpdated: timestamp }
 
+    this._clientWellKnown = undefined;
+    this._clientWellKnownPromise = undefined;
+
     // The SDK doesn't really provide a clean way for events to recalculate the push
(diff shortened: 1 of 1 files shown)


dbkr (commented):
Hmm, I wonder if a method that returns a promise that resolves with the well known result or null once it's been fetched or failed to fetch, respectively, might be simpler?

This also seems fine though, although I would add a `this._clientWellKnownFetchAttempted = false;` to the constructor so it returns `false` rather than `undefined` (and maybe add similar for the other members too, eg. `this._clientWellKnown`, even though it's been baked into the API that it returns `undefined` rather than `null`...)

jryans:
@dbkr I have switched to a promise-based API, let me know how it looks now.

dbkr (approved):
Thanks!

PR #1445, opened 2020-08-19
Fix IndexedDB store creation example

florianjacob:
This seems like a simple oversight. However, I have no idea whether this is really correct, please verify this.

Signed-off-by: Florian Jacob <florian ÄT wolkenplanet <POINT> de>

### DIFF
--- a/src/store/indexeddb.js
+++ b/src/store/indexeddb.js
@@ -51,8 +51,8 @@ const WRITE_DELAY_MS = 1000 * 60 * 5; // once every 5 minutes
  * sync from the server is not required. This does not reduce memory usage as all
  * the data is eagerly fetched when <code>startup()</code> is called.
  * <pre>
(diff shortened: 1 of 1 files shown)


florianjacob:
Seems like the example is even more broken, and missing an `opts = { indexeddb: window.indexeddb }` or something similar.

jryans:
> Seems like the example is even more broken, and missing an `opts = { indexeddb: window.indexeddb }` or something similar.

Yes, looks like that's correct, at the moment `indexedDB: window.indexedDB` is expected to be passed as well (or an equivalent when running on NodeJS). @florianjacob, would you like to add this tweak to your PR?

florianjacob:
Tweak added. :+1:

jryans (approved):
Thanks for working on this, and sorry for the long delay... 😓

PR #1450, opened 2020-08-27
Untangle cross-signing and secret storage

jryans:
This untangles cross-signing and secret storage setup into separate path that
can be invoked independently. There is no functional change with this patch, but
instead this just separates one giant monster API into two.

Part of https://github.com/vector-im/element-web/issues/13895
Used by https://github.com/matrix-org/matrix-react-sdk/pull/5149

### DIFF
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -761,7 +761,7 @@ Crypto.prototype.bootstrapSecretStorage = async function({
     if (
         !this._baseApis._cryptoCallbacks.saveCrossSigningKeys &&
         await this.isCrossSigningReady() &&
-        !await this._crossSigningInfo.isStoredInSecretStorage(secretStorage)
+        (newKeyId || !await this._crossSigningInfo.isStoredInSecretStorage(secretStorage))
     ) {
         logger.log("Storing cross-signing private keys in secret storage");
         const crossSigningPrivateKeys =
--- a/src/crypto/index.js
+++ b/src/crypto/index.js
@@ -444,7 +444,7 @@ Crypto.prototype.isCrossSigningReady = async function() {
  * bootstrapSecretStorage() completes successfully, this function should
  * return true.
  *
- * The secret storage API is currently UNSTABLE and may change without notice.
+ * The Secure Secret Storage API is currently UNSTABLE and may change without notice.
  *
  * @return {bool} True if secret storage is ready to be used on this device
  */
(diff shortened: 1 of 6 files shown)


src/crypto/index.js
```diff
@@ -422,24 +420,173 @@ Crypto.prototype.createRecoveryKeyFromPassphrase = async function(password) {
...
+    const delegateCryptoCallbacks = this._baseApis._cryptoCallbacks;
+    const builder = new EncryptionSetupBuilder(
+        this._baseApis.store.accountData,
+        delegateCryptoCallbacks,
+    );
+
+    const crossSigningInfo = new CrossSigningInfo(
+        this._userId,
+        builder.crossSigningCallbacks,
+        builder.crossSigningCallbacks);
```
dbkr, line 503: With full awareness that the first comment I'm making on this somewhat complex PR is about formatting, it would be nice if it were consistent with the line above. :)

src/crypto/index.js
```diff
@@ -422,24 +420,173 @@ Crypto.prototype.createRecoveryKeyFromPassphrase = async function(password) {
...
+    if (publicKeysOnDevice && privateKeysInCache) {
+        logger.log(
+            "Cross-signing public keys trusted and private keys found locally",
+        );
+    } else if (!privateKeysExistSomewhere || setupNewCrossSigning) {
+        logger.log(
+            "Cross-signing private keys not found locally or in secret storage, " +
+            "creating new keys",
+        );
+        await resetCrossSigning();
```
dbkr, line 546: Not certain if this is any different to what we do currently, but if we call this at startup and it automatically resets cross-signing without user verification if it sees no keys either in cache or SSSS, this will mess with people's account if they've set up on mobile and then logged in on web, 'cos it will automatically reset their keys.

dbkr (commented):
otherwise looks very plausible!

src/crypto/index.js
jryans: Ah thanks, cleaned up!

src/crypto/index.js
jryans: It does match what `bootstrapSecretStorage` would have done previously: if the cross-signing private keys are not found stored somewhere (it used to check _only_ 4S, not the local cache as well), they are reset.

You are correct this is a potential trap if keys are created on mobile and device verification is skipped... We do not call these bootstrap methods at startup currently, instead they should only be happening as part of some user interaction, like via a toast to setup or verify.

In the next phase of work, we will start calling `bootstrapCrossSigning` at registration to create keys, but in that case we know we're the first device, so should be safe.

jryans: I added a comment about the potential trap you mentioned, but at least for now, we shouldn't be triggering that case.

dbkr: ah ok - that should be fine then.

dbkr (approved):
lgtm!

PR #1452, opened 2020-08-29
Add missing options in ICreateClientOpts

dalcde:
This makes the examples in the README actually compile.


jryans (approved):
Thanks, looks good to me! 😄

Issue #1458, opened 2020-09-04
Some tables aren't rendered properly in the JSDoc. 


MaHa-Nordeck:
Here's a page where the table goes out of bounds.
http://matrix-org.github.io/matrix-js-sdk/8.2.0/module-client.MatrixClient.html

PR #1459, opened 2020-09-04
Switch to a combination of better-docs and docdash

t3chguy:
image

Fixes https://github.com/matrix-org/matrix-js-sdk/issues/1458


turt2live (approved):
it's great how the largest part of this is the lockfile

PR #1462, opened 2020-09-09
extend method redactEvent with reason

orangecms:
This implements passing a reason for the redaction as per https://matrix.org/docs/spec/client_server/r0.6.1#id254

Disclaimer: @fridtjof, @mythozz and @badgateway666 made me do this :p

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -2844,14 +2844,19 @@ function _sendEventHttpRequest(client, event) {
  * @param {string} eventId
  * @param {string} [txnId]  transaction id. One will be made up if not
  *    supplied.
- * @param {module:client.callback} callback Optional.
+ * @param {object|module:client.callback} callbackOrOpts
+ *    Options to pass on, may contain `reason`.
+ *    Can be callback for backwards compatibility.
  * @return {Promise} Resolves: TODO
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
-MatrixClient.prototype.redactEvent = function(roomId, eventId, txnId, callback) {
+MatrixClient.prototype.redactEvent = function(roomId, eventId, txnId, callbackOrOpts) {
+    const opts = typeof(callbackOrOpts) === 'object' ? callbackOrOpts : {};
+    const reason = opts.reason;
+    const callback = typeof(callbackOrOpts) === 'function' ? callbackOrOpts : undefined;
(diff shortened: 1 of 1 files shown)


turt2live (changes requested):
Thanks for taking on the fix! We should find a way to get this into the existing function to avoid building out infinite problems. My suggestion would be to support multiple call signatures by switching on the `callback` argument's type. 

We also need sign-off on these changes.

orangecms:
Alright, that we can do :) Thanks!

src/client.js
```diff
@@ -2844,16 +2844,24 @@ function _sendEventHttpRequest(client, event) {
  * @param {string} eventId
  * @param {string} [txnId]  transaction id. One will be made up if not
  *    supplied.
+ * @param {string|module:client.callback} reason Optional. Can be callback for backwards compatibility.
  * @param {module:client.callback} callback Optional.
  * @return {Promise} Resolves: TODO
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
-MatrixClient.prototype.redactEvent = function(roomId, eventId, txnId, callback) {
+MatrixClient.prototype.redactEvent = function(roomId, eventId, txnId, reason, callback) {
```
turt2live, line 2852: this isn't backwards compatible :( 

I was trying to say something along the lines of:

```
function(roomId, eventId, txnId, reason, callbackOrOpts) {
  const opts = typeof(callbackOrOpts) === 'object' ? callbackOrOpts : {};
  const reason = opts['reason'];
  const callback = typeof(callbackOrOpts) === 'function' ? callbackOrOpts : undefined;
  // ... the rest of the function
}
```

turt2live (changes requested):
Thanks for taking a look at this, and apologies for the delay in review - it's been a bit busy over here

src/client.js
turt2live: we don't need to worry about supplying a callback as a 5th argument when `opts` is used because we're slowly trying to discourage people from using the callback.

orangecms: Got it, so object and deprecating callbacks it shall be - I'll adjust it when I get around to it again. :-)

turt2live (approved):
Apologies for the long review on this - it looks good to me. Thanks for taking it on!

PR #1463, opened 2020-09-09
Inject identity server token for 3pid invites on createRoom

turt2live:
We only inject if needed to avoid potentially overwriting the value with something unknown. The docs in the react-sdk imply this was supposed to happen.

Fixes https://github.com/vector-im/element-web/issues/14814


jryans (approved):
Seems reasonable and necessary, thanks! A bit surprised we've been without it this long... 🤔

PR #1466, opened 2020-09-14
some TypeScript and doc fixes

uhoreg:
tell TypeScript about olm, add all the options for clients, and make options in docs camelCase


t3chguy (approved):
Awesome!!

PR #1467, opened 2020-09-15
Add support for olm fallback keys

uhoreg:
Requires https://github.com/matrix-org/synapse/pull/8312 and https://gitlab.matrix.org/matrix-org/olm/-/merge_requests/13 (Should not be merged until a new olm release is made and the olm dependency is bumped.)

Implements [MSC2732](https://github.com/matrix-org/matrix-doc/pull/2732)


src/crypto/index.js
```diff
@@ -1865,27 +1873,31 @@ function _maybeUploadOneTimeKeys(crypto) {
...
+            const res = await _uploadOneTimeKeys(crypto);
             if (res.one_time_key_counts && res.one_time_key_counts.signed_curve25519) {
                 // if the response contains a more up to date value use this
                 // for the next loop
-                return uploadLoop(res.one_time_key_counts.signed_curve25519);
+                keyCount = res.one_time_key_counts.signed_curve25519;
             } else {
                 throw new Error("response for uploading keys does not contain "
-                              + "one_time_key_counts.signed_curve25519");
+                                + "one_time_key_counts.signed_curve25519");
```
jryans, line 1898: Maybe move the `+` to the previous line, so the quotes can line up with the line above.

src/sync.js
```diff
@@ -1361,6 +1361,15 @@ SyncApi.prototype._processSyncResponse = async function(
         const currentCount = data.device_one_time_keys_count.signed_curve25519 || 0;
         this.opts.crypto.updateOneTimeKeyCount(currentCount);
     }
+    if (this.opts.crypto && data.device_unused_fallback_keys) {
+        // The presence of device_unused_fallback_keys indicates that the
+        // server supports fallback keys. If there's no unused
+        // signed_curve25519 fallback key we need a new one.
+        this.opts.crypto.setNeedsNewFallback(
+            data.device_unused_fallback_keys instanceof Array
+                && !data.device_unused_fallback_keys.includes("signed_curve25519"),
```
jryans, line 1370: Maybe move `&&` to the previous line and then dedent this line to align with the one above.

src/sync.js
```diff
@@ -1361,6 +1361,15 @@ SyncApi.prototype._processSyncResponse = async function(
         const currentCount = data.device_one_time_keys_count.signed_curve25519 || 0;
         this.opts.crypto.updateOneTimeKeyCount(currentCount);
     }
+    if (this.opts.crypto && data.device_unused_fallback_keys) {
```
jryans, line 1364: Hmm, so a supporting server will list your fallback key in every sync response indefinitely? It would nice if there were some way for it to only appear when changed, rather than every time, but I'll leave that to the spec and server discussion.

jryans (approved):
Looks good overall!

src/sync.js
jryans: Added a [comment](https://github.com/matrix-org/matrix-doc/pull/2732#discussion_r489294662) to the MSC for this.

src/crypto/index.js
uhoreg: Not sure why this line even got changed.  :shrug:

uhoreg:
@jryans already approved this, but it has probably changed enough since then that it should be reapproved

jryans (approved):

PR #1469, opened 2020-09-15
Add option to allow users of pantialaimon to use the SDK

Half-Shot:
matrix-js-sdk will block attempts to send events to encrypted rooms by default if you've not enabled the crypto parts of the library. There are special reasons for which I need to use Pantalaimon as a proxy for some bridge work, so I need the SDK to not block attempts in this case, as the upstream library will take care of it.

I'd like to propose this flag to let developers opt to ignore such warnings. I also couldn't really decide upon a flag name, so suggestions welcome!

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -182,6 +182,10 @@ function keyFromRecoverySession(session, decryptionKey) {
  * Optional. Whether to allow a fallback ICE server should be used for negotiating a
  * WebRTC connection if the homeserver doesn't provide any servers. Defaults to false.
  *
+ * @param {boolean} [opts.usingE2EProxy]
+ * Optional. Whether to allow sending messages to encrypted rooms when encryption
+ * hasn't been enabled. This is useful if you are using a E2E proxy. Defaults to false.
+ *
  * @param {object} opts.cryptoCallbacks Optional. Callbacks for crypto and cross-signing.
  *     The cross-signing API is currently UNSTABLE and may change without notice.
  *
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -182,6 +182,10 @@ function keyFromRecoverySession(session, decryptionKey) {
  * Optional. Whether to allow a fallback ICE server should be used for negotiating a
  * WebRTC connection if the homeserver doesn't provide any servers. Defaults to false.
  *
+ * @param {boolean} [opts.usingE2EProxy]
```
jryans, line 185: Perhaps this is a bit more general?

```suggestion
 * @param {boolean} [opts.usingExternalCrypto]
```

src/client.js
```diff
@@ -182,6 +182,10 @@ function keyFromRecoverySession(session, decryptionKey) {
...
+ * Optional. Whether to allow sending messages to encrypted rooms when encryption
+ * hasn't been enabled. This is useful if you are using a E2E proxy. Defaults to false.
```
jryans, line 187:
```suggestion
 * is not available internally within this SDK. This is useful if you are using an external 
 * E2E proxy, for example. Defaults to false.
```

jryans (changes requested):
Thanks, looks generally okay to add, but I think it's better for naming to be more about what it means for the JS SDK instead of your specific scenario.

src/client.js
Half-Shot: Yeah, works for me.

jryans (approved):
Thanks, seems good to me! 😄

PR #1472, opened 2020-09-18
Repair secret storage reset, cache keys when missing

jryans:
This includes changes to avoid an error when resetting if secret storage if you only have some but not all cross-signing keys. It also changes the path for caching from secret storage so it will try to cache private keys whenever they are missing as well as on private key change.

Fixes https://github.com/vector-im/element-web/issues/15230


turt2live (approved):
seems plausible

PR #1478, opened 2020-09-24
Always push docs if they are generated

t3chguy:


JorikSchellekens:
Why would we want to push the docs for a release that might fail, in the unlikely event that it fails the final push?

jryans:
> Why would we want to push the docs for a release that might fail, in the unlikely event that it fails the final push?

By this point, the release has already been pushed to the GH release page and NPM as a new version. The main sources of failure left are the docs generation itself (but then we would not get as far as pushing), and occasional conflicts merging back to develop (but those will always need manual intervention).

So, I don't think this changes the risk profile much, and it has the advantage of sharing pre-release docs the same day we build them.

jryans (approved):
Looks good to me, thanks!

PR #1481, opened 2020-09-25
Only sign key backup with cross-signing keys when available

jryans:
This changes the key backup setup step to only sign with cross-signing keys when
both the public and private keys are already available without prompting. In
many cases down these paths, the cross-signing keys either may not exist or may
not be accessible. We always sign the key backup with your device key as well,
so there is always a route to trust the key backup even if this is skipped.

Fixes https://github.com/vector-im/element-web/issues/15230


turt2live (approved):
I don't know enough about the keys, but skipping things when they won't work seems reasonable

PR #1483, opened 2020-09-28
Ignore invalid candidates

dbkr:
Ignore ICE candidates with no m-line index or mid


jryans (approved):
Looks okay, but seems like test failures

PR #1484, opened 2020-09-28
Catch exception from call event handler

dbkr:
Otherwise they leak out ot the sync loop. It's unfortunate that
exceptions from event handlers leak out to the emitter, and this
seems a bit clumsy, but having to wrap eveery event emit in a try
block seems worse.


jryans (approved):
Seems reasonable, thanks!

PR #1487, opened 2020-09-30
Convert call.js to Typescript & update WebRTC APIs

dbkr:
Two large commits here: a78b590 which converts to typescript and cc4656b which updates to changes in the WebRTC API (mostly using promises instead of callbacks and not using `addStream()`.

This PR also should make the output setting a bit more reliable, as this appears to be quite sensitive on Chrome (it's still not implemented on Firefox).


src/webrtc/call.ts
```diff
@@ -84,32 +84,73 @@ enum CallParty {
     REMOTE = 'remote',
 }
 
+enum MediaQueueId {
+    REMOTE_VIDEO = 'remote_video',
+    REMOTE_AUDIO = 'remote_audio',
+    LOCAL_VIDEO = 'local_video',
```
t3chguy, line 90: I think our convention is PascalCasing for enum members

src/webrtc/call.ts
```diff
@@ -84,32 +84,73 @@ enum CallParty {
...
+}
+
 enum CallErrorCode {
     /** An error code when the local client failed to create an offer. */
-    ERR_LOCAL_OFFER_FAILED = "local_offer_failed",
+    LOCAL_OFFER_FAILED = 'local_offer_failed',
```
t3chguy, line 95: same re PascalCasing

src/webrtc/call.ts
```diff
@@ -399,8 +367,18 @@ export class MatrixCall extends EventEmitter {
      * @param {Element} element The <code>&lt;video&gt;</code> DOM element.
      */
     setRemoteVideoElement(element : HTMLVideoElement) {
```
t3chguy, line 368: conventionally we don't tend to have a space after the parameter name

src/webrtc/call.ts
```diff
@@ -409,27 +387,41 @@ export class MatrixCall extends EventEmitter {
...
+            return element.play();
+        });
     }
 
     /**
      * Configure this call from an invite event. Used by MatrixClient.
      * @param {MatrixEvent} event The m.call.invite event
      */
-    initWithInvite(event : any) {
+    async initWithInvite(event : any) {
```
t3chguy, line 407: can this get a type?

src/webrtc/call.ts
```diff
@@ -75,8 +75,8 @@ enum CallType {
 }
 
 enum CallDirection {
-    INBOUND,
-    OUTBOUND,
+    INBOUND = 'inbound',
+    OUTBOUND = 'outbound',
```
t3chguy, line 79: Same re PascalCasing

src/@types/global.d.ts
```diff
@@ -36,7 +36,8 @@ declare global {
     }
 
     interface HTMLAudioElement {
-        // setSinkId is experimental and typescript doesn't know about it
+        // sinkId & setSinkId are experimental and typescript doesn't know about it
```
t3chguy, line 39: them?

t3chguy (commented):
Looks sane otherwise

src/webrtc/call.ts
dbkr: These ones are part of API unfortunately so need to stay for backwards-compat

src/webrtc/call.ts
dbkr: do we have a type for events in the js-sdk yet?

src/webrtc/call.ts
dbkr: oh wait, you mean the enum names not the values

src/webrtc/call.ts
t3chguy: No but you can use `MatrixEvent` (imported) and it'll work magically as soon as that gets Typescripted, helps prevent future issues from cropping up during MatrixEvent being TSified

t3chguy (approved):

PR #1489, opened 2020-10-01
Enforce logger module via lint rules

jryans:
This adds lint rules (and fixes various errors) to ensure we use the `logger`
intermediary module, rather than accessing the console directly.


t3chguy (approved):
Awesome!

PR #1490, opened 2020-10-05
Keep local device after processing device list sync

jryans:
This change ensure we always keep the local device in the device list, even if
the server claims it doesn't exist. This is important for initial startup, as it
is possible the device key upload takes a while to apply, and thus might not
appear in the first device list sync.

Fixes https://github.com/vector-im/element-web/issues/15319


t3chguy (approved):
Looks sane!

PR #1491, opened 2020-10-06
Fix re-emit of Event.replaced to be on client and not room

t3chguy:
The one listener in react-sdk listens on the MatrixClient and it never gets re-emitted there.
It isn't very useful on Room as during app-launch we do not have the Room available to bind the re-emmission


turt2live (approved):
re-fire that room event off the room - it'll be fine :P

PR #1492, opened 2020-10-06
Fix editing local echoes not updating them in real time

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/10545

Not sure if it covers all the various states but works with a simulated slow connection in both plaintext and e2ee rooms

### DIFF
--- a/spec/browserify/sync-browserify.spec.js
+++ b/spec/browserify/sync-browserify.spec.js
@@ -99,5 +99,5 @@ describe("Browserify Test", function() {
                 client.once("sync.unexpectedError", reject);
             }),
         ]);
-    }, 20000);
+    }, 20000); // additional timeout as this test can take quite a while
 });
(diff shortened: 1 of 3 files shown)


spec/browserify/sync-browserify.spec.js
```diff
@@ -99,5 +99,5 @@ describe("Browserify Test", function() {
                 client.once("sync.unexpectedError", reject);
             }),
         ]);
-    }, 10000);
+    }, 20000);
```
turt2live, line 102: I know there wasn't one before, but a comment would help a lot.

src/models/room.js
```diff
@@ -373,6 +373,20 @@ Room.prototype.hasPendingEvent = function(eventId) {
     return this._pendingEventList.some(event => event.getId() === eventId);
 };
 
+/**
+ * Get a specific event from the pending event list, if configured, undefined otherwise.
```
turt2live, line 377: but it's null

turt2live (approved):
otherwise lgtm

PR #1494, opened 2020-10-06
Convert call.js to Typescript & update WebRTC APIs (re-apply)

dbkr:
Re-applies #1487 - merge after cutting release branch


t3chguy (approved):

src/webrtc/call.ts
```diff
@@ -0,0 +1,1299 @@
...
+ * @param {Object?} options DEPRECATED optional options map.
+ * @param {boolean} options.forceTURN DEPRECATED whether relay through TURN should be
+ * forced. This option is deprecated - use opts.forceTURN when creating the matrix client
+ * since it's only possible to set this option on outbound calls.
+ * @return {MatrixCall} the call or null if the browser doesn't support calling.
+ */
+export function createNewMatrixCall(client: any, roomId: string, options: CallOpts) {
+    // typeof prevents Node from erroring on an undefined reference
+    if (typeof(window) === 'undefined' || typeof(document) === 'undefined') {
+        logger.error("No window or document object: WebRTC is not supported in this environement");
```
dalcde, line 1267: This change means that when using `matrix-js-sdk` in a non-browser context (e.g. appservice bridges), we get an error every time a client is created, which pollutes the log when nothing really went wrong. Is it possible to lower the log level of this line?

PR #1495, opened 2020-10-06
Make an accurate version of 'age' for events

dbkr:
We've always had 'age' in events but it's never really been an
accurate representation of the event's age because we never did
anything with it. This transforms it into a local clock timestamp
when the event arives and when it comes out of the sync store, and
changes getLocalAge() to use it.

react-sdk doesn't appear to use getLocalAge() but any 3rd party apps
that do may notice a slight change in bahaviour.


turt2live (approved):
seems sensible, though a bit unfortunate that we're sending a flag through half the sync accumulator

PR #1502, opened 2020-10-09
Adjust types and APIs to match React SDK

jryans:
Various small tweaks and alignments to match React SDK as part of TypeScript
conversion.

Part of https://github.com/vector-im/element-web/issues/15350
Used by https://github.com/matrix-org/matrix-react-sdk/pull/5307

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -235,7 +235,14 @@ function keyFromRecoverySession(session, decryptionKey) {
  *       {
  *           keys: {
  *               <key name>: {
- *                   pubkey: {UInt8Array}
+ *                   "algorithm": "m.secret_storage.v1.aes-hmac-sha2",
+ *                   "passphrase": {
+ *                       "algorithm": "m.pbkdf2",
+ *                       "iterations": 500000,
+ *                       "salt": "..."
+ *                   },
+ *                   "iv": "...",
+ *                   "mac": "..."
  *               }, ...
  *           }
  *       }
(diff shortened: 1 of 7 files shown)


src/matrix.ts
```diff
@@ -136,26 +137,54 @@ interface ICreateClientOpts {
...
-        onSecretRequested?: (
-            name: string, userId: string, deviceId: string,
-            requestId: string, deviceTrust: any
-        ) => Promise<string>;
+    cryptoCallbacks?: ICryptoCallbacks;
+}
+
+export interface ICryptoCallbacks {
+    getCrossSigningKey?: (keyType: string, pubKey: Uint8Array) => Promise<Uint8Array>;
+    saveCrossSigningKeys?: (keys: Record<string, Uint8Array>) => unknown;
```
turt2live, line 151: presumably:
```suggestion
    saveCrossSigningKeys?: (keys: Record<string, Uint8Array>) => void;
```

src/matrix.ts
```diff
@@ -136,26 +137,54 @@ interface ICreateClientOpts {
...
+    shouldUpgradeDeviceVerifications?: (
+        users: Record<string, any>
+    ) => Promise<Array<string>>;
```
turt2live, line 154: we generally prefer `string[]` as the type over `Array<string>`

src/client.js
```diff
@@ -1449,7 +1457,9 @@ wrapCryptoFuncs(MatrixClient, [
  * @param {string} [keyName] the name of the key.  If not given, a random
  *     name will be generated.
  *
- * @return {string} the name of the key
+ * @return {object} An object with:
+ *     keyId: {string} the ID of the key
+ *     keyInfo: {object} details about the key (iv, mac, passphrase)
```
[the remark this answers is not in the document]
jryans: A good question, all of the cross-signing / secret storage APIs currently shout "UNSTABLE" in docs, so hopefully okay. Beyond that, I'd be quite surprised if anyone else is using these specific APIs based on their complexity. A quick GitHub search suggests no use so far publicly at least, so I think it's okay.

src/matrix.ts
jryans: Ah right, most of these are extracted from above, will tune.

PR #1508, opened 2020-10-13
Typescriptify some crypto stuffs

t3chguy:


t3chguy:
Blocking merge until after RC

uhoreg (approved):
Looks plausible.  Nothing too suprising.  I didn't review the megolm.ts and olm.ts changes thoroughly because it isn't showing a helpful diff, but if it was mostly a mechanical transformation, then it should be fine.

t3chguy:
Yeah, the only actual changes are the ones you made in that one commit

PR #1510, opened 2020-10-15
Support m.call.reject

dbkr:
Start the migration to v1 VoIP by supporting m.call.reject, which
we'll send if the caller says they're v1. Our version stays as v0
for now, until we speak the rest of v1.

Honour the default reaosn in a hangup being user_hangup.

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -164,6 +164,14 @@ export enum CallErrorCode {
     Replaced = 'replaced',
 }
 
+/**
+ * The version field that we set in m.call.* events
+ * Once we are able to speak v1 VoIP sufficiently, this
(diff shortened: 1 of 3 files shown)


src/webrtc/call.ts
```diff
@@ -164,6 +164,14 @@ export enum CallErrorCode {
     Replaced = 'replaced',
 }
 
+/**
+ * The version field that we set in m.call.* events
+ * Once we are able to speak v1 VoIP sufficiently, this
+ * bumped to 1. While we partially speak v1 VoIP, it remains
+ * as 0.
+ */
+const VOIP_PROTO_VERSION = 0;
```
turt2live, line 173: refactoring this to support strings sounds fun

turt2live (approved):
lgtm

PR #1512, opened 2020-10-16
Support party_id

### LINKED ISSUE #1511: Support party_id in VoIP events
As part of VoIP v1 support

dbkr:
Send party_id on events and check the party_id of incoming events matches

Includes a basic test to assert that it actually does: we should
build out a decent test suite for calls as there's a lot of edge-case
functionality that can break and slip through the cracks (eg. glare).
This is a start (although it would be nice if we could write the tests
in TypeScript).

Fixes #1511
Based on: https://github.com/matrix-org/matrix-js-sdk/pull/1510 (https://github.com/matrix-org/matrix-js-sdk/compare/pull/1510/head...pull/1512/head)

### DIFF
--- a/spec/TestClient.js
+++ b/spec/TestClient.js
@@ -69,6 +69,9 @@ export function TestClient(
 
     this.deviceKeys = null;
     this.oneTimeKeys = {};
+    this._callEventHandler = {
+        calls: new Map(),
+    };
 }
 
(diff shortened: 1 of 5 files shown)


turt2live (approved):
I assume this was meant to go up for review?

Issue #1504, opened 2020-10-11
Error: Cannot call hasPendingEvent with pendingEventOrdering == chronological


zhaytee:
**Describe the bug**
sendReadReceipt() fails on matrix-js-sdk 8.4.1 with the following error:
```
Error: Cannot call hasPendingEvent with pendingEventOrdering == chronological
    at Room.hasPendingEvent (/node_modules/matrix-js-sdk/lib/models/room.js:380:11)
    at MatrixClient.sendReadReceipt (/node_modules/matrix-js-sdk/lib/client.js:3153:20)
```

I'm not sure what pendingEventOrdering is. I have not encountered that field while creating my app with matrix-js-sdk, so the default value must be set (wherever it is).

**To Reproduce**
1. Create a basic client which joins a room and calls sendReadReceipt() on incoming room message events.
2. Have another client send a normal room message.

**Expected behavior**
sendReadReceipt() should be called without error.

**Versions**
NodeJS: 12.19.0
matrix-js-sdk: 8.4.1

**Additional context**
Thanks for taking a look! Please let me know if you need any more information.

PR #1517, opened 2020-10-19
[Breaking] Change hasPendingEvent to return false if pending ordering !detached

t3chguy:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/1504

It is called in 3 places which assumed it would not throw, IMO its a little bit strict for it to throw.

### DIFF
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -358,16 +358,14 @@ Room.prototype.getPendingEvents = function() {
 
 /**
  * Check whether the pending event list contains a given event by ID.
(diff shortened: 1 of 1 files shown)


turt2live (approved):
we may need to highlight this more on the changelog, though I'm honestly still surprised that people use this thing with non-detached ordering.

t3chguy:
We are said people :P

`sendReadReceipt` calls upon it

turt2live:
well damn.

PR #1520, opened 2020-10-20
Fix syntax error

onagurna:


onagurna:
Fix syntax error

t3chguy:
Neither of those is correct. The export is right as it was.

image

PR #1521, opened 2020-10-20
Don't cache failures when fetching /versions

dbkr:
Otherwise we never recover

Fixes https://github.com/vector-im/element-web/issues/15509


turt2live (approved):

jryans (approved):
Thanks, seems like a reasonable fix! 😄

enzingerm:
I just tested it, works and fixes vector-im/element-web#15509 for me!

PR #1522, opened 2020-10-21
Support m.call.select_answer

dbkr:
Only send one if the answer has a party_id set

Also change some event types to the enum and rename receivedAnswer
to be more consistent


t3chguy (approved):

dbkr:
sorry - some more stuff for the failing test

t3chguy (approved):

PR #1527, opened 2020-10-22
Convert logger to typescript

dbkr:
Because it's annoying for the IDE to have no idea what methods are
on the logger. `loglevel` has types so it should just pass them
through.

Requires https://github.com/matrix-org/matrix-js-sdk/pull/1525


t3chguy (approved):
Yay

PR #1531, opened 2020-10-28
Fix stuck never-sending messages

dbkr:
Another cached promise that wasn't cleared on failure


turt2live (approved):
The happy path is the only path.

PR #1532, opened 2020-10-29
Implement call holding functionality

dbkr:
Using m.call.negotiate

VoIP protocol version remains at zero so this won't be enabled unless talking to an element-web from the future (it will still mute, it just won't signal to the other side).

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -91,6 +91,9 @@ export enum CallEvent {
     State = 'state',
     Error = 'error',
     Replaced = 'replaced',
+
+    // The value of isLocalOnHold() has changed
+    HoldUnhold = 'hold_unhold',
 }
 
 enum MediaQueueId {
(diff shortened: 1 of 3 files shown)


src/webrtc/call.ts
```diff
@@ -236,7 +244,17 @@ export class MatrixCall extends EventEmitter {
     // The party ID of the other side: undefined if we haven't chosen a partner
     // yet, null if we have but they didn't send a party ID.
     private opponentPartyId: string;
-    private inviteTimeout;
+    private inviteTimeout: NodeJS.Timeout;
```
turt2live, line 247: is this actually a NodeJS timeout or a web browser timeout?

turt2live (approved):
looks like it does a thing with call holding, though I can't claim to be a voip expert

src/webrtc/call.ts
dbkr: here it's a browser one since webrtc is only in the browser, but in general in the js-sdk it could be either - I thought this is the type I'd seen used elsewhere though?

turt2live: it mostly depends on how we're clearing it - the node timeouts have a different contract than the browser ones, which can lead to runtime errors.

dbkr: have gone with the 'commenting it' option

turt2live (approved):

PR #1534, opened 2020-11-02
Client set profile methods update own user

t3chguy:
Synthesise profile updates for ourselves to work around issues when you are in 0 rooms.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -3702,10 +3702,16 @@ MatrixClient.prototype.setProfileInfo = function(info, data, callback) {
  * @return {Promise} Resolves: TODO
  * @return {module:http-api.MatrixError} Rejects: with an error response.
  */
-MatrixClient.prototype.setDisplayName = function(name, callback) {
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -3702,10 +3702,16 @@ MatrixClient.prototype.setProfileInfo = function(info, data, callback) {
  * @return {Promise} Resolves: TODO
```
turt2live, line 3702: let's fix these while we're in the area

src/client.js
```diff
@@ -3714,10 +3720,16 @@ MatrixClient.prototype.setDisplayName = function(name, callback) {
...
```
turt2live, line 3720: please fix

turt2live (changes requested):
it's not clear from the PR what this is trying to accomplish - can we get a description, issue link, or something that says why we need to manipulate the user object directly?

turt2live (approved):
Thanks :D

PR #1549, opened 2020-11-26
Add API for listening to remote hold status, advertise VoIP V1

dbkr:
And avoid isLocalOnHold() returning true whilst the remote side
is un-held.

Start advertising VoIP V1 so that hold functionality can be used.


src/webrtc/call.ts
```diff
@@ -179,7 +184,7 @@ export enum CallErrorCode {
  * bumped to 1. While we partially speak v1 VoIP, it remains
  * as 0.
```
jryans, line 185: This comment might need an update as well.

jryans (approved):

jryans (approved):

PR #1553, opened 2020-12-03
Fixes to support line 1 / 2

dbkr:
* Resume playing audio at the appropriate time
 * Re-emit call events (they were the exception before - all other events
   were re-emitted through the MatrixClient)
 * Fix an audio/video typo

Based on: https://github.com/matrix-org/matrix-js-sdk/pull/1549
Diff: https://github.com/matrix-org/matrix-js-sdk/compare/pull/1549/head...pull/1553/head


src/webrtc/call.ts
```diff
@@ -1019,7 +1032,14 @@ export class MatrixCall extends EventEmitter {
             return;
         }
 
-        const prevOnHold = this.isLocalOnHold();
+        const prevLocalOnHold = this.isLocalOnHold();
+
+        if (description.type === 'answer') {
+            // whenever we get an answer back, clear the flag we set whilst trying to un-hold
+            // the other party: the state of the channels now reflects reality
+            //nope, this doesnt work either
```
jryans, line 1040: Hmm, what does work then? 😜

jryans (approved):
Would be good to clean up the comment, otherwise seems good to me!

src/webrtc/call.ts
dbkr: haha, whoops - the reason it didn't work was because I wasn't checking the flag in the right place, and of course I forgot to remove the comment!

PR #1557, opened 2020-12-14
Skip org.matrix.msc2858 prefix as per the MSC

t3chguy:
See https://github.com/matrix-org/matrix-doc/pull/2858#discussion_r540909985


turt2live (changes requested):
blocking this behind further review on the spec side - I'm not confident that it's been reviewed enough to warrant this sort of action.

PR #1558, opened 2020-12-15
Basic call transfer initiation support

dbkr:
Sends an m.call.replaces event and flag for whether to advertise
support.

MSC2747 (https://github.com/matrix-org/matrix-doc/pull/2747)

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -180,6 +180,9 @@ function keyFromRecoverySession(session, decryptionKey) {
  * @param {boolean} [opts.forceTURN]
  * Optional. Whether relaying calls through a TURN server should be forced.
  *
+ * @param {boolean} [opts.supportsTransfer]
+ * Optional. True to advertise support for call transfers to other parties on Matrix calls.
+ *
  * @param {boolean} [opts.fallbackICEServerAllowed]
  * Optional. Whether to allow a fallback ICE server should be used for negotiating a
  * WebRTC connection if the homeserver doesn't provide any servers. Defaults to false.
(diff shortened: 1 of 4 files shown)


src/client.js
```diff
@@ -364,6 +367,7 @@ export function MatrixClient(opts) {
     this._cryptoCallbacks = opts.cryptoCallbacks || {};
 
     this._forceTURN = opts.forceTURN || false;
+    this._supportsTransfer = opts.supportsTransfer || false;
```
jryans, line 370: Just "transfer" seems too generic at this level... Some day it might have a meaning outside of calls, for example.

```suggestion
    this._supportsCallTransfer = opts.supportsCallTransfer || false;
```

src/client.js
```diff
@@ -696,6 +700,14 @@ MatrixClient.prototype.setForceTURN = function(forceTURN) {
     this._forceTURN = forceTURN;
 };
 
+/**
+ * Set whether to advertise trabsfer support to other parties on Matrix calls.
```
jryans, line 704:
```suggestion
 * Set whether to advertise transfer support to other parties on Matrix calls.
```

src/client.js
```diff
@@ -696,6 +700,14 @@ MatrixClient.prototype.setForceTURN = function(forceTURN) {
...
+ * @param {bool} supportsTransfer True to advertise the 'm.call.transeree' capability
```
jryans, line 705:
```suggestion
 * @param {bool} supportsTransfer True to advertise the 'm.call.transferee' capability
```

src/webrtc/call.ts
```diff
@@ -1370,6 +1399,28 @@ export class MatrixCall extends EventEmitter {
...
+
+        const body = {
+            replacement_id: genCallID(),
+            target_user: {
+                id: targetUserId,
+                display_name: profileInfo.display_name,
+                avatar_url: profileInfo.avatar_url,
+            },
+            create_call: replacementId,
+        } as any;
```
jryans, line 1417: Hmm, seems like we're cheating a bit to mark these as `any`... Is it possible to create an interface with the possible fields?

jryans (changes requested):
Overall, looking good so far!

PR #1561, opened 2020-12-17
Use TypeScript source for development, swap to build during release

jryans:
This changes the JS SDK to point `main` to TypeScript source and remove any
indication of `typings`. For local development and CI workflows, it means many
steps can run without building first, which saves lots of time.

During release, we still build for Node and browsers as before. The release
script adjusts the `main` and `typings` fields before publishing and
distribution to point to the built output for those that use them.

Summary of improvements:

* `yarn install` no longer runs a build step
* Linting and (most) tests don't need a build step either
* CI runs are faster by several minutes as a result

Part of https://github.com/vector-im/element-web/issues/15987

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -6,6 +6,7 @@
 #   github-changelog-generator; install via:
 #     pip install git+https://github.com/matrix-org/github-changelog-generator.git
 #   jq; install from your distribution's package manager (https://stedolan.github.io/jq/)
+#   dot-json; install via Yarn (`yarn global add dot-json`)
 #   hub; install via brew (macOS) or source/pre-compiled binaries (debian) (https://github.com/github/hub) - Tested on v2.2.9
 #   npm; typically installed by Node.js
 #   yarn; install via brew (macOS) or similar (https://yarnpkg.com/docs/install/)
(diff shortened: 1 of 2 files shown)


release.sh
```diff
@@ -15,6 +16,7 @@
 set -e
 
 jq --version > /dev/null || (echo "jq is required: please install it"; kill $$)
+dot-json --version > /dev/null || (echo "dot-json is required: please install it"; kill $$)
```
dbkr, line 19: This seems quite similar to the jq dependency - would it be possible to us the same for both?

dbkr (approved):
otherwise lgtm

release.sh
jryans: Yes, it would be possible, `dot-json` is mostly here for convenience only... With something like `jq` which only outputs to `stdout` (rather than modifying the supplied file), we would need to write to a temporary file and then move it into place (otherwise the input file is clobbered). Perhaps it's better to have just one dependency though and just make a temp file?

jryans: I have converted to `jq` equivalent commands, thanks for the nudge!

PR #1566, opened 2020-12-21
Basic call transfer initiation support

dbkr:
New PR for https://github.com/matrix-org/matrix-js-sdk/pull/1558 after accidentally merging 1559 which was based on it.

PR #1567, opened 2020-12-21
Ignore party ID if opponent is v0

dbkr:
Re-PR of https://github.com/matrix-org/matrix-js-sdk/pull/1559 which I merged but then had to revert because it also merged another thing that wasn't approved.

Based on #1566


dbkr:
(Merging because original PR was approved)

PR #1571, opened 2020-12-29
Store keys with same index but better trust level

Sorunome:
Should fix https://github.com/vector-im/element-web/issues/16025

Signed-Off-By: Sorunome <mail@sorunome.de>


uhoreg (approved):
Looks good.  Thanks

PR #1574, opened 2021-01-12
Don't log if no WebRTC

dbkr:
as hopefully explained in comment

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -1663,7 +1663,8 @@ export function setVideoInput(deviceId: string) { videoInput = deviceId; }
 export function createNewMatrixCall(client: any, roomId: string, options?: CallOpts) {
     // typeof prevents Node from erroring on an undefined reference
     if (typeof(window) === 'undefined' || typeof(document) === 'undefined') {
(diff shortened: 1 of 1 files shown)


src/webrtc/call.ts
```diff
@@ -1663,7 +1663,8 @@ export function setVideoInput(deviceId: string) { videoInput = deviceId; }
 export function createNewMatrixCall(client: any, roomId: string, options?: CallOpts) {
     // typeof prevents Node from erroring on an undefined reference
     if (typeof(window) === 'undefined' || typeof(document) === 'undefined') {
-        logger.info("No window or document object: WebRTC is not supported in this environment");
+        // NB. We don't log here as apps try to create a call object as a test for
```
t3chguy, line 1666: I was going to suggest using `console.debug` to hide it in typical envs only to find out that `console.debug` is an alias for `console.log` in Node :'(

t3chguy (approved):

src/webrtc/call.ts
t3chguy: Oh but we're not even using bare console, but we set up loglevel with DEBUG on - so either way no bueno

PR #1577, opened 2021-01-15
Queue keys for backup even if backup isn't enabled yet

uhoreg:
We may not have managed to set up the backup yet when we get keys.  So we should unconditionally queue up the keys for backup, so that when the backup is set up, they will be sent instead of dropped.

### DIFF
--- a/src/crypto/algorithms/megolm.js
+++ b/src/crypto/algorithms/megolm.js
@@ -369,17 +369,11 @@ MegolmEncryption.prototype._prepareNewSession = async function() {
         key.key, {ed25519: this._olmDevice.deviceEd25519Key},
     );
 
-    if (this._crypto.backupInfo) {
-        // don't wait for it to complete
-        this._crypto.backupGroupSession(
-            this._roomId, this._olmDevice.deviceCurve25519Key, [],
-            sessionId, key.key,
-        ).catch((e) => {
-            // This throws if the upload failed, but this is fine
-            // since it will have written it to the db and will retry.
-            logger.log("Failed to back up megolm session", e);
-        });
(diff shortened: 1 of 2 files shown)


jryans (approved):
Thanks, this seems like the right thing to do for those using backup. It does mean that for users that do not have backup and never intend to use it, they'll have this ever growing queue of things to backup, but it's at least not any added security risk, since we're copying keys from one table and putting them in another. Given that I don't think we can distinguish "will never use backup" from "has backup but it just hasn't enabled it", it seems better to be safe for backup users and always queue like this.

I guess this means though that potentially everyone using backup has a few sessions lying around that were never queued for backup (if they occurred during startup time windows before backup was enabled)? Should we add (probably in a separate PR) a one-time mitigation pass to call `scheduleAllGroupSessionsForBackup` for everyone?

uhoreg:
> I guess this means though that potentially everyone using backup has a few sessions lying around that were never queued for backup (if they occurred during startup time windows before backup was enabled)? Should we add (probably in a separate PR) a one-time mitigation pass to call `scheduleAllGroupSessionsForBackup` for everyone?

It's somewhat rare that a megolm session would come in between the time when a user logs in and when the client sets up the backup.  The message would have to be sent at just the right time -- if it's sent too early, the client wouldn't be logged in yet, so the key wouldn't be sent to that client anyways, and if it's sent too late, it will be received after the backup is set up.  So I'm not sure if it's worth doing a `scheduleAllGroupSessionsForBackup`.  This problem is more noticeable when you're using device dehydration, but that's still a labs flag, so not many people are using it.

PR #1578, opened 2021-01-18
Expose getPresence endpoint

tzyl:
Expose the `getPresence` endpoint in the client server API through `MatrixClient`:
https://matrix.org/docs/spec/client_server/latest#get-matrix-client-r0-presence-userid-status

Currently only the `setPresence` endpoint is exposed.

Signed-off-by: tzyl <tim95@hotmail.co.uk>


jryans (approved):
Thanks for adding a test as well! 😄 This seems reasonable to me.

PR #1584, opened 2021-01-26
Add support for getting call stats

dbkr:
Also add a few 'public' annotations

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -260,6 +260,11 @@ export class MatrixCall extends EventEmitter {
     private micMuted;
     private vidMuted;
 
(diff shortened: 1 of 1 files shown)


src/webrtc/call.ts
```diff
@@ -260,6 +260,11 @@ export class MatrixCall extends EventEmitter {
     private micMuted;
     private vidMuted;
 
+    // the stats for the call at the point it ended. We can't get these after we
+    // tear the call down, so we just grab a snapshot before we stop the call.
+    // The typescript definitions have this type as 'any' :(
```
t3chguy, line 265: :( INDEED

t3chguy (approved):
LGTM

PR #1587, opened 2021-01-31
Fix edge cases with peeking where a room is re-peeked

t3chguy:
but two Room instances are created and things get duplicated


turt2live (approved):
please send to the release branch :)

PR #1596, opened 2021-02-08
Remove rogue plus character

dbkr:
Apparently this is perfectly valid javascript and somehow casts
this.callId to a number... possibly it's ignoring the whitespace
and treating it as `++this.callId`?


jryans (approved):
😱

PR #1600, opened 2021-02-09
Handle undefined peerconn

SimonBrandner:
Accompanied by [#5632](https://github.com/matrix-org/matrix-react-sdk/pull/5632)

Hopefully explained by the comment

Without this, a call could fail without the `CallView` disappearing from the room


dbkr (approved):
good point - thanks!

PR #1601, opened 2021-02-10
Cancel call if no source was selected

SimonBrandner:
This avoids an error which would otherwise appear if no desktop capturer source was selected


jryans (approved):
Thanks, makes sense to me! 😄

PR #1602, opened 2021-02-12
Friendlier room names when calculating from members

tlaundal:
This is my first PR, and an attempt to solve vector-im/element-web#15022.
The implementation is driven by my own motivation for quickly being able to recognize my own rooms (bridged from Facebook Messenger). The style is inspired by Messenger and the issue mentioned above.

I know room naming is specified the client-server spec section [13.2.2.5 Calculating the display name for a room](https://matrix.org/docs/spec/client_server/r0.6.1#calculating-the-display-name-for-a-room), and I don't believe this PR breaks that spec, although "calculate display names [...] and concatenating them" might be interpreted in several ways. I am of course open to changing my implementation should someone have better ideas, though.

## In action
Here are some screenshots from Element.
*Before*
Skjermdump fra 2021-02-14 13-13-55
*After*
Skjermdump fra 2021-02-14 13-17-31

## Possible caveats
- Not all names are delimited by spaces, or are reasonable distinct in their first "word"
- The first "word" of names might be long, so the total length of the room name is longer than ideal for an UI. This problem also exists in the existing implementation

## Commit message
The following changes are made to the room name calculation from
list of members:
 1. Only show the first word of each member name (before first space)
 3. Show up to three members listed explicitly
 2. Show two names before appending "and n others"

The possible forms are now:
 * Empty room
 * Paul Brown
   - When there is only one member except you
 * Paul and John
   - When there are two members except you
 * Paul, John and Will
   - When there are three members except you
 * Paul, John and 3 others
   - When there are more than three members except you
   - Or, only two members have actually joined the room
 * Paul, John and 1 other
   - When the third member except you has not yet joined the room
 * Paul and n other(s)
   - When only one member except you has actually joined the room

Signed-off-by: Tobias Laundal <tobias@laundal.no>

### DIFF
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -1926,6 +1926,16 @@ function calculateRoomName(room, userId, ignoreRoomNameEvent) {
     }
 }
 
+/**
+ * Simplifies a full name to a single word.
+ *
+ * @param {string} name The full name of a room member
+ * @return {string} A single word of the name
+ */
+function shortName(name) {
+    return name.split(" ")[0];
+}
+
 function memberNamesToRoomName(names, count = (names.length + 1)) {
     const countWithoutMe = count - 1;
     if (!names.length) {
@@ -1933,13 +1943,24 @@ function memberNamesToRoomName(names, count = (names.length + 1)) {
     } else if (names.length === 1 && countWithoutMe <= 1) {
         return names[0];
     } else if (names.length === 2 && countWithoutMe <= 2) {
-        return `${names[0]} and ${names[1]}`;
-    } else {
-        const plural = countWithoutMe > 1;
+        return `${shortName(names[0])} and ${shortName(names[1])}`;
+    } else if (names.length === 3 && countWithoutMe <= 3) {
+        return `${shortName(names[0])}, ${shortName(names[1])} and ${shortName(names[2])}`;
+    } else if (names.length >= 2) {
+        const restCount = countWithoutMe - 2;
+        const plural = restCount > 0;
         if (plural) {
-            return `${names[0]} and ${countWithoutMe} others`;
+            return `${shortName(names[0])}, ${shortName(names[1])} and ${restCount} others`;
         } else {
-            return `${names[0]} and 1 other`;
+            return `${shortName(names[0])}, ${shortName(names[1])} and 1 other`;
+        }
+    } else { // Implicit name.length === 1
+        const restCount = countWithoutMe - 1;
+        const plural = restCount > 0;
+        if (plural) {
+            return `${shortName(names[0])} and ${restCount} others`;
+        } else {
+            return `${shortName(names[0])} and 1 other`;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/models/room.js
```diff
@@ -1866,8 +1866,7 @@ function calculateRoomName(room, userId, ignoreRoomNameEvent) {
 
     const joinedMemberCount = room.currentState.getJoinedMemberCount();
     const invitedMemberCount = room.currentState.getInvitedMemberCount();
-    // -1 because these numbers include the syncing user
-    const inviteJoinCount = joinedMemberCount + invitedMemberCount - 1;
```
tlaundal, line 1870: This was actually a bug it seems. `inviteJoinCount` is only used in the changed if statement below and for passing to `memberNamesToRoomName`, but `memberNamesToRoomName` expects the count to be including the current user.

SimonBrandner:
@tlaundal, may I ask why have you close this?

ShadowJonathan:
@SimonBrandner bitrot + no time?

tlaundal:
> @tlaundal, may I ask why have you closed this?

Sure. After six months of no feedback I thought the likelihood of getting a merge was pretty low.

If there is interest for this PR and it actually has a chance to be merged, I am happy to reopen and fix the conflicts.

SimonBrandner:
> thought the likelihood of getting a merge was pretty low.

The PR was stuck on product review which often takes a lot of time, the team is super busy, so this is mostly about waiting... I think it would make sense to leave the PR open and wait for the product review. If they (eventually) approve, you can either resolve the conflicts or someone from the team will do it instead (if they are resolvable). But it's of course up to you

Sorry for the trouble

tlaundal:
I'll reopen this and await product review, then :)

Perhaps some info about this product review and expected wait time could go in `CONTRIBUTING.md`, or maybe one of those "Thanks for your PR"-bots could help with setting expectations for contributors?

pioneer:
For European languages, I personally think that's ok, but would like to hear feedback from someone who speaks other languages. Also, in post-USSR countries, there's sometimes a tradition of writing the surname prior to the name for some reason, this could break the implemented approach as well. Is the first word always the best representation of the full name?

tlaundal:
> Is the first word always the best representation of the full name?

I truly don't know, but I think it's the best compromise between simplicity and usability. It's fine for the few languages I know, but it would be very interesting to get some feedback with other perspectives.

MightyCreak:
> > Is the first word always the best representation of the full name?
> 
> I truly don't know, but I think it's the best compromise between simplicity and usability. It's fine for the few languages I know, but it would be very interesting to get some feedback with other perspectives.

I, at least, have one concrete example of a nickname that won't pass that: "The Chose" (I redacted the nickname, but the idea is here). We would have "The, Paul and 4 others", that won't fly IMHO.

Isn't it possible to give the complete list and let the client do the job?

MadLittleMods:
Related to https://github.com/matrix-org/synapse/issues/12835

PR #1611, opened 2021-02-21
Debuglog as events

FlyveHest:
Changed debuglog in sync module to be user configurable via `debugLogType` option given to `createClient` call.

Can be one of
- 'stdout' : Default if not specified, behaves as module has always done, outputting debug logging to console
- 'event' : Emits a `debuglog` event with string to be logged as data, so the client can incorporate debug logging into their own logging framework, if they so wish
- 'off' : Turns off debugloggin in sync module


SimonBrandner:
@FlyveHest, please include a sign-off as described [here](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.rst#sign-off)

FlyveHest:
> 
> 
> @FlyveHest, please include a sign-off as described [here](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.rst#sign-off)

Is editing the PR text sufficient?

SimonBrandner:
@FlyveHest, yes! :+1:

jryans (changes requested):
Hmm, why only in the sync module and not the entire JS SDK...?

I'd also like to see an approach that doesn't require adding several arguments to every logging call if possible.

FlyveHest:
Mostly because with my, at the time, more limited exposure to the SDK, hadn't seen the output issue elsewhere.

It seems like debug logging has been up to the individual part developers, as there are multiple ways of doing it, depending on where in the project you are.

I recently got encryption working, and got a ton of debug noise there as well, and the way its implemented there is different than in the sync module.

But, i'll try and take a look at a broader scope when I have some more time.

If this PR is rejected, it would be really nice if DEBUG coulde be set to false in the committed code, at least. (Personal opinion)

MadLittleMods:
Slightly related to https://github.com/vector-im/element-web/issues/21532

richvdh:
It looks to me like this PR isn't really going anywhere, so I'm going to go ahead and close it, though I certainly agree that better control over the logging is required. https://github.com/matrix-org/matrix-js-sdk/issues/1970 is maybe a better tracker for this.

PR #1613, opened 2021-02-23
An attempt to  cleanup how constraints are handled in calls

SimonBrandner:
This is an _attempt_ to make the handling of constraints a bit cleaner. It should also be useful for implementing midcall device changes.

@dbkr is probably the best reviewer for this.

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -177,6 +177,7 @@ export enum CallErrorCode {
 enum ConstraintsType {
     Audio = "audio",
     Video = "video",
+    AudioVideo = "audiovideo",
     Screenshare = "screenshare",
 }
 
@@ -355,7 +356,7 @@ export class MatrixCall extends EventEmitter {
         this.checkForErrorListener();
         this.localVideoElement = localVideoElement;
         this.remoteVideoElement = remoteVideoElement;
-        const constraints = await getUserMediaContraints(ConstraintsType.Video);
+        const constraints = await getUserMediaContraints(ConstraintsType.AudioVideo);
         this.placeCallWithConstraints(constraints);
         this.type = CallType.Video;
     }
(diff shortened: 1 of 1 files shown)


src/webrtc/call.ts
```diff
@@ -1713,17 +1700,34 @@ function setTracksEnabled(tracks: Array<MediaStreamTrack>, enabled: boolean) {
...
             return {
                 audio: {
                     deviceId: audioInput ? {ideal: audioInput} : undefined,
-                }, video: false,
+                },
+                video: false,
             };
-        case CallType.Video:
+        }
+        case ConstraintsType.Video: {
```
dbkr, line 1715: Doesn't look like this is used - would it ever be - or is this intended for future in-call device changes? If so my hunch would be that it would be better to ask for both audio & video again, so we always get them as one stream, etc.

If it is necessary, could we factor out the duplicated device ID / resolution between the two?

SimonBrandner: In my head, this would be also useful for adding a video track to an already existing stream that only has an audio track (turning an audio call into a video call). Replacing the whole stream is also an option but I like this one better. But I don't have much WebRTC experience, so my feeling might be wrong...

I can definitely factor out the duplicate stuff

dbkr: OK, yeah - my thinking there is that by requesting audio and video together, its clear that you want audio+video, ie. with the same base timestamp, whereas by asking for a video stream separately there's more scope for the provider to do something weird like decide to give you a video stream with timestamps that bear no resemblance to the audio stream (especially since they'll usually come from separate devices anyway).

SimonBrandner: Ok, that makes sense, I'll rework this later

dbkr (approved):

PR #1616, opened 2021-02-25
Ignore punctuation in strings to be filtered

robintown:
Closes vector-im/element-web#16511.

I believe this is the desired behavior, though let me know if I should add or remove certain characters.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -667,15 +667,16 @@ export function removeHiddenChars(str: string): string {
     return "";
 }
 
-// Regex matching bunch of unicode control characters and otherwise misleading/invisible characters.
+// Regex matching bunch of unicode control characters, punctuation, and otherwise
+// misleading/invisible characters.
 // Includes:
-// various width spaces U+2000 - U+200D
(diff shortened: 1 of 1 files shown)


t3chguy:
This'll have an interesting side effect which may or may not be desirable.

If a room has a user called `DrMike` and another called `Dr. @? Mike` then disambiguation will kick in and force both users to have their MXID appended to their displayname.

t3chguy:
Please fix the lint issue identified by tests and perform the DCO sign-off as per https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.rst#sign-off

SimonBrandner:
> perform the DCO sign-off as per

The commit seems to include a sign-off already

robintown:
Lint has been fixed.

turt2live (commented):
I'm not sure we'd want to trigger disambiguation for users like this. Instead, I think we can apply this in the react-sdk where the rooms are filtered by the room list store.

robintown:
@turt2live Okay, I can do that then

PR #1617, opened 2021-02-25
Add logging to crypto store transactions

jryans:
We churn through a huge number of crypto store transactions during startup, which may be the cause of the symptoms in https://github.com/vector-im/element-web/issues/16194.


t3chguy (approved):
Looks sane

PR #1618, opened 2021-02-25
Fix ContentHelpers export

robintown:
This was previously exporting a promise, since it called the import function manually but didn't await the result. However, since we have Babel we can just use the new export … as … from syntax instead.


jryans (approved):
Thanks, this seems reasonable to me!

PR #1624, opened 2021-02-26
Support sending invite reasons

robintown:
Added as the final argument to `invite` in order to keep backwards compatibility.


jryans (approved):
Thanks, this looks good to me! 😄

PR #1625, opened 2021-02-28
Add function to determine who has read at least a given event

turt2live:
**Required by https://github.com/matrix-org/matrix-react-sdk/pull/5699**

### DIFF
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -1585,6 +1585,44 @@ Room.prototype.recalculate = function() {
 };
 
 /**
+ * Get a list of user IDs who have read up to and including the given
+ * event. This will look at the most recent maxEvents to identify the
+ * given event, and any users who have read events leading up to that
+ * event. If the function exceeds the given limit, an empty array will
+ * be returned. Otherwise, an array with at least the event sender will
+ * be returned.
+ * @param {MatrixEvent} event The event to search for.
+ * @param {number} maxEvents The maximum number of events to search within.
+ * @returns {string[]} A list of user IDs who have read the event.
+ */
+Room.prototype.getUsersWhoHaveRead = function(event, maxEvents) {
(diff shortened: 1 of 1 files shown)


src/models/room.js
```diff
@@ -1584,6 +1584,44 @@ Room.prototype.recalculate = function() {
...
+ * event. This will look at the most recent maxEvents to identify the
+ * given event, and any users who have read events leading up to that
+ * event. If the function exceeds the given limit, an empty array will
+ * be returned. Otherwise, an array with at least the event sender will
+ * be returned.
+ * @param {MatrixEvent} event The event to search for.
+ * @param {number} maxEvents The maximum number of events to search within.
+ * @returns {string[]} A list of user IDs who have read the event.
+ */
+Room.prototype.getUsersWhoHaveRead = function(event, maxEvents) {
```
jryans, line 1598: Why do you want this API to collect _all_ users who have read the event? In the one caller so far, you actually only care if _someone_ has read it (other than yourself), but the actual list of senders goes unused... Is there a plan to use it in the future?

jryans (commented):
I'm being cautious on this, as I'd like to work out if there's any possible way to do less work to support the use case.

src/models/room.js
turt2live: It's largely the balance of usefulness and API scope. We could add a `hasAnyoneRead` function, but the react SDK would be the only consumer. In this approach, we're adding something theoretically useful for other applications, and maybe even ourselves if we decide to add a "who has read this" API. 

Ideally this whole problem is solved by matrix rather than a loop in our code, but this is the first time we've needed a different kind of receipt calculation.

jryans (approved):
Okay, let's go with this for now.

turt2live:
This ultimately isn't used.

PR #1627, opened 2021-03-02
Avoid deadlocks when ensuring Olm sessions for devices

jryans:
This reworks tracking the Olm sessions a particular task is updating to avoid deadlocks. By ensuring we synchronously mark all sessions a task cares about as in progress from the start, we know that no other tasks will own updating a session in common, which avoids deadlocks across multiple tasks that might be working on a shared set of devices.

The potential for deadlocks first appeared with https://github.com/matrix-org/matrix-js-sdk/pull/857 (2 years ago), but I believe it would be unlikely for most users to trigger this deadlock during normal usage. The app would need to be sluggish enough to permit stacking up several messages in different rooms with devices in common, which is most likely only possible on larger accounts. In any case, with this change, this process should now avoid deadlocks in all cases.

Fixes https://github.com/vector-im/element-web/issues/16194

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "build:minify-browser": "terser dist/browser-matrix.js --compress --mangle --source-map --output dist/browser-matrix.min.js",
     "gendoc": "jsdoc -c jsdoc.json -P package.json",
     "lint": "yarn lint:types && yarn lint:js",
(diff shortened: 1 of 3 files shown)


dbkr (approved):
Phew, well done for tracking that one down.

PR #1628, opened 2021-03-02
Extract display name patterns to constants

jryans:
This changes to constant patterns for calculating display names, which cuts out ~18% of the time spent and reduces GC pressure as well.


turt2live (approved):
yay, faster speed :D

PR #1629, opened 2021-03-03
Add a function to share decryption keys for room history

uhoreg:
Needed for https://github.com/matrix-org/matrix-react-sdk/pull/5711

### DIFF
--- a/src/crypto/algorithms/megolm.js
+++ b/src/crypto/algorithms/megolm.js
@@ -1685,9 +1685,11 @@ MegolmDecryption.prototype.retryDecryptionFromSender = async function(senderKey)
  * Share the keys with the given users for the given messages.
  *
  * @param {object} devicesByUser a map of user to array of module:crypto/deviceinfo.
- * @param {object} messages an iterator giving the Matrix messages to share.
+ * @param {function} nextMessage a function that returns the next Matrix message to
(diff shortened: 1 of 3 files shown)


src/crypto/algorithms/megolm.js
```diff
@@ -1681,6 +1681,140 @@ MegolmDecryption.prototype.retryDecryptionFromSender = async function(senderKey)
     return !this._pendingEvents[senderKey];
 };
 
+/**
+ * Share the keys with the given users for the given messages.
+ *
+ * @param {object} devicesByUser a map of user to array of module:crypto/deviceinfo.
+ * @param {function} nextMessage a function that returns the next Matrix message to
```
jryans, line 1688: How would a general user of the JS SDK (outside of Element) have any idea what this function should be...?

Either we need much more extensive documentation here, or we should more likely move this function into the JS SDK itself, since at least the draft version doesn't seem to have a UI component.

jryans (changes requested):
Thanks, the core of this looks good, but I wonder if it would be best to move the `nextMessage` bit here instead of leaving it at the React layer.

jryans:
I assume this can be closed since we merged https://github.com/matrix-org/matrix-js-sdk/pull/1640 instead.

PR #1632, opened 2021-03-08
Expose APIs needed for reworked cross-signing login flow

ara4n:
...and fix a daft bug where we'd cancel ourselves when selecting a self-verification partner during cross-signed verification.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -570,6 +570,27 @@ MatrixClient.prototype.rehydrateDevice = async function() {
 };
 
 /**
+ * Get the current dehydrated device, if any
+ * @return {Promise} A promise of an object containing the dehydrated device
+ */
+MatrixClient.prototype.getDehydratedDevice = async function() {
+    try {
+        return await this._http.authedRequest(
+            undefined,
+            "GET",
+            "/dehydrated_device",
+            undefined, undefined,
+            {
+                prefix: "/_matrix/client/unstable/org.matrix.msc2697.v2",
+            },
+        );
+    } catch (e) {
+        logger.info("could not get dehydrated device", e.toString());
+        return;
+    }
+};
+
+/**
  * Set the dehydration key.  This will also periodically dehydrate devices to
  * the server.
  *
--- a/src/client.js
+++ b/src/client.js
@@ -498,19 +498,8 @@ MatrixClient.prototype.rehydrateDevice = async function() {
         return;
     }
 
-    let getDeviceResult;
-    try {
-        getDeviceResult = await this._http.authedRequest(
-            undefined,
-            "GET",
-            "/dehydrated_device",
-            undefined, undefined,
-            {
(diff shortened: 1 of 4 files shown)


src/client.js
```diff
@@ -572,6 +572,27 @@ MatrixClient.prototype.rehydrateDevice = async function() {
     }
 };
 
+/**
+ * Get the current dehydrated device, if any
+ * @return {Promise} A promise of an object containing the dehydrated device
+ */
+MatrixClient.prototype.getDehydratedDevice = async function() {
```
jryans, line 579: The boundary between base APIs and client is quite blurry, but perhaps this one is more base APIs, since it's just a straightforward API access...?

src/client.js
```diff
@@ -572,6 +572,27 @@ MatrixClient.prototype.rehydrateDevice = async function() {
...
+    try {
+        return await this._http.authedRequest(
```
jryans, line 581: This same code block exists in `MatrixClient.prototype.rehydrateDevice`, could you update it to call the new function you've added?

src/crypto/verification/request/ToDeviceChannel.js
```diff
@@ -179,7 +179,9 @@ export class ToDeviceChannel {
         const isAcceptingEvent = type === START_TYPE || type === READY_TYPE;
         // the request has picked a ready or start event, tell the other devices about it
         if (isAcceptingEvent && !wasStarted && isStarted && this._deviceId) {
-            const nonChosenDevices = this._devices.filter(d => d !== this._deviceId);
+            const nonChosenDevices = this._devices.filter(
+                d => d !== this._deviceId && d !== this._client.getDeviceId(),
```
jryans, line 183: Hmm, how was this manifesting...? A bit surprised this wasn't more a of wide spread problem...

jryans (changes requested):
Thanks, looking good overall, just a few tweaks to make.

src/crypto/verification/request/ToDeviceChannel.js
ara4n: on accepting a self-verification, the verifier sent the cancellation for all devices other than the verifiee.  Which included the verifier, who saw the cancellation and promptly cancelled the verification.  I've seen it in the wild a few times before.  The workaround is to manually ignore cancellations to yourself whilst you're self-verifying, but frankly I think it's safer to do it in here.

src/client.js
ara4n: hm, i think it's probably the lesser evil to leave it next to the rest of the dehydration functions, even though it's just an API wrapper?  and it makes the code easier to read...

ara4n:
@jryans PTAL again

t3chguy:
image

(little refresh icon re-requests review)

src/client.js
jryans: Okay, seems alright then. As mentioned, it's quite a blurry line anyway.

PR #1634, opened 2021-03-08
Check TURN servers periodically, and at start of calls

dbkr:
Hopefully this should make our turn-credential checking code a bit
more robust (and possibly fix a seconds / ms mismatch).

See also https://github.com/matrix-org/matrix-react-sdk/pull/5730

### DIFF
--- a/spec/unit/webrtc/call.spec.ts
+++ b/spec/unit/webrtc/call.spec.ts
@@ -79,6 +79,7 @@ class MockRTCPeerConnection {
         return Promise.resolve();
     }
     close() {}
+    getStats() { return []; }
 }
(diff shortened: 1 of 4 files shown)


t3chguy:
If we're checking at the start of calls when why do we need to do it periodically too?

dbkr:
We may have failed to get credentials, in which case checking before the call will cover us for the case where the user tries to call before we retry fetching credentials.

t3chguy (approved):

dbkr:
after much fighting with tests, some more changes

t3chguy (approved):

Issue #1619, opened 2021-02-26
Key querying is not chunked and servers might not support an arbitrary amount of users in the request.


poljar:
**Describe the bug**
During my testing using Complement with this PR https://github.com/matrix-org/complement/pull/75, I hit a bug when element would need to send out a `/keys/query` request. Complement creates an server with N E2EE capable users that are joined into a single room.

When I increased the number of users to 1000, synapse failed to return a response for the initial `/keys/query` request which contains all 1000 users.

**To Reproduce**
1. Freshly log in to an account that shares an E2EE room with 1000 users.
2. Try to send a message in the room.
3. Observe the 500 `/keys/query` response in the network tab.

This completely wedges the E2EE support and the message can't ever be sent.

I'm not sure if this is a specific limitation of the Synapse install Complement was running but it seems a good idea to chunk `/keys/query` requests anyways. It seems that chunking the requests out at 200 or 500 users (or anything between that) would work.

PR #1639, opened 2021-03-10
Download device keys in chunks of 250

Johennes:
As described in #1619, the server might overload if the number of users in the key request is too large. To prevent this, the download is broken into chunks of 250 users each. Responses are processed once all chunks have been downloaded.

I picked the chunk size more or less arbitrarily following the observations in #1619. Happy to change it if a different value seems more appropriate.

Also, this is my first ever contribution so I'm hoping I at least didn't screw it up totally. :see_no_evil: 

Signed-off-by: Johannes Marbach <n0-0ne+github@mailbox.org>

### DIFF
--- a/src/crypto/DeviceList.js
+++ b/src/crypto/DeviceList.js
@@ -28,7 +28,7 @@ import {DeviceInfo} from './deviceinfo';
 import {CrossSigningInfo} from './CrossSigning';
 import * as olmlib from './olmlib';
 import {IndexedDBCryptoStore} from './store/indexeddb-crypto-store';
-import {defer, sleep} from '../utils';
+import {chunkPromises, defer, sleep} from '../utils';
 
 
 /* State transition diagram for DeviceList._deviceTrackingStatus
@@ -62,7 +62,7 @@ const TRACKING_STATUS_UP_TO_DATE = 3;
  * @alias module:crypto/DeviceList
  */
 export class DeviceList extends EventEmitter {
-    constructor(baseApis, cryptoStore, olmDevice) {
+    constructor(baseApis, cryptoStore, olmDevice, keyDownloadChunkSize = 250) {
         super();
 
         this._cryptoStore = cryptoStore;
@@ -97,6 +97,9 @@ export class DeviceList extends EventEmitter {
 
         // userId -> promise
         this._keyDownloadsInProgressByUser = {};
+
+        // Maximum number of user IDs per request to prevent server overload (#1619)
+        this._keyDownloadChunkSize = keyDownloadChunkSize;
 
         // Set whenever changes are made other than setting the sync token
         this._dirty = false;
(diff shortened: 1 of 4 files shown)


src/crypto/DeviceList.js
```diff
@@ -780,13 +783,17 @@ class DeviceListUpdateSerialiser {
...
-            downloadUsers, opts,
-        ).then(async (res) => {
-            const dk = res.device_keys || {};
-            const masterKeys = res.master_keys || {};
-            const ssks = res.self_signing_keys || {};
-            const usks = res.user_signing_keys || {};
+        const promises = [];
+        for (let i = 0; i < downloadUsers.length; i += this._deviceList._keyDownloadChunkSize) {
+            const userSlice = downloadUsers.slice(i, i + this._deviceList._keyDownloadChunkSize);
+            promises.push(this._baseApis.downloadKeysForUsers(userSlice, opts));
```
turt2live, line 789: our next problem is probably going to be rate limiting. In large rooms (>1000 users), this will be only 4 requests, but in massive rooms (>5000 users) this would be 20. Servers tend to allow a burst of up to 5 requests, with later ones being rate limited. We should parallelize at most 2-3 requests (probably 3?) to avoid a rate limit problem.

turt2live (changes requested):
otherwise seems to do the right thing - thanks for working on this!

src/crypto/DeviceList.js
Johennes: Ah, good call. Do you want this to be implemented just here for the key downloads? Or would it make sense to push the rate limiting down into `MatrixHttpApi` so that it affects _all_ requests?

turt2live: probably just here for now - putting it at the MatrixHttpApi layer is probably going to cause unexpected problems.

src/utils.ts
```diff
@@ -745,6 +745,15 @@ export function promiseTry<T>(fn: () => T): Promise<T> {
...
 }
 
+// Creates and awaits all promises, running no more than `chunkSize` at the same time
+export async function chunkPromises<T>(fns: (() => Promise<T>)[], chunkSize: number): Promise<T[]> {
+    const results: T[] = [];
+    for (let i = 0; i < fns.length; i += chunkSize) {
+        results.push(...(await Promise.all(fns.slice(i, i + chunkSize).map(fn => fn()))));
+    }
+    return results;
+}
```
Johennes, line 755: I couldn't find any existing logic for chunking promises so I added this. It does feel a little bit clumsy admittedly. Not sure if there's a more elegant way of doing this?

turt2live: it seems fine to me - actually pretty clever for using a for loop in this way. Took me a minute to figure out it'd be fine though 😅

turt2live (approved):
Thanks!

PR #1640, opened 2021-03-11
Add function to share megolm keys for historical messages, take 2

uhoreg:
Implements https://github.com/matrix-org/matrix-doc/pull/3061

### DIFF
--- a/src/crypto/algorithms/megolm.js
+++ b/src/crypto/algorithms/megolm.js
@@ -38,8 +38,8 @@ import {WITHHELD_MESSAGES} from '../OlmDevice';
 
 // determine whether the key can be shared with invitees
 function isRoomKeyShareable(room) {
-    const visibilityEvent = room.currentState
-          .getStateEvents("m.room.history_visibility", "");
+    const visibilityEvent = room.currentState &&
+          room.currentState.getStateEvents("m.room.history_visibility", "");
     // NOTE: if the room visibility is unset, it would normally default to
     // "world_readable".
     // (https://spec.matrix.org/unstable/client-server-api/#server-behaviour-5)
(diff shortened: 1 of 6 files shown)


dbkr (commented):
I think this looks plausible - may not be a bad idea for ryan to take a quick look too as he has context on the the recents deadlock bugs.

Also important to note that this increases the crypto db version so there's no going back once you've run this.

src/crypto/algorithms/megolm.js
```diff
@@ -192,7 +210,7 @@ utils.inherits(MegolmEncryption, EncryptionAlgorithm);
  *    OutboundSessionInfo when setup is complete.
  */
 MegolmEncryption.prototype._ensureOutboundSession = async function(
-    devicesInRoom, blocked, singleOlmCreationPhase,
+    room, devicesInRoom, blocked, singleOlmCreationPhase,
```
jryans, line 213: Would it make sense for `EncryptionAlgorithm` to hold the room model from the start at construction time, rather than expecting in certain functions...? It already has the room ID. Just an idea, perhaps there's a reason to prefer this instead.

src/crypto/algorithms/megolm.js
```diff
@@ -1370,10 +1400,14 @@ MegolmDecryption.prototype.onRoomKeyEvent = function(event) {
         keysClaimed = event.getKeysClaimed();
     }
 
+    const extraSessionData = {};
+    if (content["org.matrix.msc3061.shared_history"]) {
+        extraSessionData.sharedHistory = true;
```
jryans, line 1405: Should this locally stored session data be prefixed with the MSC while it's an experiment, or do we assume it's not necessary to guard local data like that?

jryans (approved):
Looks good overall, thanks! 😄

src/crypto/algorithms/megolm.js
uhoreg: In general, local data doesn't need to be guarded in the same way as data sent between client, as the local client can usually manage any changes itself.

src/crypto/algorithms/megolm.js
uhoreg: Yeah, I don't know if there's a particular reason why the `EncryptionAlgorithm` only holds the room ID rather than the whole room model.  But I wanted to avoid making too many unrelated changes.

PR #1641, opened 2021-03-11
docs: registerGuest()

NicolaiSoeborg:
Otherwise the client will have the default `_guest = false` and it will try to pushrules when doing the initial sync which will fail

This will set `client._guest = true` on the client even if it fails to register as a guest.
I could wrap the callback in some fancy logic (help needed)... buuut I guess people shouldn't rely on `.registerGuest()` failing and not having side-effects

Signed-off-by: Nicolai Søborg <matrix@xn--sb-lka.org>

### DIFF
--- a/src/base-apis.js
+++ b/src/base-apis.js
@@ -246,16 +246,30 @@ MatrixBaseApis.prototype.register = function(
 
 /**
  * Register a guest account.
+ * This method returns the auth info needed to create a new authenticated client,
+ * Remember to call `setGuest(true)` on the (guest-)authenticated client, e.g:
+ * ```javascript
+ * const tmpClient = await sdk.createClient(MATRIX_INSTANCE);
+ * const { user_id, device_id, access_token } = tmpClient.registerGuest();
+ * const client = createClient({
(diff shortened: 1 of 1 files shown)


t3chguy:
The typical usage is to use a throwaway MatrixClient instance to perform login/registration and create another one with the output credentials. As you can see the registerGuest method just returns credentials. If the method was updating the client's Access Token etc then yes it would make sense to also set `_guest` but it does not so I think it just muddies the usage further.

NicolaiSoeborg:
Thanks for the review.

To avoid spamming new users, I added some logic in my code to create the MatrixClient from old saved credentials, but forgot about it, so I thought my MatrixClient was created by `registerGuest()`, but I was indeed doing the above throwaway trick.

I changed this PR to just add some docs. I guess I might not be the only one forgetting to call `setGuest(true)` :wink: 

Hope the styling is okay? Not sure what you feel about having example code in doc comments?

turt2live (approved):
Woo, thanks! The styling is probably fine - it's not my favourite thing to have examples in function docs, but we don't really have a better system so this makes it fine.

PR #1642, opened 2021-03-12
Migrate to `eslint-plugin-matrix-org`

jryans:
This migrates to the new plugin form of our custom ESLint configs. As part of this, some packages are de-duplicated, configs streamlined, etc.

Related to https://github.com/vector-im/element-web/issues/16276
Depends on https://github.com/matrix-org/eslint-plugin-matrix-org/pull/5


package.json
```diff
@@ -72,15 +74,16 @@
...
+    "@typescript-eslint/parser": "^4.17.0",
     "babel-jest": "^26.6.3",
     "babelify": "^10.0.0",
     "better-docs": "^2.3.2",
     "browserify": "^17.0.0",
     "docdash": "^1.2.0",
     "eslint": "7.18.0",
-    "eslint-config-matrix-org": "^0.2.0",
-    "eslint-plugin-babel": "^5.3.1",
+    "eslint-config-google": "^0.14.0",
```
turt2live, line 85: hmm, why are we pulling in the google config here?

jryans: Our configs [extend from it](https://github.com/matrix-org/eslint-plugin-matrix-org/pull/5/files#diff-92ae86c53e137522d28b1e7e7d87ecba8c83f4f71baa1dce5142a1c788fa7d51R7), and since I've [removed all the deps](https://github.com/matrix-org/eslint-plugin-matrix-org/pull/5/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R75-R88) from our own package, it appears here.

package.json
```diff
@@ -72,15 +74,16 @@
...
+    "eslint-plugin-matrix-org": "github:matrix-org/eslint-plugin-matrix-org#737c55547881d4f5616b7783d4b66b09d0c7b357",
```
jryans, line 86: I'll change this to point to the `main` branch once the upstream changes are merged.

package.json
turt2live: hmm, interesting

turt2live (approved):
I meant to approve this

jryans:
Thanks! I think I'll hold onto this for a moment, until I can come back around and make the same changes in the other repos, as that's likely less confusing...

jryans:
Well, I should have expected the React SDK side would turn into a much larger dance... I'll get this one updated so it can merge.

PR #1647, opened 2021-03-20
Don't send m.call.hangup if m.call.invite wasn't sent either

SimonBrandner:
Don't send `m.call.hangup` if `m.call.invite` wasn't sent either (this happens when getting user media fails before sending an invite)

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -660,6 +660,8 @@ export class MatrixCall extends EventEmitter {
 
         logger.debug("Ending call " + this.callId);
         this.terminate(CallParty.Local, reason, !suppressEvent);
+        // We don't want to send hangup here if we didn't even get to sending an invite
+        if (this.state === CallState.WaitLocalMedia) return;
         const content = {};
         // Continue to send no reason for user hangups temporarily, until
         // clients understand the user_hangup reason (voip v1)


src/webrtc/call.ts
```diff
@@ -1446,7 +1450,11 @@ export class MatrixCall extends EventEmitter {
 
         // party ID must match (our chosen partner hanging up the call) or be undefined (we haven't chosen
         // a partner yet but we're treating the hangup as a reject as per VoIP v0)
-        if (this.partyIdMatches(msg) || this.state === CallState.Ringing) {
+        if (
+            this.partyIdMatches(msg) ||
+            this.state === CallState.Ringing ||
+            this.state === CallState.InviteSent
```
SimonBrandner, line 1456: After thinking this over again, this probably doesn't make sense, does it?

SimonBrandner: I wasn't thinking clearly. To explain this: we shouldn't end the call here because we want to give other devices a chance to answer the call

turt2live (approved):
Changes seem fine, though @SimonBrandner you'll have to sign off on https://github.com/matrix-org/matrix-js-sdk/pull/1647/commits/74a31d814f92730e8093c3e1c6883a35e7917c13 too for this to land.

turt2live (approved):
thanks!

PR #1649, opened 2021-03-23
Cache cross-signing private keys if needed on bootstrap

jryans:
This is a revised version of https://github.com/matrix-org/matrix-js-sdk/pull/1472 which was previously reverted for causing security prompts to appear on device list sync. In this version, we only allow private key requests (which are likely to trigger user dialogs) if we are coming from the bootstrap path.

This allows sessions that have already synced cross-signing public keys but never got the private keys for some reason to make forward progress when e.g. the user triggers bootstrap from security settings.


turt2live (approved):
looks plausible

jryans:
I think I'll hold onto this until after the RC, since it's been troublesome in the past.

PR #1651, opened 2021-03-25
Fixed the media fail error on caller's side

DantrazTrev:
Fixes vector-im/element-web/issues/16747

Switched `CallType` before async function calls  in `/src/webrtc/call.ts` that were causing the `showMediaCaptureError` in `matrix-react-sdk/src/CallHandler.tsx` to skip all the conditions and hence causing a generic error.

Signed-off-by: Ayush Pratap Singh ayushpratap16@gmail.com

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -396,7 +396,6 @@ export class MatrixCall extends EventEmitter {
 
             logger.debug("Got screen stream, requesting audio stream...");
             const audioConstraints = getUserMediaContraints(ConstraintsType.Audio);
-            this.type = CallType.Video;
             this.placeCallWithConstraints(audioConstraints);
         } catch (err) {
             this.emit(CallEvent.Error,
@@ -406,7 +405,7 @@ export class MatrixCall extends EventEmitter {
                 ),
             );
         }
-        
+        this.type = CallType.Video;
     }
 
     public getOpponentMember() {
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -337,8 +337,9 @@ export class MatrixCall extends EventEmitter {
         logger.debug("placeVoiceCall");
         this.checkForErrorListener();
         const constraints = getUserMediaContraints(ConstraintsType.Audio);
-        await this.placeCallWithConstraints(constraints);
         this.type = CallType.Voice;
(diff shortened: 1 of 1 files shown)


t3chguy:
Needs DCO sign-off as per https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.rst#sign-off

src/webrtc/call.ts
```diff
@@ -394,6 +396,7 @@ export class MatrixCall extends EventEmitter {
 
             logger.debug("Got screen stream, requesting audio stream...");
             const audioConstraints = getUserMediaContraints(ConstraintsType.Audio);
+            this.type = CallType.Video;
```
SimonBrandner, line 399: Is there any reason to put this inside and not before the try-catch block? I am probably just missing something here :sweat_smile:

DantrazTrev: Oh no , sorry I would just fix it

DantrazTrev:
> Needs DCO sign-off as per https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.rst#sign-off

Done

src/webrtc/call.ts
SimonBrandner: Hmm, I've just realized this might actually do something we want by accident :D

DantrazTrev: Oh, just a sec let me just check . I think it would work the either way

SimonBrandner: I think an intelligent thing to do here would be to decouple the call type and the `NoUserMedia` error. But I wouldn't worry about it if you don't want to

SimonBrandner: The current solution will work fine

DantrazTrev: I could do that but I've a naive question how do you make screen share calls here? I couldn't figure out

SimonBrandner: Screen-share calls are a bit of a hack right now... You can shift-click the video call button to initiate a screen-sharing call

DantrazTrev: > The current solution will work fine

Alright

SimonBrandner (approved):
LGTM, just fix the unnecessary new lines

turt2live (approved):
Seems plausible, thanks!

PR #1655, opened 2021-04-01
Persist unsent messages for subsequent sessions

gsouquet:
Fixes vector-im/element-web#7177

This pull requests hooks into the `Room#_pendingEventList` flow to store a serialised version of an unsent `MatrixEvent`.

The storage key is: `mx_pending_events_${roomId}`

Upon instanciation the `Room` model will query the storage and populate `_pendingEventList` if required. Element then picks up the unsent messages and display the `RoomStatusBar` and a user can decide whether to send or cancel the queued messages

### DIFF
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -189,6 +189,17 @@ export function Room(roomId, client, myUserId, opts) {
 
     if (this._opts.pendingEventOrdering == "detached") {
         this._pendingEventList = [];
+        const serializedPendingEventList = client._sessionStore.store.getItem(pendingEventsKey(this.roomId));
+        if (serializedPendingEventList) {
+            const self = this;
+            JSON.parse(serializedPendingEventList)
+                .forEach(serializedEvent => {
+                    const event = new MatrixEvent(serializedEvent);
+                    event.setStatus(EventStatus.NOT_SENT);
+                    const txnId = client.makeTxnId();
+                    self.addPendingEvent(event, txnId);
+                });
+        }
     }
 
     // read by megolm; boolean value - null indicates "use global value"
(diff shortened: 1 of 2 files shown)


src/models/room.js
```diff
@@ -189,6 +189,17 @@ export function Room(roomId, client, myUserId, opts) {
 
     if (this._opts.pendingEventOrdering == "detached") {
         this._pendingEventList = [];
+        const serializedPendingEventList = client._sessionStore.store.getItem(pendingEventsKey(this.roomId));
+        if (serializedPendingEventList) {
+            const self = this;
```
dbkr, line 194: You should just be able to use `this` since we have arrow functions now? (which indeed is what you've used)

dbkr (approved):
otherwise lgtm!

jryans:
@gsouquet What's the status of this one? Looks like perhaps merging develop to get the cross-signing test fix would get things moving again?

gsouquet:
> @gsouquet What's the status of this one? Looks like perhaps merging develop to get the cross-signing test fix would get things moving again?

Yeah, I was waiting on ops to give me access to BuildKite so that I can re-run the pipelines, but can do that too 👍

src/models/room.js
```diff
@@ -189,6 +191,16 @@ export function Room(roomId, client, myUserId, opts) {
 
     if (this._opts.pendingEventOrdering == "detached") {
         this._pendingEventList = [];
+        const serializedPendingEventList = client._sessionStore.store.getItem(pendingEventsKey(this.roomId));
+        if (serializedPendingEventList) {
+            JSON.parse(serializedPendingEventList)
+                .forEach(serializedEvent => {
+                    const event = new MatrixEvent(serializedEvent);
+                    event.setStatus(EventStatus.NOT_SENT);
+                    const txnId = client.makeTxnId();
```
t3chguy, line 200: I think it would be good to have the txnId be persisted to ensure idempotency even if the user refreshed during a server timeout/whilst sending rather than generating a fresh one

gsouquet: Addressed in matrix-org/matrix-js-sdk#663

PR #1657, opened 2021-04-06
Only try to cache private keys we know exist

jryans:
This tweaks https://github.com/matrix-org/matrix-js-sdk/pull/1649 to only try caching private keys locally once we've confirmed they exist first. This most likely only an issue in tests, where we sometimes create only some but not all keys.


gsouquet (approved):
Looks good to me, 💯

PR #1660, opened 2021-04-07
Support for multiple streams (not MSC3077)

SimonBrandner:
### Description

This moves the handling of video elements out of the `js-sdk` and makes `js-sdk` more flexible in terms of multi-stream support. This is done by adding an array of type `CallFeed` to `MatrixCall`. The `CallFeed` class represents a WebRTC `MediaStream` with some additional useful methods. The `CallFeed` array can then be accessed by the `react-sdk` (and others) which can then display each feed on the user's screen.

### Dependencies

+ [#5833](https://github.com/matrix-org/matrix-react-sdk/pull/5833)

### Merging and review

+ This is a **very dangerous** change - it might be a good idea for multiple people to review it. 
+ It should also be merged right after the first RC so that it has a lot of time to be tested on develop
+ This will require a major version bump because it drastically changes everything


src/webrtc/callFeed.ts
```diff
@@ -0,0 +1,86 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import EventEmitter from "events";
+
+export enum CallFeedPurpose {
+    Usermedia = "usermedia",
+    Screenshare = "screenshare",
+}
```
SimonBrandner, line 22: It would make sense to put these into `callEventTypes.ts` as `SDPStreamMetadataPurpose` to match MSC3077

src/webrtc/call.ts
```diff
@@ -363,22 +360,11 @@ export class MatrixCall extends EventEmitter {
...
-     * @param {Element} localVideoElement a <code>&lt;video&gt;</code> DOM element
-     * to render the local camera preview.
      * @throws If you have not specified a listener for 'error' events.
      */
-    async placeScreenSharingCall(
-        remoteVideoElement: HTMLVideoElement,
-        localVideoElement: HTMLVideoElement,
-        selectDesktopCapturerSource?: () => Promise<DesktopCapturerSource>,
-    ) {
+    async placeScreenSharingCall(selectDesktopCapturerSource: () => Promise<DesktopCapturerSource>) {
```
SimonBrandner, line 365:
```suggestion
    async placeScreenSharingCall(selectDesktopCapturerSource?: () => Promise<DesktopCapturerSource>) {
```

jryans:
Routing to @dbkr for a first pass, feel free to send back to the queue in another reviewer is needed.

src/webrtc/callFeed.ts
```diff
@@ -0,0 +1,82 @@
...
+export enum CallFeedEvent {
+    NewStream = "new_stream",
+}
+
+export class CallFeed extends EventEmitter {
+    constructor(
+        public stream: MediaStream,
+        public userId: string,
+        public purpose: SDPStreamMetadataPurpose,
+        private client: any, // Fix when client is TSified
```
dbkr, line 29: fwiw you can still define this as a `MatrixClient` and when it becomes ts, it will do the right thing

src/webrtc/callFeed.ts
```diff
@@ -0,0 +1,82 @@
...
+        private roomId: string,
+    ) {
+        super()
+    }
+
+    /**
+     * Returns callRoom member
+     * @returns member of the callRoom
+     */
+    public getMember() {
```
dbkr, line 39: return type please :)

src/webrtc/callFeed.ts
```diff
@@ -0,0 +1,82 @@
...
+        const callRoom = this.client.getRoom(this.roomId);
+        return callRoom.getMember(this.userId);
+    }
+
+    /**
+     * Returns true if CallFeed is local, otherwise returns false
+     * @returns {boolean} is local?
+     */
+    public isLocal() {
```
dbkr, line 48: type again

dbkr (commented):
on a first pass of review: looks a lot neater but this will be one of the biggest API changes I think we've done on the js-sdk so we might need to do developer-advocacy or something, at least in the sense of figuring out how many people this is going to impact.

src/webrtc/callFeed.ts
SimonBrandner: Oops :D

src/webrtc/call.ts
```diff
@@ -1354,32 +1260,31 @@ export class MatrixCall extends EventEmitter {
             logger.warn(`Streamless ${ev.track.kind} found: ignoring.`);
             return;
         }
+
+        let remoteStream = this.feeds.find((feed) => {return !feed.isLocal()})?.stream;
```
dbkr, line 1264: I think this would be less confusing if this were just `oldRemoteStream` and then not reassigned later to be the new stream.

dbkr (approved):
OK - I think this is looking good now, thanks for all the iterations. Let's figure out how we land the API change.

jryans:
Okay, let's go ahead and merge this! 😄 After looking around for other usages of these APIs, we believe it's enough to have a major release, and don't need to do further advocacy beyond that.

SimonBrandner:
[PR with the example update](https://github.com/matrix-org/matrix-js-sdk/pull/1680)

PR #1663, opened 2021-04-12
Fix timeline jumpiness by setting correct txnId

gsouquet:
Fixes vector-im/element-web#16914

Regression introduced by matrix-org/matrix-js-sdk#1655

I incorrectly tried to remove pending events in `updatePendingEvent` when this reconciliation is the responsability of that transaction ID. This PR sets that `txnId` attribute properly and remove the need to explicitely call `removePendingEvent` as the event is now going through the usual timeline flow

@t3chguy raised an [interesting comment](https://github.com/matrix-org/matrix-js-sdk/pull/1655#discussion_r611412326) in my previous pull request that is adressed here. The `txnId` is now persisted in storage and re-used at a later stage to ensure idempotency

### DIFF
--- a/src/models/event.js
+++ b/src/models/event.js
@@ -1098,7 +1098,6 @@ utils.extend(MatrixEvent.prototype, {
             origin_server_ts: this.getTs(),
             unsigned: this.getUnsigned(),
             room_id: this.getRoomId(),
-            txn_id: this.getTxnId(),
         };
 
         // if this is a redaction then attach the redacts key
--- a/src/models/event.js
+++ b/src/models/event.js
@@ -168,7 +168,7 @@ export const MatrixEvent = function(
     /* The txnId with which this event was sent if it was during this session,
        allows for a unique ID which does not change when the event comes back down sync.
      */
-    this._txnId = null;
+    this._txnId = event.txn_id || null;
 
     /* Set an approximate timestamp for the event relative the local clock.
      * This will inherently be approximate because it doesn't take into account
@@ -1098,6 +1098,7 @@ utils.extend(MatrixEvent.prototype, {
             origin_server_ts: this.getTs(),
             unsigned: this.getUnsigned(),
             room_id: this.getRoomId(),
+            txn_id: this.getTxnId(),
         };
 
         // if this is a redaction then attach the redacts key
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -195,9 +195,10 @@ export function Room(roomId, client, myUserId, opts) {
         if (serializedPendingEventList) {
             JSON.parse(serializedPendingEventList)
                 .forEach(serializedEvent => {
+                    const txnId = client.makeTxnId();
                     const event = new MatrixEvent(serializedEvent);
                     event.setStatus(EventStatus.NOT_SENT);
-                    const txnId = client.makeTxnId();
+                    event.setTxnId(txnId);
                     this.addPendingEvent(event, txnId);
                 });
         }
@@ -1471,9 +1472,6 @@ Room.prototype.updatePendingEvent = function(event, newStatus, newEventId) {
         // timeline map.
         for (let i = 0; i < this._timelineSets.length; i++) {
             this._timelineSets[i].replaceEventId(oldEventId, newEventId);
-        }
-        if (this._opts.pendingEventOrdering === "detached") {
-            this.removePendingEvent(event.event.event_id);
         }
     } else if (newStatus == EventStatus.CANCELLED) {
         // remove it from the pending event list, or the timeline.
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -1303,6 +1303,19 @@ Room.prototype.addPendingEvent = function(event, txnId) {
     this.emit("Room.localEchoUpdated", event, this, null, null);
 };
 
+/**
+ * Persists all pending events to local storage
+ *
+ * If the current room is encrypted only encrypted events will be persisted
+ * all messages that are not yet encrypted will be discarded
+ *
+ * This is because the flow of EVENT_STATUS transition is
+ * queued => sending => encrypting => sending => sent
+ *
+ * Steps 3 and 4 are skipped for unencrypted room.
+ * It is better to discard an unencrypted message rather than persisting
+ * it locally for everyone to read
+ */
 Room.prototype._savePendingEvents = function() {
     const pendingEvents = this._pendingEventList.map(event => {
         return {
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -195,11 +195,9 @@ export function Room(roomId, client, myUserId, opts) {
         if (serializedPendingEventList) {
             JSON.parse(serializedPendingEventList)
                 .forEach(serializedEvent => {
-                    const txnId = client.makeTxnId();
                     const event = new MatrixEvent(serializedEvent);
                     event.setStatus(EventStatus.NOT_SENT);
-                    event.setTxnId(txnId);
-                    this.addPendingEvent(event, txnId);
+                    this.addPendingEvent(event, event.getTxnId());
                 });
         }
     }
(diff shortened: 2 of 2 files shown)


t3chguy (approved):

src/models/event.js
```diff
@@ -1098,6 +1098,7 @@ utils.extend(MatrixEvent.prototype, {
             origin_server_ts: this.getTs(),
             unsigned: this.getUnsigned(),
             room_id: this.getRoomId(),
+            txn_id: this.getTxnId(),
```
jryans, line 1101: This JSON view (which drives view source features) is meant to "simulate" the format of events flowing through the protocol, so I am a bit unsure about adding this here... It might be useful, but I don't think it matches the format... 🤔

jryans: If the goal of adding this here is purely debugging, then it might be better to extend the view source display in React SDK to show it somewhere outside of the JSON directly.

t3chguy: it isn't for debugging though, this JSON representation is what is serialised for persistence.

jryans: Well, it wasn't intended for persistence when I first added it... 😬

t3chguy: Well it being used for such as of https://github.com/matrix-org/matrix-js-sdk/pull/1655/files#diff-0b31693bb53246dda6511feb2fecc466483c08239e007ce35e067eefb13e967cR403

jryans: Maybe we need separate functions for persisting an `Event` model object vs. showing a view source formatted view? Previously we were cheating, because the `toJSON` was never used for persistence like this until https://github.com/matrix-org/matrix-js-sdk/pull/1655.

jryans: The `toJSON` function also currently "makes up" whole extra blocks with `decrypted` and `encrypted` fields for all encrypted events... I don't think that's what we'd want to parse by into the event model when restoring it.

t3chguy: nor would we really want to persist in localStorage decrypted events

gsouquet: Thank you for the comments here! I'll look into that part a bit more thoroughly. 
Totally missed the "only for debugging" comment. I'll see what is usually done in the rest of the app and try to mimic that

gsouquet:
@t3chguy / @jryans thank you for your review earlier.

I have undone my changes to `event.toJSON` to avoid changing the "View Source" payload.
In addition to that the messages persisted to `localStorage` respect the "encryption" setting set by the room they were sent in

Using the spread operator I can persist all the enumerable properties on the event and add the `txnId`. Failing to do that rings all alarm bells about a potential replay attack

src/models/room.js
```diff
@@ -1311,6 +1302,28 @@ Room.prototype.addPendingEvent = function(event, txnId) {
 
     this.emit("Room.localEchoUpdated", event, this, null, null);
 };
+
+Room.prototype._savePendingEvents = async function() {
+    const pendingEvents = await Promise.all(
+        this._pendingEventList.map(async (event) => {
```
jryans, line 1308: Doesn't seem to need `async`?

src/models/room.js
```diff
@@ -1311,6 +1302,28 @@ Room.prototype.addPendingEvent = function(event, txnId) {
 
     this.emit("Room.localEchoUpdated", event, this, null, null);
 };
+
+Room.prototype._savePendingEvents = async function() {
+    const pendingEvents = await Promise.all(
```
jryans, line 1307: Do we need this `await`/ `Promise` here? It seems like the inner part is sync...

src/models/room.js
```diff
@@ -194,11 +194,13 @@ export function Room(roomId, client, myUserId, opts) {
         const serializedPendingEventList = client._sessionStore.store.getItem(pendingEventsKey(this.roomId));
         if (serializedPendingEventList) {
             JSON.parse(serializedPendingEventList)
-                .forEach(serializedEvent => {
+                .forEach(async serializedEvent => {
                     const event = new MatrixEvent(serializedEvent);
+                    if (event.getType() === "m.room.encrypted") {
+                        await event.attemptDecryption(this._client._crypto);
+                    }
                     event.setStatus(EventStatus.NOT_SENT);
```
t3chguy, line 202: what if the pendingEvent hasn't yet been encrypted, e.g the app got restarted mid-encryption?

src/models/room.js
gsouquet: You're absolutely right, some left over when I was attempting to encrypt the messages there.

src/models/room.js
gsouquet: If that is the case at the moment my code will have an unencrypted version persisted in the local storage which is far from ideal. Thank you for pointing that out

I'm going to filter out unencrypted messages from ever being persisted to local storage

src/models/room.js
```diff
@@ -1303,15 +1303,19 @@ Room.prototype.addPendingEvent = function(event, txnId) {
...
-    );
+Room.prototype._savePendingEvents = function() {
+    const pendingEvents = this._pendingEventList.map(event => {
+        return {
+            ...event.event,
+            txn_id: event.getTxnId(),
+        };
+    }).filter(event => {
+        // Filter out the unencrypted messages if the room is encrypted
+        const isEventEncrypted = event.getType !== "m.room.encrypted";
```
t3chguy, line 1314: presumably you want `getType()` here - but not sure about this line in general, event `isEncrypted` has the opposite check

src/models/room.js
```diff
@@ -1303,15 +1303,19 @@ Room.prototype.addPendingEvent = function(event, txnId) {
...
+        const isRoomEncrypted = this._client.isRoomEncrypted(this.roomId);
+
+        return !isRoomEncrypted || (isRoomEncrypted && isEventEncrypted);
```
t3chguy, line 1317: can't this be simplified to `return isEventEncrypted || !isRoomEncrypted;`

src/models/room.js
```diff
@@ -1312,6 +1302,31 @@ Room.prototype.addPendingEvent = function(event, txnId) {
...
 };
+
+Room.prototype._savePendingEvents = function() {
+    const pendingEvents = this._pendingEventList.map(event => {
+        return {
+            ...event.event,
+            txn_id: event.getTxnId(),
+        };
+    }).filter(event => {
+        // Filter out the unencrypted messages if the room is encrypted
```
jryans, line 1313: Hmm, so why is this filtering what we want? I think I am just a little unsure what the expected behaviour for this persistence is in an encrypted room... I guess we could have both unencrypted and encrypted events queued up for an encrypted room, depending on whether we encrypted them before persisting...? Maybe we need a few comments on what's expected...

gsouquet: Fair enough, 
So at this point this is my understanding of how things work
- In an unencrypted room, all messages are unencrypted
- In an encrypted room, all messages are encrypted

The flow in that `Room` model is a bit odd as a message, will have the following flow, right after a user hits send:
1. `EVENT_STATUS=sending` (message unencrypted)
2. `EVENT_STATUS=encrypting` (message unencrypted)
3. `EVENT_STATUS=sending` (message encrypted)
4. `EVENT_STATUS=sent`

on an unencrypted room, the steps 2 and 3 are skipped.
It is a bit difficult in that model to know at which stage we are as `EVENT_STATUS` in step 1 and 3 are the same.
It seems like a sensible thing to do to not persist events in `localStorage` if they are unencrypted in a room that requires encryption (the use case that @t3chguy raised)

So I can definitely add comments in code to highlight my reasoning, but please do let me know if I'm missing a key part of the logic

jryans: Hmm okay, right, that helps me follow the context a bit better. Seems good indeed to do this kind of filtering.

jryans (approved):
Perhaps sprinkle in some comments on encryption expectations, but otherwise seems good to me. 😄

PR #1666, opened 2021-04-15
Discard fallback body for HTML messages, and generate a spoiler-free plaintext representation to use locally in the UI

xenofem:
This fixes vector-im/element-web#12034 , I'm unclear on how to reproduce vector-im/element-web#14447 but I'm fairly sure this will fix that issue as well.

I've deliberately done the HTML-to-text conversion in a very fast-and-loose way, since the text version is only needed in a few limited contexts and doesn't need to do much more than give a general gist of the message content. If you'd rather retain a bit more formatting information, perhaps in a markdown-like syntax, I can adjust things.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "bs58": "^4.0.1",
     "content-type": "^1.0.4",
     "loglevel": "^1.7.1",
+    "parse5": "^6.0.1",
     "qs": "^6.9.6",
     "request": "^2.88.2",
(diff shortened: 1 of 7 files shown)


xenofem:
Retracting this PR, the `textForMessageEvent` function in matrix-react-sdk is a better place to do this.

gabrc52:
Did you make a PR in matrix-react-sdk?

PR #1667, opened 2021-04-15
Fix failure to lookup timeline in handleRemoteEcho

hannojg:
### Reproduction / Use case

I was trying to implement the "sending of voice message". I am:
* (creating a voice record)
* calling `room.addPendingEvent`, so a local message with the state being "sending" is shown to the user (UX: immediate feedback after stopping the recording). I create a dummy event with a txnId.
* uploading the voice record file
* sending an audio message using the txnId of the dummy event (and the content URL received from the upload)

### The issue

The problem was that once the audio message has been sent it would appear as a new message in chat, not updating the local echo (e.g. removing it).

### What I find out

The local echo event will be updated earlier when the message is being marked as "sent". The code `handleRemoteEcho` function was just looking up using the old event id (and handleRemoteEcho has been called with the old event id being the txnId, although the local event's id has already been updated).

With this fix, this doesn't happen. I can imagine that this might happens in other cases as well.
I am just surfacing the sdk's code, maybe this is just a workaround for a symptom, and did not fix the root cause? 

_Side note:_ There is a different issue that prevents the above-described flow/reproduction from working. The fix is in [this](https://github.com/matrix-org/matrix-js-sdk/pull/1668) PR.

### DIFF
--- a/src/models/event-timeline-set.js
+++ b/src/models/event-timeline-set.js
@@ -576,7 +576,7 @@ EventTimelineSet.prototype.addEventToTimeline = function(event, timeline,
 EventTimelineSet.prototype.handleRemoteEcho = function(localEvent, oldEventId,
                                                         newEventId) {
     // XXX: why don't we infer newEventId from localEvent?
-    const existingTimeline = this._eventIdToTimeline[oldEventId];
+    const existingTimeline = this._eventIdToTimeline[oldEventId] || this._eventIdToTimeline[newEventId];
     if (existingTimeline) {
         delete this._eventIdToTimeline[oldEventId];
         this._eventIdToTimeline[newEventId] = existingTimeline;


SimonBrandner:
@hannojg, thanks for working on this! Could you please include a sign-off as described [here](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.rst#sign-off)?

hannojg:
done

src/models/event-timeline-set.js
```diff
@@ -576,7 +576,7 @@ EventTimelineSet.prototype.addEventToTimeline = function(event, timeline,
 EventTimelineSet.prototype.handleRemoteEcho = function(localEvent, oldEventId,
                                                         newEventId) {
     // XXX: why don't we infer newEventId from localEvent?
-    const existingTimeline = this._eventIdToTimeline[oldEventId];
+    const existingTimeline = this._eventIdToTimeline[oldEventId] || this._eventIdToTimeline[newEventId];
     if (existingTimeline) {
         delete this._eventIdToTimeline[oldEventId];
         this._eventIdToTimeline[newEventId] = existingTimeline;
```
hannojg, line 582: hmm, in the light of the new day I just realized that in case `this._eventIdToTimeline[newEventId]` we would just need to do nothing. Like, the delete operation won't do anything & L582 doesn't change anything either. Maybe it would be better to handle this case separately?

jryans:
For future PRs, please use title case headlines, as they currently go directly into our changelog. I'll fix these up.

turt2live (commented):
Like on 1668: The changes are a bit hard to review without seeing the code that is leading to this problem. Is that code somewhere public we can see to aid review?

turt2live:
@hannojg are you able to take a look at this?

hannojg:
No, sorry for letting this go stale - we don't use matrix in production anymore

t3chguy:
Closing due to being abandoned; given it has been signed off, another contributor is welcome to cherry-pick the commits over and continue

PR #1668, opened 2021-04-15
Fix sending when pending event exists

hannojg:
### Reproduction / Use case

I was trying to implement the "sending of voice message". I am:

* (creating a voice record)
* calling room.addPendingEvent, so a local message with the state being "sending" is shown to the user (UX: immediate feedback after stopping the recording). I create a dummy event with a txnId.
* uploading the voice record file
* sending an audio message using the txnId of the dummy event (and the content URL received from the upload)

### The issue

When sending the audio message event with the same txnId of the event we added earlier as pending the sending of the whole event will fail, as (the changed code line) will throw that the event already exists. It's fine that it exists and it should proceed.

_Side note:_ There is a different issue that prevents the above-described flow/reproduction from working. The fix is in [this](https://github.com/matrix-org/matrix-js-sdk/pull/1667) PR.

Signed-off-by: Hanno Gödecke <hgoedecke@cuvent.com>

---

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -2857,7 +2857,9 @@ MatrixClient.prototype._sendCompleteEvent = function(roomId, eventObject, txnId,
         try {
             room.addPendingEvent(localEvent, txnId);
         } catch (e) {
-            logger.warn(`_sendCompleteEvent adding pending event failed because of: ${e ? e.message : "unknown reason"}`); 
+            logger.warn(
+                `_sendCompleteEvent adding pending event failed because of: ${e ? e.message : "unknown reason"}`
+            );
         }
     }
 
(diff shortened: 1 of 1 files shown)


SimonBrandner:
@hannojg, thanks for working on this! Could you please include a sign-off as described [here](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.rst#sign-off)?

hannojg:
did something changed? 😅 not sure if I did it correctly

SimonBrandner:
I don't think anything did. You can just add this to your PR description:
```
Signed-off-by: Your Name <your@email.example.org>
```

turt2live (commented):
The changes are a bit hard to review without seeing the code that is leading to this problem. Is that code somewhere public we can see to aid review?

hannojg:
no, but i can create a local reproduction if that helps?

turt2live:
A minimal case would help too, yea

t3chguy:
Closing as abandoned

PR #1670, opened 2021-04-18
Fix `/search` with no results field work again

timokoesters:
Also see https://github.com/vector-im/element-web/issues/15484

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -4799,7 +4799,8 @@ MatrixClient.prototype._processRoomEventsSearch = function(searchResults, respon
     searchResults.highlights = Object.keys(highlights);
 
     // append the new results to our existing results
-    for (let i = 0; i < room_events.results.length; i++) {
+    const resultsLength = room_events.results ? room_events.results.length : 0;
+    for (let i = 0; i < resultsLength; i++) {
         const sr = SearchResult.fromJson(room_events.results[i], this.getEventMapper());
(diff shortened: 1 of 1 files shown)


timokoesters:
There might be more problems, I'm testing it right now

timokoesters:
Okay, should be good now

timokoesters:
Signed-off-by: Timo Kösters <timo@koesters.xyz>

src/client.js
```diff
@@ -4799,7 +4799,7 @@ MatrixClient.prototype._processRoomEventsSearch = function(searchResults, respon
     searchResults.highlights = Object.keys(highlights);
 
     // append the new results to our existing results
-    for (let i = 0; i < room_events.results.length; i++) {
+    for (let i = 0; i < (room_events.results ? room_events.results.length : 0); i++) {
```
jryans, line 4802: Could you extract this to a `resultsLength` or similar constant above the loop? Maybe personal style, but I think it would be nice to keep a ternary out of the loop header itself.

jryans (changes requested):
Makes sense to me, just a tiny tweak requested.

src/client.js
timokoesters: Thanks for the suggestion. I implemented it now.

jryans (approved):
Thanks! 😄

PR #1673, opened 2021-04-18
Rework how disambiguation is handled

SimonBrandner:
This adds a `disambiguate` property to the `RoomMember`. This is useful when a client wants to handle more things on its own. For example, a client may wish to format the `displayName` differently than the `userId` and therefore has to retrieve the `displayName` and `userName` separately - it can't rely on `js-sdk`'s calculation of display names.


src/models/room-member.js
```diff
@@ -45,6 +45,7 @@ import * as utils from "../utils";
  * @prop {string} membership The membership state for this room member e.g. 'join'.
  * @prop {Object} events The events describing this RoomMember.
  * @prop {MatrixEvent} events.member The m.room.member event for this RoomMember.
+ * @prop {boolean} disambiguate True if the client show disambiguate display names
```
ShadowJonathan, line 48:
```suggestion
 * @prop {boolean} disambiguate True if the client should disambiguate display names
```

SimonBrandner: Thanks for catching that!

src/models/room-member.js
```diff
@@ -45,6 +45,7 @@ import * as utils from "../utils";
  * @prop {string} membership The membership state for this room member e.g. 'join'.
  * @prop {Object} events The events describing this RoomMember.
  * @prop {MatrixEvent} events.member The m.room.member event for this RoomMember.
+ * @prop {boolean} disambiguate True if the client should disambiguate display names
```
turt2live, line 48: This is an indicator flag and not something that can be set external to the room member, so we should word it as such.
```suggestion
 * @prop {boolean} disambiguate True if the member's name is disambiguated.
```

turt2live (approved):
otherwise lgtm, thanks!

turt2live (approved):
looks good - we'll probably land this at the same time as the react-sdk, just in case.

SimonBrandner:
New PR at: https://github.com/matrix-org/matrix-js-sdk/pull/1730

PR #1674, opened 2021-04-19
Support MSC3086 asserted identity

dbkr:
Adds support for receiving MSC3086 Asserted Identity events & emitting event from the call object.

Also moves the method for creating new calls to the client so it can be stubbed out when testing applications using the js-sdk (the client was passed in before so I'm not sure what the benefit was of it being separate).


jryans (approved):
Thanks, great to have tests for this new work! 😄

PR #1675, opened 2021-04-20
Restrict event emit for room members that had power levels changed

gsouquet:
Fixes vector-im/element-web#16967

The `RoomState.members` event was propagated for every single room member even though their power level was not impacted.

This was tested on a room on a local synapse instance with 2000 members
The frame takes **90% less time** to process you can see in the two recording screenshots below

# 🐌 Before

Screen Shot 2021-04-20 at 11 26 27

# 🐎 After

Screen Shot 2021-04-20 at 11 25 41

### DIFF
--- a/src/models/room-state.js
+++ b/src/models/room-state.js
@@ -351,8 +351,14 @@ RoomState.prototype.setStateEvents = function(stateEvents) {
         } else if (event.getType() === "m.room.power_levels") {
             const members = utils.values(self.members);
             utils.forEach(members, function(member) {
+                // We only propagate `RoomState.members` event if the
+                // power levels has been changed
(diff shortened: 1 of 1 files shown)


t3chguy (approved):
This looks like a nice shortcut!

Thanks

t3chguy:
Hmm, thinking about this some more, this might break some things, now it only fires if the member's PL or the room's normal PL changes.
But if PL to do XYZ changes and the user could not perform XYZ but now can then it will not fire, things may be relying on this.

> Fires whenever a member in the members dictionary is updated in any way.

This is a bit open ended, in any case maybe this qualifies for better documentation and a major release cut as a breaking change.

PR #1679, opened 2021-04-30
Switch from MSC1772 unstable prefixes to stable

t3chguy:
~~Blocked on FCP https://github.com/matrix-org/matrix-doc/pull/1772~~

This will render incompatibility for existing spaces


jryans (approved):
Hooray! 🎉

PR #1684, opened 2021-05-06
Prioritise and reduce the amount of events decrypted on application startup

gsouquet:
To review alongside matrix-org/matrix-react-sdk#5980

When loading element, it will either read from cache the entire `/sync` blob that has been persisted or will query that information from Synapse.
The idea is to implement a priority queue that will cap the maximum of events that will be decrypted on application startup. Everything else will be lazily decrypted when a user views a room

At the moment, from clicking on a tile to having all messages displayed on screen it takes roughly a second. With lazy decryption it will approximately take 2 seconds but the most recent messages will start to appear on screen from the first second.

# Critical events

This pull request brings the concept of critical events. These are the events that a client requires to have decrypted to ensure it can run and provide all the required features to the user

At the moment a critical event is an event that falls under the following category
* Last event of a room (to ensure previews are working)
* All events that occured after the read marker (to ensure correct-ness in notification count)

Let me know if you can think about anything else that would benefit from being decrypted ahead of time

In addition to that, I am using breadcrumb rooms to detect which rooms a user might be likely to interact with again in the future and proactively decrypt the live timeline. 
This could be scrapped if we wanted to further improve app start performance at the cost of a few layout shift as highlighted in [this video](https://github.com/matrix-org/matrix-js-sdk/pull/1684#issuecomment-833583200)

# Benchmark

| # of e2ee rooms  | load time (before) | load time (after) | % difference | 
| ---------------- | ------------------ | ----------------- | ------------ |
| 250 | 4mn50 | 31s | **89%** |
| 43 | 37s | 12s | **67%** |

These recordings are taken on an existing session by looking at the time between a hard refresh until the console stops spamming "decrypted events" messages by the thousands.

Please share with me your recording and I will add them to the table above, the code is [up and running on a netlify preview](https://deploy-preview-5980--matrix-react-sdk.netlify.app/)
If you want to know the number of encrypted rooms in your account you can run the following snippet in your terminal

```javascript
Object.values(mxMatrixClientPeg.get().store.rooms).filter(room => {
  return room._client.isRoomEncrypted(room.roomId)
}).length;
```

### DIFF
--- a/spec/test-utils.js
+++ b/spec/test-utils.js
@@ -212,6 +212,9 @@ MockStorageApi.prototype = {
  * @returns {Promise} promise which resolves (to `event`) when the event has been decrypted
  */
 export function awaitDecryption(event) {
+    // An event is not always decrypted ahead of time
+    // getClearContent is a good signal to know whether an event has been decrypted
+    // already
     if (event.getClearContent() !== null) {
         return event;
     } else {
--- a/spec/test-utils.js
+++ b/spec/test-utils.js
@@ -212,18 +212,18 @@ MockStorageApi.prototype = {
  * @returns {Promise} promise which resolves (to `event`) when the event has been decrypted
  */
 export function awaitDecryption(event) {
-    if (!event.isBeingDecrypted()) {
-        return Promise.resolve(event);
-    }
+    if (event.getClearContent() !== null) {
+        return event;
+    } else {
+        logger.log(`${Date.now()} event ${event.getId()} is being decrypted; waiting`);
 
-    logger.log(`${Date.now()} event ${event.getId()} is being decrypted; waiting`);
-
-    return new Promise((resolve, reject) => {
-        event.once('Event.decrypted', (ev) => {
-            logger.log(`${Date.now()} event ${event.getId()} now decrypted`);
-            resolve(ev);
+        return new Promise((resolve, reject) => {
+            event.once('Event.decrypted', (ev) => {
+                logger.log(`${Date.now()} event ${event.getId()} now decrypted`);
+                resolve(ev);
+            });
         });
-    });
+    }
 }
 
 
--- a/src/models/room.js
+++ b/src/models/room.js
@@ -228,18 +228,32 @@ function pendingEventsKey(roomId) {
 
 utils.inherits(Room, EventEmitter);
 
-Room.prototype.lazyDecryptEvents = async function() {
-    return Promise.allSettled(this
+Room.prototype.decryptCriticalEvents = function() {
+    const readReceiptEventId = this.getEventReadUpTo(this._client.getUserId(), true);
+    const events = this.getLiveTimeline().getEvents();
+    const readReceiptTimelineIndex = events.findIndex(matrixEvent => {
+        return matrixEvent.event.event_id === readReceiptEventId
+    });
+
+    const decryptionPromises = events
+        .slice(readReceiptTimelineIndex)
+        .filter(event => event.isEncrypted() && !event.isBeingDecrypted())
+        .reverse()
+        .map(event => event.attemptDecryption(this._client._crypto, true));
+
+    return Promise.allSettled(decryptionPromises);
+}
+
+Room.prototype.decryptAllEvents = function() {
+    const decryptionPromises = this
         .getUnfilteredTimelineSet()
-        .getTimelines()
-        .reduce((decryptionPromises, timeline) => {
-            return decryptionPromises.concat(
-                timeline
-                    .getEvents()
-                    .filter(event => event.isEncrypted() && !event.isBeingDecrypted())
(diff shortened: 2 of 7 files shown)


src/sync.js
```diff
@@ -701,7 +701,31 @@ SyncApi.prototype._syncFromCache = async function(savedSync) {
...
 
     try {
-        await this._processSyncResponse(syncEventData, data);
+        await this._processSyncResponse(syncEventData, data, false);
+
+        const breadcrumbs = this.client.store.getAccountData("im.vector.setting.breadcrumbs");
+        const breadcrumbsRooms = breadcrumbs?.getContent().recent_rooms || [];
+
+        this.client.getRooms().forEach(room => {
+            const readReceipt = room.getAccountData("m.fully_read");
```
dbkr, line 710: this is the read marker rather than the read receipt and it's the read receipt that counts for clearing notifications so may want the other one

dbkr:
in general this looks like a good approach I think

src/sync.js
```diff
@@ -701,7 +701,31 @@ SyncApi.prototype._syncFromCache = async function(savedSync) {
     };
 
     try {
-        await this._processSyncResponse(syncEventData, data);
+        await this._processSyncResponse(syncEventData, data, false);
+
+        const breadcrumbs = this.client.store.getAccountData("im.vector.setting.breadcrumbs");
```
jryans, line 706: Breadcrumbs are an Element Web specific feature that JS SDK shouldn't have to concern itself with directly, so we may need some option / callback / hook that handles this a bit differently. Perhaps there's an optional function the app can pass to SDK which gives back a generic array of "eagerly decrypted rooms"...? And then in React SDK, we would implement that to look at the breadcrumbs.

jryans: Or instead, maybe the React SDK just does this work itself after some `sync` event is emitted from JS SDK, since none of this seems specific to the data being manipulated as part of the sync action.

jryans (commented):
Overall, this is looking reasonable so far. Do you have any data on how much performance improves with this change in isolation?

gsouquet:
> Do you have any data on how much performance improves with this change in isolation?

I have just done a benchmark using the same account. It takes down the load time from **4 minutes 50 seconds** down to **30 seconds**. During which the application is still usable, but with a degraded experience (sometimes the room spinner is missing, ...)

And below is a screen recording of the experience when we load a room where we need to lazily decrypt events


https://user-images.githubusercontent.com/769871/117318458-56ca1d80-ae82-11eb-9141-4c5eb226832e.mov

src/models/room.js
```diff
@@ -228,6 +228,20 @@ function pendingEventsKey(roomId) {
 
 utils.inherits(Room, EventEmitter);
 
+Room.prototype.lazyDecryptEvents = async function() {
```
jryans, line 231: Perhaps just call it `decryptEvents`...? There's nothing inherently "lazy" about the function body that I can see, it's more about the timing of when it is called, right?

```suggestion
Room.prototype.decryptEvents = async function() {
```

src/models/room.js
```diff
@@ -228,6 +228,34 @@ function pendingEventsKey(roomId) {
 
 utils.inherits(Room, EventEmitter);
 
+Room.prototype.decryptCriticalEvents = function() {
```
jryans, line 231: It would be good to have a comment describing the meaning of "critical" that we're aiming for here, so future readers know what properties to maintain.

jryans (commented):
Looking good overall to me! 😄

jryans:
Ah yeah, as mentioned before, it would be good to have some kind of automated to verify rooms still decrypt as expected, either here or in the React layer (whichever makes more sense).

turt2live (commented):
ftr, this seems to give about a 15x improvement for decrypting the important rooms on my account. Clicking on a room and having it pop in at that second is amazing 😍

I have profiles I can share privately of before and after if people are interested (though you need *a ton* of CPU to load them).

spec/test-utils.js
```diff
@@ -212,18 +212,18 @@ MockStorageApi.prototype = {
  * @returns {Promise} promise which resolves (to `event`) when the event has been decrypted
  */
 export function awaitDecryption(event) {
-    if (!event.isBeingDecrypted()) {
-        return Promise.resolve(event);
-    }
-
-    logger.log(`${Date.now()} event ${event.getId()} is being decrypted; waiting`);
+    if (event.getClearContent() !== null) {
```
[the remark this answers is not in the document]
gsouquet: Because this PR does not proactively trigger decryption for all events. 
It is not possible that an event is encrypted and not in the process of being decrypted.

`getClearContent` is the function that returns the decrypted content of an event. I'll add some code comment to clarify

turt2live (approved):
thanks :)

PR #1688, opened 2021-05-11
👕 New linting rules

gsouquet:
To apply when matrix-org/eslint-plugin-matrix-org#6 gets merged into `main`


jryans:
Hmm, we probably should have just merged this right away, as there were no conflicts under the previous rule config and it will just accumulate conflicts... In any case, @gsouquet can you resolve the conflicts and merge?

PR #1693, opened 2021-05-13
Use native Object and Array methods

gsouquet:

A lot of `utils` were once very useful for cross browser compatibility. However vendors have come a long way, and every methods listed below are now supported

Replacing them with their native counterpart shaves off 10% of my `_syncFromCache` call.


t3chguy (approved):
🎉

PR #1696, opened 2021-05-17
Decrypt relations before applying them to target event

gsouquet:
Fixes vector-im/element-web#17290
Regressed by matrix-org/matrix-js-sdk#1684

The code was expecting events to already be decrypted when it reaches that block of code. I have added a bit of logic to wait for relations to be decrypted before it mutates the target event

### DIFF
--- a/src/webrtc/callEventHandler.ts
+++ b/src/webrtc/callEventHandler.ts
@@ -55,7 +55,9 @@ export class CallEventHandler {
     private evaluateEventBuffer = () => {
         if (this.client.getSyncState() === "SYNCING") {
             // don't process any events until they are all decrypted
-            if (this.callEventBuffer.some((e) => e.isBeingDecrypted())) return;
+            if (this.callEventBuffer.some((e) => {
+                return e.isBeingDecrypted() || e.shouldAttemptDecryption()
+            })) return;
 
             const ignoreCallIds = new Set<String>();
             // inspect the buffer and mark all calls which have been answered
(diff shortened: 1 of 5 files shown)


jryans:
Is there a similar issue with aggregations generally in this block as well?

https://github.com/matrix-org/matrix-js-sdk/blob/2d5857f1454b6292b26af7cb219b635d517e71ea/src/models/event-timeline-set.js#L772-L777

It seems like it might assume that every event is either already decrypted or in the process of being decrypted... This variant might affect reactions more than edits.

t3chguy:
The signature changes on public methods might mandate a minor version change given the backwards incompatibility of 3 sync methods going async

gsouquet:
> Is there a similar issue with aggregations generally in this block as well?
> 
> https://github.com/matrix-org/matrix-js-sdk/blob/2d5857f1454b6292b26af7cb219b635d517e71ea/src/models/event-timeline-set.js#L772-L777
> 
> It seems like it might assume that every event is either already decrypted or in the process of being decrypted... This variant might affect reactions more than edits.

That makes a lot of sense! I've added an update to ensure that it sets the listener is the event is not yet decrypted too


> The signature changes on public methods might mandate a minor version change given the backwards incompatibility of 3 sync methods going async

Yes! Sorry I forgot what's the correct course of actions to notify @jryans of which version increment he should use

jryans:
@gsouquet I wonder if we need examine effectively all call sites that were checking `isBeingDecrypted` and / or adding a listener for `Event.decrypted`...? For example:

https://github.com/matrix-org/matrix-js-sdk/blob/52a893a8116d60bb76f1b015d3161a15404b3628/src/webrtc/callEventHandler.ts#L56-L58

https://github.com/matrix-org/matrix-js-sdk/blob/52a893a8116d60bb76f1b015d3161a15404b3628/src/webrtc/callEventHandler.ts#L90-L102

There are various other call sites too. I am a bit worried there are various operations which are not being scheduled any more at the moment.

gsouquet:
Fair enough, I did a project wide search on `Event.decrypted` but did not go as far as looking at `isBeingDecrypted`.
I'll take a closer look before moving this forward

src/webrtc/callEventHandler.ts
```diff
@@ -87,17 +89,19 @@ export class CallEventHandler {
...
-            || event.isBeingDecrypted()
+            || isBeingDecrypted || shouldAttemptDecryption
         ) {
             // queue up for processing once all events from this sync have been
             // processed (see above).
             this.callEventBuffer.push(event);
         }
 
-        if (event.isBeingDecrypted() || event.isDecryptionFailure()) {
+        if (event.isDecryptionFailure() || isBeingDecrypted || shouldAttemptDecryption) {
```
jryans, line 104: So for call events, we're relying on `decryptCriticalEvents` to attempt the decryption, correct? (I think it seems right, I just wanted to confirm my understanding.)

jryans (approved):
Overall this is shaping up nicely, thanks!

I'll track this as a major version change.

jryans:
> I'll track this as a major version change.

Actually, since this only impacts relations, which are both documented as unstable in the SDK and unstable at the spec level, I think we can ignore this for versioning purposes.

PR #1698, opened 2021-05-19
Change call event handlers to adapt to undecrypted events

gsouquet:
Fixes vector-im/element-web#17393
Regressed by matrix-org/matrix-js-sdk#1684

The call event handler was expecting all messages to be decrypted or in the process of being decrypted and would not process any call events until everything has been processed.

The fix is to ignore all historical data by setting the listeners only after initial client sync

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -356,7 +356,13 @@ export function MatrixClient(opts) {
         // Start listening for calls after the initial sync is done
         // We do not need to backfill the call event buffer
         // with encrypted events that might never get decrypted
-        this.once("sync", () => this._callEventHandler.start());
+        function startCallEventHandler() {
+            if (this.isInitialSyncComplete()) {
+                this._callEventHandler.start();
+                this.off("sync", startCallEventHandler);
+            }
+        }
+        this.on("sync", startCallEventHandler);
     } else {
         this._callEventHandler = null;
     }
--- a/src/client.js
+++ b/src/client.js
@@ -353,6 +353,10 @@ export function MatrixClient(opts) {
     if (call) {
         this._callEventHandler = new CallEventHandler(this);
         this._supportsVoip = true;
+        // Start listening for calls after the initial sync is done
+        // We do not need to backfill the call event buffer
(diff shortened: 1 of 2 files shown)


SimonBrandner:
Fixes [#17393](https://github.com/vector-im/element-web/issues/17393)

src/client.js
```diff
@@ -353,6 +353,10 @@ export function MatrixClient(opts) {
     if (call) {
         this._callEventHandler = new CallEventHandler(this);
         this._supportsVoip = true;
+        // Start listening for calls after the initial sync is done
+        // We do not need to backfill the call event buffer
+        // with encrypted events that might never get decrypted
```
dbkr, line 358: I think we do want to start listening before the catchup sync to catch calls placed shortly before the user opened their client, but its probably safe to start listening after the events are loaded from the database, so after the state changes to PREPARED I think?

src/webrtc/callEventHandler.ts
```diff
@@ -85,37 +88,16 @@ export class CallEventHandler {
         }
     }
 
-    private onEvent = (event: MatrixEvent) => {
-        // any call events or ones that might be once they're decrypted
+    private onEvent = async (event: MatrixEvent) => {
+        await this.client.decryptEventIfNeeded(event);
```
dbkr, line 92: So the call state machine relies on the call events arriving in the right order: I think think this will mean that the order they get processed in will be whatever order these promises resolve in, which *should* be the same order they were started in but I'm not really confident since it will depend on the js engine's microtask and indexeddb operation scheduling. Now we can just `await` for an event to be decrypted, we could probably just shove events into the buffer whether they're encrypted or not, then await on that promise serially for each event as we work our way through the buffer.

dbkr (approved):
lgtm!

SimonBrandner:
I suppose this should be sent to release, right?

jryans:
> I suppose this should be sent to release, right?

Yes, I'll work on that now.

PR #1700, opened 2021-05-19
Only log once if a Room lacks an m.room.create event

ara4n:
Currently my account is logging ~20,000 warnings a minute about getType() being called
on invites which have no m.room.create events, which crashes my inspector and makes
my console unusable.  We still want the logline though, as it helps diagnose
space invite problems.  So instead, log this only once.  (Untested).


jryans (approved):
Looks reasonable to me, thanks!

PR #1701, opened 2021-05-20
Cache normalized room name

gsouquet:
Fixes vector-im/element-web#17370

Approximatly divides the function execution time by three. Here is the difference of the two call stacks

# Before 🐌 

Screen Shot 2021-05-20 at 10 42 26

# After 🐎 

Screen Shot 2021-05-20 at 10 26 33

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -347,7 +347,7 @@ export function removeHiddenChars(str: string): string {
 }
 
 export function normalize(str: string): string {
-    // Note: we have to match the filter with the removeHiddenChars() room name because the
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -346,6 +346,16 @@ export function removeHiddenChars(str: string): string {
     return "";
 }
 
+export function normalize(str: string): string {
+    // Note: we have to match the filter with the removeHiddenChars() room name because the
```
t3chguy, line 350: this probably ought not to talk about room names given its a util

t3chguy (approved):
:shipit:

PR #1712, opened 2021-05-27
Make consistent call event type checks

gsouquet:
Fixes vector-im/element-web#17393

This issue has been spotted when transfering calls which uses asserted identity.


jryans (approved):
Thanks, seems reasonable to me! 😄

PR #1714, opened 2021-05-28
Apply new Babel linting config

jryans:
This applies the new Babel linting config layer from https://github.com/matrix-org/eslint-plugin-matrix-org/pull/7.

In addition, the local project TypeScript config is expanded to cover tests as well.

A few linting errors were revealed and fixed.


gsouquet (approved):
🙌 Looks good to me!

PR #1716, opened 2021-05-31
Fix relations getRelationsForEvent create relation

hannojg:
### The problem

Currently, when calling `getRelationsForEvent` on a timeline set it will return undefined when no relation event exists.
In our app, we have the following logic:

*Message component:*

```jsx
const Message = (props) => {
  ...
  useEffect(() => {
    const client = RNMWrapper.getClient();
    const relation = client
      .getRoom(matrixEvent.getRoomId())
      ?.getLiveTimeline()
      .getTimelineSet()
      .getRelationsForEvent(matrixEvent.getId(), "m.moment.seen", "m.reaction");
    relation?.addListener("Relations.add", (e) => {
      // process newly added relation on message
    });
  }, [matrixEvent]);
```

When a relation is added for `matrixEvent` we won't know it, as `relation` is `undefined`.

### The change

I implemented (and wrote tests) that when `getRelationsForEvent` a new relation will be created if none exists. I think this behaviour is fine, because we can then do `relation.getRelations()` which will return an empty array, which is more descriptive than just returning undefined.

However, I am not sure whether this change would be breaking?

Happy to discuss this change with you ✌️

### DIFF
--- a/spec/unit/relations.spec.ts
+++ b/spec/unit/relations.spec.ts
@@ -18,25 +18,39 @@ import { EventTimelineSet } from "../../src/models/event-timeline-set";
 import { MatrixEvent } from "../../src/models/event";
 import { Relations } from "../../src/models/relations";
 
+// Helper
+function createRelationEvent() {
+    return new MatrixEvent({
+        "sender": "@bob:example.com",
+        "type": "m.reaction",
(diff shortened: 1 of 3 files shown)


hannojg:
ah sorry, commit history is not clean. Let me recreate the pr ...

PR #1717, opened 2021-05-31
Fix relations getRelationsForEvent create relation 

hannojg:
### The problem

Currently, when calling `getRelationsForEvent` on a timeline set it will return undefined when no relation event exists.
In our app, we have the following logic:

*Message component:*

```jsx
const Message = (props) => {
  ...
  useEffect(() => {
    const client = RNMWrapper.getClient();
    const relation = client
      .getRoom(matrixEvent.getRoomId())
      ?.getLiveTimeline()
      .getTimelineSet()
      .getRelationsForEvent(matrixEvent.getId(), "m.moment.seen", "m.reaction");
    relation?.addListener("Relations.add", (e) => {
      // process newly added relation on message
    });
  }, [matrixEvent]);
```

When a relation is added for `matrixEvent` we won't know it, as `relation` is `undefined`.

### The change

I implemented (and wrote tests) that when `getRelationsForEvent` a new relation will be created if none exists. I think this behaviour is fine because we can then do `relation.getRelations()` which will return an empty array, which is more descriptive than just returning undefined.

However, I am not sure whether this change would be breaking?

Happy to discuss this change with you ✌️ 

---

## 🐛 Bug Fixes
 * Fix relations getRelationsForEvent create relation  ([\#1717](https://github.com/matrix-org/matrix-js-sdk/pull/1717)). Contributed by @hannojg.

### DIFF
--- a/spec/unit/relations.spec.ts
+++ b/spec/unit/relations.spec.ts
@@ -18,25 +18,39 @@ import { EventTimelineSet } from "../../src/models/event-timeline-set";
 import { MatrixEvent } from "../../src/models/event";
 import { Relations } from "../../src/models/relations";
 
+// Helper
+function createRelationEvent() {
+    return new MatrixEvent({
+        "sender": "@bob:example.com",
+        "type": "m.reaction",
+        "event_id": "$cZ1biX33ENJqIm00ks0W_hgiO_6CHrsAc3ZQrnLeNTw",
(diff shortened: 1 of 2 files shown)


t3chguy (commented):
Can you please resolve the merge conflicts

t3chguy:
Closing as abandoned

PR #1718, opened 2021-06-01
[BREAKING] Convert MatrixClient to TypeScript

turt2live:
***This is considered breaking due to an API contract change (merge of two classes): it is unknown if external consumers will have extended these classes.***

**Accompanied by https://github.com/matrix-org/matrix-react-sdk/pull/6132**

----

**Reviewer**: This is a *massive* diff, and one that doesn't line up with the old source. I would highly suggest doing a kind of review where simple sanity is checked rather than exact function definitions being translated to the right place.

Also, probably take breaks. It took over a week to write this, so will probably take a few hours to review.

This is unlikely to be reviewable commit-by-commit, sorry.

----

The approach for this started with optimism and quickly dwindled to being as much as a straight-through conversion as possible. This means that after the first little bit of the type conversion things started getting `// TODO: Types` instead of interfaces when they weren't plainly obvious. Trying to type *everything* out would likely make this PR literally impossible to review, so I've settled for comments for now. 

This also merges the MatrixBaseApis with MatrixClient given they were always the same thing. Nothing else extended MatrixBaseApis. This also saves us from having to use mixins/multiple `extends`, which is arguably a win. The downside is the MatrixClient class is even longer now.

This includes a brief pass over jsdoc just to ensure the type warnings line up. No realistic effort was made to improve the state of the documentation as a whole.

**Note**: Groups have had the absolute bare minimum amount of effort, even going so far as to *not* document types or put TODOs on them. This is because they're being ripped out sooner than is worth documenting.

TODO:
* [x] Unbreak linter
* [x] Unbreak any tests
* [x] Unbreak build
* [x] 1-2 passes over the documentation (skipped: time)
* [x] 1-2 passes over the code style
* [x] 1-2 passes over obvious types (skipped: time)
* [x] Validate react-sdk happy
* [x] Validate element-web happy
* [x] Validate element-desktop happy

Note: while I have verified that Element Web/Desktop doesn't explode, there's almost certainly areas I missed in manual testing. We should watch for bugs.


jryans:
@turt2live Thanks for working on this big change! 😄 

One thing I've noticed with the current branch is Git seems unable to follow the .js -> .ts rename of `client.js`, so all blame history would be lost, which seems quite unfortunate for such a key file... 😭 Is it possible to add an early commit to the stack that does a simple rename without content changes, and then rebase the work on top of that? Usually there's a higher chance Git can follow along if the renaming vs. large content changes occur in separate commits.

turt2live:
I'll timebox doing that, but have little hope of it working. The conversion was deliberately done without a rename due to complexity of the source material: blame would have been lost regardless. This also merges two files into one, making it even harder for the blame to operate...

turt2live:
@jryans the commit history is a bit fucky, but the commit is there now - does it look okay/different to you? 

Github doesn't count it as a rename due to the massive file diff (+8432 -6057 on client.ts / client.js), but the commit trail seems to count it in my git tree.

jryans:
Thanks, looks like the line history is indeed now preserved across the rename at least locally (where it matters most), so that's great, thanks! 😁

jryans:
I've grabbed this one from the queue explicitly... Will start on it after the sync meeting.

jryans (approved):
Looks great! 😄 Obviously I've only done a quick skim, but this is a great foundation, and I think it would be better to merge sooner (avoiding conflicts) and then improve from there.

turt2live:
*brace for impact*

PR #1723, opened 2021-06-02
Ignore power_levels events with unknown state_key on room-state initialization

### LINKED ISSUE #1722: m.room.power_levels events with unknown state_key are processed by room-state
**Describe the bug**
In order to set the power_levels of a room, you can send a request to `/_matrix/client/r0/rooms/{roomId}/state/m.room.power_levels/{stateKey}`.
If you happen to set a stateKey by mistake, there will be two power_level states existing at the same time in the current state.

I do not see a possibility to remove a state event from the currentState in a room. Redaction does not seem to do the job.

As far as I understood, state events with unknown `state_keys` should just be ignored, by server and by client in the processing steps.
Example for this behaviour found in synapse here:
https://github.com/matrix-org/synapse/blob/0085dc5abc614579f3adbd9e6d2cbdd41facef00/synapse/rest/admin/rooms.py#L487
Example in the js-sdk:
https://github.com/matrix-org/matrix-js-sdk/blob/3a3be36f4c274da2c8233271605f43adeae226e5/src/models/room-state.js#L406

When first setting the state events of a `room-state` model, the state_key will only be checked for existance, not for content.
https://github.com/matrix-org/matrix-js-sdk/blob/3a3be36f4c274da2c8233271605f43adeae226e5/src/models/room-state.js#L351

This leads to the `room.currentState.members` to sometimes have wrong power_levels.

**To Reproduce**
The order in which `room-state` processes `m.room.power_levels` events with different state_keys is important.
I do not know what determines that order, therefore the steps to reproduce are kind of incomplete.

I used synapse as the server to reproduce the bug.

1. Create a room, set someone to owner via `/_matrix/client/r0/rooms/{roomId}/state/m.room.power_levels` (no state_key)
2. Make someone (person B) else Moderator (power_level 50) via /_matrix/client/r0/rooms/{roomId}/state/m.room.power_levels/{stateKey}` with stateKey being some non-empty string
3. Demote this person again using /_matrix/client/r0/rooms/{roomId}/state/m.room.power_levels` (no state_key)

**Expected behavior**
I expect that person B will be projected as demoted with power_level 0 in `room.currentState.members[<person B id>].powerLevel`.

**Experienced behaviour**
person B sometimes has the old power_level of 50.

**Suggestions**

I suggest that we implement one of the following:
* Sort the state_events by some temporal parameter, like a timestamp or streamingId.
    * This will not lead to the power_level events with state_key to be ignored, but then they can be overwritten
* Ignore all power_level events with state_key not equal to an empty string

schmop:
Signed-off-by: Lars Richard <lars.richard@iserv.eu>

Closes #1722

### DIFF
--- a/src/models/room-state.js
+++ b/src/models/room-state.js
@@ -349,6 +349,11 @@ RoomState.prototype.setStateEvents = function(stateEvents) {
             self._updateMember(member);
             self.emit("RoomState.members", event, self, member);
         } else if (event.getType() === "m.room.power_levels") {
+            // events with unknown state keys should be ignored
+            // and should not aggregate onto members power levels
+            if (event.getStateKey() !== "") {
+                return;
+            }
             const members = Object.values(self.members);
             members.forEach(function(member) {
                 // We only propagate `RoomState.members` event if the


turt2live (approved):
Good catch! Thanks for the quick fix.

PR #1727, opened 2021-06-04
Stop breeding sync listeners

t3chguy:


jryans (approved):
Thanks, looks like the correct fix to me! 😄

dbkr (approved):

PR #1732, opened 2021-06-10
Unstable implementation of MSC3089: File Trees

turt2live:
This is a ***heavily opinionated*** set of classes to help downstream projects implement file trees. There is no UI component to this (for now). As an unstable implementation, this has known gaps that will be filled (eventually, probably). 

This has a very specific usecase (managed file trees) and is not entirely ready for usecases like Element or bridges yet - its opinions are formed out of this usecase.

Such opinionated code is a bit of a first in the js-sdk stack, but for this sort of downstream project it helps with development. The downstream project using this is even more opinionated, and it is expected that the before dropping the unstable flag that this implementation would be refactored to match whatever structure the js-sdk has at that time. Please ping me on Matrix for more information/reference material (if appropriate).

**There is no expectation or guarantee that this API surface be maintained as such. Breaking changes are encouraged and expected.**

The NamespacedValue stuff is borrowed from the Extensible Events implementation PR in the react-sdk as a useful thing to put here. When/if the MSCs become stable, we'd convert the values to enums and drop the constants (or keep them for migration purposes, but switch to stable). 

----

MSC: https://github.com/matrix-org/matrix-doc/pull/3089
Includes part of MSC3088 (room subtyping): https://github.com/matrix-org/matrix-doc/pull/3088

*Note for diligent community members*: This does not count as an MSC implementation due to the lack of UI component.

### DIFF
--- a/spec/unit/utils.spec.js
+++ b/spec/unit/utils.spec.js
@@ -1,4 +1,12 @@
 import * as utils from "../../src/utils";
+import {
+    alphabetPad,
+    averageBetweenStrings,
+    baseToString,
+    DEFAULT_ALPHABET,
+    nextString, prevString,
+    stringToBase
+} from "../../src/utils";
 
 describe("utils", function() {
     describe("encodeParams", function() {
@@ -257,6 +265,75 @@ describe("utils", function() {
 
             await utils.chunkPromises([fn1, fn2], 1);
             expect(promiseCount).toEqual(2);
+        });
+    });
+
+    describe('DEFAULT_ALPHABET', () => {
+        it('should be usefully printable ASCII in order', () => {
+            expect(DEFAULT_ALPHABET).toEqual(" !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~");
+        });
+    });
+
+    describe('alphabetPad', () => {
+        it('should pad to the alphabet length', () => {
+            const defaultPrefixFor1char = [""].reduce(() => {
+                let s = "";
+                for (let i = 0; i < DEFAULT_ALPHABET.length - 1; i++) {
+                    s += DEFAULT_ALPHABET[0];
+                }
+                return s;
+            }, "");
+            expect(alphabetPad("a")).toEqual(defaultPrefixFor1char + "a");
+            expect(alphabetPad("a", "123")).toEqual("11a");
+        });
+    });
+
+    describe('baseToString', () => {
+        it('should calculate the appropriate string from numbers', () => {
+            expect(baseToString(10)).toEqual(DEFAULT_ALPHABET[10]);
+            expect(baseToString(10, "abcdefghijklmnopqrstuvwxyz")).toEqual('k');
+            expect(baseToString(6241)).toEqual("ab");
+            expect(baseToString(53, "abcdefghijklmnopqrstuvwxyz")).toEqual('cb');
+        });
+    });
+
+    describe('stringToBase', () => {
+        it('should calculate the appropriate number for a string', () => {
+            expect(stringToBase(" ")).toEqual(0);
+            expect(stringToBase("a", "abcdefghijklmnopqrstuvwxyz")).toEqual(0);
+            expect(stringToBase("a")).toEqual(65);
+            expect(stringToBase("c", "abcdefghijklmnopqrstuvwxyz")).toEqual(2);
+            expect(stringToBase("ab")).toEqual(6241);
+            expect(stringToBase("cb", "abcdefghijklmnopqrstuvwxyz")).toEqual(53);
+        });
+    });
+
+    describe('averageBetweenStrings', () => {
+        it('should average appropriately', () => {
+            expect(averageBetweenStrings('A', 'z')).toEqual(']');
+            expect(averageBetweenStrings('a', 'z', "abcdefghijklmnopqrstuvwxyz")).toEqual('m');
+            expect(averageBetweenStrings('AA', 'zz')).toEqual('^.');
+            expect(averageBetweenStrings('aa', 'zz', "abcdefghijklmnopqrstuvwxyz")).toEqual('mz');
+            expect(averageBetweenStrings('cat', 'doggo')).toEqual("BH65B");
+            expect(averageBetweenStrings('cat', 'doggo', "abcdefghijklmnopqrstuvwxyz")).toEqual("buedq");
+        });
+    });
+
+    describe('nextString', () => {
+        it('should find the next string appropriately', () => {
+            expect(nextString('A')).toEqual('B');
+            expect(nextString('b', 'abcdefghijklmnopqrstuvwxyz')).toEqual('c');
+            expect(nextString('cat')).toEqual('cau');
+            expect(nextString('cat', 'abcdefghijklmnopqrstuvwxyz')).toEqual('cau');
+        });
+    });
+
+    describe('prevString', () => {
+        it('should find the next string appropriately', () => {
+            expect(prevString('B')).toEqual('A');
+            expect(prevString('c', 'abcdefghijklmnopqrstuvwxyz')).toEqual('b');
+            expect(prevString('cau')).toEqual('cat');
+            expect(prevString('cau', 'abcdefghijklmnopqrstuvwxyz')).toEqual('cat');
         });
     });
 });
--- a/spec/unit/utils.spec.js
+++ b/spec/unit/utils.spec.js
@@ -9,6 +9,7 @@ import {
     prevString,
     stringToBase,
 } from "../../src/utils";
+import { logger } from "../../src/logger";
 
 describe("utils", function() {
     describe("encodeParams", function() {
@@ -344,7 +345,7 @@ describe("utils", function() {
             for (let i = 0; i < 100; i++) {
                 const next = nextString(midpoint);
                 const prev = prevString(midpoint);
-                console.log({i, midpoint, next, prev}); // for test debugging
+                logger.log({ i, midpoint, next, prev }); // for test debugging
 
                 expect(lexicographicCompare(midpoint, next) < 0).toBe(true);
                 expect(lexicographicCompare(midpoint, prev) > 0).toBe(true);
--- a/spec/unit/utils.spec.js
+++ b/spec/unit/utils.spec.js
@@ -4,6 +4,7 @@ import {
     averageBetweenStrings,
     baseToString,
     DEFAULT_ALPHABET,
+    lexicographicCompare,
     nextString,
     prevString,
     stringToBase,
@@ -279,15 +280,9 @@ describe("utils", function() {
 
     describe('alphabetPad', () => {
         it('should pad to the alphabet length', () => {
-            const defaultPrefixFor1char = [""].reduce(() => {
-                let s = "";
-                for (let i = 0; i < DEFAULT_ALPHABET.length - 1; i++) {
-                    s += DEFAULT_ALPHABET[0];
-                }
-                return s;
-            }, "");
-            expect(alphabetPad("a")).toEqual(defaultPrefixFor1char + "a");
-            expect(alphabetPad("a", "123")).toEqual("11a");
+            const len = 12;
+            expect(alphabetPad("a", len)).toEqual("a" + ("".padEnd(len - 1, DEFAULT_ALPHABET[0])));
+            expect(alphabetPad("a", len, "123")).toEqual("a" + ("".padEnd(len - 1, '1')));
         });
     });
 
@@ -313,12 +308,12 @@ describe("utils", function() {
 
     describe('averageBetweenStrings', () => {
         it('should average appropriately', () => {
-            expect(averageBetweenStrings('A', 'z')).toEqual(']');
-            expect(averageBetweenStrings('a', 'z', "abcdefghijklmnopqrstuvwxyz")).toEqual('m');
+            expect(averageBetweenStrings('A', 'z')).toEqual('^');
+            expect(averageBetweenStrings('a', 'z', "abcdefghijklmnopqrstuvwxyz")).toEqual('n');
             expect(averageBetweenStrings('AA', 'zz')).toEqual('^.');
-            expect(averageBetweenStrings('aa', 'zz', "abcdefghijklmnopqrstuvwxyz")).toEqual('mz');
-            expect(averageBetweenStrings('cat', 'doggo')).toEqual("BH65B");
-            expect(averageBetweenStrings('cat', 'doggo', "abcdefghijklmnopqrstuvwxyz")).toEqual("buedq");
+            expect(averageBetweenStrings('aa', 'zz', "abcdefghijklmnopqrstuvwxyz")).toEqual('na');
+            expect(averageBetweenStrings('cat', 'doggo')).toEqual("d9>Cw");
+            expect(averageBetweenStrings('cat', 'doggo', "abcdefghijklmnopqrstuvwxyz")).toEqual("cumqh");
         });
     });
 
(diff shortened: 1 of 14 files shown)


src/utils.ts
```diff
@@ -456,3 +456,105 @@ export function setCrypto(c: Object) {
...
+ * Pads a string using the given alphabet as a base. The returned string will be the
+ * same length as the alphabet, and padded with the first character in the alphabet.
+ *
+ * This is intended for use with string averaging.
+ * @param {string} s The string to pad.
+ * @param {string} alphabet The alphabet to use as a single string.
+ * @returns {string} The padded string.
+ */
+export function alphabetPad(s: string, alphabet = DEFAULT_ALPHABET): string {
+    return s.padStart(alphabet.length, alphabet[0]);
```
t3chguy, line 487: padding at the start will cause the wrong behaviour for lexicographic sorting with mismatched string lengths, the sort will treat the strings differently to your `averageBetweenStrings` where `bb` ... `z` will be lexicographically sorted as such but `averageBetweenStrings` will treat `z` as ` z` which will be earlier than `bb`.

src/client.ts
```diff
@@ -7709,6 +7717,74 @@ export class MatrixClient extends EventEmitter {
...
+    public unstableGetFileTreeSpace(roomId: string): MSC3089TreeSpace {
+        const room = this.getRoom(roomId);
+        if (!room) return null;
+
+        const createEvent = room.currentState.getStateEvents(EventType.RoomCreate, "");
+        const purposeEvent = room.currentState.getStateEvents(
+            UNSTABLE_MSC3088_PURPOSE.name,
+            UNSTABLE_MSC3089_TREE_SUBTYPE.name);
+
+        if (!createEvent || Array.isArray(createEvent)) throw new Error("Expected single room create event");
```
t3chguy, line 7779: I don't think the array case is possible given `getStateEvents` only ever returns an array if the 2nd arg is undefined.

src/models/MSC3089TreeSpace.ts
```diff
@@ -0,0 +1,433 @@
...
+        ordered.sort((a, b) => {
+            if (a.order && !b.order) {
+                return -1;
+            } else if (!a.order && b.order) {
+                return 1;
+            } else if (!a.order && !b.order) {
+                const roomA = this.client.getRoom(a.roomId);
+                const roomB = this.client.getRoom(b.roomId);
+                if (!roomA || !roomB) { // just don't bother trying to do more partial sorting
+                    return a.roomId.localeCompare(b.roomId);
```
t3chguy, line 236: reminder that this is stupidly expensive in JS

src/utils.ts
t3chguy: Also padding to alphabet's length (94) is longer than MSC1772 allows (50) so this won't work for ordering space children.

spec/unit/utils.spec.js
```diff
@@ -259,4 +268,75 @@ describe("utils", function() {
...
+    });
+
+    describe('averageBetweenStrings', () => {
+        it('should average appropriately', () => {
+            expect(averageBetweenStrings('A', 'z')).toEqual(']');
+            expect(averageBetweenStrings('a', 'z', "abcdefghijklmnopqrstuvwxyz")).toEqual('m');
+            expect(averageBetweenStrings('AA', 'zz')).toEqual('^.');
+            expect(averageBetweenStrings('aa', 'zz', "abcdefghijklmnopqrstuvwxyz")).toEqual('mz');
+            expect(averageBetweenStrings('cat', 'doggo')).toEqual("BH65B");
+            expect(averageBetweenStrings('cat', 'doggo', "abcdefghijklmnopqrstuvwxyz")).toEqual("buedq");
```
t3chguy, line 321: this is not an expected output, `buedq` does not fall between `cat` and `doggo` lexicographically.

src/utils.ts
```diff
@@ -456,3 +456,105 @@ export function setCrypto(c: Object) {
...
+ * @param {string} alphabet The alphabet to use as a single string.
+ * @returns {number} The baseN number.
+ */
+export function stringToBase(s: string, alphabet = DEFAULT_ALPHABET): number {
+    s = alphabetPad(s, alphabet);
+    const len = alphabet.length;
+    const reversedStr = Array.from(s).reverse();
+    let result = 0;
+    for (let i = 0; i < len; i++) {
+        result += alphabet.indexOf(reversedStr[i]) * (len ** i);
```
t3chguy, line 520: this `indexOf` is a relatively expensive scan, given we know the alphabet array to be sequential, consecutive and ordered we can (should) just use math here.

`reversedStr.charCodeAt(i) - alphabet[0].charCodeAt(0)` or similar

src/models/MSC3089TreeSpace.ts
gsouquet, line 236:
```suggestion
                    return compare(a.roomId, b.roomId);
```

To import from `src/utils/strings`. They do pretty much the same thing but reusing a `Intl.Collator` object is improving performance tremendously

src/models/MSC3089TreeSpace.ts
```diff
@@ -0,0 +1,433 @@
...
+                }
+
+                const createTsA = roomA.currentState.getStateEvents(EventType.RoomCreate, "")?.getTs() ?? 0;
+                const createTsB = roomB.currentState.getStateEvents(EventType.RoomCreate, "")?.getTs() ?? 0;
+                if (createTsA === createTsB) {
+                    return a.roomId.localeCompare(b.roomId);
```
gsouquet, line 242:
```suggestion
                    return compare(a.roomId, b.roomId);
```

src/models/MSC3089TreeSpace.ts
```diff
@@ -0,0 +1,433 @@
...
+                }
+                return createTsA - createTsB;
+            } else { // both not-null orders
+                return a.order.localeCompare(b.order);
```
gsouquet, line 246:
```suggestion
                return compare(a.order, b.order);
```

src/models/MSC3089TreeSpace.ts
t3chguy: This is js-sdk, your string util is in react-sdk, maybe it ought to be moved.

src/models/MSC3089TreeSpace.ts
```diff
@@ -0,0 +1,433 @@
...
+
+        // TODO: Deal with order conflicts by reordering
+
+        // Now we can finally update our own order state
+        const currentChild = parentRoom.currentState.getStateEvents(EventType.SpaceChild, this.roomId);
+        const content = currentChild?.getContent() ?? { via: [this.client.getDomain()] };
+        await this.client.sendStateEvent(parentRoom.roomId, EventType.SpaceChild, {
+            ...content,
+            order: newOrder,
+        }, this.roomId);
```
t3chguy, line 381: I managed to massively simplify this by use of a method which creates a range of strings in between a gap and using an array to diff instead of trying to loop with prev/nextString to move one-by-one.

```typescript
const reorder = <T extends any>(list: T[], startIndex: number, endIndex: number): T[] => {
    const result = Array.from(list);
    const [removed] = result.splice(startIndex, 1);
    result.splice(endIndex, 0, removed);

    return result;
};

export const midPointsBetweenStrings = (a: string, b: string, count: number, alphabet = ALPHABET): string[] => {
    const n = Math.max(a.length, b.length);
    const aBase = stringToBase(pad(a, n, alphabet), alphabet);
    const bBase = stringToBase(pad(b, n, alphabet), alphabet);
    const step = (bBase - aBase) / (count + 1);
    if (step < 1) {
        return [];
    }
    return Array(count).fill(undefined).map((_, i) => baseToString(aBase + step + (i * step), alphabet));
};

public moveRootSpace(fromIndex: number, toIndex: number): void {
    if (
        fromIndex < 0 || toIndex < 0 ||
        fromIndex > this.rootSpaces.length || toIndex > this.rootSpaces.length ||
        fromIndex === toIndex
    ) {
        return;
    }
	const orders = this.rootSpaces.map(this.getSpaceTagOrdering);
	  
	const newRootSpaces = reorder(this.rootSpaces, fromIndex, toIndex);
	let firstIncorrectIdx = -1;
	let lastIncorrectIdx = -1;
	for (let i = 0; i < this.rootSpaces.length; i++) {
	    if (this.rootSpaces[i] !== newRootSpaces[i] || (i <= toIndex && orders[i] === undefined)) {
	        if (firstIncorrectIdx < 0) firstIncorrectIdx = i;
	        lastIncorrectIdx = i;
	    }
	}
	  
	if (firstIncorrectIdx < 0) return;
	  
	const prevOrder = firstIncorrectIdx > 0
	    ? orders[firstIncorrectIdx]
	    : String.fromCharCode(ALPHABET_START).repeat(5);
	const nextOrder = orders[lastIncorrectIdx + 1]
	    ?? String.fromCharCode(ALPHABET_END).repeat(prevOrder.length + 1);
	  
	const midpoints = midPointsBetweenStrings(prevOrder, nextOrder, 1 + lastIncorrectIdx - firstIncorrectIdx);
	for (let i = firstIncorrectIdx, j = 0; i <= lastIncorrectIdx; i++, j++) {
	    this.setRootSpaceOrder(newRootSpaces[i], midpoints[j]);
	}
```

src/client.ts
turt2live: hmm... looks like my comfort with the function is bleeding through in PRs now... (I'm really uncomfortable with "array or sometimes not" contracts)

t3chguy: If it was TS it'd be an overload, its not that insane

src/utils.ts
turt2live: The padding is to remove floating point issues we have because JS doesn't trigger integer division the same way Python does, but I suppose we can throw in more `floor()`s.

Will take a look.

Also, 50 is a bit tiny if the character space is 95...

spec/unit/utils.spec.js
turt2live: I hate this.

src/utils.ts
turt2live: Turns out I just padded the wrong value.

src/models/MSC3089TreeSpace.ts
turt2live: replaced with simple math

spec/unit/utils.spec.js
```diff
@@ -259,4 +270,111 @@ describe("utils", function() {
...
+        it('should find the next string appropriately', () => {
+            expect(prevString('B')).toEqual('A');
+            expect(prevString('c', 'abcdefghijklmnopqrstuvwxyz')).toEqual('b');
+            expect(prevString('cau')).toEqual('cat');
+            expect(prevString('cau', 'abcdefghijklmnopqrstuvwxyz')).toEqual('cat');
+        });
+    });
+
+    // Let's just ensure the ordering is sensible for lexicographic ordering
+    describe('string averaging unified', () => {
```
t3chguy, line 340: similar testing at the extremes of the alphabet (e.g `"~".repeat(50)`) would be good given the risk of overflows/loss of accuracy

src/client.ts
```diff
@@ -7709,6 +7717,74 @@ export class MatrixClient extends EventEmitter {
...
+
+        const createEvent = room.currentState.getStateEvents(EventType.RoomCreate, "");
+        const purposeEvent = room.currentState.getStateEvents(
+            UNSTABLE_MSC3088_PURPOSE.name,
+            UNSTABLE_MSC3089_TREE_SUBTYPE.name);
+
+        if (!createEvent) throw new Error("Expected single room create event");
+        if (!purposeEvent) return null;
+
+        if (!purposeEvent.getContent()?.[UNSTABLE_MSC3088_ENABLED.name]) return null;
```
t3chguy, line 7782:
```suggestion
        if (!purposeEvent?.getContent()?.[UNSTABLE_MSC3088_ENABLED.name]) return null;
```

src/models/MSC3089TreeSpace.ts
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    /**
+     * Whether or not this is a top level space.
+     */
+    public get isTopLevel(): boolean {
+        const parentEvents = this.room.currentState.getStateEvents(EventType.SpaceParent);
+        if (!parentEvents?.length) return true;
+        return parentEvents.every(e => !e.getContent()?.['via']);
+    }
```
t3chguy, line 96: This inherently isn't what determines whether a space is top-level or not. MSC3089 doesn't seem to change that either.

turt2live: Correct. Will add a comment, and it should be in MSC3089

src/NamespacedValue.ts
```diff
@@ -0,0 +1,95 @@
...
+*/
+
+/**
+ * Represents a simple Matrix namespaced value. This will assume that if a stable prefix
+ * is provided that the stable prefix should be used when representing the identifier.
+ */
+export class NamespacedValue<S extends string, U extends string> {
+    public constructor(public readonly stable: S, public readonly unstable?: U) {
+        if (!this.unstable && !this.stable) {
+            throw new Error("One of stable or unstable values must be supplied");
```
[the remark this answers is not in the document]
turt2live: stable is optional, but the type hints are trying to get you to provide one of the two. Will try to clarify the types.

src/NamespacedValue.ts
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    public get tsType(): U | S {
+        return null; // irrelevant return
+    }
```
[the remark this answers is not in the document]
turt2live: this was used in a prior iteration of Extensible Events, but has since been... lost? Not sure where it went.

src/utils.ts
```diff
@@ -456,3 +456,122 @@ export function setCrypto(c: Object) {
...
+ * Converts a string to a baseN number, where N is the alphabet's length.
+ *
+ * This is intended for use with string averaging.
+ * @param {string} s The string to convert to a number.
+ * @param {string} alphabet The alphabet to use as a single string.
+ * @returns {number} The baseN number.
+ */
+export function stringToBase(s: string, alphabet = DEFAULT_ALPHABET): number {
+    const len = alphabet.length;
+    const reversedStr = Array.from(s).reverse().join(""); // keep as string
```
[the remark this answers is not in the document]
turt2live: mostly because the math is already complex enough to understand: trying to increase density isn't exactly a good thing, and in practice this is under 1ms for the most complex strings. I'm not too concerned about improving it.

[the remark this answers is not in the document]
turt2live: gah, but I guess this is outside the range of "unstable, your fault for using" :/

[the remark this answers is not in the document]
t3chguy: if this bit is meant to be unstable then can it be named as such or live in an `unstableUtils.ts` or something? Other unstable things mention they instability

turt2live:
@t3chguy ptal. I've punted a proper conversion of the utils test out to https://github.com/vector-im/element-web/issues/17630 in the interest of speed.

spec/unit/utils.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+    });
+
+    describe('averageBetweenStrings', () => {
+        it('should average appropriately', () => {
+            expect(averageBetweenStrings('A', 'z')).toEqual(']');
+            expect(averageBetweenStrings('a', 'z', "abcdefghijklmnopqrstuvwxyz")).toEqual('m');
+            expect(averageBetweenStrings('AA', 'zz')).toEqual('^.');
+            expect(averageBetweenStrings('aa', 'zz', "abcdefghijklmnopqrstuvwxyz")).toEqual('mz');
+            expect(averageBetweenStrings('cat', 'doggo')).toEqual("d9>Cw");
+            expect(averageBetweenStrings('cat', 'doggo', "abcdefghijklmnopqrstuvwxyz")).toEqual("cumqh");
```
t3chguy, line 313: Can we get an extra test which checks avg between `     ` (5 spaces) and `!'Tu:}` - this doesn't result in a lexicographical average in my testing

t3chguy: Or simpler case: `expect(average("  ", "!!")).toBe("!");`

src/utils.ts
```diff
@@ -456,3 +456,133 @@ export function setCrypto(c: Object) {
...
+ * The default alphabet used by string averaging in this SDK. This matches
+ * all usefully printable ASCII characters (0x20-0x7E, inclusive).
+ */
+export const DEFAULT_ALPHABET = [""].reduce(() => {
+    let str = "";
+    for (let c = 0x20; c <= 0x7E; c++) {
+        str += String.fromCharCode(c);
+    }
+    return str;
+}, "");
```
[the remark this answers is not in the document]
turt2live: The diff highlight is surprisingly less helpful than I thought

spec/unit/utils.spec.ts
turt2live: That's an invalid test case due to the exponents involved. 2 spaces is a `0n` and `!!` comes out as `96n`, making the average `P`. Repeated blank spaces do not have an impact on the numbering.

t3chguy: it still breaks sort order and the `setOrder` method if it does happen.

turt2live: This will probably require a whole new algorithm to implement properly then.

t3chguy: Right, which should be done here because as implemented here any time an `order` field with a leading space is encountered it'll create an order which does not fit lexicographic ordering.

turt2live: it was close: https://github.com/matrix-org/matrix-js-sdk/pull/1732/commits/d3027e1fe879bd63b8cb3960547d651c9f0ce019

spec/unit/utils.spec.ts
```diff
@@ -285,26 +285,49 @@ describe("utils", function() {
...
+            expect(stringToBase("ab")).toEqual(BigInt(6337));
+            expect(stringToBase("cb", "abcdefghijklmnopqrstuvwxyz")).toEqual(BigInt(80));
         });
     });
 
     describe('averageBetweenStrings', () => {
         it('should average appropriately', () => {
+            console.log(stringToBase("  "));
+            console.log(stringToBase("!!"));
+            expect(averageBetweenStrings("  ", "!!")).toEqual(" P");
```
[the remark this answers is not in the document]
t3chguy: what happened to the cat and doggo test?

spec/unit/utils.spec.ts
```diff
@@ -285,26 +285,49 @@ describe("utils", function() {
 
     describe('baseToString', () => {
         it('should calculate the appropriate string from numbers', () => {
-            expect(baseToString(BigInt(10))).toEqual(DEFAULT_ALPHABET[10]);
-            expect(baseToString(BigInt(10), "abcdefghijklmnopqrstuvwxyz")).toEqual('k');
-            expect(baseToString(BigInt(6241))).toEqual("ab");
-            expect(baseToString(BigInt(53), "abcdefghijklmnopqrstuvwxyz")).toEqual('cb');
+            // Verify the whole alphabet
+            for (let i = BigInt(1); i <= DEFAULT_ALPHABET.length; i++) {
+                console.log({i}); // for debugging
```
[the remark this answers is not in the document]
turt2live: keeping it around for the debugging side, like the other test which does this. Can't have expect messages without a library, so out of scope for this PR.

spec/unit/utils.spec.ts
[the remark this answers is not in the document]
turt2live: cat and doggo test didn't change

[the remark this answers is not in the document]
t3chguy: arg, I was looking at the wrong diff

t3chguy (approved):
I hate everything about lexicographic ordering

:shipit:

PR #1736, opened 2021-06-15
Convert additional files to typescript

t3chguy:


src/models/relations.ts
```diff
@@ -27,8 +30,16 @@ import { logger } from '../logger';
  * EventTimelineSet#getRelationsForEvent.
  */
 export class Relations extends EventEmitter {
+    private relationEventIds = new Set<string>();
+    private relations = new Set<MatrixEvent>();
+    private annotationsByKey: Record<string, Set<MatrixEvent>> = {};
+    private annotationsBySender: Record<string, Set<MatrixEvent>> = {};
+    private sortedAnnotationsByKey: [string, MatrixEvent][] = [];
```
jryans, line 37: I think the second part is a `Set`...?

```suggestion
    private sortedAnnotationsByKey: [string, Set<MatrixEvent>][] = [];
```

jryans (approved):
Wow, looks great, thanks for all this! 😄

src/models/relations.ts
t3chguy: Yup! Once MatrixEvent gets typed that'd have been caught

PR #1739, opened 2021-06-16
Fix typescript return types for membership update events

t3chguy:
Fixed some jsdoc and ts return types.


src/client.ts
```diff
@@ -3940,76 +3940,33 @@ export class MatrixClient extends EventEmitter {
...
-            $roomId: roomId,
-        });
-        const data = {
-            user_id: userId,
-        };
-        return this.http.authedRequest(
-            callback, "POST", path, undefined, data,
-        );
+        // kick / ban someone could end up banning and then un-banning them.
+        return this.membershipChange(roomId, userId, "unban", undefined, callback);
```
turt2live, line 3948: unban is not a membership: https://github.com/vector-im/element-web/issues/17510

turt2live (changes requested):
otherwise lgtm

src/client.ts
t3chguy: Okay for `forget` but for ban/unban/kick/invite a re-usable helper seems clean, they all have the same signature (some omit reason) and instead of membership it could be `verb`

turt2live (approved):

PR #1740, opened 2021-06-17
Add invite retries to file trees

turt2live:


src/utils.ts
```diff
@@ -443,6 +444,26 @@ export async function chunkPromises<T>(fns: (() => Promise<T>)[], chunkSize: num
...
+ * @returns {Promise<T>} The promise for the retried operation.
+ */
+export function simpleRetryOperation<T>(promiseFn: (attempt: number) => Promise<T>): Promise<T> {
+    return promiseRetry((attempt: number) => {
+        return promiseFn(attempt);
+    }, {
+        forever: true,
+        factor: 2,
+        minTimeout: 3000, // ms
+        maxTimeout: 15000, // ms
```
turt2live, line 463: pre-emptive comment: we could probably expose this as an argument, but for now we don't need to change this. It'd be trivial to do later, and likely would consider a library-agnostic way of doing it.

jryans (approved):

PR #1741, opened 2021-06-17
Convert IndexedDBStore to TS

t3chguy:
Some types have been omitted as getting the generics right around the helper method would take a lot longer, task for later.


src/store/indexeddb.ts
```diff
@@ -0,0 +1,323 @@
...
+            userTuples.push([u.userId, u.events.presence.event]);
+
+            // note that we've saved this version of the user
+            this.userModifiedMap[u.userId] = u.getLastModifiedTime();
+        }
+
+        return this.backend.syncToDatabase(userTuples);
+    });
+
+    public setSyncData = this.degradable((syncData: object) => {
```
gsouquet, line 238: Is it possible to set return types on all the places that uses degradable ?

gsouquet (approved):

PR #1742, opened 2021-06-17
Convert Event Context to TS

t3chguy:
Requires https://github.com/matrix-org/matrix-react-sdk/pull/6208


src/models/event.ts
```diff
@@ -1070,33 +1148,33 @@ utils.extend(MatrixEvent.prototype, {
...
-    },
+    }
 
     /**
      * Gets whether or not the event is flagged as cancelled. See flagCancelled() for
      * more information.
      * @returns {boolean} True if the event is cancelled, false otherwise.
      */
-    isCancelled() {
+    isCancelled(): boolean {
```
gsouquet, line 1175: This can be removed and `_isCancelled` made public

gsouquet (approved):
Nice one 🏆 !

src/models/event.ts
t3chguy: that changes the signature from a function to a field which would be a breaking change

PR #1744, opened 2021-06-17
Add keysharing on invites to File Tree Spaces

turt2live:


src/models/MSC3089TreeSpace.ts
```diff
@@ -120,15 +121,28 @@ export class MSC3089TreeSpace {
...
+     * @param {boolean} shareHistoryKeys True (default) to share encryption keys
+     * with the invited user. This will allow them to decrypt the events (files)
+     * in the tree. Keys will not be shared if the room is lacking appropriate
+     * history visibility (by default, history visibility is "shared" in trees,
+     * which is an appropriate visibility for these purposes).
      * @returns {Promise<void>} Resolves when complete.
      */
-    public invite(userId: string, andSubspaces = true): Promise<void> {
-        // TODO: [@@TR] Share keys
+    public invite(userId: string, andSubspaces = true, shareHistoryKeys = true): Promise<void> {
```
jryans, line 131: Perhaps consider an options object, instead of a growing sequence of booleans which all look the same at the call site... just a suggestion though. 😇

jryans (approved):
Looks good overall, thanks! 😄

src/models/MSC3089TreeSpace.ts
turt2live: will keep it under advisement for the proof of concept :p 

There might still be one more option...

PR #1746, opened 2021-06-18
Convert Room and RoomState to Typescript

t3chguy:


tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "esModuleInterop": true,
     "module": "commonjs",
     "moduleResolution": "node",
-    "target": "es2016",
+    "target": "es2020",
```
jryans, line 8: I am personally okay with this, but it changes the compatibility surface of the SDK, esp. for Node where we'd now be targetting Node 14. As that's the current LTS version, it seems safe but it's still a change. Please update the README.

jryans (approved):
Looks great overall, thanks for working on this! 😄

tsconfig.json
t3chguy: I mean I can just write the types definition for `allSettled` as I already did in matrix-react-sdk if that simplifies things. Though if we're happy with es2020 then we should probably change on both layers, babel will probably be the configuration affecting the final output form anyhow

PR #1749, opened 2021-06-19
Convert crypto index to TS

t3chguy:


dbkr (commented):
Yay, 3700 line files. Let's say this is probably fine. Thanks for doing this, this will make life immeasurably more pleasant.

dbkr (approved):

PR #1754, opened 2021-06-23
Do not honor string power levels

aaronraimist:
Requires https://github.com/matrix-org/matrix-react-sdk/pull/6245


src/models/room-member.ts
```diff
@@ -163,7 +163,7 @@ export class RoomMember extends EventEmitter {
         const oldPowerLevel = this.powerLevel;
         const oldPowerLevelNorm = this.powerLevelNorm;
 
-        if (users[this.userId] !== undefined) {
+        if (users[this.userId] !== undefined && Number.isInteger(users[this.userId])) {
```
jryans, line 166: I noticed we also have related but different checks in `room-state.ts` as part of `maySendEventOfType`, where `Number.isFinite` is used... Perhaps those should also require integers as well?

jryans (changes requested):
Thanks, seems reasonable overall, but would be good to use the same checks in all power level paths.

aaronraimist:
I'm not sure where else you want changed. As far as I can tell this is the only place Element uses the JS SDK to show a PL.

These string power levels aren't going to last a much longer anyway since I also made a PR to enforce integers in Synapse https://github.com/matrix-org/synapse/pull/10232 so I don't think it's worth putting a whole lot more time in on the client side.

jryans (approved):
That's all I had in mind, thanks. Just seems good to have similar checks all around.

PR #1759, opened 2021-06-29
Fix broken /messages filtering due to internal field changes in FilterComponent

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/17753#event-4952613253

Also no longer sends `filter_json` redundantly, this halves the size of the filter GET field being sent, yay?


jryans (approved):
Looks great, thanks! 😄

PR #1763, opened 2021-07-01
Convert Sync and SyncAccumulator to Typescript

t3chguy:


gsouquet (approved):
That looks pretty sensible to me.
I did not manage to get a proper look at `sync.ts` as the diff view is broken but did skim over and that appears to be doing what we want

PR #1771, opened 2021-07-06
Functional members

jaller94:
At Element Matrix Services we're experimenting with "functional members", e.g. bots you want to have in a DM room with someone else.

Currently, if I have a DM room with Alice and my bot "Helper Bot" the DM room will be called "Alice and 1 other" (or "Helper Bot" and 1 other").
We want to filter bots based on a state event so that the room name will be just "Alice".

The state event content for the type `io.element.functional_members` would look like this:

```
{
    "service_members": [
        "@helperbot:localhost",
        "@reminderbot:alice.tdl"
    ]
}
```

----

I'm writing this PR as an employee at Element, so I'm not sure the Sign-Off is required.

Signed-off-by: Christian Paul <christianp@element.io>


src/@types/event.ts
```diff
@@ -33,6 +33,7 @@ export enum EventType {
     RoomGuestAccess = "m.room.guest_access",
     RoomServerAcl = "m.room.server_acl",
     RoomTombstone = "m.room.tombstone",
+    RoomFunctionalMembers = "io.element.functional_members",
```
t3chguy, line 36: Ideally only `m.` prefixed items live in the js-sdk for a clean separation but there is a single other `io.element` instance already here

turt2live: `UnstablePrefix` should help a lot here.

src/models/room.ts
```diff
@@ -2039,20 +2039,41 @@ export class Room extends EventEmitter {
         // -1 because these numbers include the syncing user
         const inviteJoinCount = joinedMemberCount + invitedMemberCount - 1;
 
+        // get service members (e.g. helper bots) for exclusion
+        let excludedUserIds: string[] = [];
+        const mFunctionalMembers = this.currentState.getStateEvents(EventType.RoomFunctionalMembers, "");
+        if (mFunctionalMembers && mFunctionalMembers.getContent() && Array.isArray(mFunctionalMembers.getContent().service_members)) {
```
t3chguy, line 2045: Just FYI, you have access to TS here. And getContent is never falsey

```suggestion
        if (Array.isArray(mFunctionalMembers?.getContent().service_members)) {
```

t3chguy:
(Updated base branch so it doesn't accidentally go to the wrong place.)

src/@types/event.ts
t3chguy: @turt2live it isn't unstable though, it seems to be Element-specific functionality which doesn't appear to be intended for MSC'ing

jaller94: Correct, for now we don't have plans to standardise this.
If it proves to be useful to a wider audience, I'm sure an MSC will follow at some point.

src/models/room.ts
jaller94: I blame line 2019 for bad influence.
https://github.com/matrix-org/matrix-js-sdk/blob/dc90115a1bbd733351097fc97d93f47b0e31fa8d/src/models/room.ts#L2019

src/@types/event.ts
t3chguy: From a surface glance it'll likely be handed in standard by https://github.com/matrix-org/matrix-doc/blob/travis/msc/immutable-dms/proposals/2199-canonical-dms.md `unimportant` users

src/models/room.ts
t3chguy: Oh for sure, the TS conversion I did definitely did not exhaustively improve every single possible line, but rather focused on types.

src/@types/event.ts
turt2live: Anything off-spec is considered unstable by the spec, so it would still apply here. You don't need to specify a stable value to `UnstablePrefix`

turt2live: Or it might be `UnstableNamespace`, but either way a class exists for this exact reason

jaller94: This is my first PR to the repo. Can you point me to the place of unstable event types?
I see one UPPER_CASE unstable constant at the bottom of `event.ts`. Is that where this type is supposed to go?

turt2live: It's not a well-defined system just yet, but something like that would be best: goal is to not have it in the enum until it is a stable identifier, at which point there's two references to it while people transition to the enum.

jaller94: I renamed it to `UNSTABLE_ELEMENT_FUNCTIONAL_USERS`.

src/@types/event.ts
```diff
@@ -145,6 +144,13 @@ export const UNSTABLE_MSC3089_LEAF = new UnstableValue("m.leaf", "org.matrix.msc
  */
 export const UNSTABLE_MSC3089_BRANCH = new UnstableValue("m.branch", "org.matrix.msc3089.branch");
 
+/**
+ * Functional members type for declaring a purpose of room members (e.g. helpful bots).
+ * Note that this reference is UNSTABLE and subject to breaking changes, including its
+ * eventual removal.
+ */
+export const UNSTABLE_ELEMENT_FUNCTIONAL_USERS = "io.element.functional_members");
```
turt2live, line 152: you're missing `new UnstableValue(` from this, but have the closing brace :)

jaller94:
```suggestion
export const UNSTABLE_ELEMENT_FUNCTIONAL_USERS = new UnstableValue("io.element.functional_members");
```

src/@types/event.ts
```diff
@@ -144,6 +144,13 @@ export const UNSTABLE_MSC3089_LEAF = new UnstableValue("m.leaf", "org.matrix.msc
  */
 export const UNSTABLE_MSC3089_BRANCH = new UnstableValue("m.branch", "org.matrix.msc3089.branch");
 
+/**
+ * Functional members type for declaring a purpose of room members (e.g. helpful bots).
+ * Note that this reference is UNSTABLE and subject to breaking changes, including its
+ * eventual removal.
+ */
+export const UNSTABLE_ELEMENT_FUNCTIONAL_USERS = new UnstableValue("io.element.functional_members");
```
jaller94, line 152: @turt2live Can I give it the same value as the first and second parameter?

jaller94: In general, how to use this `UnstableValue`?
TypeScript complains.
https://buildkite.com/matrix-dot-org/matrix-js-sdk/builds/3082#5440f81c-35a8-44de-a76d-7013d2f8b207

t3chguy:
Adding tests for this behaviour would be great

src/@types/event.ts
```diff
@@ -144,6 +144,15 @@ export const UNSTABLE_MSC3089_LEAF = new UnstableValue("m.leaf", "org.matrix.msc
...
 export const UNSTABLE_MSC3089_BRANCH = new UnstableValue("m.branch", "org.matrix.msc3089.branch");
 
+/**
+ * Functional members type for declaring a purpose of room members (e.g. helpful bots).
+ * Note that this reference is UNSTABLE and subject to breaking changes, including its
+ * eventual removal.
+ */
+export const UNSTABLE_ELEMENT_FUNCTIONAL_USERS = new UnstableValue(
+    "io.element.functional_members",
+    "io.element.functional_members");
```
turt2live, line 154: I'd sooner hop into `UnstableValue` and mark the second field as optional, fixing any docs along the way.

src/models/room.ts
```diff
@@ -2037,22 +2037,43 @@ export class Room extends EventEmitter {
         const joinedMemberCount = this.currentState.getJoinedMemberCount();
         const invitedMemberCount = this.currentState.getInvitedMemberCount();
         // -1 because these numbers include the syncing user
-        const inviteJoinCount = joinedMemberCount + invitedMemberCount - 1;
+        let inviteJoinCount = joinedMemberCount + invitedMemberCount - 1;
+
+        // get service members (e.g. helper bots) for exclusion
+        let excludedUserIds: string[] = [];
+        const mFunctionalMembers = this.currentState.getStateEvents(UNSTABLE_ELEMENT_FUNCTIONAL_USERS.unstable, "");
```
turt2live, line 2044: We should use `name` here as otherwise the whole point is lost.
```suggestion
        const mFunctionalMembers = this.currentState.getStateEvents(UNSTABLE_ELEMENT_FUNCTIONAL_USERS.name, "");
```

src/@types/event.ts
```diff
@@ -144,6 +144,15 @@ export const UNSTABLE_MSC3089_LEAF = new UnstableValue("m.leaf", "org.matrix.msc
  */
 export const UNSTABLE_MSC3089_BRANCH = new UnstableValue("m.branch", "org.matrix.msc3089.branch");
 
+/**
+ * Functional members type for declaring a purpose of room members (e.g. helpful bots).
+ * Note that this reference is UNSTABLE and subject to breaking changes, including its
+ * eventual removal.
+ */
```
turt2live, line 151: A schema of the event content would be appreciated here given there's no MSC backing it.

turt2live (changes requested):
as t3chguy says, tests would be appreciated for this. From my part, it'd be good to see tests for the abusive cases, and a paragraph somewhere on this PR acknowledging the room for abuse and how that is mitigated (if at all).

Some abusive cases include:
* Existing DMs with participants where excluding the member might cause an empty name (eg: 1:1 DM and manually excluding a non-service user). Alternatively, putting all users from the DM into the excluded users list.
* Excluding users which are not present in the room (would expect the performance to not be impacted, and for the code to not explode)

It's particularly concerning that the exclusion is based on user ID only as that isn't a reliable factor for whether or not the user is truly a service/bot user or not. It's largely fine for this case, I think, but acknowledging the vector is appreciated.

src/@types/event.ts
jaller94: I'd make it a type of its own, as UnstableValue has a different meaning, is exported and also requires `unstable` to be defined.

Or do you mean to set `.unstable` to the same value as `.name` if no second parameter is provided to the constructor?

turt2live (approved):
Thanks!

Kingcrypto38 (approved):

PR #1778, opened 2021-07-12
Add minimal types for "notification settings" UI

turt2live:
For https://github.com/matrix-org/matrix-react-sdk/pull/6352

**Note**: These aren't meant to be comprehensive types. They are just enough to get the notification settings UI refresh done, plus a little around the edges for ease of use. A future PR will have to go through fix all the missing types, still.


src/client.ts
```diff
@@ -7235,7 +7216,13 @@ export class MatrixClient extends EventEmitter {
      * @return {Promise} Resolves: an empty object {}
      * @return {module:http-api.MatrixError} Rejects: with an error response.
      */
-    public addPushRule(scope: string, kind: string, ruleId: string, body: any, callback?: Callback): Promise<{}> {
+    public addPushRule(
+        scope: string,
+        kind: PushRuleKind,
+        ruleId: RuleId | string,
```
t3chguy, line 7222: this actually can't be RuleId as you cannot add those given the `.` prefix - maybe `Exclude<string, RuleId>` or something

src/client.ts
```diff
@@ -7252,7 +7239,12 @@ export class MatrixClient extends EventEmitter {
      * @return {Promise} Resolves: an empty object {}
      * @return {module:http-api.MatrixError} Rejects: with an error response.
      */
-    public deletePushRule(scope: string, kind: string, ruleId: string, callback?: Callback): Promise<{}> {
+    public deletePushRule(
+        scope: string,
+        kind: PushRuleKind,
+        ruleId: RuleId | string,
```
t3chguy, line 7245: same again

t3chguy (approved):
LGTM otherwise

PR #1779, opened 2021-07-12
Convert IndexedDB store & worker to Typescript

t3chguy:
Requires https://github.com/vector-im/element-web/pull/17962

Avoids using `event.target` on IndexedDB events due to https://github.com/microsoft/TypeScript/issues/28293


src/store/indexeddb-backend.ts
```diff
@@ -0,0 +1,36 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { ISavedSync } from "./index";
+import { IEvent, IStartClientOpts, ISyncResponse } from "..";
+
+export interface IIndexeddbBackend {
```
jryans, line 20: The types in the next file use the spelling `IDB` as opposed to `Idb`, so I feel like this would match more as...

```suggestion
export interface IIndexedDBBackend {
```

...but feel free to ignore if you feel strongly. Even though we use write e.g. `Url`, we seem to also write `IndexedDB` as the spelling in most places.

jryans (approved):
Looks great overall, thanks! 😄

src/store/indexeddb-backend.ts
t3chguy: oh yeah, not sure how I messed that one up

PR #1780, opened 2021-07-12
Use webpack worker-loader instead of homegrown hack

t3chguy:
# Requires a version bump

Switches to a more generic approach to passing a worker to the indexeddb remote backend by means of a factory


gsouquet (approved):
:shipit:

PR #1784, opened 2021-07-14
Expose MatrixEvent's internal clearEvent as a function

turt2live:
For https://github.com/vector-im/element-web/issues/17615
For https://github.com/matrix-org/matrix-react-sdk/pull/6371

### DIFF
--- a/src/models/event.ts
+++ b/src/models/event.ts
@@ -263,6 +263,15 @@ export class MatrixEvent extends EventEmitter {
     }
 
     /**
+     * Gets the clear event as it would be received over the wire. If the event
+     * is not encrypted, this simply returns the event as-is.
+     * @returns {Partial<IEvent>} The clear event, as known by the SDK.
+     */
+    public getClearEvent(): Partial<IEvent> {
+        return this.clearEvent ?? this.event;
+    }
+
+    /**
      * Get the event_id for this event.
      * @return {string} The event ID, e.g. <code>$143350589368169JsLZx:localhost
      * </code>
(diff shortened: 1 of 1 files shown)


gsouquet (approved):
🙌

turt2live:
(rr @t3chguy due to comment on the react-sdk side)

src/models/event.ts
```diff
@@ -1240,15 +1240,7 @@ export class MatrixEvent extends EventEmitter {
...
-            sender: this.getSender(),
-            content: this.getContent(),
-            event_id: this.getId(),
-            origin_server_ts: this.getTs(),
-            unsigned: this.getUnsigned(),
-            room_id: this.getRoomId(),
-        };
+        const event = this.getEffectiveEvent();
 
         // if this is a redaction then attach the redacts key
```
t3chguy, line 1245: this is already achieved by your getEffectiveEvent

src/models/event.ts
```diff
@@ -262,6 +262,16 @@ export class MatrixEvent extends EventEmitter {
...
     }
 
+    /**
+     * Gets the event as though it would appear unencrypted. If the event is already not
+     * encrypted, it is simply returned as-is.
+     * @returns {IEvent} The event in wire format.
+     */
+    public getEffectiveEvent(): IEvent {
+        // clearEvent doesn't have all the fields, so we'll copy what we can from this.event
+        return Object.assign({}, this.event, this.clearEvent) as IEvent;
```
t3chguy, line 272: are you allergic to the nicer syntax :P?

```
return {
    ...this.event,
    ...this.clearEvent,
}
```

t3chguy (approved):
LGTM :shipit:

src/models/event.ts
turt2live: "nicer" - I feel it doesn't communicate the order properly given JS objects are unordered, so there's no implicit guarantee that the spread operation will be applied correctly. The spec says it will, but that's a detail someone has to remember when writing/reading the code.

PR #1788, opened 2021-07-15
Use an unstable prefix for MSC2885: Hidden read receipts  

SimonBrandner:
We should be using an unstable prefix as per [`a4e5b45` (#2285)](https://github.com/matrix-org/matrix-doc/pull/2285/commits/a4e5b45204b243b6b23226a0fcf5cf8a9b10cdc0)

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -3650,7 +3650,7 @@ export class MatrixClient extends EventEmitter {
         }
 
         const addlContent = {
-            "m.hidden": Boolean(opts.hidden),
+            "org.matrix.msc2885.hidden": Boolean(opts.hidden),
         };
 
         return this.sendReceipt(event, "m.read", addlContent, callback);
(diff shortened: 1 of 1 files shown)


t3chguy (approved):
Yup, thanks. Synapse doesn't support this anyway so not super critical :P

src/client.ts
```diff
@@ -3650,7 +3650,7 @@ export class MatrixClient extends EventEmitter {
         }
 
         const addlContent = {
-            "m.hidden": Boolean(opts.hidden),
+            "org.matrix.msc2885.hidden": Boolean(opts.hidden),
```
t3chguy, line 3653:
```suggestion
            "org.matrix.msc2285.hidden": Boolean(opts.hidden),
```

src/client.ts
```diff
@@ -6429,7 +6429,7 @@ export class MatrixClient extends EventEmitter {
         const content = {
             "m.fully_read": rmEventId,
             "m.read": rrEventId,
-            "m.hidden": Boolean(opts ? opts.hidden : false),
+            "org.matrix.msc2885.hidden": Boolean(opts ? opts.hidden : false),
```
t3chguy, line 6432:
```suggestion
            "org.matrix.msc2285.hidden": Boolean(opts ? opts.hidden : false),
```

t3chguy (changes requested):
THAT COULD HAVE BEEN A FUN ONE TO DEBUG

SimonBrandner:
@t3chguy, thanks for the massive save!

t3chguy (approved):

PR #1789, opened 2021-07-15
Switch callEventHandler from listening on `event` to `Room.timeline`

t3chguy:
so that it doesn't end up processing ephemerals, account data, remote echoes, etc etc

This was causing remote-echoes to be pointlessly decrypted


dbkr (commented):
That was simple... hopefully that's all there is to it!

dbkr (approved):

PR #1792, opened 2021-07-16
Deprecate groups APIs

t3chguy:
Notes: Deprecate the groups APIs. Groups are no longer supported, only Synapse has support. They are being replaced by Spaces which build off of Rooms and are far more flexible.

---


jryans (approved):
Should it be recorded any special way in the release notes?

t3chguy:
I'll do one of them changeloggy things, not merging it just yet

t3chguy:
Not sure how to sanely expose this to the changelog.

PR #1794, opened 2021-07-18
Add node-es6 example

theawless:
Hi! While trying to use the SDK, I faced various issues. I tried to find the answers in the documentation and comments without any luck. I thought adding this example with tons of comments would help me and future users as well. I have put a few questions in the code comments which I will update once they are answered. I hope the questions would be obvious to a regular contributor.

Please feel free to close this PR if the example is not worthy to be added.

Through this example, I have tried to demonstrate:
* log in with a password
* change log level
* create a filter (please let me know if this is not the right way)
* use the SDK with ES6
* document delivery and ordering guarantees

It is a simple piece of code where multiple users join a room and send a hello message, and once other users receive this message they send a hello back. Similar thing happens when a user leaves.

Open questions (mostly mentioned in the code comments):
* Does `startClient` save the filter on the server?
* When is the client ready to listen to new events?
  Just after the `startClient` method or after the `sync` event?
* When is the client ready to join new rooms?
  Just after the `startClient` method or after the `sync` event?
* Can we get duplicate message events?
* Is there at least once delivery guarantee for message events?
  Example:
  -> user1 is already in a room
  -> user2 joins and sends a message
  -> user1 sends first message
  -> user2 doesn't receive the message event (how can this happen?)
  -> user1 sends second message
  -> user2 receives the second message event

Notes: none
Labels: T-Task
Signed-off-by: Abhinav Singh <theawless@gmail.com>

### DIFF
--- a/examples/node-es6/README.md
+++ b/examples/node-es6/README.md
@@ -1,0 +1,56 @@
+This is a sample script to showcase server side filtering, node es6, and a few other things.
+
+To try it out, first edit `app.js` to configure your `baseUrl` and `room`.
+
+Next build the example using this command:
+
+```shell
+npm install
+```
+
+Next edit and run this command for multiple users on multiple terminals:
+
+```shell
+node app user1 pass1
+```
+
+Example output (after running for three different users):
+
+```shell
+@user-1:matrix.org: hello everyone
+@user-2:matrix.org: hello everyone
+@user-1:matrix.org: hello @user-2:matrix.org
+@user-3:matrix.org: hello everyone
+@user-1:matrix.org: hello @user-3:matrix.org
+@user-2:matrix.org: hello @user-3:matrix.org
+@user-1:matrix.org: adios everyone
+```
+
+```shell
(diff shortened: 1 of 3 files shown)


gsouquet (commented):
Thank you for working on that, I can totally understand how JavaScript syntax can be off-putting sometimes given it's changed so much in recent years.

At this stage I'm not sure I would create a separate example to showcase how to use the SDK with newer JavaScript feature. I would rather make the existing examples evolve and be refreshed with that `async/await` syntax you're using here for example.

For that reason I'll close this PR, but feel free to re-open another one and update the current examples 👍

theawless:
@gsouquet I'm sorry if my description was not clear enough. The example was not only adding node-es6 demonstration but also answering questions on how to use filters, and what kind of guarantees (one time delivery/duplicate event) we get for receiving messages. I didn't find any examples for that in the documentation.

The name of the example - I just chose because the other examples were named like that, but it's irrelevant to the main gist of the example.

PR #1801, opened 2021-07-21
Improve calculateRoomName performances by using Intl.Collator

gsouquet:


src/utils.ts
```diff
@@ -683,3 +683,13 @@ export function lexicographicCompare(a: string, b: string): number {
...
+
+const collator = new Intl.Collator();
+/**
+ * Performant language-sensitive string comparison
+ * @param a the first string to compare
+ * @param b the second string to compare
+ */
+export function compare(a: string, b: string): number {
+    return collator.compare(a, b);
+}
```
t3chguy, line 695: kind of annoying that we're duplicating this in react-sdk now, can you open a maintenance issue to use the js-sdk util instead of duplicating along with merging this please

t3chguy (approved):

bradjones1:
Dropping a quite note here to hopefully save others time - `Intl` isn't present in React Native < [65.0](https://reactnative.dev/blog/2021/08/17/version-065#whats-new-in-hermes-08), so this will result in errors like `Can't find variable: Intl` when running on Expo SDK <= 44. (There's not yet an Expo SDK rolled that includes RN 65.) I ended up monkey-patching this with a revert since this is (hopefully) a temporary condition. Whew.

byanes:
I am facing the same issue in React Native with Expo. Are there any plans of changing/patching this?

t3chguy:
This project aims for Node 10 compatibility, Intl has been in Node since 0.12.0 so you'll need to find a way of having Expo polyfill it.

bradjones1:
@byanes Per my comment above, you'll have Intl in Expo once they roll an SDK release which includes React Native 65. See https://github.com/expo/fyi/blob/main/react-native-releases.md#so-when-will-expo-support-react-native-latest_version for their policy (which isn't much help for understanding "when" you'll get it, but it does set expectations.)

You can maintain a fork of this project and revert the applicable commit, as I am doing [here](https://github.com/bradjones1/matrix-js-sdk/tree/v99)

bradjones1:
Expo SDK 45 [includes RN 0.68](https://blog.expo.dev/expo-sdk-45-f4e332954a68), FYI.

byanes:
> Expo SDK 45 [includes RN 0.68](https://blog.expo.dev/expo-sdk-45-f4e332954a68), FYI.

Thanks for the heads up! I'll give it a try.

bradjones1:
I've confirmed this works with Hermes JS engine in RN 0.68.

jpschw:
I'm using 0.68.2 and still getting the error...

jpschw:
For everyone using Expo - you can enable Hermes in app.json to get rid of the error:
https://docs.expo.dev/guides/using-hermes/

Thank you @bradjones1

PR #1803, opened 2021-07-21
Migrate room-state members object to a Map

gsouquet:
For matrix-org/matrix-react-sdk#6429

### DIFF
--- a/src/models/room-state.ts
+++ b/src/models/room-state.ts
@@ -53,7 +53,7 @@ export class RoomState extends EventEmitter {
     private modified: number;
 
     // XXX: Should be read-only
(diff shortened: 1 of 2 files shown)


src/models/room-state.ts
```diff
@@ -53,7 +53,7 @@ export class RoomState extends EventEmitter {
     private modified: number;
 
     // XXX: Should be read-only
-    public members: Record<string, RoomMember> = {}; // userId: RoomMember
+    public members = new Map<string, RoomMember>(); // userId: RoomMember
```
t3chguy, line 56: We access this field in a few places so this would need a version bump

image

PR #1807, opened 2021-07-23
Clean up Event.clearEvent handling to fix a bug where malformed events with falsey content wouldn't be considered decrypted

bradtgmurray:
`shouldAttemptDecryption` was checking to see if we had a `content` property on our `clearEvent`, but it's possible that we've already been decrypted and the result still has a falsey `content` property. Instead, only set the `clearEvent` class member after we've successfully decrypted, and leave it undefined otherwise.

Signed-off-by: Brad Murray <brad@beeper.com>


src/models/event.ts
```diff
@@ -486,7 +486,7 @@ export class MatrixEvent extends EventEmitter {
     }
 
     public shouldAttemptDecryption() {
-        return this.isEncrypted() && !this.isBeingDecrypted() && this.getClearContent() === null;
+        return this.isEncrypted() && !this.isBeingDecrypted() && !this.clearEvent;
```
bradtgmurray, line 489: This is the "fix" here.

bradtgmurray (commented):
Most of this PR is just me reviewing references to `clearEvent` and making sure the code path still behaves if it's now undefined instead of an empty object.

src/models/event.ts
```diff
@@ -285,7 +285,7 @@ export class MatrixEvent extends EventEmitter {
      * @return {string} The event type, e.g. <code>m.room.message</code>
      */
     public getType(): EventType | string {
-        return this.clearEvent.type || this.event.type;
+        return this.clearEvent?.type || this.event.type;
```
t3chguy, line 288: this falling-through for a decrypted event with malformed `type` in the encrypted payload is not ideal, it'll behave inconsistently where getType() may be for wire or payload and getContent() for the other

src/models/event.ts
```diff
@@ -334,7 +334,7 @@ export class MatrixEvent extends EventEmitter {
         if (this._localRedactionEvent) {
             return {} as T;
         }
-        return (this.clearEvent.content || this.event.content || {}) as T;
+        return (this.clearEvent?.content || this.event.content || {}) as T;
```
t3chguy, line 337: see above

src/models/event.ts
```diff
@@ -334,7 +337,10 @@ export class MatrixEvent extends EventEmitter {
         if (this._localRedactionEvent) {
             return {} as T;
         }
-        return (this.clearEvent.content || this.event.content || {}) as T;
+        if (this.clearEvent) {
+           return this.clearEvent.content as T;
```
t3chguy: this also needs the `|| {}` bit given content can be `falsey` to match the behaviour applied to the this.event.content below

src/models/event.ts
```diff
@@ -729,8 +732,10 @@ export class MatrixEvent extends EventEmitter {
      * @returns {Object} The cleartext (decrypted) content for the event
      */
     public getClearContent(): IContent | null {
-        const ev = this.clearEvent;
-        return ev && ev.content ? ev.content : null;
+        if (this.clearEvent) {
+            return this.clearEvent.content;
+        }
+        return null;
```
t3chguy, line 738: the behaviour here is slightly different, where falsey things got coalesced into `null` but now they won't. For matching the signature better it would probably be good to maintain that

t3chguy (commented):
Other than the minor nits this is great, thanks!

src/models/event.ts
bradtgmurray: Makes sense to me, made the change.

What do you think of tweaking the comment on the function? I like maintaining the old behaviour, but the comment doesn't quite match the old behaviour.

t3chguy: Huh, that's a good spot. I'm actually tempted to side with the comment in that case rather than the old behaviour

bradtgmurray: Going to move over to respecting the comment I think. The behaviour is only different in the case of falsey content, which was already being weirdly handled.

spec/integ/megolm-integ.spec.js
```diff
@@ -1025,4 +1025,68 @@ describe("megolm", function() {
...
+            return aliceTestClient.flushSync();
+        }).then(function() {
+            const room = aliceTestClient.client.getRoom(ROOM_ID);
+            const event = room.getLiveTimeline().getEvents()[0];
+            expect(event.isEncrypted()).toBe(true);
+            return testUtils.awaitDecryption(event);
+        }).then((event) => {
+            expect(event.getRoomId()).toEqual(ROOM_ID);
+            expect(event.getContent()).toEqual({});
+            expect(event.getClearContent()).toBeUndefined();
```
bradtgmurray, line 1089: This test would fail on develop because getClearContent would continue to be null before and now it actually has the correct(?) value.

Also, `testUtils.awaitDecryption` would have thought that the event wasn't decoded yet because `getClearContent` would still be null, and therefore would wait for an "Event.decrypted" event that would never come as it was already decrypted.

src/models/event.ts
```diff
@@ -528,10 +534,7 @@ export class MatrixEvent extends EventEmitter {
             throw new Error("Attempt to decrypt event which isn't encrypted");
         }
 
-        if (
-            this.clearEvent && this.clearEvent.content &&
-            this.clearEvent.content.msgtype !== "m.bad.encrypted"
-        ) {
+        if (this.clearEvent?.content && this.clearEvent.content.msgtype !== "m.bad.encrypted") {
```
t3chguy, line 537: Shouldn't `this.clearEvent?.content` now just be `this.clearEvent` given the presence of clearEvent is now the signal for whether something has or has not been decrypted - `this.clearEvent?.content` was the old signal for this

t3chguy (approved):
Great, thanks for this!

:shipit:

bradtgmurray:
Thanks for the feedback and the speedy review!

PR #1812, opened 2021-07-27
Support for MSC3291: Muting in VoIP calls

SimonBrandner:
An implementation of https://github.com/matrix-org/matrix-doc/pull/3291
Requires https://github.com/matrix-org/matrix-react-sdk/pull/6486


src/@types/event.ts
```diff
@@ -53,6 +53,7 @@ export enum EventType {
     CallReject = "m.call.reject",
     CallSelectAnswer = "m.call.select_answer",
     CallNegotiate = "m.call.negotiate",
+    CallSDPStreamMetadataChanged = "org.matrix.call.sdp_stream_metadata_changed",
```
dbkr, line 56: Best practice is to *send* only the prefixed version but support *receiving* both, that way once it's unprefixed, old clients will work seamlessly with new ones.

dbkr (approved):

PR #1813, opened 2021-07-27
Handle DTMF support

SimonBrandner:
For https://github.com/matrix-org/matrix-react-sdk/pull/6472

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -394,6 +394,10 @@ export class MatrixCall extends EventEmitter {
         return Boolean(this.opponentCaps && this.opponentCaps["m.call.transferee"]);
     }
 
+    public opponentSupportsDTMF(): boolean {
+        return Boolean(this.opponentCaps && this.opponentCaps["m.call.dtmf"]);
+    }
+
     public getRemoteAssertedIdentity(): AssertedIdentity {
         return this.remoteAssertedIdentity;
     }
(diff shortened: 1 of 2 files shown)


Palid (commented):
Looks fine to me according to the [spec](https://github.com/matrix-org/matrix-doc/blob/dbkr/msc2746/proposals/2746-reliable-voip.md#add-dtmf) but I don't really feel confident enough about this as of this moment. @dbkr please additionally have a look at this, so I don't do a stupid merge 😄

src/webrtc/call.ts
```diff
@@ -1032,11 +1036,10 @@ export class MatrixCall extends EventEmitter {
             [SDPStreamMetadataKey]: this.getLocalSDPStreamMetadata(),
         } as MCallAnswer;
 
-        if (this.client.supportsCallTransfer) {
```
Palid, line 1035: Is there a reason why this check has been removed?

src/webrtc/call.ts
```diff
@@ -1374,11 +1377,10 @@ export class MatrixCall extends EventEmitter {
             content.description = this.peerConn.localDescription;
         }
 
-        if (this.client.supportsCallTransfer) {
```
Palid, line 1377: Is there a reason why this check has been removed?

dbkr (approved):

PR #1817, opened 2021-07-29
Add strictNullChecks to tsconfig.json

Palid:


hughns:
@Palid FYI - I have recently encountered some issues with incorrect typings that are masked by `strictNullChecks` being `false`. The total error count in matrix-js-sdk with `strictNullChecks` being `true` is now 950+ which I believe indicates the situation has gotten worse. 👎

MadLittleMods:
Feels like we can close this and come back to it whenever we want to address this.

The new issue to track around this is https://github.com/vector-im/element-web/issues/21967

PR #1818, opened 2021-07-29
Add support for /hierarchy API and fall back to /spaces API if needed

t3chguy:

### DIFF
--- a/src/@types/spaces.ts
+++ b/src/@types/spaces.ts
@@ -43,8 +43,9 @@ export interface ISpaceSummaryEvent {
     };
 }
 
-export interface IRoomChildState extends IStrippedState {
+export interface IHierarchyRelation extends IStrippedState {
     room_id: string;
+    origin_server_ts: number;
     content: {
         order?: string;
         suggested?: boolean;
@@ -52,8 +53,8 @@ export interface IRoomChildState extends IStrippedState {
     };
 }
 
-export interface IRoomChild extends IPublicRoomsChunkRoom {
+export interface IHierarchyRoom extends IPublicRoomsChunkRoom {
     room_type?: RoomType | string;
-    children_state: IRoomChildState[];
+    children_state: IHierarchyRelation[];
 }
 /* eslint-enable camelcase */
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -8008,6 +8009,60 @@ export class MatrixClient extends EventEmitter {
...
 
+    /**
+     * Fetches or paginates a room hierarchy as defined by MSC2946.
+     * Falls back gracefully to sourcing its data from `getSpaceSummary` if this API is not yet supported by the server.
+     * @param {string} roomId The ID of the space-room to use as the root of the summary.
+     * @param {number?} limit The maximum number of rooms to return per page.
+     * @param {number?} maxDepth The maximum depth in the tree from the root room to return.
+     * @param {boolean?} suggestedOnly Whether to only return rooms with suggested=true.
+     * @param {string?} fromToken The opaque token to paginate a previous request.
+     * @returns {Promise} the response, with next_token & rooms fields.
```
clokep, line 8020: We updated this to `next_batch` in matrix-org/synapse#10579!

src/@types/spaces.ts
```diff
@@ -15,26 +15,46 @@ limitations under the License.
...
 // Types relating to Rooms of type `m.space` and related APIs
 
 /* eslint-disable camelcase */
+/** @deprecated */
 export interface ISpaceSummaryRoom extends IPublicRoomsChunkRoom {
     num_refs: number;
     room_type: string;
 }
 
+/** @deprecated */
```
turt2live, line 30:
```suggestion
/** @deprecated Use hierarchy instead where possible. */
```

src/@types/spaces.ts
```diff
@@ -15,26 +15,46 @@ limitations under the License.
...
 export interface ISpaceSummaryEvent {
```
turt2live, line 31: I'm surprised we don't have a generic event interface yet

t3chguy: We do, this predates it, keeping the identifier until deprecation expires and this gets canned

PR #1819, opened 2021-07-29
Only clear bit 63 when we create the IV

uhoreg:
If we got the IV from somewhere else, we should just use it as-is, rather than trying to re-clear the bit.

This fixes compatibility with other clients that happen to clear different bits.


t3chguy (approved):
Looks sane :shipit:

PR #1822, opened 2021-08-03
Bump eslint deps

Palid:


src/@types/PushRules.ts
```diff
@@ -91,10 +91,10 @@ export interface ISenderNotificationPermissionCondition
...
 
 export type PushRuleCondition = IPushRuleCondition<string>
-    | IEventMatchCondition
-    | IContainsDisplayNameCondition
-    | IRoomMemberCountCondition
-    | ISenderNotificationPermissionCondition;
+| IEventMatchCondition
+| IContainsDisplayNameCondition
+| IRoomMemberCountCondition
+| ISenderNotificationPermissionCondition;
```
t3chguy, line 97: can't say I agree with this indentation

src/client.ts
```diff
@@ -7990,9 +7990,9 @@ export class MatrixClient extends EventEmitter {
         limit?: number,
         batch?: string,
     ): Promise<{
-        rooms: ISpaceSummaryRoom[];
-        events: ISpaceSummaryEvent[];
-    }> {
+            rooms: ISpaceSummaryRoom[];
+            events: ISpaceSummaryEvent[];
+        }> {
```
t3chguy, line 7995: can't say I agree with this indentation at all

t3chguy (changes requested):

t3chguy:
Warrants some code style discussion

src/client.ts
Palid: Do we just set it to 2 for objects? Turning it off entirely isn't good either.

PR #1824, opened 2021-08-03
Switch to new changelog generator

dbkr:
This can just be a package.json dep now, so removing the checks for the tool existing (both of them...)


gsouquet (approved):
That looks good to me 👍

PR #1827, opened 2021-08-04
Support upgrading voice calls to video calls 

SimonBrandner:


src/webrtc/call.ts
```diff
@@ -774,6 +787,49 @@ export class MatrixCall extends EventEmitter {
         this.sendVoipEvent(EventType.CallReject, {});
     }
 
+    /**
+     * Adds a video track to a voice call
+     * @returns new mute state
+     */
```
turt2live, line 793: These docs look copy/pasted :)

src/webrtc/call.ts
```diff
@@ -774,6 +787,49 @@ export class MatrixCall extends EventEmitter {
...
+        audio: boolean, video: boolean,
+    ): Promise<void> {
+        // We don't do call downgrades
+        if (!audio && !video) return;
+        if (!this.opponentSupportsSDPStreamMetadata()) return;
+
+        try {
+            const upgradeAudio = audio && !this.hasLocalUserMediaAudioTrack;
+            const upgradeVideo = video && !this.hasLocalUserMediaVideoTrack;
+            logger.debug(`Upgrading call audio=${upgradeAudio} video=${upgradeVideo}`);
```
turt2live, line 804: The log line isn't clear that the upgrade variables are being used rather than the target audio/video state, so something like this would be appreciated: 
```suggestion
            logger.debug(`Upgrading call: audio?=${upgradeAudio} video?=${upgradeVideo}`);
```

turt2live (commented):
otherwise looks good! Not sure if we want to squeeze this into the RC or if we should delay it. Very keen not to stick it into the release just because we have another breaking change: version numbers are relatively free, and we can afford back-to-back breaking releases.

src/webrtc/call.ts
SimonBrandner: No just old :P

SimonBrandner:
> otherwise looks good! Not sure if we want to squeeze this into the RC or if we should delay it. Very keen not to stick it into the release just because we have another breaking change: version numbers are relatively free, and we can afford back-to-back breaking releases.

Yeah, makes sense, let's wait a bit

src/webrtc/call.ts
```diff
@@ -788,8 +788,7 @@ export class MatrixCall extends EventEmitter {
     }
 
     /**
-     * Adds a video track to a voice call
-     * @returns new mute state
+     * Adds an audio and/or video track - upgrades the call
```
turt2live, line 791: can we get `@param` documentation please?

turt2live (commented):
otherwise looks good. 

and yes, we'll be waiting until after the RC on this.

turt2live (approved):
thanks :)

PR #1828, opened 2021-08-04
Thread store and event parsing to create threads

gsouquet:
This is a first implementation for thread in the `js-sdk`. There is currently no UI associated yet but will follow shortly

Given that the concept of thread will be completely client-side for now, references to the root and tails of the thread are kept, to be able to rebuild the reply chain upon a refresh or so that it can be synchronized between multiple clients

```
EV1 (root)
 |
 |
EV2
 |\__
 |   \ 
EV3   EV4 (tail)
 |
EV5 (tail)
```


spec/unit/models/thread.spec.ts
```diff
@@ -0,0 +1,105 @@
+import { MatrixEvent } from "../../../src/models/event";
```
turt2live, line 1: we should probably start introducing copyright to our tests instead of copy/pasting history, but that might also need to go to the team for general notice.

I'm happy either way for this PR, but definitely something we should deal with *eventually*.

src/models/event.ts
```diff
@@ -190,6 +190,10 @@ export class MatrixEvent extends EventEmitter {
      */
     private txnId: string = null;
 
+    /** A reference to the event ID making the root of the thread
+     */
```
turt2live, line 194: doc alignment mismatch from surrounding code

turt2live (commented):
Generally it looks sane, though a couple things: is this meant to be unstable functionality? If so, annotating the functions aggressively with unstable warnings would be appreciated (to allow for breaking changes during development / discouraging use until it's ready). 

More documentation on the public fields/functions would be appreciated as well, to help with understanding and those who use this thing outside of us.

PR #1829, opened 2021-08-05
Include all TypeScript @types listed in project dependency

gsouquet:
Fixes vector-im/element-web#17743

The package @types/jest was not discovered because if types is specified, only packages listed will be included in the global scope.


Palid (approved):

Palid:
@gsouquet tests aren't passing

PR #1834, opened 2021-08-09
Fix temporary call messages being handled without call 

Palid:
In cases where a rogue client, or just a simple bug, sends a temporary call message, like CallNegotiate the messages should just be discarded if there's no actual p2p connection created.

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -1286,7 +1286,7 @@ export class MatrixCall extends EventEmitter {
         // https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Perfect_negotiation
         const offerCollision = (
             (description.type === 'offer') &&
-            (this.makingOffer || this.peerConn.signalingState != 'stable')
+            (this.makingOffer || this.peerConn.signalingState !== 'stable')
         );
 
         this.ignoreOffer = !polite && offerCollision;
@@ -1934,6 +1934,10 @@ export class MatrixCall extends EventEmitter {
                 }
             }
         }
+    }
+
+    get hasPeerConnection() {
+        return Boolean(this.peerConn);
     }
 }
 
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -1936,7 +1936,7 @@ export class MatrixCall extends EventEmitter {
         }
     }
 
-    get hasPeerConnection() {
+    public get hasPeerConnection() {
         return Boolean(this.peerConn);
     }
 }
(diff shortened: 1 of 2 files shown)


src/webrtc/call.ts
```diff
@@ -1286,7 +1286,7 @@ export class MatrixCall extends EventEmitter {
         // https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Perfect_negotiation
         const offerCollision = (
             (description.type === 'offer') &&
-            (this.makingOffer || this.peerConn.signalingState != 'stable')
+            (this.makingOffer || this.peerConn?.signalingState != 'stable')
```
dbkr, line 1289: I might be missing some context on the errors, but if we're getting negotiate events without a call ongoing this feels like something we should be catching at the top of the function. It's entirely possible, if they race, and doesn't necessarily indicate an error, but rolling it into the state of whether an offer collided or not feels like mixing semantics up.

Palid: I'll try to figure out what's going on here, but that was the easiest solution if we were to merge HMR.

src/webrtc/call.ts
```diff
@@ -1935,6 +1935,10 @@ export class MatrixCall extends EventEmitter {
             }
         }
     }
+
+    get hasPeerConnection() {
```
SimonBrandner, line 1939:
```suggestion
    public get hasPeerConnection() {
```

src/webrtc/callEventHandler.ts
```diff
@@ -255,10 +257,14 @@ export class CallEventHandler {
                     this.calls.delete(content.call_id);
                 }
             }
+            return;
         }
 
-        // The following events need a call
-        if (!call) return;
+        // The following events need a call and a peer connection
+        if (!call && !call.hasPeerConnection) {
```
SimonBrandner, line 264:
```suggestion
        if (!call || !call.hasPeerConnection) {
```

Sounds saner to me, am I missing something?

src/webrtc/callEventHandler.ts
```diff
@@ -255,10 +257,14 @@ export class CallEventHandler {
...
+            logger.warn('Discarding an event', type);
```
SimonBrandner, line 265:
```suggestion
            logger.warn('Discarding an event, we don't have a call/peerConn', type);
```

src/webrtc/callEventHandler.ts
Palid: Oh yeah, right.

SimonBrandner (approved):

dbkr (approved):
Looking at call.js again, the most most likely cause of this was old calls ie. inited with `initWithHangup()`. It is worrying that we're seeing these spurious events at all and if it happened with a non-hung-up call we'd get very strange behaviour, but ok, let's see if we start getting this log message a lot.

PR #1835, opened 2021-08-09
Add support for reading permissions from tree spaces

turt2live:
This was originally thought to be able to be performed in another way, but unfortunately isn't as easy as originally predicted.


dbkr (approved):

dbkr:
merging because it's marked a a release blocker

PR #1838, opened 2021-08-11
fix TURN by fixing regression preventing multiple ICE candidates from sending.

ara4n:
In https://github.com/matrix-org/matrix-js-sdk/commit/eafecd36bc860e79d9aa280a7e60455dbd7c6f67#diff-df33fb6c18a9b5896cd500875824d6c10980d42c92b21cedac6d8f570a8d52b7L1639
the candidateSendTries reset was accidentally lost, meaning that we never send more than one m.call.candidates, breaking trickle ICE.

Also, we leak blank candidates out in m.call.candidates on Safari and FF, so fix that too.

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -1640,9 +1640,14 @@ export class MatrixCall extends EventEmitter {
     }
 
     queueCandidate(content: RTCIceCandidate) {
-        // Sends candidates with are sent in a special way because we try to amalgamate
-        // them into one message
-        this.candidateSendQueue.push(content);
+        // We partially de-trickle candidates by waiting for `delay` before sending them
+        // amalgamated, in order to avoid sending too many m.call.candidates events and hitting
+        // rate limits in Matrix.
+        // In practice, it'd be better to remove rate limits for m.call.*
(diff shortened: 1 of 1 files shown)


gsouquet (approved):
Looks sane, 💯

src/webrtc/call.ts
```diff
@@ -1640,9 +1640,14 @@ export class MatrixCall extends EventEmitter {
...
 
     queueCandidate(content: RTCIceCandidate) {
-        // Sends candidates with are sent in a special way because we try to amalgamate
-        // them into one message
-        this.candidateSendQueue.push(content);
+        // We partially de-trickle candidates by waiting for `delay` before sending them
+        // amalgamated, in order to avoid sending too many m.call.candidates events and hitting
+        // rate limits in Matrix.
+        // In practice, it'd be better to remove rate limits for m.call.*
+        // ensure the 'ICE complete' dummy event never leaks out into m.call.candidates
```
dbkr, line 1647: We deliberately send this to signal the end of candidates to the other side (often it's important, eg. if we're bridging to something without trickle ICE and the bridge has to wait for all the candidates).

PR #1842, opened 2021-08-12
Fix the types in shipped package

dbkr:
1829 added the spec files to 'src', presumably because they just
were't type checked otherwise, but this chaqnged the implicit rootDir
so the lib directory we shipped has types in src/ and spec/ which
meant lib/index.d.ts was at lib/src/index.d.ts.

Specify the rootDir on the build tsconfig to put it back in the
right place, and also try to make the base tsconfig never emit
to avoid it ever generating .d.ts files in silly places. Move all
the options related to emitting to -build since they aren't relevant
in the main one.

Is this a sensible way to do this? It doesn't feel terribly elegant.

Fixes https://github.com/vector-im/element-web/issues/18503
Regressed in https://github.com/matrix-org/matrix-js-sdk/pull/1829

---

## 🐛 Bug Fixes
 * Fix the types in shipped package ([\#1842](https://github.com/matrix-org/matrix-js-sdk/pull/1842)). Fixes vector-im/element-web#18503.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "clean": "rimraf lib dist",
     "build": "yarn build:dev && yarn build:compile-browser && yarn build:minify-browser",
     "build:dev": "yarn clean && git rev-parse HEAD > git-revision.txt && yarn build:compile && yarn build:types",
-    "build:types": "tsc --emitDeclarationOnly",
+    "build:types": "tsc -p tsconfig-build.json --emitDeclarationOnly",
     "build:compile": "babel -d lib --verbose --extensions \".ts,.js\" src",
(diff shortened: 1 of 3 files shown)


Palid:
@dbkr I think in this case the easiest solution would be to make a `base.tsconfig.json` somewhere else (like we have with eslint?) and just extend it everywhere we need. 
Current solution isn't bad, but I wouldn't remove all those compiler options from the `tsconfig.json`, as it may break type reporting if you were to run `typescript:watch` task in vscode.

turt2live (approved):
This is unfortunately the way you do this on TS.

dbkr:
Fair enough :/

PR #1846, opened 2021-08-13
SSSSCryptoCallbacks.getSecretStorageKey return Promise<null> if no key was found

### LINKED ISSUE #1848: SSSSCryptoCallbacks.getSecretStorageKey may return Promise<void>
**Describe the bug**
If no result is found, the function may reach the last line without return ever being called.
This violates the function's TypeScript signature.

jaller94:
Fixes #1848

If no result is found, the function may reach the last line without `return` ever being called.

Note: SSSSCryptoCallbacks.getSecretStorageKey now returns Promise<null> instead of Promise<void>, if no key was found.

---

Add one of: `T-Enhancement`, `T-Defect`, `T-Task` to indicate what type of change this is plus `X-Breaking-Change` if it's a breaking change.

### DIFF
--- a/src/crypto/EncryptionSetup.ts
+++ b/src/crypto/EncryptionSetup.ts
@@ -356,7 +356,7 @@ class SSSSCryptoCallbacks {
     public async getSecretStorageKey(
         { keys }: { keys: Record<string, ISecretStorageKeyInfo> },
         name: string,
-    ): Promise<[string, Uint8Array]> {
+    ): Promise<[string, Uint8Array]|void> {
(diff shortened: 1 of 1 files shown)


Palid:
@jaller94 can you satisfy our CI? Seems like a very good change. We're lacking `strictNullChecks` in our typescript configs so far, so finding this is very useful!

jaller94:
@Palid It's happy now.
I changed it to match the signature of ICryptoCallbacks and return `null` instead of `void`.

Palid (approved):
LGTM

PR #1847, opened 2021-08-13
SSSSCryptoCallbacks.getSecretStorageKey: Test if delegateCryptoCallbacks has getSecretStorageKey

### LINKED ISSUE #1845: SSSSCryptoCallbacks: this.delegateCryptoCallbacks.getSecretStorageKey is not a function
**Describe the bug**
`delegateCryptoCallbacks.getSecretStorageKey` is used without a check of whether it is defined.

https://github.com/matrix-org/matrix-js-sdk/blob/a69d9a9f20b680e3710e0f43cca68a2ce31ba030/src/crypto/EncryptionSetup.ts#L369-L370

**To Reproduce**
1. Call `createClient` with `cryptoCallbacks: undefined` or `cryptoCallbacks: {}` (both are valid and don't have a `getSecretStorageKey` property)

**Expected behavior**
It either works or a proper error is thrown.

** Additional notes**
`ICryptoCallbacks.getSecretStorageKey` is declared optional.
https://github.com/matrix-org/matrix-js-sdk/blob/b63de6a902a9e1f8ffd7697dea33820fc04f028e/src/matrix.ts#L109

jaller94:
Fixes #1845 

Notes: Fix a JavaScript error on matrixClient.bootstrapSecretStorage() if matrixClient.cryptoCallbacks.getSecretStorageKey is not defined.

---

### DIFF
--- a/src/crypto/EncryptionSetup.ts
+++ b/src/crypto/EncryptionSetup.ts
@@ -365,7 +365,7 @@ class SSSSCryptoCallbacks {
         }
         // if we don't have the key cached yet, ask
         // for it to the general crypto callbacks and cache it
-        if (this.delegateCryptoCallbacks?.getSecretStorageKey) {
+        if (this?.delegateCryptoCallbacks?.getSecretStorageKey) {
             const result = await this.delegateCryptoCallbacks.
                 getSecretStorageKey({ keys }, name);
             if (result) {
--- a/src/crypto/EncryptionSetup.ts
+++ b/src/crypto/EncryptionSetup.ts
@@ -365,7 +365,7 @@ class SSSSCryptoCallbacks {
         }
         // if we don't have the key cached yet, ask
         // for it to the general crypto callbacks and cache it
-        if (this.delegateCryptoCallbacks) {
+        if (this.delegateCryptoCallbacks && this.delegateCryptoCallbacks.getSecretStorageKey) {
             const result = await this.delegateCryptoCallbacks.
                 getSecretStorageKey({ keys }, name);
             if (result) {
--- a/src/crypto/EncryptionSetup.ts
+++ b/src/crypto/EncryptionSetup.ts
@@ -365,7 +365,7 @@ class SSSSCryptoCallbacks {
         }
         // if we don't have the key cached yet, ask
         // for it to the general crypto callbacks and cache it
-        if (this.delegateCryptoCallbacks && this.delegateCryptoCallbacks.getSecretStorageKey) {
+        if (this.delegateCryptoCallbacks?.getSecretStorageKey) {
             const result = await this.delegateCryptoCallbacks.
                 getSecretStorageKey({ keys }, name);
             if (result) {


src/crypto/EncryptionSetup.ts
```diff
@@ -365,7 +365,7 @@ class SSSSCryptoCallbacks {
         }
         // if we don't have the key cached yet, ask
         // for it to the general crypto callbacks and cache it
-        if (this.delegateCryptoCallbacks) {
+        if (this.delegateCryptoCallbacks && this.delegateCryptoCallbacks.getSecretStorageKey) {
```
Palid, line 368:
```suggestion
        if (this?.delegateCryptoCallbacks?.getSecretStorageKey) {
```
Makes it slightly easier for the eyes. :)

jaller94: No need to check `this`.
```suggestion
        if (this.delegateCryptoCallbacks?.getSecretStorageKey) {
```

src/crypto/EncryptionSetup.ts
```diff
@@ -365,7 +365,7 @@ class SSSSCryptoCallbacks {
         }
         // if we don't have the key cached yet, ask
         // for it to the general crypto callbacks and cache it
-        if (this.delegateCryptoCallbacks) {
+        if (this.delegateCryptoCallbacks?.getSecretStorageKey) {
```
Palid, line 368: can `delegateCryptoCallbacks` fail? If yes we need to make `this?.delegateCryptoCallbacks?.getSecretStorageKey` instead to check for both cases.

src/crypto/EncryptionSetup.ts
Palid: @jaller94 Could you change it in line 381 too, so the behavior's the same in both if we don't need to check for `this`? It seems that it'll always be defined, but may be an empty object sometimes.

src/crypto/EncryptionSetup.ts
jaller94: As long as `this` is truthy, it won't throw.

jaller94, line 368: According to the constructor, we shouldn't need this `?` as `delegateCryptoCallbacks` must be a `ICryptoCallbacks`.

There's another issue that
* either the constructor's signature of `SSSSCryptoCallbacks` and `EncryptionSetupBuilder` should allow `ICryptoCallbacks|undefined` or
* [`MatrixClient.cryptoCallbacks`](https://github.com/matrix-org/matrix-js-sdk/blob/a69d9a9f20b680e3710e0f43cca68a2ce31ba030/src/client.ts#L315) must not be undefined.

It's handed over here.
https://github.com/matrix-org/matrix-js-sdk/blob/a69d9a9f20b680e3710e0f43cca68a2ce31ba030/src/crypto/index.ts#L616-L620

jaller94: https://github.com/matrix-org/matrix-js-sdk/pull/1847/files#diff-2ed9f095536152a0f96b01826f2bc6fa8d04591b7c5bc02782804d4e12bdbc31R382

regards `this.delegateCryptoCallbacks` as optional, so I guess version 1 is/was supposed to be the truth.

jaller94: As that's yet a different issue, I solved that in a separate PR: #1849

src/crypto/EncryptionSetup.ts
jaller94: If #1849 gets accepted, `delegateCryptoCallbacks` becomes optional and we should keep them like this.

src/crypto/EncryptionSetup.ts
Palid: @jaller94 so let's go back to checking `this?.delegateCryptoCallbacks?.getSecretStorageKey`. You can just click commit suggestion. 😄

jaller94: I don't agree. According to our types, `this` and `this.delegateCryptoCallbacks` are safe.
However, it doesn't hurt to add the question marks.

```suggestion
        if (this?.delegateCryptoCallbacks?.getSecretStorageKey) {
```

jaller94:
@Palid Could you have another look at this?

Palid (approved):

Palid:
> @Palid Could you have another look at this?

LGTM. Sorry!

PR #1849, opened 2021-08-13
Make delegateCryptoCallbacks an optional parameter of EncryptionSetupBuilder and SSSSCryptoCallbacks

jaller94:
https://github.com/matrix-org/matrix-js-sdk/blob/a69d9a9f20b680e3710e0f43cca68a2ce31ba030/src/client.ts#L315

`cryptoCallbacks` may be undefined but the constructor of `EncryptionSetupBuilder` requires it to be defined.

https://github.com/matrix-org/matrix-js-sdk/blob/a69d9a9f20b680e3710e0f43cca68a2ce31ba030/src/crypto/index.ts#L616-L620

---

### DIFF
--- a/src/crypto/EncryptionSetup.ts
+++ b/src/crypto/EncryptionSetup.ts
@@ -351,7 +351,7 @@ class CrossSigningCallbacks implements ICryptoCallbacks, ICacheCallbacks {
 class SSSSCryptoCallbacks {
     private readonly privateKeys = new Map<string, Uint8Array>();
 
-    constructor(private readonly delegateCryptoCallbacks: ICryptoCallbacks) {}
+    constructor(private readonly delegateCryptoCallbacks?: ICryptoCallbacks) {}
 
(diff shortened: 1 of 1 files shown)


Palid (approved):
LGTM, even though it seems like we should instead fix the instances to make sure that it's not undefined?

jaller94:
If these classes don't make sense without the callbacks, we should not create the builder, if we don't have `delegateCryptoCallbacks` is undefined/null.

This line should be conditional:
https://github.com/matrix-org/matrix-js-sdk/blob/a69d9a9f20b680e3710e0f43cca68a2ce31ba030/src/crypto/index.ts#L616-L620

jaller94:
@Palid I think this is the right thing to do. I was worried that the classes may not make sense without the callbacks, but they seem to have other functionality.

Palid (approved):

Palid:
We can always revert this if it turns out to be a problem.

PR #1861, opened 2021-08-17
Fix release script to handle no signing ID

dbkr:
We now use the file for stuff other than just the signing ID, so
the file may be present but with no signing ID: handle this case.

---


dbkr:
...except apparently I failed to change the target branch

PR #1865, opened 2021-08-19
Give `CallFeed` the capability to emit on volume changes

SimonBrandner:
Type: enhancement 
For https://github.com/matrix-org/matrix-react-sdk/pull/6639

<hr>

To be transparent, I have no idea what I am doing but it seems to work :sweat_smile: 

---

## ✨ Features
 * Give `CallFeed` the capability to emit on volume changes ([\#1865](https://github.com/matrix-org/matrix-js-sdk/pull/1865)). Contributed by [SimonBrandner](https://github.com/SimonBrandner).

### DIFF
--- a/src/@types/global.d.ts
+++ b/src/@types/global.d.ts
@@ -34,6 +34,7 @@ declare global {
 
     interface Window {
         electron?: Electron;
+        webkitAudioContext: typeof AudioContext;
     }
 
     interface Electron {
--- a/src/webrtc/callFeed.ts
+++ b/src/webrtc/callFeed.ts
@@ -121,16 +121,28 @@ export class CallFeed extends EventEmitter {
         }
     }
 
+    /**
+     * Set feed's internal audio mute state
+     * @param muted is the feed's audio muted?
+     */
     public setAudioMuted(muted: boolean): void {
         this.audioMuted = muted;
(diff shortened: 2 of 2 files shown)


dbkr (commented):
Nice! It might be nice to actually shift more of this towards the js-sdk side and have it emit active-speaker events so the js-sdk handles the threshold etc rather than just throwing volume events at the app. Also unsure if we'll want to highlight all people who are actively speaking or just decide which person is the active speaker (like jitsi does). In a way I sort of like highlighting all the people we think are speaking, but probably a product call.

SimonBrandner:
> It might be nice to actually shift more of this towards the js-sdk side and have it emit active-speaker events so the js-sdk handles the threshold etc rather than just throwing volume events at the app.

It was intentional in case some other app building on the js-sdk can access the volume directly but we can emit both `volume_changed` and speaking `speaking` and solve the problem that way

> Also unsure if we'll want to highlight all people who are actively speaking or just decide which person is the active speaker (like jitsi does). In a way I sort of like highlighting all the people we think are speaking, but probably a product call.

Agreed, I would also prefer highlighting all speaking users. Guessing the main speaker feels a bit useless now and it doesn't feel like an important thing to have. We can always add it in the future. Figma also seems to agree: _When a user speaks, its videofeed should be highlighted in Accent colour_

dbkr (approved):

PR #1872, opened 2021-08-26
Count notifications in encrypted rooms client-side

SimonBrandner:
Fixes https://github.com/vector-im/element-web/issues/15393
Closes https://github.com/matrix-org/matrix-react-sdk/pull/6267
Type: enhancement
element-web notes: Don't notify for hidden events sent during calls

<hr>

For the reviewer: This is a change that has a big potential to break things - please review carefully

---

## ✨ Features
 * Count notifications in encrypted rooms client-side ([\#1872](https://github.com/matrix-org/matrix-js-sdk/pull/1872)). Fixes vector-im/element-web#15393 and vector-im/element-web#15393. Contributed by [SimonBrandner](https://github.com/SimonBrandner).

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -838,36 +838,7 @@ export class MatrixClient extends EventEmitter {
         // We do this so that push rules are correctly executed on events in their decrypted
         // state, such as highlights when the user's name is mentioned.
         this.on("Event.decrypted", (event: MatrixEvent) => {
-            const oldActions = event.getWirePushActions();
-            const actions = this.pushProcessor.actionsForEvent(event);
-            event.setPushActions(actions); // Might as well while we're here
-
-            const room = this.getRoom(event.getRoomId());
-            if (!room) return;
-
-            const currentHighlightCount = room.getUnreadNotificationCount(NotificationCountType.Highlight);
-
-            // Ensure the unread counts are kept up to date if the event is encrypted
-            // We also want to make sure that the notification count goes up if we already
-            // have encrypted events to avoid other code from resetting 'highlight' to zero.
-            const oldHighlight = Boolean(oldActions?.tweaks?.highlight);
-            const newHighlight = Boolean(actions?.tweaks?.highlight);
-            if (oldHighlight !== newHighlight || currentHighlightCount > 0) {
-                // TODO: Handle mentions received while the client is offline
-                // See also https://github.com/vector-im/element-web/issues/9069
-                if (!room.hasUserReadEvent(this.getUserId(), event.getId())) {
-                    let newCount = currentHighlightCount;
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -837,39 +837,8 @@ export class MatrixClient extends EventEmitter {
...
-            // Ensure the unread counts are kept up to date if the event is encrypted
-            // We also want to make sure that the notification count goes up if we already
-            // have encrypted events to avoid other code from resetting 'highlight' to zero.
-            const oldHighlight = oldActions && oldActions.tweaks
-                ? !!oldActions.tweaks.highlight : false;
-            const newHighlight = actions && actions.tweaks
-                ? !!actions.tweaks.highlight : false;
-            if (oldHighlight !== newHighlight || currentCount > 0) {
-                // TODO: Handle mentions received while the client is offline
-                // See also https://github.com/vector-im/element-web/issues/9069
```
turt2live, line 859: This TODO and issue reference I think need to stay still, as we don't handle limited (gappy) syncs, even in the new code.

src/client.ts
```diff
@@ -837,39 +837,8 @@ export class MatrixClient extends EventEmitter {
...
-                if (!room.hasUserReadEvent(this.getUserId(), event.getId())) {
-                    let newCount = currentCount;
-                    if (newHighlight && !oldHighlight) newCount++;
-                    if (!newHighlight && oldHighlight) newCount--;
```
turt2live, line 863: I swear there was a reason why we did a `-1` in addition to `+1`, but clearly 2019 me didn't bother to document it. Not really worth exploring in this PR, but something to watch for after this lands: it'll probably be some subtle bug that has been supressed from my nightmares for 2 years.

turt2live: Though, a quick manual test of scrolling up in an encrypted room with messages sitting in that backlog would be appreciated as I think it might be for that?

turt2live (commented):
Overall looks fine. Terrifying, but fine.

src/client.ts
SimonBrandner: Oops! Sorry

src/client.ts
SimonBrandner: Scrolling through an encrypted room isn't causing any weirdness for me. I am also a bit scared about what this could cause. Also, I am away next week so if this sets Element on fire I won't be there to fix it :sweat_smile: So feel free to delay this or whatever you find suitable

turt2live (approved):
The RC is scheduled for Tuesday, so for maximum testing let's land it after the branches are cut.

turt2live:
Reverting this: found out what the `-1` did. It was for reading the room on other devices, decrementing the unread count. 

cc @SimonBrandner

PR #1874, opened 2021-08-28
Improve TypeScript in `MatrixCall` 

SimonBrandner:
Type: task

---

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -31,14 +31,21 @@ import { randomString } from '../randomstring';
 import {
     MCallReplacesEvent,
     MCallAnswer,
-    MCallOfferNegotiate,
+    MCallInviteNegotiate,
     CallCapabilities,
     SDPStreamMetadataPurpose,
     SDPStreamMetadata,
     SDPStreamMetadataKey,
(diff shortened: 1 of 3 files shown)


src/webrtc/call.ts
```diff
@@ -330,33 +339,14 @@ export class MatrixCall extends EventEmitter {
...
-        this.usermediaSenders = [];
-        this.screensharingSenders = [];
     }
 
     /**
      * Place a voice call to this room.
      * @throws If you have not specified a listener for 'error' events.
      */
-    async placeVoiceCall() {
+    async placeVoiceCall(): Promise<void> {
```
t3chguy, line 349: public?

src/webrtc/call.ts
```diff
@@ -368,19 +358,19 @@ export class MatrixCall extends EventEmitter {
      * Place a video call to this room.
      * @throws If you have not specified a listener for 'error' events.
      */
-    async placeVideoCall() {
+    async placeVideoCall(): Promise<void> {
```
t3chguy, line 361: public?

t3chguy (approved):
Looks sane to me otherwise

Issue #1645, opened 2021-03-13
DesktopCapturerSource is not properly exported


SimonBrandner:
**Describe the bug**
Creating a new TS project and running `tsc` results in:
```ts
node_modules/matrix-js-sdk/lib/webrtc/call.d.ts:122:68 - error TS2304: Cannot find name 'DesktopCapturerSource'.

122 export declare function getDesktopCapturerSources(): Promise<Array<DesktopCapturerSource>>;
                                                                              ~~~~~~~~~~~~~~~~~~~~~

node_modules/matrix-js-sdk/lib/webrtc/call.d.ts:202:146 - error TS2304: Cannot find name 'DesktopCapturerSource'.

202     placeScreenSharingCall(remoteVideoElement: HTMLVideoElement, localVideoElement: HTMLVideoElement, selectDesktopCapturerSource: () => Promise<DesktopCapturerSource>): Promise<void>;
                                                                                                                                                                   ~~~~~~~~~~~~~~~~~~~~
```

**To Reproduce**
+ `npm init`
+ `npm install typescript matrix-js-sdk node @types/node`
+ Create file `main.ts` with `import * as sdk from "matrix-js-sdk"`
+ `npx tsc main.ts`
+ See errors

**Additional context**
This seems to be caused by the fact that [`DesktopCapturerSource` isn't properly exported.](https://github.com/matrix-org/matrix-js-sdk/blob/0b7b35f80084f5c46ac7a20bb4539049654106f3/src/%40types/global.d.ts#L59). I don't know how it should be exported. Some of the declarations in `global.d.ts` depend on it so it can't be moved to `call.ts`. Adding the export keyword before the declaration doesn't solve the problem. It used to be that way and the problem was present

enekonieto:
Same happens to me with 9.9.0

enekonieto:
I got it working! Changes I did:

Added in `src/polyfills.ts` of my Angular app:
```
(window as any)['global'] = window;
```

Added in `node_modules/matrix-js-sdk/lib/webrtc/call.d.ts`:
```
interface DesktopCapturerSource {
    id: string;
    name: string;
    thumbnailURL: string;
}
```

jryans:
React SDK seems to avoid hitting this because it explicitly imports JS SDK's global types:

```js
import "matrix-js-sdk/src/@types/global"
```

It seems strange to make every downstream package do that though...

jryans:
[...]

PR #1878, opened 2021-09-01
Remove Element-specifc screen-sharing code out of the js-sdk 

SimonBrandner:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/1645
**Requires https://github.com/matrix-org/matrix-react-sdk/pull/6727**
Type: task

<hr>

Notes: This changes the `selectDesktopCapturerSource` callback in `setScreensharingEnabled()` to a `string` parameter `desktopCapturerSourceId`. This way you can handle choosing a source more easily on your side and just pass the id of the chosen source. This also removes the `getDesktopCapturerSources()` method which was Element specific and shouldn't have been in the `js-sdk` in the first place.

<hr>

This fixes some of my awful 8 months old code.

---

### DIFF
--- a/src/@types/global.d.ts
+++ b/src/@types/global.d.ts
@@ -33,12 +33,7 @@ declare global {
     }
 
     interface Window {
-        electron?: Electron;
         webkitAudioContext: typeof AudioContext;
-    }
-
-    interface Electron {
-        getDesktopCapturerSources(options: GetSourcesOptions): Promise<Array<DesktopCapturerSource>>;
     }
 
     interface Crypto {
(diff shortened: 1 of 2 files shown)


gsouquet (approved):

gsouquet:
Code wise this looks fine, however we might want to flag this as a breaking change as it changes some public function signatures

Having this pull request as a task means that it won't be included in the changelog. Looping @dbkr to make sure we take the correct approach to notify developers of this change

Palid:
Could we just add a little deprecation warning instead and remove it a quarter later or so? This seems like the safest way to go with breaking changes for the SDK.

SimonBrandner:
Please do not merge, I think we can do a bit better. Will comment on the other stuff when I get home

SimonBrandner:
> Could we just add a little deprecation warning instead and remove it a quarter later or so? This seems like the safest way to go with breaking changes for the SDK.

I'd like to but that is probably much more trouble than what it is worth, we might need to implement several helper methods and it would just be a pain :(

PR #1884, opened 2021-09-07
Convert OlmDevice to Typescript

t3chguy:
Requires https://github.com/matrix-org/matrix-react-sdk/pull/6756

https://github.com/matrix-org/matrix-js-sdk/commit/68a2762b75b92710df08c6b0c1d5a57937b973f0 seems to fix a possible bug

---


src/crypto/index.ts
```diff
@@ -130,13 +130,14 @@ export interface IRoomKeyRequestBody extends IRoomKey {
 }
 
 export interface IMegolmSessionData {
+    [key: string]: any;
```
gsouquet, line 133: That sounds fairly dangerous? If I understand correctly we can set any property to any value?

Is it by any chance possible to know the type ahead of time here?

src/crypto/OlmDevice.ts
```diff
@@ -0,0 +1,1518 @@
...
+}
+
+/**
+ * data stored in the session store about an inbound group session
+ *
+ * @typedef {Object} InboundGroupSessionData
+ * @property {string} room_id
+ * @property {string} session   pickled Olm.InboundGroupSession
+ * @property {Object<string, string>} keysClaimed
+ * @property {Array<string>} forwardingCurve25519KeyChain  Devices involved in forwarding
```
gsouquet, line 79: Looks like this comment is missing a couple properties that are now listed in the interface

src/crypto/OlmDevice.ts
```diff
@@ -0,0 +1,1518 @@
...
+        this.deviceCurve25519Key = e2eKeys.curve25519;
+        this.deviceEd25519Key = e2eKeys.ed25519;
+    }
+
+    /**
+     * Populates the crypto store using data that was exported from an existing device.
+     * Note that for now only the “account” and “sessions” stores are populated;
+     * Other stores will be as with a new device.
+     *
+     * @param {Object} exportedData Data exported from another device
```
gsouquet, line 236:
```suggestion
     * @param {IExportedDevice} exportedData Data exported from another device
```

gsouquet (approved):
Looks pretty good to me, awesome job migrating this to TypeScript 👏 !

I noticed a couple of disconnect between the types defined in JSDoc and the ones defined in TypeScript. I stopped commenting on them as I don't feel like this PR is the right forum to discuss this, but I'll bring that up internally to see how we want to approach that in the future

src/crypto/index.ts
t3chguy: Well it can be extended with arbitrary fields which are namespaced, js-sdk might not handle all of them but consumers might.

Palid:
Wouldn't it make sense to slightly modify the API so that use extensions would have to be separately namespaced in a map or so? This probably is too big of a change for typescript migration though.On Sep 7, 2021 15:28, Michael Telatynski ***@***.***> wrote:
@t3chguy commented on this pull request.



In src/crypto/index.ts:
> @@ -130,13 +130,14 @@ export interface IRoomKeyRequestBody extends IRoomKey {
 }
 
 export interface IMegolmSessionData {
+    [key: string]: any;

Well it can be extended with arbitrary fields which are namespaced, js-sdk might not handle all of them but consumers might.

t3chguy:
That would be a rather large change given there are around 10 different methods returning/accepting that iface, and would be a breaking change at that. Also inconsistent with our other usages of openly extensible structures

gsouquet (approved):

PR #1661, opened 2021-04-09
Fix calling addEventListener if it does not exist

janpawellek:
Some platforms (e.g. React Native) register global.window, but do not have global.window.addEventListener. In this case, this function should not be called.

Signed-off-by: Jan Pawellek <jpawellek@churchtools.de>

PR #1886, opened 2021-09-07
Fix calling removeEventListener if it does not exist

janpawellek:
Some platforms (e.g. React Native) register global.window, but do not have global.window.removeEventListener. In this case, this function should not be called.

This has already been fixed for addEventListener (https://github.com/matrix-org/matrix-js-sdk/pull/1661), however not yet for removeEventListener.

Signed-off-by: Jan Pawellek jpawellek@churchtools.de

---

### DIFF
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -704,7 +704,7 @@ export class SyncApi {
      */
     public stop(): void {
         debuglog("SyncApi.stop");
-        if (global.window) {
+        if (global.window && global.window.removeEventListener) {
             global.window.removeEventListener("online", this.onOnline, false);
         }
         this.running = false;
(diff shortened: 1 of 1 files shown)


gsouquet:
Awesome, thank you for your contribution!
Could you maybe add a comment explaining the reasons behind this, as it will probably not be obvious and could get pruned inadvertently

gsouquet (approved):

janpawellek:
> Awesome, thank you for your contribution!
> Could you maybe add a comment explaining the reasons behind this, as it will probably not be obvious and could get pruned inadvertently

Hi @gsouquet  
Sure, thank you for your review! I just added a comment why the check is necessary.

PR #1900, opened 2021-09-09
Disable pending events for threads temporarily

gsouquet:
Pre-requisite for vector-im/element-web#18947

- Make timelineSet public readonly
- Add helper to check threads support
- Disable pending events for threads

---


turt2live (approved):
looks like the CI isn't happy though

PR #1902, opened 2021-09-09
Group Calls (WIP)

robertlong:
# Group Calls

This PR expands upon Matrix calls and adds support for establishing calls to multiple members in a room. The code stems from my work on [matrix-video-chat](https://github.com/vector-im/matrix-video-chat).

### `GroupCall`

Created by calling `client.createGroupCall(roomId, type)` where type can be either `"voice"` or `"video"`. This class essentially sets up a bunch of `MatrixCall`s to room members who have set a participant state event.

### `GroupCallParticipant`

This is a wrapper around `MatrixCall` that represents the state for a participant who has entered a `GroupCall`.

### Addition of the `invitee` field to `m.call.invite`

This `invitee` field allows the recipients of the `m.call.invite` event to filter out invites not intended for them.

### Ice Disconnected Timeout

Rather than wait on the Ice `failed` connection state, I added a 30 second timeout which ensures that calls hang up if you've been disconnected for too long. If the connection state changes within this 30 second period, this timeout resets.

### Active Speaker

@SimonBrandner recently added speaking and volume events to `CallFeed`s, this PR uses that work to track the "active speaker" in a group call. This is primarily intended to be used to show what user should be focused in a group call.

### WebRTC Datachannels

Adding datachannels to calls is currently possible by calling `call.peerConn.createDataChannel()`. This PR just formalizes that process by adding a public API method for doing so and a `datachannel` event for listening for new datachannels on a `MatrixCall` or `GroupCall`.

## To Do:
- [x] Split out unrelated code into separate PRs
- [x] Refactor `getLocalAudioStream`/ `getLocalVideoStream`
- [x] Revert `CallFeed` export
- [ ] Figure out final conference call join/leave/presence events (device events etc.)
- [ ] Write an MSC for the new group call events and `invitee` field.
- [x] Fix known regression with microphone/video mute state
- [ ] Add docs for all new public methods/strings/classes
- [ ] Write release notes

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1026,7 +1026,6 @@ export class MatrixClient extends EventEmitter {
 
         this.callEventHandler?.stop();
         this.groupCallEventHandler?.stop();
-        this.callEventHandler = null;
 
         global.clearInterval(this.checkTurnServersIntervalID);
         if (this.clientWellKnownIntervalID !== undefined) {
--- a/src/client.ts
+++ b/src/client.ts
@@ -1028,6 +1028,8 @@ export class MatrixClient extends EventEmitter {
         this.groupCallEventHandler?.stop();
         this.callEventHandler = null;
 
+        this.mediaHandler.stop();
+
         global.clearInterval(this.checkTurnServersIntervalID);
         if (this.clientWellKnownIntervalID !== undefined) {
             global.clearInterval(this.clientWellKnownIntervalID);
--- a/src/client.ts
+++ b/src/client.ts
@@ -4632,7 +4632,12 @@ export class MatrixClient extends EventEmitter {
                     const stateEvents = res.state.map(this.getEventMapper());
                     room.currentState.setUnknownStateEvents(stateEvents);
                 }
-                room.addEventsToTimeline(matrixEvents, true, room.getLiveTimeline());
+
+                const [timelineEvents, threadedEvents] = this.partitionThreadedEvents(matrixEvents);
+
+                room.addEventsToTimeline(timelineEvents, true, room.getLiveTimeline());
+                this.processThreadEvents(room, threadedEvents);
+
                 room.oldState.paginationToken = res.end;
                 if (res.chunk.length === 0) {
                     room.oldState.paginationToken = null;
@@ -4738,7 +4743,11 @@ export class MatrixClient extends EventEmitter {
                 const stateEvents = res.state.map(this.getEventMapper());
                 timeline.getState(EventTimeline.BACKWARDS).setUnknownStateEvents(stateEvents);
             }
-            timelineSet.addEventsToTimeline(matrixEvents, true, timeline, res.start);
+
+            const [timelineEvents, threadedEvents] = this.partitionThreadedEvents(matrixEvents);
+
+            timelineSet.addEventsToTimeline(timelineEvents, true, timeline, res.start);
+            this.processThreadEvents(timelineSet.room, threadedEvents);
 
             // there is no guarantee that the event ended up in "timeline" (we
             // might have switched to a neighbouring timeline) - so check the
@@ -4871,8 +4880,11 @@ export class MatrixClient extends EventEmitter {
                     matrixEvents[i] = event;
                 }
 
-                eventTimeline.getTimelineSet()
-                    .addEventsToTimeline(matrixEvents, backwards, eventTimeline, token);
+                const [timelineEvents, threadedEvents] = this.partitionThreadedEvents(matrixEvents);
+
+                const timelineSet = eventTimeline.getTimelineSet();
+                timelineSet.addEventsToTimeline(timelineEvents, backwards, eventTimeline, token);
+                this.processThreadEvents(timelineSet.room, threadedEvents);
 
                 // if we've hit the end of the timeline, we need to stop trying to
                 // paginate. We need to keep the 'forwards' token though, to make sure
@@ -4905,8 +4917,12 @@ export class MatrixClient extends EventEmitter {
                 }
                 const token = res.end;
                 const matrixEvents = res.chunk.map(this.getEventMapper());
+
+                const [timelineEvents, threadedEvents] = this.partitionThreadedEvents(matrixEvents);
+
                 eventTimeline.getTimelineSet()
-                    .addEventsToTimeline(matrixEvents, backwards, eventTimeline, token);
+                    .addEventsToTimeline(timelineEvents, backwards, eventTimeline, token);
+                this.processThreadEvents(room, threadedEvents);
 
                 // if we've hit the end of the timeline, we need to stop trying to
                 // paginate. We need to keep the 'forwards' token though, to make sure
@@ -5476,6 +5492,14 @@ export class MatrixClient extends EventEmitter {
         const resultsLength = roomEvents.results ? roomEvents.results.length : 0;
         for (let i = 0; i < resultsLength; i++) {
             const sr = SearchResult.fromJson(roomEvents.results[i], this.getEventMapper());
+            const room = this.getRoom(sr.context.getEvent().getRoomId());
+            if (room) {
+                // Copy over a known event sender if we can
+                for (const ev of sr.context.getTimeline()) {
+                    const sender = room.getMember(ev.getSender());
+                    if (!ev.sender && sender) ev.sender = sender;
+                }
+            }
             searchResults.results.push(sr);
         }
         return searchResults;
@@ -6046,7 +6070,9 @@ export class MatrixClient extends EventEmitter {
         if (fetchedEventType === EventType.RoomMessageEncrypted) {
             const allEvents = originalEvent ? events.concat(originalEvent) : events;
             await Promise.all(allEvents.map(e => {
-                return new Promise(resolve => e.once("Event.decrypted", resolve));
+                if (e.isEncrypted()) {
+                    return new Promise(resolve => e.once("Event.decrypted", resolve));
+                }
             }));
             events = events.filter(e => e.getType() === eventType);
         }
@@ -8597,6 +8623,60 @@ export class MatrixClient extends EventEmitter {
             qsStringifyOptions: { arrayFormat: 'repeat' },
             prefix: "/_matrix/client/unstable/im.nheko.summary",
         });
+    }
+
+    public partitionThreadedEvents(events: MatrixEvent[]): [MatrixEvent[], MatrixEvent[]] {
+        // Indices to the events array, for readibility
+        const ROOM = 0;
+        const THREAD = 1;
+        const threadRoots = new Set<string>();
+        if (this.supportsExperimentalThreads()) {
+            return events.reduce((memo, event: MatrixEvent) => {
+                const room = this.getRoom(event.getRoomId());
+                // An event should live in the thread timeline if
+                // - It's a reply in thread event
+                // - It's related to a reply in thread event
+                let shouldLiveInThreadTimeline = event.isThreadRelation;
+                if (shouldLiveInThreadTimeline) {
+                    threadRoots.add(event.relationEventId);
+                } else {
+                    const parentEventId = event.parentEventId;
+                    const parentEvent = room?.findEventById(parentEventId) || events.find((mxEv: MatrixEvent) => {
+                        return mxEv.getId() === parentEventId;
+                    });
+                    shouldLiveInThreadTimeline = parentEvent?.isThreadRelation;
+
+                    // Copy all the reactions and annotations to the root event
+                    // to the thread timeline. They will end up living in both
+                    // timelines at the same time
+                    const targetingThreadRoot = parentEvent?.isThreadRoot || threadRoots.has(event.relationEventId);
+                    if (targetingThreadRoot && !event.isThreadRelation && event.relationEventId) {
+                        memo[THREAD].push(event);
+                    }
+                }
+                const targetTimeline = shouldLiveInThreadTimeline ? THREAD : ROOM;
+                memo[targetTimeline].push(event);
+                return memo;
+            }, [[], []]);
+        } else {
+            // When `experimentalThreadSupport` is disabled
+            // treat all events as timelineEvents
+            return [
+                events,
+                [],
+            ];
+        }
+    }
+
+    /**
+     * @experimental
+     */
+    public processThreadEvents(room: Room, threadedEvents: MatrixEvent[]): void {
+        threadedEvents
+            .sort((a, b) => a.getTs() - b.getTs())
+            .forEach(event => {
+                room.addThreadedEvent(event);
+            });
     }
 }
 
--- a/src/client.ts
+++ b/src/client.ts
@@ -758,6 +758,7 @@ export class MatrixClient extends EventEmitter {
     protected exportedOlmDeviceToImport: IOlmDevice;
     protected txnCtr = 0;
     protected mediaHandler = new MediaHandler(this);
+    protected sessionId: string;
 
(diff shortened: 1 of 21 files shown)


src/client.ts
```diff
@@ -1258,15 +1261,68 @@ export class MatrixClient extends EventEmitter {
         this.supportsCallTransfer = support;
     }
 
+    public async getLocalVideoStream() {
```
robertlong, line 1264: This is an artifact of the [original group call implementation](https://github.com/vector-im/matrix-video-chat/blob/4d7e5583eb055d0fa2940ac0201f344e70cab588/src/Room.jsx#L232) in matrix-video-chat. I used this method to preview the user's local video stream. I think with the addition of `groupCall.initLocalParticipant()` these public methods are no longer necessary. However, we should still keep track of the `localAVStream`. We want to reuse the `localAVStream` and call `localAVStream.clone()` to get the streams to use in additional calls. In Safari, without calling `localAVStream.clone()` you'll get a black video.

src/client.ts
```diff
@@ -6107,11 +6163,11 @@ export class MatrixClient extends EventEmitter {
     public register(
         username: string,
         password: string,
-        sessionId: string,
+        sessionId: string | null,
```
robertlong, line 6166: Note for myself, these type changes should be split out into another PR. I need them for matrix-video-call.

src/matrix.ts
```diff
@@ -51,8 +51,14 @@ export {
     createNewMatrixCall,
     setAudioInput as setMatrixCallAudioInput,
     setVideoInput as setMatrixCallVideoInput,
+    CallType,
 } from "./webrtc/call";
 
+// TODO: This export is temporary and is only used for the local call feed for conference calls
+// Ideally conference calls will become a first-class concept and we will have a local call feed with
+// a lifecycle that matches the conference call, not individual calls to members.
+export { CallFeed } from "./webrtc/callFeed";
```
robertlong, line 60: Note for myself. Now that GroupCall has it's own `localParticipant` `CallFeed` we don't need to export this.

src/webrtc/call.ts
```diff
@@ -697,7 +712,14 @@ export class MatrixCall extends EventEmitter {
             this.waitForLocalAVStream = true;
 
             try {
-                const mediaStream = await navigator.mediaDevices.getUserMedia(constraints);
+                let mediaStream: MediaStream;
+
+                if (this.type === CallType.Voice) {
+                    mediaStream = await this.client.getLocalAudioStream();
```
robertlong, line 718: If we remove the public methods for getting local audio stream, we need to add some private method/state in its place.

robertlong: Maybe `localAVStream` / `localAVStreamPromise` could just be moved to a variable within this module? We only ever want to reference it from the call class. Also then we don't need to export `getUserMediaContraints`

src/webrtc/callFeed.ts
```diff
@@ -188,4 +189,8 @@ export class CallFeed extends EventEmitter {
             this.volumeLooper();
         }, POLLING_INTERVAL);
     }
+
+    public dispose(): void {
+        clearTimeout(this.volumeLooperTimeout);
```
robertlong, line 194: This dispose method just makes sure we don't keep the timeout alive after we've removed the call feed. I can split this out into another PR if we'd like.

src/webrtc/groupCall.ts
```diff
@@ -0,0 +1,762 @@
...
+export enum GroupCallEvent {
+    Entered = "entered",
+    Left = "left",
+    ActiveSpeakerChanged = "active_speaker_changed",
+    ParticipantsChanged = "participants_changed",
+    LocalMuteStateChanged = "local_mute_state_changed",
+}
+
+const CONF_ROOM = "me.robertlong.conf";
+const CONF_PARTICIPANT = "me.robertlong.conf.participant";
```
robertlong, line 22: TODO: I need to move this to the proper MSC event once we've created it.

src/webrtc/groupCall.ts
```diff
@@ -0,0 +1,762 @@
...
+            ACTIVE_SPEAKER_INTERVAL,
+        );
+    };
+
+    /**
+     * Utils
+     */
+
+    // TODO: move this elsewhere or get rid of the retry logic. Do we need it?
+    sendStateEventWithRetry(
```
robertlong, line 732: I don't actually know that we still want this for this PR. I don't think it was actually making calls any more stable, but I don't know for sure.

src/webrtc/call.ts
```diff
@@ -366,9 +370,15 @@ export class MatrixCall extends EventEmitter {
     public async placeVideoCall(): Promise<void> {
         logger.debug("placeVideoCall");
         this.checkForErrorListener();
-        const constraints = getUserMediaContraints(ConstraintsType.Video);
         this.type = CallType.Video;
-        await this.placeCallWithConstraints(constraints);
+        await this.placeCallWithConstraints(ConstraintsType.Video);
+    }
+
+    public createDataChannel(label: string, options: RTCDataChannelInit) {
```
robertlong, line 377: This method and corresponding event could be split into another PR if desired. I would like to reserve the last two parameters to `createGroupCall` for enabling datachannels for all calls and datachannel options though.

src/webrtc/call.ts
robertlong: Spent some time trying to figure out a better way to do this and I couldn't come up with a great solution. We actually want to call `getLocalAudioStream` from both `call.ts` and `groupCall.ts` so it can't be local to `call.ts`. I guess I need to know if it's alright to leave these as public methods or not.

src/client.ts
robertlong: Split out into https://github.com/matrix-org/matrix-js-sdk/pull/1903

src/webrtc/callFeed.ts
robertlong: Moved to https://github.com/matrix-org/matrix-js-sdk/pull/1904

src/client.ts
SimonBrandner: I think it would make sense to move this to the [`MediaHandler`](https://github.com/SimonBrandner/matrix-js-sdk/commit/a1f73eee86aff8b5b3c667573b44942c3c4e23f9) which will be introduced by https://github.com/matrix-org/matrix-js-sdk/pull/1911

src/webrtc/call.ts
SimonBrandner: I would argue for splitting this out as this PR is already large enough and the data channel stuff feels like a separate thing

src/webrtc/call.ts
SimonBrandner: As mentioned above the [`MediaHandler`](https://github.com/SimonBrandner/matrix-js-sdk/commit/a1f73eee86aff8b5b3c667573b44942c3c4e23f9) from https://github.com/matrix-org/matrix-js-sdk/pull/1911 feels like the way to go to me

src/webrtc/groupCall.ts
```diff
@@ -0,0 +1,565 @@
...
+                sessionId,
+                call,
+            );
+            // TODO: Should we wait until the call has been answered to push the participant?
+            // Or do we hide the participant until their stream is live?
+            // Does hiding a participant without a stream present a privacy problem because
+            // a participant without a stream can still listen in on other user's streams?
+            this.participants.push(participant);
+            this.reEmitter.reEmit(participant, Object.values(GroupCallParticipantEvent));
+            this.emit(GroupCallEvent.ParticipantsChanged), this.participants;
```
SimonBrandner, line 495: This looks weird

SimonBrandner (commented):
Generally, this looking absolutely awesome! I'll do a more thorough review when this isn't a draft but I've got a few more general points:

+ [ ] It might be beneficial for `localParticipant` to be a getter the same way `localUserMediaStream` and similar are
+ [x] `GroupCall` should emit `feeds_changed` events and have a `getFeeds()` method which is more friendly for making `CallView` components. It might even allow us to make the already existing `CallView` in the `react-sdk` work without many changes
+ [ ] The TypeScript could be improved a bit (params and return types)
+ [x] [`MediaHandler`](https://github.com/SimonBrandner/matrix-js-sdk/commit/a1f73eee86aff8b5b3c667573b44942c3c4e23f9) from https://github.com/matrix-org/matrix-js-sdk/pull/1911 should be somehow used when it gets merged

src/webrtc/call.ts
robertlong: Started here https://github.com/matrix-org/matrix-js-sdk/pull/1929

src/webrtc/call.ts
robertlong: I moved everything to use #1911 and #1928

src/webrtc/groupCall.ts
robertlong: You're talking about the comment right? I could use some feedback on that.

src/webrtc/groupCall.ts
```diff
@@ -0,0 +1,565 @@
...
+    Left = "left",
+    ActiveSpeakerChanged = "active_speaker_changed",
+    ParticipantsChanged = "participants_changed",
+    LocalMuteStateChanged = "local_mute_state_changed",
+}
+
+const CONF_ROOM = "me.robertlong.conf";
+const CONF_PARTICIPANT = "me.robertlong.conf.participant";
+
+export class GroupCall extends EventEmitter {
```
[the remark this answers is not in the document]
robertlong: Yeah, I'll document all the methods before this goes out. Added to the PR's todo list.

src/webrtc/groupCall.ts
SimonBrandner: I meant it probably should be like this, right? 

```suggestion
            this.emit(GroupCallEvent.ParticipantsChanged, this.participants);
```

Or is this some syntax, I am not aware of :sweat_smile:

src/webrtc/groupCall.ts
```diff
@@ -0,0 +1,559 @@
...
+            participant = new GroupCallParticipant(
+                this,
+                member,
+                sessionId,
+                call,
+            );
+            // TODO: Should we wait until the call has been answered to push the participant?
+            // Or do we hide the participant until their stream is live?
+            // Does hiding a participant without a stream present a privacy problem because
+            // a participant without a stream can still listen in on other user's streams?
```
SimonBrandner, line 486: If we end up doing this using `m.call.join` events in the way, I talked about, I think, we can push to participants when we start the call, without waiting for the answer, though I don't think we should show any feeds for them at that point. 

I've been actually thinking that it could be interesting trying to push new feeds once we receive `SDPStreamMetadata` and give `CallFeed` a loading state but I guess that is a topic for another time

SimonBrandner: We could even try to send `m.call.sdp_stream_metadata_changed` before the actual negotiation to get the info about new loading streams to the other side more quickly though I am not sure how much real benefit would that have. 

I am going to try to do some experiments with this

src/webrtc/groupCall.ts
robertlong: Ah, good catch haha :sweat_smile:

src/webrtc/groupCall.ts
```diff
@@ -0,0 +1,559 @@
...
+
+    private speakerMap: Map<RoomMember, number[]> = new Map();
+    private presenceLoopTimeout?: number;
+    private activeSpeakerLoopTimeout: number;
+    private reEmitter: ReEmitter;
+
+    constructor(
+        private client: MatrixClient,
+        roomId: string,
+        public type: CallType,
```
SimonBrandner, line 44: We are slowly killing of `CallType`s in `MatrixCall` and I think we should do the same here. This feels even more important for conferences - users should be allowed to join in whatever way they want (no camera, no mic, no nothing, whatever :D)

SimonBrandner: We've decided to keep this as the initial state in the MSC but I think the js-doc should reflect this is just the initial type

src/webrtc/callEventHandler.ts
```diff
@@ -273,7 +273,6 @@ export class CallEventHandler {
                         " and canceling outgoing call " + existingCall.callId,
                     );
                     existingCall.replacedBy(call);
-                    call.answer();
```
SimonBrandner, line 215: Won't this break regular VoIP calls?

src/webrtc/call.ts
```diff
@@ -1997,7 +1997,10 @@ export class MatrixCall extends EventEmitter {
         if (this.opponentDeviceId) {
             return this.client.sendToDevice(eventType, {
                 [this.invitee || this.getOpponentMember().userId]: {
-                    [this.opponentDeviceId]: realContent,
+                    [this.opponentDeviceId]: {
+                        ...realContent,
+                        device_id: this.client.deviceId,
```
SimonBrandner, line 2002: I am wondering if the js-sdk wouldn't allow you to get the deviceId from the `MatrixEvent`, so that you don't have to send it

no-1ne:
Looking forward

Will there be limitations on number of members in the call before quality degrades from the perspective of someone sharing a screen to 10-20 participants...

can getting the participants to act as relay to 1-2 other participants help.

SimonBrandner:
> Will there be limitations on number of participants in the call before quality degrades especially the scenario of someone sharing a screen to 10-20..50...100 participants...

At first, this will work as a full-mesh - all participants call all other participants, which will limit the number of users to ~8, afaik, before the quality starts to get worse, of course, this will depend on your computer and connection. Then the work on an SFU is planned which should solve this and possibly allow hundreds of users in the future if done well. See https://github.com/matrix-org/matrix-doc/pull/3401

no-1ne:
SFU makes sense scaling wise, thought matrix want to avoid anything centralised

t3chguy:
Each call/room/server could define its own SFU to use

SimonBrandner:
> Each call/room/server could define its own SFU to use

Also, there can be multiple SFUs involved in a call

no-1ne:
Aiming for 64-128 participants, Can participants  act as a relay...each relay  node forming a 2-4 full mesh. There by presenter to 4 nodes, those 4 nodes to 4 other nodes. Next hop reaching 64 participants. 

The reason being, Not many has access to a publicly accessible droplet to create a Mediasoup instance

SimonBrandner:
> Aiming for 64-128 participants, Can participants act as a relay...each relay node forming a 2-4 full mesh. There by presenter to 4 nodes, those 4 nodes to 4 other nodes. Next hop reaching 64 participants.
> 
> The reason being, Not many has access to a publicly accessible droplet to create a Mediasoup instance

This isn't implemented atm but might be in the future

src/webrtc/call.ts
```diff
@@ -630,6 +630,11 @@ export class MatrixCall extends EventEmitter {
      * @param {boolean} addToPeerConnection whether to add the tracks to the peer connection
      */
     public pushLocalFeed(callFeed: CallFeed, addToPeerConnection = true): void {
+        if (this.feeds.some((feed) => callFeed.stream.id === feed.stream.id)) {
```
SimonBrandner, line 633: Does the same stream id always mean the same feed?

src/webrtc/call.ts
bwindels: Indeed, you should be able to find out through the sender key of the olm decryption (unsure if/how the js-sdk exposes that though), which has the added benefit that it can't be spoofed. See https://github.com/matrix-org/matrix-doc/pull/3401#discussion_r805893936

dbkr:
Closing in favour of https://github.com/matrix-org/matrix-js-sdk/pull/2355

PR #1903, opened 2021-09-10
Add types for the MatrixClient register method

robertlong:
Added types for the `MatrixClient` `register()` method. Typescript projects importing matrix-js-sdk that don't set all of the fields don't pass type checking.

If there's a better place for the types to be converted to type or interface definitions, let me know. I could just put them at the top of `client.ts` if that works?

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -6109,7 +6109,7 @@ export class MatrixClient extends EventEmitter {
         password: string,
         sessionId: string | null,
         auth: { session?: string, type: string },
-        bindThreepids?: boolean | null | { email?: boolean | boolean, msisdn?: boolean },
+        bindThreepids?: boolean | null | { email?: boolean, msisdn?: boolean },
         guestAccessToken?: string,
         inhibitLogin?: boolean,
         callback?: Callback,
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -6107,17 +6107,17 @@ export class MatrixClient extends EventEmitter {
     public register(
         username: string,
         password: string,
-        sessionId: string,
-        auth: any,
-        bindThreepids: any,
-        guestAccessToken: string,
-        inhibitLogin: boolean,
+        sessionId: string | null,
```
turt2live, line 6110: we don't generally describe `null`, but this is fine

src/client.ts
```diff
@@ -6107,17 +6107,17 @@ export class MatrixClient extends EventEmitter {
...
+        auth: { session?: string, type: string },
+        bindThreepids?: boolean | null | { email?: boolean | boolean, msisdn?: boolean },
```
turt2live, line 6112:
```suggestion
        bindThreepids?: boolean | null | { email?: boolean, msisdn?: boolean },
```

turt2live (approved):
Otherwise looks good - thanks!

src/client.ts
Palid: @turt2live It makes sense to add nulls when possible, as this will make it easier for us to turn on strictNullChecks in the future

PR #1904, opened 2021-09-10
Properly dispose of CallFeeds

robertlong:
A `setTimeout` loop was recently added to the `CallFeed` class to poll volume levels. When we remove call feeds we never clean it up. This PR properly disposes of call feeds by immediately calling `clearTimeout` when they are removed.

---


turt2live (approved):

SimonBrandner (approved):
Thanks

PR #1911, opened 2021-09-11
Improve `MatrixCall` media handling and internally remove `CallType`s

SimonBrandner:
Type: task
Split out of https://github.com/matrix-org/matrix-js-sdk/pull/1827
**Requires https://github.com/matrix-org/matrix-react-sdk/pull/6781**

<hr>

Notes: This PR removes `setMatrixCallAudioInput()` and `setMatrixCallVideoInput()` in favour of `MediaHandler::setAudioInput()` and `MediaHandler::setVideoInput()`. The `MediaHandler` can be accessed through the `MatrixClient` using `MatrixClient::getMediaHandler()`

---

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -212,11 +212,6 @@ export enum CallErrorCode {
     Transfered = 'transferred',
 }
 
-enum ConstraintsType {
-    Audio = "audio",
-    Video = "video",
-}
-
 /**
  * The version field that we set in m.call.* events
  */
@@ -337,11 +332,7 @@ export class MatrixCall extends EventEmitter {
      * @throws If you have not specified a listener for 'error' events.
      */
     public async placeVoiceCall(): Promise<void> {
-        logger.debug("placeVoiceCall");
-        this.checkForErrorListener();
-        const constraints = getUserMediaContraints(ConstraintsType.Audio);
-        this.type = CallType.Voice;
-        await this.placeCallWithConstraints(constraints);
+        await this.placeCall(true, false);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


robertlong:
This looks really good @SimonBrandner! I'll let someone else review the breaking change, but it looks good to me.

One question, if we wanted to expose changing audio/video devices mid-call, I assume we'd want to expose that here. It'd also need a reference to the active calls to swap out the media streams and renegotiate. That could be a good future PR building on this work.

src/webrtc/call.ts
```diff
@@ -1809,8 +1809,17 @@ export class MatrixCall extends EventEmitter {
...
-        logger.log("Getting user media with constraints", constraints);
+    /**
+     * Place a call to this room.
+     * @throws if you have not specified a listener for 'error' events.
+     * @throws if have passed audio=false.
+     */
+    public async placeCall(audio: boolean, video: boolean): Promise<void> {
+        logger.debug(`placeCall audio=${audio} video=${video}`);
+        if (!audio) {
+            throw new Error("You CANNOT start a call without audio");
```
turt2live, line 1820: hmm.. why would it be a boolean option then?

turt2live (approved):
Overall this looks fine, thank you! The breaking change feels like a good quality of life improvement and version numbers are cheap, so I'm okay with this going through.

src/webrtc/call.ts
SimonBrandner: Because this should be possible in the future and I'd like to avoid another breaking change. In general, I feel like this will make things smoother

PR #1916, opened 2021-09-14
Tests for key sharing security fix

dbkr:
Tests for 894c24880da0e1cc81818f51c0db80e3c9fb2be9

---
This change is marked as an *internal change* (Task), so will not be included in the changelog.

PR #1917, opened 2021-09-14
Tests for key sharing security fix

dbkr:
https://github.com/matrix-org/matrix-js-sdk/pull/1916 to release branch

---


dbkr:
Merging because changes already approved elsewhere

PR #1928, opened 2021-09-15
Clone and stop local media streams from mediaHandler

robertlong:
This PR is an addition to #1911 that clones the local usermedia stream or screensharing stream instead of calling getUserMedia again when there is already an active stream. This addresses a bug on iOS where multiple calls to `getUserMedia()` will disable existing media streams. You need to call `mediaStream.clone()` instead. This is important for group calls (#1902) where we need to initiate `RTCPeerConnection`s to multiple group call participants. This PR also keeps track of the active streams so that they can be explicitly stopped when leaving a call. For group calls this means that we can call `stopAllStreams()` and ensure that all local media streams are stopped. This will also ensure that the webcam/microphone indicators will be turned off.

---

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -1747,18 +1747,14 @@ export class MatrixCall extends EventEmitter {
         logger.debug(`stopAllMedia (stream=${this.localUsermediaStream})`);
 
         for (const feed of this.feeds) {
-            if (feed.isLocal()) {
-                if (feed.purpose === SDPStreamMetadataPurpose.Usermedia) {
-                    this.client.getMediaHandler().stopUserMediaStream(feed.stream);
-                    continue;
-                } else if (feed.purpose === SDPStreamMetadataPurpose.Screenshare) {
-                    this.client.getMediaHandler().stopScreensharingStream(feed.stream);
-                    continue;
+            if (feed.isLocal() && feed.purpose === SDPStreamMetadataPurpose.Usermedia) {
+                this.client.getMediaHandler().stopUserMediaStream(feed.stream);
+            } else if (feed.isLocal() && feed.purpose === SDPStreamMetadataPurpose.Screenshare) {
+                this.client.getMediaHandler().stopScreensharingStream(feed.stream);
+            } else {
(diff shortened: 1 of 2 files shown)


src/webrtc/call.ts
```diff
@@ -1751,6 +1747,16 @@ export class MatrixCall extends EventEmitter {
         logger.debug(`stopAllMedia (stream=${this.localUsermediaStream})`);
 
         for (const feed of this.feeds) {
```
SimonBrandner, line 1749: I think we could use `this.getLocalFeeds()` to make this a bit simpler

src/webrtc/call.ts
```diff
@@ -1751,6 +1747,16 @@ export class MatrixCall extends EventEmitter {
...
+            if (feed.isLocal()) {
+                if (feed.purpose === SDPStreamMetadataPurpose.Usermedia) {
+                    this.client.getMediaHandler().stopUserMediaStream(feed.stream);
+                    continue;
```
SimonBrandner, line 1753: The `continue` doesn't seem to be necessary here?

SimonBrandner (approved):
LGTM otherwise, thanks!

src/webrtc/call.ts
robertlong: Do we need to cover unknown `feed.purpose` values? That's the only reason I did it. It avoids calling `stop()` on the tracks multiple times.

robertlong: I think the new version is a bit better

SimonBrandner: Oops, I missed a part of the code, the current solution seems great 👍

SimonBrandner (approved):

turt2live (approved):
Looks good - thank you!

For code review, we generally avoid requesting specific member review as anyone on the team should be able to satisfy the review requirement. Though it's not uncommon for the same person to review certain areas of the app :)

robertlong:
Got it 👍 yeah I'm used to assigning a reviewer and often picking people who have recently seen that code. I'll avoid it in the future. Thanks!

PR #1930, opened 2021-09-16
Typescriptify thread events a bit

Palid:
Move stringly typed events to enum, so it's harder to make a development mistake

---

### DIFF
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -20,9 +20,9 @@ import { MatrixEvent } from "./event";
 import { EventTimelineSet } from './event-timeline-set';
 import { Room } from './room';
 
-export enum THREAD_EVENTS {
-    ready = "Thread.ready",
-    update = "Thread.update"
+export enum ThreadEvents {
+    Ready = "Thread.ready",
+    Update = "Thread.update"
 }
 
 /**
@@ -78,7 +78,7 @@ export class Thread extends EventEmitter {
         if (this.ready) {
             this.client.decryptEventIfNeeded(event, {});
         }
-        this.emit(THREAD_EVENTS.ready, this);
+        this.emit(ThreadEvents.Ready, this);
     }
 
     /**
@@ -101,7 +101,7 @@ export class Thread extends EventEmitter {
                 await this.fetchReplyChain();
             } else {
                 await this.decryptEvents();
-                this.emit(THREAD_EVENTS.ready, this);
+                this.emit(ThreadEvents.Ready, this);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/models/thread.ts
```diff
@@ -20,6 +20,11 @@ import { MatrixEvent } from "./event";
 import { EventTimelineSet } from './event-timeline-set';
 import { Room } from './room';
 
+export enum THREAD_EVENTS {
```
t3chguy, line 23: we normally use Camel Casing for our enum names in the singular pluralisation

src/models/thread.ts
```diff
@@ -20,6 +20,11 @@ import { MatrixEvent } from "./event";
...
+    ready = "Thread.ready",
+    update = "Thread.update"
```
t3chguy, line 25: We normally use Camel Casing for our enum member naming

Palid: Done.

src/models/thread.ts
Palid: Done.

src/models/thread.ts
```diff
@@ -20,6 +20,11 @@ import { MatrixEvent } from "./event";
 import { EventTimelineSet } from './event-timeline-set';
 import { Room } from './room';
 
+export enum ThreadEvents {
```
t3chguy, line 23: normally we'd use `ThreadEvent.Ready` etc, rather than pluralising

src/models/thread.ts
```diff
@@ -73,7 +78,7 @@ export class Thread extends EventEmitter {
         if (this.ready) {
             this.client.decryptEventIfNeeded(event, {});
         }
-        this.emit("Thread.update", this);
+        this.emit(ThreadEvents.Ready, this);
```
t3chguy, line 81: this seems to have gone from update to ready, is this an intentional change? definitely doesn't look like straight TSification

src/models/thread.ts
```diff
@@ -193,4 +198,25 @@ export class Thread extends EventEmitter {
...
+        return this;
+    }
+    public addListener(event: ThreadEvents, listener: (...args: any[]) => void): this {
+        super.addListener(event, listener);
+        return this;
+    }
+    public removeListener(event: ThreadEvents, listener: (...args: any[]) => void): this {
+        super.removeListener(event, listener);
+        return this;
+    }
```
t3chguy, line 221: Wonder if we can do this with a generic base class or something, as this is kinda boilerplatey, e.g `EventEmitter<ThreadEvents>` etc

t3chguy (approved):
LGTM otherwise

src/models/thread.ts
Palid: I think we can definitely do this. Is it useful to do it within this PR?

src/models/thread.ts
Palid: That's a mistake, thanks for catching! Probably went a bit too far with replacements.

src/models/thread.ts
t3chguy: Up to you, I already approved

PR #1934, opened 2021-09-17
Switch to /kick API

aaronraimist:
Fixes https://github.com/vector-im/element-web/issues/19018

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -4288,7 +4288,16 @@ export class MatrixClient extends EventEmitter {
      * @return {module:http-api.MatrixError} Rejects: with an error response.
      */
     public kick(roomId: string, userId: string, reason?: string, callback?: Callback): Promise<{}> {
-        return this.setMembershipState(roomId, userId, "leave", reason, callback);
+        const path = utils.encodeUri("/rooms/$roomId/kick", {
+            $roomId: roomId,
+        });
+        const data = {
+            user_id: userId,
+            reason: reason,
+        };
+        return this.http.authedRequest(
+            callback, "POST", path, undefined, data,
+        );
     }
 
     /**


turt2live (approved):
This sort of issue is a bit of a catch-22: landing this ahead of the Synapse change means Element is at risk of reopening the bug regarding unintentional unbans, but the same happens if Synapse lands their changes first.

I think it's best if we figure out coordination at a higher level than code review in the independent projects (cc @anoadragon453 ) - will raise internally.

aaronraimist:
Actually I think it should be fine to land this first. `/kick`ing someone who is banned doesn't seem to do anything right now.

novocaine:
Finding this hard to evaluate this because the description of the user impact isn't all in one place - there's a rolling discussion across many tickets, and the tickets are spec-centric, not user centric. Can we summarise the user impact of this change here please? 

There are other home servers and other clients to consider, as well:

* What's the user impact of releasing the Synapse change on other clients in the ecosystem, prior to them updating to the latest `matrix-js-sdk`?
* What's the user impact of updating Element if the home server administrator does not update Synapse in a timely fashion, or if another home server is used such as conduit or dendrite?

aaronraimist:
There should be no impact on users. This is an internal change that just switches from manually sending `m.room.member` events to using the proper Matrix API for kicking. The spec says

> It is preferable to use the membership APIs (/rooms/<room id>/invite etc) when performing membership actions rather than adjusting the state directly

https://spec.matrix.org/unstable/client-server-api/#room-membership

There is no need to update Synapse for this change to work. This API has been around in the spec like 2016 https://matrix.org/docs/spec/client_server/r0.1.0#post-matrix-client-r0-rooms-roomid-kick. Any server not implementing it would be at fault, not Element. I have not tested against Dendrite or Conduit but Element already uses the proper APIs for /ban /unban /leave and /invite so I'm sure those servers implement /kick as well.

Palid:
@novocaine @aaronraimist would it make sense to have a fallback for the old way in case the new and proper way fails? I don't think it'll hurt, but this way we'll have proper compatibility even if something implements it incorrectly.

novocaine:
> I have not tested against Dendrite or Conduit but Element already uses the proper APIs for /ban /unban /leave and /invite so I'm sure those servers implement /kick as well.

I think we should test this or at least check if their code looks like they implement /kick - it would be quite unfortunate if EW inadvertently lost the ability to kick people on certain home servers.

turt2live:
Conduit and Dendrite both have route handlers for `/kick`. 

Merging this as Synapse appears to have landed their side. We should watch out for issues.

PR #1938, opened 2021-09-18
Ensure unencrypted fields get exposed by getEffectiveEvent()

turt2live:
Fixes https://github.com/vector-im/element-web/issues/19062

---

## 🐛 Bug Fixes
 * Ensure unencrypted fields get exposed by getEffectiveEvent() ([\#1938](https://github.com/matrix-org/matrix-js-sdk/pull/1938)). Fixes vector-im/element-web#19062.

### DIFF
--- a/src/models/event.ts
+++ b/src/models/event.ts
@@ -285,8 +285,28 @@ export class MatrixEvent extends EventEmitter {
      * @returns {IEvent} The event in wire format.
      */
     public getEffectiveEvent(): IEvent {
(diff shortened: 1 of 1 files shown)


gsouquet (approved):
Looks good to me!
Wondering whether there's any room to add a test case to reproduce this scenario as it looks quite tricky to spot, or even be aware of

turt2live:
I'd love for this code to be tested, but unfortunately it'd not be a super great test at the moment. Most likely when other work in the events area is done (another side project of mine) then it'll get covered with testing of that.

PR #1944, opened 2021-09-22
Share room state events with thread's timelines

gsouquet:
Fixes vector-im/element-web#18719

Events created in a thread do not receive the correct meta data, resulting in an incomplete sender profile as you can see on the screenshot below

134323765-b7ac6104-0d02-44d0-914e-9ea9db6eabda

The reason is that when an event is added to the timeline, the `event-timeline-set` looks for room state events. This works perfectly well for events in the room's timeline, however threaded events are part of their own timelines which contains no state events.

This pull request adds a way to pass the room's state events down so that the event metadata can be set the same way they would be in the room timeline.
The approach looks a bit odd but at this stage I'm not entirely sure about alternative approaches hence why this is a draft.

Keen to hear your ideas/alternatives to what I'm doing here

---

### DIFF
--- a/src/models/event-timeline-set.ts
+++ b/src/models/event-timeline-set.ts
@@ -27,6 +27,7 @@ import { Relations } from './relations';
 import { Room } from "./room";
 import { Filter } from "../filter";
 import { EventType, RelationType } from "../@types/event";
+import { RoomState } from "./room-state";
 
 // var DEBUG = false;
 const DEBUG = true;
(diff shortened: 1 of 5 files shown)


src/models/event-timeline-set.ts
```diff
@@ -509,8 +515,14 @@ export class EventTimelineSet extends EventEmitter {
      * @param {MatrixEvent} event Event to be added
      * @param {string?} duplicateStrategy 'ignore' or 'replace'
      * @param {boolean} fromCache whether the sync response came from cache
+     * @param {boolean} fromCache whether the sync response came from cache
```
turt2live, line 518: looks like this was meant to be for `roomState`?

turt2live (changes requested):
Looks sensible, thank you! Just the one comment about docs.

Was this meant to be a not-draft PR?

turt2live (approved):

PR #1950, opened 2021-09-25
Allow answering calls without audio/video

SimonBrandner:
Type: enhancement

---

## ✨ Features
 * Allow answering calls without audio/video ([\#1950](https://github.com/matrix-org/matrix-js-sdk/pull/1950)). Contributed by @SimonBrandner.

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -696,13 +696,34 @@ export class MatrixCall extends EventEmitter {
         this.setState(CallState.Ended);
     }
 
+    private shouldAnswerWithMediaType(
+        wantedValue: boolean, valueOfTheOtherSide: boolean, type: "audio" | "video",
+    ): boolean {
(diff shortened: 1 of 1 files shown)


src/webrtc/call.ts
```diff
@@ -696,13 +696,34 @@ export class MatrixCall extends EventEmitter {
         this.setState(CallState.Ended);
     }
 
+    private shouldAnswerWithMediaType(
+        wantedValue: boolean, valueOfTheOtherSide: boolean, type: "audio" | "video",
```
turt2live, line 700: Because these are allowed to be undefined, but we can't use optionals:
```suggestion
        wantedValue: boolean | undefined, valueOfTheOtherSide: boolean | undefined, type: "audio" | "video",
```

turt2live (commented):
Just a small note about types, otherwise looks good - thanks!

turt2live (approved):

PR #1957, opened 2021-09-28
Self-destructing messages (partial MSC2228)

alangecker:
Signed-off-by: chandi Langecker <git@chandi.it>

implements partially [MSC2228: Self destructing events](https://github.com/matrix-org/matrix-doc/pull/2228), only its `m.self_destruct_after` field because it is already supported by synapse since [Add ephemeral messages support (MSC2228)](https://github.com/matrix-org/synapse/pull/6409). There is currently no server side support of `m.self_destruct`.

## Open obscurities
1. Is creating redaction event which only exists locally the best way to do that?
1. ~~How to redact the event content from the stored sync?~~
1. How should an `event_id` be generated for an event which only exists locally?
1. Can we safely treat an event without any content as unencrypted?

## Reviewing hints
- it requires a synapse server with `enable_ephemeral_messages: yes`
- create an event with an `org.matrix.self_destruct_after` field in the content set to a future timestamp in msecs

### ToDo's
- [x] Basic implementation
- [x] Redaction from store
- [x] Tests
- [x] Changelog entry
- [ ] Review

Notes: Self destructing events (partial implementation of MSC2228)
Type: enhancement

---

## ✨ Features
 * Self destructing events (partial implementation of MSC2228) ([\#1957](https://github.com/matrix-org/matrix-js-sdk/pull/1957)). Contributed by @alangecker.

### DIFF
--- a/spec/unit/crypto/algorithms/megolm.spec.js
+++ b/spec/unit/crypto/algorithms/megolm.spec.js
@@ -68,6 +68,9 @@ describe("MegolmDecryption", function() {
             // 'crypto' implementation.
             const event = new MatrixEvent({
                 type: 'm.room.encrypted',
+                content: {
+                    algorithm: 'm.megolm.v1.aes-sha2',
+                },
             });
             const decryptedData = {
                 clearEvent: {
--- a/spec/unit/crypto/backup.spec.js
+++ b/spec/unit/crypto/backup.spec.js
@@ -224,6 +224,9 @@ describe("MegolmBackup", function() {
             // 'crypto' implementation.
             const event = new MatrixEvent({
                 type: 'm.room.encrypted',
+                content: {
+                    algorithm: 'm.megolm.v1.aes-sha2',
+                },
             });
             const decryptedData = {
                 clearEvent: {
--- a/spec/unit/room.spec.ts
+++ b/spec/unit/room.spec.ts
@@ -42,6 +42,8 @@ import { emitPromise } from "../test-utils/test-utils";
 import { ReceiptType } from "../../src/@types/read_receipts";
 import { Thread, ThreadEvent } from "../../src/models/thread";
 
+jest.useFakeTimers();
+
 describe("Room", function() {
     const roomId = "!foo:bar";
     const userA = "@alice:bar";
@@ -2429,7 +2431,7 @@ describe("Room", function() {
                     content: {
                         msgtype: "m.text",
                         body: "TOPSECRET",
-                        [UNSTABLE_MSC2228_SELF_DESTRUCT_AFTER.name]: Date.now()+50,
+                        [UNSTABLE_MSC2228_SELF_DESTRUCT_AFTER.name]: Date.now()+10000,
                     },
                 }),
             ]);
@@ -2446,6 +2448,8 @@ describe("Room", function() {
                 expect(client.store.replaceEvent).toBeCalled();
                 done();
             });
+
+            jest.runAllTimers();
         });
     });
 });
--- a/spec/unit/room.spec.ts
+++ b/spec/unit/room.spec.ts
@@ -2446,7 +2446,7 @@ describe("Room", function() {
             expect(event.isRedacted()).toBeFalsy();
             expect(event.getContent().body).toBe("TOPSECRET");
 
-            // deleted soon
+            // deleted soon after the self destruct time elapses
             room.on(RoomEvent.Redaction, function(redactionEvent, room) {
                 expect(redactionEvent.getType()).toBe("m.room.redaction");
                 expect(event.getContent().body).toBeFalsy();
--- a/spec/unit/room.spec.ts
+++ b/spec/unit/room.spec.ts
@@ -26,6 +26,7 @@ import {
     EventTimelineSet,
     EventType,
     JoinRule,
+    MatrixClient,
     MatrixEvent,
     PendingEventOrdering,
     RelationType,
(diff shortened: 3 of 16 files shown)


gsouquet:
Thank you for creating your first pull request in `matrix-js-sdk`!

I have converted it to a draft and removed the `element-web` team as reviewer as it's still a work in progress. Ping me when you'd like someone to review your work

Feel free to drop by https://matrix.to/#/#element-dev:matrix.org for dev work coordination

alangecker:
I've managed to build now a way to redact the event from the IndexedDB store / SyncAccumulator.

@gsouquet Apart from the tests and a changelog entry, I'm not planning any more changes at the moment.
I would therefore appreciate a review whether the implementation makes sense at all, before I take the effort to build tests :)

alangecker:
@turt2live 
good point with the app reload, but no matter what I do, on a page reload there is always `Room.addLiveEvent` once called for every event, so the timers are always getting triggered.

Is this intended or is there also a case where the method does not get executed?

turt2live:
Timers being restarted from `Room.addLiveEvent` sounds like the magic I wasn't seeing - thanks. 

I think this PR just needs tests and such to fill out the checkboxes and then it is good to go?

alangecker:
@turt2live tests and changelog entry added! Hope it can be merged as it is :)

spec/unit/crypto/algorithms/megolm.spec.js
```diff
@@ -68,6 +68,9 @@ describe("MegolmDecryption", function() {
             // 'crypto' implementation.
```
MadLittleMods, line 68: This PR has fallen a bit stale and has some conflicts now. Can you merge in the latest `develop` here?

Conflicting files:

 - `spec/unit/room.spec.js`
 - `src/@types/event.ts`
 - `src/crypto/index.ts`
 - `src/models/room.ts`

spec/unit/room.spec.js
```diff
@@ -1757,4 +1760,70 @@ describe("Room", function() {
...
+            };
+            const room = new Room(roomId, client, userA);
+            room.addLiveEvents([
+                utils.mkEvent({
+                    type: "m.room.message",
+                    room: roomId, event: true,
+                    content: {
+                        msgtype: "m.text",
+                        body: "TOPSECRET",
+                        [UNSTABLE_MSC2228_SELF_DESTRUCT_AFTER.name]: Date.now()+50,
```
MadLittleMods, line 1810: I'm wondering if this `50` magic value would be flakey at all if it takes more than 50ms to add the event to the live timeline. It's probably rare though if this is sufficiently big enough.

We can control the `Date` in Jest if we want to be absolute about this.

Would be good to at least comment about the `50` magic value here. Something like, 

> // We chose `50` because it gives a small buffer for us to check that the message is not deleted and then deleted a few moments later.
> // FIXME: This is potentially flakey if it takes more than 50ms for the event to be added to the live timeline

spec/unit/room.spec.js
```diff
@@ -1757,4 +1760,70 @@ describe("Room", function() {
...
+                    },
+                }),
+            ]);
+
+            // not deleted yet
+            const event = room.timeline[0];
+            expect(event.isRedacted()).toBeFalsy();
+            expect(event.getContent().body).toBe("TOPSECRET");
+
+            // deleted soon
```
MadLittleMods, line 1820:
```suggestion
            // deleted soon after the self destruct time elapses
```

spec/unit/crypto/algorithms/megolm.spec.js
alangecker: @MadLittleMods I did now a rebase :)

spec/unit/room.spec.js
alangecker: good point! I've switched now to `jest.useFakeTimers()` and `jest.runAllTimers()` which works nicely :)

alangecker:
@MadLittleMods many thanks for the review! :)

I've resolved now all the points and would be very happy to see this finally merged 🙂

alangecker:
the tests fail at the `yarn cache` step with:
`Error: getCacheEntry failed: Cache service responded with 503`

this seems unrelated to the PR, maybe someone could rerun the tests? :)

spec/unit/crypto/algorithms/megolm.spec.js
MadLittleMods, line 68: > this seems unrelated to the PR, maybe someone could rerun the tests? :)

Try merging the latest `develop` again

MadLittleMods, line 68: > I've resolved now all the points and would be very happy to see this finally merged 🙂

For reference, I'm not an Element-Web maintainer to merge. Just gave it a review glance 🙂

shoetten:
What's the status on this? It's a feature needed by activists and journalists worldwide and it would be a shame to waste an (almost) finished PR.

dbkr:
It's a shame this looks to have stalled. Effort for implementing the MSC is very appreciated. That said, the MSC hasn't really progressed either. It also looks quite conflicty. It's very useful for this to exist to be an example implementation for the MSC, but since it's not currently in a mergeable state, I'm going to hit the close button.

For anyone who'd like to pick this up & open a new one, some thing to bear in mind:
 * We probably would not merge this until the MSC is accepted, or at best, it would need to be behind a labs flag in Element, so at the js-sdk level that means it would need a way to enable/disable at the API level if it were to be merged before the MSC were accepted.
 * Digging into stores to remove the message is logical, except that we don't do this for regular redacted events which seems like a weird disparity, especially since this is based on redactions. It might make more sense to talk about whether the js-sdk should be doing this for all redactions first, then implement self-destructing message separately based on this.

PR #1959, opened 2021-09-29
Fix incorrect return value type in getJoinedRooms()

psrpinto:
[`_matrix/client/r0/joined_rooms`](https://matrix.org/docs/spec/client_server/latest#get-matrix-client-r0-joined-rooms) returns an object with a single `joined_rooms` property, which is an array of strings:

```json
{
  "joined_rooms": [
    "!foo:example.com"
  ]
}
```

However, the [`getJoinedRooms()` function](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/client.ts#L6677) returns a promised array of strings:

```
public getJoinedRooms(): Promise<string[]> {
```

This PR changes the return value to `Promise<IJoinedRoomsResponse>`, with `IJoinedRoomsResponse` being:

```typescript
interface IJoinedRoomsResponse {
    joined_rooms: string[];
}
```

---

## 🐛 Bug Fixes
 * Fix incorrect return value type in getJoinedRooms() ([\#1959](https://github.com/matrix-org/matrix-js-sdk/pull/1959)). Contributed by @psrpinto.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -568,6 +568,10 @@ interface IRoomInitialSyncResponse {
     presence: Partial<IEvent>; // legacy and undocumented, api is deprecated so this won't get attention
 }
 
+interface IJoinedRoomsResponse {
+    joined_rooms: string[];
+}
+
 interface IJoinedMembersResponse {
     joined: {
         [userId: string]: {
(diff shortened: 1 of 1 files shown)


psrpinto:
I'm wondering, is this a backwards-compatibility break?

gsouquet (approved):
That looks good to me!
Regarding the backwards compatibility I'm inclined to say that we don't want to support erroneous return types for the sake of backwards compat, but it might be best to gather a bit more opinions on that in the element-dev room.

We could always make the return type
```
Promise<IJoinedRoomsResponse | string[]
```
And add comments explaining the issue.

psrpinto:
[Asked on #element-dev](https://matrix.to/#/!bEWtlqtDwCLFIAKAcv:matrix.org/$kMs_aKMETexK7XaRm1CpdBoYQLgLScHgvO89s8uCFjg?via=matrix.org&via=envs.net&via=supercable.onl), consensus appears to be changing types isn't a BC-break.

Screenshot 2021-09-30 at 17 50 48

So I guess this can be merged? 😇

PR #1963, opened 2021-10-01
Prepend events to thread when fetching the reply chain

gsouquet:
Fixes vector-im/element-web#19225

---


Palid:
@gsouquet linter is unhappy. LGTM otherwise

Palid (approved):
Make linter happy please, LGTM otherwise

PR #1973, opened 2021-10-09
Fix re-emitter memory leak

Palid:
Seems that in `ReEmitter.ts` we have been listening to things, but never un-listening after re-emitting events, which made circular events listeners in threads. 
It doesn't seem like an expected behavior.

@dbkr can you double-check that we should've un-listened to the events after emitting them again, or was this behavior correct and threads are using it wrong?

---


gsouquet:
As mentioned by Michael and Dave, the re-emitter is expected to emit events multiple times and not only once.
Threads do not suffer from circular updates anymore.

Do you still believe like we need to change something in relation to this or are we good to close this pull request?

t3chguy:
Seems no longer needed.

PR #2012, opened 2021-11-05
Copy relations to thread root in the thread timeline

gsouquet:
Fixes vector-im/element-web#19638

---


jryans (approved):
Seems reasonable, thanks! 😄 l

PR #2014, opened 2021-11-08
Standardise content type handling in MSC3089 createFile() and createNewVersion()

hughns:
The existing implementations are replaced with versions that handle the same content types as MatrixClient.uploadContent(). This breaks uses of `ArrayBuffer` but enables support for Node.js environment.

Notes: Remove support for `ArrayBuffer` in unstable MSC3089 `createFile()` and `createNewVersion()` and instead use same content types as handled by `MatrixClient.uploadContent()`. This enables support for Node.js.

---

## ✨ Features
 * Remove support for `ArrayBuffer` in unstable MSC3089 `createFile()` and `createNewVersion()` and instead use same content types as handled by `MatrixClient.uploadContent()`. This enables support for Node.js. ([\#2014](https://github.com/matrix-org/matrix-js-sdk/pull/2014)). Contributed by @hughns.


turt2live (commented):
I think this is fine overall. Because the entire API surface here is unstable, we can break it without deprecating the old versions. I'd recommend we do that and let the downstream (hidden) dependents know that things are slightly different now.

hughns:
@turt2live I've made this a breaking change in the unstable API as you suggest. How should this change now be labelled in the changelog?

spec/unit/models/MSC3089TreeSpace.spec.ts
```diff
@@ -997,9 +988,10 @@ describe("MSC3089TreeSpace", () => {
             });
         client.sendStateEvent = sendStateFn;
 
-        const buf = Uint8Array.from(Array.from(fileContents).map((_, i) => fileContents.charCodeAt(i)));
+        const buf = Buffer.from(fileContents);
 
         // We clone the file info just to make sure it doesn't get mutated for the test.
+        // TODO: use non-deprecated function
```
turt2live, line 994: I believe this no longer applies?

turt2live (approved):
For the changelog, I think it's probably fine as-is.

PR #2019, opened 2021-11-12
Add LocalStorageErrorEventListener

Palid:
Add a way to listen to LocalStorage error events from element-web and matrix-react-sdk in the same manner.
Related: https://github.com/vector-im/element-web/issues/18423
Required by later implementations in `matrix-react-sdk` and `element-web`.

Requires https://github.com/matrix-org/matrix-js-sdk/pull/2018 to be merged in first.

---

### DIFF
--- a/src/store/local-storage-events-emitter.ts
+++ b/src/store/local-storage-events-emitter.ts
@@ -1,0 +1,37 @@
+/*
+Copyright 2021 The Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { BaseModel } from "../models/base-model";
+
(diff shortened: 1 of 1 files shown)


src/store/local-storage-events-emitter.ts
```diff
@@ -0,0 +1,29 @@
...
+export enum LocalStorageErrors {
+    Global = 'Global',
+    SetItemError = 'setItem',
+    GetItemError = 'getItem',
+    RemoveItemError = 'removeItem',
+    ClearError = 'clear',
+    QuotaExceededError = 'QuotaExceededError'
+}
+
+// Used in element-web as a temporary hack to handle all the localStorage errors on the highest level possible
```
gsouquet, line 28: It would be interesting to capture the goal we're aiming for and reference the issue here.
If I was to come across this comment, I would be unsure about why this is hacky and what we're trying to achieve

gsouquet (approved):
Looks good to me 👍 

I changed the base branch for an easier review. I believe GitHub should change it back to `develop` once the pull request gets merged

src/store/local-storage-events-emitter.ts
Palid: @gsouquet that's why we have git blame! 😄 
I'll add the issue here though, good idea.

Palid: See the comment now, should be much better.

gsouquet: The comment does not address why this is a temporary hack nor the ideal solution we're aiming for

The git blame did not mention any of that before you force pushed...

gsouquet (approved):

PR #2023, opened 2021-11-17
Exporting missing types

skyka13711:
These types are needed to describe function signatures that wrap matrix-sdk methods

---

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -32,8 +32,8 @@ try {
 export * from "./matrix";
 export default matrixcs;
 
-export { EventType } from './@types/event'
-export { PushRuleKind } from './@types/PushRules'
-export { ISendEventResponse } from './@types/requests'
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -31,3 +31,9 @@ try {
 
 export * from "./matrix";
 export default matrixcs;
+
+export { EventType } from './@types/event';
```
turt2live, line 35: we should probably be using `export *` for all of these.

turt2live (changes requested):
Thanks for the contribution! It would be great if we can get a more complete title for this PR so it shows up in the changelog correctly, and it looks like the `yarn.lock` was accidentally committed with no relevant `package.json` changes.

skyka13711:
fixed

turt2live (approved):
Thanks!

PR #2024, opened 2021-11-17
Fix call upgrades

SimonBrandner:
Type: defect

<hr>

Due to `CallFeed`s mute state being set to muted based on their tracks at the time of their creation we can get stuck with them being muted. This happens because the code that determines if a `CallFeed` is muted looks like this: 

```
return this.stream.getVideoTracks().length === 0 || this.videoMuted;
```

This means that if `videoMuted` is `true` the feed will be muted no matter the present tracks. The `videoMuted` property should only work as an override based on the remote `SPDStreamMetadata`

---

## 🐛 Bug Fixes
 * Fix call upgrades ([\#2024](https://github.com/matrix-org/matrix-js-sdk/pull/2024)). Contributed by @SimonBrandner.

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -583,8 +583,8 @@ export class MatrixCall extends EventEmitter {
                 new CallFeed({
                     client: this.client,
                     roomId: this.roomId,
-                    audioMuted: stream.getAudioTracks().length === 0,
-                    videoMuted: stream.getVideoTracks().length === 0,
+                    audioMuted: false,
+                    videoMuted: false,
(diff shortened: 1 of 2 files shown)


turt2live (changes requested):
The changes look good, though the PR description could do with a description of what is broken or an issue link which describes that.

SimonBrandner:
> The changes look good, though the PR description could do with a description of what is broken or an issue link which describes that.

Should have thought of that... I've updated the description

turt2live (approved):
Thanks!

PR #2025, opened 2021-11-18
Update sync state to error when aborting due to invalid auth token

robertlong:
Currently the sync state doesn't change when sync fails due to an invalid auth token. This PR ensures that the sync state changes to error when we abort the sync so we can ask the user to login again.

---


gsouquet (approved):

sanjarbarakayev:
How can I handle this error? I have the same problem. I can't catch errors on sync event when the token is not valid.

PR #2026, opened 2021-11-18
Make local echo work for threads 

gsouquet:
Related to https://github.com/vector-im/element-web/issues/19616

- Use PendingEventOrdering enum rather than hardcoded values
- Make local echo work for threads

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -3411,7 +3411,7 @@ export class MatrixClient extends EventEmitter {
      * @return {module:http-api.MatrixError} Rejects: with an error response.
      */
     private sendCompleteEvent(
-        roomId: string,
+        compoundId: string,
         eventObject: any,
         txnId?: string,
         callback?: Callback,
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -3398,20 +3399,45 @@ export class MatrixClient extends EventEmitter {
...
+    public sendEvent(
+        roomId: string,
+        threadId: string | null,
+        eventType: string | IContent,
+        content: IContent | string,
+        txnId?: string | Callback,
+        callback?: Callback,
     ): Promise<ISendEventResponse> {
-        return this.sendCompleteEvent(roomId, { type: eventType, content }, txnId, callback);
+        if (!threadId?.startsWith("$") && threadId !== null) {
```
turt2live, line 3419: I'm really not a fan of string matching these things - can we pass an object of some kind for `instanceof` checks instead? The types will be a lot easier to work with.

turt2live (changes requested):

src/client.ts
gsouquet: They could be, but the reason why I opted out from that was to keep it simple to send an event. 
Not requiring a developer to instanciate 5 different models before they can even send something

The way the app is architectured too, the composer is not really aware of the `Thread` model either, but is aware of a relation to attach to all event it emits out.
I decided to go with that solution to keep the door open to more changes in the future and not tie down the composer to the current model of threads

turt2live (approved):
Thanks! This is a lot more comfortable feeling :)

PR #2027, opened 2021-11-18
Add optional force parameter when ensuring Olm sessions

turt2live:
This is useful for debugging crypto where Element's `/discardsession` doesn't reset the Olm session.

---

## ✨ Features
 * Add optional force parameter when ensuring Olm sessions ([\#2027](https://github.com/matrix-org/matrix-js-sdk/pull/2027)).


andybalaam (approved):
Looks good, thanks!

PR #2032, opened 2021-11-22
getStateEvent should return any

Half-Shot:
getStateEvent returns the content of an event, not the whole state event as per https://matrix.org/docs/spec/client_server/r0.6.0#get-matrix-client-r0-rooms-roomid-state-eventtype-statekey

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -6613,7 +6613,7 @@ export class MatrixClient extends EventEmitter {
         eventType: string,
         stateKey: string,
         callback?: Callback,
-    ): Promise<any> {
+    ): Promise<Record<string, any>> {
         const pathParams = {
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -6613,7 +6613,7 @@ export class MatrixClient extends EventEmitter {
         eventType: string,
         stateKey: string,
         callback?: Callback,
-    ): Promise<IStateEventWithRoomId> {
+    ): Promise<any> {
```
t3chguy, line 6616:
```suggestion
    ): Promise<Record<string, any>> {
```

Half-Shot: Fair point!

t3chguy (approved):

PR #2034, opened 2021-11-23
Comprehensive events emitter typing

gsouquet:
Following up on https://github.com/matrix-org/matrix-js-sdk/pull/2018

It was missing a fair few methods as listed here https://nodejs.org/api/events.html
This PR updates all the methods that can benefit from the newly introduced typing.

Once thing that I have not been able to express is how to extends the `enum` passed down to make sure that it can include the two generic event `newListener` and `removeListener` that `EventEmitter` can emit.

---


andybalaam (approved):
Looks good, thanks!

Palid (approved):

PR #2037, opened 2021-11-23
Improve fallback key behaviour

uhoreg:
- tracks whether the fallback key has already been published
- clear out old fallback key some time after it's been replaced

~Note: do not merge until libolm 3.2.7 is released.  (Also, needs a change to `package.json` to use the newer libolm) (test failures are due to this)~ Done

---

## ✨ Features
 * Improve fallback key behaviour ([\#2037](https://github.com/matrix-org/matrix-js-sdk/pull/2037)).

### DIFF
--- a/src/crypto/index.ts
+++ b/src/crypto/index.ts
@@ -1957,12 +1957,11 @@ export class Crypto extends EventEmitter {
 
         const requestBody: Record<string, any> = {
             "one_time_keys": oneTimeJson,
-            "org.matrix.msc2732.fallback_keys": fallbackJson,
-            "fallback_keys": fallbackJson,
         };
 
         if (fallbackJson) {
             requestBody["org.matrix.msc2732.fallback_keys"] = fallbackJson;
+            requestBody["fallback_keys"] = fallbackJson;
         }
 
         const res = await this.baseApis.uploadKeysRequest(requestBody);
--- a/src/crypto/index.ts
+++ b/src/crypto/index.ts
@@ -1928,8 +1928,9 @@ export class Crypto extends EventEmitter {
     private async uploadOneTimeKeys() {
         const promises = [];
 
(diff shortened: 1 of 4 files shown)


src/crypto/index.ts
```diff
@@ -1864,8 +1865,23 @@ export class Crypto extends EventEmitter {
...
 
                 if (this.getNeedsNewFallback()) {
-                    logger.info("generating fallback key");
-                    await this.olmDevice.generateFallbackKey();
+                    const fallbackKeys = await this.olmDevice.getFallbackKey();
+                    // if fallbackKeys is non-empty, we've already generated a
+                    // fallback key, but it hasn't been published yet, so we
+                    // can use that instead of generating a new one
+                    if (!fallbackKeys.curve25519 ||
+                        [...Object.keys(fallbackKeys.curve25519)].length == 0) {
```
turt2live, line 1873:
```suggestion
                        Object.keys(fallbackKeys.curve25519).length == 0) {
```

turt2live (approved):
yay, improved fallback keys

src/crypto/index.ts
uhoreg: Huh. I was under the impression that `Object.keys` returns an iterator rather than an array, but looks like I was wrong.

uhoreg:
Apparently, https://github.com/matrix-org/matrix-js-sdk/pull/2037/commits/ae68f000435959d48e88aa925f8210d4e337a354 makes jest run for over an hour, which is probably not ideal.  Will try to figure out a workaround.

t3chguy:
Look into jest.useFakeTimers()

uhoreg:
I think my changes exceed the "needs re-review" threshold.  (I also added support for unprefixed properties, to match https://github.com/matrix-org/synapse/pull/11541)

turt2live (approved):
Thanks :)

PR #2041, opened 2021-11-29
Implement MSC3531: Hiding messages during moderation

Yoric:
Matrix-js-sdk-level support for hiding messages.

This is my first patch in matrix-js-sdk, please accept my apologies for anything I have done wrong.

MSC: https://github.com/matrix-org/matrix-doc/pull/3531

Signed-off-by: David Teller <davidt@element.io>

---

### DIFF
--- a/src/@types/event.ts
+++ b/src/@types/event.ts
@@ -177,6 +177,16 @@ export const UNSTABLE_ELEMENT_FUNCTIONAL_USERS = new UnstableValue(
     "io.element.functional_members",
     "io.element.functional_members");
 
+/**
+ * A type of message that affects visibility of a message,
+ * as per https://github.com/matrix-org/matrix-doc/pull/3531
+ *
+ * @experimental
+ */
+export const EVENT_VISIBILITY_CHANGE_TYPE = new UnstableValue(
+    "m.visibility",
+    "org.matrix.msc3531.visibility");
+
 export interface IEncryptedFile {
     url: string;
     mimetype?: string;
--- a/src/models/event.ts
+++ b/src/models/event.ts
@@ -28,6 +28,7 @@ import {
     EventType,
     MsgType,
     RelationType,
+    EVENT_VISIBILITY_CHANGE_TYPE,
 } from "../@types/event";
 import { Crypto, IEventDecryptionResult } from "../crypto";
 import { deepSortedObjectEntries } from "../utils";
@@ -125,6 +126,35 @@ export interface IEventRelation {
     key?: string;
 }
 
+export interface IVisibilityEventRelation extends IEventRelation {
+    visibility: "visible" | "hidden";
+    reason?: string;
+}
+
+/**
+ * When an event is a visibility change event, as per MSC3531,
+ * the visibility change implied by the event.
+ */
+export interface IVisibilityChange {
+    /**
+     * If `true`, the target event should be made visible.
+     * Otherwise, it should be hidden.
+     */
+    visible: boolean;
+
+    /**
+     * The event id affected.
+     */
+    eventId: string;
+
+    /**
+     * Optionally, a human-readable reason explaining why
+     * the event was hidden. Ignored if the event was made
+     * visible.
+     */
+    reason: string | null;
+}
+
 export interface IClearEvent {
     room_id?: string;
     type: string;
@@ -143,12 +173,42 @@ export interface IDecryptOptions {
     isRetry?: boolean;
 }
 
+/**
+ * Message hiding, as specified by https://github.com/matrix-org/matrix-doc/pull/3531.
+ */
+export type MessageVisibility = IMessageVisibilityHidden | IMessageVisibilityVisible;
+/**
+ * Variant of `MessageVisibility` for the case in which the message should be displayed.
+ */
+export interface IMessageVisibilityVisible {
+    readonly visible: true;
+}
+/**
+ * Variant of `MessageVisibility` for the case in which the message should be hidden.
+ */
+export interface IMessageVisibilityHidden {
+    readonly visible: false;
+    /**
+     * Optionally, a human-readable reason to show to the user indicating why the
+     * message has been hidden (e.g. "Message Pending Moderation").
+     */
+    readonly reason: string | null;
+}
+// A singleton implementing `IMessageVisibilityVisible`.
+const MESSAGE_VISIBLE: IMessageVisibilityVisible = Object.freeze({ visible: true });
+
 export class MatrixEvent extends EventEmitter {
     private pushActions: IActionsObject = null;
     private _replacingEvent: MatrixEvent = null;
     private _localRedactionEvent: MatrixEvent = null;
     private _isCancelled = false;
     private clearEvent?: IClearEvent;
+
+    /* Message hiding, as specified by https://github.com/matrix-org/matrix-doc/pull/3531.
+
+    Note: We're returning this object, so any value stored here MUST be frozen.
+    */
+    private visibility: MessageVisibility = MESSAGE_VISIBLE;
 
     /* curve25519 key which we believe belongs to the sender of the event. See
      * getSenderKey()
@@ -924,6 +984,53 @@ export class MatrixEvent extends EventEmitter {
     }
 
     /**
+     * Change the visibility of an event, as per https://github.com/matrix-org/matrix-doc/pull/3531 .
+     *
+     * @fires module:models/event.MatrixEvent#"Event.visibilityChange" if `visibilityEvent`
+     *   caused a change in the actual visibility of this event, either by making it
+     *   visible (if it was hidden), by making it hidden (if it was visible) or by
+     *   changing the reason (if it was hidden).
+     * @param visibilityEvent event holding a hide/unhide payload, or nothing
+     *   if the event is being reset to its original visibility (presumably
+     *   by a visibility event being redacted).
+     */
+    public applyVisibilityEvent(visibilityChange?: IVisibilityChange): void {
+        const visible = visibilityChange ? visibilityChange.visible : true;
+        const reason = visibilityChange ? visibilityChange.reason : null;
+        let change = false;
+        if (this.visibility.visible !== visibilityChange.visible) {
+            change = true;
+        } else if (!this.visibility.visible && this.visibility["reason"] !== reason) {
+            change = true;
+        }
+        if (change) {
+            if (visible) {
+                this.visibility = MESSAGE_VISIBLE;
+            } else {
+                this.visibility = Object.freeze({
+                    visible: false,
+                    reason: reason,
+                });
+            }
+            if (change) {
+                this.emit("Event.visibilityChange", this, visible);
+            }
+        }
+    }
+
+    /**
+     * Return instructions to display or hide the message.
+     *
+     * @returns Instructions determining whether the message
+     * should be displayed.
+     */
+    public messageVisibility(): MessageVisibility {
+        // Note: We may return `this.visibility` without fear, as
+        // this is a shallow frozen object.
+        return this.visibility;
+    }
+
+    /**
      * Update the content of an event in the same way it would be by the server
      * if it were redacted before it was sent to us
      *
@@ -990,6 +1097,54 @@ export class MatrixEvent extends EventEmitter {
      */
     public isRedaction(): boolean {
         return this.getType() === EventType.RoomRedaction;
+    }
+
+    /**
+     * Return the visibility change caused by this event,
+     * as per https://github.com/matrix-org/matrix-doc/pull/3531.
+     *
+     * @returns If the event is a well-formed visibility change event,
+     * an instance of `IVisibilityChange`, otherwise `null`.
+     */
+    public asVisibilityChange(): IVisibilityChange | null {
+        if (!EVENT_VISIBILITY_CHANGE_TYPE.matches(this.getType())) {
+            // Not a visibility change event.
+            return null;
+        }
+        const relation = this.getRelation();
+        if (!relation || relation.rel_type != "m.reference") {
+            // Ill-formed, ignore this event.
+            return null;
+        }
+        const eventId = relation.event_id;
+        if (!eventId) {
+            // Ill-formed, ignore this event.
+            return null;
+        }
+        const content = this.getWireContent();
+        const visible = !!content.visible;
+        const reason = content.reason;
+        if (reason && typeof reason != "string") {
+            // Ill-formed, ignore this event.
+            return null;
+        }
+        // Well-formed visibility change event.
+        return {
+            visible,
+            reason,
+            eventId,
+        };
+    }
+
+    /**
+     * Check if this event alters the visibility of another event,
+     * as per https://github.com/matrix-org/matrix-doc/pull/3531.
+     *
+     * @returns {boolean} True if this event alters the visibility
+     * of another event.
+     */
+    public isVisibilityEvent(): boolean {
+        return EVENT_VISIBILITY_CHANGE_TYPE.matches(this.getType());
     }
 
     /**
--- a/src/models/room-state.ts
+++ b/src/models/room-state.ts
@@ -618,14 +618,14 @@ export class RoomState extends EventEmitter {
     }
 
     /**
-     * Returns true if the given MatrixClient has permission to send a state
-     * event of type `stateEventType` into this room.
-     * @param {string} stateEventType The type of state events to test
-     * @param {MatrixClient} cli The client to test permission for
-     * @return {boolean} true if the given client should be permitted to send
-     *                        the given type of state event into this room,
-     *                        according to the room's state.
-     */
+      * Returns true if the given MatrixClient has permission to send a state
+      * event of type `stateEventType` into this room.
+      * @param {string} stateEventType The type of state events to test
+      * @param {MatrixClient} cli The client to test permission for
+      * @return {boolean} true if the given client should be permitted to send
+      *                        the given type of state event into this room,
+      *                        according to the room's state.
+      */
     public mayClientSendStateEvent(stateEventType: EventType | string, cli: MatrixClient): boolean {
         if (cli.isGuest()) {
             return false;
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -30,7 +30,10 @@ import { RoomMember } from "./room-member";
 import { IRoomSummary, RoomSummary } from "./room-summary";
 import { logger } from '../logger';
 import { ReEmitter } from '../ReEmitter';
-import { EventType, RoomCreateTypeField, RoomType, UNSTABLE_ELEMENT_FUNCTIONAL_USERS } from "../@types/event";
+import {
+    EventType, RoomCreateTypeField, RoomType, UNSTABLE_ELEMENT_FUNCTIONAL_USERS,
+    EVENT_VISIBILITY_CHANGE_TYPE,
+} from "../@types/event";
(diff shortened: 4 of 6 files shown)


src/models/event.ts
```diff
@@ -154,13 +154,27 @@ export interface IDecryptOptions {
...
+}
+
 export class MatrixEvent extends EventEmitter {
     private pushActions: IActionsObject = null;
     private _replacingEvent: MatrixEvent = null;
     private _localRedactionEvent: MatrixEvent = null;
     private _isCancelled = false;
     private clearEvent?: IClearEvent;
 
+    /* Message hiding, as specified by MSC3515.
```
turt2live, line 168: wrong MSC number? I suggest links to avoid people having to convert between these manually.

src/models/event.ts
```diff
@@ -934,6 +948,28 @@ export class MatrixEvent extends EventEmitter {
         this.event.unsigned.redacted_because = redactionEvent.event as IEvent;
     }
 
+    /**
+     * Apply a MSC3531 event to hide/unhide the message.
```
turt2live, line 952: links please

src/models/event.ts
```diff
@@ -934,6 +948,28 @@ export class MatrixEvent extends EventEmitter {
...
+     *
+     * @param visibilityEvent event causing the hide/unhide change.
+     * @return true if the event caused a change, false if it should be ignored.
+     */
+    public applyVisibilityChange(visible: boolean, reason?: string): boolean {
+        let change;
```
turt2live, line 958: types on `change` please, or update the ladder to early return.

src/models/event.ts
```diff
@@ -934,6 +948,28 @@ export class MatrixEvent extends EventEmitter {
         this.event.unsigned.redacted_because = redactionEvent.event as IEvent;
     }
 
+    /**
+     * Apply a MSC3531 event to hide/unhide the message.
+     *
+     * @param visibilityEvent event causing the hide/unhide change.
```
turt2live, line 954: This isn't a parameter

src/models/event.ts
```diff
@@ -1003,6 +1039,14 @@ export class MatrixEvent extends EventEmitter {
         return this.getType() === "m.room.redaction";
     }
 
+    public isVisibilityChange(): boolean {
```
turt2live, line 1042: Visibility can mean all sorts of things, so I'd suggest we name this to avoid confusion with directory and room visibility.
```suggestion
    public isEventVisibilityChange(): boolean {
```

src/models/event.ts
```diff
@@ -1003,6 +1039,14 @@ export class MatrixEvent extends EventEmitter {
...
+        const relation = this.getRelation();
+        if (!relation) {
+            return false;
+        }
+        return relation.rel_type == "m.visibility";
```
turt2live, line 1047: This cannot and should not be using the `m.*` namespace yet. Please use your unstable prefix, ideally through the `UnstableValue` class.

src/models/room.ts
```diff
@@ -151,6 +156,14 @@ export class Room extends EventEmitter {
      */
     public threads = new Map<string, Thread>();
 
+    /**
+     * A mapping of eventId to the latest visibility change to apply
+     * to the event.
+     *
+     * @experimental
```
turt2live, line 163: Would be good to link to the MSC here

src/models/room.ts
```diff
@@ -1365,6 +1378,71 @@ export class Room extends EventEmitter {
             // this may be needed to trigger an update.
         }
 
+        if (event.isVisibilityChange()) {
+            // Apply MSC3531-style visibility changes.
+            (() => {
```
turt2live, line 1383: Why is this a closure? (comments please)

src/models/room.ts
```diff
@@ -1365,6 +1378,71 @@ export class Room extends EventEmitter {
...
+                // Perform sanity checks on the event.
+                const relation = event.getRelation();
+                const originalEventId = relation.event_id;
+                let visible: boolean;
+                switch (relation["visibility"]) {
```
turt2live, line 1388: The MSC requires an unstable prefix for this field.

turt2live (changes requested):
As a first pass there's mostly spec-related workflow concerns on this, but the change overall seems sensible. I've also left a review on the MSC itself which might ultimately impact the review of the actual visibility change code.

Yoric:
Apologies, I realized too late that I was missing a big part of the patch to handle redactions. Added.

src/models/room.ts
Yoric: Moved to a method, it's more readable that way.

src/models/event.ts
```diff
@@ -1003,6 +1065,80 @@ export class MatrixEvent extends EventEmitter {
...
+            UNSTABLE_MSC3531_REASON_FIELD.unstable]) {
+            if (!key) {
+                continue;
+            }
+            const reason = content[key];
+            if (!reason && typeof reason == "string") {
+                return reason;
+            }
+        }
+        return null;
```
turt2live, line 1139:
```suggestion
        return UNSTABLE_MSC3531_REASON_FIELD.findIn(content);
```

src/models/room.ts
```diff
@@ -2249,6 +2288,184 @@ export class Room extends EventEmitter {
...
+
+    /**
+     * When we receive a new visibility change event:
+     *
+     * - store this visibility change alongside the timeline, in case we
+     *   later need to apply it to an event that we haven't received yet;
+     * - if we have already received the event whose visibility has changed,
+     *   patch it to reflect the visibility change and inform listeners.
+     */
+    private applyNewVisibilityChangeEvent(event: MatrixEvent): void {
```
turt2live, line 2300: from before: we should be consistent in the naming. If we're calling this an EventVisibility, then we should always use "EventVisibility".

Applies to the other functions as well.

src/models/room.ts
```diff
@@ -2249,6 +2288,184 @@ export class Room extends EventEmitter {
...
+            case "hidden":
+                visible = false;
+                break;
+            case "visible":
+                visible = true;
+                break;
+            default:
+                // Event is ill-formed.
+                return;
+        }
```
turt2live, line 2320: This can be shortened:

```suggestion
        visible = event.getEventVisibilityChange();
        if (!visible && visible !== false) return; // event is ill-formed
```

src/models/room.ts
```diff
@@ -2249,6 +2288,184 @@ export class Room extends EventEmitter {
...
+        const reason = visible ? event.getContent()["reason"] : null;
+        if (reason != null && reason !== undefined && typeof reason != "string") {
```
turt2live, line 2322:
```suggestion
        if (reason && typeof reason != "string") {
```

src/models/room.ts
```diff
@@ -2249,6 +2288,184 @@ export class Room extends EventEmitter {
...
+            // Event is ill-formed.
+            return;
+        }
+
+        // Ignore visibility change events that are not emitted by moderators.
+        const powerLevelsEvents = this.currentState.getStateEvents(EventType.RoomPowerLevels, "");
+        const powerLevels = powerLevelsEvents && powerLevelsEvents.getContent();
+        if (!powerLevels || powerLevels["org.matrix.msc3531.visibility"] < event.sender.powerLevel) {
```
turt2live, line 2330: unstable value for the namespace please :)

src/models/room.ts
```diff
@@ -2249,6 +2288,184 @@ export class Room extends EventEmitter {
...
+            // 2 visibility changes, so linear scans through `visibilityChanges`
+            // should be fast. It might be possible to DoS this feature, though.
+            let index = visibilityChangesOnOriginalEvent.length - 1;
+            for (; index >= 0; --index) {
+                const target = visibilityChangesOnOriginalEvent[index];
+                if (target.getTs() < event.getTs()) {
+                    break;
+                }
+            }
+            if (index == -1) {
```
turt2live, line 2353:
```suggestion
            if (index === -1) {
```

src/models/room.ts
```diff
@@ -2249,6 +2288,184 @@ export class Room extends EventEmitter {
...
+        // we may need to apply the visibility change at a later date.
+
+        const visibilityChangesOnOriginalEvent = this.visibilityChanges.get(originalEventId);
+        if (visibilityChangesOnOriginalEvent) {
+            // It would be tempting to simply erase the latest visibility change
+            // but we need to record all of the changes in case the latest change
+            // is ever redacted.
+            // In practice, it should be unusual for an event to have more than
+            // 2 visibility changes, so linear scans through `visibilityChanges`
+            // should be fast. It might be possible to DoS this feature, though.
```
turt2live, line 2345: We can always apply a limit to the loop, and we probably should if it's highlighted as an attack here.

src/models/room.ts
```diff
@@ -2249,6 +2288,184 @@ export class Room extends EventEmitter {
...
+        const relation = event.getRelation();
+        const originalEventId = relation.event_id;
+        const visibilityChangesOnOriginalEvent = this.visibilityChanges.get(originalEventId);
+        if (!visibilityChangesOnOriginalEvent) {
+            // No visibility changes on the original event.
+            // In particular, this change event was not recorded,
+            // most likely because it was ill-formed.
+            return;
+        }
+        const index = visibilityChangesOnOriginalEvent.findIndex(change => change.getId() == event.getId());
```
turt2live, line 2388:
```suggestion
        const index = visibilityChangesOnOriginalEvent.findIndex(change => change.getId() === event.getId());
```

src/models/room.ts
```diff
@@ -2249,6 +2288,184 @@ export class Room extends EventEmitter {
...
+        if (index == -1) {
```
turt2live, line 2389:
```suggestion
        if (index === -1) {
```

turt2live: this sort of comment applies throughout - going to stop posting about it for this review :)

src/models/room.ts
```diff
@@ -2249,6 +2288,184 @@ export class Room extends EventEmitter {
...
+    /**
+     * When we receive an event whose visibility has been altered by
+     * a (more recent) visibility change event, patch the event in
+     * place so that clients now not to display it.
+     *
+     * @param event Any matrix event. If this event has at least one a
+     * pending visibility change event, apply the latest visibility
+     * change event.
+     */
+    private applyPendingVisibilityChanges(event: MatrixEvent): void {
```
turt2live, line 2447: same comment as the other function: EventVisibility please :)

turt2live (commented):
overall this is looking fine - just a few more nitpicks, mostly around the project structure

src/models/room.ts
Yoric: Not sure I understand how you go from a string to a boolean.

src/models/room.ts
Yoric: That would mean that if `reason == ""`, we behave as if `reason` was `null`? That makes sense.

src/models/room.ts
Yoric: How should we react when we reach the limit?

turt2live: I'd just truncate. If someone is flipping visibility more than 50 times, they kinda deserve whatever state we ended up on.

src/@types/event.ts
```diff
@@ -180,26 +180,32 @@ export const UNSTABLE_ELEMENT_FUNCTIONAL_USERS = new UnstableValue(
 /**
  * A rel_type that affects visibility of a message,
  * as per https://github.com/matrix-org/matrix-doc/pull/3531
+ *
+ * @experimental
  */
-export const UNSTABLE_MSC3531_VISIBILITY_CHANGE_REL_TYPE = new UnstableValue(
+export const MSC3531_VISIBILITY_CHANGE_REL_TYPE = new UnstableValue(
```
turt2live, line 186: sorry, also meant to say that the MSC number should be dropped from the constant's name as well. Also, `EVENT` should be somewhere in there to line up with all the other EventVisibility things.

src/models/room.ts
turt2live: per internal: I'm not really seeing the value of the string, so would change this to be a boolean.

src/models/room.ts
turt2live: there are entire swathes of code in our projects which rely on empty strings being falsy, so I wouldn't worry about it.

src/models/event.ts
```diff
@@ -923,6 +983,53 @@ export class MatrixEvent extends EventEmitter {
...
+     *   changing the reason (if it was hidden).
+     * @param visibilityEvent event holding a hide/unhide payload, or nothing
+     *   if the event is being reset to its original visibility (presumably
+     *   by a visibility event being redacted).
+     */
+    public applyVisibilityEvent(visibilityChange?: IVisibilityChange): void {
+        const visible = visibilityChange ? visibilityChange.visible : true;
+        const reason = visibilityChange ? visibilityChange.reason : null;
+        let change = false;
+        if (this.visibility.visible != visibilityChange.visible) {
```
turt2live, line 1001: triple equals throughout please.

src/models/room-state.ts
```diff
@@ -607,13 +608,15 @@ export class RoomState extends EventEmitter {
...
-     * @param {string} eventType The type of event to test
+     * @param {string} eventType The type of event to test.
      * @param {string} userId The user ID of the user to test permission for
      * @return {boolean} true if the given user ID should be permitted to send
      *                        the given type of event into this room,
      *                        according to the room's state.
      */
-    public maySendEvent(eventType: EventType | string, userId: string): boolean {
+    public maySendEvent<S extends string, T extends string>(
+        eventType: EventType | string | UnstableValue<S, T>,
```
turt2live, line 618: These functions should not be taking `UnstableValue` directly. The caller will need to call the function for both types if it wants to.

This should also remove the iterator requirement on `UnstableValue` - something that shouldn't be present there.

yarn.lock
```diff
@@ -1210,9 +1210,9 @@
     "@types/yargs" "^15.0.0"
     chalk "^4.0.0"
 
-"@matrix-org/olm@https://gitlab.matrix.org/api/v4/projects/27/packages/npm/@matrix-org/olm/-/@matrix-org/olm-3.2.3.tgz":
-  version "3.2.3"
-  resolved "https://gitlab.matrix.org/api/v4/projects/27/packages/npm/@matrix-org/olm/-/@matrix-org/olm-3.2.3.tgz#cc332fdd25c08ef0e40f4d33fc3f822a0f98b6f4"
+"@matrix-org/olm@https://gitlab.matrix.org/api/v4/projects/27/packages/npm/@matrix-org/olm/-/@matrix-org/olm-3.2.7.tgz":
```
turt2live, line 1213: not sure what happened here, but this seems unrelated.

turt2live (changes requested):
almost there :)

There's also MSC feedback which will need addressing ftr. Those aspects are not covered in code review.

src/@types/event.ts
```diff
@@ -177,6 +177,16 @@ export const UNSTABLE_ELEMENT_FUNCTIONAL_USERS = new UnstableValue(
...
 
+/**
+ * A type of message that affects visibility of a message,
+ * as per https://github.com/matrix-org/matrix-doc/pull/3531
+ *
+ * @experimental
+ */
+export const EVENT_VISIBILITY_CHANGE_TYPE = new UnstableValue(
+    "org.matrix.msc3531.visibility",
+    "m.visibility");
```
turt2live, line 188: These arguments are backwards

src/models/room-state.ts
Yoric: > These functions should not be taking `UnstableValue` directly. The caller will need to call the function for both types if it wants to.

Why? It feels to me like this is both a common operation when prototyping events and something error-prone (and slower) to reimplement manually.

> This should also remove the iterator requirement on UnstableValue - something that shouldn't be present there.

I don't understand that remark, either.

t3chguy: The logic would end up duplicated anyway as the user may have permission to send one but not the other so the called needs to know which one in the case of knowing whether it may have to send the unstable one because thats the only one it has perms to send to, if that's even a thing it wishes to do.

Yoric: > 

How's that different from the code I wrote that we're discussing?

turt2live: Behaviourally it's fine for your PR, but it isn't necessarily safe for other implementations or features. It's very possible that a feature down the line needs explicitly one or the other, and this signature would introduce a potential false sense of security to a developer.

Yoric: Ok, I've changed the implementation.

src/models/room-state.ts
```diff
@@ -605,27 +605,27 @@ export class RoomState extends EventEmitter {
...
-     *                        according to the room's state.
-     */
+      * Returns true if the given user ID has permission to send a normal
+      * event of type `eventType` into this room.
+      * @param {string} eventType The type of event to test
+      * @param {string} userId The user ID of the user to test permission for
+      * @return {boolean} true if the given user ID should be permitted to send
+      *                        the given type of event into this room,
+      *                        according to the room's state.
+      */
```
turt2live, line 615: all of the jsdoc comments appear to have gotten mangled by the revert

src/models/room.ts
```diff
@@ -2254,6 +2311,152 @@ export class Room extends EventEmitter {
...
+            // The event is ill-formed.
+            return;
+        }
+
+        // Ignore visibility change events that are not emitted by moderators.
+        const userId = event.getSender();
+        if (!userId) {
+            return;
+        }
+        if (!EVENT_VISIBILITY_CHANGE_TYPE.some(name => this.currentState.maySendStateEvent(name, userId))) {
```
turt2live, line 2335: I'm not really convinced that we need an iterator on this - we can just call `maySendStateEvent` twice. It'd be a bit different if we were supporting more than two namespaces or the classes had a purpose outside of detecting presence within an object.

Or in other words: the usage of `some` feels backwards here. A namespaced value is intended to retrieve values from an object/array, not process an entity to check compliance. 

Storing two variables and checking either is more than fine, I think. The power level check uses local cache so should have no performance impact.

yarn.lock
```diff
@@ -1210,9 +1210,9 @@
     "@types/yargs" "^15.0.0"
     chalk "^4.0.0"
 
-"@matrix-org/olm@https://gitlab.matrix.org/api/v4/projects/27/packages/npm/@matrix-org/olm/-/@matrix-org/olm-3.2.3.tgz":
-  version "3.2.3"
-  resolved "https://gitlab.matrix.org/api/v4/projects/27/packages/npm/@matrix-org/olm/-/@matrix-org/olm-3.2.3.tgz#cc332fdd25c08ef0e40f4d33fc3f822a0f98b6f4"
+"@matrix-org/olm@https://gitlab.matrix.org/api/v4/projects/27/packages/npm/@matrix-org/olm/-/@matrix-org/olm-3.2.8.tgz":
```
turt2live, line 1213: This seems unrelated.

turt2live (commented):
one more step, and then we're there :)

src/models/room.ts
Yoric: It strikes me as error-prone, but I'll stop insisting.

yarn.lock
Yoric: Gasp. I'm almost sure that I have reverted that change several times already. Not sure what's going on.

turt2live (approved):
alright, let's send it.

turt2live:
... after you fix merge conflicts and the build :D (try merging develop - should fix both, hopefully)

turt2live:
(please don't force push onto reviewed branches - it invalidates review, thus making it required for re-review. In this case, the changes appear fine)

PR #2043, opened 2021-12-01
Threads notifications after app startup

gsouquet:
Related to https://github.com/vector-im/element-web/issues/19126

- Allow for push actions refresh for event decryption
- Emit event when a new reply is added to a thread

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -881,8 +881,8 @@ export class MatrixClient extends EventEmitter {
             // Ensure the unread counts are kept up to date if the event is encrypted
             // We also want to make sure that the notification count goes up if we already
             // have encrypted events to avoid other code from resetting 'highlight' to zero.
-            const oldHighlight = !!oldActions?.tweaks?.highlight ?? false;
-            const newHighlight = !!actions?.tweaks?.highlight ?? false;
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -882,10 +881,8 @@ export class MatrixClient extends EventEmitter {
             // Ensure the unread counts are kept up to date if the event is encrypted
             // We also want to make sure that the notification count goes up if we already
             // have encrypted events to avoid other code from resetting 'highlight' to zero.
-            const oldHighlight = oldActions && oldActions.tweaks
-                ? !!oldActions.tweaks.highlight : false;
-            const newHighlight = actions && actions.tweaks
-                ? !!actions.tweaks.highlight : false;
+            const oldHighlight = !!oldActions?.tweaks?.highlight ?? false;
+            const newHighlight = !!actions?.tweaks?.highlight ?? false;
```
t3chguy, line 885: This transpiles to `(_b = !!((_a = oldActions === null || oldActions === void 0 ? void 0 : oldActions.tweaks) === null || _a === void 0 ? void 0 : _a.highlight)) !== null && _b !== void 0 ? _b : false;`

Given the `!!` it'll never be null/undefined

t3chguy (approved):
LGTM otherwise

PR #2047, opened 2021-12-02
FIX exports type

skyka13711:
webpack couldn't retrieve the previous import structure
https://github.com/matrix-org/matrix-js-sdk/pull/2023

---

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -32,8 +32,3 @@ try {
 export * from "./matrix";
 export default matrixcs;
 
-export * from './@types/event';
-export * from './@types/PushRules';
-export * from './@types/requests';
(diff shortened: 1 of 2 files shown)


turt2live:
Do you have more information as to what the issue with webpack is? The project needs a bit of special care to include in a webpack project, which we can help out with in #element-dev:matrix.org on Matrix.

skyka13711:
We use matrix-js-sdk in our project and build it with webpack. When we import, for example, enum SearchOrderBy with webpack, it is resolved as undefined. Also, it would be much more uniform if all modules were exported the same way

turt2live (approved):

PR #2048, opened 2021-12-03
Finish adding registration token UIA type

govynnus:
Signed-off-by: Callum Brown <callum@calcuode.com>

I think this should complete what I started in #2020 but there might still be something missing....

---


turt2live (approved):
seems reasonable as a types change - thanks!

PR #2049, opened 2021-12-03
Fix incorrect MSC3089 typings and add null checks

hughns:
---

## 🐛 Bug Fixes
 * Fix incorrect MSC3089 typings and add null checks ([\#2049](https://github.com/matrix-org/matrix-js-sdk/pull/2049)). Contributed by @hughns.


gsouquet (approved):
Looks great to me! ✨

PR #2050, opened 2021-12-06
typescript does not find declarations

aanation:
In this [commit](https://github.com/matrix-org/matrix-js-sdk/commit/55b489b17a519bc0d557c47ac9fe7ce54690b602#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) you changed the "main" option in package.json. 

For this reason, the package in the develop branch does not work (typescript in my project does't know where to take the declarations, and tries to compile the .ts files from the src directory). 
Your comment says "Resetting package fields for development". Are you sure you haven't forgotten about it? 
Maybe we should add types/typings section in the develop branch?

I am worried that this change will be released into the npm package

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
   "matrix_src_browser": "./src/browser-index.js",
   "matrix_lib_main": "./lib/index.js",
   "matrix_lib_typings": "./lib/index.d.ts",
+  "types": "./lib",
   "author": "matrix.org",
   "license": "Apache-2.0",
   "files": [


aanation:
> This is done automatically by the release script (adds it for the the release then removes it again for development) so that, on the develop branch, the default entry point is the typescript version and the types come directly from the source rather than from built types. It's far from an ideal solution, but yes, it is intentional.

Okay, that makes me feel better :)
However, what prevents you from using the d.ts in the development environment instead of sources? 

And another question, if we're talking about declarations. Have you thought about generating single file d.ts + one bundle .js file?

t3chguy:
> However, what prevents you from using the d.ts in the development environment instead of sources?

Having to run tsc in watch mode as a separate process rather than doing everything in a single webpack

turt2live (commented):
marking as reviewed while the change is discussed.

richvdh:
superceded by https://github.com/matrix-org/matrix-js-sdk/pull/4357 I think

PR #2051, opened 2021-12-07
Add new room event filter fields

gsouquet:
As they are described in MSC3440. Useful to fetch all threads in a room

---

## ✨ Features
 * Add new room event filter fields ([\#2051](https://github.com/matrix-org/matrix-js-sdk/pull/2051)).

### DIFF
--- a/src/filter.ts
+++ b/src/filter.ts
@@ -66,7 +66,7 @@ export interface IRoomEventFilter extends IFilterComponent {
     lazy_load_members?: boolean;
     include_redundant_members?: boolean;
     types?: EventType[] | string[];
-    [UNSTABLE_FILTER_RELATION_TYPES.name]?: RelationType[] | string[];
+    [UNSTABLE_FILTER_RELATION_TYPES.name]?: Array<RelationType | string>;
     [UNSTABLE_FILTER_RELATION_SENDERS.name]?: string[];
 }
 
--- a/src/filter.ts
+++ b/src/filter.ts
@@ -18,9 +18,23 @@ limitations under the License.
  * @module filter
  */
 
-import { EventType, RelationType } from "./@types/event";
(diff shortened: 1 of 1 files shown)


src/filter.ts
```diff
@@ -18,8 +18,23 @@ limitations under the License.
...
+import {
+    EventType,
+    RelationType,
+} from "./@types/event";
 import { FilterComponent, IFilterComponent } from "./filter-component";
 import { MatrixEvent } from "./models/event";
+import { UnstableValue } from "./NamespacedValue";
+
+export const UNSTABLE_FILTER_RELATION_SENDERS = new UnstableValue(
+    "relation_senders",
```
andybalaam, line 30: Wouldn't this be something like `m.relation_senders` ?  Or am I misunderstanding?

andybalaam (approved):

src/filter.ts
t3chguy: Nope, the filter fields aren't prefixed with `m.` https://github.com/matrix-org/matrix-doc/blob/gsouquet/threading-via-relations/proposals/3440-threading-via-relations.md

src/filter.ts
```diff
@@ -52,8 +66,8 @@ export interface IRoomEventFilter extends IFilterComponent {
     lazy_load_members?: boolean;
     include_redundant_members?: boolean;
     types?: EventType[] | string[];
-    relation_types?: RelationType[] | string[];
-    relation_senders?: string[];
+    [UNSTABLE_FILTER_RELATION_TYPES.name]?: RelationType[] | string[];
```
t3chguy, line 69: should be `Array<RelationType | string>` so that you can mix-n-match

t3chguy (approved):
LGTM

PR #2055, opened 2021-12-09
Improve typing

t3chguy:
Very small step of the way towards noImplicitAny

---


andybalaam (approved):
Thanks, nice work.

PR #2056, opened 2021-12-09
Fix types

t3chguy:
---

### DIFF
--- a/src/crypto/SecretStorage.ts
+++ b/src/crypto/SecretStorage.ts
@@ -37,7 +37,7 @@ export interface ISecretRequest {
 
 export interface IAccountDataClient extends EventEmitter {
     // Subset of MatrixClient (which also uses any for the event content)
-    getAccountDataFromServer: <T>(eventType: string) => Promise<T>;
+    getAccountDataFromServer: <T extends {[k: string]: any}>(eventType: string) => Promise<T>;
     getAccountData: (eventType: string) => MatrixEvent;
     setAccountData: (eventType: string, content: any) => Promise<{}>;
 }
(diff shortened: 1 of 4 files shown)


src/crypto/SecretStorage.ts
```diff
@@ -277,7 +277,7 @@ export class SecretStorage {
      * @return {string} the contents of the secret
      */
     public async get(name: string): Promise<string> {
-        const secretInfo = await this.accountDataAdapter.getAccountDataFromServer<any>(name); // TODO types
+        const secretInfo = await this.accountDataAdapter.getAccountDataFromServer(name);
```
turt2live, line 280: we should be adding types, not removing them.

src/crypto/SecretStorage.ts
```diff
@@ -339,7 +339,7 @@ export class SecretStorage {
      */
     public async isStored(name: string, checkKey: boolean): Promise<Record<string, ISecretStorageKeyInfo>> {
         // check if secret exists
-        const secretInfo = await this.accountDataAdapter.getAccountDataFromServer<any>(name); // TODO types
```
turt2live, line 342: we should be adding types, not removing them.

src/crypto/SecretStorage.ts
```diff
@@ -352,7 +352,7 @@ export class SecretStorage {
         // filter secret encryption keys with supported algorithm
         for (const keyId of Object.keys(secretInfo.encrypted)) {
             // get key information from key storage
-            const keyInfo = await this.accountDataAdapter.getAccountDataFromServer<any>( // TODO types
+            const keyInfo = await this.accountDataAdapter.getAccountDataFromServer(
```
turt2live, line 355: we should be adding types, not removing them.

turt2live (changes requested):

turt2live (approved):
Thanks :)

PR #2057, opened 2021-12-10
Thread view integration with HomeServer API

gsouquet:
Partially addresses https://github.com/vector-im/element-web/issues/19588 and https://github.com/vector-im/element-web/issues/19961

Integrates with the `/relations` API and the bundled relationships
Also provides a fallback for when a server does not have support for those APIs

- Integrate threads with homeserver APIs
- Store pagination token directly on timeline
- Update for threads with server side support

---


src/@types/event.ts
```diff
@@ -114,6 +115,12 @@ export enum MsgType {
     KeyVerificationRequest = "m.key.verification.request",
 }
 
+export interface ThreadBundledRelation {
```
t3chguy, line 118: We normally prefix our interfaces with an `I`

src/sync.ts
```diff
@@ -1704,8 +1710,8 @@ export class SyncApi {
     /**
      * @experimental
      */
-    private processThreadEvents(room: Room, threadedEvents: MatrixEvent[]): void {
-        return this.client.processThreadEvents(room, threadedEvents);
+    private processThreadEvents(room: Room, threadedEvents: MatrixEvent[], opts): void {
```
t3chguy, line 1713: type?

t3chguy (approved):
Looks sane otherwise

gsouquet:
Asking for a re-review as some things to do with back pagination have significantly changed since the last review

gsouquet:
Superseded by https://github.com/matrix-org/matrix-js-sdk/pull/2142

PR #2061, opened 2021-12-13
Iterate typing to work towards noImplicitAny

t3chguy:
---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1424,11 +1424,10 @@ export class MatrixClient extends EventEmitter {
         // We swallow errors because we need a default object anyhow
         return this.http.authedRequest(
             undefined, "GET", "/capabilities",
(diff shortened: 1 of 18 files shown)


src/client.ts
```diff
@@ -1424,11 +1453,10 @@ export class MatrixClient extends EventEmitter {
         // We swallow errors because we need a default object anyhow
         return this.http.authedRequest(
             undefined, "GET", "/capabilities",
-        ).catch((e: Error) => {
+        ).catch((e: Error): {} => {
```
jryans, line 1456: What is the `: {}` doing...? Is that return type? But it doesn't return an object... 🤔

jryans (approved):
Looks good overall, thanks! 😄

t3chguy:
Going to land after RC

PR #2065, opened 2021-12-13
Update filters to reflect MSC3440 (threads)

gsouquet:
- Allow omitting the "fromToken" when making a message request
- Make event filtering work for threads

---


src/models/room.ts
```diff
@@ -109,6 +109,10 @@ export enum NotificationCountType {
     Total = "total",
 }
 
+export interface ICreateFilterOpts {
+    prepopulateTimeline?: boolean;
```
turt2live, line 113: this should get docs

src/filter-component.ts
```diff
@@ -89,6 +90,8 @@ export class FilterComponent {
             senders: this.filterJson.senders || null,
             not_senders: this.filterJson.not_senders || [],
             contains_url: this.filterJson.contains_url || null,
+            [UNSTABLE_FILTER_RELATION_SENDERS.name]: this.filterJson[UNSTABLE_FILTER_RELATION_SENDERS.name],
+            [UNSTABLE_FILTER_RELATION_TYPES.name]: this.filterJson[UNSTABLE_FILTER_RELATION_TYPES.name],
```
turt2live, line 94:
```suggestion
            [UNSTABLE_FILTER_RELATION_SENDERS.name]: UNSTABLE_FILTER_RELATION_SENDERS.findIn(this.filterJson),
            [UNSTABLE_FILTER_RELATION_TYPES.name]: UNSTABLE_FILTER_RELATION_TYPES.findIn(this.filterJson),
```

src/client.ts
```diff
@@ -4991,21 +4990,22 @@ export class MatrixClient extends EventEmitter {
     // XXX: Intended private, used in code.
     public createMessagesRequest(
         roomId: string,
-        fromToken: string,
-        limit: number,
+        fromToken: string | null,
```
turt2live, line 4993: This is relying on unspecced behaviour of Synapse. A `from` token is required by the spec.

https://spec.matrix.org/v1.1/client-server-api/#get_matrixclientv3roomsroomidmessages

turt2live (changes requested):
The code overall seems fine, though does rely on unspecced behaviour. We need to figure out a way forward on that before continuing here, I think.

As a gauge of difficulty, the spec has previously rejected the change to allow optional `from`. As a reference SDK we cannot (or rather, shouldn't) let ourselves become bound to Synapse behaviour.

src/models/room.ts
```diff
@@ -1209,7 +1213,10 @@ export class Room extends EventEmitter {
      * @param {Filter} filter The filter to be applied to this timelineSet
      * @return {EventTimelineSet} The timelineSet
      */
-    public getOrCreateFilteredTimelineSet(filter: Filter): EventTimelineSet {
+    public getOrCreateFilteredTimelineSet(
+        filter: Filter,
+        { prepopulateTimeline = true }: ICreateFilterOpts = {},
```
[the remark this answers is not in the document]
gsouquet: Should we continue to document all params on top of the typescript type definition? I have witnessed a disconnect between the comments and the function declaration in a couple of places.
This is probably not the right forum to make that decision, will open that up to the rest of the team

src/client.ts
gsouquet: Thanks, I did realise that afterwards. It seems that synapse supports that because the android client used to make some calls without that parameter.
Me and @clokep are discussing how EW could get its hands on a pagination token without going through the whole sync flow as it won't fit the bill here.

Won't merge this pull request until we have solved this

clokep: To cross-reference, see matrix-org/synapse#5538. There's desire to remove this capability from Synapse (in order to be spec compliant).

It was suggested that there should be a sync token somewhere to use in order to use this API, but I think @gsouquet wants to not have to sync the entire filtered request, but simply paginate through a filtered set of messages, which seems like quite a reasonable thing to do?

turt2live: it's very reasonable in my opinion to have `from` be optional, largely for cases like this. It just needs to be in the spec as such, or at least as an MSC with net-positive feedback.

clokep:
> Blocked, waiting on `from` token made optional. @clokep is going to open an MSC to propose an update to the Matrix spec

See [MSC3567](https://github.com/matrix-org/matrix-doc/pull/3567).

src/client.ts
gsouquet: This has been proposed in https://github.com/matrix-org/matrix-doc/pull/3567
I will consider this pull request unblocked as it's likely to be approved and enter FCP very soon

turt2live (approved):
This seems fine, though if https://github.com/matrix-org/matrix-doc/pull/3567 fails for some reason then we'll need to backpedal on this change, so I would not recommend merging until the MSC has at least passed FCP. At that point it's stable enough for usage and we don't need to do awkward steering here in the js-sdk. Before FCP finishes, we have to do quite a few things that honestly aren't worth the effort.

PR #2066, opened 2021-12-13
Load room history if necessary when searching for MSC3089 getFileEvent()

hughns:
---

## ✨ Features
 * Load room history if necessary when searching for MSC3089 getFileEvent() ([\#2066](https://github.com/matrix-org/matrix-js-sdk/pull/2066)). Contributed by @hughns.


src/models/MSC3089Branch.ts
```diff
@@ -140,7 +140,14 @@ export class MSC3089Branch {
         const room = this.client.getRoom(this.roomId);
         if (!room) throw new Error("Unknown room");
 
-        const event = room.getUnfilteredTimelineSet().findEventById(this.id);
+        let event: MatrixEvent | undefined = room.getUnfilteredTimelineSet().findEventById(this.id);
+
+        // keep scrolling back if needed until we find the event or reach the start of the room:
+        while (!event && room.oldState.paginationToken) {
```
turt2live, line 146: do we specifically want `oldState`, or `currentState`? Using `oldState` is sufficiently rare where it should get a comment when we deliberately use it.

turt2live (changes requested):
otherwise lgtm

src/models/MSC3089Branch.ts
hughns: Good spot. `currentState` makes more sense and I've also used the non-legacy implementation.

turt2live (approved):
Thanks :D

PR #2067, opened 2021-12-14
Move Room properties JSDoc into more usable comments on actual props

hughns:
This makes the docs more accessible from IDEs. However, will this impact the auto-generated JSDoc docset negatively?

---


gsouquet (approved):

t3chguy:
Yeah this completely butchers the generated docs 

Before
http://matrix-org.github.io/matrix-js-sdk/15.3.0-rc.1/module-models_room.html
image


After
image

PR #2076, opened 2021-12-15
Provide a makeLocationContent to create a location content event

andybalaam:
---

### DIFF
--- a/src/@types/location.ts
+++ b/src/@types/location.ts
@@ -19,7 +19,6 @@ limitations under the License.
 import { UnstableValue } from "../NamespacedValue";
 import { IContent } from "../models/event";
 import { TEXT_NODE_TYPE } from "./extensible_events";
-import { MsgType } from "./event";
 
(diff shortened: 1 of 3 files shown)


src/@types/location.ts
```diff
@@ -50,9 +51,29 @@ export interface ILocationContent extends IContent {
...
+        "geo_uri": uri,
+        [LOCATION_EVENT_TYPE.name]: {
+            uri,
+            description,
+        },
+        [TIMESTAMP_NODE_TYPE.name]: ts,
+        [TEXT_NODE_TYPE.name]: text,
+        // TODO: MSC1767 fallbacks m.image thumbnail
+    };
+}
```
t3chguy, line 79: feels a bit strange to have utility methods inside `/@types/`

andybalaam: Where would you suggest?

t3chguy: `content-helpers.ts`

t3chguy (approved):

PR #2079, opened 2021-12-17
Fix TypeScript issues

zhaytee:
Signed-off-by: Joe Toscano <joetoscano@pm.me>

Notes: These minor typing issues prevent building dependent projects which need `"strict": true` in tsconfig.json.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -455,7 +455,7 @@ interface IUploadKeySignaturesResponse {
 }
 
 export interface IPreviewUrlResponse {
-    [key: string]: string | number;
+    [key: string]: string | number | undefined;
     "og:title": string;
     "og:type": string;
     "og:url": string;
--- a/src/store/memory.ts
+++ b/src/store/memory.ts
@@ -334,7 +334,7 @@ export class MemoryStore implements IStore {
      * @param {string} eventType The event type being queried
      * @return {?MatrixEvent} the user account_data event of given type, if any
      */
-    public getAccountData(eventType: EventType | string): MatrixEvent | undefined {
+    public getAccountData(eventType: EventType | string): MatrixEvent {
         return this.accountData[eventType];
     }
 


src/store/memory.ts
```diff
@@ -334,7 +334,7 @@ export class MemoryStore implements IStore {
      * @param {string} eventType The event type being queried
      * @return {?MatrixEvent} the user account_data event of given type, if any
      */
-    public getAccountData(eventType: EventType | string): MatrixEvent | undefined {
+    public getAccountData(eventType: EventType | string): MatrixEvent {
```
turt2live, line 337: it can return undefined though? The docs would need updating otherwise.

src/client.ts
```diff
@@ -455,7 +455,7 @@ interface IUploadKeySignaturesResponse {
 }
 
 export interface IPreviewUrlResponse {
-    [key: string]: string | number;
+    [key: string]: string | number | undefined;
```
turt2live, line 458: this seems a bit silly to type - what's the error from typescript?

turt2live (changes requested):
Thanks for taking a look! I'm a little confused as to why strict mode is causing these problems, but happy to be convinced :)

src/client.ts
zhaytee: 6 of the values are marked as optional with `?` (such as `"og:image"?`) so tsc says:
```
node_modules/matrix-js-sdk/lib/client.d.ts:305:5 - error TS2411: Property '"og:image"' of type 'string | undefined' is not assignable to 'string' index type 'string | number'.
```

src/store/memory.ts
zhaytee:
```
node_modules/matrix-js-sdk/lib/store/memory.d.ts:168:5 - error TS2416: Property 'getAccountData' in type 'MemoryStore' is not assignable to the same property in base type 'IStore'.
  Type '(eventType: string) => MatrixEvent | undefined' is not assignable to type '(eventType: string) => MatrixEvent'.
    Type 'MatrixEvent | undefined' is not assignable to type 'MatrixEvent'.
      Type 'undefined' is not assignable to type 'MatrixEvent'.
```

turt2live: This sounds like an issue with the upper interface, not with the store.

MadLittleMods: To expand on what @turt2live mentioned, it sounds like we need to update `IStore` to allow `undefined` as well,

https://github.com/matrix-org/matrix-js-sdk/blob/871149912102b7fe65dc49df35be033f7cea5e62/src/store/index.ts#L160-L164

PR #2086, opened 2021-12-22
Fix more function typings relating to key backup

hughns:
These typings should now match JSDoc and the implementation.

---

## 🐛 Bug Fixes
 * Fix more function typings relating to key backup ([\#2086](https://github.com/matrix-org/matrix-js-sdk/pull/2086)). Contributed by @hughns.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2658,6 +2658,9 @@ export class MatrixClient extends EventEmitter {
         );
     }
 
+    private makeKeyBackupPath(roomId: undefined, sessionId: undefined, version: string): IKeyBackupPath;
+    private makeKeyBackupPath(roomId: string, sessionId: undefined, version: string): IKeyBackupPath;
+    private makeKeyBackupPath(roomId: string, sessionId: string, version: string): IKeyBackupPath;
     private makeKeyBackupPath(
         roomId: string | undefined,
         sessionId: string | undefined,
@@ -2689,10 +2692,13 @@ export class MatrixClient extends EventEmitter {
      * @return {Promise} a promise that will resolve when the keys
      * are uploaded
      */
+    public sendKeyBackup(roomId: undefined, sessionId: undefined, version: string, data: IKeyBackup): Promise<void>;
+    public sendKeyBackup(roomId: string, sessionId: undefined, version: string, data: IKeyBackup): Promise<void>;
+    public sendKeyBackup(roomId: string, sessionId: string, version: string, data: IKeyBackup): Promise<void>;
     public sendKeyBackup(
-        roomId: string | undefined,
+        roomId: string,
         sessionId: string | undefined,
-        version: string,
+        version: string | undefined,
         data: IKeyBackup,
     ): Promise<void> {
         if (!this.crypto) {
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -2658,7 +2658,11 @@ export class MatrixClient extends EventEmitter {
         );
     }
 
-    private makeKeyBackupPath(roomId: string, sessionId: string, version: string): IKeyBackupPath {
+    private makeKeyBackupPath(
+        roomId: string | undefined,
+        sessionId: string | undefined,
+        version: string,
+    ): IKeyBackupPath {
```
t3chguy, line 2665: this isn't quite right, `roomId` can only be undefined if `sessionId` is undefined too, this should be an overload signature

src/client.ts
```diff
@@ -2685,7 +2689,12 @@ export class MatrixClient extends EventEmitter {
      * @return {Promise} a promise that will resolve when the keys
      * are uploaded
      */
-    public sendKeyBackup(roomId: string, sessionId: string, version: string, data: IKeyBackup): Promise<void> {
+    public sendKeyBackup(
+        roomId: string | undefined,
+        sessionId: string | undefined,
+        version: string,
+        data: IKeyBackup,
+    ): Promise<void> {
```
t3chguy, line 2697: Same as above

src/client.ts
```diff
@@ -2773,8 +2782,8 @@ export class MatrixClient extends EventEmitter {
      */
     public async restoreKeyBackupWithPassword(
         password: string,
-        targetRoomId: string,
-        targetSessionId: string,
+        targetRoomId: string | undefined,
+        targetSessionId: string | undefined,
```
t3chguy, line 2786: Same as above

src/client.ts
```diff
@@ -2835,8 +2844,8 @@ export class MatrixClient extends EventEmitter {
      */
     public restoreKeyBackupWithRecoveryKey(
         recoveryKey: string,
-        targetRoomId: string,
-        targetSessionId: string,
+        targetRoomId: string | undefined,
+        targetSessionId: string | undefined,
```
t3chguy, line 2848: Same as above

src/client.ts
```diff
@@ -2845,8 +2854,8 @@ export class MatrixClient extends EventEmitter {
     }
 
     public async restoreKeyBackupWithCache(
-        targetRoomId: string,
-        targetSessionId: string,
+        targetRoomId: string | undefined,
+        targetSessionId: string | undefined,
```
t3chguy, line 2858: Same as above

src/client.ts
```diff
@@ -2859,8 +2868,8 @@ export class MatrixClient extends EventEmitter {
 
     private async restoreKeyBackup(
         privKey: ArrayLike<number>,
-        targetRoomId: string,
-        targetSessionId: string,
+        targetRoomId: string | undefined,
+        targetSessionId: string | undefined,
```
t3chguy, line 2872: Same as above

src/client.ts
```diff
@@ -2953,7 +2962,11 @@ export class MatrixClient extends EventEmitter {
         return { total: totalKeyCount, imported: keys.length };
     }
 
-    public deleteKeysFromBackup(roomId: string, sessionId: string, version: string): Promise<void> {
+    public deleteKeysFromBackup(
+        roomId: string | undefined,
+        sessionId: string | undefined,
+        version: string,
+    ): Promise<void> {
```
t3chguy, line 2969: Same as above

src/client.ts
hughns: Hopefully these are all now as expected?

t3chguy: Yup, shame that the optional params were put first :(

t3chguy (approved):

PR #2094, opened 2022-01-07
Don't consider alt_aliases when calculating room name

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/13887

---

## ✨ Features
 * Don't consider alt_aliases when calculating room name ([\#2094](https://github.com/matrix-org/matrix-js-sdk/pull/2094)). Fixes vector-im/element-web#13887.

### DIFF
--- a/spec/unit/room.spec.js
+++ b/spec/unit/room.spec.js
@@ -622,6 +622,11 @@ describe("Room", function() {
                 }, event: true,
             })]);
         };
+        const setAlias = function(alias) {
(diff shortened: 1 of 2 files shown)


turt2live (commented):
lgtm, but the tests are unhappy. I think this is actually tested code from memory, so would expect the test failure to be "legitimate" and need fixing.

daniellekirkwood (approved):
This is great, thank you!

turt2live (approved):

PR #2099, opened 2022-01-10
Use session ids to resolve refresh during invite/answer

robertlong:
This PR is a fix for #1902 that resolves split-brains when a client refreshes before they have answered an incoming call.

---


turt2live (approved):
it is a session ID.

Given this is going to a custom branch, approved.

PR #2093, opened 2022-01-06
Fetch server capabilities during client initialisation

gsouquet:
For https://github.com/vector-im/element-web/issues/20408

---
This PR currently has no changelog labels, so will not be included in changelogs.

Add one of: `T-Deprecation`, `T-Enhancement`, `T-Defect`, `T-Task` to indicate what type of change this is plus `X-Breaking-Change` if it's a breaking change.

PR #2100, opened 2022-01-11
Remove getCapabilities call for guest users

gsouquet:
Regressed by https://github.com/matrix-org/matrix-js-sdk/pull/2093

Capabilities requires authentication and therefore can not be called for guests, https://spec.matrix.org/v1.1/client-server-api/#capabilities-negotiation

---


jryans (approved):

t3chguy:
I think this was the wrong thing to do, caps work for guests

image

PR #2109, opened 2022-01-18
Support m.asset in m.location event content

andybalaam:
See https://github.com/matrix-org/matrix-doc/blob/matthew/location/proposals/3488-location.md - we have a new requirement to support a `m.asset` property in location event content.

---

## ✨ Features
 * Support m.asset in m.location event content ([\#2109](https://github.com/matrix-org/matrix-js-sdk/pull/2109)). Contributed by @andybalaam.


t3chguy (approved):
Looks sane

Issue #1893, opened 2021-09-08
TS Conversion: browser-index, errors


dbkr:
Misc little TS conversions: src/browser-index.js and src/errors.js

ryushar:
Hi,
I was planning to take on the task of converting src/error.js to TS. While examining the file, I found that there were some constructor functions which were using Reflect to kind of create extended Error classes.

The reason for not using ES6 classes was for backward compatibility ([as written here](https://github.com/matrix-org/matrix-js-sdk/blob/230e3b4acee3f0802e2b36fca2dd9f1e81c27a19/src/errors.js#L1)) but I don't think there is a proper way to Typescriptify these constructor functions. Also, in the Contribution.md file, [Code Style section](https://github.com/matrix-org/matrix-js-sdk/blob/230e3b4acee3f0802e2b36fca2dd9f1e81c27a19/src/errors.js#L1), it is mentioned that this project aims to target ES6.

So, my question is should I convert these constructor functions to ES6 classes? In my opinion, the code gets alot more easier to understand this way. Also, the support of Reflect and ES6 classes came almost at a similar time in most of the browsers (Chrome added both of them together in v49).

PR #2111, opened 2022-01-22
Typescriptify src/errors.js

ryushar:
Partially addresses #1893 

Signed-off-by: Tushar <tusharxdev@gmail.com>

Notes: none


src/errors.ts
```diff
@@ -0,0 +1,34 @@
+export class InvalidStoreError extends Error {
```
SimonBrandner, line 1: We should probably have a licences here :thinking:

SimonBrandner: _resolved_ - thanks!

ryushar: Forgot about the license since it wasn't there previously, should be good now~

src/errors.ts
```diff
@@ -0,0 +1,50 @@
+/*
+Copyright 2015-2022 The Matrix.org Foundation C.I.C.
```
turt2live, line 2: We can't guarantee that it was 2015 - 2022 because the previous file didn't have any headers. If you're assigning your copyright to the foundation, it'd just be 2022.

src/errors.ts
```diff
@@ -0,0 +1,50 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+export class InvalidStoreError extends Error {
+    static TOGGLED_LAZY_LOADING = "TOGGLED_LAZY_LOADING";
+    reason: string;
+    value: unknown;
```
turt2live, line 20: explicit member access (`public` / `private`) please

src/errors.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    constructor(reason: string, value: unknown) {
+        const message = `Store is invalid because ${reason}, `
+            + `please stop the client, delete all data and start the client again`;
+        super(message);
+        this.reason = reason;
+        this.value = value;
```
turt2live, line 27: we should use the shortcuts in the constructor rather than this

src/errors.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+}
+
+export class InvalidCryptoStoreError extends Error {
+    static TOO_NEW = "TOO_NEW";
+    reason: string;
```
turt2live, line 33: explicit member access

src/errors.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    constructor(reason: string) {
+        const message = `Crypto store is invalid because ${reason}, ` +
+            `please stop the client, delete all data and start the client again`;
+        super(message);
+        this.reason = reason;
```
turt2live, line 39: constructor shortcuts

src/errors.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+}
+
+export class KeySignatureUploadError extends Error {
+    value: unknown;
+
+    constructor(message: string, value: unknown) {
+        super(message);
+        this.value = value;
```
turt2live, line 48: constructor shortcuts

src/sync.ts
```diff
@@ -602,7 +602,7 @@ export class SyncApi {
             if (shouldClear) {
                 this.storeIsInvalid = true;
                 const reason = InvalidStoreError.TOGGLED_LAZY_LOADING;
-                const error = new InvalidStoreError(reason, !!this.opts.lazyLoadMembers);
+                const error = new InvalidStoreError(reason, !!this.opts.lazyLoadMembers) as any;
```
turt2live, line 605: need a comment to explain why we're casting to `any`, or better yet: fix the error that prevents the casting from working.

src/errors.ts
turt2live, line 20: this and the `value` in a later definition are `unknown` - they should be typed.

turt2live (changes requested):
thank you! Just a few code styling issues to resolve before we merge this.

src/errors.ts
ryushar: Can't do that as there seems to be a bug in typescript which isn't allowing even statements like variable assignment (`const message = ...`) before the `super()` call.
image
[Bug reference](https://github.com/microsoft/TypeScript/issues/3311)

t3chguy: Why not inline the variable into the super call then?

ryushar: Done, although imo this way makes it more unclear, since the parameter isn't a small simple string.

Also, shouldn't these rules be enforced using eslint? Else these small coding style issues will keep on appearing in future pull requests too.

src/sync.ts
```diff
@@ -602,6 +602,9 @@ export class SyncApi {
             if (shouldClear) {
                 this.storeIsInvalid = true;
                 const reason = InvalidStoreError.TOGGLED_LAZY_LOADING;
+                // TODO - InvalidStoreError, or rather all the custom error classes, should extend
+                // the MatrixError class. For a quickfix, `error` has been casted to `any` because `updateSyncState` requires
+                // it to be an instance of MatrixError class.
```
turt2live, line 607: This is a sign that we should extend the class now, rather than assume the types will work. The function wants a specific contract to be met, and we're intentionally not providing that :(

src/errors.ts
```diff
@@ -0,0 +1,41 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { IUploadKeySignaturesResponse } from "./client";
+
+export class InvalidStoreError extends Error {
+    public static TOGGLED_LAZY_LOADING = "TOGGLED_LAZY_LOADING";
```
turt2live, line 20:
```suggestion
    public static readonly TOGGLED_LAZY_LOADING = "TOGGLED_LAZY_LOADING";
```

src/errors.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    constructor(public reason: string, public value: boolean) {
+        super(`Store is invalid because ${reason}, `
+        + `please stop the client, delete all data and start the client again`);
+    }
+}
+
+export class InvalidCryptoStoreError extends Error {
+    public static TOO_NEW = "TOO_NEW";
```
turt2live, line 29:
```suggestion
    public static readonly TOO_NEW = "TOO_NEW";
```

turt2live (changes requested):
almost there - just a couple other changes to fix. 

Apologies for the slow turnaround time on this - it's hard to get review time, sometimes.

src/sync.ts
ryushar: That would require code changes in multiple other files, and this pull request is meant for only converting error.js to Typescript, which is a cosmetic change as it doesn't change the working of the original code. Imho, this should be done in a separate pull request.

turt2live: Conversions to typescript can and should affect other files: the point is to fix bugs, not check boxes :)

turt2live (changes requested):
existing review still stands

t3chguy:
This happened

PR #2129, opened 2022-01-24
Make threads reply chain filter out local event ID

gsouquet:
Fixes https://github.com/vector-im/element-web/issues/20644

---

### DIFF
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -138,10 +138,13 @@ export class Thread extends TypedEventEmitter<ThreadEvent> {
     /**
      * Return last reply to the thread
      */
-    public get lastReply(): MatrixEvent {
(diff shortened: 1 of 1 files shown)


t3chguy:
A public signature change which is technically breaking - also where's the counterpart react-sdk PR?

turt2live (approved):
This seems fine for two reasons:

1. The whole class is flagged as experimental, so breaking signature changes are okay.
2. A default is set.

Still curious as to why it's needed without a counterpart PR which uses it though...

gsouquet:
> Still curious as to why it's needed without a counterpart PR which uses it though...

I forgot to push the counterpart PR, but it's up right here https://github.com/matrix-org/matrix-react-sdk/pull/7620

PR #2132, opened 2022-01-26
Improved threads reliability with/without server side support

gsouquet:
Fixes https://github.com/vector-im/element-web/issues/20708
Fixes https://github.com/vector-im/element-web/issues/20709
Fixes https://github.com/vector-im/element-web/issues/20607
Fixes https://github.com/vector-im/element-web/issues/20605
Fixes https://github.com/vector-im/element-web/issues/20633

- Make threadRootId work for relation of thread relations
- Avoid race conditions creating multiple thread of the same type
- Improve typing for server side aggregated relations
- Use bundled relationships for thread reply count
- Count all events when threads has not server support'
- Improved thread root detection using server side aggregated relations

The tackles three main things that were the root of a lot of recent bugs we have seen in threads

- Better leverages bundled relationships way more than I used to which has a 100% accurate reply count when server side support is enabled.
- Multiple threads model were created for the same thread root during the initial sync due to some race condition. Thread creation used to be synchronous but that can not be the case anymore
- Some `m.replace` and `m.room.redaction` were not partitioned correctly

---

### DIFF
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -16,7 +16,8 @@ limitations under the License.
 
 import { MatrixClient } from "../matrix";
 import { ReEmitter } from "../ReEmitter";
-import { MatrixEvent } from "./event";
+import { RelationType } from "../@types/event";
+import { MatrixEvent, IThreadBundledRelationship } from "./event";
 import { EventTimeline } from "./event-timeline";
 import { EventTimelineSet } from './event-timeline-set';
 import { Room } from './room';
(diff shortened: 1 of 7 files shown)


src/models/thread.ts
```diff
@@ -47,6 +48,9 @@ export class Thread extends TypedEventEmitter<ThreadEvent> {
 
     private reEmitter: ReEmitter;
 
+    private _lastReply: MatrixEvent;
+    private _replyCount = 0;
```
jryans, line 52: I believe `_` prefix is discouraged for our TS files (unless there's a getter with the same name), since `private` should be enough.

```suggestion
    private lastReply: MatrixEvent;
    private replyCount = 0;
```

jryans (approved):
Looks good overall, thanks! 😄

src/models/thread.ts
t3chguy, line 52: Any need to underscore these private fields?

t3chguy (approved):
LGTM otherwise

Issue #2124, opened 2022-01-23
Maybe enable `strict-boolean-expressions`


ShadowJonathan:
https://typescript-eslint.io/rules/strict-boolean-expressions

Together with https://github.com/matrix-org/matrix-js-sdk/issues/2120 and https://github.com/matrix-org/matrix-js-sdk/issues/2121, this'd do with all truthiness/falsiness checks, and requires these to be explicitly checked by themselves.

ShadowJonathan:
This would remove truthy/falsy expressions, such as;

```ts
let num: number | undefined = 0;
if (num) {
  console.log('num is defined');
}
```

These assertions may be difficult to follow, and also catch a lot more than just nullability; the above branch would never pass, `0` is falsy.

Instead, this lint would enforce the following;

```ts
let num: number | undefined = 0;
if (num !== undefined) {  // or `!= null`, as `undefined == null` 
  console.log('num is defined');
}
```

This would enable a ton more correctness in the code, and help subtle edgecases around falsifiable non-undefined values.

PR #2136, opened 2022-01-26
Assume per-user deviceID uniqueness in encryptAndSendKeysToDevices

### LINKED ISSUE #2135: `encryptAndSendKeysToDevices` assumes device IDs are globally unique
`encryptAndSendKeysToDevices` uses two different mappings for devices and user+devices;

```ts
const contentMap: Record<string, Record<string, IEncryptedContent>> = {};
const deviceInfoByDeviceId = new Map<string, DeviceInfo>();
```

These two are filled independently, with the latter being queried in this spot at the end of the function;

```ts
 return this.baseApis.sendToDevice("m.room.encrypted", contentMap).then(() => {
    // store that we successfully uploaded the keys of the current slice
    for (const userId of Object.keys(contentMap)) {
        for (const deviceId of Object.keys(contentMap[userId])) {
            session.markSharedWithDevice(
                userId,
                deviceId,
                deviceInfoByDeviceId.get(deviceId).getIdentityKey(),
                chainIndex,
            );
        }
    }
});
```

This assumes that no two users share a device with the same deviceID, though that is possible given a large enough room and a large enough time for users to keep generating deviceIDs.

AFAIK, this does not expose a security vulnerability, worst this could do is junk up the message ratchet for either of the two users' specific devices in that particular room, making messages undecryptable.

ShadowJonathan:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/2135

Notes: Make sure `encryptAndSendKeysToDevices` assumes devices are unique per-user.

Type: defect

`Signed-off-by: Jonathan de Jong <jonathan@automatia.nl>`

---

## 🐛 Bug Fixes
 * Make sure `encryptAndSendKeysToDevices` assumes devices are unique per-user. ([\#2136](https://github.com/matrix-org/matrix-js-sdk/pull/2136)). Fixes #2135. Contributed by @ShadowJonathan.


src/crypto/algorithms/megolm.ts
```diff
@@ -591,8 +591,9 @@ class MegolmEncryption extends EncryptionAlgorithm {
         userDeviceMap: IOlmDevice[],
         payload: IPayload,
     ): Promise<void> {
-        const contentMap: Record<string, Record<string, IEncryptedContent>> = {};
-        const deviceInfoByDeviceId = new Map<string, DeviceInfo>();
+        const userDeviceMapping: Record<string, Record<string, {
+            encryptedContent: IEncryptedContent; deviceInfo: DeviceInfo;
+        }>> = {};
```
t3chguy, line 596: Could we use the alternate format which allows identifiers?

```suggestion
        const userDeviceMapping: {
            [userId: string]: {
                [deviceId: string]: {
                    encryptedContent: IEncryptedContent;
                    deviceInfo: DeviceInfo;
                };
            };
        } = {};
```

src/crypto/algorithms/megolm.ts
```diff
@@ -605,12 +606,11 @@ class MegolmEncryption extends EncryptionAlgorithm {
             const userId = val.userId;
             const deviceInfo = val.deviceInfo;
             const deviceId = deviceInfo.deviceId;
-            deviceInfoByDeviceId.set(deviceId, deviceInfo);
 
-            if (!contentMap[userId]) {
-                contentMap[userId] = {};
+            if (userDeviceMapping[userId] == null) {
```
t3chguy, line 610: relying on type coercion due to `==` is meh, it'd be `undefined` if anything but also can just be treated as falsey

ShadowJonathan: I personally prefer testing for nullish over falsey explicitly, also see #2124 where I got this from.

However, if this still isn't alright, ill just change it to falsey checking.

t3chguy: I think conflating this otherwise useful change with a slightly controversial typing nit would be a waste, it'd just bikeshed it a little bit

ShadowJonathan: Alright, I'll change it to falsey then :+1:

src/crypto/algorithms/megolm.ts
```diff
@@ -630,37 +635,47 @@ class MegolmEncryption extends EncryptionAlgorithm {
...
 
+            const contentMap: Record<string, Record<string, IEncryptedContent>> = {};
+
+            for (const userId of Object.keys(userDeviceMapping)) {
+                contentMap[userId] = {};
+
+                for (const deviceId of Object.keys(userDeviceMapping[userId])) {
+                    contentMap[userId][deviceId] = userDeviceMapping[userId][deviceId].encryptedContent;
+                }
+            }
```
t3chguy, line 669: merging an object and a map only just to reconstruct that object again seems a little silly.
Whilst yes this PR fixes the bug, it makes the code harder to follow IMO.

ShadowJonathan: I'm not sure what to do here, as keeping the deviceInfo and encryptedContent together is also more correct, as it ensures one does not exist without the other.

This object mapping (of sorts) is needed for the sendToDevice dispatch.

turt2live (commented):
ideally we try and construct a better data type, per comments already made.

src/crypto/algorithms/megolm.ts
dkasak: It's unfortunate that we have to construct this mapping ourselves here at all. Do we not have a method somewhere which returns a `user_id -> device_id -> DeviceInfo` map given a list of user IDs?

In the absence of such a mapping, I don't really see a particular advantage in grouping these two things together. Can't we simply do something like https://github.com/dkasak/matrix-js-sdk/commit/ea6bd5dcb8be48f3024f9ad3f1e44072ea542b92 instead?

dkasak (commented):
This PR looks functionally correct and will fix the issue, but see my PR comments.

To expand on the reasoning from #2135, I agree this bug shouldn't have security consequences, though the reasoning is more subtle; I don't think this would lead to a ratchet corruption and undecryptable messages.

Rather, it allows a malicious server to confuse Element Web into recording the wrong device identity key for re-sharing (forwarding) purposes. However it can only confuse it into recording the key of another device *present in the room at the time the message was sent*, so that device would've received the message and the message key anyway, even without this trick.

src/crypto/algorithms/megolm.ts
ShadowJonathan: I think that https://github.com/dkasak/matrix-js-sdk/commit/ea6bd5dcb8be48f3024f9ad3f1e44072ea542b92 could instead be a better solution for this, could I force-push my PR to only include that commit instead?

ShadowJonathan:
Reviewers, I have reset this PR to @dkasak's [suggested commit](https://github.com/dkasak/matrix-js-sdk/commit/ea6bd5dcb8be48f3024f9ad3f1e44072ea542b92), as I believe that is a better approach. Approval got granted via DM.

t3chguy:
(Sonar didn't run due to missing sonar-project.properties file in this old PR)

ShadowJonathan:
Sonar now ants 80% coverage on this new code, while I'm not sure how to achieve that. Any suggestions? Adding specific tests for this?

src/crypto/algorithms/megolm.ts
```diff
@@ -592,7 +592,7 @@ class MegolmEncryption extends EncryptionAlgorithm {
         payload: IPayload,
     ): Promise<void> {
         const contentMap: Record<string, Record<string, IEncryptedContent>> = {};
-        const deviceInfoByDeviceId = new Map<string, DeviceInfo>();
+        const deviceInfoByDeviceId = new Map<string, Map<string, DeviceInfo>>();
```
MadLittleMods, line 595: This seems more accurate now:

```suggestion
        const deviceInfoByUserIdAndDeviceId = new Map<string, Map<string, DeviceInfo>>();
```

src/crypto/algorithms/megolm.ts
```diff
@@ -592,7 +592,7 @@ class MegolmEncryption extends EncryptionAlgorithm {
         payload: IPayload,
```
MadLittleMods, line 592: > Sonar now ants 80% coverage on this new code, while I'm not sure how to achieve that. Any suggestions? Adding specific tests for this?

It would be good to add a `encryptAndSendKeysToDevices` test which hits on this edge case you're trying to solve:

> This assumes that no two users share a device with the same deviceID, though that is possible given a large enough room and a large enough time for users to keep generating deviceIDs.
>
> *-- https://github.com/matrix-org/matrix-js-sdk/issues/2135#issue-1115248936*

ShadowJonathan: Sonar seems to be okay with the new lines, i had this problem before in another PR, i merged `develop` into here again to use the new rules.

PR #2142, opened 2022-01-28
Refactor thread model to be created from the root event

gsouquet:
To review alongside https://github.com/matrix-org/matrix-react-sdk/pull/7670

- Refactor thread model to be created from the root event
- Integrate threads with the /relations API

---

### DIFF
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -135,7 +135,7 @@ export class Thread extends TypedEventEmitter<ThreadEvent> {
         }
 
         if (this.hasServerSideSupport && this.initialEventsFetched) {
-            if (event.getTs() > this.lastReply().getTs() && !this.findEventById(event.getId())) {
+            if (event.localTimestamp > this.lastReply().localTimestamp && !this.findEventById(event.getId())) {
                 this.timelineSet.addEventToTimeline(
                     event,
                     this.liveTimeline,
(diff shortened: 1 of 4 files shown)


src/models/thread.ts
```diff
@@ -91,85 +99,111 @@ export class Thread extends TypedEventEmitter<ThreadEvent> {
...
     /**
      * Add an event to the thread and updates
      * the tail/root references if needed
      * Will fire "Thread.update"
      * @param event The event to add
      */
     public async addEvent(event: MatrixEvent, toStartOfTimeline = false): Promise<void> {
-        if (this.timelineSet.findEventById(event.getId())) {
-            return;
+        // When there's no server side support we man
```
t3chguy, line 113: we man what?

src/models/thread.ts
```diff
@@ -91,85 +99,111 @@ export class Thread extends TypedEventEmitter<ThreadEvent> {
...
+            await this.client.decryptEventIfNeeded(event, {});
         }
 
-        if (!this.root) {
-            if (event.isThreadRelation) {
-                this.root = event.threadRootId;
-            } else {
-                this.root = event.getId();
+        if (this.hasServerSideSupport && this.initialEventsFetched) {
+            if (event.getTs() > this.lastReply().getTs() && !this.findEventById(event.getId())) {
```
t3chguy, line 137: relying on timestamps between events for ordering is a no-go given different servers will have different clocks

src/models/thread.ts
```diff
@@ -91,85 +99,111 @@ export class Thread extends TypedEventEmitter<ThreadEvent> {
...
-
         const isThreadReply = event.getRelation()?.rel_type === RelationType.Thread;
         // If no thread support exists we want to count all thread relation
         // added as a reply. We can't rely on the bundled relationships count
         if (!this.hasServerSideSupport && isThreadReply) {
             this.replyCount++;
         }
 
-        if (!this.lastEvent || (isThreadReply && event.getTs() > this.lastEvent.getTs())) {
+        if (!this.lastEvent || (isThreadReply && event.getTs() > this.replyToEvent.getTs())) {
```
t3chguy, line 159: same as above

src/models/thread.ts
gsouquet: What would be the alternative? How can I know whether a message is newly incoming?

gsouquet: I'm planning to merge this implementation using the `age`
It has cases in which is could cause the count to drift compared to the value returned by the HS. But this is never going to happen pre-release as we're going to use two homeservers with relatively synced clocks.
I'll create an issue to revisit that before the launch and will see whether we can find a better approach

t3chguy (approved):

PR #2147, opened 2022-02-01
Add thread relation to sticker message

gsouquet:
Fixes https://github.com/vector-im/element-web/issues/20641

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -4212,6 +4212,18 @@ export class MatrixClient extends EventEmitter {
             info: info,
             body: text,
         };
+
+        const thread = this.getRoom(roomId)?.threads.get(threadId);
+        if (thread) {
+            content["m.relates_to"] = {
+                "rel_type": RelationType.Thread,
+                "event_id": threadId,
+                "m.in_reply_to": {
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -4212,6 +4212,18 @@ export class MatrixClient extends EventEmitter {
...
+
+        const thread = this.getRoom(roomId)?.threads.get(threadId);
+        if (thread) {
+            content["m.relates_to"] = {
+                "rel_type": RelationType.Thread,
+                "event_id": threadId,
+                "m.in_reply_to": {
+                    "event_id": thread.replyToEvent.getId(),
+                },
+            };
```
t3chguy, line 4224: kind of surprised there isn't a helper to generate this given a `thread`

t3chguy (approved):
LGTM

src/client.ts
gsouquet: This is done in the composer for all other event types. But stickers work slightly differently and it constructs the event in the "send" method.
If that starts being used elsewhere in the SDK I will create a helper to avoid duplication

PR #2149, opened 2022-02-02
Add thread relation in sendEvent if it's missing

gsouquet:
---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -3593,6 +3593,19 @@ export class MatrixClient extends EventEmitter {
             eventType = threadId;
             threadId = null;
         }
+
+        if (threadId && content["m.relates_to"]?.rel_type !== RelationType.Thread) {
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -3593,6 +3593,19 @@ export class MatrixClient extends EventEmitter {
...
         }
+
+        if (threadId && content["m.relates_to"]?.rel_type !== RelationType.Thread) {
+            const thread = this.getRoom(roomId)?.threads.get(threadId);
+            content["m.relates_to"] = {
+                ...content["m.relates_to"],
+                "rel_type": RelationType.Thread,
+                "event_id": threadId,
+                "m.in_reply_to": {
+                    "event_id": thread.replyToEvent.getId(),
```
t3chguy, line 3604: `thread` may be null given you're null-ish coalescing on the getRoom call

t3chguy (approved):

PR #2153, opened 2022-02-04
Out of sync thread summary in main timeline

gsouquet:
Fixes https://github.com/vector-im/element-web/issues/20873

- Fix localTimestamp for local event
- Initialise thread after determining the root ID
- Re-use existing MatrixEvent if it's already known locally

---

### DIFF
--- a/src/event-mapper.ts
+++ b/src/event-mapper.ts
@@ -38,7 +38,9 @@ export function eventMapperFor(client: MatrixClient, options: MapperOpts): Event
             event = room.findEventById(plainOldJsObject.event_id);
         }
 
-        if (!event) {
+        // If no event is found or if the event found was only local we can
+        // safely create a new model
+        if (!event || event.status) {
             event = new MatrixEvent(plainOldJsObject);
         }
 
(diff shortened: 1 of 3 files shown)


src/event-mapper.ts
```diff
@@ -29,7 +29,19 @@ export function eventMapperFor(client: MatrixClient, options: MapperOpts): Event
     const decrypt = options.decrypt !== false;
 
     function mapper(plainOldJsObject: Partial<IEvent>) {
-        const event = new MatrixEvent(plainOldJsObject);
+        const room = client.getRoom(plainOldJsObject.room_id);
+        let event;
```
t3chguy, line 33:
```suggestion
        let event: MatrixEvent;
```

src/models/event.ts
```diff
@@ -584,9 +584,10 @@ export class MatrixEvent extends EventEmitter {
      * Get the age of this event. This represents the age of the event when the
      * event arrived at the device, and not the age of the event when this
      * function was called.
+     * Can only be returned once the server has echo'ed back
      * @return {Number} The age of this event in milliseconds.
```
t3chguy, line 588: type needs updating

t3chguy (approved):
LGTM otherwise

PR #2155, opened 2022-02-04
Fix error in uploadContent() when file is empty under Node.js

hughns:
---

## 🐛 Bug Fixes
 * Fix error in uploadContent() when file is empty under Node.js ([\#2155](https://github.com/matrix-org/matrix-js-sdk/pull/2155)). Contributed by @hughns.


gsouquet (approved):
Looks good to me 💯

PR #2156, opened 2022-02-04
Send references to thread root to threads, even out of order

andybalaam:
Helps a bit with https://github.com/vector-im/element-web/issues/20282

---

## 🐛 Bug Fixes
 * Send references to thread root to threads, even out of order ([\#2156](https://github.com/matrix-org/matrix-js-sdk/pull/2156)). Contributed by @andybalaam.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -9105,7 +9105,7 @@ export class MatrixClient extends EventEmitter {
                         const copiedEvent = event.toSnapshot();
 
                         // The copied event is in this thread:
-                        copiedEvent.setThreadId(parentEvent.getId());
+                        copiedEvent.setThreadId(parentEventId);
                         memo[THREAD].push(copiedEvent);
                     } else if (parentEvent?.isThreadRelation) {
                         // If our parent is in a thread, we are in that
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -9065,36 +9065,53 @@ export class MatrixClient extends EventEmitter {
...
+
                     if (targetingThreadRoot && !event.isThreadRelation && event.relationEventId) {
-                        memo[THREAD].push(event);
+                        // If we refer to the thread root, we should be copied
+                        // into the thread as well as the main timeline.
+                        // This happens for reactions, annotations, poll votes etc.
+                        const copiedEvent = event.toSnapshot();
+
+                        // The copied event is in this thread:
+                        copiedEvent.setThreadId(parentEvent.getId());
```
gsouquet, line 9108: It is possible that `parentEvent` is not loaded. It's probably fine to use `parentEventId` here

```suggestion
                        copiedEvent.setThreadId(parentEventId);
```

andybalaam:
Converted to draft since I can't actually validate that this fixes the problems any more :-(

andybalaam:
We have concluded that this makes the situation better, even though it doesn't fix everything, so undrafting.

gsouquet (approved):

PR #2159, opened 2022-02-07
Account for encryption in `maySendMessage()`

SimonBrandner:
Type: defect

---

## 🐛 Bug Fixes
 * Account for encryption in `maySendMessage()` ([\#2159](https://github.com/matrix-org/matrix-js-sdk/pull/2159)). Contributed by @SimonBrandner.

### DIFF
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -2199,9 +2199,9 @@ export class Room extends EventEmitter {
      *                   message events into the room.
      */
     public maySendMessage(): boolean {
-        return this.getMyMembership() === 'join' && this.client.isRoomEncrypted(this.roomId)
+        return this.getMyMembership() === 'join' && (this.client.isRoomEncrypted(this.roomId)
             ? this.currentState.maySendEvent(EventType.RoomMessageEncrypted, this.myUserId)
-            : this.currentState.maySendEvent(EventType.RoomMessage, this.myUserId);
+            : this.currentState.maySendEvent(EventType.RoomMessage, this.myUserId));
     }
 
     /**
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -2199,8 +2199,9 @@ export class Room extends EventEmitter {
      *                   message events into the room.
      */
     public maySendMessage(): boolean {
-        return this.getMyMembership() === 'join' &&
-            this.currentState.maySendEvent(EventType.RoomMessage, this.myUserId);
+        return this.getMyMembership() === 'join' && this.client.isRoomEncrypted(this.roomId)
+            ? this.currentState.maySendEvent(EventType.RoomMessageEncrypted, this.myUserId)
+            : this.currentState.maySendEvent(EventType.RoomMessage, this.myUserId);
     }
 
     /**
--- a/spec/unit/room.spec.js
+++ b/spec/unit/room.spec.js
@@ -1461,7 +1461,7 @@ describe("Room", function() {
     describe("maySendMessage", function() {
         it("should return false if synced membership not join",
         function() {
-            const room = new Room(roomId, null, userA);
+            const room = new Room(roomId, { isRoomEncrypted: () => false }, userA);
             room.updateMyMembership("invite");
             expect(room.maySendMessage()).toEqual(false);
             room.updateMyMembership("leave");


src/models/room.ts
```diff
@@ -2199,8 +2199,9 @@ export class Room extends EventEmitter {
      *                   message events into the room.
      */
     public maySendMessage(): boolean {
-        return this.getMyMembership() === 'join' &&
-            this.currentState.maySendEvent(EventType.RoomMessage, this.myUserId);
+        return this.getMyMembership() === 'join' && this.client.isRoomEncrypted(this.roomId)
```
t3chguy, line 2202: should also check if the client has encryption enabled, otherwise the messages sent won't be encrypted and this check would be wrong

src/models/room.ts
```diff
@@ -2199,9 +2199,15 @@ export class Room extends EventEmitter {
...
     public maySendMessage(): boolean {
-        return this.getMyMembership() === 'join' && this.client.isRoomEncrypted(this.roomId)
-            ? this.currentState.maySendEvent(EventType.RoomMessageEncrypted, this.myUserId)
-            : this.currentState.maySendEvent(EventType.RoomMessage, this.myUserId);
+        if (this.getMyMembership() !== "join") return false;
+
+        if (this.client.isRoomEncrypted(this.roomId)) {
+            return this.client.crypto
+                ? this.currentState.maySendEvent(EventType.RoomMessageEncrypted, this.myUserId)
+                : this.client.usingExternalCrypto;
```
t3chguy, line 2207: I don't think this is right, should probably be `(this.client.crypto || this.client.usingExternalCrypto) && this.currentState...`

SimonBrandner: https://github.com/matrix-org/matrix-js-sdk/blob/2d23330b74daad730c9122f5ee09391c6caf653e/src/client.ts#L3785 seems to suggest we don't need to encrypt if `this.client.usingExternalCrypto` is truthy and therefore I assume that the event type wouldn't be `m.room.encrypted`?

t3chguy: That is correct, the external crypto service would do the encryption instead of us.

SimonBrandner: So, it should actually be something like this? What am I missing?

```
return this.client.crypto
	? this.currentState.maySendEvent(EventType.RoomMessageEncrypted, this.myUserId)
	: this.client.usingExternalCrypto && this.currentState.maySendEvent(EventType.RoomMessage, this.myUserId);
```

t3chguy: No, if `usingExternalCrypto` then we (js-sdk) send `m.room.message` but before it hits the HS will be encrypted (so turned into `m.room.encrypted`)

SimonBrandner: I agree, is my code doing something else? I am probably overlooking something obvious, sorry

t3chguy: Your code will say you always have no permission if `!this.client.crypto` && `!this.client. usingExternalCrypto`

SimonBrandner: And when should this return `true`?

t3chguy:
```
return (this.client.crypto || this.client.usingExternalCrypto)
	? this.currentState.maySendEvent(EventType.RoomMessageEncrypted, this.myUserId)
	:  this.currentState.maySendEvent(EventType.RoomMessage, this.myUserId);
```

I think it should be this tbh

t3chguy: well if the room isn't encrypted then we need to check `m.room.message` permissions

SimonBrandner: Argh, I was looking at the old version of your comment, all makes sense now :D

SimonBrandner: Though what if don't have crypto (nor external crypto) and the room is encrypted, then we shouldn't return true even if we can send `m.room.message`

src/models/room.ts
```diff
@@ -2199,8 +2199,14 @@ export class Room extends EventEmitter {
      *                   message events into the room.
      */
     public maySendMessage(): boolean {
-        return this.getMyMembership() === 'join' &&
-            this.currentState.maySendEvent(EventType.RoomMessage, this.myUserId);
+        if (this.getMyMembership() !== "join") return false;
+
+        if (this.client.isRoomEncrypted(this.roomId)) {
+            return (this.client.crypto || this.client.usingExternalCrypto) &&
```
t3chguy, line 2205: I actually think `(this.client.crypto || this.client.usingExternalCrypto)` may be redundant and you were right initially, as js-sdk won't let you send unencrypted content in encrypted rooms (unless usingExternalCrypto but that's special because we trust the crypto service to encrypt for us)

So I think reverting back to your first solution is sane.

t3chguy (approved):
Looks sane to me! thanks

PR #2165, opened 2022-02-08
Improve thread partitioning for 2nd degree relations

gsouquet:
Fixes https://github.com/vector-im/element-web/issues/20966
Fixes https://github.com/vector-im/element-web/issues/20915
Fixes https://github.com/vector-im/element-web/issues/20921
Fixes https://github.com/vector-im/element-web/issues/20282
Fixes https://github.com/vector-im/element-web/issues/20848

- Cater for 2-nd degree relations in thread partitionning
- Fix event edits in a thread

---

### DIFF
--- a/spec/integ/matrix-client-methods.spec.js
+++ b/spec/integ/matrix-client-methods.spec.js
@@ -396,6 +396,236 @@ describe("MatrixClient", function() {
     });
 
     describe("partitionThreadedEvents", function() {
+
+        beforeEach(() => {
+            eventMessageInThread = new MatrixEvent({
+                "age": 80098509,
+                "content": {
+                    "algorithm": "m.megolm.v1.aes-sha2",
+                    "ciphertext": "ENCRYPTEDSTUFF",
+                    "device_id": "XISFUZSKHH",
+                    "m.relates_to": {
+                        "event_id": "$VLS2ojbPmxb6x8ECetn45hmND6cRDcjgv-j-to9m7Vo",
+                        "m.in_reply_to": {
+                            "event_id": "$VLS2ojbPmxb6x8ECetn45hmND6cRDcjgv-j-to9m7Vo",
+                        },
+                        "rel_type": "io.element.thread",
+                    },
+                    "sender_key": "i3N3CtG/CD2bGB8rA9fW6adLYSDvlUhf2iuU73L65Vg",
+                    "session_id": "Ja11R/KG6ua0wdk8zAzognrxjio1Gm/RK2Gn6lFL804",
+                },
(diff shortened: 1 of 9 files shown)


spec/integ/matrix-client-methods.spec.js
```diff
@@ -396,6 +395,236 @@ describe("MatrixClient", function() {
     });
 
     describe("partitionThreadedEvents", function() {
+
+        beforeEach(() => {
```
andybalaam, line 398: I really hate beforeEach etc. - could we turn each variable into a function that returns a MatrixEvent instead?

(Reasoning: https://www.artificialworlds.net/blog/2013/03/27/setup-and-teardown-considered-harmful/ )

src/client.ts
```diff
@@ -9079,6 +9083,52 @@ export class MatrixClient extends EventEmitter {
...
+        }
+
+        // If the parent event also has an associated ID we want to re-run the
+        // computation for that parent event.
+        // In the case of the redaction of a reaction that targets a root event
+        // we want that redaction to be pushed to both timeline
+        if (parentEvent?.getAssociatedId()) {
+            return this.eventShouldLiveIn(parentEvent, room, events, roots);
+        } else {
+            // We've exhausted all scenarios, can safely same that this event
```
andybalaam, line 9123: same => assume

andybalaam (approved):
Looks good - minor comments above.

andybalaam:
@gsouquet I see that you are marking a lot of bugs fixed with this change.  Did you get a chance to test all those cases?  I have been testing only the responses-to-the-thread-root part.

gsouquet:
>  Did you get a chance to test all those cases? 

I did test all of them, and the partitioning was the root of all those issues. 
There's a couple more issues that could be solved by this, but did not include them because I have trouble reproducing them reliably

src/models/event-timeline-set.ts
```diff
@@ -159,17 +158,12 @@ export class EventTimelineSet extends EventEmitter {
...
-    public getPendingEvents(thread?: Thread): MatrixEvent[] {
+    public getPendingEvents(): MatrixEvent[] {
         if (!this.room || !this.displayPendingEvents) {
             return [];
         }
 
-        const pendingEvents = this.room.getPendingEvents(thread);
-        if (this.filter) {
-            return this.filter.filterRoomTimeline(pendingEvents);
-        } else {
```
richvdh, line 170: @germain-gg I'm trying to understand this change. I note it changed the behaviour of this method so that it no longer applies the TimelineSet's filter (despite what the documentation says), but  really my question is: why doesn't the calling code just call `room.getPendingEvents`, if it actually wants all the pending events rather than going through the filter?

Issue #2141, opened 2022-01-28
Support refresh tokens


bradjones1:
**Is your feature request related to a problem? Please describe.**

Refresh tokens were introduced in [MSC2918](https://github.com/matrix-org/matrix-doc/pull/2918) but do not appear to be supported in the JS SDK.

**Describe the solution you'd like**

Support for refresh tokens and automatic refresh based on known expiration or auth failure using expired token.

**Describe alternatives you've considered**

This is a spec support feature so I do not believe there is an alternative.

**Additional context**

In more traditional OAuth world, it's possible to use a middleware (such as https://github.com/badgateway/fetch-mw-oauth2) for HTTP transport, but I don't think that would be an option here.

bradjones1:
I'm not generally a TypeScript developer so my JS skills are... lacking but it looks as though this could be supported in a way by creating a middleware-d `fetch` type function that is then set as the request function, per: https://github.com/matrix-org/matrix-js-sdk/blob/164320136391e6dc92ef77323e311f0299bd104e/src/matrix.ts#L59-L92

I think it would be nice to have first-class support for refresh tokens in the SDK itself, but it seems as though telling the SDK to not bother with access tokens, and then handling them in the request function, could be an approach. (Or at least prove a POC.)

bradjones1:
Related to https://github.com/matrix-org/matrix-js-sdk/issues/801 - `request` is deprecated and there is discussion to change HTTP libs, which would help enable things like middleware-d `fetch` or swapping in `axios`.

bradjones1:
Rewriting the SDK's `request` implementation for `fetch` is a greater lift/risk than I think I can handle on my own right now, but I am going to try to do something similar to the [refactoring this project did](https://github.com/huo-ju/mixin-node/commit/02cb3b50f7ad09d477e95f6e7f18d2354221c77e#diff-0e93da6b52116b9c84ab8969554261c34897dcebd86221fb838260f18601eaf3) to move from the `request` request/promise/rejection model with a shim that has a similar function signature.

bradjones1:
Functions added in https://github.com/matrix-org/matrix-js-sdk/pull/2178
React client got it in https://github.com/matrix-org/matrix-react-sdk/pull/7802 but then this was reverted in https://github.com/matrix-org/matrix-react-sdk/pull/7821

Issue #801, opened 2018-12-06
Consider switching http libs


turt2live:
something like https://github.com/visionmedia/superagent works nicely in both the browser and in node. `request` does work in both, but is painful and large for the browser (and causes weird problems sometimes).

turt2live:
Related: https://github.com/matrix-org/matrix-js-sdk/issues/244

turt2live:
Replacing `request` would be a huge undertaking as the js-sdk and its consumers make a huge number of assumptions about the library being used. We'd almost certainly need to cut a major release, although that's not impossible.

Most of the HTTP code should be in one place, with the exception of all the places that rely on the response/error objects from `request`. Riot would also need to make use of the new library.

The shorter answer is that I unfortunately don't have a list, but I do know it is a long one.

stalniy:
I would recommend to use [axios](https://github.com/axios/axios). [request is maintance mode](https://github.com/request/request/issues/3142)

> The most valuable thing request can do for the JavaScript ecosystem is to go into maintenance mode and stop considering new features or major releases

jonassmedegaard:
Maybe this feature comparison is helpful in deciding which alternative to switch to: https://github.com/sindresorhus/got#comparison

jc-delvalle:
I was having some issues login in with JWT token and request. 

I ended up doing this: 
`const client = sdk.createClient({
      baseUrl: 'https://matrix-dev.mydomain.com',
      request: async function (opts, fn) {                                                                                                        
        opts.url = opts.uri;  
        opts.data = opts.body;  
        const res = await axios(opts);  
        fn(null, res, JSON.stringify(res.data));  
      },  
    });`

Since I was not able to make this endpoint work with request in any way, I even tried isolated tests and I always received  'Error: Server returned 403 error' until I switched to axios, this endpoint is working with browser-request.

Since it seems like a very simple change in interface from request to axios, is there any effort already going on switching into another library?

I would love to contribute to this if you think this is the right approach.

t3chguy:
There is a WIP PR to move over to fetch, which is a more likely candidate as it is built-in in the web

jc-delvalle:
Great! Thanks.

bradjones1:
This is the aforementioned PR, though it is 4 years old now: https://github.com/matrix-org/matrix-js-sdk/pull/797

bradjones1:
[...]

PR #2166, opened 2022-02-08
WIP: Fetch support

bradjones1:
This PR is very much WIP, but I'm trying to push the ball forward for #801 and #2141

Type: enhancement
Notes: Breaking change: Transition from deprecated/unsupported request library to fetch

---

## 🚨 BREAKING CHANGES
 * Breaking change ([\#2166](https://github.com/matrix-org/matrix-js-sdk/pull/2166)). Contributed by @bradjones1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,11 +55,11 @@
     "browser-request": "^0.3.3",
     "bs58": "^4.0.1",
     "content-type": "^1.0.4",
+    "cross-fetch": "^3.1.5",
     "loglevel": "^1.7.1",
     "matrix-events-sdk": "^0.0.1-beta.6",
     "p-retry": "^4.5.0",
-    "qs": "^6.9.6",
-    "request": "^2.88.2",
+    "qs": "^6.10.3",
     "unhomoglyph": "^1.0.6"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -80,7 +80,6 @@
     "@types/content-type": "^1.1.5",
     "@types/jest": "^26.0.20",
     "@types/node": "12",
-    "@types/request": "^2.48.5",
     "@typescript-eslint/eslint-plugin": "^5.6.0",
     "@typescript-eslint/parser": "^5.6.0",
     "allchange": "^1.0.6",
--- a/src/client.ts
+++ b/src/client.ts
@@ -161,6 +161,8 @@ import { IThreepid } from "./@types/threepids";
 import { CryptoStore } from "./crypto/store/base";
 import { MediaHandler } from "./webrtc/mediaHandler";
 
+import fetch from 'cross-fetch';
+
 export type Store = IStore;
 export type SessionStore = WebStorageSessionStore;
 
@@ -212,12 +214,10 @@ export interface ICreateClientOpts {
     scheduler?: MatrixScheduler;
 
     /**
-     * The function to invoke for HTTP
-     * requests. The value of this property is typically <code>require("request")
-     * </code> as it returns a function which meets the required interface. See
-     * {@link requestFunction} for more information.
+     * The function to invoke for HTTP requests.
+     * See {@link fetch} for more information.
      */
-    request?: IHttpOpts["request"];
+    fetch?: typeof fetch;
 
     userId?: string;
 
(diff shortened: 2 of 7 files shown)


bradjones1:
So I spent a few hours hacking away on this and it seems to be a heavier lift than the initial work that was done in #797, which was before this library's move to TypeScript. There is quite a bit of logic that follows the callback pattern from `request` rather than a promise-based approach a la `fetch`. I've pushed my latest efforts to this branch, but it's not usable and I have reached the limits of my TS knowledge (which is not saying much.)

I think we should really move away from `request`, but this would likely be a significant breaking change and I'm not sure the maintainers' plan for handling those.

package.json
```diff
@@ -55,11 +55,11 @@
     "browser-request": "^0.3.3",
     "bs58": "^4.0.1",
     "content-type": "^1.0.4",
+    "cross-fetch": "^3.1.5",
```
t3chguy, line 58: Might be worth investigating but this package uses whatwg-fetch (browser polyfill) which may create downstream issues in Element Web inflating bundle size with a polyfill which we never care about as all compatible browsers already support fetch.

src/client.ts
```diff
@@ -212,12 +214,10 @@ export interface ICreateClientOpts {
...
     /**
-     * The function to invoke for HTTP
-     * requests. The value of this property is typically <code>require("request")
-     * </code> as it returns a function which meets the required interface. See
-     * {@link requestFunction} for more information.
+     * The function to invoke for HTTP requests.
+     * See {@link fetch} for more information.
      */
-    request?: IHttpOpts["request"];
+    fetch?: typeof fetch;
```
t3chguy, line 220: I think maybe instead of depending on cross-fetch we could just have the caller pass their favourite fetch, as for typing a shim in `/@types/` would be fine

src/http-api.ts
```diff
@@ -35,6 +34,7 @@ import { IDeferred } from "./utils";
 import { Callback } from "./client";
 import * as utils from "./utils";
 import { logger } from './logger';
+const queryString = require('qs');
```
t3chguy, line 37: imports not requires please

bradjones1:
@t3chguy thank you very much for the look.

As I hope was clear, this is very much POC/WIP and I do not have any working code leveraging this yet. As explained in some of the linked issues, my personal motivation here is to be able to support refresh tokens via a `fetch` middleware, similar to what I can do with OAuth2 services for other purposes.

Anyway, I'm curious your thoughts on what this change would mean re: backwards compatibility, particularly as it seems all of the callers to the HTTP client's public methods should change away from the `request`-style callback functions to use a promise. There's also the question of where the JSON serialization can/should happen, as well as handling for query strings, which had a lot of special-casing in play.

For the moment I have backed out of this rabbit trail for my project and am working on shimming in fetch as a custom function that can be used with the same signature... we'll see how that works out!

t3chguy:
> support refresh tokens via a fetch middleware, similar to what I can do with OAuth2 services for other purposes.

I think it won't be that simple, I think @turt2live is also working on refresh tokens, but the issue is if you have multiple tabs open then they will compete with each other and all battle for refreshing so you need a more resilient way to make sure it only gets done by a single tab.

turt2live:
refresh tokens are sufficiently complicated where they cannot be handled by fetch middleware. The concern isn't that they'll overwrite each other (ultimately it's not bad if they do) - it's that both the old and new token have a strong likelihood of being used, which will trigger the server to do a proper logout.

The code on my end already exists. It's just undergoing polish and cleanup before going up for review.

bradjones1:
Thanks for the feedback.

> the issue is if you have multiple tabs open then they will compete with each other and all battle for refreshing so you need a more resilient way to make sure it only gets done by a single tab.

In my case I'm working so far in a React Native context, so I don't have to worry about coordinating multiple tabs... but I will have a web version and this is indeed an issue, there. Thanks for helping me better understand the trade-offs.

@turt2live looking forward to seeing your work on refresh tokens.

Beyond this question of course is changing away from `request` due to its maintenance status, but that's a broader question/refactor.

bradjones1:
> The code on my end already exists. It's just undergoing polish and cleanup before going up for review.

@turt2live, did your work on refresh tokens ever land anywhere? My shim is kinda-sorta doing the trick in React Native but I'd like to move along to something less janky if it will land in the core SDK. Let me know how I can help.

bradjones1:
Ah, ok... I noted the related commits at https://github.com/matrix-org/matrix-js-sdk/issues/2141#issuecomment-1103241588 - looks like functions to help perform refresh were added in the SDK and then the implementation was added then reverted in the react implementation.

I'm going to close this since my code is very half-baked and I'm not actively working on this. Switching to `fetch` or something similar would be good, but that's already open on https://github.com/matrix-org/matrix-js-sdk/issues/801.

PR #2169, opened 2022-02-10
Log when member event membership is undefined

dbkr:
To diagnose https://github.com/vector-im/element-web/issues/20962

---

### DIFF
--- a/src/models/room-member.ts
+++ b/src/models/room-member.ts
@@ -25,6 +25,7 @@ import * as utils from "../utils";
 import { User } from "./user";
 import { MatrixEvent } from "./event";
 import { RoomState } from "./room-state";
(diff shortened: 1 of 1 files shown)


src/models/room-member.ts
```diff
@@ -116,6 +117,15 @@ export class RoomMember extends EventEmitter {
 
         const oldMembership = this.membership;
         this.membership = event.getDirectionalContent().membership;
+        if (this.membership === undefined) {
+            // logging to diagnose https://github.com/vector-im/element-web/issues/20962
+            // (logs event content, although only of membership events)
+            logger.error(
```
turt2live, line 123: `.trace` to try and get some semblance of a stacktrace?

PR #2174, opened 2022-02-14
Fix synthetic read receipt handling

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/21016

The complexity in this area of code is worth the performance gains, with more testing and comments now it should be more maintainable too.

---

## 🐛 Bug Fixes
 * Fix synthetic read receipt handling ([\#2174](https://github.com/matrix-org/matrix-js-sdk/pull/2174)). Fixes vector-im/element-web#21016.

### DIFF
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -1981,12 +1981,12 @@ export class Room extends EventEmitter {
     }
 
     public getReadReceiptForUserId(userId: string, ignoreSynthesized = false): IWrappedReceipt | null {
-        const [realReceipt, fakeReceipt] = this.receipts["m.read"]?.[userId] ?? [];
+        const [realReceipt, syntheticReceipt] = this.receipts["m.read"]?.[userId] ?? [];
         if (ignoreSynthesized) {
             return realReceipt;
         }
 
-        return fakeReceipt ?? realReceipt;
+        return syntheticReceipt ?? realReceipt;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/models/room.ts
```diff
@@ -1980,11 +1981,12 @@ export class Room extends EventEmitter {
     }
 
     public getReadReceiptForUserId(userId: string, ignoreSynthesized = false): IWrappedReceipt | null {
-        const [realReceipt, syntheticReceipt] = this.receipts["m.read"]?.[userId] ?? [];
-        if (ignoreSynthesized || realReceipt) {
+        const [realReceipt, fakeReceipt] = this.receipts["m.read"]?.[userId] ?? [];
+        if (ignoreSynthesized) {
```
turt2live, line 1985: it's a little bit awkward to have the thing called different names in the function body - can we restore `syntheticReceipt` or rename `ignoreSynthesized` please?

turt2live: this sort of dual naming appears to apply throughout

turt2live (commented):
overall seems fine - just concerned about the dual naming of "fake" and "synthetic" read receipts - we should use a single term to describe them.

My preference is towards "synthetic" largely for historical purposes, but ultimately I don't mind if we pick "fake" instead (especially considering "synthetic" is a bit unfriendly to not-English languages)

turt2live (approved):

PR #2178, opened 2022-02-15
Add functions to support refresh tokens

turt2live:
MSC: https://github.com/matrix-org/matrix-doc/pull/2918 (stable, but not formalized)
For https://github.com/vector-im/element-web/issues/18698
For https://github.com/matrix-org/matrix-react-sdk/pull/7802

---

## ✨ Features
 * Add functions to support refresh tokens ([\#2178](https://github.com/matrix-org/matrix-js-sdk/pull/2178)).

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -51,7 +51,8 @@ import {
     PREFIX_IDENTITY_V2,
     PREFIX_MEDIA_R0,
     PREFIX_R0,
-    PREFIX_UNSTABLE,
+    PREFIX_UNSTABLE, 
+    PREFIX_V1,
     retryNetworkOperation,
     UploadContentResponseType,
 } from "./http-api";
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -6772,6 +6775,29 @@ export class MatrixClient extends EventEmitter {
...
+    public refreshToken(refreshToken: string): Promise<IRefreshTokenResponse> {
+        return this.http.authedRequest(
+            undefined, Method.Post, "/_matrix/client/v1/refresh",
+            undefined, {
+                refresh_token: refreshToken,
+            }, {
+                prefix: "", // set manually
+                inhibitLogoutEmit: true, // we don't want to cause logout loops
+            },
+        );
```
t3chguy, line 6798: can we be consistent with how many args we do per line?

src/client.ts
```diff
@@ -6772,6 +6775,29 @@ export class MatrixClient extends EventEmitter {
...
+     * @return {Promise<IRefreshTokenResponse>} Resolves to the new token.
+     * @return {module:http-api.MatrixError} Rejects with an error response.
+     */
+    public refreshToken(refreshToken: string): Promise<IRefreshTokenResponse> {
+        return this.http.authedRequest(
+            undefined, Method.Post, "/_matrix/client/v1/refresh",
+            undefined, {
+                refresh_token: refreshToken,
+            }, {
+                prefix: "", // set manually
```
t3chguy, line 6795: This should be using `PREFIX_V1` and `/refresh` as the path

t3chguy (approved):
LGTM otherwise

src/client.ts
turt2live: can I say no? :P 

(I don't know what I was doing here tbh)

PR #2179, opened 2022-02-15
Formalise guidelines around writing tests

novocaine:
As discussed internally and in #element-dev

---

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,6 +110,13 @@ must include:
    These are located in `/test/end-to-end-tests`. Ideally, you would also
    include tests for edge and error cases.
 
+Unit tests are expected even when the feature is in labs. It's good practice
+to write tests alongside the code as it ensures the code is testable from
+the start, and gives you a fast feedback loop while you're developing the
+functionality. End-to-end tests should be added prior to the feature
+leaving labs, but don't have to be present from the start (although it might
+be beneficial to have some running early so you can test things faster).
+
 For bugs in those repos, your change must include at least one unit test or
 end-to-end test; which is best depends on what sort of test most concisely
 exercises the area.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,7 +110,7 @@ must include:
    These are located in `/test/end-to-end-tests`. Ideally, you would also
    include tests for edge and error cases.
 
-For bugs in those repos, your change must include at least one unit tests or
+For bugs in those repos, your change must include at least one unit test or
 end-to-end test; which is best depends on what sort of test most concisely
 exercises the area.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -100,18 +100,34 @@ checks, so please check back after a few minutes.
 
 Tests
 -----
-If your PR is a feature (ie. if it's being labelled with the 'T-Enhancement'
-label) then we require that the PR also includes tests. These need to test that
-your feature works as expected and ideally test edge cases too. For the js-sdk
-itself, your tests should generally be unit tests. matrix-react-sdk also uses
-these guidelines, so for that your tests can be unit tests using
-react-test-utils, snapshot tests or screenshot tests.
+Your PR should include tests.
 
-We don't require tests for bug fixes (T-Defect) but strongly encourage regression
-tests for the bug itself wherever possible.
+For new user facing features in `matrix-react-sdk` or `element-web`, you
+must include:
 
-In the future we may formalise this more with a minimum test coverage
-percentage for the diff.
+1. Comprehensive unit tests written in Jest. These are located in `/test`.
+2. "happy path" end-to-end tests.
+   These are located in `/test/end-to-end-tests`. Ideally, you would also
+   include tests for edge and error cases.
+
+For bugs in those repos, your change must include at least one unit tests or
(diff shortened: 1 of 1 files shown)


kerryarchibald (approved):
LGTM

CONTRIBUTING.md
```diff
@@ -100,18 +100,34 @@ checks, so please check back after a few minutes.
...
+
+For new user facing features in `matrix-react-sdk` or `element-web`, you
+must include:
+
+1. Comprehensive unit tests written in Jest. These are located in `/test`.
+2. "happy path" end-to-end tests.
+   These are located in `/test/end-to-end-tests`. Ideally, you would also
+   include tests for edge and error cases.
+
+For bugs in those repos, your change must include at least one unit tests or
```
andybalaam, line 113:
```suggestion
For bugs in those repos, your change must include at least one unit test or
```

andybalaam (approved):
One typo - otherwise wholeheartedly agree!

CONTRIBUTING.md
```diff
@@ -100,18 +100,34 @@ checks, so please check back after a few minutes.
...
-percentage for the diff.
+Your PR should include tests.
+
+For new user facing features in `matrix-react-sdk` or `element-web`, you
+must include:
+
+1. Comprehensive unit tests written in Jest. These are located in `/test`.
+2. "happy path" end-to-end tests.
+   These are located in `/test/end-to-end-tests`. Ideally, you would also
+   include tests for edge and error cases.
```
t3chguy, line 111: this only applies to react-sdk

t3chguy (approved):

CONTRIBUTING.md
turt2live, line 111: we'll need to fix the test suite to actually run easily on all environments to make this enforceable.

CONTRIBUTING.md
```diff
@@ -100,18 +100,34 @@ checks, so please check back after a few minutes.
...
-
-We don't require tests for bug fixes (T-Defect) but strongly encourage regression
-tests for the bug itself wherever possible.
-
-In the future we may formalise this more with a minimum test coverage
-percentage for the diff.
+Your PR should include tests.
+
+For new user facing features in `matrix-react-sdk` or `element-web`, you
+must include:
```
turt2live, line 106: does this include features behind labs? Not sure where the clause went, but previously we had a (potentially informal) policy that tests had to be written for a feature to leave labs, but not to enter labs. I'd like to keep this clause in place.

turt2live (changes requested):
mostly wording clarifications - the principle is agreeable.

CONTRIBUTING.md
novocaine: We can discuss, but we want to shoot for writing unit tests alongside new work, not after the fact; this means the code is much more likely to be testable and you get the benefit of having the tests there to make sure you're not breaking stuff as you go.

For e2e tests, I can see the argument for adding them at the end of feature development.

CONTRIBUTING.md
novocaine: Do we have issues tracking the current problems?

gsouquet (approved):

CONTRIBUTING.md
jryans: The previous policy @turt2live mentions is described in https://github.com/vector-im/element-web/blob/develop/docs/review.md#code-quality, which I guess partially overlaps with this file (and perhaps conflicts with it).

Not sure which repo should be the canonical home of team process docs, but Element Web seems a tad more logical to me... 🤷 There are other useful tidbits in https://github.com/vector-im/element-web/blob/develop/docs/review.md, so it would seem best to preserve and update it where needed.

novocaine: Thanks. That policy isn't linked to from the element-web README afaict, which links to the develop guide for matrix-react-sdk, which links to this file.

jryans: Yes, but all of them are documents we can edit... 😜 

Perhaps good to add separate document clean up task. At the moment, there are enough separate documents in different repos it's likely overwhelming for someone new (contributor or Element employee alike), especially when there are several covering similar topics with different text.

dbkr (approved):
Seems reasonable! Some good points about the various different docs at different levels of the tree: we should try & clean these up a bit (I think I'm convinced that the right place for things to live would be element-web).

CONTRIBUTING.md
turt2live: not off the top of my head, but the problems seem easily encountered. We also discussed a few of them in other channels.

It's very possible that it's just a situation of us needing to try installing them again on affected platforms and see how it goes (and report bugs).

CONTRIBUTING.md
```diff
@@ -110,6 +110,13 @@ must include:
    These are located in `/test/end-to-end-tests`. Ideally, you would also
    include tests for edge and error cases.
 
+Unit tests are expected even when the feature is in labs. It's good practice
```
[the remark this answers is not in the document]
novocaine: I think if we're saying some code can escape the threshold of typical quality, we need a more powerful distinction for marking it as "experimental" than just labs. A lot of labs code these days really is intended for eventual production, and skipping tests during development is problematic because adding tests later usually doesn't work - the code won't be very testable and will need reworking.

SimonBrandner (approved):

PR #2181, opened 2022-02-16
Stabilise token authenticated registration support

govynnus:
Token authenticated registration was added to the Matrix specification in v1.2:
https://spec.matrix.org/v1.2/client-server-api/#token-authenticated-registration

Signed-off-by: Callum Brown <callum@calcuode.com>

---

## ✨ Features
 * Stabilise token authenticated registration support ([\#2181](https://github.com/matrix-org/matrix-js-sdk/pull/2181)). Contributed by @govynnus.

### DIFF
--- a/src/interactive-auth.ts
+++ b/src/interactive-auth.ts
@@ -60,7 +60,7 @@ export enum AuthType {
     Sso = "m.login.sso",
     SsoUnstable = "org.matrix.login.sso",
     Dummy = "m.login.dummy",
-    RegistrationToken = "org.matrix.msc3231.login.registration_token",
+    RegistrationToken = "m.login.registration_token",
 }
 
 export interface IAuthDict {
@@ -79,7 +79,8 @@ export interface IAuthDict {
     // eslint-disable-next-line camelcase
     threepid_creds?: any;
     threepidCreds?: any;
-    registrationToken?: string;
+    // For m.login.registration_token type
+    token?: string;
 }
 
 class NoAuthFlowFoundError extends Error {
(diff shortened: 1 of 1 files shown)


govynnus:
I believe Synapse is the only server implementing this, and there is a release candidate (1.53.0rc1) which uses the stable version:
https://github.com/matrix-org/synapse/blob/develop/CHANGES.md#synapse-1530rc1-2022-02-15

Ruma has also made the change, but I don't think it's released yet:
https://github.com/ruma/ruma/commit/f63168538970c76bdd1639c27606442a80a567a9

govynnus:
Ok.
How would supporting both types work?

Also, Synapse 1.53.0.rc1 does not maintain backwards compatibility, so without support for the stable version the functionality will break for matrix-js-sdk users of updated servers.

turt2live:
The server doesn't need to maintain backwards compatibility in this case, but we as a client won't know what version of the server we're talking to so need to handle the case where legacy servers offer the unstable flows. 

We've implicitly made the decision to support the unstable identifiers by implementing it, so need to maintain that until servers have upgraded enough where it's reasonable for us to remove the backwards compatibility ourselves. 

Usually we avoid this problem entirely by not using enums (because they can't support stable and unstable at the same time), but given we don't really have a choice we end up with `UnstableRegistrationToken` and `RegistrationToken`, then annotate and document the differences between the flows.

govynnus:
How does that look?

govynnus:
@turt2live Hi, any chance you could have a look at this this week?

turt2live (approved):
Thanks!

PR #2186, opened 2022-02-18
Return send event response from MSC3089Branch.createNewVersion()

hughns:
---

## ✨ Features
 * Return send event response from MSC3089Branch.createNewVersion() ([\#2186](https://github.com/matrix-org/matrix-js-sdk/pull/2186)). Contributed by @hughns.


src/models/MSC3089Branch.ts
```diff
@@ -171,7 +172,7 @@ export class MSC3089Branch {
         encryptedContents: FileType,
         info: Partial<IEncryptedFile>,
         additionalContent?: IContent,
-    ): Promise<void> {
+    ): Promise<ISendEventResponse> {
```
gsouquet, line 175: We need to update the JSDoc return type too

gsouquet (approved):
Otherwise LGTM

PR #2192, opened 2022-02-22
Add new room state emit RoomStateEvent.Update for lower-frequency hits

t3chguy:
---

## ✨ Features
 * Add new room state emit RoomStateEvent.Update for lower-frequency hits ([\#2192](https://github.com/matrix-org/matrix-js-sdk/pull/2192)).

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -800,7 +800,8 @@ type RoomEvents = RoomEvent.Name
 type RoomStateEvents = RoomStateEvent.Events
     | RoomStateEvent.Members
     | RoomStateEvent.NewMember
-    | RoomStateEvent.Update;
+    | RoomStateEvent.Update
+    ;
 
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -799,7 +799,8 @@ type RoomEvents = RoomEvent.Name
 
 type RoomStateEvents = RoomStateEvent.Events
     | RoomStateEvent.Members
-    | RoomStateEvent.NewMember;
+    | RoomStateEvent.NewMember
+    | RoomStateEvent.Update;
```
turt2live, line 803: stylistically questionable, but this avoids future cases of having to juggle the semicolon:

```suggestion
    | RoomStateEvent.Update
    ;
```

turt2live (approved):

PR #2193, opened 2022-02-22
Fix bug with one-way audio after a transfer

dbkr:
Seems chrome at least will give you a disabled audio track if you
already had another user media audio track and disabled it, so make
sure our tracks are enabled when we add them. We already did this
on one code path but it didn't get moved over when a new code path
was added.

On the plus side, we now know the reason for the ancient code that
had the comment asking what it was for, so update that.

---

## 🐛 Bug Fixes
 * Fix bug with one-way audio after a transfer ([\#2193](https://github.com/matrix-org/matrix-js-sdk/pull/2193)).


SimonBrandner (approved):

turt2live (approved):
looks like a voip thing

PR #2203, opened 2022-02-24
Add codecov badge to README

andybalaam:
---


t3chguy:
Codecov is gone and we have a readme badge now

PR #2204, opened 2022-02-25
Stop using development /spaces API in favour of v1 /hierarchy API

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/19738

---


gsouquet (approved):
🔥

PR #2209, opened 2022-02-28
Update thread info after MSC3440 updates

gsouquet:
- Update thread filter stable property name
- Change server support to be detected off /versions

---


spec/unit/room.spec.ts
```diff
@@ -1,10 +1,14 @@
+import MockHttpBackend from "matrix-mock-request";
```
t3chguy: this file is missing a copyright header, could you add one please

t3chguy (approved):
LGTM otherwise

PR #2211, opened 2022-03-01
Throw error if both browser-index and (node) index are loaded

t3chguy:
This reverts commit 7fa01ffb also as the fix is no longer necessary

image

---


turt2live (approved):

tiennguyen1293:
Hi, could you please help me how i can resolve to throw error?

`Multiple matrix-js-sdk entrypoints detected!`

image

PR #2213, opened 2022-03-02
Fix wrong event_id being sent for m.in_reply_to of threads

t3chguy:
---

## 🐛 Bug Fixes
 * Fix wrong event_id being sent for m.in_reply_to of threads ([\#2213](https://github.com/matrix-org/matrix-js-sdk/pull/2213)).


dbkr (approved):

andybalaam:
Can we test this?

t3chguy:
Its a type failure, should be caught by better typing rather than tests imo

andybalaam:
I mean, stronger typing would have prevented this bug, but a test of this function could have prevented many others, right? (E.g. missing this field altogether, or supplying "".)

PR #2217, opened 2022-03-04
Fix wrongly asserting that PushRule::conditions is non-null

t3chguy:
---

## 🐛 Bug Fixes
 * Fix wrongly asserting that PushRule::conditions is non-null ([\#2217](https://github.com/matrix-org/matrix-js-sdk/pull/2217)).


kerryarchibald (approved):
LGTM

andybalaam:
Seems likely we could have a unit test for this case?

PR #2218, opened 2022-03-04
Update thread relation fields to match MSC3440 changes

t3chguy:
Based on [matrix-org/matrix-spec-proposals@`9159a5a` (#3440)](https://github.com/matrix-org/matrix-spec-proposals/pull/3440/commits/9159a5a7381a1ba5a7416e78a2a8d80c5a4cffd5)

---


turt2live (changes requested):
clearing review request on this because the MSC has changed.

gsouquet (approved):

PR #2219, opened 2022-03-04
Add test coverage around push rules with no conditions

t3chguy:
Add tests for https://github.com/matrix-org/matrix-js-sdk/pull/2217

---


andybalaam (approved):
Brilliant, thanks.

andybalaam:
Really? 
image

t3chguy:
@andybalaam probably because of how the coverage is gathered, the `?.` won't be distinct from `.` so doesn't improve coverage because the same if branch is being hit. Another test must already be hitting the `[]` case, so the `null` case won't improve coverage.

PR #2223, opened 2022-03-08
Add lint for unused locals

jryans:
TypeScript supports checking for "unused locals", which actually includes
locals, private fields, private methods, etc. If a private thing is actually
needed as part of the API surface, you can change it to public to resolve the
error.

This seems like a good check to have, as it highlights unused bits of code that
perhaps got lost while refactoring. This changes our config to reject them and
fixes up a few existing cases.

---


t3chguy (approved):
Epic!

PR #2226, opened 2022-03-09
Fix missing threads in thread list

gsouquet:
A couple of issues were causing this. We can now paginate in a timeline with no initial pagination token,  the JS-SDK implementation would unfortunately not allow for that. And there was a bug of implementation in the local filter

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -5383,7 +5383,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
                 only: 'highlight',
             };
 
-            if (token && token !== "end") {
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -5383,7 +5383,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
                 only: 'highlight',
             };
 
-            if (token && token !== "end") {
+            if (token !== null && token !== "end") {
                 params.from = token;
             }
```
t3chguy, line 5388: I'm not sure this bit is really needed

PR #2231, opened 2022-03-11
Fix incorrect toJSON for filter-component

gsouquet:
Issue introduced by https://github.com/matrix-org/matrix-js-sdk/pull/2228

---

### DIFF
--- a/src/filter-component.ts
+++ b/src/filter-component.ts
@@ -16,7 +16,11 @@ limitations under the License.
 
 import { RelationType } from "./@types/event";
 import { MatrixEvent } from "./models/event";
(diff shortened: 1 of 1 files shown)


src/filter-component.ts
```diff
@@ -103,6 +103,7 @@ export class FilterComponent {
      * Converts the filter component into the form expected over the wire
      */
     public toJSON(): object {
+        FILTER_RELATED_BY_REL_TYPES;
```
t3chguy, line 106: hm?

t3chguy (approved):
looks sane otherwise

PR #2232, opened 2022-03-11
Send and handle stable name for withheld codes

uhoreg:
Since MSC2399 is finished FCP and it's in the spec, we can use the stable name now.

---

### DIFF
--- a/src/crypto/index.ts
+++ b/src/crypto/index.ts
@@ -3115,7 +3115,8 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
                 this.secretStorage.onRequestReceived(event);
             } else if (event.getType() === "m.secret.send") {
                 this.secretStorage.onSecretReceived(event);
-            } else if (event.getType() === "org.matrix.room_key.withheld") {
+            } else if (event.getType() === "m.room_key.withheld"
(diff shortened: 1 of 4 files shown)


src/crypto/index.ts
```diff
@@ -3115,7 +3115,8 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
                 this.secretStorage.onRequestReceived(event);
             } else if (event.getType() === "m.secret.send") {
                 this.secretStorage.onSecretReceived(event);
-            } else if (event.getType() === "org.matrix.room_key.withheld") {
+            } else if (event.getType() === "m.room_key.withheld"
+                || event.getType() === "org.matrix.room_key.withheld") {
```
turt2live, line 3119: These sorts of checks get easier with a `NamespacedValue`, but considering it'd be 1 line of code to help 2 other lines, the value is questionable.

turt2live (approved):
lgtm - just my eternal quest to keep `*Value` on people's minds, though obviously not super applicable here. 

Can we also get an issue to remind ourselves to remove the unstable stuff in X time?

uhoreg:
> Can we also get an issue to remind ourselves to remove the unstable stuff in X time?

https://github.com/matrix-org/matrix-js-sdk/issues/2233

PR #2234, opened 2022-03-11
Remove groups and groups-related APIs

turt2live:
Paired with but not blocked alongside https://github.com/matrix-org/matrix-react-sdk/pull/8027
Paired with https://github.com/matrix-org/synapse/pull/12200 to a degree
For https://github.com/vector-im/element-web/issues/17532

**Must not land before https://github.com/matrix-org/matrix-react-sdk/pull/8027**

-----

This API is due for removal in Synapse and has been deprecated for a very long time. People should move away from it soon, but just in case we'll declare this as a breaking change.

There is no impact on sync storage here: we happen to store the data in a way that is backwards-compatible for group-supporting clients, and the code guards against missing data from the stores. So, if someone were to revert, they'd be "safe" (probably lose all their group info, but the app wouldn't crash).

---

## 🚨 BREAKING CHANGES
 * Remove groups and groups-related APIs ([\#2234](https://github.com/matrix-org/matrix-js-sdk/pull/2234)).


src/client.ts
```diff
@@ -8794,368 +8764,6 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         return new MSC3089TreeSpace(this, roomId);
     }
 
-    // TODO: Remove this warning, alongside the functions
-    // See https://github.com/vector-im/element-web/issues/17532
-    // ======================================================
-    // **                ANCIENT APIS BELOW                **
-    // ======================================================
```
turt2live, line 8801: I would like to point out that code got added below this warning that is not subject to the "ancient apis" warning. Not sure if we should do a better job communicating these things in the future.

gsouquet (approved):
💯

PR #2235, opened 2022-03-14
Fix 'Empty Room (was %user)' is not translated

Aman-Godara:
Fixes: [21369](https://github.com/vector-im/element-web/issues/21369)

---

## ✨ Features
 * Fix 'Empty Room (was %user)' is not translated ([\#2235](https://github.com/matrix-org/matrix-js-sdk/pull/2235)). Contributed by @Aman-Godara.

### DIFF
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -2416,7 +2416,7 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap> 
         }
 
         if (inviteJoinCount) {
-            return memberNamesToRoomName(otherNames, inviteJoinCount);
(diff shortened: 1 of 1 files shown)


src/models/room.ts
```diff
@@ -2647,15 +2647,15 @@ function memberNamesToRoomName(names: string[], count = (names.length + 1)) {
     if (!names.length) {
```
MadLittleMods, line 2647: There is a `// TODO i18n` comment above we can remove now that this PR is translated it

Aman-Godara:
At first I thought I solved it by raising this PR but I think this is not as straight forward as I thought it to be. Hence, I am closing this PR.

PR #2236, opened 2022-03-14
Move test-utils into a directory

kerryarchibald:
I want to add new typescript test helpers for location work. Move existing js test utilities into a directory so new typescript can live in the same place.

---


andybalaam (approved):
Looks good!

PR #2240, opened 2022-03-15
Fix threads reply count sometimes off by one

gsouquet:
Fixes https://github.com/vector-im/element-web/issues/21415

---

### DIFF
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -191,7 +191,9 @@ export class Thread extends TypedEventEmitter<EmittedEvents, EventHandlerMap> {
         // There is a risk that the `localTimestamp` approximation will not be accurate
         // when threads are used over federation. That could results in the reply
         // count value drifting away from the value returned by the server
-        if (!this.lastEvent || (isThreadReply && event.localTimestamp > this.lastEvent.localTimestamp)) {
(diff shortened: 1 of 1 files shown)


src/models/thread.ts
```diff
@@ -191,7 +191,10 @@ export class Thread extends TypedEventEmitter<EmittedEvents, EventHandlerMap> {
         // There is a risk that the `localTimestamp` approximation will not be accurate
         // when threads are used over federation. That could results in the reply
         // count value drifting away from the value returned by the server
-        if (!this.lastEvent || (isThreadReply && event.localTimestamp > this.lastEvent.localTimestamp)) {
+        if (!this.lastEvent || (isThreadReply
+            && (event.getId() !== this.lastEvent.getId())
```
turt2live, line 195: would be good to get a comment here to say why we are checking the IDs (it implies there is code somewhere that is dealing with the `lastEvent`'s count, but not sure)

turt2live (approved):
lgtm either way

turt2live:
/me notes to watch for auto-merge (it's not really a problem in this case)

gsouquet:
@turt2live argh, sorry, when i enabled auto-merge I thought I had already had an approve and was waiting just for the CI.
Will try to address your comment as part of another PR

PR #2242, opened 2022-03-16
Add support for MSC3575: Sliding Sync

kegsay:
This PR adds API support for Sliding Sync. This is a big change and can be broken down into sliding sync specific changes and general JS SDK changes (which are inherently more risky and needs closer attention).

JS SDK changes:
 - `IRequestOpts` in `http-api.ts` now accepts an optional `baseUrl`. It already accepts a `prefix` but that isn't enough for sliding sync because we want to redirect `/sync` requests to a different server entirely (the proxy server), hence this was added.
 - `slidingSync` has been added as a function to `MatrixClient`. This function just makes a `/sync` request: it doesn't handle the loop or retries.
 - `IStartClientOpts` in `client.ts` now accepts an optional `slidingSync` object, which must be of type `SlidingSync`. If this is present, JS SDK operates in Sliding Sync mode.

Sliding Sync changes:
 - Added new files:
     * `sliding-sync.ts`: contains the core sliding sync API and event loop handling. Does not handle interactions with JS SDK innards. Exposes an event emitter for sliding sync data.
     * `siding-sync-sdk.ts`: contains JS SDK <--> sliding sync data mappings. Listens for events emitted by `SlidingSync`.
     * `conn-management.ts`: factored out keep-alive code. Same as sync v2 otherwise.

Limitations:
 - The data mappings in `sliding-sync-sdk.ts` are very bare bones currently - only state/timeline events are handled (not ephemeral/account/to-device/E2EE). This will be expanded upon in the future.

Contains integration tests for `SlidingSync`. No tests exist yet for `SlidingSyncSdk`.

### DIFF
--- a/spec/integ/matrix-client-sliding-sync.spec.js
+++ b/spec/integ/matrix-client-sliding-sync.spec.js
@@ -31,14 +31,15 @@ describe("SlidingSync", () => {
                 room_subscriptions: {},
             };
             httpBackend.when("POST", syncUrl).respond(200, fakeResp);
-            let deliver = callbackPromise(500, "lifecycle callback was not called");
-            slidingSync.on("SlidingSync.Lifecycle", deliver.callback);
+            let p = listenUntil(slidingSync, "SlidingSync.Lifecycle", (state, resp, err) => {
+                expect(state).toEqual(SlidingSyncState.RequestFinished);
+                expect(resp).toEqual(fakeResp);
+                expect(err).toBeFalsy();
+                return true;
+            });
             slidingSync.start();
             await httpBackend.flush(syncUrl, 1);
-            let lifecycleData = await deliver.promise;
-            expect(lifecycleData[0]).toEqual(SlidingSyncState.RequestFinished);
-            expect(lifecycleData[1]).toEqual(fakeResp);
-            expect(lifecycleData[2]).toBeFalsy();
+            await p;
             slidingSync.stop();
             done();
         });
@@ -80,13 +81,15 @@ describe("SlidingSync", () => {
                 expect(body.room_subscriptions).toBeTruthy();
                 expect(body.room_subscriptions[roomId]).toEqual(roomSubInfo);
             }).respond(200, fakeResp);
-            let deliver = callbackPromise(500, "room callback was not called");
-            slidingSync.on("SlidingSync.RoomData", deliver.callback);
+
+            let p = listenUntil(slidingSync, "SlidingSync.RoomData", (gotRoomId, gotRoomData) => {
+                expect(gotRoomId).toEqual(roomId);
+                expect(gotRoomData).toEqual(wantRoomData);
+                return true;
+            });
             slidingSync.start();
             await httpBackend.flush(syncUrl, 1);
-            let roomData = await deliver.promise;
-            expect(roomData[0]).toEqual(roomId);
-            expect(roomData[1]).toEqual(wantRoomData);
+            await p;
 
             httpBackend.when("POST", syncUrl).check(function(req) {
                 let body = req.data;
@@ -98,8 +101,9 @@ describe("SlidingSync", () => {
                 expect(body.unsubscribe_rooms).toEqual([roomId]);
             }).respond(200, fakeResp);
         
-            deliver = callbackPromise(500, "lifecycle callback was not called");
-            slidingSync.on("SlidingSync.Lifecycle", deliver.callback);
+            p = listenUntil(slidingSync, "SlidingSync.Lifecycle", (state) => {
+                return state === SlidingSyncState.Complete;
+            });
         
             // remove the subscription
             slidingSync.modifyRoomSubscriptions(new Set());
@@ -107,9 +111,16 @@ describe("SlidingSync", () => {
             // kick the connection to resend the unsub
             slidingSync.resend();
             await httpBackend.flush(syncUrl, 2); // flush 2, the one made before the req change and the req change
-            await deliver.promise;
+            await p;
             slidingSync.stop();
             done();
+        });
+
+        it("should be possible to adjust room subscription info whislt syncing", () => {
+            // add 1 sync, modify info, sync, check it is resent
+        });
+        it("should be possible to add room subscriptions whilst syncing", () => {
+            // add 1, sync, add 1, sync, check only 1 is sent
         });
     });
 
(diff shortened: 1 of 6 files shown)


spec/integ/matrix-client-sliding-sync.spec.js
```diff
@@ -0,0 +1,499 @@
+import { SlidingSync, SlidingSyncState } from "../../src/sliding-sync";
+import { TestClient } from "../TestClient";
+import { logger } from '../../src/logger';
+
+/**
+ * Tests for sliding sync. These tests are broken down into sub-tests which are reliant upon one another.
```
dbkr, line 6: huh - I would have thought jest would just split them up and run them across multiple workers, but they seem to have passed...

spec/integ/matrix-client-sliding-sync.spec.js
```diff
@@ -0,0 +1,499 @@
...
+                },
+            });
+
+            const p = listenUntil(slidingSync, "SlidingSync.RoomData", (gotRoomId, gotRoomData) => {
+                expect(gotRoomId).toEqual(roomId);
+                expect(gotRoomData).toEqual(wantRoomData);
+                return true;
+            });
+
+            slidingSync.modifyRoomSubscriptionInfo(newSubInfo);
```
dbkr, line 148: how does this assert that the subscription has actually changed?

src/sliding-sync.ts
```diff
@@ -0,0 +1,611 @@
...
+/**
+ * Internal Class. SlidingList represents a single list in sliding sync. The list can have filters,
+ * multiple sliding windows, and maintains the index->room_id mapping.
+ */
+class SlidingList {
+    private list: MSC3575List;
+    private isModified: boolean;
+
+    // returned data
+    roomIndexToRoomId: Record<number, string>;
```
dbkr, line 117: are these public (or, should they be?)

dbkr (approved):
few minor comments but otherwise looks plausible!

src/sliding-sync.ts
```diff
@@ -0,0 +1,611 @@
...
+                }
+                if (newSubscriptions.size > 0) {
+                    reqBody.room_subscriptions = {};
+                    for (const roomId of newSubscriptions) {
+                        reqBody.room_subscriptions[roomId] = this.roomSubscriptionInfo;
+                    }
+                }
+                this.pendingReq = this.client.slidingSync(reqBody, this.proxyBaseUrl);
+                resp = await this.pendingReq;
+                debuglog(resp);
```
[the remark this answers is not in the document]
kegsay: Long term, no. For now, yes. It helps to identify whether the problem is the client or the server.

spec/integ/matrix-client-sliding-sync.spec.js
```diff
@@ -0,0 +1,604 @@
+import { SlidingSync, SlidingSyncState, ExtensionState } from "../../src/sliding-sync";
```
t3chguy, line 1: Needs copyright header

t3chguy: All new code should be in typescript files (`.ts` extension)

t3chguy (dismissed):

src/conn-management.ts
```diff
@@ -0,0 +1,158 @@
...
+import { Method } from "./http-api";
+import { SyncState } from "./sync";
+
+export type Callback = (newState: SyncState, data?: ISyncStateData) => void;
+
+/**
+ * ConnectionManagement is a class which handles keep-alives and invokes a callback when the connection
+ * is alive/dead. Uses /versions as a keep-alive endpoint.
+ */
+export class ConnectionManagement {
```
t3chguy, line 30: This class is deeply flawed and provides questionable value.

1. `pokeKeepAlive` is only ever called with an arg of `0` so the param is useless as is the internal default value of 2000 + rand
2. Unless the `online` event fires (not available in RN) this class does nothing. `retryImmediately` will immediately return false unless `startKeepAlives` has already ran (via `online` firing)
3. The callback is only fired on error, not when the error goes away, so this cannot be used for "connection management"

Given a /sync failing 3 times already fires the sync error, this class does not provide any real gains beyond maybe improving the latency on a server error once the window fires an `online` event. We also never listen for the `offline` event wtf?

src/client.ts
```diff
@@ -5888,6 +5899,9 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         // There can be only room-kind push rule per room
         // and its id is the room id.
         if (this.pushRules) {
+            if (!this.pushRules[scope] || !this.pushRules[scope].room) {
+                return;
+            }
```
kegsay, line 5904: Defensive programming when the server returns `{}`. There's no guarantees that `scope` will exist.

spec/unit/room.spec.ts
```diff
@@ -361,6 +361,23 @@ describe("Room", function() {
...
+
+            // first add the local echo
+            room.addPendingEvent(localEvent, "TXN_ID");
+            expect(room.timeline.length).toEqual(1);
+
+            // then the remoteEvent, it should de-dupe based on the event ID.
+            const remoteEvent = new MatrixEvent(eventJson);
+            room.addLiveEvents([remoteEvent]);
+            expect(room.timeline.length).toEqual(1);
+        });
```
t3chguy, line 380: This test passes even without any of the changes in this PR

src/conn-management.ts
t3chguy: Yes

src/models/room.ts
```diff
@@ -1917,6 +1917,27 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
...
+                if (localEvent.getId() === event.getId()) {
+                    logger.debug("processLiveEvent: found sent event without txn ID: ", tid, event.getId());
+                    // update the unsigned field so we can re-use the same codepaths
+                    const unsigned = event.getUnsigned();
+                    unsigned.transaction_id = tid;
+                    event.setUnsigned(unsigned);
+                    break;
+                }
+            }
+        }
```
t3chguy, line 1939: This seems like it'll only work if the send request returns before the sliding sync response?

kegsay: Yes. The scenarios are:
 - /send returns then /sync returns : rely on the event ID as you cannot always rely on the txn ID existing.
 - /sync returns then /send returns : rely on txn ID to de-dupe as you don't know the event ID yet.

Given this is a proxy problem, and the proxy sits on the /sync hot-path, it's unlikely that /sync would return before /send. We can guard against it fully by checking the `txnToEvent` map when /send returns, then it will work in all scenarios.

t3chguy: Given the packet switching nature of the network, and especially with the proxy likely running on a different server, you very much could experience higher latency on /send than /sync
Fixing this for all edges would be great

t3chguy (approved):
LGTM, will block until after RC branch cut tomorrow though!

PR #2244, opened 2022-03-17
Create threads event timeline set in the room model

gsouquet:
Related to https://github.com/vector-im/element-web/issues/21439
Related to https://github.com/vector-im/element-web/issues/21442

Some of the ordering was incorrect because some listeners were only active when viewing the thread list. This PR moves the logic to the JS-SDK so that those timelines always listen to incoming events and populate the timeline sets accordingly.

There's no logic change, only moving the code at a place that will be able to listen at all time

This could also make some improvements with the room counter issues, but I have not had a chance to confirm it with a 100% certainty

---


kerryarchibald (approved):
I think there are some tests for the filter logic in react-sdk that could be moved over too.
Otherwise looks good!

PR #2249, opened 2022-03-18
Voice rooms prototype

robintown:
See https://github.com/matrix-org/matrix-react-sdk/pull/8084 for the full details.

Notes: Support call room type from MSC3417

---

### DIFF
--- a/src/@types/event.ts
+++ b/src/@types/event.ts
@@ -119,7 +119,7 @@ export const RoomCreateTypeField = "type";
 
 export enum RoomType {
     Space = "m.space",
-    Call = "org.matrix.msc3417.call",
+    UnstableCall = "org.matrix.msc3417.call",
 }
 
 /**
--- a/src/@types/event.ts
+++ b/src/@types/event.ts
@@ -119,6 +119,7 @@ export const RoomCreateTypeField = "type";
 
 export enum RoomType {
     Space = "m.space",
+    Call = "org.matrix.msc3417.call",
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/models/room.ts
```diff
@@ -2344,6 +2344,14 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
         return this.getType() === RoomType.Space;
     }
 
+    /**
+     * Returns whether the room is a call-room as defined by MSC3417.
+     * @returns {boolean} true if the room's type is RoomType.Call
```
SimonBrandner, line 2349: I don't know why but this feels better here?

```suggestion
     * @returns {boolean} true if the room's type is m.call
```

SimonBrandner (approved):
LGTM - thank you!

src/models/room.ts
robintown: The room type isn't `m.call` at the moment, though - it's `org.matrix.msc3417.call`. Also, the `isSpaceRoom` docs above say the same thing.

SimonBrandner: Ah, ok, _resolved_

src/@types/event.ts
```diff
@@ -119,6 +119,7 @@ export const RoomCreateTypeField = "type";
 
 export enum RoomType {
     Space = "m.space",
+    Call = "org.matrix.msc3417.call",
```
turt2live, line 122: I'd suggest this be called `UnstableCall` so we can introduce the stable version more easily down the line. Alternatively, `UnstableValue` usage outside of this enum (like we do for event types).

turt2live (commented):
generally looks fine on this side - thanks!

turt2live (approved):
this side lgtm

PR #2250, opened 2022-03-20
Don't cancel SAS verifications if `ready` is received after `start`

ara4n:
it's completely valid to receive a `ready` event after having received a
`start` event as messages may be received or decrypted in any order.

partial (but possibly sufficient?) fix for https://github.com/vector-im/element-web/issues/21488

---

## 🐛 Bug Fixes
 * Don't cancel SAS verifications if `ready` is received after `start` ([\#2250](https://github.com/matrix-org/matrix-js-sdk/pull/2250)).


turt2live (approved):
The spec wants us to validate the order, but it also doesn't say how to deal with the ordering or if order is guaranteed, so I suppose this is fine.

Given the risk of touching verification though, I'm tempted to merge this post-RC rather than now for absolute safety (though I also have zero hope of it being tested while on develop/staging anyways).

ara4n:
this is still a valid fix. the fact the spec asks us to validate ordering on something which has no ordering is a spec bug to be fixed.

PR #2251, opened 2022-03-21
Use beacon info event type as beacon identifier

kerryarchibald:
---


turt2live (approved):
seems fine code-wise - thanks!

I've left a comment on the MSC about the event type, as it's now trying to say too much, but that's not a reason to block the functionality from landing.

PR #2255, opened 2022-03-23
Update dependencies

gsouquet:
Update dependencies following RC

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -26,18 +26,19 @@
   dependencies:
     tunnel "0.0.6"
 
-"@ampproject/remapping@^2.0.0":
-  version "2.1.0"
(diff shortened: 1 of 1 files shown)


src/crypto/verification/SAS.ts
```diff
@@ -193,6 +193,7 @@ function calculateMAC(olmSAS: OlmSAS, method: string) {
 }
 
 const calculateKeyAgreement = {
+    // eslint-disable-next-line
```
turt2live, line 196: this needs to be scoped to the exact rule we're disabling

turt2live (changes requested):

gsouquet:
Another set of dependency upgrades to be done after today's release that will superseed this PR

PR #2256, opened 2022-03-23
Fix issues with duplicated MatrixEvent objects around threads

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/21466
Believably fixes https://github.com/vector-im/element-web/issues/21533

---


src/event-mapper.ts
```diff
@@ -29,9 +29,22 @@ export function eventMapperFor(client: MatrixClient, options: MapperOpts): Event
     const decrypt = options.decrypt !== false;
 
     function mapper(plainOldJsObject: Partial<IEvent>) {
```
gsouquet, line 31: I think there would be value in testing that we're not creating dupes. 
Especially given some of the nasty that can bring

gsouquet (changes requested):
Otherwise looks good to me!

src/event-mapper.ts
t3chguy: Good call, will do so shortly

gsouquet (approved):
LGTM 👍  thank you for the tests

PR #2261, opened 2022-03-28
Fix issues with /search and /context API handling for threads

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/21543

---

## 🐛 Bug Fixes
 * Fix issues with /search and /context API handling for threads ([\#2261](https://github.com/matrix-org/matrix-js-sdk/pull/2261)). Fixes vector-im/element-web#21543.


turt2live (approved):
I think this is fine, though there's a lot of formatting changes which make it a bit difficult to see what's going on. The timeline logic looks sane in the new diff at least.

PR #2267, opened 2022-04-01
Fix handling of threaded messages around edits & echoes

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/21633

---

### DIFF
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -2192,11 +2192,60 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap> 
             }
         }
 
-        const threadRoots = this.client.findThreadRoots(events);
+        const threadRoots = this.findThreadRoots(events);
         for (let i = 0; i < events.length; i++) {
             // TODO: We should have a filter to say "only add state event types X Y Z to the timeline".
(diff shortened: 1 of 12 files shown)


src/models/room.ts
```diff
@@ -1630,48 +1639,75 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
...
      * @experimental
      */
-    public async addThreadedEvent(event: MatrixEvent, toStartOfTimeline: boolean): Promise<void> {
-        this.applyRedaction(event);
-        let thread = this.findThreadForEvent(event);
-        if (thread) {
-            await thread.addEvent(event, toStartOfTimeline);
-        } else {
-            thread = await this.createThreadFetchRoot(event.threadRootId, [event], toStartOfTimeline);
+    public async processThreadedEvents(events: MatrixEvent[], toStartOfTimeline: boolean): Promise<unknown> {
```
gsouquet, line 1696: is the type really unknown here? It looks like it's an array of `void` ?

gsouquet (approved):
Looks good at large! I'll give this a go and test it further

src/models/room.ts
t3chguy: Yeah but unknown can be used to denote don't care, alternatively we can await rather than return. Promise<void[]> is just nonsensical imo

PR #1417, opened 2020-07-05
Add _unstable_getSharedRooms

Half-Shot:
This is needed for https://github.com/matrix-org/matrix-react-sdk/pull/4897.

Synapse PR: https://github.com/matrix-org/synapse/pull/7785
MSC: matrix-org/matrix-doc#2666

PR #2271, opened 2022-04-02
Change `_unstable_getSharedRooms` to `_unstable_getMutualRooms`

ShadowJonathan:
In accordance with changes to https://github.com/matrix-org/matrix-spec-proposals/pull/2666, renaming from "shared" to "mutual".

This updates https://github.com/matrix-org/matrix-js-sdk/pull/1417
This blocks https://github.com/matrix-org/matrix-react-sdk/pull/4897

Notes: Change `_unstable_getSharedRooms` to `_unstable_getMutualRooms` and update unstable URL.
Type: task

`Signed-off-by: Jonathan de Jong <jonathan@automatia.nl>`

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -6408,11 +6408,11 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @return {Promise<string[]>} Resolves to a set of rooms
      * @return {module:http-api.MatrixError} Rejects: with an error response.
      */
-    public async _unstable_getSharedRooms(userId: string): Promise<string[]> { // eslint-disable-line
+    public async _unstable_getMutualRooms(userId: string): Promise<string[]> { // eslint-disable-line
         if (!(await this.doesServerSupportUnstableFeature("uk.half-shot.msc2666"))) {
-            throw Error('Server does not support shared_rooms API');
+            throw Error('Server does not support mutual_rooms API');
         }
-        const path = utils.encodeUri("/uk.half-shot.msc2666/user/shared_rooms/$userId", {
+        const path = utils.encodeUri("/uk.half-shot.msc2666/user/mutual_rooms/$userId", {
             $userId: userId,
         });
         const res = await this.http.authedRequest<{ joined: string[] }>(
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -6408,11 +6408,11 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @return {Promise<string[]>} Resolves to a set of rooms
      * @return {module:http-api.MatrixError} Rejects: with an error response.
      */
-    public async _unstable_getSharedRooms(userId: string): Promise<string[]> { // eslint-disable-line
+    public async _unstable_getMutualRooms(userId: string): Promise<string[]> { // eslint-disable-line
         if (!(await this.doesServerSupportUnstableFeature("uk.half-shot.msc2666"))) {
```
turt2live, line 6412: This check is no longer valid, because the flag won't have changed to support the new endpoint: old servers will be advertising that they support the API, but the code won't believe them.

A new flag is required, and both will need to be checked. It looks like the API is equivalent otherwise, so just a matter of using the now-old flag for the `shared_rooms` version and the new flag (preferring it) for the `mutual_rooms` version.

src/client.ts
```diff
@@ -6408,11 +6408,11 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @return {Promise<string[]>} Resolves to a set of rooms
      * @return {module:http-api.MatrixError} Rejects: with an error response.
      */
-    public async _unstable_getSharedRooms(userId: string): Promise<string[]> { // eslint-disable-line
+    public async _unstable_getMutualRooms(userId: string): Promise<string[]> { // eslint-disable-line
```
turt2live, line 6411: I'm not seeing the benefit of the function rename for this change. Even with the fallback semantics implied by the server-side feature flags, we don't need to rename this function.

turt2live (changes requested):
note also that this sort of rename is *not* required in order for the MSC to pass the implementation check. The name of the feature is less important than the function of it.

ShadowJonathan:
(Synapse (re)implementation will happen concurrently)

turt2live (approved):
Thanks!

PR #2272, opened 2022-04-02
Add support for HTML renderings of room topics

Johennes:
For matrix-org/matrix-react-sdk#8215
MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/3765

Relates to vector-im/element-web#5180

This introduces `m.topic` as an extension in `m.room.topic` to allow for both plain text and HTML rendering of room topics.

Eventually most of the logic should be pushed down into matrix-events-sdk (see matrix-org/matrix-events-sdk#5). However, in order not to complicate future architectural changes of this project, @turt2live advised to put the code here for now.

---

## ✨ Features
 * Add support for HTML renderings of room topics ([\#2272](https://github.com/matrix-org/matrix-js-sdk/pull/2272)). Contributed by @Johennes.


turt2live (approved):
this side lgtm

src/client.ts
```diff
@@ -3534,12 +3534,19 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
-        return this.sendStateEvent(roomId, EventType.RoomTopic, { topic: topic }, undefined, callback);
+    public setRoomTopic(
+        roomId: string,
+        topic: string,
+        htmlTopic?: string,
+        callback?: Callback,
+    ): Promise<ISendEventResponse> {
+        const content = ContentHelpers.makeTopicContent(topic, htmlTopic);
+        return this.sendStateEvent(roomId, EventType.RoomTopic, content, undefined, callback);
     }
```
t3chguy, line 3550: This is a breaking change

Johennes: Is it actually? The content is just added to. The `"topic": "..."` field is still there.

Or is adding fields already considered a breaking change?

t3chguy: The `callback` parameter is no longer the 3rd parameter

Johennes: Oh, you're talking about the method, not the content. Sorry.

I guess still better to keep it like this because otherwise the two topic parameters would be spread out in the signature. But happy to swap the 3rd and 4th parameter if you think that'd be better.

t3chguy: I have no real opinions, I commented to explain the addition of the X-Breaking-Change label

turt2live: Ideally we don't break the API surface - good spot @t3chguy 

@Johennes if we can make the third parameter support dual types like we do elsewhere, that'd be best. We'd be overriding the function definition to show the various signatures someone can use and the function itself would have to pick the types apart manually.

Johennes: Good idea. I added the overloads so this should not be a breaking change anymore.

Sunaandrei (approved):

Johennes:
Could I get a pass with "only" 73% coverage? I think I've covered the meaningful bits with tests.

t3chguy:
The gate being 80% is actually considered low, James wanted the gate for the js-sdk to be 100% given it is a reusable SDK. The bits in topic.ts are fully covered, content-helpers is lacking one line and client.ts is at 59.3%.

https://sonarcloud.io/code?id=matrix-js-sdk&pullRequest=2272&selected=matrix-js-sdk%3Asrc%2Fcontent-helpers.ts&line=217
`setRoomTopic` looks entirely untested? https://sonarcloud.io/code?id=matrix-js-sdk&pullRequest=2272&selected=matrix-js-sdk%3Asrc%2Fclient.ts&line=3573

t3chguy:
Could you also please pull latest develop, some of the checks aren't running seemingly

Johennes:
> Are there any parts here which are infeasible to cover for some reason?

I suppose not infeasible, no. The main untested chunk `setRoomTopic` just seemed like it might not be worth covering because it mainly just calls `sendStateEvent`. I see the rationale for the coverage threshold though so let me see what I can do to meet it.

> Could you also please pull latest develop, some of the checks aren't running seemingly

Hm, I had merged in develop just last night. 🤔 I'll merge it again when working on the tests. Hopefully that fixes it.

t3chguy:
> The main untested chunk setRoomTopic just seemed like it might not be worth covering because it mainly just calls sendStateEvent.

The added backwards compatibility overload is certainly worth testing, those are super fragile.

t3chguy:
(Adding the label re-ran the missing CI job)

Johennes:
Hm, something's wrong. I've added tests for `setRoomTopic` and it's now fully covered. However, SonarCloud still rejects the PR with

> 74.6% Coverage on New Code (is less than 80%)

When I click [through](https://sonarcloud.io/component_measures?id=matrix-js-sdk&pullRequest=2272&metric=new_coverage&view=list), it lists a lot of uncovered lines in "new" code for files that I haven't actually touched, like 81 lines in `room.ts`.

@t3chguy any idea what's wrong here?

turt2live:
I wonder if it just doesn't understand merge commits...

t3chguy:
Not sure what happened here - https://github.com/matrix-org/matrix-js-sdk/pull/2366 was unaffected, showing 81.6% covg before & after the merge commit

t3chguy:
image

Wonder if something has gone a bit strange on this branch given gh can't even rebase it

t3chguy (approved):

t3chguy:
Skipping Sonar check due to weirdness, coverage is appropriate already

Johennes:
Thanks @t3chguy!

PR #2277, opened 2022-04-06
Make self membership less prone to races

robintown:
Previously, if you tried to peek a room while the sync loop was still pulling in your own join event for that room, the peek could end up overwriting the synced membership with `null`, causing the room to get stuck in a 'peeking' state. By bundling the call to `updateMyMembership` with `recalculate`, we ensure that it the actual latest membership value gets set.

Type: defect

Closes https://github.com/vector-im/element-web/issues/21661.


gsouquet (approved):
Looks great to me 👍

PR #2281, opened 2022-04-07
Live location sharing - update beacon_info implementation to latest MSC

kerryarchibald:
https://github.com/matrix-org/matrix-spec-proposals/pull/3672

- changes unstable event prefixes to use `msc3672`
- changes beacon identifier logic. Previously beacons used unique event types to enable multiple state events per user per room so event type could be used as an identifier. According to msc3672 plain event types and state_key are used for beacons, identifier is now `<roomId>_<state_key>`. 
- add extra logic about beacon update

---

### DIFF
--- a/spec/test-utils/beacon.ts
+++ b/spec/test-utils/beacon.ts
@@ -42,7 +42,6 @@ export const makeBeaconInfoEvent = (
     roomId: string,
     contentProps: Partial<InfoContentProps> = {},
     eventId?: string,
-    eventTypeSuffix?: string,
 ): MatrixEvent => {
     const {
         timeout, isLive, description, assetType,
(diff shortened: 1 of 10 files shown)


spec/test-utils/beacon.ts
```diff
@@ -51,12 +50,14 @@ export const makeBeaconInfoEvent = (
         ...contentProps,
     };
     const event = new MatrixEvent({
-        type: `${M_BEACON_INFO.name}.${sender}.${eventTypeSuffix || Date.now()}`,
+        type: M_BEACON_INFO.name,
```
turt2live, line 53: 🎉

turt2live (approved):

PR #2282, opened 2022-04-08
Draft: Refresh room timeline when we see a MSC2716 marker event

MadLittleMods:
Refresh room timeline when we see a MSC2716 marker event.

The "marker" events are being used as state now because this way they can't be lost in the timeline. Regardless of when they were sent, we will still have the latest version of the state to compare against. Any time we see our latest state value change for marker events, refresh the timeline.

> In a [sync meeting with @ara4n](https://docs.google.com/document/d/1KCEmpnGr4J-I8EeaVQ8QJZKBDu53ViI7V62y5BzfXr0/edit#bookmark=id.67nio1ka8znc), we came up with the idea to make the `marker` events as state events. When the client sees that the `m.room.marker` state changed to a different event ID, it can throw away all of the timeline and re-fetch as needed.
>
> For homeservers where the [same problem](https://github.com/matrix-org/matrix-doc/pull/2716#discussion_r782499674) can happen, we probably don't want to throw away the whole timeline but it can go up the `unsigned.replaces_state` chain of the `m.room.marker` state events to get them all.
>
> In terms of state performance, there could be thousands of `marker` events in a room but it's no different than room members joining and leaving over and over like an IRC room.
>
> *-- https://github.com/matrix-org/matrix-spec-proposals/pull/2716#discussion_r782629097*

## Todo

 - [ ] Add tests
    - [x] When already joined to room, new `marker` event will cause `timelineNeedsRefresh`
    - [ ] `marker` event already sent in room, user joins room where the `/sync` `state` includes `marker` should *NOT* trigger `timelineNeedsRefresh`
    - [ ] When already joined to room, `marker` event already sent in room, initial `/sync` `state` includes `marker` should *NOT* trigger `timelineNeedsRefresh`
    - [ ] When `syncFromCache` where the state includes marker should *NOT* trigger `timelineNeedsRefresh`
    - [ ] In normal room version, `marker` from creator will cause `timelineNeedsRefresh`
    - [ ] In normal room version, `marker` not from creator  should *NOT* trigger `timelineNeedsRefresh`
 - [ ] Add method to refresh the timeline to use in https://github.com/matrix-org/matrix-react-sdk/pull/8303#discussion_r848999579

### Dev notes

```
yarn jest spec/integ/matrix-client-syncing.spec.js --silent=false --verbose=false
```

---

```js
        Error.stackTraceLimit = Infinity;
        console.log(`setStateEvents timelineWasEmpty=${markerFoundOptions && markerFoundOptions.timelineWasEmpty} stateEvents:\n`, stateEvents.map((ev) => {
            return `\t${ev.getType()} ${ev.getId()}`;
        }).join('\n'), new Error().stack);
```

---

```
room.resetLiveTimeline(null, null);
```

```
addEventsToTimeline
addEventToTimeline
addLiveEvent

resetLiveTimeline

resetAllTimelines

removeFilteredTimelineSet


syncFromCache
processSyncResponse
processRoomEvents
room.addLiveEvents


sessionStore


processSyncResponse
processRoomEvents
initialiseState
setStateEvents

load
getEventTimeline
initialiseState
setStateEvents
```

---

## ✨ Features
 * Draft: Refresh room timeline when we see a MSC2716 marker event ([\#2282](https://github.com/matrix-org/matrix-js-sdk/pull/2282)). Contributed by @MadLittleMods.

### DIFF
--- a/src/models/event-timeline-set.ts
+++ b/src/models/event-timeline-set.ts
@@ -639,7 +639,7 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
     ) {
         const eventId = event.getId();
         timeline.addEvent(event, {
-            atStart: toStartOfTimeline,
+            toStartOfTimeline,
             roomState,
             fromInitialState,
         });
--- a/src/models/event-timeline-set.ts
+++ b/src/models/event-timeline-set.ts
@@ -641,7 +641,13 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
         timeline.addEvent(event, {
             toStartOfTimeline,
             roomState,
-            fromInitialState,
+            // We don't want to refresh the timeline:
+            //  1. If it's persons first time syncing the room, they won't have
+            //     any old events cached to refresh. This could be from initial
+            //     sync or just the first time syncing the room since joining.
+            //  2. If we're re-hydrating from `syncFromCache` because we already
+            //     processed any marker event state that was in the cache
+            fromInitialState: fromInitialState || fromCache,
         });
         this._eventIdToTimeline[eventId] = timeline;
 
--- a/src/models/event-timeline-set.ts
+++ b/src/models/event-timeline-set.ts
@@ -56,6 +56,20 @@ export interface IRoomTimelineData {
     liveEvent?: boolean;
 }
 
+export interface IAddLiveEventOptions {
+    duplicateStrategy?: DuplicateStrategy;
+    fromCache?: boolean;
+    roomState?: RoomState;
+    fromInitialState?: boolean;
+}
+
+export interface IAddEventToTimelineOptions {
+    toStartOfTimeline: boolean;
+    fromCache?: boolean;
+    roomState?: RoomState;
+    fromInitialState?: boolean;
+}
+
 type EmittedEvents = RoomEvent.Timeline | RoomEvent.TimelineReset;
 
 export type EventTimelineSetHandlerMap = {
@@ -431,7 +445,9 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
 
             if (!existingTimeline) {
                 // we don't know about this event yet. Just add it to the timeline.
-                this.addEventToTimeline(event, timeline, toStartOfTimeline);
+                this.addEventToTimeline(event, timeline, {
+                    toStartOfTimeline
+                });
                 lastEventWasNew = true;
                 didUpdate = true;
                 continue;
@@ -529,9 +545,12 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
      */
     public addLiveEvent(
         event: MatrixEvent,
-        duplicateStrategy: DuplicateStrategy = DuplicateStrategy.Ignore,
-        fromCache = false,
-        roomState?: RoomState,
+        {
+            duplicateStrategy = DuplicateStrategy.Ignore,
+            fromCache = false,
+            roomState,
+            fromInitialState
+        }: IAddLiveEventOptions = {}
     ): void {
         if (this.filter) {
             const events = this.filter.filterRoomTimeline([event]);
@@ -570,7 +589,12 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
             return;
         }
 
-        this.addEventToTimeline(event, this.liveTimeline, false, fromCache, roomState);
+        this.addEventToTimeline(event, this.liveTimeline, {
+            toStartOfTimeline: false,
+            fromCache,
+            roomState,
+            fromInitialState,
+        });
     }
 
     /**
@@ -589,12 +613,19 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
     public addEventToTimeline(
         event: MatrixEvent,
         timeline: EventTimeline,
-        toStartOfTimeline: boolean,
-        fromCache = false,
-        roomState?: RoomState,
+        {
+            toStartOfTimeline,
+            fromCache = false,
+            roomState,
+            fromInitialState = false
+        }: IAddEventToTimelineOptions,
     ) {
         const eventId = event.getId();
-        timeline.addEvent(event, toStartOfTimeline, roomState);
+        timeline.addEvent(event, {
+            atStart: toStartOfTimeline, 
+            stateContext: roomState,
+            fromInitialState,
+        });
         this._eventIdToTimeline[eventId] = timeline;
 
         this.setRelationsTarget(event);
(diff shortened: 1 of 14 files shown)


src/sync.ts
```diff
@@ -239,6 +246,27 @@ export class SyncApi {
...
+                // marker. All of the historical content will be in the
+                // `/messsages` responses from here on out.
+                event.getId() !== room.getLastMarkerEventIdProcessed()
+            ) {
+                // Saw new marker event, refreshing the timeline
+                console.log('Saw new marker event, refreshing the timeline');
+                room.resetLiveTimeline(null, null);
+                room.setLastMarkerEventIdProcessed(event.getId())
+            }
+        });
```
MadLittleMods, line 269: TODO: We don't want to refresh the timeline when:

 - [ ] it's the persons first time syncing the room (and got the state for the first time). They won't have any old events cached so there is nothing to refresh
 - [ ] Someone is doing a `syncFromCache` and seeing the state that way. We already processed the marker event state that was in the cache
    - We could solve this by saving `lastMarkerEventIdProcessed` in localStorage so it's persistent across browser refreshes.

MadLittleMods: Addressed in https://github.com/matrix-org/matrix-js-sdk/pull/2282/commits/ffbcbff97fc9f08136d0bdd18e956051e3c93feb

src/sync.ts
```diff
@@ -239,6 +247,37 @@ export class SyncApi {
...
+            if(
+                isValidMsc2716Event &&
+                // We only need to throw the timeline away once, when we see a
+                // marker. All of the historical content will be in the
+                // `/messsages` responses from here on out.
+                event.getId() !== room.getLastMarkerEventIdProcessed()
+            ) {
+                // Saw new marker event, refreshing the timeline
+                console.log('Saw new marker event, refreshing the timeline', new Error().stack);
+                room.resetLiveTimeline(null, null);
```
MadLittleMods, line 277: TODO: Test if this is the function we actually want to use.

Our goal is to throw away the current timeline and re-fetch events for the room from the server.

MadLittleMods: It would probably be better if we re-fetched the new events first, then replaced the timeline

turt2live (commented):
review to clear codeowners. When leaving draft it'll need to be re-reviewed.

src/sync.ts
MadLittleMods: Thinking about this more, do we even want to do this automatically? Someone could cause your Element client to constantly refresh the timeline by just sending marker events over and over. Granted, you probably want to leave a room like this 🤷. Perhaps also some sort of DOS vector since everyone will be refreshing and hitting the server at the exact same time.

I think we might want to approach this to be a similar flow to how Element receives app updates. When the client receives a marker event, we first notify the user that history was imported and is available to see after they refresh their timeline (which can be a button action). The UI can probably use something similar to the "Connectivity to the server has been lost." banner.

src/models/event-timeline-set.ts
```diff
@@ -56,6 +56,20 @@ export interface IRoomTimelineData {
     liveEvent?: boolean;
 }
 
+export interface IAddLiveEventOptions {
+    duplicateStrategy?: DuplicateStrategy;
+    fromCache?: boolean;
+    roomState?: RoomState;
+    fromInitialState?: boolean;
```
MadLittleMods, line 63: `fromFirstSyncInRoom` is maybe a better name

src/sync.ts
MadLittleMods: Updated to prompt the user to refresh their timeline: https://github.com/matrix-org/matrix-react-sdk/pull/8303





> We can detect whether the history imported affects any timelines on the client and only ask to refresh then. To detect, when we receive a marker event, make a request to `/context/{insertionEventId}` with the `insertionEventId` the marker is pointing to and check if any its `prev_events` are in the timeline. `/context` doesn't work over federation but the homeserver backfills the insertion event when it receives the marker so it already has it available.

We can't do this because `prev_events` aren't available in the client API's 😢. I'm not too keen to add an extra field to the marker or insertion event indicating where the history is imported next to as there isn't a way to enforce that it's true (a homeserver could lie and make it up). I'd rather keep the source of truth in the `prev_events`.

src/models/event-timeline-set.ts
MadLittleMods: That isn't a good name either 🤷‍♀️

src/models/event-timeline-set.ts
```diff
@@ -56,6 +56,38 @@ export interface IRoomTimelineData {
...
+    fromCache?: boolean;
+    /** The state events to reconcile metadata from */
+    roomState?: RoomState;
+    /** Whether the state is part of the first state snapshot we're seeing in
+     *  the room. This could be happen in a variety of cases:
+     *  1. From the initial sync
+     *  2. It's the first state we're seeing after joining the room
+     *  3. Or whether it's coming from `syncFromCache` */
+    fromInitialState?: boolean;
+}
```
MadLittleMods, line 77: Instead of dealing with unwieldy long function signatures with unlabeled and optional arguments in the usage, I've refactored some of these into option objects now that I am adding yet another `boolean` here.

Before:

```js
room.addLiveEvents(mainTimelineEvents, null, fromCache, timelineWasEmpty);
```

After:

```js
room.addLiveEvents(mainTimelineEvents, {
    duplicateStrategy: null,
    fromCache,
    fromInitialState: timelineWasEmpty
});
```

src/sync.ts
```diff
@@ -240,6 +249,82 @@ export class SyncApi {
                 RoomMemberEvent.Membership,
             ]);
         });
+
+        // TODO: Should we just move this to `room.ts`?
```
MadLittleMods, line 253: Might be fine. This is within the `registerStateListeners` function which seems pretty relevant 🤷

src/sync.ts
```diff
@@ -240,6 +249,87 @@ export class SyncApi {
...
+
+        // When we see the marker state change in the room, we know there is
+        // some new historical messages imported by MSC2716 `/batch_send`
+        // somewhere in the room and we need to throw away the timeline to make
+        // sure the historical messages are shown when we paginate `/messages`
+        // again.
+        room.currentState.on(RoomStateEvent.Marker, async function(
+            markerEvent,
+            { fromInitialState }: ISetStateOptions = {},
+        ) {
```
MadLittleMods, line 261: Calling out one points of interest in the PR (new logic)

src/models/room-state.ts
```diff
@@ -401,6 +414,8 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
 
                 // assume all our sentinels are now out-of-date
                 this.sentinels = {};
+            } else if (event.getType() === EventType.Marker) {
+                this.emit(RoomStateEvent.Marker, event, setStateOptions);
```
MadLittleMods, line 418: When we see a new marker state event, emit an event so we can listen for it in `sync.ts` -> `registerStateListeners` and see whether we need to refresh the timeline

MadLittleMods:
Closing in favor of https://github.com/matrix-org/matrix-js-sdk/pull/2299

PR #2285, opened 2022-04-08
Support for MSC2457 logout_devices param for setPassword()

hughns:
---

## ✨ Features
 * Support for MSC2457 logout_devices param for setPassword() ([\#2285](https://github.com/matrix-org/matrix-js-sdk/pull/2285)). Contributed by @hughns.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -7815,15 +7815,35 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * Make a request to change your password.
      * @param {Object} authDict
      * @param {string} newPassword The new desired password.
+     * @param {boolean} logoutDevices Should all sessions be logged out after the password change. Defaults to true.
      * @param {module:client.callback} callback Optional.
      * @return {Promise} Resolves: TODO
      * @return {module:http-api.MatrixError} Rejects: with an error response.
      */
-    public setPassword(authDict: any, newPassword: string, callback?: Callback): Promise<any> { // TODO: Types
+    public setPassword(
+        authDict: any,
+        newPassword: string,
+        logoutDevices: boolean,
+        callback?: Callback
+        ): Promise<any>;
+    public setPassword(
+        authDict: any,
+        newPassword: string,
+        logoutDevices?: Callback | boolean,
+        callback?: Callback,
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -7815,15 +7815,35 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
      * @return {Promise} Resolves: TODO
      * @return {module:http-api.MatrixError} Rejects: with an error response.
      */
-    public setPassword(authDict: any, newPassword: string, callback?: Callback): Promise<any> { // TODO: Types
+    public setPassword(
+        authDict: any,
+        newPassword: string,
+        logoutDevices: boolean,
+        callback?: Callback
+        ): Promise<any>;
```
t3chguy, line 7828:
```suggestion
        callback?: Callback,
    ): Promise<any>;
```

src/client.ts
```diff
@@ -7815,15 +7815,35 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+        newPassword: string,
+        logoutDevices?: Callback | boolean,
+        callback?: Callback,
+    ): Promise<any> { // TODO: Types
+        if (typeof logoutDevices !== 'boolean') {
+            logoutDevices = true;
+        }
+        if (typeof logoutDevices === 'function') {
+            callback = logoutDevices;
+        }
```
t3chguy, line 7840: I think these might be backwards, clobbering the callback to be `true` in the backwards compat case

t3chguy (changes requested):

src/client.ts
hughns: Well spotted, thank you.

t3chguy (approved):

t3chguy:
Optional: could we get test coverage over the backwards compatible signature?

hughns:
@t3chguy i added tests trying to follow existing patterns.

In doing so I noted that it would be better to make the default behaviour actually mimic the current behaviour. Specifically making it so that if the existing setPassword signature is used then the `logout_devices` param is not specified in the `POST` request.

t3chguy (approved):
:shipit:

PR #2286, opened 2022-04-08
Fix issues around echo & redaction handling in threads

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/21716
Fixes https://github.com/vector-im/element-web/issues/21662
Fixes https://github.com/vector-im/element-web/issues/21711

---

### DIFF
--- a/spec/test-utils/test-utils.ts
+++ b/spec/test-utils/test-utils.ts
@@ -80,7 +80,7 @@ interface IEventOpts {
     redacts?: string;
 }
 
-let i = 1; // counter for events, easier for comparison of randomly generated events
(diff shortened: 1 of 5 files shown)


spec/test-utils/test-utils.ts
```diff
@@ -79,6 +80,7 @@ interface IEventOpts {
     redacts?: string;
 }
 
+let i = 1; // counter for events, easier for comparison of randomly generated events
```
turt2live, line 83: if it's going in global scope it should get a better name

turt2live (approved):
lgtm otherwise

PR #2287, opened 2022-04-08
Ignore `eslint --fix` formatting changes in git blame

MadLittleMods:
Ignore `eslint --fix` formatting changes in git blame

Docs: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

## Dev notes

```
git log --all --grep='eslint'
```

---

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -32,7 +32,7 @@ f4a7395e3a3751a1a8e92dd302c49175a3296ad2
 cee7f7a280a8c20bafc21c0a2911f60851f7a7ca
 # eslint --fix
 0fa9f7c6098822db1ae214f352fd1fe5c248b02c
-# eslint --fix for lots of whitespace
+# eslint --fix for lots of white-space
 5abf6b9f208801c5022a47023150b5846cb0b309
 # eslint --fix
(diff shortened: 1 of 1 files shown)


.git-blame-ignore-revs
```diff
@@ -0,0 +1,41 @@
...
+# Fix linting on tests
+cee7f7a280a8c20bafc21c0a2911f60851f7a7ca
+# eslint --fix
+0fa9f7c6098822db1ae214f352fd1fe5c248b02c
+# eslint --fix for lots of white-space
+5abf6b9f208801c5022a47023150b5846cb0b309
+# eslint --fix
+7ed65407e6cdf292ce3cf659310c68d19dcd52b2
+# Switch to ESLint from JSHint (Google eslint rules as a base)
+e057956ede9ad1a931ff8050c411aca7907e0394
```
MadLittleMods, line 40: I went through all of the entries for `git log --all --grep='eslint'` and added anything which was making formatting changes.

turt2live (approved):
I haven't checked the commits, but hopefully it's fine.

MadLittleMods:
Awesome! Just tried out the blame it led me straight to the commit instead of 4 different layers of cruft.

Thanks for the review and merge @turt2live 🦏

PR #2297, opened 2022-04-13
Add MatrixClient.doesServerSupportLogoutDevices() for MSC2457

hughns:
---

## ✨ Features
 * Add MatrixClient.doesServerSupportLogoutDevices() for MSC2457 ([\#2297](https://github.com/matrix-org/matrix-js-sdk/pull/2297)). Contributed by @hughns.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -6598,8 +6598,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @return {Promise<boolean>} true if server supports the `logout_devices` parameter
      */
     public async doesServerSupportLogoutDevices(): Promise<boolean> {
-        const response = await this.getVersions();
-        return response?.versions?.includes("r0.6.1");
+        return this.isVersionSupported("r0.6.1");
     }
 
     /**
--- a/src/client.ts
+++ b/src/client.ts
@@ -6594,6 +6594,15 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 
     /**
+     * Query the server to see if it supports the MSC2457 `logout_devices` parameter when setting password
+     * @return {Promise<boolean>} true if server supports the `logout_devices` parameter
+     */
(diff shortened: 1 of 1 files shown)


t3chguy (approved):

t3chguy (changes requested):

turt2live:
out of interest, why are we adding more of these `doesServerSupportX` functions for things that are released? We should be able to do it inline at the code site rather than have a dedicated function.

t3chguy:
Multiple places call it, where you put the utility function might as well be js-sdk for re-use

t3chguy (approved):
:shipit:

turt2live:
it's quite questionable as a utility function even still, honestly. These functions make sense for unstable features, not for released things.

The check is also a bit awkward because it doesn't account for the fact that a server doesn't need to implement r0.6.1 but can implement v1.1 or v1.2. This isn't a problem unique to this change though, so won't require it being fixed here.

t3chguy:
> The check is also a bit awkward because it doesn't account for the fact that a server doesn't need to implement r0.6.1 but can implement v1.1 or v1.2. This isn't a problem unique to this change though, so won't require it being fixed here.

Agreed, even more reason for them to happen in js-sdk where they can be fixed holistically.

Though would be nice if they could be extracted from the MatrixClient into a separate Features class or similar to make this class less huge

hughns:
> 



> Multiple places call it, where you put the utility function might as well be js-sdk for re-use

Exactly + ...

What I actually wanted to check is whether version `>=r0.6.1` is supported. So that if a future server removed support for `r0.6.1` but continued to support `v1.1` then the function should still return `true`.

However, I didn't find such existing logic and so decided it would be better to leave the "flawed" version checking logic alongside the other "flawed" logic rather than strewn across the codebase (and without bloating my oneliner PR).

turt2live:
I would still argue that a purpose-built `doesServerSupportX` function is well over the top for the intended case here :p

hughns:
Haha. Like great minds adrift in the night.

PR #2298, opened 2022-04-13
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) or [String.prototype.substring()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) which work similarily but aren't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

---

### DIFF
--- a/examples/node/app.js
+++ b/examples/node/app.js
@@ -341,7 +341,7 @@ function printLine(event) {
 
     var maxNameWidth = 15;
     if (name.length > maxNameWidth) {
-        name = name.substr(0, maxNameWidth-1) + "\u2026";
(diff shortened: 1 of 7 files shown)


turt2live (approved):
Thanks!

PR #2299, opened 2022-04-14
Timeline needs to refresh when we see a MSC2716 marker event 

MadLittleMods:
Inform the client that historical messages were imported in the timeline and they should refresh the timeline in order to see the new events.

Companion `matrix-react-sdk` PR: https://github.com/matrix-org/matrix-react-sdk/pull/8354

The `marker` events are being used as state now because this way they can't be lost in a timeline gap. Regardless of when they were sent, we will still have the latest version of the state to compare against. Any time we see our latest state value change for marker events, prompt the user that the timeline needs to refresh.

> In a [sync meeting with @ara4n](https://docs.google.com/document/d/1KCEmpnGr4J-I8EeaVQ8QJZKBDu53ViI7V62y5BzfXr0/edit#bookmark=id.67nio1ka8znc), we came up with the idea to make the `marker` events as state events. When the client sees that the `m.room.marker` state changed to a different event ID, it can throw away all of the timeline and re-fetch as needed.
>
> For homeservers where the [same problem](https://github.com/matrix-org/matrix-doc/pull/2716#discussion_r782499674) can happen, we probably don't want to throw away the whole timeline but it can go up the `unsigned.replaces_state` chain of the `m.room.marker` state events to get them all.
>
> In terms of state performance, there could be thousands of `marker` events in a room but it's no different than room members joining and leaving over and over like an IRC room.
>
> *-- https://github.com/matrix-org/matrix-spec-proposals/pull/2716#discussion_r782629097*

### Why are we just setting `timlineNeedsRefresh` (and [prompting the user](https://github.com/matrix-org/matrix-react-sdk/pull/8354)) instead of automatically refreshing the timeline for the user?

If we refreshed the timeline automatically, someone could cause your Element client to constantly refresh the timeline by just sending marker events over and over. Granted, you probably want to leave a room like this 🤷. Perhaps also some sort of DOS vector since everyone will be refreshing and hitting the server at the exact same time.

In order to avoid the timeline maybe going blank during the refresh, we could re-fetch the new events first, then replace the timeline. But the points above still stand on why we shouldn't.

## Todo

 - [x] Add tests
 - [x] Add tests for `refreshLiveTimeline`
    - Added as e2e tests in https://github.com/matrix-org/matrix-react-sdk/pull/8354
 - [x] Add method to refresh the timeline to use in https://github.com/matrix-org/matrix-react-sdk/pull/8303#discussion_r848999579

## Dev notes

```
yarn jest spec/integ/matrix-client-syncing.spec.js --silent=false --verbose=false
```

---

```js
        Error.stackTraceLimit = Infinity;
        console.log(`setStateEvents timelineWasEmpty=${markerFoundOptions && markerFoundOptions.timelineWasEmpty} stateEvents:\n`, stateEvents.map((ev) => {
            return `\t${ev.getType()} ${ev.getId()}`;
        }).join('\n'), new Error().stack);
```

---

Functions with overloads:

```
eventTimelineset.addLiveEvent
eventTimelineSet.addEventToTimeline
eventTimeline.addEvent
room.addLiveEvent
room.addLiveEvents
```

---

```
room.resetLiveTimeline(null, null);
```

```
addEventsToTimeline
addEventToTimeline
addLiveEvent

resetLiveTimeline

resetAllTimelines

removeFilteredTimelineSet


syncFromCache
processSyncResponse
processRoomEvents
room.addLiveEvents


sessionStore


processSyncResponse
processRoomEvents
initialiseState
setStateEvents

load
getEventTimeline
initialiseState
setStateEvents
```

```
hasSyncedBefore
getSyncState()
```

```js
function serializeDebugInfoFromTimelineSets(timelineSets): { [key: string]: string[] }[] {
    const serializedEventIdsInTimelineSet = timelineSets.map((timelineSet) => {
        const timelineMap = {};

        const timelines = timelineSet.getTimelines();
        const liveTimeline = timelineSet.getLiveTimeline();

        timelines.forEach((timeline, index) => {
            const forwardPaginationToken = timeline.getPaginationToken(EventTimeline.FORWARDS);
            const backwardPaginationToken = timeline.getPaginationToken(EventTimeline.BACKWARDS);

            // Add a special label when it is the live timeline so we can tell
            // it apart from the others
            const isLiveTimeline = timeline === liveTimeline;
            const paginationString = `--f=${forwardPaginationToken}--b=${backwardPaginationToken}`;
            timelineMap[isLiveTimeline ? `${index}-${timeline.toString()}--liveTimeline${paginationString}` : `${index}-${timeline.toString()}${paginationString}`] = timeline.getEvents().map(ev => ev.getId());
        });

        return timelineMap;
    });

    return serializedEventIdsInTimelineSet;
}

console.log('refresh timeline serializeDebugInfoFromTimelineSets', serializeDebugInfoFromTimelineSets(this.timelineSets));
```

---

## ✨ Features
 * Timeline needs to refresh when we see a MSC2716 marker event  ([\#2299](https://github.com/matrix-org/matrix-js-sdk/pull/2299)). Contributed by @MadLittleMods.


src/sync.ts
```diff
@@ -251,6 +264,92 @@ export class SyncApi {
...
+        // timeline or not. The problem is we can't see the prev_events of
+        // the base insertion event that the marker was pointing to because
+        // prev_events aren't available in the client API's. In most cases,
+        // the history won't be in people's locally cached timelines in the
+        // client, so we don't need to bother everyone about refreshing
+        // their timeline. This works for a v1 though and there are use
+        // cases like initially bootstrapping your bridged room where people
+        // are likely to encounter the historical messages affecting their
+        // current timeline (think someone signing up for Beeper and
+        // importing their Whatsapp history).
```
MadLittleMods, line 338: Ideas welcome ^

src/models/event-timeline-set.ts
```diff
@@ -56,6 +56,40 @@ export interface IRoomTimelineData {
...
+    fromCache?: boolean;
+    /** The state events to reconcile metadata from */
+    roomState?: RoomState;
+    /** Whether the timeline was empty before the marker arrived in
+     *  the room. This could be happen in a variety of cases:
+     *  1. From the initial sync
+     *  2. It's the first state we're seeing after joining the room
+     *  3. Or whether it's coming from `syncFromCache` */
+    timelineWasEmpty?: boolean;
+}
```
MadLittleMods, line 77: Instead of dealing with unwieldy long function signatures with unlabeled and optional arguments in the usage, I've refactored some of these into option objects now that I am adding yet another `boolean` here (adding `fromInitialState`).

Before:

```js
room.addLiveEvents(mainTimelineEvents, null, fromCache, timelineWasEmpty);
```

After:

```js
room.addLiveEvents(mainTimelineEvents, {
    fromCache,
    fromInitialState: timelineWasEmpty
});
```

Is it acceptable to make the refactor? Do we have any external API consumers that rely on those function signatures not changing?

If necessary, this refactor can be split out to another PR with some effort.

src/sync.ts
```diff
@@ -251,6 +264,92 @@ export class SyncApi {
...
+     * @param {Room} room The room where the marker event was sent
+     * @param {MatrixEvent} markerEvent The new marker event
+     * @param {ISetStateOptions} setStateOptions When `timelineWasEmpty` is set
+     * as `true`, the given marker event will be ignored
+    */
+    private onMarkerStateEvent(
+        room: Room,
+        markerEvent: MatrixEvent,
+        { timelineWasEmpty }: IMarkerFoundOptions = {},
+    ): void {
```
MadLittleMods, line 283: Calling out one of the main points of interest in the PR (new logic)

src/models/room-state.ts
```diff
@@ -401,6 +415,8 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
 
                 // assume all our sentinels are now out-of-date
                 this.sentinels = {};
+            } else if (event.getType() === EventType.Marker) {
+                this.emit(RoomStateEvent.Marker, event, markerFoundOptions);
```
MadLittleMods, line 419: When we see a new marker state event, emit an event so we can listen for it in `sync.ts` -> `onMarkerStateEvent` and see whether we need to refresh the timeline

src/@types/event.ts
```diff
@@ -33,6 +33,7 @@ export enum EventType {
     RoomGuestAccess = "m.room.guest_access",
     RoomServerAcl = "m.room.server_acl",
     RoomTombstone = "m.room.tombstone",
+    Marker = "org.matrix.msc2716.marker", // MSC2716
```
t3chguy, line 36: This should be using UnstableValue

src/models/event-timeline-set.ts
t3chguy: > Is it acceptable to make the refactor? Do we have any external API consumers that rely on those function signatures not changing?

Yes we do, this is a library used by more than just Element, so this would be a breaking change so please mark it as such (via the issue label) so the release process knows to bump the version. Or better yet, make it backwards compatible via an overload.

src/models/event-timeline-set.ts
```diff
@@ -56,6 +56,41 @@ export interface IRoomTimelineData {
...
+
+export interface IAddEventToTimelineOptions {
+    /** Whether to insert the new event at the start of the timeline */
+    toStartOfTimeline: boolean;
```
MadLittleMods, line 81: `toStartOfTimeline` is ambiguous to what the start is. The timeline could be in chronological or reverse chronological order.

We should clarify whether this adds to the most recent side or the oldest side.

MadLittleMods: It looks like a timeline is in chronological order with the most recent events being at the end.

`toStartOfTimeline` adds to the beginning of the list (index `0`) where the oldest events are.

src/models/event-timeline-set.ts
MadLittleMods: > Or better yet, make it backwards compatible via an overload.

Added some overloads 👍 but it would be good to make sure these get a second set of eyes to make sure these look sane 👀

```
eventTimelineset.addLiveEvent
eventTimelineSet.addEventToTimeline
eventTimeline.addEvent
room.addLiveEvent
room.addLiveEvents
```

src/models/room.ts
```diff
@@ -904,6 +929,71 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
...
+        // available
+        this.fixUpLegacyTimelineFields();
+
+        // The timeline has now been refreshed ✅
+        this.setTimelineNeedsRefresh(false);
+
+        // Emit an event which clients can react to and re-load the timeline
+        // from the SDK
+        this.emit(RoomEvent.TimelineRefresh, this, timelineSet);
+    }
```
MadLittleMods, line 995: What's a decent way to test this? I could write something in the room specs but it would just be asserting a bunch of internal structures which guarantees nothing to how it will behave in a client like Element. (Like checking that the `timelineSet` only holds the new timeline we constructed, has the correct forward pagination token, set as the live timeline, `room.timelineNeedsRefresh = true`, and that it emits events).

Could this be covered by an e2e test in another repo? (see https://github.com/matrix-org/matrix-react-sdk/pull/8354#discussion_r852588155)

src/models/room.ts
```diff
@@ -164,6 +165,8 @@ export enum RoomEvent {
     LocalEchoUpdated = "Room.localEchoUpdated",
     Timeline = "Room.timeline",
     TimelineReset = "Room.timelineReset",
+    TimelineRefresh = "Room.TimelineRefresh",
+    historyImportedWithinTimeline = "Room.historyImportedWithinTimeline",
```
t3chguy, line 169: UpperCamelCase to match please.

src/models/event-timeline-set.ts
```diff
@@ -581,20 +658,46 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
...
-     * @param {boolean} fromCache whether the sync response came from cache
+     * @param {IAddEventToTimelineOptions} options addEventToTimeline options
      *
      * @fires module:client~MatrixClient#event:"Room.timeline"
      */
     public addEventToTimeline(
         event: MatrixEvent,
         timeline: EventTimeline,
         toStartOfTimeline: boolean,
+        fromCache?,
```
t3chguy, line 669: type?

src/models/event-timeline-set.ts
```diff
@@ -581,20 +658,46 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
...
+    public addEventToTimeline(
+        event: MatrixEvent,
+        timeline: EventTimeline,
+        toStartOfTimelineOrOpts: boolean | IAddEventToTimelineOptions,
         fromCache = false,
         roomState?: RoomState,
-    ) {
+    ): void {
+        let toStartOfTimeline = !!toStartOfTimelineOrOpts;
+        let timelineWasEmpty;
```
t3chguy, line 690: type?

src/models/event-timeline-set.ts
```diff
@@ -523,16 +571,40 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
...
+        }: IAddLiveEventOptions,
+    ): void;
+    public addLiveEvent(
+        event: MatrixEvent,
+        duplicateStrategyOrOpts?: DuplicateStrategy | IAddLiveEventOptions,
         fromCache = false,
         roomState?: RoomState,
     ): void {
+        let duplicateStrategy = duplicateStrategyOrOpts as DuplicateStrategy || DuplicateStrategy.Ignore;
+        let timelineWasEmpty;
```
t3chguy, line 599: type?

src/models/event-timeline.ts
```diff
@@ -345,24 +373,49 @@ export class EventTimeline {
...
+            timelineWasEmpty,
+        }: IAddEventOptions,
+    ): void;
+    public addEvent(
+        event: MatrixEvent,
+        toStartOfTimelineOrOpts: boolean | IAddEventOptions,
+        roomState?: RoomState,
+    ): void {
+        let toStartOfTimeline = !!toStartOfTimelineOrOpts;
+        let timelineWasEmpty;
```
t3chguy, line 397: type?

src/models/room-state.ts
```diff
@@ -405,6 +419,8 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
 
                 // assume all our sentinels are now out-of-date
                 this.sentinels = {};
+            } else if (event.getType() === UNSTABLE_MSC2716_MARKER.name) {
```
t3chguy, line 422:
```suggestion
            } else if (UNSTABLE_MSC2716_MARKER.matches(event.getType())) {
```

src/models/room.ts
```diff
@@ -904,6 +929,71 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
...
+        // and doesn't ignore the room events we emit because it checks that the
+        // `timelineSet` references are the same. We need the `timelineSet`
+        // empty so that the `client.getEventTimeline(...)` call later, will
+        // call `/context` and create a new timeline instead of returning the
+        // same one.
+        this.resetLiveTimeline(null, null);
+
+        // Get a reference to the emptied out `timelineSet`
+        //
+        // TODO: Do we want to use `this.getLiveTimeline().getTimelineSet()`
```
t3chguy, line 967: Needs resolving

src/models/room.ts
```diff
@@ -904,6 +929,71 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
...
+        // Use `client.getEventTimeline(...)` to construct a new timeline from a
+        // `/context` response state and events for the most recent event before
+        // we reset everything. The `timelineSet` we pass in needs to be empty
+        // in order for this function to call `/context` and generate a new
+        // timeline.
+        const newTimeline = await this.client.getEventTimeline(timelineSet, mostRecentEventInTimeline.getId());
+        // Set the pagination token back to the live sync token instead of using
+        // the `/context` historical token so that it matches the next response
+        // from `/sync` and we can properly continue the timeline.
+        newTimeline.setPaginationToken(forwardPaginationToken, EventTimeline.FORWARDS);
```
t3chguy, line 980: This seems like it could be problematic, by mixing pagination tokens between timelines you're basically acknowledging that events may be missing or duplicated due to race issues between /sync and /context

src/models/room.ts
```diff
@@ -2174,18 +2325,27 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
...
-    public addLiveEvents(events: MatrixEvent[], duplicateStrategy?: DuplicateStrategy, fromCache = false): void {
+    public addLiveEvents(events: MatrixEvent[], duplicateStrategy?: DuplicateStrategy, fromCache?: boolean): void;
+    public addLiveEvents(events: MatrixEvent[], addLiveEventOptions?: IAddLiveEventOptions): void;
+    public addLiveEvents(
+        events: MatrixEvent[],
+        duplicateStrategyOrOpts?: DuplicateStrategy | IAddLiveEventOptions,
+        fromCache = false,
+    ): void {
+        let duplicateStrategy = duplicateStrategyOrOpts as (DuplicateStrategy | null);
+        let timelineWasEmpty;
```
t3chguy, line 2339: type?

src/models/room.ts
```diff
@@ -1839,15 +1966,35 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
...
-    private addLiveEvent(event: MatrixEvent, duplicateStrategy: DuplicateStrategy, fromCache = false): void {
+    private addLiveEvent(event: MatrixEvent, duplicateStrategy: DuplicateStrategy, fromCache?): void;
+    private addLiveEvent(event: MatrixEvent, addLiveEventOptions: IAddLiveEventOptions): void;
+    private addLiveEvent(
+        event: MatrixEvent,
+        duplicateStrategyOrOpts: DuplicateStrategy | IAddLiveEventOptions,
+        fromCache = false,
+    ): void {
+        let duplicateStrategy = duplicateStrategyOrOpts as DuplicateStrategy;
+        let timelineWasEmpty;
```
t3chguy, line 1981: type?

src/models/room.ts
```diff
@@ -2174,18 +2325,28 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
...
      */
-    public addLiveEvents(events: MatrixEvent[], duplicateStrategy?: DuplicateStrategy, fromCache = false): void {
+    public addLiveEvents(events: MatrixEvent[], duplicateStrategy?: DuplicateStrategy, fromCache?: boolean): void;
+    public addLiveEvents(events: MatrixEvent[], addLiveEventOptions?: IAddLiveEventOptions): void
+    public addLiveEvents(
+        events: MatrixEvent[],
+        duplicateStrategyOrOpts?: DuplicateStrategy | IAddLiveEventOptions,
+        fromCache = false,
+    ): void {
+        let duplicateStrategy = duplicateStrategyOrOpts as (DuplicateStrategy | null);
```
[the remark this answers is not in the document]
t3chguy: `duplicateStrategyOrOpts` is defined as `DuplicateStrategy | IAddLiveEventOptions` - never `null`

src/sync.ts
```diff
@@ -69,14 +70,32 @@ const BUFFER_PERIOD_MS = 80 * 1000;
...
+     */
     Reconnecting = "RECONNECTING",
 }
 
+// Room versions where "insertion", "batch", and "marker" events are controlled
+// by power-levels. MSC2716 is supported in existing room versions but they
+// should only have special meaning when the room creator sends them.
+const MSC2716_ROOM_VERSIONS = [
+    'org.matrix.msc2716v3',
+];
```
t3chguy, line 97: This should really be signalled via the `m.room_versions` capability

src/sync.ts
```diff
@@ -251,6 +274,92 @@ export class SyncApi {
...
+    private onMarkerStateEvent(
+        room: Room,
+        markerEvent: MatrixEvent,
+        { timelineWasEmpty }: IMarkerFoundOptions = {},
+    ): void {
+        // We don't need to refresh the timeline if it was empty before the
+        // marker arrived. This could be happen in a variety of cases:
+        //  1. From the initial sync
+        //  2. If it's from the first state we're seeing after joining the room
+        //  3. Or whether it's coming from `syncFromCache`
```
t3chguy, line 298: not sure how true this is, you'd always see your membership join event in syncs regardless of if its the first one after joining the room

src/models/room.ts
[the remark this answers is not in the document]
MadLittleMods: `duplicateStrategyOrOpts?` is optional which is what I was trying to convey. How do I type that `duplicateStrategy` could also be `undefined`/`null`?

src/models/room.ts
MadLittleMods: I'm looking for advice here. What are your thoughts?

t3chguy: `this.getLiveTimeline().getTimelineSet()` is identical to `this.getUnfilteredTimelineSet()` as the former is defined as `this.getUnfilteredTimelineSet().getLiveTimeline()` and an event-timeline is only part of one event-timeline-set 

So might as well go with the current approach which involves less indirection

src/sync.ts
MadLittleMods: I'm not following. We don't care about member events here and historical state does not affect the current state.

We only care whether it's a new marker event that we're seeing. And if marker is part of the first timeline chunk that was fetched, then the timeline doesn't need to be refreshed (it's already new).

src/sync.ts
MadLittleMods: Good shout 👍 Will create a Synapse PR to add it

src/models/room.ts
MadLittleMods: We could add some better merging logic here. If after `getEventTimeline` runs, we see another timeline (indicating that a sync happened in between), we can skip setting the forward pagination token and setting it as the live timeline.

We can rely on the latest sync and since we cleared out the other timelines, it will paginate back from the latest sync.

Definitely want to add some tests for these scenarios, https://github.com/matrix-org/matrix-react-sdk/pull/8354#discussion_r852588155

src/models/room.ts
MadLittleMods: @t3chguy Any thoughts here to move this forward?

t3chguy: Yeah I think for something with quite a few moving pieces, would be great to have an e2e test for it simply due to how easy to break it could be

src/sync.ts
MadLittleMods: Synapse PR: https://github.com/matrix-org/synapse/pull/12516

MadLittleMods: I think we may have to opt for this static list for now. The consensus on the PR is not to stack the MSC behaviors, see https://github.com/matrix-org/synapse/pull/12516#pullrequestreview-948231079 ([MSC3244](https://github.com/matrix-org/matrix-doc/pull/3244) to add `room_capabilities` hasn't landed yet)

src/models/room.ts
MadLittleMods: Added Cypress e2e tests in https://github.com/matrix-org/matrix-react-sdk/pull/8354

src/sync.ts
```diff
@@ -252,6 +275,92 @@ export class SyncApi {
...
+                `by the room creator.`,
+            );
+        }
+
+        // Don't process a marker event multiple times; we only need to
+        // throw the timeline away once, when we see a marker. All of the
+        // historical content will be in the `/messsages` responses from
+        // here on out.
+        const markerAlreadyProcessed = markerEvent.getId() === room.getLastMarkerEventIdProcessed();
+        if (markerAlreadyProcessed) {
```
MadLittleMods, line 331: We moved to marker events as state with each one having a unique `state_key` so they all appear readily to homeservers/clients without getting lost in timeline gaps. This means there will be multiple marker events in the state at a time.

TODO: Add a test to make sure this plays well with multiple markers and we don't just get in a loop of processing them back and fourth as they invalidate each other. In other words, does `onMarkerStateEvent` only get called for new marker state, or all marker events from any time we see it (like when we jump to a permalink with `/context` for example)?

I think `setStateEvents` (which emits `RoomStateEvent.Marker` -> `onMarkerStateEvent`) currently doesn't discriminate between new vs old state we already have based on the comment doc, "Add an array of one or more state MatrixEvents, overwriting any existing state with the same {type, stateKey} tuple." So we do have to worry about this.

Ideally, we could adjust `setStateEvents` to store the state before applying the new state, then we can compare and only emit `RoomStateEvent.Marker` if we didn't already have it. This way, we can also get rid of the `getLastMarkerEventIdProcessed` stuff.

Otherwise, we have to do keep track of things we already processed differently which doesn't sound fun. List of every marker event, using `origin_server_ts` which is flawed, etc.

src/models/room.ts
MadLittleMods: In https://github.com/matrix-org/matrix-react-sdk/pull/8354, we have an e2e Cypress test for a `/sync` racing between the `this.resetLiveTimeline(null, null);` and when the `client.getEventTimeline()` `/context` request completes. Everything seems to behave as expected ✅

The live timeline has a forward pagination token of `null` so it merges perfectly without issue.

Here is a breakdown of the timelines:
```
// After `this.resetLiveTimeline(null, null);`
[
    {
        "0-!OIGyXdhYVmWEAByyII:localhost:2022-05-26T02:39:10.737Z--liveTimeline--f=null--b=null": []
    }
]

// After racey `/sync` completes
[
    {
        "0-!OIGyXdhYVmWEAByyII:localhost:2022-05-26T02:39:10.737Z--liveTimeline--f=null--b=null": [
            "$LpjtBslib3J0kPzL1J8EbiAg7mNCNQfHoki7nh9ejH8"
        ]
    }
]

// After we use `client.getEventTimeline(...)` to construct a new timeline from a `/context` response
[
    {
        "0-!OIGyXdhYVmWEAByyII:localhost:2022-05-26T02:39:10.737Z--liveTimeline--f=null--b=t10--6_0_0_0_0_0_0_0_0": [
            "$8-8QjB-U0BRFcmpY39FkMpdt4-R8ux5XbHL0B0LXZ00",
            "$7DrVapInIVSKaM8MWQZDi15_ruTIUQfyY4grQu-NOiw",
            "$bFyHxdn9Y8wosMxnvA-HdXaGSHpT1VridJnGUvsBGWU",
            "$wvoRm9JI5uMeWxVm49mny2YtthfYKdEAZ9g3-gbeCJc",
            "$kOFxIReYWJK-M7JmNyFGINkq4Y7FBlD3jMEFzk1GqNg",
            "$3ZcZZe9MKr35wVVkI-1HoGem-WRdyEtOMJttCQq9QUU",
            "$LpjtBslib3J0kPzL1J8EbiAg7mNCNQfHoki7nh9ejH8"
        ]
    }
]

// After things settle
[
    {
        "0-!OIGyXdhYVmWEAByyII:localhost:2022-05-26T02:39:10.737Z--liveTimeline--f=null--b=t10--6_0_0_0_0_0_0_0_0": [
            "$8-8QjB-U0BRFcmpY39FkMpdt4-R8ux5XbHL0B0LXZ00",
            "$7DrVapInIVSKaM8MWQZDi15_ruTIUQfyY4grQu-NOiw",
            "$bFyHxdn9Y8wosMxnvA-HdXaGSHpT1VridJnGUvsBGWU",
            "$wvoRm9JI5uMeWxVm49mny2YtthfYKdEAZ9g3-gbeCJc",
            "$kOFxIReYWJK-M7JmNyFGINkq4Y7FBlD3jMEFzk1GqNg",
            "$3ZcZZe9MKr35wVVkI-1HoGem-WRdyEtOMJttCQq9QUU",
            "$LpjtBslib3J0kPzL1J8EbiAg7mNCNQfHoki7nh9ejH8"
        ]
    }
]
```

MadLittleMods: Let me know if there are other scenarios you are particularly worried about and we can add an e2e test for them.

src/sync.ts
MadLittleMods: > I think `setStateEvents` (which emits `RoomStateEvent.Marker` -> `onMarkerStateEvent`) currently doesn't discriminate between new vs old state

We do discriminate when `timelineWasEmpty` and ignore those which seems to do exactly what we need.

In all of my tests, we don't ever trigger this scenario and now looking at it, I'm not sure how we would. I'm going to remove for now. If we ever need it again, we'll understand why and can write a test for it ⏩

src/models/room.ts
```diff
@@ -933,6 +1051,10 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
             .getState(EventTimeline.BACKWARDS);
         this.currentState = this.getLiveTimeline()
             .getState(EventTimeline.FORWARDS);
+
+        // Let people know to register listeners for the new state references
+        this.emit(RoomEvent.OldStateUpdated, this, previousOldState, this.currentState);
+        this.emit(RoomEvent.CurrentStateUpdated, this, previousCurrentState, this.currentState);
```
MadLittleMods, line 1057: Calling out this new behavior. See how it's used in `sync.ts` to re-register the state listeners.

src/models/room.ts
```diff
@@ -933,6 +1051,10 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
             .getState(EventTimeline.BACKWARDS);
         this.currentState = this.getLiveTimeline()
             .getState(EventTimeline.FORWARDS);
+
+        // Let people know to register listeners for the new state references
+        this.emit(RoomEvent.OldStateUpdated, this, previousOldState, this.currentState);
```
MadLittleMods, line 1056: `RoomEvent.OldStateUpdated` is not used but seemed applicable to add the equivalent to `RoomEvent.CurrentStateUpdated` since people could be doing the same thing with `oldState` for whatever reason. Also willing to remove since we aren't exercising it ourselves.

src/sync.ts
```diff
@@ -1254,7 +1356,6 @@ export class SyncApi {
                 }
 
                 if (limited) {
-                    this.deregisterStateListeners(room);
```
MadLittleMods, line 1257: Removing the deregister and register here because it can now be handled automatically in one spot by `RoomEvent.CurrentStateUpdated` (see above)

src/sync.ts
```diff
@@ -209,6 +228,15 @@ export class SyncApi {
...
         this.registerStateListeners(room);
+        // Register listeners again after the state reference changes
+        room.on(RoomEvent.CurrentStateUpdated, (targetRoom, previousCurrentState) => {
+            if (targetRoom !== room) {
+                return;
+            }
+
+            this.deregisterStateListeners(previousCurrentState);
+            this.registerStateListeners(room);
+        });
```
MadLittleMods, line 239: Calling out this new behavior.

This is necessary because `room.fixUpLegacyTimelineFields()` (and `room.resetLiveTimeline()` which calls that function) will change the `room.currentState` reference and we will fail listen to any new state events.

The one place we deregistered and registered before was for a `limited` `/sync` response but that was only possible because that is handled in `sync.ts` (see below in this file https://github.com/matrix-org/matrix-js-sdk/pull/2299#discussion_r882335575)

Now we're also calling `fixUpLegacyTimelineFields` inside of `room.ts` so we need to make sure any consumers of the room know about the reference change.

t3chguy (changes requested):
> Failed condition 66.9% 66.9% Coverage on New Code (is less than 80%)

src/models/room.ts
```diff
@@ -933,6 +1051,17 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
             .getState(EventTimeline.BACKWARDS);
         this.currentState = this.getLiveTimeline()
             .getState(EventTimeline.FORWARDS);
+
+        // Let people know to register new listeners for the new state
+        // references. The reference won't necessarily change every time so only
+        // emit when we see a change.
+        if (previousOldState !== this.oldState) {
+            this.emit(RoomEvent.OldStateUpdated, this, previousOldState, this.oldState);
```
MadLittleMods, line 1059: `RoomEvent.OldStateUpdated` is not used but seemed applicable to add the equivalent to `RoomEvent.CurrentStateUpdated` since people could be doing the same thing with `oldState` for whatever reason. Also willing to remove since we aren't exercising it ourselves.

src/models/room.ts
MadLittleMods: Re-commented on latest diff: https://github.com/matrix-org/matrix-js-sdk/pull/2299/files#r883240213

src/sync.ts
```diff
@@ -51,10 +51,11 @@ import { MatrixError, Method } from "./http-api";
 import { ISavedSync } from "./store";
 import { EventType } from "./@types/event";
 import { IPushRules } from "./@types/PushRules";
-import { RoomStateEvent } from "./models/room-state";
+import { RoomState, RoomStateEvent } from "./models/room-state";
 import { RoomMemberEvent } from "./models/room-member";
 import { BeaconEvent } from "./models/beacon";
 import { IEventsResponse } from "./@types/requests";
+import { IMarkerFoundOptions } from "./models/room-state";
```
t3chguy, line 58: Can these not be in the existing import statement? (Sonar smell)

src/models/room.ts
```diff
@@ -49,6 +49,7 @@ import {
 import { TypedEventEmitter } from "./typed-event-emitter";
 import { ReceiptType } from "../@types/read_receipts";
 import { IStateEventWithRoomId } from "../@types/search";
+import { IAddLiveEventOptions } from "./event-timeline-set";
```
t3chguy, line 52: Can this please go in the existing event-timeline-set import, it is creating a Sonar smell

spec/integ/matrix-client-syncing.spec.js
```diff
@@ -582,6 +583,477 @@ describe("MatrixClient syncing", function() {
...
+                                    // should trigger `timelineNeedsRefresh=true`
+                                    markerEventFromRoomCreator,
+                                ],
+                            },
+                        };
+
+                        httpBackend.when("GET", "/sync").respond(200, normalFirstSync);
+                        httpBackend.when("GET", "/sync").respond(200, nextSyncData);
+
+                        client.startClient();
```
t3chguy, line 879: all these clients also need stopping to not continue running after the test is done (this upsets future versions of Jest)

src/sync.ts
t3chguy: This feels like a breaking change, all downstream clients will need updating if any hold references to the currentState

src/sync.ts
```diff
@@ -241,17 +269,91 @@ export class SyncApi {
...
+        // the base insertion event that the marker was pointing to because
+        // prev_events aren't available in the client API's. In most cases,
+        // the history won't be in people's locally cached timelines in the
+        // client, so we don't need to bother everyone about refreshing
+        // their timeline. This works for a v1 though and there are use
+        // cases like initially bootstrapping your bridged room where people
+        // are likely to encounter the historical messages affecting their
+        // current timeline (think someone signing up for Beeper and
+        // importing their Whatsapp history).
+        if (isValidMsc2716Event) {
```
t3chguy, line 347: Can we do if-else or if-return, rather than `if (foo)` and `if(!foo)`

src/client.ts
```diff
@@ -5361,6 +5364,48 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+     *
+     * @param {EventTimelineSet} timelineSet  The timelineSet to look for the
+     * event in
+     * @param {string} eventId  The ID of the event to look for
+     *
+     * @return {Promise} Resolves:
+     *    {@link module:models/event-timeline~EventTimeline} including the given
+     *    event
+     */
+    public async getLatestTimeline(timelineSet: EventTimelineSet): Promise<EventTimeline> {
```
t3chguy, line 5380: This feels like a racey concept, whilst you're waiting for the response from /context new messages may have come in which now won't be on the tip of this timeline.

src/models/event-timeline-set.ts
```diff
@@ -55,6 +55,43 @@ export interface IRoomTimelineData {
...
+    roomState?: RoomState;
+    /** Whether the timeline was empty before the marker arrived in
+     *  the room. This could be happen in a variety of cases:
+     *  1. From the initial sync
+     *  2. It's the first state we're seeing after joining the room
+     *  3. Or whether it's coming from `syncFromCache` */
+    timelineWasEmpty?: boolean;
+}
+
+export interface IAddEventToTimelineOptions {
```
t3chguy, line 78: A lot of options are shared between this and the interface above, maybe we can have a base `IAddEventOptions` (unexported) just to save having to duplicate all those detailed comments (and someone's mental understanding)

src/models/event-timeline-set.ts
```diff
@@ -55,6 +55,43 @@ export interface IRoomTimelineData {
...
+     * the existing event in the timeline. If this is not specified, or is
+     * 'ignore', then the event passed to this function will be ignored
+     * entirely, preserving the existing event in the timeline. Events are
+     * identical based on their event ID <b>only</b>. */
+    duplicateStrategy?: DuplicateStrategy;
+    /** Whether the sync response came from cache */
+    fromCache?: boolean;
+    /** The state events to reconcile metadata from */
+    roomState?: RoomState;
+    /** Whether the timeline was empty before the marker arrived in
```
t3chguy, line 70: This comment (and same below) doesn't make sense without explaining `marker`

src/models/event-timeline-set.ts
```diff
@@ -180,6 +217,15 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
...
     }
 
+    /**
+     * Set the live timeline for this room.
+     *
+     * @return {module:models/event-timeline~EventTimeline} live timeline
+     */
+    public setLiveTimeline(timeline: EventTimeline): void {
+        this.liveTimeline = timeline;
+    }
```
t3chguy, line 227: How are downstream consumers which may be holding a reference to live timeline meant to know it got updated from the context of an EventTimelineSet (feels breaking)

src/models/event-timeline-set.ts
```diff
@@ -522,16 +570,40 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
      * Add an event to the end of this live timeline.
      *
      * @param {MatrixEvent} event Event to be added
-     * @param {string?} duplicateStrategy 'ignore' or 'replace'
-     * @param {boolean} fromCache whether the sync response came from cache
+     * @param {IAddLiveEventOptions} options addLiveEvent options
      * @param roomState the state events to reconcile metadata from
      */
     public addLiveEvent(
```
t3chguy, line 576: It'd be grand to get a test for this overload, they can be easy to break

src/models/event-timeline-set.ts
```diff
@@ -580,20 +657,46 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
      *
      * @param {MatrixEvent} event
      * @param {EventTimeline} timeline
-     * @param {boolean} toStartOfTimeline
-     * @param {boolean} fromCache whether the sync response came from cache
+     * @param {IAddEventToTimelineOptions} options addEventToTimeline options
      *
      * @fires module:client~MatrixClient#event:"Room.timeline"
      */
     public addEventToTimeline(
```
t3chguy, line 664: Would be great to get a test for this overload, they are easy to break

src/models/event-timeline.ts
```diff
@@ -24,6 +24,30 @@ import { MatrixEvent } from "./event";
...
+export interface IInitialiseStateOptions {
+    /** Whether the timeline was empty before the marker arrived in
+     *  the room. This could be happen in a variety of cases:
+     *  1. From the initial sync
+     *  2. It's the first state we're seeing after joining the room
+     *  3. Or whether it's coming from `syncFromCache` */
+    timelineWasEmpty?: boolean;
+}
+
+export interface IAddEventOptions {
```
t3chguy, line 36: This looks like a good base interface for the two lengthier ones in event-timeline-set

src/models/event-timeline.ts
```diff
@@ -24,6 +24,30 @@ import { MatrixEvent } from "./event";
 import { Filter } from "../filter";
 import { EventType } from "../@types/event";
 
+export interface IInitialiseStateOptions {
+    /** Whether the timeline was empty before the marker arrived in
+     *  the room. This could be happen in a variety of cases:
+     *  1. From the initial sync
+     *  2. It's the first state we're seeing after joining the room
+     *  3. Or whether it's coming from `syncFromCache` */
```
t3chguy, line 32: We have 4 copies of this exact comment, and its kinda vague to the reader `the marker` doesn't mean much in isolation.

src/models/event-timeline.ts
```diff
@@ -345,24 +373,49 @@ export class EventTimeline {
      * Add a new event to the timeline, and update the state
      *
      * @param {MatrixEvent} event   new event
-     * @param {boolean}  atStart     true to insert new event at the start
+     * @param {IAddEventOptions} options addEvent options
      */
-    public addEvent(event: MatrixEvent, atStart: boolean, stateContext?: RoomState): void {
-        if (!stateContext) {
-            stateContext = atStart ? this.startState : this.endState;
+    public addEvent(
```
t3chguy, line 378: Would be great to get a test for this overload, they're high-debt

src/models/room-state.ts
```diff
@@ -30,6 +30,15 @@ import { Beacon, BeaconEvent, BeaconEventHandlerMap, getBeaconInfoIdentifier, Be
 import { TypedReEmitter } from "../ReEmitter";
 import { M_BEACON, M_BEACON_INFO } from "../@types/beacon";
 
+export interface IMarkerFoundOptions {
+    /** Whether the timeline was empty before the marker arrived in
```
t3chguy, line 34: I take it back, there's 5 copies of this comment, and this is the first place that `the marker` makes sense. Can we please somehow DRY this to prevent 5 comment copies from falling out of sync.

t3chguy: Remember an interface can extend many interfaces, so you can assemble the bigger ones really nicely

src/models/room.ts
```diff
@@ -441,6 +458,19 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
...
+     * Gets the creator of the room
+     * @returns {string} The creator of the room, or null if it could not be determined
+     */
+    public getCreator(): string | null {
+        const createEvent = this.currentState.getStateEvents(EventType.RoomCreate, "");
+        if (!createEvent) {
+            return null;
+        }
+        const roomCreator = createEvent.getContent()['creator'];
+        return roomCreator;
```
t3chguy, line 471: This technically doesn't match its signature, creator could be `undefined`

```suggestion
        return createEvent?.getContent()['creator'] ?? null;
```

src/models/room.ts
```diff
@@ -897,6 +927,91 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
...
+            `mostRecentEventInTimeline=${mostRecentEventInTimeline && mostRecentEventInTimeline.getId()} ` +
+            `liveTimelineBefore=${liveTimelineBefore.toString()} ` +
+            `forwardPaginationToken=${forwardPaginationToken} ` +
+            `backwardPaginationToken=${backwardPaginationToken}`,
+        );
+
+        // Get the main TimelineSet
+        const timelineSet = this.getUnfilteredTimelineSet();
+
+        let newTimeline;
```
t3chguy, line 960: Type this please

src/models/room.ts
```diff
@@ -1778,15 +1925,35 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
...
      *
      * @param {MatrixEvent} event Event to be added
-     * @param {string?} duplicateStrategy 'ignore' or 'replace'
-     * @param {boolean} fromCache whether the sync response came from cache
+     * @param {IAddLiveEventOptions} options addLiveEvent options
      * @fires module:client~MatrixClient#event:"Room.timeline"
      * @private
      */
-    private addLiveEvent(event: MatrixEvent, duplicateStrategy: DuplicateStrategy, fromCache = false): void {
+    private addLiveEvent(event: MatrixEvent, duplicateStrategy: DuplicateStrategy, fromCache?): void;
```
t3chguy, line 1932: Why do we need an overload on a private internal method? We have no need to maintain the old contract here

src/models/room.ts
```diff
@@ -2113,18 +2284,27 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
...
-     * timeline. If this is not specified, or is 'ignore', then the event passed to
-     * this function will be ignored entirely, preserving the existing event in the
-     * timeline. Events are identical based on their event ID <b>only</b>.
-     *
-     * @param {boolean} fromCache whether the sync response came from cache
+     * @param {IAddLiveEventOptions} options addLiveEvent options
      * @throws If <code>duplicateStrategy</code> is not falsey, 'replace' or 'ignore'.
      */
-    public addLiveEvents(events: MatrixEvent[], duplicateStrategy?: DuplicateStrategy, fromCache = false): void {
+    public addLiveEvents(events: MatrixEvent[], duplicateStrategy?: DuplicateStrategy, fromCache?: boolean): void;
```
t3chguy, line 2290: Can we get a test exercising the overload please

t3chguy (commented):
Thanks for the additional test coverage!

src/sync.ts
MadLittleMods: @t3chguy All existing functionality still works.

Clients who choose to implement timeline refreshing will also need to take `room.currentState` changing though.

src/sync.ts
MadLittleMods: Yes! 👍 I wish there was a lint for this

src/client.ts
MadLittleMods: We should continue this conversation in the other thread, https://github.com/matrix-org/matrix-js-sdk/pull/2299#discussion_r852891150

If you have certain scenarios you're worried about, I can add a test for it.

src/models/event-timeline-set.ts
MadLittleMods: Clients which choose to implement timeline refreshing can also listen for `RoomEvent.TimelineRefresh`.

For example, Element listens for `RoomEvent.TimelineRefresh` and tells `TimelineWindow` to load again which picks up from the same `timelineSet`

src/models/room.ts
MadLittleMods: Good shout, removed the old signatures 👍

src/models/event-timeline-set.ts
MadLittleMods: How far do we want to go with these overload tests? I can add a simple test making sure it doesn't throw an error in some specific scenarios.

But if we want to make sure both method signatures behave the same, we will need to run all of the tests against both overloads.

```js
function addLiveEventOverloadWithOptions(event, opts: IAddLiveEventOptions) {
  return room.addLiveEvent(event, opts);
}

function addLiveEventOverloadWithDirectParameters(event, opts: IAddLiveEventOptions) {
  return room.addLiveEvent(event, opts.duplicateStrategy, opts.duplicateStrategy, opts.duplicateStrategy);
}

[addLiveEventOverloadWithOptions, addLiveEventOverloadWithDirectParameters].forEach((addLiveEvent) => {
  it('foo', () => {
     addLiveEvent(event, {
       duplicateStrategy: 'replace',
       timelineWasEmpty: true,
     });
  });
});

```

src/models/event-timeline-set.ts
```diff
@@ -522,16 +570,40 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
...
+     * @param {IAddLiveEventOptions} options addLiveEvent options
      * @param roomState the state events to reconcile metadata from
      */
     public addLiveEvent(
         event: MatrixEvent,
-        duplicateStrategy: DuplicateStrategy = DuplicateStrategy.Ignore,
+        duplicateStrategy?: DuplicateStrategy,
+        fromCache?: boolean,
+        roomState?: RoomState,
+    ): void;
```
MadLittleMods, line 581: Should we add deprecation warnings for these old overloads now? This way we have some pathway to cleaning up this technical debt in a year or two.

spec/integ/matrix-client-syncing.spec.js
MadLittleMods: Should be happening in the root-level `afterEach`,

https://github.com/matrix-org/matrix-js-sdk/blob/c36e1b9e8d3773db2861ad97fac7790bbb8523b7/spec/integ/matrix-client-syncing.spec.js#L43-L47

src/sync.ts
MadLittleMods: 👍 Makes sense now. Previously there was some extra `markerAlreadyProcessed` logic in this if-statement.

src/models/event-timeline-set.ts
MadLittleMods: Added an extra explanation of a `marker` in the `IMarkerFoundOptions` options 👍

src/models/room-state.ts
MadLittleMods: I first tried drying things out by a [naive extending of everything](https://github.com/matrix-org/matrix-js-sdk/pull/2299/commits/58fa63c27a21aa2721b2c178caa7b2ea68cfca77) but the nature of this kinda sucks to follow (OOP hell).

I [updated it to use `extends Pick<Type, Keys>`](https://github.com/matrix-org/matrix-js-sdk/pull/2299/commits/d0466a1605d4ed68d49db5c4157c608a97137db0) instead. This way, we can better compose stuff together and share the comment from a single place and it's obvious which properties are defined for each interface (no need to follow the chain across files). This also avoids the problem of stuff unintentionally cascading across interfaces which have disparate concerns when we add properties in the future.

Seem reasonable?

src/models/event-timeline-set.ts
MadLittleMods: More context in https://github.com/matrix-org/matrix-js-sdk/pull/2299#discussion_r883502814

src/models/event-timeline.ts
MadLittleMods: More context in https://github.com/matrix-org/matrix-js-sdk/pull/2299#discussion_r883502814

src/models/event-timeline.ts
MadLittleMods: More context in https://github.com/matrix-org/matrix-js-sdk/pull/2299#discussion_r883502814

Addressed vague `marker` in https://github.com/matrix-org/matrix-js-sdk/pull/2299#discussion_r883499258

src/models/event-timeline-set.ts
t3chguy: Good idea

src/models/event-timeline-set.ts
t3chguy: Ideally we'd implement overloads as wrappers, then we could mock the wrapped function and validate that all overloads result in the same call to the wrapped method, but this is not how the SDK does existing overloads so anything is better than no coverage

src/models/room-state.ts
t3chguy: Yup, thanks!

src/models/room.ts
t3chguy: It'd be better to have a test at the js-sdk layer for this, given that a change in the js-sdk being merged should not cause cascading failures to downstream consumers

t3chguy (changes requested):
This is getting very close now

Remaining things:
https://github.com/matrix-org/matrix-js-sdk/pull/2299#discussion_r852891150

src/sync.ts
MadLittleMods: > We can also iterate this in another PR.

Opting for this option ⏩. This is a nice to have.

src/models/event-timeline-set.ts
MadLittleMods: Done ✅ https://github.com/matrix-org/matrix-js-sdk/pull/2299/commits/e4bd5188735e614a814f19ce4fa616243b052481

 1. Moved our preferred overload to the top so it's suggested first
 1. Added deprecation warnings in usage via `logger.warning`
 1. I've also added [`@deprecated`](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#deprecated) TypeScript decorations

src/models/room.ts
MadLittleMods: 👍 Added tests in the `matrix-js-sdk` and fixed up the merging scenario which is tested harder, https://github.com/matrix-org/matrix-js-sdk/pull/2299#discussion_r886552876

t3chguy (approved):
:shipit: thanks for perservering with this!

src/client.ts
```diff
@@ -5374,6 +5377,48 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+     * call `/messages` to get the latest message in the room, then use
+     * `client.getEventTimeline(...)` to construct a new timeline from it.
+     *
+     * @param {EventTimelineSet} timelineSet  The timelineSet to look for the
+     * event in
+     * @param {string} eventId  The ID of the event to look for
+     *
+     * @return {Promise} Resolves:
+     *    {@link module:models/event-timeline~EventTimeline} including the given
+     *    event
```
MadLittleMods, line 5391:
```suggestion
     *    {@link module:models/event-timeline~EventTimeline} timeline with the latest events in the room
```

spec/unit/room-state.spec.js
```diff
@@ -258,6 +258,29 @@ describe("RoomState", function() {
             );
         });
 
+        it("should emit 'RoomStateEvent.Marker' for each marker event", function() {
```
MadLittleMods, line 261:
```suggestion
        it("should emit `RoomStateEvent.Marker` for each marker event", function() {
```

spec/integ/matrix-client-event-timeline.spec.js
```diff
@@ -540,6 +540,77 @@ describe("MatrixClient event timelines", function() {
...
+                        ],
+                        end: "end_token",
+                    };
+                });
+
+            return Promise.all([
+                client.getLatestTimeline(timelineSet).then(function(tl) {
+                    // Instead of this assertion logic, we could just add a spy
+                    // for `getEventTimeline` and make sure it's called with the
+                    // correct parameters. This doesn't feel to bad to make sure
```
MadLittleMods, line 578:
```suggestion
                    // correct parameters. This doesn't feel too bad to make sure
```

src/client.ts
```diff
@@ -5374,6 +5377,48 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
 
+    /**
+     * Get an EventTimeline for the latest events in the room. This will just
+     * call `/messages` to get the latest message in the room, then use
+     * `client.getEventTimeline(...)` to construct a new timeline from it.
+     *
+     * @param {EventTimelineSet} timelineSet  The timelineSet to look for the
+     * event in
+     * @param {string} eventId  The ID of the event to look for
+     *
```
MadLittleMods, line 5388:
```suggestion
     * @param {EventTimelineSet} timelineSet  The timelineSet to find or add the timeline to
     *
```

MadLittleMods:
Thanks for the all of the review @t3chguy! 🐝 Lots of strain to get it up to snuff but came out with a lot more polish and more confidence in this new timeline concept working!

PR #2301, opened 2022-04-14
fix getUrlPreview typeError

skyka13711:
When using a strict mod, this line causes a typeError

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Read-only

---


t3chguy (changes requested):
https://developer.mozilla.org/en-US/docs/Web/API/URL/hash does not document it to be read-only. Neither does https://url.spec.whatwg.org/#dom-url-hash

t3chguy:
Closing as a non-issue per the MDN doc, if you have a more complete example for why this is useful do let us know

PR #2311, opened 2022-04-20
Add PTT call mode & mute by default in PTT calls

dbkr:
No other parts of PTT calls implemented yet

For https://github.com/vector-im/element-call/issues/294

---

## ✨ Features
 * Add PTT call mode & mute by default in PTT calls ([\#2311](https://github.com/matrix-org/matrix-js-sdk/pull/2311)).


gsouquet (approved):
Looks good to me 👍

PR #2312, opened 2022-04-21
Add maximum trasmit time for PTT

dbkr:
on sender side by muting mic after the max transmit time has elapsed.

For https://github.com/vector-im/element-call/issues/297

---
Here's what your changelog entry will look like:

## ✨ Features
 * Add maximum trasmit time for PTT ([\#2312](https://github.com/matrix-org/matrix-js-sdk/pull/2312)).

PR #2313, opened 2022-04-21
Don't allow user to unmute if another user is speaking 

dbkr:
Based on https://github.com/matrix-org/matrix-js-sdk/pull/2312
For https://github.com/vector-im/element-call/issues/298

---

## ✨ Features
 * Don't allow user to unmute if another user is speaking  ([\#2313](https://github.com/matrix-org/matrix-js-sdk/pull/2313)).

### DIFF
--- a/src/webrtc/groupCall.ts
+++ b/src/webrtc/groupCall.ts
@@ -112,6 +112,7 @@ export class GroupCall extends EventEmitter {
     public activeSpeakerInterval = 1000;
     public retryCallInterval = 5000;
     public participantTimeout = 1000 * 15;
+    public pttMaxTransmitTime = 1000 * 20;
 
     public state = GroupCallState.LocalCallFeedUninitialized;
     public activeSpeaker?: string; // userId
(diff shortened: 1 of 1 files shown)


gsouquet (approved):

src/webrtc/groupCall.ts
```diff
@@ -408,6 +414,12 @@ export class GroupCall extends EventEmitter {
 
         // set a timer for the maximum transmit time on PTT calls
         if (this.isPtt) {
+            // if anoher user is currently unmuted, we can't unmute
+            if (!muted && this.userMediaFeeds.some(f => !f.isAudioMuted())) {
+                throw new OtherUserSpeakingError();
+            }
```
MadLittleMods, line 420: This seems susceptible to ping response times? If multiple people unmute at the same time, they all would be allowed.

src/webrtc/groupCall.ts
```diff
@@ -408,6 +414,12 @@ export class GroupCall extends EventEmitter {
 
         // set a timer for the maximum transmit time on PTT calls
         if (this.isPtt) {
+            // if anoher user is currently unmuted, we can't unmute
```
MadLittleMods, line 417: This seems a bit weird to force on everyone.

If I was using this for game callouts, I would want to interrupt the other person.

PR #2314, opened 2022-04-21
Don't decrement the length count of a thread when root redacted

t3chguy:
The count is number of replies, in which the root is not included


kerryarchibald (approved):
test?

PR #2319, opened 2022-04-25
De-duplicate code and fix types

t3chguy:
---

### DIFF
--- a/spec/unit/room.spec.ts
+++ b/spec/unit/room.spec.ts
@@ -1471,16 +1471,13 @@ describe("Room", function() {
                 isRoomEncrypted: function() {
                     return false;
                 },
-                http: {
-                    serverResponse,
-                    authedRequest: function() {
-                        if (this.serverResponse instanceof Error) {
-                            return Promise.reject(this.serverResponse);
-                        } else {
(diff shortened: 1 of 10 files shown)


kerryarchibald (approved):
LGTM

YousefED:
I don't think the change in types for `members()` parameters in this PR is correct.

```
includeMembership?: string,
excludeMembership?: string,
```
should both still accept a `string[]`, because we can filter multiple membership types. If the expected input is a comma-separated list, this should be documented? (but afaik passing in an array works)

t3chguy:
@YousefED https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3roomsroomidmembers says the field takes a single enum value, not an array of values

YousefED:
> @YousefED https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3roomsroomidmembers says the field takes a single enum value, not an array of values - if Synapse supports an array then the spec needs to consider updating to document that

Damn, you're right! My bad. I was probably confused by the previous (wrong) signature though. I think the spec is a bit weird (why would you need an OR between the two parameters), but it seems like Synapse is also built to handle it this way.

I'll fix my application instead. Appreciate your reply

PR #2322, opened 2022-04-26
Add support for sending encrypted to-device events with OLM

robertlong:
This PR encrypts to-device events used in GroupCall signalling.

---

## ✨ Features
 * Add support for sending encrypted to-device events with OLM ([\#2322](https://github.com/matrix-org/matrix-js-sdk/pull/2322)). Contributed by @robertlong.

### DIFF
--- a/src/crypto/index.ts
+++ b/src/crypto/index.ts
@@ -3036,7 +3036,7 @@ export class Crypto extends EventEmitter {
      * set of devices.  Factored out from encryptAndSendKeysToDevices in
      * megolm.ts.
      *
-     * @param {object<userId, deviceInfo>} userDeviceMap
+     * @param {object[]} userDeviceInfoArr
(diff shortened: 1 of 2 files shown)


t3chguy:
Would be good to get a description in the PR why we're breaking public interfaces, what does it gain us

t3chguy (approved):
I think the PR title is unclear, the sdk obviously already supports sending encrypted to-device events with olm, this is specific to call signalling. The PR titles go into the changelog, please fix

LGTM otherwise

dbkr (approved):

dbkr:
looks good, although also in the title, Olm isn't an acronym (https://en.wikipedia.org/wiki/Olm).

t3chguy:
@dbkr does this want to be part of https://github.com/matrix-org/matrix-js-sdk/pull/2355 ?

PR #2330, opened 2022-04-29
Try SonarCloud

t3chguy:
---


dbkr (approved):
Exciting - I guess the problem if forked PRs still applies, but we can try it for non-forked ones for now.

PR #2333, opened 2022-04-29
Don't notify on room server ACL changes

SimonBrandner:
Fixes https://github.com/vector-im/element-web/issues/20788
Type: defect

### DIFF
--- a/src/pushprocessor.ts
+++ b/src/pushprocessor.ts
@@ -34,6 +34,7 @@ import {
     PushRuleSet,
     TweakName,
 } from "./@types/PushRules";
+import { EventType } from "./@types/event";
 
 /**
  * @module pushprocessor
@@ -90,6 +91,22 @@ const DEFAULT_OVERRIDE_RULES: IPushRule[] = [
                 kind: ConditionKind.EventMatch,
                 key: "type",
                 pattern: "m.reaction",
+            },
+        ],
+        actions: [
+            PushRuleActionName.DontNotify,
+        ],
+    },
+    {
+        // Make sure we don't notify on ACL changes - https://github.com/vector-im/element-web/issues/20788
+        rule_id: ".m.rule.server_acl",
+        default: true,
+        enabled: true,
+        conditions: [
+            {
+                kind: ConditionKind.EventMatch,
+                key: "type",
+                pattern: EventType.RoomServerAcl,
             },
         ],
         actions: [
--- a/spec/unit/pushprocessor.spec.js
+++ b/spec/unit/pushprocessor.spec.js
@@ -1,5 +1,6 @@
 import * as utils from "../test-utils/test-utils";
 import { PushProcessor } from "../../src/pushprocessor";
+import { EventType } from "../../src";
 
 describe('NotificationService', function() {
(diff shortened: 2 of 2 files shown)


src/pushprocessor.ts
```diff
@@ -96,6 +97,22 @@ const DEFAULT_OVERRIDE_RULES: IPushRule[] = [
             PushRuleActionName.DontNotify,
         ],
     },
+    {
+        // Make sure we don't notify on ACL changes - https://github.com/vector-im/element-web/issues/20788
+        rule_id: ".m.rule.server_acl",
```
t3chguy, line 102: Where is this rule defined? I'm not seeing it in https://spec.matrix.org/v1.2/client-server-api/

SimonBrandner: It isn't - we'll either go without it being there or wait for an MSC (which I will write soon, hopefully :sweat_smile:)

t3chguy: We can't use `.m.rule` prefixes without it being specced

SimonBrandner: I suppose a sensible way would be to make it `io.element.rule.server_acl` while we don't have an MSC?

t3chguy: This alone won't be sufficient, because the server will be incrementing the notification_count so causing invalid unread counts without related notifications

SimonBrandner: Ah, true :+1: , I'll write an MSC then, I suppose

t3chguy:
I think the client has to more actively create the rule, otherwise client & server will be applying different push rules causing the notification_count being wrong

src/pushprocessor.ts
```diff
@@ -96,6 +97,22 @@ const DEFAULT_OVERRIDE_RULES: IPushRule[] = [
             PushRuleActionName.DontNotify,
         ],
     },
+    {
+        // For homeservers which don't support MSC3786 yet
```
t3chguy, line 101: I'm not sure this will work as you hope, maybe will only work in e2ee rooms, otherwise the server's notifications count will still be inflated if it doesn't support this MSC

SimonBrandner: That is expected... (it probably isn't possible to get this working in a way we'd like with the current notification system...) This PR is analogues to https://github.com/matrix-org/matrix-js-sdk/pull/976: it expects the MSC implementation in Synapse to land, but in case your specific Synapse does not have it yet, it hopes to at least avoid the annoying notifications.

At least that is my understanding of why this code is here

src/pushprocessor.ts
```diff
@@ -96,6 +97,22 @@ const DEFAULT_OVERRIDE_RULES: IPushRule[] = [
...
+        rule_id: ".org.matrix.msc0000.rule.room.server_acl",
```
t3chguy, line 102:
```suggestion
        rule_id: ".org.matrix.msc3786.rule.room.server_acl",
```

t3chguy (approved):
Looks sane

t3chguy:
Hmm @SimonBrandner this says `Requires https://github.com/matrix-org/synapse/pull/12601` yet landed before it?

PR #2334, opened 2022-04-30
Switch coverage to SonarQube

t3chguy:
image

---


t3chguy:
image

SonarQube has a slightly more pessimistic analysis of our coverage, though it is founded in the exact same data.

image

See that line coverage & condition (branch) coverage match

It must be using a different metric/weighting for the main coverage percentage

turt2live (approved):

package.json
```diff
@@ -116,8 +117,14 @@
       "<rootDir>/src/**/*.{js,ts}"
     ],
     "coverageReporters": [
-      "text",
-      "json"
-    ]
+      "text-summary",
+      "json",
```
t3chguy, line 121:
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -28,11 +23,12 @@ jobs:
...
+      - name: Upload Artifact
+        uses: actions/upload-artifact@v2
         with:
-          fail_ci_if_error: false
-          verbose: true
-          override_commit: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || '' }}
+          name: coverage
+          path: |
+            coverage
+            !coverage/lcov-reportw
```
t3chguy, line 34:
```suggestion
            !coverage/lcov-report
```

PR #2340, opened 2022-05-03
Apply suggestions from SonarQube

t3chguy:
---

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -711,6 +711,7 @@ export function recursivelyAssign(target: Object, source: Object, ignoreNullish 
         }
         if ((sourceValue !== null && sourceValue !== undefined) || !ignoreNullish) {
             target[sourceKey] = sourceValue;
+            continue;
         }
     }
     return target;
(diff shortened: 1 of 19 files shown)


src/utils.ts
```diff
@@ -676,7 +676,13 @@ export function prevString(s: string, alphabet = DEFAULT_ALPHABET): string {
...
     // Dev note: this exists because I'm sad that you can use math operators on strings, so I've
     // hidden the operation in this function.
-    return (a < b) ? -1 : ((a === b) ? 0 : 1);
+    if (a < b) {
+        return -1;
+    } else if (a > b) {
+        return 1;
+    } else {
+        return 0;
+    }
```
turt2live, line 685: I mean, sure. The code was clear enough before though I think.

src/utils.ts
```diff
@@ -705,7 +711,6 @@ export function recursivelyAssign(target: Object, source: Object, ignoreNullish
         }
         if ((sourceValue !== null && sourceValue !== undefined) || !ignoreNullish) {
             target[sourceKey] = sourceValue;
-            continue;
```
turt2live, line 708: merge blocker: the benefit of being explicit here really helps, I think.

turt2live (commented):
only the one actual blocker to merge

src/utils.ts
t3chguy: It got upset about nested ternaries, and I don't blame it

turt2live (approved):

PR #2349, opened 2022-05-05
Helper to help components spit out debug logs

MadLittleMods:
Helper to help components spit out debug logs

Part of https://github.com/matrix-org/matrix-react-sdk/pull/8502

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1298,9 +1298,9 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * Get the current dehydrated device, if any
      * @return {Promise} A promise of an object containing the dehydrated device
      */
-    public getDehydratedDevice(): Promise<IDehydratedDevice> {
+    public async getDehydratedDevice(): Promise<IDehydratedDevice> {
         try {
-            return this.http.authedRequest<IDehydratedDevice>(
+            return await this.http.authedRequest<IDehydratedDevice>(
                 undefined,
                 Method.Get,
                 "/dehydrated_device",
@@ -3367,7 +3367,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * data event.
      * @return {module:http-api.MatrixError} Rejects: with an error response.
      */
-    public getAccountDataFromServer<T extends {[k: string]: any}>(eventType: string): Promise<T> {
+    public async getAccountDataFromServer<T extends {[k: string]: any}>(eventType: string): Promise<T> {
         if (this.isInitialSyncComplete()) {
             const event = this.store.getAccountData(eventType);
             if (!event) {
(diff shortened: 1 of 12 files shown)


src/models/room.ts
```diff
@@ -1216,6 +1216,13 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
         return Array.from(this.threads.values());
     }
 
+    /**
+     * @experimental
+     */
+    public getThreadsMap(): Map<string, Thread> {
+        return this.threads;
+    }
```
MadLittleMods, line 1224: The alternative to this is using `getThreads()` but this just returns unhelpful `[[], []]` blank arrays of arrays if you haven't opened the threads yet.

It's useful to see which events are threads so we can return the raw map which has a map of `threadID` -> `Thread`

```json
{
    "$BWFwOdh4QmmUpEHNdjAYXEXM5jue3ZcTtLKhc7YIusw": [],
    "$VknYvjxNw2HKifHp_8C5F_WqDMXBTfdDYYT3-AVnNhY": ["$VknYvjxNw2HKifHp_8C5F_WqDMXBTfdDYYT3-AVnNhY","$c-6O4zVYEpEuXU5bWfqFtJXXKXARxJDn64BO_9HnP4A"]
}
```

src/timeline-window.ts
```diff
@@ -41,10 +41,14 @@ const DEFAULT_PAGINATE_LOOP_LIMIT = 5;
 
 interface IOpts {
     windowLimit?: number;
+    // A label used in the logs to namespace and differentiate the timelines.
+    //ex. main timeline from the thread timeline.
+    contextLabel?: string;
```
MadLittleMods, line 46: TODO: Remove as this turned out to not be necessary and we can log everything from `TimelinePanel` in https://github.com/matrix-org/matrix-react-sdk/pull/8502

src/models/room.ts
t3chguy: Your snippet shows an object string->string[]

But the map is threadId -> Thread

The key in the object id Thread::id so you gain 0 additional information from having the map vs the array of threads

src/client.ts
```diff
@@ -8963,6 +8965,11 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             },
         );
     }
+
+    public dumpDebugLogs() {
+        logger.debug('matrix-js-sdk: dumping extra debug logs to the console');
+        this.emit(ClientEvent.DumpDebugLogs);
+    }
```
t3chguy, line 8972: using logs is a really bad mechanism given that it relies on consumers monkey-patching `console` which we don't say anywhere in the generated docs to do

src/models/room.ts
MadLittleMods: Ahh cool, `thread.id` is perfect 👌

src/client.ts
MadLittleMods: This doesn't have to live in `matrix-js-sdk`. What event emitter pattern can we use exclusively in `matrix-react-sdk`? This point is tracked by https://github.com/matrix-org/matrix-react-sdk/pull/8502#discussion_r865555576

turt2live (changes requested):
I don't think we need this at this level - will detail on the react-sdk PR.

turt2live (changes requested):
cc @MadLittleMods : I don't think we need this PR anymore?

Issue #2351, opened 2022-05-07
Overriding of setTimeout makes js-sdk import order-dependent


dbkr:
The js-sdk uses types defined in `src/@types/global.d.ts` so any typescript project importing the js-sdk needs to explicitly import this file - this is a bit questionable in itself, but this file also redefines `setTimeout` and `setInterval` to definition which the js-sdk then depends on. I can't find any docs on what Typescript is supposed to do in this situation, but empirically it seems to take the last definition it processed. The result of this is that:

```
import { MatrixClient } from "matrix-js-sdk/src/client";
import "matrix-js-sdk/src/@types/global";
```
...works, but:

```
import "matrix-js-sdk/src/@types/global";
import { MatrixClient } from "matrix-js-sdk/src/client";
```
...fails with `matrix-js-sdk/src/client.ts:979:13 - error TS2322: Type 'Timer' is not assignable to type 'number'` (and several other such errors).

dbkr:
...actually no, this only works if matrix-js-sdk is `yarn link`ed into the project (ie. has its own node type definitions in its own node_modules directory). If you just install matrix-js-sdk as a normal node module, it doesn't work either way around.

PR #2352, opened 2022-05-09
Improve typing

t3chguy:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/2351

---

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import request from "request";
+import * as request from "request";
 
 import * as matrixcs from "./matrix";
 import * as utils from "./utils";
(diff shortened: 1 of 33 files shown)


kerryarchibald (approved):
LGTM

dbkr:
merging because this fixes the issue I was having with ts-ifying element-call and I want the change :)

src/index.ts
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import request from "request";
+import * as request from "request";
```
andrevmatos, line 17: This broke our client. `request` here now is an object containing `get`, `post`, etc, instead of a function.
Repro:
```js
const { getRequest } = require('matrix-js-sdk'); // notice we're importing from index.js, as expected
typeof getRequest(); // gives `object`, instead of `function`, as before
```

andrevmatos: It seems the transpiler helper `_interopRequireWildcard` converts the function to object

PR #2354, opened 2022-05-09
Remove spec v1.3 check for threads

turt2live:
Citation: https://matrix.to/#/!ewdjhNcPcEmYNKzlWp:t2l.io/$CkPuvKdFZyFL547JCy5J3MfvLaWUo_a1XEdmiop1PKc?via=matrix.org&via=element.io&via=envs.net

---


t3chguy (approved):

t3chguy (changes requested):
This no longer handles the non `.stable` variant, we need to match mobile behaviour otherwise it'll split-brain threads for a given user of a given server

t3chguy (approved):

PR #2357, opened 2022-05-10
registration: add function to re-request email token

justjanne:
Type: Enhancement
Related: https://github.com/vector-im/element-web/issues/21984
Related: https://github.com/matrix-org/matrix-react-sdk/pull/8554

---

## ✨ Features
 * registration: add function to re-request email token ([\#2357](https://github.com/matrix-org/matrix-js-sdk/pull/2357)). Contributed by @justjanne.

### DIFF
--- a/src/interactive-auth.ts
+++ b/src/interactive-auth.ts
@@ -203,6 +203,8 @@ export class InteractiveAuth {
     private chosenFlow: IFlow = null;
     private currentStage: string = null;
 
+    private emailAttempt = 1;
+
     // if we are currently trying to submit an auth dict (which includes polling)
     // the promise the will resolve/reject when it completes
     private submitPromise: Promise<void> = null;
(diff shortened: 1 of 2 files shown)


src/interactive-auth.ts
```diff
@@ -408,6 +410,34 @@ export class InteractiveAuth {
         this.emailSid = sid;
     }
 
+    /**
+     * Requests a new email token and sets the email sid for the validation session
+     */
+    public requestEmailToken = async () => {
```
t3chguy, line 416: out of interest is there any reason this is an arrow func?

t3chguy (commented):
Looks sane otherwise, but as pointed out by Sonar, does not hit the required mark of 80% coverage on new code.

src/interactive-auth.ts
justjanne: We need to use `this` in the function, and while I personally would prefer to use a regular function and binding it in the constructor, across the codebase I’ve found arrow functions to be used more commonly to accomplish the same.

t3chguy: Sure but the function is always called with the correct `this` context already (L493) so would work in either way. Its not being copied to lose its bound context anywhere yet

t3chguy: The SDK as a whole makes the assumption that the caller will always call with the correct `this` context, we tend to use arrow funcs for when we want to use it as a local handler. Arrow functions here are not wrong by any means, was just curious.

t3chguy (approved):
🥳 Thanks for being the first to brave the better tested new world order!

src/interactive-auth.ts
justjanne: Oh, that’s an interesting assumption! Good to know, I’ve never seen a codebase that assumed this before.

justjanne:
Sadly I couldn’t migrate the tests to typescript cause they do some type trickery :(

src/interactive-auth.ts
t3chguy: It is very much a bad assumption to have, but one we make anyway. If you feel strongly about it we could make a rule that all public methods have to be arrow functions to remedy this

PR #2366, opened 2022-05-13
More sonar tweaks and typing improvements

t3chguy:
---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -32,7 +32,7 @@ import {
     MatrixEventHandlerMap,
 } from "./models/event";
 import { StubStore } from "./store/stub";
-import { CallEvent, CallEventHandlerMap, createNewMatrixCall, MatrixCall } from "./webrtc/call";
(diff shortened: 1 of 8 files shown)


src/client.ts
```diff
@@ -8037,7 +8025,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         kind: PushRuleKind,
         ruleId: Exclude<string, RuleId>,
         callback?: Callback,
-    ): Promise<any> { // TODO: Types
+    ): Promise<{}> {
```
turt2live, line 8028: Mostly to show it's deliberate and not an attempt to evade types:
```suggestion
    ): Promise<{}> { // API returns an empty object
```

turt2live (approved):
lgtm generally

src/client.ts
t3chguy: JSDoc a few lines up says it is deliberate

PR #2370, opened 2022-05-13
Fix races when muting/unmuting

dbkr:
await on the async operation so the promise we return resolves once
everything's actuall complete

---

## 🐛 Bug Fixes
 * Fix races when muting/unmuting ([\#2370](https://github.com/matrix-org/matrix-js-sdk/pull/2370)).


t3chguy:
image

Huh, TIL

t3chguy (approved):

PR #2373, opened 2022-05-16
Try to load keys from key backup when a message fails to decrypt

duxovni:
Fixes vector-im/element-web#21026.

Also typescriptified `TestClient` while I was in the area, I can split that into a separate PR if that'd be preferable.

---

## ✨ Features
 * Try to load keys from key backup when a message fails to decrypt ([\#2373](https://github.com/matrix-org/matrix-js-sdk/pull/2373)). Fixes vector-im/element-web#21026. Contributed by @duxovni.


spec/integ/megolm-backup.spec.ts
```diff
@@ -0,0 +1,165 @@
+/*
+Copyright 2019 The Matrix.org Foundation C.I.C.
```
turt2live, line 2:
```suggestion
Copyright 2022 The Matrix.org Foundation C.I.C.
```

src/crypto/algorithms/megolm.ts
```diff
@@ -1298,6 +1298,8 @@ class MegolmDecryption extends DecryptionAlgorithm {
 
         if (res === null) {
             // We've got a message for a session we don't have.
+
+            this.crypto.backupManager.queryKeyBackupRateLimited(event.getRoomId(), content.session_id).catch(() => {});
             //
             // (XXX: We might actually have received this key since we started
```
turt2live, line 1304:
```suggestion
            // We've got a message for a session we don't have.
            // try and get the missing key from the backup first
            this.crypto.backupManager.queryKeyBackupRateLimited(event.getRoomId(), content.session_id).catch(() => {});

            // (XXX: We might actually have received this key since we started
```

src/crypto/backup.ts
```diff
@@ -111,6 +112,7 @@ export class BackupManager {
     public backupInfo: IKeyBackupInfo | undefined; // The info dict from /room_keys/version
     public checkedForBackup: boolean; // Have we checked the server for a backup we can use?
     private sendingBackups: boolean; // Are we currently sending backups?
+    private sessionLastCheckAttemptedTime: Record<string, number> = {}; // When did we last try to check the server for a given session id?
     constructor(private readonly baseApis: MatrixClient, public readonly getKey: GetKey) {
```
turt2live, line 116: blank line to appease future linter:
```suggestion

    constructor(private readonly baseApis: MatrixClient, public readonly getKey: GetKey) {
```

src/crypto/backup.ts
```diff
@@ -35,6 +35,7 @@ import { UnstableValue } from "../NamespacedValue";
 import { CryptoEvent, IMegolmSessionData } from "./index";
 
 const KEY_BACKUP_KEYS_PER_REQUEST = 200;
+const KEY_BACKUP_CHECK_RATE_LIMIT = 5000;
```
turt2live, line 38:
```suggestion
const KEY_BACKUP_CHECK_RATE_LIMIT = 5000; // ms
```

src/crypto/backup.ts
```diff
@@ -282,6 +284,24 @@ export class BackupManager {
...
 
+    /**
+     * Attempts to retrieve a session from a key backup, if enough time
+     * has elapsed since the last check for this session id.
+     */
+    public async queryKeyBackupRateLimited(
+        targetRoomId: string | undefined,
+        targetSessionId: string | undefined,
+    ): Promise<void> {
+        const now = new Date().getTime();
```
turt2live, line 295: I think we're also missing a check to see if key backup is even enabled/possible? This whole area of code is a bit difficult to follow, but I think a null check on `this.backupInfo` or `this.algorithm` will do the job.

turt2live (commented):
Overall seems fine, thank you! Just some random suggestions and an edge case to consider.

spec/integ/matrix-client-retrying.spec.ts
```diff
@@ -1,10 +1,10 @@
-import { EventStatus, RoomEvent } from "../../src/matrix";
+import { EventStatus, RoomEvent, MatrixClient } from "../../src/matrix";
```
t3chguy, line 1: Missing copyright

src/crypto/backup.ts
duxovni: Good catch, thanks!

src/crypto/algorithms/megolm.ts
```diff
@@ -1298,7 +1298,9 @@ class MegolmDecryption extends DecryptionAlgorithm {
 
         if (res === null) {
             // We've got a message for a session we don't have.
-            //
+            // try and get the missing key from the backup first
+            this.crypto.backupManager.queryKeyBackupRateLimited(event.getRoomId(), content.session_id).catch(() => {});
```
BillCarsonFr, line 1302: In order to reduce unneeded to_device traffic, on EA if we have a positive reply from the backup we don't call requestKeysForEvent. Is it possible to do the same here?

duxovni: Not trivially. If we naively `await` this operation, we get stuck in a deadlock because the key backup query operation includes importing the keys, which includes re-attempting decryption of the messages, which means waiting for the current decryption attempt to finish. So we can't use, or even know, the results of the backup query during the current decryption attempt, without a bunch of messy re-architecting. In order to still avoid sending key requests, we'd need to add more complicated state tracking to check whether we've already tried the key backups on a previous decryption attempt :/

turt2live (approved):
would recommend waiting for the RC to be cut first, but lgtm

PR #2374, opened 2022-05-17
Ensure rooms are recalculated on re-invites

turt2live:
Fixes https://github.com/vector-im/element-web/issues/22106

**Reviewer**: This has two commits: a failing test and the fix. This is to demonstrate that the test actually works, because I don't exactly have confidence in it myself.

---

## 🐛 Bug Fixes
 * Ensure rooms are recalculated on re-invites ([\#2374](https://github.com/matrix-org/matrix-js-sdk/pull/2374)). Fixes vector-im/element-web#22106.

### DIFF
--- a/spec/integ/matrix-client-syncing.spec.js
+++ b/spec/integ/matrix-client-syncing.spec.js
@@ -1,5 +1,4 @@
-import { MatrixEvent } from "../../src/models/event";
-import { EventTimeline } from "../../src/models/event-timeline";
+import { EventTimeline, MatrixEvent, RoomEvent } from "../../src";
 import * as utils from "../test-utils/test-utils";
 import { TestClient } from "../TestClient";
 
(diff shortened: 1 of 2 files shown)


spec/integ/matrix-client-syncing.spec.js
```diff
@@ -1,5 +1,4 @@
-import { MatrixEvent } from "../../src/models/event";
-import { EventTimeline } from "../../src/models/event-timeline";
+import { EventTimeline, MatrixEvent, RoomEvent } from "../../src";
```
t3chguy, line 1: Missing copyright

src/sync.ts
```diff
@@ -1166,6 +1166,9 @@ export class SyncApi {
                 room.recalculate();
                 client.store.storeRoom(room);
                 client.emit(ClientEvent.Room, room);
+            } else {
+                // Update room state for invite->reject->invite cycles
+                room.recalculate();
```
t3chguy, line 1171: Can we not just move the recalculate out to before the `if`?

t3chguy (approved):
LGTM otherwise

turt2live:
Slightly worried about this change: if people notice performance regressions, please flag.

PR #2382, opened 2022-05-18
Initial attempt at automating jsdoc

t3chguy:
Test example https://github.com/t3chguy/matrix-js-sdk/commit/8f67b714968f4c8db83a6c2a2c2e497fda0a337b

---


turt2live (approved):
seems like a thing that does a jsdoc thing

PR #2383, opened 2022-05-18
Handle other members having no e2e keys

dbkr:
Fetch the device info once at the start of the cal and cache it
rather than fetching every time, and throw if we're supposed to be
using e2e but the other end has no e2e keys.

Fixes https://github.com/vector-im/element-call/issues/338

---

## 🐛 Bug Fixes
 * Handle other members having no e2e keys ([\#2383](https://github.com/matrix-org/matrix-js-sdk/pull/2383)). Fixes vector-im/element-call#338.


robertlong:
Looks good 👍

robertlong (approved):

PR #2385, opened 2022-05-19
Catch promise errors in degradable (fixes #2384)

### LINKED ISSUE #2384: Degration of indexeddb to MemoryStore does not work anymore
**Describe the bug**
If the indexeddb throws an exception, normally the store would have degraded on a MemoryStore.
This does not longer work, because of the switch to the async-notation. The exceptions do not get caught and the degration will not be started.

**To Reproduce**
1. Client startup with custom indexeddb in store
```js
import db from './customindexeddb';

let opts = {
  store: new matrixcs.IndexedDBStore(
    indexeddb: db,
    dbName: 'matrix-store',
  ),
  baseUrl: "https://matrix.org",
};

var client = matrixcs.createClient(opts);
client.store.startup();
```
2. customindexeddb is programmed to throw on reading from object stores
3. MatrixClient crashes, but does not degrade to MemoryStore

**Expected behavior**
An error is thrown in the console, indexeddb will not be used anymore, and the backend store falls back to  the MemoryStore.

schmop:
Signed-off-by: Lars Richard <lars.richard@iserv.de>

Fixes #2384

---

## 🐛 Bug Fixes
 * Fixes #2384 ([\#2385](https://github.com/matrix-org/matrix-js-sdk/pull/2385)). Fixes undefined/matrix-js-sdk#2384. Contributed by @schmop.

### DIFF
--- a/src/store/indexeddb.ts
+++ b/src/store/indexeddb.ts
@@ -297,7 +297,7 @@ export class IndexedDBStore extends MemoryStore {
 
         return async (...args) => {
             try {
-                return func.call(this, ...args);
+                return await func.call(this, ...args);
             } catch (e) {
                 logger.error("IndexedDBStore failure, degrading to MemoryStore", e);
                 this.emitter.emit("degraded", e);


t3chguy (approved):
Thanks for this contribution, I'll make sure it makes it into the next release which is currently in RC

PR #2389, opened 2022-05-20
Fix behaviour of isRelation with relation m.replace for state events

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/22280

---

## 🐛 Bug Fixes
 * Fix behaviour of isRelation with relation m.replace for state events ([\#2389](https://github.com/matrix-org/matrix-js-sdk/pull/2389)). Fixes vector-im/element-web#22280.


andybalaam (approved):
LGTM, thanks

PR #2391, opened 2022-05-20
Add GHA for requiring all reviews to be dismissed or approving before merging

t3chguy:
---


turt2live (approved):
I have it on good word that this is a safe change

RiotRobot (changes requested):
test

RiotRobot (dismissed):
Test

justjanne:
Does this also count "reviews" from random users who’ve got nothing to do with matrix/element?

t3chguy:
> Does this also count "reviews" from random users who’ve got nothing to do with matrix/element?

Yup, intentional that those would need explicit dismissal

justjanne (approved):

gsouquet (approved):

PR #2392, opened 2022-05-21
Prevent overlapping sync accumulator persists

ara4n:
Adds a flag to stop the sync worker trying to persist to indexeddb if there is already a persist in flight.  This means that if indexeddb is being very slow (e.g. due to the browser running for a few seconds every ~10 mins while a macbook is asleep), you won't end up with overlapping write operations to indexeddb which could consume heap and cause an OOM.  It's okay to do this in the sync worker rather than the main thread, as it turns out that all the accumulation happens inside the worker, so we never pass large accumulated sync responses over postmessage to the worker.

If the persist is skipped, we accumulate user presence updates in RAM to stop them being lost (given the sync accumulator doesn't handle them itself for some reason).

Hopefully fixes https://github.com/vector-im/element-web/issues/21541

---

## 🐛 Bug Fixes
 * Prevent overlapping sync accumulator persists ([\#2392](https://github.com/matrix-org/matrix-js-sdk/pull/2392)). Fixes vector-im/element-web#21541.

### DIFF
--- a/src/store/indexeddb-local-backend.ts
+++ b/src/store/indexeddb-local-backend.ts
@@ -412,13 +412,15 @@ export class LocalIndexedDBStoreBackend implements IIndexedDBBackend {
             this.isPersisting = true;
         }
 
-        await Promise.all([
-            this.persistUserPresenceEvents(userTuples),
-            this.persistAccountData(syncData.accountData),
-            this.persistSyncData(syncData.nextBatch, syncData.roomsData),
-        ]).finally(() => {
+        try {
+            await Promise.all([
+                this.persistUserPresenceEvents(userTuples),
+                this.persistAccountData(syncData.accountData),
+                this.persistSyncData(syncData.nextBatch, syncData.roomsData),
+            ]);
+        } finally {
             this.isPersisting = false;
-        });
+        }
     }
 
     /**
--- a/src/store/indexeddb-local-backend.ts
+++ b/src/store/indexeddb-local-backend.ts
@@ -127,6 +127,8 @@ export class LocalIndexedDBStoreBackend implements IIndexedDBBackend {
     private db: IDBDatabase = null;
     private disconnected = true;
     private _isNewlyCreated = false;
+    private _isPersisting = false;
+    private _pendingUserPresenceData: UserTuple[] = [];
 
     /**
      * Does the actual reading from and writing to the indexeddb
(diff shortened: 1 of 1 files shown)


src/store/indexeddb-local-backend.ts
```diff
@@ -127,6 +127,8 @@ export class LocalIndexedDBStoreBackend implements IIndexedDBBackend {
     private db: IDBDatabase = null;
     private disconnected = true;
     private _isNewlyCreated = false;
+    private _isPersisting = false;
+    private _pendingUserPresenceData: UserTuple[] = [];
```
SimonBrandner, line 131: I think we don't prefix private members with `_` anymore; the only reason it is used for `_isNewlyCreated` is that there is a getter with the name `isNewlyCreated`

SimonBrandner (approved):
Otherwise LGTM!

src/store/indexeddb-local-backend.ts
```diff
@@ -401,11 +403,22 @@ export class LocalIndexedDBStoreBackend implements IIndexedDBBackend {
...
         const syncData = this.syncAccumulator.getJSON(true);
 
+        if (this.isPersisting) {
+            logger.warn("Skipping syncToDatabase() as persist already in flight");
+            this.pendingUserPresenceData.push(...userTuples);
+            return;
+        } else {
+            userTuples.unshift(...this.pendingUserPresenceData);
+            this.isPersisting = true;
+        }
```
t3chguy, line 413: given the return we could outdent this `else` case

src/store/indexeddb-local-backend.ts
```diff
@@ -401,11 +403,22 @@ export class LocalIndexedDBStoreBackend implements IIndexedDBBackend {
...
+
         await Promise.all([
             this.persistUserPresenceEvents(userTuples),
             this.persistAccountData(syncData.accountData),
             this.persistSyncData(syncData.nextBatch, syncData.roomsData),
-        ]);
+        ]).finally(() => {
+            this.isPersisting = false;
+        });
```
t3chguy, line 421: given the async-await we could try `try-finally`

src/store/indexeddb-local-backend.ts
ara4n: *shrug* - the intent is to explicitly show the alternation. not sure what benefit there is to outdenting it, other than make the flow of control harder to grok.

src/store/indexeddb-local-backend.ts
ara4n: hm, i might be missing how finally interacts with async-await. how would try-finally help here?

ara4n: you're saying this could be `try { await Promise.all(); } finally { this.isPersisting = false }` - is this just stylistic? feels a bit weird to be mixing promise-chaining style with block-style just because there's an `await` in there? or is the intention to avoid ambiguity over the precedence of the `await`?

t3chguy: We're tending towards async-await vs promise-chaining and I'm not seeing any promise chaining in this block. The ambiguity aspect is minimal. I will bring it up on the code style doc to codify something, for now this is merely a comment

ara4n: right, so i guess the concern is mixing imperative await-style with chain-style (.finally()). will switch to blocks for the `finally`

ara4n:
going to merge this so i don't have to run a custom Nightly to see if it's fixed things.

PR #2398, opened 2022-05-24
Implementation of MSC3824 to add action= param on SSO login

hughns:
---

## ✨ Features
 * Implementation of MSC3824 to add action= param on SSO login ([\#2398](https://github.com/matrix-org/matrix-js-sdk/pull/2398)). Contributed by @hughns.


src/client.ts
```diff
@@ -876,6 +877,8 @@ export type ClientEventHandlerMap = {
     & HttpApiEventHandlerMap
     & BeaconEventHandlerMap;
 
+const UNSTABLE_MSC3824_ACTION_PARAM = new UnstableValue("action", "org.matrix.msc3824.action");
```
turt2live, line 880: we typically just call it something like `OIDC_ACTION_PARAM` so we can change the type to a `NamespacedValue` without breaking the SDK itself.

hughns: I've renamed it `SSO_ACTION_PARAM`.

src/client.ts
```diff
@@ -7143,15 +7146,24 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
      * @param {string} idpId The ID of the Identity Provider being targeted, optional.
+     * @param {"login"|"register"} action the SSO flow to indicate to the IdP, optional.
      * @return {string} The HS URL to hit to begin the SSO login process.
      */
-    public getSsoLoginUrl(redirectUrl: string, loginType = "sso", idpId?: string): string {
+    public getSsoLoginUrl(
+        redirectUrl: string,
+        loginType = "sso",
+        idpId?: string,
+        action?: "login" | "register",
```
SimonBrandner, line 7156: Would we benefit from this being an enum?

src/client.ts
```diff
@@ -7143,15 +7146,24 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+    ): string {
         let url = "/login/" + loginType + "/redirect";
         if (idpId) {
             url += "/" + idpId;
         }
 
-        return this.http.getUrl(url, { redirectUrl }, PREFIX_R0);
+        const params = { redirectUrl };
+        params[SSO_ACTION_PARAM.unstable] = action;
```
SimonBrandner, line 7164:
```suggestion
        const params = { 
            redirectUrl,
            [SSO_ACTION_PARAM.unstable]: action,
        };
```

I think this looks a bit nicer

<hr>

Also, are we ok with sending the server `undefined`s?

SimonBrandner (approved):

weeman1337 (changes requested):
@hughns can you add a test for it?

src/client.ts
hughns: I've made it an enum.

hughns:
> @hughns can you add a test for it?

I've added some some simple test cases.

weeman1337 (approved):
:+1:

PR #2412, opened 2022-05-30
Refactor Relations to not be per-EventTimelineSet

### LINKED ISSUE #2399: EventTimelineSet constructor and docs are contradictory
**Describe the bug**
The docblock of `EventTimelineSet`s constructor states that the room should be nullable, the parameters however state that it is not. Also the usage within the class is not clear how it should be used. (`getPendingEvents` has a guard clause, while `aggregateRelations` does not)

This leads to errors if you use the Timelineset without providing a room.

**To Reproduce**
1. Create a new EventTimelineSet with room set to null
2. Try to add an Event with a relation (in our case `m.replace`)

**Expected behavior**
The class should work correctly without a room or the docblock should be changed.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser: chrome, firefox

t3chguy:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/2399
Fixes https://github.com/vector-im/element-web/issues/22298

All the "breaking" changes are on code which was marked as experimental so does not need a major version bump.

---

## 🐛 Bug Fixes
 * Refactor Relations to not be per-EventTimelineSet ([\#2412](https://github.com/matrix-org/matrix-js-sdk/pull/2412)). Fixes #2399 and vector-im/element-web#22298.


t3chguy:
Blocking until after RC

src/models/relations-container.ts
```diff
@@ -0,0 +1,153 @@
...
+    // this.relations[relatesToEventId][relationType][relationEventType]
+    private relations: {
+        [relatesToEventId: string]: {
+            [relationType: RelationType | string]: {
+                [eventType: EventType | string]: Relations;
+            };
+        };
+    } = {};
+
+    constructor(private readonly client: MatrixClient, private readonly room?: Room) {
```
turt2live, line 35: `client` is nullable thanks to how event-timeline-set works on the constructor (and possibly other places)

src/models/relations-container.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * Get a collection of relations to a given event in this timeline set.
```
turt2live, line 39: There's a reason why the spec decided to abandon all hope of documenting what a "relation event", "related event", and "relatable event" are: it gets quite complicated. 


Having reviewed this PR, I think it's quite important that we match the spec on this as it's nearly impossible to understand what event is being talked about with the current overlap in vocabulary. The spec uses "parent" and "child" to denote the relationship between the events, described here: https://github.com/matrix-org/matrix-spec/pull/1062

src/models/relations-container.ts
```diff
@@ -0,0 +1,153 @@
...
+     */
+    public getRelationsForEvent(
+        eventId: string,
+        relationType: RelationType | string,
+        eventType: EventType | string,
+    ): Relations | undefined {
+        return this.relations[eventId]?.[relationType]?.[eventType];
+    }
+
+    public getAllRelationsEventForEvent(eventId: string): MatrixEvent[] {
```
turt2live, line 63:
```suggestion
    public getAllRelationsEventsForEvent(eventId: string): MatrixEvent[] {
```

.. or are they related events? :P

turt2live (changes requested):
generally looks sane - just a few nitpicks and a whole naming scheme change 😇

src/models/relations-container.ts
t3chguy: It is required, the jsdoc for EventTimelineSet says client or room must be provided (or both, but that isn't stated)

turt2live (approved):

Issue #2353, opened 2022-05-09
Convert all matrix-js-sdk tests to use Typescript


novocaine:
JS-SDK version of https://github.com/vector-im/element-web/issues/21965

PR #2416, opened 2022-05-30
matrix-mock-request to 2.0.1

kerryarchibald:
https://github.com/matrix-org/matrix-js-sdk/issues/2353

with https://github.com/matrix-org/matrix-mock-request/pull/14

- update matrix-mock-request to v2.0.1
- fix issue with tests logging after completing that surfaced after mock-req update. Track offending timeouts from mock clients and clear in afterEach
- upgrade new ts TestClient for typed mock backend

---


turt2live (approved):

src/client.ts
```diff
@@ -590,13 +590,9 @@ export interface IRequestMsisdnTokenResponse extends IRequestTokenResponse {
     intl_fmt: string;
 }
 
-interface IUploadKeysRequest {
+export interface IUploadKeysRequest {
```
kerryarchibald, line 593: This type was incorrect, the usage here matches the new type [psf-874/test-messageactionbar](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/crypto/index.ts#L2003-L2026)

kerryarchibald: Added breaking change label

turt2live: I don't think this is a breaking change? The interface wasn't exported before, so it'd just be a regular bug fix.

turt2live (approved):
otherwise lgtm

src/client.ts
kerryarchibald: I wasn't really sure here so erred on the safe side. Removed the label.

PR #2419, opened 2022-05-30
Update renovate.json

t3chguy:
---


turt2live (approved):
this is in fact an update

PR #2421, opened 2022-05-30
Mute disconnected peers in PTT mode

dbkr:
When we lose ICE connection to peers, set the status of their feeds
to muted so to end their talking session.

For https://github.com/vector-im/element-call/issues/331

---

## ✨ Features
 * Mute disconnected peers in PTT mode ([\#2421](https://github.com/matrix-org/matrix-js-sdk/pull/2421)).

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -292,6 +292,10 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
     public peerConn?: RTCPeerConnection;
     public toDeviceSeq = 0;
 
+    // whether this call should have push-to-talk semantics
+    // This should be set by the consumer on incoming & outgoing calls.
+    public isPtt = false;
+
     private client: MatrixClient;
     private forceTURN: boolean;
     private turnServers: Array<TurnServer>;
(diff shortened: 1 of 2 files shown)


src/webrtc/call.ts
```diff
@@ -292,6 +292,10 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
     public peerConn?: RTCPeerConnection;
     public toDeviceSeq = 0;
 
+    // whether this call should have push-to-talk semantics
+    // This should be set by the consumer on incoming & outgoing calls.
+    public isPtt = false;
```
kerryarchibald, line 297: nitpick: would this be more legible long term as `isRadioMode` or similar?

kerryarchibald (approved):
Looks reasonable to me. Left one naming nitpick, feel free to ignore.

(Also, small poke about tests, not sure how you're handling that in the ptt branch)

src/webrtc/call.ts
dbkr: We've used 'ptt' in Element Call internally, so I think we'll probably stick with 'ptt' in the code for (hopefully) less confusion, irrespective of what it ends up being called to the user. And yep, I'm not yet 100% sure how we'll get the test coverage up on the voip code, but it's definitely a thing on my mind - thanks for the poke. :)

PR #2424, opened 2022-05-31
Typings for MSC2965 OIDC provider discovery

hughns:
---

## ✨ Features
 * Typings for MSC2965 OIDC provider discovery ([\#2424](https://github.com/matrix-org/matrix-js-sdk/pull/2424)). Contributed by @hughns.


robintown (approved):
I have my doubts about whether we should be using UnstableValue like this (including the stable identifier before it's ever part of the spec), but given that we do that throughout the codebase, I think this is fine

hughns:
> I have my doubts about whether we should be using UnstableValue like this (including the stable identifier before it's ever part of the spec), but given that we do that throughout the codebase, I think this is fine

I was confused by this too... 🤷‍♂️

PR #2430, opened 2022-06-02
Don't bug the user while re-checking key backups after decryption failures

duxovni:
Fixes https://github.com/vector-im/element-web/issues/22416


robintown (approved):
Looks good

PR #2435, opened 2022-06-06
Add tests for sendEvent threadId handling

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/22433

---

## 🐛 Bug Fixes
 * Add tests for sendEvent threadId handling ([\#2435](https://github.com/matrix-org/matrix-js-sdk/pull/2435)). Fixes vector-im/element-web#22433.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -3785,7 +3785,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
                 content["m.relates_to"]["m.in_reply_to"] = {
                     "event_id": thread.lastReply((ev: MatrixEvent) => {
                         return ev.isRelation(THREAD_RELATION_TYPE.name) && !ev.status;
-                    })?.getId(),
+                    })?.getId() ?? threadId,
                 };
             }
         }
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -3772,13 +3772,16 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         // If we expect that an event is part of a thread but is missing the relation
         // we need to add it manually, as well as the reply fallback
         if (threadId && !content["m.relates_to"]?.rel_type) {
+            const isReply = !!content["m.relates_to"]?.["m.in_reply_to"];
             content["m.relates_to"] = {
                 ...content["m.relates_to"],
                 "rel_type": THREAD_RELATION_TYPE.name,
                 "event_id": threadId,
+                // Set is_falling_back to true unless this is actually intended to be a reply
```
turt2live, line 3780:
```suggestion
                // Set is_falling_back to true unless this is actually intended to be a thread
```

because it's a reply unless it's a thread, essentially

turt2live (approved):
otherwise lgtm, though I'm not super confident on the comment myself.

src/client.ts
t3chguy: Reply is right here. The `is_falling_back` is to say, is the m.in_reply_to intentional, or just for the fallback

t3chguy: conferred in DM, it is not backwards, spec just took some weird turns

turt2live: apparently we specified this field to go the inverse to what I thought, so the sentence is in fact correct.

PR #2437, opened 2022-06-07
Live location sharing - monitor liveness of beacons yet to start

kerryarchibald:
https://github.com/vector-im/element-web/issues/22437

- When a beacon's start timestamp is in the future, add a timeout to refresh the beacon's state after it becomes live


weeman1337 (approved):
LGTM :+1:

PR #2441, opened 2022-06-08
Test typescriptification - cross-signing.spec

kerryarchibald:
https://github.com/matrix-org/matrix-js-sdk


spec/unit/crypto/cross-signing.spec.ts
```diff
@@ -1023,7 +1068,7 @@ describe("Cross Signing", function() {
             },
         };
         alice.crypto.deviceList.storeDevicesForUser("@alice:example.com", {
-            Dynabook: aliceNotCrossSignedDevice,
+            Dynabook: aliceNotCrossSignedDevice as unknown as IDevice,
```
t3chguy, line 1071: Can we get a comment for why this is needed

t3chguy (approved):
LGTM otherwise

spec/unit/crypto/cross-signing.spec.ts
kerryarchibald: Fixed the object instead

PR #2444, opened 2022-06-08
Fix issues with getEventTimeline and thread roots

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/21613

### DIFF
--- a/spec/integ/matrix-client-event-timeline.spec.js
+++ b/spec/integ/matrix-client-event-timeline.spec.js
@@ -1,5 +1,5 @@
 import * as utils from "../test-utils/test-utils";
-import { EventTimeline } from "../../src/matrix";
+import { EventTimeline, MatrixEvent } from "../../src/matrix";
 import { logger } from "../../src/logger";
 import { TestClient } from "../TestClient";
 import { Thread, THREAD_RELATION_TYPE } from "../../src/models/thread";
@@ -500,7 +500,8 @@ describe("MatrixClient event timelines", function() {
             Thread.setServerSideSupport(true);
             client.stopClient(); // we don't need the client to be syncing at this time
             const room = client.getRoom(roomId);
-            const timelineSet = room.getTimelineSets()[0];
+            const thread = room.createThread(THREAD_ROOT.event_id, undefined, [], false);
+            const timelineSet = thread.timelineSet;
 
             httpBackend.when("GET", "/rooms/!foo%3Abar/context/" + encodeURIComponent(THREAD_REPLY.event_id))
                 .respond(200, function() {
@@ -537,6 +538,39 @@ describe("MatrixClient event timelines", function() {
 
             expect(timeline.getEvents().find(e => e.getId() === THREAD_ROOT.event_id)).toBeTruthy();
             expect(timeline.getEvents().find(e => e.getId() === THREAD_REPLY.event_id)).toBeTruthy();
+        });
+
+        it("should return undefined when event is not within a thread but timelineSet is", async () => {
+            client.clientOpts.experimentalThreadSupport = true;
+            Thread.setServerSideSupport(true);
+            client.stopClient(); // we don't need the client to be syncing at this time
+            const room = client.getRoom(roomId);
+            const threadRoot = new MatrixEvent(THREAD_ROOT);
+            const thread = room.createThread(THREAD_ROOT.event_id, threadRoot, [threadRoot], false);
+            const timelineSet = thread.timelineSet;
+
+            httpBackend.when("GET", "/rooms/!foo%3Abar/event/" + encodeURIComponent(THREAD_ROOT.event_id))
+                .respond(200, function() {
+                    return THREAD_ROOT;
+                });
+
+            httpBackend.when("GET", "/rooms/!foo%3Abar/context/" + encodeURIComponent(EVENTS[0].event_id))
+                .respond(200, function() {
+                    return {
+                        start: "start_token0",
+                        events_before: [],
+                        event: THREAD_ROOT,
+                        events_after: [],
+                        end: "end_token0",
+                        state: [],
+                    };
+                });
+
+            const timelinePromise = client.getEventTimeline(timelineSet, EVENTS[0].event_id);
+            await httpBackend.flushAllExpected();
+
+            const timeline = await timelinePromise;
+            expect(timeline).toBeUndefined();
         });
     });
 


turt2live (approved):

src/models/event-timeline-set.ts
```diff
@@ -119,13 +120,15 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
      * Set to true to enable improved timeline support.
      * @param {Object} [opts.filter = null]
      * The filter object, if any, for this timelineSet.
-     * @param {MatrixClient} client the Matrix client which owns this EventTimelineSet,
+     * @param {MatrixClient=} client the Matrix client which owns this EventTimelineSet,
```
MadLittleMods, line 123: Is this only necessary for threads? That's the only place we use this functionality. We should add a commentthe reasoning for when this is necessary and the thread use case.

src/models/event-timeline-set.ts
```diff
@@ -119,13 +120,15 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
...
      * can be omitted if room is specified.
+     * @param {Thread=} thread the thread to which this timeline set relates.
      */
     constructor(
         public readonly room: Room | undefined,
         opts: IOpts = {},
         client?: MatrixClient,
+        public readonly thread?: Thread,
```
MadLittleMods, line 131: Seems like we should put this stuff in `opts`

src/client.ts
```diff
@@ -5245,14 +5245,15 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * <p>If the EventTimelineSet object already has the given event in its store, the
      * corresponding timeline will be returned. Otherwise, a /context request is
      * made, and used to construct an EventTimeline.
+     * If the event does not belong to this EventTimelineSet then undefined will be returned.
```
MadLittleMods, line 5248:
```suggestion
     * If the event does not belong to this EventTimelineSet then undefined will be returned
     * (like an `eventId` that doesn't belong in the thread represented by the given `EventTimelineSet`).
```

src/client.ts
```diff
@@ -5245,14 +5245,15 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
      *
-     * @param {EventTimelineSet} timelineSet  The timelineSet to look for the event in
+     * @param {EventTimelineSet} timelineSet  The timelineSet to look for the event in, must be bound to a room
```
MadLittleMods, line 5250: We should add an assertion or throw a useful error when `timelineSet.room` is not set. Currently it will blow up with a `undefined` exception.

spec/integ/matrix-client-event-timeline.spec.js
```diff
@@ -538,6 +539,92 @@ describe("MatrixClient event timelines", function() {
             expect(timeline.getEvents().find(e => e.getId() === THREAD_ROOT.event_id)).toBeTruthy();
             expect(timeline.getEvents().find(e => e.getId() === THREAD_REPLY.event_id)).toBeTruthy();
         });
+
+        it("should return undefined when event is not within a thread but timelineSet is", async () => {
```
MadLittleMods, line 543: The current title makes sense but I had to read it a couple times to get it.

Maybe this:

```suggestion
        it("should return undefined when event is not within the thread that the given timelineSet is representing", async () => {
```

src/client.ts
```diff
@@ -5297,38 +5298,44 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
             }
 
-            const opts: IRelationsRequestOpts = {
-                direction: Direction.Backward,
-                limit: 50,
-            };
+            if (timelineSet.thread?.id !== threadId) {
+                // Event does not belong in this timelineSet
+                return undefined;
+            }
```
MadLittleMods, line 5312: These two conditional statements are the main part of the fix.

Feels like we should have a `shouldEventBelongTimelineSet(timelineSet, event);` function to wrap up this logic in an understandable label.

src/client.ts
```diff
@@ -5297,38 +5298,44 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            if (!timelineSet.thread && !shouldLiveInRoom) {
+                // Thread response does not belong in this timelineSet
+                return undefined;
             }
 
-            const opts: IRelationsRequestOpts = {
-                direction: Direction.Backward,
-                limit: 50,
-            };
+            if (timelineSet.thread?.id !== threadId) {
```
MadLittleMods, line 5309:
```suggestion
            // Prevent thread events mixing in the main room timeline or the wrong thread timeline
            if (timelineSet.thread?.id !== threadId) {
```

src/client.ts
MadLittleMods, line 5312: I think this second conditional covers all of the cases:

 - If main timeline and thread event, return `undefined`
 - If thread timeline and main timeline event, return `undefined`
 - If thread timeline and event from different thread, return `undefined`
 - If the main timeline and main timeline event, return timeline
 - If thread timeline and event from same thread, return timeline
 
I feel like the first conditional doesn't add anything more? Am I missing something?

src/client.ts
MadLittleMods, line 5309: A `threadRoot` returns:
```
{
    shouldLiveInRoom: true,
    shouldLiveInThread: true,
    threadId: event.getId(),
};
```

Won't this logic prevent a `threadRoot` from showing in the main timeline? `undefined !== $abc`

MadLittleMods, line 5309: I think this second conditional covers all of the cases:

 - If main timeline and thread event, return `undefined`
 - If thread timeline and main timeline event, return `undefined`
 - If thread timeline and event from different thread, return `undefined`
 - If the main timeline and main timeline event, return timeline
 - If thread timeline and event from same thread, return timeline
 
I feel like the first conditional doesn't add anything more? Am I missing something?

spec/integ/matrix-client-event-timeline.spec.js
```diff
@@ -1,5 +1,5 @@
 import * as utils from "../test-utils/test-utils";
```
MadLittleMods, line 1: The title and changelog of this PR is not every descriptive: "Fix issues with getEventTimeline and thread roots". It just describes the code and not what's being fixed.

Potential option:

> Fix events from main timeline getting mixed in threads

t3chguy: But this does more than that, that's just one symptom it fixes, changelog entries should be relatively short even if that's vague

src/client.ts
t3chguy: Indeed, I think this is the cause of https://github.com/vector-im/element-web/issues/22539

The if check should be `timelineSet.thread && timelineSet.thread.id !== threadId`

t3chguy: Indeed, that is only because of it being incorrect, as per the above comment

src/models/event-timeline-set.ts
t3chguy: Its used for the notification timeline (or any without a room) where a `client` is still necessary but cannot be accessed via `room.client`

spec/integ/matrix-client-event-timeline.spec.js
MadLittleMods: What other symptoms does it fix? The other problems are all variations of that AFAICT.

The problem is that following the link to this PR from the changelog also doesn't have any of the extra details.

t3chguy: > The problem is that following the link to this PR from the changelog also doesn't have any of the extra details either.

Except it does, in the canonical source, the issue it links to

image

MadLittleMods: Why doesn't "Fix events from main timeline getting mixed in threads" work to describe what this does then?

t3chguy: Because it also stops other effects, like stops an unrelated timeline being returned if you call with a threaded-timelineset and an event id from a diff thread/main timeline

MadLittleMods: The description and linked issue don't explain this. And my changelog example can be extended to something like:

> Fix events from main timeline getting mixed in threads and generally events being mixed across timelines.

Which is way more useful than `Fix issues with getEventTimeline and thread roots` where [thread roots are one of the edge cases that it actually doesn't fix](https://github.com/matrix-org/matrix-js-sdk/pull/2444#discussion_r893025870) anyway. And doesn't describe the main issue it fixes that everyone is running into in the first place.

I had to decode all of this information from the diff and when I did, I found bugs. The expectations and why were not spelled out in the code or description where we can easily compare how the code behaves to those expectations. One (maybe 2) symptoms are explain in the linked issue.

t3chguy: > The description and linked issue don't explain this.

Nor do they need to given threads in the js-sdk is marked `@experimental`

> where https://github.com/matrix-org/matrix-js-sdk/pull/2444#discussion_r893025870 anyway.

This isn't true, it fixed the issue for event timeline sets accessing thread roots which caused the issue you saw, and created a regression around accessing thread roots from main timeline event timeline sets

src/models/event-timeline-set.ts
t3chguy: > Seems like we should put this stuff in opts

Why? https://github.com/vector-im/element-meta/pull/216/files doesn't favour it in any way shape or form

MadLittleMods: Feel free to defer in favor of the style-guide.

The function signature here is getting pretty complex and we already have `opts` where this seems like it would fit.

spec/integ/matrix-client-event-timeline.spec.js
MadLittleMods: > Nor do they need to given threads in the js-sdk is marked `@experimental`

While compromises and known problems are probably part of an `@experimental` feature, seems like we shouldn't just drop the ball on all of it. I think this problem generally exists in most PR's though. And our current plan/goal for threads is to make this not experimental where this code is bound to live on.

We should so we can better maintain this. We're not going to remember this fresh context that I am driving for in 6 months when we find another regression. I understand that some of this code will change when we get backend changes to better support threads but that info will just make it easier to refactor this in the future as well to see that we're still maintaining all of the previous behavior plus the extra benefits.


> This isn't true, it fixed the issue for event timeline sets accessing thread roots which caused the issue you were hitting with main timeline events showing up in threads, and created a regression around accessing thread roots from main timeline event timeline sets

I see what you mean. This is an example of what we should be explaining though.

src/client.ts
MadLittleMods: Done in https://github.com/matrix-org/matrix-js-sdk/pull/2454#discussion_r896279149 ✅

src/client.ts
MadLittleMods: Addressed in https://github.com/matrix-org/matrix-js-sdk/pull/2454#discussion_r896234228 ✅

src/client.ts
MadLittleMods: List fleshed out in https://github.com/matrix-org/matrix-js-sdk/pull/2454#discussion_r896282288

spec/integ/matrix-client-event-timeline.spec.js
MadLittleMods: Addressed by https://github.com/matrix-org/matrix-js-sdk/pull/2454#discussion_r896284318

src/models/event-timeline-set.ts
MadLittleMods: Can we add that statement to the JSDoc here? It's great context for why we use it.

PR #2447, opened 2022-06-09
Emit unknown device errors for group call participants without e2e

dbkr:
There are a number of different cases here: there were some before
when dealing with versions that didn't send deviceId. This catches
all of them and makes all these cases emit the same error.

---

## ✨ Features
 * Emit unknown device errors for group call participants without e2e ([\#2447](https://github.com/matrix-org/matrix-js-sdk/pull/2447)).


src/webrtc/callEventHandler.ts
```diff
@@ -210,8 +210,9 @@ export class CallEventHandler {
 
         let opponentDeviceId: string | undefined;
 
+        let groupCall;
```
robintown, line 213: could use a type

robintown (approved):
Looks reasonable otherwise

PR #2454, opened 2022-06-13
Fix issue with `getEventTimeline` returning undefined for thread roots in main timeline

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/22539
Continuation of https://github.com/matrix-org/matrix-js-sdk/pull/2444 (fix https://github.com/matrix-org/matrix-js-sdk/pull/2444#discussion_r893025870)

---

## 🐛 Bug Fixes
 * Fix issue with `getEventTimeline` returning undefined for thread roots in main timeline ([\#2454](https://github.com/matrix-org/matrix-js-sdk/pull/2454)). Fixes vector-im/element-web#22539.


dbkr (approved):

t3chguy:
@MadLittleMods thanks for spotting the culprit in the previous PR, does this address all your concerns?

src/models/event-timeline-set.ts
```diff
@@ -842,6 +842,31 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
...
+     * for timeline sets relating to a thread, only return true for events in the same
+     * thread timeline, for timeline sets not relating to a thread only return true
+     * for events which should be shown in the main room timeline.
+     * Requires the `room` property to have been set at EventTimelineSet construction time.
+     *
+     * @param event {MatrixEvent} the event to check whether it belongs to this timeline set.
+     * @throws {Error} if `room` was not set when constructing this timeline set.
+     * @return {boolean} whether the event belongs to this timeline set.
+     */
+    public eventBelongs(event: MatrixEvent): boolean {
```
MadLittleMods, line 857: `eventAllowedToBelong` seems more accurate.

src/models/event-timeline-set.ts
```diff
@@ -842,6 +842,31 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
...
+        if (!this.room) {
+            throw new Error("Cannot call `EventTimelineSet::eventBelongs without a `room` set. " +
+                "Set the room when creating the EventTimelineSet to call this method.");
+        }
+
+        const { threadId, shouldLiveInRoom } = this.room.eventShouldLiveIn(event);
+
+        if (this.thread) {
+            return this.thread.id === threadId;
+        }
```
MadLittleMods, line 867: I think this will break if the `event` is from a thread but is a main `timelineSet` because this check is skipped in that case.

Previously, always checking `if (timelineSet.thread?.id !== threadId) {` saved us here.

src/models/event-timeline-set.ts
```diff
@@ -842,6 +842,31 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
...
+     *
+     * @param event {MatrixEvent} the event to check whether it belongs to this timeline set.
+     * @throws {Error} if `room` was not set when constructing this timeline set.
+     * @return {boolean} whether the event belongs to this timeline set.
+     */
+    public eventBelongs(event: MatrixEvent): boolean {
+        if (!this.room) {
+            throw new Error("Cannot call `EventTimelineSet::eventBelongs without a `room` set. " +
+                "Set the room when creating the EventTimelineSet to call this method.");
+        }
```
MadLittleMods, line 861: This addresses https://github.com/matrix-org/matrix-js-sdk/pull/2444#discussion_r893003867

src/models/event-timeline-set.ts
```diff
@@ -842,6 +842,31 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
...
+
+        const { threadId, shouldLiveInRoom } = this.room.eventShouldLiveIn(event);
+
+        if (this.thread) {
+            return this.thread.id === threadId;
+        }
+        return shouldLiveInRoom;
+    }
```
MadLittleMods, line 869: We should explain these scenarios that we need to worry about in the JSDoc here. The function description tries to describe this already but we're not able to keep all of these in our head and bugs are creeping in.

 1. If main timeline and thread event, return `undefined`
 1. If thread timeline and main timeline event, return `undefined`
 1. If thread timeline and event from different thread, return `undefined`
 1. If thread timeline and `threadRoot` from different thread, return `undefined`
 1. If the main timeline and main timeline event, return timeline
 1. If thread timeline and event from same thread, return timeline
 1. If main timeline and `threadRoot`, return timeline
 1. If thread timeline and `threadRoot` from same thread, return timeline
 
 Tests cross-referencing the numbered list from the JSDoc would also be good.

spec/integ/matrix-client-event-timeline.spec.js
```diff
@@ -540,20 +555,46 @@ describe("MatrixClient event timelines", function() {
...
+                client.getEventTimeline(timelineSet, THREAD_ROOT.event_id),
+                httpBackend.flushAllExpected(),
+            ]);
+
+            expect(timeline).not.toBe(thread.liveTimeline);
+            expect(timelineSet.getTimelines()).toContain(timeline);
+            expect(timeline.getEvents().find(e => e.getId() === THREAD_ROOT.event_id)).toBeTruthy();
+        });
+
+        it("should return undefined when event is not in the thread that the given timelineSet is representing", () => {
```
MadLittleMods, line 589: Addresses https://github.com/matrix-org/matrix-js-sdk/pull/2444#discussion_r893005753

spec/integ/matrix-client-event-timeline.spec.js
```diff
@@ -540,20 +555,46 @@ describe("MatrixClient event timelines", function() {
             expect(timeline.getEvents().find(e => e.getId() === THREAD_REPLY.event_id)).toBeTruthy();
         });
 
-        it("should return undefined when event is not within a thread but timelineSet is", async () => {
+        it("should return relevant timeline from non-thread timelineSet when asking for the thread root", async () => {
```
MadLittleMods, line 558: The current title is accurate, just trying avoid "non-thread" as we're really dealing with `timelineSet`'s which either represent a thread or the main timeline.

I think this is still accurate:

```suggestion
        it("should return relevant main timeline when event is a threadRoot given the main timelineSet (threadRoots belong to the main timeline and thread)", async () => {
```

spec/integ/matrix-client-event-timeline.spec.js
```diff
@@ -566,14 +607,13 @@ describe("MatrixClient event timelines", function() {
...
-            const timeline = await timelinePromise;
-            expect(timeline).toBeUndefined();
+            return Promise.all([
+                expect(client.getEventTimeline(timelineSet, EVENTS[0].event_id)).resolves.toBeUndefined(),
+                httpBackend.flushAllExpected(),
+            ]);
         });
 
-        it("should return undefined when event is within a thread but timelineSet is not", async () => {
+        it("should return undefined when event is within a thread but timelineSet is not", () => {
```
MadLittleMods, line 616: We can update this to match the pattern of the test name above (outlined in https://github.com/matrix-org/matrix-js-sdk/pull/2444#discussion_r893005753)

src/models/event-timeline-set.ts
t3chguy: > I think this will break with the main timelineSet if the event is from a thread. The check is skipped in that case and we will return true for shouldLiveInRoom

`shouldLiveInRoom` will be false for threaded events unless that event is a thread-root.

MadLittleMods: Thanks for clarifying! Got a bit mixed up after looking at this again. `shouldLiveInRoom` is a bit ambiguous when it means `shouldLiveInMainTimeline`

PR #2455, opened 2022-06-13
Remove unused sessionStorage layer

t3chguy:
None of the methods were called in it anymore, it was used as a way to pass a Web `Storage` object around without a proper abstraction for storing the pending event list, will do it properly instead


robintown (approved):
Looks fine, though it looks like allchange won't do anything with the breaking change tag as long as it's marked as just a task

PR #2461, opened 2022-06-14
Don't block muting on determining whether the device exists

robintown:
This was one thing that could allow you to keep transmitting past 20 seconds in PTT mode on Firefox.

---

## 🐛 Bug Fixes
 * Don't block muting on determining whether the device exists ([\#2461](https://github.com/matrix-org/matrix-js-sdk/pull/2461)).


SimonBrandner (approved):
Otherwise looks sane

PR #2465, opened 2022-06-16
Live location share - add start time leniency (PSF-1081)

kerryarchibald:
Adds leniency so start timestamps that are slightly in the future are still considered live.

---

## ✨ Features
 * Live location share - add start time leniency ([\#2465](https://github.com/matrix-org/matrix-js-sdk/pull/2465)). Contributed by @kerryarchibald.


andybalaam (approved):
Great, thanks!

PR #2470, opened 2022-06-23
Prevent infinite loop in getPushRules

texuf:
During integration tests that start and stop the client quickly, getPushRules fails and doesn't exit cleanly. 

getPushRules fails with `Getting push rules failed M_MISSING_TOKEN: missing access token`, but shouldAbortSync only checks for `M_UNKNOWN_TOKEN`

Tests: pulled down project locally and ran all tests.

Signed-off-by: Austin Ellis <austin@hntlabs.com>

---

## 🐛 Bug Fixes
 * Prevent infinite loop in getPushRules ([\#2470](https://github.com/matrix-org/matrix-js-sdk/pull/2470)). Contributed by @texuf.

### DIFF
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -604,7 +604,10 @@ export class SyncApi {
     }
 
     private shouldAbortSync(error: MatrixError): boolean {
-        if (error.errcode === "M_UNKNOWN_TOKEN") {
+        if (!this.running) {
+            logger.info("No longer running - assuming logout");
+            return true;
+        } else if (error.errcode === "M_UNKNOWN_TOKEN") {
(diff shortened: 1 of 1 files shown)


andybalaam (changes requested):
This looks sensible, but to help me understand it I'd really like to see a new test that failed before your change, and passes now.

texuf:
> This looks sensible, but to help me understand it I'd really like to see a new test that failed before your change, and passes now.

That is a reasonable request. I will give it a shot when I get a minute.

texuf:
I don't currently have this issue. closing.

PR #2482, opened 2022-06-29
Enable DTX on audio tracks in calls

robintown:
This greatly reduces the amount of bandwidth used when transmitting silence.

Here's some sample stats where Firefox and Chrome were both transmitting silence to each other:

.|Before|After
-|-|-
Firefox|50 packets/s, 13 Kb/s|5 packets/s, 700 b/s
Chrome|50 packets/s, 1.2 Kb/s|5 packets/s, 80 b/s

We should investigate at some point why Firefox's bandwidth usage for audio is an order of magnitude higher than Chrome's.

---

## ✨ Features
 * Enable DTX on audio tracks in calls ([\#2482](https://github.com/matrix-org/matrix-js-sdk/pull/2482)).

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
   ],
   "dependencies": {
     "@babel/runtime": "^7.12.5",
+    "@types/sdp-transform": "^2.4.5",
     "another-json": "^0.2.0",
     "browser-request": "^0.3.3",
     "bs58": "^4.0.1",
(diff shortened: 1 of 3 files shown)


dbkr (approved):
Any idea if there are likely to be compatibility issues or anything? Mostly wondering why it isn't enabled by default. If in doubt we could just enable it for PTT mode, but if we think it's safe enough then would be good to have for all calls.

May want to combine all the SDP munging together at some point (ie. with `getRidOfRTXCodecs`).

robintown:
From what I can tell, browsers generally ignore any fmtp params that they don't recognize, so at worst the `usedtx` option will just have no effect. And on the receiving side, I'm pretty sure that any Opus decoder should be able to handle DTX'd audio transparently.

So, yes, I think it's safe enough. The only risk I could find mentioned is slightly lower audio quality, but in the "indistinguishable to the human ear" kind of sense.

PR #2483, opened 2022-07-01
Base support for MSC3840: Ignore invites

Yoric:
This PR expects that most of the work will be done at higher level,
where the user can decide to look at a list of ignored invites,
un-ignore them individually, etc.

MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/3840

---

## ✨ Features
 * Base support for MSC3840: Ignore invites ([\#2483](https://github.com/matrix-org/matrix-js-sdk/pull/2483)). Contributed by @Yoric.


src/client.ts
```diff
@@ -3424,6 +3424,23 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+        if (!event || !event.getContent()) return {};
+        return event.getContent();
+    }
+
+    /**
+     * Update the list of ignored invites.
+     * @param invites The new list of ignored invites.
+     */
+    public setIgnoredInvites(invites: IgnoredInvites): Promise<{}> {
+        return this.setAccountData(IGNORED_INVITES_ACCOUNT_DATA_KEY, invites);
```
Gnuxie, line 3441: Will you de duplicate the ignored rooms? Or is that the responsibility of the caller?

turt2live (commented):
overall lgtm, though needs tests :)

src/client.ts
Yoric: I could change this, but I'm assuming that it is the responsibility of the caller.

Yoric:
Adding a few unit tests.
I'm not convinced that they're testing anything useful, unfortunately.

src/client.ts
```diff
@@ -3424,6 +3424,23 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         return this.getIgnoredUsers().includes(userId);
     }
 
+    /**
+     * Gets the list of currently ignored invites.
+     */
+    public getIgnoredInvites(): IgnoredInvites {
+        const event = this.getAccountData(EventType.IgnoredInvites);
```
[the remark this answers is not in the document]
Yoric: I haven't been able to find exactly where this is implemented. However, as far as I understand, `this.getAccountData` should be fast (because everything is cached to memory).

Also note that you can subscribe to `client.on(ClientEvent.AccountData, event => { ... })` to be informed whenever a piece of account data changes.

spec/unit/matrix-client.spec.ts
```diff
@@ -1297,4 +1297,37 @@ describe("MatrixClient", function() {
             expect(result!.aliases).toEqual(response.aliases);
         });
     });
+
+    describe("ignoring invites", () => {
+        it('stores ignored invites', async function() {
```
turt2live, line 1302: we should be using `should`, and arrow functions:
```suggestion
        it('should store ignored invites', async () => {
```

spec/unit/matrix-client.spec.ts
```diff
@@ -1297,4 +1297,37 @@ describe("MatrixClient", function() {
...
+            // Mockup account data storage.
+            const dataStore = new Map();
+            client.setAccountData = function(eventType, content) {
```
turt2live, line 1305:
```suggestion
            client.setAccountData = (eventType, content) => {
```

spec/unit/matrix-client.spec.ts
```diff
@@ -1297,4 +1297,37 @@ describe("MatrixClient", function() {
...
+                dataStore.set(eventType, content);
+                return Promise.resolve();
+            };
+            client.getAccountData = function(eventType) {
```
turt2live, line 1309:
```suggestion
            client.getAccountData = (eventType) => {
```

spec/unit/matrix-client.spec.ts
```diff
@@ -1297,4 +1297,37 @@ describe("MatrixClient", function() {
...
+                const data = dataStore.get(eventType);
+                return new MatrixEvent({
+                    content: data,
+                });
+            };
+
+            // Initially, the invite list should be empty but not `null`.
+            expect(await client.getIgnoredInvites()).toEqual({});
```
turt2live, line 1317:
```suggestion
            expect(await client.getIgnoredInvites()).toStrictEqual({});
```

src/@types/event.ts
```diff
@@ -81,6 +81,7 @@ export enum EventType {
     PushRules = "m.push_rules",
     Direct = "m.direct",
     IgnoredUserList = "m.ignored_user_list",
+    IgnoredInvites = "org.matrix.msc3840.ignored_invites",
```
turt2live, line 84: we're not supposed to define unstable event types here - we should be using `UnstableValue().name`

src/client.ts
```diff
@@ -3424,6 +3424,23 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         return this.getIgnoredUsers().includes(userId);
     }
 
+    /**
+     * Gets the list of currently ignored invites.
+     */
+    public getIgnoredInvites(): IgnoredInvites {
```
turt2live, line 3430: apologies, should have caught this in the first round of review: this function and the setter will need to be named `unstable_getIgnoredInvites` (and similar for the setter) to make it obvious that the functions can be ripped out. 

Additionally, the return value for both will need to be documented in the jsdoc - see other functions for examples

turt2live, line 3430: also something I should have caught in the previous review was the usefulness of these functions: do we really need them, or can we just have the consumer call `get/setAccountData` on its own?

I realize that ignoring users is here already, but I honestly kinda question that too.

At the very least, I think the getter and setter should return a `{ [roomId]: Metadata }` object rather than the raw account data.

src/client.ts
```diff
@@ -9320,3 +9337,35 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+/**
+ * Metadata on ignored invites.
+ *
+ * This interface is expected to gain additional fields, all
+ * of them optional.
+ */
+export interface IgnoreInviteMetadata {
+    /**
+     * The instant the user decided to ignore this invite,
+     * as a Matrix timestamp.
```
turt2live, line 9350:
```suggestion
     * The timestamp for when the user decided to ignore this invite.
```

src/client.ts
```diff
@@ -9320,3 +9337,35 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+    readonly ts: number;
+
+     /**
+      * A human-readable reason for ignoring this invite.
+      */
+    readonly reason?: string;
+}
+
+/**
+ * Invites to ignore across all devices/sessions, as per MSC 3840.
```
turt2live, line 9361:
```suggestion
 * Invites to ignore across all devices/sessions, as per MSC3840.
 *
 * See https://github.com/matrix-org/matrix-spec-proposals/pull/3840
```

src/client.ts
```diff
@@ -9320,3 +9337,35 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
  * @param {object} data The JSON object returned by the server
  */
+
+/**
+ * Metadata on ignored invites.
+ *
+ * This interface is expected to gain additional fields, all
+ * of them optional.
+ */
+export interface IgnoreInviteMetadata {
```
turt2live, line 9347: All of these interfaces will need a jsdoc line that says they can be removed at any point without notice due to being unstable.

src/client.ts
```diff
@@ -9320,3 +9337,35 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+
+/**
+ * Invites to ignore across all devices/sessions, as per MSC 3840.
+ *
+ * This interface is expected to gain additional fields, all of
+ * them optional.
+ */
+export interface IgnoredInvites {
+    /**
+     * A map of room ids => metadata for rooms the user does not wish to be invited to.
```
turt2live, line 9368:
```suggestion
     * A map of room IDs => metadata for rooms the user does not wish to be invited to.
```

turt2live (changes requested):
generally this seems fine - thanks for adding the test, and apologies for missing some things in my last review.

src/client.ts
Yoric: > also something I should have caught in the previous review was the usefulness of these functions: do we really need them, or can we just have the consumer call get/setAccountData on its own?

In my mind, there are a few reasons to put it in the matrix-js-sdk:

- there may come a time when part of the spec is implemented by the server;
- the consumers for `getIgnoredInvites` and that for `setIgnoredInvites` look like they are going to be in very different parts of the code ([possibly in matrix-js-sdk](https://github.com/matrix-org/matrix-js-sdk/pull/2496)), so having a single source of truth with a single type signature, rather than inlining two untyped calls to `get/setAccountData` feels less accident-prone;

> At the very least, I think the getter and setter should return a { [roomId]: Metadata } object rather than the raw account data.

I don't understand that line. Do you mean that I should make a shallow copy of the account data rather than trusting the `readonly` in the type signature? A deep copy?

turt2live: should be the equivalent of `this.getAccountData("whatever")?.ignored_invites` - no explicit copying needed. The idea is to return the inner structure because the caller shouldn't be concerned with the intricacies of how account data is formed.

Yoric:
Closing in favour of another implementation based on MSC3847.

t3chguy:
@Yoric you said closing in favour of X but didn't actually close it?

Yoric:
(oops)

PR #2484, opened 2022-07-01
Fix some MatrixCall leaks and use a shared AudioContext

robintown:
These leaks, combined with the dozens of AudioContexts floating around in memory across different CallFeeds, could cause some really bad performance issues and audio crashes on Chrome.


dbkr (approved):
lgtm, although tests presumably just need adjusting

PR #2488, opened 2022-07-01
Fix return type on funcs in matrixClient to be optionally null

texuf:
The value can be undefined, but the type was just string

tests: Ran all tests

signed-off-by: austin ellis austin@hntlabs.com

notes: none

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1403,7 +1403,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      *
      * @return {?string} MXID for the logged-in user, or null if not logged in
      */
-    public getUserId(): string {
+    public getUserId(): string | null {
         if (this.credentials && this.credentials.userId) {
             return this.credentials.userId;
         }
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -6806,7 +6806,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * Get the access token associated with this account.
      * @return {?String} The access_token or null
      */
-    public getAccessToken(): string {
+    public getAccessToken(): string | undefined {
         return this.http.opts.accessToken || null;
```
t3chguy, line 6810: These lines contradict each other, undefined != null

texuf: good point! still learning typescript. 
a brief tour of the docs says
1) undefined is a type, null is an object,  https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript Declaring the type to be undefined and then returning null makes sense to me but it's wrong! and will fail type checking in later versions of typescript

2) we should probably use undefined everywhere https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined but we're not, and I see the utility of null

3) there are a couple more places we do this, i'll update this pr to use `| null` and fix the rest

t3chguy (approved):
Looks sane to me, thanks for this!!

PR #2492, opened 2022-07-04
Add config option for e2e group call signalling

dbkr:
---

## ✨ Features
 * Add config option for e2e group call signalling ([\#2492](https://github.com/matrix-org/matrix-js-sdk/pull/2492)).

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -346,6 +346,12 @@ export interface ICreateClientOpts {
      */
     fallbackICEServerAllowed?: boolean;
 
+    /**
+     * If true, to-device signalling for group calls will be encrypted
+     * with Olm. Default: true.
+     */
+    useE2eForGroupCall?: boolean;
+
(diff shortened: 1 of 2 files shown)


SimonBrandner (commented):
Would it make more sense for this to be a per-call setting?

dbkr:
So the reason I didn't do it per-call or anywhere in the signalling is because that then allows the server to do a downgrade attack. This feels a bit harder to force clients to downgrade to not using encryption.

SimonBrandner:
> So the reason I didn't do it per-call or anywhere in the signalling is because that then allows the server to do a downgrade attack. This feels a bit harder to force clients to downgrade to not using encryption.

Sorry, I am not very familiar with this, so I am not seeing how exactly this should happen; though my main concern now is https://github.com/vector-im/element-call/pull/438#discussion_r913241512

SimonBrandner (approved):
LGTM and thank you for answering all my questions!

PR #2495, opened 2022-07-06
Improve VoIP integrations testing

SimonBrandner:
---


spec/unit/webrtc/call.spec.ts
```diff
@@ -506,6 +405,383 @@ describe('Call', function() {
...
+                call_id: call.callId,
+                party_id: 'party_id',
+                selected_party_id: "different_party_id",
+            }),
+        });
+
+        expect(callHangupCallback).toHaveBeenCalled();
+    });
+
+    describe("should handle turn servers", () => {
```
weeman1337, line 733: This block is testing `turnServers` so I would suggest simply putting `turnServers` into `describe`.

Then it reads more fluent like for example: „Call turnServers should fallback if allowed“.
Compared to: „Call should handle turn servers should fallback if allowed“.

weeman1337 (approved):
LGTM :+1:

PR #2496, opened 2022-07-06
don't consider room invites that are ignored to be "visible"

jesopo:
to be rebased on #2483


Yoric (approved):
Looks good to me. There's an ongoing debate about `getIgnoredInvites()` and whether it might become more costly, but for the time being, this should be efficient.

PR #2500, opened 2022-07-07
Don't crash with undefined room in `processBeaconEvents()`

### LINKED ISSUE #2494: Crash on processBeaconEvents due to null room in notificationTimeline
**Describe the bug**
When loading the client timeline set data, `processBeaconEvents` is called on the matrix events for the notification timeline. However, notification timeline has no room associated with it.

The `processBeaconEvents` function expects a room if there are any matrix events.

Here is the callback on the sync event:
image

And the call to processBeaconEvents:
image

**To Reproduce**
Create / start the client
Create a notification timelineset
Request timeline notifications

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 103

This is not an issue in version 16 of the SDK. It is present in 17, 18, and 19.

SimonBrandner:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/2494

---

## 🐛 Bug Fixes
 * Don't crash with undefined room in `processBeaconEvents()` ([\#2500](https://github.com/matrix-org/matrix-js-sdk/pull/2500)). Fixes #2494.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -8947,9 +8947,9 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         room: Room,
         events?: MatrixEvent[],
     ): void {
-        if (!events?.length) {
-            return;
-        }
+        if (!events?.length) return;
+        if (!room) return;
+
         room.currentState.processBeaconEvents(events, this);
     }
 
--- a/src/client.ts
+++ b/src/client.ts
@@ -8944,7 +8944,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 
     public processBeaconEvents(
-        room: Room,
+        room?: Room,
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -8947,9 +8947,9 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         room: Room,
         events?: MatrixEvent[],
     ): void {
-        if (!events?.length) {
-            return;
-        }
+        if (!events?.length) return;
+        if (!room) return;
```
t3chguy, line 8951: L8947 says room must be provided

SimonBrandner: Well, it feels safer to have the if statement inside the method in case something like this happens again though it feels weird to say the room can be falsey, so I am not sure which one I should go with...

t3chguy (approved):

PR #2502, opened 2022-07-07
Prevent double mute status changed events

dbkr:
Audio & video mute status were set in separate calls but share a
mute status changed event, so you'd always get two mute status
changed events emitted. We could suppress events where the mute
status didn't change, but this would still get two events saying
the same thing when they both changed. Instead, merge setAudioMuted
& setVideoMuted into a single call that sets either or both.

---

## 🐛 Bug Fixes
 * Prevent double mute status changed events ([\#2502](https://github.com/matrix-org/matrix-js-sdk/pull/2502)).


SimonBrandner (approved):
Could we please also get this into `develop`? :pray:

PR #2505, opened 2022-07-07
Properly re-insert room ID in bundled thread relation messages from sync

duxovni:
Fixes vector-im/element-web#22094.

---

## 🐛 Bug Fixes
 * Properly re-insert room ID in bundled thread relation messages from sync ([\#2505](https://github.com/matrix-org/matrix-js-sdk/pull/2505)). Fixes vector-im/element-web#22094. Contributed by @duxovni.

### DIFF
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -49,13 +49,14 @@ import {
 import { IThreadBundledRelationship, MatrixEvent } from "./models/event";
 import { MatrixError, Method } from "./http-api";
 import { ISavedSync } from "./store";
-import { EventType, RelationType } from "./@types/event";
+import { EventType } from "./@types/event";
 import { IPushRules } from "./@types/PushRules";
 import { RoomState, RoomStateEvent, IMarkerFoundOptions } from "./models/room-state";
 import { RoomMemberEvent } from "./models/room-member";
 import { BeaconEvent } from "./models/beacon";
 import { IEventsResponse } from "./@types/requests";
 import { IAbortablePromise } from "./@types/partials";
+import { THREAD_RELATION_TYPE } from "./models/thread";
 
 const DEBUG = true;
 
@@ -1601,7 +1602,8 @@ export class SyncApi {
             if (room) {
                 e["room_id"] = room.roomId;
                 if ("unsigned" in e) {
-                    const threadRel: IThreadBundledRelationship = e.unsigned?.["m.relations"]?.[RelationType.Thread];
+                    const threadRel: IThreadBundledRelationship =
+                        e.unsigned?.["m.relations"]?.[THREAD_RELATION_TYPE.name];
                     if (threadRel) {
                         threadRel.latest_event.room_id = room.roomId;
                     }
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -46,10 +46,10 @@ import {
     ISyncResponse,
     ITimeline,
 } from "./sync-accumulator";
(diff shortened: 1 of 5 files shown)


src/sync.ts
```diff
@@ -1600,6 +1601,13 @@ export class SyncApi {
         return (obj.events as Array<IStrippedState | IRoomEvent | IStateEvent | IMinimalEvent>).map(function(e) {
             if (room) {
                 e["room_id"] = room.roomId;
+                if ("unsigned" in e) {
+                    const threadRel: IThreadBundledRelationship =
+                        e.unsigned?.["m.relations"]?.[THREAD_RELATION_TYPE.name];
+                    if (threadRel) {
+                        threadRel.latest_event.room_id = room.roomId;
+                    }
+                }
```
t3chguy, line 1610:
```suggestion
                  const threadRel: IThreadBundledRelationship =
                      e.unsigned?.["m.relations"]?.[THREAD_RELATION_TYPE.name];
                  if (threadRel) {
                      threadRel.latest_event.room_id = room.roomId;
                  }
```

You're already using `e.unsigned?.` so the if check wasn't adding anything

t3chguy: So the spec is unclear on this, it only has a single example of `latest_event` and it doesn't have a room_id. Should we be doing similar trickery when we get the latest event via the /event, /context and /relations APIs? IIRC those do include a room_id but that could be a Synapse idiom

t3chguy: The spec says

> The latest event should be serialised in the same form as the event itself; this includes adding any bundled aggregations for the event (and applying edits).[1](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3440-threading-via-relations.md#user-content-fn-1-b89185a683eddef9c6c0acdc33023cfd)

https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3440-threading-via-relations.md

So I would argue this is a Synapse bug

duxovni: I tried moving the fix to `thread.ts` just now, but that makes my new test fail. It looks like there are other places where we process bundled relations that don't specifically name `latest_event`, so I'd need to track down all of those and apply a fix at each point. I think this *is* the most resilient solution, because it re-inserts the room ID at the earliest point possible before anything else can start trying to process the event. I'm confused about what would be wrong with this approach; worst case, this event already has a room ID and re-inserting the room ID is a no-op. If you'd like, I can add a check so that we only insert a room ID if there isn't one present already?

duxovni: And the purpose of the `if` statement is for type narrowing; ultimately it's somewhat moot because of the `any` in the record type further down the line, but it does give IDEs etc. the ability to type-check things up to `m.relations`. I can remove it if you don't think it's worth it.

duxovni: Ah, I see your point about making sure we're handling bundled events from endpoints other than `/sync` correctly. It looks like in the spec, `/sync` is the only place where `ClientEventWithoutRoomID` is used at all, FWIW.

t3chguy: Yeah but equally the MSC is vague on the language, and also mutating the unsigned field returned from the server isn't something we do elsewhere to my knowledge, so would be good to stay consistent with that as it may confuse developers making use of /devtools

t3chguy (approved):
LGTM but SonarQube is demanding more tests

PR #2508, opened 2022-07-08
Typescriptify crypto integration tests

duxovni:
---

### DIFF
--- a/spec/integ/matrix-client-crypto.spec.ts
+++ b/spec/integ/matrix-client-crypto.spec.ts
@@ -32,24 +32,25 @@ import { logger } from '../../src/logger';
 import * as testUtils from "../test-utils/test-utils";
 import { TestClient } from "../TestClient";
 import { CRYPTO_ENABLED } from "../../src/client";
+import { ClientEvent, IContent, MatrixClient, MatrixEvent } from "../../src/matrix";
 
-let aliTestClient;
+let aliTestClient: TestClient;
 const roomId = "!room:localhost";
 const aliUserId = "@ali:localhost";
 const aliDeviceId = "zxcvb";
 const aliAccessToken = "aseukfgwef";
-let bobTestClient;
+let bobTestClient: TestClient;
 const bobUserId = "@bob:localhost";
 const bobDeviceId = "bvcxz";
 const bobAccessToken = "fewgfkuesa";
-let aliMessages;
-let bobMessages;
+let aliMessages: IContent[];
+let bobMessages: IContent[];
 
 function bobUploadsDeviceKeys() {
     bobTestClient.expectDeviceKeyUpload();
     return Promise.all([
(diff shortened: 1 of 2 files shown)


spec/integ/matrix-client-crypto.spec.ts
```diff
@@ -32,24 +32,25 @@ import { logger } from '../../src/logger';
...
+let bobTestClient: TestClient;
 const bobUserId = "@bob:localhost";
 const bobDeviceId = "bvcxz";
 const bobAccessToken = "fewgfkuesa";
-let aliMessages;
-let bobMessages;
+let aliMessages: IContent[];
+let bobMessages: IContent[];
 
 function bobUploadsDeviceKeys() {
```
turt2live, line 49: return type

duxovni:
Oh right, I was working on top of #2505, I'll merge in develop once that gets merged and that should fix the type errors.

spec/integ/matrix-client-crypto.spec.ts
turt2live: applies throughout to other functions as well (unreachable with line comments)

spec/integ/matrix-client-crypto.spec.ts
```diff
@@ -240,7 +234,7 @@ function bobSendsReplyMessage() {
  * @return {promise} which resolves to the ciphertext for Bob's device.
  */
 function expectAliSendMessageRequest() {
-    return expectSendMessageRequest(aliTestClient.httpBackend).then(function(content) {
+    return expectSendMessageRequest(aliTestClient.httpBackend).then(function(content: IContent) {
```
turt2live, line 237: all these inline functions should probably be converted to arrow functions, but not the end of the world.

spec/integ/matrix-client-crypto.spec.ts
```diff
@@ -593,16 +584,16 @@ describe("MatrixClient crypto", function() {
                 };
                 bobTestClient.httpBackend.when("GET", "/sync").respond(200, syncData);
 
-                const eventPromise = new Promise((resolve, reject) => {
-                    const onEvent = function(event) {
+                const eventPromise = new Promise<MatrixEvent>((resolve) => {
+                    const onEvent = function(event: MatrixEvent) {
                         logger.log(bobUserId + " received event",
-                                    event);
+                            event);
```
turt2live, line 590: I'm surprised `event` is even on its own line?

turt2live (changes requested):
overall looks good - thank you! Mostly just need to add a bunch of return types :)

turt2live (approved):
lgtm.

For future PRs, it's generally best to break each refactor into its own commit so the review is easier, though in this case it worked out well enough. We end up collapsing the commits down to one at merge, so they're effectively free :D

PR #2512, opened 2022-07-11
Remove unstable support for `m.room_key.withheld`

### LINKED ISSUE #2233: Stop using unstable names for withheld codes
https://github.com/matrix-org/matrix-js-sdk/pull/2232 (merged 2022-03-11) added support for using the stable name `m.room_key.withheld`.  We should check if other clients support the stable names too.  And after some time, we should stop sending events with the unstable name.

turt2live:
We no longer send or receive the unstable type.

The [MSC](https://github.com/matrix-org/matrix-spec-proposals/pull/2399) was *merged* to the spec about 2 years ago, so we should be more than safe to remove this sort of support. It was formally released in v1.1 of the spec in November of 2021, well beyond the 2 month window the spec requests of implementation for a grace period.

The only implementation I can see which hasn't changed over is Element Android: https://github.com/matrix-org/matrix-android-sdk2/blob/3adfe4d01c8ff5189ada1373015501256e2a2d57/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/events/model/EventType.kt#L90

Given the massive time scale and spec process here, I'm keen to land this without waiting for Android to fix their event type usage. Otherwise, we risk having to put `org.matrix.room_key.withheld` into the spec, which I am absolutely not going to be happy about :)

Fixes https://github.com/matrix-org/matrix-js-sdk/issues/2233

---

## 🦖 Deprecations
 * Remove unstable support for `m.room_key.withheld` ([\#2512](https://github.com/matrix-org/matrix-js-sdk/pull/2512)). Fixes #2233.

### DIFF
--- a/spec/integ/megolm-integ.spec.js
+++ b/spec/integ/megolm-integ.spec.js
@@ -616,9 +616,6 @@ describe("megolm", function() {
                 event_id: '$event_id',
             });
             aliceTestClient.httpBackend.when(
-                'PUT', '/sendToDevice/org.matrix.room_key.withheld/',
-            ).respond(200, {});
-            aliceTestClient.httpBackend.when(
                 'PUT', '/sendToDevice/m.room_key.withheld/',
             ).respond(200, {});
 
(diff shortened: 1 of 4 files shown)


robintown (approved):
Haven't seen that changelog section before :eyes:

turt2live:
(leaving open so this doesn't go into the RC, giving the crypto team a chance to at least protest my decisions here 😇 )

BillCarsonFr:
Android Matching PR https://github.com/vector-im/element-android/pull/6520

For now will support receiving both unstable and stable, but only send stable. will clear after more clients have updated

turt2live:
release scheduling appears to work out on this - discussed oob.

PR #2518, opened 2022-07-12
Typescriptify megolm integration tests

duxovni:
## Checklist

* [x] Tests written for new code (and old code if feasible)
* [x] Linter and other CI checks pass
* [ ] ~~Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))~~ N/A

I have a separate PR ready to rewrite the tests with async arrow functions.

---
This change is marked as an *internal change* (Task), so will not be included in the changelog.

PR #2519, opened 2022-07-12
Rewrite megolm integration tests with async arrow functions

duxovni:
## Checklist

* [ ] ~~Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))~~ N/A

Pending on #2518.

---

### DIFF
--- a/spec/integ/megolm-integ.spec.ts
+++ b/spec/integ/megolm-integ.spec.ts
@@ -24,13 +24,13 @@ import {
     IContent,
     IEvent,
     IClaimOTKsResult,
+    IJoinedRoom,
     ISyncResponse,
     IDownloadKeyResult,
     MatrixEvent,
     MatrixEventEvent,
 } from "../../src/matrix";
 import { IDeviceKeys } from "../../src/crypto/dehydration";
-import { ISignatures } from "../../src/@types/signed";
 
 const ROOM_ID = "!room:id";
 
@@ -75,18 +75,17 @@ function encryptOlmEvent(opts: {
         type: opts.plaintype || 'm.test',
     };
 
-    const event = {
+    return {
         content: {
             algorithm: 'm.olm.v1.curve25519-aes-sha2',
-            ciphertext: {},
+            ciphertext: {
+                [opts.recipient.getDeviceKey()]: opts.p2pSession.encrypt(JSON.stringify(plaintext)),
+            },
             sender_key: opts.senderKey,
         },
         sender: opts.sender || '@bob:xyz',
         type: 'm.room.encrypted',
     };
-    event.content.ciphertext[opts.recipient.getDeviceKey()] =
-        opts.p2pSession.encrypt(JSON.stringify(plaintext));
-    return event;
 }
 
 // encrypt an event with megolm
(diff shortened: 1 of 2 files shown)


spec/integ/megolm-integ.spec.ts
```diff
@@ -672,118 +672,116 @@ describe("megolm", () => {
...
-            // the completion of the first initialsync hould make Alice
-            // invalidate the device cache for all members in e2e rooms (ie,
-            // herself), and do a key query.
-            aliceTestClient.expectKeyQuery(
-                getTestKeysQueryResponse(aliceTestClient.userId),
-            );
+        // start out with the device unknown - the send should be rejected.
+        try {
+            await aliceTestClient.client.sendTextMessage(ROOM_ID, 'test');
+            throw new Error("sendTextMessage succeeded on an unknown device");
```
robintown, line 717: Rather than putting a `throw` inside the `try` block like this, I think it would be clearer to use Jest's `expect(...).rejects.toThrow(...)` condition

spec/integ/megolm-integ.spec.ts
```diff
@@ -823,89 +823,76 @@ describe("megolm", () => {
...
+            next_batch: 1,
+            rooms: {
+                join: {},
+            },
+        };
+        syncResponse.rooms.join[ROOM_ID] = {
+            timeline: {
+                events: [messageEncrypted],
+            },
+        };
```
robintown, line 889: Could use `join: { [ROOM_ID]: ... }` here again

spec/integ/megolm-integ.spec.ts
```diff
@@ -951,67 +951,64 @@ describe("megolm", () => {
...
+            },
+            rooms: {
+                join: {},
+            },
+        };
+        syncResponse.rooms.join[ROOM_ID] = {
+            timeline: {
+                events: [messageEncrypted],
+            },
+        };
```
robintown, line 1001: same here

spec/integ/megolm-integ.spec.ts
duxovni: Hmmm, it looks like I can use Jest conditions to check that it throws the right kind of error, but I'm not seeing a way to inspect the error's properties in more detail like we're doing here; do you know a way to do that in Jest?

robintown: I think you should be able to do something like

```typescript
.toThrow({
  name: "UnknownDeviceError",
  devices: { [aliceTestClient.userId]: { DEVICE_ID: expect.anything() } },
})
```

robintown:
@duxovni I'm off this week, so I'd recommend asking one of the others for review

duxovni:
Understood, have a good week off! Simon's also showing as busy, @andybalaam do you have spare cycles?

spec/integ/megolm-integ.spec.ts
duxovni: Nope, looks like Jest will only let you test the error message, or test that the error object is an instance of a specific class, but nothing more detailed than that :/ https://jestjs.io/docs/expect#tothrowerror

andybalaam (approved):
Huge diff, so I didn't look at every line, but looks good.

PR #2521, opened 2022-07-13
Fix `getLatestTimeline` not working when the latest event in the room is a threaded message

MadLittleMods:
Fix `getLatestTimeline` not working when the latest event in the room is a threaded message

 - Follow-up to https://github.com/matrix-org/matrix-js-sdk/pull/2299
 - `matrix-react-sdk` PR: https://github.com/matrix-org/matrix-react-sdk/pull/8354

---

See https://github.com/matrix-org/matrix-react-sdk/pull/8354#discussion_r892481473

We also have to keep in mind that we don't want to mix messages in the wrong timelines (main vs threaded timeline):

 - https://github.com/matrix-org/matrix-js-sdk/pull/2444
 - https://github.com/matrix-org/matrix-js-sdk/pull/2454

## Dev notes

## Checklist

---

## 🐛 Bug Fixes
 * Fix `getLatestTimeline` not working when the latest event in the room is a threaded message ([\#2521](https://github.com/matrix-org/matrix-js-sdk/pull/2521)). Contributed by @MadLittleMods.


spec/unit/event-timeline-set.spec.ts
```diff
@@ -143,6 +167,41 @@ describe('EventTimelineSet', () => {
...
+        it("should not add a normal message to the timelineSet representing a thread", () => {
+            const eventTimelineSetForThread = new EventTimelineSet(room, {}, client, thread);
+            const liveTimeline = eventTimelineSetForThread.getLiveTimeline();
+            expect(liveTimeline.getEvents().length).toStrictEqual(0);
+
+            eventTimelineSetForThread.addEventToTimeline(messageEvent, liveTimeline, {
+                toStartOfTimeline: true,
+            });
+            expect(liveTimeline.getEvents().length).toStrictEqual(0);
+        });
```
MadLittleMods, line 204: This tests the two main mixing scenarios. We could add tests for the other mixing scenarios but we're using `canContain` which already has it's own set of tests, https://github.com/matrix-org/matrix-js-sdk/blob/5112340040dbb09e5d9b71cdec183d87ec17638e/spec/unit/event-timeline-set.spec.ts#L227-L293

src/client.ts
```diff
@@ -5288,10 +5288,6 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         ];
 
         if (this.supportsExperimentalThreads()) {
-            if (!timelineSet.canContain(event)) {
-                return undefined;
-            }
```
MadLittleMods, line 5293: Instead of returning `undefined`, we now just ignore the event and don't add it to the given `timelineSet`

MadLittleMods, line 5293: In order to make sure we're not regressing https://github.com/matrix-org/matrix-js-sdk/pull/2444 and https://github.com/matrix-org/matrix-js-sdk/pull/2454 for sure, do we have a specific testing strategy in the Element app to reproduce previously?

src/client.ts
```diff
@@ -5338,9 +5334,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         // There is no guarantee that the event ended up in "timeline" (we might have switched to a neighbouring
         // timeline) - so check the room's index again. On the other hand, there's no guarantee the event ended up
         // anywhere, if it was later redacted, so we just return the timeline we first thought of.
-        return timelineSet.getTimelineForEvent(eventId)
-            ?? timelineSet.room.findThreadForEvent(event)?.liveTimeline // for Threads degraded support
```
MadLittleMods, line 5342: `?? timelineSet.room.findThreadForEvent(event)?.liveTimeline // for Threads degraded support` was added in [#2261](https://github.com/matrix-org/matrix-js-sdk/pull/2261/files#diff-25d66d74617fe2e23d7946bd6e3ba95640ab1b9bc8947445d604fc271c7c1f12R5328). Is it necessary? I'm just trying to get the context behind why we have it.

What does `Threads degraded support` exactly map to?

If `supportsExperimentalThreads` is disabled all events will be added to the main timeline ([`eventShouldLiveIn` always returns `shouldLiveInRoom: true`](https://github.com/matrix-org/matrix-js-sdk/blob/5112340040dbb09e5d9b71cdec183d87ec17638e/src/models/room.ts#L1718-L1720))

t3chguy: > Threads degraded support

Server has no APIs to help the client so the client has to build the threads objects without filters & relations support on the server

src/client.ts
```diff
@@ -5234,15 +5234,15 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * <p>If the EventTimelineSet object already has the given event in its store, the
      * corresponding timeline will be returned. Otherwise, a /context request is
      * made, and used to construct an EventTimeline.
-     * If the event does not belong to this EventTimelineSet then undefined will be returned.
+     * If the event does not belong to this EventTimelineSet then it will ignored.
```
t3chguy, line 5237: The event will be ignored? What does that even mean? This method isn't meant to process the given event, just use it as a pointer. What event timeline will be returned to the caller? The caller would now have need to manually check that the returned timeline is valid for what they asked for, I guess by your other change lower down by attempting to add an event and by asserting that it worked, that seems rather strange

MadLittleMods: Let's first discuss the previous behavior. Previously, when the event didn't belong to the `timelineSet`, we would return `undefined` and create no additional timelines. Returning `undefined` has no special meaning given that all of the usage is always fire and forget meaning we don't use the `timeline` returned by `client.getEventTimeline()` and only use the function to load the event so it's available to the client.

In our usage, there is only a single spot where the caller [uses the `timeline` returned by `client.getEventTimeline()`](https://github.com/matrix-org/matrix-react-sdk/blob/f5f79158ed5a950626bc89439b8319e362a75de1/src/components/structures/LoggedInView.tsx#L356). This usage should be replaced by `room.findEventById(eventId)` because it doesn't even use the `timeline`, it just wants the event that was loaded in as well.

It's probably a misnomer to call it `getEventTimeline(): timeline` in the first place as it's more accurately used as `loadEventInTimeline(): Promise<void>` everywhere.

---

We're only working within the given `timelineSet` that was passed in. If `client.getEventTimeline()` is really meant to just give the timeline for the `eventId`, then we should just provide the `room` instead which can look at all of the timelineSets (`room.timelineSets`).

Previously, we would return `undefined` in the case where the event doesn't belong in the `timelineSet`. Now we're returning a `timeline` in the `timelineSet` where all of the events returned by `/context` can go. This means events that we fetched, are actually added and not wasted. And it means that the main room timeline can be populated regardless if the `eventId` passed in was a threaded reply.

By ignored, I mean if the `eventId` is a threaded reply, it won't be added to the main room `timelineSet` for example.

src/client.ts
MadLittleMods: @t3chguy Any hints here?

src/client.ts
MadLittleMods: Thanks for the context!

I think it's safe to drop this case given `getTimelineForEvent` is always used in a fire and forget manner, see https://github.com/matrix-org/matrix-js-sdk/pull/2521#discussion_r920214132

src/client.ts
turt2live: Our usage might be to be a fire-and-forget pattern (or wrong, in the case of needing to use `findEventById` instead), but as a public function and SDK we have to maintain a rationalized contract for the function name: it says it gets an event timeline, so it should do that (returning `undefined` if needed)

We can adjust our code to instead use a new `updateEventTimeline()` function or similar, but the existing `getEventTimeline` function can't realistically have a behavioural change like this.

turt2live (changes requested):

src/client.ts
MadLittleMods: The `return undefined` part is part of `supportsExperimentalThreads` which changed in https://github.com/matrix-org/matrix-js-sdk/pull/2444/files

Can we change the experimental implementation?

spec/integ/matrix-client-event-timeline.spec.ts
```diff
@@ -649,13 +649,40 @@ describe("MatrixClient event timelines", function() {
                     };
                 });
 
-            return Promise.all([
-                expect(client.getEventTimeline(timelineSet, EVENTS[0].event_id)).resolves.toBeUndefined(),
-                httpBackend.flushAllExpected(),
-            ]);
+            // getEventTimeline -> thread.fetchInitialEvents
+            httpBackend.when("GET", "/rooms/!foo%3Abar/relations/" +
+                encodeURIComponent(THREAD_ROOT.event_id) + "/" +
```
MadLittleMods, line 654: There are some lints in CI: https://github.com/matrix-org/matrix-js-sdk/runs/8077711103?check_suite_focus=true

But I don't see them locally even after re-installing `node_modules` to ensure correct versions.
```
$ yarn lint
yarn run v1.22.18
$ yarn lint:types && yarn lint:js
$ tsc --noEmit
$ eslint --max-warnings 0 src spec
✨  Done in 21.60s.
```

---

This is also the same pattern we use in the existing tests here but don't appear because they're not in the diff so will need to refactor this.

t3chguy: Those are shown if you use `tsc --strict` - hence being reported by the `Typescript Strict Error Checker`

src/client.ts
turt2live: I'm still not comfortable with the behavioural change here, sorry. While our usage might be fire-and-forget, we can't guarantee that *all* usages of the function are fire-and-forget. The documentation and function itself are not experimental in nature as well, preventing us from making arbitrary breaking changes.

Adding a function is more code, but I think it's worthwhile here. It can even call `getEventTimeline()` and ignore the return value - it looks a bit silly, but it's how we avoid unnecessary major version releases.

turt2live (changes requested):
apologies for the late review on this - it somehow got lost in my queue :(

spec/integ/matrix-client-event-timeline.spec.ts
MadLittleMods: Can probably merge `develop` to resolve all of these unrelated errors now that https://github.com/matrix-org/matrix-js-sdk/pull/2835 fixed them up

src/timeline-window.ts
```diff
@@ -135,11 +135,12 @@ export class TimelineWindow {
...
             }
 
-            return this.client.getEventTimeline(this.timelineSet, initialEventId).then(initFields);
+            await this.client.getEventTimeline(this.timelineSet, initialEventId).then(initFields);
+            return;
         } else {
             const tl = this.timelineSet.getLiveTimeline();
             initFields(tl);
-            return Promise.resolve();
+            return;
```
MadLittleMods, line 143: Split out to https://github.com/matrix-org/matrix-js-sdk/pull/2849

src/models/event-timeline-set.ts
```diff
@@ -695,6 +695,19 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
...
+                "in timelineSet(threadId=${this.thread?.id})`);
+        }
+
+        // Make sure events don't get mixed in timelines they shouldn't be in
+        // (e.g. a threaded message should not be in the main timeline).
+        if (!this.canContain(event)) {
+            logger.warn(`EventTimelineSet.addEventToTimeline: Ignoring event=${event.getId()} that does not belong " + 
+                "in timeline=${timeline.toString()} timelineSet(threadId=${this.thread?.id})`);
+            return;
+        }
```
MadLittleMods, line 709: Split out to https://github.com/matrix-org/matrix-js-sdk/pull/2848

src/client.ts
MadLittleMods: Conversation continued at https://github.com/matrix-org/matrix-js-sdk/pull/2852#discussion_r1013894123

MadLittleMods:
Closed as I went a different direction in this v2 PR, https://github.com/matrix-org/matrix-js-sdk/pull/2852

PR #2522, opened 2022-07-13
Prevent double mute status changed events

dbkr:
Audio & video mute status were set in separate calls but share a
mute status changed event, so you'd always get two mute status
changed events emitted. We could suppress events where the mute
status didn't change, but this would still get two events saying
the same thing when they both changed. Instead, merge setAudioMuted
& setVideoMuted into a single call that sets either or both.

Port of https://github.com/matrix-org/matrix-js-sdk/pull/2502 from
group call branch

## Checklist

---

### DIFF
--- a/src/webrtc/callFeed.ts
+++ b/src/webrtc/callFeed.ts
@@ -184,21 +184,18 @@ export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap> 
     }
 
     /**
-     * Set feed's internal audio mute state
-     * @param muted is the feed's audio muted?
-     */
-    public setAudioMuted(muted: boolean): void {
-        this.audioMuted = muted;
-        this.speakingVolumeSamples.fill(-Infinity);
-        this.emit(CallFeedEvent.MuteStateChanged, this.audioMuted, this.videoMuted);
-    }
-
-    /**
-     * Set feed's internal video mute state
+     * Set one or both of feed's internal audio and video video mute state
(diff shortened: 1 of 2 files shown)


src/webrtc/callFeed.ts
```diff
@@ -184,21 +184,18 @@ export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap>
...
-
-    /**
-     * Set feed's internal video mute state
+     * Set one or both of feed's internal audio and video video mute state
+     * Either value may be null to leave it as-is
      * @param muted is the feed's video muted?
      */
-    public setVideoMuted(muted: boolean): void {
-        this.videoMuted = muted;
+    public setAudioVideoMuted(audioMuted: boolean, videoMuted: boolean): void {
```
t3chguy, line 191: Marked this as a breaking change, any chance you fancy keeping the old ones around for compatibility but just calling the new one under the hood?

t3chguy (approved):
LGTM otherwise

src/webrtc/callFeed.ts
dbkr: This is just internal js-sdk API, so shouldn't need to be a breaking change (although of course it's a public method on an object that the API consumer gets given, so they could call it, but it wouldn't really do anything useful).

weeman1337 (approved):

SimonBrandner:
@dbkr, I think this can be merged now?

dbkr:
so it can!

PR #2526, opened 2022-07-14
Make SDP munging media type specific

dbkr:
We were trying to apply modifications to all media types which led
to confusing warning messages saying opus wasn't present (when it
was for the video stream). Make the modifications media-type specific
to avoid this.

## Checklist

---

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -93,10 +93,20 @@ interface AssertedIdentity {
     displayName: string;
 }
 
+enum MediaType {
+    AUDIO = "audio",
+    VIDEO = "video",
+}
+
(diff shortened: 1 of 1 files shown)


src/webrtc/call.ts
```diff
@@ -94,13 +94,13 @@ interface AssertedIdentity {
 }
 
 // Used internally to specify modifications to codec parameters in SDP
-interface CodecParams {
+interface CodecParamsMod {
+    mediaType: string;
+    codec: string;
```
robintown, line 99: I think an enum would make sense for `mediaType`, if not also `codec`

robintown (approved):

PR #2528, opened 2022-07-15
Add support for sending user-defined encrypted to-device messages

robintown:
This is a port of the same changes from the robertlong/group-call branch.

Depends on https://github.com/matrix-org/matrix-mock-request/pull/18

## Checklist

---

## ✨ Features
 * Add support for sending user-defined encrypted to-device messages ([\#2528](https://github.com/matrix-org/matrix-js-sdk/pull/2528)).

### DIFF
--- a/spec/unit/crypto.spec.ts
+++ b/spec/unit/crypto.spec.ts
@@ -65,6 +65,10 @@ describe("Crypto", function() {
         return Olm.init();
     });
 
+    afterEach(() => {
+        jest.useRealTimers();
+    });
+
     it("Crypto exposes the correct olm library version", function() {
         expect(Crypto.getOlmVersion()[0]).toEqual(3);
     });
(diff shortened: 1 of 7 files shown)


turt2live (commented):
Clearing code review while draft

spec/unit/crypto.spec.ts
```diff
@@ -451,8 +454,10 @@ describe("Crypto", function() {
             // cancelAndResend will call sendToDevice twice:
             // the first call to sendToDevice will be the cancellation
             // the second call to sendToDevice will be the key request
-            expect(aliceClient.sendToDevice).toBeCalledTimes(3);
-            expect(aliceClient.sendToDevice.mock.calls[2][2]).not.toBe(txnId);
+            expect(aliceSendToDevice).toBeCalledTimes(3);
+            expect(aliceSendToDevice.mock.calls[2][2]).not.toBe(txnId);
+
+            jest.useRealTimers();
```
dbkr, line 460: I'd suggest putting this in the afterEach - it doesn't matter if it's called after some tests where it doesn't need to be, but if tests throw exceptions then this would need to be in a try/finally, so afterEach is tidier.

dbkr (approved):
otherwise lgtm, thanks!

PR #2553, opened 2022-07-29
Add support for group calls using MSC3401

SimonBrandner:
Implementation of https://github.com/matrix-org/matrix-spec-proposals/pull/3401

---

## ✨ Features
 * Add support for group calls using MSC3401 ([\#2553](https://github.com/matrix-org/matrix-js-sdk/pull/2553)).


.github/CODEOWNERS
```diff
@@ -1 +1 @@
-* @matrix-org/element-web
+* @matrix-org/element-call-reviewers
```
turt2live, line 1: this change needs reverting, unless you're planning on taking ownership for the whole js-sdk 😇

src/webrtc/callEventHandler.ts
```diff
@@ -40,6 +40,10 @@ export class CallEventHandler {
     calls: Map<string, MatrixCall>;
     callEventBuffer: MatrixEvent[];
     candidateEventsByCall: Map<string, Array<MatrixEvent>>;
+    nextSeqByCall: Map<string, number> = new Map();
+    toDeviceEventBuffers: Map<string, Array<MatrixEvent>> = new Map();
```
turt2live, line 44: public / private / protected (applies throughout)

turt2live (commented):
I've spot-checked a handful of files and aside from the comments raised here it looks like there's a lot of needless import changes in files - it'd be nice to clean those up for a clean diff, but not required.

A question would also be whether we want this to be a useful commit history and do a merge commit on the PR or hide it all behind a squashed commit.

.github/CODEOWNERS
SimonBrandner: I think we should perhaps take ownership of `/webrtc`?

dbkr: yeah, that would probably make sense.

dbkr:
I think the import changes are all valid as far as I can see, and are changing 'matrix' import to import directly from the original place instead.

robintown:
The import changes were necessary to prevent import loops, if I recall. It might not be possible to remove them, but I suppose we could put them in a separate PR if necessary.

richvdh:
sorry to weigh in, but

> I'd vote for a merge commit - there's lots of useful history here in the individual PRs to the branch so it would be a shame to lose that.

I have a fairly strong counteropinion on this. Each git commit that gets merged to your project's mainline should be an atomic change that leaves the system in an operational state. There are hundreds of commits in this branch, many of them called things like "fix the tests". If the branch had been maintained with a coherent history then it might be reasonable to merge it, but that's not what's happened here.

"it would be a shame to lose this" is a spurious argument. Github keeps all of the commits comprising a squash-merged PR forever, so if at some point in the future we want to understand how this evolved, we can retrieve it.

I've previously written about this at https://matrix-org.github.io/synapse/latest/development/git.html#on-keeping-the-commit-history-clean.

dbkr:
Ah, the initial work on this branch was individual commits to it before we started PRing into it, which is annoying. We could bundle those commits up into a separate PR and reconstruct another branch, although that sounds like a way to lose another afternoon on this in addition to the hours of merging. Squash-merge it is, then.

turt2live:
(my counter-counter-argument for doing a merge commit would have been `git blame` support, but it is indeed a bit of a smell if we're considering a merge commit at all - normally that'd mean we should have had 300+ PRs. For this case though, any merge strategy seems not-great, so picking the one the button defaults to would be fine)

turt2live (approved):
spot check looks fine

PR #2556, opened 2022-08-01
Feat: Voice activity threshold

hugohutri:
Adding voice activity detection feature

More information at https://github.com/vector-im/element-call/pull/492

Notes: none

---
This change has no change notes, so will not be included in the changelog.


dbkr (commented):
I've left more feedback over on the element call PR, but specifically to this PR, remember to check for your editor reformatting files before committing.

hugohutri:
Signed-off-by: Hugo Hutri hugo.hutri98@gmail.com

DashieTM:
Signed-off-by: Fabio Lenherr / DashieTM [fabio.lenherr@gmail.com](mailto:fabio.lenherr@gmail.com)

spec/unit/webrtc/callFeed.spec.ts
```diff
@@ -85,5 +85,60 @@ describe("CallFeed", () => {
...
+                feed.setVoiceActivityThreshold(Infinity);
+
+                setTimeout(() => {
+                    expect(feed.stream.getAudioTracks()[0].enabled).toBe(false);
+                }, 1000);
+            });
+
+            it("should enable track", () => {
+                feed.stream.addTrack(
+                    //@ts-ignore
```
SimonBrandner, line 105: What are we ignoring here?

spec/unit/webrtc/callFeed.spec.ts
```diff
@@ -85,5 +85,60 @@ describe("CallFeed", () => {
                 expect(feed.isVideoMuted()).toBeTruthy();
             });
         });
+
+        describe("voice activity detection", () => {
+            it("should disable track", () => {
+                feed.stream.addTrack(
+                    //@ts-ignore
```
SimonBrandner, line 92: What are we ignoring here?

spec/unit/webrtc/callFeed.spec.ts
```diff
@@ -85,5 +85,60 @@ describe("CallFeed", () => {
...
+
+                feed.setVoiceActivityThreshold(-50);
+                feed.speakingVolumeSamples = [-40];
+
+                setTimeout(() => {
+                    expect(feed.stream.getAudioTracks()[0].enabled).toBe(true);
+                }, 1000);
+            });
+
+            it("should enable track", () => {
```
SimonBrandner, line 129:
```suggestion
            it("should disable track", () => {
```

src/webrtc/callFeed.ts
```diff
@@ -21,7 +21,7 @@ import { RoomMember } from "../models/room-member";
 import { logger } from "../logger";
 import { TypedEventEmitter } from "../models/typed-event-emitter";
 
-const POLLING_INTERVAL = 200; // ms
+const POLLING_INTERVAL = 1; // ms
```
SimonBrandner, line 24: I wonder what is the correct value to put here - has there been research done on this?

src/webrtc/callFeed.ts
```diff
@@ -39,6 +39,8 @@ export interface ICallFeedOpts {
      * Whether or not the remote SDPStreamMetadata says video is muted
      */
     videoMuted: boolean;
+
+    setVADMute?: (muted: boolean) => void;
```
SimonBrandner, line 43: This might be obvious but what does VAD mean/stand for?

src/webrtc/callFeed.ts
```diff
@@ -47,26 +49,33 @@ export enum CallFeedEvent {
...
     VolumeChanged = "volume_changed",
     Speaking = "speaking",
+    VoiceActivityThresholdChanged = "voice_activity_threshold_changed",
 }
 
 type EventHandlerMap = {
     [CallFeedEvent.NewStream]: (stream: MediaStream) => void;
-    [CallFeedEvent.MuteStateChanged]: (audioMuted: boolean, videoMuted: boolean) => void;
     [CallFeedEvent.LocalVolumeChanged]: (localVolume: number) => void;
+    [CallFeedEvent.MuteStateChanged]: (audioMuted: boolean, videoMuted: boolean) => void;
```
SimonBrandner, line 58: I think there is no need to change the line order here

src/webrtc/callFeed.ts
```diff
@@ -47,26 +49,33 @@ export enum CallFeedEvent {
     LocalVolumeChanged = "local_volume_changed",
     VolumeChanged = "volume_changed",
     Speaking = "speaking",
+    VoiceActivityThresholdChanged = "voice_activity_threshold_changed",
```
SimonBrandner, line 52: I can't see this ever being emitted - am I blind?

src/webrtc/callFeed.ts
SimonBrandner, line 43: I feel like this should be an event instead of a call back

src/webrtc/groupCall.ts
```diff
@@ -472,6 +473,12 @@ export class GroupCall extends TypedEventEmitter<
 
         for (const call of this.calls) {
             call.localUsermediaFeed.setAudioVideoMuted(muted, null);
+
+            // Disable voice activity detection when user has muted their microphone
+            // (This can't happen when push-to-talk is enabled)
```
SimonBrandner, line 478: How so?

src/webrtc/groupCall.ts
```diff
@@ -509,6 +516,40 @@ export class GroupCall extends TypedEventEmitter<
...
+        if (!(await this.client.getMediaHandler().hasAudioDevice())) {
+            return false;
+        }
+
+        for (const call of this.calls) {
+            call.localUsermediaFeed.setVadMuted(muted, null);
+        }
+
+        if (this.localCallFeed) {
+            // logger.log(
```
SimonBrandner, line 532: This seams like a leftover

SimonBrandner (commented):
Thank you for your work! Will review EC tomorrow 

(might add more comments once on a computer)

src/webrtc/callFeed.ts
DashieTM: Okay, this change was necessary in order to achieve low latency activation. Before the looper just checked whether or not you are loud enough (and for an extended period) to be considered speaking. 
Now we want instant checks on whether or not the user is loud enough to activate their audio track which needs this low latency change. Otherwise as dbkr mentioned, the first part of a sentence etc will be cut off.

src/webrtc/callFeed.ts
DashieTM: Voice Activation Detection

We can change it, or perhaps just make a comment on it?

src/webrtc/groupCall.ts
DashieTM: Push to talk and voice activity detection kinda together kind of don't make sense.
When you use push to talk, you explicitly want to transmit. When using voice activity, you just want to send when the you reach the threshold.

We can change this obviously, but this is how it is done everywhere else as well.

src/webrtc/callFeed.ts
DashieTM: Element-Call side. Used to set the threshold from the settings menu.

SimonBrandner: Ah, of course!

src/webrtc/groupCall.ts
hugohutri: Now when I think about it, (this is more of element call thing) but we should probably clearly indicate that it is either PPT or VAD in the settings modal, so the user would clearly know that which one is used

src/webrtc/groupCall.ts
```diff
@@ -472,6 +473,12 @@ export class GroupCall extends TypedEventEmitter<
 
         for (const call of this.calls) {
             call.localUsermediaFeed.setAudioVideoMuted(muted, null);
+
+            // Disable voice activity detection when user has muted their microphone
+            // (This can't happen when push-to-talk is enabled)
+            if (!this.isPtt) {
+                call.localUsermediaFeed.VADEnabled = !muted;
+            }
```
SimonBrandner, line 481: So I don't think we should be doing anything here - I assume we would only want VAD enabled in the `GroupCall`'s `localUsermediaFeed` and then use that to inform if the other clones should be muted

SimonBrandner: I.e. this should be under `if (this.localCallFeed) {...`

SimonBrandner (commented):
A few more comments :)

I am also re-requesting the team review again, as I would like to know if they have other thoughts

src/webrtc/callFeed.ts
DashieTM: Will test, with 10ms small things like snipping fingers will no longer be recognized.

DashieTM: Yeah, so I did some more testing, and sometimes it does indeed cut off a very small portion of the start.
I think the only solution here is to A: directly manipulate the audio, which I have no idea how to do. 
or B: introduce a small amount of latency to properly handle the audio before sending. 

Perhaps you have an idea here, sorry.

SimonBrandner:
(click the wheel next to my name in the top right corner to re-request my review once ready, same for EC PR)

DashieTM:
Will do, for now I have to find a way to remove the delay.

src/webrtc/callFeed.ts
```diff
@@ -55,18 +65,26 @@ type EventHandlerMap = {
...
 export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap> {
     public stream: MediaStream;
+    public secondStream: MediaStream;
+    public audioDelay: DelayNode;
     public sdpMetadataStreamId: string;
     public userId: string;
     public purpose: SDPStreamMetadataPurpose;
     public speakingVolumeSamples: number[];
+    public voiceActivityThreshold: number;
+    public setVoiceActivityDetectionMute: (muted: boolean) => void;
```
SimonBrandner, line 80: I think you started working on replacing this with `VADMuteStateChanged` but did not finish it?

SimonBrandner, line 80: I think you started working on replacing this with `VADMuteStateChanged` but did not finish it?

src/webrtc/callFeed.ts
```diff
@@ -76,10 +94,16 @@ export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap>
...
+    /**
+     * Cooldown for voice activity detection, so that we don't mute immediately when the user stops speaking
+     * But when he has been silent 200ms
+     */
+    private VADCooldownStarted = new Date();
 
     constructor(opts: ICallFeedOpts) {
         super();
 
+        this.stream = opts.stream;
```
SimonBrandner, line 106: This should be handled by `updateStream()`

SimonBrandner, line 106: This should be handled by `updateStream()`

src/webrtc/callFeed.ts
```diff
@@ -129,9 +167,19 @@ export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap>
         this.analyser.fftSize = 512;
         this.analyser.smoothingTimeConstant = 0.1;
 
-        const mediaStreamAudioSourceNode = this.audioContext.createMediaStreamSource(this.stream);
+        this.secondStream = this.stream.clone();
```
SimonBrandner, line 170: Could this have a bit more descriptive name than `secondStream` please?

src/webrtc/callFeed.ts
```diff
@@ -129,9 +167,19 @@ export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap>
...
+        const mediaStreamAudioSourceNode = this.audioContext.createMediaStreamSource(this.secondStream);
         mediaStreamAudioSourceNode.connect(this.analyser);
 
+        const streamNode = this.audioContext.createMediaStreamSource(this.stream);
+        this.audioDelay = this.audioContext.createDelay(0.001);
```
SimonBrandner, line 175: Could the delay be a `const` please?

src/webrtc/callFeed.ts
```diff
@@ -129,9 +167,19 @@ export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap>
...
+        const gainNode = this.audioContext.createGain();
```
SimonBrandner, line 176: Why do we need a gain node?

SimonBrandner, line 176: Why do we need a gain node?

src/webrtc/callFeed.ts
```diff
@@ -204,6 +252,13 @@ export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap>
         }
         if (videoMuted !== null) this.videoMuted = videoMuted;
         this.emit(CallFeedEvent.MuteStateChanged, this.audioMuted, this.videoMuted);
+        this.VADEnabled = !audioMuted;
+    }
+
+    public setVoiceActivityDetectionMuteLocal(audioMuted: boolean | null, videoMuted: boolean | null): void {
```
SimonBrandner, line 258:
```suggestion
    public setVoiceActivityDetectionMuteLocal(audioMuted: boolean | null): void {
```

src/webrtc/groupCall.ts
```diff
@@ -509,6 +518,31 @@ export class GroupCall extends TypedEventEmitter<
         return true;
     }
 
+    public async setVadMicrophoneMuted(muted: boolean): Promise<boolean> {
+        // hasAudioDevice can block indefinitely if the window has lost focus,
+        // and it doesn't make much sense to keep a device from being muted, so
+        // we always allow muted = true changes to go through
+        if (!(await this.client.getMediaHandler().hasAudioDevice())) {
+            return false;
+        }
```
SimonBrandner, line 527: We need to check `muted` here, like in the other methods

src/webrtc/groupCall.ts
```diff
@@ -509,6 +518,31 @@ export class GroupCall extends TypedEventEmitter<
...
+
+        if (this.localCallFeed) {
+            this.localCallFeed.setVoiceActivityDetectionMuteLocal(muted, null);
+
+            setTracksEnabled(this.localCallFeed.stream.getAudioTracks(), !muted);
+        }
+
+        for (const call of this.calls) {
+            setTracksEnabled(call.localUsermediaFeed.stream.getAudioTracks(), !muted);
+        }
```
SimonBrandner, line 541:
```suggestion
        for (const call of this.calls) {
            call.localUsermediaFeed.setVoiceActivityDetectionMuteLocal(muted, null);
            setTracksEnabled(call.localUsermediaFeed.stream.getAudioTracks(), !muted);
        }

        if (this.localCallFeed) {
            this.localCallFeed.setVoiceActivityDetectionMuteLocal(muted, null);
            setTracksEnabled(this.localCallFeed.stream.getAudioTracks(), !muted);
        }
```

Why not the above?

spec/unit/webrtc/callFeed.spec.ts
```diff
@@ -85,5 +85,60 @@ describe("CallFeed", () => {
                 expect(feed.isVideoMuted()).toBeTruthy();
             });
         });
+
+        describe("voice activity detection", () => {
```
SimonBrandner, line 89: The test names could be more descriptive here, also, would it be possible to test the cooldown?

SimonBrandner (commented):
Thank you for the continued work on this!

SimonBrandner:
Also, please resolve the merge conflicts

src/webrtc/callFeed.ts
DashieTM: This was the hard part to understand, as far as I am aware now, if you don't create a sourcenode, then you can leave this out and it will automatically take care of your audio. This means add automatic gain and more.
If you however do it manually, then you have to add a gainnode yourself.
I noticed that since the audio was suddenly different to what it was before.

With the gainnode the audio goes back to normal, although it would be very beneficial if other people could test this out as well, since I only have 1 pair of ears.

src/webrtc/groupCall.ts
DashieTM: I have a check on Callfeed, or rather muting manually turns vad off, as we do not need to check whether or not the user is loud enough when we are muted either way. 
This means this function won't be called when you are muted manually.

src/webrtc/callFeed.ts
```diff
@@ -76,6 +85,12 @@ export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap>
     private speakingThreshold = SPEAKING_THRESHOLD;
     private speaking = false;
     private volumeLooperTimeout: ReturnType<typeof setTimeout>;
+    /**
+     * Cooldown for voice activity detection, so that we don't mute immediately when the user stops speaking
+     * But when he has been silent 100ms
+     */
+    private VADCooldownStarted = new Date();
```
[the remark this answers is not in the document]
hugohutri: Yea I think this is more readable and easier to debug if something goes wrong. Also setTimeout could be called multiple times and I think it would become a little more complex. 

I think we should leave this as is, and potentially change it later if someone really wants to

SimonBrandner (changes requested):
There still are a few unresolved threads (e.g. test naming, emitting events)... do you need help with them?

spec/unit/webrtc/callFeed.spec.ts
hugohutri: Done :heavy_check_mark:

src/webrtc/callFeed.ts
hugohutri: Ah, Now we have replaced it :turtle:

DashieTM:
Sorry missed 2 of those, should be addressed now.
Unless we have missed some more. 
Most are outdated, which means we did work on it.

spec/unit/webrtc/callFeed.spec.ts
```diff
@@ -85,5 +85,129 @@ describe("CallFeed", () => {
...
+                setTimeout(() => {
+                    feed.speakingVolumeSamples = [-Infinity];
+                }, 100);
+
+                setTimeout(() => {
+                    expect(feed.stream.getAudioTracks()[0].enabled).toBe(false);
+                }, 310);
+            });
+
+            it("voice activity cooldown should be reseted when speaking", async () => {
```
SimonBrandner, line 188:
```suggestion
            it("voice activity cooldown should be reset when speaking", async () => {
```

spec/unit/webrtc/callFeed.spec.ts
```diff
@@ -85,5 +85,129 @@ describe("CallFeed", () => {
...
+                // Cooldown is reseted after speaking again
```
SimonBrandner, line 189:
```suggestion
                // Cooldown is reset after speaking again
```

spec/unit/webrtc/callFeed.spec.ts
```diff
@@ -85,5 +85,129 @@ describe("CallFeed", () => {
...
+                    new MockMediaStreamTrack("track", "audio", true),
+                );
+
+                feed.setVoiceActivityThreshold(-Infinity);
+                setTimeout(() => {
+                    expect(feed.stream.getAudioTracks()[0].enabled).toBe(true);
+                }, 1000);
+            });
+
+            it("voice activity should enable audio track", () => {
```
SimonBrandner, line 115:
```suggestion
            it("enables track when volume is above threshold", () => {
```

spec/unit/webrtc/callFeed.spec.ts
```diff
@@ -85,5 +85,129 @@ describe("CallFeed", () => {
...
+
+                feed.setVoiceActivityThreshold(-50);
+                feed.speakingVolumeSamples = [-40];
+
+                setTimeout(() => {
+                    expect(feed.stream.getAudioTracks()[0].enabled).toBe(true);
+                }, 1000);
+            });
+
+            it("voice activity should disable audio track", () => {
```
SimonBrandner, line 129:
```suggestion
            it("disables track when volume is below threshold", () => {
```

spec/unit/webrtc/callFeed.spec.ts
```diff
@@ -85,5 +85,129 @@ describe("CallFeed", () => {
...
+                feed.stream.addTrack(
+                    //@ts-ignore Mock
+                    new MockMediaStreamTrack("track", "audio", true),
+                );
+
+                feed.setVoiceActivityThreshold(-Infinity);
+                setTimeout(() => {
+                    expect(feed.stream.getAudioTracks()[0].enabled).toBe(true);
+                }, 1000);
+            });
```
SimonBrandner, line 113: Do these tests test something specific that is not tested by the two tests below?

SimonBrandner: Are they testing if `VADEnabled === true` ? Could that be more explicit?

src/webrtc/callFeed.ts
SimonBrandner: Ok, that is interesting... Could you please add a comment about this?

(I'll try to test it at some point but lack the time atm)

src/webrtc/callFeed.ts
```diff
@@ -268,6 +336,11 @@ export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap>
         this.volumeLooperTimeout = setTimeout(this.volumeLooper, POLLING_INTERVAL);
     };
 
+    private isVADinCooldown(): boolean {
```
SimonBrandner, line 339:
```suggestion
    private isVADInCooldown(): boolean {
```

spec/unit/webrtc/callFeed.spec.ts
```diff
@@ -85,5 +85,129 @@ describe("CallFeed", () => {
...
+
+            it("voice activity should not disable audio track after a few milliseconds", async () => {
+                // Someone speaking
+                // Stops speaking for a few milliseconds
+                // Starts speaking again
+                // -> Is not muted in between (VAD_COOLDOWN)
+
+                feed.stream.addTrack(
+                    //@ts-ignore Mock
```
SimonBrandner, line 150: Could we avoid this by adding (or using, if it already exists) a `typed()` method on the mock class that would return the object with correct type cast?

spec/unit/webrtc/callFeed.spec.ts
```diff
@@ -85,5 +85,129 @@ describe("CallFeed", () => {
...
+
+                feed.setVoiceActivityThreshold(-50);
+                feed.speakingVolumeSamples = [-60];
+
+                setTimeout(() => {
+                    expect(feed.stream.getAudioTracks()[0].enabled).toBe(false);
+                }, 1000);
+            });
+
+            it("voice activity should not disable audio track after a few milliseconds", async () => {
```
SimonBrandner, line 143: To be completely honest, it's tough for me to follow what part of the cooldown tests does what (i.e. start/stop speaking) - could you please add comments and also avoid leaving things out that don't need to be in the test but would be better if they were there to be explicit (if I do understand everything correctly, you're not adding samples for all the start/stop speaking changes?)

src/webrtc/callFeed.ts
```diff
@@ -237,17 +286,36 @@ export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap>
...
-        this.emit(CallFeedEvent.VolumeChanged, maxVolume);
+        this.speakingVolumeSamples.push(this.maxVolume);
+
+        this.emit(CallFeedEvent.VolumeChanged, this.maxVolume);
+
+        // Handle voice activity detection only if it is enabled and user has not manually muted themselves
+        if (this.VADEnabled) {
+            // If the user is speaking
+            if (this.maxVolume > this.voiceActivityThreshold) {
+                this.VADCooldownStarted = new Date();
```
SimonBrandner, line 305: Would it make more sense to set the cooldown when the user stops speaking? It somehow feels better to only mute the user if we have seen they did stop speaking

src/webrtc/groupCall.ts
```diff
@@ -517,6 +532,28 @@ export class GroupCall extends TypedEventEmitter<
         return true;
     }
 
+    public async setVadMicrophoneMuted(muted: boolean): Promise<boolean> {
+        // hasAudioDevice can block indefinitely if the window has lost focus
+        if (!(await this.client.getMediaHandler().hasAudioDevice())) {
```
SimonBrandner, line 537: Even if not necessarily necessary (_words_), I think it would be good to have the `muted` check here, just to be sure

src/webrtc/callFeed.ts
DashieTM: It is not much different, since it will still be called every time the user is below threshold but I changed it.

spec/unit/webrtc/callFeed.spec.ts
DashieTM: I added some comments, they are very likely a bit too much but should at least give a better picture, I can then remove the superfluous ones later.

With leaving things out, is there something we can combine here? Like only create one stream?
Not sure what exactly you meant with that.

spec/unit/webrtc/callFeed.spec.ts
DashieTM: They would be the baseline, aka volume is higher than threshold therefore send, the new ones are specifically made for the cooldown, but they of course will also use vad.
We can also remove these if you want that, since technically they do the same, the ones below just add another step.

src/webrtc/callFeed.ts
DashieTM: After some testing, this seems to be another quirk that this delaynode has. If I create a const delay, then it doesn't work.
I need to create a delay inside the class that can be overwritten with a new stream instead of creating a new one. 
Otherwise the connection between the nodes get lost and the old behavior with voice cutting off at the start is back.

spec/unit/webrtc/callFeed.spec.ts
DashieTM: To be honest we simply took the existing mute implementation and copied the style from that to ensure we do it in the same manner.
The only thing we did different is that hugos formatting tool kinda broke the style again :D

I will talk to hugo though to see what we can do.

SimonBrandner (commented):
Just a few nitpicks, we're getting really close with this!

One more tiny nit: we tend to write comments with a space after the slashes (i.e. `// comment` not `//comment`)

spec/unit/webrtc/callFeed.spec.ts
```diff
@@ -85,5 +85,132 @@ describe("CallFeed", () => {
...
+                //@ts-ignore Mock
+                feed.stream.addTrack(new MockMediaStreamTrack("track", "audio", true));
+
+                //set threshold to infinity, this ensures we hit the threshold.
+                //then we check if the user is unmuted.
+                feed.setVoiceActivityThreshold(-Infinity);
+                setTimeout(() => {
+                    expect(feed.stream.getAudioTracks()[0].enabled).toBe(true);
+                }, 1000);
+            });
```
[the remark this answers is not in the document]
DashieTM: Yeah that was my oversight, the first 2 were unnecessary. I removed one of them and changed the other to be a simple check if the VAD does what it should when it's disabled, aka just not do anything.

DashieTM:
> Just a few nitpicks, we're getting really close with this!
> 
> One more tiny nit: we tend to write comments with a space after the slashes (i.e. `// comment` not `//comment`)

No worries, changed the ones in the tests.
Just so I don't make another new comment, I also put the audio delay on the same spot as the cooldown and treshold.
This number works ofc, just not the node itself :P

spec/unit/webrtc/callFeed.spec.ts
```diff
@@ -85,5 +85,125 @@ describe("CallFeed", () => {
                 expect(feed.isVideoMuted()).toBeTruthy();
             });
         });
+
+        describe("voice activity detection", () => {
+            it("If voice activity is disabled we should not mute/unmute with it", () => {
```
SimonBrandner, line 90:
```suggestion
            it("should not mute/unmute if voice activity is disabled", () => {
```

SimonBrandner (approved):
LGTM now! Will also request Dave's review, just to be sure

spec/unit/webrtc/callFeed.spec.ts
SimonBrandner: The words should make a sentence i.e.: _should not mute/unmute if voice activity is disabled_

spec/unit/webrtc/callFeed.spec.ts
hugohutri: Do you mean that we should remove `//@ts-ignore Mock`?
The implementation is the same as in the other tests, so maybe this could be something to be changed in another pull request?

DashieTM:
> LGTM now! Will also request Dave's review, just to be sure

Awesome!!
One thing though: 
we likely need some human feedback on whether or not the settings like the cooldown and the quality is ok. 
I will try to personally test this out with friends, but more people would always be nice.
Is the intention to mass test this alongside other features in a hosted version of element-call?

SimonBrandner:
> > LGTM now! Will also request Dave's review, just to be sure
> 
> Awesome!! One thing though: we likely need some human feedback on whether or not the settings like the cooldown and the quality are ok. I will try to personally test this out with friends, but more people would always be nice. (although I am behind a NAT with my low power server, so this will prob be a pain. Until now I have tested it locally.)
> 
> Is the intention to mass test this alongside other features in a hosted version of element-call?

Hmm, I think this would be best to discuss over at element-call with Jake since he's the one on the product team. Sorry, for late response...

DashieTM:
> Hmm, I think this would be best to discuss over at element-call with Jake since he's the one on the product team. Sorry, for late response...

No worries, also got lot's of stuff to do :P

dbkr (commented):
I was attempting to test this but I couldn't get it to work as it was NPEing on the local call feed:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'setVoiceActivityThreshold')
    at GroupCall.setVoiceActivityThreshold (groupCall.ts:556:35)
    at GroupCallView.tsx:92:15
    at commitHookEffectListMount (react-dom.development.js:23150:26)
    at commitPassiveMountOnFiber (react-dom.development.js:24926:13)
    at commitPassiveMountEffects_complete (react-dom.development.js:24891:9)
    at commitPassiveMountEffects_begin (react-dom.development.js:24878:7)
    at commitPassiveMountEffects (react-dom.development.js:24866:3)
    at flushPassiveEffectsImpl (react-dom.development.js:27039:3)
    at flushPassiveEffects (react-dom.development.js:26984:14)
    at react-dom.development.js:26769:9
```

Also, more points that are really for design, but I wasn't sure from the UI when I'd be muted/unmuted based on voice activity. It also seems a bit odd that it's in 'advanced' and not 'audio'.

DashieTM:
I will see what causes the error, that ofc shouldn't happen. 

As for why it is in an advanced tab. We were asked to do this: 
image

PR #2559, opened 2022-08-03
Read receipts for threads proof of concept (MSC3771) 

gsouquet:
Related to https://github.com/matrix-org/matrix-spec-proposals/pull/3771
Fixes https://github.com/vector-im/element-web/issues/22980

Update the `sendReceipt` function to include the `threadId` in the read receipt tuple. Also duplicates the read receipts room structure to all threads model

---

## ✨ Features
 * Read receipts for threads proof of concept (MSC3771)  ([\#2559](https://github.com/matrix-org/matrix-js-sdk/pull/2559)). Fixes vector-im/element-web#22980.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -4535,11 +4535,18 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             return Promise.resolve({}); // guests cannot send receipts so don't bother.
         }
 
-        const path = utils.encodeUri("/rooms/$roomId/receipt/$receiptType/$eventId", {
+        let path = utils.encodeUri("/rooms/$roomId/receipt/$receiptType/$eventId", {
             $roomId: event.getRoomId(),
             $receiptType: receiptType,
             $eventId: event.getId(),
         });
+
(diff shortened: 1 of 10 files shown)


src/client.ts
```diff
@@ -4535,11 +4535,18 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
             $roomId: event.getRoomId(),
             $receiptType: receiptType,
             $eventId: event.getId(),
         });
+
+        if (event.threadRootId) {
+            path += utils.encodeUri("/$threadId", {
+                $threadId: event.threadRootId,
+            });
+        }
```
clokep, line 4548: We should figure out if we want to keep sending this as part of the path (or the body or a GET parameter), but this matches the Synapse proof-of-concept for now. See https://github.com/matrix-org/matrix-spec-proposals/pull/3771#discussion_r918213697.

gsouquet: Changed to match what the MSC says

gsouquet:
There is now a production implementation of this proof of concept

Issue #2561, opened 2022-08-03
The queueToDevice.spec.ts test suite is flaky


robintown:
Me:

> We seem to have an issue in our js-sdk tests somewhere with the new to-device message queue:
> 
> ```
>   ●  Cannot log after tests are done. Did you forget to wait for something async in your test?
>     Attempted to log "All queued to-device messages sent".
> 
>       50 |             return console[methodName](...args);
>       51 |         } else {
>     > 52 |             return console.log(...args);
>          |                            ^
>       53 |         }
>       54 |         /* eslint-enable no-console */
>       55 |     };
> 
>       at console.log (node_modules/@jest/console/build/BufferedConsole.js:187:10)
>       at Logger.log (src/logger.ts:52:28)
>       at ToDeviceMessageQueue.debug [as sendQueue] (src/ToDeviceMessageQueue.ts:84:20)
> ```
> 
> This is coming out of the CI which suppresses most output so I can't see what test suite is causing it, and I can't reproduce it locally either. So, I guess this is a note to shout if anyone runs into it locally.

Travis:

> ![image(1)](https://user-images.githubusercontent.com/48614497/182710467-c05d3036-a596-49d8-afb7-ea43c87c7015.png)
> I get different failures as a flake 😛
> Expected 1 got 0 on a flush for the first one. Expected 20 got 1 on the object size for the second.

dbkr:
I've increased the timeout on the flush in the above PR, but I can't repro the failure so I'm struggling to see how it would be possible for the code to only send one message in the request. It looks like mock-request does flush the requests in order so  I don't think it's that the requests are getting mixed up.

robintown:
I still got it to fail, even with the new timeout in place:

```
Summary of all failing tests
 FAIL  spec/unit/queueToDevice.spec.ts
  ● queueToDevice (IndexedDB store) › retries when a message is retried

    expect(received).toEqual(expected) // deep equality

    Expected: 1
    Received: 0

      301 |         });
      302 |
    > 303 |         expect(await httpBackend.flush(null, 1, 1)).toEqual(1);
          |                                                     ^
      304 |         await flushPromises();
      305 |
      306 |         const dummyEvent = new MatrixEvent({

      at Object.toEqual (spec/unit/queueToDevice.spec.ts:303:53)

  ● queueToDevice (IndexedDB store) › splits many messages into multiple HTTP requests

    expect(received).toEqual(expected) // deep equality

    Expected: 20
[...]

PR #2564, opened 2022-08-04
Fix output after test finished

dbkr:
As per comment

https://github.com/matrix-org/matrix-js-sdk/issues/2561 (fixes the logging after tests but not Travis's flakes)

## Checklist

---


robintown:
Are there other tests that need that, or is that first test the only one?

dbkr:
There does appear to be another one - I'll add it to this PR since the tests have duly failed

dbkr:
Done - still need to look into the flakes from Travis, but this should fix the logging after tests finished.

PR #2567, opened 2022-08-05
Add txn_id support to sliding sync

kegsay:
This allows clients to know when a request has been applied
on the server. This allows us to change `resend(): void` to
`resend(): Promise<string>` which resolves/rejects with the
transaction ID when it has been applied/not.

## Checklist

---

## ✨ Features
 * Add txn_id support to sliding sync ([\#2567](https://github.com/matrix-org/matrix-js-sdk/pull/2567)).

### DIFF
--- a/spec/integ/sliding-sync.spec.ts
+++ b/spec/integ/sliding-sync.spec.ts
@@ -584,7 +584,7 @@ describe("SlidingSync", () => {
             let txnId;
             httpBackend.when("POST", syncUrl).check(function(req) {
                 const body = req.data;
-                logger.log("txn got ", body);
+                logger.debug("got ", body);
                 expect(body.room_subscriptions).toBeTruthy();
                 expect(body.room_subscriptions[roomId]).toEqual(roomSubInfo);
                 expect(body.txn_id).toBeTruthy();
@@ -616,7 +616,7 @@ describe("SlidingSync", () => {
             let txnId;
             httpBackend.when("POST", syncUrl).check(function(req) {
                 const body = req.data;
-                logger.log("txn got ", body);
+                logger.debug("got ", body);
                 expect(body.room_subscriptions).toBeFalsy();
                 expect(body.lists[0]).toEqual(newList);
                 expect(body.txn_id).toBeTruthy();
@@ -638,7 +638,7 @@ describe("SlidingSync", () => {
             let txnId;
             httpBackend.when("POST", syncUrl).check(function(req) {
                 const body = req.data;
-                logger.log("txn got ", body);
+                logger.debug("got ", body);
                 expect(body.room_subscriptions).toBeFalsy();
                 expect(body.lists[0]).toEqual({
                     ranges: [[20, 40]],
@@ -664,7 +664,7 @@ describe("SlidingSync", () => {
             let txnId;
             httpBackend.when("POST", syncUrl).check(function(req) {
                 const body = req.data;
-                logger.log("txn got ", body);
+                logger.debug("got ", body);
                 expect(body.room_subscriptions).toBeTruthy();
                 expect(body.room_subscriptions[roomId]).toEqual({
                     timeline_limit: 99,
@@ -759,7 +759,7 @@ describe("SlidingSync", () => {
             let txnId;
             httpBackend.when("POST", syncUrl).check(function(req) {
                 const body = req.data;
-                logger.log("txn got ", body);
+                logger.debug("got ", body);
                 expect(body.room_subscriptions).toBeFalsy();
                 expect(body.lists[0]).toEqual({
                     ranges: [[20, 40]],
--- a/spec/integ/sliding-sync.spec.ts
+++ b/spec/integ/sliding-sync.spec.ts
@@ -562,6 +562,225 @@ describe("SlidingSync", () => {
         });
     });
 
+    describe("transaction IDs", () => {
+        beforeAll(setupClient);
+        afterAll(teardownClient);
+        const roomId = "!foo:bar";
+
+        let slidingSync: SlidingSync;
+
+        // really this applies to them all but it's easier to just test one
+        it("should resolve modifyRoomSubscriptions after SlidingSync.start() is called", async () => {
+            const roomSubInfo = {
+                timeline_limit: 1,
+                required_state: [
+                    ["m.room.name", ""],
+                ],
+            };
+            // add the subscription
+            slidingSync = new SlidingSync(proxyBaseUrl, [], roomSubInfo, client, 1);
+            // modification before SlidingSync.start()
+            const subscribePromise = slidingSync.modifyRoomSubscriptions(new Set([roomId]));
+            let txnId;
+            httpBackend.when("POST", syncUrl).check(function(req) {
+                const body = req.data;
+                logger.log("txn got ", body);
+                expect(body.room_subscriptions).toBeTruthy();
+                expect(body.room_subscriptions[roomId]).toEqual(roomSubInfo);
+                expect(body.txn_id).toBeTruthy();
+                txnId = body.txn_id;
+            }).respond(200, function() {
+                return {
+                    pos: "aaa",
+                    txn_id: txnId,
+                    lists: [],
+                    extensions: {},
+                    rooms: {
+                        [roomId]: {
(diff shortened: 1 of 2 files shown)


src/sliding-sync.ts
```diff
@@ -615,11 +622,52 @@ export class SlidingSync extends TypedEventEmitter<SlidingSyncEvent, SlidingSync
...
     /**
-     * Resend a Sliding Sync request. Used when something has changed in the request.
+     * Resend a Sliding Sync request. Used when something has changed in the request. Resolves with
+     * the transaction ID of this request on success. Rejects with the transaction ID of this request
+     * on failure.
      */
-    public resend(): void {
+    public resend(): Promise<string> {
         this.needsResend = true;
+        this.txnId = ""+Math.random();
```
dbkr, line 631: Client (/baseApis) has `makeTxnId()` so may as well use that here.

src/sliding-sync.ts
```diff
@@ -615,11 +622,52 @@ export class SlidingSync extends TypedEventEmitter<SlidingSyncEvent, SlidingSync
...
+        const p: Promise<string> = new Promise((resolve, reject) => {
```
dbkr, line 632: You can also just put the template param on the Promise constructor, like so...

```suggestion
        const p = new Promise<string>((resolve, reject) => {
```

src/sliding-sync.ts
```diff
@@ -334,6 +336,11 @@ export class SlidingSync extends TypedEventEmitter<SlidingSyncEvent, SlidingSync
     private terminated = false;
     // flag set when resend() is called because we cannot rely on detecting AbortError in JS SDK :(
     private needsResend = false;
+    // the txn_id to send with the next request.
+    private txnId?: string = null;
+    // a list (in chronological order of when they were sent) of objects containing the txn ID and
+    // a defer to resolve/reject depending on whether they were successfully sent or not.
+    private txnIdDefers: {txnId: string, resolve: Function, reject: Function}[] = [];
```
dbkr, line 343: This might be worth promoting to it s own named type, especially if you type the functions (ie. `(s: string) => void` presumably).

src/sliding-sync.ts
```diff
@@ -615,11 +622,52 @@ export class SlidingSync extends TypedEventEmitter<SlidingSyncEvent, SlidingSync
...
+                resolve: resolve,
+                reject: reject,
+            });
+        });
         this.pendingReq?.abort();
+        return p;
+    }
+
+    private resolveTransactionDefers(txnId?: string) {
+        if (!txnId) {
```
dbkr, line 644: Is this a thing we expect to happen or just paranoia? If the latter, might be worth a log line, otherwise this will just change it from something accessing undefined to a completely silent failure.

src/sliding-sync.ts
```diff
@@ -615,11 +622,52 @@ export class SlidingSync extends TypedEventEmitter<SlidingSyncEvent, SlidingSync
...
+        if (txnIndex === -1) {
+            // this shouldn't happen; we shouldn't be seeing txn_ids for things we don't know about,
+            // whine about it.
+            logger.warn(`resolveTransactionDefers: seen ${txnId} but it isn't a pending txn, ignoring.`);
+            return;
+        }
+        // This list is sorted in time, so if the input txnId ACKs in the middle of this array,
+        // then everything before it that hasn't been ACKed yet never will and we should reject them.
+        for (let i = 0; i < txnIndex; i++) {
+            if (i < txnIndex) {
```
dbkr, line 664: Why do we need to check this again?

spec/integ/sliding-sync.spec.ts
```diff
@@ -562,6 +562,225 @@ describe("SlidingSync", () => {
...
+                ],
+            };
+            // add the subscription
+            slidingSync = new SlidingSync(proxyBaseUrl, [], roomSubInfo, client, 1);
+            // modification before SlidingSync.start()
+            const subscribePromise = slidingSync.modifyRoomSubscriptions(new Set([roomId]));
+            let txnId;
+            httpBackend.when("POST", syncUrl).check(function(req) {
+                const body = req.data;
+                logger.log("txn got ", body);
```
dbkr, line 587: was this left in deliberately?

src/sliding-sync.ts
t3chguy: also if we're calling it "defers" can we use our "defer" (IDeferred) utility?

spec/integ/sliding-sync.spec.ts
kegsay: Yes, it's useful to see this should it fail so you know whether the request was at fault or the response.

src/sliding-sync.ts
```diff
@@ -615,11 +622,52 @@ export class SlidingSync extends TypedEventEmitter<SlidingSyncEvent, SlidingSync
...
 
     /**
-     * Resend a Sliding Sync request. Used when something has changed in the request.
+     * Resend a Sliding Sync request. Used when something has changed in the request. Resolves with
+     * the transaction ID of this request on success. Rejects with the transaction ID of this request
+     * on failure.
      */
-    public resend(): void {
+    public resend(): Promise<string> {
         this.needsResend = true;
```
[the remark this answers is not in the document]
kegsay: Yes, which will cause the 1st request to not be sent. This is "fine" in that this class remembers what the first request contained so it will be send with the 2nd request. It's not fine as well because then the promise will not know when to resolve - it will end up rejecting after the 2nd request is resolved, which in this case is fine as that 2nd request will contain the 1st request's data, so it'll reject at the "right" time for UI purposes.

src/sliding-sync.ts
kegsay: It's not paranoia, as it's valid for responses to not have `txn_id` if one was not sent. This is why I don't log/whine about it.

src/sliding-sync.ts
kegsay: If you send 5 requests immediately, then there will be 5 elements in `this.txnIdDefers`, sorted in time so `[0]` is the earliest. Some of these requests may have made it to the server, some may not have. If for example, the 3rd request made it, then we want to reject the defers for requests 1 and 2 as they were lost (this relies on a single in-flight request at a time), and resolve the defer for request 3. This conditional is doing that.

kegsay: In practice this can happen very easily when users move around the app - each time they change the UI (space, room, etc) we cancel the current long-poll, modify the req params and send a new request. Rapidly switching rooms will cause many consecutive request -> cancel operations, and we don't know if these cancelled requests made it to the server.

spec/integ/sliding-sync.spec.ts
dbkr: OK - just looked a bit brief. Also generally you'd choose a logger level (ie. log.info/warn/debug etc) - logger/log is mostly for console.log compatibility.

src/sliding-sync.ts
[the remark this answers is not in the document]
dbkr: OK - just seemed like a `if (needsResend) return <the previous promise>;` would make for more sensible behaviour (ie. it's already going to resend, so no need to schedule another one). If not, could you add a comment to this effect in the code, as this is sort of weird and surprising that the request would reject because you called resend again too quickly.

src/sliding-sync.ts
dbkr: I was thinking much simpler: the loop condition is `i < txnIndex` so the the if statement just does the same test as the loop condition has just done each iteration of the loop? The loop only runs for i = 0, 1 ... txnIndex anyway.

src/sliding-sync.ts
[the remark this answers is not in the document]
kegsay: This is a good point: if we know we haven't sent the request yet we can bundle updates more efficiently. Will fix.

src/sliding-sync.ts
kegsay: Derp, thanks!

spec/integ/sliding-sync.spec.ts
kegsay: Using debug level now.

dbkr (approved):
Thanks - lgtm!

PR #2569, opened 2022-08-07
Don't load the sync accumulator if there's already a sync persist in flight

ara4n:
This should hopefully reduce chances of
https://github.com/vector-im/element-web/issues/21541 a bit more,
as we were incorrectly loading the sync accumulator even
if a sync persist was already in flight, thus wasting RAM
and increasing the chance of the renderer process OOMing

---

## 🐛 Bug Fixes
 * Don't load the sync accumulator if there's already a sync persist in flight ([\#2569](https://github.com/matrix-org/matrix-js-sdk/pull/2569)).


SimonBrandner (approved):
Looks sane

PR #2572, opened 2022-08-09
eslint: Enforce no-mixed-operators for boolean operators

3nprob:
## Checklist

* [x] Tests written for new code (and old code if feasible)
* [x] Linter and other CI checks pass
* [x] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))

Type: Task

---
This change is marked as an *internal change* (Task), so will not be included in the changelog.

PR #2574, opened 2022-08-09
Improve test coverage and modernize style for interactive-auth

3nprob:
#2572 but without the actual rule change. Will make the rule pass here once merged and has some additional cleanup and modernization around the changed lines.

## Checklist

Type: task

---

### DIFF
--- a/spec/unit/autodiscovery.spec.ts
+++ b/spec/unit/autodiscovery.spec.ts
@@ -153,27 +153,26 @@ describe("AutoDiscovery", function() {
         ]);
     });
 
-    it("should return FAIL_PROMPT when .well-known returns not-JSON", function() {
+    it("should return FAIL_PROMPT when .well-known returns not-JSON", async () => {
         const httpBackend = getHttpBackend();
-        httpBackend.when("GET", "/.well-known/matrix/client").respond(200, "abc");
+        httpBackend.when("GET", "/.well-known/matrix/client").respond(200, "abc", true);
+        const expected = {
+            "m.homeserver": {
+                state: "FAIL_PROMPT",
+                error: AutoDiscovery.ERROR_INVALID,
+                base_url: null,
+            },
+            "m.identity_server": {
+                state: "PROMPT",
+                error: null,
+                base_url: null,
+            },
+        };
         return Promise.all([
             httpBackend.flushAllExpected(),
-            AutoDiscovery.findClientConfig("example.org").then((conf) => {
-                const expected = {
-                    "m.homeserver": {
-                        state: "FAIL_PROMPT",
-                        error: AutoDiscovery.ERROR_INVALID,
-                        base_url: null,
-                    },
-                    "m.identity_server": {
-                        state: "PROMPT",
-                        error: null,
-                        base_url: null,
-                    },
-                };
-
-                expect(conf).toEqual(expected);
-            }),
+            AutoDiscovery.findClientConfig("example.org").then(
+              expect(expected).toEqual
+            )
         ]);
     });
 
--- a/spec/unit/autodiscovery.spec.ts
+++ b/spec/unit/autodiscovery.spec.ts
@@ -171,8 +171,8 @@ describe("AutoDiscovery", function() {
         return Promise.all([
             httpBackend.flushAllExpected(),
             AutoDiscovery.findClientConfig("example.org").then(
-              expect(expected).toEqual
-            )
+                expect(expected).toEqual,
+            ),
         ]);
     });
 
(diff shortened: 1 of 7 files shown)


3nprob:
```
Signed-off-by: 3nprob <git@3n.anonaddy.com>
```

src/interactive-auth.ts
```diff
@@ -539,19 +539,17 @@ export class InteractiveAuth {
             return;
         }
 
-        if (this.data && this.data.errcode || this.data.error) {
```
3nprob, line 542: The existing version will actually throw if `data` is falsey here.

3nprob:
I may be misinterpreting what Sonarcloud is reporting here, but it seems the lines it is reporting in should already be covered by existing tests (e.g. https://github.com/matrix-org/matrix-js-sdk/blob/3e37c7426420786c53992d48b85fe3ed37fcf40f/spec/unit/autodiscovery.spec.ts#L161)

src/autodiscovery.ts
```diff
@@ -409,39 +411,41 @@ export class AutoDiscovery {
...
 
                     try {
-                        resolve({ raw: JSON.parse(body), action: AutoDiscoveryAction.SUCCESS });
-                    } catch (e) {
-                        let reason = AutoDiscovery.ERROR_INVALID;
-                        if (e.name === "SyntaxError") {
-                            reason = AutoDiscovery.ERROR_INVALID_JSON;
-                        }
-                        resolve({
+                        return resolve({
```
dbkr, line 437: What's the value in returning the result of the `resolve()` function?

dbkr (commented):
May also want to update the PR title as it's a bit more extensive than just boolean logic operators, ie. adding types (yay!)

src/autodiscovery.ts
3nprob: In isolation here not much but I find it a good general pattern to remove the cognitive need of considering accidental fallthroughs.

dbkr: Fair enough - it does seem quite nice as a compact shorthand I guess.

dbkr (approved):
lgtm then, you have changed code that isn't covered by tests, so normally we'd ask to get the test coverage up to the required level. <del>I guess we can skip that in a PR that's just tidying up code, although we're pretty keen to get test coverage, so that may not be the case in general.</del> Actually, the sonarcloud check is required, so we need tests on this before it can be merged I'm afaraid.

3nprob:
So it turned out a reason why I couldn't seem to increase test coverage initially was a couple of small bugs which made the test silently pass despite not covering the intended path. This is now fixed and a handful of unit tests have been added. SonarCloud is now content.

Is there anything particular in here which needs to be broken out into a separate PR or is it OK to have it bundled up here? It would be really nice if we can keep this at least mostly in one piece as most of the code is tests to satisfy sonarcloud detecting existing untested code as new untested lines for the style change

3nprob:
Renamed the PR - suggestions that can add more clarity welcome

spec/unit/autodiscovery.spec.ts
```diff
@@ -153,27 +153,26 @@ describe("AutoDiscovery", function() {
         ]);
     });
 
-    it("should return FAIL_PROMPT when .well-known returns not-JSON", function() {
+    it("should return FAIL_PROMPT when .well-known returns not-JSON", async () => {
         const httpBackend = getHttpBackend();
-        httpBackend.when("GET", "/.well-known/matrix/client").respond(200, "abc");
+        httpBackend.when("GET", "/.well-known/matrix/client").respond(200, "abc", true);
```
3nprob, line 158: without the last flag it actually parsed as valid json and ended up returning the same error for a different reason.

spec/unit/interactive-auth.spec.ts
```diff
@@ -174,9 +343,108 @@ describe("InteractiveAuth", function() {
             throw err;
         });
 
-        return ia.attemptAuth().catch(function(error) {
-            expect(error.message).toBe('No appropriate authentication flow found');
```
3nprob, line 178: This was not actually asserted since it was never asserted that the `catch` happens in the first place. In fact it did not but does now.

src/utils.ts
```diff
@@ -328,22 +328,28 @@ export function escapeRegExp(string: string): string {
 }
 
 export function globToRegexp(glob: string, extended?: any): string {
```
3nprob, line 330: Any particular reason why `extended` is not restricted to be a boolean?

dbkr: Not sure - I guess whoever converted it to typescript just didn't manage to work out the correct type.

dbkr (approved):
Thanks so much for writing / fixing the tests!

PR #2583, opened 2022-08-10
sliding sync bugfix: ensure history is treated as history and not live events

kegsay:
Notes: fixed a sliding sync issue where history could be interpreted as live events.

---

## 🐛 Bug Fixes
 * fixed a sliding sync issue where history could be interpreted as live events. ([\#2583](https://github.com/matrix-org/matrix-js-sdk/pull/2583)).

### DIFF
--- a/src/sliding-sync-sdk.ts
+++ b/src/sliding-sync-sdk.ts
@@ -415,9 +415,9 @@ export class SlidingSyncSdk {
             // we should not know about any of these timeline entries if this is a genuinely new room.
             // If we do, then we've effectively done scrollback (e.g requesting timeline_limit: 1 for
             // this room, then timeline_limit: 50).
-            const knownEvents: Record<string, boolean> = {};
+            const knownEvents = new Set<string>();
             room.getLiveTimeline().getEvents().forEach((e) => {
-                knownEvents[e.getId()] = true;
+                knownEvents.add(e.getId());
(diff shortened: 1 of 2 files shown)


src/sliding-sync-sdk.ts
```diff
@@ -406,9 +406,51 @@ export class SlidingSyncSdk {
...
+        // TODO: handle threaded / beacon events
+
+        if (roomData.initial) {
+            // we should not know about any of these timeline entries if this is a genuinely new room.
+            // If we do, then we've effectively done scrollback (e.g requesting timeline_limit: 1 for
+            // this room, then timeline_limit: 50).
+            const knownEvents: Record<string, boolean> = {};
+            room.getLiveTimeline().getEvents().forEach((e) => {
+                knownEvents[e.getId()] = true;
+            });
```
t3chguy, line 421: Should this not be a Set?

t3chguy (approved):
looks sane otherwise

src/sliding-sync-sdk.ts
kegsay: Ah yes, too much Go... (which has no `Set` type!!!!)

t3chguy: You have Generics now, can now finally make a generic Set! :D

PR #2586, opened 2022-08-11
utils: Fix bug in deepCompare which would incorrectly return objects with disjoint keys as equal

3nprob:
## Checklist

Type: Defect

---

## 🐛 Bug Fixes
 * utils: Fix bug in deepCompare which would incorrectly return objects with disjoint keys as equal ([\#2586](https://github.com/matrix-org/matrix-js-sdk/pull/2586)). Contributed by @3nprob.


spec/unit/utils.spec.ts
```diff
@@ -152,6 +152,9 @@ describe("utils", function() {
             assert.isTrue(utils.deepCompare({ a: 1, b: 2 }, { a: 1, b: 2 }));
             assert.isTrue(utils.deepCompare({ a: 1, b: 2 }, { b: 2, a: 1 }));
             assert.isFalse(utils.deepCompare({ a: 1, b: 2 }, { a: 1, b: 3 }));
+            assert.isFalse(utils.deepCompare({ a: 1, b: 2 }, { a: 1 }));
```
3nprob, line 155: Without the fix, this assert will fail.

3nprob:
```
Signed-off-by: 3nprob <git@3n.anonaddy.com>
```

3nprob:
This function is used in these places in this repo. While it's involved in cryptographic key management I don't immediately see how this can become an actual security issue.

https://github.com/matrix-org/matrix-js-sdk/blob/9eb72908a73fc17af3f5c78845affa3c6865cc66/src/crypto/store/memory-crypto-store.ts#L149

https://github.com/matrix-org/matrix-js-sdk/blob/9eb72908a73fc17af3f5c78845affa3c6865cc66/src/crypto/store/indexeddb-crypto-store-backend.ts#L171

https://github.com/matrix-org/matrix-js-sdk/blob/9eb72908a73fc17af3f5c78845affa3c6865cc66/src/client.ts#L6265

3nprob:
@turt2live @richvdh (Since I see you as the users of this function when looking at the blame)

turt2live:
hey @3nprob - thanks for the PR. Would it be possible to get a de-anonymized signoff on your PRs, or a form filled out with dco@matrix.org (email) please? Thanks

3nprob:
> hey @3nprob - thanks for the PR. Would it be possible to get a de-anonymized signoff on your PRs, or a form filled out with [dco@matrix.org](mailto:dco@matrix.org) (email) please? Thanks

I'd rather stay pseudonymous for the time being. Invited you for DM on Matrix just now.

turt2live:
If you've invited me to a DM, I did not get it 😅 - I'm `@travis:t2l.io` on Matrix

3nprob:
So we did have a brief chat just now.

Unfortunately I had missed this before starting work: https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md#sign-off

> We accept contributions under a legally identifiable name, such as your name on government documentation or common-law names (names claimed by legitimate usage or repute). Unfortunately, we cannot accept anonymous contributions at this time.

I submit and make my patches available under the Apache License and have agreed to the DCO, which I have signed off on. Element is free to use these patches under the license, all I ask is that the license terms including attribution are followed.

However, requiring contributors to doxx themselves (in a way legally recognized in the UK, I assume here) is not something I find comfortable with for a project such as this (which is kind of OT here; just explaining my rationale for any onlookers).

turt2live (commented):
not sure I see the actual code change which matches the title here - it looks like a general refactoring rather than a bug fix?

3nprob:
> not sure I see the actual code change which matches the title here - it looks like a general refactoring rather than a bug fix?

There existing code is asymmetric. Check out and run just the added regression cases and you should see them fail.

3nprob:
For anyone wondering what the status is re the sign-off conversation above: Got put in touch with someone else from the Matrix Foundation and they came up with something that was acceptable so this PR should be good to go in that regard.

3nprob:
Rebased on `develop`

t3chguy:
@3nprob my PR was related to the test failure, but not due to the code being buggy, but the test making the wrong assertions

src/utils.ts
```diff
@@ -238,33 +238,24 @@ export function deepCompare(x: any, y: any): boolean {
...
-        let p;
-        for (p in y) {
+        for (const p in y) {
             if (y.hasOwnProperty(p) !== x.hasOwnProperty(p)) {
                 return false;
             }
         }
 
         // finally, compare each of x's keys with y
-        for (p in y) { // eslint-disable-line guard-for-in
```
turt2live, line 254: right, it was pointed out to me that this is the bug I couldn't see: `y` is meant to be `x`.

turt2live (approved):
Thanks! Apologies it took so long for this one to go through: it managed to hit just about every internal process unexpectedly :)

3nprob:
> @3nprob my PR was related to the test failure, but not due to the code being buggy, but the test making the wrong assertions

Which assertions specifically?

3nprob:
> @3nprob my PR was related to the test failure, but not due to the code being buggy, but the test making the wrong assertions

Or more precisely: The bug existed before and after your PR but some assertions were missing. This PR added those assertions.

3nprob:
Just curious; Why did e808e0ff18922030c74d5ead33d9e7fda3f8f600 become part of this PR and not opened as its own? Looks like a completely separate bug.

t3chguy:
@3nprob it became part of this PR to get your tests passing, we squash merge by policy hence it becoming a single commit

PR #2597, opened 2022-08-17
Add tests for incoming calls in group calls

dbkr:
Inspiration wwlecome for the renamed describe group which we're
really abusing for a bunch of things that happen to have the same
dependencies.

Fixes https://github.com/vector-im/element-call/issues/532

## Checklist

---

### DIFF
--- a/spec/unit/webrtc/groupCall.spec.ts
+++ b/spec/unit/webrtc/groupCall.spec.ts
@@ -620,7 +620,7 @@ describe('Group Call', function() {
             expect(mockCall.answerWithCallFeeds).not.toHaveBeenCalled();
         });
 
-        it("answers calls for the right room & group call ID", async () => { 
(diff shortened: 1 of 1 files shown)


spec/unit/webrtc/groupCall.spec.ts
```diff
@@ -545,5 +568,87 @@ describe('Group Call', function() {
                 groupCall.terminate();
             });
         });
+
+        describe("Incoming Calls", function() {
+            it("ignores incoming calls for other rooms", async () => {
+                const groupCall = await createAndEnterGroupCall(mockClient, room);
```
robintown, line 574: Extract this into a `beforeEach` and `groupCall.leave()` into an `afterEach` since they're the same for every test in the block?

robintown (approved):
Good to go otherwise

PR #2603, opened 2022-08-18
Update lockfile

robintown:
---


SimonBrandner (approved):

t3chguy:
We should be using Renovate for such tasks, no?

robintown:
Potentially? I never considered that we could make it work for feature branches.

PR #2604, opened 2022-08-19
Add `disposed` to `CallFeed`

SimonBrandner:


src/webrtc/callFeed.ts
```diff
@@ -295,6 +298,12 @@ export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap>
             this.analyser = null;
             releaseContext();
         }
+        this.disposed = true;
+        this.emit(CallFeedEvent.Disposed, this.disposed);
+    }
+
+    public idDisposed(): boolean {
```
robintown, line 305: I think this would be a good use case for the public getter + private setter pattern, i.e.

```typescript
private _disposed = false;
public get disposed(): boolean {
    return this._disposed;
}
private set disposed(value: boolean) {
    this._disposed = value;
}

File: src/webrtc/callFeed.ts [hunk 2]
```diff
@@ -55,6 +56,7 @@ type EventHandlerMap = {
     [CallFeedEvent.LocalVolumeChanged]: (localVolume: number) => void;
     [CallFeedEvent.VolumeChanged]: (volume: number) => void;
     [CallFeedEvent.Speaking]: (speaking: boolean) => void;
+    [CallFeedEvent.Disposed]: (disposed: boolean) => void;
```
robintown, line 59: I don't entirely understand why the `disposed: boolean` argument exists

SimonBrandner: I was thinking of an SFU use case but I'd actually do the thing differently in the end anyway :sweat_smile:

src/webrtc/callFeed.ts
```diff
@@ -295,6 +298,16 @@ export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap>
...
         }
+        this._disposed = true;
+        this.emit(CallFeedEvent.Disposed);
+    }
+
+    public get disposed(): boolean {
+        return this._disposed;
+    }
+
+    public set disposed(value: boolean) {
```
dbkr, line 309: No need for this to be public though I think?

SimonBrandner: Argh, my mind a few seconds ago: "this should definitely be private" and then I just forgot

robintown (approved):

PR #2606, opened 2022-08-19
Add unit tests for hangup / reject

dbkr:
Fixes https://github.com/vector-im/element-call/issues/537

## Checklist

---

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -1569,12 +1569,19 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
 
         try {
             await this.peerConn.setLocalDescription(answer);
+
+            // make sure we're still going
+            if (this.callHasEnded()) return;
+
             this.setState(CallState.Connecting);
 
             // Allow a short time for initial candidates to be gathered
             await new Promise(resolve => {
                 setTimeout(resolve, 200);
             });
+
+            // make sure the call hasn't ended before we continue
+            if (this.callHasEnded()) return;
 
             this.sendAnswer();
         } catch (err) {
--- a/spec/test-utils/webrtc.ts
+++ b/spec/test-utils/webrtc.ts
@@ -109,6 +109,12 @@ export class MockRTCPeerConnection {
             sdp: DUMMY_SDP,
         });
     }
+    createAnswer() {
+        return Promise.resolve({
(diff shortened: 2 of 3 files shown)


SimonBrandner (approved):

SimonBrandner:
Except the tests actually don't seem to be passing :/

src/webrtc/call.ts
```diff
@@ -1569,13 +1569,20 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
 
         try {
             await this.peerConn.setLocalDescription(answer);
+
+            // make sure we're still going
+            if (this.callHasEnded()) return;
+
```
SimonBrandner, line 1575: I wonder if this should be in `setState()`

src/webrtc/call.ts
```diff
@@ -1569,13 +1569,20 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
...
             this.setState(CallState.Connecting);
 
             // Allow a short time for initial candidates to be gathered
             await new Promise(resolve => {
                 setTimeout(resolve, 200);
             });
 
+            // make sure the call hasn't ended before we continue
+            if (this.callHasEnded()) return;
+
```
SimonBrandner, line 1585: I wonder if this should be in `sendAnswer()`

src/webrtc/call.ts
SimonBrandner: I should go sleep, anyway...

SimonBrandner (approved):

src/webrtc/call.ts
dbkr: Yeah, I did wonder about this and it seems quite elegant to basically just disallow state transitions out of 'ended', at the price of having to wrap all our setState calls in a try/catch (or some way to make the caller stop what it's doing).

src/webrtc/call.ts
dbkr: Possibly - I went for here as it's straight after it comes back from await-ing.

src/webrtc/call.ts
SimonBrandner: Yeah, so what I meant by _I should go to sleep, anyway..._ is that there is no simple way to do this, so I think this is actually fine (same for the other comment)

PR #2607, opened 2022-08-19
Re-emit room state events on rooms

robintown:
This also fixes some potential memory leaks and abuse of `removeAllListeners` in sync.ts.

## Checklist

---

## ✨ Features
 * Re-emit room state events on rooms ([\#2607](https://github.com/matrix-org/matrix-js-sdk/pull/2607)).

### DIFF
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -199,47 +199,7 @@ export class SyncApi {
      * @return {Room}
      */
     public createRoom(roomId: string): Room {
-        const client = this.client;
-        const { timelineSupport } = client;
-
-        const room = new Room(roomId, client, client.getUserId(), {
-            lazyLoadMembers: this.opts.lazyLoadMembers,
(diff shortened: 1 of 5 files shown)


src/sync.ts
```diff
@@ -1792,3 +1720,49 @@ function createNewUser(client: MatrixClient, userId: string): User {
     return user;
 }
 
+// /!\ This function is not intended for public use! It's only exported from
+// here in order to share some common logic with sliding-sync-sdk.ts.
+export function createRoom(client: MatrixClient, roomId: string, opts: Partial<IStoredClientOpts>): Room {
```
SimonBrandner, line 1725: I wonder if we could name this a bit better to avoid someone using it by accident - `createRoomAndSetupClientReEmitter()`?

SimonBrandner (commented):
LGTM, otherwise

SimonBrandner (approved):

PR #2609, opened 2022-08-22
Add ability to override built in room name generator for an i18n'able one

t3chguy:
---

## ✨ Features
 * Add ability to override built in room name generator for an i18n'able one ([\#2609](https://github.com/matrix-org/matrix-js-sdk/pull/2609)).

### DIFF
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -2914,6 +2914,33 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap> 
         return this.getType() === RoomType.ElementVideo;
     }
 
+    private roomNameGenerator(state: RoomNameState): string {
+        if (this.client.roomNameGenerator) {
+            const name = this.client.roomNameGenerator(this.roomId, state);
+            if (name !== null) {
+                return name;
+            }
+        }
+
+        switch (state.type) {
+            case RoomNameType.Actual:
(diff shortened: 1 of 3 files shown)


src/models/room.ts
```diff
@@ -2928,14 +2955,20 @@ export class Room extends TypedEventEmitter<EmittedEvents, RoomEventHandlerMap>
             // check for an alias, if any. for now, assume first alias is the
             // official one.
             const mRoomName = this.currentState.getStateEvents(EventType.RoomName, "");
-            if (mRoomName && mRoomName.getContent() && mRoomName.getContent().name) {
-                return mRoomName.getContent().name;
+            if (mRoomName?.getContent().name) {
```
justjanne, line 2958: The previous code did the equivalent of 
```suggestion
            if (mRoomName?.getContent()?.name) {
```
are you sure that getContent() always returns an object? We don’t have strict null checking yet in js-sdk.

justjanne (approved):

weeman1337 (approved):

src/models/room.ts
t3chguy: image
image

It doesn't always return an object, `content` or `m.new_content` could be a boolean, because we don't do runtime type checking on incoming events, but it does the `|| {}` that this was effectively doing already

PR #2612, opened 2022-08-22
Add tags and not_tags to the list of valid sliding sync filters

kegsay:
Notes: Add `tags` and `not_tags` to the list of valid sliding sync filters.

---


turt2live (approved):
tests wouldn't make sense for this change, so are not needed.

weeman1337 (approved):

PR #2613, opened 2022-08-23
Fix room membership race with PREPARED event

jotto:
See the call site of the original triggering event of this function: https://github.com/matrix-org/matrix-js-sdk/blob/b265d795a427c6d30ccdf279a09f7836509df863/src/sliding-sync.ts#L789-L806

I think the bug is current code assumes downstream event listeners of `SlidingSyncEvent.RoomData` have synchronous execution so that by the time it emits `SlidingSyncState.Complete`, and eventually `SyncState.Prepared` the room state is correct. But since SlidingSyncSdk's `processRoomData` is async, and the membership field was being set after the async, it looks like `SlidingSyncState.Complete` was being fired before the membership field was set.

Signed-off-by: Jonathan Otto <jonathan.otto@gmail.com>

---

## 🐛 Bug Fixes
 * Fix room membership race with PREPARED event ([\#2613](https://github.com/matrix-org/matrix-js-sdk/pull/2613)). Contributed by @jotto.


kegsay (approved):
Thanks for this. I thought I did basically what sync.ts did here, but this makes sense to set it prior to invoking listeners. Needs linting fixes and a regression test though ideally.

t3chguy (approved):
Thanks for this

PR #2622, opened 2022-08-25
Fix ICE end-of-candidates messages

dbkr:
We were casting a POJO to an RTCIceCandidate for the dummy
end-of-candidates candidate, but https://github.com/matrix-org/matrix-js-sdk/pull/2473
started calling .toJSON() on these objects.

Store separately whether we've seen the end of candidates rather than
adding on a dummy candidate object.

A test for this will follow, but a) I want to get this fix out and
b) I'm currently rewriting the call test file to add typing.

Fixes https://github.com/vector-im/element-call/issues/553

## Checklist

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -343,7 +343,7 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
     // possible
     private candidateSendQueue: Array<RTCIceCandidate> = [];
     private candidateSendTries = 0;
-    private sentEndOfCandidates = false;
+    private candidatesEnded = false;
     private feeds: Array<CallFeed> = [];
     private usermediaSenders: Array<RTCRtpSender> = [];
     private screensharingSenders: Array<RTCRtpSender> = [];
@@ -1597,6 +1597,11 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
      */
     private gotLocalIceCandidate = (event: RTCPeerConnectionIceEvent): Promise<void> => {
         if (event.candidate) {
+            if (this.candidatesEnded) {
+                logger.warn("Got candidate after candidates have ended - ignoring!");
+                return;
+            }
+
             logger.debug(
                 "Call " + this.callId + " got local ICE " + event.candidate.sdpMid + " candidate: " +
                 event.candidate.candidate,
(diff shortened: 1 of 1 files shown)


src/webrtc/call.ts
```diff
@@ -2257,19 +2256,23 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
...
         // to send if we successfully send the answer.
         // Equally don't send if we haven't yet sent the answer because we can send the
         // first batch of candidates along with the answer
-        if (this.state === CallState.Ringing || !this.inviteOrAnswerSent) return;
+        //if (this.state === CallState.Ringing || !this.inviteOrAnswerSent) return;
 
         // MSC2746 recommends these values (can be quite long when calling because the
         // callee will need a while to answer the call)
-        const delay = this.direction === CallDirection.Inbound ? 500 : 2000;
+        const delay = 0;//this.direction === CallDirection.Inbound ? 500 : 2000;
```
SimonBrandner, line 2275: What's going on here?

dbkr: oops, left in while debugging - thanks for spotting

src/webrtc/call.ts
```diff
@@ -1606,29 +1611,18 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
 
             // As with the offer, note we need to make a copy of this object, not
             // pass the original: that broke in Chrome ~m43.
-            if (event.candidate.candidate !== '' || !this.sentEndOfCandidates) {
+            if (event.candidate.candidate !== '') {
                 this.queueCandidate(event.candidate);
-
-                if (event.candidate.candidate === '') this.sentEndOfCandidates = true;
+            } else {
+                this.queueCandidate(null);
```
SimonBrandner, line 1617: I'd switch this up for clarity

src/webrtc/call.ts
```diff
@@ -2247,7 +2241,12 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
         }
     }
 
-    private queueCandidate(content: RTCIceCandidate): void {
+    /**
+     * Queue a candidate to be sent
+     * @param content The candidate to queue up, or null if candidates have finished being generated
+     *                and end-of-candidates should be signalled
+     */
+    private queueCandidate(content: RTCIceCandidate | null): void {
```
SimonBrandner, line 2249: I wonder if we could avoid the step with `null` and set `candidatesEnded` right at the start?

dbkr: Not sure what you mean here - at the start of what?

SimonBrandner: I was thinking we would set `candidatesEnded = true` right after receiving the last candidate and avoid calling `queueCandidate()` with null though since that might require a bit more work (if even possible) and you're on holiday, we can go with this as is

SimonBrandner (approved):

src/webrtc/call.ts
dbkr: oh I see - we need to schedule it to be sent too though, so this seemed the easiest way of re-using that code.

PR #2624, opened 2022-08-27
Remove support for unstable private read receipts

SimonBrandner:
Requires https://github.com/matrix-org/matrix-react-sdk/pull/9225

<hr>

We want there to be a reasonable transition time to switch to stable private read receipts, so this is blocked until that time passes

---


weeman1337 (approved):
LGTM :+1:

PR #2626, opened 2022-08-30
Base support for MSC3847: Ignore invites with policy rooms

Yoric:
This is a second take on the matrix-js-sdk-level machinery needed to be able to ignore invites, this time based on MSC3847.

The intention is that decisions on whether/how to display ignored invites is left in the hands of matrix-react-sdk.

## Checklist

Type: enhancement

---

## ✨ Features
 * Base support for MSC3847: Ignore invites with policy rooms ([\#2626](https://github.com/matrix-org/matrix-js-sdk/pull/2626)). Contributed by @Yoric.

### DIFF
--- a/spec/unit/matrix-client.spec.ts
+++ b/spec/unit/matrix-client.spec.ts
@@ -39,8 +39,11 @@ import { M_BEACON_INFO } from "../../src/@types/beacon";
 import { ContentHelpers, EventTimeline, Room } from "../../src";
 import { supportsMatrixCall } from "../../src/webrtc/call";
 import { makeBeaconEvent } from "../test-utils/beacon";
-import { IGNORE_INVITES_ACCOUNT_EVENT_KEY, POLICIES_ACCOUNT_EVENT_TYPE, PolicyScope }
-    from "../../src/models/invites-ignorer";
+import {
+    IGNORE_INVITES_ACCOUNT_EVENT_KEY,
+    POLICIES_ACCOUNT_EVENT_TYPE,
+    PolicyScope,
+} from "../../src/models/invites-ignorer";
 
 jest.useFakeTimers();
 
@@ -1498,12 +1501,14 @@ describe("MatrixClient", function() {
                 }
             };
         });
+
         it("should initialize and return the same `target` consistently", async () => {
             const target1 = await client.ignoredInvites.getOrCreateTargetRoom();
             const target2 = await client.ignoredInvites.getOrCreateTargetRoom();
             expect(target1).toBeTruthy();
             expect(target1).toBe(target2);
         });
+
         it("should initialize and return the same `sources` consistently", async () => {
             const sources1 = await client.ignoredInvites.getOrCreateSourceRooms();
             const sources2 = await client.ignoredInvites.getOrCreateSourceRooms();
@@ -1511,6 +1516,7 @@ describe("MatrixClient", function() {
             expect(sources1).toHaveLength(1);
             expect(sources1).toEqual(sources2);
         });
+
         it("should initially not reject any invite", async () => {
             const rule = await client.ignoredInvites.getRuleForInvite({
                 sender: "@foobar:example.org",
@@ -1518,6 +1524,7 @@ describe("MatrixClient", function() {
             });
             expect(rule).toBeFalsy();
         });
+
         it("should reject invites once we have added a matching rule in the target room (scope: user)", async () => {
             await client.ignoredInvites.addRule(PolicyScope.User, "*:example.org", "just a test");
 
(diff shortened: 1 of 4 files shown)


src/models/invites-ignorer.ts
```diff
@@ -0,0 +1,245 @@
...
+                    }
+                    let regexp;
+                    try {
+                        regexp = globToRegexp(glob, false);
+                    } catch (ex) {
+                        // Assume invalid event.
+                        continue;
+                    }
+                    for (const entity of entities) {
+                        if (entity && entity.search(regexp) >= 0) {
```
duxovni, line 147:
```suggestion
                        if (entity && regexp.test(entity)) {
```

duxovni (approved):
Looks reasonable overall; should this baseline support also include a method for adding rooms to the sources list?

src/models/invites-ignorer.ts
Yoric: So that would be `(new Regexp(regexp)).test(entity)`, since `regexp` is a `string` at that stage. Do you think it's worth it?

Yoric:
> Looks reasonable overall; should this baseline support also include a method for adding rooms to the sources list?

I can do that. I don't see any scenario for using this method in the near future, except perhaps for testing, though.

Yoric:
I took the opportunity to add a few more tests.

src/models/invites-ignorer.ts
duxovni: Oh, good point; maybe the thing to do is compile the regexp outside the for loop at the same time we're doing globToRegexp?

duxovni: From what I'm seeing, [the `new Regexp()` conversion happens implicitly anyway when you call `string.search(otherString)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search#parameters), so if anything, that would shave off one regexp parse in the server-scope case? It's a tiny nitpick regardless.

Yoric: Ah, right, this would help a bit. Fixed!

Yoric:
Are we waiting for any further reviews or can this be merged?

src/models/invites-ignorer.ts
```diff
@@ -0,0 +1,301 @@
...
+
+import { MatrixClient } from "../client";
+import { EventTimeline, MatrixEvent, Preset } from "../matrix";
+import { globToRegexp } from "../utils";
+import { Room } from "./room";
+
+/// The event type storing the user's individual policies.
+///
+/// Exported for testing purposes.
+export const POLICIES_ACCOUNT_EVENT_TYPE = "org.matrix.msc3847.policies";
```
turt2live, line 25: these should be `UnstableValue`s

spec/unit/matrix-client.spec.ts
```diff
@@ -1412,4 +1414,292 @@ describe("MatrixClient", function() {
...
+                }
+            };
+        });
+        it("should initialize and return the same `target` consistently", async () => {
+            const target1 = await client.ignoredInvites.getOrCreateTargetRoom();
+            const target2 = await client.ignoredInvites.getOrCreateTargetRoom();
+            expect(target1).toBeTruthy();
+            expect(target1).toBe(target2);
+        });
+        it("should initialize and return the same `sources` consistently", async () => {
```
turt2live, line 1507: please put a blank line between each test case

spec/unit/matrix-client.spec.ts
```diff
@@ -36,9 +36,11 @@ import { ReceiptType } from "../../src/@types/read_receipts";
 import * as testUtils from "../test-utils/test-utils";
 import { makeBeaconInfoContent } from "../../src/content-helpers";
 import { M_BEACON_INFO } from "../../src/@types/beacon";
-import { ContentHelpers, Room } from "../../src";
+import { ContentHelpers, EventTimeline, Room } from "../../src";
 import { supportsMatrixCall } from "../../src/webrtc/call";
 import { makeBeaconEvent } from "../test-utils/beacon";
+import { IGNORE_INVITES_ACCOUNT_EVENT_KEY, POLICIES_ACCOUNT_EVENT_TYPE, PolicyScope }
+    from "../../src/models/invites-ignorer";
```
turt2live, line 43: not sure why the linter allowed this:
```suggestion
import { 
    IGNORE_INVITES_ACCOUNT_EVENT_KEY, 
    POLICIES_ACCOUNT_EVENT_TYPE, 
    PolicyScope,
} from "../../src/models/invites-ignorer";
```

turt2live (changes requested):

turt2live (approved):
Thanks!

PR #2641, opened 2022-09-06
Add client.waitUntilRoomReadyForGroupCalls()

dbkr:
See comment, although this still feels like a poor solution to the
problem. Might be better if the js-sdk processed everything internally
before emitting the 'Room' event (or indeed before joinRoom resolved)
so the app knows everything is ready when it gets that event.

Also removes the corresponding event in favour of this method.

## Checklist

---

## 🐛 Bug Fixes
 * Add client.waitUntilRoomReadyForGroupCalls() ([\#2641](https://github.com/matrix-org/matrix-js-sdk/pull/2641)).


dbkr:
Merging on out-of-band approval my @ara4n

PR #2653, opened 2022-09-08
Making IgnoredInvites mostly sync.

Yoric:
Experience and feedback from https://github.com/matrix-org/matrix-react-sdk/pull/9255 indicates that we cannot integrate an async `IgnoredInvites.getRuleForInvite`. This PR:

- rewrites `getRuleForInvite` to be sync;
- adds some locking within `IgnoredInvites` to avoid race conditions noticed by @t3chguy.

## Checklist

* [X] Tests written for new code (and old code if feasible)
* [X] Linter and other CI checks pass
* [X] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))

---

Type: task

### DIFF
--- a/src/models/invites-ignorer.ts
+++ b/src/models/invites-ignorer.ts
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import AwaitLock from "await-lock";
 import { UnstableValue } from "matrix-events-sdk";
 
 import { MatrixClient } from "../client";
@@ -73,6 +74,16 @@ export enum PolicyScope {
  * our data structures.
  */
 export class IgnoredInvites {
+    // A lock around method `getOrCreateTargetRoom`.
+    // Used to ensure that only one async task of this class
+    // is creating a new target room and modifying the
+    // `target` property of account key `IGNORE_INVITES_POLICIES`.
+    private _getOrCreateTargetRoomLock = new AwaitLock();
+
+    // A lock around method `withIgnoreInvitesPoliciesLock`.
+    // Used to ensure that only one async task of this class is
+    // modifying `IGNORE_INVITES_POLICIES` at any point in time.
+    private _withIgnoreInvitesPoliciesLock = new AwaitLock();
     constructor(
         private readonly client: MatrixClient,
     ) {
@@ -85,6 +96,12 @@ export class IgnoredInvites {
      * @param entity The entity covered by this rule. Globs are supported.
      * @param reason A human-readable reason for introducing this new rule.
      * @return The event id for the new rule.
+     *
+     * # Safety
+     *
+     * This method will rewrite the `Policies` object in the user's account data.
+     * This rewrite is inherently racy and could overwrite or be overwritten by
+     * other concurrent rewrites of the same object.
      */
     public async addRule(scope: PolicyScope, entity: string, reason: string): Promise<string> {
         const target = await this.getOrCreateTargetRoom();
@@ -121,11 +138,10 @@ export class IgnoredInvites {
      */
     public async addSource(roomId: string): Promise<boolean> {
         // We attempt to join the room *before* calling
-        // `await this.getOrCreateSourceRooms()` to decrease the duration
+        // `await this.getSourceRooms()` to decrease the duration
         // of the racy section.
         await this.client.joinRoom(roomId);
-        // Race starts.
-        const sources = (await this.getOrCreateSourceRooms())
+        const sources = this.getSourceRooms()
             .map(room => room.roomId);
         if (sources.includes(roomId)) {
             return false;
@@ -135,7 +151,6 @@ export class IgnoredInvites {
             ignoreInvitesPolicies.sources = sources;
         });
 
-        // Race ends.
         return true;
     }
 
@@ -146,10 +161,10 @@ export class IgnoredInvites {
      * @param roomId The room to which the user is invited.
      * @returns A rule matching the entity, if any was found, `null` otherwise.
      */
-    public async getRuleForInvite({ sender, roomId }: {
+    public getRuleForInvite({ sender, roomId }: {
         sender: string;
         roomId: string;
-    }): Promise<Readonly<MatrixEvent | null>> {
+    }): Readonly<MatrixEvent | null> {
         // In this implementation, we perform a very naive lookup:
         // - search in each policy room;
         // - turn each (potentially glob) rule entity into a regexp.
@@ -160,7 +175,7 @@ export class IgnoredInvites {
         // - match several entities per go;
         // - pre-compile each rule entity into a regexp;
         // - pre-compile entire rooms into a single regexp.
-        const policyRooms = await this.getOrCreateSourceRooms();
+        const policyRooms = this.getSourceRooms();
         const senderServer = sender.split(":")[1];
         const roomServer = roomId.split(":")[1];
         for (const room of policyRooms) {
@@ -229,21 +244,30 @@ export class IgnoredInvites {
             const room = this.client.getRoom(target);
             if (room) {
                 return room;
-            } else {
-                target = null;
             }
         }
-        // We need to create our own policy room for ignoring invites.
-        target = (await this.client.createRoom({
-            name: "Individual Policy Room",
-            preset: Preset.PrivateChat,
-        })).room_id;
-        await this.withIgnoreInvitesPolicies(ignoreInvitesPolicies => {
-            ignoreInvitesPolicies.target = target;
-        });
+        try {
+            // We need to create our own policy room for ignoring invites.
+            await this._getOrCreateTargetRoomLock.acquireAsync();
+            target = (await this.client.createRoom({
+                name: "Individual Policy Room",
+                preset: Preset.PrivateChat,
+            })).room_id;
+            await this.withIgnoreInvitesPolicies(ignoreInvitesPolicies => {
+                ignoreInvitesPolicies.target = target;
+                if (!("sources" in ignoreInvitesPolicies)) {
+                    // `[target]` is a reasonable default for `sources`.
+                    ignoreInvitesPolicies.sources = [target];
+                }
+            });
 
-        // Since we have just called `createRoom`, `getRoom` should not be `null`.
-        return this.client.getRoom(target)!;
+            // Since we have just called `createRoom`, `getRoom` should not be `null`.
+            // Note that this is unavoidably racy, e.g. another client could have left
+            // the room during the call to `this.withIgnoreInvitesPolicies`.
+            return this.client.getRoom(target)!;
+        } finally {
+            this._getOrCreateTargetRoomLock.release();
+        }
     }
 
     /**
@@ -260,40 +284,21 @@ export class IgnoredInvites {
      * This rewrite is inherently racy and could overwrite or be overwritten by
      * other concurrent rewrites of the same object.
      */
-    public async getOrCreateSourceRooms(): Promise<Room[]> {
+    public getSourceRooms(): Room[] {
         const ignoreInvitesPolicies = this.getIgnoreInvitesPolicies();
         let sources = ignoreInvitesPolicies.sources;
 
         // Validate `sources`. If it is invalid, trash out the current `sources`
         // and create a new list of sources from `target`.
-        let hasChanges = false;
         if (!Array.isArray(sources)) {
             // `sources` could not be an array.
-            hasChanges = true;
             sources = [];
         }
-        let sourceRooms: Room[] = sources
+        const sourceRooms: Room[] = sources
             // `sources` could contain non-string / invalid room ids
             .filter(roomId => typeof roomId === "string")
             .map(roomId => this.client.getRoom(roomId))
             .filter(room => !!room);
-        if (sourceRooms.length != sources.length) {
-            hasChanges = true;
-        }
-        if (sourceRooms.length == 0) {
-            // `sources` could be empty (possibly because we've removed
-            // invalid content)
-            const target = await this.getOrCreateTargetRoom();
-            hasChanges = true;
-            sourceRooms = [target];
-        }
-        if (hasChanges) {
-            // Reload `policies`/`ignoreInvitesPolicies` in case it has been changed
-            // during or by our call to `this.getTargetRoom()`.
-            await this.withIgnoreInvitesPolicies(ignoreInvitesPolicies => {
-                ignoreInvitesPolicies.sources = sources;
-            });
-        }
         return sourceRooms;
     }
 
@@ -315,10 +320,15 @@ export class IgnoredInvites {
      * Modify in place the `IGNORE_INVITES_POLICIES` object from account data.
      */
     private async withIgnoreInvitesPolicies(cb: (ignoreInvitesPolicies: {[key: string]: any}) => void) {
-        const { policies, ignoreInvitesPolicies } = this.getPoliciesAndIgnoreInvitesPolicies();
-        cb(ignoreInvitesPolicies);
-        policies[IGNORE_INVITES_ACCOUNT_EVENT_KEY.name] = ignoreInvitesPolicies;
-        await this.client.setAccountData(POLICIES_ACCOUNT_EVENT_TYPE.name, policies);
+        await this._withIgnoreInvitesPoliciesLock.acquireAsync();
+        try {
+            const { policies, ignoreInvitesPolicies } = this.getPoliciesAndIgnoreInvitesPolicies();
+            cb(ignoreInvitesPolicies);
+            policies[IGNORE_INVITES_ACCOUNT_EVENT_KEY.name] = ignoreInvitesPolicies;
+            await this.client.setAccountData(POLICIES_ACCOUNT_EVENT_TYPE.name, policies);
+        } finally {
+            this._withIgnoreInvitesPoliciesLock.release();
+        }
     }
 
     /**
--- a/src/models/invites-ignorer.ts
+++ b/src/models/invites-ignorer.ts
@@ -153,11 +153,25 @@ export class IgnoredInvites {
     /**
      * Find out whether an invite should be ignored.
      *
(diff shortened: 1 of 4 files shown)


src/models/invites-ignorer.ts
```diff
@@ -227,21 +242,30 @@ export class IgnoredInvites {
...
-        await this.withIgnoreInvitesPolicies(ignoreInvitesPolicies => {
-            ignoreInvitesPolicies.target = target;
-        });
+        try {
+            // We need to create our own policy room for ignoring invites.
+            await this._getOrCreateTargetRoomLock.acquireAsync();
+            target = (await this.client.createRoom({
+                name: "Individual Policy Room",
+                preset: Preset.PrivateChat,
+            })).room_id;
```
t3chguy, line 253: Does this need to lock? We could instead return the same promise, no?

Yoric: Which promise? That one?

```ts
this.client.createRoom({
                name: "Individual Policy Room",
                preset: Preset.PrivateChat,
            })
```
We could probably store it and use it as a makeshift lock, but it feels more fragile/harder to read. Since we don't expect that this block will be performance-critical (in most runs, it should be executed either never or once), I'd favour the most readable solution, whichever it is.

t3chguy: We have that pattern in other places - https://github.com/matrix-org/matrix-js-sdk/blob/8aee884d03d79934e8944b05c22643c1bb480a64/src/models/room.ts#L409-L427 

I wouldn't suggest just storing the promise of the createRoom call, but rather this whole method by wrapping it.

Yoric: Yeah, that would work but it feels a bit more brittle in presence of exception.

Your call.

Yoric: Done.
As always, documenting concurrency is an interesting pastime :)

src/models/invites-ignorer.ts
```diff
@@ -313,10 +349,15 @@ export class IgnoredInvites {
...
-        await this.client.setAccountData(POLICIES_ACCOUNT_EVENT_TYPE.name, policies);
+        await this._withIgnoreInvitesPoliciesLock.acquireAsync();
+        try {
+            const { policies, ignoreInvitesPolicies } = this.getPoliciesAndIgnoreInvitesPolicies();
+            cb(ignoreInvitesPolicies);
+            policies[IGNORE_INVITES_ACCOUNT_EVENT_KEY.name] = ignoreInvitesPolicies;
+            await this.client.setAccountData(POLICIES_ACCOUNT_EVENT_TYPE.name, policies);
+        } finally {
+            this._withIgnoreInvitesPoliciesLock.release();
+        }
```
t3chguy, line 360: Not sure a lock is better here, as it yield N `setAccountData` calls which don't de-dup lower down so the promise-d/deferred approach probably works better here also

Yoric: I'm not sure what you have in mind.

Here, if the code is called N times, it's because we have N different things to write, so writing N times kinda makes sense. I suppose that I could accumulate them into a single call to `setAccountData` by storing the changes somewhere, to decrease network I/O in the unlikely case `withIgnoreInvitesPolicies` is called concurrently?

This would definitely complicate the code, though, and I don't really see what it optimizes for.

Or am I misunderstanding what you have in mind?

t3chguy: This lock is just providing linearisation, which can be accomplished with a Promise. 

```
withIgnoreInvitesPoliciesPromise = Promise.resolve<void>();
...
await this.withIgnoreInvitesPoliciesPromise;
...
this.withIgnoreInvitesPoliciesPromise = this.client.setAccountData(...);
```

t3chguy: Also there are missing types in `getPoliciesAndIgnoreInvitesPolicies`

src/models/invites-ignorer.ts
```diff
@@ -71,6 +72,16 @@ export enum PolicyScope {
  * our data structures.
  */
 export class IgnoredInvites {
+    // A lock around method `getOrCreateTargetRoom`.
+    // Used to ensure that only one async task of this class
+    // is creating a new target room and modifying the
+    // `target` property of account key `IGNORE_INVITES_POLICIES`.
+    private _getOrCreateTargetRoomPromise: Promise<Room> | null = null;
```
t3chguy, line 79: This shouldn't be underscored as per the code style

src/models/invites-ignorer.ts
```diff
@@ -71,6 +72,16 @@ export enum PolicyScope {
...
+
+    // A lock around method `withIgnoreInvitesPoliciesLock`.
+    // Used to ensure that only one async task of this class is
+    // modifying `IGNORE_INVITES_POLICIES` at any point in time.
+    private _withIgnoreInvitesPoliciesLock = new AwaitLock();
```
t3chguy, line 84: This shouldn't be underscored as per the code style

t3chguy: There should also be a blank line between this and the constructor

src/models/invites-ignorer.ts
Yoric: > Also there are missing types in `getPoliciesAndIgnoreInvitesPolicies`

Not following. Do you wish me to give names to `{[key: string]: any}` types?

Yoric:
@t3chguy Is this any better? The `any` code makes it clearer what data has been validated and what data hasn't.

t3chguy:
@Yoric sorry going to have to bump this back to the queue due to other obligations

src/models/invites-ignorer.ts
```diff
@@ -83,6 +94,12 @@ export class IgnoredInvites {
      * @param entity The entity covered by this rule. Globs are supported.
      * @param reason A human-readable reason for introducing this new rule.
      * @return The event id for the new rule.
+     *
+     * # Safety
+     *
+     * This method will rewrite the `Policies` object in the user's account data.
+     * This rewrite is inherently racy and could overwrite or be overwritten by
+     * other concurrent rewrites of the same object.
```
turt2live, line 102: shouldn't this be part of the description? At the moment it's part of the `@return`

src/models/invites-ignorer.ts
```diff
@@ -227,21 +255,48 @@ export class IgnoredInvites {
...
+            // Merge the calls.
+            return this.getOrCreateTargetRoomPromise;
+        }
+        try {
+            // Nobody is calling the method at the moment.
+            // Register ourselves as the leader and start creating our policy room.
+            this.getOrCreateTargetRoomPromise = (async () => {
+                // We need to create our own policy room for ignoring invites.
+                target = (await this.client.createRoom({
+                    name: "Individual Policy Room",
```
turt2live, line 275: Should this be translated?

src/models/invites-ignorer.ts
```diff
@@ -227,21 +255,48 @@ export class IgnoredInvites {
...
+        // may be racing with us.
+        if (this.getOrCreateTargetRoomPromise) {
+            // Another caller is already calling this method.
+            // Merge the calls.
+            return this.getOrCreateTargetRoomPromise;
+        }
+        try {
+            // Nobody is calling the method at the moment.
+            // Register ourselves as the leader and start creating our policy room.
+            this.getOrCreateTargetRoomPromise = (async () => {
```
turt2live, line 272: typically our pattern is `= new Promise((resolve, reject) => { ... });` instead

src/models/invites-ignorer.ts
```diff
@@ -303,9 +339,9 @@ export class IgnoredInvites {
      *
      * The result is *not* validated but is guaranteed to be a non-null object.
      *
-     * @returns A non-null object.
+     * @returns A non-null object. NOT validated.
      */
-    private getIgnoreInvitesPolicies(): {[key: string]: any} {
+    private getIgnoreInvitesPolicies(): any {
```
turt2live, line 344: this needs a type - what is it?

src/models/invites-ignorer.ts
```diff
@@ -315,16 +351,18 @@ export class IgnoredInvites {
...
     }
 
     /**
      * As `getIgnoreInvitesPolicies` but also return the `POLICIES_ACCOUNT_EVENT_TYPE`
-     * object.
+     * object. Content is NOT validated.
      */
     private getPoliciesAndIgnoreInvitesPolicies():
-        {policies: {[key: string]: any}, ignoreInvitesPolicies: {[key: string]: any}} {
+        {policies: any, ignoreInvitesPolicies: any} {
```
turt2live, line 365: these also need types

yarn.lock
```diff
@@ -1410,7 +1410,6 @@
 
 "@matrix-org/olm@https://gitlab.matrix.org/api/v4/projects/27/packages/npm/@matrix-org/olm/-/@matrix-org/olm-3.2.12.tgz":
   version "3.2.12"
-  uid "0bce3c86f9d36a4984d3c3e07df1c3fb4c679bd9"
```
turt2live, line 1413: this seems unrelated?

src/models/invites-ignorer.ts
```diff
@@ -144,10 +159,10 @@ export class IgnoredInvites {
      * @param roomId The room to which the user is invited.
      * @returns A rule matching the entity, if any was found, `null` otherwise.
      */
-    public async getRuleForInvite({ sender, roomId }: {
+    public getRuleForInvite({ sender, roomId }: {
```
turt2live, line 162: this is a breaking change because the class is not marked as unstable

src/models/invites-ignorer.ts
```diff
@@ -71,6 +71,17 @@ export enum PolicyScope {
...
+    // A lock around method `getOrCreateTargetRoom`.
+    // Used to ensure that only one async task of this class
+    // is creating a new target room and modifying the
+    // `target` property of account key `IGNORE_INVITES_POLICIES`.
+    private getOrCreateTargetRoomPromise: Promise<Room> | null = null;
+
+    // A lock around method `withIgnoreInvitesPoliciesLock`.
+    // Used to ensure that only one async task of this class is
+    // modifying `IGNORE_INVITES_POLICIES` at any point in time.
+    private withIgnoreInvitesPoliciesPromise: Promise<{}> = Promise.resolve({});
```
turt2live, line 83: these look like functions if the reader isn't extremely careful: can we call them something like `targetRoomLock` instead?

src/models/invites-ignorer.ts
```diff
@@ -119,11 +136,10 @@ export class IgnoredInvites {
...
     public async addSource(roomId: string): Promise<boolean> {
         // We attempt to join the room *before* calling
-        // `await this.getOrCreateSourceRooms()` to decrease the duration
+        // `await this.getSourceRooms()` to decrease the duration
         // of the racy section.
         await this.client.joinRoom(roomId);
-        // Race starts.
-        const sources = (await this.getOrCreateSourceRooms())
+        const sources = this.getSourceRooms()
             .map(room => room.roomId);
```
turt2live, line 143: why is this on two lines? It's difficult to read this way :(

src/models/invites-ignorer.ts
```diff
@@ -119,11 +136,10 @@ export class IgnoredInvites {
      */
     public async addSource(roomId: string): Promise<boolean> {
         // We attempt to join the room *before* calling
-        // `await this.getOrCreateSourceRooms()` to decrease the duration
+        // `await this.getSourceRooms()` to decrease the duration
```
turt2live, line 139: we don't await it though?

src/models/invites-ignorer.ts
```diff
@@ -215,6 +230,19 @@ export class IgnoredInvites {
...
      */
     public async getOrCreateTargetRoom(): Promise<Room> {
+        // Synchronous code. Do NOT introduce any `await` before locking
+        // or there will be race conditions on both in-memory data and
+        // homeserver-stored data.
+        //
+        // Thanks to run-to-completion, the uninterruptible behavior of this
+        // method is the following:
+        // 1. Execute all the code until the first `await`, including
+        //    * all the code before testing whether `this._getOrCreateTargetRoomPromise`
```
turt2live, line 240: this is the wrong field name, likewise throughout the rest of the comment block

src/models/invites-ignorer.ts
```diff
@@ -215,6 +230,19 @@ export class IgnoredInvites {
      * other concurrent rewrites of the same object.
      */
     public async getOrCreateTargetRoom(): Promise<Room> {
+        // Synchronous code. Do NOT introduce any `await` before locking
+        // or there will be race conditions on both in-memory data and
+        // homeserver-stored data.
+        //
+        // Thanks to run-to-completion, the uninterruptible behavior of this
+        // method is the following:
```
turt2live, line 238: This comment (and the lines which follow) aren't really describing anything: it's just saying what we're about to do, not why it's important to have a lock. We know what it's about to do: there's code to tell us :)

turt2live: We should also describe why we don't need cross-client locking, considering this code is seemingly important.

It's further a bad smell that we need a lock at all - why would this code be called multiple times?

turt2live (changes requested):
sorry, I'm coming at this blind with no real idea what troubles you're facing on the react-sdk side: it may be worth a visit to Element Web Internal to talk about those challenges instead, as making things synchronous can be annoyingly more difficult than understanding the async code path.

This is also much harder to review because there's no MSC reference within the area of the diff - it's hard to see what the underlying data structures are without such a reference.

Yoric:
> sorry, I'm coming at this blind with no real idea what troubles you're facing on the react-sdk side: it may be worth a visit to Element Web Internal to talk about those challenges instead, as making things synchronous can be annoyingly more difficult than understanding the async code path.

Well, I've been asked by my react-sdk reviewer to make things sync :)

> This is also much harder to review because there's no MSC reference within the area of the diff - it's hard to see what the underlying data structures are without such a reference.

Should be https://github.com/matrix-org/matrix-spec-proposals/pull/3847, iirc.

src/models/invites-ignorer.ts
Yoric: Oh, good point. Is there any way to do that in matrix-js-sdk? Or should I somehow pass that string from react-sdk?

src/models/invites-ignorer.ts
Yoric: > this needs a type - what is it?

It's data coming from Matrix, therefore untrusted. Until I have checked it (which is what the caller functions do), I have no clue. The `any` is there specifically to mark that it's untrusted. Is there a better way to write it down?

src/models/invites-ignorer.ts
Yoric: Well, this is a breaking change on my code that afaict isn't used anywhere yet, because the react-sdk patch was rejected because of `async` :)

src/models/invites-ignorer.ts
Yoric: :shrug: Different codebases, different conventions. I'll adapt :)

src/models/invites-ignorer.ts
t3chguy: The issue is this wasn't marked experimental and there's 70+ dependants on matrix-js-sdk, any of which could use a release where the API was synchronous and then have major issues when updating within the same semver range, so it is a breaking change.

src/models/invites-ignorer.ts
Yoric: > This comment (and the lines which follow) aren't really describing anything: it's just saying what we're about to do, not why it's important to have a lock. We know what it's about to do: there's code to tell us :)

...because of race conditions? I'm reading, modifying, writing some user account data, that's inherently racy.

> We should also describe why we don't need cross-client locking, considering this code is seemingly important.

Wait, we have cross-client locking? Where do I find the documentation on these?

> It's further a bad smell that we need a lock at all - why would this code be called multiple times?

I come from a safety background. Any code is considered to be called multiple times unless proven otherwise :)

src/models/invites-ignorer.ts
Yoric: With two calls to `then` after the `Promise`? Or is it an `async` `(resolve, reject) => { ... }`?

I've tried rewriting this call to use the `Promise` constructor but it didn't really feel readable.

src/models/invites-ignorer.ts
Yoric: Same response, sadly. If there is a better way to mark untrusted-deserialized-json than `any`, I'll be happy to use it!

src/models/invites-ignorer.ts
Yoric: @t3chguy So how do I handle this change? Should I add a new method instead of replacing the old one?

t3chguy: Ideally yes, or mark it as X-Breaking-Change but then it might get held until we have a few to go together for a major version bump

turt2live (changes requested):
this appears to have failing checks, and it's still not really clear why this PR needs to exist. Usually we're heavily in support of async code, so I'm surprised we want to make it sync.

It's also hard to re-review this at the moment: all the conversations have been left unresolved, making it unclear which need further comment.

Yoric:
> this appears to have failing checks, and it's still not really clear why this PR needs to exist. Usually we're heavily in support of async code, so I'm surprised we want to make it sync.
> 
> It's also hard to re-review this at the moment: all the conversations have been left unresolved, making it unclear which need further comment.

To clarify: I re-requested review because at least 3 conversations are waiting for a response on your part. I realize that I should have cleared up the resolved conversations, my bad.

Also, I'm making the code sync because @t3chguy requested that I do so [here](https://github.com/matrix-org/matrix-react-sdk/pull/9255#discussion_r964931213).

florianduros:
Hi!

The PR is 2 years old and stuck. 
I'm closing it. Feel free to reopen it if you want to continue to work on it.

PR #2662, opened 2022-09-12
Fix race in creating calls

dbkr:
We ran an async function between checking for an existing call and adding the new one to the map, so it would have been possible to start creating another call while we were placing the first call. This changes the code to add the call to the map as soon as we've created it.

Also adds more logging.

## Checklist

---

## 🐛 Bug Fixes
 * Fix race in creating calls ([\#2662](https://github.com/matrix-org/matrix-js-sdk/pull/2662)).


dbkr:
This is causing test breakage

SimonBrandner (approved):

PR #2663, opened 2022-09-12
Fix race in creating calls

dbkr:
Second attempt, just put a mutex around the method that processes member state events. The previous change meant that uninitialised calls were sitting in the map which is not a thing this code is prepared to handle and was causing test failures.

Also more logging.

Supersedes https://github.com/matrix-org/matrix-js-sdk/pull/2662

## Checklist

---

## 🐛 Bug Fixes
 * Fix race in creating calls ([\#2663](https://github.com/matrix-org/matrix-js-sdk/pull/2663)).

### DIFF
--- a/src/webrtc/groupCall.ts
+++ b/src/webrtc/groupCall.ts
@@ -185,6 +185,9 @@ export class GroupCall extends TypedEventEmitter<
     private resendMemberStateTimer: ReturnType<typeof setInterval> | null = null;
     private initWithAudioMuted = false;
     private initWithVideoMuted = false;
+    // We use this as a set of mutexes, effectively, to make sure we only enter the function
+    // to process a member state event once at any one time for each member
+    private processingMemberStateEvent = new Map<string, boolean>();
 
     constructor(
         private client: MatrixClient,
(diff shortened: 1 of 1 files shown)


src/webrtc/groupCall.ts
```diff
@@ -185,6 +185,9 @@ export class GroupCall extends TypedEventEmitter<
     private resendMemberStateTimer: ReturnType<typeof setInterval> | null = null;
     private initWithAudioMuted = false;
     private initWithVideoMuted = false;
+    // We use this as a set of mutexes, effectively, to make sure we only enter the function
+    // to process a member state event once at any one time for each member
+    private processingMemberStateEvent = new Map<string, boolean>();
```
SimonBrandner, line 190: Why is this a map? Wouldn't a `Set` work?

src/webrtc/groupCall.ts
```diff
@@ -754,6 +759,25 @@ export class GroupCall extends TypedEventEmitter<
...
+            logger.debug(`Already processing meber state event for ${event.getStateKey()}`);
+            return;
+        }
+
+        this.processingMemberStateEvent.set(event.getStateKey(), true);
+        try {
+            return this.processMemberStateEventInternal(event);
+        } finally {
+            this.processingMemberStateEvent.delete(event.getStateKey());
+        }
```
SimonBrandner, line 777: What's the benefit of try/finally here?

SimonBrandner (commented):
Otherwise, LGTM

SimonBrandner:
Also, what if the event changes while we're processing it?

dbkr:
Hmm, yes, you're right. We need to handle the newer ones somehow.

dbkr:
Going back to https://github.com/matrix-org/matrix-js-sdk/pull/2662 again

PR #2670, opened 2022-09-15
Target widget actions at a specific room

robintown:
Otherwise, the RoomWidgetClient class can end up accidentally sending and receiving events from rooms it didn't intend to, if it's an always-on-screen widget.

## Checklist

---

## 🐛 Bug Fixes
 * Target widget actions at a specific room ([\#2670](https://github.com/matrix-org/matrix-js-sdk/pull/2670)).

### DIFF
--- a/src/embedded.ts
+++ b/src/embedded.ts
@@ -24,6 +24,7 @@ import {
     ISendToDeviceToWidgetActionRequest,
 } from "matrix-widget-api";
 
+import type { IEvent, IContent } from "./models/event";
 import { ISendEventResponse } from "./@types/requests";
 import { EventType } from "./@types/event";
 import { logger } from "./logger";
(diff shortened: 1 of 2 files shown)


src/embedded.ts
```diff
@@ -69,27 +70,30 @@ export class RoomWidgetClient extends MatrixClient {
         super(opts);
 
         // Request capabilities for the functionality this client needs to support
-        this.capabilities.sendState?.forEach(({ eventType, stateKey }) =>
-            this.widgetApi.requestCapabilityToSendState(eventType, stateKey),
+        if (capabilities.sendState?.length || capabilities.receiveState?.length) {
+            widgetApi.requestCapabilityForRoomTimeline(roomId);
```
SimonBrandner, line 74: Why is this necessary now and it wasn't before?

src/embedded.ts
```diff
@@ -215,11 +218,20 @@ export class RoomWidgetClient extends MatrixClient {
...
+        // send us events from other rooms if this widget is always on screen
+        if (ev.detail.data.room_id === this.roomId) {
+            const event = new MatrixEvent(ev.detail.data as Partial<IEvent>);
+            await this.syncApi.injectRoomEvents(this.room, [], [event]);
+            this.emit(ClientEvent.Event, event);
+            this.setSyncState(SyncState.Syncing);
+            logger.info(`Received event ${event.getId()} ${event.getType()} ${event.getStateKey()}`);
+        } else {
+            const { event_id: eventId, room_id: roomId } = ev.detail.data;
+            logger.info(`Received event ${eventId} for a different room ${roomId}; discarding`);
```
SimonBrandner, line 232: Nit: I think it might be nicer for this to be in the `if` followed by `return`

SimonBrandner (approved):

src/embedded.ts
robintown: Because the timeline capability is a requirement for accessing a specific room, even if that room is normally what you'd think of as "the current room". The widget API rejects any requests directed at a specific room if the widget doesn't have this capability.

src/embedded.ts
robintown: Returning early would mean that we skip the `ack` call though, which is still something we should do

PR #2679, opened 2022-09-18
Add a few group call event handler tests

SimonBrandner:
---

### DIFF
--- a/spec/test-utils/webrtc.ts
+++ b/spec/test-utils/webrtc.ts
@@ -436,7 +436,6 @@ export function installWebRTCMocks() {
 export function makeMockGroupCallStateEvent(roomId: string, groupCallId: string, content: IContent = {
     "m.type": GroupCallType.Video,
     "m.intent": GroupCallIntent.Prompt,
-    "m.terminated": false,
 }): MatrixEvent {
     return {
         getType: jest.fn().mockReturnValue(EventType.GroupCallPrefix),
--- a/spec/test-utils/webrtc.ts
+++ b/spec/test-utils/webrtc.ts
@@ -21,6 +21,7 @@ import {
     GroupCall,
     GroupCallIntent,
     GroupCallType,
+    IContent,
     ISendEventResponse,
     MatrixClient,
     MatrixEvent,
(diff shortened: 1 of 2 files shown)


spec/test-utils/webrtc.ts
```diff
@@ -421,15 +422,16 @@ export function installWebRTCMocks() {
     };
 }
 
-export function makeMockGroupCallStateEvent(roomId: string, groupCallId: string): MatrixEvent {
+export function makeMockGroupCallStateEvent(roomId: string, groupCallId: string, content: IContent = {
+    "m.type": GroupCallType.Video,
+    "m.intent": GroupCallIntent.Prompt,
+    "m.terminated": false,
```
robintown, line 428: According to MSC3401, the presence of the `m.terminated` field is supposed to be enough to mark the call as terminated, which I presume isn't what you were going for here. If the group call handler doesn't treat this as terminated, I think we should change it to do that

spec/unit/webrtc/groupCallEventHandler.spec.ts
```diff
@@ -137,6 +183,42 @@ describe('Group Call Event Handler', function() {
...
+                    "m.intent": GroupCallIntent.Prompt,
+                    "dataChannelsEnabled": true,
+                    dataChannelOptions,
+                },
+            ),
+            {
+                roomId: FAKE_ROOM_ID,
+            } as unknown as RoomState,
+            null,
+        );
```
robintown, line 212: Maybe this should be factored out into a method on MockCallMatrixClient, given that you're sending state events like this multiple times

spec/unit/webrtc/groupCallEventHandler.spec.ts
```diff
@@ -137,6 +183,42 @@ describe('Group Call Event Handler', function() {
...
+
+        // @ts-ignore Mock
```
robintown, line 214: It's not obvious what this is bypassing TS for

spec/unit/webrtc/groupCallEventHandler.spec.ts
```diff
@@ -159,4 +241,86 @@ describe('Group Call Event Handler', function() {
...
+            expect(mockClientEmit).not.toHaveBeenCalledWith(
+                GroupCallEventHandlerEvent.Incoming,
+                expect.objectContaining({
+                    groupCallId: FAKE_GROUP_CALL_ID,
+                }),
+            );
+        });
+
+        it("ignores calls without a room", async () => {
+            (mockClient as any).getRoom = jest.fn().mockReturnValue(undefined);
```
robintown, line 314: Is it possible to spy this instead?

spec/test-utils/webrtc.ts
SimonBrandner: Argh, good point! (typing...)

robintown (approved):

PR #2694, opened 2022-09-23
Jest silence console on successful test suites

t3chguy:
---


t3chguy:
Abandoning

PR #2709, opened 2022-09-28
Rename redecryption-related function arguments for clarity

duxovni:
## Checklist

---

### DIFF
--- a/src/crypto/algorithms/megolm.ts
+++ b/src/crypto/algorithms/megolm.ts
@@ -1795,12 +1795,16 @@ class MegolmDecryption extends DecryptionAlgorithm {
      * @private
      * @param {String} senderKey
      * @param {String} sessionId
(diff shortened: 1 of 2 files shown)


src/crypto/algorithms/megolm.ts
```diff
@@ -1795,12 +1795,16 @@ class MegolmDecryption extends DecryptionAlgorithm {
      * @private
      * @param {String} senderKey
      * @param {String} sessionId
-     * @param {Boolean} keyTrusted
+     * @param {Boolean} forceRedecryptIfUntrusted
```
t3chguy, line 1798: I'm not sure this is much better, filling the docstring would be more appreciated

duxovni: The rename was suggested by @BillCarsonFr on the private gitlab repo; adding the docstring is a good idea in any case :+1:

PR #2710, opened 2022-09-28
Fix missing return when receiving an invitation without shared history

duxovni:
## Checklist

---

## 🐛 Bug Fixes
 * Fix missing return when receiving an invitation without shared history ([\#2710](https://github.com/matrix-org/matrix-js-sdk/pull/2710)). Contributed by @duxovni.

### DIFF
--- a/spec/integ/matrix-client-syncing.spec.ts
+++ b/spec/integ/matrix-client-syncing.spec.ts
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import 'fake-indexeddb/auto';
+
 import { Optional } from "matrix-events-sdk/lib/types";
 import HttpBackend from "matrix-mock-request";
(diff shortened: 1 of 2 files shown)


spec/integ/matrix-client-syncing.spec.ts
```diff
@@ -194,6 +195,46 @@ describe("MatrixClient syncing", () => {
             expect(fires).toBe(3);
         });
 
+        it("should emit ClientEvent.Room when invited", async () => {
```
t3chguy, line 198: This test isn't exercising the code path, probably due to it not having IDB, maybe we need to import `import 'fake-indexeddb/auto';`

t3chguy: You can ignore the strict errors stuff, its a bit faulty at excluding things you haven't touched

PR #2713, opened 2022-09-28
Calculate IndexedDB versions automatically to reduce repeated information and possibilities for error

duxovni:
## Checklist

---


SimonBrandner (commented):
I am confused as to why the version checks are not necessary?

duxovni:
The version checks are all still there, we're just using an actual for loop now, instead of writing out a for loop by hand and manually counting the number of iterations.

SimonBrandner (approved):
Oh, turns out I am very blind - sorry!

PR #2715, opened 2022-09-30
Fix issue in sync when crypto is not supported by client

stas-demydiuk:
When the Matrix client is initialized without crypto support it now breaks in the sync as starting from matrix-js-sdk 19.7.0 the additional code was added that directly access the client's crypto feature. 

This PR is aimed to fix the issue by introducing the additional check that crypto is actually available before accessing it

Signed-off-by: Stanislav Demydiuk <s.demydiuk@gmail.com>

## Checklist

* [x] Sign-off given on the changes

---

## 🐛 Bug Fixes
 * Fix issue in sync when crypto is not supported by client ([\#2715](https://github.com/matrix-org/matrix-js-sdk/pull/2715)). Contributed by @stas-demydiuk.

### DIFF
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -1201,7 +1201,7 @@ export class SyncApi {
 
             const inviter = room.currentState.getStateEvents(EventType.RoomMember, client.getUserId())?.getSender();
 
-            if (client.crypto) {
+            if (client.isCryptoEnabled()) {
                 const parkedHistory = await client.crypto.cryptoStore.takeParkedSharedHistory(room.roomId);
                 for (const parked of parkedHistory) {
                     if (parked.senderId === inviter) {
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -1200,19 +1200,22 @@ export class SyncApi {
             await this.processRoomEvents(room, stateEvents);
 
             const inviter = room.currentState.getStateEvents(EventType.RoomMember, client.getUserId())?.getSender();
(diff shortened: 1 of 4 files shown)


src/sync.ts
```diff
@@ -1200,19 +1200,22 @@ export class SyncApi {
...
-                        parked.senderKey,
-                        parked.forwardingCurve25519KeyChain,
-                        parked.sessionId,
-                        parked.sessionKey,
-                        parked.keysClaimed,
-                        true,
-                        { sharedHistory: true, untrusted: true },
-                    );
+
+            if (client.crypto) {
```
gsouquet, line 1204:
```suggestion
            if (client.isCryptoEnabled()) {
```

gsouquet (changes requested):
Thank you for your pull request!

I believe it would be important to add a smoke test to ensure that the js-sdk can process a sync when the crypto layer is missing

If I look a little further at this, I can see that the type definition here fails to mention that the `crypto` property can be `undefined` (see https://github.com/matrix-org/matrix-js-sdk/blob/e1edd84700ef0b2b871185a5e93bfb2b3192fb79/src/client.ts#L390) I would suggest changing that too

t3chguy:
/me will handle this and write a test, thanks for review @gsouquet and thanks @stas-demydiuk for the report & fix

gsouquet (approved):
Looks good to me 👍  Good luck with the strict null checks

PR #2723, opened 2022-10-03
Introduce Membership TS type

stas-demydiuk:
According to the [Matrix Spec](https://spec.matrix.org/v1.4/client-server-api/#room-membership) the room membership is enum with finite numbers of possible states. 

This PR introduces a new typescript type `Membrship` to be able to strictly specify what the exact membership is instead of the generic `string` type. The scope includes:
- New Membership type
- Update type definitions in `Client`, `Room`, `RoomMember` and `EventContent` interfaces
- Update existing conditions to use the new type instead of hardcoded strings
- Update spec files with the new type

## Checklist

* [X] Tests written for new code (and old code if feasible) - NA

---

## ✨ Features
 * Introduce Membership TS type ([\#2723](https://github.com/matrix-org/matrix-js-sdk/pull/2723)). Contributed by @stas-demydiuk.

### DIFF
--- a/src/@types/partials.ts
+++ b/src/@types/partials.ts
@@ -86,3 +86,5 @@ export interface IUsageLimit {
     // eslint-disable-next-line camelcase
     admin_contact?: string;
 }
+
+export type Membership = "ban" | "invite" | "join" | "knock" | "leave" | string;
--- a/spec/test-utils/test-utils.ts
+++ b/spec/test-utils/test-utils.ts
@@ -6,7 +6,7 @@ import '../olm-loader';
 
 import { logger } from '../../src/logger';
 import { IContent, IEvent, IUnsigned, MatrixEvent, MatrixEventEvent } from "../../src/models/event";
-import { ClientEvent, EventType, IPusher, MatrixClient, MsgType } from "../../src";
+import { ClientEvent, EventType, IPusher, MatrixClient, Membership, MsgType } from "../../src";
 import { SyncState } from "../../src/sync";
 import { eventMapperFor } from "../../src/event-mapper";
 
@@ -180,7 +180,7 @@ export function mkPresence(opts: IPresenceOpts & { event?: boolean }): Partial<I
 
 interface IMembershipOpts {
     room?: string;
-    mship: string;
+    mship: Membership;
     sender?: string;
     user?: string;
     skey?: string;
@@ -226,7 +226,7 @@ export function mkMembership(opts: IMembershipOpts & { event?: boolean }): Parti
 }
 
 export function mkMembershipCustom<T>(
-    base: T & { membership: string, sender: string, content?: IContent },
+    base: T & { membership: Membership, sender: string, content?: IContent },
 ): T & { type: EventType, sender: string, state_key: string, content: IContent } & GeneratedMetadata {
     const content = base.content || {};
     return mkEventCustom({
--- a/src/client.ts
+++ b/src/client.ts
@@ -166,7 +166,7 @@ import {
     UNSTABLE_MSC3088_PURPOSE,
     UNSTABLE_MSC3089_TREE_SUBTYPE,
 } from "./@types/event";
-import { IdServerUnbindResult, IImageInfo, Preset, Visibility } from "./@types/partials";
+import { IdServerUnbindResult, IImageInfo, Membership, Preset, Visibility } from "./@types/partials";
 import { EventMapper, eventMapperFor, MapperOpts } from "./event-mapper";
 import { randomString } from "./randomstring";
 import { BackupManager, IKeyBackup, IKeyBackupCheck, IPreparedKeyBackupVersion, TrustInfo } from "./crypto/backup";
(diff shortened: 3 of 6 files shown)


stas-demydiuk:
I see several TS errors in the lint report but they weren't introduced by the PR. Wonder if I should fix them as well?

image

src/@types/partials.ts
```diff
@@ -91,3 +91,11 @@ export interface IUsageLimit {
     // eslint-disable-next-line camelcase
     admin_contact?: string;
 }
+
+export enum Membership {
```
turt2live, line 95: I'm not sure an enum is neccessarily best for this sort of data type: enums get annoyingly large when transpiled and we're using it as a simple mapping: a plain `type` might be more suitable here instead.

turt2live (commented):
The lint errors are from the strict checker, which I think will be important to fix here given they're all to do with the new `Membership` enum.

src/@types/partials.ts
stas-demydiuk: I'm not sure that it is going to be that large for just 5 items, but from my point of view enums are much easier and cleaner from the readability and code maintenance perspective

stas-demydiuk:
> The lint errors are from the strict checker, which I think will be important to fix here given they're all to do with the new Membership enum.

@turt2live I've updated the PR with types fixes for all types errors related to membership

src/@types/partials.ts
turt2live: Ther ewas a time when I'd agree, though enums appear to be the way we've gone regardless. At 5 items, it's about 25 lines of boilerplate - enums make a lot more sense when it's >10-15 items.

turt2live (commented):
this is looking good - would it be possible to get a similar change on the react-sdk side which uses this new enum too?

stas-demydiuk:
> this is looking good - would it be possible to get a similar change on the react-sdk side which uses this new enum too?

Ok, I'll take a look there too

stas-demydiuk:
@turt2live I wonder what's the plan for this PR, is it going to be merged? as because it touches lots of files it becomes outdated pretty quickly

turt2live:
Sorry, without a review request these things tend to fall off my radar. I'll try to take a look this week, and can take on updating the PR if needed.

I do kinda want input from the wider team though as to whether this a welcome change or if they'd prefer some other approach though, as this will eternally affect their workflow.

turt2live:
Right, after talking to the team on this it sounds like we'd prefer union types when they make sense, and this is a case where they make sense. We wouldn't be leveraging the benefit of an enum here (ordered keys, primarily), and IDEs should be smart enough these days to autocomplete it.

Would you be interested in making this a union type instead? The notable change would be that we'd likely have to add `| string` to the end of it, given arbitrary memberships are *theoretically* possible.

stas-demydiuk:
> we'd likely have to add `| string` to the end of it, given arbitrary memberships are _theoretically_ possible.
not sure I understand why do we need `| string` as well, and what's the benefit of a union type in that case?

t3chguy:
Because memberships get added over time, the client may only support A, B, C but the server sends D. Using a union type without the general case means clients might wrongly assume no other values are possible.

stas-demydiuk:
> Because memberships get added over time, the client may only support A, B, C but the server sends D. Using a union type without the general case means clients might wrongly assume no other values are possible.

But as I understand no other values are possible at the moment with the current matrix specification, am I missing something?

t3chguy:
Correct. But the servers and clients don't update at the same time. So a client may receive data it doesn't believe possible leading to bugs, e.g if...else the else clause would need to handle extra memberships it was never designed to

stas-demydiuk:
I'm not sure I agree with that logic as we anyway cannot (and probably shouldn't) guarantee that the current `matrx-js-sdk` version will be compatible with future protocol versions out of the box as there might be changes that require the logic change and not only type update. But if you insist on `| string` I'll add it as well

t3chguy:
I don't disagree but the spec has no way to filter rooms out if you don't support their version

stas-demydiuk:
@turt2live / @t3chguy do you have any updates regarding this PR's future? it's the fourth time I'm fixing merge conflicts in it :)

turt2live:
Each force push makes it harder to review - we don't mind merge commits as they'll be squashed out of existence during merge of the PR. 

In general, this is probably good to go but not at the top of our list. As suggested before, we're happy to take on merge conflicts for the change, particularly for something as invasive as this.

stas-demydiuk:
Got it, thank you for the feedback!

src/@types/partials.ts
t3chguy: const enums are better at this

t3chguy:
> According to the [Matrix Spec](https://spec.matrix.org/v1.4/client-server-api/#room-membership) the room membership is enum with finite numbers of possible states.

This isn't really true, Matrix is extensible, custom memberships and latter-added memberships (like `knock`) can get added and you wouldn't want your types to fall out of sync with reality as it leads to brittle code.

andybalaam:
Closed in favour of https://github.com/matrix-org/matrix-js-sdk/pull/4107

PR #2726, opened 2022-10-03
Add support for unread thread notifications

gsouquet:
Implements https://github.com/matrix-org/matrix-spec-proposals/pull/3773
Related to https://github.com/vector-im/element-web/issues/23192

Passes the quality gate but still has optional strict null checks failures

## Checklist

---

## ✨ Features
 * Add support for unread thread notifications ([\#2726](https://github.com/matrix-org/matrix-js-sdk/pull/2726)).

### DIFF
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -1268,7 +1268,7 @@ export class SyncApi {
                 }
             }
 
-            room.threadNotifications = {};
+            room.resetThreadUnreadNotificationCount();
             const unreadThreadNotifications = joinObj[UNREAD_THREAD_NOTIFICATIONS.name];
             if (unreadThreadNotifications) {
                 Object.entries(unreadThreadNotifications).forEach(([threadId, unreadNotification]) => {
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -707,7 +707,8 @@ export class SyncApi {
                 initialFilter.setDefinition(filter.getDefinition());
                 initialFilter.setTimelineLimit(this.opts.initialSyncLimit);
                 const supportsThreadNotifications =
-                    await this.client.doesServerSupportUnstableFeature("org.matrix.msc3773");
+                    await this.client.doesServerSupportUnstableFeature("org.matrix.msc3773")
+                 || await this.client.isVersionSupported("v1.4");
                 initialFilter.setUnreadThreadNotifications(supportsThreadNotifications);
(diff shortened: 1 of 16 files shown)


src/sync.ts
```diff
@@ -1264,6 +1268,28 @@ export class SyncApi {
                 }
             }
 
+            room.resetThreadUnreadNotificationCount();
+            const unreadThreadNotifications = joinObj[UNREAD_THREAD_NOTIFICATIONS.name];
```
weeman1337, line 1272: Shouldn't we try to retrieve the value by `altName` as well?

src/sync.ts
```diff
@@ -1264,6 +1268,28 @@ export class SyncApi {
...
+            if (unreadThreadNotifications) {
+                Object.entries(unreadThreadNotifications).forEach(([threadId, unreadNotification]) => {
+                    room.setThreadUnreadNotificationCount(
+                        threadId,
+                        NotificationCountType.Total,
+                        unreadNotification.notification_count,
+                    );
+
+                    const hasUnreadNotification =
+                        room.getThreadUnreadNotificationCount(threadId, NotificationCountType.Highlight) <= 0;
```
weeman1337, line 1282: Is `<= 0` correct here? I would expect to check for `> 0`.

weeman1337 (commented):
:+1: just some small questions

src/sync.ts
gsouquet: I was confused when I named that variable. I replicated the logic from above
https://github.com/matrix-org/matrix-js-sdk/blob/7d5adf48bcca8fcdcefea707528aa906c71929e2/src/sync.ts#L1263

weeman1337 (approved):

t3chguy:
This breaks initialSync filtering support

https://github.com/matrix-org/matrix-js-sdk/commit/21a6f61b7b7bccb899c2aa34e6c881b729524bcd#diff-1b0cb46cd1fcdaacf73c771bf7eed6e00a12ee030da1dbcca4ae12842e1ecd81R709-R712

image

This may need reverting if not remedied

t3chguy:
Inherently this is broken

1. it never sets the unstable-prefixed filter
2. it sets the filter with a value of `false` which is redundant
3. it only sets it for the initial sync and not any later syncs which means the counts won't ever update on incremental/gappy syncs.

PR #2730, opened 2022-10-04
Use the correct sender key when checking shared secret

uhoreg:
fixes https://github.com/vector-im/element-web/issues/23374

Bug was introduced in the recent security release.

The incorrect code is checking the `sender_key` in the encrypted content, which is normally not set.  It should have been checking the `sender_key` in the cleartext content, or `event.getSenderKey()`.

---

## 🐛 Bug Fixes
 * Use the correct sender key when checking shared secret ([\#2730](https://github.com/matrix-org/matrix-js-sdk/pull/2730)). Fixes vector-im/element-web#23374.

### DIFF
--- a/src/crypto/SecretStorage.ts
+++ b/src/crypto/SecretStorage.ts
@@ -549,7 +549,7 @@ export class SecretStorage {
 
         const senderKeyUser = this.baseApis.crypto.deviceList.getUserByIdentityKey(
             olmlib.OLM_ALGORITHM,
-            event.getSenderKey(),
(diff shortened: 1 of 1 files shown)


turt2live (approved):

turt2live:
does this need including in the RC that went out today?

uhoreg:
> does this need including in the RC that went out today?

If possible, this would probably be a good one to get in.

Johennes:
Can / should this receive test coverage?

PR #2734, opened 2022-10-05
Encryption should not hinder verification

BillCarsonFr:
Fixes [#13656](https://github.com/vector-im/element-web/issues/13656)

Verification events sent in e2ee rooms should bypass the encrypt to verified devices only option. If not it renders verification impossible.

As per spec:
> When using in-room messages and the room has encryption enabled, clients should ensure that encryption does not hinder the verification. For example, if the verification messages are encrypted, clients must ensure that all the recipient’s unverified devices receive the keys necessary to decrypt the messages, even if they would normally not be given the keys to decrypt messages in the room. Alternatively, verification messages may be sent unencrypted, though this is not encouraged.

## Checklist

---

## 🐛 Bug Fixes
 * Encryption should not hinder verification ([\#2734](https://github.com/matrix-org/matrix-js-sdk/pull/2734)).


duxovni:
(Mostly looking for a quick reality-check on https://github.com/matrix-org/matrix-js-sdk/pull/2734/commits/8d29d5fe60a9432a1ead68cf60fb0a9fb64bf140)

SimonBrandner (approved):
Looks like an E2EE thing

PR #2738, opened 2022-10-07
Update hidden characters regex

turt2live:
These characters are hard to type, but can result in empty display name problems - let's force the app to use the user ID in these cases.

---


SimonBrandner (approved):

turt2live:
sorry sonarcloud, I'm not writing tests for this - the tests would be pretty pointless ("does it remove X character? Sure does.")

PR #2739, opened 2022-10-07
Fix sync init when thread unread notif is not supported

gsouquet:
Fixes https://github.com/vector-im/element-web/issues/23435

Creates a solution to detect features available on a given homeserver.
Developers can now document what matrix version will support a feature as well as the unstable prefixes. This code path is then accessible synchronously

## Checklist

---

## 🐛 Bug Fixes
 * Fix sync init when thread unread notif is not supported ([\#2739](https://github.com/matrix-org/matrix-js-sdk/pull/2739)). Fixes vector-im/element-web#23435.

### DIFF
--- a/src/feature.ts
+++ b/src/feature.ts
@@ -1,0 +1,56 @@
+/*
+Copyright 2022 The Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { IServerVersions } from "./client";
+
+export enum Feature {
+    Thread = "Thread",
+    ThreadUnreadNotifications = "ThreadUnreadNotifications",
+}
+
+type FeatureSupportCondition = {
+    unstablePrefixes?: string[];
+    matrixVersion?: string;
+};
+
+const featureSupportResolver: Record<string, FeatureSupportCondition> = {
+    [Feature.Thread]: {
+        unstablePrefixes: ["org.matrix.msc3440"],
+        matrixVersion: "v1.3",
+    },
+    [Feature.ThreadUnreadNotifications]: {
+        unstablePrefixes: ["org.matrix.msc3771", "org.matrix.msc3773"],
+        matrixVersion: "v1.4",
+    },
+};
+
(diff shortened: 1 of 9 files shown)


src/sync.ts
```diff
@@ -562,7 +563,11 @@ export class SyncApi {
     };
 
     private buildDefaultFilter = () => {
-        return new Filter(this.client.credentials.userId);
+        const filter = new Filter(this.client.credentials.userId);
+        if (this.client.canSupport.get(Feature.ThreadUnreadNotifications)) {
+            filter.setUnreadThreadNotifications(true);
```
t3chguy, line 568: `setUnreadThreadNotifications` only uses the stable non-prefixed version but `this.client.canSupport.get(Feature.ThreadUnreadNotifications)` will return true if unstable is supported.

src/store/memory.ts
```diff
@@ -227,7 +227,7 @@ export class MemoryStore implements IStore {
      * @param {Filter} filter
      */
     public storeFilter(filter: Filter): void {
-        if (!filter) {
+        if (!filter || !filter.userId) {
```
t3chguy, line 230:
```suggestion
        if (!filter?.userId) {
```

t3chguy (changes requested):

src/sync.ts
gsouquet: Good catch! I've changed the code to use `UnstableValue`. 
I believe I had not done that in the first place after seeing how `setProp` works... it essentially treats `.` in property names as a nested level

src/filter.ts
```diff
@@ -227,7 +228,16 @@ export class Filter {
...
      */
     public setUnreadThreadNotifications(enabled: boolean): void {
-        setProp(this.definition, "room.timeline.unread_thread_notifications", !!enabled);
+        this.definition = {
+            ...this.definition,
+            room: {
+                ...this.definition?.room,
+                timeline: {
+                    ...this.definition?.room?.timeline,
+                    [UNREAD_THREAD_NOTIFICATIONS.name]: !!enabled,
```
t3chguy, line 237: This'll now always use the unstable. If your HS only has stable support and not unstable it'll cause /sync to explode.

src/sync.ts
t3chguy: Maybe we should update `setProp` to take a variadic list of keys rather than a string to avoid this problem

src/filter.ts
gsouquet: I might be misunderstanding the use of `UnstableValue` then 🤔

src/feature.ts
```diff
@@ -0,0 +1,62 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { IServerVersions } from "./client";
+
+export enum ServerSupport {
+    Stable,
+    Experimental,
```
t3chguy, line 21: Can we call it `Unstable` to match `unstablePrefixes` etc?

t3chguy (approved):
Looks sane to me! Any chance of an extra test around `/sync` actually including this filter?

PR #2745, opened 2022-10-10
Correct the dir parameter of MSC3715

dhenneke:
The implementation of [MSC3715](https://github.com/matrix-org/matrix-doc/pull/3715) was stabilized in [synapse](https://github.com/matrix-org/synapse/issues/13919) and we might want to use the new parameter. Note that the old name `direction` was always wrong and never worked so I don't think we need the fallback to the prefixed version. This change would also need a PR at the react-sdk but there only seems to be a [single usage](https://github.com/matrix-org/matrix-react-sdk/blob/cf029c51dc466db448c7324d3dda1fa484a9f3ee/src/stores/widgets/StopGapWidgetDriver.ts#L452). If you accept this change I can do a second PR to update it. I assume merging would need to be coordinated to not break the build(?).

An alternative could be to keep the name `direction` in the interface and convert it to `dir` when adding it to the query params. That wouldn't require a potential breaking change and a change in a downstream project. Let me know if you would prefer that solution.

I added some primitive tests that only focus on the HTTP calls and not on the actual behavior of the relations function.

## Checklist

---

## 🐛 Bug Fixes
 * Correct the dir parameter of MSC3715 ([\#2745](https://github.com/matrix-org/matrix-js-sdk/pull/2745)). Contributed by @dhenneke.

### DIFF
--- a/spec/integ/matrix-client-relations.spec.ts
+++ b/spec/integ/matrix-client-relations.spec.ts
@@ -1,0 +1,112 @@
+/*
+Copyright 2022 Dominik Henneke
+Copyright 2022 Nordeck IT + Consulting GmbH.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


SimonBrandner (approved):
LGTM - thank you!

SimonBrandner:
SonarCloud seems to require a bit more tests for the `MatrixClient` class - if you could please add those, it would be awesome!

dhenneke:
I added a test and hope SonarCloud is satisfied with that.

I also created the PR on the react-sdk: https://github.com/matrix-org/matrix-react-sdk/pull/9391

SimonBrandner (approved):

PR #2747, opened 2022-10-12
Support sign in + E2EE set up using QR code implementing MSC3886, MSC3903 and MSC3906

hughns:
Implements [MSC3886](https://github.com/matrix-org/matrix-spec-proposals/pull/3886), [MSC3903](https://github.com/matrix-org/matrix-spec-proposals/pull/3903) and [MSC3906](https://github.com/matrix-org/matrix-spec-proposals/pull/3906)

## Checklist

---

## ✨ Features
 * Support sign in + E2EE set up using QR code implementing MSC3886, MSC3903 and MSC3906 ([\#2747](https://github.com/matrix-org/matrix-js-sdk/pull/2747)). Contributed by @hughns.


hughns:
I will add some more tests, but if a review could start before that it would be much appreciated.

gsouquet (approved):
Looks pretty sensible to me!

I have not reviewed `channels/ecdhV1.ts` as I do not have the necessery context, but besides that I believe it looks good and I added some minor non-blocking comments.

Some places could benefit from JSDoc, but I believe that is intentional and you tried to keep that in the interfaces?

turt2live (changes requested):
This cannot land with stable prefixes

src/rendezvous/channels/ecdhV1.ts
```diff
@@ -0,0 +1,298 @@
...
+export interface ECDHv1RendezvousCode extends RendezvousCode {
+    rendezvous: {
+        transport: RendezvousTransportDetails;
+        algorithm: SecureRendezvousChannelAlgorithm.ECDH_V1;
+        key: string;
+    };
+}
+
+// The underlying algorithm is the same as:
+// https://github.com/matrix-org/matrix-js-sdk/blob/75204d5cd04d67be100fca399f83b1a66ffb8118/src/crypto/verification/SAS.ts#L54-L68
```
uhoreg, line 44: It seems like it should be calling this function, rather than copying the code, even if it is a (almost) one-liner

src/rendezvous/channels/ecdhV1.ts
```diff
@@ -0,0 +1,298 @@
...
+        (sasBytes[0] << 5 | sasBytes[1] >> 3) + 1000,
+        ((sasBytes[1] & 0x7) << 10 | sasBytes[2] << 2 | sasBytes[3] >> 6) + 1000,
+        ((sasBytes[3] & 0x3f) << 7 | sasBytes[4] >> 1) + 1000,
+    ];
+
+    return digits.join('-');
+}
+
+async function importKey(key: Uint8Array): Promise<CryptoKey | Uint8Array> {
+    if (getCrypto()) {
```
uhoreg, line 64: https://github.com/matrix-org/matrix-js-sdk/pull/2762 removes this function, so this may need to be updated (as well as the definition of `subtleCrypto`, and changing things in general to use only the WebCrypto API).  This will need some coordination between this PR and 2762.

hughns: (The irony of having created more work for myself by suggesting the migration to Web Crypto this morning)

src/rendezvous/rendezvous.ts
```diff
@@ -0,0 +1,222 @@
...
+        this.code = JSON.stringify(await this.channel.generateCode(this.ourIntent));
+    }
+
+    async startAfterShowingCode(): Promise<string | undefined> {
+        const checksum = await this.channel.connect();
+
+        logger.info(`Connected to secure channel with checksum: ${checksum} our intent is ${this.ourIntent}`);
+
+        // determine available protocols
+        if (!(await this.client.doesServerSupportUnstableFeature('org.matrix.msc3882'))) {
```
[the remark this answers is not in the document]
hughns: Version is unknown. I have refactored to use `feature.ts`.

src/rendezvous/cancellationReason.ts
```diff
@@ -0,0 +1,31 @@
+/*
```
turt2live, line 1: This file is improperly named, I think. Should be called `RendezvousFailureReason.ts`

src/rendezvous/channel.ts
```diff
@@ -0,0 +1,54 @@
+/*
```
turt2live, line 1: `RendezvousChannel.ts` please

src/rendezvous/channels/ecdhV1.ts
```diff
@@ -0,0 +1,298 @@
+/*
```
turt2live, line 1: `MSC3903ECDHv1RendezvousChannel.ts` (further reinforcing how unstable it is)

src/rendezvous/channels/ecdhV1.ts
```diff
@@ -0,0 +1,298 @@
...
+                key: encodeBase64(this.ourPublicKey),
+                transport: await this.transport.details(),
+            },
+            intent,
+        };
+
+        return rendezvous;
+    }
+
+    async connect(): Promise<string> {
```
turt2live, line 102:
```suggestion
    public async connect(): Promise<string> {
```

src/rendezvous/channel.ts
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * @returns the checksum/confirmation digits to be shown to the user
+     */
+    connect(): Promise<string>;
+
+    /**
+     * Send a payload via the channel.
+     * @param data payload to send
+     */
+    send(data: any): Promise<void>;
```
turt2live, line 35:
```suggestion
    send(data: object): Promise<void>;
```

or other more suitable type - `any` is not really acceptable as a type.

src/rendezvous/channel.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Receive a payload from the channel.
+     * @returns the received payload
+     */
+    receive(): Promise<any>;
```
turt2live, line 41: likewise on `any` - what is it?

src/rendezvous/channels/ecdhV1.ts
```diff
@@ -0,0 +1,298 @@
...
+
+        const aesKeyBytes = this.olmSAS.generate_bytes(aesInfo, 32);
+
+        this.aesKey = await importKey(aesKeyBytes);
+
+        const rawChecksum = this.olmSAS.generate_bytes(aesInfo, 5);
+        return generateDecimalSas(Array.from(rawChecksum));
+    }
+
+    private async encrypt(data: any): Promise<string> {
```
turt2live, line 149: type for `data`?

src/rendezvous/channels/ecdhV1.ts
```diff
@@ -0,0 +1,298 @@
...
+        const stringifiedData = JSON.stringify(data);
+
+        if (this.aesKey) {
+            await this.transport.send('application/json', await this.encrypt(stringifiedData));
+        } else {
+            await this.transport.send('application/json', stringifiedData);
+        }
+    }
+
+    private async decrypt({ iv, ciphertext }: { iv: string, ciphertext: string }): Promise<any> {
```
turt2live, line 206: Type for return value?

src/rendezvous/channels/ecdhV1.ts
hughns: Fixed. n.b. the function is now exported from `SAS.ts`.

src/rendezvous/code.ts
```diff
@@ -0,0 +1,31 @@
+/*
```
turt2live, line 1: This should be bundled into a `types.ts` file, not `code.ts`

src/rendezvous/error.ts
```diff
@@ -0,0 +1,23 @@
+/*
```
turt2live, line 1: `RendezvousError.ts` or bundled into a `types.ts` file

src/rendezvous/index.ts
```diff
@@ -0,0 +1,22 @@
+/*
```
turt2live, line 1: Generally shouldn't need index files, fwiw. Not a blocker for this PR.

src/rendezvous/rendezvous.ts
```diff
@@ -0,0 +1,228 @@
+/*
```
turt2live, line 1: `MSC3906Rendezvous.ts`

src/rendezvous/rendezvous.ts
```diff
@@ -0,0 +1,228 @@
...
+ * This implementation only supports generating a QR code on a device that is already signed in.
+ * Note that this is UNSTABLE and may have breaking changes without notice.
+ */
+export class MSC3906Rendezvous {
+    private newDeviceId?: string;
+    private newDeviceKey?: string;
+    private ourIntent: RendezvousIntent = RendezvousIntent.RECIPROCATE_LOGIN_ON_EXISTING_DEVICE;
+    public code?: string;
+
+    constructor(
```
turt2live, line 48:
```suggestion
    public constructor(
```

src/rendezvous/rendezvous.ts
```diff
@@ -0,0 +1,228 @@
...
+/**
+ * Implements MSC3906 to allow a user to sign in on a new device using QR code.
+ * This implementation only supports generating a QR code on a device that is already signed in.
+ * Note that this is UNSTABLE and may have breaking changes without notice.
+ */
+export class MSC3906Rendezvous {
+    private newDeviceId?: string;
+    private newDeviceKey?: string;
+    private ourIntent: RendezvousIntent = RendezvousIntent.RECIPROCATE_LOGIN_ON_EXISTING_DEVICE;
+    public code?: string;
```
turt2live, line 46: Can this variable get a better name please? or at least a comment to indicate what's contained in it (base64 QR code?)

src/rendezvous/rendezvous.ts
```diff
@@ -0,0 +1,228 @@
...
+
+    constructor(
+        public channel: RendezvousChannel,
+        public client: MatrixClient,
+        public onFailure?: RendezvousFailureListener,
+    ) {}
+
+    async generateCode(): Promise<void> {
```
turt2live, line 54:
```suggestion
    public async generateCode(): Promise<void> {
```

src/rendezvous/rendezvous.ts
```diff
@@ -0,0 +1,228 @@
...
+        if (this.code) {
+            return;
+        }
+
+        this.code = JSON.stringify(await this.channel.generateCode(this.ourIntent));
+    }
+
+    async startAfterShowingCode(): Promise<string | undefined> {
```
turt2live, line 62:
```suggestion
    public async startAfterShowingCode(): Promise<string | undefined> {
```

src/rendezvous/rendezvous.ts
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        return checksum;
+    }
+
+    private async send({ type, ...payload }: { type: PayloadType, [key: string]: any }) {
+        await this.channel.send({ type, ...payload });
+    }
+
+    async declineLoginOnExistingDevice() {
```
turt2live, line 110:
```suggestion
    public async declineLoginOnExistingDevice() {
```

src/rendezvous/rendezvous.ts
```diff
@@ -0,0 +1,228 @@
...
+        logger.info('User declined sign in');
+        await this.send({ type: PayloadType.Finish, outcome: 'declined' });
+    }
+
+    async approveLoginOnExistingDevice(loginToken: string): Promise<string | undefined> {
```
turt2live, line 115:
```suggestion
    public async approveLoginOnExistingDevice(loginToken: string): Promise<string | undefined> {
```

src/rendezvous/channels/ecdhV1.ts
```diff
@@ -0,0 +1,298 @@
...
+                key: encodeBase64(this.ourPublicKey),
+            });
+        }
+
+        this.olmSAS.set_their_key(encodeBase64(this.theirPublicKey!));
+
+        const initiatorKey = isInitiator ? this.ourPublicKey : this.theirPublicKey!;
+        const recipientKey = isInitiator ? this.theirPublicKey! : this.ourPublicKey;
+        let aesInfo = SecureRendezvousChannelAlgorithm.ECDH_V1.toString();
+        aesInfo += `|${encodeBase64(initiatorKey)}`;
```
uhoreg, line 138: I'd suggest using unpadded base64 encoding in general for the final specced version, but this may be OK for now (and of course changing it here would require changing it in the other implementations)

src/rendezvous/rendezvous.ts
```diff
@@ -0,0 +1,228 @@
...
+            outcome: 'verified',
+            verifying_device_id: this.client.getDeviceId(),
+            verifying_device_key: this.client.getDeviceEd25519Key(),
+            master_key: masterPublicKey,
+        });
+
+        return info;
+    }
+
+    async verifyNewDeviceOnExistingDevice(timeout = 10 * 1000): Promise<DeviceInfo | CrossSigningInfo | undefined> {
```
turt2live, line 178:
```suggestion
    public async verifyNewDeviceOnExistingDevice(timeout = 10 * 1000): Promise<DeviceInfo | CrossSigningInfo | undefined> {
```

src/rendezvous/rendezvous.ts
```diff
@@ -0,0 +1,228 @@
...
+            throw new Error('No user ID set');
+        }
+
+        {
+            const deviceInfo = this.client.crypto.getStoredDevice(userId, this.newDeviceId);
+
+            if (deviceInfo) {
+                return await this.verifyAndCrossSignDevice(deviceInfo);
+            }
+        }
```
turt2live, line 204: Why is there a scope defined here? (if intentional, please add a comment. Otherwise, please remove)

src/rendezvous/rendezvous.ts
```diff
@@ -0,0 +1,228 @@
...
+        logger.info("Going to wait for new device to be online");
+        await sleep(timeout);
+
+        {
+            const deviceInfo = this.client.crypto.getStoredDevice(userId, this.newDeviceId);
+
+            if (deviceInfo) {
+                return await this.verifyAndCrossSignDevice(deviceInfo);
+            }
+        }
```
turt2live, line 215: another random scope?

src/rendezvous/rendezvous.ts
```diff
@@ -0,0 +1,228 @@
...
+
+        throw new Error('Device not online within timeout');
+    }
+
+    async cancel(reason: RendezvousFailureReason) {
```
turt2live, line 220:
```suggestion
    public async cancel(reason: RendezvousFailureReason) {
```

src/rendezvous/rendezvous.ts
```diff
@@ -0,0 +1,228 @@
...
+        this.onFailure?.(reason);
+        await this.channel.cancel(reason);
+    }
+
+    async close() {
```
turt2live, line 225:
```suggestion
    public async close() {
```

src/rendezvous/transport.ts
```diff
@@ -0,0 +1,59 @@
+/*
```
turt2live, line 1: `RendezvousTransport.ts`

src/rendezvous/transports/simpleHttpTransport.ts
```diff
@@ -0,0 +1,191 @@
+/*
```
turt2live, line 1: `MSC3886SimpleHttpRendezvousTransport.ts`

src/rendezvous/transports/simpleHttpTransport.ts
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/**
+ * Implementation of the unstable [MSC3886](https://github.com/matrix-org/matrix-spec-proposals/pull/3886)
+ * simple HTTP rendezvous protocol.
+ * Note that this is UNSTABLE and may have breaking changes without notice.
+ */
+export class MSC3886SimpleHttpRendezvousTransport implements RendezvousTransport {
+    ready = false;
+    cancelled = false;
```
turt2live, line 39:
```suggestion
    public ready = false;
    public cancelled = false;
```

also, it's a bit of a smell if these are mutable externally - should they be `private` with only a getter accessible?

src/rendezvous/transports/simpleHttpTransport.ts
```diff
@@ -0,0 +1,191 @@
...
+    private uri?: string;
+    private etag?: string;
+    private expiresAt?: Date;
+    public onFailure?: RendezvousFailureListener;
```
turt2live, line 43: please group public with public. Also same concern about mutability.

src/rendezvous/transports/simpleHttpTransport.ts
```diff
@@ -0,0 +1,191 @@
...
+    private client: MatrixClient;
+    private fallbackRzServer?: string;
+    private fetchFn?: typeof global.fetch;
+
+    constructor({
```
turt2live, line 48:
```suggestion
    public constructor({
```

src/rendezvous/transports/simpleHttpTransport.ts
```diff
@@ -0,0 +1,191 @@
...
+        fallbackRzServer?: string;
+    }) {
+        this.fetchFn = fetchFn;
+        this.onFailure = onFailure;
+        this.client = client;
+        this.fallbackRzServer = fallbackRzServer;
+        this.ready = !!this.uri;
+    }
+
+    async details(): Promise<MSC3886SimpleHttpRendezvousTransportDetails> {
```
turt2live, line 65:
```suggestion
    public async details(): Promise<MSC3886SimpleHttpRendezvousTransportDetails> {
```

src/rendezvous/transports/simpleHttpTransport.ts
```diff
@@ -0,0 +1,191 @@
...
+                return `${this.client.baseUrl}${ClientPrefix.Unstable}/org.matrix.msc3886/rendezvous`;
+            }
+        } catch (err) {
+            logger.warn('Failed to get unstable features', err);
+        }
+
+        return this.fallbackRzServer;
+    }
+
+    async send(contentType: string, data: any) {
```
turt2live, line 95:
```suggestion
    public async send(contentType: string, data: any) {
```

src/rendezvous/transport.ts
```diff
@@ -0,0 +1,59 @@
...
+   * @returns the transport details that can be encoded in a QR or similar
+   */
+  details(): Promise<RendezvousTransportDetails>;
+
+  /**
+   * Send data via the transport.
+   * @param contentType the content type of the data being sent
+   * @param data the data itself
+   */
+  send(contentType: string, data: any): Promise<void>;
```
turt2live, line 47: type for `any` please

src/rendezvous/transport.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Receive data from the transport.
+   */
+  receive(): Promise<any>;
```
turt2live, line 52: type for `any` please

src/rendezvous/transports/simpleHttpTransport.ts
```diff
@@ -0,0 +1,191 @@
...
+            // we would usually expect the final `url` to be set by a proper fetch implementation.
+            // however, if a polyfill based on XHR is used it won't be set, we we use existing URI as fallback
+            const baseUrl = res.url ?? uri;
+            // resolve location header which could be relative or absolute
+            this.uri = new URL(location, `${baseUrl}${baseUrl.endsWith('/') ? '' : '/'}`).href;
+            this.ready =true;
+        }
+    }
+
+    async receive(): Promise<any> {
```
turt2live, line 138:
```suggestion
    public async receive(): Promise<any> {
```

src/rendezvous/transports/simpleHttpTransport.ts
```diff
@@ -0,0 +1,191 @@
...
+                this.etag = poll.headers.get("etag") ?? undefined;
+                const data = await poll.json();
+
+                return data;
+            }
+            await sleep(1000);
+        }
+    }
+
+    async cancel(reason: RendezvousFailureReason) {
```
turt2live, line 172:
```suggestion
    public async cancel(reason: RendezvousFailureReason) {
```

src/rendezvous/channels/ecdhV1.ts
uhoreg, line 206: Shouldn't this return a `Promise<string>`?

turt2live (changes requested):
Nearly all of my remaining concerns are with the code style: while not enforced by linters, it's fairly important on large changes like this to not make the problem of inconsistent code worse. 

Overall, the code itself seems fine (pending other reviewer's concerns)

src/rendezvous/channels/ecdhV1.ts
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    public async receive(): Promise<any> {
+        if (!this.olmSAS) {
+            throw new Error('Channel closed');
+        }
+
+        const data = await this.transport.receive();
+        if (!data) {
+            return data;
```
uhoreg, line 249: Probably should return a fixed value (e.g. `null`?) to indicate that nothing was received, rather than returning the empty data as-is?

src/rendezvous/rendezvous.ts
hughns: It's because there was a variable name re-use. I've refactored to be more elegant.

src/rendezvous/rendezvous.ts
hughns: It's because there was a variable name re-use. I've refactored to be more elegant.

src/rendezvous/channels/ecdhV1.ts
hughns: Yeah, it's a bit late. Please do feedback into the actual MSCs and I can update.

turt2live:
Per internal conversation, this will land after the RC is cut.

src/rendezvous/channels/MSC3903ECDHv1RendezvousChannel.ts
```diff
@@ -0,0 +1,279 @@
...
+
+        const initiatorKey = isInitiator ? this.ourPublicKey : this.theirPublicKey!;
+        const recipientKey = isInitiator ? this.theirPublicKey! : this.ourPublicKey;
+        let aesInfo = RendezvousChannelAlgorithm.ECDH_V1.toString();
+        aesInfo += `|${encodeBase64(initiatorKey)}`;
+        aesInfo += `|${encodeBase64(recipientKey)}`;
+
+        const aesKeyBytes = this.olmSAS.generate_bytes(aesInfo, 32);
+
+        this.aesKey = await importKey(aesKeyBytes);
```
uhoreg, line 143: (After you switch things over to WebCrypto, ) you should clear out `aesKeyBytes` by [`fill`-ing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/fill) it with 0's, to make sure that the key doesn't remain in memory..

src/rendezvous/channels/MSC3903ECDHv1RendezvousChannel.ts
```diff
@@ -0,0 +1,279 @@
...
+
+        const rawChecksum = this.olmSAS.generate_bytes(aesInfo, 5);
+        return generateDecimalSas(Array.from(rawChecksum)).join('-');
+    }
+
+    private async encrypt(data: any): Promise<string> {
```
uhoreg, line 149: We'll probably eventually want to factor this function (and `decrypt`) out into `crypto/aes.ts`, but it should be OK to keep it here for now.

src/rendezvous/channels/MSC3903ECDHv1RendezvousChannel.ts
```diff
@@ -0,0 +1,279 @@
...
+
+        if (!subtleCrypto) {
+            throw new Error('Neither Web Crypto nor Node.js crypto are available');
+        }
+
+        const plaintext = await subtleCrypto.decrypt(
+            {
+                name: "AES-GCM",
+                iv: decodeBase64(iv),
+                tagLength: 128,
```
uhoreg, line 233: [MDN](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/decrypt) doesn't have `tagLength` documented.  Is it necessary?

uhoreg (commented):
Looks mostly OK from a crypto point of view

src/rendezvous/channels/MSC3903ECDHv1RendezvousChannel.ts
hughns: Agreed. You had mentioned that you wanted to migrate to AES-GCM for the file attachments at some point. I figured that would be a good time to do it.

src/rendezvous/channels/MSC3903ECDHv1RendezvousChannel.ts
hughns: Confusingly it is in the [AesGcmParams](https://developer.mozilla.org/en-US/docs/Web/API/AesGcmParams) doc on MDN instead.

src/rendezvous/channels/ecdhV1.ts
hughns: Now undefined.

src/rendezvous/rendezvous.ts
hughns: Comments added.

src/rendezvous/channels/ecdhV1.ts
hughns: Now rebased to use Web Crypto exclusively.

src/rendezvous/channels/MSC3903ECDHv1RendezvousChannel.ts
uhoreg: > Confusingly it is in the [AesGcmParams](https://developer.mozilla.org/en-US/docs/Web/API/AesGcmParams) doc on MDN instead.

Huh, strange. :shrug:

src/rendezvous/MSC3906Rendezvous.ts
```diff
@@ -0,0 +1,269 @@
...
+        const features = await buildFeatureSupportMap(await this.client.getVersions());
+        // determine available protocols
+        if (features.get(Feature.LoginTokenRequest) === ServerSupport.Unsupported) {
+            logger.info("Server doesn't support MSC3882");
+            await this.send({ type: PayloadType.Finish, outcome: Outcome.Unsupported });
+            await this.cancel(RendezvousFailureReason.HomeserverLacksSupport);
+            return undefined;
+        }
+
+        await this.send({ type: PayloadType.Progress, protocols: [LOGIN_TOKEN_PROTOCOL.name] });
```
t3chguy, line 113: It is strange that this call mixes stable and unstable-prefixed identifiers, both which are introduced in the same changeset, this seems wrong

src/rendezvous/MSC3906Rendezvous.ts
```diff
@@ -0,0 +1,269 @@
...
+            throw new Error('Linking failed');
+        }
+
+        this.newDeviceId = deviceId;
+        this.newDeviceKey = deviceKey;
+
+        return deviceId;
+    }
+
+    private async verifyAndCrossSignDevice(deviceInfo: DeviceInfo) {
```
t3chguy, line 177: Missing return type

src/rendezvous/MSC3906Rendezvous.ts
```diff
@@ -0,0 +1,269 @@
...
+
+        throw new Error('Device not online within timeout');
+    }
+
+    public async cancel(reason: RendezvousFailureReason) {
+        this.onFailure?.(reason);
+        await this.channel.cancel(reason);
+    }
+
+    public async close() {
```
t3chguy, line 266: Missing return type

src/rendezvous/MSC3906Rendezvous.ts
```diff
@@ -0,0 +1,269 @@
...
+        }
+
+        if (deviceInfo) {
+            return await this.verifyAndCrossSignDevice(deviceInfo);
+        }
+
+        throw new Error('Device not online within timeout');
+    }
+
+    public async cancel(reason: RendezvousFailureReason) {
```
t3chguy, line 261: Missing return type

src/rendezvous/RendezvousChannelAlgorithm.ts
```diff
@@ -0,0 +1,19 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+export enum RendezvousChannelAlgorithm {
+  ECDH_V1 = "org.matrix.msc3903.rendezvous.v1.curve25519-aes-sha256"
```
t3chguy, line 18: This should be an UnstableValue, to aid the future where the client has to be able to send stable but handle both

src/rendezvous/channels/MSC3903ECDHv1RendezvousChannel.ts
```diff
@@ -0,0 +1,255 @@
...
+        algorithm: RendezvousChannelAlgorithm.ECDH_V1;
+        key: string;
+    };
+}
+
+interface ECDHPayload {
+    algorithm?: RendezvousChannelAlgorithm.ECDH_V1;
+    key?: string;
+    iv?: string;
+    ciphertext?: string;
```
t3chguy, line 45: Seems strange for all these to be optional, you won't get far with them missing, what's the rationale?

src/rendezvous/channels/MSC3903ECDHv1RendezvousChannel.ts
```diff
@@ -0,0 +1,255 @@
...
+            }
+            return this.decrypt(data);
+        } else if (this.aesKey) {
+            throw new Error('Data received but no ciphertext');
+        }
+
+        return data;
+    }
+
+    public async close() {
```
t3chguy, line 241: Missing return type

src/rendezvous/channels/MSC3903ECDHv1RendezvousChannel.ts
```diff
@@ -0,0 +1,255 @@
...
+        if (this.olmSAS) {
+            this.olmSAS.free();
+            this.olmSAS = undefined;
+        }
+    }
+
+    public async cancel(reason: RendezvousFailureReason) {
```
t3chguy, line 248: Missing return type

src/rendezvous/channels/MSC3903ECDHv1RendezvousChannel.ts
```diff
@@ -0,0 +1,255 @@
...
+                tagLength: 128,
+            },
+            this.aesKey as CryptoKey,
+            ciphertextBytes,
+        );
+
+        return JSON.parse(new TextDecoder().decode(new Uint8Array(plaintext)));
+    }
+
+    public async receive(): Promise<object | undefined> {
```
t3chguy, line 219: This needs better typing, e.g. Generics, see `MatrixEvent::getContent` as an example

src/rendezvous/channels/MSC3903ECDHv1RendezvousChannel.ts
```diff
@@ -0,0 +1,255 @@
...
+            encodedData,
+        );
+
+        return {
+            iv: encodeBase64(iv),
+            ciphertext: encodeBase64(ciphertext),
+        };
+    }
+
+    public async send(payload: object) {
```
t3chguy, line 185: This needs better typing, e.g. Generics

src/rendezvous/transports/MSC3886SimpleHttpRendezvousTransport.ts
```diff
@@ -0,0 +1,197 @@
...
+    }) {
+        this.fetchFn = fetchFn;
+        this.onFailure = onFailure;
+        this.client = client;
+        this.fallbackRzServer = fallbackRzServer;
+    }
+
+    get ready() {
+        return this._ready;
+    }
```
t3chguy, line 71: Missing visibilty & return type

src/rendezvous/transports/MSC3886SimpleHttpRendezvousTransport.ts
```diff
@@ -0,0 +1,197 @@
...
+                return `${this.client.baseUrl}${ClientPrefix.Unstable}/org.matrix.msc3886/rendezvous`;
+            }
+        } catch (err) {
+            logger.warn('Failed to get unstable features', err);
+        }
+
+        return this.fallbackRzServer;
+    }
+
+    public async send(data: object) {
```
t3chguy, line 103: Missing return type, lacking types, we should avoid things like `any` and `object` - preferring Generics

src/rendezvous/transports/MSC3886SimpleHttpRendezvousTransport.ts
```diff
@@ -0,0 +1,197 @@
...
+            // we would usually expect the final `url` to be set by a proper fetch implementation.
+            // however, if a polyfill based on XHR is used it won't be set, we we use existing URI as fallback
+            const baseUrl = res.url ?? uri;
+            // resolve location header which could be relative or absolute
+            this.uri = new URL(location, `${baseUrl}${baseUrl.endsWith('/') ? '' : '/'}`).href;
+            this._ready = true;
+        }
+    }
+
+    public async receive(): Promise<object | undefined> {
```
t3chguy, line 146: lacking types as above

src/rendezvous/transports/MSC3886SimpleHttpRendezvousTransport.ts
```diff
@@ -0,0 +1,197 @@
...
+                this.etag = poll.headers.get("etag") ?? undefined;
+            } else if (poll.status === 200) {
+                this.etag = poll.headers.get("etag") ?? undefined;
+                return poll.json();
+            }
+            await sleep(1000);
+        }
+    }
+
+    public async cancel(reason: RendezvousFailureReason) {
```
t3chguy, line 179: Missing return type

t3chguy (changes requested):

src/rendezvous/channels/MSC3903ECDHv1RendezvousChannel.ts
t3chguy: Yes, please use Partial on the receive side in that case, we want to leverage types to ensure we don't miss sending the *actually required* fields

hughns: Done.

src/rendezvous/RendezvousChannelAlgorithm.ts
hughns: Done.

src/rendezvous/MSC3906Rendezvous.ts
hughns: Well, one could argue that everything needs unstable-prefixes as everything is new. I think that is overkill.

t3chguy: That's not necessarily the case. If the API you are sending on has an unstable prefix part of the same MSC then all identifiers within can be stable. Otherwise all new identifiers must be unstable.

turt2live (approved):

PR #2748, opened 2022-10-12
Emit events when setting unread notifications

gsouquet:
Paves the way for https://github.com/vector-im/element-web/issues/23192

## Checklist

---


src/models/room.ts
```diff
@@ -127,6 +128,7 @@ export enum RoomEvent {
     OldStateUpdated = "Room.OldStateUpdated",
     CurrentStateUpdated = "Room.CurrentStateUpdated",
     HistoryImportedWithinTimeline = "Room.historyImportedWithinTimeline",
+    UnreadNotifications = "Room.UnreadNotifications",
```
turt2live, line 131: it bothers me that we mix `Subject.lowerCamelCase` and `Subject.UpperCamelCase`, ftr. Personally I prefer `lowerCamelCase`, but that seems to be outvoted here.

src/models/room.ts
```diff
@@ -186,7 +192,8 @@ export class Room extends ReadReceipt<EmittedEvents, RoomEventHandlerMap> {
     public readonly reEmitter: TypedReEmitter<EmittedEvents, RoomEventHandlerMap>;
     private txnToEvent: Record<string, MatrixEvent> = {}; // Pending in-flight requests { string: MatrixEvent }
     private notificationCounts: NotificationCount = {};
-    private threadNotifications: Map<string, NotificationCount> = new Map();
+    private readonly threadNotifications: Map<string, NotificationCount> = new Map();
```
turt2live, line 195: Shorthand would be this fwiw:
```suggestion
    private readonly threadNotifications = new Map<string, NotificationCount>();
```

src/models/room.ts
```diff
@@ -1193,8 +1206,17 @@ export class Room extends ReadReceipt<EmittedEvents, RoomEventHandlerMap> {
      * @returns The notification count, or undefined if there is no count
      *          for this type.
      */
-    public getThreadUnreadNotificationCount(threadId: string, type = NotificationCountType.Total): number | undefined {
-        return this.threadNotifications.get(threadId)?.[type];
+    public getThreadUnreadNotificationCount(threadId: string, type = NotificationCountType.Total): number {
```
turt2live, line 1209: this looks like a breaking change - the function is not flagged as experimental, it's public, and it has a behavioural change.

src/models/room.ts
```diff
@@ -1193,8 +1206,17 @@ export class Room extends ReadReceipt<EmittedEvents, RoomEventHandlerMap> {
...
+        return this.threadNotifications.get(threadId)?.[type] ?? 0;
+    }
+
+    public hasThreadUnreadNotification(): boolean {
```
turt2live, line 1213: jsdoc please

turt2live (commented):
generally seems fine, though the breaking change is a bit uncomfortable.

src/models/room.ts
gsouquet: It should have been marked as experimental. And this is only behind a labs flag.
I believe this should not be marked as a breaking change.

src/models/room.ts
gsouquet: This kind of style comments should be enforced as part of the code style and not be part of reviews IMO.

gsouquet:
@turt2live the Typescript Strict Error CI has gone a bit coocoo. Besides that I believe I've adressed all your comments

src/models/room.ts
turt2live: The js-sdk doesn't have a concept of labs flags though, and we can't guarantee that people *aren't* using it. There's other breaking changes in the SDK already landed anyways, so I think we can just apply the label here and move on.

src/models/room.ts
turt2live: This was a suggestion, not a requirement.

turt2live (approved):

gsouquet:
@turt2live again, this change is NOT a breaking change. This function is behind a labs flag, and on top of that it has never made it to a release as it was introduced 7 days ago.

I will be removing the label

src/models/room.ts
```diff
@@ -1182,8 +1189,14 @@ export class Room extends ReadReceipt<EmittedEvents, RoomEventHandlerMap> {
      * @return {Number} The notification count, or undefined if there is no count
      *                  for this type.
      */
-    public getUnreadNotificationCount(type = NotificationCountType.Total): number | undefined {
-        return this.notificationCounts[type];
+    public getUnreadNotificationCount(type = NotificationCountType.Total): number {
```
turt2live, line 1192: This would be a breaking change, @gsouquet

turt2live (changes requested):
We should probably take this conversation OOB at this point - there are breaking changes in this PR, in at least one of the two functions, and there's no such thing as a labs flag here.

If the functions are intended to be marked experimental, please do so.

turt2live (approved):

PR #2749, opened 2022-10-13
Improve MatrixError message

t3chguy:
Unfortunately `fetch` also does not maintain stack traces, so errors are super opaque without this

---


weeman1337 (approved):
LGTM :+1:

PR #2753, opened 2022-10-13
Always send back an httpStatus property if one is known

kegsay:
Previously, non-JSON responses would be missing the `httpStatus` property, which was different to how `request()` used to work.

Ensure we always send this property, even for non-JSON responses.

---

## 🐛 Bug Fixes
 * Always send back an httpStatus property if one is known ([\#2753](https://github.com/matrix-org/matrix-js-sdk/pull/2753)).


gsouquet (approved):
LGTM 🚀

PR #2762, opened 2022-10-17
Remove node-specific crypto bits, use Node 16's WebCrypto

### LINKED ISSUE #2760: Simplify crypto by leveraging Node 16's WebCrypto implementation
migrate all the Node.js specific crypto implementations to be based on web crypto - https://nodejs.org/api/webcrypto.html

t3chguy:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/2760
Requires https://github.com/vector-im/element-web/pull/23506

---

## ✨ Features
 * Remove node-specific crypto bits, use Node 16's WebCrypto ([\#2762](https://github.com/matrix-org/matrix-js-sdk/pull/2762)). Fixes #2760.

### DIFF
--- a/src/crypto/aes.ts
+++ b/src/crypto/aes.ts
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 import { decodeBase64, encodeBase64 } from './olmlib';
-import { subtleCrypto } from "./crypto";
+import { subtleCrypto, crypto, TextEncoder } from "./crypto";
 
 // salt for HKDF, with 8 bytes of zeros
 const zeroSalt = new Uint8Array(8);
@@ -46,7 +46,7 @@ export async function encryptAES(
         iv = decodeBase64(ivStr);
     } else {
         iv = new Uint8Array(16);
-        window.crypto.getRandomValues(iv);
+        crypto.getRandomValues(iv);
 
         // clear bit 63 of the IV to stop us hitting the 64-bit counter boundary
         // (which would mean we wouldn't be able to decrypt on Android). The loss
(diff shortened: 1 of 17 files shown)


src/crypto/aes.ts
```diff
@@ -180,8 +90,8 @@ async function encryptBrowser(data: string, key: Uint8Array, name: string, ivStr
  * @param {Uint8Array} key the encryption key to use
```
robintown, line 90: The doc comment above still says 'browser'

src/crypto/store/memory-crypto-store.ts
```diff
@@ -292,17 +292,21 @@ export class MemoryCryptoStore implements CryptoStore {
         func(this.crossSigningKeys);
     }
 
-    public getSecretStorePrivateKey(txn: unknown, func: (key: IEncryptedPayload | null) => void, type: string): void {
-        const result = this.privateKeys[type];
+    public getSecretStorePrivateKey<T = IEncryptedPayload>(
```
robintown, line 295: Are these type parameters meant to be `extends IEncryptedPayload`, or can they really be anything?

robintown (commented):
Looks fine overall

src/crypto/store/memory-crypto-store.ts
t3chguy: They really can be anything, its entirely opaque object storage

uhoreg (approved):
Looks fine from a crypto standpoint.

src/crypto/store/memory-crypto-store.ts
t3chguy: I have improved this using a mapped type, so now everything we store has to be typed better

robintown (approved):

PR #2765, opened 2022-10-17
Re-emit unread notifs

kegsay:
This is required for sliding sync because currently the proxy does not send read receipts, which would otherwise update the unread counts. Without this, when you read a room (setting the counter to 0) on a different device, it does not live update. Needs react SDK changes as well.

Notes: Re-emit RoomEvent.UnreadNotifications

---

## 🐛 Bug Fixes
 * Re-emit RoomEvent.UnreadNotifications ([\#2765](https://github.com/matrix-org/matrix-js-sdk/pull/2765)).


kegsay:
Don't need this, just need the React SDK bits as that listens for Room events.

PR #2778, opened 2022-10-19
Apply more strict typescript around the codebase

t3chguy:
Requires https://github.com/matrix-org/matrix-react-sdk/pull/9474

---


t3chguy:
Increases overall coverage from 67.6% to 68% - infeasible to hit 80% mark with this typing fix.

src/NamespacedValue.ts
```diff
@@ -109,10 +109,10 @@ export class UnstableValue<S extends string, U extends string> extends Namespace
     }
 
     public get name(): U {
-        return this.unstable;
+        return this.unstable!;
```
kerryarchibald, line 112: Should be `U | undefined` return type rather than non-null assertion

src/NamespacedValue.ts
```diff
@@ -109,10 +109,10 @@ export class UnstableValue<S extends string, U extends string> extends Namespace
...
     }
 
     public get altName(): S {
-        return this.stable;
+        return this.stable!;
```
kerryarchibald, line 116: same

src/models/event.ts
```diff
@@ -685,14 +686,6 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
      * attempt is completed.
      */
     public async attemptDecryption(crypto: Crypto, options: IDecryptOptions = {}): Promise<void> {
-        // For backwards compatibility purposes
```
kerryarchibald, line 688: Should this qualify as a breaking change even though the signature is unchanged?

kerryarchibald (approved):
Looks reasonable to me, left a couple small questions.

src/models/event.ts
t3chguy: The type didn't allow for the backwards compatibility so it got partially removed at some point in the past

src/NamespacedValue.ts
t3chguy: The constructor says

```
// Stable is optional, but one of the two parameters is required, hence the weird-looking types.
// Goal is to to have developers explicitly say there is no stable value (if applicable).
```

So we can assert it will be defined given the `stable` wasn't. Will try express it with an overload

PR #2780, opened 2022-10-19
Fix connectivity regressions

dbkr:
Switches back to addTrack, digging the transceivers out manually to re-use, because the only way to group tracks into streams re-using trasceivers from the offer is to use setStreams which FF doesn't implement.

## Checklist

---

## 🐛 Bug Fixes
 * Fix connectivity regressions ([\#2780](https://github.com/matrix-org/matrix-js-sdk/pull/2780)).

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -647,20 +647,6 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
             videoMuted,
         }));
 
-        // gather transceivers from the new tracks so that we can use the same ones for tracks that
-        // we add later. We only do this for user media streams though: screenshare streams just always
-        // get their own unidirectional transceiver since a bidirectional screen share is pretty rare
-        // (we *could* re-use an existing recvonly transceiver for this, but it's simpler to just not).
-        // Actually, we don't do this since firefox doesn't implement setStreams() yet, so the only way
-        // we can group the tracks into streams is to use addTrack(). It remains here to hopefully be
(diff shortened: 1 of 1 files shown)


src/webrtc/call.ts
```diff
@@ -651,12 +651,15 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
         // we add later. We only do this for user media streams though: screenshare streams just always
         // get their own unidirectional transceiver since a bidirectional screen share is pretty rare
         // (we *could* re-use an existing recvonly transceiver for this, but it's simpler to just not).
-        if (purpose == SDPStreamMetadataPurpose.Usermedia) {
+        // Actually, we don't do this since firefox doesn't implement setStreams() yet, so the only way
+        // we can group the tracks into streams is to use addTrack(). It remains here to hopefully be
+        // resurrected once Firefox sort themselves out.
+        /*if (purpose == SDPStreamMetadataPurpose.Usermedia) {
```
SimonBrandner, line 657: Comments

src/webrtc/call.ts
```diff
@@ -700,10 +703,14 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
             purpose,
         }));
 
+        /*
+        Another commented block for retrieving existing transceivers: re-enable once we can use
```
SimonBrandner, line 707: I don't think we should leave in comments like this we can always use git to get the code back

SimonBrandner (changes requested):
Otherwise LGTM

PR #2783, opened 2022-10-21
fix build error caused by wrong ts-strict improvements

justjanne:
Type: Defect

---

## 🐛 Bug Fixes
 * fix build error caused by wrong ts-strict improvements ([\#2783](https://github.com/matrix-org/matrix-js-sdk/pull/2783)). Contributed by @justjanne.

### DIFF
--- a/src/crypto/DeviceList.ts
+++ b/src/crypto/DeviceList.ts
@@ -83,7 +83,7 @@ export class DeviceList extends TypedEventEmitter<EmittedEvents, CryptoEventHand
     // moment represented by the snapshot in the db.
     private syncToken: string | null = null;
 
-    private keyDownloadsInProgressByUser = new Map<string, Promise<void>>;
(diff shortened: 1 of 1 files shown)


src/crypto/DeviceList.ts
```diff
@@ -83,7 +83,7 @@ export class DeviceList extends TypedEventEmitter<EmittedEvents, CryptoEventHand
     // moment represented by the snapshot in the db.
     private syncToken: string | null = null;
 
-    private keyDownloadsInProgressByUser = new Map<string, Promise<void>>;
+    private keyDownloadsInProgressByUser: Map<string, Promise<void>> = new Map();
```
t3chguy, line 86:
```suggestion
    private keyDownloadsInProgressByUser = new Map<string, Promise<void>>();
```

t3chguy (approved):

Issue #2788, opened 2022-10-22
Crypto unavailable in React Native (non-node, non-browser) runtimes


bradjones1:
The change in https://github.com/matrix-org/matrix-js-sdk/pull/2762 is great, but it underscores an important limitation: The SDK cannot provide E2E encryption support in a runtime that lacks a WebCrypto implementation, e.g. React Native.

This SDK is amazing and I totally appreciate all the hard work that goes into it by the Matrix team. I did, however, need to do a bit of creative work to shim in at least _some_ sort of `crypto` module to get my RN project building, due to the new [`src/crypto/crypto.ts`](https://github.com/matrix-org/matrix-js-sdk/pull/2762/files#diff-67a1f31727c538badd116d232d024d7db6dd119eaa26b0c84aaeb08ecfccc9ee) being evaluated at build-time.

I wonder if it's possible to somehow tighten this up to bail out of loading any Crypto functionality if it's unsupported by the runtime?

t3chguy:
> need to do a bit of creative work to shim in at least some sort of crypto module to get my RN project building

How is it failing otherwise?

Is it just the require failing? We should wrap it in a try-catch
Should also add setter for the crypto, subtleCrypto & TextEncoder variables to allow easier shimming

bradjones1:
The related PR is a good step to address

> add setter for the crypto, subtleCrypto & TextEncoder variables to allow easier shimming

however to your question: 

> Is it just the require failing?

I _think_ so. I ended up shimming this to get my app building but I will report back here when I do remove the shim and see if these changes fixed this in my particular case.

PR #2791, opened 2022-10-24
Improve crypto init code and allow easier shimming

t3chguy:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/2788

---


weeman1337 (approved):
:+1:

PR #2794, opened 2022-10-24
Fix incorrect prevEv being sent in ClientEvent.AccountData events

t3chguy:
Account data events don't have an ID, clobber them on type instead
If multiple account data events came down sync, the previous event of the last event in the list was used for all prevEv's sent

---

## 🐛 Bug Fixes
 * Fix incorrect prevEv being sent in ClientEvent.AccountData events ([\#2794](https://github.com/matrix-org/matrix-js-sdk/pull/2794)).


weeman1337 (approved):
looks good :+1:

PR #2812, opened 2022-10-26
Replace `instanceof Array` with `Array.isArray`

### LINKED ISSUE #2811: SAS.ts assumes arrays are from the same JS realm
SAS.ts assumes the array from the SAS response is from the same JS realm, since it uses `instanceof Array` to detect if it really got an array:
https://github.com/matrix-org/matrix-js-sdk/blob/dddc0aeccb32c8efc1cbb7e7bd033c6bd251e375/src/crypto/verification/SAS.ts#L220

However, with the switch to fetch the arrays in parsed JSON responses don't come from the same JS realm as the SDK is running in (apparently). Changing the code to use `Array.isArray` fixes the issues for us.

t3chguy:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/2811

---

## 🐛 Bug Fixes
 * Replace `instanceof Array` with `Array.isArray` ([\#2812](https://github.com/matrix-org/matrix-js-sdk/pull/2812)). Fixes #2811.

### DIFF
--- a/src/crypto/verification/SAS.ts
+++ b/src/crypto/verification/SAS.ts
@@ -217,7 +217,7 @@ const MAC_SET = new Set(MAC_LIST);
 const SAS_SET = new Set(SAS_LIST);
 
 function intersection<T>(anArray: T[], aSet: Set<T>): T[] {
-    return anArray instanceof Array ? anArray.filter(x => aSet.has(x)) : [];
+    return Array.isArray(anArray) ? anArray.filter(x => aSet.has(x)) : [];
 }
 
 export enum SasEvent {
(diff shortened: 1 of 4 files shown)


turt2live (approved):
we should probably lint for this, but happy for that to be a different/future PR.

freaktechnik:
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-instanceof-array.md is an eslint rule for this I'm aware of.

t3chguy:
@freaktechnik thanks, will inherit

PR #2814, opened 2022-10-26
Switch ESLint warnings to be errors instead

gsouquet:
For https://github.com/vector-im/element-web/issues/23600

---


.eslintrc.js
```diff
@@ -21,12 +21,12 @@ module.exports = {
     // New changes belong in https://github.com/matrix-org/eslint-plugin-matrix-org/
     rules: {
         "no-var": ["warn"],
```
t3chguy, line 23: should this be error too?

t3chguy (approved):

PR #2834, opened 2022-11-01
sliding sync: add custom room subscriptions support

kegsay:
This is mostly useful when you need to change the subscription depending on the room. For example, unencrypted rooms have lazy-loaded members, but encrypted rooms do not. Previously it was not possible to express this as all rooms had to use the same subscription, even though the HTTP API is more flexible.

## Checklist

---

## ✨ Features
 * sliding sync: add custom room subscriptions support ([\#2834](https://github.com/matrix-org/matrix-js-sdk/pull/2834)).


t3chguy (approved):
Looks sane, might be worth adding a comment that a custom subscription overrides rather than mixes with the typical subscription

Issue #2116, opened 2022-01-23
Have js-sdk pass TSconfig `strict: true`


ShadowJonathan:
This will remove tons of bugs simply by virtue of function types becoming more correct.

Sub-issues;
- [ ] `strictNullChecks`: https://github.com/matrix-org/matrix-js-sdk/issues/2112
- [ ] `strictFunctionTypes`: TBD
  - [ ] https://github.com/matrix-org/matrix-js-sdk/issues/2114 
  (to properly check interface implementations)
- [ ] `noImplicitAny`: TBD
- [ ] ... (more from https://www.typescriptlang.org/tsconfig#strict)

Related: https://github.com/matrix-org/matrix-js-sdk/issues/2115

turt2live:
What are the benefits to doing this? I've yet to be convinced that `strict: true` is a good idea :/

ShadowJonathan:
`strict: true` will enable 8 other linters by default, this issue is meant to be partially a tracker for these as well.

All of these linters make TypeScript apply more scrutiny and static analysis to the TS source files, which in turn would make the overall program more deterministic and type-safe.

These are the applied linters, i'll explain each in turn;
- `alwaysStrict`
- `strictNullChecks`
- `strictBindCallApply`
- `strictFunctionTypes`
- `strictPropertyInitialization`
- `noImplicitAny`
- `noImplicitThis`
- `useUnknownInCatchVariables`

---

### `alwaysStrict`

This'll make Typescript emit its javascript in strict mode, and type-check it the same way as it would be in strict mode.

This has a number of advantages, all of which are listed [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode), but a few highlights;

- Unique function parameters
- Throw on assignment to get/readonly property
- Throw on assignment to a nonasignable global (`undefined`, `Infinity`, etc.)

Ultimately, the lint description highlights it as thus;

> ECMAScript strict mode was introduced in ES5 and provides behavior tweaks to the runtime of the JavaScript engine to improve performance, and makes a set of errors throw instead of silently ignoring them.

### `strictNullChecks`

This'll be explained in https://github.com/matrix-org/matrix-js-sdk/issues/2112.

### `strictBindCallApply`

Will apply stricter checking to `.call`, `.apply`, and `.bind`, making sure that the types given to its arguments are the same as given on the function signature.

[...]

PR #2835, opened 2022-11-01
Make the js-sdk conform to tsc --strict

### LINKED ISSUE #2112: Have js-sdk pass `strictNullChecks: true`
When checking https://github.com/matrix-org/matrix-js-sdk/pull/1817, and then looking at https://github.com/matrix-org/matrix-js-sdk/issues/983, i decided to take it upon myself to add a little bit more null-safety to the library.

When i added `strictNullChecks` to the config, 1147 errors popped up, most of these cascading from other errors.

t3chguy:
Requires https://github.com/matrix-org/matrix-react-sdk/pull/9528
~Requires https://github.com/matrix-org/matrix-events-sdk/pull/16~
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/2112
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/2116
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/2124

---

## ✨ Features
 * Make the js-sdk conform to tsc --strict ([\#2835](https://github.com/matrix-org/matrix-js-sdk/pull/2835)). Fixes #2112 #2116 and #2124.


src/autodiscovery.ts
```diff
@@ -448,12 +453,17 @@ export class AutoDiscovery {
                 };
             }
         } catch (err) {
-            const error = err as Error | string | undefined;
+            const error = err as AutoDiscoveryError | string | undefined;
+            let reason = "";
```
duxovni, line 457:
```suggestion
            let reason: string | undefined;
```

src/client.ts
```diff
@@ -3121,7 +3122,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         opts: IKeyBackupRestoreOpts,
     ): Promise<IKeyBackupRestoreResult> {
         const privKey = decodeRecoveryKey(recoveryKey);
-        return this.restoreKeyBackup(privKey, targetRoomId, targetSessionId, backupInfo, opts);
+        return this.restoreKeyBackup(privKey, targetRoomId!, targetSessionId!, backupInfo, opts);
```
duxovni, line 3125: I'm confused why we're saying that `targetRoomId` and `targetSessionId` are defined, aren't we explicitly allowing them to be undefined in the function's type signature and in the documentation?

src/client.ts
```diff
@@ -3283,7 +3284,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             throw new Error("End-to-end encryption disabled");
         }
 
-        const path = this.makeKeyBackupPath(roomId, sessionId, version);
+        const path = this.makeKeyBackupPath(roomId!, sessionId!, version!);
```
duxovni, line 3287: Same here

src/client.ts
```diff
@@ -3787,18 +3788,18 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
 
         // If we expect that an event is part of a thread but is missing the relation
         // we need to add it manually, as well as the reply fallback
-        if (threadId && !content["m.relates_to"]?.rel_type) {
-            const isReply = !!content["m.relates_to"]?.["m.in_reply_to"];
-            content["m.relates_to"] = {
-                ...content["m.relates_to"],
+        if (threadId && !content!["m.relates_to"]?.rel_type) {
+            const isReply = !!content!["m.relates_to"]?.["m.in_reply_to"];
```
duxovni, line 3792: Shouldn't these be `?` not `!`?

src/client.ts
```diff
@@ -3787,18 +3788,18 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            content!["m.relates_to"] = {
```
duxovni, line 3793: If we're making `content` optional, what's supposed to happen in this case?

src/client.ts
```diff
@@ -4097,7 +4098,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         if (!eventId?.startsWith(EVENT_ID_PREFIX)) {
             opts = txnId as IRedactOpts;
             txnId = eventId;
-            eventId = threadId;
+            eventId = threadId!;
```
duxovni, line 4101: Likewise, why does it make sense to say this is guaranteed non-null?

tsconfig.json
```diff
@@ -8,7 +8,8 @@
     "noImplicitAny": false,
     "noUnusedLocals": true,
     "noEmit": true,
-    "declaration": true
+    "declaration": true,
+      "strict": true
```
duxovni, line 12:
```suggestion
    "strict": true
```

duxovni (changes requested):
It seems like there are a lot of places where we're using `!` without any reason to be sure that values are actually defined and non-null; I'm not clear on why this is an improvement from a type-safety standpoint?

src/client.ts
t3chguy: Because of the complex method signature overload, otherwise we'd need to have 3 if-cases (one for each overload) to be able to call it without error

src/client.ts
t3chguy: Same as above

src/client.ts
t3chguy: `content` is only optional in the final signature, in all the callable overloads it must be specified, therefore it isn't really optional but overloads are strange in typescript

src/client.ts
t3chguy: See below

src/client.ts
t3chguy: It might be cleaner if the arg changes name to `contentOrTxnId` and then we create a better-typed `content` variable at the top of the method but then the param names get messy in the overload itself

src/client.ts
t3chguy: Due to the overload signatures, if eventId doesn't start with event id prefix then we know which overload we are in and can make more deductions than typescript is able to do for us

duxovni (approved):

package.json
```diff
@@ -7,6 +7,7 @@
   },
```
MadLittleMods, line 7: > It seems like there are a lot of places where we're using ! without any reason to be sure that values are actually defined and non-null; I'm not clear on why this is an improvement from a type-safety standpoint?
>
> *-- @duxovni, https://github.com/matrix-org/matrix-js-sdk/pull/2835#pullrequestreview-1163892857*

What's the thoughts/summary here?

duxovni: The upshot from the individual comments was basically, we've got a lot of messy stuff with overloads where it'd be a pain to convince typescript that we're doing the right thing, so it's best to just tell typescript it's fine so we can go ahead and enable strict mode globally which will make life easier in general.

MadLittleMods: And I assume we should avoid this situation more in future code if possible?

Do our current overloads doom us to `!` usage forever? Is there certain overload patterns we should avoid in the future?

t3chguy: Overloads with different combinations of optionals, e.g. `A is optional & B is mandatory` | `A is mandatory & B is optional` will either need an if-case per combination to handle or `!`s

PR #2842, opened 2022-11-02
Run tests in multiple supported node versions

t3chguy:


turt2live (commented):
clearing code review while draft

.github/workflows/tests.yml
```diff
@@ -8,25 +8,40 @@ concurrency:
   cancel-in-progress: true
 jobs:
   jest:
-    name: Jest
+    name: 'Jest [${{ matrix.specs }}] (Node ${{ matrix.node }})'
     runs-on: ubuntu-latest
+    timeout-minutes: 10
+    strategy:
+      matrix:
+        specs: [browserify, integ, unit]
```
richvdh, line 16: slightly worried about this getting out of sync. Presumably this is an attempt to speed up the build, by parallelising more? Was the time to run all the tests prohibitive?

It's not entirely obvious that waiting for 9 GHA workers to be available is actually going to be any quicker than just waiting for 3 to be available and doing more with each worker. Given it means that you have to do all the setup actions more often, it also means you're using more GHA worker-minutes in total, which has a negative effect on all the other projects using GHA.

Just some thoughts, not a blocking comment.

spec/setupTests.ts
```diff
@@ -17,3 +17,9 @@ limitations under the License.
 import DOMException from "domexception";
 
 global.DOMException = DOMException;
+
+jest.mock("../src/http-api/utils", () => ({
+    ...jest.requireActual("../src/http-api/utils"),
+    // We mock timeoutSignal otherwise it causes tests to leave timers running
+    timeoutSignal: () => new AbortController().signal,
+}));
```
richvdh, line 25: I'm getting pretty lost here. This stuff doesn't really seem to be related the stated purpose of this PR (running the tests in multiple node versions): is it about making it possible to run the suites separately?  My instinct is to say that it should be pulled out to a separate PR.

Alternatively: can you explain what this is doing, and why it is needed now where it wasn't before?

richvdh (commented):
I started reviewing this and got a bit lost :/

spec/setupTests.ts
t3chguy: It is for Node 19 compatibility, where things are stricter

.github/workflows/tests.yml
t3chguy: The browserify test is super basic, requires the most setup, and takes more than half the test time, it is nice to not have it block so you get a more responsive CI.

t3chguy: The longer term plan is to only test one node version on every commit and then run the full test of suites when the PR enters the [Merge Queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) - once this is available

spec/setupTests.ts
richvdh: so are all the changes to the `spec` files node 19 compat? or are there some changes for other things?

it really feels like this PR is conflating 3 things and I would love to see it split up

t3chguy: Node 19 compat, tests were hanging and Jest was failing them due to timers being left hanging

richvdh (approved):
seems plausible, though I would beg you to split this sort of thing up in future.

PR #2848, opened 2022-11-04
Extra insurance that we don't mix events in the wrong timelines

MadLittleMods:
Add checks to `addEventToTimeline` as extra insurance that we don't mix events in the wrong timelines (main timeline vs thread timeline).

Previously, we just relied on the callers to make sure they're doing the right thing and since it's easy to get it wrong, we mixed and bugs happened.

## Dev notes

---

```
yarn jest spec/unit/event-timeline-set.spec.ts
```

## Checklist

---

## 🐛 Bug Fixes
 * Extra insurance that we don't mix events in the wrong timelines ([\#2848](https://github.com/matrix-org/matrix-js-sdk/pull/2848)). Contributed by @MadLittleMods.


src/models/room.ts
```diff
@@ -1862,7 +1862,7 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         shouldLiveInThread: boolean;
         threadId?: string;
     } {
-        if (!this.client.supportsExperimentalThreads()) {
+        if (!this.client?.supportsExperimentalThreads()) {
```
MadLittleMods, line 1865: Same crutch used elsewhere in the file,
https://github.com/matrix-org/matrix-js-sdk/blob/4a33e584b01d0dd2f905a29a48bd98aa1458f155/src/models/room.ts#L382

The alternative is to fix the tests to not pass in `null` for the client but since we already made the decision to cut corners, I'm doing it here too.

florianduros: Maybe we should also show that the `client` is optional in the attribute definition. We kind of straddling between two waters here.

florianduros (approved):

src/models/room.ts
t3chguy: Yeah, honestly I'm surprised Sonar isn't flagging this like it does with extraneous `!`

MadLittleMods: `client` isn't optional and there is plenty of logic in `room.ts` isn't prepared for that yet. The real fix we can do is with the tests to not pass `null` for the `client`.

Perhaps I misunderstood but I'll leave this follow-up change to someone else.

MadLittleMods:
Thanks for the review @florianduros and @t3chguy 🐿

PR #2849, opened 2022-11-04
We don't care about the timeline (just fire and forget to load the event)

MadLittleMods:
We don't care about the timeline (just fire and forget to load the event).

> It's probably a misnomer to call it `getEventTimeline(): timeline` in the first place as it's more accurately used as `loadEventInTimeline(): Promise<void>` everywhere.
>
> *-- https://github.com/matrix-org/matrix-js-sdk/pull/2521#discussion_r920214132**

## Checklist

---

### DIFF
--- a/src/timeline-window.ts
+++ b/src/timeline-window.ts
@@ -101,7 +101,7 @@ export class TimelineWindow {
      *
      * @return {Promise}
      */
-    public load(initialEventId?: string, initialWindowSize = 20): Promise<void> {
+    public async load(initialEventId?: string, initialWindowSize = 20): Promise<void> {
         // given an EventTimeline, find the event we were looking for, and initialise our
         // fields so that the event in question is in the middle of the window.
         const initFields = (timeline: Optional<EventTimeline>) => {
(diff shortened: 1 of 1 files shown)


src/timeline-window.ts
```diff
@@ -135,13 +135,14 @@ export class TimelineWindow {
         // which is important to keep room-switching feeling snappy.
         if (this.timelineSet.getTimelineForEvent(initialEventId)) {
             initFields(this.timelineSet.getTimelineForEvent(initialEventId));
-            return Promise.resolve();
+            return;
         } else if (initialEventId) {
-            return this.client.getEventTimeline(this.timelineSet, initialEventId)
+            await this.client.getEventTimeline(this.timelineSet, initialEventId)
                 .then(initFields);
+            return;
```
MadLittleMods, line 142: The function signature is `Promise<void>` here. We shouldn't return anything here.

src/timeline-window.ts
```diff
@@ -135,13 +135,14 @@ export class TimelineWindow {
         // which is important to keep room-switching feeling snappy.
         if (this.timelineSet.getTimelineForEvent(initialEventId)) {
             initFields(this.timelineSet.getTimelineForEvent(initialEventId));
-            return Promise.resolve();
+            return;
         } else if (initialEventId) {
-            return this.client.getEventTimeline(this.timelineSet, initialEventId)
+            await this.client.getEventTimeline(this.timelineSet, initialEventId)
                 .then(initFields);
```
robintown, line 141: `this.client.getEventTimeline(...).then(initFields)` is already a `Promise<void>`?

robintown (commented):
I'm confused. It doesn't look like this changes any behavior?

src/timeline-window.ts
MadLittleMods: Sorry, I think I confused my sentiment with [this spot](https://github.com/matrix-org/matrix-react-sdk/blob/f5f79158ed5a950626bc89439b8319e362a75de1/src/components/structures/LoggedInView.tsx#L356) (which is totally different, oops) that I mentioned in https://github.com/matrix-org/matrix-js-sdk/pull/2521#discussion_r920214132. I'm trying to get back to speed with everything going on in that stale PR.

And in order to refactor to a fire and forget `getEventTimeline` pattern here, I'd need to adjust how we `initFields` is done here.

I'll have a think on this one whether it's relevant. Good shout!

richvdh:
removing review requests while this is in draft

PR #2852, opened 2022-11-04
Fix `refreshLiveTimeline(...)` not working when the latest event in the room is a threaded message

MadLittleMods:
Fix `refreshLiveTimeline(...)` not working when the latest event in the room is a threaded message.

 - Follow-up to https://github.com/matrix-org/matrix-js-sdk/pull/2299
 - `matrix-react-sdk` PR: https://github.com/matrix-org/matrix-react-sdk/pull/8354
 - This PR is a version 2 of https://github.com/matrix-org/matrix-js-sdk/pull/2521

This works because `fetchLatestLiveTimeline()` will get an `EventTimeline` for the latest events in the room. It first calls `/messages?dir=b` to find the latest message, `/context` to get the proper pagination tokens, sets those pagination tokens on the timeline so it can paginate appropriately. It doesn't matter whether the latest event in the room is a thread because if it is, then we will just create a new `EventTimeline` in the `EventTimelineSet`, otherwise will get the existing timeline.

---

See https://github.com/matrix-org/matrix-react-sdk/pull/8354#discussion_r892481473

We also have to keep in mind that we don't want to mix messages in the wrong timelines (main vs threaded timeline):

 - https://github.com/matrix-org/matrix-js-sdk/pull/2444
 - https://github.com/matrix-org/matrix-js-sdk/pull/2454

## Checklist

---

## 🐛 Bug Fixes
 * Fix `refreshLiveTimeline(...)` not working when the latest event in the room is a threaded message ([\#2852](https://github.com/matrix-org/matrix-js-sdk/pull/2852)). Contributed by @MadLittleMods.


spec/integ/matrix-client-event-timeline.spec.ts
```diff
@@ -1178,7 +1241,7 @@ describe("MatrixClient event timelines", function() {
                 await flushHttp(client.getEventTimeline(timelineSet, THREAD_ROOT.event_id!));
 
                 respondToMessagesRequest();
-                const timeline = await flushHttp(client.getLatestTimeline(timelineSet));
```
MadLittleMods, line 1181: It looks like some non-refresh-timeline tests got infected with `getLatestTimeline` usage in the months while these PR's have sat.

But I don't really want to support `getLatestTimeline` for threads which these tests rely on.

Either we have to maintain the old `getLatestTimeline(...)` alongside `getLatestLiveTimeline(...)` forever or we can refactor these tests away from it.

src/models/room.ts
```diff
@@ -913,9 +913,10 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         const backwardPaginationToken = liveTimelineBefore.getPaginationToken(EventTimeline.BACKWARDS);
         const eventsBefore = liveTimelineBefore.getEvents();
         const mostRecentEventInTimeline = eventsBefore[eventsBefore.length - 1];
+        const mostRecentEventIdInTimeline = mostRecentEventInTimeline?.getId();
```
MadLittleMods, line 916: I am surprised that TypeScript didn't catch this before. I would have assumed that TypeScript would infer `undefined` as a possibility from an array lookup but I guess not.

Related https://stackoverflow.com/questions/59479171/typescript-does-not-infer-undefined-as-a-return-type-from-array-prototype-find which mentions that `strictNullChecks` needs to be enabled for it to infer `undefined` there.

Is `strictNullChecks` separate from `strict`? We added `strict` recently in https://github.com/matrix-org/matrix-js-sdk/pull/2835

Related to https://github.com/vector-im/element-web/issues/21967 for adding `strictNullChecks`

src/webrtc/call.ts
```diff
@@ -411,10 +411,13 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
...
 
-        if (!this.client.deviceId) throw new Error("Client must have a device ID to start calls");
+        if (!this.client.getDeviceId()) throw new Error("Client must have a device ID to start calls");
 
         this.forceTURN = opts.forceTURN ?? false;
-        this.ourPartyId = this.client.deviceId;
+        this.ourPartyId = this.client.deviceId!;
+        if (!this.ourPartyId) {
+            logger.warn('MatrixCall: ourPartyId is null which may have unintended consequences');
+        }
```
MadLittleMods, line 420: Random change but the strict tsc was complaining here.

There is probably a better value to fallback to here? `this.client.deviceId`  can be `null`

robintown (commented):
Clearing element-call-reviewers review

src/webrtc/call.ts
MadLittleMods: Good point about the check above. I've updated it to use the same `const deviceId = this.client.getDeviceId();` reference so that the type inference can properly happen from that logic escape hatch

https://github.com/matrix-org/matrix-js-sdk/blob/4c49e68a865fc765e9028a1a2df9dcd7a1dd93ad/src/webrtc/call.ts#L414-L418

src/client.ts
```diff
@@ -5564,6 +5564,140 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            const backwardsState = timeline.getState(EventTimeline.BACKWARDS);
+            if (backwardsState) {
+                backwardsState.setUnknownStateEvents(contextRes.state.map(mapper));
+            } else {
+                throw new Error(
+                    `fetchLatestLiveTimeline: While updating backwards state of the existing ` +
+                    ` timeline=${timeline.toString()}, it unexpectedly did not have any backwards ` +
+                    `state. Something probably broke with the timeline earlier for this to happen.`,
+                );
+            }
```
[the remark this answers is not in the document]
MadLittleMods: Related to https://github.com/matrix-org/matrix-react-sdk/pull/9558#discussion_r1017295497 about not stressing over these safety checks

kerryarchibald (commented):
This PR duplicates a lot of code. I think it would be worth extracting and sharing the bulk of the `getEventTimeline`/`fetchLatestLiveTimeline ` functions that is afaict identical

spec/integ/matrix-client-event-timeline.spec.ts
```diff
@@ -910,6 +910,200 @@ describe("MatrixClient event timelines", function() {
...
+                undefined,
+                { timelineSupport: false },
+            );
+            client = testClient.client;
+            httpBackend = testClient.httpBackend;
+            await startClient(httpBackend, client);
+
+            const room = client.getRoom(roomId)!;
+            const timelineSet = room.getTimelineSets()[0]!;
+            await expect(client.fetchLatestLiveTimeline(timelineSet)).rejects.toBeTruthy();
```
robintown, line 935: I think there's a little room for deduplicating code in these first 3 tests

florianduros:
Hi!

Are you still working on it?

PR #2856, opened 2022-11-04
Extra insurance that we don't mix events in the wrong timelines - v2

MadLittleMods:
Add checks to `addEventToTimeline` as extra insurance that we don't mix events in the wrong timelines (main timeline vs thread timeline).

This PR is a v2 of https://github.com/matrix-org/matrix-js-sdk/pull/2848 since it was reverted in https://github.com/matrix-org/matrix-js-sdk/pull/2853

Previously, we just relied on the callers to make sure they're doing the right thing and since it's easy to get it wrong, we mixed and bugs happened.

## Dev notes

---

```
yarn jest spec/unit/event-timeline-set.spec.ts
```

## Todo

 - [x] Ensure the `matrix-react-sdk` Cypress tests pass with these changes
 - [x] Ensure the `matrix-react-sdk`, `yarn jest test/components/structures/ThreadView-test.tsx` pass with these changes
 - [x] Ensure the `matrix-react-sdk`, `yarn jest test/Notifier-test.ts` pass with these changes

## Checklist

---

## 🐛 Bug Fixes
 * Extra insurance that we don't mix events in the wrong timelines - v2 ([\#2856](https://github.com/matrix-org/matrix-js-sdk/pull/2856)). Contributed by @MadLittleMods.


src/models/event-timeline-set.ts
```diff
@@ -702,6 +702,25 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
...
+            throw new Error(`EventTimelineSet.addEventToTimeline: Timeline=${timeline.toString()} does not belong " + 
+                "in timelineSet(threadId=${this.thread?.id})`);
+        }
+
+        // Make sure events don't get mixed in timelines they shouldn't be in (e.g. a
+        // threaded message should not be in the main timeline).
+        //
+        // We can only run this check for timelines with a `room` because `canContain`
+        // requires it
+        if (this.room && !this.canContain(event)) {
```
MadLittleMods, line 715: Main change is here. We added the extra `this.room` check as `canContain` requires it

spec/unit/event-timeline-set.spec.ts
```diff
@@ -144,6 +168,58 @@ describe('EventTimelineSet', () => {
...
+                );
+                const nonRoomEventTimeline = new EventTimeline(nonRoomEventTimelineSet);
+
+                expect(nonRoomEventTimeline.getEvents().length).toStrictEqual(0);
+                nonRoomEventTimelineSet.addEventToTimeline(messageEvent, nonRoomEventTimeline, {
+                    toStartOfTimeline: true,
+                });
+                expect(nonRoomEventTimeline.getEvents().length).toStrictEqual(1);
+            });
+        });
```
MadLittleMods, line 222: Added a test for a non-room timeline scenario that would fail without the additional `this.room` check additonally added in this PR compared to v1.

src/models/event-timeline-set.ts
```diff
@@ -702,6 +702,28 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
             );
         }
 
+        if (timeline.getTimelineSet() !== this) {
+            throw new Error(`EventTimelineSet.addEventToTimeline: Timeline=${timeline.toString()} does not belong " + 
+                "in timelineSet(threadId=${this.thread?.id})`);
+        }
```
MadLittleMods, line 708: I'm confused by this SonarQube code coverage analysis, https://sonarcloud.io/component_measures?id=matrix-js-sdk&metric=new_coverage&pullRequest=2856&view=list

We have tests for both of these new conditions.

MadLittleMods: Latest results: https://sonarcloud.io/component_measures?id=matrix-js-sdk&metric=new_coverage&pullRequest=2856&view=list

spec/unit/event-timeline-set.spec.ts
```diff
@@ -144,6 +168,58 @@ describe('EventTimelineSet', () => {
...
+            expect(liveTimeline.getEvents().length).toStrictEqual(0);
+
+            eventTimelineSetForThread.addEventToTimeline(messageEvent, liveTimeline, {
+                toStartOfTimeline: true,
+            });
+            expect(liveTimeline.getEvents().length).toStrictEqual(0);
+        });
+
+        describe('non-room timeline', () => {
+            fit('Adds event to timeline', () => {
```
t3chguy, line 208: Only this test is being run due to `fit` - hence Sonar is unhappy with coverage.

src/models/event-timeline-set.ts
MadLittleMods: The cause was a left-in focused test `fit` causing the others not to be run, see https://github.com/matrix-org/matrix-js-sdk/pull/2856#discussion_r1015415938

t3chguy:
@MadLittleMods I suggest waiting on the auto assigned reviewers, if you're waiting on me it'll be end of week due to the fires I'm dealing with

src/models/room.ts
```diff
@@ -1862,7 +1862,7 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         shouldLiveInThread: boolean;
         threadId?: string;
     } {
-        if (!this.client.supportsExperimentalThreads()) {
+        if (!this.client?.supportsExperimentalThreads()) {
```
robintown, line 1865: Why was this change made, if the types say `client` is non-null? IMO we should either fix the places that pass in a falsy `client`, or change the type to match (though I strongly prefer the former)

src/models/event.ts
```diff
@@ -107,7 +107,7 @@ export interface IEventRelation {
     event_id?: string;
     is_falling_back?: boolean;
     "m.in_reply_to"?: {
-        event_id: string;
+        event_id?: string;
```
robintown, line 110: I'm also confused about this change, as the spec implies it's required

robintown (commented):
The main substance of the change looks good

src/models/room.ts
MadLittleMods: See https://github.com/matrix-org/matrix-js-sdk/pull/2848#discussion_r1013505393

src/models/event.ts
MadLittleMods: It's required for a valid relation but someone can choose to include it or not.

And judging by the other fields above being optional, this one also seems optional.

robintown: I believe the other fields such as `rel_type` are only optional because of replies being a special case. However in the case of replies, we know for certain that the `event_id` subkey is required, so I don't really see the point of letting invalid relations typecheck

MadLittleMods: There is no guarantee the the field will be present. AFAIK, we don't have any validation to make sure this is true. The only place we use this type is for the following:

https://github.com/matrix-org/matrix-js-sdk/blob/df2b65f111cf09df19b12fbe8210a54a1a23b3d8/src/models/event.ts#L1298-L1303

There is no guarantee that `m.in_reply_to` -> `event_id` is present in downstream code and our downstream code should treat it that way.

robintown: I see, arguably that method needs to do more validation, but I guess this is fine for now

robintown (approved):

MadLittleMods:
Thanks for the review @t3chguy and @robintown 🐄

spec/unit/event-timeline-set.spec.ts
MadLittleMods: Adding the lint via https://github.com/matrix-org/eslint-plugin-matrix-org/pull/33

PR #2861, opened 2022-11-08
Use a temporary Git dependency instead of patch-package

robintown:
patch-package is [only designed for use in top-level applications, not in libraries which get repackaged](https://github.com/ds300/patch-package/issues/198#issuecomment-579386635), and so will fail to run in upstream consumers of matrix-js-sdk. Until the [issues with matrix-events-sdk](https://github.com/matrix-org/matrix-events-sdk/pull/16) are resolved, this replaces the patch-package workaround with a temporary Git dependency to unblock upstream packages such as Element Call, while keeping the typechecker happy.

Notes: none

---
This change has no change notes, so will not be included in the changelog.


robintown:
Well, this doesn't work either

robintown:
I've checked the build artifacts for matrix-events-sdk into a fork, which seems to have worked (but feels very wrong)

PR #2867, opened 2022-11-10
Add browserslist configuration

gsouquet:
Added browserslist config that matches the supported browsers defined in https://github.com/matrix-org/matrix-js-sdk/#in-a-browser

## Checklist

---

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,1 @@
+> 0.5%, last 2 versions, Firefox ESR, not dead, node 18


.browserslistrc
```diff
@@ -0,0 +1 @@
+> 0.5%, last 2 versions, Firefox ESR, not dead, node 18
```
t3chguy, line 1: https://github.com/matrix-org/matrix-js-sdk/blob/develop/package.json#L6 we actually target Node 16+

gsouquet: Hm that contradicts what https://github.com/matrix-org/matrix-js-sdk#in-nodejs says.
Wondering whether we can claim node 16 support if we don't ship a ponyfill for `fetch` wdyt?

t3chguy: As part of constructing a MatrixClient you can pass a `fetchFn`

PR #2871, opened 2022-11-11
Remove `Crypto.start()`

richvdh:
This isn't really needed, and its semantics are poorly defined. (Contrary to the comment, it dos *not* set any background processes running).

---

### DIFF
--- a/spec/unit/crypto.spec.ts
+++ b/spec/unit/crypto.spec.ts
@@ -1122,4 +1122,22 @@ describe("Crypto", function() {
             expect(free).toHaveBeenCalled();
         });
     });
+
+    describe("start", () => {
+        let client: TestClient;
+
+        beforeEach(async () => {
+            client = new TestClient("@alice:example.org", "aliceweb");
+            await client.client.initCrypto();
+        });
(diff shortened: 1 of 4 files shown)


t3chguy:
`crypto` is a public member so this might be a breaking change

richvdh:
> crypto is a public member so this might be a breaking change

In theory yes. We're going to have to break a few eggs to make this omelette though, and this one seems like it should be pretty uncontentious: there should have been no good reason for clients to call `crypto.start()`.

t3chguy:
Can we bundle all the breaking changes into one release to not have to semver major bump multiple times?

richvdh:
> Can we bundle all the breaking changes into one release to not have to semver major bump multiple times?

I don't think that's going to be viable. This project is going to span multiple weeks, and I really don't want to have to work in a long-lived branch while that is going on.

For now I've added a stub back in for compatibility, but I think in general we're going to have to resign ourselves to breaking compat for applications that do weird stuff directly with the Crypto interface.

t3chguy (approved):

PR #2872, opened 2022-11-11
Upload device keys during initCrypto

richvdh:
Rather than waiting for the application to call `.startClient`, upload the device keys during `initCrypto()`. Element-R is going to approach this slightly differently (it wants to manage the decision on key uploads itself), so this lays some groundwork by collecting the libolm-specific bits together.

Based on https://github.com/matrix-org/matrix-js-sdk/pull/2871

Part of https://github.com/vector-im/element-web/issues/21972

---


SimonBrandner (approved):

richvdh:
(or I will, if I can ever get the typescript checks to pass. Looks like we need to wait for #2985)

PR #2873, opened 2022-11-11
Use an underride rule for Element Call notifications

SimonBrandner:
Fixes https://github.com/vector-im/element-web/issues/23691

---

## 🐛 Bug Fixes
 * Use an underride rule for Element Call notifications ([\#2873](https://github.com/matrix-org/matrix-js-sdk/pull/2873)). Fixes vector-im/element-web#23691.


t3chguy (approved):
Looks sane to me

PR #2876, opened 2022-11-13
fix registration add phone number not working

bagvand:
This PR will resolve below defect

Step to repro the defect
1-Go to register
2-Add phone number when register
3-enter correct OTP code and submit
4-server will return error because we are sending POST data via url params.

---

## 🐛 Bug Fixes
 * fix registration add phone number not working ([\#2876](https://github.com/matrix-org/matrix-js-sdk/pull/2876)). Contributed by @bagvand.


weeman1337 (approved):
Looks good to me @bagvand . Thanks for your contribution.

Tested sending a request using your branch locally. The data is now part of the body and no longer inside the URL.

PR #2880, opened 2022-11-15
Make calls go back to 'connecting' state when media lost

dbkr:
This is a change in how the state machine works: technically it's a breaking change, although likely not worth a specific major release given the level of breakage. Calls will now now go back into the connecting state if the media connection is lost (they'll try to re-establish the connection).

## Checklist

---

## ✨ Features
 * Make calls go back to 'connecting' state when media lost ([\#2880](https://github.com/matrix-org/matrix-js-sdk/pull/2880)).


SimonBrandner (approved):
LGTM

PR #2892, opened 2022-11-18
add-privileged-users-in-room

GoodGuyMarco:
This PR is required by https://github.com/matrix-org/matrix-react-sdk/pull/9596.

Notes: Enable multiple users' power levels to be set at once

## Checklist

Signed-off-by: Marco Bartelt marco.bartelt.ext@bwi.de

---

## ✨ Features
 * Enable multiple users' power levels to be set at once ([\#2892](https://github.com/matrix-org/matrix-js-sdk/pull/2892)). Contributed by @GoodGuyMarco.

### DIFF
--- a/spec/integ/matrix-client-methods.spec.ts
+++ b/spec/integ/matrix-client-methods.spec.ts
@@ -1337,27 +1337,38 @@ describe("MatrixClient", function() {
         });
     });
 
-    describe("registerWithIdentityServer", () => {
-        it("should pass data to POST request", async () => {
-            const token = {
-                access_token: "access_token",
-                token_type: "Bearer",
-                matrix_server_name: "server_name",
-                expires_in: 12345,
-            };
+    describe("setPowerLevel", () => {
+        it.each([
+            {
+                userId: "alice@localhost",
+                expectation: {
+                    "alice@localhost": 100,
+                },
+            },
+            {
+                userId: ["alice@localhost", "bob@localhost"],
+                expectation: {
+                    "alice@localhost": 100,
+                    "bob@localhost": 100,
+                },
+            },
+        ])("should modify power levels of $userId correctly", async ({ userId, expectation }) => {
+            const event = {
+                getType: () => "m.room.power_levels",
+                getContent: () => ({
+                    users: {
+                        "alice@localhost": 50,
+                    },
+                }),
+            } as MatrixEvent;
 
-            httpBackend!.when("POST", "/account/register").check(req => {
-                expect(req.data).toStrictEqual(token);
-            }).respond(200, {
-                access_token: "at",
-                token: "tt",
-            });
+            httpBackend!.when("PUT", "/state/m.room.power_levels").check(req => {
+                expect(req.data.users).toStrictEqual(expectation);
+            }).respond(200, {});
 
-            const prom = client!.registerWithIdentityServer(token);
+            const prom = client!.setPowerLevel("!room_id:server", userId, 100, event);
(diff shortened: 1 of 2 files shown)


robintown (approved):
Looks fine to me. Sorry this took a while to be reviewed!

robintown:
@GoodGuyMarco Our CI seems to be stuck. Could you please update your branch to restart it?

GoodGuyMarco:
@robintown @t3chguy Thank you for your review! I've just rebased this PR against `develop`. I also added a unit test covering the new changes. Please look into that aswell 🕵️

src/client.ts
```diff
@@ -3795,29 +3795,38 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
      */
     public setPowerLevel(
         roomId: string,
-        userId: string,
+        userId: string | string[],
         powerLevel: number,
         event: MatrixEvent,
     ): Promise<ISendEventResponse> {
-        let content = {
+        type EventContent = {
```
t3chguy, line 3812: the utils.deepCopy later on breaks this type, given it'll include other fields like `events`, `state_default`, etc. Once the typing for `deepCopy` is fixed this would no longer comply.

GoodGuyMarco: I see, any idea how to satisfy the strict type checker? I don't get why it fails. Locally it does not have any TypeScript issues in strict mode.

t3chguy: Make sure to also run `--noImplicitAny` given we (wrongly) specify that as `false` in tsconfig

GoodGuyMarco: Ah, thanks for the hint 💡  Is this a better solution: https://github.com/matrix-org/matrix-js-sdk/pull/2892/commits/763532f84a5774b90ce62208e8908545ec16ed7c?

src/client.ts
```diff
@@ -3795,29 +3795,36 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
     public setPowerLevel(
         roomId: string,
-        userId: string,
+        userId: string | string[],
         powerLevel: number,
         event: MatrixEvent,
     ): Promise<ISendEventResponse> {
+        type UsersWithPowerLevel = Record<string, number>;
         let content = {
             users: {},
```
t3chguy, line 3814: Wonder if you could do `users: {} as Record<string, number>,` and then not need the type assertions later

GoodGuyMarco: Yes, that seems to work aswell 👍

spec/integ/matrix-client-methods.spec.ts
```diff
@@ -1337,27 +1337,38 @@ describe("MatrixClient", function() {
...
 
-    describe("registerWithIdentityServer", () => {
-        it("should pass data to POST request", async () => {
-            const token = {
-                access_token: "access_token",
-                token_type: "Bearer",
-                matrix_server_name: "server_name",
-                expires_in: 12345,
-            };
+    describe("setPowerLevel", () => {
```
t3chguy, line 1340: It looks like you replaced an existing unrelated test?

GoodGuyMarco: Yes, there were two completely identical tests, see: https://github.com/matrix-org/matrix-js-sdk/blob/develop/spec/integ/matrix-client-methods.spec.ts#L1316

t3chguy (approved):
Looks sane to me, thanks for iterating!!

Issue #2894, opened 2022-11-18
Enable tsc `noImplicitAny`


t3chguy:
Requires https://gitlab.matrix.org/matrix-org/olm/-/merge_requests/63

PR #2895, opened 2022-11-18
Enable noImplicitAny

t3chguy:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/2894
Requires https://gitlab.matrix.org/matrix-org/olm/-/merge_requests/65

---


turt2live (approved):
pre-approved

PR #2897, opened 2022-11-18
Add jest metrics

florianduros:
## Checklist

FYI: jest doesn't seems to support TS file for custom reporter, so I have to stick with a js file.

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,7 +38,11 @@ jobs:
 
       - name: Run tests with coverage
         run: |
-          yarn coverage --ci --reporters github-actions --max-workers ${{ steps.cpu-cores.outputs.count }} ./spec/${{ matrix.specs }}
+          if [[ ${{ github.ref }} == refs/heads/develop ]]; then
+            yarn coverage --ci --reporters github-actions '--reporters=<rootDir>/spec/slowReporter.js' --max-workers ${{ steps.cpu-cores.outputs.count }} ./spec/${{ matrix.specs }}
+          else
+            yarn coverage --ci --reporters github-actions --max-workers ${{ steps.cpu-cores.outputs.count }} ./spec/${{ matrix.specs }}
+          fi
           mv coverage/lcov.info coverage/${{ matrix.node }}-${{ matrix.specs }}.lcov.info
         env:
           JEST_SONAR_UNIQUE_OUTPUT_NAME: true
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -38,7 +38,7 @@ jobs:
 
       - name: Run tests with coverage
         run: |
-          yarn coverage --ci --reporters github-actions --max-workers ${{ steps.cpu-cores.outputs.count }} ./spec/${{ matrix.specs }}
+          yarn coverage --ci --reporters github-actions '--reporters=<rootDir>/spec/slowReporter.js' --max-workers ${{ steps.cpu-cores.outputs.count }} ./spec/${{ matrix.specs }}
```
t3chguy, line 41: IMO this reporter shouldn't be used in PRs unless it was to fail PRs based on runtime. Though on develop runs it'd make sense.

spec/slowReporter.js
```diff
@@ -0,0 +1,100 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/* eslint-disable no-console */
+
+class JestSlowTestReporter {
```
turt2live, line 19: if we can make this TS, would be good for our overall project health.

florianduros: As I said in the PR description, jest doesn't seems to support reporter in ts

.github/workflows/tests.yml
florianduros: @t3chguy I made it run only on develop branch

.github/workflows/tests.yml
```diff
@@ -38,7 +38,11 @@ jobs:
 
       - name: Run tests with coverage
         run: |
-          yarn coverage --ci --reporters github-actions --max-workers ${{ steps.cpu-cores.outputs.count }} ./spec/${{ matrix.specs }}
+          if [[ ${{ github.ref }} == refs/heads/develop ]]; then
```
t3chguy, line 41: this will probably explode if someone puts quotes in their branch names which seem to be valid, maybe we should use https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif

t3chguy (approved):

PR #2898, opened 2022-11-18
Don't hang up calls that haven't started yet

dbkr:
## Checklist

---

## 🐛 Bug Fixes
 * Don't hang up calls that haven't started yet ([\#2898](https://github.com/matrix-org/matrix-js-sdk/pull/2898)).


robintown (commented):
I'd be interested to know how it's possible for `disposeCall` to get called before the call starts, just to make sure that this is the right place to fix this

dbkr:
Sondarcloud seems to want me to refactor `onCallFeedsChanged` which hasn't changed in this PR...

dbkr:
I strongly suspect it's if the call gets replaced before it's had a chance to send the invite

t3chguy:
> Sondarcloud seems to want me to refactor onCallFeedsChanged which hasn't changed in this PR...


Likely your base branch is out of date and thus re-introduces the issue

robintown:
@dbkr Then would https://github.com/matrix-org/matrix-js-sdk/blob/c389de98f36787f862173973b22079248eb351ae/src/webrtc/call.ts#L1082 perhaps be a more appropriate place to put this? Otherwise the replaced call would never clean up its holds on the streams, I think

dbkr:
Very good point - in fact that check should have included both the states in which we wouldn't have sent an invite anyway.

robintown (approved):

PR #2900, opened 2022-11-21
Further improvements to e2ee logging

richvdh:
A followup to #2884. In particular, it is not always the case that the
`sender_key` in a `m.room_key_withheld` message is the sender of that message.

---

## ✨ Features
 * Further improvements to e2ee logging ([\#2900](https://github.com/matrix-org/matrix-js-sdk/pull/2900)).

### DIFF
--- a/src/crypto/index.ts
+++ b/src/crypto/index.ts
@@ -3269,9 +3269,9 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
         }
 
         logger.info(
-            `Got room key withheld event from ${event.getSender()} (${content.sender_key}) `
-            + `for ${content.algorithm}/${content.room_id}/${content.session_id} `
-            + `with reason ${content.code} (${content.reason})`,
+            `Got room key withheld event from ${event.getSender()}q `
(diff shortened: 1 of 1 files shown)


src/crypto/index.ts
```diff
@@ -3269,9 +3269,9 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
         }
 
         logger.info(
-            `Got room key withheld event from ${event.getSender()} (${content.sender_key}) `
-            + `for ${content.algorithm}/${content.room_id}/${content.session_id} `
-            + `with reason ${content.code} (${content.reason})`,
+            `Got room key withheld event from ${event.getSender()}q `
```
weeman1337, line 3272: Is there `q` here by purpose?

weeman1337 (approved):
:+1:

one remark

src/crypto/index.ts
richvdh: absoulutely not :facepalm:

richvdh:
```suggestion
            `Got room key withheld event from ${event.getSender()} `
```

PR #2901, opened 2022-11-21
Improve perceived performance for threads

justjanne:
Type: Task

---

### DIFF
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -127,7 +127,7 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
 
         // even if this thread is thought to be originating from this client, we initialise it as we may be in a
         // gappy sync and a thread around this event may already exist.
-        this.initialiseThread();
+        this.updateThreadMetadata();
         this.setEventMetadata(this.rootEvent);
     }
 
@@ -188,7 +188,7 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
             this._currentUserParticipated = false;
             this.emit(ThreadEvent.Delete, this);
         } else {
-            await this.initialiseThread();
+            await this.updateThreadMetadata();
         }
     };
 
@@ -209,7 +209,7 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
         if (this.lastEvent === event) return;
         if (!event.isRelation(THREAD_RELATION_TYPE.name)) return;
 
-        await this.initialiseThread();
+        await this.updateThreadMetadata();
         this.emit(ThreadEvent.NewReply, this, event);
     };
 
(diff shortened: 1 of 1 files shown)


src/models/thread.ts
```diff
@@ -281,21 +281,32 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+    public async initialiseThread(): Promise<void> {
+        if (Thread.hasServerSideSupport) {
+            // Ensure we show *something* as soon as possible, we'll update it as soon as we get better data, but we
+            // don't want the thread preview to be empty if we can avoid it
+            if (!this.initialEventsFetched) {
+                await this.processRootEvent();
+            }
+            await this.fetchRootEvent();
+        }
+        await this.processRootEvent();
```
weeman1337, line 309: If both ifs above are truthy, then `await this.processRootEvent();` will be called twice.
Is this intended?

src/models/thread.ts
```diff
@@ -281,21 +281,32 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
             await this.processEvent(this.lastEvent);
         }
+    }
+
+    public async initialiseThread(): Promise<void> {
+        if (Thread.hasServerSideSupport) {
+            // Ensure we show *something* as soon as possible, we'll update it as soon as we get better data, but we
+            // don't want the thread preview to be empty if we can avoid it
+            if (!this.initialEventsFetched) {
+                await this.processRootEvent();
```
weeman1337, line 305: Does it make sense to process the root event before fetching it after this if?

weeman1337 (commented):
Tested threads on the deployment. They still seem  to be working.
Some comments though…

src/models/thread.ts
justjanne: Yes, because this allows us to show the bundled content, if available, while loading the updated content, instead of showing nothing. In the vast majority of situations, this should be the same content.

weeman1337 (approved):

src/models/thread.ts
t3chguy: Why call it twice though? Looks like it may cause a memory leak, as the `mapper` will be ran an extra time, causing a new MatrixEvent to get created which never has its listeners cleaned up so will avoid GC

src/models/thread.ts
```diff
@@ -317,15 +325,18 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+    }
 
-        if (Thread.hasServerSideSupport && bundledRelationship) {
-            this.replyCount = bundledRelationship.count;
-            this._currentUserParticipated = !!bundledRelationship.current_user_participated;
-
-            const mapper = this.client.getEventMapper();
-            this.lastEvent = mapper(bundledRelationship.latest_event);
-            await this.processEvent(this.lastEvent);
+    private async updateThreadMetadata(): Promise<void> {
```
0xM4m4n, line 330: was this meant to be turned private? initialiseThread() was previously used if somehow the threadtimeline hasn't loaded when requested so was called then. Is there another way to force a fetch/updating the thread data since `updateThreadMetadata` is now private?

justjanne: updateThreadMetadata shouldn't be loading the thread timeline in the first place, that timeline should be handled the same way as all other timelines nowadays. You should be able to just use `client.getLatestTimeline()` on the threads' timelineset, or `client.paginateEventTimeline()` on the threads' liveTimeline to force it to load.

PR #2902, opened 2022-11-21
Refactor GroupCall participant management

robintown:
This refactoring brings a number of improvements to `GroupCall`, which I've unfortunately had to combine into a single commit due to coupling:

- Moves the expiration timestamp field on call membership state to be per-device
- Makes the participants of a group call visible without having to enter the call yourself
- Enables users to join group calls from multiple devices
- Identifies active speakers by their call feed, rather than just their user ID
- Plays nicely with clients that can be in multiple calls in a room at once
- Fixes a memory leak caused by the call retry loop never stopping
- Changes `GroupCall` to update its state synchronously, and write back to room state asynchronously
  - This was already sort of halfway being done, but now we'd be committing to it
  - Generally improves the robustness of the state machine
  - It means that group call joins will appear instant, in a sense

For many reasons, this is a breaking change.

Notes: Enable users to join group calls from multiple devices

---

## 🚨 BREAKING CHANGES
 * Enable users to join group calls from multiple devices ([\#2902](https://github.com/matrix-org/matrix-js-sdk/pull/2902)).

### DIFF
--- a/spec/test-utils/webrtc.ts
+++ b/spec/test-utils/webrtc.ts
@@ -431,6 +431,7 @@ export class MockCallMatrixClient extends TypedEventEmitter<EmittedEvents, Emitt
 export class MockCallFeed {
     constructor(
         public userId: string,
+        public deviceId: string | undefined,
         public stream: MockMediaStream,
     ) {}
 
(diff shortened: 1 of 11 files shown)


robintown:
I've removed the review request for the general element-web team, since the only change outside of `src/webrtc` is the addition of a small utility function

dbkr (approved):
Phew - big PR! Looks plausible I think.

PR #2906, opened 2022-11-23
Add prettier

weeman1337:
https://github.com/matrix-org/eslint-plugin-matrix-org/pull/35 to be merged before
 
Should best be reviewed commit wise

## Checklist

---


spec/integ/megolm-integ.spec.ts
```diff
@@ -1010,70 +1010,73 @@ describe("megolm", () => {
...
-        });
-
-        // encrypt a message with the group session
-        const messageEncrypted = encryptMegolmEvent({
-            senderKey: testSenderKey,
-            groupSession: groupSession,
-            room_id: ROOM_ID,
-        });
+    it(
+        "should successfully decrypt bundled redaction events " + "that don't include a room_id in their /sync data",
```
weeman1337, line 1014: Looks silly, but if you remove the concat, prettier puts `it`, the string and `async () => {` into a single line. Then ESLint starts complaining about hitting the hard line length limit. :man_shrugging:

t3chguy: If prettier doesn't comprehend our hard line limit then this will be really painful to have to conjure

weeman1337: They added something about it in their docs
https://prettier.io/docs/en/options.html#print-width

As far as I understand that means use prettier or disable length checks in your other linters.

t3chguy: Sounds like we'll need to ask for people's opinions on this, conveniently it is thursday

turt2live (commented):
I haven't looked at this in depth, but believe we should probably land it soon (particularly considering the project is now fully TS --strict compliant, which means a boatload of conflicts are pending)

spec/integ/megolm-integ.spec.ts
weeman1337: We decided to disable `max-len` in ESLint

andybalaam (approved):
Do it!

.eslintrc.js
```diff
@@ -1,13 +1,6 @@
 module.exports = {
```
MadLittleMods, line 1: We should add these bulk fix commits to [`.git-blame-ignore-revs`](https://github.com/matrix-org/matrix-js-sdk/blob/193c38523c6304c86e56d6bdc5089ece26ad34da/.git-blame-ignore-revs) so we can easily skip over them in the git blame interface

(same with other bulk Typescript fixes that have happened recently)

weeman1337: Yes, definitively. I will open PRs later :+1:

PR #2908, opened 2022-11-24
Don't swallow errors coming from the shareSession call

poljar:
This PR intends to fix https://github.com/vector-im/element-web/issues/23792.

What should have been an easy fix sadly turned out to be a deciphering of the logic here. It seems to me that the `ensureOutboundSession()` has a bug which would wedge the outbound group session in the case of an error.

This wasn't problematic until now, since almost no errors happen in the `prepareSession()` call, but as we found out may happen in the `shareSession` call.

The first patch fixes the mentioned bug, aligning the code with the comments.
The second patch now propagates errors from the `shareSession` call to the caller.

The individual commits have even more info around what each of them does and means.

I tested this in Element Web, if there's an error in `shareSession` we get prompted to retry the sending of the event. If the error happened to be transient, a press of the retry button will successfully share the outbound session and send the event.

This closes:  https://github.com/vector-im/element-web/issues/23792.

---

## 🐛 Bug Fixes
 * Don't swallow errors coming from the shareSession call ([\#2908](https://github.com/matrix-org/matrix-js-sdk/pull/2908)). Fixes vector-im/element-web#23792. Contributed by @poljar.

### DIFF
--- a/src/crypto/algorithms/megolm.ts
+++ b/src/crypto/algorithms/megolm.ts
@@ -265,7 +265,9 @@ export class MegolmEncryption extends EncryptionAlgorithm {
      * Using `>>=` to represent the promise chaining operation, it does the
      * following:
      *
+     * ```
      * setupPromise = previousSetupPromise >>= setup >>= discardErrors
+     * ```
      *
      * The initial value for the `setupPromise` is a promise that resolves to
      * `null`. The forceDiscardSession() resets setupPromise to this initial
--- a/src/crypto/algorithms/megolm.ts
+++ b/src/crypto/algorithms/megolm.ts
@@ -256,6 +256,21 @@ export class MegolmEncryption extends EncryptionAlgorithm {
      * @param singleOlmCreationPhase - Only perform one round of olm
      *     session creation
      *
+     * This method updates the setupPromise field of the class by chaining a new
+     * call on top of the existing promise, and then catching and discarding any
+     * errors that might happen while setting up the outbound group session. This
+     * is done to ensure that `setupPromise` always resolves to `null` or the
+     * `OutboundSessionInfo`.
+     *
+     * Using `>>=` to represent the promise chaining operation, it does the
+     * following:
+     *
+     * setupPromise = previousSetupPromise >>= setup >>= discardErrors
+     *
+     * The initial value for the `setupPromise` is a promise that resolves to
+     * `null`. The forceDiscardSession() resets setupPromise to this initial
+     * promise.
+     *
      * @returns Promise which resolves to the
      *    OutboundSessionInfo when setup is complete.
      */
@@ -286,18 +301,20 @@ export class MegolmEncryption extends EncryptionAlgorithm {
         };
 
         // first wait for the previous share to complete
-        const prom = this.setupPromise.then(setup);
+        const fallible = this.setupPromise.then(setup);
 
-        // Ensure any failures are logged for debugging
-        prom.catch(e => {
+        // Ensure any failures are logged for debugging and make sure that the
+        // promise chain remains unbroken
+        //
+        // setupPromise resolves to `null` or the `OutboundSessionInfo` whether
+        // or not the share succeeds
+        this.setupPromise = fallible.catch(e => {
             logger.error(`Failed to setup outbound session in ${this.roomId}`, e);
+            return null;
         });
 
-        // setupPromise resolves to `session` whether or not the share succeeds
-        this.setupPromise = prom;
-
         // but we return a promise which only resolves if the share was successful.
-        return prom;
+        return fallible;
     }
 
     private async prepareSession(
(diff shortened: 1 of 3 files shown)


src/crypto/algorithms/megolm.ts
```diff
@@ -265,6 +265,21 @@ class MegolmEncryption extends EncryptionAlgorithm {
...
+     * `OutboundSessionInfo`.
+     *
+     * Using `>>=` to represent the promise chaining operation, it does the
+     * following:
+     *
+     * setupPromise = previousSetupPromise >>= setup >>= discardErrors
+     *
+     * The initial value for the `setupPromise` is a promise that resolves to
+     * `null`. The forceDiscardSession() resets setupPromise to this initial
+     * promise.
```
richvdh, line 281: it seems like this is an implementation detail which belongs in a comment inside the function, rather than in the doc-comment?

src/crypto/algorithms/megolm.ts
```diff
@@ -287,26 +300,28 @@ class MegolmEncryption extends EncryptionAlgorithm {
 
             try {
                 await this.shareSession(devicesInRoom, sharedHistory, singleOlmCreationPhase, blocked, session);
+                return session;
             } catch (e) {
                 logger.error(`Failed to ensure outbound session in ${this.roomId}`, e);
```
richvdh, line 305: Given that the error will be logged in the `catch` handler later, this seems redundant.

src/crypto/algorithms/megolm.ts
```diff
@@ -287,26 +300,28 @@ class MegolmEncryption extends EncryptionAlgorithm {
...
             }
-
-            return session;
         };
 
         // first wait for the previous share to complete
-        const prom = this.setupPromise.then(setup);
+        const fallible = this.setupPromise.then(setup);
 
         // Ensure any failures are logged for debugging
```
richvdh, line 313: ... and make sure that the promise chain remains unbroken.

src/crypto/algorithms/megolm.ts
```diff
@@ -287,26 +300,28 @@ class MegolmEncryption extends EncryptionAlgorithm {
...
+        const infallible = fallible.catch(e => {
             logger.error(`Failed to setup outbound session in ${this.roomId}`, e);
+            return null;
         });
 
-        // setupPromise resolves to `session` whether or not the share succeeds
-        this.setupPromise = prom;
+        // setupPromise resolves to `null` or the `OutboundSessionInfo` whether
+        // or not the share succeeds
+        this.setupPromise = infallible;
```
richvdh, line 321: I'd be tempted to inline `infallible` here.

richvdh (changes requested):

src/crypto/algorithms/megolm.ts
poljar: Considering that this method is private, and modifies a field of the class, I would say that it's perfectly reasonable to explain how it does so in the doc comment.

richvdh: Not entirely convinced, but I don't feel that strongly.

src/crypto/algorithms/megolm.ts
poljar: Removed in https://github.com/matrix-org/matrix-js-sdk/pull/2908/commits/d5074776f117afe2af16756bfb74a3c9063740da.

richvdh (approved):
lgtm

poljar:
Force pushed to get rid of the fixup commits.

poljar:
Force pushed to resolve a merge conflict.

What is needed to get this merged? I don't have write access in this repo, if the expectations is that I'm going to do that.

richvdh:
Oh, sorry, I didn't realise you didn't have write access. And now we have lint failures :(.

> Force pushed to resolve a merge conflict.

It should be possible to resolve a conflict without force-pushing?

poljar:
> It should be possible to resolve a conflict without force-pushing?

Sure, but I posted some fixup commits that required a rebase/autosquash workflow so I continued it. Are rebases frowned upon in this repo, should I stop using them as well as fixup commits?

richvdh:
Rebases after a review make it hard to see what has been changed since the review, so one tends to have to do the review all over again.

poljar:
> Rebases after a review make it hard to see what has been changed since the review, so one tends to have to do the review all over again.

I understand this, that's why the fixup commits were used before the approval. Are you re-reviewing my merge-conflict resolution? Should I not use fixup commits for this lint issue now?

richvdh:
@poljar unfortunately this is failing the test-coverage gate :(

poljar:
This has now a unrelated change that I'm trying to merge with https://github.com/matrix-org/matrix-js-sdk/pull/2959.

I'm closing this and going to reopen two separate PRs, one for the fix for `setupPromise` becoming wedged, and one to stop swallowing the errors.

PR #2912, opened 2022-11-25
sliding sync: add receipts extension

kegsay:
## Checklist

---

## ✨ Features
 * sliding sync: add receipts extension ([\#2912](https://github.com/matrix-org/matrix-js-sdk/pull/2912)).

### DIFF
--- a/spec/integ/sliding-sync-sdk.spec.ts
+++ b/spec/integ/sliding-sync-sdk.spec.ts
@@ -961,7 +961,7 @@ describe("SlidingSyncSdk", () => {
                     {
                         type: EventType.RoomMember,
                         state_key: alice,
-                        content: {membership: "join"},
+                        content: { membership: "join" },
                         sender: alice,
                         origin_server_ts: Date.now(),
                         event_id: "$alice",
@@ -982,7 +982,7 @@ describe("SlidingSyncSdk", () => {
                                 "m.read": {
                                     [alice]: {
                                         ts: 1234567,
-                                    }
+                                    },
                                 },
                             },
                         },
@@ -1009,7 +1009,7 @@ describe("SlidingSyncSdk", () => {
                                 "m.read": {
                                     [alice]: {
                                         ts: 1234567,
-                                    }
+                                    },
                                 },
                             },
                         },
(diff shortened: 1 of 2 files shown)


spec/integ/sliding-sync-sdk.spec.ts
```diff
@@ -928,4 +928,95 @@ describe("SlidingSyncSdk", () => {
...
+                                "m.read": {
+                                    [alice]: {
+                                        ts: 1234567,
+                                    },
+                                },
+                            },
+                        },
+                    },
+                },
+            });
```
alunturner, line 991: The object that's used here is reused on line 1003. It looks to me like tests in this area might quite often have to generate this object, so I was wondering if you think it warrants making a tiny util function that takes in the `roomId,  eventId, username` and returns the required object?

kegsay: I don't anticipate using this object a bunch, but given sonarcloud is whining I probably will be forced to make a util function for it.

src/sliding-sync-sdk.ts
```diff
@@ -253,22 +253,48 @@ class ExtensionTyping implements Extension {
...
+        if (isInitial) {
+            return {
+                enabled: true,
+            };
+        }
+        return undefined;
+    }
+
+    public onResponse(data: {rooms: Record<string, IMinimalEvent>}): void {
+        if (!data || !data.rooms) {
```
richvdh, line 290: given empty objects are truthy... if the type annotation is correct then this condition will always fail?

src/sliding-sync-sdk.ts
```diff
@@ -253,22 +253,48 @@ class ExtensionTyping implements Extension {
...
+        return ExtensionState.PostProcess;
+    }
+
+    public onRequest(isInitial: boolean): object | undefined {
+        if (isInitial) {
+            return {
+                enabled: true,
+            };
+        }
+        return undefined;
```
richvdh, line 286: I wouldn't have minded a comment to explain why this logic is correct, but that could be due to unfamiliarity with sliding sync and the modules therein

richvdh (approved):
seems generally plausible

PR #2910, opened 2022-11-25
Process `m.room.encryption` events before emitting `RoomMember` events

richvdh:
Fixes https://github.com/vector-im/element-web/issues/23819.

---
Here's what your changelog entry will look like:

## 🐛 Bug Fixes
 * Process `m.room.encryption` events before emitting `RoomMember` events ([\#2910](https://github.com/matrix-org/matrix-js-sdk/pull/2910)). Fixes vector-im/element-web#23819.

PR #2914, opened 2022-11-28
Process `m.room.encryption` events before emitting `RoomMember` events

richvdh:
Fixes https://github.com/vector-im/element-web/issues/23819

The root cause of https://github.com/vector-im/element-web/issues/23819 is as follows:
 * In `matrix-react-sdk`, [`ensureDMExists`](https://github.com/matrix-org/matrix-react-sdk/blob/v3.62.0-rc.1/src/createRoom.ts#L418-L433) tries to create an encrypted DM room, and assumes it is ready for use (including sending encrypted events) as soon as it receives a `RoomStateEvent.NewMember` notification indicating that the other user has been invited or joined. 
 * However, in `sync.ts`, we process the membership events in a `/sync` response  (including emitting `RoomStateEvent.NewMember` notifications) [here](https://github.com/matrix-org/matrix-js-sdk/blob/v22.0.0-rc.1/src/sync.ts#L1366), which is long before we process any `m.room.encryption` event ([here](https://github.com/matrix-org/matrix-js-sdk/blob/v22.0.0-rc.1/src/sync.ts#L1395)).
 * The upshot is that we can end up trying to send an encrypted event in the new room before processing the `m.room.encryption` event, which causes the crypto layer to blow up with the error in https://github.com/vector-im/element-web/issues/23819.

Strictly speaking, `ensureDMExists` probably ought to be listening for `ClientEvent.Room` as well as `RoomStateEvent.NewMember`; but that doesn't help us, because `ClientEvent.Room` is *also* emitted before we process the crypto event.

So, we need to process the crypto event before we start emitting these other events; but a corollary of that is that we need to do so *before* we store the new room in the client's store. That makes things tricky, because currently the crypto layer expects the room to have been stored in the client first.

So... we have to rearrange everything to pass the newly-created `Room` object into the crypto layer, rather than just the room id, so that it doesn't need to rely on getting the `Room` from the client's store.

I've tried to structure the commits in this PR in a way to make them easy to review.

(This is another go at https://github.com/matrix-org/matrix-js-sdk/pull/2910).

See also: 
 * https://github.com/matrix-org/matrix-react-sdk/pull/9650, which is a dummy PR to ensure that the current react-sdk and EW tests continue to pass with this change
 * https://github.com/matrix-org/matrix-react-sdk/pull/9619, which adds a cypress test that this works.

---

## 🚨 BREAKING CHANGES
 * Process `m.room.encryption` events before emitting `RoomMember` events ([\#2914](https://github.com/matrix-org/matrix-js-sdk/pull/2914)). Fixes vector-im/element-web#23819.

### DIFF
--- a/spec/integ/megolm-integ.spec.ts
+++ b/spec/integ/megolm-integ.spec.ts
@@ -21,16 +21,19 @@ import * as testUtils from "../test-utils/test-utils";
 import { TestClient } from "../TestClient";
 import { logger } from "../../src/logger";
 import {
-    IContent,
-    IEvent,
     IClaimOTKsResult,
-    IJoinedRoom,
-    ISyncResponse,
+    IContent,
     IDownloadKeyResult,
+    IEvent,
+    IJoinedRoom,
+    IndexedDBCryptoStore,
+    ISyncResponse,
+    IUploadKeysRequest,
     MatrixEvent,
     MatrixEventEvent,
-    IndexedDBCryptoStore,
     Room,
+    RoomMember,
+    RoomStateEvent,
 } from "../../src/matrix";
 import { IDeviceKeys } from "../../src/crypto/dehydration";
 import { DeviceInfo } from "../../src/crypto/deviceinfo";
@@ -327,7 +330,9 @@ describe("megolm", () => {
         const room = aliceTestClient.client.getRoom(ROOM_ID)!;
         const event = room.getLiveTimeline().getEvents()[0];
         expect(event.isEncrypted()).toBe(true);
-        const decryptedEvent = await testUtils.awaitDecryption(event);
+
+        // it probably won't be decrypted yet, because it takes a while to process the olm keys
+        const decryptedEvent = await testUtils.awaitDecryption(event, { waitOnDecryptionFailure: true });
         expect(decryptedEvent.getContent().body).toEqual('42');
     });
 
@@ -873,7 +878,12 @@ describe("megolm", () => {
 
         const room = aliceTestClient.client.getRoom(ROOM_ID)!;
         await room.decryptCriticalEvents();
-        expect(room.getLiveTimeline().getEvents()[0].getContent().body).toEqual('42');
+
+        // it probably won't be decrypted yet, because it takes a while to process the olm keys
+        const decryptedEvent = await testUtils.awaitDecryption(
+            room.getLiveTimeline().getEvents()[0], { waitOnDecryptionFailure: true },
+        );
+        expect(decryptedEvent.getContent().body).toEqual('42');
 
         const exported = await aliceTestClient.client.exportRoomKeys();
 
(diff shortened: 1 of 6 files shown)


src/crypto/index.ts
```diff
@@ -2972,10 +2972,11 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
     /**
      * handle an m.room.encryption event
      *
-     * @param {module:models/event.MatrixEvent} event encryption event
+     * @param room in which the event was received
+     * @param event encryption event to be processed
      */
-    public async onCryptoEvent(event: MatrixEvent): Promise<void> {
-        const roomId = event.getRoomId()!;
+    public async onCryptoEvent(room: Room, event: MatrixEvent): Promise<void> {
```
richvdh, line 3017: strictly speaking, this is a breaking change, because this method is accessible outside the js-sdk. But it's really not intended to be called anywhere except internally.

If we really want I could add a new `onCryptoEventWithARealRoom` method, use that from `sync.ts` etc, and deprecate the existing method. But it feels to me like that would be adding a bunch of dead code for very little gain.

spec/integ/megolm-integ.spec.ts
```diff
@@ -327,7 +330,9 @@ describe("megolm", () => {
         const room = aliceTestClient.client.getRoom(ROOM_ID)!;
         const event = room.getLiveTimeline().getEvents()[0];
         expect(event.isEncrypted()).toBe(true);
-        const decryptedEvent = await testUtils.awaitDecryption(event);
+
+        // it probably won't be decrypted yet, because it takes a while to process the olm keys
+        const decryptedEvent = await testUtils.awaitDecryption(event, { waitOnDecryptionFailure: true });
```
richvdh, line 335: this is needed due to timing differences in processing `/sync` responses: there are fewer `await`s, which means that we finish processing the `/sync` while there are still to-device messages in the queue for processing.

src/crypto/index.ts
```diff
@@ -2844,7 +2883,7 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
             delete content['io.element.performance_metrics'];
         }
 
-        const encryptedContent = await alg.encryptMessage(room, event.getType(), content);
+        const encryptedContent = await alg.encryptMessage(room, event.getType(), content) as IContent;
```
weeman1337, line 2886: Not sure about this cast. `encryptMessage` returns an object. That is all we know. Wouldn't it be better to update the `encryptMessage` definition instead?

weeman1337 (approved):
Thanks for all the comments. Helped me to review this one.

Looks good to me. :+1:  Also tested a bit using the dummy PR deployment. Everything seems to be working. Did some verifications (without having a DM). All succeeded.

src/crypto/index.ts
richvdh: yes, but honestly I've not touched this code and I'm sick to the back teeth of trying to land this fix so I just wanted to get something that passed the typechecker.

PR #2915, opened 2022-11-29
Prefer stable `/timestamp_to_event` endpoint first - MSC3030

MadLittleMods:
Prefer stable `/timestamp_to_event` endpoint first, then fallback to the unstable MSC3030 prefixed version.

[MSC3030](https://github.com/matrix-org/matrix-spec-proposals/pull/3030) has been merged into the spec.

Stabilized in Synapse via https://github.com/matrix-org/synapse/pull/14471

### Dev notes

Add console color codes
```ts
enum AnsiColorCode {
    Red = 31,
    Green = 32,
    Yellow = 33,
}
// Add color to text in the terminal output
function decorateStringWithAnsiColor(inputString: string, decorationColor: AnsiColorCode): string {
    return `\x1b[${decorationColor}m${inputString}\x1b[0m`;
}
```

### Todo

 - [ ] Create follow-up PR to add ANSI color codes back to test assertions
    - https://github.com/matrix-org/matrix-js-sdk/pull/2915/commits/bf78a64d821ab6581932624a27dbe00185ebac2b

## Checklist

---


src/client.ts
```diff
@@ -9297,31 +9297,51 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            return await this.http.authedRequest(
+                Method.Get,
+                path,
+                queryParams,
+                undefined,
+                {
+                    prefix: ClientPrefix.V1,
+                },
+            );
+        } catch (err) {
```
MadLittleMods, line 9327: Need exception for SonarCloud, https://sonarcloud.io/dashboard?id=matrix-js-sdk&pullRequest=2915

No interest in making a test for testing the prefixed fallback. It's all mocked garbage anyway.

t3chguy: Why not instead check the versions endpoint to see which spec releases the server supports and deduce API support level that way? One less round-trip per request and more testable.

MadLittleMods: @t3chguy `/timestamp_to_event` isn't in a Matrix spec version yet to deduce from and the stable version is not in `unstable_features`.

t3chguy: So how is the client meant to know whether to show the feature to the user or not? Normally unstable_features includes both the unstable and stable-but-not-in-a-spec-release flags.

https://github.com/matrix-org/synapse/blob/13aa29db1ddc925beb35f5f1da8fd1a1bcc91373/synapse/rest/client/versions.py#L82-L123

turt2live: fwiw, `stable-but-not-in-spec-release` flags are only meant to be created in two scenarios:
1. The stable release of the feature is expected to span multiple spec releases
2. The feature cannot be detected any other way (4xx error, etc)

We (the SCT) generally don't like to assign such flags because they're confusing and complicated to manage at the implementation level.

Instead, the recommendation is to convert the now-stable MSC to actual spec *before* updating the implementations to help build confidence in the spec version the changes will land in, allowing clients to do a simpler check for stability. In this case however, it appears the implementations are being updated first, which can cause process problems.

With my SCT hat on, I'd recommend pausing work on this PR and instead refocusing on writing up the MSC as actual spec. Currently the SCT is scheduled to look at it, so please drop by [#sct-office:matrix.org](https://matrix.to/#/#sct-office:matrix.org) to confirm if we're too far along the spec writing process for it to reasonably be taken over (the SCT's latest shift in priorities does not include writing up spec PRs for MSCs, so if the author or another person is more interested, we'll generally stop working on it in favour of someone else picking it up).

turt2live (changes requested):
With my web hat on, this does need a test to ensure the correct endpoints are called in fallback scenarios (if we're going that route). Otherwise, the unstable-stable flag or version checks need to be tested.

src/client.ts
MadLittleMods: > Normally unstable_features includes both the unstable and stable-but-not-in-a-spec-release flags.

I asked about "advertising support" in https://github.com/matrix-org/synapse/issues/14390. Using `org.matrix.msc3030.stable` would work well until we get a spec release 👍. Thanks for the example! I'll create a Synapse PR for that.

The alternative to determine support is to run `timestampToEvent(...)` against one of the users rooms and see if it succeeds or gives us a `M_UNRECOGNIZED`. If they haven't joined a room, then we just assume no support (small edge case until we get a spec release).


For reference, [`getRoomHierarchy`](https://github.com/matrix-org/matrix-js-sdk/blob/6611cfa2537ef14ce8c8f7d76a45dad4a4e26167/src/client.ts#L9116-L9146) uses this fallback pattern.

---

@turt2live We're supposed to be able to move fast with all of these things. Even if the spec PR was written, it seems like your suggestion is to wait for a Matrix spec release which isn't the fast thing.

It's unclear to me what your suggestion for moving fast is at least.

t3chguy: > The feature cannot be detected any other way (4xx error, etc)

This is fine for deciding which API to use, but not for whether or not to show the feature at all, having to call it with dummy params may fail due to stringent input checking on the server

turt2live: > @turt2live We're supposed to be able to move fast with all of these things. Even if the spec PR was written, it seems like your suggestion is to wait for a Matrix spec release which isn't the fast thing.
> 
> It's unclear to me what your suggestion for moving fast is at least.

By opening the spec PR, we'd move it from "will probably get written up eventually" to "next release material", allowing you to consider what that version number would be. The spec PR would also have to understand which spec release it's going into for the `added-in` flags sprinkled throughout the spec. 

Writing the spec PR first is the faster approach compared to managing 2 unstable flags and a spec release.

MadLittleMods: Quarterly spec releases is not fast for progressively using features.

I could have the stable endpoint used tomorrow using `org.matrix.msc3030.stable`.

turt2live: In case it's not clear, I'm strongly advocating *against* adding a `stable-but-unreleased` flag as it ultimately causes more problems than it solves for a feature of this size.

turt2live: > Quarterly spec releases is not fast for progressively using features.
> 
> I could have the stable endpoint used tomorrow using `org.matrix.msc3030.stable`.

is there a reason why you need the feature before a spec release? The unstable endpoint should be just as sufficient for this feature?

fwiw, the SCT is completely unaware of your timelines or priorities on this. If there's something driving this work which demands a faster timeline than the one available, we need to know by having it raised in the SCT office linked above.

MadLittleMods: > is there a reason why you need the feature before a spec release? The unstable endpoint should be just as sufficient for this feature?

The unstable endpoint is gone in the next Synapse release, https://github.com/matrix-org/synapse/pull/14471 (only stable endpoint)

> fwiw, the SCT is completely unaware of your timelines or priorities on this. If there's something driving this work which demands a faster timeline than the one available, we need to know by having it raised in the SCT office linked above.

We can move fast on features outside of the SCT. Nothing is needed from the SCT end for this to work.

The drive to do this work is to get it done so people can enjoy the feature when `matrix.org` gets the stable endpoint in the next Synapse release. Get a feature out the door.

I'm aware you want the spec PR written from [MSC3030](https://github.com/matrix-org/matrix-spec-proposals/pull/3030) but that's independent of us being able to accomplish things now.

turt2live: > > is there a reason why you need the feature before a spec release? The unstable endpoint should be just as sufficient for this feature?
> 
> The unstable endpoint is gone in the next Synapse release, [matrix-org/synapse#14471](https://github.com/matrix-org/synapse/pull/14471) (only stable endpoint)
> 

This feels premature and like a mistake, considering the impact here :(

The unstable and stable endpoints provide the exact same functionality (they did not change between implementation and acceptance of the MSC), and it would be more valuable to the end user to continue using the feature while the spec process does its thing: the unstable endpoint is more than able to be used for this.

Releasing the feature does *not* have to mean using stable endpoints.

> > fwiw, the SCT is completely unaware of your timelines or priorities on this. If there's something driving this work which demands a faster timeline than the one available, we need to know by having it raised in the SCT office linked above.
> 
> We can move fast on features outside of the SCT. Nothing is needed from the SCT end for this to work.
> 
> The drive to do this work is to get it done so people can enjoy the feature when `matrix.org` gets the stable endpoint in the next Synapse release. Get a feature out the door.
> 
> I'm aware you want the spec PR written from [MSC3030](https://github.com/matrix-org/matrix-spec-proposals/pull/3030) but that's independent of us being able to accomplish things now.

The SCT does need to be involved in consideration for unstable feature flags like this, as it affects the MSC's overall timeline and text. Adding a flag after acceptance is a process burden, as it needs to be formally documented and could even be rejected.

Aside from the SCT's involvement, the product and design teams also need to be consulted on this if you're aiming to release the feature as it's currently behind labs. Labs features are intended to break from time to time (so it's feasible to remove the unstable endpoint if you like, though expectations of having a stable endpoint can cause users to lose access to the feature), and removing those features from labs requires explicit sign off: the feature wouldn't be "shipped" just because a stable endpoint is being used.

MadLittleMods: > This feels premature and like a mistake, considering the impact here :(
> 
> The unstable and stable endpoints provide the exact same functionality (they did not change between implementation and acceptance of the MSC), and it would be more valuable to the end user to continue using the feature while the spec process does its thing: the unstable endpoint is more than able to be used for this.
> 
> Releasing the feature does not have to mean using stable endpoints.

It won't be enabled on `matrix.org` unless it's stable. With the next Synapse release, it will be stable.

This means a lot of people can actually use the jump to date lab on Element if they want.

We also need `/timestamp_to_event` enabled for the [canonical instance of the Matrix Public Archive](https://github.com/matrix-org/matrix-ansible-private/issues/5325) which means having it enabled on `matrix.org`.

As I mentioned [here](https://github.com/matrix-org/synapse/pull/14471#discussion_r1024523025), I'm not worried if there is a period where that functionality disappears. And there doesn't even have to be a period where it disappears if we merge something like this PR.


> Aside from the SCT's involvement, the product and design teams also need to be consulted on this if you're aiming to release the feature as it's currently behind labs. Labs features are intended to break from time to time (so it's feasible to remove the unstable endpoint if you like, though expectations of having a stable endpoint can cause users to lose access to the feature), and removing those features from labs requires explicit sign off: the feature wouldn't be "shipped" just because a stable endpoint is being used.

Whatever product/design input that needs to happen can happen on the de-labs PR for the jump to date feature.


---

I still see no alternative suggested that provides an equivalent speed to iteration for shipping a feature.

turt2live: To bring the quote wall back into a single thread: I'm trying to point out that there's nothing being shipped here. It's completely valid to have *both* unstable and stable endpoints supported by Synapse *without a config flag to disable the unstable feature* - if the stable endpoint is exposed, that should be more than enough reason to expose the unstable one too. 

The Synapse team's restrictions are also managed more easily when going through the spec writing path rather than the feature flag path. It's considerably less effort to write the spec up, get a known spec release assigned to the feature, and check for that spec release in client code (while still using the unstable endpoint until spec release) than it is to try and manage the heavyweight process of a stable feature flag.

This also applies to the Matrix Public Archive project and any other implementation which we might not be aware of. 

I'm not sure what planning has been done so far, but it really feels like this discussion is rooted in misinterpretation of how all the different processes interact and their purposes. Please also understand that I'm not making these suggestions to send you down the slowest or most inconvenient possible process: I genuinely believe that there is a better (and faster) way to get the feature *fully released* than the plan suggested by this PR.

turt2live: on this specific point though:

> And there doesn't even have to be a period where it disappears if we merge something like this PR.

Not everyone uses matrix.org, the latest Synapse release, or the latest Element release. Users are ultimately affected more than usual by the path this PR (and the associated Synapse PR) take.

turt2live: Further, while we've been having this conversation, Hubert has been working away at writing the spec PR - much of the discussion is effectively moot now that it's formally on the table for the next spec release.

MadLittleMods: > It's completely valid to have *both* unstable and stable endpoints supported by Synapse *without a config flag to disable the unstable feature* - if the stable endpoint is exposed, that should be more than enough reason to expose the unstable one too. 

Leaving the unstable endpoint in place until a spec release happens is a reasonable option 👍 - I can create a PR to add it back.

But I'd still like to move forward with a change like this PR so we can also use `/timestamp_to_event` on `matrix.org` ahead of the spec release (whether that includes `org.matrix.msc3030.stable` or however).



> Further, while we've been having this conversation, Hubert has been working away at writing the spec PR - much of the discussion is effectively moot now that it's formally on the table for the next spec release.

Having a spec PR is fantastic!

But this involves a lot more than a spec PR. We have to wait for the next Matrix spec release to be cut, we have to create an issue in Synapse to track changes for that new spec version, we have to implement everything, then advertise support for that spec version.

(these things aren't fast)

There is no question that this is the completely stable future but there is space to move without it.

turt2live: Reinstating the unstable endpoint means there's no need for an unstable-stable flag, as implementations can simply use the unstable endpoint. We know from the spec PR that it'll be in v1.6 of Matrix, so we can write client code to detect that specific version and use the stable endpoint in that case, which will automatically take effect when the server supports the spec version.

When the server adds support for advertising the spec version, it can also remove the unstable endpoint, leaving the feature fully functional for everyone. On matrix.org, the unstable endpoint would also be available with this setup, meaning a larger testing group for the Product team's consideration.

The feature can be de-labsed (released) at any point from here on out - with or without a spec release happening, and with or without a stable endpoint to call.

While these steps individually are not fast, when considered as a whole they are extremely efficient as they are designed for this exact scenario: wanting to get a feature live *before* the spec release even happens. An unstable-stable feature flag ends up *adding* steps, which means additional workload for all the teams involved and therefore more time.

Regardless, if you'd like to still continue with this PR, tests will be required per my other thread/review. I'd suggest adapting the PR to meet the plan I've outlined above, but ultimately the concern is lack of tests rather than the approach (as hopefully highlighted by the formal review left with my web team hat on).

MadLittleMods: > Reinstating the unstable endpoint means there's no need for an unstable-stable flag, as implementations can simply use the unstable endpoint. We know from the spec PR that it'll be in v1.6 of Matrix, so we can write client code to detect that specific version and use the stable endpoint in that case, which will automatically take effect when the server supports the spec version.

Thinking about this more, while it's still reasonable to leave the unstable endpoint alongside the stable one, I don't want to maintain both. This involves complexity on the server to choose the right federation endpoint depending on the entry and if we want a seamless transition while we support both, we have to do some fallbacks on the server as well. And underlying everything is that we don't have to let breaking unstable and labs features slow us down (they have no guarantees).

Things flow just fine on the client if we use the stable endpoint and fallback to unstable one.

For deciding whether to show a feature or not, I'd even lean towards just calling `timestampToEvent(firstRoom, 0, 'f')` and seeing if we get a successful response or an `M_UNRECOGNIZED` error. This avoids the unstable-stable flag as well.

Then when Matrix v1.6 is out, we can remove the quick and dirty support check in favor of a `v1.6` check. We could also do this first and fallback to the quick and dirty check.


> While these steps individually are not fast, when considered as a whole they are extremely efficient as they are designed for this exact scenario: wanting to get a feature live before the spec release even happens. An unstable-stable feature flag ends up adding steps, which means additional workload for all the teams involved and therefore more time.

No denying that waiting for every piece to be in place is the most efficient and the perfect thing to do if you have time to wait around 👍

Waiting for Matrix spec release:

 1. PR for unstable implementation on server
 1. PR for unstable implementation on client
 1. PR for stable implementation on server (dropping unstable implementation)
 1. PR for stable implementation on client with fallback to unstable (because easy)
 1. PR to remove unstable fallback on client

But that actually turns out to be almost the same amount of work as I'm trying to approach things. We just lose a little bit of compatibility in the transition:

 1. PR for unstable implementation on server
 1. PR for unstable implementation on client
 1. PR for stable implementation on server (dropping unstable implementation)
 1. PR for stable implementation on client with fallback to unstable (because easy)
 1. PR to remove unstable fallback and any updates for final stable checks on client

And if we try to be the most compatible by supporting both unstable and stable at the same time, it's actually more work:

 1. PR for unstable implementation on server
 1. PR for unstable implementation on client
 1. PR for stable implementation on server (with complexity of supporting both)
 1. PR for stable implementation on client with fallback to unstable
 1. PR to remove unstable implementation from server
 1. PR to remove unstable fallback on client

(hopefully these are right but it's kinda a whatever point. I don't disagree basically)


> Regardless, if you'd like to still continue with this PR, tests will be required per my other thread/review. I'd suggest adapting the PR to meet the plan I've outlined above, but ultimately the concern is lack of tests rather than the approach (as hopefully highlighted by the formal review left with my web team hat on).

👍 Sounds good, I'll add something here ⏩

MadLittleMods: > I'd also recommend just trying the real request instead of a limit of `0` - worst case is you look for [MSC3743](https://github.com/matrix-org/matrix-spec-proposals/pull/3743)'s error codes and repeat the same exact request.

For reference, `timestampToEvent(firstRoom, 0, 'f')` is a real request and `0` is the timestamp in this example. It will find the first event in the room.

Looks like we're thinking the same on all things around the check though 👍


> fwiw, in an unexpected discussion with the Synapse team around reinstating the unstable endpoint for this MSC in particular, the team is highly resistant to enable it by default due to factors I wasn't aware of. Those factors are being considered externally, but not in time for this PR's future.

Where did this discussion happen (link please, even if internal)? I'm looking for more context and what the factors are.

turt2live: > > fwiw, in an unexpected discussion with the Synapse team around reinstating the unstable endpoint for this MSC in particular, the team is highly resistant to enable it by default due to factors I wasn't aware of. Those factors are being considered externally, but not in time for this PR's future.
> 
> Where did this discussion happen (link please, even if internal)? I'm looking for more context and what the factors are.

It's specifically internal to the Spec Core Team at the moment. The results of the conversation will be made public as needed.

spec/unit/matrix-client.spec.ts
```diff
@@ -165,12 +203,15 @@ describe("MatrixClient", function() {
...
-        return new Promise(() => {});
+        // If you're seeing this then you forgot to handle at least 1 pending request.
+        const receivedRequest = decorateStringWithAnsiColor(`${method} ${prefix}${path}${new URLSearchParams(convertQueryDictToStringRecord(queryParams)).toString()}`, AnsiColorCode.Red);
+        const expectedRequest = decorateStringWithAnsiColor(`${next.method} ${next.prefix ?? ''}${next.path}${new URLSearchParams(convertQueryDictToStringRecord(next.expectQueryParams)).toString()}`, AnsiColorCode.Green);
+        throw new Error(
+            `A pending request was not handled: ${receivedRequest} ` +
+            `(next request expected was ${expectedRequest})\n` +
+            `Check your tests to ensure your number of expectations lines up with your number of requests ` +
+            `made, and that those requests match your expectations.`,
+        );
```
MadLittleMods, line 214: Better error message:

Colored terminal output of what this assertion looks like

spec/unit/matrix-client.spec.ts
```diff
@@ -104,7 +132,14 @@ describe("MatrixClient", function() {
...
         path: string;
     } | null = null;
-    function httpReq(method, path, qp, data, prefix) {
+    function httpReq(
+        method: Method,
+        path: string,
+        queryParams?: QueryDict,
+        body?: Body,
+        requestOpts: IRequestOpts = {}
+    ) {
```
MadLittleMods, line 141: Updated to match the actual thing:

https://github.com/matrix-org/matrix-js-sdk/blob/d692a5dbe239d06e04159c40f282159725415689/src/http-api/fetch.ts#L137-L143

spec/unit/matrix-client.spec.ts
```diff
@@ -231,6 +272,118 @@ describe("MatrixClient", function() {
         client.stopClient();
     });
 
+    describe('timestampToEvent', () => {
```
MadLittleMods, line 275: Calling out the new tests here since the last review cycle

spec/unit/matrix-client.spec.ts
```diff
@@ -165,12 +205,28 @@ describe("MatrixClient", function() {
...
+        // If you're seeing this then you forgot to handle at least 1 pending request.
+        throw new Error(
+            `A pending request was not handled: ${receivedRequest} ` +
+            `(next request expected was ${expectedRequest})\n` +
+            `Check your tests to ensure your number of expectations lines up with your number of requests ` +
+            `made, and that those requests match your expectations.`,
+        );
     }
 
     function makeClient() {
```
MadLittleMods, line 232: The `Typescript Strict Error Checker` CI job is failing (which is required), https://github.com/matrix-org/matrix-js-sdk/actions/runs/3588517236/jobs/6039998561

Since this file is a minefield, it has thrown a bunch of error. I've gone through and made sure the remaining examples are part of completely unrelated adjacent code.

One of the heavy hitters is the untyped `client` but that lets out a barrage of more errors if we try to simply correct it by adding a type. I'm going to defer all of those to the future.

Need an exception here.

spec/unit/matrix-client.spec.ts
```diff
@@ -165,12 +205,28 @@ describe("MatrixClient", function() {
...
+            `${next.method} ${next.prefix ?? ''}${next.path}${expectedQueryString}`,
+            AnsiColorCode.Green,
+        );
+        // If you're seeing this then you forgot to handle at least 1 pending request.
+        throw new Error(
+            `A pending request was not handled: ${receivedRequest} ` +
+            `(next request expected was ${expectedRequest})\n` +
+            `Check your tests to ensure your number of expectations lines up with your number of requests ` +
+            `made, and that those requests match your expectations.`,
+        );
```
MadLittleMods, line 229: Better error message:

Colored terminal output of what this assertion looks like

spec/unit/matrix-client.spec.ts
MadLittleMods: Re-commented on diff since this got outdated, https://github.com/matrix-org/matrix-js-sdk/pull/2915#discussion_r1036602902

spec/unit/matrix-client.spec.ts
```diff
@@ -104,7 +134,14 @@ describe("MatrixClient", function() {
...
         path: string;
     } | null = null;
-    function httpReq(method, path, qp, data, prefix) {
+    function httpReq(
+        method: Method,
+        path: string,
+        queryParams?: QueryDict,
+        body?: Body,
+        requestOpts: IRequestOpts = {},
+    ) {
```
MadLittleMods, line 143: Updated to match the actual thing:

https://github.com/matrix-org/matrix-js-sdk/blob/d692a5dbe239d06e04159c40f282159725415689/src/http-api/fetch.ts#L137-L143

spec/unit/matrix-client.spec.ts
MadLittleMods: Re-commented on the diff since this got outdated, https://github.com/matrix-org/matrix-js-sdk/pull/2915#discussion_r1036603134

spec/unit/matrix-client.spec.ts
```diff
@@ -52,6 +58,29 @@ jest.mock("../../src/webrtc/call", () => ({
...
 
+enum AnsiColorCode {
+    Red = 31,
+    Green = 32,
+    Yellow = 33,
+}
+// Add color to text in the terminal output
+function decorateStringWithAnsiColor(inputString: string, decorationColor: AnsiColorCode): string {
+    return `\x1b[${decorationColor}m${inputString}\x1b[0m`;
+}
```
turt2live, line 69: I'm hesitant to add console colouring code as there's about a million different runtimes we have to consider - this would be best placed in a dedicated PR to consider impact, avoiding a review delay here.

spec/unit/matrix-client.spec.ts
```diff
@@ -52,6 +58,29 @@ jest.mock("../../src/webrtc/call", () => ({
...
+function convertQueryDictToStringRecord(queryDict?: QueryDict): Record<string, string> {
+    if (!queryDict) {
+        return {};
+    }
+
+    return Object.entries(queryDict).reduce((resultant, [key, value]) => {
+        resultant[key] = String(value);
+        return resultant;
+    }, {} as Record<string, string>);
+}
```
turt2live, line 82: We should be using a `Map` instead of a `Record` - please ping internally on this if you have questions.

spec/unit/matrix-client.spec.ts
```diff
@@ -231,6 +287,130 @@ describe("MatrixClient", function() {
         client.stopClient();
     });
 
+    describe('timestampToEvent', () => {
+        const roomId = '!room:server.org';
+        const eventId = "$eventId:example.org";
+        const unstableMsc3030Prefix = "/_matrix/client/unstable/org.matrix.msc3030";
```
turt2live, line 293:
```suggestion
        const unstableMSC3030Prefix = "/_matrix/client/unstable/org.matrix.msc3030";
```

spec/unit/matrix-client.spec.ts
```diff
@@ -231,6 +287,130 @@ describe("MatrixClient", function() {
...
+            });
+        });
+
+        it('should fallback to unstable endpoint when no support for stable endpoint', async () => {
+            httpLookups = [{
+                method: "GET",
+                path: `/rooms/${encodeURIComponent(roomId)}/timestamp_to_event`,
+                prefix: ClientPrefix.V1,
+                error: {
+                    httpStatus: 404,
```
turt2live, line 327: we should also be testing `405 M_UNRECOGNIZED` per https://github.com/matrix-org/matrix-spec-proposals/pull/3743

spec/unit/matrix-client.spec.ts
```diff
@@ -231,6 +287,130 @@ describe("MatrixClient", function() {
...
+            });
+        });
+
+        it('should not fallback to unstable endpoint when stable endpoint returns an error', async () => {
+            httpLookups = [{
+                method: "GET",
+                path: `/rooms/${encodeURIComponent(roomId)}/timestamp_to_event`,
+                prefix: ClientPrefix.V1,
+                error: {
+                    httpStatus: 500,
```
turt2live, line 389: for safety, we should also test 429, 502, and 400

src/client.ts
```diff
@@ -9297,31 +9297,60 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            // Fallback to the prefixed unstable endpoint. Since the stable endpoint is
+            // new, we should also try the unstable endpoint before giving up. We can
+            // remove this fallback request in a year (remove after 2023-11-28).
+            if (
+                (<MatrixError>err).errcode === "M_UNRECOGNIZED" && (
+                    // XXX: The 400 status code check should be removed in the future
+                    // when Synapse is compliant with MSC3743.
+                    (<MatrixError>err).httpStatus === 400 ||
```
turt2live, line 9335: I'd somewhat suggest not going for lowest common denominator here and just look for MSC3743 instead. Synapse should have support as of recently, and the spec says MSC3743 is stable, and for the feature this impacts I'd expect the user to upgrade both their Element and Synapse before trying to use it (or their admin will do the update, at least).

turt2live (changes requested):
Overall this is looking better - thanks for adding the test! Just some points on the code itself and this should be fine to land.

spec/unit/matrix-client.spec.ts
MadLittleMods: This doesn't seem compatible with `URLSearchParams` out of the box and I haven't bogged myself into the details on the the type compatibility here. It might be best for you to just push up a commit for me to learn from here.

What benefit are we trying to gain with `Map` as well?

spec/unit/matrix-client.spec.ts
MadLittleMods: I don't see the benefit to this test. We use the same `GET` method for stable and unstable.

spec/unit/matrix-client.spec.ts
MadLittleMods: Are you suggesting that we shouldn't fallback to the unstable endpoint for `429` (rate-limit) and `502` (bad-gateway, probably load balancer timeout type problem)? That seems reasonable 👍

For `400`, I don't think we can do the same thing yet given MSC3743 isn't a thing just yet. We still have to rely on that to show that there isn't support.

MadLittleMods: > Are you suggesting that we shouldn't fallback to the unstable endpoint for `429` (rate-limit) and `502` (bad-gateway, probably load balancer timeout type problem)? That seems reasonable 👍

Diving into the tests/code again. We already do this. I don't see a benefit to adding extra tests for those particular status codes.

src/client.ts
MadLittleMods: All versions of Synapse that had unstable [MSC3030](https://github.com/matrix-org/matrix-spec-proposals/pull/3030) support, didn't support [MSC3743](https://github.com/matrix-org/matrix-spec-proposals/pull/3743). If we removed falling back when `400`, that means that the fallback would never be called and we might as well remove the fallback altogether.

spec/unit/matrix-client.spec.ts
```diff
@@ -268,6 +304,130 @@ describe("MatrixClient", function() {
...
+                expectQueryParams: {
+                    ts: '0',
+                    dir: 'f',
+                },
+            }];
+
+            await client.timestampToEvent(roomId, 0, Direction.Forward);
+
+            expect(mocked(client.http.authedRequest).mock.calls.length).toStrictEqual(1);
+            const [method, path, queryParams,, { prefix }] = mocked(client.http.authedRequest).mock.calls[0];
```
MadLittleMods, line 326: What's a good way to solve this Typescript lint?

https://github.com/matrix-org/matrix-js-sdk/actions/runs/3653310171/jobs/6172633351#step:5:7
```
Property 'prefix' does not exist on type 'IRequestOpts | undefined'.
```

This seems to work but is there something better?
```suggestion
            const [method, path, queryParams,, { prefix }  = { prefix: undefined }] = mocked(client.http.authedRequest).mock.calls[0];
```

spec/unit/matrix-client.spec.ts
MadLittleMods: Looks like I bunch of this was also done in https://github.com/matrix-org/matrix-js-sdk/pull/2895

turt2live:
for some reason this PR isn't listed in my review queue, but I am apparently getting notifications for it.

@MadLittleMods when it is ready for re-review, please ping me oob and I'll take a look. The CI looks a bit angry at the moment.

spec/unit/matrix-client.spec.ts
MadLittleMods: > @MadLittleMods when it is ready for re-review, please ping me oob and I'll take a look. The CI looks a bit angry at the moment.

@turt2live The CI is talking about this point. I've pushed the first change suggested here but still curious if anything comes to mind?

spec/unit/matrix-client.spec.ts
MadLittleMods: This has resolved itself now that the codebase was updated to be strict.

spec/unit/matrix-client.spec.ts
turt2live: I'll have to look into this one more another time.

spec/unit/matrix-client.spec.ts
turt2live: Like above, we're testing that the specific handling we have in place is compliant with the MSC - a relatively important test imo.

src/client.ts
turt2live: That sounds regrettable, but to be honest it's pretty appealing to not have a fallback then. We're primarily a Matrix client, not a Synapse client, and 400 is a really awful error code to do logic like this on, so instead there's likely a path where we just don't fall back and point at MSC3743 for not existing sooner.

I'm somewhat easily convinced on either side though, if you have thoughts on the matter.

turt2live (commented):
review per comments - hopefully re-review shows up in my queue this time

src/client.ts
MadLittleMods: Opting to leave the fallback in place for now ⏩

spec/unit/matrix-client.spec.ts
turt2live: The code says we're following MSC3743 though? We should be making sure we actually are...

turt2live: That MSC isn't set up to be broken down like that though - we really do need a test case here, sorry.

spec/unit/matrix-client.spec.ts
turt2live: @MadLittleMods was this resolved? Github appears to think it is.

MadLittleMods: @turt2live The lint itself is resolved because I made the change as described above (see comment just above, "I've pushed the first change suggested here").

But I had left this unresolved ("but still curious if anything comes to mind?") in hopes there might be a better taste-making option here. This whole point is asking if there are better options.

It can be resolved if you have no further input to add.

spec/unit/matrix-client.spec.ts
MadLittleMods: This isn't something I will be addressing with the current reasoning.

turt2live: The current reasoning is spec/MSC compliance, which is a blocker for this PR landing, sorry.

andybalaam: Fixed in 628bcbf33a3dfa3d685ee5bf4bc933beca30590f

spec/unit/matrix-client.spec.ts
andybalaam: Fixed in 12cc7be31cd7f7891e1f89bdd1636d536e596e75

spec/unit/matrix-client.spec.ts
andybalaam: Hopefully you prefer the code after my little refactoring:

```
const [callMethod, callPath, callQueryParams, , callOpts] = call;
const callPrefix = callOpts?.prefix;
```

andybalaam (approved):
@MadLittleMods thank you for all your work on this. With the little changes I added, it looks great.

spec/unit/matrix-client.spec.ts
```diff
@@ -87,6 +100,42 @@ type WrappedRoom = Room & {
     _state: Map<string, any>;
 };
 
+describe("convertQueryDictToStringRecord", () => {
```
MadLittleMods, line 103:
```suggestion
describe("convertQueryDictToMap", () => {
```

spec/unit/matrix-client.spec.ts
MadLittleMods: What benefit are we gaining from using a `Map`? It looks like we just convert it back to an array to use it anyway

Better semantics?

andybalaam: Just that it fits the semantics of what we are doing.

spec/unit/matrix-client.spec.ts
MadLittleMods: Adding support for `405` seems like the wrong move. If we ever see a `405` response here, then this is most likely an issue with Element itself (we used the wrong method in our first API call against the stable endpoint) and we should try to fallback. It was better to blow-up the app so the user has feedback to report or ideally auto-report the issue.

spec/unit/matrix-client.spec.ts
MadLittleMods: While I appreciate that I didn't add the tests here, adding tests specifically for `429` and `502` feels like an extra burden we will have to maintain in the future. The shared `assertRequestsMade` test harness makes this burden of supporting any extra status code responses less though.

We already had tests for when we should fallback and a test to cover the generic case of *any other* status codes where we shouldn't fallback.

With `429`/`502`, we're covering the main scenarios we're likely to run into so it's not unreasonable but it's sliding into the territory of "why not have a test for every possible status code?"

Anyway, if no hearts and minds we're convinced ⏩

---

(adding the test for `400` alongside `404` for when we should fallback makes sense)

src/client.ts
```diff
@@ -9391,7 +9391,10 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
                 ((<MatrixError>err).httpStatus === 400 ||
                     // This the correct standard status code for an unsupported
                     // endpoint according to MSC3743.
-                    (<MatrixError>err).httpStatus === 404)
+                    (<MatrixError>err).httpStatus === 404 ||
+                    // This the correct standard status code for an invalid
+                    // method according to MSC3743.
```
turt2live, line 9429: both error codes are the correct standard status codes - comment should be moved up a line and adjusted

turt2live (approved):
otherwise lgtm - thanks for taking a look

src/client.ts
andybalaam: Fixed in c7c16256d

spec/unit/matrix-client.spec.ts
MadLittleMods: @andybalaam came up with a great example for a slightly different case but makes this code make more sense. Imagine a scenario where the server supports `GET /foo` and there is a MSC that went stable for `POST /foo`. In this case, making a `POST /foo` could return a `405` unsupported method error and we would want to fallback to the unstable endpoint.

In the `/timestamp_to_event` case because only one HTTP method is supported, it's more obvious that a `405` would be an error in our code, but we can opt to leave the general pattern here ⏩

PR #2916, opened 2022-11-29
Fix synthesizeReceipt

gsouquet:
Fixes https://github.com/vector-im/element-web/issues/23827
Fixes https://github.com/vector-im/element-web/issues/23754
Fixes https://github.com/vector-im/element-web/issues/23847

This is still missing a couple of tests to ensure synthetic receipts are added when paginating through a thread timeline set. But otherwise should be ready for review

## Checklist

---

## 🐛 Bug Fixes
 * Fix synthesizeReceipt ([\#2916](https://github.com/matrix-org/matrix-js-sdk/pull/2916)). Fixes vector-im/element-web#23827 vector-im/element-web#23754 and vector-im/element-web#23847.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -5830,8 +5830,14 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             ).then(async (res) => {
                 const mapper = this.getEventMapper();
                 const matrixEvents = res.chunk.map(mapper);
-                for (const event of matrixEvents) {
-                    await eventTimeline.getTimelineSet()?.thread?.processEvent(event);
+
+                // Process latest events first
+                for (const event of matrixEvents.slice().reverse()) {
+                    await thread?.processEvent(event);
+                    const sender = event.getSender()!;
+                    if (!backwards || thread?.getEventReadUpTo(sender) === null) {
+                        room.addLocalEchoReceipt(sender, event, ReceiptType.Read);
+                    }
                 }
 
                 const newToken = res.next_batch;
--- a/src/client.ts
+++ b/src/client.ts
@@ -9325,6 +9325,19 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 }
 
+export function getCount({ type, room, event }: {
+    type: NotificationCountType; room: Room; event: MatrixEvent;
+}): number {
+    const isThreadEvent = !!event.threadRootId && !event.isThreadRoot;
+
+    return (isThreadEvent
+        ? room.getThreadUnreadNotificationCount(
+            event.threadRootId,
+            type,
+        )
+        : room.getRoomUnreadNotificationCount(type)) ?? 0;
+}
+
 /**
  * recalculates an accurate notifications count on event decryption.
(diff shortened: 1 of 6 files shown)


src/client.ts
```diff
@@ -9325,6 +9331,19 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 }
 
+export function getCount({ type, room, event }: {
```
t3chguy, line 9334: Could this be moved to somewhere more logical, `getCount` in `client.ts` implies it gets a count of clients/a client, which it doesn't

src/models/thread.ts
```diff
@@ -192,6 +192,18 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
     };
 
+    private onTimelineEvent = (
+        event: MatrixEvent,
+        room: Room | undefined,
+        toStartOfTimeline: boolean | undefined,
+    ): void => {
+        // Add a synthesized receipt to all live events added to the timeline
+        if (!toStartOfTimeline) {
+            room!.addLocalEchoReceipt(event.getSender()!, event, ReceiptType.Read);
```
t3chguy, line 202: Won't this possibly clobber a main room timeline RR given that `threadId` prefers threads and falls back to `MAIN_ROOM_TIMLINE`

src/client.ts
```diff
@@ -9325,6 +9331,19 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+    type: NotificationCountType; room: Room; event: MatrixEvent;
+}): number {
+    const isThreadEvent = !!event.threadRootId && !event.isThreadRoot;
+
+    return (isThreadEvent
+        ? room.getThreadUnreadNotificationCount(
+            event.threadRootId,
+            type,
+        )
```
t3chguy, line 9343:
```suggestion
        ? room.getThreadUnreadNotificationCount(event.threadRootId, type)
```

src/models/thread.ts
gsouquet: I'm not sure I follow the scenario you're talking about, nor what the risk would be?

t3chguy: that a RR the thread root in the main timeline will vanish as soon as it's sender sends a message into the thread and a synthetic RR is generated with a `thread_id` which isn't `MAIN_ROOM_TIMELINE`

src/models/thread.ts
```diff
@@ -192,6 +192,18 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
         }
     };
 
+    private onTimelineEvent = (
+        event: MatrixEvent,
+        room: Room | undefined,
+        toStartOfTimeline: boolean | undefined,
+    ): void => {
+        // Add a synthesized receipt to all live events added to the timeline
+        if (!toStartOfTimeline) {
```
t3chguy, line 201: I don't think this means it is live, forward paginating will be included too, while that is fine I think this comment is not

t3chguy (approved):
:shipit:

PR #2918, opened 2022-11-29
Update CODEOWNERS

t3chguy:
Take the dependency upgrades under the WAT

---


turt2live (approved):
I don't want to know why the Jest tests are failing on this.

PR #2920, opened 2022-11-30
Resume to-device message queue after resumed sync

Anderas:
Fixes https://github.com/matrix-org/element-web-rageshakes/issues/17170 and https://github.com/matrix-org/element-web-rageshakes/issues/17169

If there is any server connectivity issue whilst to-device message queue is processing batches of messages, the queue will be stopped and only resumed once further to-device messages are enqueued. This can lead to UISIs if the sender does not send any new keys immediately after connection is restored.

The solution is to observe client sync status and automatically resume the queue via `sendQueue` when we first recieve a `SyncState.Syncing`. If the queue is already empty, `sendQueue` will exit gracefully.

## Checklist

---

## 🐛 Bug Fixes
 * Resume to-device message queue after resumed sync ([\#2920](https://github.com/matrix-org/matrix-js-sdk/pull/2920)). Fixes matrix-org/element-web-rageshakes#17170. Contributed by @Anderas.

### DIFF
--- a/spec/unit/ToDeviceMessageQueue.spec.ts
+++ b/spec/unit/ToDeviceMessageQueue.spec.ts
@@ -1,0 +1,106 @@
+import { ConnectionError } from "../../src/http-api/errors";
+import { ClientEvent, MatrixClient, Store } from "../../src/client";
+import { ToDeviceMessageQueue } from "../../src/ToDeviceMessageQueue";
+import { getMockClientWithEventEmitter } from "../test-utils/client";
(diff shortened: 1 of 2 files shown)


gsouquet (changes requested):
Looks sensible to me!
With a couple of test to cover the new scenario, we should be good to go 👍

gsouquet (approved):
I think that makes sense! I tried dabbling with the timeout and believe you've landed in a good spot!

PR #2923, opened 2022-11-30
Move updated threads to the end of the thread list

justjanne:
Type: Defect
Fixes: https://github.com/vector-im/element-web/issues/23876

---

## 🐛 Bug Fixes
 * Move updated threads to the end of the thread list ([\#2923](https://github.com/matrix-org/matrix-js-sdk/pull/2923)). Fixes vector-im/element-web#23876. Contributed by @justjanne.

### DIFF
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -1843,7 +1843,7 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
     }
 
     private onThreadNewReply(thread: Thread): void {
-        this.updateThreadRootEvents(thread, false);
+        this.updateThreadRootEvents(thread, false, true);
     }
 
     private onThreadDelete(thread: Thread): void {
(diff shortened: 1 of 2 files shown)


gsouquet (approved):
Looks good to me, given there will be tests to assert the changes

src/models/room.ts
```diff
@@ -1968,11 +1968,11 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         ));
     }
 
-    private updateThreadRootEvents = (thread: Thread, toStartOfTimeline: boolean): void => {
+    private updateThreadRootEvents = (thread: Thread, toStartOfTimeline: boolean, recreateEvent: boolean): void => {
```
richvdh, line 1971: I can't help but wonder why some functions are defined with regular `private func(args) { .... }` method syntax, and others are defined as properties with `private func = (args) => {...}`. But I guess that's not new.

src/models/room.ts
```diff
@@ -1981,8 +1981,12 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         timelineSet: Optional<EventTimelineSet>,
         thread: Thread,
         toStartOfTimeline: boolean,
+        recreateEvent: boolean,
     ): void => {
         if (timelineSet && thread.rootEvent) {
+            if (recreateEvent) {
+                timelineSet.removeEvent(thread.id);
+            }
```
richvdh, line 1989: could this not be made unconditional? As I understand it, the only time that `updateThreadRootEvent` is called with `recreateEvent=False`, the event will not be in the timelineset anyway.

richvdh (commented):
It would be nice to add a comment somewhere saying *why* we recreate the event.

It's disappointing to see new code being added without new tests, but maybe it's unreasonable to demand a test.

src/models/room.ts
t3chguy: [Arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) have the benefit of being strongly bound to the context of the class. So even if you pass the method as a callback/handler it'll have the correct `this` context, unlike `function` syntax (without `.bind`

richvdh: makes sense I guess, thanks.

src/models/room.ts
justjanne: `createThread` is also called for every thread we load in the thread list. That means when scrolling through the thread list, it'll get called repeatedly, with old threads. In that case we want to avoid moving the event to the bottom of the list.

PR #2937, opened 2022-12-02
Fix highlight notifications increasing when total notification is zero

gsouquet:
Fixes https://github.com/vector-im/element-web/issues/23885

This was originally as an attempt to be more defensive and prevent highlight notification to appear for already read events! 
The synthetic read receipts should help

This clears the release blocker and will monitor the tracker

It also fixes an issue where `public get timeline(): MatrixEvent[]` would always return `undefined` because the `public abstract timeline: MatrixEvent[]` declaration in the parent was overriding it.
It created a lot of inconsistencies when trying to determine whether a message has been previously read or not!

## Checklist

---

## 🐛 Bug Fixes
 * Fix highlight notifications increasing when total notification is zero ([\#2937](https://github.com/matrix-org/matrix-js-sdk/pull/2937)). Fixes vector-im/element-web#23885.

### DIFF
--- a/spec/unit/notifications.spec.ts
+++ b/spec/unit/notifications.spec.ts
@@ -114,8 +114,8 @@ describe("fixNotificationCountOnDecryption", () => {
 
         fixNotificationCountOnDecryption(mockClient, event);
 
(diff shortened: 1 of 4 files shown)


t3chguy (approved):
Looks sane

PR #2938, opened 2022-12-02
Add a message ID on each to-device message

richvdh:
To make it easier to track down where to-device messages are getting lost,
add a custom property to each one, and log its value. Synapse will also log
this property.

We will emit log lines such as:

```
Enqueuing batch of to-device messages. type=m.room.encrypted txnid=m1670257675639.2 [ "@alice:localhost:8482/QSEDQLTFJA (msgid d0bd05e5-1fa3-410c-b1b5-c541ec99aa1b)",  "@bob:localhost:8482/SJGHSKJDH (msgid 0852cdaa-1546-4af7-87fb-07f3350be64f)" ]
```

and

```
received to-device m.room.encrypted from: @testuser82:localhost:8482 id: e354f925-a583-4b71-94d8-7864589ded29
```

part of https://github.com/vector-im/element-meta/issues/558

See also https://github.com/matrix-org/synapse/pull/14598 for the server side bits.

---

## ✨ Features
 * Add a message ID on each to-device message ([\#2938](https://github.com/matrix-org/matrix-js-sdk/pull/2938)).

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,6 @@
   "dependencies": {
     "@babel/runtime": "^7.12.5",
     "@types/sdp-transform": "^2.4.5",
-    "@types/uuid": "7",
     "another-json": "^0.2.0",
     "bs58": "^5.0.0",
     "content-type": "^1.0.4",
@@ -88,6 +87,7 @@
     "@types/domexception": "^4.0.0",
     "@types/jest": "^29.0.0",
     "@types/node": "18",
+    "@types/uuid": "7",
     "@typescript-eslint/eslint-plugin": "^5.6.0",
     "@typescript-eslint/parser": "^5.6.0",
     "allchange": "^1.0.6",
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
   "dependencies": {
     "@babel/runtime": "^7.12.5",
     "@types/sdp-transform": "^2.4.5",
+    "@types/uuid": "7",
     "another-json": "^0.2.0",
     "bs58": "^5.0.0",
     "content-type": "^1.0.4",
@@ -64,7 +65,8 @@
     "p-retry": "4",
     "qs": "^6.9.6",
     "sdp-transform": "^2.14.1",
-    "unhomoglyph": "^1.0.6"
+    "unhomoglyph": "^1.0.6",
+    "uuid": "7"
   },
   "devDependencies": {
     "@babel/cli": "^7.12.10",
--- a/spec/unit/crypto.spec.ts
+++ b/spec/unit/crypto.spec.ts
@@ -1010,18 +1010,24 @@ describe("Crypto", function() {
 
         it("encrypts and sends to devices", async () => {
             client.httpBackend
-                .when("PUT", "/sendToDevice/m.room.encrypted", {
-                    messages: {
-                        "@bob:example.org": {
-                            bobweb: encryptedPayload,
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -55,6 +55,7 @@
   "dependencies": {
     "@babel/runtime": "^7.12.5",
     "@types/sdp-transform": "^2.4.5",
+    "@types/uuid": "7",
```
t3chguy, line 58: This should go into dev deps `@types`

t3chguy: https://github.com/matrix-org/matrix-js-sdk/pull/2927 fixes it for the naughty stragglers

src/@types/event.ts
```diff
@@ -120,6 +120,8 @@ export enum RoomType {
     ElementVideo = "io.element.video",
 }
 
+export const ToDeviceMessageId = "org.matrix.msgid";
```
t3chguy, line 123: Is this specced anywhere to be using `org.matrix`? Doth it need be?

spec/unit/crypto.spec.ts
```diff
@@ -1010,18 +1010,24 @@ describe("Crypto", function() {
...
-                        "@bob:example.org": {
-                            bobweb: encryptedPayload,
-                            bobmobile: encryptedPayload,
-                        },
-                        "@carol:example.org": {
-                            caroldesktop: encryptedPayload,
+                .when("PUT", "/sendToDevice/m.room.encrypted")
+                .check((request) => {
+                    const data = request.data;
+                    delete data.messages["@bob:example.org"]["bobweb"]["org.matrix.msgid"];
```
robintown, line 1016: Alternatively you can use `"org.matrix.msgid": expect.any(String)` in the `expect` statement

robintown (commented):
Greatly appreciated, I do wonder though if there's a more foolproof way of doing this such that IDs don't have to be sprinkled throughout the SDK. Would it be possible to insert the ID in `MatrixClient.sendToDevice`?

src/@types/event.ts
richvdh: Well, no it's not, and I hope not. I'm trying to keep it out of the spec as a quick thing implementations can add if they so choose (but nobody should rely on it in any way).

t3chguy: I was just wondering if `org.matrix` had any special sentiment in the spec, given we do spec & use it - `org.matrix.custom.html` for example

richvdh:
> Would it be possible to insert the ID in MatrixClient.sendToDevice?

Well, it turns out that actually breaks some to-device messages (in particular, some of those used in verification), because it breaks their signatures, so I don't think that would be a good approach. Also, for messages that go via the queue, we want to assign IDs to messages *before* they are queued, to make sure we can trace them right through.

I do agree though, it feels very repetitive and boilerplatey. I considered some sort of `assignIDandSendToDevice` method, but decided that was probably more opaque and no more reliable than just generating an ID alongside the message.

src/@types/event.ts
richvdh: No, `org.matrix` is not special in the spec.

`org.matrix.custom.html` was *supposed* to be an "unstable development feature" and get turned into something else before speccing, but by the time that was on the cards it was way too late. It is literally the poster child for why I get cranky about using unstable prefixes in production.

robintown (approved):
In that case this is fine

PR #2940, opened 2022-12-02
Improve tsdoc types

t3chguy:
Requires https://github.com/matrix-org/matrix-js-sdk/pull/2895
Removed any comments which added no value e.g. `@param foo - ` as typedoc should generate those implicitly 

This fixes a lot of mis-associated things, e.g. https://pr2947--js-sdk-docs-previews.netlify.app/interfaces/_internal_.callopts attaching the Event onto CallOpts because it was floating above it - https://github.com/matrix-org/matrix-js-sdk/blob/8e6e7c8/src/webrtc/call.ts#L75

---


src/scheduler.ts
```diff
@@ -116,16 +103,52 @@ export class MatrixScheduler<T = ISendEventResponse> {
...
+         * `-1`. If this event was part of a queue, it will be removed from
+         * the queue.
+         * @param event - The event being retried.
+         * @param attempts - The number of failed attempts. This will always be \>= 1.
+         * @param err - The most recent error message received when trying
+         * to send this event.
+         * @returns The number of milliseconds to wait before trying again. If
+         * this is 0, the request will be immediately retried. If this is
+         * `-1`, the event will be marked as
+         * {@link EventStatus.NOT_SENT} and will not be retried.
```
richvdh, line 129: something weird has happened to the  documentation here: it seems to be showing the docs for `RETRY_BACKOFF_RATELIMIT`: https://pr2940--js-sdk-docs-previews.netlify.app/classes/matrixscheduler

richvdh (changes requested):
wow, this is great.

A few comments and thoughts.

Generally the EventHandlerMap docs either dont' work at all, or give unclear output. Could you look at them again?

spec/TestClient.ts
```diff
@@ -135,7 +135,7 @@ export class TestClient {
      * set up an expectation that the keys will be uploaded, and wait for
      * that to happen.
      *
-     * @returns {Promise} for the one-time keys
+     * @returns for the one-time keys
```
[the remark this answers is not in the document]
t3chguy: > Generally the EventHandlerMap docs either dont' work at all, or give unclear output. Could you look at them again?

They didn't work to begin with, and even marked unrelated things as Events - https://pr2947--js-sdk-docs-previews.netlify.app/interfaces/_internal_.callopts

I'd rather not have this rot too much

src/client.ts
```diff
@@ -916,13 +925,183 @@ export type EmittedEvents = ClientEvent
...
+     *       // update UI to remove any "Connection Lost" message
+     *       break;
+     *     case "PREPARED":
+     *       // the client instance is ready to be queried.
+     *       var rooms = matrixClient.getRooms();
+     *       break;
+     *   }
+     * });
+     * ```
+     */
```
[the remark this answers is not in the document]
t3chguy: Then linter won't like it because an enum value can't have `@param` fields

src/crypto/deviceinfo.ts
```diff
@@ -79,28 +54,34 @@ export class DeviceInfo {
...
+    // list of algorithms supported by this device
     public algorithms: string[] = [];
+    // a map from &lt;key type&gt;:&lt;id&gt; -> &lt;base64-encoded key&gt;>
     public keys: Record<string, string> = {};
+    // whether the device has been verified/blocked by the user
     public verified = DeviceVerification.Unverified;
+    // whether the user knows of this device's existence
+    // (useful when warning the user that a user has added new devices)
     public known = false;
+    // additional data from the homeserver
```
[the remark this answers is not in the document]
t3chguy: I fully expected that normal comments would work to, as they do for Jetbrains Intellisense

src/crypto/olmlib.ts
```diff
@@ -55,22 +53,20 @@ export const MEGOLM_ALGORITHM = Algorithm.Megolm;
 export const MEGOLM_BACKUP_ALGORITHM = Algorithm.MegolmBackup;
 
 export interface IOlmSessionResult {
+    // device info
```
t3chguy, line 56:
```suggestion
    /** device info */
```

src/crypto/olmlib.ts
```diff
@@ -55,22 +53,20 @@ export const MEGOLM_ALGORITHM = Algorithm.Megolm;
...
     device: DeviceInfo;
+    // base64 olm session id; null if no session could be established
```
t3chguy, line 58:
```suggestion
    /** base64 olm session id; null if no session could be established */
```

src/crypto/deviceinfo.ts
```diff
@@ -79,28 +54,34 @@ export class DeviceInfo {
...
-    /**
-     * @enum
-     */
     public static DeviceVerification = {
         VERIFIED: DeviceVerification.Verified,
         UNVERIFIED: DeviceVerification.Unverified,
         BLOCKED: DeviceVerification.Blocked,
     };
 
+    // list of algorithms supported by this device
```
t3chguy, line 63:
```suggestion
    /** list of algorithms supported by this device */
```

src/crypto/deviceinfo.ts
```diff
@@ -79,28 +54,34 @@ export class DeviceInfo {
...
     public algorithms: string[] = [];
+    // a map from &lt;key type&gt;:&lt;id&gt; -> &lt;base64-encoded key&gt;>
```
t3chguy, line 65:
```suggestion
    /** a map from &lt;key type&gt;:&lt;id&gt; -> &lt;base64-encoded key&gt;> */
```

src/crypto/deviceinfo.ts
```diff
@@ -79,28 +54,34 @@ export class DeviceInfo {
...
     public keys: Record<string, string> = {};
+    // whether the device has been verified/blocked by the user
```
t3chguy, line 67:
```suggestion
    /** whether the device has been verified/blocked by the user */
```

src/crypto/deviceinfo.ts
t3chguy, line 72:
```suggestion
    /** additional data from the homeserver */
```

src/crypto/api.ts
```diff
@@ -131,6 +130,7 @@ export interface IImportOpts {
 }
 
 export interface IImportRoomKeysOpts {
+    // called with an object that has a "stage" param
```
t3chguy, line 133:
```suggestion
    /** called with an object that has a "stage" param */
```

src/crypto/algorithms/base.ts
```diff
@@ -35,46 +33,36 @@ import { DeviceInfo } from "../deviceinfo";
...
     deviceId: string;
+    // crypto core
     crypto: Crypto;
+    // olm.js wrapper
     olmDevice: OlmDevice;
+    // base matrix api interface
     baseApis: MatrixClient;
+    // The ID of the room we will be sending to
     roomId?: string;
+    // The body of the m.room.encryption event
```
t3chguy, line 60:
```suggestion
    /** The UserID for the local user */
    userId: string;
    /** The identifier for this device. */
    deviceId: string;
    /** crypto core */
    crypto: Crypto;
    /** olm.js wrapper */
    olmDevice: OlmDevice;
    /** base matrix api interface */
    baseApis: MatrixClient;
    /** The ID of the room we will be sending to */
    roomId?: string;
    /** The body of the m.room.encryption event */
```

src/crypto/aes.ts
```diff
@@ -22,18 +22,21 @@ const zeroSalt = new Uint8Array(8);
 
 export interface IEncryptedPayload {
     [key: string]: any; // extensible
+    // the initialization vector in base64
     iv: string;
+    // the ciphertext in base64
     ciphertext: string;
+    // the HMAC in base64
```
t3chguy, line 29:
```suggestion
    /** the initialization vector in base64 */
    iv: string;
    /** the ciphertext in base64 */
    ciphertext: string;
    /** the HMAC in base64 */
```

src/crypto/OlmDevice.ts
```diff
@@ -53,43 +53,22 @@ function checkPayloadLength(payloadString: string): void {
...
     room_id: string; // eslint-disable-line camelcase
+    // pickled Olm.InboundGroupSession
     session: string;
     keysClaimed: Record<string, string>;
+    // Devices involved in forwarding this session to us (normally empty).
     forwardingCurve25519KeyChain: string[];
+    // whether this session is untrusted.
     untrusted?: boolean;
+    // whether this session exists during the room being set to shared history.
     sharedHistory?: boolean;
```
t3chguy, line 72:
```suggestion
/** data stored in the session store about an inbound group session */
export interface InboundGroupSessionData {
    room_id: string; // eslint-disable-line camelcase
    /** pickled Olm.InboundGroupSession */
    session: string;
    keysClaimed: Record<string, string>;
    /** Devices involved in forwarding this session to us (normally empty). */
    forwardingCurve25519KeyChain: string[];
    /** whether this session is untrusted. */
    untrusted?: boolean;
    /** whether this session exists during the room being set to shared history. */
    sharedHistory?: boolean;
```

src/crypto/OlmDevice.ts
```diff
@@ -134,20 +113,14 @@ type OneTimeKeys = { curve25519: { [keyId: string]: string } };
...
- * @property {string} deviceCurve25519Key   Curve25519 key for the account
- * @property {string} deviceEd25519Key      Ed25519 key for the account
  */
 export class OlmDevice {
     public pickleKey = "DEFAULT_KEY"; // set by consumers
 
     // don't know these until we load the account from storage in init()
+    // Curve25519 key for the account
     public deviceCurve25519Key: string | null = null;
+    // Ed25519 key for the account
```
t3chguy, line 123:
```suggestion
    /** Curve25519 key for the account, unknown until we load the account from storage in init() */
    public deviceCurve25519Key: string | null = null;
    /** Ed25519 key for the account, unknown until we load the account from storage in init() */
```

src/crypto/algorithms/megolm.ts
```diff
@@ -123,37 +121,27 @@ interface SharedWithData {
...
- *    devices with which we have shared the session key
- *        userId -> {deviceId -> SharedWithData}
+ * @internal
  */
 class OutboundSessionInfo {
+    // number of times this session has been used
     public useCount = 0;
+    // when the session was created (ms since the epoch)
     public creationTime: number;
+    // devices with which we have shared the session key `userId -> {deviceId -> SharedWithData}`
```
t3chguy, line 131:
```suggestion
    /** number of times this session has been used */
    public useCount = 0;
    /** when the session was created (ms since the epoch) */
    public creationTime: number;
    /** devices with which we have shared the session key `userId -> {deviceId -> SharedWithData}` */
```

src/crypto/index.ts
```diff
@@ -128,7 +124,12 @@ interface IInitOpts {
 }
 
 export interface IBootstrapCrossSigningOpts {
+    // Optional. Reset even if keys already exist.
```
t3chguy, line 127:
```suggestion
    /** Optional. Reset even if keys already exist. */
```

src/store/memory.ts
```diff
@@ -43,16 +42,10 @@ function isValidFilterId(filterId?: string | number | null): boolean {
 }
 
 export interface IOpts {
+    // The local storage instance to persist some forms of data such as tokens. Rooms will NOT be stored.
```
t3chguy, line 45:
```suggestion
    /** The local storage instance to persist some forms of data such as tokens. Rooms will NOT be stored. */
```

src/pushprocessor.ts
```diff
@@ -116,25 +112,27 @@ const DEFAULT_UNDERRIDE_RULES: IPushRule[] = [
 ];
 
 export interface IActionsObject {
+    // Whether this event should notify the user or not.
     notify: boolean;
+    // How this event should be notified.
```
t3chguy, line 117:
```suggestion
    /** Whether this event should notify the user or not. */
    notify: boolean;
    /** How this event should be notified. */
```

src/models/room-summary.ts
```diff
@@ -14,36 +14,30 @@ See the License for the specific language governing permissions and
...
+    // The title of the room (e.g. `m.room.name`)
     title: string;
+    // The description of the room (e.g. `m.room.topic`)
     desc?: string;
+    // The number of joined users.
     numMembers?: number;
+    // The list of aliases for this room.
     aliases?: string[];
+    // The timestamp for this room.
     timestamp?: number;
```
t3chguy, line 33:
```suggestion
    /** The title of the room (e.g. `m.room.name`) */
    title: string;
    /** The description of the room (e.g. `m.room.topic`) */
    desc?: string;
    /** The number of joined users. */
    numMembers?: number;
    /** The list of aliases for this room. */
    aliases?: string[];
    /** The timestamp for this room. */
    timestamp?: number;
```

src/models/user.ts
```diff
@@ -30,51 +26,101 @@ export enum UserEvent {
...
+     * We can subtract lastActiveAgo from this to approximate an absolute value for when a user was last active.
+     */
     public lastActiveAgo = 0;
+    // The time elapsed in ms since the user interacted proactively with the server, or we saw a message from the user
     public lastPresenceTs = 0;
+    // Whether we should consider lastActiveAgo to be an approximation and that the user should be seen as active 'now'
     public currentlyActive = false;
+    // The events describing this user.
     public events: {
+        // The m.presence event for this user.
```
t3chguy, line 116:
```suggestion
    /**
     * The 'displayname' of the user if known.
     * @privateRemarks
     * Should be read-only
     */
    public displayName?: string;
    public rawDisplayName?: string;
    /**
     * The 'avatar_url' of the user if known.
     * @privateRemarks
     * Should be read-only
     */
    public avatarUrl?: string;
    /**
     * The presence status message if known.
     * @privateRemarks
     * Should be read-only
     */
    public presenceStatusMsg?: string;
    /**
     * The presence enum if known.
     * @privateRemarks
     * Should be read-only
     */
    public presence = "offline";
    /**
     * Timestamp (ms since the epoch) for when we last received presence data for this user.
     * We can subtract lastActiveAgo from this to approximate an absolute value for when a user was last active.
     * @privateRemarks
     * Should be read-only
     */
    public lastActiveAgo = 0;
    /**
     * The time elapsed in ms since the user interacted proactively with the server,
     * or we saw a message from the user
     * @privateRemarks
     * Should be read-only
     */
    public lastPresenceTs = 0;
    /**
     * Whether we should consider lastActiveAgo to be an approximation
     * and that the user should be seen as active 'now'
     * @privateRemarks
     * Should be read-only
     */
    public currentlyActive = false;
    /**
     * The events describing this user.
     * @privateRemarks
     * Should be read-only
     */
    public events: {
        /** The m.presence event for this user. */
```

src/models/event.ts
```diff
@@ -272,11 +280,33 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
     public localTimestamp: number;
 
     // XXX: these should be read-only
+    /**
+     * The room member who sent this event, or null e.g.
+     * this is a presence event. This is only guaranteed to be set for events that
+     * appear in a timeline, ie. do not guarantee that it will be set on state
+     * events.
+     */
```
t3chguy, line 288:
```suggestion
    /**
     * The room member who sent this event, or null e.g.
     * this is a presence event. This is only guaranteed to be set for events that
     * appear in a timeline, ie. do not guarantee that it will be set on state
     * events.
     * @privateRemarks
     * Should be read-only
     */
```

src/models/event.ts
```diff
@@ -272,11 +280,33 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
...
+     * the invitee, the person being banned, etc.
+     */
     public target: RoomMember | null = null;
+    /**
+     * The sending status of the event.
+     */
     public status: EventStatus | null = null;
+    /**
+     * most recent error associated with sending the event, if any
+     */
```
t3chguy, line 301:
```suggestion
    /**
     * The room member who is the target of this event, e.g.
     * the invitee, the person being banned, etc.
     * @privateRemarks
     * Should be read-only
     */
    public target: RoomMember | null = null;
    /**
     * The sending status of the event.
     * @privateRemarks
     * Should be read-only
     */
    public status: EventStatus | null = null;
    /**
     * most recent error associated with sending the event, if any
     * @privateRemarks
     * Should be read-only
     */
```

src/models/event.ts
```diff
@@ -272,11 +280,33 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
...
     public error: MatrixError | null = null;
-    public forwardLooking = true; // only state events may be backwards looking
+    /**
+     * True if this event is 'forward looking', meaning
+     * that getDirectionalContent() will return event.content and not event.prev_content.
+     * Only state events may be backwards looking
+     * Default: true. <strong>This property is experimental and may change.</strong>
```
t3chguy, line 307:
```suggestion
     * Default: true. <strong>This property is experimental and may change.</strong>
     * @privateRemarks
     * Should be read-only
```

t3chguy:
If possible I'd like to avoid perfecting the types here, as long as we're making them better, it is a valuable PR but at huge risk of rot. Perfecting event emitter documentation is critical and deserves its own PR IMO.

spec/TestClient.ts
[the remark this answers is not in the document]
richvdh: yes fair, we can improve this later

src/client.ts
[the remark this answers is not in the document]
richvdh: Well, I still think that's where they belong; we might just have to find another way to format the params. But we can do this on a later pass.

src/http-api/fetch.ts
```diff
@@ -104,35 +103,29 @@ export class FetchHttpApi<O extends IHttpOpts> {
...
-     * If not specified, uses this.opts.baseUrl
-     *
-     * @param {Object=} opts.headers map of additional request headers
-     *
-     * @return {Promise} Resolves to <code>{data: {Object},
-     * headers: {Object}, code: {Number}}</code>.
-     * If <code>onlyData</code> is set, this will resolve to the <code>data</code>
-     * object only.
-     * @return {module:http-api.MatrixError} Rejects with an error if a problem
-     * occurred. This includes network problems and Matrix-specific error JSON.
```
MadLittleMods, line 135: Do we really want to get rid of the reject errors? Those aren't implicit by the return type so I think it would be nice to keep this information around

MadLittleMods: This was discussed [today in the Web Chapter Sync](https://docs.google.com/document/d/11-rd31tZxCvtp-8hfGj3sKlAE1m1r9L1n8aklR5ZLqU/edit#bookmark=id.tn0uhbinedyj). The general consensus was that this type of information is valuable to keep around 👍. We agreed that we will try to document what can be thrown from a function, but we won’t claim that we always document everything that is thrown.

Unclear if we should use `@return` or `@throws` 🤷 and I'm not that familiar with JSDoc to know for sure

PR #2941, opened 2022-12-04
Don't expose `calls` on `GroupCall`

SimonBrandner:
---

### DIFF
--- a/spec/unit/webrtc/groupCall.spec.ts
+++ b/spec/unit/webrtc/groupCall.spec.ts
@@ -479,9 +479,10 @@ describe('Group Call', function() {
 
             it("sends metadata updates before unmuting in PTT mode", async () => {
                 const mockCall = new MockMatrixCall(FAKE_ROOM_ID, groupCall.groupCallId);
-                (groupCall as any).calls.set(
+                // @ts-ignore
+                groupCall.calls.set(
                     mockCall.getOpponentMember() as RoomMember,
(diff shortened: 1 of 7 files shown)


spec/unit/webrtc/groupCall.spec.ts
```diff
@@ -698,7 +662,7 @@ describe('Group Call', function() {
 
                 expect(client1.sendToDevice).toHaveBeenCalled();
 
-                const oldCall = groupCall1.calls.get(
+                const oldCall = (groupCall1 as any).calls.get(
```
robintown, line 665: When accessing private fields we should always prefer a `ts-ignore` comment over casting to `any`, since `ts-ignore` preserves type information

src/webrtc/callFeed.ts
```diff
@@ -101,6 +111,21 @@ export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap>
...
+    }
+
+    public get connected(): boolean {
+        // Local feeds are always considered connected
+        return this.isLocal() || this._connected;
+    }
+
+    private set connected(connected: boolean) {
+        this._connected = connected;
+        this.emit(CallFeedEvent.ConnectedChanged, connected);
```
robintown, line 128: This may emit the wrong value for local feeds

robintown (approved):

PR #2943, opened 2022-12-05
Upload device keys during initCrypto

richvdh:
Rather than waiting for the application to call `.startClient`, upload the
device keys during `initCrypto()`. Element-R is going to approach this slightly
differently (the rust-sdk wants to manage the decision on key uploads itself),
so this lays some groundwork by collecting the libolm-specific bits together.

---


richvdh:
Turns out I already have a PR open for this :facepalm: : https://github.com/matrix-org/matrix-js-sdk/pull/2872

PR #2946, opened 2022-12-07
Minor VoIP stack improvements

SimonBrandner:
---

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -263,7 +263,11 @@ const VOIP_PROTO_VERSION = "1";
 const FALLBACK_ICE_SERVER = 'stun:turn.matrix.org';
 
 /** The length of time a call can be ringing for. */
-const CALL_TIMEOUT_MS = 60000;
+const CALL_TIMEOUT_MS = 60 * 1000; // ms
+/** The time after which we increment callLength */
+const CALL_LENGTH_INTERVAL = 1000; // ms
+/** The time after which we end the call, if ICE got disconnected */
+const ICE_DISCONNECTED_TIMEOUT = 30 * 1000; // ms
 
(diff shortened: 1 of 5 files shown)


src/webrtc/call.ts
```diff
@@ -2106,7 +2110,7 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
                 this.callLengthInterval = setInterval(() => {
                     this.callLength++;
                     this.emit(CallEvent.LengthChanged, this.callLength);
-                }, 1000);
+                }, CALL_LENGTH_INTERVAL);
```
dbkr, line 2113: Hmm, this feels a bit weird and is essentially defining the number of milliseconds in a second. Also, a setInterval won't execute after exactly the amount of time specified (especially when browsers do funny things with suspending timer execution in background pages), so this might have a tendency to drift less accurate over time.

I'd probably just store the timestamp when the call started and recalculate the length based on `now - startTime` each time, then this constant really is just the update interval.

SimonBrandner: Good point!

src/webrtc/call.ts
```diff
@@ -401,7 +401,7 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
     private remoteSDPStreamMetadata?: SDPStreamMetadata;
 
     private callLengthInterval?: ReturnType<typeof setInterval>;
-    private callLength = 0;
+    private callLStartTime?: number;
```
dbkr, line 404: Extra 'L'?

dbkr (commented):
Otherwise lgtm - thanks for the fix

dbkr (approved):

PR #2947, opened 2022-12-07
Tweak how the SonarCloud scan fails

t3chguy:
---


turt2live (approved):
I'm a bit worried github will take this to mean that the job never fails, but let's find out :D

t3chguy:
The main check *can* fail, its the 2nd-ary check that it creates on the associated commit that is manually managed in `Sibz/github-status-action@v1` that we actually protect using

PR #2950, opened 2022-12-07
Expose a new 'userHasCrossSigningKeys' method

richvdh:
... to save clients digging into the guts of `crypto`

---


turt2live (commented):
clearing code review while draft

PR #2951, opened 2022-12-07
Fix #23916: Prevent edits of the last message in a thread getting lost

justjanne:
Type: Defect
Fixes: https://github.com/vector-im/element-web/issues/23916
Fixes: https://github.com/vector-im/element-web/issues/23942
Requires: https://github.com/matrix-org/matrix-js-sdk/pull/2965

Prevent confusion between synthesized lastEvent and the same event in the actual timeline

---

## 🐛 Bug Fixes
 * Fix #23916: Prevent edits of the last message in a thread getting lost ([\#2951](https://github.com/matrix-org/matrix-js-sdk/pull/2951)). Fixes vector-im/element-web#23916 and vector-im/element-web#23942. Contributed by @justjanne.

### DIFF
--- a/spec/integ/matrix-client-event-timeline.spec.ts
+++ b/spec/integ/matrix-client-event-timeline.spec.ts
@@ -27,12 +27,11 @@ import {
     MatrixEvent,
     PendingEventOrdering,
     Room,
-    RoomEvent,
 } from "../../src/matrix";
 import { logger } from "../../src/logger";
 import { encodeParams, encodeUri, QueryDict, replaceParam } from "../../src/utils";
 import { TestClient } from "../TestClient";
-import { FeatureSupport, Thread, THREAD_RELATION_TYPE } from "../../src/models/thread";
+import { FeatureSupport, Thread, THREAD_RELATION_TYPE, ThreadEvent } from "../../src/models/thread";
 import { emitPromise } from "../test-utils/test-utils";
 
 const userId = "@alice:localhost";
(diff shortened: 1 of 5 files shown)


gsouquet (commented):
clearing the review request. Request another one when it's ready 👍

src/models/thread.ts
```diff
@@ -317,31 +322,33 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
             this.replyCount = bundledRelationship.count;
             this._currentUserParticipated = !!bundledRelationship.current_user_participated;
 
-            const mapper = this.client.getEventMapper();
             // re-insert roomId
-            this.lastEvent = mapper({
+            this.lastEvent = new MatrixEvent({
```
gsouquet, line 326: Why not using the `mapper` ? It's actually setting some event re-emitting stuff that are quite important, like dispatching `MatrixEventEvent.Decrypted`, ...

src/models/thread.ts
```diff
@@ -317,31 +322,33 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
-        }
+    private updatePendingReplyCount(): void {
+        const unfilteredPendingEvents =
+            this.pendingEventOrdering === PendingEventOrdering.Detached ? this.room.getPendingEvents() : this.events;
+        const pendingEvents = unfilteredPendingEvents.filter(
+            (ev) =>
+                ev.threadRootId === this.id &&
+                ev.isRelation(THREAD_RELATION_TYPE.name) &&
+                ev.status !== null &&
+                ev.getId() !== this.lastEvent?.getId(),
```
gsouquet, line 343: This looks awfully similar to what is in `onEcho`, any chance to consolidate the two under one roof?

gsouquet (approved):
Otherwise looks good to me!

src/models/thread.ts
justjanne: This is something I’ve been struggling a bit with. We need those events where possible (or make sure no one uses this as if it was a full, actual event), but at the same time we need to make sure we're not leaking memory. And as techguy told me, if we just create events via the mapper but don't clean them up, we're gonna leak quite a bit of memory.

src/models/thread.ts
justjanne: They do look similar, but so far they're not similar enough to easily unify them :/

duxovni:
This seems to have broken [this matrix-react-sdk test](https://github.com/matrix-org/matrix-react-sdk/blob/develop/test/components/views/rooms/EventTile-test.tsx#L91). It looks like after the redaction redacts its parent, we can no longer tell that the redaction belongs to the thread, which messes up some logic down the line.

duxovni:
nvm, the issue was on the react-sdk side :slightly_smiling_face:

spec/integ/matrix-client-event-timeline.spec.ts
```diff
@@ -1692,27 +1652,11 @@ describe("MatrixClient event timelines", function () {
...
-                        buildParams(Direction.Forward, "end_token"),
-                )
-                .respond(200, function () {
-                    return {
-                        original_event: THREAD_ROOT,
-                        chunk: [THREAD_REPLY],
+                        chunk: [THREAD_ROOT],
                     };
                 });
             const timeline = await flushHttp(client.getEventTimeline(timelineSet, THREAD_ROOT.event_id!));
```
MadLittleMods, line 1662: Where is the test coverage for the issues that this claims to fix?

 - https://github.com/vector-im/element-web/issues/23942
 - https://github.com/vector-im/element-web/issues/23916
 
I don't see it amongst these random edits here.

PR #2952, opened 2022-12-07
Fix `examples/browser`

richvdh:
This seems to have been broken for ages

---


weeman1337 (approved):
Thanks for fixing these :+1:

PR #2954, opened 2022-12-08
Implement MSC3912: Relation-based redactions

weeman1337:
SDK support for [MSC3912](https://github.com/matrix-org/matrix-spec-proposals/pull/3912)

Should be reviewed commitwise, because it contains a heavy refactoring of the client spec.

## Checklist

---

## ✨ Features
 * Implement MSC3912: Relation-based redactions ([\#2954](https://github.com/matrix-org/matrix-js-sdk/pull/2954)).

### DIFF
--- a/src/@types/requests.ts
+++ b/src/@types/requests.ts
@@ -21,7 +21,7 @@ import { IRoomEventFilter } from "../filter";
 import { Direction } from "../models/event-timeline";
 import { PushRuleAction } from "./PushRules";
 import { IRoomEvent } from "../sync-accumulator";
-import { EventType, RoomType } from "./event";
+import { EventType, RelationType, RoomType } from "./event";
 
 // allow camelcase as these are things that go onto the wire
 /* eslint-disable camelcase */
@@ -47,6 +47,8 @@ export interface IJoinRoomOpts {
 
 export interface IRedactOpts {
     reason?: string;
+    // also delete related events, if the server supports it (MSC3912)
+    with_relations?: Array<RelationType | string>;
 }
 
 export interface ISendEventResponse {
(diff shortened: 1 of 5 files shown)


gsouquet (commented):
Changes related to MSC3912 make sense

Glancing at the test changes, seem to make relative sense, but it would be interesting to have a better description of what the test changes are actually doing given the very large diff.

weeman1337:
> Glancing at the test changes, seem to make relative sense, but it would be interesting to have a better description of what the test changes are actually doing given the very large diff.

The test did set up one specific client. I just moved the current client setup into a `describe` block. This way you can add more blocks with other initial configurations. Alternatively it could be a separate file for reach client setup, because the client test is already huge. Would also include some refactoring in form of extraction of some test utilities.

weeman1337:
@gsouquet did a force push because everything changed based on the feedback or your previous review. No crazy client refactoring any more :slightly_smiling_face: . Also using `features.ts` like you suggested :+1:

gsouquet (approved):

src/@types/requests.ts
```diff
@@ -47,7 +47,13 @@ export interface IJoinRoomOpts {
 
 export interface IRedactOpts {
     reason?: string;
-    // also delete related events, if the server supports it (MSC3912)
+    /**
+     * Also try to delete related events of the relation types in this array.
```
richvdh, line 51: I think we should be more explicit about what this actually means:

```suggestion
     * Whether events related to the redacted event should be redacted.
     *
     * If specified, then any events which relate to the event being redacted with
     * any of the relationship types listed will also be redacted.
     *
```

richvdh (approved):

src/@types/requests.ts
weeman1337: Thanks for your suggestion :+1: Updated accordingly

PR #2955, opened 2022-12-08
Begin factoring out a CryptoBackend interface

richvdh:
Part of https://github.com/vector-im/element-web/issues/21972. Eventually I want to replace the whole of the current `Crypto` implementation with an alternative implementation, but in order to get from here to there, I'm factoring out a common interface which will be implemented by both implementations.

I'm also determined to fix the problem where the innards of the crypto implementation are exposed to applications via the `MatrixClient.crypto` property.

It's not (yet) entirely clear what shape this interface should be, so I'm going with a minimal approach and adding things as we know we need them. This means that we need to keep the old `client.crypto` property around as well as a new `client.cryptoBackend` property. Eventually `client.crypto` will go away, but that will be a breaking change in the js-sdk.

~~Currently based on #2958~~

---

### DIFF
--- a/spec/integ/megolm-integ.spec.ts
+++ b/spec/integ/megolm-integ.spec.ts
@@ -650,6 +650,70 @@ describe("megolm", () => {
         ]);
     });
 
+    describe("get|setGlobalBlacklistUnverifiedDevices", () => {
+        it("should raise an error if crypto is disabled", () => {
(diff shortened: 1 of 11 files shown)


dbkr (approved):

richvdh:
I've rebased this on develop now that #2958 has merged.

PR #2961, opened 2022-12-09
Exclude generated files from prettier

richvdh:
The example directories include symlinks to the generated matrix.js, which we
should not prettify.

---


gsouquet (approved):
lgtm

PR #2965, opened 2022-12-12
Fix #23919: Root message for new thread loaded from network

justjanne:
Type: Defect
Fixes: https://github.com/vector-im/element-web/issues/23919

---

## 🐛 Bug Fixes
 * Fix #23919: Root message for new thread loaded from network ([\#2965](https://github.com/matrix-org/matrix-js-sdk/pull/2965)). Fixes vector-im/element-web#23919. Contributed by @justjanne.

### DIFF
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -20,7 +20,7 @@ import { MatrixClient, PendingEventOrdering } from "../client";
 import { TypedReEmitter } from "../ReEmitter";
 import { RelationType } from "../@types/event";
 import { EventStatus, IThreadBundledRelationship, MatrixEvent, MatrixEventEvent } from "./event";
-import { EventTimeline } from "./event-timeline";
+import { Direction, EventTimeline } from "./event-timeline";
 import { EventTimelineSet, EventTimelineSetHandlerMap } from "./event-timeline-set";
 import { NotificationCountType, Room, RoomEvent } from "./room";
 import { RoomState } from "./room-state";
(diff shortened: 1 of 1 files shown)


gsouquet (approved):
Looks good to me!

src/models/thread.ts
```diff
@@ -358,7 +358,15 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
-                await this.client.paginateEventTimeline(this.liveTimeline, { backwards: true, limit: 1 });
+                if (this.replyCount === 0 && this.rootEvent) {
+                    this.timelineSet.addEventsToTimeline([this.rootEvent], true, this.liveTimeline, null);
+                    this.liveTimeline.setPaginationToken(null, Direction.Backward);
+                } else {
+                    await this.client.paginateEventTimeline(this.liveTimeline, {
+                        backwards: true,
+                        limit: Math.max(1, this.length),
+                    });
+                }
```
MadLittleMods, line 369: What was wrong before that caused https://github.com/vector-im/element-web/issues/23934?

What changed so https://github.com/vector-im/element-web/issues/23934 now works?

If there is some tricky order of operations here, we should explain why/what, etc

MadLittleMods: @justjanne ^

justjanne: @MadLittleMods I'm currently on vacation, I'll address it in January.

PR #2966, opened 2022-12-12
Fix issue where threads weren't correctly removed from My Threads tab

justjanne:
Type: Defect

---

## 🐛 Bug Fixes
 * Fix issue where threads weren't correctly removed from My Threads tab ([\#2966](https://github.com/matrix-org/matrix-js-sdk/pull/2966)). Contributed by @justjanne.


florianduros:
Can we add a non-regression test for this case ? It will be great! Thanks

justjanne:
This PR requires more work than expected, I'll handle it after the RC

turt2live (commented):
clearing review while in draft

PR #2967, opened 2022-12-12
Support MSC3391: Account data deletion

kerryarchibald:
MSC: https://github.com/ShadowJonathan/matrix-doc/blob/account-data-delete/proposals/3391-account-data-delete.md

## Checklist

---

## ✨ Features
 * Support MSC3391: Account data deletion ([\#2967](https://github.com/matrix-org/matrix-js-sdk/pull/2967)). Contributed by @kerryarchibald.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -3765,10 +3765,9 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 
     public async deleteAccountData(eventType: string): Promise<void> {
-        const supportsMsc3391DeleteAccountData =
-            this.canSupport.get(Feature.AccountDataDeletion) !== ServerSupport.Unsupported;
+        const msc3391DeleteAccountDataServerSupport = this.canSupport.get(Feature.AccountDataDeletion);
         // if deletion is not supported overwrite with empty content
-        if (!supportsMsc3391DeleteAccountData) {
+        if (msc3391DeleteAccountDataServerSupport === ServerSupport.Unsupported) {
             await this.setAccountData(eventType, {});
             return;
         }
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -3764,6 +3764,23 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+    public async deleteAccountData(eventType: string): Promise<void> {
+        const supportsMsc3391DeleteAccountData =
+            this.canSupport.get(Feature.AccountDataDeletion) !== ServerSupport.Unsupported;
+        // if deletion is not supported overwrite with empty content
+        if (!supportsMsc3391DeleteAccountData) {
+            await this.setAccountData(eventType, {});
+            return;
+        }
+        const path = utils.encodeUri("/user/$userId/account_data/$type", {
+            $userId: this.credentials.userId!,
```
weeman1337, line 3776: This is a public method and can be called when no user is logged in. Means the `!` is probably incorrect here. We should handle the case when it is null.

src/client.ts
```diff
@@ -3764,6 +3764,23 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            $type: eventType,
+        });
+        return await this.http.authedRequest(Method.Delete, path, undefined, undefined, {
```
weeman1337, line 3779: Not sure about the unstable process: But shouldn't we here not call the endpoint depending on the `this.canSupport.get(Feature.AccountDataDeletion)` stable/unstable value?

weeman1337 (commented):
Some details questions. Looks good otherwise :+1:

src/client.ts
kerryarchibald: True, I just copied this part from the `setAccountData` method, which I assume relies on `authedRequest` to fail nicely when there is no logged in user. Added `getSafeUserId` to use here

weeman1337 (approved):
Looks good :rocket:

PR #2969, opened 2022-12-12
Introduce a mechanism for using the rust-crypto-sdk

richvdh:
This PR introduces `MatrixClient.initRustCrypto`, which is similar to `initCrypto`, except that it will use the Rust crypto SDK instead of the old libolm-based implementation.

This is very much *not* something you want to use in production code right now, because the integration with the rust sdk is extremely skeletal and almost everything crypto-related will raise an exception rather than doing anything useful.

It is, however, enough to demonstrate the loading of the wasmified rust sdk in element web, and a react sdk with light modifications can successfully log in and out.

Element-web notes: none

---

## ✨ Features
 * Introduce a mechanism for using the rust-crypto-sdk ([\#2969](https://github.com/matrix-org/matrix-js-sdk/pull/2969)).

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2091,6 +2091,8 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * *WARNING*: this API is very experimental, should not be used in production, and may change without notice!
      *    Eventually it will be deprecated and `initCrypto` will do the same thing.
      *
+     * @experimental
+     *
      * @returns a Promise which will resolve when the crypto layer has been
      *    successfully initialised.
      */
--- a/src/client.ts
+++ b/src/client.ts
@@ -211,6 +211,7 @@ import { LocalNotificationSettings } from "./@types/local_notifications";
 import { UNREAD_THREAD_NOTIFICATIONS } from "./@types/sync";
 import { buildFeatureSupportMap, Feature, ServerSupport } from "./feature";
 import { CryptoBackend } from "./common-crypto/CryptoBackend";
+import { RUST_SDK_STORE_PREFIX } from "./rust-crypto/constants";
 
 export type Store = IStore;
 
@@ -1667,7 +1668,10 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
                 // No indexeddb support
                 return;
             }
-            for (const dbname of ["matrix-js-sdk::matrix-sdk-crypto", "matrix-js-sdk::matrix-sdk-crypto-meta"]) {
+            for (const dbname of [
+                `${RUST_SDK_STORE_PREFIX}::matrix-sdk-crypto`,
+                `${RUST_SDK_STORE_PREFIX}::matrix-sdk-crypto-meta`,
+            ]) {
                 const prom = new Promise((resolve, reject) => {
                     logger.info(`Removing IndexedDB instance ${dbname}`);
                     const req = indexedDB.deleteDatabase(dbname);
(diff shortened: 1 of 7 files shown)


src/client.ts
```diff
@@ -2051,6 +2083,44 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         });
     }
 
+    /**
+     * Initialise support for end-to-end encryption in this client, using the rust matrix-sdk-crypto.
+     *
+     * An alternative to {@link initCrypto}.
+     *
+     * *WARNING*: this API is very experimental and should not yet be used in production code! Eventually
+     *    it will be deprecated and `initCrypto` will do the same thing.
```
turt2live, line 2092:
```suggestion
     * *WARNING*: this API is very experimental, should not be used in production, and may change without notice!
     *    Eventually it will be deprecated and `initCrypto` will do the same thing.
```

just an added note that we're not going to declare API changes as breaking, as some people might assume experimental==safe enough to use (unfortunately).

src/rust-crypto/rust-crypto.ts
```diff
@@ -0,0 +1,60 @@
...
+import * as RustSdkCryptoJs from "@matrix-org/matrix-sdk-crypto-js";
+
+import { IEventDecryptionResult } from "../@types/crypto";
+import { MatrixEvent } from "../models/event";
+import { CryptoBackend } from "../common-crypto/CryptoBackend";
+
+// import { logger } from "../logger";
+
+/**
+ * An implementation of {@link CryptoBackend} using the Rust matrix-sdk-crypto.
```
turt2live, line 26:
```suggestion
 * An implementation of {@link CryptoBackend} using the Rust matrix-sdk-crypto.
 * @experimental This entire class is subject to breaking changes, without notice.
```

or whatever jsdoc wants to say this.

turt2live: alternatively, apparently `@internal` will exclude it from public types, despite being public itself.

turt2live (commented):
The changes overall lgtm, with minor nits for declaring things breaking for overly clear communication. The `yarn.lock` is also so massive it broke something, and I'm concerned there's a package manager mismatch somewhere: I've sent you a DM to try and figure this out OOB, though is not necessarily a blocker for this PR landing.

src/client.ts
```diff
@@ -1657,6 +1657,38 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+        // delete the stores used by the rust matrix-sdk-crypto, in case they were used
+        const deleteRustSdkStore = async (): Promise<void> => {
+            let indexedDB: IDBFactory;
+            try {
+                indexedDB = global.indexedDB;
+            } catch (e) {
+                // No indexeddb support
+                return;
+            }
+            for (const dbname of ["matrix-js-sdk::matrix-sdk-crypto", "matrix-js-sdk::matrix-sdk-crypto-meta"]) {
```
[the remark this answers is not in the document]
richvdh: not really, tbh: they are defined in the rust-sdk [here](https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk-indexeddb/src/crypto_store.rs#L145) and [here](https://github.com/matrix-org/matrix-rust-sdk/blob/main/crates/matrix-sdk-indexeddb/src/crypto_store.rs#L237).

> alternatively, we should consider making them constants for reasons of reliability.

Given they are only used once, I'm unconvinced that a constant will help reliability.

The prefix does come from the js-sdk so I can factor that out to a constant.

src/rust-crypto/rust-crypto.ts
richvdh: Well, it's not exported explicitly, and it's only dynamically referenced, so it doesn't actually appear in the generated documentation.

src/client.ts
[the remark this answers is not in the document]
richvdh: > The prefix does come from the js-sdk so I can factor that out to a constant.

done in `df8d44b`.

turt2live (approved):
thanks!

PR #2970, opened 2022-12-12
Add `exportRoomKeys` to `CryptoBackend`

richvdh:
Element-web calls `exportRoomKeys` on logout, so we need a stub implementation to get it EW working with the rust crypto sdk.

---


weeman1337 (approved):
Looks good :+1:

PR #2971, opened 2022-12-13
Fix messages loaded during initial fetch ending up out of order

justjanne:
Type: Defect
Fixes: https://github.com/vector-im/element-web/issues/23972

### DIFF
--- a/spec/integ/matrix-client-event-timeline.spec.ts
+++ b/spec/integ/matrix-client-event-timeline.spec.ts
@@ -1249,6 +1249,7 @@ describe("MatrixClient event timelines", function () {
                     },
                     event: true,
                 });
+                THREAD_REPLY2.localTimestamp += 1000;
 
                 // Test data for the first thread, with the second reply
                 const THREAD_ROOT_UPDATED = {
@@ -1305,6 +1306,7 @@ describe("MatrixClient event timelines", function () {
 
                 // Test adding a second event to the first thread
                 const thread = room.getThread(THREAD_ROOT.event_id!)!;
+                thread.initialEventsFetched = true;
                 const prom = emitPromise(room, ThreadEvent.NewReply);
                 respondToEvent(THREAD_ROOT_UPDATED);
                 respondToEvent(THREAD_ROOT_UPDATED);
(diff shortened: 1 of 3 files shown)


spec/unit/room.spec.ts
```diff
@@ -2518,7 +2533,11 @@ describe("Room", function () {
             prom = emitPromise(room, ThreadEvent.Update);
             room.addLiveEvents([threadResponseEdit]);
             await prom;
+            /*
```
andybalaam, line 2536: Commenting out an assertion?

spec/integ/matrix-client-event-timeline.spec.ts
```diff
@@ -1624,10 +1626,35 @@ describe("MatrixClient event timelines", function () {
                     },
                 },
             });
+            httpBackend
```
andybalaam, line 1629: Please could you explain to me how these changes to the test check we are now behaving correctly?

I was expecting to see a test which failed before because events were out of order, but now passes.

andybalaam (changes requested):
I love that this is a one-line fix!

One commented assertion, and a question.

spec/integ/matrix-client-event-timeline.spec.ts
justjanne: The issue with that is that the tests often omit certain parts of the way threads are loaded, to reduce complexity. Otherwise our tests would be much larger than the are. Ideally, each of these tests only tests the part of the functionality that it's concerned with, and mocks the rest.

But as this PR changes a minor part of the original functionality, the way tests mock the loading has to be adapted.

There's no single test that can easily test this issue yet, because the way we mock the loading can mask the issue itself.

andybalaam: We talked over Matrix and agreed that @justjanne will work on a test that failed before and passes now.

The new test will either be part of this PR, or a new one as soon as possible afterwards.

andybalaam (approved):
Approving so long as a new test comes soon afterwards.

Thanks!

PR #2982, opened 2022-12-14
Stop requesting room keys from other users

poljar:
Requesting forwarded room keys from other users is quite pointless since we don't accept them. This mainly produces unnecessary to-device traffic.

The main commit in this PR is the third one, where we finally remove the requesting of room keys from other users. The previous commits are refactors of the forwarded room key handling and unit tests validating the forwarded room key handling.

Fixes: https://github.com/vector-im/element-web/issues/23766
Fixes: https://github.com/vector-im/element-web/issues/24681

---


src/crypto/algorithms/megolm.ts
```diff
@@ -1419,154 +1438,318 @@ export class MegolmDecryption extends DecryptionAlgorithm {
...
+        roomKey.exportFormat = true;
+        roomKey.forwardingKeyChain = forwardingKeyChain;
+        // forwarded keys are always untrusted
+        roomKey.extraSessionData.untrusted = true;
+
+        return roomKey;
+    }
+
+    /**
+     * Should we accept the forwarded room key that was found in the given
```
richvdh, line 1532:
```suggestion
     * Determine if we should accept the forwarded room key that was found in the given
```

src/crypto/algorithms/megolm.ts
```diff
@@ -1419,154 +1438,318 @@ export class MegolmDecryption extends DecryptionAlgorithm {
...
-        if (content["org.matrix.msc3061.shared_history"]) {
-            extraSessionData.sharedHistory = true;
+    private wasRoomKeyForwardedAsHistory(roomKey: RoomKey): boolean {
+        const room = this.baseApis.getRoom(roomKey.roomId);
+
+        if (room && roomKey.extraSessionData.sharedHistory) {
+            return true;
+        } else {
+            return false;
+        }
```
richvdh, line 1629: making this dependent on `room` feels surprising, given its name. I'm also a bit unclear why it is needed: `wasRoomKeyForwardedByInviter` will return false for an unknown room, so `(fromInviter && sharedAsHistory)` already checks for a known room via `fromInviter`. Why not set `sharedAsHistory = roomKey.extraSessionData.sharedHistory` ?

src/crypto/algorithms/megolm.ts
```diff
@@ -1419,154 +1438,318 @@ export class MegolmDecryption extends DecryptionAlgorithm {
...
-            };
+        // We received this room key from event.getSenderKey(), but the original
+        // creator of the room key is claimed in the content. Let's try to fetch
+        // those keys now.
+        const claimedCurve25519Key = content.sender_key;
+        const claimedEd25519Key = content.sender_claimed_ed25519_key;
 
-            // If this is a key for a room we're not in, don't load it
-            // yet, just park it in case *this sender* invites us to
-            // that room later
```
richvdh, line 1495: this comment was somewhat useful and has been lost. Reinstate it in `shouldParkForwardedKey` ?

richvdh (changes requested):
looks generally good to me, modulo a few suggestions/nits as below.

The logic in `shouldAcceptForwardedKey` seems a little bit magical and I'm not entirely convinced we have comprehensive tests for it: it might be nice to make sure we do. But I'll not insist if you need to move on to other things.

src/crypto/algorithms/megolm.ts
```diff
@@ -1419,154 +1438,318 @@ export class MegolmDecryption extends DecryptionAlgorithm {
...
+     * `m.forwarded_room_key` event.
+     *
+     * Forwarded room keys need special handling since we have no way of knowing
+     * who the original creator of the room key was. This naturally means that
+     * forwarded room keys are always untrusted and should only be accepted in
+     * some cases.
+     *
+     * @param event - An `m.forwarded_room_key` event.
+     *
+     * @internal
```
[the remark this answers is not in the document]
poljar: Hmm, other private methods, not written by me, seem to have the `@internal` marker as well. Given this I'm inclined to leave them in, unless you insist of course.

src/crypto/algorithms/megolm.ts
poljar: I'm not sure that it needs to check for the room either but I don't think that because some other method, that coincidentally gets called alongside this one, checks for a room is a particularly good reason to remove the check.

I think what we should be asking us is the following: "Do we want to consider forwarded room keys shared as history if we're not part of the room".

This flag is controlled by the user that does the forwarding, in the spirit of minimizing the amount of forwarded keys we accept, I decided that we only consider forwarded room keys to be accepted as historic ones only if there's a room present.

No room, no room history and the flag is meaningless to us. It makes it a bit harder for users to try to insert arbitrary room keys into our store. 

Now the point is mostly moot, since the homeserver controls our room membership and it can insert us into a room to get us to consider the flag anyways. This check would make a bit more sense in a rosy world where the homeserver can't insert us  into arbitrary rooms.

Considering the above I'm fine with removing it, though I'll remove the whole method in that case and just access the flag directly in the `(fromInviter && sharedAsHistory)` line. That way the method can't be used in a place where there's no `room` check alongside it.

src/crypto/algorithms/megolm.ts
poljar: Added it back as a doc comment in https://github.com/matrix-org/matrix-js-sdk/pull/2982/commits/fb4d9b662158911dbb5eb5cb8af685a94b55d0fe.

src/crypto/algorithms/megolm.ts
richvdh: > I'll remove the whole method in that case and just access the flag directly in the `(fromInviter && sharedAsHistory)` line.

That sounds like a good approach to me.

I don't have a strong feeling here, but if you want to keep the method, I would ask that you add a comment explaining some of this:

```javascript
// If the key is not for a known room, then something fishy is going on, so we reject the key out of caution.
// In practice, this is a bit moot because we'll only accept shared_history forwarded by the inviter, and we won't 
// know who was the inviter for an unknown room, so we'll reject it anyway.
if (!room) {
   return false;
}
```

src/crypto/algorithms/megolm.ts
```diff
@@ -1419,154 +1450,326 @@ export class MegolmDecryption extends DecryptionAlgorithm {
...
+        //
+        // It would still be far nicer if the `sendingDevice` had a user ID
+        // attached to it that went through signature checks.
+        const fromUs = event.getSender() === this.baseApis.getUserId();
+        const keyFromOurVerifiedDevice = deviceTrust.isVerified() && fromUs;
+        const weRequested = await this.wasRoomKeyRequested(event, roomKey);
+        const fromInviter = this.wasRoomKeyForwardedByInviter(event, roomKey);
+        const sharedAsHistory = this.wasRoomKeyForwardedAsHistory(roomKey);
+
+        return (weRequested && keyFromOurVerifiedDevice) || (fromInviter && sharedAsHistory);
```
richvdh, line 1578: Another thing that might be clearer here is to break this out, allowing the addition of some comments:

```suggestion
        // we trust the forwarded key if we asked for it and it's from one of our own devices...
        if (weRequested && keyFromOurVerifiedDevice) {
            return true;
        }
        
        // ... or if it's from the user that invited us to the room, and it is being forwarded as part 
        // of the "shared history" of the room ...
        if (fromInviter && sharedAsHistory) {
            return true;
        }
        
        // ... otherwise, we ignore it.
        return false;
```

YMMV though; keep what you have if you prefer.

richvdh:
Fixes: https://github.com/vector-im/element-web/issues/23766
Fixes: https://github.com/vector-im/element-web/issues/24681

t3chguy:
@richvdh would be worth updating the OP otherwise github won't auto-close

src/crypto/algorithms/megolm.ts
poljar: I opted for the additional comment in the end https://github.com/matrix-org/matrix-js-sdk/pull/2982/commits/dd4973fd4c0ede7b195f3a7a584f376d925d52a9.

src/crypto/algorithms/megolm.ts
poljar: I prefer the one-liner, at least unless more cases were to be added.

poljar:
I think that this has addressed all the issues and I merged `develop` in.

richvdh (approved):
lgtm!

andybalaam:
Adding back to the merge queue after the change that I am hoping fixes the quality metrics.

PR #2983, opened 2022-12-14
Add support for simulcast

SimonBrandner:


src/webrtc/call.ts
```diff
@@ -839,24 +867,30 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
...
+                    });
 
-                    // now go & fish for the new transceiver
-                    const newTransciever = this.peerConn!.getTransceivers().find((t) => t.sender === newSender);
-                    if (newTransciever) {
-                        this.transceivers.set(tKey, newTransciever);
-                    } else {
-                        logger.warn("Didn't find a matching transceiver after adding track!");
+                    const parameters = transceiver.sender.getParameters();
+                    if (isFirefox()) {
```
dbkr, line 876: Could do with a comment on why the firefox-specific hackery is needed and what has to change so we can get rid of it, especially as it looks like we're doing the same unconditionally on line 856?

src/webrtc/call.ts
```diff
@@ -839,24 +867,30 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
                     // doesn't yet implement RTCRTPSender.setStreams()
                     // (https://bugzilla.mozilla.org/show_bug.cgi?id=1510802) so we'd have no way to group the
                     // two tracks together into a stream.
```
dbkr, line 869: Comment needs updating as this is now using addTransceiver

src/webrtc/call.ts
```diff
@@ -839,24 +867,30 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
...
-                        logger.warn("Didn't find a matching transceiver after adding track!");
+                    const parameters = transceiver.sender.getParameters();
+                    if (isFirefox()) {
+                        transceiver.sender.setParameters({
+                            ...transceiver.sender.getParameters(),
+                            encodings: encodings ?? parameters.encodings,
+                        });
                     }
+
+                    // now go & fish for the new transceiver
```
dbkr, line 883: Comment is now obsolete

src/webrtc/call.ts
```diff
@@ -828,9 +849,16 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
 
                     // this is what would allow us to use addTransceiver(), but it's not available
                     // on Firefox yet. We call it anyway if we have it.
-                    if (transceiver.sender.setStreams) transceiver.sender.setStreams(callFeed.stream);
+                    if (transceiver.sender.setStreams) transceiver.sender.setStreams(callFeed.stream!);
```
dbkr, line 852: Comment again suggests this will break if we switch to addTransceiver, but we have done - will this work?

src/webrtc/call.ts
```diff
@@ -839,24 +867,30 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
                     // doesn't yet implement RTCRTPSender.setStreams()
                     // (https://bugzilla.mozilla.org/show_bug.cgi?id=1510802) so we'd have no way to group the
                     // two tracks together into a stream.
-                    const newSender = this.peerConn!.addTrack(track, callFeed.stream);
+                    const transceiver = this.peerConn!.addTransceiver(track, {
+                        streams: [callFeed.stream!],
+                        sendEncodings: isFirefox() ? undefined : encodings,
```
dbkr, line 872: Are we passing sendEncodings even if we're not using an SFU?

src/webrtc/call.ts
```diff
@@ -1379,7 +1433,19 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
...
+                    if (transceiver.sender.setStreams) transceiver.sender.setStreams(this.localUsermediaStream!);
+
                     await oldSender.replaceTrack(track);
+
+                    const parameters = transceiver.sender.getParameters();
+                    transceiver.sender.setParameters({
+                        ...transceiver.sender.getParameters(),
+                        encodings: encodings ?? parameters.encodings,
+                    });
+
```
dbkr, line 1448: Feels like we're duplicating quite a lot of stuff between here & pushLocalFeed - would be good to re-use some of that, but that doesn't have to be now.

src/webrtc/call.ts
```diff
@@ -2038,53 +2111,121 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
...
+    /**
+     * Send an m.call.track_subscription event to the focus. The method is
+     * throttled to avoid spamming the focus with events when scrolling or
+     * resizing the window.
+     * The m.call.track_subscription lets the focus know about the visibility
+     * and size of different tracks.
+     * @param force - whether or not to force the request to be sent immediately
+     */
+    public subscribeToFocus(force = false): void {
+        // TODO: Can we throttle this better?
```
dbkr, line 2123: lodash.throttle / debounce are your friends

src/webrtc/call.ts
```diff
@@ -267,6 +273,34 @@ const CALL_TIMEOUT_MS = 60 * 1000; // ms
...
+        maxBitrate: 1_500_000,
+        rid: SimulcastResolution.Half,
+        scaleResolutionDownBy: 2.0,
+    },
+    {
+        maxBitrate: 300_000,
+        rid: SimulcastResolution.Quarter,
+        scaleResolutionDownBy: 4.0,
+    },
+];
```
dbkr, line 303: Just to note that the bitrates here are probably defined assuming 720p (ie. what we ask for in getUserMedia) but screenshares will be different resolutions, so these figures might not be as sensible for those (plus of course we may not actually get 720p).

dbkr (commented):
Generally looks good. I don't fully understand the reason for moving from pushRemoteFeed to pushRemoteStream though?

src/webrtc/call.ts
SimonBrandner: So I think they are actually not necessary and we should be fine without setting them - let's see :crossed_fingers:

src/webrtc/call.ts
SimonBrandner: I've created https://github.com/vector-im/element-call/issues/877

src/webrtc/call.ts
SimonBrandner: Looking at all of this now, it seems we could probably remove some of the hackery but I am not fully confident yet - have created https://github.com/vector-im/element-call/issues/878 and will look at it ASAP

SimonBrandner:
> Generally looks good. I don't fully understand the reason for moving from pushRemoteFeed to pushRemoteStream though?

I liked the name better with the changes made

src/webrtc/call.ts
dbkr: Cool, that's good too.

src/webrtc/call.ts
dbkr: Ah, ok - ftr I'd vote for adding it: it will only pull in the bits you use so it's not expensive.

dbkr (approved):
Thanks for the fast turnaround - lgtm

SimonBrandner:
> Thanks for the fast turnaround - lgtm

Thanks to you, too!

PR #2991, opened 2022-12-16
sliding sync: Fix issue where no unsubs are sent when switching rooms

S7evinK:
Should fix an issue where no `unsubscribe_rooms` are sent when switching rooms. 
This happens because we're deleting the room from the `confirmedRoomSubscription` list when switching between unencrypted rooms and the `unsubscribe_rooms` list can't be calculated correctly, resulting in an ever growing subscriptions list on the server side.

## Checklist

---

## 🐛 Bug Fixes
 * sliding sync: Fix issue where no unsubs are sent when switching rooms ([\#2991](https://github.com/matrix-org/matrix-js-sdk/pull/2991)). Contributed by @S7evinK.

### DIFF
--- a/src/sliding-sync.ts
+++ b/src/sliding-sync.ts
@@ -397,6 +397,10 @@ export class SlidingSync extends TypedEventEmitter<SlidingSyncEvent, SlidingSync
      * @param sub - The subscription information.
      */
     public addCustomSubscription(name: string, sub: MSC3575RoomSubscription): void {
+        if (this.customSubscriptions.get(name) !== undefined) {
+            logger.warn(`addCustomSubscription: ${name} already exists as a custom subscription, ignoring.`);
+            return;
+        }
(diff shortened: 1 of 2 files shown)


kegsay (approved):
You'll need some Jest tests for this though.

src/sliding-sync.ts
```diff
@@ -397,6 +397,10 @@ export class SlidingSync extends TypedEventEmitter<SlidingSyncEvent, SlidingSync
      * @param sub - The subscription information.
      */
     public addCustomSubscription(name: string, sub: MSC3575RoomSubscription): void {
+        if (this.customSubscriptions.get(name) !== undefined) {
```
t3chguy, line 400:
```suggestion
        if (this.customSubscriptions.has(name)) {
```

richvdh (approved):

PR #2992, opened 2022-12-16
Close all streams when a call ends

dbkr:
We didn't close streams in group calls (presumably from back when we used the same stream for all calls rather than cloning?) but this left stray screenshare streams in the mediahandler when a participant left whilst we were screensharing.

Fixes https://github.com/vector-im/element-call/issues/742

## Checklist

---

## 🐛 Bug Fixes
 * Close all streams when a call ends ([\#2992](https://github.com/matrix-org/matrix-js-sdk/pull/2992)). Fixes vector-im/element-call#742.


SimonBrandner (approved):
Thanks!

PR #2998, opened 2022-12-20
Break coupling between `Room` and `Crypto.trackRoomDevices`

richvdh:
`Room` and `Crypto` currently have some tight coupling in the form of a call to `trackRoomDevices` when out-of-band members are loaded. We can improve this by instead having Crypto listen out for a `RoomSateEvent.Update` notification.

---


t3chguy (approved):
Looks sane!

Issue #1648, opened 2021-03-22
Unable to decrypt: Unknown encryption algorithm \"undefined\".


michel-zimmer:
**Describe the bug**
Decrypted event source:

```json
{
  "type": "m.room.message",
  "content": {
    "msgtype": "m.bad.encrypted",
    "body": "** Unable to decrypt: Unknown encryption algorithm \"undefined\". **"
  }
}
```

Original event source:

```json
{
  "content": {},
  "origin_server_ts": 1593761398124,
  "room_id": "!…:matrix.org",
  "sender": "@michel-zimmer:matrix.org",
  "type": "m.room.encrypted",
  "unsigned": {
    "redacted_by": "$6M1oJkRMLfn3k1ALmnLbbWq2vemoO6cCLomowO4BkbM",
    "redacted_because": {
      "content": {},
      "origin_server_ts": 1593764558519,
      "redacts": "$VmprAKEQ9RA_NloLleUbkCKPpf8FWVf5yly2gD_mW58",
      "room_id": "!…:matrix.org",
      "sender": "@michel-zimmer:matrix.org",
      "type": "m.room.redaction",
      "unsigned": {
        "age": 22680076733
      },
      "event_id": "$6M1oJkRMLfn3k1ALmnLbbWq2vemoO6cCLomowO4BkbM",
      "user_id": "@michel-zimmer:matrix.org",
      "age": 22680076733
    },
    "age": 22683237128
  },
  "event_id": "$VmprAKEQ9RA_NloLleUbkCKPpf8FWVf5yly2gD_mW58",
  "user_id": "@michel-zimmer:matrix.org",
  "age": 22683237128,
  "redacted_because": {
    "content": {},
    "origin_server_ts": 1593764558519,
    "redacts": "$VmprAKEQ9RA_NloLleUbkCKPpf8FWVf5yly2gD_mW58",
    "room_id": "!…:matrix.org",
    "sender": "@michel-zimmer:matrix.org",
    "type": "m.room.redaction",
    "unsigned": {
      "age": 22680076733
    },
    "event_id": "$6M1oJkRMLfn3k1ALmnLbbWq2vemoO6cCLomowO4BkbM",
    "user_id": "@michel-zimmer:matrix.org",
    "age": 22680076733
  }
}
```
2nd example for an original event source:

```json
{
  "content": {},
  "origin_server_ts": 1616336830758,
  "sender": "@michel-zimmer:matrix.org",
[...]

PR #1589, opened 2021-02-01
Decrypt redaction events

SimonBrandner:
This PR adds support for decrypting redaction events. This is useful for displaying encrypted redaction reasons

PR #3004, opened 2022-12-21
Avoid triggering decryption errors when decrypting redacted events

uhoreg:
Currently, when processing a redacted event, it tries to decrypt the `redacted_because`  (https://github.com/matrix-org/matrix-js-sdk/pull/1589).  However, since this often is not encrypted, the decryption will fail, and this failure will bubble up to the parent event.  This patch simply uses the plain `redacted_because` if it is unable to decrypt it.

fixes https://github.com/vector-im/element-web/issues/24084 and https://github.com/matrix-org/matrix-js-sdk/issues/1648

---

## 🐛 Bug Fixes
 * Avoid triggering decryption errors when decrypting redacted events ([\#3004](https://github.com/matrix-org/matrix-js-sdk/pull/3004)). Fixes vector-im/element-web#24084.

### DIFF
--- a/spec/unit/crypto.spec.ts
+++ b/spec/unit/crypto.spec.ts
@@ -167,6 +167,35 @@ describe("Crypto", function () {
 
             client.stopClient();
         });
+
+        it("doesn't throw an error when attempting to decrypt a redacted event", async () => {
+            const client = new TestClient("@alice:example.com", "deviceid").client;
+            await client.initCrypto();
+
+            const event = new MatrixEvent({
+                content: {},
+                event_id: "$event_id",
+                room_id: "!room_id",
+                sender: "@bob:example.com",
+                type: "m.room.encrypted",
+                unsigned: {
+                    redacted_because: {
+                        content: {},
+                        event_id: "$redaction_event_id",
+                        redacts: "$event_id",
+                        room_id: "!room_id",
+                        origin_server_ts: 1234567890,
+                        sender: "@bob:example.com",
+                        type: "m.room.redaction",
+                        unsigned: {},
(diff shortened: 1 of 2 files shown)


spec/unit/crypto.spec.ts
```diff
@@ -167,6 +167,35 @@ describe("Crypto", function () {
 
             client.stopClient();
         });
+
+        it("doesn't throw an error when attempting to decrypt a redacted event", async () => {
```
gsouquet, line 171: Could we also assess that `redacted_because` is correctly set? That sounds like an important edge to test against

gsouquet (changes requested):
Otherwise looks good to me

src/crypto/index.ts
```diff
@@ -2866,19 +2866,28 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
...
                 ...event.getUnsigned().redacted_because,
             });
-            const decryptedEvent = await this.decryptEvent(redactionEvent);
+            let redactedBecause: IEvent;
+            try {
+                const decryptedEvent = await this.decryptEvent(redactionEvent);
+                redactedBecause = decryptedEvent.clearEvent as IEvent;
+            } catch {
+                redactedBecause = event.getUnsigned().redacted_because!;
+            }
```
richvdh, line 2882: I feel like we should make this conditional on `redactionEvent.isEncrypted()`, and log a warning if we find it is encrypted but we can't decrypt.

gsouquet (approved):

richvdh (approved):

PR #3005, opened 2022-12-21
Add `device_id` to `/account/whoami` types

turt2live:
https://spec.matrix.org/v1.5/client-server-api/#get_matrixclientv3accountwhoami

---

## ✨ Features
 * Add `device_id` to `/account/whoami` types ([\#3005](https://github.com/matrix-org/matrix-js-sdk/pull/3005)).

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -9383,7 +9383,6 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * Fetches information about the user for the configured access token.
      */
     public async whoami(): Promise<IWhoamiResponse> {
-        // eslint-disable-line camelcase
         return this.http.authedRequest(Method.Get, "/account/whoami");
     }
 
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -9377,7 +9377,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     /**
      * Fetches the user_id of the configured access token.
```
robintown, line 9378: could use amending

src/client.ts
```diff
@@ -9377,7 +9377,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
      */
-    public async whoami(): Promise<{ user_id: string }> {
+    public async whoami(): Promise<{ user_id: string; device_id?: string }> {
```
robintown, line 9380: Our style guide says we shouldn't be defining interfaces inline

robintown (approved):

PR #3009, opened 2022-12-23
Factor `SyncApi` options out of `IStoredClientOptions`

richvdh:
There are a couple of callback interfaces which are currently stuffed into `IStoredClientOpts` to make it easier to pass them into the `SyncApi` constructor. This isn't really a sensible place for them, because they aren't application options in the same way as the rest of `IStoredClientOpts` are, nor are they 'stored' in indexedDB as the other options are.

I want to add `cryptoBackend` to the things passed into `SyncApi`, so rather than perpetuating the mess, define a new interface `SyncApiOptions` to be passed in separately.

(sorry, this isn't quite passing the code coverage gate. But it almost does, and it's super-annoying to add the tests to make it do so.)

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -21,7 +21,7 @@ limitations under the License.
 import { EmoteEvent, IPartialEvent, MessageEvent, NoticeEvent, Optional } from "matrix-events-sdk";
 
 import type { IMegolmSessionData } from "./@types/crypto";
-import { ISyncStateData, SyncApi, SyncState } from "./sync";
+import { ISyncStateData, SyncApi, SyncApiOptions, SyncState } from "./sync";
 import {
     EventStatus,
     IContent,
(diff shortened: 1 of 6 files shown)


src/client.ts
```diff
@@ -1433,20 +1423,18 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
-        // shallow-copy the opts dict before modifying and storing it
-        this.clientOpts = Object.assign({}, opts) as IStoredClientOpts;
-        this.clientOpts.crypto = this.crypto;
-        this.clientOpts.canResetEntireTimeline = (roomId): boolean => {
-            if (!this.canResetTimelineCallback) {
-                return false;
-            }
-            return this.canResetTimelineCallback(roomId);
-        };
+        this.clientOpts = opts ?? {};
```
SimonBrandner, line 1426: Are we sure we don't want to copy here?

SimonBrandner (approved):
LGTM, otherwise

src/client.ts
richvdh: given we aren't going to modify it, I figured there wasn't much point.

SimonBrandner: Sounds good then

PR #3016, opened 2023-01-02
Use correct /v3 prefix for /refresh

davidisaaclee:
## Checklist

-   [x] Tests written for new code (and old code if feasible) – **I only added tests for the default behavior, not for all configurations of the options I added.** Will wait on review to add any more.

Synapse v1.71 and below incorrectly exposed `/refresh` under the `/v1` endpoint. This was fixed in https://github.com/matrix-org/synapse/pull/14364, which was released with v1.72.0. This fix causes `matrix-js-sdk`'s `MatrixClient#refreshToken` to fail when targeting newer Synapse instances (or any server which correctly implements the spec). `matrix-js-sdk` currently fails `refreshToken` when targeting the matrix.org homeserver.

I've supplied some alternative levels of fixing in this PR, since I don't know how committed you are to preserving backwards compatibility / matching the spec; my goal was that we could drop or squash commits as maintainers desire to match the expectations of this repo.
- a8df2a9ee2d9fd10a37ed3e32d3b294ef7c358af simply swaps out the `/v1` prefix to instead use `/v3`. This is correct, but breaks clients targeting older Synapses.
- 2c85e750ba37d60c2301d8c274fbf53fbccd8820 gives an option to `refreshToken` for library consumers to explicitly ask for `/v1`; will break things by default, but consumers can individually manage calling the incorrect `/v1` prefix.
- c8aa908 adds an option to first try the correct `/v3` prefix, then retry with `/v1` on `M_UNRECOGNIZED`, defaulting to false. Default behavior is correct to spec without extra weight, but breaks users on old Synapses; but users can opt in to an automatic fix by adding the option.
- 39edcd5 enables the retry option by default, which should work for everyone, but adds complexity by default (e.g. this could be confusing if `/refresh` raises an `M_UNRECOGNIZED` which is not relevant to this issue)

I noted that `getRoomHierarchy` has a similar approach, but for switching between unstable and stable endpoints: https://github.com/matrix-org/matrix-js-sdk/blob/e6bf5ebd0eccd8746175337935623a6551c619f5/src/client.ts#L9266-L9279

---

## ✨ Features
 * Use correct /v3 prefix for /refresh ([\#3016](https://github.com/matrix-org/matrix-js-sdk/pull/3016)). Contributed by @davidisaaclee.


spec/unit/login.spec.ts
```diff
@@ -57,3 +57,46 @@ describe("SSO login URL", function () {
...
+            access_token: "access_token",
+            refresh_token: "refresh_token",
+            expires_in_ms: 30000,
+        };
+
+        client.httpBackend.when("POST", "/_matrix/client/v3/refresh").respond(400, { errcode: "M_UNRECOGNIZED" });
+        client.httpBackend.when("POST", "/_matrix/client/v1/refresh").respond(200, response);
+
+        const refreshResultPromise = client.client.refreshToken("initial_refresh_token");
+        await client.httpBackend.flushAllExpected();
```
davidisaaclee, line 99: Is there a better way to assert "these two `ExpectedRequests` were received"?

richvdh (commented):
@davidisaaclee thanks very much for picking this up, and sorry that it appears to have languished rather than getting a prompt review.

In general, I'd say that exposing both `forcePrefixV1 and `retryWithPrefixV1OnUnrecognizedRequest` to the application are unnecessary flexibility. The application should expect the SDK to take care of picking the right endpoint to use.

I think your default behaviour of retrying would be fine.

However, in this particular case, I think there's a strong argument that we shouldn't support `/v1/refresh` at all. It was never written down in the spec, and I would expect any users trying to use refresh tokens to be doing it with recent versions of Synapse.  So personally I would drop all the fancy logic and just use `/v3/refresh`. YMMV though.

spec/unit/login.spec.ts
richvdh: I suspect that, in view of my other comment, your question becomes moot, but to answer it: this is fine as it is: `flushAllExpected` will ensure that both of the expected calls have been made, so I don't think you need to do anything further.

A preferable solution IMHO is not to use `TestClient` (which comes with a load of baggage, and forces you to use the rather painful `matrix-mock-request`), and instead to instantiate a MatrixClient yourself and use `fetch-mock-jest` to intercept the HTTP calls. We're gradually trying to transition away from `TestClient` and `matrix-mock-request`.

davidisaaclee:
@richvdh Thank you for reviewing. It makes sense that the library should _just work_, so I'll remove these options.

> I think there's a strong argument that we shouldn't support /v1/refresh at all [...] I would expect any users trying to use refresh tokens to be doing it with recent versions of Synapse

My motivation for this PR is that I'm trying to use refresh tokens with an older version of Synapse _as well as_ with the matrix.org server, so the retry logic is directly helping my use case and likely would help others like me (probably what you meant by "YMMV") — so I would selfishly prefer to keep the default behavior of "try /v3/refresh first, then /v1/refresh on `M_UNRECOGNIZED`." Without that behavior, I would need to manually catch the error and do a raw HTTP request from my application – yuck!

I'll plan to make these code changes soon (requires more than reverts to add retry without exposing configuration param).

spec/unit/login.spec.ts
davidisaaclee: Switched to using `fetch-mock-jest` in [Add tests to ensure /v3/refresh is called + automatic /v1 retry](https://github.com/matrix-org/matrix-js-sdk/pull/3016/commits/036c0e03bc9b64ebad9ddaa1db629dab5f450520) – easy to use, thanks for the recommendation.

davidisaaclee:
I rewrote git history on this branch since the changes were significant. To recap:

- [Request /refresh with v3 prefix, and quietly fall back to v1](https://github.com/matrix-org/matrix-js-sdk/pull/3016/commits/0d06e7713c761ee69daead1dd3c4c858b7ea8578) is the simpler approach of "try /v3 first; if that fails with `M_UNRECOGNIZED`, fall back to /v1" – no user configuration
- [Add tests to ensure /v3/refresh is called + automatic /v1 retry](https://github.com/matrix-org/matrix-js-sdk/pull/3016/commits/036c0e03bc9b64ebad9ddaa1db629dab5f450520) checks the fallback behavior, and [Add tests checking re-raising errors](https://github.com/matrix-org/matrix-js-sdk/pull/3016/commits/fe46e0c817b64da1ee50d772897c41b70c7394a7) checks that errors raised in each prefixed endpoint are raised appropriately (which seemed like a potential bug given my implementation).

src/client.ts
```diff
@@ -7569,16 +7569,25 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+                },
+            );
+
+        return performRefreshRequestWithPrefix(ClientPrefix.V3).catch((e) => {
+            if (e.errcode === "M_UNRECOGNIZED") {
+                // fall back to the /v1 prefixed API, which was incorrectly used in Synapse before 1.72.0
+                return performRefreshRequestWithPrefix(ClientPrefix.V1);
+            }
+            throw e;
+        });
```
richvdh, line 7590: I think we could clarify this a bit with a comment:

```suggestion
        // First try with the (specced) /v3/ prefix.
        // However, before Synapse 1.72.0, Synapse incorrectly required a /v1/ prefix, so we fall
        // back to that if the request fails, for backwards compatibility.
        return performRefreshRequestWithPrefix(ClientPrefix.V3).catch((e) => {
            if (e.errcode === "M_UNRECOGNIZED") {
                return performRefreshRequestWithPrefix(ClientPrefix.V1);
            }
            throw e;
        });
```

spec/unit/login.spec.ts
```diff
@@ -57,3 +66,84 @@ describe("SSO login URL", function () {
         });
     });
 });
+
+describe("refreshToken", () => {
```
richvdh, line 70: a great set of tests, thank you!

richvdh (commented):
other than my comments on comments, this looks great.

Please could you sign off your contribution by adding a comment to the PR with "Signed-off-by: ...", as per https://github.com/vector-im/element-web/blob/develop/CONTRIBUTING.md#sign-off.

spec/unit/login.spec.ts
```diff
@@ -1,6 +1,15 @@
+import fetchMock from "fetch-mock-jest";
+
+import { ClientPrefix, MatrixClient, MatrixError } from "../../src";
```
richvdh, line 3:
```suggestion
import { ClientPrefix, MatrixClient } from "../../src";
```

src/client.ts
davidisaaclee: makes sense - committed this suggestion [0ddebe9](https://github.com/matrix-org/matrix-js-sdk/pull/3016/commits/0ddebe956e84d5b75c179748f1b0c1562e1c6312)

davidisaaclee:
@richvdh added sign-off to PR description

thanks again for review. I think everything is addressed now, lmk what's next!

richvdh (approved):
looks great, thank you!

davidisaaclee:
Hey – is there anything I can do to help merge this?

richvdh:
Thank you for the reminder; I don't know why github hasn't merged it. Hopefully it will work this time; if it hasn't auto-merged in a couple of hours feel free to ping me again

PR #3017, opened 2023-01-03
In release.yml use an env section to get ref_name

andybalaam:
This is completely untested, but follows the example in https://github.com/vector-im/element-call/blob/main/.github/workflows/netlify-pr.yaml#L49-L51

Related to https://github.com/matrix-org/internal-config/issues/1388

---

## 🐛 Bug Fixes
 * In release.yml use an env section to get ref_name ([\#3017](https://github.com/matrix-org/matrix-js-sdk/pull/3017)). Contributed by @andybalaam.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,7 +33,7 @@ jobs:
 
             - name: 🔪 Prepare
               env:
-                GITHUB_REF_NAME: ${{ github.ref_name }}
+                  GITHUB_REF_NAME: ${{ github.ref_name }}
               run: |
                   tag="${GITHUB_REF_NAME}"
                   VERSION="${tag#v}"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -32,8 +32,10 @@ jobs:
                   ref: gh-pages
 
             - name: 🔪 Prepare
+              env:
+                  GITHUB_REF_NAME: ${{ github.ref_name }}
               run: |
-                  tag="${{ github.ref_name }}"
+                  tag="${GITHUB_REF_NAME}"
```
t3chguy, line 38: This var is now redundant, can be accessed directly in L39 to strip the leading `v`

t3chguy (approved):

PR #3018, opened 2023-01-04
Add alt event type matching in Relations model

kerryarchibald:
Allow one set of relations to support both stable and unstable variants of an event type.
Enables nicer poll response aggregation.

`RelatedRelations` can be used to combine `Relations` instances, but it produces a new object on every call of `getRelations`. 

Also removes unnecessary (copy-pasted?) code from `removeEvent`

## Checklist

### DIFF
--- a/spec/unit/relations.spec.ts
+++ b/spec/unit/relations.spec.ts
@@ -14,13 +14,21 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import { M_POLL_START } from "matrix-events-sdk";
+
 import { EventTimelineSet } from "../../src/models/event-timeline-set";
 import { MatrixEvent, MatrixEventEvent } from "../../src/models/event";
 import { Room } from "../../src/models/room";
 import { Relations } from "../../src/models/relations";
 import { TestClient } from "../TestClient";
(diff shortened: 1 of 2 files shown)


src/models/relations.ts
```diff
@@ -122,20 +127,6 @@ export class Relations extends TypedEventEmitter<RelationsEvent, EventHandlerMap
             return;
         }
 
-        const relation = event.getRelation();
```
kerryarchibald, line 125: Events can only be added to `this.relations` if they pass the identical checks in `addEvent`, so these checks are superfluous.

spec/unit/relations.spec.ts
```diff
@@ -75,6 +83,92 @@ describe("Relations", function () {
...
+                        rel_type: relationType,
+                    },
+                },
+            });
+
+            await relations.addEvent(event);
+
+            // event added
+            expect(relations.getRelations()).toEqual([event]);
+            expect(emitSpy).toHaveBeenCalled();
```
robintown, line 146: Might as well make this a `toHaveBeenCalledWith` assertion since the event is right there

robintown (approved):

PR #3021, opened 2023-01-04
Pass to-device messages into rust crypto-sdk

richvdh:
This introduces a new method in `SyncCryptoCallbacks`. We can't just piggyback onto `ClientEvent.ToDeviceEvent`, because that is only emitted for successfully decrypted to-device events.

We also don't really want to hook into the EventMapper mechanism (as the current `Crypto` implementation does), because the rust crypto api expects the raw to-device events, rather than those that have been mangled into MatrixEvents.

---


weeman1337 (approved):
LGTM :+1:

PR #3024, opened 2023-01-04
Make error handling in decryptionLoop more generic

richvdh:
Not everything is a `DecryptionError`, and there's no real reason that we should only do retries for `DecryptionError`s.

---


andybalaam (approved):
Agreed, so long as we won't spin doing rapid retries for e.g. a coding error?

richvdh:
> Agreed, so long as we won't spin doing rapid retries for e.g. a coding error?

We'll only do a retry if `retryDecryption` gets set while the decryption takes place, and that only happens if we receive a key that might help decryption. So we might now do a single extra retry for a coding error, but that's not the end of the world.

PR #3028, opened 2023-01-05
Remove video track when muting video

SimonBrandner:
Fixes https://github.com/vector-im/element-call/issues/209

---

## ✨ Features
 * Remove video track when muting video ([\#3028](https://github.com/matrix-org/matrix-js-sdk/pull/3028)). Fixes vector-im/element-call#209.

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -1280,7 +1280,10 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
         setTracksEnabled(stream.getAudioTracks(), audioEnabled);
         setTracksEnabled(stream.getVideoTracks(), videoEnabled);
 
-        // We want to keep the same stream id, so we replace the tracks rather than the whole stream
+        // We want to keep the same stream id, so we replace the tracks rather
+        // than the whole stream.
+
+        // Firstly, we replace the tracks in our localUsermediaStream.
         for (const track of this.localUsermediaStream!.getTracks()) {
             this.localUsermediaStream!.removeTrack(track);
             track.stop();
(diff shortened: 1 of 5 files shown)


src/webrtc/groupCall.ts
```diff
@@ -609,6 +609,9 @@ export class GroupCall extends TypedEventEmitter<
             logger.log(
                 `groupCall ${this.groupCallId} setLocalVideoMuted stream ${this.localCallFeed.stream.id} muted ${muted}`,
             );
+
+            const stream = await this.client.getMediaHandler().getUserMediaStream(true, !muted);
```
robintown, line 613: Since the audio flag is set to `true`, I take it that the code was already requesting audio, no matter what?

src/webrtc/call.ts
```diff
@@ -1349,7 +1369,12 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
             await this.upgradeCall(false, true);
             return this.isLocalVideoMuted();
         }
-        this.localUsermediaFeed?.setAudioVideoMuted(null, muted);
+        if (this.opponentSupportsSDPStreamMetadata()) {
+            const stream = await this.client.getMediaHandler().getUserMediaStream(true, !muted, false);
```
robintown, line 1373: Why can't we let the stream be reusable?

robintown (commented):
Looks fine, I just want to double-check the parts that I don't quite understand

src/webrtc/call.ts
SimonBrandner: Good catch!

src/webrtc/groupCall.ts
SimonBrandner: Yeah, we basically always request audio since even when muted we want the track to always be there for quick toggling

robintown (approved):

PR #2921, opened 2022-11-30
Redo key sharing after own device verification

weeman1337:
If you don't set up key backup and verify one of your other devices, you won't receive keys for the messages of already viewed rooms. This PR triggers key requests after verifying an own device.

See the linked issue for a detailed description on how to reproduce it.

Fixes https://github.com/vector-im/element-web/issues/23333

Before

https://user-images.githubusercontent.com/6216686/204786166-4bce51c9-81a8-4b22-8d8a-a0417d002abe.mp4

After

https://user-images.githubusercontent.com/6216686/204786195-32c7692a-6562-446b-b6e5-fc2a9ab50b8b.mp4

## Checklist

* [x] Tests written for new code (and old code if feasible)
* [x] Linter and other CI checks pass
* [x] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))

---
Here's what your changelog entry will look like:

## 🐛 Bug Fixes
 * Redo key sharing after own device verification ([\#2921](https://github.com/matrix-org/matrix-js-sdk/pull/2921)). Fixes vector-im/element-web#23333.

PR #3029, opened 2023-01-05
Fix false key requests after verifying new device

weeman1337:
Reverts https://github.com/matrix-org/matrix-js-sdk/pull/2921

closes https://github.com/vector-im/element-web/issues/24167

The reverted PR tried to fix https://github.com/vector-im/element-web/issues/23333 . Unfortunately it triggers key requests from all existing devices.

## Checklist

---

## 🐛 Bug Fixes
 * Fix false key requests after verifying new device ([\#3029](https://github.com/matrix-org/matrix-js-sdk/pull/3029)). Fixes vector-im/element-web#24167 and vector-im/element-web#23333.


alunturner (approved):

BillCarsonFr:
Late comment.
It is possible to just clear the database of existing outgoing requests (instead of re-requesting).
So like that, next time there will be an attempt to decrypt a request will be sent (when you open the timeline). This would greatly reduce the number of key requests sent at once

PR #3031, opened 2023-01-05
Improve hasUserReadEvent and getUserReadUpTo realibility with threads

gsouquet:
Fixes https://github.com/vector-im/element-web/issues/24164
Helps solving https://github.com/vector-im/element-web/issues/23907

Marks all threads before the first threaded read receipts in a room as read. deals with threads that have no read receipts by checking the room unthreaded receipt for that user.

## Checklist

---

## ✨ Features
 * Improve hasUserReadEvent and getUserReadUpTo realibility with threads ([\#3031](https://github.com/matrix-org/matrix-js-sdk/pull/3031)). Fixes vector-im/element-web#24164.


src/models/room.ts
```diff
@@ -309,6 +309,10 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
     private notificationCounts: NotificationCount = {};
     private readonly threadNotifications = new Map<string, NotificationCount>();
     public readonly cachedThreadReadReceipts = new Map<string, CachedReceiptStructure[]>();
+    // Useful to know at what point the current user has started using thread in this room
+    public oldestThreadedReceiptTs = Infinity;
+    // Important to compute compute `hasUserReadEvent` and similar method correctly
+    public unthreadedReceipts = new Map<string, Receipt>();
```
clokep, line 315:
```suggestion
    // Useful to know at what point the current user has started using threads in this room
    public oldestThreadedReceiptTs = Infinity;
    // Important to compute `hasUserReadEvent` and similar methods correctly.
    public unthreadedReceipts = new Map<string, Receipt>();
```

src/models/room.ts
```diff
@@ -2730,6 +2734,15 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
                             { eventId, receiptType, userId, receipt, synthetic },
                         ]);
                     }
+
+                    const me = this.client.getSafeUserId();
+                    if (userId === me && !receiptForMainTimeline && receipt.ts < this.oldestThreadedReceiptTs) {
+                        this.oldestThreadedReceiptTs = receipt.ts;
+                    }
```
clokep, line 2741:
```suggestion
                    // Track the time of the current user's oldest threaded receipt in the room.
                    if (userId === me && !receiptForMainTimeline && receipt.ts < this.oldestThreadedReceiptTs) {
                        this.oldestThreadedReceiptTs = receipt.ts;
                    }
```

src/models/room.ts
```diff
@@ -2730,6 +2734,15 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+
+                    if (!receipt.thread_id && receipt.ts > (this.unthreadedReceipts.get(userId)?.ts ?? 0)) {
+                        this.unthreadedReceipts.set(userId, receipt);
+                    }
```
clokep, line 2745:
```suggestion
                    // Track each user's unthreaded read receipt.
                    if (!receipt.thread_id && receipt.ts > (this.unthreadedReceipts.get(userId)?.ts ?? 0)) {
                        this.unthreadedReceipts.set(userId, receipt);
                    }
```

src/models/thread.ts
```diff
@@ -504,8 +504,42 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
         throw new Error("Unsupported function on the thread model");
     }
 
+    public getEventReadUpTo(userId: string, ignoreSynthesized?: boolean): string | null {
+        const isCurrentUser = userId === this.client.getUserId();
+        if (isCurrentUser && this.lastReply()) {
+            // If a thread last activity is prior the first read receipt sent in a thread
+            // we want to consider that this thread has been read
```
clokep, line 511:
```suggestion
            // If the last activity in a thread is prior to the first threaded read receipt
            // sent in the room we want to consider this thread as read.
```

src/models/thread.ts
```diff
@@ -504,8 +504,42 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+            const beforeFirstThreadedReceipt = (this.lastReply()?.getTs() ?? 0) < this.room.oldestThreadedReceiptTs;
```
clokep, line 512: We know `lastReply()` exists?

```suggestion
            const beforeFirstThreadedReceipt = (this.lastReply().getTs() ?? 0) < this.room.oldestThreadedReceiptTs;
```

src/models/thread.ts
```diff
@@ -504,8 +504,42 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+                return this.timeline.at(-1)?.getId() ?? null;
+            }
+        }
+
+        const readUpToId = super.getEventReadUpTo(userId, ignoreSynthesized);
+
+        // Checking whether the unthreaded read receipt for that user is more recent
+        // than the read receipt inside that thread.
+        if (isCurrentUser && this.lastReply()) {
+            const unthreadedReceiptTs = this.room.unthreadedReceipts.get(userId)?.ts ?? Infinity;
```
clokep, line 523: There's no reason to iterate the timeline if there's no unthreaded receipt:

```suggestion
            if (!this.room.unthreadedReceipts.has(userId)) {
                return readUpToId;
            }
            const unthreadedReceiptTs = this.room.unthreadedReceipts.get(userId).ts;
```

src/models/thread.ts
```diff
@@ -504,8 +504,42 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+            for (let i = this.timeline?.length - 1; i >= 0; --i) {
+                const ev = this.timeline[i];
+                if (ev.getTs() > unthreadedReceiptTs) return ev.getId() ?? readUpToId;
+                if (ev.getId() === readUpToId) return readUpToId;
```
clokep, line 527: Also I wonder if we should check the event ID against `readUpToId` first since timestamps are theoretically not 100% reliable?

I'm not sure this is correct -- I think you want `ev.getTs() <= unthreadedReceipTs` as we want to return the last read event? I think the current code will almost always return the latest event in the timeline? (I think this would then need to default to `0` if there is no unthreaded receipt, this better matches what happens in `hasUserReadEvent` though?)

src/models/thread.ts
```diff
@@ -504,8 +504,42 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+            if (beforeFirstThreadedReceipt) {
+                return this.timeline.at(-1)?.getId() ?? null;
+            }
+        }
+
+        const readUpToId = super.getEventReadUpTo(userId, ignoreSynthesized);
+
+        // Checking whether the unthreaded read receipt for that user is more recent
+        // than the read receipt inside that thread.
+        if (isCurrentUser && this.lastReply()) {
```
clokep, line 522: I'm having trouble following this comment, but have been unable to reword it in a way I'm happy with.

Why do we do this just for the current user? Why do we care about `lastReply()` (which is unused in the if-block)?

src/client.ts
```diff
@@ -4840,13 +4840,8 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
 
-        const supportsThreadRR = this.canSupport.get(Feature.ThreadUnreadNotifications) !== ServerSupport.Unsupported;
-        if (supportsThreadRR && !unthreaded) {
-            const isThread = !!event.threadRootId;
-            body = {
-                ...body,
-                thread_id: isThread ? event.threadRootId : MAIN_ROOM_TIMELINE,
-            };
+        if (this.supportsExperimentalThreads() && !unthreaded) {
+            body.thread_id = event.threadRootId ? event.threadRootId : MAIN_ROOM_TIMELINE;
```
clokep, line 4844: I'm not really sure this change is related?

gsouquet: True, that is not 100% related. I could remove it for now

src/models/thread.ts
gsouquet: You're right, we probably should apply that logic for all user and not only the current one.
Good spot

src/models/thread.ts
gsouquet: I thought this wasn't needed, but it actually is. 
TypeScript has no way to know whether `this.lastReply()` will return the same thing when called twice consecutively

src/client.ts
clokep: It might be fine to keep, I just didn't really understand it. 😄

src/models/thread.ts
clokep: I'm still not sure why we use `lastReply()` though? Should it be `readUpToId` instead?

gsouquet: Hmm maybe we don't need to check `this.lastReply()` in the end.

If `readUpToId` is the last event of the timeline we should return `readUpToId` as no further computation is needed. However for all other cases we want to run the logic below.

And yeah that's slightly different than what is in the diff above. Will update

src/models/thread.ts
richvdh: then it should be `!` ?

```suggestion
            const beforeFirstThreadedReceipt = this.lastReply()!.getTs() < this.room.oldestThreadedReceiptTs;
```

or factor out `const lastReply = this.lastReply()` ?

spec/unit/models/thread.spec.ts
```diff
@@ -96,21 +137,151 @@ describe("Thread", () => {
...
+                client,
+                authorId: "@alice:example.org",
+                participantUserIds: ["@alice:example.org"],
+                length: 2,
+                ts: 10,
+            });
+
+            // The 10th event has been read, as alice's last unthreaded receipt is at ts 200
+            // and `mkThread` increment every thread response by 1ms.
+            expect(thread.getEventReadUpTo(myUserId)).toBe(events.at(-1)!.getId());
```
richvdh, line 272: this comment seems a bit confused:
 * there are only 2 events in `events`, not 10
 * *both* of them have been read, because they all have ts < 200.

src/client.ts
```diff
@@ -601,9 +601,9 @@ interface IJoinRequestBody {
     third_party_signed?: IThirdPartySigned;
 }
 
-interface ITagMetadata {
+export interface ITagMetadata {
```
richvdh, line 604: why is this changing, out of interest?

src/models/room.ts
```diff
@@ -309,6 +309,10 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
     private notificationCounts: NotificationCount = {};
     private readonly threadNotifications = new Map<string, NotificationCount>();
     public readonly cachedThreadReadReceipts = new Map<string, CachedReceiptStructure[]>();
+    // Useful to know at what point the current user has started using threads in this room
+    public oldestThreadedReceiptTs = Infinity;
+    // Important to compute `hasUserReadEvent` and similar methods correctly.
+    public unthreadedReceipts = new Map<string, Receipt>();
```
richvdh, line 315: please can you give this a proper doc-comment? What is the map to and from?

src/models/thread.ts
```diff
@@ -504,17 +504,43 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
         throw new Error("Unsupported function on the thread model");
     }
 
+    public getEventReadUpTo(userId: string, ignoreSynthesized?: boolean): string | null {
```
richvdh, line 507: It would be nice to give this a doc-comment. What does it return? what does a result of `null` mean? what does ignoreSynthesized do?

richvdh: also, uh, where is it actually used? It seems to be a new method but I can't find any new calls outside the tests?

src/models/thread.ts
```diff
@@ -504,17 +504,43 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+        const isCurrentUser = userId === this.client.getUserId();
+        if (isCurrentUser && this.lastReply()) {
+            // If the last activity in a thread is prior to the first threaded read receipt
+            // sent in the room we want to consider this thread as read.
```
richvdh, line 511: *why* do we want to do this?

```suggestion
            // If the last activity in a thread is prior to the first threaded read receipt
            // sent in the room (suggesting that it was sent before the user started
            // using a client that supported threaded read receipts), we want to
            // consider this thread as read.
```

is that right?

src/models/thread.ts
```diff
@@ -504,17 +504,43 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+            const beforeFirstThreadedReceipt = (this.lastReply()?.getTs() ?? 0) < this.room.oldestThreadedReceiptTs;
+            if (beforeFirstThreadedReceipt) {
+                return this.timeline.at(-1)?.getId() ?? null;
+            }
+        }
+
+        const readUpToId = super.getEventReadUpTo(userId, ignoreSynthesized);
+
+        // Checking whether the unthreaded read receipt for that user is more recent
```
richvdh, line 520:
```suggestion
        // Check whether the unthreaded read receipt for that user is more recent
```

src/models/thread.ts
```diff
@@ -504,17 +504,43 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+            if (!this.room.unthreadedReceipts.has(userId)) {
+                return readUpToId;
+            }
+
+            const unthreadedReceiptTs = this.room.unthreadedReceipts.get(userId)!.ts;
+            for (let i = this.timeline?.length - 1; i >= 0; --i) {
+                const ev = this.timeline[i];
+                if (ev.getId() === readUpToId) return readUpToId;
+                if (ev.getTs() < unthreadedReceiptTs) return ev.getId() ?? readUpToId;
+            }
```
richvdh, line 532: please can you add some comments in here explaining this logic.

src/models/thread.ts
```diff
@@ -504,17 +504,43 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
         if (userId === this.client.getUserId()) {
-            const publicReadReceipt = this.getReadReceiptForUserId(userId, false, ReceiptType.Read);
-            const privateReadReceipt = this.getReadReceiptForUserId(userId, false, ReceiptType.ReadPrivate);
-            const hasUnreads = this.room.getThreadUnreadNotificationCount(this.id, NotificationCountType.Total) > 0;
-
-            if (!publicReadReceipt && !privateReadReceipt && !hasUnreads) {
-                // Consider an event read if it's part of a thread that has no
-                // read receipts and has no notifications. It is likely that it is
-                // part of a thread that was created before read receipts for threads
-                // were supported (via MSC3771)
```
richvdh, line 517: it seems a shame to lose this comment. is it no longer correct?

src/models/room.ts
richvdh, line 315: wouldn't it be better to make these private and give them accessors, to ensure they can't be modified outside this class?

src/models/thread.ts
richvdh: oh wait, `Thread` extends `ReadReceipt`? that implies a thread is a particular type read receipt? WTF does "read up to" mean for a single read receipt? I am so confused.

src/models/thread.ts
```diff
@@ -504,17 +504,43 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+            const beforeFirstThreadedReceipt = (this.lastReply()?.getTs() ?? 0) < this.room.oldestThreadedReceiptTs;
+            const unthreadedReceiptTs = this.room.unthreadedReceipts.get(userId)?.ts ?? 0;
+            const beforeLastUnthreadedReceipt = (this?.lastReply()?.getTs() ?? 0) < unthreadedReceiptTs;
```
richvdh, line 542: Why are we replacing the `getReadReceiptForUserId` accessors with new maps in `this.room`? how do the two differ? why do we need both? Please could we have comments to document this stuff?

richvdh (changes requested):
Really struggling to grok this, I'm afraid.

src/models/thread.ts
gsouquet: Correct, we're changing how this is approached, and this comment is no longer relevant

src/models/thread.ts
gsouquet: I've added doc comment, documenting them similarly to what the super class does.

`ReadReceipt` is pretty poorly named I have to admit. It's an abstract class that hosts all the logic related to read receipts between a `Thread` and a `Room`.

`Thread` require a bit of extra logic as clients have gained threaded abilities somewhat recently, and it is possible for a thread to have no receipts at all

src/models/thread.ts
gsouquet: I'm not sure I follow where you're getting at with the comment?

src/client.ts
gsouquet: That should not have ended up here, removed 🔥

spec/unit/models/thread.spec.ts
```diff
@@ -71,17 +75,54 @@ describe("Thread", () => {
...
+                            [client.getUserId()!]: { ts: 200 },
+                            ["@alice:example.org"]: { ts: 200 },
+                        },
+                    },
+                    // last threaded receipt
+                    "$event2:localhost": {
+                        [ReceiptType.Read]: {
+                            [client.getUserId()!]: { ts: 300 },
+                        },
+                    },
```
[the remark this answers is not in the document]
gsouquet: Definitely does, thanks for the heads up

spec/unit/models/thread.spec.ts
gsouquet: Sorry, wrong copy paste! I've ammended that

spec/unit/models/thread.spec.ts
```diff
@@ -96,21 +137,151 @@ describe("Thread", () => {
                 authorId: myUserId,
                 participantUserIds: [myUserId],
                 length: 2,
+                ts: 201,
```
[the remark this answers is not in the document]
gsouquet: This one is slightly different as you don't need a read receipt as the sender is the current user. Therefore no read receipt is needed. 
I've added a comment and removed the `ts`

gsouquet:
@richvdh thank you for the review, that's super helpful.
Managed to address everything, but I couldn't grasp what you were asking with https://github.com/matrix-org/matrix-js-sdk/pull/3031#discussion_r1063403340

src/models/thread.ts
richvdh: yes sorry, this wasn't very well expressed.

My comment was really saying two things, neither of which relate directly to this bit of code.

 1. It felt like a retrograde step to go from accessors like `getReadReceiptForUserId` to directly accessing the properties on Room. You've now changed the direct access to go via accessors, so that part is now resolved.

 2. I was struggling to understand how the read receipt information we now track in `Room` related to the information we already track in `ReadReceipt` (and, by inheritance, Thread). I still am a bit confused about that, to be honest. Do the `receipts` in `ReadReceipt` only include those related to either a particular thread, or no thread at all?

src/models/thread.ts
```diff
@@ -504,17 +504,71 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+        const isCurrentUser = userId === this.client.getUserId();
+        const lastReply = this.lastReply();
+        if (isCurrentUser && lastReply) {
+            // If the last activity in a thread is prior to the first threaded read receipt
+            // sent in the room (suggesting that it was sent before the user started
+            // using a client that supported threaded read receipts), we want to
+            // consider this thread as read.
+            const beforeFirstThreadedReceipt = (lastReply.getTs() ?? 0) < this.room.getOldestThreadedReceiptTs();
+            if (beforeFirstThreadedReceipt) {
+                return this.timeline.at(-1)?.getId() ?? null;
```
richvdh, line 526: is it even possible for `this.timeline.at(-1)` to be nullish here, given lastReply is truthy? If not, let's avoid the `?? null` please.

richvdh (changes requested):
Seems somewhat clearer now - thanks for the updates. Still a few questions though.

src/models/thread.ts
gsouquet: `getId()` can be `undefined`

src/models/thread.ts
```diff
@@ -504,41 +504,69 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
             }
         }
 
         return readUpToId;
     }
 
+    /**
+     * Determines if the given user has read a particular event ID with the known
+     * history of the room. This is not a definitive check as it relies only on
+     * what is available to the thread at the time of execution.
```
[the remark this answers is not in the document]
gsouquet: It was quite confusing indeed, I have changed "history of the room" to "history of the thread" which is more relevant in this context.

gsouquet:
> I was struggling to understand how the read receipt information we now track in Room related to the information we already track in ReadReceipt (and, by inheritance, Thread). I still am a bit confused about that, to be honest. Do the receipts in ReadReceipt only include those related to either a particular thread, or no thread at all?

`ReadReceipt` (which is very ill named, and should be changed, but that won't be part of this PR) is a structure that host all the logic to have a timeline that you want to track read receipts in.
So the main room timeline, or each individual thread.

It's not related to a single read receipt, but it's more the structure to host them, and check the `MatrixEvent` in a timeline

src/models/thread.ts
richvdh: Hrm, I think it should only be `undefined` for unsent events, which should not be relevant here?

In any case we can still assume that `this.timeline.at(-1)` is non-null?

If you want to keep the `?? null` please can you document it in the doc-comment: ie, that this function can return null under particular circumstances (and what they are)? It's super-frustrating to realise that a function sometimes return null and then have to spend hours reverse-engineering when it happens and whether we need to care about it.

src/models/thread.ts
```diff
@@ -505,8 +505,8 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
     }
 
     /**
-     * Get the ID of the event that a given user has read up to, or null if we
-     * have received no read receipts from them.
+     * Get the ID of the event that a given user has read up to within that thread,
```
richvdh, line 508:
```suggestion
     * Get the ID of the event that a given user has read up to within this thread,
```

src/models/thread.ts
```diff
@@ -505,8 +505,8 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+     * or null if we have received no read receipts from them.
```
richvdh, line 509:
```suggestion
     * or null if we have received no read receipt (at all) from them.
```

src/models/thread.ts
```diff
@@ -556,14 +556,18 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+     * history of the thread. This is not a definitive check as it relies only on
      * what is available to the thread at the time of execution.
      * @param userId - The user ID to check the read state of.
      * @param eventId - The event ID to check if the user read.
      * @returns True if the user has read the event, false otherwise.
      */
     public hasUserReadEvent(userId: string, eventId: string): boolean {
         if (userId === this.client.getUserId()) {
+            // Consider an event read if it's part of a thread that has no
+            // read receipts and has no notifications. It is likely that it is
```
richvdh, line 568: This doesn't seem to match the logic?

> part of a thread that has no read receipts

We don't seem to be looking at read receipts on the thread itself?

> has no notifications

... or considering notifications?

src/models/thread.ts
```diff
@@ -556,14 +556,18 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
 
     /**
      * Determines if the given user has read a particular event ID with the known
-     * history of the room. This is not a definitive check as it relies only on
+     * history of the thread. This is not a definitive check as it relies only on
```
richvdh, line 559: sorry, I still don't understand this sentence. What does it mean to have read an event with the known history of the thread (or a room?)

richvdh (changes requested):

src/models/thread.ts
gsouquet: I've updated the logic slightly to avoid the optional chaining here

src/models/room.ts
```diff
@@ -3292,6 +3295,24 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+     * @returns the timestamp of the oldest threaded receipt
+     */
+    public getOldestThreadedReceiptTs(): number {
+        return this.oldestThreadedReceiptTs;
+    }
+
+    /**
+     * Returns the most receipt unthreaded receipt for a given user
+     * @param userId - the MxID of the User
+     * @returns an unthreaded Receipt
```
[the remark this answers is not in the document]
gsouquet: I've added a comment. But it's because some folks can use private read receipts, or disable sending receipts altogether

src/models/thread.ts
gsouquet: It can only iterate through the events that are loaded client-side, and is not going to backfill the room to find a definitive answer. It will only work against what is given to it.

src/models/room.ts
```diff
@@ -3331,7 +3331,8 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
     /**
      * Returns the most recent unthreaded receipt for a given user
      * @param userId - the MxID of the User
-     * @returns an unthreaded Receipt
+     * @returns an unthreaded Receipt, can be undefined if receipts have been disabled
```
richvdh, line 3334:
```suggestion
     * @returns an unthreaded Receipt. Can be undefined if receipts have been disabled
```

src/models/room.ts
```diff
@@ -3331,7 +3331,8 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+     * or a user chooses to use private read receipts.
```
richvdh, line 3335: or the user has never sent a receipt?

src/models/thread.ts
```diff
@@ -504,17 +504,77 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+            }
+        }
+
+        return readUpToId;
+    }
+
+    /**
+     * Determines if the given user has read a particular event ID with the known
+     * history of the thread. This is not a definitive check as it relies only on
+     * what is available to the thread at the time of execution.
```
richvdh, line 562:
```suggestion
     * Determine if the given user has read a particular event.
     *
     * It is invalid to call this method with an event that is not part of this thread.
     *
     * This is not a definitive check as it only checks the events that have been
     * loaded client-side at the time of execution.
```

I think this matches what we're trying to say here?

src/models/thread.ts
richvdh: Ok, well, the doc-comment needs to say that. I've added a suggestion.

src/models/room.ts
richvdh:
```suggestion
     * or a user chooses to use private read receipts (or we have simply not received
     * a receipt from this user yet).
```

src/models/room.ts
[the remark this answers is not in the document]
richvdh: :+1: thanks. In case it wasn't obvious, the question was somewhat rhetorical, and what I should have said was "please *document* what it means if this returns undefined" :smile:

PR #2719, opened 2022-09-30
Modernize http-api - move from `browser-request` to `fetch`

t3chguy:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/2415
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/801

Notes: Changes the `uploadContent` API, kills off `request` and `browser-request` in favour of `fetch`, removed callback support on a lot of the methods, adds a lot of tests.

---
Here's what your changelog entry will look like:

## 🚨 BREAKING CHANGES
 * Changes the `uploadContent` API, kills off `request` and `browser-request` in favour of `fetch`, removed callback support on a lot of the methods, adds a lot of tests. ([\#2719](https://github.com/matrix-org/matrix-js-sdk/pull/2719)). Fixes #2415 and #801.

Issue #244, opened 2016-10-24
Reduce dep tree


kegsay:
This problem seems to almost entirely be down to 'request' which is pulling in things like har-validator

https://matrix.org/jira/browse/SYJS-10

Kegsay:
Short of using a different request library, there isn't much we can do about this.

Kegsay:
Matthew says: This is a major problem for running on constrained devices like Tessels. I'd rather we didn't resolve valid bugs as won't-fix, especially when there's a solution ('using a different request library'). Will bump down the priority instead.

fred-wang:
I think it would indeed be great to reduce dependencies! In https://bugzilla.mozilla.org/show_bug.cgi?id=1199855, I'm using browser-request for now. In addition, I see the following deps:
- browserify: events, punycode, querystring and url (I haven't tried to see if they can be replaced with native support)
- q (see #290 for native support)
- another-json (see #291 for native support)

fred-wang:
cc'ing @clokep , this is about patch "Part 1 - Import Matrix JS SDK from github.com/matrix-org"

t3chguy:
We killed all of the deps mentioned above except browserify which is leaving in October

PR #3033, opened 2023-01-05
Remove 'qs' dependency

clarkf:
'qs' ([package](https://www.npmjs.com/package/qs), [source](https://github.com/ljharb/qs)) appears to be unused since 34c5598 (PR #2719), and can be safely removed.

It's worth noting that 'qs' has the deepest dependency graph of all (non-development) dependencies.

```
$ npm ls --depth=32 --omit=dev
...
├─┬ qs@6.11.0
│ └─┬ side-channel@1.0.4
│   ├─┬ call-bind@1.0.2
│   │ ├── function-bind@1.1.1
│   │ └── get-intrinsic@1.1.3 deduped
│   ├─┬ get-intrinsic@1.1.3
│   │ ├── function-bind@1.1.1 deduped
│   │ ├── has-symbols@1.0.3
│   │ └─┬ has@1.0.3
│   │   └── function-bind@1.1.1 deduped
│   └── object-inspect@1.12.2
...
```

No other dependencies have more than one level of nesting.

#244 is relevant here (maybe?), though this doesn't go particularly far. A full dev install still weighs 372M on my machine, though this reduces a production install from 9.3M to 8.6M.

## Checklist

---

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5872,13 +5872,6 @@ punycode@^2.1.0, punycode@^2.1.1:
   resolved "https://registry.yarnpkg.com/punycode/-/punycode-2.1.1.tgz#b58b010ac40c22c5657616c8d2c2c02c7bf479ec"
   integrity sha512-XRsRjdf+j5ml+y/6GKHPZbrF/8p2Yga0JPtdqTIY2Xe5ohJPD9saDJJLPvp9+NSBprVvevdXZybnj2cv8OEd0A==
 
-qs@^6.9.6:
-  version "6.11.0"
-  resolved "https://registry.yarnpkg.com/qs/-/qs-6.11.0.tgz#fd0d963446f7a65e1367e01abd85429453f0c37a"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -1389,7 +1389,6 @@
 
 "@matrix-org/olm@https://gitlab.matrix.org/api/v4/projects/27/packages/npm/@matrix-org/olm/-/@matrix-org/olm-3.2.14.tgz":
   version "3.2.14"
-  uid acd96c00a881d0f462e1f97a56c73742c8dbc984
```
turt2live, line 1392: this looks like an unrelated change

turt2live (commented):
Thanks! Looks like we indeed don't use this dependency.

Just one comment about the lockfile before we can merge this.

yarn.lock
clarkf: You're right, I'm not sure why yarn (1.22.19, latest as of now) decided to do that. I only `yarn remove`d qs and committed whatever it spat out.

I'll revert the line, but it may crop back up later, I genuinely have no idea. `@matrix-org/olm` seems to be the only entry in that file that has a `uid` value.

turt2live (approved):
thanks :)

Issue #1255, opened 2020-03-10
modify megolm's `prepareToEncrypt` to be cancellable


uhoreg:
`prepareToEncrypt` is introduced in https://github.com/matrix-org/matrix-js-sdk/pull/1252 so that applications can pre-send keys to recipients before the message itself is ready to send, reducing the amount of work that needs to be done when the message is sent.

However, the current function cannot be cancelled if the key cannot be used after all (e.g. a user/device has left, so the key needs to be rotated).

clarkf:
I believe this is the root cause of vector-im/element-web#21612. Or, well, sorta. Mind if I take a stab at it?

[`MegolmEncryption#prepareToEncrypt`][prepareToEncrypt] does a bunch of expensive, blocking crypto on the main thread per untrusted device (see [`CrossSigningInfo#checkDeviceTrust`][checkDeviceTrust]). I don't think a simple cancellation callback will work, since it mostly blocks the caller.

The most straightforward option would be to break the work up into chunks, wrapped in [`setImmediate`](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate) (or similar). Allow preemption between each `checkDeviceTrust` call. Then, cancellation could be added, and vector-im/element-web#21612 resolved (probably).

The "correct" solution would probably be to move the expensive crypto off the main thread, using a WebWorker or something. I don't think that's super reasonable, though, and would require some really sweeping changes.

[prepareToEncrypt]: https://github.com/matrix-org/matrix-js-sdk/blob/c3d422f5fb1efac400da4c4ade592db3831445f9/src/crypto/algorithms/megolm.ts#L944-L989
[checkDeviceTrust]: https://github.com/matrix-org/matrix-js-sdk/blob/c3d422f5fb1efac400da4c4ade592db3831445f9/src/crypto/CrossSigning.ts#L519-L555

PR #3035, opened 2023-01-08
Reduce Megolm blocking and add cancellation

clarkf:
This was originally intended to help with vector-im/element-web#21612, but ended up addressing #1255 as well.

The input delay was addressed by adding a `setTimeout` to the cpu-bound portion of the code. That should allow rendering code and event handlers to preempt key preparation periodically. Should help somewhat, but probably not a full solution. Ideally, `setImmediate` should be used or approximated, but it's basically not supported in any browser, so 0ms timeouts it is.

That allowed for cancellation to be addressed. There are multiple ways this could have been achieved, and I wasn't able to find a clear precedent. I opted for passing around stateful closures. It had the most minimal impact on the API that I could come up with.

I've tried my best to follow the contribution guidelines and code style documentation, but this is my first (substantial) contribution. Any and all feedback welcome!

## Checklist

Notes: Introduces a backwards-compatible API change. `MegolmEncrypter#prepareToEncrypt`'s return type has changed from `void` to `() => void`.

---

## ✨ Features
 * Introduces a backwards-compatible API change. `MegolmEncrypter#prepareToEncrypt`'s return type has changed from `void` to `() => void`. ([\#3035](https://github.com/matrix-org/matrix-js-sdk/pull/3035)). Contributed by @clarkf.

### DIFF
--- a/src/crypto/algorithms/megolm.ts
+++ b/src/crypto/algorithms/megolm.ts
@@ -1,5 +1,5 @@
 /*
-Copyright 2015 - 2021 The Matrix.org Foundation C.I.C.
+Copyright 2015 - 2021, 2023 The Matrix.org Foundation C.I.C.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -43,6 +43,7 @@ import { IMegolmEncryptedContent, IncomingRoomKeyRequest, IEncryptedContent } fr
 import { RoomKeyRequestState } from "../OutgoingRoomKeyRequestManager";
 import { OlmGroupSessionExtraData } from "../../@types/crypto";
 import { MatrixError } from "../../http-api";
+import { immediate } from "../../utils";
 
 // determine whether the key can be shared with invitees
 export function isRoomSharedHistory(room: Room): boolean {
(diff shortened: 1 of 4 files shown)


src/crypto/algorithms/megolm.ts
```diff
@@ -1177,6 +1204,11 @@ export class MegolmEncryption extends EncryptionAlgorithm {
                     continue;
                 }
 
+                // Yield prior to checking each device so that we don't block
+                // updating/rendering for too long.
+                // See https://github.com/vector-im/element-web/issues/21612
+                await sleep(0);
```
robintown, line 1210: While I see how this would unblock the event loop and fix the linked issue, I also want to make sure this won't create a performance regression in `encryptMessage`. It's quite possible that there will be a lot of devices processed here, and [the HTML spec](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timers) claims that after 5 nested calls, any further `setTimeout` calls are clamped to a minimum delay, which could add up.

Perhaps it would be more appropriate for the function to only yield on cancelable operations?

```suggestion
                if (isCancelled !== undefined) await sleep(0);
```

robintown (commented):
The changes look pretty straightforward, thanks.

For the record, you're actually free to use `setImmediate` here, since any time the js-sdk gets used in a browser, it invariably gets run through Babel which has polyfills for that.

src/crypto/algorithms/megolm.ts
clarkf: Interesting, I'd never heard about the minimum delay!

robintown (approved):
Sorry for the wait, this looks good to land now 👍

PR #3036, opened 2023-01-09
Poll model

kerryarchibald:
Adds a Poll model
Create and add polls to room state
Use relations API to aggregate poll responses

## Checklist


src/models/room.ts
```diff
@@ -1880,6 +1888,41 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+            if (event.isBeingDecrypted() || event.isDecryptionFailure()) {
+                // add an event listener for once the event is decrypted.
+                event.once(MatrixEventEvent.Decrypted, async () => {
+                    processPollEvent(event);
+                });
+            } else {
+                processPollEvent(event);
+            }
+        });
+    }
```
gsouquet, line 1924:
```suggestion
	for (const event of events) {
            try {
                await matrixClient.decryptEventIfNeeded(event);
                processPollEvent(event);
            } catch {}
        }
```

`decryptEventIfNeeded` returns a resolved promise for unencrypted or already decrypted events. The above might read a bit better. But that surely shouldn't be blocking this pull request

src/models/room.ts
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { Optional } from "matrix-events-sdk";
+import { M_POLL_START, Optional } from "matrix-events-sdk";
```
gsouquet, line 17: Needs copyright update to 2023 in this file

src/models/relations.ts
```diff
@@ -122,7 +122,7 @@ export class Relations extends TypedEventEmitter<RelationsEvent, EventHandlerMap
      *
      * @param event - The relation event to remove.
      */
-    private async removeEvent(event: MatrixEvent): Promise<void> {
+    public async removeEvent(event: MatrixEvent): Promise<void> {
```
gsouquet, line 125: Also needs copyright update to 2023

src/models/poll.ts
```diff
@@ -0,0 +1,178 @@
...
+export enum PollEvent {
+    New = "Poll.new",
+    End = "Poll.end",
+    Update = "Poll.update",
+    Responses = "Poll.Responses",
+    Destroy = "Poll.Destroy",
+}
+
+export type PollEventHandlerMap = {
+    [PollEvent.Update]: (event: MatrixEvent, poll: Poll) => void;
```
gsouquet, line 33: The handler map is missing a definition for `PollEvent.New`. Is that expected?

src/models/poll.ts
```diff
@@ -0,0 +1,178 @@
...
+
+        this.responses = responses;
+        this.endEvent = pollEndEvent;
+        if (this.endEvent) {
+            this.emit(PollEvent.End);
+        }
+        this.emit(PollEvent.Responses, this.responses);
+    }
+
+    private refilterResponsesOnEnd(): void {
```
gsouquet, line 164: Can we add a comment explaining why it is important to do so?

src/models/poll.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    public get pollId(): string {
+        return this.rootEvent.getId()!;
+    }
+
+    public get isEnded(): boolean {
+        return !!this.endEvent;
+    }
+
+    public setPollStartEvent(event: MatrixEvent): void {
```
weeman1337, line 83: Should it really be possible to change the start events after creating a Poll?
Eventually → `private`

src/models/poll.ts
```diff
@@ -0,0 +1,178 @@
...
+    private refilterResponsesOnEnd(): void {
+        if (!this.responses) {
+            return;
+        }
+
+        const pollEndTimestamp = this.endEvent?.getTs() || Number.MAX_SAFE_INTEGER;
+        this.responses.getRelations().forEach((event) => {
+            if (event.getTs() > pollEndTimestamp) {
+                this.responses?.removeEvent(event);
+            }
```
gsouquet, line 173: For the read receipt code (defining what has been read or not) it usually prefers looking at the order in the DAG and as a last result will use the timestamp.

Could we take a similar approach here?
You have functions like `compareEventOrdering` in `EventTimelineSet` that could be good to use.

`getEventReadUpTo` is a good place to look at the logic in place that favours ordering first, and timestamp as a fallback

src/models/poll.ts
```diff
@@ -0,0 +1,178 @@
...
+    public get isEnded(): boolean {
+        return !!this.endEvent;
+    }
+
+    public setPollStartEvent(event: MatrixEvent): void {
+        this.pollEvent = event.unstableExtensibleEvent as PollStartEvent;
+    }
+
+    public getPollStartEvent(): PollStartEvent {
+        return this.pollEvent!;
```
weeman1337, line 88: Had the same „issue“ with all these `!` while working on broadcasts. What I did is not setting the value in a method. Instead I changed it to return the value and assigned it inside the ctor.

Here this would mean something like `setPollStartEvent` → `determinePollStartEvent` and inside the ctor `this.pollEvent = this.determinePollStartEvent(…)`.

src/models/poll.ts
```diff
@@ -0,0 +1,178 @@
...
+        if (responseEvents.length) {
+            responseEvents.forEach((event) => {
+                this.responses!.addEvent(event);
+            });
+            this.emit(PollEvent.Responses, this.responses);
+        }
+    }
+
+    private async fetchResponses(): Promise<void> {
+        this.fetchingResponsesPromise = new Promise<void>(() => {});
```
gsouquet, line 133: I don't think this is actually needed from what I see?
At least this promise will never be settled

src/models/poll.ts
```diff
@@ -0,0 +1,178 @@
...
+export class Poll extends TypedEventEmitter<Exclude<PollEvent, PollEvent.New>, PollEventHandlerMap> {
+    public readonly roomId: string;
+    private pollEvent: PollStartEvent | undefined;
+    private fetchingResponsesPromise: null | Promise<void> = null;
+    private responses: null | Relations = null;
+    private endEvent: MatrixEvent | undefined;
+
+    public constructor(private rootEvent: MatrixEvent, private matrixClient: MatrixClient) {
+        super();
+        this.roomId = this.rootEvent.getRoomId()!;
```
weeman1337, line 71: Technically an undefined room-Id is allowed here. We should eventually check for it and raise an error, if we don't want to allow polls without rooms (same applies to event Id).

src/models/poll.ts
```diff
@@ -0,0 +1,178 @@
...
+            // From MSC3381:
+            // "Votes sent on or before the end event's timestamp are valid votes"
+            event.getTs() <= pollEndTimestamp
+        );
+    });
+
+    return { responseEvents };
+};
+
+// https://github.com/matrix-org/matrix-spec-proposals/pull/3672
```
gsouquet, line 61: This MSC is talking about

spec/unit/room.spec.ts
```diff
@@ -19,6 +19,7 @@ limitations under the License.
  */
 
 import { mocked } from "jest-mock";
+import { M_POLL_KIND_DISCLOSED, M_POLL_RESPONSE, PollStartEvent } from "matrix-events-sdk";
```
gsouquet, line 22: Copyright update to 2023 here

src/models/poll.ts
```diff
@@ -0,0 +1,178 @@
...
+        if (!this.responses) {
+            return;
+        }
+
+        const pollEndTimestamp = this.endEvent?.getTs() || Number.MAX_SAFE_INTEGER;
+        const { responseEvents } = filterResponseRelations([event], pollEndTimestamp);
+
+        if (responseEvents.length) {
+            responseEvents.forEach((event) => {
+                this.responses!.addEvent(event);
```
weeman1337, line 126: Is the `!` necessary here because of this check https://github.com/matrix-org/matrix-js-sdk/pull/3036/files#diff-6d5fabe09478e35a518f33e36f448a7e7c3e224ffa9a4a726fe8ccfeefd7b4f0R117 ?

gsouquet (changes requested):
Look like a good start 👍 
A few comments that probably need to be looked into, the biggest one being how we compare events with each others

weeman1337 (commented):
I checked out all branches and tested polls locally. Looks good :+1: 

Some minor comments. Mainly about strict typing.

src/models/poll.ts
kerryarchibald: Yep, `Room` emits the new event when creating the poll

src/models/poll.ts
kerryarchibald: Good point, I've replaced this with a `public readonly pollEvent` that's assigned in the constructor.

src/models/poll.ts
kerryarchibald: Yeah, I also thought ts would be able to infer that. Even wrapping this block in `if (this.responses) {...}` doesn't help

weeman1337: :man_shrugging:

src/models/poll.ts
kerryarchibald: I understand why this is probably more correct, but since the polls spec specifically mentions the poll end event timestamp, and comparing events by `origin_server_ts`, I think it is best to stay with strict timestamp comparison.

> Only a user's most recent vote (by `origin_server_ts`) is accepted, even if that event is invalid or redacted. Votes with timestamps after the poll has closed are ignored, as if they never happened.

> The m.poll.end's origin_server_ts determines when the poll closes exactly: if no valid end event is received, the poll is still open. If the poll is closed, only votes sent on or before that timestamp are considered, even if those votes are from before the start event. This is to handle clock drift over federation as gracefully as possible.

https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/polls/proposals/3381-polls.md

src/client.ts
```diff
@@ -9354,10 +9354,15 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 
     public processBeaconEvents(room?: Room, events?: MatrixEvent[]): void {
+        this.processAggregatedTimelineEvents(room, events);
+    }
+
+    public processAggregatedTimelineEvents(room?: Room, events?: MatrixEvent[]): void {
```
gsouquet, line 9360: Could we add a tsdoc comment to explain what this is supposed to do?

gsouquet (approved):
Looks good to me ✅

PR #3038, opened 2023-01-09
Tests for getVisibleRooms

andybalaam:
This is prep for an implementation of [MSC3946](https://github.com/matrix-org/matrix-spec-proposals/pull/3946)

---

### DIFF
--- a/spec/unit/matrix-client.spec.ts
+++ b/spec/unit/matrix-client.spec.ts
@@ -60,6 +60,7 @@ import { IOlmDevice } from "../../src/crypto/algorithms/megolm";
 import { QueryDict } from "../../src/utils";
 import { SyncState } from "../../src/sync";
 import * as featureUtils from "../../src/feature";
+import { StubStore } from "../../src/store/stub";
 
 jest.useFakeTimers();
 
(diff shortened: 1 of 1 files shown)


richvdh (approved):
nice and clear!

spec/unit/matrix-client.spec.ts
```diff
@@ -1967,4 +1968,82 @@ describe("MatrixClient", function () {
...
+
+        it("Returns all non-replaced rooms", () => {
+            const room1 = new Room("room1", client, "@carol:alexandria.example.com");
+            const room2 = new Room("room2", client, "@daryl:alexandria.example.com");
+            client.store = new StubStore();
+            client.store.getRooms = () => [room1, room2];
+            const rooms = client.getVisibleRooms();
+            expect(rooms).toContain(room1);
+            expect(rooms).toContain(room2);
+            expect(rooms).toHaveLength(2);
```
alunturner, line 2023: Does order matter here? Just wondering whether a single check of `.toEqual([room1, room2])` (or equivalent) would work or not.

alunturner (approved):
Looks good to me, just a single (non-blocking) question

spec/unit/matrix-client.spec.ts
alunturner: Too slow

andybalaam: I think order is unimportant, but maybe that would be nicer...

PR #3039, opened 2023-01-09
Factor out a (public) function to find a room's predecessor

andybalaam:
This is prep for an implementation of [MSC3946](https://github.com/matrix-org/matrix-spec-proposals/pull/3946)

---

### DIFF
--- a/spec/unit/room.spec.ts
+++ b/spec/unit/room.spec.ts
@@ -33,6 +33,7 @@ import {
     IRelationsRequestOpts,
     IStateEventWithRoomId,
     JoinRule,
+    MatrixClient,
     MatrixEvent,
     MatrixEventEvent,
     PendingEventOrdering,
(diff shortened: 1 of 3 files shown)


spec/unit/room.spec.ts
```diff
@@ -3225,4 +3226,52 @@ describe("Room", function () {
...
+                predecessor: { event_id: string; room_id: string } | undefined;
+            } = {
+                "creator": "@daryl:alexandria.example.com",
+                "predecessor": undefined,
+                "m.federate": true,
+                "room_version": "9",
+            };
+            if (predecessorRoomId) {
+                content.predecessor = {
+                    event_id: "spec_is_not_clear_what_id_this_is",
```
richvdh, line 3245: For the record: the spec says "The event ID of the last known event in the old room.". Generally it's the tombstone event.

richvdh (approved):
MSC3946 seems to be under fierce debate, so I'd be hesitant about implementing it right now.

Nevertheless factoring code out to functions is a win!

spec/unit/room.spec.ts
andybalaam: Thank you! Where did you find that? I couldn't see much info in the Client-Server API spec for room upgrades.

richvdh: on the spec for `m.room.create` (https://spec.matrix.org/v1.5/client-server-api/#mroomcreate)

PR #3040, opened 2023-01-09
Remove extensible events v1 field population on legacy events

turt2live:
With extensible events v2, affected events are now gated by a room version, so we don't need this code anymore. 

The proposal has generally moved away from mixing m.room.message with extensible fields as well.

---

## 🦖 Deprecations
 * Remove extensible events v1 field population on legacy events ([\#3040](https://github.com/matrix-org/matrix-js-sdk/pull/3040)).


SimonBrandner (approved):
Looks sane

PR #3045, opened 2023-01-10
Handle edits which are bundled with an event, per MSC3925

richvdh:
MSC3925 suggests that the server should no longer replace the body of edited events, so let's get ready for that. https://github.com/matrix-org/synapse/pull/14811 has already laid the groundwork, by bundling the whole of the edited event.

See also https://github.com/matrix-org/matrix-react-sdk/pull/9886 for some tests.

---

## ✨ Features
 * Handle edits which are bundled with an event, per MSC3925 ([\#3045](https://github.com/matrix-org/matrix-js-sdk/pull/3045)).


SimonBrandner (approved):
Looks sane, afaict

Issue #3030, opened 2023-01-05
`mkdirp` not found when running `npm run build`


pedro-nonfree:
`mkdirp` not found when running `npm run build`

I solved it running `npm install mkdirp`, so it should be added to dependencies

PR #3050, opened 2023-01-11
Avoid use of mkdirp

t3chguy:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/3030

---


andybalaam (approved):
Looks good.

PR #3056, opened 2023-01-12
Correctly handle limited sync responses by resetting the thread timeline

justjanne:
Type: Defect
Fixes: https://github.com/vector-im/element-web/issues/23952

---

## 🐛 Bug Fixes
 * Correctly handle limited sync responses by resetting the thread timeline ([\#3056](https://github.com/matrix-org/matrix-js-sdk/pull/3056)). Fixes vector-im/element-web#23952. Contributed by @justjanne.

### DIFF
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -361,7 +361,10 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
     /**
      * Reset the live timeline of all timelineSets, and start new ones.
      *
-     * <p>This is used when /sync returns a 'limited' timeline.
+     * <p>This is used when /sync returns a 'limited' timeline. 'Limited' means that there's a gap between the messages
+     * /sync returned, and the last known message in our timeline. In such a case, our live timeline isn't live anymore
+     * and has to be replaced by a new one. To make sure we can continue paginating our timelines correctly, we have to
+     * set new pagination tokens on the old and the new timeline.
      *
      * @param backPaginationToken -   token for back-paginating the new timeline
      * @param forwardPaginationToken - token for forward-paginating the old live timeline,
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -358,6 +358,21 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
         this.pendingReplyCount = pendingEvents.length;
     }
 
+    /**
+     * Reset the live timeline of all timelineSets, and start new ones.
+     *
+     * <p>This is used when /sync returns a 'limited' timeline.
+     *
+     * @param backPaginationToken -   token for back-paginating the new timeline
+     * @param forwardPaginationToken - token for forward-paginating the old live timeline,
+     * if absent or null, all timelines are reset, removing old ones (including the previous live
+     * timeline which would otherwise be unable to paginate forwards without this token).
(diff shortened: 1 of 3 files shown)


src/models/thread.ts
```diff
@@ -358,6 +358,55 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+     * We try to prevent that by detecting such cases and resetting the live timeline accordingly.
+     * @private
+     */
+    private async tryResetLiveTimeline(): Promise<void> {
+        // if we're still syncing, let the sync complete first
+        if (!this.client.isInitialSyncComplete()) {
+            return;
+        }
+
+        const lastKnownEvent = this.events[this.events.length - 1];
```
gsouquet, line 375: This won't work, because the message in the bundled relationship of the root event is only going to be the last event with a `thread` relation type.
However the last known event in the timeline could be a reaction that is a second degree relation to a thread event

src/models/thread.ts
```diff
@@ -358,6 +358,56 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
         this.pendingReplyCount = pendingEvents.length;
     }
 
+    /**
+     * Resets the live timeline if we're desynced.
+     *
+     * During connection failures, sometimes not all sync state for a thread is saved.
+     * After a reconnect, the thread will be missing some messages.
+     * We try to prevent that by detecting such cases and resetting the live timeline accordingly.
```
gsouquet, line 366:
```suggestion
     * After connection failures, the sync loop will not always return all missed messages
     * the thread cou.d be missing some messages. We try to detect those and reset the 
     * live timeline accordingly.
     * Comparing the `lastEvent` advertised in the root event bundled relationship to the 
     * the known thread relation in the timeline is usually a good signal
```

src/models/thread.ts
```diff
@@ -358,6 +358,56 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
+        // If they have, abort and try again.
+        if (lastKnownEvent !== this.lastReply((ev) => ev.isRelation(THREAD_RELATION_TYPE.name))
+            || lastActualEvent !== this.lastEvent
+            || !this.client.isInitialSyncComplete()) {
+            return this.tryResetLiveTimeline();
+        }
+
+        logger.warn("Resetting live timeline for thread " + this.id + " due to a desync");
+
+        // This turns the old livetimeline into a regular timeline, moves it, and creates a new livetimeline for us
```
gsouquet, line 403:
```suggestion
        // This turns the old live timeline into a regular timeline, moves it, and creates a new live timeline for us
```

justjanne:
For this to be ready, we need https://github.com/matrix-org/synapse/issues/14830 to be resolved first.

t3chguy:
@justjanne does that mean we also need to wait for a Synapse release and sufficient roll-out or?

justjanne:
@t3chguy That, as well as whether we can solve this in alternative ways, is what I'll try to figure out today. I suspect we'll ship a workaround for now with a TODO to remove it at a later time.

gsouquet (changes requested):
Code looks good ! But the tests are quite hard to follow as the code is not really documented and the test case names are not very explanatory

gsouquet (approved):
Looks good to me 👍

PR #3058, opened 2023-01-12
Fix failure to start in firefox private browser

richvdh:
Fixes https://github.com/vector-im/element-web/issues/24216

This is a quick fix so we can unblock the release. I'd like to investigate whether we can (a) report errors in this code, rather than swallowing them; (b) test that app load and login work with an indexeddb impl that behaves this way.

---

## 🐛 Bug Fixes
 * Fix failure to start in firefox private browser ([\#3058](https://github.com/matrix-org/matrix-js-sdk/pull/3058)). Fixes vector-im/element-web#24216.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1690,12 +1690,16 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
                         resolve(0);
                     };
                     req.onerror = (e): void => {
-                        logger.error(`Failed to remove IndexedDB instance ${dbname}: ${e}`);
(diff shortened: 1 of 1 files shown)


t3chguy (approved):
:shipit:

richvdh:
> @t3chguy added T-Defect and removed T-Task

By the way, the reason I labelled this as T-Task was that (afaik) the defect hasn't made it into a release of EW so it seemed odd to include it in the EW changelog as a fix. YMMV though...

ashughes1:
Does not look like this 100% fixed the issue. I have filed issue 24234.

PR #3060, opened 2023-01-12
Improvements to megolm integration tests

richvdh:
The megolm tests were making a few assumptions which they really shouldn't; in particular:

 * They were creating mock events with event_ids not starting `$`, and lacking `sender`, `origin_server_ts` and `unsigned` properties

 * They were not including the (now) required `keys.ed25519` property inside the ciphertext of an olm message.

These work ok currently, but they aren't really correct, and they cause problems when testing the new rust implementation.

---


andybalaam (approved):
Looks good, thanks!

PR #3061, opened 2023-01-12
Fix spurious "Decryption key withheld" messages

richvdh:
Fixes https://github.com/vector-im/element-web/issues/23803

Do not store `m.room_key.withheld` events with a code of `m.unavailable` in the crypto store: doing so causes a spurious "key withheld" message, and may overwrite more useful messages.

This also includes a change to ensure we update the error message when we receive an `m.room_key.withheld` - mostly driven by the need to test the change, but seems like an improvement anyway.

---

## 🐛 Bug Fixes
 * Fix spurious "Decryption key withheld" messages ([\#3061](https://github.com/matrix-org/matrix-js-sdk/pull/3061)). Fixes vector-im/element-web#23803.


alunturner (approved):
Looks good

PR #3062, opened 2023-01-12
Extract v1 extensible events polls types out of the events-sdk

turt2live:
These types are disappearing, so rescue them in the name of the react-sdk's feature set.

**Reviewer**: Please be aware that these are copy/pasted into the project, with alterations made to appease our linters.

---

### DIFF
--- a/src/extensible_events_v1/PollStartEvent.ts
+++ b/src/extensible_events_v1/PollStartEvent.ts
@@ -63,8 +63,8 @@ export class PollAnswerSubevent extends MessageEvent {
 
     /**
      * Creates a new PollAnswerSubevent from ID and text.
-     * @param id The answer ID (unique within the poll).
-     * @param text The text.
+     * @param id - The answer ID (unique within the poll).
+     * @param text - The text.
      * @returns The representative answer.
      */
     public static from(id: string, text: string): PollAnswerSubevent {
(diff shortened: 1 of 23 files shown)


src/extensible_events_v1/PollStartEvent.ts
```diff
@@ -0,0 +1,207 @@
...
+    ): PollStartEvent {
+        return new PollStartEvent({
+            type: M_POLL_START.name,
+            content: {
+                [M_TEXT.name]: question, // unused by parsing
+                [M_POLL_START.name]: {
+                    question: { [M_TEXT.name]: question },
+                    kind: kind instanceof NamespacedValue ? kind.name : kind,
+                    max_selections: maxSelections,
+                    answers: answers.map((a) => ({ id: makeId(), [M_TEXT.name]: a })),
```
kerryarchibald, line 197: Should we just use `uuid` here instead of more custom code?

kerryarchibald (approved):
LGTM

src/extensible_events_v1/PollStartEvent.ts
turt2live: probably, or one of the other thousand random string libraries available. Equally, hopefully polls are only unstable for another few months and we can delete this code soon. 

For now, I'm opting to leave this for a future PR to consider to narrowly keep the scope of this one to just airlifting the types over here.

PR #2531, opened 2022-07-18
Re-insert room IDs when decrypting bundled redaction events returned by `/sync`

duxovni:
## Checklist

* [x] Tests written for new code (and old code if feasible)
* [x] Linter and other CI checks pass
* [ ] ~~Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))~~ N/A

---
Here's what your changelog entry will look like:

## 🐛 Bug Fixes
 * Re-insert room IDs when decrypting bundled redaction events returned by `/sync` ([\#2531](https://github.com/matrix-org/matrix-js-sdk/pull/2531)). Contributed by @duxovni.

PR #3070, opened 2023-01-16
Remove brokenheaded encryption test

richvdh:
This test seemed to be testing the behaviour of decrypting redacted events, but
that seems... strange. A redaction event cannot be encrypted (at least, there
is no spec for it), and it should be impossible to decrypt a (correctly)
redacted event, because such an event will lack a `ciphertext` property.

This test is just sticking a "redacted_because" property into a regular event,
which is a bit of a nonsense.

---


t3chguy (approved):

richvdh:
For links: the test was added in https://github.com/matrix-org/matrix-js-sdk/pull/2531.

PR #3072, opened 2023-01-17
Poll model - validate end events

kerryarchibald:
~~Includes https://github.com/matrix-org/matrix-js-sdk/pull/3036~~

> Repeated end events are ignored - only the first (valid) closure event by origin_server_ts is counted.

> If a m.poll.end event is received from someone other than the poll creator or user with permission to redact other's messages in the room, the event must be ignored by clients due to being invalid.

## Checklist

---

## ✨ Features
 * Poll model - validate end events ([\#3072](https://github.com/matrix-org/matrix-js-sdk/pull/3072)). Contributed by @kerryarchibald.


SimonBrandner:
Any chance the base of this could be changed for easier review?

src/models/poll.ts
```diff
@@ -154,7 +154,7 @@ export class Poll extends TypedEventEmitter<Exclude<PollEvent, PollEvent.New>, P
     }
 
     /**
-     * Only responses made before the poll ended are valid
+     * Only responses made before the poll ended are vali
```
richvdh, line 157: looks like a mistake

richvdh (commented):
this seems to be mixing in a bunch of changes to threads - is that deliberate?

kerryarchibald:
@richvdh @SimonBrandner This is ready for another look!

src/models/poll.ts
```diff
@@ -172,4 +173,26 @@ export class Poll extends TypedEventEmitter<Exclude<PollEvent, PollEvent.New>, P
...
+         * only the first (valid) closure event by origin_server_ts is counted.
+         */
+        if (this.endEvent && this.endEvent.getTs() < endEvent.getTs()) {
+            return false;
+        }
+
+        /**
+         * MSC3381
+         * If a m.poll.end event is received from someone other than the poll creator or user with permission to redact
+         * other's messages in the room, the event must be ignored by clients due to being invalid.
```
richvdh, line 192:
```suggestion
         * others' messages in the room, the event must be ignored by clients due to being invalid.
```

src/models/poll.ts
```diff
@@ -172,4 +173,26 @@ export class Poll extends TypedEventEmitter<Exclude<PollEvent, PollEvent.New>, P
...
+         */
+        const roomCurrentState = this.room.currentState;
+        const endEventSender = endEvent.getSender();
+        return !!endEventSender && roomCurrentState.maySendRedactionForEvent(this.rootEvent, endEventSender);
```
richvdh, line 196: this doesn't seem to be quite the same thing as the comment above. It's possible that `endEventSender` is the same as the sender of the root event, and yet they lack permission to redact that create event (for example, because sending `m.room.redaction` events has been disabled).

richvdh (changes requested):

src/models/poll.ts
kerryarchibald: good catch, have added a `sender === creator` check here

src/models/poll.ts
```diff
@@ -189,10 +189,14 @@ export class Poll extends TypedEventEmitter<Exclude<PollEvent, PollEvent.New>, P
...
          * If a m.poll.end event is received from someone other than the poll creator or user with permission to redact
-         * other's messages in the room, the event must be ignored by clients due to being invalid.
+         * others' messages in the room, the event must be ignored by clients due to being invalid.
          */
         const roomCurrentState = this.room.currentState;
         const endEventSender = endEvent.getSender();
-        return !!endEventSender && roomCurrentState.maySendRedactionForEvent(this.rootEvent, endEventSender);
+        return (
+            !!endEventSender &&
+            (endEventSender === this.matrixClient.getSafeUserId() ||
```
richvdh, line 198: this is checking if the sender of the end event is the *current* user? Don't we want to check if it was the sender of the `rootEvent`?

```suggestion
            (endEventSender === this.rootEvent.getSender() ||
```

richvdh (approved):
lgtm otherwise

src/models/poll.ts
kerryarchibald: 🤦‍♀️ fixed and improved the tests to catch this. Thanks!

PR #3076, opened 2023-01-18
refactor: sliding sync: swap to lists-as-keys

kegsay:
Update the request/response API shape to match the latest MSC3575 version, which converts `lists` from being an array of list objects to being a map of list objects.

## Checklist

---


t3chguy (approved):
Needs a smidge more coverage

PR #3019, opened 2023-01-04
Handle outgoing requests from rust crypto SDK

richvdh:
The rust matrix-sdk-crypto has an `outgoingRequests()` method which we need to poll, and make the requested requests.

Part of https://github.com/vector-im/element-web/issues/21972.

---
This change is marked as an *internal change* (Task), so will not be included in the changelog.

PR #3083, opened 2023-01-20
Element-R: implement remaining OutgoingMessage request types

richvdh:
This is a follow-up to https://github.com/matrix-org/matrix-js-sdk/pull/3019: it implements the remaining two types of message types, now that rust SDK has sensibly-shaped types for them.

It turns out that I need to pull this stuff out to a separate class, so this PR also includes that.

Part of vector-im/element-web#21972.

---

### DIFF
--- a/spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
+++ b/spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
@@ -1,0 +1,112 @@
+/*
+Copyright 2023 The Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import MockHttpBackend from "matrix-mock-request";
+import { Mocked } from "jest-mock";
+import * as RustSdkCryptoJs from "@matrix-org/matrix-sdk-crypto-js";
+import {
+    KeysBackupRequest,
+    KeysClaimRequest,
+    KeysQueryRequest,
+    KeysUploadRequest,
+    SignatureUploadRequest,
+} from "@matrix-org/matrix-sdk-crypto-js";
+
+import { TypedEventEmitter } from "../../../src/models/typed-event-emitter";
+import { HttpApiEvent, HttpApiEventHandlerMap, MatrixHttpApi } from "../../../src";
+import { OutgoingRequestProcessor } from "../../../src/rust-crypto/OutgoingRequestProcessor";
+
+describe("OutgoingRequestProcessor", () => {
+    /** the OutgoingRequestProcessor implementation under test */
+    let processor: OutgoingRequestProcessor;
+
+    /** A mock http backend which processor is connected to */
+    let httpBackend: MockHttpBackend;
+
+    /** a mocked-up OlmMachine which processor is connected to */
+    let olmMachine: Mocked<RustSdkCryptoJs.OlmMachine>;
+
+    /** wait for a call to olmMachine.markRequestAsSent */
+    function awaitCallToMarkAsSent(): Promise<void> {
+        return new Promise((resolve, _reject) => {
+            olmMachine.markRequestAsSent.mockImplementationOnce(async () => {
+                console.log("received call to markAsSent");
+                resolve(undefined);
+            });
+        });
+    }
+
+    beforeEach(async () => {
+        httpBackend = new MockHttpBackend();
+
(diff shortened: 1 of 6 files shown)


spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+    let httpBackend: MockHttpBackend;
+
+    /** a mocked-up OlmMachine which processor is connected to */
+    let olmMachine: Mocked<RustSdkCryptoJs.OlmMachine>;
+
+    /** wait for a call to olmMachine.markRequestAsSent */
+    function awaitCallToMarkAsSent(): Promise<void> {
+        return new Promise((resolve, _reject) => {
+            olmMachine.markRequestAsSent.mockImplementationOnce(async () => {
+                console.log("received call to markAsSent");
```
andybalaam, line 48: eslint error is the console.log here.

spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+    /* simple requests that map directly to the request body */
+    const tests: Array<[any, "POST" | "PUT", string]> = [
+        [KeysUploadRequest, "POST", "https://example.com/_matrix/client/v3/keys/upload"],
+        [KeysQueryRequest, "POST", "https://example.com/_matrix/client/v3/keys/query"],
+        [KeysClaimRequest, "POST", "https://example.com/_matrix/client/v3/keys/claim"],
+        [SignatureUploadRequest, "POST", "https://example.com/_matrix/client/v3/keys/signatures/upload"],
+        [KeysBackupRequest, "PUT", "https://example.com/_matrix/client/v3/room_keys/keys"],
+    ];
+
+    for (const [RequestClass, expectedMethod, expectedPath] of tests) {
```
andybalaam, line 81: Can we use [test.each](https://jestjs.io/docs/api#testeachtablename-fn-timeout) here?

spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+        it(`should handle ${RequestClass.name}s`, async () => {
+            const testBody = '{ "foo": "bar" }';
+            const outgoingRequest = new RequestClass("1234", testBody);
+
+            const reqProm = processor.makeOutgoingRequest(outgoingRequest);
```
andybalaam, line 86: I'd like a comment along the lines of "This is what we are testing" here - it took me a while to figure out what this test is really doing.

spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+    RoomMessageRequest,
+    SignatureUploadRequest,
+    ToDeviceRequest,
+} from "@matrix-org/matrix-sdk-crypto-js";
+
+import { TypedEventEmitter } from "../../../src/models/typed-event-emitter";
+import { HttpApiEvent, HttpApiEventHandlerMap, MatrixHttpApi } from "../../../src";
+import { OutgoingRequestProcessor } from "../../../src/rust-crypto/OutgoingRequestProcessor";
+
+describe("OutgoingRequestProcessor", () => {
```
andybalaam, line 34: Is this a Rusty equivalent to an existing JS class? If so, it would be awesome if I could compare the JS behaviour with the Rust.

andybalaam (changes requested):
One suggestion and some questions, but looks good.

spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
richvdh: indeed, thanks.

spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
richvdh: yes, though giving the tests a sensible name automatically is a fiddle. Have had a go, anyway.

spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
richvdh: Yes, sorry about the lack of comments. Have added some. Is it clearer now?

spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
richvdh: No, there's not really any equivalent to this in the legacy implementation. 

This is where the rust sdk makes a request to the application for an outgoing HTTP request, so all the different request types are being channelled through a single bit of code. That's not really the case for the legacy implementation, which is much more distributed.

spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
andybalaam: Yes, thank you.

spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
andybalaam: Looks good

PR #3091, opened 2023-01-24
Remove video tracks on video mute without renegotiating

dbkr:
Builds on https://github.com/matrix-org/matrix-js-sdk/pull/3028 but without removing the sender and causing a renegotiation.

## Checklist

---

## ✨ Features
 * Remove video tracks on video mute without renegotiating ([\#3091](https://github.com/matrix-org/matrix-js-sdk/pull/3091)).

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -510,7 +510,7 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
     }
 
     private get hasUserMediaAudioSender(): boolean {
-        return this.transceivers.has(getTransceiverKey(SDPStreamMetadataPurpose.Usermedia, "audio"));
+        return Boolean(this.transceivers.get(getTransceiverKey(SDPStreamMetadataPurpose.Usermedia, "audio"))?.sender);
     }
 
     private get hasUserMediaVideoSender(): boolean {
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -387,6 +387,10 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
     private opponentSessionId?: string;
     public groupCallId?: string;
 
+    // Used to keep the timer for the delay before actually stopping our
+    // video track after muting (see setLocalVideoMuted)
+    private stopVideoTrackTimer?: ReturnType<typeof setTimeout>;
+
     /**
      * Construct a new Matrix Call.
      * @param opts - Config options.
@@ -480,7 +484,9 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
     }
 
     public get type(): CallType {
-        return this.hasLocalUserMediaVideoTrack || this.hasRemoteUserMediaVideoTrack ? CallType.Video : CallType.Voice;
+        // we may want to look for a video receiver here rather than a track to match the
+        // sender behaviour, although in practice they should be the same thing
+        return this.hasUserMediaVideoSender || this.hasRemoteUserMediaVideoTrack ? CallType.Video : CallType.Voice;
     }
 
     public get hasLocalUserMediaVideoTrack(): boolean {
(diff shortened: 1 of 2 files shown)


src/webrtc/call.ts
```diff
@@ -1361,22 +1364,51 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
...
+        // if we're muting video, set a timeout to stop & remove the video track so we release
+        // the camera. We wait a short time to do this because when we disable a track, WebRTC
+        // will send black video for it. If we just stop and remove it straight away, the video
+        // will just freeze which means that when we unmute video, the other side will briefly
+        // get a static frame of us from before we muted. This way, the still frame is just black.
+        // A very small delay is not always enough so the theory here is that it needs to be long
+        // enough for WebRTC to encode a frame: 120ms should be long enough even if we're only
+        // doing 10fps.
+        if (muted) {
+            this.stopVideoTrackTimer = setTimeout(() => {
```
robintown, line 1404: For the sake of timer hygene I'd suggest clearing the timeout in `terminate`

robintown (approved):

src/webrtc/call.ts
```diff
@@ -2465,6 +2498,8 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
     private async terminate(hangupParty: CallParty, hangupReason: CallErrorCode, shouldEmit: boolean): Promise<void> {
         if (this.callHasEnded()) return;
 
+        if (this.stopVideoTrackTimer !== undefined) clearTimeout(this.stopVideoTrackTimer);
```
SimonBrandner, line 2501: Could this please be at the same place as the other `clearTimeout()`s

src/webrtc/call.ts
```diff
@@ -503,6 +509,14 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
         });
     }
 
+    private get hasUserMediaAudioSender(): boolean {
+        return this.transceivers.has(getTransceiverKey(SDPStreamMetadataPurpose.Usermedia, "audio"));
+    }
+
+    private get hasUserMediaVideoSender(): boolean {
+        return this.transceivers.has(getTransceiverKey(SDPStreamMetadataPurpose.Usermedia, "video"));
+    }
```
SimonBrandner, line 518: This seems to only be checking for `transceiver`s not `sender`s - that doesn't feel right

src/webrtc/call.ts
```diff
@@ -1404,7 +1436,7 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
             return this.isMicrophoneMuted();
         }
 
-        if (!this.hasLocalUserMediaAudioTrack && !muted) {
+        if (!this.hasUserMediaAudioSender && !muted) {
```
SimonBrandner, line 1439: The fact we have an audio sender doesn't mean we have a track to enable, right?

SimonBrandner (commented):
At first, I had trouble understanding what this is exactly trying to do but I think I am starting to see now but a closer PR description could be helpful here

src/webrtc/call.ts
dbkr: Oops, yes

src/webrtc/call.ts
dbkr: True, it should probably be checking the sender

src/webrtc/call.ts
dbkr: True, we should probably check both

src/webrtc/call.ts
```diff
@@ -480,7 +484,9 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
     }
 
     public get type(): CallType {
-        return this.hasLocalUserMediaVideoTrack || this.hasRemoteUserMediaVideoTrack ? CallType.Video : CallType.Voice;
+        // we may want to look for a video receiver here rather than a track to match the
+        // sender behaviour, although in practice they should be the same thing
+        return this.hasUserMediaVideoSender || this.hasRemoteUserMediaVideoTrack ? CallType.Video : CallType.Voice;
```
SimonBrandner, line 489: The sender will be there even if the track isn't, so this could return `CallType.Video` even if we have no video in the call, afaict

dbkr: I'd argue this is correct - a video call is still a video call if everyone has their video muted, although really the lines of what is a voice and video call are very blurry.

src/webrtc/call.ts
```diff
@@ -503,6 +509,14 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
         });
     }
 
+    private get hasUserMediaAudioSender(): boolean {
+        return this.transceivers.has(getTransceiverKey(SDPStreamMetadataPurpose.Usermedia, "audio"));
```
SimonBrandner, line 513: This still isn't checking the sender

dbkr: Oops, yes - missed that one

SimonBrandner (approved):

PR #3092, opened 2023-01-24
Implement MSC3952: intentional mentions

clokep:
## Checklist

-   [x] ~~Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))~~

See matrix-org/matrix-react-sdk#9983 & matrix-org/matrix-spec-proposals#3952.

Fixes vector-im/element-web#24376

---

## ✨ Features
 * Implement MSC3952: intentional mentions ([\#3092](https://github.com/matrix-org/matrix-js-sdk/pull/3092)). Fixes vector-im/element-web#24376.

### DIFF
--- a/spec/unit/pushprocessor.spec.ts
+++ b/spec/unit/pushprocessor.spec.ts
@@ -1,6 +1,6 @@
 import * as utils from "../test-utils/test-utils";
 import { IActionsObject, PushProcessor } from "../../src/pushprocessor";
-import { ConditionKind, EventType, IContent, MatrixClient, MatrixEvent, PushRuleActionName } from "../../src";
+import { ConditionKind, EventType, IContent, MatrixClient, MatrixEvent, PushRuleActionName, RuleId } from "../../src";
 
 describe("NotificationService", function () {
     const testUserId = "@ali:matrix.org";
@@ -48,7 +48,7 @@ describe("NotificationService", function () {
         credentials: {
             userId: testUserId,
         },
-        supportsIntentionalMentions: () => false,
+        supportsIntentionalMentions: () => true,
         pushRules: {
             device: {},
             global: {
@@ -712,6 +712,37 @@ describe("NotificationService", function () {
                 rule: msc3914RoomCallRule,
             });
         });
+    });
+
+    describe("test intentional mentions behaviour", () => {
+        it.each([RuleId.ContainsUserName, RuleId.ContainsDisplayName, RuleId.AtRoomNotification])(
+            "disable legacy rule: %s",
+            (ruleId) => {
+                const rule = {
+                    rule_id: ruleId,
+                    actions: [],
+                    conditions: [],
+                    default: false,
+                    enabled: true,
+                };
+                expect(pushProcessor.ruleMatchesEvent(rule, testEvent)).toBe(true);
+
+                // Add the mentions property to the event and the rule is now disabled.
+                testEvent = utils.mkEvent({
+                    type: "m.room.message",
+                    room: testRoomId,
+                    user: "@alfred:localhost",
+                    event: true,
+                    content: {
+                        "body": "",
+                        "msgtype": "m.text",
+                        "org.matrix.msc3952.mentions": {},
+                    },
+                });
+
+                expect(pushProcessor.ruleMatchesEvent(rule, testEvent)).toBe(false);
+            },
+        );
     });
 });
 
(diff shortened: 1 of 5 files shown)


spec/unit/pushprocessor.spec.ts
```diff
@@ -712,6 +713,37 @@ describe("NotificationService", function () {
             });
         });
     });
+
+    describe("test intentional mentions behaviour", () => {
+        it.each([RuleId.ContainsUserName, RuleId.ContainsDisplayName, RuleId.AtRoomNotification])(
+            "disable legacy rule: %s",
```
andybalaam, line 719: Can we find a fuller description? It took me a while to work out what this test does.

How about "Rule %s matches in legacy mode and does not in intentional mentions mode"

(Or something better...)

src/pushprocessor.ts
```diff
@@ -70,6 +71,36 @@ const DEFAULT_OVERRIDE_RULES: IPushRule[] = [
...
         actions: [PushRuleActionName.DontNotify],
     },
+    {
+        rule_id: RuleId.IsUserMention,
+        default: true,
+        enabled: true,
+        conditions: [
+            {
+                kind: ConditionKind.EventPropertyContains,
+                key: "content.org\\.matrix\\.msc3952\\.mentions.user_ids",
```
andybalaam, line 81: Do you have a plan for when this lands? E.g. in some places we use `matrix-js-sdk/src/NamespacedValue.ts` 's `UnstableValue`, which is intended to help make the transition easier. Not convinced it would help here...

src/pushprocessor.ts
```diff
@@ -174,9 +206,21 @@ export class PushProcessor {
...
             const existingRule = globalOverrides.find((r) => r.rule_id === override.rule_id);
 
+            // Dynamically add the user ID as the value for the is_user_mention rule.
+            if (override.rule_id === RuleId.IsUserMention) {
+                // If the user ID wasn't provided, skip the rule.
+                if (!userId) {
+                    continue;
+                }
+
+                override = JSON.parse(JSON.stringify(override)); // deep clone
```
andybalaam, line 219: I think our support policy still says last 3 versions of Safari, which rules out https://developer.mozilla.org/en-US/docs/Web/API/structuredClone which was introduced in Safari 15.4. Otherwise I'd be recommending it :-(

src/pushprocessor.ts
```diff
@@ -174,9 +206,21 @@ export class PushProcessor {
...
+                const conditions = override.conditions!;
+                conditions[0].value = userId;
```
andybalaam, line 221: Why not:

```
override.conditions?.[0].value = userId;
```

To me it's a little clearer that we are modifying override.

src/pushprocessor.ts
```diff
@@ -174,9 +206,21 @@ export class PushProcessor {
 
         // Merge the client-level defaults with the ones from the server
         const globalOverrides = newRules.global.override;
-        for (const override of DEFAULT_OVERRIDE_RULES) {
+        for (let override of DEFAULT_OVERRIDE_RULES) {
```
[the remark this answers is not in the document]
andybalaam: (And doing `else { override = originalOverride }` after the `if (override.rule_id === RuleId.IsUserMention )`

andybalaam (changes requested):
A few very minor comments and questions.

src/pushprocessor.ts
clokep: FTR I copied this code from a few lines above. 😄

src/pushprocessor.ts
clokep: Likely because I didn't realize you could use `?.` or `!.` in front of `[0]`.

src/pushprocessor.ts
clokep: I don't have a strong plan, do you know what's been done for push rules in the past? I think the transition will be relatively difficult here, we could maybe do something like:

1. Server side, just jump to the new rule (so remove the unstable `/versions` flag).
2. That should disable the current rules.
3. Client side, add support for the stable rule -- maybe based on some flag or spec version? 🤷 
4. Support unstable version if the unstable flag is seen.

I'll try to think a bit more about this though and see if I can come up with a better rollout plan.

src/pushprocessor.ts
andybalaam: Yeah it is a little surprising, but useful here.

src/pushprocessor.ts
andybalaam: Wow, yeah it's much more difficult than I was thinking about for this comment. Will be happy to discuss the plan when we get to it, but no need to delay this change for that conversation.

andybalaam (approved):

PR #3108, opened 2023-01-26
Combine sonarqube CI jobs into Tests to avoid dependency issue

hughns:
Might resolve "no artifacts found" issue

## Checklist


.github/workflows/tests.yml
```diff
@@ -59,3 +59,33 @@ jobs:
                   path: |
                       coverage
                       !coverage/lcov-report
+
```
andybalaam, line 62: Please could you update the name of this job to "Tests and SonarQube" or similar?

Thank you!

andybalaam (approved):

t3chguy (changes requested):
We cannot do this as it'll break Community PRs which don't have access to the Sonarcloud secret credential, hence the workflow_run requirement.

andybalaam (changes requested):
As per Michael's comment, we can't do this :-(

t3chguy:
https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow for the spiel

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

hughns:
Yeah, I came to the same conclusion that this approach wouldn't work and that there was a good reason for it having been split out in the first place. 😞

PR #3112, opened 2023-01-28
Don't use ideal/exact in case of screen-sharing

SimonBrandner:
---

### DIFF
--- a/src/webrtc/mediaHandler.ts
+++ b/src/webrtc/mediaHandler.ts
@@ -23,17 +23,19 @@ import { logger } from "../logger";
 import { MatrixClient } from "../client";
 
 const isWebkit = (): boolean => Boolean(navigator.webkitGetUserMedia);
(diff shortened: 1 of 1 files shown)


robintown:
What does it mean when the constraint is just a number? MDN is failing me.

SimonBrandner:
> What does it mean when the constraint is just a number? MDN is failing me.

Likewise, but it seems to work fine for any size of window

src/webrtc/mediaHandler.ts
```diff
@@ -442,7 +442,9 @@ export class MediaHandler extends TypedEventEmitter<
         return {
             audio: audio ?? false,
             video: {
-                ...getDimensionConstraints(1920, 1080, 30),
```
robintown, line 445: Why can't we use ideal/exact constraints? I think this is worth a comment.

PR #3114, opened 2023-01-30
Move the logic of roomPredecessor into the RoomState class

andybalaam:
Part of https://github.com/vector-im/element-web/issues/24323

---

### DIFF
--- a/src/models/room-state.ts
+++ b/src/models/room-state.ts
@@ -986,7 +986,7 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
             const predecessorEvent = this.getStateEvents(EventType.RoomPredecessor, "");
             if (predecessorEvent) {
                 const roomId = predecessorEvent.getContent()["predecessor_room_id"];
-                if (roomId) {
+                if (typeof roomId === "string") {
                     return { roomId, eventId: null };
                 }
             }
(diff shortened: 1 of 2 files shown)


src/models/room-state.ts
```diff
@@ -964,6 +964,48 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
...
+        msc3946ProcessDynamicPredecessor = false,
+    ): { roomId: string; eventId: string | null } | null {
+        // Note: the tests for this function are against Room.findPredecessor,
+        // which just calls through to here.
+
+        if (msc3946ProcessDynamicPredecessor) {
+            const predecessorEvent = this.getStateEvents(EventType.RoomPredecessor, "");
+            if (predecessorEvent) {
+                const roomId = predecessorEvent.getContent()["predecessor_room_id"];
+                if (roomId) {
```
robintown, line 989:
```suggestion
                if (typeof roomId === "string") {
```

For safety, since we can't trust the contents of events to match the types we want

src/models/room-state.ts
```diff
@@ -964,6 +964,48 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
...
+        }
+
+        const createEvent = this.getStateEvents(EventType.RoomCreate, "");
+        if (createEvent) {
+            const predecessor = createEvent.getContent()["predecessor"];
+            if (predecessor) {
+                const roomId = predecessor["room_id"];
+                if (roomId) {
+                    const eventId = predecessor["event_id"] || null;
+                    return { roomId, eventId };
```
robintown, line 1002: These two are also missing type checks

src/models/room-state.ts
andybalaam: Thank you, fixed in 558392c

src/models/room-state.ts
andybalaam: Thank you, fixed in 558392c

gsouquet (approved):
Looks good to me 💯

PR #3115, opened 2023-01-30
Cleanup pre MSC3773 thread unread notif logic

gsouquet:
A lot of logic regarding servers that do not support thread notifications (pre MSC3773) exists. This PR removes some of it. 
Comes with its react-sdk counterpart too.

## Checklist

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -4840,8 +4840,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             $eventId: event.getId()!,
         });
 
-        const supportsThreadRR = this.canSupport.get(Feature.ThreadUnreadNotifications) !== ServerSupport.Unsupported;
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -7045,9 +7044,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         const serverVersions = await this.serverVersionsPromise;
         this.canSupport = await buildFeatureSupportMap(serverVersions);
 
-        // We can set flag values to use their stable or unstable version
-        const support = this.canSupport.get(Feature.ThreadUnreadNotifications);
-        UNREAD_THREAD_NOTIFICATIONS.setPreferUnstable(support === ServerSupport.Unstable);
+        UNREAD_THREAD_NOTIFICATIONS.setPreferUnstable(false);
```
richvdh, line 7047: is this not the default? Can we remove this line?

PR #3117, opened 2023-01-31
Element R: Fix obscure errors when we fail to decrypt to-device events

richvdh:
Previously, if we failed to decrypt a to-device event, we would raise an "expected a string" error when we later tried to decrypt it as a room event. This at least makes the error clearer.

---


gsouquet (approved):
Looks sane! Thank you

PR #3118, opened 2023-01-31
Don't require e2e keys for foci

dbkr:
Since we don't speak e2e to them anyway

## Checklist

---

## 🐛 Bug Fixes
 * Don't require e2e keys for foci ([\#3118](https://github.com/matrix-org/matrix-js-sdk/pull/3118)).

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -676,6 +676,9 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
     private async initOpponentCrypto(): Promise<void> {
         if (!this.opponentDeviceId) return;
         if (!this.client.getUseE2eForGroupCall()) return;
+        // We (currently) don't speak e2e with foci. It's debateable whether there would be
+        // any benefit in doing so.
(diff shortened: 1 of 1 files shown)


src/webrtc/call.ts
```diff
@@ -676,6 +676,9 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
     private async initOpponentCrypto(): Promise<void> {
         if (!this.opponentDeviceId) return;
         if (!this.client.getUseE2eForGroupCall()) return;
+        // We (currently) don't speak e2e with foci. It's debateable whether there would be
+        // any benefit in doing so.
+        if (this.isFocus) return;
```
EnricoSchw, line 681: Just to be sure. We explicitly do not want encrypted messages with the SFU?

EnricoSchw (approved):

src/webrtc/call.ts
dbkr: Not currently - we do the same check later to see if we're talking to a focus and send unencrypted messages.

PR #3121, opened 2023-02-02
Element-R: Fix invite processing

richvdh:
Currently, whenever we receive an invite on element R, it crashes the sync loop. A quick fix to make it not do that.

---

## 🐛 Bug Fixes
 * Element-R: Fix invite processing ([\#3121](https://github.com/matrix-org/matrix-js-sdk/pull/3121)).

### DIFF
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -1254,11 +1254,12 @@ export class SyncApi {
 
             const inviter = room.currentState.getStateEvents(EventType.RoomMember, client.getUserId()!)?.getSender();
 
-            if (client.isCryptoEnabled()) {
-                const parkedHistory = await client.crypto!.cryptoStore.takeParkedSharedHistory(room.roomId);
+            const crypto = client.crypto;
+            if (crypto) {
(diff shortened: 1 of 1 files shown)


src/sync.ts
```diff
@@ -1254,11 +1254,12 @@ export class SyncApi {
 
             const inviter = room.currentState.getStateEvents(EventType.RoomMember, client.getUserId()!)?.getSender();
 
-            if (client.isCryptoEnabled()) {
-                const parkedHistory = await client.crypto!.cryptoStore.takeParkedSharedHistory(room.roomId);
```
richvdh, line 1258: For context, `client.isCryptoEnabled` checks if `client.cryptoBackend` is defined, rather than `client.crypto`. 

`CryptoBackend` doesn't have a `cryptoStore`. Eventually we can look at implementing `takeParkedSharedHistory` on `CryptoBackend`, but not today.

gsouquet (approved):
🚀

PR #3122, opened 2023-02-02
Element-R: implement encryption of outgoing events

richvdh:
This PR wires up the Rust-SDK into the event encryption path. I'm sorry it's a bit meaty - I've tried to break it down into bite-size commits and I recommend reviewing them in turn.

I'm also sorry that it's not as well tested as I would like, and in particular there are no integration tests. I've had trouble getting the tests to work for reasons including, but not limited to, https://github.com/matrix-org/matrix-rust-sdk/issues/1443. However, empirically it works in a browser, and I'm keen to see this shipped. I'll return to the tests soon.

---

## ✨ Features
 * Element-R: implement encryption of outgoing events ([\#3122](https://github.com/matrix-org/matrix-js-sdk/pull/3122)).

### DIFF
--- a/src/rust-crypto/RoomEncryptor.ts
+++ b/src/rust-crypto/RoomEncryptor.ts
@@ -20,6 +20,7 @@ import { IContent, MatrixEvent } from "../models/event";
 import { Room } from "../models/room";
 import { logger, PrefixedLogger } from "../logger";
 import { KeyClaimManager } from "./KeyClaimManager";
+import { RoomMember } from "../models/room-member";
 
 /**
  * RoomEncryptor: responsible for encrypting messages to a given room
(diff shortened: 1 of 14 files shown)


src/rust-crypto/RoomEncryptor.ts
```diff
@@ -0,0 +1,129 @@
...
+        await this.ensureEncryptionSession();
+
+        const encryptedContent = await this.olmMachine.encryptRoomEvent(
+            new RoomId(this.room.roomId),
+            event.getType(),
+            JSON.stringify(event.getContent()),
+        );
+
+        event.makeEncrypted(
+            "m.room.encrypted",
```
t3chguy, line 123: EventType enum?

t3chguy (commented):
Looks sane, could we hit 80% covg on the bits outside of rust-crypto at least?

src/rust-crypto/RoomEncryptor.ts
richvdh: good point, done.

richvdh:
> Looks sane, could we hit 80% covg on the bits outside of rust-crypto at least?

mmkay. I believe we're now there: 14 lines are changing outside rust-crypto, of which 12 are tested.

t3chguy (approved):

PR #3123, opened 2023-02-03
Improve WebRTC logging

SimonBrandner:
This should hopefully make debugging calls at least a bit easier 

# Please review carefully, we don't want our own logs to confuse us

---

### DIFF
--- a/spec/unit/webrtc/groupCall.spec.ts
+++ b/spec/unit/webrtc/groupCall.spec.ts
@@ -151,6 +151,13 @@ describe("Group Call", function () {
             },
         );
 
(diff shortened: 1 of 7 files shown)


EnricoSchw (approved):

dbkr (approved):

dbkr:
Merging this anyway because it's improving our logging and we shouldn't need to add tests cases to all the obscure error cases just to have better logging.

ejke:
Although it is said that this should be an internal change, it affects the users - when using matrix-js-sdk in service, those logs are generated and fill up the console. Please change to keep those logs internally and not trickle down to end users.

https://github.com/matrix-org/matrix-js-sdk/issues/3342

PR #3130, opened 2023-02-06
Fix notification counts for encrypted rooms with ignored event rules

Half-Shot:
Fixes https://github.com/vector-im/element-web/issues/24448

## Checklist

-   [N/A] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))

Notes: Fix a bug where events in encrypted rooms would sometimes erroneously increment the total unread counter after being processed locally.

This PR aims to fix the problem where currently events with pushrules calculated clientside that emit a `{notify: false}` are still calculated as increasing the unread counter in a room. This refactors one of the critical functions for dealing with it, so please review with care.

In terms of the changes here, we:

- Stop relying on the Total notification count entirely for encrypted rooms because unlike with the highlight count where the client generally has a higher count than the server, the total count is most likely to be lower in reality. Therefore, we ignore the server and calculate our own every time.
- There are now a lot of comments rationalising the various if blocks.
- I've slightly adjusted some of the ordering in the `fixNotificationCountOnDecryption` function to avoid doing so many non-null assertions.

---

## 🐛 Bug Fixes
 * Fix a bug where events in encrypted rooms would sometimes erroneously increment the total unread counter after being processed locally. ([\#3130](https://github.com/matrix-org/matrix-js-sdk/pull/3130)). Fixes vector-im/element-web#24448. Contributed by @Half-Shot.


spec/unit/notifications.spec.ts
```diff
@@ -134,7 +134,7 @@ describe("fixNotificationCountOnDecryption", () => {
 
         fixNotificationCountOnDecryption(mockClient, event);
 
-        expect(room.getUnreadNotificationCount(NotificationCountType.Total)).toBe(2);
+        expect(room.getUnreadNotificationCount(NotificationCountType.Total)).toBe(3);
```
Half-Shot, line 137: I think this is meant to be 3. The test sets up an event with a `getPushActions()` value of `{notify: false}` but `getPushActionsForEvent` returns `{notify: true}`. Ergo, I think this is meant to increase the room's total count.

spec/unit/notifications.spec.ts
```diff
@@ -154,7 +154,7 @@ describe("fixNotificationCountOnDecryption", () => {
 
         fixNotificationCountOnDecryption(mockClient, threadEvent);
 
-        expect(room.getThreadUnreadNotificationCount(THREAD_ID, NotificationCountType.Total)).toBe(1);
+        expect(room.getThreadUnreadNotificationCount(THREAD_ID, NotificationCountType.Total)).toBe(2);
```
Half-Shot, line 157: Ditto as above.

spec/unit/notifications.spec.ts
```diff
@@ -134,7 +134,7 @@ describe("fixNotificationCountOnDecryption", () => {
```
Half-Shot, line 134: These tests are broken and wrong right now, I'm refining the logic first.

src/client.ts
```diff
@@ -9483,64 +9483,77 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+    } else {
+        hasReadEvent = room.hasUserReadEvent(ourUserId, eventId);
+    }
+
+    if (hasReadEvent) {
+        // If the event has been read, ignore it.
+        return;
+    }
+
+    if (oldHighlight !== newHighlight || currentHighlightCount > 0) {
```
gsouquet, line 9526: Should it also check for `actions.notify` or does it affect, `total` and `total` only?

gsouquet (approved):
Otherwise looks sensible!
Sorry that took a little while for me to review

src/client.ts
Half-Shot: I believe this is handled by the if statement below. This just checks the highlights specifically.

t3chguy:
Context https://gist.github.com/Half-Shot/f9501916363894761a1659250aa25181

PR #3134, opened 2023-02-07
Implement MSC3873 to handle escaped dots in push rule keys

### LINKED ISSUE #1454: Pushrules with Dotted Value keys break
I wanted to configure a pushrule ignoring "m.replace" events.
Therefore i used this configuration:

```
{
    scope: 'global',
    kind: 'override',
    ruleId: 'rule.no_edits',
    body: {
        actions: [
            'dont_notify'
        ],
        conditions: [
            {
                'kind': 'event_match',
                'key': 'type',
                'pattern': 'm.room.message'
            },
            {
                'kind': 'event_match',
                'key': 'content.m.relates_to.rel_type',
                'pattern': 'm.replace'
            },
        ]
    },
}
```

But the behaviour of Synapse and the matrix-js-sdk didn't match.
The condition key `content.m.relates_to.rel_type` is interpreted in the valueDottedForKey method wrong:

Actual part splitting:
`['content', 'm', 'relates_to', 'rel_type']` 
Needed splitting:
`['content', 'm.relates_to', 'rel_type']`

https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/pushprocessor.js#L292
```
        const parts = key.split('.');
```

Maybe the approach of just splitting by the dot is undesirable.
Synapse takes the apporach the other way around, converting the nested object to a flat object and afterwards just accessing by the key.

Pseudo:
```js
function accessDottedKey(evt, key) {
   flatEvt = makeFlat(evt);
   return flatEvt[key];
}
```

clokep:
This implements MSC3873 which allows for disambiguating the `key` used in `event_match` between fields that have a `.` in them vs. using `.` as a field separator.

The MSC doesn't have any backwards compatibility clauses as none of the default push rules are affected. Some custom rules might behave differently after this change. This would fix #1454.

## Checklist

-   [x] ~~Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))~~

---

## ✨ Features
 * Implement MSC3873 to handle escaped dots in push rule keys ([\#3134](https://github.com/matrix-org/matrix-js-sdk/pull/3134)). Fixes undefined/matrix-js-sdk#1454.


spec/unit/pushprocessor.spec.ts
```diff
@@ -287,6 +287,7 @@ describe("NotificationService", function () {
         expect(actions.tweaks.highlight).toEqual(true);
     });
 
+    // TODO: This is not spec compliant behaviour.
```
andybalaam, line 290: Please can we have a link to the relevant part of the spec and a sentence on what we do wrong here?

spec/unit/pushprocessor.spec.ts
```diff
@@ -493,4 +498,77 @@ describe("NotificationService", function () {
...
+
+        expect(
+            pushProcessor.ruleMatchesEvent(
+                {
+                    rule_id: "rule1",
+                    actions: [],
+                    conditions: [
+                        {
+                            kind: ConditionKind.EventMatch,
+                            key: "content.m\\.test.foo",
```
andybalaam, line 524: Worth also checking that `content.m.test.foo` does not match?

src/pushprocessor.ts
```diff
@@ -433,8 +433,74 @@ export class PushProcessor {
         return PushProcessor.cachedGlobToRegex[glob];
     }
 
+    /**
+     * Parse a dotted key into parts following the escape rules.
+     *
+     * @param str - The key of the push rule condition: a dotted field.
+     * @returns The unescaped parts to fetch.
+     */
+    private partsFotDottedKey(str: string): string[] {
```
andybalaam, line 442: Should this name be `PartsForDottedKey` (Fot->For)?

src/pushprocessor.ts
```diff
@@ -433,8 +433,74 @@ export class PushProcessor {
...
+
+        // The current field and whether the previous character was the escape
+        // character (a backslash).
+        let part = "";
+        let escaped = false;
+
+        // Iterate over each character, and decide whether to append to the current
+        // part (following the escape rules) or to start a new part (based on the
+        // field separator).
+        for (let i = 0; i < str.length; ++i) {
```
andybalaam, line 453: This iterates code units (I think...). I suspect we want to iterate code points, which I'm fairly sure we get with `for (const ch of str) {`.

src/pushprocessor.ts
```diff
@@ -433,8 +433,74 @@ export class PushProcessor {
...
+                // A backslash adds no characters, but starts an escape sequence.
+                escaped = true;
+            } else {
+                // Otherwise, just add the current character.
+                part += c;
+            }
+        }
+
+        // Ensure the final part is included, if there is one.
+        if (part.length) {
```
andybalaam, line 485: Technically if part is empty, pushing it doesn't do any harm, right? So I might be tempted to remove the if (but just a thought).

src/pushprocessor.ts
```diff
@@ -433,8 +433,74 @@ export class PushProcessor {
...
+        const result = [];
+
+        // The current field and whether the previous character was the escape
+        // character (a backslash).
+        let part = "";
+        let escaped = false;
+
+        // Iterate over each character, and decide whether to append to the current
+        // part (following the escape rules) or to start a new part (based on the
+        // field separator).
```
andybalaam, line 452: Do we know how hot the paths are that run this code? We could be introducing a performance problem here, and might need to e.g. try for a regex-based solution.

src/pushprocessor.ts
andybalaam, line 442: I'd really like to see some unit tests for this specific function. It definitely has some edge cases and feels very testable... E.g. what if a string ends with \ or . or \. ?

andybalaam (changes requested):
Some questions, and an ask for more unit tests.

src/pushprocessor.ts
clokep: This would run once per push rule per message, so fairly hot. I briefly attempted a regular expression solution and couldn't get it working.

Which aspect are you concerned about? The string concatenation or the iterating each character or something else?

andybalaam: It might well be fine, but the iteration per-character made me wonder what the perf impact would be. There's no guarantee that a regular expression would be better, but I imagine regex engines are going to be good at iterating characters so maybe.

If we had nice unit tests for this function it would make it nice and easy to swap out the impl :-)

andybalaam: For a regex solution I was thinking something dumb like

```
const r = /\\\./g;
const key = "foo.bar\\.baz";
const tmpkey = key.replaceAll(r, "__PLACEHOLDER__");
const parts = tmpkey.split(".").map(p => p.replaceAll("__PLACEHOLDER__", "."));
```

I haven't thought about edge cases, double-escaping etc.

andybalaam: No idea whether it would be faster, and I generally prefer real parsing to solve a real parsing problem, but I am just worried about introducing a perf bomb :-)

clokep: I've usually found the double escaping to break horribly with regular expression solutions, will take another look though!

andybalaam: Yeah, it does feel dodgy.

src/pushprocessor.ts
clokep: Talking to @andybalaam out of band and the best way to do this is to make it public, but put a warning in the comment saying that it isn't for public consumption (or something to that effect).

clokep: Will do, thanks for the hint! 👍

src/pushprocessor.ts
clokep: Good spot -- this was a bug. There's no reason you can't have empty parts here, e.g. `m..foo` to match:

```json
{
  "m": {
    "": {
      "foo"
    }
  }
}
```

Definitely a (silly) edgecase, but it simplifies the code to support it!

andybalaam: Brilliant! I didn't spot that, but am happy I perturbed your brain sufficiently for you to :-)

src/pushprocessor.ts
clokep: I went through a few iterations.

| Commit | Rate | Comment  |
|---|---|---|
| develop | 1.0 | Baseline |
| 7e11af82506d0393233047276c24784496b0ee60 | 0.25 | Looping solution |
| 97fe8e944506be01c95437c76617186fff3d79e8 | 0.20 | Regex solution |
| 322913ec114aa91533dbaffe5c614b857b124e15 | 0.50 | Split on regex solution |

Let me know what you think of them. I suspect the latest one is the easiest to understand?

clokep: (Those timings are fairly unscientific. 🤷 )

clokep: Although it seems SonarCloud doesn't like the splitting solution...

andybalaam: Given a good set of unit tests, we are reasonably confident that all 3 are correct, I'd go for the looping solution for 2 main reasons:

1. It's the most direct implementation of the problem, so we can feel most confident it does what is intended. The regexes feel like hacks in comparison.
2. It performs approximately as fast as the others.

But, I think you're closest to it, so would like to hear what you think.

In terms of whether the performance is good enough, I guess the point is that we need to do this functionality, so we're going to need to find a way to make it work, so choosing something that performs about as well as we can is our best first guess.

andybalaam (changes requested):
One question, and whatever is needed for the outcome of the choice of implementation discussion.

src/pushprocessor.ts
andybalaam: (I reviewed the looping solution and am happy with it.)

clokep: I had an ah-ha moment last night about this -- we should be cached this on the push rule and then we only need to evaluate it once per (`event_match`) rule, not once per event per rule.

With that in mind, I definitely think the direct implementation is easiest to understand (and I'm most confident it isn't hitting other edge-cases).

andybalaam: Brilliant, good thinking.

src/pushprocessor.ts
```diff
@@ -123,6 +123,12 @@ export class PushProcessor {
      */
     public constructor(private readonly client: MatrixClient) {}
 
+    /**
+     * Maps the original key from the push rules to a list of property names
+     * after unescaping.
+     */
+    private readonly parsedKeys = new Map<string, string[]>();
```
clokep, line 130: Originally I was going to add some sort of cached property to the `IEventMatchCondition`, but that's just an interface so this doesn't seem doable? Let me know if I'm wrong!

Downside of this approach is that if you delete a push rule then the key would still be cached.

t3chguy: You could add a field to IEventMatchCondition which is optional, e.g. `memoizedCheckFn?(ev: MatrixEvent) => boolean` and set it on all such conditions as part of parsing/processing them whenever push rules change. Or have another interface which extends IEventMatchCondition e.g. `interface EventMatchConditionWithMemoizedHandler extends IEventMatchCondition` and "upgrade" them as part of processing push rules

clokep: I figured there was a way there, but does it make sense to add it there? It seemed odd to add functionality to an interface like that.

I did consider the upgrade path, but it seemed potentially fragile, I'll take another look.

t3chguy: Better than a leaky map :)

Could also prune the map when receiving new push rules via /sync

andybalaam: If I understand what you mean, you'll be adding the data to the object that implements the interface, and adding code to the interface to manage that data, which both seem fine. If you'd like a call to talk through how to implement this more, ping me.

clokep: > You could add a field to IEventMatchCondition which is optional

My fear with this is that if we somehow got a push rule with this field set we wouldn't calculate it. This might require a malicious homeserver, but 🤷 

> Or have another interface which extends IEventMatchCondition [...] and "upgrade" them as part of processing push rules whenever they change

This looks doable in `rewriteDefaultRules`, I think there might be some issues with multiple interfaces sharing the same `ConditionKind`? I did try creating a non-interface type which contained an `IEventMatchCondition` as a field, but adding that to `PushRuleCondition` caused issues with various methods which took an `IPushRuleCondition`.

I think we can do it in `rewriteDefaultRules` though and set a field, similar to the first approach. That seems simpler.

clokep:
@andybalaam I think this is ready for another look.

spec/unit/pushprocessor.spec.ts
```diff
@@ -287,6 +287,13 @@ describe("NotificationService", function () {
         expect(actions.tweaks.highlight).toEqual(true);
     });
 
+    // TODO: This is not spec compliant behaviour.
+    //
+    // See https://spec.matrix.org/v1.5/client-server-api/#conditions-1 which
+    // describes pattern should glob:
+    //
+    // 1. * matches 0 or ore characters;
```
andybalaam, line 295: ore -> more

andybalaam (approved):
One typo, looks great!

spec/unit/pushprocessor.spec.ts
clokep:
```suggestion
    // 1. * matches 0 or more characters;
```

andybalaam:
The downstream test failed https://github.com/matrix-org/matrix-js-sdk/actions/runs/4260935958/jobs/7414674597

(i.e. matrix-react-sdk tests failed with this change)

clokep:
> The downstream test failed [matrix-org/matrix-js-sdk/actions/runs/4260935958/jobs/7414674597](https://github.com/matrix-org/matrix-js-sdk/actions/runs/4260935958/jobs/7414674597)
> 
> (i.e. matrix-react-sdk tests failed with this change)

matrix-org/matrix-react-sdk#10237 should fix it (passes locally). I'm not sure if CI will pick-up on the "fix" though or not?

clokep:
> > The downstream test failed [matrix-org/matrix-js-sdk/actions/runs/4260935958/jobs/7414674597](https://github.com/matrix-org/matrix-js-sdk/actions/runs/4260935958/jobs/7414674597)
> > (i.e. matrix-react-sdk tests failed with this change)
> 
> [matrix-org/matrix-react-sdk#10237](https://github.com/matrix-org/matrix-react-sdk/pull/10237) should fix it (passes locally). I'm not sure if CI will pick-up on the "fix" though or not?

Although this now seems to be failing lint because we don't provide the `keyParts` when *creating* push rules. I think we really need to separate what we're using "internally" vs. what the protocol expects here. Everything kind of assumes they're one and the same though.

andybalaam:
@clokep are you blocked on me/us for this? Give me a ping if so.

clokep:
> @clokep are you blocked on me/us for this? Give me a ping if so.

No, just trying to figure out how to refactor tests after the changes I made.

src/pushprocessor.ts
```diff
@@ -162,7 +168,7 @@ export class PushProcessor {
         if (!newRules) newRules = {} as IPushRules;
         if (!newRules.global) newRules.global = {} as PushRuleSet;
         if (!newRules.global.override) newRules.global.override = [];
-        if (!newRules.global.override) newRules.global.underride = [];
+        if (!newRules.global.underride) newRules.global.underride = [];
```
clokep, line 171: This seems to have been a separate, long-standing bug?

andybalaam (approved):
Looks good to me.

src/pushprocessor.ts
```diff
@@ -202,6 +208,53 @@ export class PushProcessor {
...
+     */
+    public updateCachedPushRuleKeys(newRules: IPushRules): void {
+        // These lines are mostly to make the tests happy. We shouldn't run into these
+        // properties missing in practice.
+        if (!newRules) newRules = {} as IPushRules;
+        if (!newRules.global) newRules.global = {} as PushRuleSet;
+        if (!newRules.global.override) newRules.global.override = [];
+        if (!newRules.global.room) newRules.global.room = [];
+        if (!newRules.global.sender) newRules.global.sender = [];
+        if (!newRules.global.underride) newRules.global.underride = [];
```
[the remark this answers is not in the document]
andybalaam: Seems fine like this, if it's more consistent.

src/pushprocessor.ts
andybalaam: Yeah, how exciting. What would the symptoms have been? Maybe we can find an issue to close....

clokep: It was added in #2873, but that also added (on line 191) a `newRules.global.underride ?? []`, likely to solve the same problem? 🤷 (I don't think this would have a user-visible bug.)

PR #3146, opened 2023-02-13
Fix reactions in threads sometimes causing stuck notifications

justjanne:
Type: Defect
Fixes: https://github.com/vector-im/element-web/issues/24000

---

## 🐛 Bug Fixes
 * Fix reactions in threads sometimes causing stuck notifications ([\#3146](https://github.com/matrix-org/matrix-js-sdk/pull/3146)). Fixes vector-im/element-web#24000. Contributed by @justjanne.

### DIFF
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -2152,6 +2152,8 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         this.cachedThreadReadReceipts.delete(threadId);
 
         // If we managed to create a thread and figure out its `id` then we can use it
+        // This has to happen before thread.addEvents, because that adds events to the eventtimeline, and the
+        // eventtimeline sometimes looks up thread information via the room.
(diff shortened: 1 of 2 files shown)


src/models/room.ts
```diff
@@ -2151,14 +2151,15 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         // a reference to the cached receipts anymore.
         this.cachedThreadReadReceipts.delete(threadId);
 
+        // If we managed to create a thread and figure out its `id` then we can use it
+        this.threads.set(thread.id, thread);
+
```
gsouquet, line 2156: So if i remember correctly the fix works because of the order of execution. 
This needs to happen before another thing, so another code path manages to retrieve the thread model?

Adding a couple of lines about that flow will probably saves us a few gray hair in the future

gsouquet (approved):
Otherwise looks good to me, thank you for investigating and fixing the issue 💯

PR #3149, opened 2023-02-14
Include 'browser' in list of adjusted properties in release.sh

andybalaam:
---


t3chguy (approved):

richvdh:
@andybalaam could you add some notes for future archaeologists about what this fixed and why it was important?

andybalaam:
> @andybalaam could you add some notes for future archaeologists about what this fixed and why it was important?

@richvdh thank you for the prompt. I have added what I know to the PR description.

andybalaam:
I've added one sentence to the summary above but I think we've reached the extent of my understanding.

PR #3153, opened 2023-02-14
Refactor `CallFeed`s

SimonBrandner:
Todos:
- [ ] Fix tests
- [ ] Perform further tests
- [ ] Draw a diagram of the thing, somehow

Follow up work
- Move mute handling into `CallTrack`s
- Possibly abstract publishing into a  class
- Possibly abstract subscribing into a class
- Possibly abstract different `CallFeed` purposes into separate classes

---

### DIFF
--- a/src/webrtc/callFeed.ts
+++ b/src/webrtc/callFeed.ts
@@ -15,13 +15,15 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { SDPStreamMetadataPurpose, SDPStreamMetadataTracks } from "./callEventTypes";
+import { SDPStreamMetadataPurpose } from "./callEventTypes";
 import { acquireContext, releaseContext } from "./audioContext";
 import { MatrixClient } from "../client";
 import { RoomMember } from "../models/room-member";
-import { logger } from "../logger";
 import { TypedEventEmitter } from "../models/typed-event-emitter";
-import { CallEvent, CallState, MatrixCall } from "./call";
+import { MatrixCall } from "./call";
+import { CallTrack } from "./callTrack";
+import { randomString } from "../randomstring";
+import { logger } from "../logger";
 
 const POLLING_INTERVAL = 200; // ms
 export const SPEAKING_THRESHOLD = -60; // dB
@@ -30,16 +32,6 @@ const SPEAKING_SAMPLE_COUNT = 8; // samples
 export interface ICallFeedOpts {
     client: MatrixClient;
     roomId?: string;
-    userId: string;
-    deviceId: string | undefined;
-    /**
-     * Now, this should be the same as streamId but in the future we might want
-     * to use something different
-     */
-    feedId: string;
-    stream?: MediaStream;
-    purpose: SDPStreamMetadataPurpose;
-    tracksMetadata?: SDPStreamMetadataTracks;
     /**
      * Whether or not the remote SDPStreamMetadata says audio is muted
      */
@@ -48,10 +40,6 @@ export interface ICallFeedOpts {
      * Whether or not the remote SDPStreamMetadata says video is muted
      */
     videoMuted: boolean;
-    /**
-     * The MatrixCall which is the source of this CallFeed
-     */
-    call?: MatrixCall;
 }
 
 export enum CallFeedEvent {
@@ -76,20 +64,28 @@ type EventHandlerMap = {
     [CallFeedEvent.Disposed]: () => void;
 };
 
-export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap> {
-    public feedId: string;
-    public readonly userId: string;
-    public readonly deviceId: string | undefined;
-    public purpose: SDPStreamMetadataPurpose;
-    public speakingVolumeSamples: number[];
-    public tracksMetadata: SDPStreamMetadataTracks = {};
+export abstract class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap> {
+    public abstract readonly id?: string;
+    public abstract readonly streamId?: string;
+    public abstract readonly purpose: SDPStreamMetadataPurpose;
+    public abstract readonly connected: boolean;
+    public abstract readonly userId: string;
+    public abstract readonly deviceId?: string;
 
-    private _stream?: MediaStream;
-    private client: MatrixClient;
-    private call?: MatrixCall;
-    private roomId?: string;
-    private audioMuted: boolean;
-    private videoMuted: boolean;
+    public abstract isLocal: boolean;
+    public abstract isRemote: boolean;
+
+    public speakingVolumeSamples: number[];
+
+    protected readonly _id: string;
+    protected _tracks: CallTrack[] = [];
+    protected _stream?: MediaStream;
+    protected call?: MatrixCall;
+    protected roomId?: string;
+    protected client: MatrixClient;
+    protected audioMuted: boolean;
+    protected videoMuted: boolean;
+
     private localVolume = 1;
     private measuringVolumeActivity = false;
     private audioContext?: AudioContext;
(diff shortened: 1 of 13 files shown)


src/webrtc/callFeed.ts
```diff
@@ -76,20 +64,28 @@ type EventHandlerMap = {
...
 };
 
-export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap> {
-    public feedId: string;
-    public readonly userId: string;
-    public readonly deviceId: string | undefined;
-    public purpose: SDPStreamMetadataPurpose;
+export abstract class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap> {
+    public abstract readonly id?: string;
+    public abstract readonly streamId?: string;
```
EnricoSchw, line 69: Since we also have the stream object in the `CallFeed`, we should not set the stream ID as an extra property. We could make them accessible via a `getter`. This has the advantage if the stream change to not having to update this property too.

src/webrtc/callFeed.ts
```diff
@@ -76,20 +64,28 @@ type EventHandlerMap = {
...
-    private _stream?: MediaStream;
-    private client: MatrixClient;
-    private call?: MatrixCall;
-    private roomId?: string;
-    private audioMuted: boolean;
-    private videoMuted: boolean;
+
+    protected readonly _id: string;
+    protected _tracks: CallTrack[] = [];
+    protected _stream?: MediaStream;
```
EnricoSchw, line 82: One suggestion: we should be consistent with underscores. In my opinion, underscores are not necessary in typescript because we have operators like `private`, `public readonly`, `public`.

I don't mind using underscores to indicate that it's a very private property either. Only there doesn't seem to be a rule in `matrix-js-sdk` to use them or not.

src/webrtc/callFeed.ts
```diff
@@ -158,24 +130,31 @@ export class CallFeed extends TypedEventEmitter<CallFeedEvent, EventHandlerMap>
...
     private get hasAudioTrack(): boolean {
         return this.stream ? this.stream.getAudioTracks().length > 0 : false;
     }
 
-    private updateStream(oldStream?: MediaStream, newStream?: MediaStream): void {
+    protected get tracks(): CallTrack[] {
+        return [...this._tracks];
+    }
+
+    protected updateStream(oldStream?: MediaStream, newStream?: MediaStream): void {
```
EnricoSchw, line 149: This method allows calling like `updateStream(oldStream)`
In this case `this._stream = undefined`.

I am confused that the method has two optional parameters. So her behavior is not very clear to me. The method can add, update or delete a stream.  Do you wont such a powerful method because to make it easier to handle add, update and delete more common? (Because I saw `public setNewStream()`)

src/webrtc/localCallTrack.ts
```diff
@@ -0,0 +1,261 @@
...
+        return true;
+    }
+
+    public publish(call: MatrixCall): void {
+        if (this.published) {
+            throw new Error("Cannot publish already published track");
+        }
+
+        try {
+            this._transceiver = call.publishTrack(this);
```
EnricoSchw, line 183: Ok we get the `transceiver` per publishing 👍

src/webrtc/remoteCallFeed.ts
```diff
@@ -0,0 +1,228 @@
...
+    call: MatrixCall;
+
+    /**
+     * @deprecated addTransceiver() should be used instead
+     */
+    stream?: MediaStream;
+}
+
+export class RemoteCallFeed extends CallFeed {
+    private _connected = false;
```
EnricoSchw, line 35: `RemoteCallFeed` holds as well the info of the `CallState`? What is with `Screen Share`? We cover this as well in a extra `RemoteCallFeed` with a extra `CallState`?

EnricoSchw:
I like that you break down `call.ts` and `groupCall.ts` and split the Feeds into multiple files. That makes it much easier to read. ❤️

src/webrtc/remoteCallFeed.ts
SimonBrandner: Sorry, I am not sure I understand... `CallState` has nothing to do with screen-sharing - the possible values are connected, connecting, ended...

src/webrtc/callFeed.ts
SimonBrandner: Since this is the way it's always been, I'd leave it alone for now and revisit it in another PR for the sake of fewer changed lines. Does that sound fine?

src/webrtc/callFeed.ts
SimonBrandner: I think we're generally against using underscores for private props, the only case where we use them is when we have a getter and a private prop of the same name, which is the case here

EnricoSchw: OK, I see 👍

src/webrtc/callFeed.ts
EnricoSchw: fine

src/webrtc/callTrack.ts
```diff
@@ -0,0 +1,47 @@
...
+    public abstract get track(): MediaStreamTrack | undefined;
+    public abstract get trackId(): string | undefined;
+    public abstract get metadata(): SDPStreamMetadataTrack | undefined;
+    public abstract get kind(): string | undefined;
+
+    protected readonly _id: string;
+    protected _transceiver?: RTCRtpTransceiver;
+
+    public constructor(opts: CallTrackOpts) {
+        this._id = randomString(32);
```
dbkr, line 33: Is this another ID independent of the webrtc track id?

src/webrtc/remoteCallFeed.ts
```diff
@@ -0,0 +1,228 @@
...
+    public constructor(opts: RemoteCallFeedOpts) {
+        super(opts);
+
+        if (!opts.metadata && opts.call.opponentSupportsSDPStreamMetadata()) {
+            throw new Error(
+                "Cannot create RemoteCallFeed without metadata if the opponents supports sending sdp_stream_metadata",
+            );
+        }
+
+        this.streamId = opts.streamId;
```
dbkr, line 55: Is this also the same as the webrtc stream ID or different? If it's the same, why track it separately?

src/webrtc/remoteCallFeed.ts
```diff
@@ -0,0 +1,228 @@
...
+        this.call = opts.call;
+        this.metadata = opts.metadata;
+
+        this._stream = opts.stream || new window.MediaStream();
```
dbkr, line 59: I think probably just `MediaStream` as there's no `window` in node (assuming we'd ever do webrtc in node, but may as well follow best practice).

src/webrtc/remoteCallFeed.ts
```diff
@@ -0,0 +1,228 @@
...
+        return [...this._tracks];
+    }
+
+    public get connected(): boolean {
+        return this._connected;
+    }
+
+    private set connected(connected: boolean) {
+        this._connected = connected;
+        this.emit(CallFeedEvent.ConnectedChanged, this.connected);
```
dbkr, line 142: Should this only emit if it's changed?

src/webrtc/remoteCallFeed.ts
EnricoSchw: Maybe my question have been misleading.  I explain in some detail.

Why i have used the Screen Share example:
If a user receives a ScreenShare, we also cover this MediaStream as a `RemoteCallFeed`. Receiving Screen Share is an example that's an user receiving two media streams from the same remote user as well in the non SFU case. 

To the property itself: 
Your property `connected` maps two states into one: 'CallState' and basically `remote Track received (yes|no)`.

I was confused that the connection status is shown in the remote call feed because:
In the case of non SFU the change of `CallState` is the same for all `RemoteCallFeeds` from the same Remote User and in case of SFU the change of `CallState` is the same for all `RemoteCallFeeds`. 

With the cover of the state: `remote Track received (yes|no)`, you only cover the condition that the metadata event arrives before the remote media stream. This is the only case in which a `RemoteCallFeed` without a remote MediaStream without tracks exists.

In summary, `connected ` essentially maps the ICE connection state of the underlying WebRTC connection. Putting this state in the feeds is a fundamental design decision, so I wanted to question that.

dbkr (approved):
Thanks — couple of minor comments but this looks generally less repeated code & smaller functions, so codecov should be happier.

src/webrtc/remoteCallFeed.ts
SimonBrandner: This is analogous to `new window.RTCPeerConnection()` in `MatrixCall::createPeerConnection`. When running jest, `MediaStream` was undefined and the only way to get it working was this along with the mock in `installWebRTCMocks()`

src/webrtc/remoteCallFeed.ts
SimonBrandner: In this specific case, the reason is we create the `RemoteCallFeed` based on the metadata and give it the `streamId` as specified in the metadata and only then add tracks/transceivers to it based on the `streamId` it has as a prop

src/webrtc/remoteCallFeed.ts
SimonBrandner: I am still having a bit of trouble but do I understand you correctly you're wondering why the `connected` property is on all `CallFeed`s?

The answer there is that I wanted to hide all the underlying connection logic from the js-sdk consumer. Basically, the consumer only needs to create a group call and not care if it's full-mesh or SFU as it can read the `connected` property from the individual feeds instead of getting the individual calls and figuring out if they're connected or not. This doesn't feel right as it makes the consumer have to care about the full-mesh vs SFU difference and other stuff

Does that make sense?

EnricoSchw: Exactly, I'm asking about the API design. And yes, your intention makes absolutely sense to me. It's also the software design how Element Call works. If a connection is aborted, we will see it in the media stream tile. 

It's not a small design decision and it affects how the SDK is used. If this design decision is general consensus then that's ok and we can close this as solved.

EnricoSchw (approved):

src/webrtc/remoteCallFeed.ts
SimonBrandner: I believe it is, since it passed review before, shout if you disagree 😅

src/webrtc/callTrack.ts
dbkr: OK - I wondered if that was the case. Let's leave it like this for now & rename / comment it as appropriate when we switch over then.

src/webrtc/remoteCallFeed.ts
dbkr: oh, if we use `window` elsewhere then fair enough.

dbkr (approved):

PR #3154, opened 2023-02-14
Allow disabling VoIP support

stas-demydiuk:
Allow to manually disable VoIP support by passing a `{voipSupport: false}` option during matrix client initialization.

Signed-off-by: Stanislav Demydiuk <s.demydiuk@gmail.com>

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -342,6 +342,13 @@ export interface ICreateClientOpts {
     verificationMethods?: Array<VerificationMethod>;
 
     /**
+     * Set to false to disable the VoIP functionality of the client.
+     *
+     * Default true.
+     */
+    voipSupport?: boolean;
+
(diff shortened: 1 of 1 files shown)


SimonBrandner (commented):
I am not sure I see what would be the use case - could you please explain this?

stas-demydiuk:
> I am not sure I see what would be the use case - could you please explain this?

Yes, for cases when the server doesn't support the VoIP, it will prevent the client from making additional requests to `/client/r0/voip/turnServer`

t3chguy:
> for cases when the server doesn't support the VoIP

This isn't something which is allowed by the Matrix spec.

> A compliant server implementation MUST support all modules and supporting specification

We shouldn't be catering for non-compliant implementations

PR #3159, opened 2023-02-15
Fix predecessor types, nowhere does the spec say it can be `null`

t3chguy:
---

### DIFF
--- a/src/models/room-state.ts
+++ b/src/models/room-state.ts
@@ -976,19 +976,21 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
      * Note: if an m.predecessor event is used, eventId is null since those
      * events do not include an event_id property.
      */
-    public findPredecessor(
-        msc3946ProcessDynamicPredecessor = false,
-    ): { roomId: string; eventId: string | null } | null {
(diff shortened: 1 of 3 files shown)


src/models/room-state.ts
```diff
@@ -976,19 +976,21 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
      * Note: if an m.predecessor event is used, eventId is null since those
```
andybalaam, line 976: Comment should be updated to say undefined instead of null, but also I just realised this comment is out of date, because MSC3946 is being updated to allow eventId, and this is what the code supports, so it should read:

```
Note: if an m.predecessor event is used, eventId may be undefined since last_known_event_id is optional.
```

src/models/room.ts
```diff
@@ -3055,9 +3055,7 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
      * Note: if an m.predecessor event is used, eventId is null since those
```
andybalaam, line 3055: Same comment about the comment.

andybalaam (approved):
All good with the comment update I asked for.

PR #3161, opened 2023-02-16
Run matrix-react-sdk tests on merge queue

t3chguy:
Requires https://github.com/matrix-org/matrix-react-sdk/pull/10174
Requires https://github.com/matrix-org/matrix-react-sdk/pull/10198

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -72,10 +72,10 @@ jobs:
     # Hook for branch protection to work outside merge queues
     downstream:
         name: Downstream tests
+        if: always()
         runs-on: ubuntu-latest
         needs:
             - matrix-react-sdk
-        if: always()
         steps:
             - if: needs.matrix-react-sdk.result != 'skipped' && needs.matrix-react-sdk.result != 'success'
               run: exit 1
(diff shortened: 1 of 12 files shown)


.github/workflows/tests.yml
```diff
@@ -59,3 +61,21 @@ jobs:
                   path: |
                       coverage
                       !coverage/lcov-report
+
+    matrix-react-sdk:
+        name: Downstream test matrix-react-sdk
+        if: github.event_name == 'merge_group'
+        uses: matrix-org/matrix-react-sdk/.github/workflows/tests.yml@develop
```
gsouquet, line 68: Could we run this workflow without any of the coverage option? 
That will significantly speed up test execution, and we don't need to check for coverage in this scenario

gsouquet (changes requested):
Otherwise looks good to me 👍

.github/workflows/tests.yml
t3chguy: Done, requires https://github.com/matrix-org/matrix-react-sdk/pull/10198

gsouquet (approved):
lgtm!

Issue #902, opened 2019-04-21
initCrypto in Apache Cordova doesn't work


danielgottbehuet:
Hi all,
i'm working on a mobile app using the matrix-js-sdk and Olm crypto support.
Everything works fine in the browser and iam able to receive crypto messages and decrypt them + show the event.content().
But when i'm try to run the app via Adobe PhoneGap or natively via `cordova run android`, there is no return from `matrixClient.initCrypto()`.
Console shows:
`connecting to indexeddb matrix-js-sdk:crypto`
`connected to indexeddb matrix-js-sdk:crypto`
And thats it.
I can't figure out what the problem is. Maybe the WebAssembly?
Maybe someone knows how to find out what the problem is.

danielgottbehuet:
`cordova Fetch API cannot load file:///android_asset/www/js/olm.wasm. URL scheme "file" is not supported.`
Ok, looks like its my fault.

PR #3162, opened 2023-02-16
Enable group calls without video and audio track by configuration of MatrixClient

EnricoSchw:
## Checklist

Allow Element Call to be started without audio / video interface for the sake of a screen-sharing only session https://github.com/vector-im/element-call/issues/902

---


src/webrtc/groupCall.ts
```diff
@@ -1461,4 +1474,19 @@ export class GroupCall extends TypedEventEmitter<
...
+     * Joining without a camera and video should only be explicitly allowed, because there can be unexpected behavior
+     * when later querying the media access rights from browser that the client cannot intercept. Join without came
+     * and audio should be a property like PTT.
+     */
+    public get initCallWithoutVideoAndAudio(): boolean {
+        return this.allowCallWithOutVideoAndAudio;
+    }
+
+    public set initCallWithoutVideoAndAudio(allowCallWithOutVideoAndAudio: boolean) {
+        this.allowCallWithOutVideoAndAudio = allowCallWithOutVideoAndAudio;
```
SimonBrandner, line 1490: Why not put this into opts and make it readonly and public?

EnricoSchw: In the beginning `createGroupCall` from the MatrixClient stopped me from doing that. But since the configuration is now done via the Matrix Client, there is no reason not to do it via the constructor. Thx :+1:

src/webrtc/groupCall.ts
```diff
@@ -609,6 +617,11 @@ export class GroupCall extends TypedEventEmitter<
      * @returns Whether muting/unmuting was successful
      */
     public async setLocalVideoMuted(muted: boolean): Promise<boolean> {
+        // Because we need a Local Call Feed to establish a call connection, we avoid muting video in case of empty
+        // video track. In this way we go sure if a client implements muting we don't raise an error.
+        if (this.initCallWithoutVideoAndAudio && this.localCallFeed?.stream.getVideoTracks().length === 0) {
+            return false;
+        }
```
SimonBrandner, line 624: Sorry, it's somehow not clear to me what this piece of code does exactly

SimonBrandner (commented):
I am not exactly confident having this as a flag on the client is the correct way to go though I can't think of alternatives :thinking: Could we quickly go over this in the EC internal room?

EnricoSchw:
> I am not exactly confident having this as a flag on the client is the correct way to go though I can't think of alternatives 🤔 Could we quickly go over this in the EC internal room?

Technically we could implement a new GroupCall Type in the same way we do this with PTT. After this we can bind it on EC internal room like we do it currently with PTT.  

However, I don't think that's a good idea. The SDK is based on Feeds with a video and a audio track. PTT is already working around this and provide Feeds with only audio tracks. The feature I'm implemented should allows calls without Feeds at all. I'am working around this in a way thats I am provide Feeds without audio and without video tracks (but this is not a Feed anymore). However this is a hack, because a User is connected in the call, when we receive a Feed from the user means a Video or Audio Track from him. As long as we don't know exactly what is a Feed and what isn't, it's better to configure this feature separately via the client.

src/webrtc/groupCall.ts
EnricoSchw: My implementation allows feeds that does not contain tracks at all. I have to make sure that the user of the SDK doesn't try to mute a feed that doesn't have a video track. This case must be caught. It would be logically correct if we would have no LocalFeed. This case is covert below. But you need a LocalFeed to innit a GroupCall. I can't go the way of PTT either, because I have to assume that no device exists.

dbkr: Do we need to check for `this.initCallWithoutVideoAndAudio`? Wouldn't it always be an error to try & mute/unmute video if there was no video track?

dbkr (commented):
I think basically either we add this flag (which does make the logic of calls a bit more complex) or accept that it's no longer an error condition to have a call with no local or remote feeds (ie. basically just make this behaviour the default. I think making it the default could be okay actually: it would be an API change (albeit not a very significant one) but I think that's OK?

src/webrtc/groupCall.ts
EnricoSchw: i'll remove it

EnricoSchw:
> I think basically either we add this flag (which does make the logic of calls a bit more complex) or accept that it's no longer an error condition to have a call with no local or remote feeds (ie. basically just make this behaviour the default. I think making it the default could be okay actually: it would be an API change (albeit not a very significant one) but I think that's OK?

I didn't want to change that much and I wanted to keep it very simple, so we can close this ticket #902.

This implementation works because no feed is expected if flag active. I have no idea how complicated it will be to add a new GrupenCall type.

dbkr:
Agreed, I don't think adding a new group call type is the way to go. I feel okay about the flag since it leaves 1:1 calls in element web etc in a mode where the call will fail if no media is negotiated, which is probably the sensible thing to do. Perhaps we could put the option either in `initLocalCallFeeds` or a separate dedicated setter for the option. @SimonBrandner how would you feel about that?

SimonBrandner:
> Agreed, I don't think adding a new group call type is the way to go. I feel okay about the flag since it leaves 1:1 calls in element web etc in a mode where the call will fail if no media is negotiated, which is probably the sensible thing to do. Perhaps we could put the option either in `initLocalCallFeeds` or a separate dedicated setter for the option. @SimonBrandner how would you feel about that?

Yeah, that sounds good

EnricoSchw:
> Agreed, I don't think adding a new group call type is the way to go. I feel okay about the flag since it leaves 1:1 calls in element web etc in a mode where the call will fail if no media is negotiated, which is probably the sensible thing to do. Perhaps we could put the option either in `initLocalCallFeeds` or a separate dedicated setter for the option.

`initCallWithoutVideoAndAudio` change the behavior of the whole SDK and not only the behavior of the Feeds.. it should be a main property and not pass by a feed init method.

And because it change the behavior of the SDK its a good idea to path it by constructor, we already have discussed

EnricoSchw:
Maybe the name `initCallWithoutVideoAndAudio` is misleading. I will change the name to `allowCallWithoutVideoAndAudio`. Its then consistent to EC.

dbkr:
I don't particularly mind if you feel strongly about it, but the flag is only used within `initLocalCallFeedInternal `.

SimonBrandner (changes requested):
Could we just no-op if we don't have the feed to mute?

EnricoSchw:
> Could we just no-op if we don't have the feed to mute?
Could you please help me to understand you. Are you referring to these lines here? [For Example: Mute Video](https://github.com/matrix-org/matrix-js-sdk/blob/c6294e046ebec1c5c479d1ceda09a1f6343cafb9/src/webrtc/groupCall.ts#L655)

SimonBrandner:
> > Could we just no-op if we don't have the feed to mute?
> 
> Could you please help me to understand you. Are you referring to these lines here? [For Example: Mute Video](https://github.com/matrix-org/matrix-js-sdk/blob/c6294e046ebec1c5c479d1ceda09a1f6343cafb9/src/webrtc/groupCall.ts#L655)

Sorry, for not being clear. Yeah, I am asking if we could just `if (...) return;` if the function won't do anything in the end instead of `try {...} catch (...) {...}`

EnricoSchw:
> > > Could we just no-op if we don't have the feed to mute?
> > 
> > 
> > Could you please help me to understand you. Are you referring to these lines here? [For Example: Mute Video](https://github.com/matrix-org/matrix-js-sdk/blob/c6294e046ebec1c5c479d1ceda09a1f6343cafb9/src/webrtc/groupCall.ts#L655)
> 
> Sorry, for not being clear. Yeah, I am asking if we could just `if (...) return;` if the function won't do anything in the end instead of `try {...} catch (...) {...}`

Our SDK needs a `LocalFeed` to init a `GroupCall`. Of course, we could do `GroupCall`s without `LocalFeed`. Because `Feed` more or less the basement of this SDK to cover a WebRTC connection I do not know how big is this efforts. This was the reason to go this way.

The Trick is to init a `GroupCall` with a `LocalFeed` with an empty Stream. Now as user, you can join a call without device permissions., because you do not need a Video or Audio.  If then the user try to unmute `this.client.getMediaHandler().hasVideoDevice()` can raise an error, because about permissions. Another way is to not display the mute buttons.

EnricoSchw:
@dbkr @SimonBrandner Would you agree with my point of view, or do you have big concerns?

src/webrtc/mediaHandler.ts
```diff
@@ -185,13 +185,23 @@ export class MediaHandler extends TypedEventEmitter<
...
     }
 
     public async hasVideoDevice(): Promise<boolean> {
-        const devices = await navigator.mediaDevices.enumerateDevices();
-        return devices.filter((device) => device.kind === "videoinput").length > 0;
+        try {
+            const devices = await navigator.mediaDevices.enumerateDevices();
+            return devices.filter((device) => device.kind === "videoinput").length > 0;
+        } catch (err) {
+            logger.log(`MediaHandler hasVideoDevice() calling navigator.mediaDevices.enumerateDevices with error`);
```
SimonBrandner, line 202:
```suggestion
            logger.log(`MediaHandler hasVideoDevice() calling navigator.mediaDevices.enumerateDevices with error`, err);
```

src/webrtc/mediaHandler.ts
```diff
@@ -185,13 +185,23 @@ export class MediaHandler extends TypedEventEmitter<
     }
 
     public async hasAudioDevice(): Promise<boolean> {
-        const devices = await navigator.mediaDevices.enumerateDevices();
-        return devices.filter((device) => device.kind === "audioinput").length > 0;
+        try {
+            const devices = await navigator.mediaDevices.enumerateDevices();
+            return devices.filter((device) => device.kind === "audioinput").length > 0;
+        } catch (err) {
+            logger.log(`MediaHandler hasAudioDevice() calling navigator.mediaDevices.enumerateDevices with error`);
```
SimonBrandner, line 192:
```suggestion
            logger.log(`MediaHandler hasAudioDevice() calling navigator.mediaDevices.enumerateDevices with error`, err);
```

spec/unit/webrtc/groupCall.spec.ts
```diff
@@ -882,11 +884,27 @@ describe("Group Call", function () {
...
                 jest.spyOn(mockClient.getMediaHandler(), "hasVideoDevice").mockResolvedValue(false);
                 expect(await groupCall.setLocalVideoMuted(false)).toBe(false);
             });
+
+            it("returns false when no permission for video stream", async () => {
+                const groupCall = await createAndEnterGroupCall(mockClient, room);
+                jest.spyOn(mockClient.getMediaHandler(), "getUserMediaStream").mockRejectedValueOnce(
+                    new Error("No Permission"),
+                );
+                expect(await groupCall.setLocalVideoMuted(false)).toBe(false);
```
SimonBrandner, line 906: Shouldn't it return `true` since the track stays muted?

spec/unit/webrtc/groupCall.spec.ts
```diff
@@ -882,11 +884,27 @@ describe("Group Call", function () {
                 expect(await groupCall.setMicrophoneMuted(false)).toBe(false);
             });
 
+            it("returns false when no permission for audio stream", async () => {
+                const groupCall = await createAndEnterGroupCall(mockClient, room);
+                jest.spyOn(mockClient.getMediaHandler(), "getUserMediaStream").mockRejectedValueOnce(
+                    new Error("No Permission"),
+                );
+                expect(await groupCall.setMicrophoneMuted(false)).toBe(false);
```
SimonBrandner, line 892: Shouldn't it return `true` since the track stays muted?

SimonBrandner (commented):
Can the code in `groupcall.ts` be simplified now?

spec/unit/webrtc/groupCall.spec.ts
EnricoSchw: The return value of the method is not the current state of mute. The return value refers to the successful execution of the method. The mute state is propagated via an event.

spec/unit/webrtc/groupCall.spec.ts
EnricoSchw: The return value of the method is not the current state of mute. The return value refers to the successful execution of the method. The mute state is propagated via an event.

EnricoSchw:
> Can the code in `groupcall.ts` be simplified now?

@SimonBrandner Can you please be more specific? Do you mean `setLocalVideoMuted` and `setMicrophoneMuted`? I would like to use `promise.then().catch()` instead of try catch blocks, but the linter forbids that.

spec/unit/webrtc/groupCall.spec.ts
SimonBrandner: Argh, that is confusing - the method on `call.ts` with the same name returns the new mute state...

spec/unit/webrtc/groupCall.spec.ts
SimonBrandner: Argh, that is confusing - the method on `call.ts` with the same name returns the new mute state...

SimonBrandner (approved):

EnricoSchw:
@dbkr  PTAL

dbkr (approved):

t3chguy:
Due to a required parameter being added

t3chguy:
https://github.com/matrix-org/matrix-js-sdk/pull/3186 brought back backwards compatibility

PR #3167, opened 2023-02-21
Add easy way to determine if the decryption failure is due to "DecryptionError: The sender has disabled encrypting to unverified devices."

florianduros:
## Checklist

---

## ✨ Features
 * Add easy way to determine if the decryption failure is due to "DecryptionError: The sender has disabled encrypting to unverified devices." ([\#3167](https://github.com/matrix-org/matrix-js-sdk/pull/3167)). Contributed by @florianduros.


alunturner (approved):

src/models/event.ts
```diff
@@ -704,6 +704,22 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
...
+     *
+     * (This implies that we might retry decryption at some point in the future)
+     *
+     * @returns boolean
+     */
+    public isEncryptedDisabledForUnverifiedDevices(): boolean {
+        return (
+            this.isDecryptionFailure() &&
+            this.clearEvent?.content?.reason ===
+                "DecryptionError: The sender has disabled encrypting to unverified devices."
```
t3chguy, line 719: Could we maybe refer to the source of truth for this string to make this less brittle?

florianduros: @t3chguy I used a boolean instead of the reason, what do you think about it ?

src/models/event.ts
```diff
@@ -913,9 +903,10 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
                 content: {
                     msgtype: "m.bad.encrypted",
                     body: "** Unable to decrypt: " + reason + " **",
-                    reason,
                 },
             },
+            isEncryptedDisabledForUnverifiedDevices:
+                reason === "DecryptionError: The sender has disabled encrypting to unverified devices.",
```
t3chguy, line 909: It'd be great here if we were using the source of truth which iirc is WITHHELD_MESSAGES

as for the previous approach vs this one, either is fine, but `MatrixEvent::isEncryptedDisabledForUnverifiedDevices` is mutable so should really be behind a getter

florianduros: When the event is decrypted, `crypto.decryptEvent(this)` is called and the string `DecryptionError: The sender has disabled encrypting to unverified devices."` is the content of the error raised. How can I have access to the WITHHELD_MESSAGES ?

t3chguy: So the reason field is a stringified `algorithms.DecryptionError`

It's still not ideal but I'd suggest `reason === "DecryptionError: " + WITHHELD_MESSAGES["m.unverified"]` or similar, rather than duplicating the user-facing string.

florianduros: Oh right thanks

florianduros: I used `WITHHELD_MESSAGES `, it should be cleaner.

src/models/event.ts
```diff
@@ -710,6 +711,10 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
         return this.clearEvent?.content?.msgtype === "m.bad.encrypted";
     }
 
+    public isEncryptedDisabledForUnverifiedDevices(): boolean {
```
t3chguy, line 714: This should be a getter rather than a function

florianduros: I used the same behaviour than `isDecryptionFailure` which is also a function, it should be also a getter in this case ?

t3chguy: The SDK predates getters being available in the language, we should be using them going forward though, not breaking backwards compatibility by mutating existing contracts

PR #3175, opened 2023-02-22
Improve types

t3chguy:
---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -4130,7 +4130,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     public setPowerLevel(
         roomId: string,
         userId: string | string[],
-        powerLevel: number,
+        powerLevel: number | undefined,
         event: MatrixEvent | null,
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -4130,7 +4130,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     public setPowerLevel(
         roomId: string,
         userId: string | string[],
-        powerLevel: number,
+        powerLevel: number | undefined,
```
gsouquet, line 4133: Wondering whether the following would be more adequate? Or whether it's better to be really explicit?

```suggestion
        powerLevel?: number,
```

gsouquet (approved):
Otherwise lgtm 👍

src/client.ts
t3chguy: Can't because the next field isn't optional, optional fields cannot be followed by mandatory fields

gsouquet: Woops, i read that wrong and thought we destructured the thing

PR #3178, opened 2023-02-27
Fix spec compliance issue around encrypted `m.relates_to`

t3chguy:
Requires https://github.com/matrix-org/matrix-react-sdk/pull/10248

image

---

## 🐛 Bug Fixes
 * Fix spec compliance issue around encrypted `m.relates_to` ([\#3178](https://github.com/matrix-org/matrix-js-sdk/pull/3178)).


gsouquet (changes requested):
Looks good to me! Thank you for fixing this
Seems to be a case where we'd benefit from adding a test

gsouquet (approved):
Thank you

PR #3183, opened 2023-03-02
Abstract publishing

SimonBrandner:
I am not 100% happy with the abstraction - let me know if you have any thoughts

---


dbkr (approved):
I think this makes sense and it seems nice to split these out. If anything, some comments on the classes talking about what the track/feedPublication objects do & represent could be helpful. For instance, why is a FeedPublication necessary over just publishing the two tracks separately?

dbkr (approved):
Perfect, thanks for the comments.

Issue #3182, opened 2023-03-01
browser-matrix.min.js file size became 4x larger, causing negative impact on web app startup time


tomo564:
At v23.0.0, browser-matrix.min.js file size was 1.37MB. The file size jumped more than 4x to 5.67MB at v23.1.0-rc.1 and now it's 5.44MB for v23.4.0.

This jump in the file size causes some significant delays in webapp startup time especially for the browsers on mobile network environment like 4G/5G data connections.

Should the browser JS bundle be offered in "light-weight" build alongside fully-featured build. Or maybe in some modular way so that the users can pick and choose what they load in their apps.

t3chguy:
This will be the new Rust crypto library being included - https://github.com/matrix-org/matrix-js-sdk/compare/v23.0.0...v23.1.0-rc.1#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R57

Pinging @richvdh - maybe the browserify build should only include Olm or we have two separate builds depending on which you prefer. This size increase isn't really expected or intended.

richvdh:
> Pinging @richvdh - maybe the browserify build should only include Olm or we have two separate builds depending on which you prefer. This size increase isn't really expected or intended.

The browserify build doesn't include libolm currently: if you want to use E2E with the browserified artifact you have to load it separately. We have a bunch of *magic* to support allowing the application to provide libolm in different ways - or not at all - and I'm really not enthusiastic about repeating that complexity for the rust implementation. We end up with lots of complexity in the application so that it behaves correctly whether or not E2E is supported by the js-sdk.

We can look at removing the rust crypto library from the browserify artifact to restore the status quo for now, but that doesn't really solve the long-term problem, as we will be dropping support for libolm. Creating different browserify artifacts is a *possibility*, but as https://github.com/vector-im/element-web/issues/11876 suggests we might want to move away from browserify anyway.

richvdh:
For now, I have stubbed out the rust crypto implementation. We'll need to figure out what we are doing here long-term though, for which I have opened #3189.

PR #3187, opened 2023-03-02
Stub out the rust crypto implementation for browserify

richvdh:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/3182.

Even this took me an annoyingly long time to figure out. We *really* need to figure out if browserify is important to us.

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,10 @@
         "matrix-org"
     ],
     "main": "./src/index.ts",
-    "browser": "./src/browser-index.ts",
+    "browser": {
+        "./src/index.ts": "./src/browser-index.ts",
+        "./src/rust-crypto": "./src/rust-crypto/browserify-index.ts"
+    },
     "matrix_src_main": "./src/index.ts",
     "matrix_src_browser": "./src/browser-index.ts",
     "matrix_lib_main": "./lib/index.js",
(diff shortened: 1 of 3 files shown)


richvdh:
gah, this breaks webpack as well

package.json
```diff
@@ -33,7 +33,10 @@
         "matrix-org"
     ],
     "main": "./src/index.ts",
-    "browser": "./src/browser-index.ts",
+    "browser": {
```
t3chguy, line 36: This field is managed by release.sh so that would need updating to not revert your fix

richvdh: gordon bennett. Why are we still using JSON as a config file format in 2023?

richvdh: thanks for the heads-up, anyway. My more recent attempt doesn't change this, so it's moot.

package.json
```diff
@@ -147,5 +148,25 @@
...
+            "browserify-swap",
+            [
+                "babelify",
+                {
+                    "sourceMaps": "inline",
+                    "presets": [
+                        "@babel/preset-env",
+                        "@babel/preset-typescript"
+                    ]
+                }
```
richvdh, line 163: moving this here wasn't a requirement, but the commandline was getting increasingly unwieldy.

src/rust-crypto/browserify-index.ts
```diff
@@ -0,0 +1,31 @@
+/*
+Copyright 2022 The Matrix.org Foundation C.I.C.
```
t3chguy, line 2:
```suggestion
Copyright 2023 The Matrix.org Foundation C.I.C.
```

t3chguy (approved):
Looks great to me, thanks!!

t3chguy:
(Feel free to bypass sonar here)

PR #3193, opened 2023-03-06
Support for MSC3906 v2

hughns:
This PR adds support for v2 of [MSC3906](https://github.com/matrix-org/matrix-spec-proposals/pull/3906) which is used for Sign in with QR.

Both v1 (default) and v2 are supported so compatibility is maintained.

Signed-off-by: Hugh Nimmo-Smith <hughns@element.io>

## Checklist

Notes: Updates to protocol used for Sign in with QR code

---

## ✨ Features
 * Updates to protocol used for Sign in with QR code ([\#3193](https://github.com/matrix-org/matrix-js-sdk/pull/3193)). Contributed by @hughns.

### DIFF
--- a/src/rendezvous/MSC3906Rendezvous.ts
+++ b/src/rendezvous/MSC3906Rendezvous.ts
@@ -134,6 +134,10 @@ export class MSC3906Rendezvous {
         this._code = JSON.stringify(raw);
     }
 
+    /**
+     *
+     * @returns the checksum of the secure channel if the rendezvous set up was successful, otherwise undefined
+     */
     public async startAfterShowingCode(): Promise<string | undefined> {
         const checksum = await this.channel.connect();
 
@@ -158,17 +162,43 @@ export class MSC3906Rendezvous {
         });
 
         logger.info("Waiting for other device to chose protocol");
-        const { type, protocol, outcome, reason, intent } = await this.receive();
+        const nextPayload = await this.receive();
 
+        this.checkForV1Fallback(nextPayload);
+
+        const protocol = this.v1FallbackEnabled
+            ? await this.handleV1ProtocolPayload(nextPayload)
+            : await this.handleV2ProtocolPayload(nextPayload);
+
+        // invalid protocol
+        if (!protocol || !LOGIN_TOKEN_PROTOCOL.matches(protocol)) {
+            await this.cancel(RendezvousFailureReason.UnsupportedAlgorithm);
+            return undefined;
+        }
+
+        return checksum;
+    }
+
+    private checkForV1Fallback({ type }: MSC3906RendezvousPayload): void {
         // even if we didn't start in v1 fallback we might detect that the other device is v1
         if (type === PayloadType.Finish || type === PayloadType.Progress) {
             // this is a PDU from a v1 flow so use fallback mode
             this.v1FallbackEnabled = true;
         }
+    }
 
-        // fallback for v1 flow
+    /**
+     *
+     * @returns true if the protocol was received successfully, false otherwise
+     */
+    private async handleV1ProtocolPayload({
+        type,
+        protocol,
+        outcome,
+        reason,
+        intent,
+    }: MSC3906RendezvousPayload): Promise<string | void> {
         if (type === PayloadType.Finish) {
-            this.v1FallbackEnabled = true;
             // new device decided not to complete
             let reason: RendezvousFailureReason;
             if (intent) {
@@ -182,9 +212,29 @@ export class MSC3906Rendezvous {
                 reason = RendezvousFailureReason.Unknown;
             }
             await this.cancel(reason);
-            return undefined;
+            return;
         }
 
+        // unexpected payload
+        if (type !== PayloadType.Progress) {
+            await this.cancel(RendezvousFailureReason.Unknown);
+            return;
+        }
+
+        return protocol;
+    }
+
+    /**
+     *
+     * @returns true if the protocol was received successfully, false otherwise
+     */
+    private async handleV2ProtocolPayload({
+        type,
+        protocol,
+        outcome,
+        reason,
+        intent,
+    }: MSC3906RendezvousPayload): Promise<string | void> {
         // v2 flow
         if (type === PayloadType.Failure) {
             // new device decided not to complete
@@ -206,28 +256,16 @@ export class MSC3906Rendezvous {
                     failureReason = RendezvousFailureReason.Unknown;
             }
             await this.cancel(failureReason);
-            return undefined;
+            return;
         }
 
-        // v1 unexpected payload
-        if (this.v1FallbackEnabled && type !== PayloadType.Progress) {
+        // unexpected payload
+        if (type !== PayloadType.Protocol) {
             await this.cancel(RendezvousFailureReason.Unknown);
-            return undefined;
+            return;
         }
 
-        // v2 unexpected payload
-        if (!this.v1FallbackEnabled && type !== PayloadType.Protocol) {
-            await this.cancel(RendezvousFailureReason.Unknown);
-            return undefined;
-        }
-
-        // invalid protocol
-        if (!protocol || !LOGIN_TOKEN_PROTOCOL.matches(protocol)) {
-            await this.cancel(RendezvousFailureReason.UnsupportedAlgorithm);
-            return undefined;
-        }
-
-        return checksum;
+        return protocol;
     }
 
     private async receive(): Promise<MSC3906RendezvousPayload> {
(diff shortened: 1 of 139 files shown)


hughns:
n.b. the code smells reported by SonarCloud are for deliberate use of deprecated MSC3906 v1 types

richvdh:
for links: previous impl was https://github.com/matrix-org/matrix-js-sdk/pull/2747

src/rendezvous/MSC3906Rendezvous.ts
```diff
@@ -25,11 +32,23 @@ import { logger } from "../logger";
 import { sleep } from "../utils";
 
 enum PayloadType {
```
richvdh, line 34: please could you give this a doc-comment with a bit text explaining what it does. (The name is *sort-of* self-explanatory, but I would prefer it to be explicit).

As I understand it it is used to populate the `type` property in the MSC3886 payload?

src/rendezvous/MSC3906Rendezvous.ts
```diff
@@ -38,10 +57,21 @@ enum Outcome {
     Unsupported = "unsupported",
 }
 
+enum FailureReason {
```
richvdh, line 60: again, let's give this a docstring please. seems to be potential `reason` properties of payloads of type `m.login.failure` ?

src/rendezvous/RendezvousFlow.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const SETUP_ADDITIONAL_DEVICE_FLOW_V1 = "org.matrix.msc3906.v1";
+
+export const SETUP_ADDITIONAL_DEVICE_FLOW_V2 = new UnstableValue(
+    "m.setup.additional_device.v2",
+    "org.matrix.msc3906.setup.additional_device.v2",
+);
+
+// v1 is never included in the JSON, but we give it a name for the sake of determining the flow to use
+export type RendezvousFlow =
```
richvdh, line 27: doc-comments for these guys please. what do they do/where are they used?

src/rendezvous/RendezvousFlow.ts
```diff
@@ -0,0 +1,30 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { UnstableValue } from "../NamespacedValue";
+
+export const SETUP_ADDITIONAL_DEVICE_FLOW_V1 = "org.matrix.msc3906.v1";
+
+export const SETUP_ADDITIONAL_DEVICE_FLOW_V2 = new UnstableValue(
+    "m.setup.additional_device.v2",
```
richvdh, line 22: (aside: why v2, given we have (presumably) never used a v1 without the `org.matrix.msc3906` unstable prefix? This is the whole point of unstable prefixes.)

src/rendezvous/MSC3906Rendezvous.ts
```diff
@@ -64,18 +94,23 @@ export class MSC3906Rendezvous {
     private newDeviceId?: string;
     private newDeviceKey?: string;
     private ourIntent: RendezvousIntent = RendezvousIntent.RECIPROCATE_LOGIN_ON_EXISTING_DEVICE;
+    private v1FallbackEnabled: boolean;
```
richvdh, line 97: for me, "fallback" suggests that we try v2 and, if it doesn't work, *fall back* to using v1.

As far as I can tell, once `v1FallbackEnabled` is true, we just use the whole of the v1 flow from that point on, rather than trying v2 at all? (In fact, it looks to me as though "v1 fallback" is *always* enabled, in that as soon as we receive a v1-like payload we switch to v1 mode.)

If so, this is maybe not a great name: `useV1Flow` or something might be better.

src/rendezvous/MSC3906Rendezvous.ts
```diff
@@ -64,18 +94,23 @@ export class MSC3906Rendezvous {
...
     private _code?: string;
 
     /**
      * @param channel - The secure channel used for communication
      * @param client - The Matrix client in used on the device already logged in
      * @param onFailure - Callback for when the rendezvous fails
+     * @param flow - The flow to use. Defaults to MSC3906 v1 for backwards compatibility.
```
richvdh, line 104: please could we have some more words explaining what a "flow" means and how I, as a user of this API, should pick one?

src/rendezvous/channels/MSC3903ECDHv2RendezvousChannel.ts
```diff
@@ -98,6 +99,7 @@ export class MSC3903ECDHv2RendezvousChannel<T> implements RendezvousChannel<T> {
                 key: encodeUnpaddedBase64(this.ourPublicKey),
                 transport: await this.transport.details(),
             },
+            flow,
             intent,
```
richvdh, line 103: While trying to get a handle on how these pieces fit together, I find it a bit odd that this is implementing a "MSC3903 rendezvous channel", but there is no mention in MSC3903 of this bit. If this is specific to MSC3906, shouldn't it be in `MSC3906Rendezvous` rather than here?

src/rendezvous/MSC3906Rendezvous.ts
```diff
@@ -92,9 +127,16 @@ export class MSC3906Rendezvous {
...
+            ? await this.channel.generateCode(this.ourIntent)
+            : await this.channel.generateCode(this.ourIntent, this.flow);
+        this._code = JSON.stringify(raw);
     }
 
+    /**
+     *
+     * @returns the checksum of the secure channel if the rendezvous set up was successful, otherwise undefined
+     */
     public async startAfterShowingCode(): Promise<string | undefined> {
```
richvdh, line 140: Particularly given this is a public function, please could it have a better doc-comment? What does it actually do, beyond returning a success flag?

src/rendezvous/MSC3906Rendezvous.ts
```diff
@@ -104,39 +146,125 @@ export class MSC3906Rendezvous {
...
+            await this.cancel(RendezvousFailureReason.UnsupportedAlgorithm);
+            return undefined;
+        }
+
+        return checksum;
+    }
+
+    private checkForV1Fallback({ type }: MSC3906RendezvousPayload): void {
+        // even if we didn't start in v1 fallback we might detect that the other device is v1
+        if (type === PayloadType.Finish || type === PayloadType.Progress) {
```
richvdh, line 183: I still see `m.login.progress` on MSC3906: is it correct that that is a v1 payload?

src/rendezvous/MSC3906Rendezvous.ts
```diff
@@ -104,39 +146,125 @@ export class MSC3906Rendezvous {
...
+            // this is a PDU from a v1 flow so use fallback mode
+            this.v1FallbackEnabled = true;
+        }
+    }
```
richvdh, line 187: I'd be inclined to inline this method, but ymmv

richvdh (commented):
Afraid I have to step away for a meeting, but I've left a few initial thoughts/stupid questions.

As a general theme: I would like to see more doc-comments explaining what things do/what they are for, *even if it seems obvious*. Writing out in words how an API is used/what it does can be really useful in terms of helping people understand, and maintaining a clean API.

src/rendezvous/MSC3906Rendezvous.ts
hughns: I had it inline but Sonarcloud complained that the function was too complicated, so I split it out!

src/rendezvous/MSC3906Rendezvous.ts
hughns: I'm renaming it to `usingV1Flow` with a comment.

src/rendezvous/RendezvousFlow.ts
hughns: IIRC, I chose `v2` in the MSC to make it explicit that the was an earlier version of the MSC in use in the wild. I appreciate that it isn't strictly necessary.

hughns: Or are you actually suggesting making the stable name not have the v2? If so that's a good idea.

src/rendezvous/MSC3906Rendezvous.ts
hughns: Yes, the code is correct - I've added comments to explain it.

justjanne:
Please make sure to rebase this PR so it can be compared against the current codebase cleanly

hughns:
Closing this in preference to moving directly to supporting Sign in with QR with OIDC.

PR #3200, opened 2023-03-08
Improve `processBeaconEvents` hotpath

t3chguy:
Improved test by @ara4n from 4230ms to 6.2ms spent in `processBeaconEvents()`

### DIFF
--- a/src/models/room-state.ts
+++ b/src/models/room-state.ts
@@ -18,7 +18,7 @@ import { RoomMember } from "./room-member";
 import { logger } from "../logger";
 import * as utils from "../utils";
 import { EventType, UNSTABLE_MSC2716_MARKER } from "../@types/event";
-import { IEvent, MatrixEvent, MatrixEventEvent } from "./event";
+import { IEvent, MatrixEvent } from "./event";
 import { MatrixClient } from "../client";
 import { GuestAccess, HistoryVisibility, IJoinRuleEventContent, JoinRule } from "../@types/partials";
 import { TypedEventEmitter } from "./typed-event-emitter";
@@ -467,7 +467,7 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
         this.emit(RoomStateEvent.Update, this);
     }
 
-    public processBeaconEvents(events: MatrixEvent[], matrixClient: MatrixClient): void {
+    public async processBeaconEvents(events: MatrixEvent[], matrixClient: MatrixClient): Promise<void> {
         if (
             !events.length ||
             // discard locations if we have no beacons
@@ -476,10 +476,10 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
             return;
         }
 
-        const beaconByEventIdDict: Record<string, Beacon> = [...this.beacons.values()].reduce(
-            (dict, beacon) => ({ ...dict, [beacon.beaconInfoId]: beacon }),
-            {},
-        );
+        const beaconByEventIdDict = [...this.beacons.values()].reduce<Record<string, Beacon>>((dict, beacon) => {
+            dict[beacon.beaconInfoId] = beacon;
+            return dict;
+        }, {});
 
         const processBeaconRelation = (beaconInfoEventId: string, event: MatrixEvent): void => {
             if (!M_BEACON.matches(event.getType())) {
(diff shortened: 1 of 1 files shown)


t3chguy:
Yup the tests were spying on `once` and asserting that is how decryption was watched for, did highlight one issue but the rest were spurious

src/models/room-state.ts
```diff
@@ -493,22 +493,24 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
             }
         };
 
-        events.forEach((event: MatrixEvent) => {
+        for (const event of events) {
```
Johennes, line 496: TIL that `forEach` is up to [3 times slower](https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/#arrayforeach-vs-for-and-forof) than the classic loops. 😳

Johennes: Which makes me wonder if other noticeable performance improvements could be had from banning it elsewhere, too?

kerryarchibald (approved):

src/models/room-state.ts
weeman1337: I tested `forEach` vs. `for of` with their code in the browser.

Interesting results:

Firefox: 
forEach of took: 47
for of took: 77

Chrome:
forEach of took: 3807
for of took: 4095

<details>

<summary>Code</summary>

```
<!DOCTYPE html>
<html lang="de">
    <head>
        <meta charset="utf-8">
        <meta name="viewport" content="width=device-width, initial-scale=1.0">
        <title>Titel</title>
    </head>
    <body>
        <div id="results">

        </div>

        <script>
            console.log("Starting");

            function generateTestArray() {
                const result = [];
                for (let i = 0; i < 10000000; ++i) {
                    result.push({
                        a: i,
                        b: i / 2,
                        r: 0,
                    });
                }
                return result;
            }

            const testArray1 = generateTestArray();
            const testArray2 = generateTestArray();

            const results = document.getElementById("results");

            const before1 = Date.now();
            testArray1.forEach((x) => {
                x.r = x.a + x.b;
            });
            const after1 = Date.now();

            const before2 = Date.now();
            for (const obj of testArray2) {
                obj.r = obj.a + obj.b;
            }
            const after2 = Date.now();

            const result1 = document.createElement("DIV");
            result1.textContent = "forEach of took: " + (after1 - before1);
            results.appendChild(result1);

            const result2 = document.createElement("DIV");
            result2.textContent = "for of took: " + (after2 - before2);
            results.appendChild(result2);
        </script>
    </body>
</html>
```

</details>

Johennes: Oh interesting. Somehow it makes no sense to me that these aren't optimized into the same machine code. 🤔

weeman1337: Yet another benchmark: https://www.measurethat.net/Benchmarks/Show/18189/0/for-vs-foreach-vs-forin-vs-forof
This is more aimed at the overhead of the different ways of iterating an array.

I would conclude: The performance of a loop highly depends on what you do in it.

weeman1337 (approved):

src/models/room-state.ts
gsouquet: It also depends on how the array is interpreted and what it contains.
I found https://v8.dev/blog/elements-kinds interesting to conceptualise some perf optimisation that happen at the JS engine level

I'd optimise for readibility in the source code, and if we found that one thing was really slower than other, looking at transpiling that at build time

gsouquet (approved):

src/models/room-state.ts
t3chguy: FTR the reason for the swap here was simply for access to async/await

Johennes: > Yet another benchmark: [measurethat.net/Benchmarks/Show/18189/0/for-vs-foreach-vs-forin-vs-forof](https://www.measurethat.net/Benchmarks/Show/18189/0/for-vs-foreach-vs-forin-vs-forof) This is more aimed at the overhead of the different ways of iterating an array.

That particular one showed a factor 100 difference between `for of` and `for in` in my browser. I find it mildly irritating that there's so much variation just from the choice of iteration mechanism.

> I'd optimise for readibility in the source code, and if we found that one thing was really slower than other, looking at transpiling that at build time

💯

PR #3203, opened 2023-03-09
Add a new test for event encryption, which works with rust

richvdh:
~~Requires fixes in the rust-sdk including https://github.com/matrix-org/matrix-rust-sdk/pull/1633, so leaving this as a draft for now.~~

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     ],
     "dependencies": {
         "@babel/runtime": "^7.12.5",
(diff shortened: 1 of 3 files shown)


richvdh:
@florianduros perhaps you could take a look at this if you get a minute, since it's related to your work?

florianduros (approved):
LGTM

justjanne (approved):

PR #3205, opened 2023-03-09
Add WebRTC media stats collector

EnricoSchw:
## Checklist

---


gsouquet (commented):
Clearing the review request as this is still a draft.

src/webrtc/stats/transportStats.ts
```diff
@@ -0,0 +1,10 @@
+export class TransportStats {
```
dbkr, line 1: This feels like something that sould be an interface rather than a class?

dbkr (commented):
Some minor comments but this looks like a good approach. Thanks!

src/webrtc/stats/statsCollector.ts
```diff
@@ -0,0 +1,476 @@
...
+
+    public constructor(
+        public readonly callId: string,
+        public readonly remoteUserId: string,
+        private readonly pc: RTCPeerConnection,
+        private readonly isFocus = false,
+    ) {}
+
+    public processStats(groupCallId: string, localUserId: string): Promise<boolean> {
+        if (this.isActive) {
```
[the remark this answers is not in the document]
EnricoSchw: Async/Await and Promises 

I like Promises more  😄  
But you are right. I will change it, because Async/Await is more used in the SDK. 👍

src/webrtc/stats/transportStats.ts
EnricoSchw: Right, I will change it 👍

spec/unit/webrtc/stats/groupCallStats.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+const LOCAL_USER_ID = "LOCAL_USER_ID";
+const TIME_INTERVAL = 10000;
+
+describe("GroupCallStats", () => {
+    let stats: GroupCallStats;
+    beforeEach(() => {
+        stats = new GroupCallStats(GROUP_CALL_ID, LOCAL_USER_ID, TIME_INTERVAL);
+        // @ts-ignore
+        // eslint-disable-next-line no-global-assign
+        global["window"] = {};
```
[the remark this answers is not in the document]
EnricoSchw: Let me try to convince you. But I am also fine to adjust all your `setTimeout` comments.

There two Issues:
The first problem with `setTimeout` is, thats it have different signatures for Browser and Node-js. The Return Value of `setTimeout` in the Browser  is a Number and in Node-js it is Timer. (There for I saw your commend: `ReturnType<typeof setTimeout>`)

The second Issue is thats we develop Typescript in the Node-Js context and not in the Browser Context. Means the global `setTimeout` is the Node-Js time out.  

With `window.setTimeout` the context and the signature is clear without casting.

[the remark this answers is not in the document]
EnricoSchw: By the way this nice theoretical discussion becomes more interesting, when we also talking about developing Workers where some Browser JS functions not existing in the Worker runtime context.

spec/unit/webrtc/stats/groupCallStats.spec.ts
```diff
@@ -0,0 +1,134 @@
+/*
+Copyright 2020 - 2023 The Matrix.org Foundation C.I.C.
```
dbkr, line 2: Probably just copyright 2023?

spec/unit/webrtc/stats/groupCallStats.spec.ts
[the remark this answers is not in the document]
EnricoSchw: Consistency is an argument. I will adjust it 👍

src/webrtc/stats/media/mediaTrackStats.ts
```diff
@@ -0,0 +1,104 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { TrackId } from "./mediaTrackHandler";
+
+export interface PackageLoos {
```
dbkr, line 19:
```suggestion
export interface PacketLoss {
```

EnricoSchw: Oh. Thx

dbkr (commented):
Thanks, looks good. Do we want to keep the stats logging to the console for now? If so, I'd change that to use the `logger` wrapper. Otherwise looks great!

EnricoSchw:
> Thanks, looks good. Do we want to keep the stats logging to the console for now? If so, I'd change that to use the `logger` wrapper. Otherwise looks great!

@dbkr 
The console log is only for debugging and to be able to see the structure of the metrics. I turned off the metric logs for now. I'm currently thinking about two options:

1. Broadcast metrics data events from SDK to ElementCall like you did it with device msg Events. So in EC we can decide how to display metrics (OTEL, Console, Posthog, Html-View ) 

- [ ] You would prefer this option?

2. Make it configurable via MatrxClient, how to display or where we send the metrics

- [ ] You would prefer this option?

(I prefer the first option, but with the option to disable the metrics as well)

The next steps for the metrics are:
-- Augment logs with feed data
-- Include pkg send/receive metric data for feed -tracks.
-- Add transceiver data in feed metrics

dbkr:
I vote option 1 personally. I think the opentelemetry stuff will probably move in to the js-sdk in the long run, but we can get it out into EC with an event in the same way, that sounds ideal for now.

PR #3211, opened 2023-03-10
Pull out a public `CryptoApi`

richvdh:
This is a first step towards fixing https://github.com/vector-im/element-web/issues/17488. We create a new `CryptoApi` which is a subset of `CryptoBackend`, and defines the crypto interfaces which are meant to be used directly by applications.

Obviously, this is just a very initial step: in time we'll need to more all the crypto-related functions from `MatrixClient` in here. But some of those are likey to change signature slightly, or not be supported at all, as we move to the Rust SDK, and I only want to add things to `CryptoApi` once we're reasonably sure that it represents the final API shape.

---


richvdh:
more tests are in https://github.com/matrix-org/matrix-js-sdk/pull/3203

gsouquet (approved):
Looks sane 🚀

PR #3218, opened 2023-03-16
Fire `closed` event when IndexedDB closes unexpectedly

t3chguy:
https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase/close_event
For https://github.com/vector-im/element-web/issues/24831

---

## ✨ Features
 * Fire `closed` event when IndexedDB closes unexpectedly ([\#3218](https://github.com/matrix-org/matrix-js-sdk/pull/3218)).

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -40,7 +40,7 @@ export interface IStore {
 
     // XXX: The indexeddb store exposes a non-standard emitter for the "degraded" event
     // for when it falls back to being a memory store due to errors.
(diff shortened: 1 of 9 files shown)


src/store/index.ts
```diff
@@ -40,7 +40,7 @@ export interface IStore {
 
     // XXX: The indexeddb store exposes a non-standard emitter for the "degraded" event
     // for when it falls back to being a memory store due to errors.
```
robintown, line 42: Should this comment also explain what `"closed"` is doing here, perhaps?

robintown (approved):

PR #3227, opened 2023-03-23
Stop doing O(n^2) work to find event's home

jryans:
In certain rooms (e.g. with many state changes hidden via user preferences), the events array presented to `eventShouldLiveIn` may contain 100s of events. As part of its various checks, `eventShouldLiveIn` would get an event's associated ID (reply / relation / redaction parent). It would then use `events.find` to search the entire (possibly large) `events` array to look for the parent. (This by itself seems sub-optimal and should probably change to use a map.)

For many events in a room, there is no associated ID. Unfortunately, `eventShouldLiveIn` did not check whether the associated ID actually exists before running off to search all of `events`, resulting in O(n^2) work.

This changes `eventShouldLiveIn` to first check that there is an associated ID before proceeding with its (slow) search. For some rooms, this change drastically improves performance from ~100% CPU usage to nearly idle.

## Profile before

image

## Profile after

image

## Checklist

---

## 🐛 Bug Fixes
 * Stop doing O(n^2) work to find event's home ([\#3227](https://github.com/matrix-org/matrix-js-sdk/pull/3227)). Contributed by @jryans.

### DIFF
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -2028,8 +2028,11 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
             };
         }
 
-        const parentEventId = event.getAssociatedId()!;
-        const parentEvent = this.findEventById(parentEventId) ?? events?.find((e) => e.getId() === parentEventId);
+        const parentEventId = event.getAssociatedId();
+        let parentEvent;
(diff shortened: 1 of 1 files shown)


t3chguy (approved):
:shipit: LGTM otherwise

Thanks for this!!

PR #3230, opened 2023-03-24
Element-R: handle events which arrive before their keys

richvdh:
Currently, any event which is received before the decryption keys will not be decrypted. We can fix this by retrying decryption after keys are received.

Fixes https://github.com/vector-im/element-web/issues/24489
~~Requires https://github.com/matrix-org/matrix-rust-sdk/pull/1704~~

---

## 🐛 Bug Fixes
 * Element-R: handle events which arrive before their keys ([\#3230](https://github.com/matrix-org/matrix-js-sdk/pull/3230)). Fixes vector-im/element-web#24489.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     ],
     "dependencies": {
         "@babel/runtime": "^7.12.5",
-        "@matrix-org/matrix-sdk-crypto-js": "^0.1.0-alpha.5",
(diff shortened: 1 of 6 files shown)


weeman1337 (approved):
:+1: looks good to me

t3chguy (approved):

PR #3234, opened 2023-03-28
Refactor the way group calls hang up

dbkr:
We previously used disposeCall to terminate the call which meant that sometimes a call would never get a hangup event. This changes it so that we always end a call by calling hangup, then do the cleanup when the hangup event arrives, so the cleanup is the same whether we hang up or the other side does.

This means we have somewhere we can reliably end OpenTelemetry spans (we were seeing them sometimes close but other times not).

## Checklist

---

### DIFF
--- a/src/webrtc/groupCall.ts
+++ b/src/webrtc/groupCall.ts
@@ -495,6 +495,11 @@ export class GroupCall extends TypedEventEmitter<
             this.retryCallLoopInterval = undefined;
         }
 
+        if (this.participantsExpirationTimer !== null) {
+            clearTimeout(this.participantsExpirationTimer);
+            this.participantsExpirationTimer = null;
+        }
+
         if (this.state !== GroupCallState.Entered) {
             return;
         }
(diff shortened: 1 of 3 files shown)


src/webrtc/groupCall.ts
```diff
@@ -495,12 +497,16 @@ export class GroupCall extends TypedEventEmitter<
...
+            clearTimeout(this.participantsExpirationTimer);
+            this.participantsExpirationTimer = null;
+        }
+
         if (this.state !== GroupCallState.Entered) {
             return;
         }
 
-        this.forEachCall((call) => this.disposeCall(call, CallErrorCode.UserHangup));
-        this.calls.clear();
```
EnricoSchw, line 503: The calls map do not need clean up anymore?

dbkr: They should each be removed when the individual call is cleaned up. This was essentially doing it in a brute force way, but more importantly it was preempting the calls cleaning themselves up.

EnricoSchw (approved):

florianduros (approved):
LGFM, thank you for the explanations!

PR #3235, opened 2023-03-28
Update changelog for v24.0.0 now the security issue is public

andybalaam:
---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 Changes in [24.0.0](https://github.com/matrix-org/matrix-js-sdk/releases/tag/v24.0.0) (2023-03-28)
 ==================================================================================================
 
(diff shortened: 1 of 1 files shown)


weeman1337 (approved):

t3chguy:
@andybalaam the changelog is also published to NPM https://www.npmjs.com/package/matrix-js-sdk?activeTab=code

andybalaam:
> @andybalaam the changelog is also published to NPM https://www.npmjs.com/package/matrix-js-sdk?activeTab=code

I'm not sure how to update this without pushing a new package. Is that what you are suggesting?

andybalaam:
> I'm not sure either, we really should not have broken the changelog in the first place.

Can you write about this in our retrospective doc? I didn't see an alternative.

PR #3237, opened 2023-03-29
Re-apply "Add the call object to Call events"

dbkr:
Re-applies https://github.com/matrix-org/matrix-js-sdk/pull/3229

Merge https://github.com/matrix-org/matrix-react-sdk/pull/10477 immediately afterwards

---


dbkr:
This has already been through review in the original PR, so merging.

PR #3238, opened 2023-03-30
Fix twice time renegotiation handling if reuse a transceiver

EnricoSchw:
If we use a new transceiver and attach a new track, then we should not swap the stream because that will trigger additional negotiation. Please check description: See: https://github.com/vector-im/element-call/issues/944

## Checklist


dbkr (approved):
Oh, of course! Well found. It stands to reason that this could end up confusing things, so I don't see why we'd try to make the stream match rather than just accepting that it will be different, just like the track ID will after a `replaceTrack`.

PR #3244, opened 2023-03-31
Move `SecretStorage`-related interfaces out to new module

richvdh:
We have a number of interfaces related to SecretStorage defined either in `crypto/api` or `crypto/SecretStorage`. This isn't entirely appropriate, because the SecretStorage implementation is orthogonal to the E2EE implementation.

This PR moves those interfaces out to `secret-storage.ts`, which was started in #3242. It is part of a larger piece of work in refactoring the SecretStorage interface (vector-im/element-web#24982).

Suggest reviewing commit-by-commit

---

### DIFF
--- a/src/secret-storage.ts
+++ b/src/secret-storage.ts
@@ -125,7 +125,7 @@ export interface AccountDataClient extends TypedEventEmitter<ClientEvent.Account
      * @param eventType - The type of account data
      * @returns The contents of the given account data event.
      */
-    getAccountDataFromServer: <T extends { [k: string]: any }>(eventType: string) => Promise<T>;
+    getAccountDataFromServer: <T extends Record<string, any>>(eventType: string) => Promise<T>;
 
     /**
      * Set account data event for the current user, with retries
--- a/src/secret-storage.ts
+++ b/src/secret-storage.ts
@@ -86,3 +86,23 @@ export interface PassphraseInfo {
     /** The number of bits to generate. Defaults to 256. */
     bits?: number;
 }
+
+/**
+ * Options for {@link SecretStorage#addKey}.
+ */
+export interface AddSecretStorageKeyOpts {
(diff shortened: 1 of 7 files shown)


src/secret-storage.ts
```diff
@@ -86,3 +91,48 @@ export interface PassphraseInfo {
...
+export interface AccountDataClient extends TypedEventEmitter<ClientEvent.AccountData, ClientEventHandlerMap> {
+    /**
+     * Get account data event of given type for the current user. This variant
+     * gets account data directly from the homeserver if the local store is not
+     * ready, which can be useful very early in startup before the initial sync.
+     *
+     * @param eventType - The type of account data
+     * @returns The contents of the given account data event.
+     */
+    getAccountDataFromServer: <T extends { [k: string]: any }>(eventType: string) => Promise<T>;
```
alunturner, line 128: I _think_ this could be written as `<T extends Record<string, any>>` if you were so inclined. I personally prefer the `Record` syntax for things like this and as far as I know they are completely equivalent in this case.

Doesn't block my approval anyhow.

alunturner (approved):
Looks good to me!

src/secret-storage.ts
richvdh: I agree with you, though in this case I was deliberately trying to move things without worrying too much about polishing their typescript definitions. (In this case, it's just been moved from https://github.com/matrix-org/matrix-js-sdk/pull/3244/files/e290ec3130a0206e1d329edf085a15662ae27cc5#diff-cef792a7fe5292ff218b44701c423e282ae1b8df338a537d576851781430da7fL46).

still, might as well fix this now you've pointed it out.

```suggestion
    getAccountDataFromServer: <T extends Record<string, any>>(eventType: string) => Promise<T>;
```

PR #3247, opened 2023-04-03
Skip sonarcloud & coverage in merge_queue

t3chguy:
They're already checked on the PR

Fixes issue with merge queue always failing if new code coverage on develop < 80%

---


andybalaam (approved):
Thanks, should improve our lives.

PR #3248, opened 2023-04-04
Implement MSC 3981

justjanne:
Type: Enhancement
Fixes: https://github.com/vector-im/element-web/issues/25021
Related: https://github.com/matrix-org/matrix-spec-proposals/pull/3981

---

## ✨ Features
 * Implement MSC 3981 ([\#3248](https://github.com/matrix-org/matrix-js-sdk/pull/3248)). Fixes vector-im/element-web#25021. Contributed by @justjanne.

### DIFF
--- a/src/@types/requests.ts
+++ b/src/@types/requests.ts
@@ -186,6 +186,7 @@ export interface IRelationsRequestOpts {
     to?: string;
     limit?: number;
     dir?: Direction;
+    recurse?: boolean;
 }
 
 export interface IRelationsResponse {
--- a/src/@types/requests.ts
+++ b/src/@types/requests.ts
@@ -186,7 +186,7 @@ export interface IRelationsRequestOpts {
     to?: string;
     limit?: number;
     dir?: Direction;
-    recurse?: boolean;
+    recurse?: boolean; // MSC3981 Relations Recursion https://github.com/matrix-org/matrix-spec-proposals/pull/3981
 }
 
 export interface IRelationsResponse {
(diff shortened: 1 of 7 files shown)


src/@types/requests.ts
```diff
@@ -186,6 +186,7 @@ export interface IRelationsRequestOpts {
     to?: string;
     limit?: number;
     dir?: Direction;
+    recurse?: boolean;
```
t3chguy, line 189: A comment linking it to MSC3981 would be good

src/models/thread.ts
```diff
@@ -458,25 +459,27 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
...
-                            if (relations.events.length) {
-                                event.makeReplaced(relations.events[0]);
-                            }
-                        })
-                        .catch((e) => {
-                            logger.error("Failed to load edits for encrypted thread event", e);
-                        });
-                }),
-        );
+        if (!this.client.canSupport.get(Feature.RelationsRecursion)) {
```
t3chguy, line 462: This looks wrong, shouldn't it be `!== Unsupported` - right now it says if not stable, so doesn't allow unstable behaviour.

justjanne: You're right! Here I actually made the very error that I used as justification for why we shouldn't coerce values to booleans like this^^

t3chguy: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md can prevent it ftr

t3chguy (approved):

PR #3249, opened 2023-04-04
Measure whether we receive media and add the mute state as an exception

EnricoSchw:
We want to measure whether we receive media. This value serves as an indicator that the media pipeline is healthy. But in case of video is muted we don't get any bytes. Therefore we check the mute state of a track and treat it as if we received bytes. This pull request contains the calculation of muted track states. The determination remote track is mutes is made using the receiver track of a WbRTC connection.

## Checklist

---


src/webrtc/stats/media/mediaTrackStats.ts
```diff
@@ -101,4 +104,28 @@ export class MediaTrackStats {
     public resetBitrate(): void {
         this.bitrate = { download: 0, upload: 0 };
     }
+
+    public set alive(isAlive: boolean) {
+        this.isAlive = isAlive;
+    }
+
+    public get alive(): boolean {
```
dbkr, line 112: The explanation of what 'alive' means could go here maybe?

dbkr (approved):
Otherwise lgtm

src/webrtc/stats/media/mediaTrackStats.ts
EnricoSchw: Good point. I will add descriptions here.

t3chguy:
Known Wysiwyg flake.

PR #3250, opened 2023-04-04
Switch some imports to type imports

dbkr:
Having these as regular imports confuses Vite for some reason.

## Checklist

---


t3chguy (approved):
Webpack was complaining too FTR

PR #3251, opened 2023-04-04
Add an event emitted when a Call creates a PeerConnection

dbkr:
For apps that need acces to the per connection as soon as it's created for debugging etc.

## Checklist

---

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -152,6 +152,10 @@ export enum CallEvent {
     DataChannel = "datachannel",
 
     SendVoipEvent = "send_voip_event",
+
+    // When the call instantiates its peer connection
+    // For apps that want to access the underlying peer connection, eg for debugging
+    PeerConnectionCreated = "peer_connection_created",
 }
 
 export enum CallErrorCode {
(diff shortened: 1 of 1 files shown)


src/webrtc/call.ts
```diff
@@ -325,6 +329,7 @@ export type CallEventHandlerMap = {
     /* @deprecated */
     [CallEvent.HoldUnhold]: (onHold: boolean) => void;
     [CallEvent.SendVoipEvent]: (event: VoipEvent, call: MatrixCall) => void;
+    [CallEvent.PeerConnectionCreated]: (peerConn: RTCPeerConnection, call: MatrixCall) => void;
```
EnricoSchw, line 332: It's a pragmatic approach.  By exposing the entire object, we could have left the stats collecting to the calling client.

EnricoSchw (approved):

src/webrtc/call.ts
dbkr: Yeah, this seemed easier and most flexible. We already expose the peer connection anyway, there was just no way to know as soon as it was created.

PR #3215, opened 2023-03-14
Send one time key count and unused fallback keys for rust-crypto

florianduros:
## Checklist

-   [x] Tests written for new code (and old code if feasible)
-   [x] Linter and other CI checks pass
-   [x] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))

Fix https://github.com/vector-im/element-web/issues/24795#event-8732738550

---
Here's what your changelog entry will look like:

## ✨ Features
 * Send one time key count and unused fallback keys for rust-crypto ([\#3215](https://github.com/matrix-org/matrix-js-sdk/pull/3215)). Fixes vector-im/element-web#24795. Contributed by @florianduros.

PR #3253, opened 2023-04-05
Use `processKeyCounts` in sliding sync

florianduros:
## Checklist

Following comment about sliding sync in https://github.com/matrix-org/matrix-js-sdk/pull/3215#discussion_r1144640847

The purpose is to use the `processKeyCounts` method of the `CryptoBackend` interface  to process one time key counts and unused fall back keys after a `/sync` instead of using a duplicated behaviour

The sliding sync tests about it are eased because the `processKeyCounts` behaviour is already tested in https://github.com/matrix-org/matrix-js-sdk/blob/develop/spec/integ/crypto.spec.ts#L1966

Also
- `processKeyCounts` is implemented by old and new crypto (`CryptoBackend` interface) and should be used instead of `updateOneTimeKeyCount` for old crypto
- `updateOneTimeKeyCount` is only used by `processKeyCounts` so I made it private (sliding sync was the last one to use it)
- Same logic for `setNeedsNewFallback` but since the method has one line of code, I removed it.

---

### DIFF
--- a/spec/integ/sliding-sync-sdk.spec.ts
+++ b/spec/integ/sliding-sync-sdk.spec.ts
@@ -671,32 +671,17 @@ describe("SlidingSyncSdk", () => {
             // TODO: more assertions?
         });
 
-        it("can update OTK counts", () => {
-            client!.crypto!.updateOneTimeKeyCount = jest.fn();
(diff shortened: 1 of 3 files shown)


richvdh (approved):
lgtm

PR #3254, opened 2023-04-05
Element-R: pass device list change notifications into rust crypto-sdk

florianduros:
## Checklist

Closes https://github.com/vector-im/element-web/issues/24795

Pass device list from `/sync` to the new crypto.

- I renamed the `handleDeviceListChanges` method into `processDeviceLists` to match the method names in the `CryptoBackend` interface. *ie:* `processKeyCounts`, `preprocessToDeviceMessages`
- I renamed the `syncDeviceList` parameter to `deviceList` to match the parameter names in `processXX` or `preprocessXX` methods.

---

## ✨ Features
 * Element-R: pass device list change notifications into rust crypto-sdk ([\#3254](https://github.com/matrix-org/matrix-js-sdk/pull/3254)). Fixes vector-im/element-web#24795. Contributed by @florianduros.


src/rust-crypto/rust-crypto.ts
```diff
@@ -229,6 +235,23 @@ export class RustCrypto implements CryptoBackend {
...
+     * @param syncData - Object containing sync tokens associated with this sync
+     * @param deviceLists - device_lists field from /sync, or response from
+     */
+    public async processDeviceLists(
+        syncData: ISyncStateData,
+        deviceLists: Required<ISyncResponse>["device_lists"],
+    ): Promise<void> {
+        // Initial syncs don't have device change lists. We'll either get the complete list
+        // of changes for the interval or will have invalidated everything in willProcessSync
+        if (!syncData.oldSyncToken || !deviceLists) return;
```
florianduros, line 249: I add the same verification that in the old crypto but I don't know if it's relevant here.

src/rust-crypto/rust-crypto.ts
```diff
@@ -229,6 +234,23 @@ export class RustCrypto implements CryptoBackend {
         }
     }
 
+    /** called by the sync loop to process
```
richvdh, line 237: to process ... ?

src/crypto/index.ts
```diff
@@ -2897,16 +2897,16 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
...
-    public async handleDeviceListChanges(
+    public async processDeviceLists(
         syncData: ISyncStateData,
-        syncDeviceLists: Required<ISyncResponse>["device_lists"],
+        deviceLists?: Required<ISyncResponse>["device_lists"],
     ): Promise<void> {
         // Initial syncs don't have device change lists. We'll either get the complete list
         // of changes for the interval or will have invalidated everything in willProcessSync
-        if (!syncData.oldSyncToken) return;
+        if (!syncData.oldSyncToken || !deviceLists) return;
```
richvdh, line 2909: I don't think we really need to check `oldSyncToken` here. As the comment says: if `oldSyncToken` is absent, then so will `deviceLists` be, so the check on `oldSyncToken` is redundant, which means that we don't need to pass in `syncData` here.

richvdh, line 2909: Suggest we move the check on whether `deviceLists` is defined or not into `sync.ts` and `sliding-sync-sdk.ts`: it simplifies the interface for this API.

src/common-crypto/CryptoBackend.ts
```diff
@@ -113,6 +115,16 @@ export interface SyncCryptoCallbacks {
...
 
+    /**
+     * Handle the notification from /sync or /keys/changes that device lists have
+     * been changed.
+     *
+     * @param syncData - Object containing sync tokens associated with this sync
+     * @param deviceLists - device_lists field from /sync, or response from
+     * /keys/changes
+     */
+    processDeviceLists(syncData: ISyncStateData, deviceLists?: Required<ISyncResponse>["device_lists"]): Promise<void>;
```
richvdh, line 126: can we define the type of `deviceLists` from first principles rather than by reference to `ISyncResponse`? It's an `IDeviceLists`, which is much simpler.

src/common-crypto/CryptoBackend.ts
```diff
@@ -113,6 +115,16 @@ export interface SyncCryptoCallbacks {
      */
     processKeyCounts(oneTimeKeysCounts?: Record<string, number>, unusedFallbackKeys?: string[]): Promise<void>;
 
+    /**
+     * Handle the notification from /sync or /keys/changes that device lists have
+     * been changed.
```
richvdh, line 120: currently we only seem to call it from `/sync`. Is there an intention to change this?

richvdh: (looks like the doc-comment on `handleDeviceListChanges` is some hangover of a long-forgotten era)

src/common-crypto/CryptoBackend.ts
```diff
@@ -113,6 +115,16 @@ export interface SyncCryptoCallbacks {
...
+     *
+     * @param syncData - Object containing sync tokens associated with this sync
+     * @param deviceLists - device_lists field from /sync, or response from
+     * /keys/changes
```
richvdh, line 124: again, not `/keys/changes`

richvdh (changes requested):

src/crypto/index.ts
florianduros: How moving the check on the upper level simplifies the interface in this case ? The `deviceLists` is still needed in the method signature

src/common-crypto/CryptoBackend.ts
florianduros: Can we use instead the `IDeviceLists` definition in the `sync-accumulator.ts` ? (Where from `IToDeviceEvent` definition is also coming)

src/rust-crypto/rust-crypto.ts
florianduros: Oops. Will fix it

src/common-crypto/CryptoBackend.ts
richvdh: I'm afraid I don't understand the question. Yes, `IDeviceLists` is defined in `sync-accumulator.ts`, and we should import it here.

florianduros: I misunderstood your first comment, I repeated the same idea :)

richvdh (approved):
lgtm otherwise

richvdh:
I'm just going to merge this, despite the quality gate not being satisfied. Ultimately this sort of thing can be covered in `devicelist-integ.spec.ts`, but that is currently tied into the old crypto, which is https://github.com/vector-im/element-web/issues/24833.

PR #3266, opened 2023-04-06
Export type for return of getCapabilities()

hughns:
Renamed because it clashes with ICapabilities from embedded. I would happily take alternative naming suggestions.

## Checklist

---


t3chguy (approved):
If we're renaming it can we drop the `I` prefix, we stopped doing that a while ago

t3chguy (approved):

PR #3267, opened 2023-04-06
Split `SecretStorage` into two parts

richvdh:
The existing `SecretStorage` class is really a combination of two orthogonal concepts: secret *storage* and secret *sharing*. There is some awful type-wrangling to make it work, and it's problematic for Element R, because the rust crypto-sdk implements the sharing part but not the storage part.

So, it's time to split it up into its component parts. This PR attempts to do so without breaking backwards-compatibity, by leaving the old `SecretStorage` class in place.

Part of https://github.com/vector-im/element-web/issues/24982, though there remains work to be done in exposing the new `SecretStorage` without a legacy `Crypto`.

Suggest review commit-by-commit

---


src/secret-storage.ts
```diff
@@ -146,3 +150,340 @@ export interface SecretStorageCallbacks {
...
+
+/**
+ * Implementation of Server-side secret storage.
+ *
+ * Secret *sharing* is *not* implemented here: this class is strictly about the storage component of
+ * SSSS.
+ *
+ * @see https://spec.matrix.org/v1.6/client-server-api/#storage
+ */
+export class SecretStorage {
```
richvdh: It's highly regrettable that this gives us two different classes called `SecretStorage`. On the other hand, I can't rename the existing one without breaking backwards compat, and I can't think of a better name for the new one. Suggestions welcome.

src/secret-storage.ts
```diff
@@ -162,6 +162,89 @@ interface Decryptors {
     decrypt: (ciphertext: IEncryptedPayload) => Promise<string>;
 }
 
+/**
+ * Interface provided by SecretStorage implementations
+ *
+ * Normally this will just be an {@link SecretStorage}, but for backwards-compatibility some methods allow other
+ * implementations.
+ */
+export interface ISecretStorage {
```
richvdh, line 171: Again, this isn't a great name: we're not supposed to call our interfaces `IFoo`.

(Maybe this should just be `SecretStorage`, and the new class should be `SecretStorageImpl` ?)

spec/unit/crypto/secrets.spec.ts
```diff
@@ -81,26 +76,6 @@ describe("Secrets", function () {
         return global.Olm.init();
     });
 
-    it("should allow storing a default key", async function () {
```
richvdh, line 84: unlike the rest of the tests in this file, this one instantiates its own SecretStorage and uses that, rather than going via the instance in the test client. In fact, I'd say that all the rest of the tests are much closer integration tests, despite the name of the file.

Anyway, this test is not like the others, so is a good candidate for moving. The rest need sorting out another time.

src/secret-storage.ts
weeman1337: What about `ServerSideSecretStorage` as a name for the server side secret storage implementation?

src/secret-storage.ts
```diff
@@ -136,3 +140,433 @@ export interface AccountDataClient extends TypedEventEmitter<ClientEvent.Account
      */
     setAccountData: (eventType: string, content: any) => Promise<{}>;
 }
+
+/**
+ *  Application callbacks for use with {@link SecretStorage}
+ */
+export interface SecretStorageCallbacks {
+    getSecretStorageKey?: (
+        keys: { keys: Record<string, SecretStorageKeyDescription> },
```
weeman1337, line 149: Can you please add docs what `string` should be here and how it relates to the `string` of the result?

src/secret-storage.ts
```diff
@@ -136,3 +140,433 @@ export interface AccountDataClient extends TypedEventEmitter<ClientEvent.Account
...
+        name: string,
+    ) => Promise<[string, Uint8Array] | null>;
```
weeman1337, line 151: A description about what the `Uint8Array` is would be helpful here.

src/secret-storage.ts
```diff
@@ -136,3 +140,433 @@ export interface AccountDataClient extends TypedEventEmitter<ClientEvent.Account
...
+    store(name: string, secret: string, keys?: string[] | null): Promise<void>;
+
+    /**
+     * Get a secret from storage.
+     *
+     * @param name - the name of the secret
+     *
+     * @returns the contents of the secret
+     */
+    get(name: string): Promise<string | undefined>;
```
weeman1337, line 234: Does `undefined` here mean there is no such secret in the store? Can you document it?

src/secret-storage.ts
```diff
@@ -136,3 +140,433 @@ export interface AccountDataClient extends TypedEventEmitter<ClientEvent.Account
...
+            keyInfo.passphrase = opts.passphrase;
+        }
+        if (opts.key) {
+            const { iv, mac } = await calculateKeyCheck(opts.key);
+            keyInfo.iv = iv;
+            keyInfo.mac = mac;
+        }
+
+        if (!keyId) {
+            do {
```
weeman1337, line 335: This construct is worth a comment. It does loop until it finds an unused key ID I think…

weeman1337 (commented):
Thanks for splitting this up @richvdh . I think it makes sense :+1: 

Some places can be improved with better docs. Also see my proposal for a name of the storage class https://github.com/matrix-org/matrix-js-sdk/pull/3267#discussion_r1159712486 . What do you think about it?

src/secret-storage.ts
richvdh: Honestly if you can figure it out you're doing better than me. The react-sdk's implementation of this method doesn't even use this parameter :(.

I'll spend some time trying to understand.

weeman1337 (approved):
Nevermind! If you can add some docs that would be great. Otherwise I would say doc blocks are out of scope of this PR.

src/secret-storage.ts
richvdh: As discussed elsewhere: I've gone for `ServerSideSecretStorage` for the interface and `ServerSideSecretStorageImpl` for the implementation.

PR #3269, opened 2023-04-11
Improve output in github actions for jest tests

richvdh:
Two functional changes here:
 * Set `silent: false` for the [GitHub Actions Reporter](https://jestjs.io/docs/configuration#github-actions-reporter), which stops it swallowing failure reasons (otherwise it only puts it in an annotation)sends it to annotations)
 * Enable the `summary` reporter, which makes it easier to see why a test run as a whole failed.

Setting `silent: false` requires us to use a config file, and using a config file overrides the config from `package.json`. So in the end it's easier just to have one config file with some intelligence for whether we are in GHA or not.

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,27 +122,11 @@
         "prettier": "2.8.7",
         "rimraf": "^4.0.0",
         "terser": "^5.5.1",
+        "ts-node": "^10.9.1",
         "tsify": "^5.0.2",
         "typedoc": "^0.23.20",
         "typedoc-plugin-missing-exports": "^1.0.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -122,28 +122,12 @@
         "prettier": "2.8.7",
         "rimraf": "^4.0.0",
         "terser": "^5.5.1",
+        "ts-node": "^10.9.1",
```
richvdh, line 125: this is needed to support having a jest config file in typescript

t3chguy (approved):
:shipit:

PR #3272, opened 2023-04-12
Element-R: wire up device lists

florianduros:
## Checklist

Related to https://github.com/vector-im/element-web/issues/24833

The `MatrixClient.downloadKeys` and `MatrixClient.getStoredDevicesForUser` methods are not available for the new crypto. Following the idea in https://github.com/vector-im/element-web/issues/24833 to add a new method `getUserDeviceInfo` to the crypto API.

Instead of changing the existing `downloadKeys` and `getStoredDevicesForUser` for both crypto and changing their behaviors, a new method `getUserDeviceInfo` is also added to the `crypto-api`.

This method has the following behavior:
1. Check if the users are currently tracked
2. If they are not tracked and `downloadUncached` is at `true` (in order to get the devices of theses untracked users)
	- *old crypto* `downloadKeys` is called (which is calling `/keys/query` later)
	- *new crypto* make a request to `/keys/query`
3. Do a device conversion (`DeviceInfo`, `QueryDevice`, `RustSdkCryptoJs.Device` ) to a unique new `Device` class
4. Return the content in maps

---

## ✨ Features
 * Element-R: wire up device lists ([\#3272](https://github.com/matrix-org/matrix-js-sdk/pull/3272)). Contributed by @florianduros.


src/models/device.ts
```diff
@@ -0,0 +1,82 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { DeviceVerification } from "../crypto/deviceinfo";
+
+// user-Id → device-Id → IDevice
```
florianduros, line 19: Replace `IDevice` by `Device`

t3chguy (approved):
Looks sane :shipit:

src/models/device.ts
```diff
@@ -0,0 +1,82 @@
...
+export type DeviceMap = Map<string, Map<string, Device>>;
+
+type DeviceParameters = Pick<Device, "deviceId" | "userId" | "algorithms" | "keys"> & Partial<Device>;
+
+export class Device {
+    /** id of the device */
+    public readonly deviceId: string;
+
+    /** id of the device user */
```
richvdh, line 28:
```suggestion
    /** id of the user that owns the device */
```

src/client.ts
```diff
@@ -2317,6 +2318,23 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+     * @returns A map `{@link DeviceMap}`.
+     */
+    public getUserDeviceInfo(userIds: string[], downloadUncached = false): Promise<DeviceMap> {
+        if (!this.cryptoBackend) {
+            throw new Error("End-to-end encryption disabled");
+        }
+
+        return this.cryptoBackend.getUserDeviceInfo(userIds, downloadUncached);
+    }
+
```
richvdh, line 2337: please don't add more methods to `MatrixClient`. callers should call `getCryptoApi.getUserDeviceInfo`

richvdh (changes requested):

src/models/device.ts
```diff
@@ -0,0 +1,82 @@
...
+*/
+
+import { DeviceVerification } from "../crypto/deviceinfo";
+
+// user-Id → device-Id → IDevice
+export type DeviceMap = Map<string, Map<string, Device>>;
+
+type DeviceParameters = Pick<Device, "deviceId" | "userId" | "algorithms" | "keys"> & Partial<Device>;
+
+export class Device {
```
richvdh, line 24: also please can you make sure this and DeviceMap get exported via `matrix.ts` so that they appear in the main bit of the typedoc (currently they are marked as "internal": https://pr3272--js-sdk-docs-previews.netlify.app/stable/classes/_internal_.device)

src/crypto/device-converter.ts
```diff
@@ -0,0 +1,45 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { Device } from "../models/device";
+import { DeviceInfo } from "./deviceinfo";
+
+/**
+ * Convert a DeviceInfo {@link DeviceInfo} to a Device {@link Device}
```
richvdh, line 21: You don't need to repeat the contents of the link:

```suggestion
 * Convert a {@link DeviceInfo} to a {@link Device}.
```

src/crypto/device-converter.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @param deviceInfo - deviceInfo to convert
+ * @param userId - id of the device user
```
richvdh, line 23:
```suggestion
 * @param userId - ID of the user that owns the device.
```

spec/unit/device-converter.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { DeviceInfo, DeviceVerification } from "../../src/crypto/deviceinfo";
+import { DeviceKeys } from "../../src";
+import { downloadDeviceToJsDevice } from "../../src/rust-crypto/device-converter";
+import { deviceInfoToDevice } from "../../src/crypto/device-converter";
+
+describe("device-converter", () => {
```
richvdh, line 22: please could you separate the tests for the rust `device-converter` and the legacy one into separate files?

src/crypto/deviceinfo.ts
```diff
@@ -25,7 +25,7 @@ export interface IDevice {
     signatures?: ISignatures;
 }
 
-enum DeviceVerification {
+export enum DeviceVerification {
```
richvdh, line 28: suggest moving this into `models/device.ts` while you're here. I want to kill off the whole `crypto` directory eventually.

src/crypto/index.ts
```diff
@@ -2063,6 +2065,54 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
...
+        // Keep the users without device to download theirs keys
+        const usersWithoutDeviceInfo: string[] = [];
+
+        for (const userId of userIds) {
+            const deviceInfos = await this.getStoredDevicesForUser(userId);
+            // If there are device infos for a userId, we transform it into a map
+            // Else, the keys will be downloaded after
+            if (deviceInfos) {
+                const deviceMap = new Map(
+                    // Convert DeviceInfo to IDevice
```
richvdh, line 2088:
```suggestion
                    // Convert DeviceInfo to Device
```

src/rust-crypto/rust-crypto.ts
```diff
@@ -162,6 +165,91 @@ export class RustCrypto implements CryptoBackend {
...
+     */
+    private isInRustUserIds(userId: string, rustUserIds: Set<RustSdkCryptoJs.UserId>): boolean {
+        for (const rustUserId of rustUserIds) {
+            if (rustUserId.toString() === userId) {
+                return true;
+            }
+        }
+
+        return false;
+    }
```
richvdh, line 221: factor this out to a function outside the class?

src/rust-crypto/rust-crypto.ts
```diff
@@ -162,6 +165,91 @@ export class RustCrypto implements CryptoBackend {
...
+        const trackedUsers: Set<RustSdkCryptoJs.UserId> = await this.olmMachine.trackedUsers();
+
+        // Keep untracked user to download their keys after
+        const untrackedUsers: Set<string> = new Set();
+
+        for (const userId of userIds) {
+            // if this is a tracked user, we can just fetch the device list from the rust-sdk
+            // (NB: this is probably ok even if we race with a leave event such that we stop tracking the user's
+            // devices: the rust-sdk will return the last-known device list, which will be good enough.)
+            if (this.isInRustUserIds(userId, trackedUsers)) {
```
richvdh, line 188: this is a bit O(N^2). it might be better to convert `trackedUsers` to a set of strings upfront.

src/rust-crypto/rust-crypto.ts
```diff
@@ -162,6 +165,91 @@ export class RustCrypto implements CryptoBackend {
...
+            if (rustUserId.toString() === userId) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * Get the user devices from the olm machine
```
richvdh, line 224:
```suggestion
     * Get the device list for the given user from the olm machine
```

src/rust-crypto/device-converter.ts
```diff
@@ -0,0 +1,129 @@
...
+                break;
+        }
+    });
+
+    // Add device display name to unsigned field
+    const unsigned = new Map<string, string>();
+    const { displayName } = device;
+    if (displayName) {
+        unsigned.set("device_display_name", displayName);
+    }
```
richvdh, line 79: hum. It might be better to give `Device` a `displayName` property, rather than an `unsigned` property which doesn't really contain the unsigned data.

richvdh (changes requested):
Thanks florian! Looks great but a few suggestions...

src/rust-crypto/rust-crypto.ts
florianduros: Oh yes, great idea

src/crypto-api.ts
```diff
@@ -77,6 +77,10 @@ export interface CryptoApi {
     /**
      * Get the device information for the given list of users.
      *
+     * First of all, the devices are retrieved from the tracked users in the local client
+     * If the users are not tracked and the `downloadUncached` parameter is set at `true`,
+     * A `/keys/query` request is made to the local user homeserver to retrieve these devices
```
richvdh, line 82: This assumes the user knows what a "tracked user" is, which is an internal detail.

```suggestion
     * For any users whose device lists are cached (due to sharing an encrypted room with the user), the
     * cached device data is returned.
     *
     * If there are uncached users,  and the `downloadUncached` parameter is set to `true`,
     * a `/keys/query` request is made to the server to retrieve these devices.
```

src/models/device.ts
```diff
@@ -14,18 +14,28 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { DeviceVerification } from "../crypto/deviceinfo";
+/** State of the verification of the device.
+ * Beware that the enum values are numbers instead of plain strings
```
richvdh, line 18: why is this important?

src/crypto/device-converter.ts
```diff
@@ -18,13 +18,13 @@ import { Device } from "../models/device";
 import { DeviceInfo } from "./deviceinfo";
 
 /**
- * Convert a DeviceInfo {@link DeviceInfo} to a Device {@link Device}
+ * Convert a {@link DeviceInfo} to a {@link Device}.
  * @param deviceInfo - deviceInfo to convert
- * @param userId - id of the device user
+ * @param userId - id of the hat owns the device.
```
richvdh, line 23: the hat?

richvdh (approved):
lgtm otherwise

PR #3273, opened 2023-04-12
Element-R: fix to device list change notifications

richvdh:
The rust crypto-sdk expects `UserId`s, not raw strings.

Follow-up to #3254

---


robintown (approved):
It's a shame that the Rust bindings don't have types, since I assume that would've helped here

richvdh:
I asked @Hywan about making these types a bit better recently. I think basically it needs some complicated support in https://github.com/rustwasm/wasm-bindgen.

PR #3274, opened 2023-04-12
Typescript fix to `AccountDataClient`

richvdh:
Obviously, `getAccountDataFromServer` can return null.

---

### DIFF
--- a/src/secret-storage.ts
+++ b/src/secret-storage.ts
@@ -132,9 +132,9 @@ export interface AccountDataClient extends TypedEventEmitter<ClientEvent.Account
      * ready, which can be useful very early in startup before the initial sync.
      *
      * @param eventType - The type of account data
-     * @returns The contents of the given account data event.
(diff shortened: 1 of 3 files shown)


src/secret-storage.ts
```diff
@@ -531,7 +531,7 @@ export class ServerSideSecretStorageImpl implements ServerSideSecretStorage {
             );
             const encInfo = secretInfo.encrypted[keyId];
             // only use keys we understand the encryption algorithm of
-            if (keyInfo.algorithm === SECRET_STORAGE_ALGORITHM_V1_AES) {
+            if (keyInfo !== null && keyInfo.algorithm === SECRET_STORAGE_ALGORITHM_V1_AES) {
```
t3chguy, line 534:
```suggestion
            if (keyInfo?.algorithm === SECRET_STORAGE_ALGORITHM_V1_AES) {
```

t3chguy (approved):
Shorthand ftw!

PR #3275, opened 2023-04-12
Fix typedoc warnings

richvdh:
There aren't that many of these, so I've gone through and fixed them, and
configured the GHA workflow to complain if any creep back in.

---


t3chguy (approved):
Awesome, thanks!

PR #3279, opened 2023-04-13
Add Room.threadsTimelineSets doc

weeman1337:
## Checklist

---

### DIFF
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -1971,7 +1971,7 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
     private async fetchRoomThreadList(filter?: ThreadFilterType): Promise<void> {
         const timelineSet = filter === ThreadFilterType.My ? this.threadsTimelineSets[1] : this.threadsTimelineSets[0];
 
-        if (!timelineSet) return;
+        if (timelineSet === undefined) return;
 
         const { chunk: events, end } = await this.client.createThreadListMessagesRequest(
             this.roomId,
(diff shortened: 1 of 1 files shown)


src/models/room.ts
```diff
@@ -1963,6 +1971,8 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
     private async fetchRoomThreadList(filter?: ThreadFilterType): Promise<void> {
         const timelineSet = filter === ThreadFilterType.My ? this.threadsTimelineSets[1] : this.threadsTimelineSets[0];
 
+        if (!timelineSet) return;
+
```
alunturner, line 1975: Following from the discussion recently about implicit casting to boolean, I have a question and also I'm wondering if this early return could be even earlier.

My thinking is along the lines of:

- this check is I think saying "if `timelineSet` is undefined, don't go any further" (I don't know if we should also stop if the array/set is empty, should we continue in that case?)
- if we're only concerned about undefined, the types would suggest that either both entries in `this.threadsTimelineSets` are defined or undefined at the same time
- in which case, could we just check if `this.timelineSet` has a length of 0 before assigning `timelineSet` to do the early return?

weeman1337: Thanks for the review. That makes sense :+1:

weeman1337: :thinking: if checking for the length it would need some non-null assertions :neutral_face: 

```
        if (this.createThreadsTimelineSets.length === 0) return;

        const timelineSet = filter === ThreadFilterType.My ? this.threadsTimelineSets[1]! : this.threadsTimelineSets[0]!;
```

I personally prefer to avoid them because they short-circuit the type system by knowing better based on knowledge.

alunturner: Perhaps I haven't got the correct settings turned on in TS config, but for me when I try that code it seems to allow TS to correctly infer that the type has been narrowed to the two item array (and so doesn't need the non-null assertion).

I definitely agree about trying to avoid the `!` though.

alunturner (approved):

PR #3280, opened 2023-04-13
Expose ServerSideSecretStorage independently of Crypto

richvdh:
There is no reason to indirect secret storage via the Crypto layer, and exposing it directly means it will work for Element-R.

Fixes: https://github.com/vector-im/element-web/issues/24982

---


richvdh:
The quality gate is being annoying on this.

t3chguy (approved):

PR #3281, opened 2023-04-13
Element-R: implement `{get,set}TrustCrossSignedDevices`

richvdh:
A precursor to https://github.com/vector-im/element-web/issues/25092

---


richvdh:
Thanks Robin!

PR #3282, opened 2023-04-13
Fix screen sharing on Firefox 113

tulir:
`getCapabilities` exists now(?), but `setCodecPreferences` doesn't, which means it would throw an error and fail the call.

---

## 🐛 Bug Fixes
 * Fix screen sharing on Firefox 113 ([\#3282](https://github.com/matrix-org/matrix-js-sdk/pull/3282)). Contributed by @tulir.


SimonBrandner (approved):
Looks sane

PR #3285, opened 2023-04-14
Fix TimelineWindow getEvents exploding if no neigbouring timeline

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/25104

---

## 🐛 Bug Fixes
 * Fix TimelineWindow getEvents exploding if no neigbouring timeline ([\#3285](https://github.com/matrix-org/matrix-js-sdk/pull/3285)). Fixes vector-im/element-web#25104.

### DIFF
--- a/src/timeline-window.ts
+++ b/src/timeline-window.ts
@@ -369,9 +369,9 @@ export class TimelineWindow {
 
         // iterate through each timeline between this.start and this.end
         // (inclusive).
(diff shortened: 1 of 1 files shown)


weeman1337 (approved):
:+1: looks good to me. Is there any way to cover it with a test case to prevent it from exploding again in the future?

PR #3286, opened 2023-04-14
Element-R: implement `MatrixClient.isSecretStorageReady`

richvdh:
Most of this is actually independent of the crypto implementation, so let's
hoist its definition up to `MatrixClient`.


richvdh:
I'm not sure I want to do this yet. I need to pull out bits of bootstrapSecretStorage, so I might pull all of that out to a separate module

PR #3287, opened 2023-04-14
New `CryptoApi.getDeviceVerificationStatus` api

richvdh:
Define a new endpoint to replace the existing `checkDeviceTrust` which is awkward to implement in rust.

Review commit-by-commit.

Based on https://github.com/matrix-org/matrix-js-sdk/pull/3281.

Part of https://github.com/vector-im/element-web/issues/25092.

---


richvdh:
The quality gate is complaining about "code smells" due to the use of deprecated APIs, but the whole point is to leave the deprecated things in place while we switch over, so I don't think we can do much about that.

robintown (approved):

PR #3307, opened 2023-04-20
Make webrtc stats collection optional

EnricoSchw:
Build a group call property to enable stats collecting

## Checklist

---

### DIFF
--- a/spec/unit/webrtc/groupCall.spec.ts
+++ b/spec/unit/webrtc/groupCall.spec.ts
@@ -1650,33 +1650,35 @@ describe("Group Call", function () {
             );
         });
         it("should be undefined if not get stats", async () => {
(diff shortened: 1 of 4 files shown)


dbkr (commented):
Will this mean you have to make sure to set the interval before initialising? Would it be better maybe to have a setGroupCallStatsInterval() that enables it or disables it as appropriate?

EnricoSchw:
> Will this mean you have to make sure to set the interval before initialising? Would it be better maybe to have a setGroupCallStatsInterval() that enables it or disables it as appropriate?

I understand: `setGroupCallStatsInterval()` would be more flexible and would allow set interval before initializing a call and as well after initializing.

With that I wouldn't need the public propery either. Ok I'll change the code a bit then. Let's do it like this.

dbkr (approved):

PR #3314, opened 2023-04-25
Adjust negotiation process for pending answer process

EnricoSchw:
Note that an answer operation pending process is not treated as an unstable negotiation state and an offer will not rejected from impolite peer. The PR fixes the issue  adding an answer is not in race with renegotiation.'

We not using `await this.peerConn!.setLocalDescription(offer);` parameter less. We would also like to adjust this. Unfortunately, we munging the local SDP before adding it. That needs to be adjust first.

## Checklist

---

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -1945,6 +1945,7 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
     }
 
     public async onNegotiateReceived(event: MatrixEvent): Promise<void> {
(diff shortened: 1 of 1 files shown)


SimonBrandner (approved):

robintown (approved):
It would be even better if this had a test, but this looks reasonable

EnricoSchw:
> It would be even better if this had a test, but this looks reasonable

Good point. I will add the missing tests, when we fix the `setLocalDescription`.

PR #3317, opened 2023-04-25
Pre-allocate transceivers

SimonBrandner:
Fixes https://github.com/vector-im/element-call/issues/1011

---

## ✨ Features
 * Pre-allocate transceivers ([\#3317](https://github.com/matrix-org/matrix-js-sdk/pull/3317)). Fixes vector-im/element-call#1011.

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -1438,8 +1438,18 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
 
         // we may not have a video track - if not, re-request usermedia
         if (!muted && this.localUsermediaStream!.getVideoTracks().length === 0) {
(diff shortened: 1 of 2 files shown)


src/webrtc/call.ts
```diff
@@ -1606,10 +1627,14 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
...
+        this.preallocateTransceivers();
+
         if (requestScreenshareFeed) {
-            this.peerConn!.addTransceiver("video", {
-                direction: "recvonly",
-            });
+            for (const kind of ["audio", "video"]) {
+                this.peerConn!.addTransceiver(kind, {
+                    direction: "recvonly",
+                });
```
EnricoSchw, line 1636: Hmm interesting. We creating every time a new transceiver, when we receive a screenshare. I was thinking thats this will done automatically.. I did not know that

PR #3320, opened 2023-04-26
Move receipt-specific logic into ReceiptAccumulator

andybalaam:

Move all the logic for processing receipts into the ReceiptAccumulator, so we can create simpler tests, and so it's easier to deal with when we start fixing bugs in it.

It's probably worth looking at this commit-by-commit.

---

### DIFF
--- a/src/sync-accumulator.ts
+++ b/src/sync-accumulator.ts
@@ -19,7 +19,7 @@ limitations under the License.
  */
 
 import { logger } from "./logger";
-import { deepCopy, isSupportedReceiptType, MapWithDefault, recursiveMapToObject } from "./utils";
+import { deepCopy, isSupportedReceiptType } from "./utils";
 import { IContent, IUnsigned } from "./models/event";
 import { IRoomSummary } from "./models/room-summary";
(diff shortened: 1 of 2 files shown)


src/sync-accumulator.ts
```diff
@@ -167,22 +168,7 @@ interface IRoom {
...
-    _threadReadReceipts: {
-        [threadId: string]: {
-            [userId: string]: {
-                data: IMinimalEvent;
-                type: ReceiptType;
-                eventId: string;
-            };
-        };
-    };
+    _receipts: ReceiptAccumulator;
```
weeman1337, line 171: Is this really something that should be exposed via the interface?
It looks more like implementation specific to me.

weeman1337 (approved):
:+1: Looks good.

src/sync-accumulator.ts
andybalaam: I didn't want to mess around with the design of the sync accumulator more than I needed to. I haven't really thought about what this interface actually does - let's discuss it sometime.

PR #3322, opened 2023-04-26
Fix various APIs which incorrectly use an `IAuthData`

richvdh:
`IAuthData` is the response, not the request. The request is an `IAuthDict`

(I feel like this is very much the sort of confusion that could have been resolved if the type had been documented in the first place.)

---


richvdh:
Possibly this is a breaking change? But if it does break things, it's probably just exposing existing brokenness?

t3chguy (approved):

PR #3326, opened 2023-04-27
Tests for the ReceiptAccumulator

andybalaam:

Set up some basic unit tests before we fix the actual bug and write tests that validate that.

---


weeman1337 (commented):
The accumulator has a special handling of `main` thread events. In my opinion tests should cover this.

andybalaam:
> The accumulator has a special handling of `main` thread events. In my opinion tests should cover this.

That's a bug I am about to fix. I'd prefer not to "lock down" the wrong behaviour then change the test later, but if you prefer that style I can do it.

weeman1337 (approved):

PR #3339, opened 2023-05-02
Accumulate receipts for the main thread and unthreaded separately

andybalaam:
Fixes https://github.com/vector-im/element-web/issues/24629

---

## 🐛 Bug Fixes
 * Accumulate receipts for the main thread and unthreaded separately ([\#3339](https://github.com/matrix-org/matrix-js-sdk/pull/3339)). Fixes vector-im/element-web#24629. Contributed by @andybalaam.

### DIFF
--- a/src/receipt-accumulator.ts
+++ b/src/receipt-accumulator.ts
@@ -18,7 +18,7 @@ import { IMinimalEvent } from "./sync-accumulator";
 import { EventType } from "./@types/event";
 import { isSupportedReceiptType, MapWithDefault, recursiveMapToObject } from "./utils";
 import { IContent } from "./models/event";
-import { MAIN_ROOM_TIMELINE, ReceiptContent, ReceiptType } from "./@types/read_receipts";
+import { ReceiptContent, ReceiptType } from "./@types/read_receipts";
 
 interface AccumulatedReceipt {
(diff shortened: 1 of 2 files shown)


gsouquet (approved):
🚀 💯 Looks great, thank you for working on this

src/receipt-accumulator.ts
```diff
@@ -118,7 +118,7 @@ export class ReceiptAccumulator {
                             eventId,
                         };
 
-                        if (!data.thread_id || data.thread_id === MAIN_ROOM_TIMELINE) {
```
MadLittleMods, line 121: A comment with some context would be helpful why "we shouldn't also be checking for `MAIN_ROOM_TIMELINE` because ... otherwise we run into bug x".

andybalaam: That's not quite it - when the user marks a room as read e.g. by pressing Esc or clicking the "x" in the top right, we do send an unthreaded receipt. I'd happily merge a PR from you with a comment like this somewhere appropriate.

MadLittleMods: Created https://github.com/matrix-org/matrix-js-sdk/pull/3378 and incorporated those caveats :+1:

PR #3341, opened 2023-05-03
Do an ICE Restart if WebRTC become disconnected

EnricoSchw:
In the case  a WebRTC connection become  'iceConnectionState' 'disconnected', we start an new ICE gathering just after two seconds. This PR is related to: https://github.com/vector-im/element-call/issues/928

The ICE Restart process can take up to 15 seconds. I will check if we can speed up this process. That's why I'm creating the PR as draft

## Checklist

---


spec/unit/webrtc/call.spec.ts
```diff
@@ -1652,19 +1652,39 @@ describe("Call", function () {
...
+            jest.advanceTimersByTime(3 * 1000);
+            expect(mockPeerConn.restartIce).toHaveBeenCalled();
         });
 
         it("should hang up after being disconnected for 30 seconds", () => {
             jest.advanceTimersByTime(31 * 1000);
             expect(call.hangup).toHaveBeenCalledWith(CallErrorCode.IceFailed, false);
         });
 
+        it("should restart ICE gathering one again after being failed for 4 seconds", () => {
```
dbkr, line 1674:
```suggestion
        it("should restart ICE gathering once again after being failed for 4 seconds", () => {
```

dbkr: Also doesn't look like the 4 second delay is part of this test?

spec/unit/webrtc/call.spec.ts
```diff
@@ -1652,19 +1652,39 @@ describe("Call", function () {
...
+            mockPeerConn.iceConnectionState = "failed";
+            mockPeerConn.iceConnectionStateChangeListener!();
+            expect(mockPeerConn.restartIce).toHaveBeenCalled();
+        });
+
+        it("should call hangup if Android after ICE being failed for 4 seconds", () => {
```
dbkr, line 1680: "if Android" == "if not restartIce" method presumably?

Also the 4 second claim doesn't seem reflected in the test.

src/webrtc/call.ts
```diff
@@ -2247,8 +2258,12 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
             // Firefox for Android does not yet have support for restartIce()
             // (the types say it's always defined though, so we have to cast
             // to prevent typescript from warning).
+            this.candidatesEnded = false;
```
dbkr, line 2261: This should probably go above the comment?

src/webrtc/call.ts
```diff
@@ -2257,7 +2272,21 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
                 this.hangup(CallErrorCode.IceFailed, false);
             }
         } else if (this.peerConn?.iceConnectionState == "disconnected") {
-            this.iceDisconnectedTimeout = setTimeout(() => {
+            this.candidatesEnded = false;
+            // noinspection TypeScriptValidateTypes
```
dbkr, line 2276: What's the invalid type here?

EnricoSchw: Always the same my IDE recognizes the node js method here. I turned it off because it was handicapping me. I forgot to remove the comments.

spec/unit/webrtc/call.spec.ts
EnricoSchw: Right, in case of "failed" we start gathering directly. I'll adjust the test description.

spec/unit/webrtc/call.spec.ts
EnricoSchw: Yes, this is more or less only in Android the case. I want to point this out. But I wouldn't commit here and would change it.

PR #3350, opened 2023-05-10
Revert "Update JS-DevTools/npm-publish action to v2 (#3336)"

andybalaam:
Attempting to fix failure to publish to NPM due to an "invalid token" error https://github.com/matrix-org/matrix-js-sdk/actions/runs/4926808655/jobs/8819822367 .

This reverts commit ca9263fa647e4cfc581a137c389ea646bcf71ce6.

---

### DIFF
--- a/.github/workflows/release-npm.yml
+++ b/.github/workflows/release-npm.yml
@@ -24,7 +24,7 @@ jobs:
 
             - name: 🚀 Publish to npm
               id: npm-publish
-              uses: JS-DevTools/npm-publish@0be441d808570461daedc3fb178405dbcac54de0 # v2
+              uses: JS-DevTools/npm-publish@0f451a94170d1699fd50710966d48fb26194d939 # v1
               with:
                   token: ${{ secrets.NPM_TOKEN }}
                   access: public


t3chguy:
Indeed, due to the major update of npm-publish, the `.npmrc` file no longer contains the token after the run of the action

> The user's ~/.npmrc file is no longer modified. Instead, a temporary .npmrc file is used. (https://github.com/JS-DevTools/npm-publish/issues/15)

So the subsequent npm commands fail.

Regressed by https://github.com/matrix-org/matrix-js-sdk/pull/3336

t3chguy (approved):

andybalaam:
> So the subsequent npm commands fail.

Which subsequent npm commands? I only see one in this workflow.

t3chguy:
@andybalaam 

https://github.com/matrix-org/matrix-js-sdk/blob/develop/.github/workflows/release-npm.yml#L25-L31 runs `npm publish`, with v1 it first sets up the `.npmrc` file - with v2 it sets up a temporary one that gets thrown away

https://github.com/matrix-org/matrix-js-sdk/blob/develop/.github/workflows/release-npm.yml#L37 is the subsequent (not publish) npm command.

andybalaam:
@t3chguy ah, thank you! I was not making the leap from release.yml to release-npm.yml

t3chguy:
https://github.com/matrix-org/matrix-js-sdk/blob/develop/.github/workflows/release.yml#L38-L42 https://docs.github.com/en/actions/using-workflows/reusing-workflows

andybalaam:
Feels like we should be able to specify multiple tags to npm-publish. I'll look at that if I have time.

t3chguy:
That would be a nice modification, but the underlying `npm publish` command doesn't support it

PR #3352, opened 2023-05-11
Rename and move `crypto.IBootstrapCrossSigningOpts`

richvdh:
Move out of the legacy `crypto` module, give it a rename, pull out a seperate `UIAuthCallback` interface

review commit-by-commit

---

### DIFF
--- a/src/crypto-api.ts
+++ b/src/crypto-api.ts
@@ -17,6 +17,7 @@ limitations under the License.
 import type { IMegolmSessionData } from "./@types/crypto";
 import { Room } from "./models/room";
 import { DeviceMap } from "./models/device";
+import { UIAuthCallback } from "./interactive-auth";
 
 /**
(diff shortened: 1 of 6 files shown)


t3chguy (approved):

src/crypto-api.ts
```diff
@@ -123,6 +124,20 @@ export interface CryptoApi {
     getDeviceVerificationStatus(userId: string, deviceId: string): Promise<DeviceVerificationStatus | null>;
 }
 
+/**
+ * Options object for {@link CryptoApi.bootstrapCrossSigning}.
```
richvdh, line 128:
```suggestion
 * Options object for `CryptoApi.bootstrapCrossSigning`.
```

richvdh: ... because that isn't written yet, and the linter is sad

PR #3354, opened 2023-05-11
Element-R: Stub out `isCrossSigningReady` and `isSecretStorageReady`

richvdh:
This is enough to get most of the cypress tests running, which is a win

---


richvdh:
will add some tests to make this pass the quality gate

t3chguy (approved):
LGTM sans sonar

PR #3360, opened 2023-05-12
Add `CryptoApi.getCrossSigningKeyId`

richvdh:
Another crypto function to stub out to get the Cypress tests running


richvdh:
I'd like to request an exemption to the quality gate, because it's legacy code we want to throw away

src/crypto-api.ts
```diff
@@ -137,6 +144,16 @@ export interface CryptoApi {
...
 
+    /**
+     * Get the ID of one of the user's cross-signing keys.
+     *
+     * @param type - The type of key to get the ID of.  One of
+     *     "master", "self_signing", or "user_signing".  Defaults to "master".
+     *
+     * @returns If cross-signing has been initialised on this device, the ID of the given key. Otherwise, null
+     */
+    getCrossSigningKeyId(type?: CrossSigningKey | string): Promise<string | null>;
```
t3chguy, line 155: Can we use a const enum instead of this union?

richvdh: actually we don't need this union any more.
```suggestion
    getCrossSigningKeyId(type?: CrossSigningKey): Promise<string | null>;
```

src/crypto/index.ts
```diff
@@ -1415,6 +1420,11 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
      *
      * @returns the key ID
      */
+    public getCrossSigningKeyId(type: CrossSigningKey | string = CrossSigningKey.Master): Promise<string | null> {
```
richvdh, line 1423:
```suggestion
    public getCrossSigningKeyId(type: CrossSigningKey = CrossSigningKey.Master): Promise<string | null> {
```

src/rust-crypto/rust-crypto.ts
```diff
@@ -324,6 +325,14 @@ export class RustCrypto implements CryptoBackend {
         return false;
     }
 
+    /**
+     * Implementation of {@link CryptoApi#getCrossSigningKeyId}
+     */
+    public async getCrossSigningKeyId(type: CrossSigningKey | string = CrossSigningKey.Master): Promise<string | null> {
```
richvdh, line 331:
```suggestion
    public async getCrossSigningKeyId(type: CrossSigningKey = CrossSigningKey.Master): Promise<string | null> {
```

src/crypto-api.ts
```diff
@@ -137,6 +144,16 @@ export interface CryptoApi {
      */
     isCrossSigningReady(): Promise<boolean>;
 
+    /**
+     * Get the ID of one of the user's cross-signing keys.
+     *
+     * @param type - The type of key to get the ID of.  One of
+     *     "master", "self_signing", or "user_signing".  Defaults to "master".
```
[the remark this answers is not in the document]
richvdh: ok, updated

t3chguy (approved):

PR #3361, opened 2023-05-12
Keep measuring a call feed's volume after a stream replacement

robintown:
Closes https://github.com/vector-im/element-call/issues/1051

---

## 🐛 Bug Fixes
 * Keep measuring a call feed's volume after a stream replacement ([\#3361](https://github.com/matrix-org/matrix-js-sdk/pull/3361)). Fixes vector-im/element-call#1051.


SimonBrandner (approved):
Thanks!

PR #3364, opened 2023-05-15
Element-R: support for manual import/export of Room keys

BillCarsonFr:
Support for importing/exporting keys manually in rust crypto

Fixes: https://github.com/vector-im/element-web/issues/25315

## Checklist

---


src/crypto-api.ts
```diff
@@ -75,6 +76,14 @@ export interface CryptoApi {
      */
     exportRoomKeys(): Promise<IMegolmSessionData[]>;
 
+    /**
+     * Import a list of room keys previously exported by exportRoomKeys
+     *
+     * @param keys - a list of session export objects
+     * @returns a promise which resolves once the keys have been imported
+     */
+    importRoomKeys(keys: IMegolmSessionData[], opts?: IImportRoomKeysOpts): Promise<void>;
```
t3chguy, line 85: docs missing for `opts`

src/rust-crypto/rust-crypto.ts
```diff
@@ -202,10 +202,23 @@ export class RustCrypto implements CryptoBackend {
     }
 
     public async exportRoomKeys(): Promise<IMegolmSessionData[]> {
-        // TODO
-        return [];
+        const raw = await this.olmMachine.exportRoomKeys(() => true);
+        return Promise.resolve(JSON.parse(raw));
+    }
+
+    public async importRoomKeys(keys: IMegolmSessionData[], opts?: IImportRoomKeysOpts | undefined): Promise<void> {
```
t3chguy, line 209: The `undefined` is superfluous given the `?:`

src/rust-crypto/rust-crypto.ts
```diff
@@ -202,10 +202,23 @@ export class RustCrypto implements CryptoBackend {
...
+                total: Number(total),
+                successes: Number(progress),
+                stage: "",
+                failures: 0,
+            };
+            opts?.progressCallback?.(importOpt);
+        });
+        return Promise.resolve();
     }
-
```
t3chguy, line 208: we should maintain spacings between methods

src/client.ts
```diff
@@ -3198,10 +3198,10 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @returns a promise which resolves when the keys have been imported
      */
     public importRoomKeys(keys: IMegolmSessionData[], opts?: IImportRoomKeysOpts): Promise<void> {
```
richvdh, line 3200: we should mark this method as deprecated while we're here (as we already have with `exportRoomKeys`)

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -47,9 +48,52 @@ describe("RustCrypto", () => {
             rustCrypto = await makeTestRustCrypto();
         });
 
-        it("should return a list", async () => {
+        it("should import and export keys", async () => {
```
richvdh, line 51: .exportRoomKeys certainly should *not* import room keys :-p

can you rename the tests?

src/crypto-api.ts
```diff
@@ -18,6 +18,7 @@ import type { IMegolmSessionData } from "./@types/crypto";
 import { Room } from "./models/room";
 import { DeviceMap } from "./models/device";
 import { UIAuthCallback } from "./interactive-auth";
+import { IImportRoomKeysOpts } from "./crypto/api";
```
richvdh, line 21: I'm trying to move things out of `crypto` at the same time as adding them to CryptoApi. Could you move `IImportRoomKeysOpts` into this file and rename it to `ImportRoomKeysOpts`?

BillCarsonFr: So:
- `crypto/` is js crypto
- `rust-crypto/` is rust crypto
- and `common-crypto` is the API

Right? So I should move IImportRoomKeysOpts to common ?

BillCarsonFr: oh i was confused with crypto/api.ts and crypto-api.ts

richvdh:
@BillCarsonFr can you please avoid force-pushing after a review: it makes it hard to see what has changed since the review.

src/rust-crypto/rust-crypto.ts
```diff
@@ -206,8 +206,22 @@ export class RustCrypto implements CryptoBackend {
...
+        await this.olmMachine.importRoomKeys(jsonKeys, (progress: BigInt, total: BigInt) => {
+            const importOpt: ImportOpts = {
+                total: Number(total),
+                successes: Number(progress),
+                stage: "load_keys",
+                failures: 0,
+            };
+            opts?.progressCallback?.(importOpt);
+        });
+        return;
```
richvdh, line 224: this `return` is redundant

src/crypto-api.ts
```diff
@@ -260,3 +269,17 @@ export class DeviceVerificationStatus {
...
+
+export interface ImportOpts {
+    stage: string; // TODO: Enum
+    successes: number;
+    failures: number;
+    total: number;
+}
+
+export interface ImportRoomKeysOpts {
+    /** called with an object that has a "stage" param */
```
richvdh, line 281: well it's called with an `ImportOpts`; that much is obvious. A more useful doc-comment here would be to say *when* it is called.

spec/unit/rust-crypto/rust-crypto.spec.ts
richvdh: this test still has a misleading name

src/client.ts
```diff
@@ -3811,7 +3811,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             algorithm.free();
         }
 
-        await this.importRoomKeys(keys, {
+        await this.getCrypto()?.importRoomKeys(keys, {
```
richvdh, line 3814: note that this method still won't work with rust crypto due to the test of `this.crypto` at the top. Arguably fixing this method is out of scope for this PR though.

src/crypto/api.ts
```diff
@@ -100,17 +100,3 @@ export interface ICreateSecretStorageOpts {
...
-    failures: number;
-    total: number;
-}
-
-export interface IImportRoomKeysOpts {
-    /** called with an object that has a "stage" param */
-    progressCallback?: (stage: IImportOpts) => void;
-    untrusted?: boolean;
-    source?: string; // TODO: Enum
-}
```
richvdh, line 116: applications might be depending on the existence of these interfaces, so we need to re-export `ImportRoomKeysOpts` and `ImportOpts` for backwards compatibility (see the existing "re-exports for backwards compatibility" at the top of the file).

src/crypto-api.ts
```diff
@@ -82,6 +82,15 @@ export interface CryptoApi {
      */
     exportRoomKeys(): Promise<IMegolmSessionData[]>;
 
+    /**
+     * Import a list of room keys previously exported by exportRoomKeys
+     *
+     * @param keys - a list of session export objects
+     * @param opts - progress reporting, safety and key source in the form of a ImportRoomKeysOpts
```
richvdh, line 89: I don't think we need to be quite this specific here: the place to document what's *in* an `ImportRoomKeysOpts` is in `ImportRoomKeysOpts`. 

```suggestion
     * @param opts - options object
```

likewise in MatrixClient

src/rust-crypto/rust-crypto.ts
```diff
@@ -206,8 +206,22 @@ export class RustCrypto implements CryptoBackend {
...
     }
```
richvdh, line 225: I'm somewhat concerned that we aren't doing anything with `opts.untrusted` and `opts.source` here. Do we know what they are meant to do? Is it a problem that we aren't implementing them? Should we deprecate them? Or at least leave a TODO to do something with them later?

src/rust-crypto/rust-crypto.ts
```diff
@@ -206,8 +206,22 @@ export class RustCrypto implements CryptoBackend {
...
+
+    public async importRoomKeys(keys: IMegolmSessionData[], opts?: ImportRoomKeysOpts): Promise<void> {
+        const jsonKeys = JSON.stringify(keys);
+        await this.olmMachine.importRoomKeys(jsonKeys, (progress: BigInt, total: BigInt) => {
+            const importOpt: ImportOpts = {
+                total: Number(total),
+                successes: Number(progress),
+                stage: "load_keys",
+                failures: 0,
+            };
```
richvdh, line 221: I'm assuming there was more than one `stage` in the legacy implementation - do we not need to replicate them? And do something with `failures`?

richvdh (changes requested):

src/rust-crypto/rust-crypto.ts
BillCarsonFr: Will add a todo.
Not sure what opts.source was supposed to be, but I suspect it is do determine if the key was received from the Backup, from a File, or from a gossip? It's most likely interesting for a dev tool. From the crypto point of view the key is trusted or not (a trusted key would be one received from a m.room.key).

For `opts.untrusted` as for now any key imported by this method will be considered as untrusted (messages decrypted with it will have the gray shield "The authenticity of this message is not guarenteed").

But I think we might be missing the mapping of the `from_backup` flag (that is used by the sdk to decide if that key needs to be uploaded to backup).

src/client.ts
BillCarsonFr: done

src/rust-crypto/rust-crypto.ts
BillCarsonFr: done

spec/unit/rust-crypto/rust-crypto.spec.ts
BillCarsonFr: Updated 
> describe(".importThenExportRoomKeys", ..)

src/rust-crypto/rust-crypto.ts
BillCarsonFr: done

src/crypto-api.ts
BillCarsonFr: Updarted

src/crypto/api.ts
BillCarsonFr: re-exported in api.ts
 ImportRoomKeyProgressData as IImportOpts,
 ImportRoomKeysOpts as IImportRoomKeysOpts,

src/crypto-api.ts
BillCarsonFr: done

src/crypto-api.ts
```diff
@@ -260,3 +269,17 @@ export class DeviceVerificationStatus {
...
 }
+
+export interface ImportOpts {
+    stage: string; // TODO: Enum
+    successes: number;
+    failures: number;
+    total: number;
+}
+
+export interface ImportRoomKeysOpts {
```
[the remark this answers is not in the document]
BillCarsonFr: done

src/rust-crypto/rust-crypto.ts
BillCarsonFr: In legacy it's the same, the stage is `load_keys` in the crypto module.
But the matrix client has a `restoreKeyBackup` method with a free form `opt` that has a similar `progressCallback` property and in this case when downloading from server the stage is `fetch`.

This would need to be refactored when implementing server backup I think.

src/crypto-api.ts
```diff
@@ -260,3 +269,17 @@ export class DeviceVerificationStatus {
         return this.localVerified || (this.trustCrossSignedDevices && this.crossSigningVerified);
     }
 }
+
+export interface ImportOpts {
```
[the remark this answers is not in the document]
BillCarsonFr: Already renamed to ImportRoomKeyProgressData, is it ok?

src/crypto-api.ts
BillCarsonFr: Updated

src/rust-crypto/rust-crypto.ts
```diff
@@ -206,8 +211,22 @@ export class RustCrypto implements CryptoBackend {
...
 
     public async exportRoomKeys(): Promise<IMegolmSessionData[]> {
-        // TODO
-        return [];
+        const raw = await this.olmMachine.exportRoomKeys(() => true);
+        return JSON.parse(raw);
+    }
+
+    public async importRoomKeys(keys: IMegolmSessionData[], opts?: ImportRoomKeysOpts): Promise<void> {
+        // TODO when backup support will be added we would need to expose from backup in bindings
```
richvdh, line 219: > expose from backup in bindings

I don't understand this. Expose *what* from backup?

richvdh (approved):
mostly looks good to me except a couple of nits

src/rust-crypto/rust-crypto.ts
BillCarsonFr: There is a [`from_backup`](https://github.com/matrix-org/matrix-rust-sdk/blob/a977b3a8d5f3dbd9dcf01196aaaf2d16f9ea64c2/bindings/matrix-sdk-crypto-ffi/src/machine.rs#L1381) flag when importing keys in olmMachine. Used as a reminder to not upload this key to backup later as part of the backup danse.

richvdh:
```suggestion
        // TODO when backup support will be added we would need to expose the `from_backup` flag in the bindings
```

PR #3368, opened 2023-05-16
Element-R: initial implementation of `bootstrapCrossSigning`

richvdh:
Fixes https://github.com/vector-im/element-web/issues/25314

---


richvdh:
hrm this seems to be less ready than I thought

dbkr (approved):

Issue #3325, opened 2023-04-27
Several places in the code use timestamp to determine event order


andybalaam:
When are processing events, we sometimes need to know what order they are in. (For example, to decide whether an event is before the last read receipt so we know whether a room is unread.)

The correct order to use is the order in which they are provided during the `sync` request from the homeserver. It is the homeserver's job to determine a linear order from the DAG, and we need to agree with the homeserver to make sure we are giving the correct results for e.g. the notifications for a room.

In some places in the code we use the timestamp (`ts` property) of events to determine their order, because we don't have a record of the order of events from `sync`.

We should keep track of the server-provided event order and replace all uses of `ts` with the correct order. In the meantime, if we add code that depends on `ts`, we should reference this issue in a comment in the code.

Impact of this issue: sometimes there may be stuck unreads or notifications due to inconsistencies between how the homeserver and the client understand the order of events. This may also cause other problems - we're not certain. We think the problems we see will be relatively rare, which is why this has not yet been fixed.

t3chguy:
> The correct order to use is the order in which they are provided during the sync request from the homeserver.

What about when you have events from `/context/` which follows a different ordering? Similar question for `/relations/`.

andybalaam:
> > The correct order to use is the order in which they are provided during the sync request from the homeserver.
> 
> What about when you have events from `/context/` which follows a different ordering? Similar question for `/relations/`.

I'm not sure, but in principle I think the answer is that we should keep hold of our current best guess of the order the server has, even if sometimes it's not 100% correct.

In future we might need to figure out an MSC that makes this easier to deal with.

t3chguy:
Related https://github.com/matrix-org/matrix-spec/issues/852

> I'm not sure, but in principle I think the answer is that we should keep hold of our current best guess of the order the server has, even if sometimes it's not 100% correct.

The server has 2 main orders though, topological ordering and stream ordering, which do you propose we try and guess?

andybalaam:
> Related [matrix-org/matrix-spec#852](https://github.com/matrix-org/matrix-spec/issues/852)
> 
> > I'm not sure, but in principle I think the answer is that we should keep hold of our current best guess of the order the server has, even if sometimes it's not 100% correct.
> 
> The server has 2 main orderings though, topological ordering and stream ordering, which do you propose we try and guess? Some APIs return one, some the other. They do not fit within the same total order.

[...]

PR #3378, opened 2023-05-17
Add more context for why threaded vs unthreaded read receipts

MadLittleMods:
Add more context for why threaded vs unthreaded read receipts

See https://github.com/matrix-org/matrix-js-sdk/pull/3339#discussion_r1195364120

## Checklist

-   [ ] ~~Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))~~

---

### DIFF
--- a/src/receipt-accumulator.ts
+++ b/src/receipt-accumulator.ts
@@ -118,19 +118,24 @@ export class ReceiptAccumulator {
                             eventId,
                         };
 
-                        // In a world that supports threads, all read receipts have a
-                        // `thread_id` which is either the thread they belong in or
-                        // `MAIN_ROOM_TIMELINE` and we should always use
-                        // `setThreaded(...)` here. The `MAIN_ROOM_TIMELINE` is just
-                        // treated as another thread.
-                        //
-                        // We still may encounter read receipts from clients that don't
-                        // support threads (unthreaded, without the `thread_id`
-                        // property) which we use `setUnthreaded(...)` for. Or even in a
-                        // threaded-world, a client may choose to use an unthreaded read
-                        // receipt to completely mark the room as read.
-                        //
-                        // Using the wrong method will cause undefined behavior like
+                        // In a world that supports threads, read receipts normally have
+                        // a `thread_id` which is either the thread they belong in or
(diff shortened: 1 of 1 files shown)


src/receipt-accumulator.ts
```diff
@@ -118,6 +118,21 @@ export class ReceiptAccumulator {
...
+                        //
+                        // We still may encounter read receipts from clients that don't
+                        // support threads (unthreaded, without the `thread_id`
+                        // property) which we use `setUnthreaded(...)` for. Or even in a
+                        // threaded-world, a client may choose to use an unthreaded read
+                        // receipt to completely mark the room as read.
+                        //
+                        // Using the wrong method will cause undefined behavior like
+                        // messages re-appearing as "new" when you already read them
+                        // previously.
```
andybalaam, line 135: In a world that supports threads, read receipts normally have a `thread_id` which is either the thread they belong in or `MAIN_ROOM_TIMELINE`, so we normally use `setThreaded(...)` here. The `MAIN_ROOM_TIMELINE` is just treated as another thread.

We still encounter read receipts that are "unthreaded" (missing the `thread_id` property). These come from clients that don't support threads, and from threaded clients that are doing a "Mark room as read" operation. Unthreaded receipts mark everything "before" them as read, in all threads, where "before" means in Sync Order i.e. the order the events were received from the homeserver in a sync. [Note: we have some bugs where we use timestamp order instead of Sync Order, because we don't correctly remember the Sync Order. See #3325.]

Calling the wrong method will cause incorrect behavior like messages re-appearing as "new" when you already read them previously.

andybalaam (changes requested):
Suggested different wording to make unthreaded receipts sound less weird (we do send them ourselves) and to explain a little about Sync Order.

src/receipt-accumulator.ts
MadLittleMods: Sounds great! Updated :white_check_mark:

andybalaam:
Great, thank you for making the effort @MadLittleMods

MadLittleMods:
Thanks for the review and merge @andybalaam 🪺

PR #3381, opened 2023-05-18
Update Mutual Rooms (MSC2666) support

ShadowJonathan:
This updates the Mutual Rooms ([MSC2666](https://github.com/matrix-org/matrix-spec-proposals/pull/2666)) implementation to match the [latest version](https://github.com/matrix-org/matrix-spec-proposals/pull/2666#issuecomment-1412236010).

This also cleans and clears up the support for older versions of the MSC. The new version paginates, and so that support has been added too.

## Checklist

Type: Task

`Signed-off-by: Jonathan de Jong <jo@jo.wtf>`

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -497,6 +497,10 @@ export interface IMSC3882GetLoginTokenCapability extends ICapability {}
 
 export const UNSTABLE_MSC3882_CAPABILITY = new UnstableValue("m.get_login_token", "org.matrix.msc3882.get_login_token");
 
+export const UNSTABLE_MSC2666_SHARED_ROOMS = "uk.half-shot.msc2666";
+export const UNSTABLE_MSC2666_MUTUAL_ROOMS = "uk.half-shot.msc2666.mutual_rooms";
+export const UNSTABLE_MSC2666_QUERY_MUTUAL_ROOMS = "uk.half-shot.msc2666.query_mutual_rooms";
+
 /**
  * A representation of the capabilities advertised by a homeserver as defined by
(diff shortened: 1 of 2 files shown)


andybalaam:
Thank you for your contribution! I'm looking at it now, but we have a (new since this was first implemented) requirement for 80% test coverage on any changes, so this will definitely need some tests before we can merge it.

src/client.ts
```diff
@@ -7146,29 +7146,75 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            );
+            extraQuery = {};
+        }
+
+        // Accumulated rooms
+        const rooms: string[] = [];
+        let token = null;
+
+        do {
+            const additionalQuery: Record<string, string> = {};
```
andybalaam, line 7196: rename to `tokenQuery`? I'm finding `extraQuery` + `additionalQuery` hard to understand. Alternatively, you could modify `extraQuery` here instead of creating a new one.

ShadowJonathan:
I'm going to try to figure out how to make tests for this, with example responses for each MSC version, and a pagination test.

andybalaam (changes requested):
Looks good, but needs tests. I suspect you could monkey-patch `this.http.authedRequest` with a mock, or maybe you can think of a better way.

andybalaam:
Focussing on stuck unreads, so not finding time for this, but I haven't forgotten :green_heart:

andybalaam (approved):

andybalaam:
Thank you!

PR #3382, opened 2023-05-18
Fix verification bug with `pendingEventOrdering: "chronological"`

richvdh:
There exists a bug (in the legacy `Crypto` implementation), in which in-room verifications-via-emoji will fail for applications which use the default `pendingEventOrdering` setting.

The basis of the problem is that the verifier sees a local echo of our own outgoing `m.key.verification.mac`. It tries to verify the MAC in that message, which of course fails.

The solution is to treat such local echoes in the same way as remote echoes.

---

## 🐛 Bug Fixes
 * Fix verification bug with `pendingEventOrdering: "chronological"` ([\#3382](https://github.com/matrix-org/matrix-js-sdk/pull/3382)).

### DIFF
--- a/src/crypto/verification/request/InRoomChannel.ts
+++ b/src/crypto/verification/request/InRoomChannel.ts
@@ -208,10 +208,17 @@ export class InRoomChannel implements IVerificationChannel {
             this.requestEventId = InRoomChannel.getTransactionId(event);
         }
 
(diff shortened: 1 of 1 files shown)


richvdh:
In an ideal world, I'd add an explicit test for this, but it's a bit of an obscure bug in the legacy Crypto implementation, and setting up a test to exercise it will be a non-trivial amount of work.

I do have a minor change to react-sdk's cypress tests which rely on it, so that will have to do.

PR #3384, opened 2023-05-18
Attempt a potential workaround for stuck notifs

andybalaam:
Fixes https://github.com/vector-im/element-web/issues/25406

This adds a new way of inserting a related event into a timeline, by adding it after its "parent", and before any event with a later timestamp. Previously a new event like a reaction would appear at the end. This makes it more likely we will guess an ordering close to the correct Sync Order.

We won't need this after MSC3981.

---

## 🐛 Bug Fixes
 * Attempt a potential workaround for stuck notifs ([\#3384](https://github.com/matrix-org/matrix-js-sdk/pull/3384)). Fixes vector-im/element-web#25406. Contributed by @andybalaam.

### DIFF
--- a/src/models/event-timeline-set.ts
+++ b/src/models/event-timeline-set.ts
@@ -757,6 +757,99 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
     }
 
     /**
+     * Insert event to the given timeline, and emit Room.timeline. Assumes
+     * we have already checked we don't know about this event.
+     *
+     * TEMPORARY: until we have recursive relations, we need this function
+     * to exist to allow us to insert events in timeline order, which is our
+     * best guess for Sync Order.
+     * This is a copy of addEventToTimeline above, modified to insert the event
+     * after the event it relates to, and before any event with a later
+     * timestamp. This is our best guess at Sync Order.
+     *
+     * Will fire "Room.timeline" for each event added.
+     *
+     * @param options - addEventToTimeline options
+     *
+     * @remarks
+     * Fires {@link RoomEvent.Timeline}
+     */
+    public insertEventIntoTimeline(event: MatrixEvent, timeline: EventTimeline, roomState: RoomState): void {
+        if (timeline.getTimelineSet() !== this) {
+            throw new Error(`EventTimelineSet.addEventToTimeline: Timeline=${timeline.toString()} does not belong " +
+                "in timelineSet(threadId=${this.thread?.id})`);
+        }
+
+        // Make sure events don't get mixed in timelines they shouldn't be in (e.g. a
+        // threaded message should not be in the main timeline).
+        //
+        // We can only run this check for timelines with a `room` because `canContain`
+        // requires it
+        if (this.room && !this.canContain(event)) {
+            let eventDebugString = `event=${event.getId()}`;
+            if (event.threadRootId) {
+                eventDebugString += `(belongs to thread=${event.threadRootId})`;
+            }
+            logger.warn(
+                `EventTimelineSet.addEventToTimeline: Ignoring ${eventDebugString} that does not belong ` +
+                    `in timeline=${timeline.toString()} timelineSet(threadId=${this.thread?.id})`,
+            );
+            return;
+        }
+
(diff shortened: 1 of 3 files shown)


src/models/thread.ts
```diff
@@ -473,6 +499,8 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
                             .then((relations) => {
                                 if (relations.events.length) {
                                     event.makeReplaced(relations.events[0]);
+                                    // TODO: check with Germain: is this right?
+                                    this.insertEventIntoTimeline(event);
```
gsouquet, line 503: That looks fine!

src/models/thread.ts
```diff
@@ -281,7 +306,8 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
                  */
                 this.replayEvents?.push(event);
             } else {
-                this.addEventToTimeline(event, toStartOfTimeline);
+                // TODO: check with Germain is this right?
+                this.insertEventIntoTimeline(event);
```
gsouquet, line 310: And that too, that pretty much matches what we discussed yesterday

src/models/event-timeline-set.ts
```diff
@@ -756,6 +756,99 @@ export class EventTimelineSet extends TypedEventEmitter<EmittedEvents, EventTime
...
+        if (!parentEvent) {
+            // Not related to anything we know about - just append
+            this.addEventToTimeline(event, timeline, {
+                toStartOfTimeline: false,
+                fromCache: false,
+                timelineWasEmpty: false,
+                roomState,
+            });
+            return;
+        }
```
justjanne, line 823: As we lazy load threads, most of the time we won't have the parent event. Are we sure that we just want to append in that case? It might be better to just purely sort it by timestamp (or prepend if all events are newer) in that situation.

src/models/thread.ts
```diff
@@ -281,7 +306,7 @@ export class Thread extends ReadReceipt<EmittedEvents, EventHandlerMap> {
                  */
                 this.replayEvents?.push(event);
             } else {
-                this.addEventToTimeline(event, toStartOfTimeline);
+                this.insertEventIntoTimeline(event);
```
justjanne, line 309: This would still be called if MSC 3981 is enabled and we get a reaction via sync. We should probably check for MSC 3981 here and just append it if MSC 3981 is enabled.

src/models/event-timeline-set.ts
andybalaam: I think if the parent event is not here yet, we will add things to `replayEvents` in the thread, and not get here until later when the root has been loaded. Maybe @gsouquet is more sure?

src/models/thread.ts
andybalaam: Yes, I will do that.

src/models/event-timeline-set.ts
justjanne: Say we've got a thread of root - A - B - C - D - E - F, and we've loaded the root and E and F. Now we get a reaction for B. What should we do?

Currently, we'd be inserting it so the total timeline looks like [root -] ... [- E - F - reactionB]. But [root -] … [- reactionB - E - F] would be better, right?

gsouquet: The `replayEvents` is only before the thread has been initialised first and we have pagination tokens. So the event would be added to the timeline.
After reading through @justjanne 's comment, I think it's probably a better approach than what we can up with.

andybalaam: Ah, so this would mean starting from the first event if the parent event is not found, but otherwise behaving the same?

andybalaam: Added my interpretation of this as https://github.com/matrix-org/matrix-js-sdk/pull/3384/commits/1aee0a84a98b61da5901f71775b5791062fb5f20 - let me know whether that's what you meant.

justjanne (approved):

andybalaam:
Force merging since we are merging this without tests to allow maximum dogfooding. I'm writing tests as we speak.

PR #3385, opened 2023-05-19
Move crypto classes into a separate namespace

richvdh:
This is mostly to tidy up the typedoc, and to make sure that all the classes in `crypto-api` are included. It doesn't have much effect on users of the library (at least, not things like react-sdk that import individual files rather than the whole of `matrixcs`.)

### DIFF
--- a/src/matrix.ts
+++ b/src/matrix.ts
@@ -64,9 +64,15 @@ export { createNewMatrixCall } from "./webrtc/call";
 export type { MatrixCall } from "./webrtc/call";
 export { GroupCallEvent, GroupCallIntent, GroupCallState, GroupCallType } from "./webrtc/groupCall";
 export type { GroupCall } from "./webrtc/groupCall";
-export type { CryptoApi } from "./crypto-api";
-export { DeviceVerificationStatus } from "./crypto-api";
 export { CryptoEvent } from "./crypto";
+
+/**
+ * Types supporting cryptography.
+ *
+ * The most important is {@link Crypto.CryptoApi}, an instance of which can be retrieved via
+ * {@link MatrixClient.getCrypto}.
+ */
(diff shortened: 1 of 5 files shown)


t3chguy:
Isn't this a breaking change by removing/moving exports?

richvdh:
> Isn't this a breaking change by removing/moving exports?

Only `DeviceVerificationStatus` and `CryptoApi`. Both were added very recently, and neither are used (via matrixcs) in react-sdk.

richvdh:
> How recently? If we backport this change can we prevent a release containing those exports then losing them?

Not that recently, I'm afraid.

t3chguy:
Can we add a deprecated re-export that we skip from typedoc?

richvdh:
Turns out these are used in the react-sdk tests at least, so that rather makes a hole in my argument that re-exports are unnecessary.

src/matrix.ts
```diff
@@ -64,10 +64,30 @@ export { createNewMatrixCall } from "./webrtc/call";
...
+ * @deprecated use {@link Crypto.CryptoApi}
+ */
+export type { CryptoApi } from "./crypto-api";
+
+/**
+ * Backwards compatibility re-export
+ * @internal
+ * @deprecated use {@link Crypto.DeviceVerificationStatus}
+ */
+export type { DeviceVerificationStatus } from "./crypto-api";
```
t3chguy, line 89: This shouldn't be a type-only import if we're to maintain non-breaking compatibility, it may have been used as a class and not a type by callers

richvdh: good spot

richvdh:
```suggestion
export { DeviceVerificationStatus } from "./crypto-api";
```

t3chguy (approved):
:shipit:

PR #3390, opened 2023-05-19
Isolate changes to Thread global into a single describe block

andybalaam:
I came across this while doing something unrelated. This test modifies a global variable and leaves it modified.

This change makes sure it is set back to what it was before after the test.

Suggested view to review this: https://github.com/matrix-org/matrix-js-sdk/pull/3390/files?diff=split&w=1

---

### DIFF
--- a/spec/unit/event-timeline-set.spec.ts
+++ b/spec/unit/event-timeline-set.spec.ts
@@ -206,85 +206,97 @@ describe("EventTimelineSet", () => {
             expect(liveTimeline.getEvents().length).toStrictEqual(0);
         });
 
-        it("should allow edits to be added to thread timeline", async () => {
-            jest.spyOn(client, "supportsThreads").mockReturnValue(true);
-            jest.spyOn(client, "getEventMapper").mockReturnValue(eventMapperFor(client, {}));
-            Thread.hasServerSideSupport = FeatureSupport.Stable;
+        describe("When the server supports threads", () => {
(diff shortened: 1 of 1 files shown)


spec/unit/event-timeline-set.spec.ts
```diff
@@ -206,85 +206,97 @@ describe("EventTimelineSet", () => {
...
-            jest.spyOn(client, "supportsThreads").mockReturnValue(true);
-            jest.spyOn(client, "getEventMapper").mockReturnValue(eventMapperFor(client, {}));
-            Thread.hasServerSideSupport = FeatureSupport.Stable;
+        describe("When the server supports threads", () => {
+            let previousThreadHasServerSideSupport: FeatureSupport;
 
-            const sender = "@alice:matrix.org";
+            beforeAll(() => {
+                previousThreadHasServerSideSupport = Thread.hasServerSideSupport;
+                Thread.hasServerSideSupport = FeatureSupport.Stable;
```
richvdh, line 214: an alternative to stashing the previous value might be to do some `jest.spyOn`, which will naturally be reset after the tests?

richvdh (approved):

spec/unit/event-timeline-set.spec.ts
andybalaam: I tried to do this with `jest.spyOn` and couldn't get it to work, so merging as-is for now.

PR #3391, opened 2023-05-19
Test for inserting events into the timeline in timestamp order

andybalaam:
This is at least a basic test for the code we added in a rush in https://github.com/matrix-org/matrix-js-sdk/pull/3384

---


t3chguy (approved):

andybalaam:
Test failures were due to several timeouts, so maybe a glitch with the runner. Trying again.

PR #3393, opened 2023-05-22
Fix mark as unread button

weeman1337:
fixes https://github.com/vector-im/element-web/issues/25411

PSF-2018

## Checklist

---

### DIFF
--- a/spec/unit/room.spec.ts
+++ b/spec/unit/room.spec.ts
@@ -51,7 +51,7 @@ import { TestClient } from "../TestClient";
 import { ReceiptType, WrappedReceipt } from "../../src/@types/read_receipts";
 import { FeatureSupport, Thread, THREAD_RELATION_TYPE, ThreadEvent } from "../../src/models/thread";
 import { Crypto } from "../../src/crypto";
-import { mkThread } from "../test-utils/thread";
+import * as threadUtils from "../test-utils/thread";
 import { getMockClientWithEventEmitter, mockClientMethodsUser } from "../test-utils/client";
 import { logger } from "../../src/logger";
 import { IMessageOpts } from "../test-utils/test-utils";
@@ -168,30 +168,45 @@ describe("Room", function () {
             room.client,
         );
 
-    const addRoomMainAndThreadMessages = (
-        room: Room,
-        tsMain?: number,
-        tsThread?: number,
-    ): { mainEvent?: MatrixEvent; threadEvent?: MatrixEvent } => {
-        const result: { mainEvent?: MatrixEvent; threadEvent?: MatrixEvent } = {};
-        const { rootEvent, thread } = mkThread({
+    /**
+     * Starts a new thread in a room by creating a message as thread root.
+     * Also creates a Thread model and adds it to the room.
+     * Does not insert the messages into a timeline.
+     */
+    const mkThread = (args: Partial<typeof threadUtils.mkThread>) => {
+        return threadUtils.mkThread({
             room,
             client: new TestClient().client,
             authorId: "@bob:example.org",
             participantUserIds: ["@bob:example.org"],
+            ...args,
         });
+    };
 
-        if (tsMain) {
-            result.mainEvent = mkMessage({ ts: tsMain });
-            room.addLiveEvents([result.mainEvent]);
-        }
+    /**
+     * Creates a message and adds it to the live timeline.
+     *
+     * @param room - Room to add the message to
+     * @param timestamp - Timestamp of the message
+     * @return The message event
+     */
+    const mkMessageInRoom = (room: Room, timestamp: number) => {
+        const message = mkMessage({ ts: timestamp });
+        room.addLiveEvents([message]);
+        return message;
+    };
 
-        if (tsThread) {
-            result.threadEvent = mkThreadResponse(rootEvent, { ts: tsThread });
-            thread.liveTimeline.addEvent(result.threadEvent, { toStartOfTimeline: true });
-        }
-
-        return result;
+    /**
+     * Creates a message in a thread and adds it to the thread live timeline.
+     *
+     * @param thread - Thread to add the message to
+     * @param timestamp - Timestamp of the message
+     * @returns The thread message event
+     */
+    const mkMessageInThread = (thread: Thread, timestamp: number) => {
+        const message = mkThreadResponse(thread.rootEvent!, { ts: timestamp });
+        thread.liveTimeline.addEvent(message, { toStartOfTimeline: true });
+        return message;
     };
 
     const addRoomThreads = (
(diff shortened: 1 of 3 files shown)


spec/unit/room.spec.ts
```diff
@@ -3562,6 +3562,11 @@ describe("Room", function () {
                 lastEventInMainTimeline = addRoomMainAndThreadMessages(room, 42, 23).mainEvent!;
                 expect(room.getLastLiveEvent()).toBe(lastEventInMainTimeline);
             });
+
+            it("and both events have the same timestamp, it should return the last event from the main timeline", () => {
```
gsouquet, line 3566: Should it? I'm not sure what bug that would fix. 
How does the main timeline take precedence? If two events have the same `ts` I don't think we have an accurate way of knowing in our current setup their order in the DAG hosted by synapse

weeman1337: I will revert this change for this PR. We need to fix the type error.

src/models/room.ts
```diff
@@ -809,7 +809,7 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
 
         const lastThreadEvent = lastThread.events[lastThread.events.length - 1];
 
-        return (lastRoomEvent?.getTs() ?? 0) > (lastThreadEvent.getTs() ?? 0) ? lastRoomEvent : lastThreadEvent;
+        return (lastRoomEvent?.getTs() ?? 0) > (lastThreadEvent?.getTs() ?? 0) ? lastRoomEvent : lastThreadEvent;
```
richvdh, line 812: Can you explain how this fixes the issue? I'm not really following.

richvdh: oh, because the issue was caused by `Cannot read properties of undefined`. Ignore me, I'm being stupid.

That said: I would expect to see a test for this situation, and I don't?

weeman1337: I encountered the same problem as Karl this morning. Console showed `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'getTs')`.

weeman1337: > oh, because the issue was caused by `Cannot read properties of undefined`. Ignore me, I'm being stupid.
> 
> That said: I would expect to see a test for this situation, and I don't?

The test is to create a thread anyway (even it is empty) : https://github.com/matrix-org/matrix-js-sdk/pull/3393/files#diff-711c041decd50c6513b88182abb538a73c5c10148570cce573b486b876359606R177 In this case it runs into the `Cannot read properties of undefined` error.

spec/unit/room.spec.ts
```diff
@@ -174,19 +174,19 @@ describe("Room", function () {
         tsThread?: number,
     ): { mainEvent?: MatrixEvent; threadEvent?: MatrixEvent } => {
         const result: { mainEvent?: MatrixEvent; threadEvent?: MatrixEvent } = {};
+        const { rootEvent, thread } = mkThread({
+            room,
+            client: new TestClient().client,
+            authorId: "@bob:example.org",
+            participantUserIds: ["@bob:example.org"],
+        });
```
richvdh, line 182: It seems to me that adding this here means that we are no longer testing the behaviour when there are no thread events? Or is that tested somewhere else?

I'm unclear what exactly `addRoomMainAndThreadMessages` is supposed to do, and what the semantics of `tsMain` and `tsThread` are supposed to do. Given the fact that we need to change this, I'm evidently not alone.

Please could you give the function a proper doc comment describing what it does, and what `tsMain` and `tsThread` do.

richvdh (changes requested):
I'm afraid I don't understand.

It's not helped by the fact that `mkThread` is also undocumented (and uses a bunch of additional undocumented methods), but that's not necessarily an issue to solve right now.

src/models/room.ts
richvdh: I don't really see how that works. `mkThread` adds two events to the thread, afaict, so we're still not testing the situation where `lastThread.events` is empty?

spec/unit/room.spec.ts
weeman1337: @richvdh take a look at the last commit. I threw away the magic setup function. It should (hopefully) be clearer now.

spec/unit/room.spec.ts
```diff
@@ -168,30 +168,45 @@ describe("Room", function () {
...
-        tsThread?: number,
-    ): { mainEvent?: MatrixEvent; threadEvent?: MatrixEvent } => {
-        const result: { mainEvent?: MatrixEvent; threadEvent?: MatrixEvent } = {};
-        const { rootEvent, thread } = mkThread({
+    /**
+     * Starts a new thread in a room by creating a message as thread root.
+     * Also creates a Thread model and adds it to the room.
+     * Does not insert the messages into a timeline.
+     */
+    const mkThread = (args: Partial<typeof threadUtils.mkThread>) => {
```
richvdh, line 176: really?

`threadUtils.mkThread` is a function type, so this is saying `args` is of type `Partial<Function<...>>` ?

spec/unit/room.spec.ts
```diff
@@ -168,30 +168,45 @@ describe("Room", function () {
...
-                room,
-                client: new TestClient().client,
-                authorId: "@bob:example.org",
-                participantUserIds: ["@bob:example.org"],
-            });
-            result.threadEvent = mkThreadResponse(rootEvent, { ts: tsThread });
-            thread.liveTimeline.addEvent(result.threadEvent, { toStartOfTimeline: true });
-        }
+    /**
+     * Creates a message and adds it to the live timeline.
```
richvdh, line 187:
```suggestion
     * Creates a message and adds it to the end of the main live timeline.
```

spec/unit/room.spec.ts
```diff
@@ -3534,32 +3534,46 @@ describe("Room", function () {
...
-            lastEventInMainTimeline = addRoomMainAndThreadMessages(room, 23).mainEvent!;
-            room.addLiveEvents([lastEventInMainTimeline]);
+            const lastEventInMainTimeline = mkMessageInRoom(room, 23);
             expect(room.getLastLiveEvent()).toBe(lastEventInMainTimeline);
         });
 
         it("when there is no event in the room live timeline but in a thread, it should return the last event from the thread", () => {
-            lastEventInThread = addRoomMainAndThreadMessages(room, undefined, 42).threadEvent!;
+            const { thread } = mkThread({ room, length: 0 });
+            const lastEventInThread = mkMessageInThread(thread, 42);
```
richvdh, line 3548: isn't it impossible for there to be events in the thread and not the main timeline, because the root event is in both?

spec/unit/room.spec.ts
```diff
@@ -3534,32 +3534,46 @@ describe("Room", function () {
...
+            });
+
+            it("and both events have the same timestamp, it should return the last event from the thread", () => {
+                mkMessageInRoom(room, 23);
+                const { thread } = mkThread({ room, length: 0 });
+                const lastEventInThread = mkMessageInThread(thread, 23);
+                expect(room.getLastLiveEvent()).toBe(lastEventInThread);
+            });
+
+            it("and there is a thread without last message, it should return the last event from the main timeline", () => {
```
richvdh, line 3574:
```suggestion
            it("and there is a thread without any messages, it should return the last event from the main timeline", () => {
```

richvdh (commented):
it's certainly much clearer now, thanks! Still a couple of questions, I'm afraid

spec/unit/room.spec.ts
weeman1337: Done

spec/unit/room.spec.ts
```diff
@@ -168,30 +168,45 @@ describe("Room", function () {
...
-        tsMain?: number,
-        tsThread?: number,
-    ): { mainEvent?: MatrixEvent; threadEvent?: MatrixEvent } => {
-        const result: { mainEvent?: MatrixEvent; threadEvent?: MatrixEvent } = {};
-        const { rootEvent, thread } = mkThread({
+    /**
+     * Starts a new thread in a room by creating a message as thread root.
+     * Also creates a Thread model and adds it to the room.
+     * Does not insert the messages into a timeline.
+     */
```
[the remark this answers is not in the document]
weeman1337: I have documented and linked the return type in the utility module.

spec/unit/room.spec.ts
weeman1337: It should also be impossible to have a thread model without events :man_shrugging: (which actually causes this bug).

 I would keep this for completeness and add a comment.

PR #3399, opened 2023-05-24
Only add a local receipt if it's after an existing receipt

andybalaam:
I suggest reviewing commit by commit.

This fixes a part that was missing from [Attempt a potential workaround for stuck notifs](https://github.com/matrix-org/matrix-js-sdk/pull/3384) - that PR inserted events that have arrived in the wrong order into (approx) the correct place in the timeline.

This change refuses to make a local echo receipt for a late-arriving event. We always consider synthesised receipts as the most important, so by making one for an old event, we were fooling ourselves into thinking newer stuff hadn't been read.

(This only applies when we don't have server support for recursive relations. When we do, we will simply trust the order of events from the server, so we won't consider timestamps - we'll just know the most recent event from the server is the most recent event, so it should get a receipt.)

---

## 🐛 Bug Fixes
 * Only add a local receipt if it's after an existing receipt ([\#3399](https://github.com/matrix-org/matrix-js-sdk/pull/3399)). Contributed by @andybalaam.

### DIFF
--- a/spec/unit/models/thread.spec.ts
+++ b/spec/unit/models/thread.spec.ts
@@ -465,58 +465,55 @@ describe("Thread", () => {
             ]);
         });
 
-        it("Creates a local echo receipt for new events", async () => {
-            // Assumption: no server side support because if we have it, events
-            // can only be added to the timeline after the thread has been
-            // initialised, and we are not properly initialising it here.
-            expect(Thread.hasServerSideSupport).toBe(FeatureSupport.None);
+        describe("Without relations recursion support", () => {
+            it("Creates a local echo receipt for new events", async () => {
+                // Assumption: no server side support because if we have it, events
+                // can only be added to the timeline after the thread has been
+                // initialised, and we are not properly initialising it here.
+                expect(Thread.hasServerSideSupport).toBe(FeatureSupport.None);
 
-            const client = createClientWithEventMapper();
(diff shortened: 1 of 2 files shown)


spec/unit/models/thread.spec.ts
```diff
@@ -503,11 +503,59 @@ describe("Thread", () => {
...
+            it("Creates a local echo receipt for new events", async () => {
+                // Assumption: no server side support because if we have it, events
+                // can only be added to the timeline after the thread has been
+                // initialised, and we are not properly initialising it here.
+                expect(Thread.hasServerSideSupport).toBe(FeatureSupport.None);
+
+                // Given a client WITH relations recursion support
+                const client = createClientWithEventMapper(
+                    new Map([[Feature.RelationsRecursion, ServerSupport.Stable]]),
+                );
```
weeman1337, line 525: This could go to a `beforeEach` block because it is shared code.

weeman1337 (approved):
Looks good :+1:

spec/unit/models/thread.spec.ts
andybalaam: I hate beforeEach blocks https://www.artificialworlds.net/blog/2013/03/27/setup-and-teardown-considered-harmful/ :-)

PR #3400, opened 2023-05-24
Fall back to receipt timestamp if we have no event (js-sdk part)

andybalaam:
There is similar logic in matrix-react-sdk, so we change that in https://github.com/matrix-org/matrix-react-sdk/pull/10974

---

## 🐛 Bug Fixes
 * Fall back to receipt timestamp if we have no event (js-sdk part) ([\#3400](https://github.com/matrix-org/matrix-js-sdk/pull/3400)). Contributed by @andybalaam.

### DIFF
--- a/src/models/read-receipt.ts
+++ b/src/models/read-receipt.ts
@@ -64,6 +64,7 @@ export abstract class ReadReceipt<
 
     public abstract getUnfilteredTimelineSet(): EventTimelineSet;
     public abstract timeline: MatrixEvent[];
+    public abstract findEventById(eventId: string): MatrixEvent | undefined;
 
     /**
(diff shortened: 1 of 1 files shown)


andybalaam:
Not doing this for now because:

1. The problem I am finding right now is to do with stuck unreads, whereas this is about notifs, so https://github.com/matrix-org/matrix-react-sdk/pull/10974 is more relevant
2. With any luck, the homeserver will give us the correct count for notifs, so if we make sure we believe that, it may be more effective than guessing as we are here
3. If we do want to guess within matrix-js-sdk, we should return an enum with e.g. `Read`, `Unread`, `ProbablyUnread`, `ProbablyRead` or similar to be clear that we are using heuristics.

PR #3411, opened 2023-05-25
Adjust fetchEditsWhereNeeded to use a clearer filter and async function

andybalaam:
The tests already cover this logic - just refactoring it to be clearer.

Slight change: before this did all events that were either in thread or not a relation. In fact, we only do this for thread events, so we explicitly check for messages that are either in the thread or thread roots. This should have the same effect.

---


weeman1337 (approved):

weeman1337:
> The tests already cover this logic - just refactoring it to be clearer.

Good! Refactor with confidence :+1:

gsouquet (approved):

PR #3413, opened 2023-05-25
Add remote user and call id in webrtc call stats

EnricoSchw:
Mark the webrtc call stats reports with the call id and the remote user id

The PR depends on: https://github.com/matrix-org/matrix-js-sdk/pull/3404

## Checklist

---

### DIFF
--- a/src/webrtc/stats/connectionStatsReportBuilder.ts
+++ b/src/webrtc/stats/connectionStatsReportBuilder.ts
@@ -16,7 +16,7 @@ limitations under the License.
 import { AudioConcealment, CodecMap, ConnectionStatsReport, FramerateMap, ResolutionMap, TrackID } from "./statsReport";
 import { MediaTrackStats, Resolution } from "./media/mediaTrackStats";
 
-export class StatsReportBuilder {
+export class ConnectionStatsReportBuilder {
     public static build(stats: Map<TrackID, MediaTrackStats>): ConnectionStatsReport {
         const report = {} as ConnectionStatsReport;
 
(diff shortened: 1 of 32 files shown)


src/webrtc/stats/connectionStatsReportBuilder.ts
```diff
@@ -17,8 +17,12 @@ import { AudioConcealment, CodecMap, ConnectionStatsReport, FramerateMap, Resolu
 import { MediaTrackStats, Resolution } from "./media/mediaTrackStats";
 
 export class ConnectionStatsReportBuilder {
-    public static build(stats: Map<TrackID, MediaTrackStats>): ConnectionStatsReport {
-        const report = {} as ConnectionStatsReport;
+    public static build(
+        callId: string,
+        opponentMemberId: string,
+        stats: Map<TrackID, MediaTrackStats>,
+    ): ConnectionStatsReport {
```
robintown, line 24: This is a breaking change, which we should try to avoid doing for something this small

src/webrtc/stats/statsReport.ts
```diff
@@ -28,10 +28,14 @@ export type TrackID = string;
...
 export interface ByteSentStatsReport extends Map<TrackID, ByteSend> {
+    callId: string;
+    opponentMemberId: string;
     // is a map: `local trackID` => byte send
 }
 
 export interface ConnectionStatsReport {
+    callId: string;
+    opponentMemberId: string;
     bandwidth: ConnectionStatsBandwidth;
```
robintown, line 39: These are also technically breaking changes

EnricoSchw: How can we avoid these?
It is necessary to mark the reports with the CallId, otherwise we cannot assign them to the call because we distribute them using events.

robintown: I'd suggest to make all of these new fields and parameters optional, with the implication that if you provide them when calling `build` they'll be set in the in output object

robintown (approved):

PR #3414, opened 2023-05-26
Element-R: Pull out an interface from `VerificationBase`

richvdh:
The abstract base class `VerificationBase` (along with its two concrete subclasses `SAS` and `ReciprocateQRCode`) implement the logic for performing a device or user verification once a specific verification method has been agreed between the two parties.

We want to reimplement this in Rust, so as a first step, pull out an interface from the current Javascript implementation, which in future we can implement with a rust-backed implementation - and deprecate direct access to the old classes..

Notes: Introduce a new `Crypto.Verifier` interface, and deprecate direct access to `VerificationBase`, `SAS` and `ReciprocateQRCode`

Part of https://github.com/vector-im/crypto-internal/issues/95.

---

## 🦖 Deprecations
 * Introduce a new `Crypto.Verifier` interface, and deprecate direct access to `VerificationBase`, `SAS` and `ReciprocateQRCode` ([\#3414](https://github.com/matrix-org/matrix-js-sdk/pull/3414)).

### DIFF
--- a/spec/integ/crypto/verification.spec.ts
+++ b/spec/integ/crypto/verification.spec.ts
@@ -18,9 +18,8 @@ import fetchMock from "fetch-mock-jest";
 import { MockResponse } from "fetch-mock";
 
 import { createClient, MatrixClient } from "../../../src";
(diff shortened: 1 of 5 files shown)


andybalaam (approved):
Great, thanks.

PR #3416, opened 2023-05-26
Support for stable MSC3882 get_login_token

hughns:
## Checklist

---

## ✨ Features
 * Support for stable MSC3882 get_login_token ([\#3416](https://github.com/matrix-org/matrix-js-sdk/pull/3416)). Contributed by @hughns.

### DIFF
--- a/spec/integ/matrix-client-methods.spec.ts
+++ b/spec/integ/matrix-client-methods.spec.ts
@@ -1206,13 +1206,11 @@ describe("MatrixClient", function () {
         it("should hit the expected API endpoint with UIA", async () => {
             httpBackend
                 .when("GET", "/capabilities")
-                .respond(200, { capabilities: { "org.matrix.msc3882.get_login_token": { enabled: true } } });
+                .respond(200, { capabilities: { "m.get_login_token": { enabled: true } } });
             const response = {};
             const uiaData = {};
             const prom = client.requestLoginToken(uiaData);
-            httpBackend
-                .when("POST", "/unstable/org.matrix.msc3882/login/get_token", { auth: uiaData })
-                .respond(200, response);
+            httpBackend.when("POST", "/v1/login/get_token", { auth: uiaData }).respond(200, response);
             await httpBackend.flush("");
             expect(await prom).toStrictEqual(response);
         });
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -8002,29 +8015,38 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+        const usesStable = !!capabilities["m.get_login_token"];
+        const usesUnstableRevision1 = !usesStable && !!capabilities["org.matrix.msc3882.get_login_token"];
+        // use stable endpoint if capability is exposed otherwise use old unstable r0 endpoint
+        const endpoint =
+            usesStable || usesUnstableRevision1
+                ? "/login/get_token" // endpoint for stable + unstable r1
+                : "/login/token"; // unstable r0 endpoint
+
+        // determine whether to use stable or unstable prefix
+        const prefix = usesStable ? ClientPrefix.V1 : `${ClientPrefix.Unstable}/org.matrix.msc3882`;
```
richvdh, line 8038: If I understand correctly, there are three different potential endpoints here? 

tbh I think it would be easier to set `prefix = ""` and embed the prefix in `endpoint`:

```suggestion
        let endpoint: String;
        if (capabilities["m.get_login_token"]) {
            // use the stable endpoint
            endpoint = ClientPrefix.V1 + "/login/get_token";
        } else if (capabilities["org.matrix.msc3882.get_login_token"]) {
            // newer unstable endpoint
            endpoint = ClientPrefix.Unstable + "/org.matrix.msc3882/login/get_token";
        } else {
            // old unstable endpoint
            endpoint = ClientPrefix.Unstable + "/org.matrix.msc3882/login/token";
        }
```

richvdh: (I also think that `ClientPrefix.V1` is an obscure way to write `"/_matrix/client/v1"`, but that's a side issue and I may be alone on this)

richvdh: Oh also, rather than hardcoding the capability names everywhere, let's use the symbolic constant, `GET_LOGIN_TOKEN_CAPABILITY.name` and `GET_LOGIN_TOKEN_CAPABILITY.altName`.

spec/integ/matrix-client-methods.spec.ts
```diff
@@ -1206,36 +1206,34 @@ describe("MatrixClient", function () {
...
             const prom = client.requestLoginToken();
-            httpBackend.when("POST", "/unstable/org.matrix.msc3882/login/get_token", {}).respond(200, response);
+            httpBackend.when("POST", "/v1/login/get_token", {}).respond(200, response);
             await httpBackend.flush("");
             // check that expires_in has been populated for compatibility with r0
             expect(await prom).toStrictEqual({ ...response, expires_in: 5 });
         });
 
-        it("should hit the r1 endpoint when capability is disabled", async () => {
+        it("should hit the stable endpoint when capability is disabled", async () => {
```
richvdh, line 1230: I had to read this test a few times to understand what it is about. I think this might help:

```suggestion
        it("should still hit the stable endpoint when capability is disabled (but present)", async () => {
```

richvdh (changes requested):
Generally seems good; a few nits

src/client.ts
hughns: > If I understand correctly, there are three different potential endpoints here?

Yes. And your proposal is much clearer to read so I have incorporated it. 👍

src/client.ts
```diff
@@ -8026,24 +8026,24 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
-                ? "/login/get_token" // endpoint for stable + unstable r1
-                : "/login/token"; // unstable r0 endpoint
-
-        // determine whether to use stable or unstable prefix
-        const prefix = usesStable ? ClientPrefix.V1 : `${ClientPrefix.Unstable}/org.matrix.msc3882`;
+        let endpoint: string;
+        if (capabilities[GET_LOGIN_TOKEN_CAPABILITY.name]) {
+            // use the stable endpoint
+            endpoint = `${ClientPrefix.V1}/login/get_token`;
+        } else if (capabilities[GET_LOGIN_TOKEN_CAPABILITY.altName as string]) {
```
richvdh, line 8033:
```suggestion
        } else if (capabilities[GET_LOGIN_TOKEN_CAPABILITY.altName!]) {
```

... I think?

src/client.ts
```diff
@@ -8026,24 +8026,24 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            endpoint = `${ClientPrefix.Unstable}/org.matrix.msc3882/login/token`;
+        }
 
         const body: UIARequest<{}> = { auth };
         const res = await this.http.authedRequest<UIAResponse<LoginTokenPostResponse>>(
             Method.Post,
             endpoint,
             undefined, // no query params
             body,
-            { prefix },
```
richvdh: does this do the right thing? I think if you don't specify `prefix` at all it defaults to `/_matrix/client/v1` or something. You'll need to set `prefix: ""`.

(Also suggests the tests could do with tightening up?)

hughns: Well spotted. Due to the `matrix-mock-request` behaviour of doing partial path matching the tests were incorrectly passing.

I hope I have found an acceptable technique to do an exact match on the complete generated URL without replacing the mock backend. 🤞

richvdh (approved):

andybalaam:
I would force merge but there is another PR in the queue, so let's try updating the branch to my latest changes in the hope that they unflake it.

PR #3419, opened 2023-05-29
OIDC: Validate `m.authentication` configuration

kerryarchibald:
For https://github.com/vector-im/element-web/issues/25466

Validates `.well-known` discovery result to check
- `m.authentication` config is valid
- `m.authentication.issuer` has valid configuration

Adds `m.authentication` section to `clientConfig`

Details in issue

Result is not used in `react-sdk` yet, but can be tested by entering a server name in the server picker (eg `synapse-oidc.lab.element.dev`)
Or configuring `default_server_config.m.authentication` in `config.json`
```
"default_server_config": {
        "m.homeserver": {
            "base_url": "https://synapse-oidc.lab.element.dev"
        },
        "org.matrix.msc2965.authentication": {
            "issuer": "https://auth-oidc.lab.element.dev/"
        }

    },
```

## Checklist

---

## ✨ Features
 * OIDC: Validate `m.authentication` configuration ([\#3419](https://github.com/matrix-org/matrix-js-sdk/pull/3419)). Contributed by @kerryarchibald.


src/autodiscovery.ts
```diff
@@ -256,10 +266,56 @@ export class AutoDiscovery {
...
         });
 
+        const authConfig = await this.validateDiscoveryAuthenticationConfig(wellknown);
+        clientConfig[M_AUTHENTICATION.stable!] = authConfig;
+
         // Step 8: Give the config to the caller (finally)
         return Promise.resolve(clientConfig);
     }
 
+    public static async validateDiscoveryAuthenticationConfig(
```
gsouquet, line 276: Can we add a JSDoc explaining what this does?
It seems to be doing a fair few things and there's a few failure scenarios that are not obvious

src/oidc/validate.ts
```diff
@@ -0,0 +1,112 @@
...
+
+import { IClientWellKnown, IDelegatedAuthConfig, M_AUTHENTICATION } from "../client";
+import { logger } from "../logger";
+
+export enum OidcDiscoveryError {
+    NotSupported = "OIDC authentication not supported",
+    Misconfigured = "OIDC is misconfigured",
+    General = "Something went wrong with OIDC discovery",
+    OpSupport = "Configured OIDC OP does not support required functions",
+}
```
gsouquet, line 25: None of those are wrapped in our i18n function. Double checking whether this is expected or not?

Do we expect the wrapping to happen in the views?

src/oidc/validate.ts
```diff
@@ -0,0 +1,112 @@
...
+export const validateWellKnownAuthentication = (wellKnown: IClientWellKnown): IDelegatedAuthConfig => {
+    const authentication = M_AUTHENTICATION.findIn<IDelegatedAuthConfig>(wellKnown);
+
+    if (!authentication) {
+        throw new Error(OidcDiscoveryError.NotSupported);
+    }
+
+    if (
+        typeof authentication.issuer === "string" &&
+        (!authentication.account || typeof authentication.account === "string")
```
gsouquet, line 49: Should we use `authentication.hasOwnProperty("account")` instead? Looks like passing `false` as an account value would make this `if` statement pass, and we'd end up with an unexpected value

src/oidc/validate.ts
```diff
@@ -0,0 +1,112 @@
...
+        return {
+            issuer: authentication.issuer,
+            account: authentication.account,
+        };
+    }
+
+    throw new Error(OidcDiscoveryError.Misconfigured);
+};
+
+// force into a record to make accessing properties easier
```
gsouquet, line 60: I'm not sure this comment matches the function below.
There's no "casting", but it's rather a check

gsouquet (changes requested):
Globally looks good, added a few comments that I believe need to be addressed before merging this 👍

src/oidc/validate.ts
kerryarchibald: These are only expected to be logged so not intended to be localised.
It's not fully planned if/how/where these errors will be surfaced to the user but they'll be mapped to localised strings in react-sdk.

PR #3420, opened 2023-05-30
Update MSC3912 implementation to use `with_rel_type` instead of `with_relations`

weeman1337:
Adapts the implementation to the latest version of [MSC3912](https://github.com/matrix-org/matrix-spec-proposals/pull/3912#issuecomment-1567173949)

- `with_relations` → `with_rel_types`
- `with_rel_types` is now always a list of strings incl. `*`
- I decided to keep `with_relations` and mark it as deprecated. It should be removed once the MSC becomes stable.

Part of https://github.com/vector-im/element-web/issues/25471

## Checklist

---

## ✨ Features
 * Update MSC3912 implementation to use `with_rel_type` instead of `with_relations` ([\#3420](https://github.com/matrix-org/matrix-js-sdk/pull/3420)).

### DIFF
--- a/spec/unit/matrix-client.spec.ts
+++ b/spec/unit/matrix-client.spec.ts
@@ -340,7 +340,7 @@ describe("MatrixClient", function () {
         store.storeClientOptions = jest.fn().mockReturnValue(Promise.resolve(null));
         store.isNewlyCreated = jest.fn().mockReturnValue(Promise.resolve(true));
 
-        // set unstableFeatures to a defined state before reach test
+        // set unstableFeatures to a defined state before each test
         unstableFeatures = {
             "org.matrix.msc3440.stable": true,
         };
--- a/spec/unit/matrix-client.spec.ts
+++ b/spec/unit/matrix-client.spec.ts
@@ -28,7 +28,6 @@ import {
     UNSTABLE_MSC3088_ENABLED,
     UNSTABLE_MSC3088_PURPOSE,
     UNSTABLE_MSC3089_TREE_SUBTYPE,
-    MSC3912_RELATION_BASED_REDACTIONS_PROP,
 } from "../../src/@types/event";
 import { MEGOLM_ALGORITHM } from "../../src/crypto/olmlib";
 import { Crypto } from "../../src/crypto";
@@ -1376,10 +1375,10 @@ describe("MatrixClient", function () {
             await client.redactEvent(roomId, eventId, txnId, { reason });
         });
 
-        describe.each(["with_rel_types", "with_relations"])("when calling with %s", (withRelTypesPropName) => {
+        describe("when calling with 'with_rel_types'", () => {
             const eventId = "$event42:example.org";
 
-            it("should raise an error if server has no support for relation based redactions", async () => {
+            it("should raise an error if the server has no support for relation based redactions", async () => {
                 // load supported features
                 await client.getVersions();
 
@@ -1387,7 +1386,7 @@ describe("MatrixClient", function () {
 
                 expect(() => {
                     client.redactEvent(roomId, eventId, txnId, {
-                        [withRelTypesPropName]: [RelationType.Reference],
+                        with_rel_types: [RelationType.Reference],
                     });
                 }).toThrow(
                     new Error(
@@ -1397,40 +1396,32 @@ describe("MatrixClient", function () {
                 );
             });
 
-            // expects both string[] and string to work
-            describe.each([[RelationType.Reference], RelationType.Reference])(
-                "and the server supports relation based redactions (%s, unstable)",
-                (relationType: RelationType | Array<RelationType>) => {
-                    beforeEach(async () => {
-                        unstableFeatures["org.matrix.msc3912"] = true;
-                        // load supported features
-                        await client.getVersions();
-                    });
+            it("and the server has unstable support for relation based redactions, it should send 'org.matrix.msc3912.with_relations' in the request body", async () => {
+                unstableFeatures["org.matrix.msc3912"] = true;
+                // load supported features
+                await client.getVersions();
 
-                    it("should send with_rel_types in the request body", async () => {
-                        const txnId = client.makeTxnId();
+                const txnId = client.makeTxnId();
 
-                        httpLookups = [
-                            {
-                                method: "PUT",
-                                path:
-                                    `/rooms/${encodeURIComponent(roomId)}/redact/${encodeURIComponent(eventId)}` +
-                                    `/${encodeURIComponent(txnId)}`,
-                                expectBody: {
-                                    reason: "redaction test",
-                                    [MSC3912_RELATION_BASED_REDACTIONS_PROP.unstable!]: [RelationType.Reference],
-                                },
-                                data: { event_id: eventId },
-                            },
-                        ];
-
-                        await client.redactEvent(roomId, eventId, txnId, {
+                httpLookups = [
+                    {
+                        method: "PUT",
+                        path:
+                            `/rooms/${encodeURIComponent(roomId)}/redact/${encodeURIComponent(eventId)}` +
+                            `/${encodeURIComponent(txnId)}`,
+                        expectBody: {
                             reason: "redaction test",
-                            [withRelTypesPropName]: relationType,
-                        });
-                    });
-                },
-            );
+                            ["org.matrix.msc3912.with_relations"]: ["m.reference"],
+                        },
+                        data: { event_id: eventId },
+                    },
+                ];
+
+                await client.redactEvent(roomId, eventId, txnId, {
+                    reason: "redaction test",
+                    with_rel_types: [RelationType.Reference],
+                });
+            });
         });
     });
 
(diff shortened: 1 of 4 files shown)


richvdh:
@weeman1337 a gentle request to try to make your PR subjects meaningful, and not to mark them as T-Task where there are functional changes :)

src/client.ts
```diff
@@ -4627,22 +4627,24 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
         const reason = opts?.reason;
+        let withRelTypes = opts?.with_rel_types ?? opts?.with_relations;
 
-        if (
-            opts?.with_relations &&
-            this.canSupport.get(Feature.RelationBasedRedactions) === ServerSupport.Unsupported
-        ) {
+        if (typeof withRelTypes === "string") {
+            withRelTypes = [withRelTypes];
+        }
```
richvdh, line 4634: why is this required? `IRedactOpts` does not permit a string here?

src/client.ts
```diff
@@ -4627,22 +4627,24 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+
+        if (withRelTypes && this.canSupport.get(Feature.RelationBasedRedactions) === ServerSupport.Unsupported) {
```
richvdh, line 4636: implicit cast to bool is forbidden by our code style.

Likewise below

spec/unit/matrix-client.spec.ts
```diff
@@ -1394,36 +1397,40 @@ describe("MatrixClient", function () {
...
-                    httpLookups = [
-                        {
-                            method: "PUT",
-                            path:
-                                `/rooms/${encodeURIComponent(roomId)}/redact/${encodeURIComponent(eventId)}` +
-                                `/${encodeURIComponent(txnId)}`,
-                            expectBody: {
-                                reason: "redaction test",
-                                [MSC3912_RELATION_BASED_REDACTIONS_PROP.unstable!]: [RelationType.Reference],
+                    it("should send with_rel_types in the request body", async () => {
```
richvdh, line 1410: ... except it doesn't, it uses the unstable prop name for now

spec/unit/matrix-client.spec.ts
```diff
@@ -1394,36 +1397,40 @@ describe("MatrixClient", function () {
...
+
+                        httpLookups = [
+                            {
+                                method: "PUT",
+                                path:
+                                    `/rooms/${encodeURIComponent(roomId)}/redact/${encodeURIComponent(eventId)}` +
+                                    `/${encodeURIComponent(txnId)}`,
+                                expectBody: {
+                                    reason: "redaction test",
+                                    [MSC3912_RELATION_BASED_REDACTIONS_PROP.unstable!]: [RelationType.Reference],
```
richvdh, line 1421: to ensure that we aren't making the same mistake twice (and to make the test clearer), I think it is better to hardcode the property names here:

```suggestion
                                    "org.matrix.msc3912.with_relations": ["m.reference"],
```

Likewise the reference to `RelationType` above.

src/client.ts
```diff
@@ -4627,22 +4627,24 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
                 "Server does not support relation based redactions " +
                     `roomId ${roomId} eventId ${eventId} txnId: ${txnId} threadId ${threadId}`,
             );
         }
 
-        const withRelations = opts?.with_relations
+        const withRelTypesContent = withRelTypes
             ? {
                   [this.canSupport.get(Feature.RelationBasedRedactions) === ServerSupport.Stable
                       ? MSC3912_RELATION_BASED_REDACTIONS_PROP.stable!
```
richvdh, line 4646: my understanding is that this line is currently unreachable, given that `Feature.RelationBasedRedactions` has no defined `matrixVersion` - is that correct?

Not necessarily asking you to change anything here. Though I do think it would be clearer to factor out a `propName` variable instead of putting the ternary condition inline.

src/client.ts
```diff
@@ -4627,22 +4627,24 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
-                      : MSC3912_RELATION_BASED_REDACTIONS_PROP.unstable!]: opts?.with_relations,
+                      : MSC3912_RELATION_BASED_REDACTIONS_PROP.unstable!]: withRelTypes,
               }
             : {};
```
richvdh, line 4649: how about enclosing all this in an `if (!!withRelTypes)` block, to save repeatedly checking it?

richvdh (changes requested):

src/client.ts
weeman1337: The previous version allowed string or list. But since it is an unstable function anyway, I will remove the fallback.

src/client.ts
weeman1337: Explicit cast to bool is forbidden by ESLint :clown_face: 

```
4633:13  error  Redundant double negation  no-extra-boolean-cast
```

src/client.ts
weeman1337: > my understanding is that this line is currently unreachable, given that Feature.RelationBasedRedactions has no defined matrixVersion - is that correct?

Yes

> Not necessarily asking you to change anything here. Though I do think it would be clearer to factor out a propName variable instead of putting the ternary condition inline.

Done

src/client.ts
richvdh: > The previous version allowed string or list.

Are you sure? https://github.com/matrix-org/matrix-js-sdk/blob/e10db6f7e9978e2d8dd3640fbe2be503710fd506/src/%40types/requests.ts#L61 requires it to be an array.

src/client.ts
richvdh: Well, that's new. Surely we use double-negation everywhere?

I don't consider this resolved: we can't just ignore the style guide. Either we agree as a team to change it, or we respect it.

Anyway, in this case an explicit test for `undefined` is more appropriate.

src/client.ts
```diff
@@ -4628,31 +4628,33 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            }
+
+            const withRelTypesPropName =
+                this.canSupport.get(Feature.RelationBasedRedactions) === ServerSupport.Stable
+                    ? MSC3912_RELATION_BASED_REDACTIONS_PROP.stable!
+                    : MSC3912_RELATION_BASED_REDACTIONS_PROP.unstable!;
+
+            withRelTypesContent = {
+                [withRelTypesPropName]: opts.with_rel_types,
+            };
```
richvdh, line 4648:
```suggestion
            withRelTypesContent[withRelTypesPropName] =  opts.with_rel_types;
```

(and make `withRelTypesContent` a `const`)?

Indeed, consider having a `const content = { reason };` object and updating that here, instead of using `...` destructuring below. But that's gold-plating

src/client.ts
```diff
@@ -4628,31 +4628,33 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+        let withRelTypesContent = {};
 
-        const withRelations = opts?.with_relations
-            ? {
-                  [this.canSupport.get(Feature.RelationBasedRedactions) === ServerSupport.Stable
-                      ? MSC3912_RELATION_BASED_REDACTIONS_PROP.stable!
-                      : MSC3912_RELATION_BASED_REDACTIONS_PROP.unstable!]: opts?.with_relations,
-              }
-            : {};
+        if (opts?.with_rel_types) {
```
richvdh, line 4633:
```suggestion
        if (opts?.with_rel_types !== undefined) {
```

spec/unit/matrix-client.spec.ts
```diff
@@ -342,6 +339,11 @@ describe("MatrixClient", function () {
         store.getClientOptions = jest.fn().mockReturnValue(Promise.resolve(null));
         store.storeClientOptions = jest.fn().mockReturnValue(Promise.resolve(null));
         store.isNewlyCreated = jest.fn().mockReturnValue(Promise.resolve(true));
+
+        unstableFeatures = {
+            "org.matrix.msc3440.stable": true,
+        };
```
richvdh, line 345:
```suggestion
        unstableFeatures["org.matrix.msc3440.stable"] = true;
```

(though I'm not sure why it is being moved at all)

richvdh (approved):
some nit-picking but generally lgtm

src/client.ts
weeman1337: You are right. Also the MSC didn't allow it to be a string. Anyway removing this block then makes event more sense.

spec/unit/matrix-client.spec.ts
weeman1337: This sets `unstableFeatures` to a defined state before each test. Another test sets an additional property on it: https://github.com/matrix-org/matrix-js-sdk/blob/deb908947b06047dc93bd5f82906e63133f9d3fd/spec/unit/matrix-client.spec.ts#L1400

weeman1337: → will add a comment

richvdh: oh, I see.

richvdh: oh, I see.

PR #3422, opened 2023-05-30
Add `getShowSasCallbacks`, `getShowQrCodeCallbacks` to VerifierBase

richvdh:
... to avoid the need for some type-casting in the application

---


florianduros (approved):
LGTM

richvdh:
This is now based on #3439

PR #3424, opened 2023-05-30
Send expected peer connections to posthog

toger5:
Send the peer expected peer connections based on roomState event in the SummaryReport.

## Checklist

### DIFF
--- a/src/webrtc/groupCall.ts
+++ b/src/webrtc/groupCall.ts
@@ -1595,6 +1595,9 @@ export class GroupCall extends TypedEventEmitter<
         });
 
         if (this.state === GroupCallState.Entered) this.placeOutgoingCalls();
+
+        // Update the participants stored in the stats object
+        this.stats?.setParticipants(this.participants);
     };
 
     private onStateChanged = (newState: GroupCallState, oldState: GroupCallState): void => {
@@ -1628,7 +1631,7 @@ export class GroupCall extends TypedEventEmitter<
     public getGroupCallStats(): GroupCallStats {
         if (this.stats === undefined) {
             const userID = this.client.getUserId() || "unknown";
-            this.stats = new GroupCallStats(this.groupCallId, userID, this.statsCollectIntervalTime);
+            this.stats = new GroupCallStats(this.groupCallId, userID, this.participants, this.statsCollectIntervalTime);
             this.stats.reports.on(StatsReport.CONNECTION_STATS, this.onConnectionStats);
             this.stats.reports.on(StatsReport.BYTE_SENT_STATS, this.onByteSentStats);
             this.stats.reports.on(StatsReport.SUMMARY_STATS, this.onSummaryStats);
--- a/src/webrtc/stats/groupCallStats.ts
+++ b/src/webrtc/stats/groupCallStats.ts
@@ -17,6 +17,8 @@ import { CallStatsReportGatherer } from "./callStatsReportGatherer";
 import { StatsReportEmitter } from "./statsReportEmitter";
 import { CallStatsReportSummary } from "./callStatsReportSummary";
 import { SummaryStatsReportGatherer } from "./summaryStatsReportGatherer";
+import { ParticipantState } from "../groupCall";
+import { RoomMember } from "../../matrix";
 
 export class GroupCallStats {
     private timer: undefined | ReturnType<typeof setTimeout>;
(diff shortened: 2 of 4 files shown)


src/webrtc/stats/statsReport.ts
```diff
@@ -79,4 +79,8 @@ export interface SummaryStatsReport {
     maxPacketLoss: number;
     percentageConcealedAudio: number;
     peerConnections: number;
+    roomStateExpectedPeerConnections: number;
+    missingPeerConnections: number;
+    percentageEstablishedPeerConnections: number;
+    // Todo: Decide if we want an index (or a timestamp) of this report in relation to the group call, to help differenciate when issues occur and ignore/track initial connection delays.
```
EnricoSchw, line 85: I think the names are interpretations of the actual statistics. I would name the values according to their real meaning and do the interpretation in Posthog. And I would also add the joined users in the room. Because it is unusual for a user to take part in a conference with several devices. If it happens too often, we can use this to identify a caching issues.
```typescript
usersInRoom: number;
devicesInRoom: number;
diffDevicesToPeerConnections: number;
ratioPeerConnectionToDevices : number;
```

src/webrtc/groupCall.ts
```diff
@@ -1629,6 +1632,7 @@ export class GroupCall extends TypedEventEmitter<
         if (this.stats === undefined) {
             const userID = this.client.getUserId() || "unknown";
             this.stats = new GroupCallStats(this.groupCallId, userID, this.statsCollectIntervalTime);
+            this.stats.setParticipants(this.participants);
```
EnricoSchw, line 1635: Just a note:
I try to avoid dependencies between the VoIP objects like `GroupCall`, `CallFeed`, ...  and the stats package for example with `GroupCallStats`. 
It's not that easy because the values have to be calculated somewhere. In addition we also want statistics based on VoIp objects. But many dependencies means `GroupCall` constantly wanting  update `GroupCallStats` .

To avoid complicated dependencies I installed the cascading event listener. So GroupCall can expand the Webrtc statistics with VoIP statistics. Here how i did it with CallFeed stats. This is also not optimal since ``CallFeedStatsReporter.expandCallFeedReport`` needs to know about `CallFeed` but it is less coupled. 

https://github.com/matrix-org/matrix-js-sdk/blob/f0673101981c5c2526568c19f31d2c55d4c0b427/src/webrtc/groupCall.ts#L291-L305

t3chguy:
Having posthog in the js-sdk changelog doesn't make much sense given that js-sdk has no concept of posthog, could you reword?

src/webrtc/stats/statsReport.ts
toger5: To make the devices more comparable with the peerConnections i only count the devices without the "local" device. I changed it to `oppDevicesInCall` and `oppUsersInCall`.

src/webrtc/stats/groupCallStats.ts
```diff
@@ -75,7 +75,9 @@ export class GroupCallStats {
             summary.push(c.processStats(this.groupCallId, this.userId));
         });
 
-        Promise.all(summary).then((s: Awaited<CallStatsReportSummary>[]) => this.summaryStatsReportGatherer.build(s));
+        Promise.all(summary)
+            .then((s: Awaited<CallStatsReportSummary>[]) => this.summaryStatsReportGatherer.build(s))
+            .catch((err) => alert(err));
```
robintown, line 80: Showing an alert when the stats gatherer fails seems like it would be a confusing experience for the user. Could we just log the errors?

src/webrtc/stats/statsReport.ts
```diff
@@ -83,4 +83,9 @@ export interface SummaryStatsReport {
     maxPacketLoss: number;
     percentageConcealedAudio: number;
     peerConnections: number;
+    oppUsersInCall?: number;
```
robintown, line 86: I'm not sure what `opp` is short for here - 'opponent'?

src/webrtc/stats/summaryStatsReporter.ts
```diff
@@ -22,7 +24,7 @@ interface CallStatsReportSummaryCounter {
     totalAudio: number;
 }
 
-export class SummaryStatsReportGatherer {
+export class SummaryStatsReporter {
```
robintown, line 27: This is a breaking change, as well as the file being renamed

src/webrtc/stats/groupCallStats.ts
```diff
@@ -77,7 +78,9 @@ export class GroupCallStats {
 
         Promise.all(summary)
             .then((s: Awaited<CallStatsReportSummary>[]) => this.summaryStatsReportGatherer.build(s))
-            .catch((err) => alert(err));
+            .catch((err) => {
+                logger.error("Cloud not build syummary stats report", err);
```
robintown, line 82:
```suggestion
                logger.error("Could not build summary stats report", err);
```

src/webrtc/stats/summaryStatsReporter.ts
robintown: Well, rather than marking the PR as a breaking change, I think my concrete suggestion would be to keep the old names so that we don't force a major version bump for something so small

robintown (approved):

robintown:
I don't think a review from the app team should be necessary here, it looks like something went wrong when you merged develop that has made GitHub think you've touched package.json

PR #3426, opened 2023-05-30
OIDC: accept homeserver URL or domain in `.well-known` discovery functions

kerryarchibald:
Part of https://github.com/vector-im/element-web/issues/25472
Used https://github.com/matrix-org/matrix-react-sdk/pull/11018 

Previously, we only lookup `.well-known` during login/registration for homeserver urls without a domain (eg `matrix.org`) in an effort to validate them as server names.
As part of OIDC support discovery, we need to lookup the client `.well-known` for all homeservers to discover `m.authentication` config on the wk.
Extend `findClientWellKnown` functions to accept both domain and url homeservers.

## Checklist

---

## ✨ Features
 * OIDC: accept homeserver URL or domain in `.well-known` discovery functions ([\#3426](https://github.com/matrix-org/matrix-js-sdk/pull/3426)). Contributed by @kerryarchibald.

### DIFF
--- a/spec/unit/autodiscovery.spec.ts
+++ b/spec/unit/autodiscovery.spec.ts
@@ -68,6 +68,26 @@ describe("AutoDiscovery", function () {
         ]);
     });
 
+    it("should handle a homeserver url that includes a protocol", async function () {
+        const httpBackend = getHttpBackend();
+        httpBackend.when("GET", "http://example.org/.well-known/matrix/client").respond(200, {});
+        AutoDiscovery.findClientConfig("http://example.org");
+        await httpBackend.flushAllExpected();
+
+        // .well-known fetched using protocol from provided domain
+        httpBackend.verifyNoOutstandingRequests();
+    });
+
+    it("should handle a homeserver domain that does not include a protocol", async function () {
+        const httpBackend = getHttpBackend();
+        httpBackend.when("GET", "https://example.org/.well-known/matrix/client").respond(200, {});
+        AutoDiscovery.findClientConfig("example.org");
+        await httpBackend.flushAllExpected();
+
+        // https protocol added to domain for request
+        httpBackend.verifyNoOutstandingRequests();
+    });
+
     it("should return PROMPT when .well-known 404s", function () {
         const httpBackend = getHttpBackend();
         httpBackend.when("GET", "/.well-known/matrix/client").respond(404, {});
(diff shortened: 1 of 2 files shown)


robintown (approved):
Looks reasonable to me!

t3chguy:
The spec only defines the well-known lookup from a server name (really from a Matrix ID) - not from a Homeserver URL. They don't always match, especially when delegation isn't configured. This doesn't seem spec compliant? https://spec.matrix.org/v1.7/client-server-api/#well-known-uri

richvdh:
> especially when delegation isn't configured.

s/isn't/is/?

As an example, the "homeserver URL" for matrix.org is https://matrix-client.matrix.org, but https://matrix-client.matrix.org/.well-known/matrix/client doesn't exist.

t3chguy:
> s/isn't/is/?

What I meant is if m.org didn't set up the client well-known file, users would need to enter `matrix-client.matrix.org` which wouldn't match the server_name. But the symptom stands either way, if the user manually inputs `matrix-client.matrix.org` which is valid, this seems like the well-known won't be discovered and thus OIDC will not function?

kerryarchibald:
> The spec only defines the well-known lookup from a server name (really from a Matrix ID) - not from a Homeserver URL. They don't always match, especially when delegation isn't configured. This doesn't seem spec compliant? https://spec.matrix.org/v1.7/client-server-api/#well-known-uri

Sure, but when the server name is unavailable we can make an assumption and try our best to discover the configuration. I couldn't find a way to discover the server name from the homeserver url (apart from registering as a guest?)
FTR Hydrogen looks up .well-known like this already.

Being unable to discover OIDC config without .well-known is [noted on the MSC](https://github.com/sandhose/matrix-doc/blob/msc/sandhose/oidc-discovery/proposals/2965-oidc-discovery.md#potential-issues) with the alternative of [a new dedicated endpoint](https://github.com/sandhose/matrix-doc/blob/msc/sandhose/oidc-discovery/proposals/2965-oidc-discovery.md#discovery-via-a-new-client-api-endpoint) suggested.


> if the user manually inputs `matrix-client.matrix.org` which is valid, this seems like the well-known won't be discovered

This is true with the existing implementation.

t3chguy:
> This is true with the existing implementation.

This sounds like a major issue, what happens if well-known (and thus OIDC) isn't discovered? User is unable to authenticate? Sounds like an issue with the MSC.

> Sure, but when the server name is unavailable we can make an assumption and try our best to discover the configuration. I couldn't find a way to discover the server name from the homeserver url (apart from registering as a guest?)

That's because a homeserver URL can actually correspond to multiple server_names (and thus well-knowns). At time of `/login` the server gives you a new URL to connect to, and its that server's server_name that matters here. The one corresponding to your account, not one guessed from a HS URL where there isn't a 1:1 mapping.

PR #3427, opened 2023-05-31
Ensure we do not add relations to the wrong timeline

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/25450
Fixes https://github.com/vector-im/element-web/issues/25494

Uses https://github.com/matrix-org/matrix-spec-proposals/pull/4023

If MSC4023 unsigned data is not included it'll use the `/event/` API to fetch the parent event to determine what thread to associate the relation with, it'll then store that unsigned in the sync accumulator to prevent needing to do that work again.

---

## 🚨 BREAKING CHANGES
 * Ensure we do not add relations to the wrong timeline ([\#3427](https://github.com/matrix-org/matrix-js-sdk/pull/3427)). Fixes vector-im/element-web#25450 and vector-im/element-web#25494.

### DIFF
--- a/src/models/event.ts
+++ b/src/models/event.ts
@@ -63,7 +63,7 @@ export interface IUnsigned {
     "transaction_id"?: string;
     "invite_room_state"?: StrippedState[];
     "m.relations"?: Record<RelationType | string, any>; // No common pattern for aggregated relations
-    "io.element.relation_thread_id"?: string;
+    "org.matrix.msc4023.thread_id"?: string;
 }
 
 export interface IThreadBundledRelationship {
--- a/src/models/event.ts
+++ b/src/models/event.ts
@@ -579,16 +579,8 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
         const relatesTo = this.getWireContent()?.["m.relates_to"];
         if (relatesTo?.rel_type === THREAD_RELATION_TYPE.name) {
             return relatesTo.event_id;
-        }
-        if (this.thread) {
-            return this.thread.id;
-        }
-        if (this.threadId) {
-            return this.threadId;
-        }
-        const unsigned = this.getUnsigned();
-        if (typeof unsigned[UNSIGNED_THREAD_ID_FIELD.name] === "string") {
-            return unsigned[UNSIGNED_THREAD_ID_FIELD.name];
+        } else {
+            return this.getThread()?.id || this.threadId;
         }
     }
 
--- a/src/models/event.ts
+++ b/src/models/event.ts
@@ -24,7 +24,13 @@ import { ExtensibleEvent, ExtensibleEvents, Optional } from "matrix-events-sdk";
 import type { IEventDecryptionResult } from "../@types/crypto";
 import { logger } from "../logger";
 import { VerificationRequest } from "../crypto/verification/request/VerificationRequest";
-import { EVENT_VISIBILITY_CHANGE_TYPE, EventType, MsgType, RelationType } from "../@types/event";
+import {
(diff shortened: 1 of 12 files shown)


src/models/event.ts
```diff
@@ -63,6 +63,7 @@ export interface IUnsigned {
     "transaction_id"?: string;
     "invite_room_state"?: StrippedState[];
     "m.relations"?: Record<RelationType | string, any>; // No common pattern for aggregated relations
+    "io.element.relation_thread_id"?: string;
```
gsouquet, line 66: I believe we should have the parent `relates_to` field copied into the children `unsigned`. That makes it less thread related, and better to write an MSC for IMO

```suggestion
    "io.element.parent.relates_to"?: {
    	rel_type: RelationType;
    	event_id: string;
    };
```

t3chguy: I'm not suggesting this be the shape the MSC uses, the MSC should probably follow https://spec.matrix.org/v1.7/client-server-api/#aggregations-of-child-events - this is just a workaround for the problem at hand

t3chguy: Updated to match https://github.com/matrix-org/matrix-spec-proposals/pull/4023

richvdh:
> If [MSC4023](https://github.com/matrix-org/matrix-spec-proposals/pull/4023) unsigned data is not included it'll use the /event/ API to fetch the parent event to determine what thread to associate the relation with, it'll then store that unsigned in the sync accumulator to prevent needing to do that work again.

According to the MSC, the `thread_id` is only included for reactions to threaded events... so we'll end up doing this for every reaction on the main thread? In which case, we might as well do it for threaded reactions too and the whole MSC is redundant.

Seems like something that needs changing in the MSC to me.

t3chguy:
> According to the MSC, the thread_id is only included for reactions to threaded events... so we'll end up doing this for every reaction on the main thread? In which case, we might as well do it for threaded reactions too and the whole MSC is redundant.

Good point, `"thread_id": null` or signalling support for the MSC could be used to catch that edge

src/models/room.ts
```diff
@@ -2141,9 +2148,19 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
             };
         }
 
-        // We've exhausted all scenarios, can safely assume that this event should live in the room timeline only
+        const unsigned = event.getUnsigned();
+        if (typeof unsigned["org.matrix.msc4023.thread_id"] === "string") {
```
gsouquet, line 2152: Are we still using the `UnstableValue` abstraction or scrapping it? I'm personally fine with this!

src/models/room.ts
```diff
@@ -2141,9 +2148,19 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+            return {
+                shouldLiveInRoom: false,
+                shouldLiveInThread: true,
+                threadId: unsigned["org.matrix.msc4023.thread_id"],
+            };
+        }
+
+        // We've exhausted all scenarios,
+        // we cannot assume that it lives in the main timeline as this may be a relation for an unknown thread
```
gsouquet, line 2161: Maybe adding a comment saying that it's better to not add the event rather thna adding it to an incorrect place as it can cause stuck notifs

gsouquet (approved):
Looks good to me! Believe this is a net improvement to what we have today

src/models/room.ts
t3chguy: Good call, forgot about it

richvdh:
Should the changelog entry not explain what exactly about this is a breaking change?

PR #3437, opened 2023-06-02
Fix edge cases around 2nd order relations and threads

t3chguy:
Mainly around missing reactions

### DIFF
--- a/src/models/event.ts
+++ b/src/models/event.ts
@@ -579,8 +579,16 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
         const relatesTo = this.getWireContent()?.["m.relates_to"];
         if (relatesTo?.rel_type === THREAD_RELATION_TYPE.name) {
             return relatesTo.event_id;
-        } else {
-            return this.getThread()?.id || this.threadId;
+        }
+        if (this.thread) {
+            return this.thread.id;
+        }
+        if (this.threadId != undefined) {
+            return this.threadId;
+        }
+        const unsigned = this.getUnsigned();
+        if (typeof unsigned[UNSIGNED_THREAD_ID_FIELD.name] === "string") {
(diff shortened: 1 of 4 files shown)


src/models/event.ts
```diff
@@ -579,9 +579,18 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
...
             return relatesTo.event_id;
-        } else {
-            return this.getThread()?.id || this.threadId;
         }
+        if (this.thread) {
+            return this.thread.id;
+        }
+        if (this.threadId != undefined) {
+            return this.threadId;
+        }
```
gsouquet, line 588: Probably not worth doing as part of this pull request, but `this.threadId` has a lot of overlap with the `unsigned.thread_id` field.
We should probably factor it out, and consolidate it under what MSC4023 proposes.

src/models/event.ts
```diff
@@ -579,9 +579,18 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
         const relatesTo = this.getWireContent()?.["m.relates_to"];
         if (relatesTo?.rel_type === THREAD_RELATION_TYPE.name) {
             return relatesTo.event_id;
-        } else {
-            return this.getThread()?.id || this.threadId;
         }
+        if (this.thread) {
+            return this.thread.id;
+        }
+        if (this.threadId != undefined) {
```
gsouquet, line 586: Why are we using `!=` over `!==` in this scenario?
If there's some special case, can we add a comment explaining it, and if not can we use `!==` instead?

gsouquet (commented):
Looks good to me 👍

src/models/event.ts
t3chguy: Fixes

gsouquet (approved):

PR #3439, opened 2023-06-02
Integration test for QR code verification

richvdh:
Followup to https://github.com/matrix-org/matrix-js-sdk/pull/3436: another integration test, this time using the QR code flow

---
This change is marked as an *internal change* (Task), so will not be included in the changelog.

PR #3441, opened 2023-06-02
Add new methods to VerificationRequest

richvdh:
I'm trying to pull out an interface from `VerificationRequest`, so that I can replace it with a Rust implementation without changes to the application (vector-im/crypto-internal#96).

Currently it exposes `channel` and `qrCodeData` which are entire objects of which only one or two fields are used. By adding more specific methods, we can reduce the amount of stuff that has to be implemented considerably.

Based on #3439 

---


gsouquet (approved):
Looks good to me

PR #3447, opened 2023-06-06
Export thread-related types from SDK

stas-demydiuk:
Export types related to threads from SDK so consumers who also use typescript can easily consume them

## Checklist

---

## 🚨 BREAKING CHANGES
 * Export thread-related types from SDK ([\#3447](https://github.com/matrix-org/matrix-js-sdk/pull/3447)). Contributed by @stas-demydiuk.

### DIFF
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -47,6 +47,11 @@ export type ThreadEventHandlerMap = {
     [ThreadEvent.Delete]: (thread: Thread) => void;
 } & EventTimelineSetHandlerMap;
 
+/**
+ * @deprecated please use ThreadEventHandlerMap instead
+ */
+export type EventHandlerMap = ThreadEventHandlerMap;
+
 interface IThreadOpts {
     room: Room;
     client: MatrixClient;
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -38,9 +38,9 @@ export enum ThreadEvent {
     Delete = "Thread.delete",
 }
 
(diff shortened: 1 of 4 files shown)


gsouquet (approved):

src/models/thread.ts
```diff
@@ -38,9 +38,9 @@ export enum ThreadEvent {
     Delete = "Thread.delete",
 }
 
-type EmittedEvents = Exclude<ThreadEvent, ThreadEvent.New> | RoomEvent.Timeline | RoomEvent.TimelineReset;
+export type ThreadEmittedEvents = Exclude<ThreadEvent, ThreadEvent.New> | RoomEvent.Timeline | RoomEvent.TimelineReset;
 
-export type EventHandlerMap = {
+export type ThreadEventHandlerMap = {
```
t3chguy, line 43: Changing this is a breaking change

t3chguy (changes requested):
Can we avoid breaking changes, maintain backwards compatibility please

src/models/thread.ts
stas-demydiuk: As I see it was never exported before, so it is just an internal type rename, or am I missing something?

t3chguy: It was exported, you could import it as `matrix-js-sdk/src/models/thread` - you just changed its name, if you prefer the new name, fine, but you must maintain the old name with a deprecation notice

t3chguy (approved):

PR #3448, opened 2023-06-06
Fix issues with implementation for MSC3981

justjanne:
Type: Defect
Uses: https://github.com/matrix-org/matrix-spec-proposals/pull/3981
Uses: https://github.com/matrix-org/matrix-spec-proposals/pull/4023
Related: https://github.com/matrix-org/matrix-js-sdk/pull/3248
Related: https://github.com/matrix-org/matrix-js-sdk/pull/3427

1. Fixes an issue where messages discovered via recursive relations would not be accurately reported as belonging to the thread via eventShouldLiveIn. This was fixed by setting the unsigned thread id field for thread timeline pagination calls if it's not yet set by the server.
3. Fixes an issue where reactions were not included in thread timelines. This was fixed by disabling the "only return m.thread events" filter on those calls if recursive relations is available.

---

## 🐛 Bug Fixes
 * Fix issues with implementation for MSC3981 ([\#3448](https://github.com/matrix-org/matrix-js-sdk/pull/3448)). Contributed by @justjanne.

### DIFF
--- a/spec/integ/matrix-client-event-timeline.spec.ts
+++ b/spec/integ/matrix-client-event-timeline.spec.ts
@@ -1046,16 +1046,20 @@ describe("MatrixClient event timelines", function () {
                 return request;
             }
 
-            function respondToThread(root: Partial<IEvent>, replies: Partial<IEvent>[], limit?: number): ExpectedHttpRequest {
+            function respondToThread(
+                root: Partial<IEvent>,
+                replies: Partial<IEvent>[],
+                limit?: number,
+            ): ExpectedHttpRequest {
                 const request = httpBackend.when(
                     "GET",
                     "/_matrix/client/v1/rooms/!foo%3Abar/relations/" +
-                    encodeURIComponent(root.event_id!) +
-                    buildRelationPaginationQuery(client, {
-                        dir: Direction.Backward,
-                        limit: limit,
-                        recurse: true,
-                    }),
+                        encodeURIComponent(root.event_id!) +
+                        buildRelationPaginationQuery(client, {
+                            dir: Direction.Backward,
+                            limit: limit,
+                            recurse: true,
+                        }),
                 );
                 request.respond(200, function () {
                     return {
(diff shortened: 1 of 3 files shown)


spec/integ/matrix-client-event-timeline.spec.ts
```diff
@@ -52,12 +53,16 @@ const withoutRoomId = (e: Partial<IEvent>): Partial<IEvent> => {
...
  * @param params query parameters
  */
-const buildRelationPaginationQuery = (params: QueryDict): string => {
+const buildRelationPaginationQuery = (client: MatrixClient, params: QueryDict): string => {
     if (Thread.hasServerSideFwdPaginationSupport === FeatureSupport.Experimental) {
         params = replaceParam("dir", "org.matrix.msc3715.dir", params);
     }
+    if (client.canSupport.get(Feature.RelationsRecursion) === ServerSupport.Unstable) {
+        params = replaceParam("recurse", "org.matrix.msc3981.recurse", params);
+    }
```
richvdh, line 65: this feels like a sledgehammer to crack a nut. Shouldn't the calling test know if we expect `recurse` or `org.matrix.msc3981.recurse`?

src/client.ts
```diff
@@ -5774,6 +5776,10 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
                     ...resOlder.chunk.map(mapper),
                 ];
                 for (const event of events) {
+                    event.setUnsigned({
+                        ...event.getUnsigned(),
+                        [UNSIGNED_THREAD_ID_FIELD.name]: timelineSet.thread.id,
```
richvdh, line 5781: isn't this overriding the response from the server? Why are we doing that?

spec/integ/matrix-client-event-timeline.spec.ts
```diff
@@ -1091,6 +1180,9 @@ describe("MatrixClient event timelines", function () {
                     event_id: THREAD_ROOT.event_id,
                 },
             },
+            unsigned: {
+                [UNSIGNED_THREAD_ID_FIELD.name]: THREAD_ROOT.event_id,
+            },
```
richvdh, line 1185: I'm somewhat concerned that by changing all these, we never test the fallback behaviour. Do we have any separate tests for the fallback? Do we actually need to change all these definitions?

richvdh (changes requested):

PR #3451, opened 2023-06-06
Fix export issue from thread module

EnricoSchw:
This was changed in prv PR and breaks the EC build. This fix it.

## Checklist

### DIFF
--- a/src/matrix.ts
+++ b/src/matrix.ts
@@ -37,7 +37,7 @@ export * from "./models/event-timeline-set";
 export * from "./models/poll";
 export * from "./models/room-member";
 export * from "./models/room-state";
-export { ThreadEvent, ThreadEmittedEvents, ThreadEventHandlerMap, Thread } from "./models/thread";
+export type { ThreadEvent, ThreadEmittedEvents, ThreadEventHandlerMap, Thread } from "./models/thread";
 export * from "./models/typed-event-emitter";
 export * from "./models/user";
 export * from "./models/device";


src/matrix.ts
```diff
@@ -37,7 +37,7 @@ export * from "./models/event-timeline-set";
 export * from "./models/poll";
 export * from "./models/room-member";
 export * from "./models/room-state";
-export { ThreadEvent, ThreadEmittedEvents, ThreadEventHandlerMap, Thread } from "./models/thread";
+export type { ThreadEvent, ThreadEmittedEvents, ThreadEventHandlerMap, Thread } from "./models/thread";
```
t3chguy, line 40: `Thread` is a class, it will want more than just a type export

src/matrix.ts
```diff
@@ -37,7 +37,8 @@ export * from "./models/event-timeline-set";
 export * from "./models/poll";
 export * from "./models/room-member";
 export * from "./models/room-state";
-export { ThreadEvent, ThreadEmittedEvents, ThreadEventHandlerMap, Thread } from "./models/thread";
+export { Thread } from "./models/thread";
+export type { ThreadEvent, ThreadEmittedEvents, ThreadEventHandlerMap } from "./models/thread";
```
gsouquet, line 41:
```suggestion
export * from "./models/thread";
```

The above seem to solve the warning and follows the pattern of what other exports do. I'd suggest moving to that so we can resolve the warning.

gsouquet (changes requested):

gsouquet (approved):

PR #3452, opened 2023-06-06
ElementR: Add `CryptoApi.getCrossSigningStatus`

florianduros:
## Checklist

- Need https://github.com/matrix-org/matrix-rust-sdk/pull/2020
- Add a new `CryptoApi.getCrossSigningStatus` API which is returning the current state of the cross signing for the current device

---

## ✨ Features
 * ElementR: Add `CryptoApi.getCrossSigningStatus` ([\#3452](https://github.com/matrix-org/matrix-js-sdk/pull/3452)). Contributed by @florianduros.


richvdh (changes requested):
looks good so far, a few suggestions!

src/@types/crypto.ts
```diff
@@ -97,3 +97,16 @@ export interface IOneTimeKey {
...
+ * The result of a call to {@link Crypto.getCrossSigningStatus}
+ */
+export interface ICrossSigningStatus {
+    publicKeyOnDevice: boolean;
+    privateKeysInSecretStorage: boolean;
+    privateKeysCachedLocally: {
+        masterKey: boolean;
+        selfSigningKey: boolean;
+        userSigningKey: boolean;
+    };
```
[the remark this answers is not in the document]
florianduros: Added some comments !

spec/integ/crypto/cross-signing.spec.ts
```diff
@@ -62,12 +62,9 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("cross-signing (%s)", (backend: s
...
 
     /**
-     * Create cross-signing keys, publish the keys
-     * Mock and bootstrap all the required steps
-     *
-     * @return the IAuthDict
+     * Mock the requests needed to set up cross signing
      */
-    async function setupCrossSigning(): Promise<IAuthDict> {
+    function mockCrossSigningRequest() {
```
richvdh, line 67:
```suggestion
    function mockSetupCrossSigningRequests(): void {
```

spec/integ/crypto/cross-signing.spec.ts
```diff
@@ -61,34 +61,54 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("cross-signing (%s)", (backend: s
...
-            );
+    /**
+     * Mock the requests needed to set up cross signing
+     */
+    function mockCrossSigningRequest() {
+        // have account_data requests return an empty object
+        fetchMock.get("express:/_matrix/client/r0/user/:userId/account_data/:type", {});
+
+        // we expect a request to upload signatures for our device ...
+        fetchMock.post({ url: "path:/_matrix/client/v3/keys/signatures/upload", name: "upload-sigs" }, {});
```
richvdh, line 72: I think we need to document these endpoint names, in the doc-comments for `mockCrossSigningRequest`. They form part of the interface of those functions, which the test is relying on.

src/crypto-api.ts
```diff
@@ -185,6 +185,17 @@ export interface CryptoApi {
      * @returns True if secret storage is ready to be used on this device
      */
     isSecretStorageReady(): Promise<boolean>;
+
+    /**
+     * Get the status of our cross-signing keys.
+     *
+     * - if there is a public key on this device
+     * - if the private master, userSigning and selfSigning keys are store in the SecretStorage
+     * - if the private master, userSigning and selfSigning keys are cached locally
```
richvdh, line 194: This feels a bit redundant? It's just repeating what's in `CrossSigningStatus`. Maybe we can instead include a summary in the `@returns` tag...

src/crypto-api.ts
```diff
@@ -185,6 +185,17 @@ export interface CryptoApi {
...
+     *
+     * @returns the current status of cross-signing keys
```
richvdh, line 196: Look at https://pr3452--js-sdk-docs-previews.netlify.app/stable/interfaces/crypto.cryptoapi#getCrossSigningStatus: this should be a proper sentence, with a capital letter and a full stop.

```suggestion
     * @returns The current status of cross-signing keys: whether we have public and private keys cached locally, and whether the private keys are in secret storage.
```

src/crypto-api.ts
```diff
@@ -263,3 +274,19 @@ export class DeviceVerificationStatus {
 }
 
 export * from "./crypto-api/verification";
+
+/**
+ * The result of a call to {@link CryptoApi.getCrossSigningStatus}
```
richvdh, line 279:
```suggestion
 * The result of a call to {@link CryptoApi.getCrossSigningStatus}.
```

src/crypto-api.ts
```diff
@@ -263,3 +274,19 @@ export class DeviceVerificationStatus {
...
+ */
+export interface CrossSigningStatus {
+    // True if the public master, self signing and user signing keys exist
```
richvdh, line 282: these all need to be doc-comments, with `/** */` delimiters. Otherwise they don't appear at https://pr3452--js-sdk-docs-previews.netlify.app/stable/interfaces/crypto.crosssigningstatus !

richvdh, line 282: exist... where? somewhere in the universe?

src/secret-storage.ts
```diff
@@ -672,3 +672,30 @@ export function trimTrailingEquals(input: string): string {
         return input;
     }
 }
+
+/**
+ * Check whether the private keys exist in secret storage.
```
richvdh, line 677: which private keys?

src/secret-storage.ts
```diff
@@ -672,3 +672,30 @@ export function trimTrailingEquals(input: string): string {
...
+ *
+ * @param secretStorage - The secret store using account data
+ * @returns map of key name to key info the secret is encrypted
+ *     with, or null if it is not present or not encrypted with a trusted
+ *     key
+ */
+export async function isStoredInSecretStorage(
```
richvdh, line 684: This needs a better name:

 * *What* is stored in secret storage?
 * It doesn't return a boolean, so we shouldn't call it `is...`.

Maybe `getCrossSigningKeysSecretStorageDetails` or something.

However...

given that all we need for the rust implementation here is a boolean, I would have left the legacy implementation alone and made a new function with a simpler interface.

richvdh, line 684: It's worth noting that the whole of the `secret-storage` module is public (https://pr3452--js-sdk-docs-previews.netlify.app/stable/modules/secretstorage) so putting this here exposes it as an external function to users of the js-sdk.

It's my opinion that the external API of the js-sdk is too cluttered as it is, and we should avoid exposing internal utility functions where we can. Can we find somewhere else for it?

spec/integ/crypto/cross-signing.spec.ts
```diff
@@ -61,34 +61,54 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("cross-signing (%s)", (backend: s
...
+            },
+            {},
+        );
+    }
+
+    /**
+     * Create cross-signing keys, publish the keys
+     * Mock and bootstrap all the required steps
+     *
+     * @return the IAuthDict - The parameters which are submitted as the `auth` dict in a UIA request
```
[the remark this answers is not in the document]
richvdh: Also, for consistency, prefer `@returns` to `@return`

richvdh (changes requested):

florianduros:
I'm adding new tests for the new `secret-storage` file.

spec/integ/crypto/cross-signing.spec.ts
richvdh: to be clear, by "endpoint names", I mean things like `upload-sigs`.

spec/integ/crypto/cross-signing.spec.ts
```diff
@@ -86,14 +90,10 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("cross-signing (%s)", (backend: s
      * Create cross-signing keys, publish the keys
      * Mock and bootstrap all the required steps
      *
-     * @return the IAuthDict - The parameters which are submitted as the `auth` dict in a UIA request
+     * @param authDict - The parameters which are submitted as the `auth` dict in a UIA request
```
richvdh, line 93:
```suggestion
     * @param authDict - The parameters to as the `auth` dict in the key upload request.
```

spec/integ/crypto/cross-signing.spec.ts
```diff
@@ -86,14 +90,10 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("cross-signing (%s)", (backend: s
...
      * @see https://spec.matrix.org/v1.6/client-server-api/#authentication-types
      */
-    async function setupCrossSigning(): Promise<IAuthDict> {
-        mockCrossSigningRequest();
-
-        // provide a UIA callback, so that the cross-signing keys are uploaded
-        const authDict = { type: "test" };
+    async function mockCrossSigningRequests(authDict: IAuthDict): Promise<void> {
```
richvdh, line 96: this no longer mocks anything.

```suggestion
    async function bootstrapCrossSigning(authDict: IAuthDict): Promise<void> {
```

src/rust-crypto/secret-storage.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ServerSideSecretStorage } from "../secret-storage";
+
+/**
+ * Check that the private cross signing keys (master, self signing, user signing) are stored into the secret storage and encrypted with the secret storage key.
+ *
+ * @param secretStorage - The secret store using account data
+ * @returns True if one of the secret storage master keys is shared with the secret storage user signing and self signing keys.
+ */
+export async function secretStorageContainsCrossSigningKeys(secretStorage: ServerSideSecretStorage): Promise<boolean> {
+    // Get the secret storage keys stored into the secret storage
```
richvdh, line 26:
```suggestion
    // Check if the master cross-signing key is stored in secret storage
```

src/rust-crypto/secret-storage.ts
```diff
@@ -0,0 +1,42 @@
...
+    const secretStorageMasterKeys = await secretStorage.isStored("m.cross_signing.master");
+
+    // Not stored keys
```
richvdh, line 29:
```suggestion
    // Master key not stored
```

src/rust-crypto/secret-storage.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (!secretStorageMasterKeys) return false;
+
+    // Get the user signing keys stored into the secret storage
+    const secretStorageUserSigningKeys = (await secretStorage.isStored(`m.cross_signing.user_signing`)) || {};
+    // Get the self signing keys stored into the secret storage
+    const secretStorageSelfSigningKeys = (await secretStorage.isStored(`m.cross_signing.self_signing`)) || {};
+
+    // Check that one of the secret storage master keys is shared with the secret storage user signing and self signing keys
```
richvdh, line 37: I think you're confusing cross-signing keys and secret-storage keys again.

To be clear:
 * We have a [master cross-signing key](https://spec.matrix.org/v1.7/client-server-api/#cross-signing). It is an ed25519 key, which means it is asymmetric (it is a key *pair*, with a public part and a private part), and it can be used for *signing* things but not encrypting them. It is used to sign the user-signing and self-signing cross-signing keys.
 * We do *not* have a master secret-storage key (or at least, we don't use that term). Instead, we just have [secret storage keys](https://spec.matrix.org/v1.7/client-server-api/#storage). There is some [fancy maths](https://spec.matrix.org/v1.7/client-server-api/#msecret_storagev1aes-hmac-sha2) but basically they are AES keys - so it is a *symmetric* key (there is no "public" part), and is used for *encryption*, not signing. They can nominally be used for encrypting any secret, but in this case we are using them to encrypt the private parts of the cross-signing keys.


```suggestion
    // Check that one of the secret storage keys used to encrypt the master key was also used to encrypt the user-signing and self-signing keys
```

src/rust-crypto/secret-storage.ts
```diff
@@ -0,0 +1,42 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { ServerSideSecretStorage } from "../secret-storage";
+
+/**
+ * Check that the private cross signing keys (master, self signing, user signing) are stored into the secret storage and encrypted with the secret storage key.
```
richvdh, line 20:
```suggestion
 * Check that the private cross signing keys (master, self signing, user signing) are stored in the secret storage and encrypted with the same secret storage key.
```

src/rust-crypto/secret-storage.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @param secretStorage - The secret store using account data
+ * @returns True if one of the secret storage master keys is shared with the secret storage user signing and self signing keys.
```
richvdh, line 23:
```suggestion
 * @returns True if the cross-signing keys are all stored and encrypted with the same secret storage key.
```

spec/unit/rust-crypto/secret-storage.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { secretStorageContainsCrossSigningKeys } from "../../../src/rust-crypto/secret-storage";
+import { ServerSideSecretStorage } from "../../../src/secret-storage";
+
+describe("secret-storage", () => {
+    describe("secretStorageContainsCrossSigningKeys", () => {
+        it("should return false when there is no secret storage master key", async () => {
```
richvdh, line 22:
```suggestion
        it("should return false when the master cross-signing key is not stored in secret storage", async () => {
```

spec/unit/rust-crypto/secret-storage.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+                    if (type === "m.cross_signing.master") return { secretStorageKey: {} };
+                    else return { secretStorageKey2: {} };
+                },
+            } as unknown as ServerSideSecretStorage;
+
+            const result = await secretStorageContainsCrossSigningKeys(secretStorage);
+            expect(result).toBeFalsy();
+        });
+
+        it("should return false when the secret storage master key is only shared by user signing the secret storage", async () => {
```
richvdh, line 44:
```suggestion
        it("should return false when the secret storage key for the master key is only shared by the user-signing key", async () => {
```

richvdh (changes requested):
we're definitely nearly there! A few more things to fix.

PR #3470, opened 2023-06-13
Remove mocking in crypto integ test

richvdh:
what happens?

---


richvdh:
> what happens?

Answer: it blows up, because on node<19, you have to jump through some hoops to get hold of a webcrypto. https://nodejs.org/docs/latest-v18.x/api/webcrypto.html#web-crypto-api: 

> Use require('node:crypto').webcrypto to access this module.

Currently my thinking is that we should make a functional `global.crypto` a requirement of running `initRustCrypto`, and leave it up to the application to arrange for such to exist.

PR #3479, opened 2023-06-15
Fix export type `GeneratedSecretStorageKey`

florianduros:
## Checklist


src/crypto/api.ts
```diff
@@ -19,7 +19,9 @@ import { IKeyBackupInfo } from "./keybackup";
 import { GeneratedSecretStorageKey } from "../crypto-api";
 
 /* re-exports for backwards compatibility. */
-export { CrossSigningKey, GeneratedSecretStorageKey as IRecoveryKey } from "../crypto-api";
+// CrossSigningKey is used as a value in `client.ts`, we can't export it as a type
```
t3chguy, line 22: It could also be used as a value by downstream consumers, not sure this comment adds value

t3chguy (approved):

src/crypto/api.ts
t3chguy: Happy either way, the lint rule will help out also

PR #3482, opened 2023-06-16
Use stricter types on login functions

kerryarchibald:
For https://github.com/vector-im/element-web/issues/21967

Pulling the loose string from strictification work in `AddThreepid` I found these untyped functions.
There is some distance between the spec and js-sdk's current implementation, for example the login type "m.login.saml2"

Breaking as far as types, no changes to logic.

With 

## Checklist

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -7798,7 +7798,6 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @returns Rejects: with an error response.
      */
     public loginWithPassword(user: string, password: string): Promise<LoginResponse> {
-        // TODO: Types
         return this.login("m.login.password", {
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -7770,40 +7771,33 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+                this.credentials = {
+                    userId: response.user_id,
+                };
+            }
+            return response;
+        });
     }
 
     /**
      * @returns Promise which resolves: TODO
```
alunturner, line 7797: Is the TODO relating to the type it returns? Not sure

alunturner (approved):
Just one comment, I think there's another TODO you can remove. Looks good though.

PR #3483, opened 2023-06-16
ElementR: Add `CryptoApi#bootstrapSecretStorage`

florianduros:
## Checklist

- Add `newBackendOnly` to run test only for the rust-crypto
- The new `bootstrapSecretStorage` implementation doesn't implement all the behavior of the old crypto. So, the tests are not expecting the same result for both crypto.

---

## 🦖 Deprecations
 * ElementR: Add `CryptoApi#bootstrapSecretStorage` ([\#3483](https://github.com/matrix-org/matrix-js-sdk/pull/3483)). Contributed by @florianduros.


src/rust-crypto/index.ts
```diff
@@ -30,12 +31,14 @@ import { ServerSideSecretStorage } from "../secret-storage";
  * @param userId - The local user's User ID.
  * @param deviceId - The local user's Device ID.
  * @param secretStorage - Interface to server-side secret storage.
+ * @param cryptoCallbacks - Crypto callbacks called during the lifetime of the e2e
```
richvdh, line 34:
```suggestion
 * @param cryptoCallbacks - Crypto callbacks provided by the application
```

src/rust-crypto/rust-crypto.ts
```diff
@@ -43,12 +43,13 @@ import {
...
 import { IDownloadKeyResult, IQueryKeysRequest } from "../client";
 import { Device, DeviceMap } from "../models/device";
-import { AddSecretStorageKeyOpts, ServerSideSecretStorage } from "../secret-storage";
+import { AddSecretStorageKeyOpts, SECRET_STORAGE_ALGORITHM_V1_AES, ServerSideSecretStorage } from "../secret-storage";
 import { CrossSigningIdentity } from "./CrossSigningIdentity";
 import { secretStorageContainsCrossSigningKeys } from "./secret-storage";
 import { keyFromPassphrase } from "../crypto/key_passphrase";
 import { encodeRecoveryKey } from "../crypto/recoverykey";
 import { crypto } from "../crypto/crypto";
+import { ICryptoCallbacks } from "../crypto";
```
richvdh, line 52: I'm trying to move stuff *out* of `src/crypto` rather than use it as-is. Basically I want to be able to delete the entire directory once ER is stable. Could you move `ICryptoCallbacks` into `src/crypto-api.ts` or somewhere like that?

We should probably move `keyFromPassphrase`, `encodeRecoveryKey` and the `crypto` shims at some point too, but that can happen separately.

src/rust-crypto/rust-crypto.ts
```diff
@@ -90,6 +91,9 @@ export class RustCrypto implements CryptoBackend {
 
         /** Interface to server-side secret storage */
         private readonly secretStorage: ServerSideSecretStorage,
+
+        /** Crypto callbacks called during the lifetime of the e2e */
```
richvdh, line 95:
```suggestion
        /** Crypto callbacks provided by the application */
```

src/crypto-api.ts
```diff
@@ -20,6 +20,7 @@ import { DeviceMap } from "./models/device";
 import { UIAuthCallback } from "./interactive-auth";
 import { AddSecretStorageKeyOpts } from "./secret-storage";
 import { VerificationRequest } from "./crypto-api/verification";
+import { ICreateSecretStorageOpts } from "./crypto/api";
```
richvdh, line 23: move it please, don't import from `./crypto`.

src/rust-crypto/rust-crypto.ts
```diff
@@ -428,6 +432,46 @@ export class RustCrypto implements CryptoBackend {
...
 
+    /**
+     * Implementation of {@link CryptoApi#bootstrapSecretStorage}
+     */
+    public async bootstrapSecretStorage({
+        createSecretStorageKey,
+        setupNewSecretStorage,
+    }: ICreateSecretStorageOpts = {}): Promise<void> {
+        // If setupNewSecretStorage is not set, we stop
+        if (!setupNewSecretStorage || !createSecretStorageKey) return;
```
richvdh, line 443: This is inconsistent with the documentation on `setupNewSecretStorage`:

> Reset even if keys already exist.

We shouldn't return unless:

 * we already have an AES key, *AND*
 * `setupNewSecretStorage` is false.

src/rust-crypto/rust-crypto.ts
```diff
@@ -428,6 +432,46 @@ export class RustCrypto implements CryptoBackend {
...
+
+        const secretStorageKeyTuple = await this.secretStorage.getKey();
```
richvdh, line 445:
```suggestion
        // See if we already have an AES secret-storage key.
        const secretStorageKeyTuple = await this.secretStorage.getKey();
```

src/rust-crypto/rust-crypto.ts
```diff
@@ -428,6 +432,46 @@ export class RustCrypto implements CryptoBackend {
...
+            // We stop
+            if (keyInfo.algorithm === SECRET_STORAGE_ALGORITHM_V1_AES) {
+                return;
+            }
+        }
+
+        const recoveryKey = await createSecretStorageKey();
+
+        // keyInfo is required to continue
+        if (!recoveryKey.keyInfo) return;
```
richvdh, line 460: IMHO would be better to raise an error here than just return silently.

richvdh (changes requested):

src/crypto-api.ts
```diff
@@ -233,14 +271,14 @@ export interface CryptoApi {
...
-     * - Do nothing if an AES key is already stored in the secret storage;
+     * - Do nothing if an AES key is already stored in the secret storage and `setupNewKeyBackup` is not set;
      * - Generate a new key {@link GeneratedSecretStorageKey} with `createSecretStorageKey`.
      * - Store this key in the secret storage and set it as the default key.
      * - Call `cryptoCallbacks.cacheSecretStorageKey` if provided.
      *
      * @param opts - Options object.
      */
-    bootstrapSecretStorage(opts: ICreateSecretStorageOpts): Promise<void>;
+    bootstrapSecretStorage(opts: CreateSecretStorageOpts): Promise<void>;
```
richvdh, line 281: please can you put this next to `isSecretStorageReady`

src/crypto-api.ts
```diff
@@ -40,6 +40,44 @@ export interface GeneratedSecretStorageKey {
     encodedPrivateKey?: string;
 }
 
+/**
+ * Parameter of {@link CryptoApi#bootstrapSecretStorage}
+ */
+export interface CreateSecretStorageOpts {
```
richvdh, line 46: please could you put this *after* CryptoApi: it is confusing when the main class in a big file is lost in the middle of the file.

src/rust-crypto/keybackup.ts
```diff
@@ -0,0 +1,42 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { ISigned } from "../@types/signed";
+
+export interface Curve25519AuthData {
```
richvdh, line 19: these public classes belong in `src/crypto-api/` (which is public), not `src/rust-crypto` (which is internal).

You can move the whole file into `src/crypto-api/` but please remember to add an `export *` to `src/crypto-api.ts`  so that they are properly exported as public identifiers from `matrix-js-sdk`.

richvdh (approved):
lgtm otherwise

Issue #1899, opened 2021-09-09
Ability to have a a logger per MatrixClient


clokep:
For a client that wishes to support multiple Matrix accounts it is useful to keep the logging separate for each account (for debugging purposes). Currently this is not possible since the matrix-js-sdk uses a single global logger.

I think a reasonable implementation of this would be for the `createClient` call to have a `logger` parameter, if not provided it falls back to a global logger.

I had briefly mentioned this as part of #1059, but that was closed. This would be particularly useful for Thunderbird (which does support multiple accounts).

A workaround for this could be to load the matrix-js-sdk multiple times into different scopes, but that feels icky.

This is filed with my community hat on. 🎩 

CC @freaktechnik

clokep:
It looks like #1435 might need fixing too for full advantage of this.

skylord123:
I would love for this to be added. I maintain node-red-contrib-matrix-chat that lets you build chat bots using simple functional block programming via Node-RED.

Currently it just logs to the console and is quite messy with multiple accounts. Most node-red log messages are prefixed with information about what node the message came from, but not having this feature makes me break that standard. This implementation would also allow me to turn debug messages on/off per matrix-client making my debugging much easier while running multiple clients.

clokep:
I was poking at this the other day (with a community hat on) and would be curious to get validation on a plan of attack. I was thinking something along the lines of:

1. Add a `ClientOpt` named `logger` of a type (I think `PrefixedLogger`, but maybe something that doesn't depend on an external dependency).
2. Store this as a public, readonly property on `MatrixClient`, falling back to the current global `logger` if it doesn't exist on `ClientOpts`.
3. Start using it everywhere:
	1. Quite a few places have reference to `MatrixClient` so these are mostly a tedious `s/logger/this.client.logger/g`.
	2. A few other places don't have easy access to this:
		1. If the class/function is private, add a logger parameter/property and start using it.
		2. If the class/function is public, add a logger parameter/property, falling back to the global logger if not provided, and start using it.
4. There's also a few odd `debuglog(...)` functions which would need to be replaced somehow.

I think the above doesn't need to be all done in a single massive PR, I'd suggest:

1. Steps 1 & 2 and a few usages (maybe inside of `MatrixClient` itself) as an initial PR.
2. Then updating things on a module level (or whatever makes sense and keeps PRs a reasonable size).

----

Any thoughts on if this is a reasonable plan which maintains backwards compatibility?

PR #3484, opened 2023-06-16
Support MatrixClient-specific loggers.

clokep:
This adds support to have a separate `PrefixedLogger` per `MatrixClient` instance. If one is not provided then it defaults back to the global `logger` instance.

This is wanted for multi-account clients, such as Thunderbird. See [bug 1837495](https://bugzilla.mozilla.org/show_bug.cgi?id=1837495), or [Phabricator D181226](https://phabricator.services.mozilla.com/D181226) for how this would be used in a downstream library.

This is initial work towards #1899.

This is done with a community hat on. 🎩 

CC @freaktechnik 

## Checklist

Support separate loggers for each `MatrixClient` instance.

---

## ✨ Features
 * Support MatrixClient-specific loggers. ([\#3484](https://github.com/matrix-org/matrix-js-sdk/pull/3484)).


src/client.ts
```diff
@@ -387,6 +387,8 @@ export interface ICreateClientOpts {
      * Default: false.
      */
     isVoipWithNoMediaAllowed?: boolean;
+
+    logger?: PrefixedLogger;
```
clokep, line 391: I suppose this needs documentation, I'm not sure if it also needs tests?

clokep: It might need tests just to pass the quality gate, I'm not really sure what to test here though? Have two matrix clients, log to them and ensure they go to the correct logger? Seems almost like testing the JavaScript engine itself though...

src/client.ts
```diff
@@ -381,6 +381,12 @@ export interface ICreateClientOpts {
      * Default: false.
      */
     isVoipWithNoMediaAllowed?: boolean;
+
+    /**
+     * A logger to associate with this MatrixClient.
+     * Defaults to the build-in global logger.
+     */
+    logger?: PrefixedLogger;
```
clokep, line 389: It is slightly dubious to use `PrefixesLogger` here since it inherits from a class outside of our control, we could use something custom if we wanted to strictly define / document the interface that is required?

fnwbr: > Yeah. It might be better to just accept a `logPrefix: string`, and make MatrixClient responsible for initialising the `PrefixedLogger`?

Personally I’d actually like more flexibility than just a custom prefix. I’d also like to disable logs altogether or configure the logger to only output certain levels.

richvdh: > Personally I’d actually like more flexibility than just a custom prefix. I’d also like to disable logs altogether or configure the logger to only output certain levels.

I totally agree with that, though I think it's orthogonal to the changes proposed here.

One idea that I had is to wire up the logging to [`debug`](https://www.npmjs.com/package/debug), and then the logging can be controlled via environment variables.

freaktechnik: The goal of the patch for "us" (TB) is to be able to log with a totally different logger. So we can control where and what levels get logged.

richvdh: > The goal of the patch for "us" (TB) is to be able to log with a totally different logger. So we can control where and what levels get logged.

Yes, fair. https://phabricator.services.mozilla.com/D181226 is instructive here.

Ok fine, I'm convinced, let's leave it as `PrefixedLogger`!

fnwbr: As long as we can access the whole logger object (rather than just setting a string) I’m all fine with a very strict interface. 🙃

src/client.ts
clokep: That makes sense, we should ensure this doesn't break. 👍

src/client.ts
clokep: > Sorry, I deleted my comment again because I think we could do with a slightly tighter interface, too, it'd just not be cleanly in "userspace" anymore.

It feels fairly dirty to me to re-use `PrefixedLogger` -- for one the Thunderbid patch doesn't even implement it as a prefix, but uses it as a suffix:

```javascript
  withPrefix(prefix) {
    return new MatrixLogger(this.prefix + prefix, this._account);
  },
```

@freaktechnik What was your thought on a tighter interface? This would also probably be desirable since it would allow changing the underlying logging library sued by matrix-js-sdk *without* being a breaking change?

freaktechnik: I'm not too familiar with `loglevel`, but something like having a function that takes a prefix, a level and then the stuff to log might be more flexible (`(string prefix, LogLevel level, ...logArgs) => void`)? That said, if the logger should expose all a JS console can do, that is not a good interface at all. Then having more of a factory style function might be preferable, where you specify prefix and get a console to log with (`(prefix: string) => Console`). In general I'd think that the library shouldn't expect to control its own log level when an external logger is used (which from what I can tell is what loglevel provides over a plain JS console).

It might also be better to call it module, context or similar instead of prefix, since as observed, not all loggers might log it before the rest of the information, and instead use it for other purposes altogether.

fnwbr: I think I can understand where the apathy for something losely defined is coming from. But in my opinion either defining a custom (tight) interface or just allowing log levels to be set via a method are perfectly fine here.
Log levels are kind of standardized as per https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.1 (bottom of 6.2.1)...

The most popular logging libraries (like Pino or Winston) are making use of that as well: https://www.npmjs.com/search?q=keywords:logger -- alternatively we can also just go with one of those and be done with this.
Or yeah, go ahead and use [loglevel](https://www.npmjs.com/package/loglevel).

fnwbr:
@justjanne @weeman1337 Can we get some official eyes on this? 🙃

weeman1337:
> @justjanne @weeman1337 Can we get some official eyes on this? upside_down_face

You have already got Rich's eyes :mag: Can you please have a look at this comments?

fnwbr:
Ah, my bad. I read the following as "it has to be either Janne or Michael":
image

weeman1337:
> Ah, my bad. I read the following as "it has to be either Janne or Michael": image

No problem. Any two reviewers are assigned to a non-draft PR. But still, anyone else can review it. If you see `blabla reviewed on behalf of matrix-org/element-web` it means that there is a reviewer. You only need an approval from one reviewer.

richvdh:
@clokep I have taken the work you did in this PR, with many thanks, and used it as a basis for #3792.

richvdh:
closing this in favour of #3792, then

clokep:
> @clokep I have taken the work you did in this PR, with many thanks, and used it as a basis for #3792.

Thanks for picking this up @richvdh! 🎉

richvdh:
It was largely selfish! Thanks for setting the ball rolling!

PR #3486, opened 2023-06-19
Fix an instance of failed to decrypt error when an in flight `/keys/query` fails.

erikjohnston:
Specifically, when checking the event sender matches who sent us the session keys we skip waiting for pending device list updates if we already know who owns the session key.

This mitigates some of https://github.com/vector-im/element-web/issues/24682.

I've also stuck an try/catch across the `downloadKeys(..)` call, on the assumption that it would make it easier to track this failure mode down. I've completely made up the new `OLM_BAD_SENDER_CHECK_FAILED` error code.

Ideally, I think instead of doing it like this we'd instead specifically track who sent us the session keys (or at least, who we believe session keys should be from), and match against those when we do the decryption. But that's a bigger job and I think this should help a bunch while this gets fixed properly.

---

## 🐛 Bug Fixes
 * Fix an instance of failed to decrypt error when an in flight `/keys/query` fails. ([\#3486](https://github.com/matrix-org/matrix-js-sdk/pull/3486)).

### DIFF
--- a/src/crypto/algorithms/olm.ts
+++ b/src/crypto/algorithms/olm.ts
@@ -194,9 +194,9 @@ class OlmDecryption extends DecryptionAlgorithm {
 
         // check that the device that encrypted the event belongs to the user that the event claims it's from.
         //
-        // To do this, we need to make sure that our device list is up-to-date. If the device is unknown, we can only
-        // assume that the device logged out and accept it anyway. Some event handlers, such as secret sharing, may be
-        // more strict and reject events that come from unknown devices.
+        // If the device is unknown then we check that we're not in the process of fetching the keys. If after that the
+        // device is still unknown, then we can only assume that the device logged out and accept it anyway. Some event
+        // handlers, such as secret sharing, may be more strict and reject events that come from unknown devices.
         //
         // This is a defence against the following scenario:
         //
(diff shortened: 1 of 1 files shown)


src/crypto/algorithms/olm.ts
```diff
@@ -209,8 +209,21 @@ class OlmDecryption extends DecryptionAlgorithm {
         //
         // In practice, it's not clear that the js-sdk would behave that way, so this may be only a defence in depth.
 
-        await this.crypto.deviceList.downloadKeys([event.getSender()!], false);
-        const senderKeyUser = this.crypto.deviceList.getUserByIdentityKey(olmlib.OLM_ALGORITHM, deviceKey);
+        let senderKeyUser = this.crypto.deviceList.getUserByIdentityKey(olmlib.OLM_ALGORITHM, deviceKey);
+        if (senderKeyUser == undefined) {
+            // Await for any ongoing key query fetches for the user before trying the lookup again.
```
richvdh, line 214:
```suggestion
            // Wait for any pending key query fetches for the user to complete before trying the lookup again.
```
because English is awful

src/crypto/algorithms/olm.ts
```diff
@@ -209,8 +209,21 @@ class OlmDecryption extends DecryptionAlgorithm {
...
+                throw new DecryptionError("OLM_BAD_SENDER_CHECK_FAILED", "Could not verify sender identity", {
+                    sender: deviceKey,
+                    err: e as Error,
+                });
+            }
+
+            senderKeyUser = this.crypto.deviceList.getUserByIdentityKey(olmlib.OLM_ALGORITHM, deviceKey);
+        }
+
         if (senderKeyUser !== event.getSender() && senderKeyUser != undefined) {
```
erikjohnston, line 227:
```suggestion
        if (senderKeyUser !== event.getSender() && senderKeyUser !== undefined) {
```

@richvdh should we also do this ^?

richvdh: probably

erikjohnston: Oh, hmm. That `senderKeyUser` can be null? Is that why we're not using strict equalities?

BillCarsonFr (changes requested):

src/crypto/algorithms/olm.ts
```diff
@@ -209,8 +209,21 @@ class OlmDecryption extends DecryptionAlgorithm {
...
+            try {
+                await this.crypto.deviceList.downloadKeys([event.getSender()!], false);
+            } catch (e) {
+                throw new DecryptionError("OLM_BAD_SENDER_CHECK_FAILED", "Could not verify sender identity", {
+                    sender: deviceKey,
+                    err: e as Error,
+                });
+            }
+
+            senderKeyUser = this.crypto.deviceList.getUserByIdentityKey(olmlib.OLM_ALGORITHM, deviceKey);
```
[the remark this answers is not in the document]
richvdh: > The only way to do that is to download the device (/keys/query),

If we have previously stored a signed copy of the device, surely there is no need to download it again?

> Regarding the solution: We need for m.room.keys to delay these checks, so not download anything at all.

Yes, that's the ideal solution, but that's really not the point here: this PR is trying to make an incremental improvement to the existing implementation.

There are two proposed improvements here:
 * Erik's change as it currently stands: if we already have a signed copy of the device, there is no need to wait for any ongoing `/keys/query` requests to complete. Can we agree this is uncontroversial?
 * My later comment: *given* that - in this code - we treat an unknown device identically to a correctly-signed device, I don't really see the value in doing a `/keys/query` request anyway. Supposing the incoming message is from a bogus device, we're not going to get any extra information by requesting the device list from the claimed user: the bogus device isn't going to be in the list, so it is treated as "unknown".

This code has been observed to cause significant numbers of UISIs. I don't believe the current implementation offers any value, and Erik is proposing a quick win here. I think it would be a mistake to refuse it because it's not the ideal solution.

richvdh (approved):
I have spoken to @BillCarsonFr and we think it's fine to merge this as-is. Per my earlier comments, I think we could go even further, but in the interests of expediency maybe we should just land it.

PR #3487, opened 2023-06-19
Element-R: wait for OlmMachine on startup

richvdh:
Previously, if you called `CryptoApi.getUserDeviceInfo()` before the first `/sync` request happened, it would return an empty list, which made a bunch of the tests racy. Add a hack to get the OlmMachine to think about its device lists during startup.

---


richvdh:
I think github flaked out. It was a very weird failure.

richvdh:
your wish is my command

t3chguy (approved):
:shipit: thanks for the test

PR #3492, opened 2023-06-20
Element-R: speed up slow unit test

richvdh:
A couple of tests were waiting for a request that wasn't happening, so timing out after 1.5 seconds. Let's avoid the extra slowth.

(This was introduced by changes in https://github.com/matrix-org/matrix-js-sdk/pull/3487, but the changes in this PR do no harm anyway)

---


kerryarchibald (approved):
lgtm

PR #3495, opened 2023-06-20
Improve intetration test for interactive verification

richvdh:
* test more things
* make it better match the spec

Suggest reviewing commit-by-commit, and turning off whitespace diffs.

### DIFF
--- a/spec/integ/crypto/verification.spec.ts
+++ b/spec/integ/crypto/verification.spec.ts
@@ -376,6 +376,64 @@ function runTests(backend: string, initCrypto: InitCrypto, methods: string[] | u
             await verificationPromise;
             expect(request.phase).toEqual(VerificationPhase.Done);
         });
+
+        oldBackendOnly("can cancel during the SAS phase", async () => {
+            // have alice initiate a verification. She should send a m.key.verification.request
+            const [, request] = await Promise.all([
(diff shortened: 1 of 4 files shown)


spec/integ/crypto/verification.spec.ts
```diff
@@ -104,180 +108,166 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
         });
 
         await initCrypto(aliceClient);
+
+        e2eKeyResponder = new E2EKeyResponder(aliceClient.getHomeserverUrl());
+        syncResponder = new SyncResponder(aliceClient.getHomeserverUrl());
+        mockInitialApiRequests(aliceClient.getHomeserverUrl());
+        aliceClient.startClient();
```
weeman1337, line 127: Isn't `startClient` `async` and shout be awaited here?

weeman1337 (commented):
Looks good overall. Thanks for digging into it and documenting it :+1: 

Found one potential bit that may introduce flakyness.

spec/integ/crypto/verification.spec.ts
richvdh: good spot!

```suggestion
        await aliceClient.startClient();
```

weeman1337 (approved):

PR #3496, opened 2023-06-21
Aggregate relations regardless of whether event fits into the timeline

t3chguy:
Relations are not timeline specific

Fixes https://github.com/vector-im/element-web/issues/25596

---

## 🐛 Bug Fixes
 * Aggregate relations regardless of whether event fits into the timeline ([\#3496](https://github.com/matrix-org/matrix-js-sdk/pull/3496)). Fixes vector-im/element-web#25596.

### DIFF
--- a/spec/unit/event-timeline-set.spec.ts
+++ b/spec/unit/event-timeline-set.spec.ts
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import { mocked } from "jest-mock";
+
 import * as utils from "../test-utils/test-utils";
 import {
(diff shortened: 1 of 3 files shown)


alunturner (approved):

Johennes:
@t3chguy out of curiosity, what causes us to not know which timeline a reaction belongs into?

t3chguy:
@Johennes as you backpaginate (`/messages` API); if the reaction comes not within the same page of events as the event to which it is replying you do not know which timeline the parent event of the reaction belongs to as the reaction only belongs to the parent event by ID. You could hit `/event` API for each and load it that way but that would slow down pagination.

PR #3501, opened 2023-06-23
Bump version of the react-sdk cypress workflow file

richvdh:
`cypress.yaml` is currently pinned to an old version of the react-sdk, meaning that each attempt to run it is currently failing with an error.

(See https://github.com/matrix-org/matrix-js-sdk/actions/runs/5355422916:

> `The workflow is not valid. .github/workflows/cypress.yml (Line: 27, Col: 28): Invalid secret, TCMS_USERNAME is not defined in the referenced workflow. .github/workflows/cypress.yml (Line: 28, Col: 28): Invalid secret, TCMS_PASSWORD is not defined in the referenced workflow.`

)

(Introduced by https://github.com/matrix-org/matrix-js-sdk/pull/3480)

---

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -15,7 +15,7 @@ concurrency:
 jobs:
     cypress:
         name: Cypress
-        uses: matrix-org/matrix-react-sdk/.github/workflows/cypress.yaml@v3.73.1
+        uses: matrix-org/matrix-react-sdk/.github/workflows/cypress.yaml@v3.74.0
         permissions:
             actions: read
             issues: read


t3chguy:
Looks like this is handled as part of https://github.com/matrix-org/matrix-js-sdk/pull/3467

germain-gg (approved):

richvdh:
Gah. this didn't fix the problem, because the change needed in the react-sdk (https://github.com/matrix-org/matrix-react-sdk/pull/11069) didn't actually land in v3.74.0.

PR #3490, opened 2023-06-19
Element-R: Basic implementation of SAS verification

richvdh:
Fixes https://github.com/vector-im/element-web/issues/25316

~~Based on https://github.com/matrix-org/matrix-js-sdk/pull/3495~~

Notes: Support for interactive device verification in Element-R.

---
Here's what your changelog entry will look like:

## ✨ Features
 * Support for interactive device verification in Element-R. ([\#3490](https://github.com/matrix-org/matrix-js-sdk/pull/3490)). Fixes vector-im/element-web#25316.

PR #3505, opened 2023-06-23
Element-R: implement cancellation of verification requests

richvdh:
Based on #3504

The final part of https://github.com/vector-im/element-web/issues/25319

Notes: Support for interactive device verification in Element-R.

---

## ✨ Features
 * Support for interactive device verification in Element-R. ([\#3505](https://github.com/matrix-org/matrix-js-sdk/pull/3505)).

### DIFF
--- a/spec/integ/crypto/verification.spec.ts
+++ b/spec/integ/crypto/verification.spec.ts
@@ -290,6 +290,10 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
             await verificationPromise;
             expect(request.phase).toEqual(VerificationPhase.Done);
 
+            // at this point, cancelling should do nothing.
+            await request.cancel();
+            expect(request.phase).toEqual(VerificationPhase.Done);
+
             // we're done with the temporary keypair
             olmSAS.free();
         });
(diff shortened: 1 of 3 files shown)


src/rust-crypto/verification.ts
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    /** once the phase is Started (and !initiatedByMe) or Ready: common methods supported by both sides */
+    public get methods(): string[] {
+        throw new Error("not implemented");
+    }
+
+    /** the method picked in the .start event */
+    public get chosenMethod(): string | null {
+        const verification: RustSdkCryptoJs.Qr | RustSdkCryptoJs.Sas | undefined = this.inner.getVerification();
```
alunturner, line 167: The `RustSdkCryptos.Qr | RustSdkCryptoJs.Sas | undefined` union appears a couple of times in this class. Do you think it would warrant becoming a type?

src/rust-crypto/verification.ts
```diff
@@ -0,0 +1,429 @@
...
+            }
+
+            this.callbacks = {
+                sas: {
+                    decimal: decimal,
+                    emoji: emoji?.map((e) => [e.symbol, e.description]),
+                },
+                confirm: async (): Promise<void> => {
+                    const requests: Array<OutgoingRequest> = await this.inner.confirm();
+                    for (const m of requests) {
```
alunturner, line 328: Maybe a bone question. Why `m`?

spec/integ/crypto/verification.spec.ts
```diff
@@ -206,15 +206,21 @@ function runTests(backend: string, initCrypto: InitCrypto, methods: string[] | u
...
-            // there should now be a verifier
-            const verifier: Verifier = request.verifier!;
-            expect(verifier).toBeDefined();
-            expect(verifier.getShowSasCallbacks()).toBeNull();
+            // as soon as the Changed event arrives, `verifier` should be defined
+            const verifier = await new Promise<Verifier>((resolve) => {
+                function onChange() {
+                    expect(request.phase).toEqual(VerificationPhase.Started);
+                    expect(request.otherPartySupportsMethod("m.sas.v1")).toBe(true);
+                    expect(request.chosenMethod).toEqual("m.sas.v1");
```
alunturner, line 204: Should this be `toBe`?

spec/integ/crypto/verification.spec.ts
```diff
@@ -156,6 +156,12 @@ function runTests(backend: string, initCrypto: InitCrypto, methods: string[] | u
         it("can verify another device via SAS", async () => {
             await waitForDeviceList();
 
+            // initially there should be no verifications in progress
+            {
+                const requests = aliceClient.getCrypto()!.getVerificationRequestsToDeviceInProgress(TEST_USER_ID);
+                expect(requests.length).toEqual(0);
```
alunturner, line 157: There is a `toHaveLength` matcher available if you prefer how that reads

alunturner (approved):
All comments are small ones, nothing blocking.

src/rust-crypto/verification.ts
richvdh: erm, good question. m for message, maybe? Though again, #3490 rather than new in this PR.

spec/integ/crypto/verification.spec.ts
richvdh: Possibly? I'm not sure it makes any difference, or if there is any particluar convention here.

alunturner: I was only asking as using `toEqual` suggests to me some sort of object comparison is going on (comparing keys and values I mean) whereas `toBe` suggests comparison of primitives. It makes no difference to the output here.

richvdh: Well, we don't normally write `method.is("m.sas.v1")` for doing string comparison, so I'm not convinced that `toBe` is, in fact, more natural for string comparison.

Also, I come from a Python world, where you definitely need `assertEqual`, and not `assertIs`:

```
>>> a = "foo"
>>> a += "bar"
>>> b = "foobar"
>>> a == b
True
>>> a is b
False
```

PR #3507, opened 2023-06-26
Fix fetchRoomThreadList emitting spurious Room.Timeline events

t3chguy:
---

## 🐛 Bug Fixes
 * Fix fetchRoomThreadList emitting spurious Room.Timeline events ([\#3507](https://github.com/matrix-org/matrix-js-sdk/pull/3507)).

### DIFF
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -2054,7 +2054,9 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         for (const rootEvent of matrixEvents) {
             timelineSet.addLiveEvent(rootEvent, {
                 duplicateStrategy: DuplicateStrategy.Replace,
-                fromCache: false,
+                // We say these events are from cache as we're abusing a single timeline to maintain events
+                // from multiple threads and as such do not want to emit Room.Timeline events for these events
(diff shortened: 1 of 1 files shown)


src/models/room.ts
```diff
@@ -2054,7 +2054,9 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         for (const rootEvent of matrixEvents) {
             timelineSet.addLiveEvent(rootEvent, {
                 duplicateStrategy: DuplicateStrategy.Replace,
-                fromCache: false,
+                // We say these events are from cache as we're abusing a single timeline to maintain events
+                // from multiple threads and as such do not want to emit Room.Timeline events for these events
+                fromCache: true,
```
germain-gg, line 2059: What is the impact if we do end up emitting a `Room.Timeline` event?
Reading through this I fail to get that context. If the conscequences are bad enough, we probably should add a test for it too

germain-gg (changes requested):

src/models/room.ts
t3chguy: Can do this in a different manner in Notifier itself, will move

PR #3517, opened 2023-06-27
Stop running jest in parallel in integ tests

richvdh:
... it makes the output impossible to understand.

---

### DIFF
--- a/spec/integ/matrix-client-syncing.spec.ts
+++ b/spec/integ/matrix-client-syncing.spec.ts
@@ -1990,5 +1990,8 @@ describe("MatrixClient syncing (IndexedDB version)", () => {
         idbHttpBackend.verifyNoOutstandingExpectation();
         await idbClient.stopClient();
         await idbHttpBackend.stop();
+
+        // force a save to indexeddb to make sure that we don't have any saves in flight when the test completes.
+        await idbClient.store.save(true);
     });
 });
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,8 +36,12 @@ jobs:
               if: matrix.specs == 'browserify'
               run: "yarn build"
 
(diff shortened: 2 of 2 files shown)


spec/integ/matrix-client-syncing.spec.ts
```diff
@@ -1990,5 +1990,8 @@ describe("MatrixClient syncing (IndexedDB version)", () => {
         idbHttpBackend.verifyNoOutstandingExpectation();
         await idbClient.stopClient();
         await idbHttpBackend.stop();
+
+        // force a save to indexeddb to make sure that we don't have any saves in flight when the test completes.
+        await idbClient.store.save(true);
```
richvdh, line 1995: once we limit jest to one worker, it realises that something is logging after shutdown, and complains about it. This makes sure that the test cleans up properly.

t3chguy (changes requested):
This goes directly against one of the goals of the Web Weekly [Reduce time required to run tests](https://github.com/vector-im/element-web/issues/21971) so needs discussion there

richvdh:
> This goes directly against one of the goals of the Web Weekly [Reduce time required to run tests](https://github.com/vector-im/element-web/issues/21971) so needs discussion there

I don't think it makes any difference. Compare https://github.com/matrix-org/matrix-js-sdk/actions/runs/5389744563/jobs/9784333178?pr=3517 *with* this PR (1m53s) with https://github.com/matrix-org/matrix-js-sdk/actions/runs/5389733941/jobs/9784477473?pr=3521 without it (1m49s)

t3chguy:
That's a 3.7% increase in time, which whilst we have few tests is negligible, but if we grow our number of tests we'd end up wanting to revert this, I don't think its the right approach personally. Happy to be overriden. -0 for me

richvdh:
closing this for now due to general lack of consensus. I may reopen in future if it continues to grind my gears

PR #3523, opened 2023-06-27
Use the right smiley in emoji verification

richvdh:
Notes: fix a bug which caused the wrong emoji to be shown during SAS device verification.

The [spec](https://spec.matrix.org/v1.7/client-server-api/#sas-method-emoji) mandates U+1F600 ("Grinning Face"), not U+1F642 "Slightly Smiling Face".

---

## 🐛 Bug Fixes
 * fix a bug which caused the wrong emoji to be shown during SAS device verification. ([\#3523](https://github.com/matrix-org/matrix-js-sdk/pull/3523)).


florianduros (approved):
Nice catch!

PR #3528, opened 2023-06-28
Element-R: support for starting a SAS verification

richvdh:
Once we have a verification request in flight, we can start a SAS flow by sending an `m.key.verification.start`. This is slightly fiddly because at the moment the js-sdk breaks this into two steps: first do a `.beginKeyVerification` on the `VerificationRequest` to immediately get a `Verifier`, and then call `.verify` on the `Verifier` to send the `start` message.

However, in the Rust world, creating the equivalent to a verifier is asynchronous and sends the `start` message.

So, we create a new method which does both, and implement it in legacy crypto for forward compatibility.

Notes: Deprecate `VerificationRequest.beginKeyVerification()` in favour of `VerificationRequest.startVerification()`.

Part of https://github.com/vector-im/element-web/issues/25320.

---

## 🦖 Deprecations
 * Deprecate `VerificationRequest.beginKeyVerification()` in favour of `VerificationRequest.startVerification()`. ([\#3528](https://github.com/matrix-org/matrix-js-sdk/pull/3528)).

### DIFF
--- a/src/crypto/verification/request/VerificationRequest.ts
+++ b/src/crypto/verification/request/VerificationRequest.ts
@@ -30,6 +30,7 @@ import {
     VerificationRequest as IVerificationRequest,
     VerificationRequestEvent,
     VerificationRequestEventHandlerMap,
+    Verifier,
 } from "../../../crypto-api/verification";
 
 // backwards-compatibility exports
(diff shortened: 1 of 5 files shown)


src/crypto/verification/request/VerificationRequest.ts
```diff
@@ -458,6 +459,13 @@ export class VerificationRequest<C extends IVerificationChannel = IVerificationC
         return this._verifier!;
     }
 
+    public async startVerification(method: string): Promise<Verifier> {
```
germain-gg, line 462: Is it desirable to have a list of supported methods here, rather than using `string` as a type?

germain-gg (approved):
Otherwise looks good to me!

src/crypto/verification/request/VerificationRequest.ts
richvdh: possibly, though we do have to support `string`s on all the other methods for backwards compatibility, so it's not an obvious win.

Issue #3463, opened 2023-06-11
MaxListenersExceededWarning on threads; many redundant fetches of thread root event


davidisaaclee:
Using an account that has some threads, I get the following warning around the first sync:

```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 Thread.update listeners added. Use emitter.setMaxListeners() to increase limit
```

<details>

<summary>Stack trace</summary>

```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 Thread.update listeners added. Use emitter.setMaxListeners() to increase limit
    _addListener events.js:211
    addListener events.js:227
    on typed-event-emitter.ts:133
    reEmit ReEmitter.ts:57
    reEmit ReEmitter.ts:84
    setThread event.ts:1618
    mapper event-mapper.ts:74
    fetchRootEvent thread.ts:148
    updateThreadMetadata thread.ts:482
    addEvents thread.ts:291
    createThread room.ts:2277
    addThreadedEvents room.ts:2183
    addLiveEvents room.ts:2843
    addLiveEvents room.ts:2842
    injectRoomEvents sync.ts:1779
    processSyncResponse sync.ts:1437
    promiseMapSeries utils.ts:425
    processSyncResponse sync.ts:1293
    doSync sync.ts:915
    sync sync.ts:798
    startClient client.ts:1513
    initializeMatrixConnection App.tsx:39
    App App.tsx:46
```
</details>

I haven't encountered any user-impacting issues around this, but it sounds like this could cause `MatrixEvent`s to fail to emit `Thread.update` to certain listeners.

**More importantly,** I believe this is caused by a bug in `Thread#fetchRootEvent` where the remote root event is fetched on every call to `updateThreadMetadata` (e.g. when adding an event to the thread, as will be done for each thread event received during a sync):

https://github.com/matrix-org/matrix-js-sdk/blob/dfb079a76fcdf2c218e8b6ba08d4d8f5f85fa10c/src/models/thread.ts#L142-L153

Note that this code both checks the cache for an event, then seems to indiscriminately fetch the root event from remote. According to the stack trace, this remote fetch is what leads to the redundant event subscriptions. (If I look at the network tab, I can see that there are many calls to `/rooms/{roomId}/event/{eventId}` for the same event, which I think is the thread root.)

I can't say for sure that this is the cause, but it at least seems like a bug in itself. If these are not related issues, I'd be happy to split this issue.

[...]

PR #3530, opened 2023-06-28
Avoid unnecessary work when processing threads

davidisaaclee:
## Checklist

---


spec/test-utils/test-utils.ts
```diff
@@ -486,3 +486,61 @@ CRYPTO_BACKENDS["rust-sdk"] = (client: MatrixClient) => client.initRustCrypto();
 if (global.Olm) {
     CRYPTO_BACKENDS["libolm"] = (client: MatrixClient) => client.initCrypto();
 }
+
```
davidisaaclee, line 489: I'd be more than happy to replace this with existing code or a third-party library if you know of one. Specifically, I needed to:
- imperatively resolve/reject promises
- manipulate promises that were created inside callbacks, so getting references to them would be difficult (especially given some of these callbacks would be called multiple times in the same test)

spec/unit/models/thread.spec.ts
```diff
@@ -77,6 +91,47 @@ describe("Thread", () => {
...
+    it("avoids fetching thread root event on addEvent in parallel", async () => {
+        const promiseFactory = controllablePromiseFactory<MatrixEvent, [roomId: string, eventId: string]>();
+        const { room, client, rootEvent, mockFetchRootEventHttp } = createRoomWithEventSituationUsingFetchMock();
+
+        const getRootEventHttp = mockFetchRootEventHttp(() =>
+            promiseFactory.makePromise(room.roomId, rootEvent.getId()!),
+        );
+
+        const thread = new Thread(rootEvent.getId()!, rootEvent, { room, client });
+        // Thread constructor has non-awaited async code, so we need to wait for it to finish
```
davidisaaclee, line 103: I think this points to another possible improvement where the async code in `Thread`'s constructor is moved to an async context. This would be a big breaking change, though.

spec/test-utils/test-utils.ts
t3chguy: We have a utility called `defer` which is likely to be helpful here

davidisaaclee: perfect – got rid of a lot of this code using that, thank you [Replace new controllablePromiseFactory utility with defer + simple cache](https://github.com/matrix-org/matrix-js-sdk/pull/3530/commits/a555e70f93c6decc562d3d9f970cb4dc56869f49)

davidisaaclee:
@t3chguy to update here: I'm running into at least one failing test starting with the first commit in this PR
[Avoid unnecessary fetch of thread root](https://github.com/matrix-org/matrix-js-sdk/pull/3530/commits/7cfedaae876d63ff5df48557c7001d3d86b9b5d3), which I'm working on fixing, but it's possible that this is opening a can of worms:
- **The test I'm focusing on** is https://github.com/davidisaaclee/matrix-js-sdk/blob/7cfedaae876d63ff5df48557c7001d3d86b9b5d3/spec/unit/room.spec.ts#L2521 – gist is that a thread is created, it gets a reply, that reply gets edited, `Thread.replyToEvent` should reflect the edited content of the reply
- **What actually happens:** the reply does not get the edit (`makeReplaced` is never called)
- **Why?** I'm still investigating this – some clues here are that [`eventShouldLiveIn`](https://github.com/davidisaaclee/matrix-js-sdk/blob/24cee68fa2023b2a8311ef335acf4f4d20118477/src/models/room.ts#L2096) returns `shouldLiveInThread: false` for edits made to a thread reply, which eventually prevents `makeReplaced` from getting called (I think? still reading...)
- **Why did this work before [7cfedaa](https://github.com/matrix-org/matrix-js-sdk/pull/3530/commits/7cfedaae876d63ff5df48557c7001d3d86b9b5d3)?** I'm also not 💯 on this, but it seems like somehow this code relied on the `fetchRootEvent` calls to operate[0]. I still think we should be able to get rid of these (imo) extra `fetchRootEvent`s, so I'm continuing to plug away here, but there is the chance that (1) this is requiring too much knowledge of matrix-js-sdk for me to implement quickly/cleanly, and (2) there will be many other broken bits due to this change. I'll update here and in https://github.com/matrix-org/matrix-js-sdk/issues/3463 if I give up :)

[0] tbc, this does not appear to be simply ["the thread root's `m.thread.latest_event` was updated in the last call to `fetchRootEvent`, so the `replyToEvent` got updated"](https://github.com/matrix-org/matrix-js-sdk/blob/24cee68fa2023b2a8311ef335acf4f4d20118477/spec/unit/room.spec.ts#L2574-L2581) – changing this mocked `fetchRootEvent` to have a different `m.thread` payload does not fail the test.

davidisaaclee:
Dumping some notes here for myself, and in case it helps someone down the line:

https://github.com/matrix-org/matrix-js-sdk/blob/c271e1533ae8eb1d1541694303d986f2e3bfc073/spec/unit/room.spec.ts#L2560

- When this is run, a Thread is created. Thread constructor runs async code in sync context; a `ThreadEvent.Update` signals its completion.
- At the same time, these events are added to the the Thread via `Room#addLiveEvents` / `Thread#addEvents`. This triggers a `ThreadEvent.Update`.
- The `emitPromise` below tries to catch the completion of the `Thread` setup. But it really catches the `addLiveEvents` `Update` event.
- The effect is that the rest of this test runs before the `Thread` constructors has completed.
- Notably, the reply edit event is sent before the thread fetches its first set of messages.
- When the thread finishes its first fetch, it emits the Update event (which we tried to listen for initially) – at this point, it has processed the edited reply as part of its construction.

1. add root event + replies
2. test hangs on first `await emitPromise`
3. thread is constructed and begins async processing...
4. Thread.addEvents emits Thread.update, fulfills the test's first `await emitPromise`
5. add edit event
6. test hangs on second `await emitPromise`
7. thread paginates in all events as part of `Thread.updateThreadMetadata`, including the edit event
8. thread construction completes, emits Thread.update, fulfills the test's second `await emitPromise`
9. `thread.replyToEvent` is checked in test: it has the edit thanks to step 7

spec/unit/room.spec.ts
```diff
@@ -2559,7 +2559,7 @@ describe("Room", function () {
             let prom = emitPromise(room, ThreadEvent.New);
             await room.addLiveEvents([randomMessage, threadRoot, threadResponse]);
             const thread: Thread = await prom;
-            await emitPromise(room, ThreadEvent.Update);
+            await Promise.all([emitPromise(room, RoomEvent.TimelineReset), emitPromise(room, ThreadEvent.Update)]);
```
davidisaaclee, line 2562: This waits for the correct signal that `Thread.constructor` has completed

src/models/room.ts
```diff
@@ -2256,6 +2256,13 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
             client: this.client,
             pendingEventOrdering: this.opts.pendingEventOrdering,
             receipts: this.cachedThreadReadReceipts.get(threadId) ?? [],
+
+            // This is necessary to be able to jump to events in threads:
+            // If we jump to an event in a thread where neither the event, nor
+            // the root, nor any thread event are loaded yet, we'll load the
+            // event as well as the thread root, create the thread, and pass
+            // the event through this.
+            events,
```
davidisaaclee, line 2265: adding `events` here (to `Thread#replayEvents`) means that `Thread.constructor` inserts these correctly regardless of async ordering; previously:
1. `addLiveEvents` called with multiple thread events
2. `Thread.constructor` is called using the thread root, starts async work...
3. the original `addLiveEvents` continues and tries to add the thread responses to the newly-created (but not fully initialized) `Thread`; fails check in `Thread#addEvent` where `this.initialEventsFetched`, so fails to add events to timeline
4. continuing `Thread.constructor` async: `updateThreadMetadata` is called, and tries to get initial fetch of events. _somehow this doesn't work – I forget now how (sorry, writing these notes a few days after investigating...)_

src/models/thread.ts
```diff
@@ -310,6 +313,11 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
         const lastReply = this.lastReply();
         const isNewestReply = !lastReply || event.localTimestamp >= lastReply!.localTimestamp;
 
+        if (isNewestReply) {
```
davidisaaclee, line 316: workaround for not refetching the thread root (on refetch, we'd get the latest reply event ID in the event's content)

davidisaaclee:
I'm giving up on this for now :( too many headaches following racing async codepaths. 

I'd love it if someone wanted to take over this branch. I added notes on this PR for my WIP, and I'd be more than happy to help handoff in any other way, since I still think this change is worthwhile.

PR #3533, opened 2023-06-29
Drop support for Node 16

richvdh:
Node 16 is quite old, and is somewhat slow. We don't appear to have any need to support it.

---

## 🦖 Deprecations
 * Drop support for Node 16 ([\#3533](https://github.com/matrix-org/matrix-js-sdk/pull/3533)).


t3chguy:
package.json engine field needs updating also

t3chguy:
beat me to it

t3chguy (approved):

PR #3537, opened 2023-07-03
Improve types around login, registration, UIA and identity servers

t3chguy:


src/interactive-auth.ts
```diff
@@ -51,22 +52,25 @@ export interface IStageStatus {
...
  */
 export interface IAuthData {
-    // XXX: many of the fields here (`type`, `available_flows`, `required_stages`, etc) look like they
-    // shouldn't be here. They aren't in the spec and it's unclear what they are supposed to do. Be wary of using them.
+    /**
+     * This is a session identifier that the client must pass back to the home server,
+     * if one is provided, in subsequent attempts to authenticate in the same API call.
+     */
     session?: string;
-    type?: string;
```
richvdh, line 57: Isn't this a breaking change? Applications may have abused this interface for other things (notably, they have confused it with `IAuthDict` in the past).

richvdh (changes requested):
this looks generally like an improvement, though it's a bit sprawling: it would have been nice to break it up into smaller changes.

src/@types/auth.ts
```diff
@@ -79,25 +81,148 @@ export interface IIdentityProvider {
...
+
+/**
+ * Request body for POST /login request
+ * See https://spec.matrix.org/v1.7/client-server-api/#post_matrixclientv3login
+ */
+export interface LoginRequest {
+    /**
+     * The login type being used.
+     */
+    type: "m.login.password" | "m.login.token" | string;
```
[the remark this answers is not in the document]
t3chguy: Yes, ideally it'd just be the string union but given the extensibility of Matrix we can't know all possible supported login types.

image

Intellisense still uses it to provide sensible suggestions.

src/client.ts
```diff
@@ -9314,8 +9321,12 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @returns Promise which resolves: Lookup results from IS.
      * @returns Rejects: with an error response.
      */
-    public async bulkLookupThreePids(query: [string, string][], identityAccessToken: string): Promise<any> {
-        // TODO: Types
+    public async bulkLookupThreePids(
+        query: [string, string][],
+        identityAccessToken: string,
+    ): Promise<{
+        threepids: [medium: string, address: string, mxid: string][];
```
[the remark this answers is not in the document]
t3chguy: Nope, I'm not changing the return type. Just documenting it.

image

Its an array of tuples.

src/interactive-auth.ts
```diff
@@ -216,20 +252,20 @@ interface IOpts {
...
-    private readonly stateUpdatedCallback: IOpts["stateUpdated"];
-    private readonly requestEmailTokenCallback: IOpts["requestEmailToken"];
+    private readonly requestCallback: IOpts<T>["doRequest"];
+    private readonly busyChangedCallback?: IOpts<T>["busyChanged"];
+    private readonly stateUpdatedCallback: IOpts<T>["stateUpdated"];
+    private readonly requestEmailTokenCallback: IOpts<T>["requestEmailToken"];
     private readonly supportedStages?: Set<string>;
 
-    private data: IAuthData;
+    private data: IAuthData | MatrixError;
```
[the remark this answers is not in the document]
t3chguy: For non-UIA (non 401) errors

src/client.ts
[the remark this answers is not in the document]
richvdh: ohhh. right.

src/@types/auth.ts
```diff
@@ -117,13 +129,17 @@ type PhoneLoginIdentifier = {
 
 type SpecUserIdentifier = UserLoginIdentifier | ThirdPartyLoginIdentifier | PhoneLoginIdentifier;
 
+/**
+ * User Identifiers usable for login & user-interactive authentication
+ * Extensibly allows more than Matrix specified identifiers
```
richvdh, line 134: Full stops are nice

```suggestion
 * User Identifiers usable for login & user-interactive authentication.
 *
 * Extensibly allows more than Matrix specified identifiers.
```

src/@types/auth.ts
```diff
@@ -90,15 +89,20 @@ export enum SSOAction {
 }
 
 /**
- * https://spec.matrix.org/v1.7/client-server-api/#matrix-user-id
+ * A client can identify a user using their Matrix ID.
+ * This can either be the fully qualified Matrix user ID, or just the localpart of the user ID.
+ * @see https://spec.matrix.org/v1.7/client-server-api/#matrix-user-id
```
richvdh, line 94: This is certainly helpful, but for bonus points it would be nice to say where this thing is used: if I look at https://pr3537--js-sdk-docs-previews.netlify.app/stable/types/_internal_.userloginidentifier it's just somewhat lacking in context.

Anyway no need to change this now but something I think we could be better at in future.

src/client.ts
```diff
@@ -7824,11 +7814,10 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 
     /**
-     * @returns Promise which resolves: TODO
+     @returns Promise which resolves to a LoginResponse object
```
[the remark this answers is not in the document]
richvdh: this is still broken I think?

src/interactive-auth.ts
```diff
@@ -262,6 +263,8 @@ export class InteractiveAuth<T> {
     private readonly requestEmailTokenCallback: IOpts<T>["requestEmailToken"];
     private readonly supportedStages?: Set<string>;
 
+    // The current latest data as part of the interactive auth
```
richvdh, line 266: the data being *sent* to the server, or received?

src/interactive-auth.ts
```diff
@@ -262,6 +263,8 @@ export class InteractiveAuth<T> {
...
+    // MatrixError can occur if the error from server is not a 401 UIA error
     private data: IAuthData | MatrixError;
```
richvdh: I might be being stupid, but can you point me to where a `MatrixError` is assigned to `data` ?

src/@types/registration.ts
```diff
@@ -0,0 +1,112 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { AuthDict } from "../interactive-auth";
+
+/**
+ * The request body of a call to POST https://spec.matrix.org/v1.7/client-server-api/#post_matrixclientv3register
```
richvdh, line 20: some words missing here

```suggestion
 * The request body of a call to `POST /_matrix/client/v3/register`.
 *
 * @see https://spec.matrix.org/v1.7/client-server-api/#post_matrixclientv3register
```

src/@types/registration.ts
```diff
@@ -0,0 +1,112 @@
...
+     */
+    bind_msisdn?: boolean;
+    /**
+     * @deprecated missing in the spec
+     */
+    bind_email?: boolean;
+}
+
+/**
+ * The result of a successful call to POST https://spec.matrix.org/v1.7/client-server-api/#post_matrixclientv3register
```
richvdh, line 71: ditto

src/interactive-auth.ts
t3chguy: Yeah honestly wasn't sure either, was inferring from private consumers. Getting rid of it seems to not break anything 🤷.

richvdh (approved):
looks grand

PR #3541, opened 2023-07-05
Prevent threads code from making identical simultaneous API hits

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/25395

Currently the `fetchRootEvent` method would get called for each emit of `Room.Timeline` which meant if a thread was found with 10 events, we'd call `/event/` 10 times. A lot of our tests seem to expect the `/event/` API being hammered which should have been a red flag much much earlier. This change doesn't fix all the issues with the Thread model (https://github.com/matrix-org/matrix-js-sdk/pull/3542 aims to do that) but helps avoid calling the init code paths multiple times (especially at the same time). The only time the init code path will be re-called is if the last event was redacted just to simplify the change. Perfect is the enemy of freaking hugely better.

---

## ✨ Features
 * Prevent threads code from making identical simultaneous API hits ([\#3541](https://github.com/matrix-org/matrix-js-sdk/pull/3541)). Fixes vector-im/element-web#25395.

### DIFF
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -362,7 +362,7 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
             return;
         }
 
-        if (event.isRelation(THREAD_RELATION_TYPE.name)) {
+        if (event.getId() !== this.id && event.isRelation(THREAD_RELATION_TYPE.name)) {
             this.replyCount++;
         }
 
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -362,9 +362,7 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
             return;
         }
 
-        // If no thread support exists we want to count all thread relation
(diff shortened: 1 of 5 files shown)


src/models/thread.ts
```diff
@@ -197,6 +198,10 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
             this._currentUserParticipated = false;
             this.emit(ThreadEvent.Delete, this);
         } else {
+            if (this.lastEvent?.getId() === event.getAssociatedId()) {
+                // XXX: If our last event got redacted we query the server for the last event once again
+                this.processRootEventPromise = undefined;
```
robintown, line 203: Are there any adverse effects that could occur if the new root event promise, racing with the previous one, completes before the previous one?

robintown (commented):
One question, though it's something I would have had about the old code too

src/models/thread.ts
t3chguy: Good point, I doubt it, but awaiting before resetting wouldn't hurt

PR #3546, opened 2023-07-06
ElementR: Add `checkOwnCrossSigningTrust` to `CryptoApi`

florianduros:
## Checklist

Content of the PR

- Add `checkOwnCrossSigningTrust` impl to rust-crypto
- Deprecated `MatrixClient.checkOwnCrossSigningTrust`
- The (unreadable) changes in `crypto.spec.ts` are documentation and function sharing. At one exception: the test `checkOwnCrossSigningTrust` is added
- Move `bootstrapSecretStorage` and `checkOwnCrossSigningTrust` into the same test suite to share utility functions

---

## ✨ Features
 * ElementR: Add `checkOwnCrossSigningTrust` to `CryptoApi` ([\#3546](https://github.com/matrix-org/matrix-js-sdk/pull/3546)). Contributed by @florianduros.


germain-gg (approved):
Looks great, thank you!

src/rust-crypto/rust-crypto.ts
```diff
@@ -640,6 +642,43 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
+        );
+
+        // Verify the device and upload the cross signing signatures
+        const request: RustSdkCryptoJs.SignatureUploadRequest = await device.verify();
+        await this.http.authedRequest(
+            Method.Post,
+            "/_matrix/client/v3/keys/signatures/upload",
+            undefined,
+            request.body,
+        );
```
richvdh, line 679: surely this should be documented in the doc-comment in cryptoApi?

src/crypto-api.ts
```diff
@@ -490,5 +501,15 @@ export interface GeneratedSecretStorageKey {
     encodedPrivateKey?: string;
 }
 
+/**
+ * Options object for {@link CryptoApi#checkOwnCrossSigningTrust}.
+ */
+export interface CheckOwnCrossSigningTrustOpts {
+    /**
+     * If true, will fetch the private cross signing keys from the secret storage
+     */
```
richvdh, line 510: what happens if it is false?

src/rust-crypto/rust-crypto.ts
```diff
@@ -640,6 +642,43 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
+     * Implementation of {@link CryptoApi#checkOwnCrossSigningTrust}
+     */
+    public async checkOwnCrossSigningTrust({ allowPrivateKeyRequests }: CheckOwnCrossSigningTrustOpts): Promise<void> {
+        // Download the device keys and add them into the olm machine
+        const queryResult = await this.downloadDeviceList(new Set([this.userId]));
+        await this.olmMachine.markRequestAsSent(
+            uuidv4(),
+            RustSdkCryptoJs.RequestType.KeysQuery,
+            JSON.stringify(queryResult),
+        );
```
richvdh, line 655: why is this necessary? making up a request id and telling the rust layer that we have sent that made-up request ID seems like an odd thing to do.

src/crypto-api.ts
```diff
@@ -229,6 +229,17 @@ export interface CryptoApi {
      */
     createRecoveryKeyFromPassphrase(password?: string): Promise<GeneratedSecretStorageKey>;
 
+    /**
+     * Check the cross signing trust of the current user
```
richvdh, line 233: > Check the cross signing trust of the current user

what happens if the check fails?

richvdh (changes requested):
I has questions

src/crypto-api.ts
florianduros: The cross signing keys are not imported

src/rust-crypto/rust-crypto.ts
florianduros: In the doc-comment: `Verify the current device`. We should also tell that we are uploading the device signatures ?

src/crypto-api.ts
florianduros: Depends when the failure occurs; the device will not be verified (the device signatures will not be uploaded), the cross signing keys or the device keys will no be imported

src/rust-crypto/rust-crypto.ts
richvdh: I would have thought so. "Verify" is quite different to "sign and upload".

src/crypto-api.ts
richvdh: ok, please could you make this explicit in the comment.

src/rust-crypto/rust-crypto.ts
richvdh: well, first of all, doesn't the OlmMachine download the device keys automatically?

src/crypto-api.ts
richvdh: please put this in the documentation

richvdh: (will an exception be thrown if the check fails?)

src/rust-crypto/rust-crypto.ts
florianduros: Oh, you mean that I don't need the download the device list manually ? It's already done by the olm machine ?

florianduros:
Abandoned: see https://github.com/vector-im/element-web/issues/25648

PR #3549, opened 2023-07-06
Don't allow Olm unwedging rate-limiting to race

uhoreg:
Previously, we only marked the time that we re-created an Olm session after the session was created.  So if we try to unwedge while we're in the middle of unwedging, it won't be rate-limited, and we end up creating multiple new sessions.  This patch sets up rate limiting before any `await`s, with a delay of 5 minutes, so it doesn't race any more.

Fixes https://github.com/vector-im/element-web/issues/25716

No tests yet because I'm not sure how to test this.  Suggestions welcome.

## Checklist

---

## 🐛 Bug Fixes
 * Don't allow Olm unwedging rate-limiting to race ([\#3549](https://github.com/matrix-org/matrix-js-sdk/pull/3549)). Fixes vector-im/element-web#25716.

### DIFF
--- a/src/crypto/index.ts
+++ b/src/crypto/index.ts
@@ -130,7 +130,12 @@ export function isCryptoAvailable(): boolean {
     return Boolean(global.Olm);
 }
 
+// minimum time between attempting to unwedge an Olm session, if we succeeded
+// in creating a new session
 const MIN_FORCE_SESSION_INTERVAL_MS = 60 * 60 * 1000;
+// minimum time between attempting to unwedge an Olm session, if we failed
+// to create a new session
+const FORCE_SESSION_RETRY_INTERVAL_MS = 5 * 60 * 1000;
 
 interface IInitOpts {
     exportedOlmDevice?: IExportedDevice;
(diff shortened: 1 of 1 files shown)


src/crypto/index.ts
```diff
@@ -3591,25 +3596,27 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
...
             logger.debug(
                 "New session already forced with device " +
                     sender +
                     ":" +
                     deviceKey +
-                    " at " +
-                    lastNewSessionForced +
-                    ": not forcing another",
+                    ": not forcing another until at least ",
+                    forceNewSessionRetryTime
```
richvdh, line 3610: suggest updating this to backtick interpolation while you are here

```
`New session already forced with device ${sender} ...`
```

etc

richvdh: also suggest `new Date(forceNewSessionRetryTime).toUTCString()` to make the timestamp readable

src/crypto/index.ts
```diff
@@ -130,7 +130,12 @@ export function isCryptoAvailable(): boolean {
     return Boolean(global.Olm);
 }
 
+// minimum time between attempting to unwedge an Olm session, if we succeeded
+// in creating a new session
 const MIN_FORCE_SESSION_INTERVAL_MS = 60 * 60 * 1000;
+// minimum time between attempting to unwedge an Olm session, if we failed
+// to create a new session
+const FORCE_SESSION_RETRY_INTERVAL_MS = 5 * 60 * 1000;
```
richvdh, line 138:
```suggestion
const FORCE_SESSION_RETRY_INTERVAL_MS = 5 * 60 * 1000;    // 5 minutes
```

richvdh (commented):
looks sensible to me. I wouldn't worry too much about a unit test; hopefully this code will only live for a couple more months.

richvdh (approved):
lgtm

PR #3550, opened 2023-07-06
Replace deprecated TestClient with fetchMock

BillCarsonFr:
Replace use of deprecated TestClient and use  fetchMock 

Fixes https://github.com/vector-im/crypto-internal/issues/103

## Checklist

---

### DIFF
--- a/spec/integ/crypto/megolm-backup.spec.ts
+++ b/spec/integ/crypto/megolm-backup.spec.ts
@@ -14,16 +14,24 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { Account } from "@matrix-org/olm";
+import fetchMock from "fetch-mock-jest";
+import "fake-indexeddb/auto";
 
 import { logger } from "../../../src/logger";
 import { decodeRecoveryKey } from "../../../src/crypto/recoverykey";
 import { IKeyBackupInfo, IKeyBackupSession } from "../../../src/crypto/keybackup";
-import { TestClient } from "../../TestClient";
-import { IEvent } from "../../../src";
+import { createClient, ICreateClientOpts, IEvent, MatrixClient } from "../../../src";
 import { MatrixEvent, MatrixEventEvent } from "../../../src/models/event";
+import { SyncResponder } from "../../test-utils/SyncResponder";
+import { E2EKeyReceiver } from "../../test-utils/E2EKeyReceiver";
+import { E2EKeyResponder } from "../../test-utils/E2EKeyResponder";
+import { mockInitialApiRequests } from "../../test-utils/mockEndpoints";
+import { syncPromise } from "../../test-utils/test-utils";
 
 const ROOM_ID = "!ROOM:ID";
+
+/** The homeserver url that we give to the test client, and where we intercept /sync, /keys, etc requests. */
+const TEST_HOMESERVER_URL = "https://alice-server.com";
 
 const SESSION_ID = "o+21hSjP+mgEmcfdslPsQdvzWnkdt0Wyo00Kp++R8Kc";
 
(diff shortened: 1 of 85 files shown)


src/crypto/backup.ts
```diff
@@ -118,10 +118,16 @@ export class BackupManager {
     public checkedForBackup: boolean; // Have we checked the server for a backup we can use?
     private sendingBackups: boolean; // Are we currently sending backups?
     private sessionLastCheckAttemptedTime: Record<string, number> = {}; // When did we last try to check the server for a given session id?
+    private clientRunning: boolean;
```
t3chguy, line 121: The changes in this file do not match the PR title

src/crypto/index.ts
```diff
@@ -1832,6 +1832,7 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
         this.outgoingRoomKeyRequestManager.stop();
         this.deviceList.stop();
         this.dehydrationManager.stop();
+        this.backupManager.stop();
```
t3chguy, line 1835: The changes in this file do not match the PR title

src/crypto/backup.ts
BillCarsonFr: The test was failing, because something was attempting to log after the test end. It appeared that the backup manager is not shutdown when the client is stopped. So this was added to make the test work.

richvdh: I think these changes are required to get the tests to exit correctly. That said, ideally they could be a separate PR (or at least a separate commit).

src/crypto/backup.ts
```diff
@@ -118,10 +118,16 @@ export class BackupManager {
...
 
     public constructor(private readonly baseApis: MatrixClient, public readonly getKey: GetKey) {
         this.checkedForBackup = false;
         this.sendingBackups = false;
+        this.clientRunning = true;
+    }
+
+    public stop(): void {
+        this.clientRunning = false;
     }
```
richvdh, line 131: needs a doc-comment please. What does this do?

src/crypto/backup.ts
t3chguy: Its a functional change which should be in the changelo

src/crypto/backup.ts
```diff
@@ -439,6 +445,11 @@ export class BackupManager {
             // the same time when a new key is sent
             const delay = Math.random() * maxDelay;
             await sleep(delay);
+            if (!this.clientRunning) {
+                logger.debug("Key backup send aborted, client stopped");
+                this.sendingBackups = false;
```
richvdh, line 450: doesn't the `finally` block take care of this?

src/crypto/backup.ts
richvdh: yeah. Which I think points towards making it a separate PR with a `T-Enhancement` label.

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -70,101 +78,133 @@ const CURVE25519_BACKUP_INFO: IKeyBackupInfo = {
...
+            ...opts,
+        });
+        await client.initCrypto();
+
+        return client;
+    }
+
+    it("Alice checks key backups when receiving a message she can't decrypt", async function () {
+        fetchMock.get("express:/_matrix/client/v3/room_keys/keys/:room_id/:session_id", CURVE25519_KEY_BACKUP_DATA);
+        fetchMock.post("express:/_matrix/client/r0/keys/upload", {
```
richvdh, line 166: no need for this to be an `express:` matcher, since there are no path params

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -70,101 +78,133 @@ const CURVE25519_BACKUP_INFO: IKeyBackupInfo = {
...
+        fetchMock.put("express:/_matrix/client/r0/sendToDevice/m.room_key_request/:txid", {});
+
+        fetchMock.get("express:/_matrix/client/v3/room_keys/version", CURVE25519_BACKUP_INFO);
+
+        aliceClient = await initTestClient();
+
+        // we need the backup to be trusted for the test to work
+        const backupDataToSign = JSON.parse(JSON.stringify(CURVE25519_BACKUP_INFO));
+
+        await aliceClient.crypto!.signObject(backupDataToSign.auth_data);
```
richvdh, line 182: I feel like we should find out why this is needed when it wasn't before.

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -70,101 +78,133 @@ const CURVE25519_BACKUP_INFO: IKeyBackupInfo = {
...
 
 describe("megolm key backups", function () {
-    if (!global.Olm) {
-        logger.warn("not running megolm tests: Olm not present");
-        return;
-    }
-    const Olm = global.Olm;
-    let testOlmAccount: Olm.Account;
-    let aliceTestClient: TestClient;
+    const syncResponse = {
```
richvdh, line 104: I feel like this still belongs inside the `it(...)`, not the `describe(...)`.

(also, it would make the diff much easier to read...)

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -70,101 +78,133 @@ const CURVE25519_BACKUP_INFO: IKeyBackupInfo = {
     version: "1",
     auth_data: {
         public_key: "hSDwCYkwp1R0i33ctD73Wg2/Og0mOBr066SpjqqbTmo",
+        // Will be updated with correct value on the fly
+        signatures: {
+            "@alice:localhost": {
+                "ed25519:BGVSTTXKZB":
+                    "RNUIJIcDrsyvKV4MReMzllZ6yx05803tR0/tdO8poS62wUsVNOgxglC2VPwGSxji/hxv7c7D8FTuQ3uQWcR1DQ",
+            },
+        },
```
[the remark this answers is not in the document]
BillCarsonFr: if there is no signatures field the backup is dimmed invalid. We could though only add signatures: {}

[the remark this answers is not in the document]
richvdh: is it not also considered invalid if it has an invalid signature?

Just wondering why an invalid signature or `signatures: {}` is better than nothing at all

spec/integ/crypto/megolm-backup.spec.ts
BillCarsonFr: The legacy test was forcing:
> aliceTestClient!.client.crypto!.backupManager.backupInfo = CURVE25519_BACKUP_INFO;

This was by-passing the checkAndStart() that is supposed to download and check (and would fail with the incomplete test data). `backupInfo` is only set if the backup is trusted

src/crypto/index.ts
BillCarsonFr: https://github.com/matrix-org/matrix-js-sdk/pull/3553

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -14,17 +14,25 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { Account } from "@matrix-org/olm";
+import fetchMock from "fetch-mock-jest";
+import "fake-indexeddb/auto";
```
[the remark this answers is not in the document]
BillCarsonFr: yes removed

spec/integ/crypto/megolm-backup.spec.ts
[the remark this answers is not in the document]
BillCarsonFr: it's in the code:
```
   if (!backupInfo || !backupInfo.algorithm || !backupInfo.auth_data || !backupInfo.auth_data.signatures) {
            logger.info("Key backup is absent or missing required data");
            return ret;
        }
```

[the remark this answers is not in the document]
BillCarsonFr: The checkAndStart is done when client is initiated, so before fetchmock can add a proper response. 
I create a correct backup then call again checkAndStart() in the code

spec/integ/crypto/megolm-backup.spec.ts
BillCarsonFr: done

spec/integ/crypto/megolm-backup.spec.ts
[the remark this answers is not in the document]
richvdh: > it's in the code:
> 
> ```
>    if (!backupInfo || !backupInfo.algorithm || !backupInfo.auth_data || !backupInfo.auth_data.signatures) {
>             logger.info("Key backup is absent or missing required data");
>             return ret;
>         }
> ```

for the record, this code is in `BackupManager.isKeyBackupTrusted`: https://github.com/matrix-org/matrix-js-sdk/blob/ef003ef337cc7540382561d07637c2efe77e4d3d/src/crypto/backup.ts#L333-L336.

I'm afraid I still don't understand. `isKeyBackupTrusted` returns `{ usable: false,  trusted_locally: false, sigs: [] }` if there is no `signatures` property on the backup data it receives. As far as I can tell, it returns exactly the same thing if `signatures` is empty.

So I ask again: why do we need a `signatures` property here?

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -134,37 +149,42 @@ describe("megolm key backups", function () {
...
-                });
-            })
-            .then((event) => {
-                expect(event.getContent()).toEqual("testytest");
+        fetchMock.get("express:/_matrix/client/v3/room_keys/keys/:room_id/:session_id", CURVE25519_KEY_BACKUP_DATA);
+
+        // mock for the outgoing key requests that will be sent
+        fetchMock.put("express:/_matrix/client/r0/sendToDevice/m.room_key_request/:txid", {});
+
+        fetchMock.get("express:/_matrix/client/v3/room_keys/version", CURVE25519_BACKUP_INFO);
```
richvdh, line 157:
```suggestion
        fetchMock.get("/_matrix/client/v3/room_keys/version", CURVE25519_BACKUP_INFO);
```

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -134,37 +149,42 @@ describe("megolm key backups", function () {
...
+
+        aliceClient = await initTestClient();
+
+        // we need the backup to be trusted for the test to work
+        const backupDataToSign = JSON.parse(JSON.stringify(CURVE25519_BACKUP_INFO));
+
+        await aliceClient.crypto!.signObject(backupDataToSign.auth_data);
+        fetchMock.get("express:/_matrix/client/v3/room_keys/version", backupDataToSign, {
+            overwriteRoutes: true,
+        });
```
richvdh, line 167: if you move backupDataToSign earlier, there is no need to redo the mock:

```suggestion
        // take a copy of the backup data; we will sign it once we start the client
        const backupData = JSON.parse(JSON.stringify(CURVE25519_BACKUP_INFO));
        fetchMock.get("/_matrix/client/v3/room_keys/version", backupData);

        aliceClient = await initTestClient();

        // we need the backup to be trusted for the test to work
        await aliceClient.crypto!.signObject(backupData.auth_data);
```

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -134,37 +149,42 @@ describe("megolm key backups", function () {
...
-            .then(() => {
-                return createOlmSession(testOlmAccount, aliceTestClient);
-            })
-            .then(() => {
-                const privkey = decodeRecoveryKey(RECOVERY_KEY);
-                return aliceTestClient!.client!.crypto!.storeSessionBackupPrivateKey(privkey);
-            })
-            .then(() => {
-                aliceTestClient!.httpBackend.when("GET", "/sync").respond(200, syncResponse);
-                aliceTestClient!.expectKeyBackupQuery(ROOM_ID, SESSION_ID, 200, CURVE25519_KEY_BACKUP_DATA);
```
richvdh, line 148: it would be nice to rip out the now-unused `TestClient.expectKeyBackupQuery`, but maybe that's a job for another day

spec/integ/crypto/megolm-backup.spec.ts
BillCarsonFr: I am surprise that we can sign before calling init crypto on the client

PR #3551, opened 2023-07-06
Fix linting errors caused by the old crypto SDK

daniel-abramov:
See https://github.com/vector-im/element-call/pull/1173#issuecomment-1621306979 for more details

---


robintown (approved):

robintown:
About the test failures: I think I remember Rich saying that he recently removed support for Node 16, which is exactly what this is failing on. So maybe we need to merge develop -> livekit?

daniel-abramov:
> About the test failures: I think I remember Rich saying that he recently removed support for Node 16 due to Rust crypto changes, which is exactly what this is failing on. So maybe we need to merge develop -> livekit?

Done.

PR #3553, opened 2023-07-06
add stop() api to BackupManager for clean shutdown

BillCarsonFr:
When the client is stopped the background upload of keys to backup is not shutdown.
Added a new stop() api that is called by crypto (similar to outgoingRoomKeyRequestManager , deviceList ..)

## Checklist

---

## ✨ Features
 * add stop() api to BackupManager for clean shutdown ([\#3553](https://github.com/matrix-org/matrix-js-sdk/pull/3553)).


t3chguy (approved):
Looks sane

PR #3558, opened 2023-07-07
add getOpenIdToken capability to embedded mode

### LINKED ISSUE #3538: Add OIDC widget api event as defined in MSC1960
The js-sdk can also be used with the widget-api backend. Instead of talking directly to the matrix server it talks to the parent client and requests events through the postmessage api.

This is used for element-call. To support the livekit sfu, we are using the oidc token that can be generated by a matrix homeserver to authenticate at the livekit sfu. (more context can be found here: https://github.com/vector-im/element-call/pull/1166)
To support this in the widget the js-sdk needs to support [MSC1960](https://github.com/matrix-org/matrix-spec-proposals/blob/old_master/proposals/1960-integrations-openid.md)
Element Web already supports this widget api feature so this change is only required in the js-sdk widget backend.
(Update from dave: this still might trigger a permissions request... so we need to deal with that or patch EW to autoapprove this permission for element-call)
 The widget backend in the js-sdk needs to support this. To make element-call with livekit work in embedded mode.

toger5:
Fixes: https://github.com/matrix-org/matrix-js-sdk/issues/3538
This maked the embedded backend support oidc tokens.

---

## ✨ Features
 * add getOpenIdToken capability to embedded mode ([\#3558](https://github.com/matrix-org/matrix-js-sdk/pull/3558)). Fixes #3538. Contributed by @toger5.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -74,7 +74,7 @@ module.exports = {
         "jest/no-standalone-expect": [
             "error",
             {
-                additionalTestBlockFunctions: ["beforeAll", "beforeEach", "oldBackendOnly"],
+                additionalTestBlockFunctions: ["beforeAll", "beforeEach", "oldBackendOnly", "newBackendOnly"],
             },
         ],
     },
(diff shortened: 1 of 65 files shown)


t3chguy (changes requested):
CI looks unhappy

toger5:
replaced with: https://github.com/matrix-org/matrix-js-sdk/pull/3571

PR #3555, opened 2023-07-07
Add `CryptoApi.getActiveSessionBackupVersion()`

BillCarsonFr:
... which replaces which replaces `BackupManager.getKeyBackupEnabled()` and `BackupManager.version`

Fixes https://github.com/vector-im/crypto-internal/issues/106

~~Based on https://github.com/matrix-org/matrix-js-sdk/pull/3631~~

---
This change is marked as an *internal change* (Task), so will not be included in the changelog.

PR #3559, opened 2023-07-07
Key backup: clean up SecureKeyBackup.prepareKeyBackupVersion

BillCarsonFr:
Depends on https://github.com/matrix-org/matrix-js-sdk/pull/3555

Fixes #https://github.com/vector-im/crypto-internal/issues/107

Moved prepareKeyBackupVersion() in the new SecureBackup API.
Moved also the part that was doing the signing to do it once inside this API
The SecureBackup API exposes a `prepareUnsignedKeyBackupVersion()` it doesn't sign here because it would require a back reference to crypto. Instead the CryptoAPi exposes now a `prepareKeyBackupVersion() ` that calls the backup API to get an unsigned version and then does the signing. In order to avoid breaking change the existing `prepareKeyBackupVersion` on backup.ts is deprecated and now calls the unsigned version.

Clients are using MatrixClient.prepareKeyBackupVersion that transparantly follow the call.

## Checklist

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     ],
     "dependencies": {
         "@babel/runtime": "^7.12.5",
-        "@matrix-org/matrix-sdk-crypto-wasm": "^1.0.1",
+        "@matrix-org/matrix-sdk-crypto-wasm": "^1.1.0",
         "another-json": "^0.2.0",
         "bs58": "^5.0.0",
         "content-type": "^1.0.4",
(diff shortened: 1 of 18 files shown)


richvdh:
Per my recent updates on https://github.com/vector-im/crypto-internal/issues/107: I don't think we want to do this as initially envisaged.

Leaving this open for now for @BillCarsonFr to decide if there is anything salvageable, but I suspect not.

PR #3562, opened 2023-07-10
Add async method for generating a QR code

richvdh:
The api to generate a QR code is async in rust, and the easiest way to deal with it is to make a new method.

Notes: Deprecate `VerificationRequest.getQRCodeBytes` and replace it with the asynchronous `generateQRCode`.

---

## 🦖 Deprecations
 * Deprecate `VerificationRequest.getQRCodeBytes` and replace it with the asynchronous `generateQRCode`. ([\#3562](https://github.com/matrix-org/matrix-js-sdk/pull/3562)).


richvdh:
ugh, the coverage gate is being an ass. Obviously I will add tests; no I'm not going to add a test that my stub method is correctly stubbed.

SimonBrandner (approved):

Issue #3506, opened 2023-06-25
Expose `EmittedEvents` subtypes


farooqkz:
Currently it is a compound of these:

```
ClientEvent | RoomEvents | RoomStateEvents | CryptoEvents | MatrixEventEvents | RoomMemberEvents | UserEvents | CallEvent | Incoming | Incoming | Outgoing | Ended | Participants | SessionLoggedOut | NoConsent | BeaconEvent
```

But few of them are exported and many like `CallEvent`, `Incoming` and `RoomEvent` are not exported and therefore cannot be used.  And we have to hardcode their values in our apps rather than use these enums which is the right way.

richvdh:
It's a general theme of matrix-js-sdk that it doesn't correctly export types that are meant to be public.

You can generally work around it by directly including the file where they are declared. For example, in the case of `CallEvent`, you can try:

```javascript
import type { CallEvent } from "matrix-js-sdk/src/webrtc/call";
```

richvdh:
Totally agree. Importing things from individual files is just a workaround.

davidisaaclee:
In my setup (Webpack 5 via create-react-app), importing an enum from a specific file (e.g. `import { ThreadEvent } from 'matrix-js-sdk/src/models/Thread'`) causes lots of build errors. I'm bad at JS build tools, so I'm not entirely sure what the problem is, but it sounds to me like the import is triggering Webpack to try to parse the referenced file, which fails.

I'm working around these issues by (1) using literal strings instead of enum members (`'Thread.new' as any`), and (2) maintaining a fork.

tbc, would PRs exposing some of these enums / types be welcome?

richvdh:
Yes, I'd be very happy to review PRs exposing these enums/types.

farooqkz:
> In my setup (Webpack 5 via create-react-app), importing an enum from a specific file (e.g. `import { ThreadEvent } from 'matrix-js-sdk/src/models/Thread'`) causes lots of build errors. I'm bad at JS build tools, so I'm not entirely sure what the problem is, but it sounds to me like the import is triggering Webpack to try to parse the referenced file, which fails.
> 
> I'm working around these issues by (1) using literal strings instead of enum members (`'Thread.new' as any`), and (2) maintaining a fork.
> 
> tbc, would PRs exposing some of these enums / types be welcome?

I have similar problems in my setup too when importing from individual SDK files. I'm using typescript to using strings directly is another problem...

[...]

PR #3565, opened 2023-07-10
Add support for scanning QR codes during verification, with Rust crypto

richvdh:
Before I change *displaying* QR codes, I need an end-to-end test for it. And before I can write an end-to-end test for it, I need to be able to *scan* QR codes. So here is support in the js-sdk for scanning QR codes. It will only work with Rust crypto, but that's enough for a Cypress test, and Rust is the future.

---

## ✨ Features
 * Add support for scanning QR codes during verification, with Rust crypto ([\#3565](https://github.com/matrix-org/matrix-js-sdk/pull/3565)).

### DIFF
--- a/src/rust-crypto/rust-crypto.ts
+++ b/src/rust-crypto/rust-crypto.ts
@@ -56,6 +56,8 @@ import { EventType } from "../@types/event";
 import { CryptoEvent } from "../crypto";
 import { TypedEventEmitter } from "../models/typed-event-emitter";
 
+const ALL_VERIFICATION_METHODS = ["m.sas.v1", "m.qr_code.scan.v1", "m.qr_code.show.v1", "m.reciprocate.v1"];
+
 /**
  * An implementation of {@link CryptoBackend} using the Rust matrix-sdk-crypto.
  */
@@ -558,7 +560,7 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
                     new RustVerificationRequest(
                         request,
                         this.outgoingRequestProcessor,
-                        this.supportedVerificationMethods,
+                        this._supportedVerificationMethods,
                     ),
             );
     }
@@ -580,10 +582,18 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
 
     /**
      * The verification methods we offer to the other side during an interactive verification.
+     */
+    private _supportedVerificationMethods: string[] = ALL_VERIFICATION_METHODS;
+
+    /**
+     * Set the verification methods we offer to the other side during an interactive verification.
      *
      * If `undefined`, we will offer all the methods supported by the Rust SDK.
      */
-    public supportedVerificationMethods: string[] | undefined;
+    public setSupportedVerificationMethods(methods: string[] | undefined): void {
+        // by default, the Rust SDK does not offer `m.qr_code.scan.v1`, but we do want to offer that.
+        this._supportedVerificationMethods = methods ?? ALL_VERIFICATION_METHODS;
+    }
 
     /**
      * Send a verification request to our other devices.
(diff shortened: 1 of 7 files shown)


src/rust-crypto/rust-crypto.ts
```diff
@@ -580,10 +582,18 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
      * The verification methods we offer to the other side during an interactive verification.
+     */
+    private _supportedVerificationMethods: string[] = ALL_VERIFICATION_METHODS;
+
+    /**
+     * Set the verification methods we offer to the other side during an interactive verification.
      *
      * If `undefined`, we will offer all the methods supported by the Rust SDK.
      */
-    public supportedVerificationMethods: string[] | undefined;
```
t3chguy, line 586: Looks like a breaking change, any reason to not use a getter+setter?

src/rust-crypto/verification.ts
```diff
@@ -56,17 +56,18 @@ export class RustVerificationRequest
     public constructor(
         private readonly inner: RustSdkCryptoJs.VerificationRequest,
         private readonly outgoingRequestProcessor: OutgoingRequestProcessor,
-        private readonly supportedVerificationMethods: string[] | undefined,
+        private readonly supportedVerificationMethods: string[],
```
t3chguy, line 59: Looks like a breaking change

src/rust-crypto/verification.ts
```diff
@@ -357,22 +387,26 @@ export class RustVerificationRequest
     }
 }
 
-export class RustSASVerifier extends TypedEventEmitter<VerifierEvent, VerifierEventHandlerMap> implements Verifier {
+/** Common base class for `Verifier` implementations which wrap rust classes.
+ *
+ * The generic parameter `InnerType` is the type of the rust Verification class which we wrap.
+ */
+class BaseRustVerifer<InnerType extends RustSdkCryptoJs.Qr | RustSdkCryptoJs.Sas> extends TypedEventEmitter<
```
t3chguy, line 394: if this is a base class should it be abstract?

src/rust-crypto/rust-crypto.ts
richvdh: this isn't a public class, so it's not a breaking change.

t3chguy: What makes it not public? Where is it marked as internal or experimental?

src/rust-crypto/verification.ts
richvdh: ditto

t3chguy: Same as above

src/rust-crypto/rust-crypto.ts
t3chguy: Neither are a bunch of the things necessary to use the SDK so I don't think that's the rule. (Namely the Typed EventEmitter events which are necessary to set up listeners0.

richvdh: I mean, I always thought that *was* the definition of what was public/private, and in line with... pretty much every other javascript library on the planet? (There might be exceptions, but I'd expect them to at least be documented).

The fact that the react-sdk gut-wrenches bits of the internals is... deeply wrong. And yes, a lint rule in the react-sdk to enforce it would be great.

richvdh: > Neither are a bunch of the things necessary to use the SDK so I don't think that's the rule. (Namely the Typed EventEmitter events which are necessary to set up listeners0.

Not quite sure what you're referring to, but anything not being exported through index.js that we expect clients to be able to refer to is very definitely a bug.

t3chguy: > Not quite sure what you're referring to

https://github.com/matrix-org/matrix-js-sdk/issues/3506

richvdh: yes well, that's a bug, and there's the issue to track it. Great.

The point remains that (a) if anybody has tried using rust crypto at all outside the react-sdk I will eat my hat; (b) even if it has, using `RustCrypto` or `RustVerificationRequest` directly is an application bug. I won't obfuscate this API just to support someone relying on the internals of the library in this way.

richvdh: https://github.com/matrix-org/matrix-js-sdk/pull/3574 then, if my argument isn't convincing.

richvdh: This is now an `@internal` class.

t3chguy (approved):

Issue #1844, opened 2021-08-12
setPowerLevel seems to clobber existing power levels


reivilibre:
`setPowerLevel` (https://github.com/matrix-org/matrix-js-sdk/blob/696b3ef1ce15a8c2c88006f6d058b3da5c4e3b32/src/client.ts#L3327-L3357) is described as setting a user's power level.

However, when using it, it seemed to discard the remaining power levels.

I notice now that it takes a slightly underdocumented event parameter. Is this meant to be the prior power level state event?

t3chguy:
> Is this meant to be the prior power level state event?

Yup

dhk2:
> Usage example https://github.com/matrix-org/matrix-react-sdk/blob/390b05617c78168aa91d128d0e3f6e9aeb89b7bf/src/SlashCommands.tsx#L783-L785

Thank you! that usage example was just what I needed.

PR #3570, opened 2023-07-10
Simplify `MatrixClient::setPowerLevel` API

t3chguy:
While making it more resilient to causing issues like nuking room state
Removes the need to pass the current power level `event` param

Fixes https://github.com/vector-im/element-web/issues/13900
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/1844

---

## 🦖 Deprecations
 * Simplify `MatrixClient::setPowerLevel` API ([\#3570](https://github.com/matrix-org/matrix-js-sdk/pull/3570)). Fixes vector-im/element-web#13900 and #1844.


kerryarchibald (approved):
LGTM

PR #3574, opened 2023-07-10
Mark all the rust crypto stuff internal

richvdh:
... for the avoidance of doubt such as that discussed at https://github.com/matrix-org/matrix-js-sdk/pull/3565#pullrequestreview-1522185838

Personally I think this is totally unnecessary but if everybody else thinks this is the only way for things to not be considered part of the public API, I guess I have to live with it.

---


andybalaam (approved):
I don't think this does any harm, and helps us be more explicit, so let's do it.

richvdh:
cypress failures appear to be long-standing and unrelated

PR #3578, opened 2023-07-11
Add methods to influence set_presence on /sync API calls

t3chguy:


src/sliding-sync-sdk.ts
```diff
@@ -463,6 +464,14 @@ export class SlidingSyncSdk {
         // TODO
     }
 
+    /**
+     * Specify the set_presence value to be used for subsequent calls to the Sync API.
+     * @param presence - the presence to specify to set_presence of sync calls
+     */
+    public setPresence(presence?: SetPresence): void {
+        // TODO
```
andybalaam, line 472: Could do with more information here. Is this impossible with sliding sync, or not yet done in the proxy, or just not yet done in our code?

andybalaam (changes requested):
Looks good, but wanting more explanation for the TODO.

src/sliding-sync-sdk.ts
t3chguy: This is just a stub like the 3 methods above it, `syncLeftRooms`, `peek`, `stopPeeking`. Sliding Sync hasn't even began to consider auxiliary features like this.

andybalaam: Can you update the TODOs to say that please?

andybalaam (changes requested):

andybalaam (approved):

PR #3588, opened 2023-07-11
Element-R: support for displaying QR codes during verification

richvdh:
Closes https://github.com/vector-im/crypto-internal/issues/124

---

## ✨ Features
 * Element-R: support for displaying QR codes during verification ([\#3588](https://github.com/matrix-org/matrix-js-sdk/pull/3588)). Fixes vector-im/crypto-internal#124.

### DIFF
--- a/src/rust-crypto/verification.ts
+++ b/src/rust-crypto/verification.ts
@@ -357,8 +357,7 @@ export class RustVerificationRequest
      * Stub implementation of {@link Crypto.VerificationRequest#getQRCodeBytes}.
      */
     public getQRCodeBytes(): Buffer | undefined {
-        // TODO
-        return undefined;
+        throw new Error("getQRCodeBytes() unsupported in Rust Crypto; use generateQRCode() instead.");
     }
 
     /**
@@ -367,8 +366,8 @@ export class RustVerificationRequest
      * Implementation of {@link Crypto.VerificationRequest#generateQRCode}.
      */
     public async generateQRCode(): Promise<Buffer | undefined> {
-        // TODO
-        return undefined;
+        const innerVerifier: RustSdkCryptoJs.Qr = await this.inner.generateQrCode();
+        return Buffer.from(innerVerifier.toBytes());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/rust-crypto/verification.ts
```diff
@@ -40,13 +41,16 @@ export class RustVerificationRequest
     extends TypedEventEmitter<VerificationRequestEvent, VerificationRequestEventHandlerMap>
     implements VerificationRequest
 {
+    /** a reëmitter which relays VerificationRequestEvent.Changed events emitted by the verifier */
```
t3chguy, line 44: `ë`?

src/rust-crypto/verification.ts
```diff
@@ -40,13 +41,16 @@ export class RustVerificationRequest
...
+    private readonly _reEmitter: TypedReEmitter<VerificationRequestEvent, VerificationRequestEventHandlerMap>;
```
t3chguy, line 45: Why underscored?

t3chguy (approved):
Looks sane otherwise

src/rust-crypto/verification.ts
richvdh: https://www.merriam-webster.com/words-at-play/mary-norris-diaeresis

my use of it is tongue-in-cheek. Will replace if you find it distracting/confusing.

src/rust-crypto/verification.ts
t3chguy: Our code style is explicit here

> Private members are private and not prefixed unless required for naming conflicts.

https://github.com/vector-im/element-web/blob/develop/code_style.md#typescript--javascript

richvdh: yes, my question is do you want me to fix the other properties here or in a separate PR

t3chguy: Happy for it to be here, approval stands

richvdh: turns out the other properties *should* have underscores because there are getter methods which conflict

PR #3485, opened 2023-06-19
Improve logging of http requests to aid debugging

BillCarsonFr:
## Checklist

-   [ ] Tests written for new code (and old code if feasible)
-   [x] Linter and other CI checks pass
-   [ ] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))

Added simple logging for all requests made from the js-sdk to ease rageshakes investigation.

~~Looks like that:~~
```
FetchHttpApi: GET --> http://localhost:8080/_matrix/client/r0/sync?filter=0&timeout=30000&since=s26_3787_3_8_29_1_12_38_0_1
```

udpated as per to remove url params that could contain secrets
```
FetchHttpApi: PUT --> /_matrix/client/v3/sendToDevice/m.room.encrypted/b8f4c87559504be794f3e1496a8970a5
rageshake.ts:77 FetchHttpApi:  <-- 200 8ms /_matrix/client/v3/sendToDevice/m.room.encrypted/b8f4c87559504be794f3e1496a8970a5
rageshake.ts:77 FetchHttpApi: PUT --> /_matrix/client/r0/rooms/!ZTAJxYLImTUucVDSZW%3Alocalhost%3A8480/typing/%40alice%3Alocalhost%3A8480
rageshake.ts:77 FetchHttpApi:  <-- 200 10ms /_matrix/client/r0/rooms/!ZTAJxYLImTUucVDSZW%3Alocalhost%3A8480/typing/%40alice%3Alocalhost%3A8480
rageshake.ts:77 FetchHttpApi:  <-- 200 2662ms /_matrix/client/r0/sync
rageshake.ts:77 FetchHttpApi: GET --> /_matrix/client/r0/sync
rageshake.ts:77 FetchHttpApi: PUT --> /_matrix/client/r0/rooms/!ZTAJxYLImTUucVDSZW%3Alocalhost%3A8480/typing/%40alice%3Alocalhost%3A8480
rageshake.ts:77 FetchHttpApi:  <-- 200 28ms /_matrix/client/r0/rooms/!ZTAJxYLImTUucVDSZW%3Alocalhost%3A8480/typing/%40alice%3Alocalhost%3A8480
rageshake.ts:77 FetchHttpApi:  <-- 200 247ms /_matrix/client/r0/sync
rageshake.ts:77 FetchHttpApi: GET --> /_matrix/client/r0/sync
[...]

PR #3591, opened 2023-07-12
Log query parameters on HTTP requests

MadLittleMods:
Log query parameters on HTTP requests

Follow-up to https://github.com/matrix-org/matrix-js-sdk/pull/3485

**Before:**

```
FetchHttpApi: --> GET https://server:8448/some/path
FetchHttpApi: <-- GET https://server:8448/some/path [1234ms 200]
```

**After:**

```
FetchHttpApi: --> GET https://server:8448/some/path?query=xxx
FetchHttpApi: <-- GET https://server:8448/some/path?query=xxx [1234ms 200]
```

All query parameters values are sanitized to `xxx` but the existence of the query parameter is helpful to distinguish between requests and a peace of mind sanity check to see that they are on the request. Like initial vs incremental `/sync` with `?since=xxx`. In the future, it would be nice to log more values.

### Dev notes

```
yarn jest spec/unit/http-api/fetch.spec.ts
```

## Checklist

-   [ ] ~~Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))~~

---

### DIFF
--- a/spec/unit/http-api/fetch.spec.ts
+++ b/spec/unit/http-api/fetch.spec.ts
@@ -300,7 +300,7 @@ describe("FetchHttpApi", () => {
         const fetchFn = jest.fn().mockReturnValue(deferred.promise);
         jest.spyOn(logger, "debug").mockImplementation(() => {});
         const api = new FetchHttpApi(new TypedEventEmitter<any, any>(), { baseUrl, prefix, fetchFn });
-        const prom = api.requestOtherUrl(Method.Get, "https://server:8448/some/path#fragment?query=param");
+        const prom = api.requestOtherUrl(Method.Get, "https://server:8448/some/path?query=param#fragment");
         jest.advanceTimersByTime(1234);
         deferred.resolve({ ok: true, status: 200, text: () => Promise.resolve("RESPONSE") } as Response);
         await prom;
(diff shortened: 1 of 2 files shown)


src/http-api/fetch.ts
```diff
@@ -301,17 +301,22 @@ export class FetchHttpApi<O extends IHttpOpts> {
...
             }
-            // get just the path to remove any potential url param that could have
-            // some potential secrets
-            return asUrl.origin + asUrl.pathname;
+            // Remove the values of any URL params that could contain potential secrets
+            const sanitizedQs = new URLSearchParams();
+            for (const key of asUrl.searchParams.keys()) {
+                sanitizedQs.append(key, "xxx");
+            }
+            const sanitizedQsUrlPiece = sanitizedQs.toString() ? `?${sanitizedQs.toString()}` : "";
```
andybalaam, line 317: I think calling `toString` twice here probably actually does the work of converting to a string twice, which seems a bit of a waste.

andybalaam (approved):
One minor improvement, but looks good (and safe) overall, thanks!

andybalaam:
Btw the Cypress tests are failing for unrelated reasons, so happy to merge when you're ready (ping me).

src/http-api/fetch.ts
MadLittleMods: Updated to only stringify once :fast_forward:

spec/unit/http-api/fetch.spec.ts
```diff
@@ -300,7 +300,7 @@ describe("FetchHttpApi", () => {
         const fetchFn = jest.fn().mockReturnValue(deferred.promise);
```
MadLittleMods, line 300: > happy to merge when you're ready (ping me).

@andybalaam Feel free to merge whenever :slightly_smiling_face:

andybalaam:
Thank you!

PR #3592, opened 2023-07-12
Element-R: check that UIA callbacks return sensible data

richvdh:
There seems to be a bug somewhere where one of our UIA callbacks returns `auth: null` and hence triggers https://github.com/matrix-org/synapse/issues/15871.

I hope this will help track it down.

---

### DIFF
--- a/src/rust-crypto/OutgoingRequestProcessor.ts
+++ b/src/rust-crypto/OutgoingRequestProcessor.ts
@@ -29,7 +29,7 @@ import {
 import { logger } from "../logger";
 import { IHttpOpts, MatrixHttpApi, Method } from "../http-api";
 import { QueryDict } from "../utils";
-import { IAuthDict, UIAuthCallback } from "../interactive-auth";
+import { AuthDict, UIAuthCallback } from "../interactive-auth";
 import { UIAResponse } from "../@types/uia";
 
 /**
@@ -116,7 +116,11 @@ export class OutgoingRequestProcessor {
         }
 
         const parsedBody = JSON.parse(body);
-        const makeRequest = async (auth: IAuthDict): Promise<UIAResponse<T>> => {
+        const makeRequest = async (auth: AuthDict): Promise<UIAResponse<T>> => {
+            // Auth dicts must have a 'type', contrary to what the type of 'AuthDict' would have you believe
+            if (!auth || !(auth as any).type) {
+                throw new Error("UIA callback returned invalid data");
+            }
             const newBody = {
                 ...parsedBody,
                 auth,
--- a/spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
+++ b/spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
@@ -29,7 +29,7 @@ import {
 } from "@matrix-org/matrix-sdk-crypto-js";
 
 import { TypedEventEmitter } from "../../../src/models/typed-event-emitter";
-import { HttpApiEvent, HttpApiEventHandlerMap, MatrixHttpApi, UIAuthCallback } from "../../../src";
+import { AuthDict, HttpApiEvent, HttpApiEventHandlerMap, MatrixHttpApi, UIAuthCallback } from "../../../src";
 import { OutgoingRequestProcessor } from "../../../src/rust-crypto/OutgoingRequestProcessor";
 
 describe("OutgoingRequestProcessor", () => {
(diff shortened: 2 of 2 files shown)


src/rust-crypto/OutgoingRequestProcessor.ts
```diff
@@ -116,7 +116,11 @@ export class OutgoingRequestProcessor {
         }
 
         const parsedBody = JSON.parse(body);
-        const makeRequest = async (auth: IAuthDict): Promise<UIAResponse<T>> => {
+        const makeRequest = async (auth: AuthDict): Promise<UIAResponse<T>> => {
+            // Auth dicts must have a 'type', contrary to what the type of 'AuthDict' would have you believe
```
t3chguy, line 120: Can we fix the type to chase the bug out then..?

richvdh: I started that and couldn't make head nor tail of how it was working in the react sdk. https://github.com/matrix-org/matrix-react-sdk/blob/cfd48b36aa0ca210157ea44a93d6d5f1963062a8/src/components/views/auth/InteractiveAuthEntryComponents.tsx#L861 is setting it to `{}` and I have no idea if/how that works.

t3chguy: Think I found it:

```patch
Index: src/interactive-auth.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/interactive-auth.ts b/src/interactive-auth.ts
--- a/src/interactive-auth.ts	(revision 1cb5fff5a17771ef0135a70653926ecbfe03c378)
+++ b/src/interactive-auth.ts	(date 1689160473254)
@@ -163,7 +163,7 @@
  *
  */
-export type UIAuthCallback<T> = (makeRequest: (authData: IAuthDict) => Promise<UIAResponse<T>>) => Promise<T>;
+export type UIAuthCallback<T> = (makeRequest: (authData: IAuthDict | null) => Promise<UIAResponse<T>>) => Promise<T>;
 
 interface IOpts<T> {
     /**
Index: src/rust-crypto/OutgoingRequestProcessor.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/rust-crypto/OutgoingRequestProcessor.ts b/src/rust-crypto/OutgoingRequestProcessor.ts
--- a/src/rust-crypto/OutgoingRequestProcessor.ts	(revision 1cb5fff5a17771ef0135a70653926ecbfe03c378)
+++ b/src/rust-crypto/OutgoingRequestProcessor.ts	(date 1689160525460)
@@ -116,11 +116,13 @@
         }
 
         const parsedBody = JSON.parse(body);
-        const makeRequest = async (auth: IAuthDict): Promise<UIAResponse<T>> => {
-            const newBody = {
+        const makeRequest = async (auth: IAuthDict | null): Promise<UIAResponse<T>> => {
+            const newBody: Record<string, any> = {
                 ...parsedBody,
-                auth,
             };
+            if (auth !== null) {
+                newBody.auth = auth;
+            }
             const resp = await this.rawJsonRequest(method, path, queryParams, JSON.stringify(newBody));
             return JSON.parse(resp) as T;
         };
Index: src/crypto/EncryptionSetup.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/crypto/EncryptionSetup.ts b/src/crypto/EncryptionSetup.ts
--- a/src/crypto/EncryptionSetup.ts	(revision 1cb5fff5a17771ef0135a70653926ecbfe03c378)
+++ b/src/crypto/EncryptionSetup.ts	(date 1689160544351)
@@ -188,7 +188,7 @@
             // We must only call `uploadDeviceSigningKeys` from inside this auth
             // helper to ensure we properly handle auth errors.
             await this.crossSigningKeys.authUpload?.((authDict) => {
-                return baseApis.uploadDeviceSigningKeys(authDict, keys as CrossSigningKeys);
+                return baseApis.uploadDeviceSigningKeys(authDict ?? undefined, keys as CrossSigningKeys);
             });
 
             // pass the new keys to the main instance of our own CrossSigningInfo.

richvdh: > For the joys, the spec seems to claim the types are optional

This turns out to be correct. `type` *is* optional in some circumstances. In particular:

> Some authentication types may be completed by means other than through the Matrix client, for example, an email confirmation may be completed when the user clicks on the link in the email. In this case, the client retries the request with an **auth dict containing only the session key**. 

(From https://spec.matrix.org/v1.7/client-server-api/#user-interactive-api-in-the-rest-api).

richvdh: > Think I found it:

I'm afraid I'm not following this.

richvdh:
closing in favour of #3594 which @t3chguy has persuaded me is better

PR #3593, opened 2023-07-12
Make some types packages regular dependencies

dbkr:
Due to the fact that we import packages like 'uuid' in files that users of the library import, these types essentially end up being part of our public library interface (to TypeScript consumers). However, some of these libraries don't come with their own types and we import the definitelytyped module, but this is a dev dependency for us so the types are not available when js-sdk is installed as a regular dependency. This means type checks on that application will fail.

I think the correct solution here (other than those libraries having their own types) is to just make the types packages regular dependencies.

This PR does that for the two types packages that element call was failing because of. I haven't added the rest as they might not all be part of external facing bits of the library (whatever that is now with js-sdk).

## Checklist

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
         "another-json": "^0.2.0",
         "bs58": "^5.0.0",
         "content-type": "^1.0.4",
+        "@types/content-type": "^1.1.5",
         "jwt-decode": "^3.1.2",
         "loglevel": "^1.7.1",
         "matrix-events-sdk": "0.0.1",
@@ -67,7 +68,8 @@
         "p-retry": "4",
         "sdp-transform": "^2.14.1",
         "unhomoglyph": "^1.0.6",
-        "uuid": "9"
+        "uuid": "9",
+        "@types/uuid": "9"
     },
     "devDependencies": {
         "@babel/cli": "^7.12.10",
@@ -85,13 +87,11 @@
         "@casualbot/jest-sonar-reporter": "^2.2.5",
         "@matrix-org/olm": "https://gitlab.matrix.org/api/v4/projects/27/packages/npm/@matrix-org/olm/-/@matrix-org/olm-3.2.14.tgz",
         "@types/bs58": "^4.0.1",
-        "@types/content-type": "^1.1.5",
         "@types/debug": "^4.1.7",
         "@types/domexception": "^4.0.0",
         "@types/jest": "^29.0.0",
         "@types/node": "18",
         "@types/sdp-transform": "^2.4.5",
-        "@types/uuid": "9",
         "@typescript-eslint/eslint-plugin": "^5.45.0",
         "@typescript-eslint/parser": "^5.45.0",
         "allchange": "^1.0.6",


t3chguy:
@dbkr hmm what interfaces use uuid types?

image

dbkr:
Ah, In think this is probably due to the fact that we're importing typescript source directly. The errors we're getting are https://github.com/vector-im/element-call/actions/runs/5522662505/jobs/10072557088?pr=1173 so I guess this is a question of whether it's expected to need to work if that's how you're consuming the library?

t3chguy:
@dbkr but matrix-react-sdk imports the TS directly and doesn't depend on `@types/uuid`

```
 ~/W/matrix-react-sdk   t3chguy/types111 *$+  yarn why @types/uuid       Wed Jul 12 11:52:41 2023
yarn why v1.22.19
[1/4] 🤔  Why do we have the module "@types/uuid"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
error We couldn't find a match!
✨  Done in 0.20s.
```

dbkr:
Yeah, I was going to just import the types in element call, but it felt wrong to have to manually copy imports from the js-sdk. If that's the better answer though, I'm also happy to do that.

dbkr:
OK, it still feels like we might want to consider allowing people to consume js-sdk as a TypeScript library without having to yarn link to get devDependencies installed, but we'll do whatever is supported.

daniel-abramov:
Just some context (note, that I'm not too much into JavaScript/TypeScript, so I might have misunderstood something, @dbkr please correct me if I'm wrong) based on our meeting sync-up: from what I understand - Element Call uses `matrix-js-sdk` "in a wrong way" which it does because of some problems in `matrix-js-sdk` that did not allow us to use it "in a proper way" (missing type definitions or something like this).

I.e. I've just tried to replace all our usages of imports like:

```javascript
import { createClient, ICreateClientOpts } from "matrix-js-sdk/src/matrix";
```

with:

```javascript
import { createClient, ICreateClientOpts } from "matrix-js-sdk";
```

Which I believe is the correct way of using TypeScript packages. After doing this change Element Call canno be built and exhibits around 250 errors when doing `tsc` that range from errors like this:

```
node_modules/matrix-js-sdk/spec/browserify/sync-browserify.spec.ts:17:25 - error TS2307: Cannot find module 'matrix-mock-request' or its corresponding type declarations.

17 import HttpBackend from "matrix-mock-request";

node_modules/matrix-js-sdk/spec/integ/crypto/olm-encryption-spec.ts:89:82 - error TS7006: Parameter '_path' implicitly has an 'any' type.

89     aliTestClient.httpBackend.when("POST", "/keys/claim").respond(200, function (_path, content: IClaimKeysRequest) {
                                                                                    ~~~~~

node_modules/matrix-js-sdk/spec/integ/crypto/olm-encryption-spec.ts:225:62 - error TS7006: Parameter '_path' implicitly has an 'any' type.

225         httpBackend.when("PUT", path).respond(200, function (_path, content) {
                                                                 ~~~~~

node_modules/matrix-js-sdk/spec/integ/crypto/verification.spec.ts:282:24 - error TS2551: Property 'advanceTimersByTimeAsync' does not exist on type 'typeof jest'. Did you mean 'advanceTimersByTime'?

282             await jest.advanceTimersByTimeAsync(10);
```

to errors like this:

```javascript
src/otel/OTelCall.ts:19:10 - error TS2614: Module '"matrix-js-sdk"' has no exported member 'CallEvent'. Did you mean to use 'import CallEvent from "matrix-js-sdk"' instead?

19 import { CallEvent } from "matrix-js-sdk";
```

Perhaps we need to address this issue at some point - it feels like we're doing something wrong and it has caused certain issues and confusion over the past few weeks (specifically when strict lints got enabled in Element Call), not to mention that it feels a bit unergonomic.

t3chguy:
@daniel-abramov https://github.com/matrix-org/matrix-js-sdk/issues/3506 is the issue for things like CallEvent

PR #3594, opened 2023-07-12
Fix sending `auth: null` due to broken types around UIA

t3chguy:
Closes https://github.com/matrix-org/matrix-js-sdk/pull/3592

---

## 🐛 Bug Fixes
 * Fix sending `auth: null` due to broken types around UIA ([\#3594](https://github.com/matrix-org/matrix-js-sdk/pull/3594)).

### DIFF
--- a/src/interactive-auth.ts
+++ b/src/interactive-auth.ts
@@ -159,7 +159,8 @@ export class NoAuthFlowFoundError extends Error {
  * The type of an application callback to perform the user-interactive bit of UIA.
  *
  * It is called with a single parameter, `makeRequest`, which is a function which takes the UIA parameters and
- * makes the HTTP request.
+ * makes the HTTP request. The `authData` parameter in `makeRequest` can be set to null to omit the `auth` field
+ * from the UIA request.
  *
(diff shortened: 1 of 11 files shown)


src/interactive-auth.ts
```diff
@@ -163,7 +163,7 @@ export class NoAuthFlowFoundError extends Error {
  *
  * The generic parameter `T` is the type of the response of the endpoint, once it is eventually successful.
  */
-export type UIAuthCallback<T> = (makeRequest: (authData: IAuthDict) => Promise<UIAResponse<T>>) => Promise<T>;
+export type UIAuthCallback<T> = (makeRequest: (authData: IAuthDict | null) => Promise<UIAResponse<T>>) => Promise<T>;
```
richvdh, line 166: please can we document what it means when `authData` is null? What is `makeRequest`'s behaviour in that case? (And why do we have `UIAuthCallback` implementations that set it to `null` rather than a real `IAuthDict`?)

richvdh (changes requested):

richvdh (approved):
thanks!

PR #3597, opened 2023-07-12
Export typed event emitter key types

### LINKED ISSUE #3506: Expose `EmittedEvents` subtypes

t3chguy:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/3506

---

## 🐛 Bug Fixes
 * Export typed event emitter key types ([\#3597](https://github.com/matrix-org/matrix-js-sdk/pull/3597)). Fixes #3506.

### DIFF
--- a/src/matrix.ts
+++ b/src/matrix.ts
@@ -73,7 +73,6 @@ export {
 export type { GroupCall } from "./webrtc/groupCall";
 export { CryptoEvent } from "./crypto";
 export { SlidingSyncEvent } from "./sliding-sync";
-export { VerificationRequestEvent, VerifierEvent } from "./crypto-api/verification";
 export { MediaHandlerEvent } from "./webrtc/mediaHandler";
 export { CallEvent } from "./webrtc/call";
 export { CallFeedEvent } from "./webrtc/callFeed";
--- a/src/matrix.ts
+++ b/src/matrix.ts
@@ -63,9 +63,23 @@ export * as SecretStorage from "./secret-storage";
 export type { ICryptoCallbacks } from "./crypto"; // used to be located here
 export { createNewMatrixCall } from "./webrtc/call";
 export type { MatrixCall } from "./webrtc/call";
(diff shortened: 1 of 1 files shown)


src/matrix.ts
```diff
@@ -63,9 +63,23 @@ export * as SecretStorage from "./secret-storage";
...
+    GroupCallEvent,
+    GroupCallIntent,
+    GroupCallState,
+    GroupCallType,
+    GroupCallStatsReportEvent,
+} from "./webrtc/groupCall";
 export type { GroupCall } from "./webrtc/groupCall";
 export { CryptoEvent } from "./crypto";
+export { SlidingSyncEvent } from "./sliding-sync";
+export { VerificationRequestEvent, VerifierEvent } from "./crypto-api/verification";
```
richvdh, line 76: these are already exposed via `export * as Crypto` below (and are documented as such in the jsdoc). Please don't clutter the main namespace more than it has to be.

richvdh (changes requested):

src/matrix.ts
richvdh: The "rule", at least in my head, is that crypto-related stuff should be contained within the `Crypto` namespace. `CryptoEvent` predates my recent work in this area so hasn't (yet) been moved. (And if it was moved, we'd still have to export it at the top level for backwards compat.)

t3chguy, line 76:
```suggestion
```

richvdh (approved):
thanks!

PR #3598, opened 2023-07-12
Don't update calls with no livekit URL

dbkr:
Expose method to update it instead and generally simplify a bit: change it to a single string rather than an array of structs.

## Checklist

---

## 🐛 Bug Fixes
 * Don't update calls with no livekit URL ([\#3598](https://github.com/matrix-org/matrix-js-sdk/pull/3598)).

### DIFF
--- a/src/webrtc/groupCall.ts
+++ b/src/webrtc/groupCall.ts
@@ -17,7 +17,7 @@ import { Room } from "../models/room";
 import { RoomStateEvent } from "../models/room-state";
 import { logger } from "../logger";
 import { ReEmitter } from "../ReEmitter";
-import { FocusInfo, SDPStreamMetadataPurpose } from "./callEventTypes";
+import { SDPStreamMetadataPurpose } from "./callEventTypes";
 import { MatrixEvent } from "../models/event";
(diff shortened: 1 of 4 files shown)


src/webrtc/groupCall.ts
```diff
@@ -338,19 +343,20 @@ export class GroupCall extends TypedEventEmitter<
...
     }
 
-    private _foci: FocusInfo[] = [];
+    public get livekitServiceURL(): string | undefined {
+        return this._livekitServiceURL;
+    }
 
-    public get foci(): FocusInfo[] {
-        return this._foci;
+    public updateLivekitServiceURL(newURL: string): Promise<void> {
```
robintown, line 357: Is this used anywhere?

robintown (commented):
This appears to only set the URL on creation, but didn't we also discuss wanting to update it on join?

src/webrtc/groupCall.ts
robintown: Oh I see, it's public and can be used in Element Call

robintown (approved):

PR #3600, opened 2023-07-12
Fix read receipt sending behaviour around thread roots

t3chguy:
Should fix issues like

> ```
> POST /_matrix/client/r0/rooms/{room_id}/receipt/m.read/{event_id} 400
>
> M_INVALID_PARAM: MatrixError: [400] event_id $foo is not related to thread $bar
> ```

image

---

## 🐛 Bug Fixes
 * Fix read receipt sending behaviour around thread roots ([\#3600](https://github.com/matrix-org/matrix-js-sdk/pull/3600)).

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -5004,6 +5004,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             const isThread = !!event.threadRootId && !event.isThreadRoot;
             body = {
                 ...body,
+                // Only thread replies should define a specific thread. Thread roots can only be read in the main timeline.
                 thread_id: isThread ? event.threadRootId : MAIN_ROOM_TIMELINE,
             };
         }
--- a/src/client.ts
+++ b/src/client.ts
@@ -5000,7 +5000,8 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         });
 
         if (!unthreaded) {
-            const isThread = !!event.threadRootId;
+            // A thread cannot be just a thread root and a thread root can only be read in the main timeline
+            const isThread = !!event.threadRootId && !event.isThreadRoot;
             body = {
                 ...body,
                 thread_id: isThread ? event.threadRootId : MAIN_ROOM_TIMELINE,


clokep (approved):
Seems reasonable to me -- we definitely want to consider a thread root on the main timeline.

src/client.ts
```diff
@@ -5000,7 +5000,8 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         });
 
         if (!unthreaded) {
-            const isThread = !!event.threadRootId;
+            // A thread cannot be just a thread root and a thread root can only be read in the main timeline
+            const isThread = !!event.threadRootId && !event.isThreadRoot;
```
MadLittleMods, line 5004: I think it might be nice if we had a `event.isThreadReply` boolean that handled this complexity in a single place

src/client.ts
```diff
@@ -5000,7 +5000,8 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
             body = {
                 ...body,
                 thread_id: isThread ? event.threadRootId : MAIN_ROOM_TIMELINE,
```
MadLittleMods, line 5007: I feel like this is a little more clear:

```suggestion
                // Only thread replies should define a specific thread. Thread roots can only be read in the main timeline.
                thread_id: isThread ? event.threadRootId : MAIN_ROOM_TIMELINE,
```

src/client.ts
t3chguy: We do, `isRelation(RelationType.Thread)` but that doesn't include 2nd order relations which we do want to include here. Reactions to thread responses aren't "thread replies" yet still want RRs

MadLittleMods: Whatever we want to call this logic `!!event.threadRootId && !event.isThreadRoot;` feels like it could be nicely encapsulated with a description of the intracies.

"thread replies" still feels applicable and a good enough name for reactions in this scenario but if there is a better name, that would be nice. I don't know how reactions to a thread root exactly works though. And those kind of details could be nicely explained there.

Johennes:
> Seems reasonable to me -- we definitely want to consider a thread root on the main timeline.

Does https://spec.matrix.org/latest/client-server-api/#threaded-read-receipts not say the opposite?

>  An event is considered to be “in a thread” if [...] it has child events with a `rel_type` of `m.thread` (in which case it’d be the thread root).
> [...]
> Events not in a thread but still in the room are considered to be part of the “main timeline”, or a special thread with an ID of `main`.

andybalaam:
> > Seems reasonable to me -- we definitely want to consider a thread root on the main timeline.
> 
> Does https://spec.matrix.org/latest/client-server-api/#threaded-read-receipts not say the opposite?
> 
> > An event is considered to be “in a thread” if [...] it has child events with a `rel_type` of `m.thread` (in which case it’d be the thread root).
> > [...]
> > Events not in a thread but still in the room are considered to be part of the “main timeline”, or a special thread with an ID of `main`.

My understanding of that is that the thread root is in _both_ timelines. My original draft of https://github.com/matrix-org/matrix-spec-proposals/pull/4033 proposed that it should be in `main` only but I took that out when I reduced the scope to just ordering.

andybalaam (approved):
Yes, thread roots should be understood to be in the main timeline, but we need to clarify this in the spec, and with Synapse devs that that is what they have implemented.

t3chguy:
Happy to undo the change if we deem the spec to be accurate. We'd then need to change the interface to have the app signal which thread to send the RR to rather than the js-sdk just guessing based on the event as it does today. And Synapse would need updating to not throw errors when we RR roots in threads. I don't mind either way which way the design goes here

PR #3601, opened 2023-07-13
ElementR: Add `CryptoApi.findVerificationRequestDMInProgress`

florianduros:
## Checklist

### DIFF
--- a/spec/integ/crypto/verification.spec.ts
+++ b/spec/integ/crypto/verification.spec.ts
@@ -22,7 +22,7 @@ import fetchMock from "fetch-mock-jest";
 import { IDBFactory } from "fake-indexeddb";
 import { createHash } from "crypto";
 
-import { createClient, CryptoEvent, ICreateClientOpts, MatrixClient } from "../../../src";
+import { Category, createClient, CryptoEvent, ICreateClientOpts, MatrixClient } from "../../../src";
 import {
     canAcceptVerificationRequest,
     ShowQrCodeCallbacks,
@@ -34,7 +34,14 @@ import {
     VerifierEvent,
 } from "../../../src/crypto-api/verification";
 import { escapeRegExp } from "../../../src/utils";
-import { CRYPTO_BACKENDS, emitPromise, InitCrypto } from "../../test-utils/test-utils";
+import {
+    CRYPTO_BACKENDS,
+    emitPromise,
+    getSyncResponse,
+    InitCrypto,
+    ROOM_ID,
+    syncPromise,
+} from "../../test-utils/test-utils";
 import { SyncResponder } from "../../test-utils/SyncResponder";
 import {
     MASTER_CROSS_SIGNING_PUBLIC_KEY_BASE64,
@@ -84,6 +91,7 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
     // oldBackendOnly is an alternative to `it` or `test` which will skip the test if we are running against the
     // Rust backend. Once we have full support in the rust sdk, it will go away.
     const oldBackendOnly = backend === "rust-sdk" ? test.skip : test;
+    const newBackendOnly = backend !== "rust-sdk" ? test.skip : test;
 
     /** the client under test */
     let aliceClient: MatrixClient;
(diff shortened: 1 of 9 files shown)


src/rust-crypto/rust-crypto.ts
```diff
@@ -94,15 +95,14 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
 
         /** The local user's User ID. */
         private readonly userId: string,
-
```
richvdh, line 97: please can you not remove these blank lines? it makes it harder to read the comments imho.

src/rust-crypto/rust-crypto.ts
```diff
@@ -94,15 +95,14 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
         /** The local user's Device ID. */
         _deviceId: string,
```
richvdh, line 99: we don't seem to need this... maybe we should consider removing it.

src/client.ts
```diff
@@ -2241,13 +2241,19 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
 
         this.cryptoBackend = rustCrypto;
 
         // attach the event listeners needed by RustCrypto
         this.on(RoomMemberEvent.Membership, rustCrypto.onRoomMembership.bind(rustCrypto));
+        // listen only key verification request
+        // other events are processed during the /sync
+        this.on(RoomEvent.Timeline, (event) => {
+            rustCrypto.onKeyVerificationRequest(event);
+        });
```
richvdh, line 2256: if we're going to feed all events into it, shouldn't it be called `onTimelineEvent` ?

src/rust-crypto/rust-crypto.ts
```diff
@@ -583,9 +583,30 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
-        // TODO
-        return;
+    public findVerificationRequestDMInProgress(roomId: string): VerificationRequest | undefined {
+        const room = this.store.getRoom(roomId);
+
+        if (!room) throw new Error(`unknown roomId ${roomId}`);
+
+        const members = room.getMembers();
+
+        // Going through all members and return the first verification request we can find
```
richvdh, line 593:
```suggestion
        // Go through all members and return the first verification request we can find
```

src/rust-crypto/rust-crypto.ts
```diff
@@ -583,9 +583,30 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
      */
-    public findVerificationRequestDMInProgress(roomId: string): undefined {
-        // TODO
-        return;
+    public findVerificationRequestDMInProgress(roomId: string): VerificationRequest | undefined {
+        const room = this.store.getRoom(roomId);
+
+        if (!room) throw new Error(`unknown roomId ${roomId}`);
+
+        const members = room.getMembers();
```
richvdh, line 591: Ironically, this method is only called in one place, where we know *exactly* which user we are interested in:

https://github.com/matrix-org/matrix-react-sdk/blob/develop/src/verification.ts#L123

Let's: 
 * add a second optional parameter `userId?: string` to  `findVerificationRequestDMInProgress`. If both `roomId` and `userId` are provided, `findVerificationRequestDMInProgress` only returns requests in that room for that user.
 * Update legacy `Crypto.findVerificationRequestDMInProgress` accordingly
 * deprecate calling `findVerificationRequestDMInProgress` without the second parameter
 * update react-sdk to pass the second parameter
 * change this code to just `raise new Error` if the second parameter is not passed
 * get rid of the IStore awfulness
 * profit

richvdh (changes requested):

src/client.ts
florianduros: Except that we don't want to process all the timeline event. They are already processed in the sync response.

I did it in my fist impl and the event was processed twice.

src/crypto-api.ts
```diff
@@ -250,9 +250,20 @@ export interface CryptoApi {
...
      * @returns the VerificationRequest that is in progress, if any
+     * @deprecated prefer `userId` parameter variant.
      */
     findVerificationRequestDMInProgress(roomId: string): VerificationRequest | undefined;
 
+    /**
+     * Finds a DM verification request that is already in progress for the given room id.
+     *
+     * @param roomId - the room to use for verification.
+     * @param userId - search the verification request for the given user.
```
richvdh, line 261:
```suggestion
     * @param userId - search for a verification request for the given user.
```

src/rust-crypto/index.ts
```diff
@@ -43,6 +45,7 @@ export async function initRustCrypto(
     secretStorage: ServerSideSecretStorage,
     cryptoCallbacks: ICryptoCallbacks,
     storePrefix: string | null,
+    store: IStore,
```
richvdh, line 48: no longer needed?

src/models/event.ts
```diff
@@ -1657,6 +1657,17 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
...
+    /**
+     * True if:
+     *  - The event type is {@link EventType.RoomMessage}
+     *  - The `msgType` of the event content is {@link MsgType.KeyVerificationRequest}
+     *
+     * @returns true if the event is key verification request
+     */
+    public isKeyVerificationRequest(): boolean {
+        return this.getType() === EventType.RoomMessage && this.getContent().msgtype === MsgType.KeyVerificationRequest;
+    }
```
richvdh, line 1670: I'd suggest you inline this. Or make it a utility function in `rust-crypto/index.ts`. I don't really think it's useful enough to form part of the public API for `Event`.

spec/integ/crypto/verification.spec.ts
```diff
@@ -800,6 +801,72 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
...
+            expect(request?.isSelfVerification).toBe(false);
+            expect(request?.otherUserId).toBe("@bob:xyz");
+        });
+
+        // Only the rust-crypto raises an error when the room id is unknown
+        newBackendOnly("Unknown RoomId", () => {
+            expect(() => aliceClient.getCrypto()!.findVerificationRequestDMInProgress(TEST_ROOM_ID)).toThrow(
+                `unknown roomId ${TEST_ROOM_ID}`,
+            );
+        });
```
richvdh, line 854: this might be more appropriate as a unit test

spec/integ/crypto/verification.spec.ts
```diff
@@ -800,6 +801,72 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
...
+
+        it("Verification request unfound", async () => {
```
richvdh, line 856:
```suggestion
        it("Verification request not found", async () => {
```

richvdh (commented):
Generally looks good, a few points.

I'm slightly confused about when exactly we are supposed to call `receiveVerificationEvent` so am waiting for poljar to help out

richvdh (changes requested):

richvdh (dismissed):

src/rust-crypto/rust-crypto.ts
```diff
@@ -100,7 +100,6 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
 
         /** Interface to server-side secret storage */
         private readonly secretStorage: ServerSideSecretStorage,
-
```
richvdh, line 103: please don't remove this blank line, either

t3chguy (approved):
Looks sane

PR #3610, opened 2023-07-19
Make sure to drop references to user device lists

richvdh:
Empirically, this seems to fix some problems with leaking references to IndexedDB.

Fixes: https://github.com/vector-im/element-web/issues/25817
Fixes: https://github.com/vector-im/element-web/issues/25779 I *hope*

---


florianduros (approved):
OH.

Good catch.  Imo, it's a bit *alarming* to have to manually free the memory.

richvdh:
> Imo, it's a bit alarming to have to manually free the memory.

It really is. And it makes me wonder how many other places this will be a problem.

PR #3613, opened 2023-07-20
Avoid running cypress tests outside of merge queue

t3chguy:
---


kerryarchibald (approved):
Seems sensible 🐌

PR #3618, opened 2023-07-25
Fix threads ending up with chunks of their timelines missing

t3chguy:
Fixes https://github.com/vector-im/element-web/issues/24466
Reverts a small part of https://github.com/matrix-org/matrix-js-sdk/pull/3541

---

## 🐛 Bug Fixes
 * Fix threads ending up with chunks of their timelines missing ([\#3618](https://github.com/matrix-org/matrix-js-sdk/pull/3618)). Fixes vector-im/element-web#24466.


germain-gg (approved):
ESLint is grumpy, otherwise looks good to me! 
Thank you for looking into this 🙏

PR #3619, opened 2023-07-25
ElementR: Add `CryptoApi.getCrossSigningKeyId`

florianduros:
## Checklist

---

## ✨ Features
 * ElementR: Add `CryptoApi.getCrossSigningKeyId` ([\#3619](https://github.com/matrix-org/matrix-js-sdk/pull/3619)). Contributed by @florianduros.

### DIFF
--- a/spec/integ/crypto/cross-signing.spec.ts
+++ b/spec/integ/crypto/cross-signing.spec.ts
@@ -318,7 +318,7 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("cross-signing (%s)", (backend: s
         it("should return the cross signing key id for each cross signing key", async () => {
             mockSetupCrossSigningRequests();
 
-            // Intercept cross singing keys upload
+            // Intercept cross signing keys upload
             const crossSigningKeysPromise = awaitCrossSigningKeysUpload();
 
             // provide a UIA callback, so that the cross-signing keys are uploaded
--- a/spec/integ/crypto/cross-signing.spec.ts
+++ b/spec/integ/crypto/cross-signing.spec.ts
@@ -21,6 +21,7 @@ import { IDBFactory } from "fake-indexeddb";
 import { CRYPTO_BACKENDS, InitCrypto } from "../../test-utils/test-utils";
 import { createClient, IAuthDict, MatrixClient } from "../../../src";
 import { mockSetupCrossSigningRequests } from "../../test-utils/mockEndpoints";
+import { CrossSigningKey } from "../../../src/crypto-api";
 
 afterEach(() => {
     // reset fake-indexeddb after each test, to make sure we don't leak connections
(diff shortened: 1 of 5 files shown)


src/crypto-api.ts
```diff
@@ -479,6 +480,17 @@ export enum CrossSigningKey {
     UserSigning = "user_signing",
 }
 
+/**
+ * Cross signing key
+ * More information https://spec.matrix.org/v1.7/client-server-api/#post_matrixclientv3keysdevice_signingupload
```
richvdh, line 485: image


```suggestion
 *
 * @see https://spec.matrix.org/v1.7/client-server-api/#post_matrixclientv3keysdevice_signingupload
```

richvdh (changes requested):
There's slightly more to it than this. The api doc for `CryptoApi.getCrossSigningKeyId` says:

> If cross-signing has been initialised on this device, the ID of the given key. Otherwise, null

... which I think means we should call `OlmMachine.crossSigningStatus()` to check that it has been initialised.

And also we should probably call `OwnUserIdentity::is_verified` to check that the  public key matches the private key?

florianduros:
> 



> There's slightly more to it than this. The api doc for `CryptoApi.getCrossSigningKeyId` says:
> 
> > If cross-signing has been initialised on this device, the ID of the given key. Otherwise, null
> 
> ... which I think means we should call `OlmMachine.crossSigningStatus()` to check that it has been initialised.
> 
> And also we should probably call `OwnUserIdentity::is_verified` to check that the public key matches the private key?
Oh, I thought that if we had an `UserIdentity`, it means that the cross signing has been initialized

src/rust-crypto/rust-crypto.ts
```diff
@@ -422,8 +422,13 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
-        if (!userIdentity) {
+        const crossSigningStatus: RustSdkCryptoJs.CrossSigningStatus = await this.olmMachine.crossSigningStatus();
+        const privateKeysOnDevice =
+            crossSigningStatus.hasMaster && crossSigningStatus.hasUserSigning && crossSigningStatus.hasSelfSigning;
+
+        // The public or private keys are not available on this device
+        // Or the current user is not verified
+        if (!userIdentity?.isVerified() || !privateKeysOnDevice) {
             return null;
         }
```
richvdh: This is a bit hard to follow
```suggestion
        if (!userIdentity || !privateKeysOnDevice) {
            // The public or private keys are not available on this device
            return null;
        }
        
        if (!userIdentity.isVerified()) {
            // We have both public and private keys, but they don't match!
            return null;
        }
```

spec/integ/crypto/cross-signing.spec.ts
```diff
@@ -288,4 +288,55 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("cross-signing (%s)", (backend: s
...
+                    // Override the routes define in `mockSetupCrossSigningRequests`
+                    { overwriteRoutes: true },
+                );
+            });
+        }
+
+        it("should return the cross signing key id for each cross signing key", async () => {
+            mockSetupCrossSigningRequests();
+
+            // Intercept cross singing keys upload
```
richvdh, line 321: :musical_note: 

```suggestion
            // Intercept cross signing keys upload
```

richvdh (approved):
It would be nice to have tests for these edge cases, but I guess it's fine for now.

PR #3623, opened 2023-07-26
Element-R: Stub interface for secure key backup

richvdh:
Based on #3624 

I realised that the BackupManager interface is only used within the js-sdk, so I have made it properly internal.

Closes https://github.com/vector-im/crypto-internal/issues/105

---

### DIFF
--- a/src/crypto-api.ts
+++ b/src/crypto-api.ts
@@ -137,6 +137,22 @@ export interface CryptoApi {
     getDeviceVerificationStatus(userId: string, deviceId: string): Promise<DeviceVerificationStatus | null>;
 
     /**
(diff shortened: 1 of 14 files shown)


src/crypto-api.ts
```diff
@@ -532,3 +532,4 @@ export interface GeneratedSecretStorageKey {
 
 export * from "./crypto-api/verification";
 export * from "./crypto-api/keybackup";
+export type { SecureKeyBackup } from "./common-crypto/SecureKeyBackup";
```
richvdh, line 535:
```suggestion
```

richvdh:
Ugh, I don't think I want to do this after all.

PR #3634, opened 2023-07-28
Ensure we don't overinflate the total notification count

t3chguy:
By correctly comparing push rules before & after decryption
I suggest reviewing commit-by-commit

Fixes https://github.com/vector-im/element-web/issues/25803

---

## 🐛 Bug Fixes
 * Ensure we don't overinflate the total notification count ([\#3634](https://github.com/matrix-org/matrix-js-sdk/pull/3634)). Fixes vector-im/element-web#25803.


germain-gg (approved):
Looks sane!

PR #3635, opened 2023-07-28
Fix age field check in event echo processing

stas-demydiuk:
Address an edge case when the server does not return the `age` field for some reason (as it seems like it is optional according to spec) to avoid `NaN` value in the event's `localTimestamp` field

Fixes https://github.com/vector-im/element-web/issues/25875

## Checklist

---

## 🐛 Bug Fixes
 * Fix age field check in event echo processing ([\#3635](https://github.com/matrix-org/matrix-js-sdk/pull/3635)). Fixes vector-im/element-web#25875. Contributed by @stas-demydiuk.


germain-gg (changes requested):
Looks sane to me, thank you for taking care of this! 

Would you be able to reference the piece of the spec in the pull request that mentions it is optional and add a test to validate the behaviour you're covering now?

Besides

isosphere:
> Looks sane to me, thank you for taking care of this!
> 
> Would you be able to reference the piece of the spec in the pull request that mentions it is optional and add a test to validate the behaviour you're covering now?

If you check https://spec.matrix.org/v1.11/client-server-api/#room-event-format

Scroll down a little to `ClientEvent`, it says of the entire `unsigned` field: 

> Contains optional extra information about the event.

Everything that is unsigned is optional.

richvdh:
Related: https://github.com/element-hq/synapse/issues/8429

dbkr (approved):
Looks like the author hasn't come back on this one, although given it's a simple enough change we think it's valuable enough to merge. A regression test would have been great, bit that's on us for not asking before.

PR #3638, opened 2023-07-31
Bump pagination limit to account for threaded events

t3chguy:
For https://github.com/vector-im/element-web/issues/25873

---

## ✨ Features
 * Bump pagination limit to account for threaded events ([\#3638](https://github.com/matrix-org/matrix-js-sdk/pull/3638)).


germain-gg (approved):
Looks good, same question as in https://github.com/matrix-org/matrix-react-sdk/pull/11342 regarding the timing difference between the two calls.

t3chguy:
@germain-gg this isn't modifying any calls - just changing how many calls we make before giving up on pagination. This code was made before threads and hidden events were a thing. Both of those can make us give up on pagination when there is actual history to find

PR #3643, opened 2023-07-31
ElementR: Add `CryptoApi.requestVerificationDM`

florianduros:
## Checklist

---

## ✨ Features
 * ElementR: Add `CryptoApi.requestVerificationDM` ([\#3643](https://github.com/matrix-org/matrix-js-sdk/pull/3643)). Contributed by @florianduros.


dbkr (approved):

src/rust-crypto/verification.ts
```diff
@@ -677,7 +677,7 @@ export class RustSASVerifier extends BaseRustVerifer<RustSdkCryptoJs.Sas> implem
 }
 
 /** For each specced verification method, the rust-side `VerificationMethod` corresponding to it */
-const verificationMethodsByIdentifier: Record<string, RustSdkCryptoJs.VerificationMethod> = {
+export const verificationMethodsByIdentifier: Record<string, RustSdkCryptoJs.VerificationMethod> = {
```
richvdh, line 680: use `verificationMethodIdentifierToMethod` instead of making this public?

spec/integ/crypto/verification.spec.ts
```diff
@@ -808,6 +810,115 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
         });
     });
 
+    describe("Send verification request in DM", () => {
+        const crossSingingKeys: Partial<IDownloadKeyResult> = {
```
richvdh, line 814: :notes: 

```suggestion
        const crossSigningKeys: Partial<IDownloadKeyResult> = {
```

spec/integ/crypto/verification.spec.ts
```diff
@@ -808,6 +810,115 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
...
+                "ed25519:bob_device": "PNsWknI+bXfeeV7GFfnuBX3sATD+/riD0gvvv8Ynd5s",
+            },
+            user_id: "@bob:xyz",
+            signatures: {
+                "@bob:xyz": {
+                    "ed25519:bob_device":
+                        "T+UUApQlgZHNfM4gdkcrJcjTmeF7wEYAZH2OS99gKKizEgEFL89acAC7yuLfm2MI5J9Ks2AGkQJp3jjNjC/kBA",
+                },
+            },
+        };
```
richvdh, line 873: Hrm. can we use the `test_data` instead of this?

spec/integ/crypto/verification.spec.ts
```diff
@@ -808,6 +810,115 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
...
+                    "express:/_matrix/client/v3/rooms/:roomId/send/m.room.message/:txId",
+                    (url: string, options: RequestInit) => {
+                        resolve(JSON.parse(options.body as string));
+                        return { event_id: "$YUwRidLecu:example.com" };
+                    },
+                );
+            });
+        }
+
+        newBackendOnly("alice send verification request in a DM to bob", async () => {
```
richvdh, line 901:
```suggestion
        newBackendOnly("alice sends a verification request in a DM to bob", async () => {
```

richvdh: doesn't this work on the old backend?

richvdh (commented):
looks good in general

spec/integ/crypto/verification.spec.ts
florianduros: The old backend fails because one of the device is not verified. Is it something we want here ?

spec/integ/crypto/verification.spec.ts
florianduros: `test_data` is about `alice` and I need keys for `bob` here. (I tested to just replace bob's username and device by Alice's username and device but it doesn't work)

spec/integ/crypto/verification.spec.ts
richvdh: it would certainly give me more confidence in the tests if we could make it work on both stacks.

I'm a bit surprised by the error: isn't the whole *point* that we haven't verified Bob's devices?

richvdh (approved):
awesomeness :heart:

PR #3645, opened 2023-08-01
Mark more logs as debug to tidy the console

t3chguy:
---

### DIFF
--- a/src/crypto/OlmDevice.ts
+++ b/src/crypto/OlmDevice.ts
@@ -1119,7 +1119,7 @@ export class OlmDevice {
                                 }
                             }
 
-                            logger.info(
+                            logger.debug(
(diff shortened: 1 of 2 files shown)


src/crypto/OlmDevice.ts
```diff
@@ -1119,7 +1119,7 @@ export class OlmDevice {
                                 }
                             }
 
-                            logger.info(
+                            logger.debug(
```
germain-gg, line 1122: Why is this not needed in the rageshakes? I'm not one to debug crypto issues, but that somewhat looks relevant to them?

t3chguy: This doesn't affect whether or not it goes into rageshakes. It *only* affects whether or not a default console configuration will show it or not.

germain-gg (approved):

robintown (approved):

PR #3646, opened 2023-08-01
ElementR: Update `CryptoApi.userHasCrossSigningKeys`

florianduros:
## Checklist

---

## ✨ Features
 * ElementR: Update `CryptoApi.userHasCrossSigningKeys` ([\#3646](https://github.com/matrix-org/matrix-js-sdk/pull/3646)). Contributed by @florianduros.


src/client.ts
```diff
@@ -2628,7 +2629,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      *
      * @returns the cross signing information for the user.
      */
-    public getStoredCrossSigningForUser(userId: string): CrossSigningInfo | null {
+    public getStoredCrossSigningForUser(userId: string): Promise<CrossSigningInfo | null> {
```
richvdh, line 2632: this is a breaking change; we can't do this. We'll need to make a new method and deprecate the old one.

src/crypto-api/CrossSigningInfo.ts
```diff
@@ -0,0 +1,31 @@
...
+*/
+
+import { CrossSigningKey } from "../crypto-api";
+
+/**
+ * Information about a user's cross-signing keys
+ */
+export interface CrossSigningInfo {
+    /**
+     * Return the public key of the cross singing key
```
richvdh, line 24:
```suggestion
     * Return the public part of  one of the cross signing keys
```

src/crypto-api/CrossSigningInfo.ts
```diff
@@ -0,0 +1,31 @@
...
+     *
+     * @param crossSigningKey - the type of cross signing key wanted
```
richvdh, line 26: It is an optional argument. What happens if it is not set?

src/crypto-api/CrossSigningInfo.ts
```diff
@@ -0,0 +1,31 @@
...
+     *
+     * @returns
```
richvdh, line 28: either write some text here, or remove the tag, please.

src/crypto-api/CrossSigningInfo.ts
```diff
@@ -0,0 +1,31 @@
...
+     */
+    getId(crossSigningKey?: CrossSigningKey): string | null;
```
richvdh, line 30: Since we're creating a new interface anyway, maybe we can make a better name for it? `getPublicKey` maybe?

src/client.ts
richvdh: (this will also mean you don't need to change `QRCode.ts` I think

src/rust-crypto/CrossSigningInfo.ts
```diff
@@ -0,0 +1,98 @@
...
+
+import { CrossSigningInfo as ICrossSigningInfo } from "../crypto-api/CrossSigningInfo";
+import { CrossSigningKey } from "../crypto-api";
+import { IHttpOpts, MatrixHttpApi, Method } from "../http-api";
+import { IDownloadKeyResult, Keys } from "../client";
+
+/**
+ * Implementation of {@link ICrossSigningInfo}
+ */
+export class CrossSigningInfo implements ICrossSigningInfo {
```
richvdh, line 25: maybe call this `CrossSigningInfoImpl` to distinguish it from the interface

src/rust-crypto/CrossSigningInfo.ts
```diff
@@ -0,0 +1,98 @@
...
+
+        const keyResult = await http.authedRequest<IDownloadKeyResult>(
+            Method.Post,
+            "/_matrix/client/v3/keys/query",
+            undefined,
+            queryBody,
+            {
+                prefix: "",
+            },
+        );
```
richvdh, line 50: given the method is called `get`**`Stored`**`CrossSigningForUser`, it seems odd that we should pull this from the server.

I think we need to call `OlmMachine::getIdentity`

richvdh (changes requested):

src/crypto-api/CrossSigningInfo.ts
```diff
@@ -0,0 +1,31 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { CrossSigningKey } from "../crypto-api";
+
+/**
+ * Information about a user's cross-signing keys
+ */
+export interface CrossSigningInfo {
```
richvdh, line 22: I would have put this in `crypto-api.ts`, but if you're going to put it here, please re-export it from `crypto-api.ts` so that it is correctly documented at https://pr3646--js-sdk-docs-previews.netlify.app/stable/modules/crypto.

richvdh: Also: can we find a name for it that distinguishes it from [`CrossSigningKeyInfo`](https://github.com/matrix-org/matrix-js-sdk/blob/0a69f6ecc540b8ff58333c487d5ad684c793ddb8/src/crypto-api.ts#L583) ?

src/crypto-api.ts
```diff
@@ -177,6 +178,15 @@ export interface CryptoApi {
      */
     getCrossSigningKeyId(type?: CrossSigningKey): Promise<string | null>;
 
+    /**
+     * Get the cross signing information for a given user.
```
richvdh, line 182:
```suggestion
     * Get the public cross-signing keys for a given user.
```

src/crypto-api/CrossSigningInfo.ts
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * Return the public part of  one of the cross signing keys
```
richvdh, line 24:
```suggestion
     * Return the public part of one of the cross signing keys
```

src/crypto-api.ts
```diff
@@ -177,6 +178,15 @@ export interface CryptoApi {
...
+     *
+     * @param userId - the user ID to get the cross-signing info for.
+     *
+     * @returns the cross signing information for the user.
```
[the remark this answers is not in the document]
richvdh: what does a return of `null` mean, and how does it differ from a `CrossSigningInfo` whose `getPublicKey` returns null?

src/crypto-api/CrossSigningInfo.ts
```diff
@@ -0,0 +1,31 @@
...
+     *
+     * @param crossSigningKey - the type of cross signing key wanted. {@link CrossSigningKey.Master} is the default value.
+     *
+     * @returns the public key.
```
richvdh, line 28: what does a `null` result mean?

src/rust-crypto/CrossSigningInfoImpl.ts
```diff
@@ -0,0 +1,98 @@
...
+
+        const getPubKey = (keys: Keys): string => {
+            // `keys` is an object with { [`ed25519:${pubKey}`]: pubKey }
+            // We assume only a single key, and we want the bare form without type
+            // prefix, so we select the values.
+            return Object.values(keys.keys)[0];
+        };
+
+        if (!keyResult.master_keys?.[userId]) return null;
+        const masterKey = getPubKey(keyResult.master_keys[userId]);
```
richvdh, line 60: any of `master_keys`, `self_signing_keys` or `user_signing_keys` could be empty, so `master_keys[userId]` may be undefined. This will throw, I think.

spec/unit/rust-crypto/CrossSigningInfo.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+                        },
+                    },
+                },
+            });
+
+            const crossSigningInfo = await CrossSigningInfoImpl.create(TEST_USER_ID, httpApi);
+            expect(crossSigningInfo).toBeNull();
+        });
+
+        it("should return null when the user signing keys are missing", async () => {
```
richvdh, line 61: It is *normal* for the `user_signing_keys` property to be *present*, but empty. The user-signing key is for the user's own use, and is not normally shared with other users.

So I think (a)  this test is testing the wrong thing; (b) it is expecting the wrong result.

richvdh (changes requested):

src/rust-crypto/rust-crypto.ts
```diff
@@ -247,15 +247,33 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
-        const userIdentity = await this.olmMachine.getIdentity(userId);
-        return userIdentity !== undefined;
+    public async userHasCrossSigningKeys(userId = this.userId, downloadUncached = false): Promise<boolean> {
+        const rustTrackedUsers: Set<RustSdkCryptoJs.UserId> = await this.olmMachine.trackedUsers();
+        const rustTrackedUser = Array.from(rustTrackedUsers).find((rustUserId) => userId === rustUserId.toString());
+
+        if (rustTrackedUser) {
+            /* make sure we have an *up-to-date* idea of the user's cross-signing keys. This is important, because if we
+             * return "false" here, we will end up generating new cross-signing keys and replacing the existing ones.
+             */
```
richvdh, line 257: I think this is only important for our own user (ie, if `userId === this.userId`). At least, the comment doesn't make any sense for other users.

src/rust-crypto/rust-crypto.ts
```diff
@@ -247,15 +247,33 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
-         */
-        const request = this.olmMachine.queryKeysForUsers([userId]);
-        await this.outgoingRequestProcessor.makeOutgoingRequest(request);
-        const userIdentity = await this.olmMachine.getIdentity(userId);
-        return userIdentity !== undefined;
+    public async userHasCrossSigningKeys(userId = this.userId, downloadUncached = false): Promise<boolean> {
+        const rustTrackedUsers: Set<RustSdkCryptoJs.UserId> = await this.olmMachine.trackedUsers();
+        const rustTrackedUser = Array.from(rustTrackedUsers).find((rustUserId) => userId === rustUserId.toString());
+
+        if (rustTrackedUser) {
```
richvdh, line 254: implicit cast to bool here. needs more `!== undefined`

src/crypto-api.ts
```diff
@@ -39,14 +39,19 @@ export interface CryptoApi {
     globalBlacklistUnverifiedDevices: boolean;
 
     /**
-     * Checks if the user has previously published cross-signing keys
+     * Check if the cross signing keys for the user are available.
```
richvdh, line 42:
```suggestion
     * Check if the given user has published cross-signing keys.
```

src/crypto-api.ts
```diff
@@ -39,14 +39,19 @@ export interface CryptoApi {
...
      *
-     * This means downloading the devicelist for the user and checking if the list includes
-     * the cross-signing pseudo-device.
+     * - If the user is tracked, a `/keys/query` request is made to update locally the cross signing keys.
+     * - If the user is not tracked locally and downloadUncached is set at true,
```
richvdh, line 45:
```suggestion
     * - If the user is not tracked locally and downloadUncached is set to true,
```

src/crypto-api.ts
```diff
@@ -39,14 +39,19 @@ export interface CryptoApi {
...
+     *   a `/keys/query` request is made to the server to retrieve the cross signing keys.
+     * - Otherwise, return false
      *
-     * @returns true if the user has previously published cross-signing keys
+     * @param userId - the user ID to check
```
richvdh, line 49: should document the default here

src/crypto-api.ts
```diff
@@ -39,14 +39,19 @@ export interface CryptoApi {
...
+     * @param downloadUncached - If true, download the cross signing keys.
```
richvdh, line 50: Let's use the same wording as `getUserDeviceInfo` here.

```suggestion
     * @param downloadUncached - If true, download the device list for users whose device list we are not
     *    currently tracking. Defaults to false, in which case `false` will be returned for such users.
```

src/crypto-api.ts
```diff
@@ -39,14 +39,19 @@ export interface CryptoApi {
...
+     *
+     * @returns true if the cross signing keys are available.
      */
-    userHasCrossSigningKeys(): Promise<boolean>;
+    userHasCrossSigningKeys(userId?: string, downloadUncached?: boolean): Promise<boolean>;
```
richvdh, line 54: I think we should move this next to `getUserDeviceInfo`, since it has quite similar functionality

richvdh (commented):
looks good, some minor tweaks which I can do

src/crypto-api.ts
richvdh:
```suggestion
     * @param userId - the user ID to check. Defaults to the local user.
```

src/rust-crypto/rust-crypto.ts
richvdh:
```suggestion
        if (rustTrackedUser !== undefined) {
```

PR #3647, opened 2023-08-02
Allow knocking rooms

charlynguyen:
We @nordeck are currently implementing the knock rooms behind the feature flag `feature_ask_to_join` (introduced in https://github.com/matrix-org/matrix-react-sdk/pull/11182).

This pull request allows knocking rooms.

## Checklist


src/@types/requests.ts
```diff
@@ -45,6 +45,18 @@ export interface IJoinRoomOpts {
     viaServers?: string[];
 }
 
+export interface IKnockRoomOpts {
```
t3chguy, line 48: We no longer use the I prefix on interfaces - see https://github.com/vector-im/element-web/blob/develop/code_style.md#typescript--javascript

> Interface names should not be marked with an uppercase I.

t3chguy (approved):
Looks sane, thanks

PR #3649, opened 2023-08-03
ElementR: Track untracked users in `CryptoApi.getUserDeviceInfo`

florianduros:
## Checklist

`getUserDeviceInfo` was downloading the device list of the untrack users but didn't add them to the `olmMachine`.

---

## 🐛 Bug Fixes
 * ElementR: Track untracked users in `CryptoApi.getUserDeviceInfo` ([\#3649](https://github.com/matrix-org/matrix-js-sdk/pull/3649)). Contributed by @florianduros.

### DIFF
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -585,6 +585,27 @@ describe("RustCrypto", () => {
             );
         });
     });
+
+    describe("getUserDeviceInfo", () => {
+        it("Add unknown users to tracked users", async () => {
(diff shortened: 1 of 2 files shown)


dbkr (approved):

richvdh (changes requested):
I'm worried about this. It feels like a very non-obvious thing for `getUserDeviceInfo` to do, *especially* when `downloadUncached` is false.

Adding users to the "tracked" list when we don't share a room with them is dangerous, because we won't receive updates for their device list, so may end up using out-of-date data.

Can you explain why this change is needed?

richvdh:
I think this has been superceded by https://github.com/matrix-org/matrix-js-sdk/pull/3646

PR #3650, opened 2023-08-04
ElementR: Process all verification events, not just requests

florianduros:
## Checklist

---

## 🐛 Bug Fixes
 * ElementR: Process all verification events, not just requests ([\#3650](https://github.com/matrix-org/matrix-js-sdk/pull/3650)). Contributed by @florianduros.

### DIFF
--- a/src/rust-crypto/rust-crypto.ts
+++ b/src/rust-crypto/rust-crypto.ts
@@ -54,8 +54,8 @@ import { secretStorageContainsCrossSigningKeys } from "./secret-storage";
 import { keyFromPassphrase } from "../crypto/key_passphrase";
 import { encodeRecoveryKey } from "../crypto/recoverykey";
 import { crypto } from "../crypto/crypto";
-import { RustVerificationRequest, verificationMethodIdentifierToMethod } from "./verification";
-import { EventType, MsgType } from "../@types/event";
+import { isVerificationEvent, RustVerificationRequest, verificationMethodIdentifierToMethod } from "./verification";
+import { EventType } from "../@types/event";
 import { CryptoEvent } from "../crypto";
 import { TypedEventEmitter } from "../models/typed-event-emitter";
 import { RustBackupManager } from "./backup";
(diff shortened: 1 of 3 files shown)


src/rust-crypto/rust-crypto.ts
```diff
@@ -1040,15 +1040,12 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
      * @param event - live event
      */
     public async onLiveEventFromSync(event: MatrixEvent): Promise<void> {
-        // Ignore state event
-        if (event.isState()) return;
+        // Ignore state event or remote echo
+        if (event.isState() || event.getUnsigned().transaction_id === null) return;
```
richvdh, line 1044: why would a remote echo have a transaction ID of `null` ?

src/rust-crypto/verification.ts
```diff
@@ -700,3 +702,31 @@ export function verificationMethodIdentifierToMethod(method: string): RustSdkCry
...
+    switch (event.getType()) {
+        case EventType.KeyVerificationCancel:
+        case EventType.KeyVerificationDone:
+        case EventType.KeyVerificationMac:
+        case EventType.KeyVerificationStart:
+        case EventType.KeyVerificationKey:
+        case EventType.KeyVerificationReady:
+        case EventType.KeyVerificationAccept: {
+            return true;
+        }
```
richvdh, line 724: are the curly braces here (and below) required by our linter? I find them a bit distracting.

```suggestion
        case EventType.KeyVerificationAccept:
            return true;

```

src/rust-crypto/verification.ts
```diff
@@ -700,3 +702,31 @@ export function verificationMethodIdentifierToMethod(method: string): RustSdkCry
     }
     return meth;
 }
+
+/**
+ * Return true if the event is a verification event
```
richvdh, line 707: Let's be a bit more specific about what we are matching here.

```suggestion
 * Return true if the event's type matches that of an in-room verification event
```

src/rust-crypto/rust-crypto.ts
florianduros: I changed, it was a misunderstanding of mine

src/rust-crypto/rust-crypto.ts
```diff
@@ -1041,7 +1041,8 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
      */
     public async onLiveEventFromSync(event: MatrixEvent): Promise<void> {
         // Ignore state event or remote echo
-        if (event.isState() || event.getUnsigned().transaction_id === null) return;
+        // transaction_id is provided in case of remote echo {@link https://spec.matrix.org/v1.7/client-server-api/#local-echo}
+        if (event.isState() || "transaction_id" in event.getUnsigned()) return;
```
richvdh, line 1045: although this is technically correct, in other places where we make this check, we look for truthiness of `transaction_id`. In other words:

```suggestion
        if (event.isState() || !!event.getUnsigned().transaction_id) return;
```

Let's be consistent.

richvdh (approved):
lgtm otherwise

PR #3633, opened 2023-07-27
Implement `CryptoApi.checkKeyBackupAndEnable`

richvdh:
Closes https://github.com/vector-im/crypto-internal/issues/111
Closes https://github.com/vector-im/crypto-internal/issues/112

~~Based on #3632~~

---
Here's what your changelog entry will look like:

## ✨ Features
 * Implement `CryptoApi.checkKeyBackupAndEnable` ([\#3633](https://github.com/matrix-org/matrix-js-sdk/pull/3633)). Fixes vector-im/crypto-internal#111 and vector-im/crypto-internal#112.

PR #3652, opened 2023-08-07
RustCrypto | Implement keybackup loop

BillCarsonFr:
Implements backup loop for rust crypto.

Fixes https://github.com/vector-im/crypto-internal/issues/113

Depends on:
- [x] Wasm bindings https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/19
- [x] Previous branch: https://github.com/matrix-org/matrix-js-sdk/pull/3633

## Checklist

---


src/rust-crypto/backup.ts
```diff
@@ -30,14 +32,26 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
     public constructor(
         private readonly olmMachine: OlmMachine,
         private readonly http: MatrixHttpApi<IHttpOpts & { onlyData: true }>,
+        private readonly outgoingRequestProcessor: OutgoingRequestProcessor,
     ) {
         super();
     }
 
+    /**
+     * Tell the RustBackupManager to immediately stop.
```
t3chguy, line 49: This comment doesn't quite match reality, it stops it on the next loop and may still process & emit events after this is called.

src/rust-crypto/backup.ts
```diff
@@ -166,9 +187,73 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+                            // wait for that and then continue?
+                            const waitTime = (<MatrixError>err).data.retry_after_ms;
+                            if (waitTime > 0) {
+                                sleep(waitTime);
+                                continue;
+                            } // else go to the normal backoff
+                        }
+                    }
+
+                    // Someo other error (mx, network, or CORS or invalid urls?) anyhow backoff
```
t3chguy, line 247: typo

src/rust-crypto/backup.ts
```diff
@@ -166,9 +187,73 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+                    await this.outgoingRequestProcessor.makeOutgoingRequest(request);
+                    numFailures = 0;
+
+                    const keyCount: RustSdkCryptoJs.RoomKeyCounts = await this.olmMachine.roomKeyCounts();
+                    const remaining = keyCount.total - keyCount.backedUp;
+                    this.emit(CryptoEvent.KeyBackupSessionsRemaining, remaining);
+                } catch (err) {
+                    numFailures++;
+                    logger.error("Error processing backup request for rust crypto-sdk", err);
+                    if ((<MatrixError>err).data) {
```
t3chguy, line 227: `err` in Typescript is unknown, you should be doing `err instanceof MatrixError` otherwise `null.data` will throw an exception

spec/test-utils/test-data/static-test-data.ts
```diff
@@ -0,0 +1,46 @@
+/*
```
t3chguy, line 1: Missing copyright

src/rust-crypto/backup.ts
BillCarsonFr: fixed

src/rust-crypto/backup.ts
BillCarsonFr: updated

spec/test-utils/test-data/static-test-data.ts
BillCarsonFr: Removed this file to use the generate-test-data.py script

t3chguy (approved):
Looks sane

spec/test-utils/test-data/generate-test-data.py
```diff
@@ -265,6 +280,35 @@ def sign_json(json_object: dict, private_key: ed25519.Ed25519PrivateKey) -> str:
 
     return signature_base64
 
+def build_exported_megolm_key() -> dict:
```
richvdh, line 283: docstring please, what does this do?

spec/test-utils/test-data/generate-test-data.py
```diff
@@ -167,6 +168,20 @@ def main() -> None:
...
+    },
+    { 
+        json.dumps(build_exported_megolm_key(), indent=8)
+    },
+];
+
+/** An exported megolm session */
+export const MEGOLM_SESSION_DATA: IMegolmSessionData = { 
+        json.dumps(build_exported_megolm_key(), indent=4)
+};
```
richvdh, line 184: do we really need this as well as MEGOLM_SESSION_DATA_ARRAY? can we just use `MEGOLM_SESSION_DATA_ARRAY[0]` ?

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -83,6 +84,58 @@ afterEach(() => {
...
 });
 
+enum MockKeyUploadEvent {
+    KeyUploaded = "KeyUploaded",
+}
+
+type MockKeyUploadEventHandlerMap = {
+    [MockKeyUploadEvent.KeyUploaded]: (roomId: string, sessionId: string, backupVersion: string) => void;
+};
+
```
richvdh, line 94: I am not a fan of the use of EventEmitter in tests like this. Is there any reason we can't just pass an `onBackupRequest` callback into `mockUploadEmitter` ?

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -83,6 +84,58 @@ afterEach(() => {
...
+/*
+ * Test helper. Will emit an event everytime fetchmock sees a request to backup a key.
```
richvdh, line 96:
```suggestion
 * Test helper. Returns an event emitter that will emit an event every time fetchmock sees a request to backup a key.
```

src/crypto/backup.ts
```diff
@@ -494,6 +498,10 @@ export class BackupManager {
                     return;
                 }
             }
+        } catch (err) {
+            // No one actually check errors on this promise, it's spawned internally.
```
richvdh, line 502:
```suggestion
            // No one actually checks errors on this promise, it's spawned internally.
```

src/crypto/backup.ts
```diff
@@ -474,6 +475,9 @@ export class BackupManager {
                             (<MatrixError>err).data.errcode == "M_NOT_FOUND" ||
                             (<MatrixError>err).data.errcode == "M_WRONG_ROOM_KEYS_VERSION"
                         ) {
+                            // fix do it now as the await check might trigger a new call to the loop before
```
richvdh, line 478: "fix do it now"?

src/rust-crypto/backup.ts
```diff
@@ -129,14 +145,10 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
                 await this.enableKeyBackup(backupInfo);
             } else if (activeVersion !== backupInfo.version) {
                 logger.log(`On backup version ${activeVersion} but found version ${backupInfo.version}: switching.`);
+                // This will remove any pending backup request, remove the backup key and reset the backup state of each room key we have.
                 await this.disableKeyBackup();
+                // Enabling will now trigger re-upload of all the keys
```
richvdh, line 150: does it? is that in the rust side?

spec/test-utils/test-data/generate-test-data.py
```diff
@@ -265,6 +280,35 @@ def sign_json(json_object: dict, private_key: ed25519.Ed25519PrivateKey) -> str:
 
     return signature_base64
 
+def build_exported_megolm_key() -> dict:
+    index = int(0)
+    private_key = ed25519.Ed25519PrivateKey.generate()
+    # Just use radom bytes for the ratchet parts
+    ratchet = randbytes(32 * 4)
```
richvdh, line 287: I'd prefer to avoid randomness in this script, so that when we re-run it, only important things change.

spec/integ/crypto/megolm-backup.spec.ts
BillCarsonFr: Mainly because I found it more easy to then create Promise that resolve when the event is emitted.
I tried the other way but the test was more annoying to write (and read), but I am not very proficient in writing promises

richvdh: mm fair enough. I probably wouldn't do this personally but I'm not sure I can come up with anything significantly better.

spec/test-utils/test-data/generate-test-data.py
```diff
@@ -167,6 +168,20 @@ def main() -> None:
 
 /** Signed backup data, suitable for return from `GET /_matrix/client/v3/room_keys/keys/{{roomId}}/{{sessionId}}` */
 export const SIGNED_BACKUP_DATA: KeyBackupInfo = { json.dumps(backup_data, indent=4) };
+
+export const MEGOLM_SESSION_DATA_ARRAY: IMegolmSessionData[] = [
```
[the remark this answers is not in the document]
BillCarsonFr: updated

spec/test-utils/test-data/generate-test-data.py
```diff
@@ -167,6 +168,20 @@ def main() -> None:
...
+    { 
+        json.dumps(build_exported_megolm_key(), indent=8)
+    },
+    { 
+        json.dumps(build_exported_megolm_key(), indent=8)
+    },
+];
```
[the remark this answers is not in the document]
BillCarsonFr: Thx, a lot better.

spec/test-utils/test-data/generate-test-data.py
BillCarsonFr: done

spec/test-utils/test-data/generate-test-data.py
BillCarsonFr: Added a static seed to have stable random at each generation

src/crypto/backup.ts
BillCarsonFr: done

spec/integ/crypto/megolm-backup.spec.ts
BillCarsonFr: done

src/crypto/backup.ts
BillCarsonFr: yeah the existing code as a problem, it's calling checkKeybackup() just there, and this will re-enable backup and start the loop;  but it will do nothing because sendingBackups is true, and anyhow it throws after and the finally reset the sendingBackups to false :/
=> I have a cleaner fix for it stashed but I am a bit reluctant to touch old code too much.
I'll push a commit after and let you review

src/rust-crypto/backup.ts
BillCarsonFr: `olmMachine.disableBackup()` reset the backed-up status of exiting keys, so next time you ask for the key to backup you will get all of the keys as candidates, so enabling again will re-upload

src/crypto/backup.ts
BillCarsonFr: done here => https://github.com/matrix-org/matrix-js-sdk/pull/3652/commits/cdde54af4eea0b4dfc5e4b0ddd8ce5150bb18550

spec/test-utils/test-data/generate-test-data.py
richvdh: can't we just hardcode the strings, rather than having to seed the random generator and call randbytes?

spec/test-utils/test-data/generate-test-data.py
```diff
@@ -265,6 +283,39 @@ def sign_json(json_object: dict, private_key: ed25519.Ed25519PrivateKey) -> str:
 
     return signature_base64
 
+def build_exported_megolm_key() -> dict:
+    """
+        Creates an exported megolm room key, as per https://gitlab.matrix.org/matrix-org/olm/blob/master/docs/megolm.md#session-export-format
+        that can be imported via importRoomKeys API.
```
richvdh, line 289:
```suggestion
    Creates an exported megolm room key, as per https://gitlab.matrix.org/matrix-org/olm/blob/master/docs/megolm.md#session-export-format
    that can be imported via importRoomKeys API.
```

richvdh (approved):
lgtm otherwise

spec/test-utils/test-data/generate-test-data.py
BillCarsonFr: Looks to me that it has best of the two words, it's stable and later on I can add more keys easilly test key upgrade and more

richvdh:
There was some confusion about why the cypress tests didn't run on this PR. So: for the record:

 1. The cypress tests only run in matrix-js-sdk as part of the merge queue (ie, after you press "Merge when ready"). This isn't *terribly* helpful, but the logic goes that (1) running them for every commit would be quite expensive in terms of Github Actions workers, and (2) the most likely tests to fail under Rust Crypto will be crypto-related(*), and we'll be running crypto-related cypress tests locally as part of development.

    *: recent experience shows that it's not actually the crypto test that fail, mostly because we *are* testing that carefully. It's all the other tests which are inherently flaky and subject to minor timing variations.

 2. The cypress tests *did* run, and fail, as pert of the merge queue on this PR. However, they are not currently a mandatory part of the merge queue, making them effectively useless. This is probably my fault, but in any case we need to get them back to a state where they are passing more often than not before we can change it.

PR #3653, opened 2023-08-08
Use types from OpenAPI Matrix C-S specification

t3chguy:
---


florianduros (approved):

richvdh (commented):
If we're going to do this, I'd like https://github.com/t3chguy/matrix-spec-npm to move into the `matrix-org` organisation. But no objections to this change.

t3chguy:
Closing as the types only giving request/response types and not types for e.g. the various Matrix Events makes this significantly less useful than first hoped.

PR #3663, opened 2023-08-18
Introduce MatrixRTCSession lower level group call primitive

dbkr:
WIP

 * Adds a layer that is just about managing the events for a call, without any actual calling
 * Switches to experimental new design for call members events, removing m.call events

## Checklist


src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -0,0 +1,358 @@
...
+    // A member joined, left, or updated a proprty of their membership
+    MembershipsChanged = "memberships_changed",
+    // We joined or left the session (our own local idea of whether we are joined, separate from MembershipsChanged)
+    JoinStateChanged = "join_state_changed",
+}
+
+export type MatrixRTCSessionEventHandlerMap = {
+    [MatrixRTCSessionEvent.MembershipsChanged]: (
+        newMemberships: CallMembership[],
+        oldMemberships: CallMembership[],
```
SimonBrandner, line 39: I think you've got the wrong order here...

andybalaam (commented):
Looks good from a quick look but I'll leave it to the VOIP team to do full review.

src/matrixrtc/CallMembership.ts
```diff
@@ -0,0 +1,95 @@
...
+    expires: number;
+    foci_active?: Focus[];
+}
+
+export class CallMembership {
+    public static equal(a: CallMembership, b: CallMembership): boolean {
+        return deepCompare(a.data, b.data);
+    }
+
+    public constructor(private parentEvent: MatrixEvent, private data: CallMembershipData) {
```
SimonBrandner, line 39: Could we please use an `opts` object here in case the amount of params grows?

src/matrixrtc/MatrixRTCSessonManager.ts
```diff
@@ -0,0 +1,130 @@
...
+    [MatrixRTCSessionManagerEvents.SessionEnded]: (roomId: string, session: MatrixRTCSession) => void;
+};
+
+export class MatrixRTCSessionManager extends TypedEventEmitter<MatrixRTCSessionManagerEvents, EventHandlerMap> {
+    // All the room-scoped sessions we know about. This will include any where the app
+    // has queried for the MatrixRTC sessions in a room, whether it's ever had any members
+    // or not)
+    private roomSessions = new Map<string, MatrixRTCSession>();
+
+    public constructor(private client: MatrixClient) {
```
SimonBrandner, line 43: Could we please use an `opts` object here in case the amount of params grows?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -0,0 +1,352 @@
...
+        const callMemberships = MatrixRTCSession.callMembershipsForRoom(room);
+
+        return new MatrixRTCSession(client, room, callMemberships);
+    }
+
+    private constructor(
+        private readonly client: MatrixClient,
+        public readonly room: Room,
+        public memberships: CallMembership[],
+    ) {
```
SimonBrandner, line 123: Could we please use an `opts` object here in case the amount of params grows?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    /**
+     * Announces this user and device as joined to the MatrixRTC session,
+     * and continues to update the membership event to keep it valid until
+     * leaveRoomSession() is called
+     * This will not subscribe to updates: remember to call subscribe() separately if
+     * desired.
+     */
+    public joinRoomSession(activeFoci: Focus[]): void {
```
SimonBrandner, line 150: Should we handle local echo in here somehow?

SimonBrandner (commented):
(/me hasn't managed to go through everything just yet...)

Requesting team review again as I feel like this should get more reviewers

src/matrixrtc/CallMembership.ts
dbkr: For this one I feel like if we're adding more params, something's gone wrong, although in any case it's not part o the library API so we can change it on a whim.

src/matrixrtc/MatrixRTCSessonManager.ts
dbkr: Again I'd argue since this isn't part of the library interface, it doesn't really matter.

src/matrixrtc/MatrixRTCSession.ts
dbkr: This is also not library interface so can change this if/when necessary.

src/matrixrtc/MatrixRTCSession.ts
dbkr: I had been wondering about this, but as yet nothing has needed it. I'm sort of inclined to say that the local client's idea of whether they're joined or not should just depend on whether they think they ought to be (is. `isJoined()`) which is effectively local echo but way simpler. Wdyt?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -0,0 +1,360 @@
...
+            expires: this.relativeExpiry,
+            foci_active: this.activeFoci,
+        };
+
+        if (prevMembership) m.created_ts = prevMembership.createdTs();
+
+        return m;
+    }
+
+    private updateCallMembershipEvent = async (): Promise<void> => {
```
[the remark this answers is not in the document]
dbkr: That should be a bit better.

src/matrixrtc/MatrixRTCSession.ts
SimonBrandner: Yeah, I guess it really depends on whether or not any of the UI depends on the `memberships`

dbkr: Yeah, if at some point it does then we can make it support local echo

SimonBrandner (approved):
LGTM, but I'd also like someone else to take a look

andybalaam (approved):

src/matrixrtc/MatrixRTCSessonManager.ts
```diff
@@ -0,0 +1,130 @@
...
+        sess.onMembershipUpdate();
+
+        const nowActive = sess.memberships.length > 0;
+
+        if (wasActive && !nowActive) {
+            this.emit(MatrixRTCSessionManagerEvents.SessionEnded, room.roomId, this.roomSessions.get(room.roomId)!);
+        } else if (!wasActive && nowActive) {
+            this.emit(MatrixRTCSessionManagerEvents.SessionStarted, room.roomId, this.roomSessions.get(room.roomId)!);
+        }
+    }
```
daniel-abramov, line 129: This function is the only place where we emit events about sessions being started and stopped from what I can see, so I assume that we must ensure that it always properly emits events.

I have a couple of notes regarding the topic of `hadSession`, let me know if they make sense to you:
1. `hadSession` seems to be `true` when the session is known to the session manager at the moment of calling this function and `false` otherwise. This means that `wasActive` will be `false` even if the session _was active_ unless the session was known to the session manager which makes the name `wasActive` a bit confusing. I think it's probably easier to split `wasActive` and `hadSession` (perhaps renaming it to `knownSession` or calling it `newSession` which will be an inverse of the current `hadSession`), otherwise `wasActive` and `nowActive` don't look symmetrical.
2. Given that `this.roomSessions` only grows during the lifetime of a session manager, won't it introduce some side effects in this function given that `hadSession` is there to specifically treat the case of session being known / not known to the session manager? (i.e. stopped sessions/sessions) with 0 members will remain on the map.

To me it feels like the most elegant to solve it would be to make sure we don't store expired sessions in the map (this may eliminate the need of the `hadSession`).

src/embedded.ts
```diff
@@ -241,6 +256,18 @@ export class RoomWidgetClient extends MatrixClient {
...
+        // the IOpenIDCredentials from the widget-api and IOpenIDToken form the matrix-js-sdk are compatible.
+        // we still recreate the token to make this transparent and catch'able by the linter in case the types change in the future.
+        return <IOpenIDToken>{
+            access_token: token.access_token,
+            expires_in: token.expires_in,
+            matrix_server_name: token.matrix_server_name,
+            token_type: token.token_type,
+        };
+    }
+
```
daniel-abramov, line 270: Was this change and the rest of the widget API (content loaded thing) changes intentional within the scope of this RP? 🙂

src/matrixrtc/MatrixRTCSessonManager.ts
```diff
@@ -0,0 +1,130 @@
+/*
```
daniel-abramov, line 1: There seems to be a typo: `MatrixRTCSessonManager` -> `MatrixRTCSessionManager`.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -0,0 +1,385 @@
...
+import { CallMembership, CallMembershipData } from "./CallMembership";
+import { Focus } from "./focus";
+import { MatrixEvent } from "../matrix";
+
+const MEMBERSHIP_EXPIRY_TIME = 60 * 60 * 1000;
+const MEMBER_EVENT_CHECK_PERIOD = 2 * 60 * 1000; // How often we check to see if we need to re-send our member event
+const CALL_MEMBER_EVENT_RETRY_DELAY_MIN = 3000;
+
+export enum MatrixRTCSessionEvent {
+    // A member joined, left, or updated a proprty of their membership
```
daniel-abramov, line 32:
```suggestion
    // A member joined, left, or updated a property of their membership.
```

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -0,0 +1,385 @@
...
+    MembershipsChanged = "memberships_changed",
+    // We joined or left the session (our own local idea of whether we are joined, separate from MembershipsChanged)
```
daniel-abramov, line 34:
```suggestion
    // We joined or left the session (our own local idea of whether we are joined, separate from MembershipsChanged).
```

OTOH: Do we want to differentiate between them? I.e. why does not membership suffice to designate our participation in a call?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -0,0 +1,385 @@
...
+import { EventTimeline } from "../models/event-timeline";
+import { Room } from "../models/room";
+import { MatrixClient } from "../client";
+import { EventType } from "../@types/event";
+import { CallMembership, CallMembershipData } from "./CallMembership";
+import { Focus } from "./focus";
+import { MatrixEvent } from "../matrix";
+
+const MEMBERSHIP_EXPIRY_TIME = 60 * 60 * 1000;
+const MEMBER_EVENT_CHECK_PERIOD = 2 * 60 * 1000; // How often we check to see if we need to re-send our member event
```
daniel-abramov, line 28: Why do we need to check it periodically by the way? Don't we have a finite set of clearly defined cases when we know that the event must be updated?

src/matrixrtc/CallMembership.ts
```diff
@@ -0,0 +1,95 @@
...
+        if (typeof data.expires !== "number") throw new Error("Malformed membership: expires must be numeric");
+        if (typeof data.device_id !== "string") throw new Error("Malformed membership event: device_id must be string");
+        if (typeof data.call_id !== "string") throw new Error("Malformed membership event: call_id must be string");
+        if (typeof data.scope !== "string") throw new Error("Malformed membership event: scope must be string");
```
daniel-abramov, line 43: Perhaps a dumb question from my side: Aren't the fields typed? - I see the types defined? (We're using TypeScript, right?) Or is it because currently, we allow type mismatches on the SDK level (no strictness?)

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -0,0 +1,385 @@
...
+        public readonly room: Room,
+        public memberships: CallMembership[],
+    ) {
+        super();
+        this.setExpiryTimer();
+    }
+
+    public isJoined(): boolean {
+        return this.relativeExpiry !== undefined;
+    }
```
daniel-abramov, line 135: What does "joined" mean in this context? - I'm asking because a matrix RTC session is a session / on-going call with N participants/members. I would assume that the session is "joined" if there is at least one participant in it, "not joined" otherwise. Or did I misunderstand the meaning of a `MatrixRTCSession`?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -0,0 +1,385 @@
...
+};
+
+/**
+ * A MatrixRTCSession manages the membership & properties of a MatrixRTC session.
+ * This class doesn't deal with media at all, just membership & properties of a session.
+ */
+export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, MatrixRTCSessionEventHandlerMap> {
+    // How many ms after we joined the call, that our membership should expire, or undefined
+    // if we're not yet joined
+    private relativeExpiry: number | undefined;
```
daniel-abramov, line 53: Sorry if the question is dumb, but does not it not mean that it's not a `MatrixRTCSession`, but rather something like a `LocalMatrixRTCSession`? (i.e. "our" part of the participation in a session instead of a `MatrixRTCSession`?)

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -0,0 +1,385 @@
...
+    public static roomSessionForRoom(client: MatrixClient, room: Room): MatrixRTCSession {
+        const callMemberships = MatrixRTCSession.callMembershipsForRoom(room);
+
+        return new MatrixRTCSession(client, room, callMemberships);
+    }
+
+    private constructor(
+        private readonly client: MatrixClient,
+        public readonly room: Room,
+        public memberships: CallMembership[],
```
daniel-abramov, line 127: As far as I can see, we never use these fields. Should we have them as an input parameter then?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -0,0 +1,385 @@
...
+        if (this.isJoined()) {
+            logger.info(`Already joined to session in room ${this.room.roomId}: ignoring join call`);
+            return;
+        }
+
+        logger.info(`Joining call session in room ${this.room.roomId}`);
+        this.activeFoci = activeFoci;
+        this.relativeExpiry = MEMBERSHIP_EXPIRY_TIME;
+        this.emit(MatrixRTCSessionEvent.JoinStateChanged, true);
+        this.updateCallMembershipEvent();
```
daniel-abramov, line 165: This is an `async` function, we don't call `await` here. Is it intentional?

daniel-abramov (commented):
This PR is quite large (in terms of newly introduced concepts), I believe it could have been split into smaller chunks for ease of verification. I have not checked everything, but I briefly checked most parts except for the tests and the `MatrixRTCSession` implementation (I checked the public interface and functions that it calls briefly though), and left some comments.

src/embedded.ts
dbkr: Oh, this will have been ported in from the livekit branch. It probably should have gone via develop first ideally rather than be part of this PR.

src/matrixrtc/CallMembership.ts
dbkr: So this is from where we're doing `getContent()` (line 73) and declaring that variable as a `CallMembershipData[]`. We don't actually know what is in the event: it could be anything, so we do need to validate it at some point. I went back & forth a few times on how to do this: we could have this just accept a `Record<string, unknown>` which is more like what the event data actually is and may be considered more correct, but it felt more helpful to define the right shape & then validate it.

src/matrixrtc/MatrixRTCSession.ts
dbkr: We do, this just seemed slightly simpler to write as a first cut rather than having to work out the time until the first one expired.

src/matrixrtc/MatrixRTCSession.ts
dbkr: I've added to the comment a bit: I also wondered this but it felt useful to distinguish whether we intend to be joined to the call and whether our membership event has actually gone through, and this seems like an easier, simpler way of doing local echo.

src/matrixrtc/MatrixRTCSession.ts
dbkr: Well, it represents both since it holds information of what other participants are in the session as well as whether we're joined, so I don't think I'd say it was specific to just our participation. That is, if we called it `LocalMatrixRTCSession`, we wouldn't have something else called `MatrixRTCSession` or `RemoteMatrixRTCSession` I think.

src/matrixrtc/MatrixRTCSession.ts
dbkr: Probably needed a comment, sorry - does that help?

src/matrixrtc/MatrixRTCSession.ts
dbkr: It is, I've added a comment. The reason is mostly because it retries, so when `updateCallMembershipEvent()` returns doesn't necessarily mean anything at all it could have just failed to send the event.

src/matrixrtc/MatrixRTCSessonManager.ts
dbkr: Hah, thank you - well spotted.

src/matrixrtc/MatrixRTCSessonManager.ts
```diff
@@ -0,0 +1,130 @@
...
+type EventHandlerMap = {
+    [MatrixRTCSessionManagerEvents.SessionStarted]: (roomId: string, session: MatrixRTCSession) => void;
+    [MatrixRTCSessionManagerEvents.SessionEnded]: (roomId: string, session: MatrixRTCSession) => void;
+};
+
+export class MatrixRTCSessionManager extends TypedEventEmitter<MatrixRTCSessionManagerEvents, EventHandlerMap> {
+    // All the room-scoped sessions we know about. This will include any where the app
+    // has queried for the MatrixRTC sessions in a room, whether it's ever had any members
+    // or not)
+    private roomSessions = new Map<string, MatrixRTCSession>();
```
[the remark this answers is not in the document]
dbkr: Yeah, although it won't get any larger than the number of rooms the client is joined to since there is at most one room session per room.

src/matrixrtc/MatrixRTCSessonManager.ts
dbkr: Yeah, this is a little confusing because `this.getRoomSession()` will create the session if it doesn't exist, and from the current room state, so sessions that were not previously in the map would be created reflecting the update that just happened. We could use `getActiveRoomSession()` instead here which would return null if we didn't know about it, but we need to create a new one anyway in order to tell it to update. I've renamed the variables, perhaps this makes it a bit clearer? I agree it's not the world's most elegant code though.

The problem with removing stopped sessions from the map is that only the sessions in this map will get updates dispatched to them, so if anything holds on to a session object that we've removed from the map, it will no longer reflect the current session in that room because that would then be a new object. Keeping the sessions around ensures that we always only ever use one object per room session for the lifetime of the client, so no possibility of stale objects. I'll add a comment to this effect.

src/matrixrtc/MatrixRTCSession.ts
daniel-abramov: Oh right, I did not know about it, thanks!

src/matrixrtc/MatrixRTCSession.ts
daniel-abramov: Yeah, this is true, but I was worried more about is because since we never `await` on it, it's possible to call this function multiple times in a row (essentially making them run in parallel), I don't know if it would have any undesired effects.

I.e. something like:
1. `this.updateCallMembershipEvent()` (gets executed and yields the control flow back to the runtime upon `await` when sending an event / when waiting to I/O).
2. `this.updateCallMembershipEvent()` is called again (while the previous is still running), while reacting to a different event.
3. Possibly `N` instance of `this.updateCallMembershipEvent()` will then run "in parallel".

May be an issue if `this.updateCallMembershipEvent()` relies upon some state that is not meant to be modified "in parallel" (while waiting on the `await`).

dbkr: Good point: awaiting on the funcvtion wouldn't have helped here since other events would still arrive and trigger it while it was awaiting, but I've put a wrapper in that prevents it from running it again if there's already one running.

src/matrixrtc/MatrixRTCSessonManager.ts
dbkr: Agreed that adding something to get a call from the Room object could be quite nice, as long as it is just a glue method and the actual code is somewhere else because rooms are already quite overloaded (room.ts is about 3k lines!). I mostly did it this was because this was similar to how the previous GroupCall stuf worked. I'm not sure how the session manager could be bottleneck though: it's not like it's running on its own thread/coroutine.

I was envisaging separate accessors for room session and user session to make it as easy as possible to get a room session, and we only ever support one of these per room by definition. I would add a separate `getUserSessions(): MatrixRTCSessiion[]` once we support user sessions (and therefore multiple sessions per room).

I'm not quite sure I understand your `MatrixRTCSession ` proposal - what are the `Member`/`LocalMember` objects / what do they represent exactly? 

Either way, this is sounding like a relatively significant refactor, so one option here could be to revisit this when we add support for user sessions?

src/matrixrtc/MatrixRTCSession.ts
daniel-abramov: Cool, I think you're referring [to this commit](https://github.com/matrix-org/matrix-js-sdk/pull/3663/commits/268a147a11b3a35e6bb7849d4f67efbd494268dd), right? - If so, are we confident that it's a safe change? - From my perspective it introduces another risk:

1. Something changed -> the `triggerCallMembershipEventUpdate()` is called to force the client to update the state.
2. The update happens, then the function stops at the `await` point awaiting the I/O to complete.
3. Something changed again and the `triggerCallMebershipEventUpdate()` is called again, but since `updateCallMembershipRunning` is `true`, we ignore that attempt to trigger an update and `return` eagerly.
4. Step (2) completes.
5. Due to the early return in step (3), we did not call `updateCallMembershipEvent()` and may have missed another update, the state of the session will be invalid (not synced-up) until the next update.

To me it feels like what we want instead is to schedule the update to happen, but just ensure that they execute one by one. In other words, if one update is running and then an event happens that wants to trigger the call membership event update function, we could either:
1. Execute the function which means that it will run "in parallel" with the already ongoing execution of this same function (that's how it was implemented initially in this PR).
2. Drop the latest request to call this function provided we know that this function is already being executed (that's how it done now with the latest change).
3. Accept the incoming request to call this update function, but don't execute it immediately, let the first function finish and only execute the latest request (that's how I imagined it should have been). There are several ways to implement it, one of which is to introduce some sort of a mutex or async queue. LiveKit JS SDK has some of similar synchronisation primitives, [`Mutex` for instance](https://github.com/livekit/client-sdk-js/blob/8042164c5f29f63e12ed91e97905d5b203088b14/src/room/utils.ts#L427).

src/matrixrtc/MatrixRTCSessonManager.ts
daniel-abramov: > I'm not quite sure I understand your MatrixRTCSession proposal - what are the Member/LocalMember objects / what do they represent exactly?

Members/participants of a call. I.e. similarly to `Participant`, `RemoteParticipant`, and `LocalParticipant` in a LiveKit. Each `Participant` holds some data (room membership data), `LocalParticipant` would then be a type that on top of the regular data that each `Participant` holds, has some additional methods, that e.g. allow changing participant's metadata. Changing it would consequently trigger a `MemberChanged` event so that other participants can react to the changes on their side.

> Either way, this is sounding like a relatively significant refactor, so one option here could be to revisit this when we add support for user sessions?

Yeah, we can do it like this I think.

src/matrixrtc/MatrixRTCSession.ts
dbkr: Yep, good point - we just need to make sure one more update happens however many are triggered, so queueing them feels overkill: I've done a simple boolean flag in 19bf62a62.

dbkr:
I've created https://github.com/vector-im/element-call/issues/1432 and https://github.com/vector-im/element-call/issues/1433 to track your last two points (the first one obviously is just in relation to the size of the PR itself).

PR #3667, opened 2023-08-22
Add E2EE for embedded mode of Element Call

SimonBrandner:
Based on https://github.com/matrix-org/matrix-js-sdk/pull/3663
For https://github.com/vector-im/element-call/pull/1350

---

## ✨ Features
 * Add E2EE for embedded mode of Element Call ([\#3667](https://github.com/matrix-org/matrix-js-sdk/pull/3667)).


richvdh:
@SimonBrandner if this is specific to voice calls, could you update the PR summary to say so?

Also:

> Requires https://github.com/vector-im/element-call/pull/1350

... and that says:

> Requires https://github.com/matrix-org/matrix-js-sdk/pull/3667

If each PR requires the other, it seems like we have a cycle. We try to avoid interdependent PRs across repos, as it makes CI and versioning tricky. I haven't looked at the changes, but is it possible to refactor the PRs so that one can safely land before the other?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -56,6 +64,13 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
 
     private activeFoci: Focus[] | undefined;
 
+    private encryptMedia = false;
+    private _encryptionKeys = new Map<{ userId: string; deviceId: string }, string>();
+
+    public get encryptionKeys(): Map<{ userId: string; deviceId: string }, string> {
```
dbkr, line 70: I don't really see the advantage of making the variable private but then just exposing it with a getter. I think my 1st choice would be to hide the map entirely with `getEncryptionKey()` & `setEncryptionKey()` functions. If that's too much boilerplate, `public readonly encryptionKeys  = ...` would be the same as this and shorter I think.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -117,7 +132,7 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
     }
 
     private constructor(
-        private readonly client: MatrixClient,
+        public readonly client: MatrixClient,
```
dbkr, line 135: aaaargh, I was able to make this private in the refactor, can't we keep it this way? :)

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -147,7 +162,7 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
      * This will not subscribe to updates: remember to call subscribe() separately if
      * desired.
      */
-    public joinRoomSession(activeFoci: Focus[]): void {
+    public async joinRoomSession(activeFoci: Focus[], encryptMedia?: boolean): Promise<void> {
```
dbkr, line 165: Is this change related? Does something need to wait until the membership event is sent?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -175,10 +191,36 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+        if (!this.isJoined()) return;
+        if (!this.encryptMedia) return;
+
+        logger.info(`MatrixRTCSession updating encryption key for room ${this.room.roomId}`);
+
+        const encryptionKey = getNewEncryptionKey();
+        await this.client.sendEvent(this.room.roomId, EventType.CallEncryptionPrefix, {
+            "m.encryption_key": encryptionKey,
+            "m.device_id": this.client.getDeviceId(),
+        } as EncryptionKeyEventContent);
```
dbkr, line 212: Feels like this probably ought to include the call ID (ie. empty string) & perhaps scope too? Enough to identify what call it's for, anyway.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -56,6 +64,13 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
 
     private activeFoci: Focus[] | undefined;
 
+    private encryptMedia = false;
+    private _encryptionKeys = new Map<{ userId: string; deviceId: string }, string>();
```
dbkr, line 68: Docs please. What keys are they? For incoming or outgoing media? Where are they used?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -175,10 +191,36 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+            "m.device_id": this.client.getDeviceId(),
+        } as EncryptionKeyEventContent);
+
+        const userId = this.client.getUserId();
+        const deviceId = this.client.getDeviceId();
+
+        if (!userId) throw new Error("No userId");
+        if (!deviceId) throw new Error("No deviceId");
+
+        this._encryptionKeys.set({ userId, deviceId }, encryptionKey);
```
dbkr, line 220: This is confusing me. Isn't `encryptionKeys` going to be for incoming media? Or do we also store our own in the same map but used for encrypting rather than decrypting? Personally I would store it in its own variable since it *is* a different thing.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -205,6 +247,26 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+        const deviceId = content["m.device_id"];
+
+        if (
+            !userId ||
+            !deviceId ||
+            !encryptionKey ||
+            typeof deviceId !== "string" ||
+            typeof encryptionKey !== "string"
+        ) {
+            throw new Error("Malformed m.call.encryption_key");
```
dbkr, line 263: I added the user ID to the equivalents of this because I wasn't finding the error messages terribly helpful, would be good to do the same here.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -205,6 +247,26 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+        }
+
+        this._encryptionKeys.set({ userId, deviceId }, encryptionKey);
```
dbkr, line 266: Worth a log line here I think.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -218,6 +280,13 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
             this.emit(MatrixRTCSessionEvent.MembershipsChanged, oldMemberships, this.memberships);
         }
 
+        const ts = (m: CallMembership): string => `${m.member.userId}:${m.deviceId}`;
```
dbkr, line 283: I'm reading `ts` as `timestamp` but this does not appear to be what this variable is...

It's also not dependent on anything from the current scope, so why not make it a separate function at the top level to assert its independence?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -218,6 +280,13 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+        const callMembersChanged = new Set(oldMemberships.map(ts)) !== new Set(this.memberships.map(ts));
```
dbkr, line 284: I'm not sure I understand how this is different to what's just been done above.

src/@types/event.ts
```diff
@@ -55,6 +55,7 @@ export enum EventType {
     CallReplaces = "m.call.replaces",
     CallAssertedIdentity = "m.call.asserted_identity",
     CallAssertedIdentityPrefix = "org.matrix.call.asserted_identity",
+    CallEncryptionPrefix = "io.element.call.encryption_key",
```
dbkr, line 58: Btw, why are these called 'prefix'? They're not prefixes, surely? They're just normal event type names.

src/matrixrtc/MatrixRTCSessonManager.ts
```diff
@@ -97,6 +100,18 @@ export class MatrixRTCSessionManager extends TypedEventEmitter<MatrixRTCSessionM
...
+    private onTimeline = (event: MatrixEvent): void => {
+        if (event.getType() !== EventType.CallEncryptionPrefix) return;
+
+        const room = this.client.getRoom(event.getRoomId());
+        if (!room) {
+            logger.error(`Got room state event for unknown room ${event.getRoomId()}!`);
+            return;
+        }
+
+        this.getRoomSession(room).onCallEncryption(event);
```
dbkr, line 112: Do we want to create a new session for this event if one doesn't exist? If not, we probably want `getActiveRoomSession()`.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -22,16 +22,23 @@ import { MatrixClient } from "../client";
...
 import { Focus } from "./focus";
+import { randomString } from "../randomstring";
+import { MatrixEvent } from "../matrix";
+import { EncryptionKeyEventContent } from "./types";
 
 const MEMBERSHIP_EXPIRY_TIME = 60 * 60 * 1000;
 const MEMBER_EVENT_CHECK_PERIOD = 2 * 60 * 1000; // How often we check to see if we need to re-send our member event
 const CALL_MEMBER_EVENT_RETRY_DELAY_MIN = 3000;
 
+const getNewEncryptionKey = (): string => randomString(32);
```
dbkr, line 33: Security horn! 🚨 This used Math.random and is not safe for a security sensitive string! Use `crypto.getRandomValues()`.

src/matrixrtc/MatrixRTCSession.ts
dbkr, line 266: Also I would expect the encryption keys map to be initialised with anything the sdk already knows about when the session object is created, otherwise isn't there a risk we'll miss them?

SimonBrandner:
> @SimonBrandner if this is specific to voice calls, could you update the PR summary to say so?

Ah, yeah, sorry for a not-very-descriptive name - should be better now

> Also:
> 
> > Requires [vector-im/element-call#1350](https://github.com/vector-im/element-call/pull/1350)
> 
> ... and that says:
> 
> > Requires #3667
> 
> If each PR requires the other, it seems like we have a cycle. We try to avoid interdependent PRs across repos, as it makes CI and versioning tricky. I haven't looked at the changes, but is it possible to refactor the PRs so that one can safely land before the other?

Ah, sorry, it's a habit from the days during which all my js-sdk and react-sdk VoIP PRs would depend on each other, haha. This does only depend on Dave's branch of the js-sdk

src/matrixrtc/MatrixRTCSession.ts
SimonBrandner: If we assume, we send new keys each time a new device joins the call, then no, we shouldn't miss them, right?

src/matrixrtc/MatrixRTCSession.ts
SimonBrandner: Good catch - thank you!

src/@types/event.ts
SimonBrandner: So this is somewhat up to discussion - I wrote this with the assumption that we would end up making this `m.` prefixed one day, so that's why I called it prefix

src/matrixrtc/MatrixRTCSession.ts
SimonBrandner: I more or less modeled this after LK which uses the same method to set the encryption key for the local user and decryption keys for remote ones. It does make sense to a degree since the cipher is symmetric, so they actually are almost the same thing

It really depends on perspective. I can change it, if you feel strongly about it

src/matrixrtc/MatrixRTCSession.ts
SimonBrandner: Good point!

src/matrixrtc/MatrixRTCSession.ts
SimonBrandner: This is a relict of the way to send the keys I used before - will revert

src/matrixrtc/MatrixRTCSessonManager.ts
SimonBrandner: :man_facepalming:

src/matrixrtc/MatrixRTCSession.ts
SimonBrandner: So the `get encryptionKeys()` method actually returns a copy of the map which the `readonly` approach doesn't, afaik. I also want EC to be able to access all keys we have using one method/getter

src/matrixrtc/MatrixRTCSessonManager.ts
SimonBrandner: Hmm, now that I am thinking about it, perhaps we want? In case the event arrives before state?

src/matrixrtc/MatrixRTCSession.ts
SimonBrandner: We don't want to send a new event if the content of the membership changes - we only care about joins/leaves

src/matrixrtc/MatrixRTCSessonManager.ts
dbkr: Yeah, agreed - if we're not checking for existing events then we probably need to be capturing them whenever they arrive, even if there's no current call. I suppose it would only happen if someone sends a key before their call membership event, but that could easily happen. This does seem like the easier way of doing it.

src/@types/event.ts
dbkr: Oh I see - that kind of prefix. That makes sense :)

src/matrixrtc/MatrixRTCSession.ts
dbkr: This still seems like a really roundabout way of achieving this to me, and clones map unnecessarily. What do you mean by, "access all keys we have", as in encryptionsKeys.values()? What's the value in it all being one getter? `getKey()` / `getAllKeys()` seems simpler & more explicit to me.

src/matrixrtc/MatrixRTCSession.ts
dbkr: Yeah, I guess it is similar to how livekit does it, so if it makes it easier there then I guess it makes sense. Otherwise it seems like it would be a map-lookup to get our own key rather than just a regular variable.

src/matrixrtc/MatrixRTCSession.ts
dbkr: Yeah, since we do this we can rely on that for now. Maybe this is good enough until we switch to todevice.

src/@types/event.ts
dbkr: Btw, `Prefixed` would probably convey this message a bit better, but no need to change them all now.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -57,6 +70,17 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
     private activeFoci: Focus[] | undefined;
 
+    private encryptMedia = false;
+    private _encryptionKeys = new Map<{ userId: string; deviceId: string }, string>();
+
+    /**
+     * A map of keys used to encrypt and decrypt (we are using a symmetric
+     * cipher) given participant's media. This also includes our own key
+     */
+    public get encryptionKeys(): Map<{ userId: string; deviceId: string }, string> {
```
dbkr, line 80: Also I've just realised that this is a Map with objects as keys. Beware that  this works because the only way it's being read is by iterating over the whole Map and getting the values out. `encryptionKeys.get({userId: foo, deviceId: bar})` won't work unless you actually pass in a reference to the same object (ie. equal according to SameValueZero equality). This definitely feels like a confusing data structure. If we['re only ever going to iterate over the whole thing, I'd suggest just using an array. If we expect to actually look up the keys in the map, I'd use a map with strings as keys.

src/@types/event.ts
SimonBrandner: FWIW, I followed the convention used by `CallAssertedIdentityPrefix` for example but yeah agreed

dbkr:
Putting all the key management stuff into js-sdk might not have been the way I'd have done it, or at least put it in a separate file from MatrixRTCSession, but I'm not sure rejigging this all is the best use of time right now anyway, and it's only my opinion anyway, so presented to the group for input.

src/randomstring.ts
```diff
@@ -15,10 +15,21 @@ See the License for the specific language governing permissions and
...
+
 const LOWERCASE = "abcdefghijklmnopqrstuvwxyz";
 const UPPERCASE = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
 const DIGITS = "0123456789";
 
+export function secureRandomBase64(len: number): string {
+    const key = new Uint8Array(len);
+    crypto.getRandomValues(key);
+
+    // encode to base64url as this value goes into URLs
```
robintown, line 29: It seems a little odd that this function should know this fact about its usage. Perhaps it should be named `secureRandomBase64Url`?

src/matrixrtc/MatrixRTCSessionManager.ts
```diff
@@ -95,6 +98,18 @@ export class MatrixRTCSessionManager extends TypedEventEmitter<MatrixRTCSessionM
...
+        if (event.getType() !== EventType.CallEncryptionKeysPrefix) return;
+
+        const room = this.client.getRoom(event.getRoomId());
+        if (!room) {
+            logger.error(`Got room state event for unknown room ${event.getRoomId()}!`);
+            return;
+        }
+
+        this.getRoomSession(room).onCallEncryption(event);
+    };
```
robintown, line 111: Non-blocking, but I'm not convinced that this callback belongs in the session manager class. I think it makes most sense if the session manager is only responsible for creating, ending, and storing sessions, so this feels better suited for the session class itself.

dbkr: Yeah, I think I would agree: I guess it was sticking to the pattern of all the events going into the manager and being dispatched from there.

robintown (approved):

PR #3668, opened 2023-08-23
Handle `localazy download` during release process

t3chguy:
---


t3chguy:
This won't work as the English translations can change too, which needs review & possibly updating snapshots/tests

PR #3675, opened 2023-08-24
Use sender instead of content.creator field on m.room.create events

Johennes:
https://github.com/matrix-org/matrix-spec-proposals/pull/2175 removed the `content.creator` field from `m.room.create` fields in favor of the `sender` field. This was released with [room version 11](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3820-rooms-v11.md).

This PR amends the JS SDK to use the `sender` field instead of the `content.creator` field _regardless of room version_. In line with https://github.com/matrix-org/matrix-spec-proposals/pull/2175#issuecomment-1669891053, the `content.creator` field is not used at all anymore.

## Checklist

### DIFF
--- a/spec/integ/matrix-client-event-emitter.spec.ts
+++ b/spec/integ/matrix-client-event-emitter.spec.ts
@@ -92,9 +92,7 @@ describe("MatrixClient events", function () {
                                     type: "m.room.create",
                                     room: "!erufh:bar",
                                     user: "@foo:bar",
-                                    content: {
-                                        creator: "@foo:bar",
-                                    },
+                                    content: {},
                                 }),
                             ],
                         },
--- a/spec/integ/matrix-client-methods.spec.ts
+++ b/spec/integ/matrix-client-methods.spec.ts
@@ -1988,10 +1988,7 @@ const buildEventMember = () =>
 const buildEventCreate = () =>
     new MatrixEvent({
         age: 80126105,
-        content: {
-            creator: "@andybalaam-test1:matrix.org",
-            room_version: "6",
-        },
+        content: {},
         event_id: "$e7j2Gt37k5NPwB6lz2N3V9lO5pUdNK8Ai7i2FPEK-oI",
         origin_server_ts: 1643815438782,
         room_id: "!STrMRsukXHtqQdSeHa:matrix.org",
--- a/spec/integ/matrix-client-syncing.spec.ts
+++ b/spec/integ/matrix-client-syncing.spec.ts
@@ -389,9 +389,7 @@ describe("MatrixClient syncing", () => {
                             type: "m.room.create",
                             room: roomOne,
                             user: selfUserId,
-                            content: {
-                                creator: selfUserId,
-                            },
+                            content: {},
                         }),
                     ],
                 },
@@ -577,9 +575,7 @@ describe("MatrixClient syncing", () => {
                                     type: "m.room.create",
                                     room: roomOne,
                                     user: selfUserId,
-                                    content: {
-                                        creator: selfUserId,
-                                    },
+                                    content: {},
                                 }),
                             ],
                         },
@@ -611,9 +607,7 @@ describe("MatrixClient syncing", () => {
                                     type: "m.room.create",
                                     room: roomTwo,
                                     user: selfUserId,
-                                    content: {
-                                        creator: selfUserId,
-                                    },
+                                    content: {},
                                 }),
                             ],
                         },
@@ -757,10 +751,7 @@ describe("MatrixClient syncing", () => {
                         type: "m.room.create",
                         room: roomOne,
                         user: otherUserId,
-                        content: {
-                            creator: otherUserId,
-                            room_version: "9",
-                        },
+                        content: {},
                     });
                     const normalFirstSync = {
                         next_batch: "batch_token",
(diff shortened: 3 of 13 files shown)


spec/integ/matrix-client-methods.spec.ts
```diff
@@ -1988,10 +1988,7 @@ const buildEventMember = () =>
 const buildEventCreate = () =>
     new MatrixEvent({
         age: 80126105,
-        content: {
-            creator: "@andybalaam-test1:matrix.org",
-            room_version: "6",
```
t3chguy, line 1993: Losing the room version here seems amiss

spec/integ/matrix-client-event-emitter.spec.ts
```diff
@@ -92,9 +92,7 @@ describe("MatrixClient events", function () {
                                     type: "m.room.create",
                                     room: "!erufh:bar",
                                     user: "@foo:bar",
-                                    content: {
-                                        creator: "@foo:bar",
-                                    },
+                                    content: {},
```
Johennes, line 95: I've opted for removing the `creator` field from all test fixtures in an attempt to expose other places that might currently rely on it.

spec/integ/matrix-client-syncing.spec.ts
```diff
@@ -843,10 +834,7 @@ describe("MatrixClient syncing", () => {
                         type: "m.room.create",
                         room: roomOne,
                         user: otherUserId,
-                        content: {
-                            creator: otherUserId,
-                            room_version: testMeta.roomVersion,
```
t3chguy, line 848: Same here

spec/integ/matrix-client-unread-notifications.spec.ts
```diff
@@ -178,10 +178,7 @@ describe("MatrixClient syncing", () => {
                         timeline: {
                             events: [
                                 {
-                                    content: {
-                                        creator: userB,
-                                        room_version: "9",
```
t3chguy, line 183: Same here

spec/unit/matrix-client.spec.ts
```diff
@@ -2266,13 +2266,11 @@ describe("MatrixClient", function () {
         function roomCreateEvent(newRoomId: string, predecessorRoomId: string): MatrixEvent {
             return new MatrixEvent({
                 content: {
-                    "creator": "@daryl:alexandria.example.com",
                     "m.federate": true,
                     "predecessor": {
                         event_id: "id_of_last_event",
                         room_id: predecessorRoomId,
                     },
-                    "room_version": "9",
```
t3chguy, line 2275: Why is this going away?

Johennes: Yeah, I wasn't sure what the implication of that would be and all the tests still run without it. So I consciously put it in here to cover it during review. Why do you think the room version should stick around here?

t3chguy: > Why do you think the room version should stick around here?

Because the PR is called `Use sender instead of content.creator field on m.room.create events` and that seems to have nothing to do with removing room versions from the content (i.e. reverting all rooms under test to v1)

Johennes: > Because the PR is called `Use sender instead of content.creator field on m.room.create events` and that seems to have nothing to do with removing room versions from the content

Yes, I had tried to explain why I included that in https://github.com/matrix-org/matrix-js-sdk/pull/3675#discussion_r1304094694 which I acknowledge happened simultaneously with your comments.

> i.e. reverting all rooms under test to v1

Oh, I had actually assumed we'd fall back to a default, not v1. In re-reading the spec, I'm realizing that the default is defined on a per-HS basis though.

So then this sounds like in order to be able to remove the `content.creator` fields from the fixtures, we'd have to explicitly include `"room_version": 11` instead.

How do you suggest we handle the fixtures in light of the spec changes in room version 11?

t3chguy: > Oh, I had actually assumed we'd fall back to a default, not v1. In re-reading the spec, I'm realizing that the default is defined on a per-HS basis though.

No, that's for the `/createRoom` API. A room without a version baked into its (immutable) `m.room.create` is `v1`

https://spec.matrix.org/v1.8/client-server-api/#mroomcreate

> The version of the room. Defaults to "1" if the key does not exist.

> So then this sounds like in order to be able to remove the content.creator fields from the fixtures, we'd have to explicitly include "room_version": 11 instead.

No as the behaviour being version-independent sounds fine to me, I just don't think your PR needs to touch the versions at all.

Johennes: > No, that's for the `/createRoom` API. A room without a version baked into its (immutable) `m.room.create` is `v1`

Ohh, right. 🤯

Johennes: > No as the behaviour being version-independent sounds fine to me, I just don't think your PR needs to touch the versions at all.

Ok, have made it so.

spec/integ/matrix-client-event-emitter.spec.ts
Johennes: This was addressed in the other thread.

PR #3683, opened 2023-08-29
`OutgoingRequestProcessor`: do not throw errors if shutting down

richvdh:
---

### DIFF
--- a/src/rust-crypto/OutgoingRequestProcessor.ts
+++ b/src/rust-crypto/OutgoingRequestProcessor.ts
@@ -105,7 +105,17 @@ export class OutgoingRequestProcessor {
         }
 
         if (msg.id) {
-            await this.olmMachine.markRequestAsSent(msg.id, msg.type, resp);
+            try {
+                await this.olmMachine.markRequestAsSent(msg.id, msg.type, resp);
+            } catch (e) {
+                // Ignore "Attempt to use moved value" errors. The most likely cause is that olmMachine has been
+                // stopped and frees, so we have nowhere to report the success of our call to.
+                if (e instanceof Error && e.message === "Attempt to use a moved value") {
+                    logger.log(`Ignoring error '${e.message}': client is likely shutting down`);
+                } else {
+                    throw e;
+                }
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/rust-crypto/OutgoingRequestProcessor.ts
```diff
@@ -105,7 +105,17 @@ export class OutgoingRequestProcessor {
         }
 
         if (msg.id) {
-            await this.olmMachine.markRequestAsSent(msg.id, msg.type, resp);
+            try {
+                await this.olmMachine.markRequestAsSent(msg.id, msg.type, resp);
+            } catch (e) {
+                // Ignore "Attempt to use moved value" errors. The most likely cause is that olmMachine has been
+                // stopped and frees, so we have nowhere to report the success of our call to.
```
andybalaam, line 112: frees->freed ?

andybalaam (approved):
Looks good, thanks. One typo.

richvdh:
@andybalaam sorry, I feel like I've changed this enough since you reviewed it that I should ask for a re-review

src/rust-crypto/OutgoingRequestProcessor.ts
```diff
@@ -105,7 +105,20 @@ export class OutgoingRequestProcessor {
         }
 
         if (msg.id) {
-            await this.olmMachine.markRequestAsSent(msg.id, msg.type, resp);
+            try {
+                await this.olmMachine.markRequestAsSent(msg.id, msg.type, resp);
+            } catch (e) {
+                // Ignore errors which are caused by the olmMachine having been freed. The exact error message depends
+                // on whether we are using a release or develop build of rust-sdk-crypto-wasm.
+                if (
```
weeman1337, line 113: It would be good if `olmMachine` can be asked about its status and return something like "shut down". Otherwise, this may hide bugs in the implementation.

weeman1337 (approved):
Code and test look good.

Just a thought on the ignored error. Maybe I am not familiar enough with the new crypto code. But I would have a better feeling if we can query the OlmMachine status additionally instead of only ignoring particular error messages. I can see Future-Rich desperately debugging the code and finding this place :laughing:

src/rust-crypto/OutgoingRequestProcessor.ts
richvdh: I don't disagree, this might be preferable. Indeed, I went looking at the (generated) OlmMachine code to see if there is a nice way before doing this.  There is in fact an `OlmMachine.ptr` property which we could check, but it's not exposed via typescript.

But then: is it worth it? I think it's vanishingly unlikely we'd see either of these errors for any reason other than the OlmMachine having been freed.

Issue #3678, opened 2023-08-27
Client keeps syncing after calling logout(true) (or stopClient())


finsterwalder:
I initialized a client with a wrong token.
I attached a listener for HttpApiEvent.SessionLoggedOut, so it got triggered and I called mxclient.logout(true).
I see in the logs:
```
stopping MatrixClient
SyncApi.stop
```
But the client keeps syncing and my listener for ClientEvent.Sync is called regularly with state ERROR.

What is going wrong? How can I ensure the client stops, no matter if it's connected or not?

finsterwalder:
I noticed, that FetchHttpApi logs an error:
```
FetchHttpApi: --> GET http://<someurl>:8008/_matrix/client/versions
FetchHttpApi: <-- GET http://<someurl>:8008/_matrix/client/versions [0ms AbortError: Failed to execute 'fetch' on 'Window': The user aborted a request.]
```
So the logout has the effect, that no requests are actually sent, but the sync is still triggered regularly...

vitonsky:
I have the same problem

finsterwalder:
Update your SDK.
I solved this issue with this Pull Request 
https://github.com/matrix-org/matrix-js-sdk/pull/3720

vitonsky:
> Please update your SDK. I solved this issue with this Pull Request #3720

It means we have regression then. I just tried it and i decided to not use current package, because i can't stop it properly.
This is why i decided to use https://github.com/turt2live/matrix-bot-sdk package instead. They are no have docs too, so it is difficult to use and still does not support encrypted rooms fine, but at least i can stop their client.

finsterwalder:
I don't think we have a regression of the same issue. I added tests to verify it. 
Your scenario must be different somehow.
Do you have a working connection that you want to stop? That works without problems and always has. My issue only happened when the initial connection had a problem and went into a retry. And during this retry a stop had no effect, which I fixed. 
Maybe you can describe exactly what is happening?

PR #3686, opened 2023-08-29
Stop keep alive when sync is aborted

finsterwalder:
Type: [defect] `T-Defect`
This PR solves issue: https://github.com/matrix-org/matrix-js-sdk/issues/3678

When you start a client, which has connection problems, a keepAlive is started.
But when the connection is actively stopped e.g. via calling stopSync(), the connection is aborted.
When the connection is aborted, it does not make sense to continue to try to connect, because it will never succeed.

This is solved by this PR.

I could not find any tests for the sync, so I didn't find a place to extend tests to check this special case.

## Checklist

---

## ✨ Features
 * Stop keep alive when sync is aborted ([\#3686](https://github.com/matrix-org/matrix-js-sdk/pull/3686)). Contributed by @finsterwalder.

### DIFF
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -1604,7 +1604,7 @@ export class SyncApi {
                     success();
                 },
                 (err) => {
-                    if (err.name === "AbortError") {
+                    if (!this.running) {
                         clearTimeout(this.keepAliveTimer);
                         if (this.connectionReturnedDefer) {
                             this.connectionReturnedDefer.reject();
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -1604,7 +1604,13 @@ export class SyncApi {
                     success();
                 },
                 (err) => {
-                    if (err.httpStatus == 400 || err.httpStatus == 404) {
+                    if (err.name === "AbortError") {
+                        clearTimeout(this.keepAliveTimer);
+                        if (this.connectionReturnedDefer) {
+                            this.connectionReturnedDefer.reject();
+                            this.connectionReturnedDefer = undefined;
+                        }
+                    } else if (err.httpStatus == 400 || err.httpStatus == 404) {
                         // treat this as a success because the server probably just doesn't
                         // support /versions: point is, we're getting a response.
                         // We wait a short time though, just in case somehow the server
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -15,7 +15,7 @@ concurrency:
 jobs:
     cypress:
         name: Cypress
-        uses: matrix-org/matrix-react-sdk/.github/workflows/cypress.yaml@v3.77.1
+        uses: matrix-org/matrix-react-sdk/.github/workflows/cypress.yaml@v3.78.0
         permissions:
             actions: read
             issues: read
--- a/.github/workflows/downstream-artifacts.yml
+++ b/.github/workflows/downstream-artifacts.yml
@@ -21,7 +21,7 @@ concurrency:
 jobs:
     build-element-web:
         name: Build element-web
-        uses: matrix-org/matrix-react-sdk/.github/workflows/element-web.yaml@v3.77.1
+        uses: matrix-org/matrix-react-sdk/.github/workflows/element-web.yaml@v3.78.0
         with:
             matrix-js-sdk-sha: ${{ github.sha }}
             react-sdk-repository: matrix-org/matrix-react-sdk
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+Changes in [28.0.0](https://github.com/matrix-org/matrix-js-sdk/releases/tag/v28.0.0) (2023-08-29)
+==================================================================================================
+
+## 🚨 BREAKING CHANGES
+ * Set minimum supported Matrix 1.1 version (drop legacy r0 versions) ([\#3007](https://github.com/matrix-org/matrix-js-sdk/pull/3007)). Fixes vector-im/element-web#16876.
+
+## 🦖 Deprecations
+ * **The Browserify artifact is being deprecated, scheduled for removal in the October 10th release cycle. (#3189)**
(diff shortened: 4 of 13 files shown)


finsterwalder:
Type: [defect]

richvdh (changes requested):
Thank you for the contribution!

However... I don't really understand this, and it seems dangerous.

First of all, please can you explain the code path more clearly. Why is an `AbortError` thrown in this situation? I'm not very familiar with this area of the code, I'm afraid. In any case, the new code needs more comments to explain what is going on.

Second: how can you be sure that this is the only reason that an error like this will be thrown? I note that the code that handles regular `/sync` requests doesn't do anything like this. Instead, it checks `this.running` (https://github.com/matrix-org/matrix-js-sdk/blob/a33edb2ee7b425a4b52dfcd0511ed99977c2ea52/src/sync.ts#L1038). It seems like we should do the same thing here.

Third: please do add some tests for this case, otherwise there is a high chance of a regression. I think the best place is `spec/integ/matrix-client-syncing.spec.ts`.

finsterwalder:
Thanks for checking my PR so quickly.

The code path is as follows.
When the sync is started, but fails initially, recoverFromSyncStartupError() is called, which indirectly calls pokeKeepAlive(), which keeps polling the versions endpoint.
This polling only stops, when we get a http response, that is either successful or has at least http status 400 or 404.
In all other cases it keeps polling.
When sync.stop() is called in this situation, abortController.abort() is called, which leads to all further http calls resulting in an AbortError (the http request itself is not even executed anymore). The above keep alive does not stop, though and will simply keep polling indefinitely.
So my rational was, that getting an AbortError is a sure sign, that polling can never succeed and should end.

But you are right. `if (!this.running)` is probably the more appropriate check.
It's set to true, right before the abortController is set and it's set to false, right before abortController.abort() is called. So it should have the same effect as checking for an AbortError.
I will update my PR accordingly.

I added a sign-off now. I did that on a different branch and then forgot it. ;-)

I looked into the test file, but I didn't see any tests, that check the keep alive feature.
That's quite a task for me to get acquainted enough to write such a test for the first time...

src/sync.ts
```diff
@@ -1604,7 +1604,13 @@ export class SyncApi {
                     success();
                 },
                 (err) => {
-                    if (err.httpStatus == 400 || err.httpStatus == 404) {
+                    if (!this.running) {
+                        clearTimeout(this.keepAliveTimer);
+                        if (this.connectionReturnedDefer) {
+                            this.connectionReturnedDefer.reject();
+                            this.connectionReturnedDefer = undefined;
+                        }
```
richvdh, line 1612: >  In any case, the new code needs more comments to explain what is going on.

please do give this code some comments to explain what it is doing. Imagine you are unfamiliar with this code and trying to understand it. When does it happen, why is this the right thing to do? Are there other places in this file that have similar logic that it might be helpful to link to?

richvdh (changes requested):

richvdh:
> I added a sign-off now. I did that on a different branch and then forgot it. ;-)

I see a sign-off on one of the commits in this PR, but not the ones with all the new code! Best thing to do is to update the description of the PR to show that your signoff covers all the changes in the PR.

finsterwalder:
Ok. Good points. I will add documentation.
I will also fix the sign-off. Sorry for that, have never done that before. :-)
I will also look into creating a test. But that might take a little while...

Thinking about my problem a little more, I have some ideas for followup improvements. Was wondering, what you think about it:

This is a keepAlive, which keeps retrying the call to the matrix server. When I stop the sync, the retry should obviously stop. That's what my fix is about.
But maybe there are other error cases, where we know that retrying is useless, because it can never succeed?
e.g. A wrong auth token will never fix itself by just retrying. That's exactly the case that is at the root of my problem. Just that I need to detect the auth error myself (listening for an HttpApiEvent.SessionLoggedOut even) and the stopping the sync myself. With my fix, that stopSync call will stop the keepAlive now. But wouldn't it make sense to stop the keepAlive already, once a broken auth token is detected? So a response with http status 401 or 403 are pointless to retry IMHO. And maybe there are other cases, where retrying is pointless? How about 407 Proxy Authentication Required or 410 Gone (for good)? 414 URI Too Long? 431 Request Header Fields Too Large?

richvdh:
> But maybe there are other error cases, where we know that retrying is useless, because it can never succeed?
> e.g. A wrong auth token will never fix itself by just retrying. That's exactly the case that is at the root of my problem. Just that I need to detect the auth error myself (listening for an HttpApiEvent.SessionLoggedOut even) and the stopping the sync myself. With my fix, that stopSync call will stop the keepAlive now. But wouldn't it make sense to stop the keepAlive already, once a broken auth token is detected? 

You're probably right that there are lots of other calls that aren't worth retrying. However: One problem is that, as things currently stand, the spec (https://spec.matrix.org/v1.8/client-server-api/#get_matrixclientv3sync) has very little to say about which error codes might be returned, and what you can infer from them.

A second problem: what should we actually do on one of these errors? If we just stop the sync loop, it just leaves the user with a dead client. Should we raise a `SessionLoggedOut` event? That's a bit of a nuclear bomb because it will make the client forget all of its state (including E2EE keys, etc). Who is to say that it is not a temporary outage?

So we might be able to improve this, but I strongly recommend focussing on `stopSync` during keepalive for now.

finsterwalder:
I understand that including other cases would need a little more thinking. I just wanted to raise awareness and start this discussion.
In this change I will only do the handling of !isRunning to stop after calling stopSync.

The 401 case is currently notified via the HttpApiEvent.SessionLoggedOut event, so there exists a possible handling strategy. And it's also clear, that this will not fix itself with retries. Would it make sense to include this case soon?

finsterwalder:
It was easier to create a new, clean Pull Request: https://github.com/matrix-org/matrix-js-sdk/pull/3720

PR #3689, opened 2023-09-01
Deprecate `MatrixClient.{prepare,create}KeyBackupVersion` in favour of new `CryptoApi.resetKeyBackup` API

BillCarsonFr:
Deprecates `prepareKeyBackupVersion`/`createKeyBackupVersion` in favor of a new api `resetKeyBackup` that deletes existing backups if needed and then creates a new one.

Rust Implemention is just stub for now.

## Checklist

---

## 🦖 Deprecations
 * Deprecate `MatrixClient.{prepare,create}KeyBackupVersion` in favour of new `CryptoApi.resetKeyBackup` API ([\#3689](https://github.com/matrix-org/matrix-js-sdk/pull/3689)).

### DIFF
--- a/spec/integ/crypto/crypto.spec.ts
+++ b/spec/integ/crypto/crypto.spec.ts
@@ -40,6 +40,7 @@ import { logger } from "../../../src/logger";
 import {
     Category,
     createClient,
+    CryptoEvent,
     IClaimOTKsResult,
     IContent,
     IDownloadKeyResult,
@@ -61,9 +62,13 @@ import { ISyncResponder, SyncResponder } from "../../test-utils/SyncResponder";
 import { escapeRegExp } from "../../../src/utils";
 import { downloadDeviceToJsDevice } from "../../../src/rust-crypto/device-converter";
 import { flushPromises } from "../../test-utils/flushPromises";
-import { mockInitialApiRequests, mockSetupCrossSigningRequests } from "../../test-utils/mockEndpoints";
+import {
+    mockInitialApiRequests,
+    mockSetupCrossSigningRequests,
+    mockSetupMegolmBackupRequests,
+} from "../../test-utils/mockEndpoints";
 import { AddSecretStorageKeyOpts, SECRET_STORAGE_ALGORITHM_V1_AES } from "../../../src/secret-storage";
-import { CryptoCallbacks } from "../../../src/crypto-api";
+import { CryptoCallbacks, KeyBackupInfo } from "../../../src/crypto-api";
 import { E2EKeyResponder } from "../../test-utils/E2EKeyResponder";
 
 afterEach(() => {
@@ -2197,11 +2202,9 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string, 
                     "express:/_matrix/client/v3/user/:userId/account_data/:type(m.secret_storage.*)",
                     (url: string, options: RequestInit) => {
                         const content = JSON.parse(options.body as string);
-
                         if (content.key) {
                             resolve(content.key);
                         }
-
                         return {};
                     },
                     { overwriteRoutes: true },
@@ -2226,6 +2229,68 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string, 
                     },
                 );
             });
+        }
+
+        function awaitMegolmBackupKeyUpload(): Promise<Record<string, {}>> {
+            return new Promise((resolve) => {
+                // Called when the megolm backup key is uploaded
+                fetchMock.put(
+                    `express:/_matrix/client/v3/user/:userId/account_data/m.megolm_backup.v1`,
+                    (url: string, options: RequestInit) => {
+                        const content = JSON.parse(options.body as string);
+                        resolve(content.encrypted);
+                        return {};
+                    },
+                    { overwriteRoutes: true },
+                );
+            });
+        }
+
+        async function bootstrapSecurity(backupVersion: string): Promise<void> {
+            mockSetupCrossSigningRequests();
+            mockSetupMegolmBackupRequests(backupVersion);
+
+            // promise which will resolve when a `KeyBackupStatus` event is emitted with `enabled: true`
+            const backupStatusUpdate = new Promise<void>((resolve) => {
+                aliceClient.on(CryptoEvent.KeyBackupStatus, (enabled) => {
+                    if (enabled) {
+                        resolve();
+                    }
+                });
+            });
+
+            const setupPromises = [
+                awaitCrossSigningKeyUpload("master"),
+                awaitCrossSigningKeyUpload("user_signing"),
+                awaitCrossSigningKeyUpload("self_signing"),
+                awaitMegolmBackupKeyUpload(),
+            ];
+
+            // Before setting up secret-storage, bootstrap cross-signing, so that the client has cross-signing keys.
+            await aliceClient.getCrypto()!.bootstrapCrossSigning({});
+
+            // Now, when we bootstrap secret-storage, the cross-signing keys should be uploaded.
+            const bootstrapPromise = aliceClient.getCrypto()!.bootstrapSecretStorage({
+                setupNewSecretStorage: true,
+                createSecretStorageKey,
+                setupNewKeyBackup: true,
+            });
+
+            // Wait for the key to be uploaded in the account data
+            const secretStorageKey = await awaitSecretStorageKeyStoredInAccountData();
+
+            // Return the newly created key in the sync response
+            sendSyncResponse(secretStorageKey);
+
+            // Wait for the cross signing keys to be uploaded
+            await Promise.all(setupPromises);
+
+            // wait for bootstrapSecretStorage to finished
+            await bootstrapPromise;
+            // Finally ensure backup is working
+            await aliceClient.getCrypto()!.checkKeyBackupAndEnable();
+
+            await backupStatusUpdate;
         }
 
         /**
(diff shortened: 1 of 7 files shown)


src/crypto-api.ts
```diff
@@ -233,6 +233,24 @@ export interface CryptoApi {
...
+     * and the backup engine will be started.
+     */
+    resetKeyBackup(): Promise<void>;
+
+    /**
+     * Deletes the given key backup.
+     *
+     * @param version - The backup version to delete.
+     */
+    deleteKeyBackupVersion(version: string): Promise<void>;
```
richvdh, line 252: There is literally a section in this file called Secure Key backup: https://github.com/matrix-org/matrix-js-sdk/blob/b39741acbd66d6cd2455363412573f13e3f46747/src/crypto-api.ts#L357

src/client.ts
```diff
@@ -3403,6 +3409,9 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      *
      * @param info - Info object from prepareKeyBackupVersion
      * @returns Object with 'version' param indicating the version created
+     *
```
richvdh, line 3412:
```suggestion
```

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2228,6 +2231,68 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
+                        const content = JSON.parse(options.body as string);
+                        resolve(content.encrypted);
+                        return {};
+                    },
+                    { overwriteRoutes: true },
+                );
+            });
+        }
+
+        async function bootstrapSecurity(backupVersion: string): Promise<void> {
```
richvdh, line 2249: I think it would be helpful to add some documentation here.

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2385,6 +2450,96 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
+        oldBackendOnly("should create a new megolm backup", async () => {
+            const backupVersion = "abc";
+            await bootstrapSecurity(backupVersion);
+
+            // Expect a backup to be available and used
+            const activeBackup = await aliceClient.getCrypto()!.getActiveSessionBackupVersion();
+            expect(activeBackup).toStrictEqual(backupVersion);
+        });
+
+        oldBackendOnly("Reset key backup should create a new backup and update 4S", async () => {
```
richvdh, line 2463: This isn't testing `bootstrapSecretStorage`, so please move it outside the `describe("bootstrapSecretStorage",...)` block.

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2385,6 +2450,96 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
+            // First set up recovery
```
richvdh, line 2464: "recovery" means nothing to me

```suggestion
            // First set up 4S and key backup
```

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2385,6 +2450,96 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
+
+            const newBackupStatusUpdate = new Promise<void>((resolve) => {
+                aliceClient.on(CryptoEvent.KeyBackupStatus, (enabled) => {
+                    if (enabled) {
+                        resolve();
+                    }
+                });
+            });
+
+            const new4SUpload = awaitMegolmBackupKeyUpload();
```
richvdh, line 2522: this is nothing to do with 4S?
```suggestion
            const newBackupUploadPromise = awaitMegolmBackupKeyUpload();
```

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2385,6 +2450,96 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
+            await new4SUpload;
+
+            const nextVersion = await aliceClient.getCrypto()!.getActiveSessionBackupVersion();
+            const nextKey = await aliceClient.getCrypto()!.getSessionBackupPrivateKey();
+
+            expect(nextVersion).toBeDefined();
+            expect(nextVersion).not.toEqual(currentVersion);
+            expect(nextKey).not.toEqual(currentBackupKey);
+
+            // The API is deprecated but has been modified to work with both crypto backend
```
richvdh, line 2536: which API?

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2385,6 +2450,96 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
+            // ensure that it works anyhow
+            await aliceClient.deleteKeyBackupVersion(nextVersion!);
+            await aliceClient.getCrypto()!.checkKeyBackupAndEnable();
+            // XXX Legacy is not updating 4S when doing that, should ensure that rust implem does it.
```
richvdh, line 2540:
```suggestion
            // XXX Legacy crypto does not update 4S when doing that; should ensure that rust implem does it.
```

richvdh: I don't really understand how or why 4S would be updated here, but I'm happy to leave that for now

richvdh (changes requested):

src/rust-crypto/rust-crypto.ts
```diff
@@ -601,6 +601,16 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
+    /**
+     * Implementation of {@link CryptoApi#resetKeyBackup}.
+     */
+    public async resetKeyBackup(): Promise<void> {
+        // stub
+    }
+
+    public async deleteKeyBackupVersion(version: string): Promise<void> {
+        // stub
+    }
```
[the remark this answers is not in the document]
BillCarsonFr: moved

src/crypto-api.ts
BillCarsonFr: moved

src/client.ts
BillCarsonFr: done

spec/integ/crypto/crypto.spec.ts
BillCarsonFr: addd

spec/integ/crypto/crypto.spec.ts
BillCarsonFr: It tightly related, the secret storage should stays in sync when reseting the backup. It shares a lot of common setup code. Moving it outside would require to move also all setup function somewhere.

spec/integ/crypto/crypto.spec.ts
BillCarsonFr: Changed but we should stop considering 4S and keybackup as something different even if technically it is. The fact that we can have key backup without 4S is legacy. I think that account recovery is a better term

spec/integ/crypto/crypto.spec.ts
BillCarsonFr: done

spec/integ/crypto/crypto.spec.ts
BillCarsonFr: deleteKeyBackupVersion, updated the comment

spec/integ/crypto/crypto.spec.ts
BillCarsonFr: The 4S stores the private keys of the backup, if you delete the current backup but don't remove the saved secret in 4S you will have a desynchronised 4S with outdated secrets, which can cause issues on restore.

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2228,6 +2231,74 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
+                        resolve(content.encrypted);
+                        return {};
+                    },
+                    { overwriteRoutes: true },
+                );
+            });
+        }
+
+        /**
+         * Add all mocks needed to setup cross-signing, key backup, 4S and then
```
richvdh, line 2250:
```suggestion
         * Add all mocks needed to set up cross-signing, key backup, 4S and then
```

spec/integ/crypto/crypto.spec.ts
richvdh: > It tightly related, the secret storage should stay in sync when reseting the backup.

It might be related, but I don't accept that it's testing `bootstrapSecretStorage`. It assumes that `bootstrapSecretStorage` works as expected.

Alternatively expressed: we should make sure that we are clear which tests are for testing which method. We should avoid having big tests which test _lots of things_ because they make it hard to see what has broken when they fail.

> It shares a lot of common setup code. Moving it outside would require to move also all setup function somewhere.

We shouldn't be afraid to clean up code. Reluctance to do so in the past is one of the reasons this entire Element R project has taken 3 times as long as we expected: nobody understands how the js-sdk works any more.

spec/integ/crypto/crypto.spec.ts
richvdh: > I think that account recovery is a better term

Well, it would be better if we said "account recovery" rather than just "recovery". Again, "recovery" on its own doesn't mean much to me.

But also: please let's not start using UX terms in code before we are *absolutely certain* they are going to stick. It is really confusing when code uses a term to mean something different from the UI, and that's not helped if it's a term we're still experimenting with.

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2385,6 +2456,96 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
+            await newBackupUploadPromise;
+
+            const nextVersion = await aliceClient.getCrypto()!.getActiveSessionBackupVersion();
+            const nextKey = await aliceClient.getCrypto()!.getSessionBackupPrivateKey();
+
+            expect(nextVersion).toBeDefined();
+            expect(nextVersion).not.toEqual(currentVersion);
+            expect(nextKey).not.toEqual(currentBackupKey);
+
+            // The `deleteKeyBackupVersion` API is deprecated but has been modified to work with both crypto backend
```
richvdh, line 2542: https://github.com/matrix-org/matrix-js-sdk/pull/3689/files#diff-ca8741e6064cda5d26243ec210815e74962e5617951c9ec99ef4877dad964b38R402 is not deprecated.

```suggestion
            // `MatrixClient.deleteKeyBackupVersion` API is deprecated but has been modified to work with both crypto backends
```

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2385,6 +2456,96 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
+            // ensure that it works anyhow
+            await aliceClient.deleteKeyBackupVersion(nextVersion!);
+            await aliceClient.getCrypto()!.checkKeyBackupAndEnable();
+            // XXX Legacy crypto does not update 4S when doing that; should ensure that rust implem does it.
```
richvdh, line 2546:
```suggestion
            // XXX Legacy crypto does not update 4S when deleting backups; should ensure that rust implem does it.
```

richvdh (changes requested):
Please do move the `resetKeyBackup` test outside the `bosstrapSecretStorage` block.

Otherwise, just a few nits.

spec/integ/crypto/crypto.spec.ts
BillCarsonFr: Created a new task to handle that after the 3 backup PRs are merged to simplify the refactoring

richvdh (approved):

PR #3692, opened 2023-09-04
Legacy crypto: re-check key backup after `bootstrapSecretStorage`

richvdh:
Fixes https://github.com/vector-im/element-web/issues/26115.

Ideally we'd have better tests for this, but since:
 1. it's a bug in legacy crypto which we hope to kill off soon
 2. the only reason I'm fixing it now is because it broke a cypress test I'm trying to write

... I'm afraid I'm not inclined to spend a lot of time on it.

element-web notes: Fix bug which caused "Set up secure backup" flow not to actually enable  backups.

---

## 🐛 Bug Fixes
 * Legacy crypto: re-check key backup after `bootstrapSecretStorage` ([\#3692](https://github.com/matrix-org/matrix-js-sdk/pull/3692)). Fixes vector-im/element-web#26115.

### DIFF
--- a/spec/unit/crypto/secrets.spec.ts
+++ b/spec/unit/crypto/secrets.spec.ts
@@ -312,6 +312,7 @@ describe("Secrets", function () {
                 this.emit(ClientEvent.AccountData, event);
                 return {};
             };
+            bob.getKeyBackupVersion = jest.fn().mockResolvedValue(null);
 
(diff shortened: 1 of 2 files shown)


weeman1337 (approved):
I've checked out the branch and tested it locally. Works :+1:  Thanks for fixing this!

PR #3706, opened 2023-09-06
Don't reset unread count when adding a synthetic receipt

### LINKED ISSUE #3684: Notification counts may be incorrectly reset by sent messages
Suppose:

 * You invite Bob to a room. Bob immediately joins and sends a message.
 * `/sync` is slow for whatever reason, and Bob's message is returned in the `/sync` response as the remote echo of your invite event.

In that scenario, `Room.getUnreadNotificationCount()` should return 1. However, it currently returns 0. This causes the client to show the room as fully read, whereas in fact there are unread messages.

andybalaam:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/3684 and there are lots more details about why we chose this solution in that issue.

---

## 🐛 Bug Fixes
 * Don't reset unread count when adding a synthetic receipt ([\#3706](https://github.com/matrix-org/matrix-js-sdk/pull/3706)). Fixes #3684. Contributed by @andybalaam.

### DIFF
--- a/spec/unit/room.spec.ts
+++ b/spec/unit/room.spec.ts
@@ -1453,6 +1453,87 @@ describe("Room", function () {
         }
 
         describe("addReceipt", function () {
+            describe("resets the unread count", () => {
+                const event1 = utils.mkMessage({ room: roomId, user: userA, msg: "1", event: true });
+                const event2 = utils.mkMessage({ room: roomId, user: userA, msg: "2", event: true });
+
+                it("should reset the unread count when our non-synthetic receipt points to the latest event", () => {
+                    // Given a room with 2 events, and an unread count set.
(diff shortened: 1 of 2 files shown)


weeman1337 (approved):
Looks good to me. :+1:  I don't know all of the read receipt details. But implementation matches [the comment from here](https://github.com/matrix-org/matrix-js-sdk/issues/3684#issuecomment-1708383374).

andybalaam:
I would say we are not 100% confident on this change, which is why we are merging soon after the RC was cut. All the tests that we have are passing, but we don't feel they cover all the cases yet. When https://github.com/vector-im/element-web/issues/25449 is done, we should be more confident.

PR #3709, opened 2023-09-06
rust backup restore support

BillCarsonFr:
Based on https://github.com/matrix-org/matrix-js-sdk/pull/3708

## Checklist

---

### DIFF
--- a/src/rust-crypto/backup.ts
+++ b/src/rust-crypto/backup.ts
@@ -17,7 +17,14 @@ limitations under the License.
 import { OlmMachine, SignatureVerification } from "@matrix-org/matrix-sdk-crypto-wasm";
 import * as RustSdkCryptoJs from "@matrix-org/matrix-sdk-crypto-wasm";
 
-import { BackupTrustInfo, Curve25519AuthData, KeyBackupCheck, KeyBackupInfo } from "../crypto-api/keybackup";
+import {
+    BackupTrustInfo,
+    Curve25519AuthData,
+    KeyBackupCheck,
+    KeyBackupInfo,
+    KeyBackupSession,
+    Curve25519SessionData,
+} from "../crypto-api/keybackup";
 import { logger } from "../logger";
 import { ClientPrefix, IHttpOpts, MatrixError, MatrixHttpApi, Method } from "../http-api";
 import { CryptoEvent, IMegolmSessionData } from "../crypto";
@@ -27,7 +34,6 @@ import { OutgoingRequestProcessor } from "./OutgoingRequestProcessor";
 import { sleep } from "../utils";
 import { BackupDecryptor } from "../common-crypto/CryptoBackend";
 import { IEncryptedPayload } from "../crypto/aes";
-import { IKeyBackupSession, Curve25519SessionData } from "../crypto/keybackup";
 
 /** Authentification of the backup info, depends on algorithm */
 type AuthData = KeyBackupInfo["auth_data"];
@@ -389,7 +395,7 @@ export class RustBackupDecryptor implements BackupDecryptor {
      * Implements {@link BackupDecryptor#decryptSessions}
      */
     public async decryptSessions(
-        ciphertexts: Record<string, IKeyBackupSession<Curve25519SessionData | IEncryptedPayload>>,
+        ciphertexts: Record<string, KeyBackupSession<Curve25519SessionData | IEncryptedPayload>>,
     ): Promise<IMegolmSessionData[]> {
         const keys: IMegolmSessionData[] = [];
         for (const [sessionId, sessionData] of Object.entries(ciphertexts)) {
--- a/src/rust-crypto/backup.ts
+++ b/src/rust-crypto/backup.ts
@@ -20,11 +20,14 @@ import * as RustSdkCryptoJs from "@matrix-org/matrix-sdk-crypto-wasm";
 import { BackupTrustInfo, Curve25519AuthData, KeyBackupCheck, KeyBackupInfo } from "../crypto-api/keybackup";
 import { logger } from "../logger";
 import { ClientPrefix, IHttpOpts, MatrixError, MatrixHttpApi, Method } from "../http-api";
-import { CryptoEvent } from "../crypto";
+import { CryptoEvent, IMegolmSessionData } from "../crypto";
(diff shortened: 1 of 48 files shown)


src/rust-crypto/backup.ts
```diff
@@ -20,11 +20,14 @@ import * as RustSdkCryptoJs from "@matrix-org/matrix-sdk-crypto-wasm";
...
 import { ClientPrefix, IHttpOpts, MatrixError, MatrixHttpApi, Method } from "../http-api";
-import { CryptoEvent } from "../crypto";
+import { CryptoEvent, IMegolmSessionData } from "../crypto";
 import { TypedEventEmitter } from "../models/typed-event-emitter";
 import { encodeUri } from "../utils";
 import { OutgoingRequestProcessor } from "./OutgoingRequestProcessor";
 import { sleep } from "../utils";
+import { BackupDecryptor } from "../common-crypto/CryptoBackend";
+import { IEncryptedPayload } from "../crypto/aes";
+import { IKeyBackupSession, Curve25519SessionData } from "../crypto/keybackup";
```
richvdh, line 30: I'd love it if we could move utilities like this out of `crypto` while we are working on them (I basically see `src/crypto` as "Legacy Crypto"), but we can do that later if you prefer

src/rust-crypto/rust-crypto.ts
```diff
@@ -151,7 +153,24 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
+        }
+
+        const authData = <Curve25519AuthData>backupInfo.auth_data;
+
+        if (!(privKey instanceof Uint8Array)) {
+            // eslint-disable-next-line @typescript-eslint/no-base-to-string
+            throw new Error(`getBackupDecryptor expects Uint8Array, got ${privKey}`);
+        }
+
+        const backupDecryptionKey = RustSdkCryptoJs.BackupDecryptionKey.fromBase64(encodeBase64(privKey));
```
richvdh, line 167: `olmlib` looks and sounds like something libolm-related, even if it's not.

`encodeBase64` is pretty simple and I prefer just to inline it:

```suggestion
        const b64PrivKey = Buffer.from(privKey).toString("base64");
        const backupDecryptionKey = RustSdkCryptoJs.BackupDecryptionKey.fromBase64(b64PrivKey);
```

Or, since we already have exactly this code in `storeSessionBackupPrivateKey`, maybe we should factor out a new utility function

richvdh (commented):
a couple of very brief things that jumped out at me

src/rust-crypto/backup.ts
```diff
@@ -370,6 +373,51 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+                    await this.decryptionKey.decryptV1(
+                        sessionData.session_data.ephemeral,
+                        sessionData.session_data.mac,
+                        sessionData.session_data.ciphertext,
+                    ),
+                );
+                decrypted.session_id = sessionId;
+                keys.push(decrypted);
+            } catch (e) {
+                logger.log("Failed to decrypt megolm session from backup", e, sessionData);
```
andybalaam, line 407: Is sessionData safe to log?

src/rust-crypto/backup.ts
```diff
@@ -370,6 +373,51 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
     }
 }
 
+/**
+ * Implementation of {@link BackupDecryptor} for the rust crypto backend.
+ */
+export class RustBackupDecryptor implements BackupDecryptor {
```
andybalaam, line 379: Naive question: why is this not in `rust-crypto.ts`?

src/rust-crypto/rust-crypto.ts
```diff
@@ -151,7 +153,24 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
-        throw new Error("Stub not yet implemented");
+        if (backupInfo.algorithm != "m.megolm_backup.v1.curve25519-aes-sha2") {
+            throw new Error(`getBackupDecryptor Unsupported algorithm ${backupInfo.algorithm}`);
+        }
+
+        const authData = <Curve25519AuthData>backupInfo.auth_data;
+
+        if (!(privKey instanceof Uint8Array)) {
+            // eslint-disable-next-line @typescript-eslint/no-base-to-string
+            throw new Error(`getBackupDecryptor expects Uint8Array, got ${privKey}`);
```
andybalaam, line 164: Avoid logging the private key.

andybalaam (changes requested):
Looks good; a couple of questions and I support Rich's comment.

src/rust-crypto/backup.ts
BillCarsonFr: Moved to crypto-api/keybackup.ts like other similar utilities. Added re-export for backward compatibility.
I used the same convention for the other utilities where the `I` prefix is removed for new names.
Updated the new files to use the new import.

src/rust-crypto/rust-crypto.ts
BillCarsonFr: Done. I created a new base64.ts in CommonCrypto. I updated rust-crypto.ts to use it. 
I haven't touch legacy code that is still using the one from olmlib.

src/rust-crypto/backup.ts
BillCarsonFr: Yes it's encrypted (would be a ciphertext, mac, ephemeral).

src/rust-crypto/backup.ts
BillCarsonFr: It's in the rust backup.ts (mirrors the legacy that has it's implem in legacy backup.ts).
Feels like it belongs there, but no particular technical reason for rust has it just pass it to the wasm olmMachine. The legacy one access some internal classes though.

src/rust-crypto/rust-crypto.ts
BillCarsonFr: done

src/rust-crypto/backup.ts
andybalaam: ah, I missed the dir name!

andybalaam (approved):

PR #3715, opened 2023-09-07
Explain why synthetic receipts don't mark the room as read

andybalaam:
---


richvdh (changes requested):

richvdh:
This is a followup to https://github.com/matrix-org/matrix-js-sdk/pull/3706, for the record

PR #3716, opened 2023-09-07
Element-R: Emit `CryptoEvent.UserTrustStatusChanged` when user identity is updated

richvdh:
Fixes https://github.com/vector-im/crypto-internal/issues/148

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     ],
     "dependencies": {
         "@babel/runtime": "^7.12.5",
-        "@matrix-org/matrix-sdk-crypto-wasm": "^1.2.1",
+        "@matrix-org/matrix-sdk-crypto-wasm": "^1.2.3-alpha.0",
         "another-json": "^0.2.0",
         "bs58": "^5.0.0",
         "content-type": "^1.0.4",
--- a/spec/unit/rust-crypto/CrossSigningIdentity.spec.ts
+++ b/spec/unit/rust-crypto/CrossSigningIdentity.spec.ts
@@ -52,7 +52,7 @@ describe("CrossSigningIdentity", () => {
                 get: jest.fn(),
             } as unknown as Mocked<ServerSideSecretStorage>;
 
-            crossSigning = new CrossSigningIdentity(olmMachine, outgoingRequestProcessor, secretStorage, jest.fn());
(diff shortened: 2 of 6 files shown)


richvdh:
This is currently awaiting a release of `rust-sdk-crypto-wasm`, but more difficult than that: it introduces a reference cycle in the rust side.

Basically, once I call `OlmMachine::user_identities_stream`, the cryptostore isn't being dropped when the OlmMachine is dropped.

What seems to be happening is as follows:

- `user_identities_stream` takes a copy of the `VerificationMachine`. The new copy holds a ref to the `CryptoStoreWrapper`.
- The new copy is used in a `Stream::map` closure ... so is retained until the stream is dropped.
- I have a reader which reads the stream to exhaustion
- The stream reaches exhaustion when the _sender_ end of the `BroadcastStream` is dropped
- However, the sender is a member of `CryptoStoreWrapper`, so is never dropped, because of the reference in the `Stream::map` closure.

So in short, we have a reference cycle which stops the store being cleaned up.

richvdh:
> but more difficult than that: it introduces a reference cycle in the rust side.

This is fixed by https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/31

andybalaam (approved):
Looks good, as far as I understand it, which is not a great deal.

andybalaam:
I suspect the test failure was flakiness, re-adding to merge queue.

For the record, the failure was:

```
FAIL test/components/views/settings/JoinRuleSettings-test.tsx (7.445 s)
  ● <JoinRuleSettings /> › knock rooms › when room does not support join rule knock › upgrades room when changing join rule to knock

    expect(element).toBeInTheDocument()

    element could not be found in the document

      227 |
      228 |                 // update spaces
    > 229 |                 expect(await screen.findByText("Updating space...")).toBeInTheDocument();
          |                                                                      ^
      230 |
      231 |                 await flushPromises();
      232 |

      at Object.toBeInTheDocument (test/components/views/settings/JoinRuleSettings-test.tsx:229:70)
```

andybalaam:
If it merges, we should add a flaky test issue, similar but not identical to https://github.com/vector-im/element-web/issues/25625

PR #3717, opened 2023-09-08
Fix potential delay in sending out requests from the rust SDK

richvdh:
There was a potential race which could cause us to be very slow to send out pending HTTP requests, particularly when handling a user verification. Add some resilience to make sure we handle it correctly.

~~Based on #3716~~

---

### DIFF
--- a/src/rust-crypto/rust-crypto.ts
+++ b/src/rust-crypto/rust-crypto.ts
@@ -1250,6 +1250,19 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
     }
 
     /**
+     * Callback for `OlmMachine.registerUserIdentityUpdatedCallback`
+     *
+     * Called by the rust-sdk whenever there is an update to any user's cross-signing status. We re-check their trust
+     * status and emit a `UserTrustStatusChanged` event.
+     *
+     * @param userId - the user with the updated identity
+     */
+    public async onUserIdentityUpdated(userId: RustSdkCryptoJs.UserId): Promise<void> {
+        const newVerification = await this.getUserVerificationStatus(userId.toString());
+        this.emit(CryptoEvent.UserTrustStatusChanged, userId.toString(), newVerification);
+    }
+
+    /**
      * Handle a live event received via /sync.
      * See {@link ClientEventHandlerMap#event}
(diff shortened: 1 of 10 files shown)


andybalaam:
Updated from develop to make the diff clearer, since the based-on PR is merged. (Hope that's ok)

src/rust-crypto/rust-crypto.ts
```diff
@@ -1323,24 +1335,43 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
+            //
+            // There is potential for a race whereby the item is added *after* `OlmMachine.outgoingRequests` checks
+            // the queue, but *before* it returns. In such a case, the item could sit there unnoticed for some time.
+            //
+            // In order to circumvent the race, we set a flag which tells the loop to go round once again even if the
+            // queue appears to be empty.
+            this.outgoingRequestLoopOneMoreLoop = true;
             return;
         }
+        this.outgoingRequestLoopInner().catch((e) => {
```
andybalaam, line 1351: Worth a comment saying we are deliberately not awaiting this here?

src/rust-crypto/rust-crypto.ts
```diff
@@ -1323,24 +1335,43 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
+
+    private async outgoingRequestLoopInner(): Promise<void> {
+        /* istanbul ignore if */
+        if (this.outgoingRequestLoopRunning) {
+            throw new Error("Cannot run two outgoing request loops");
+        }
         this.outgoingRequestLoopRunning = true;
         try {
             while (!this.stopped) {
+                this.outgoingRequestLoopOneMoreLoop = false;
```
weeman1337, line 1364: Can you add another short comment here? Because if you oversee the `await` one may think that `this.outgoingRequestLoopOneMoreLoop` is always false in the `if` below.

andybalaam (approved):
Looks sensible and as described. Any way to test this?

weeman1337 (approved):
Looks good to me. I've also tested device verification on the deployment. Works.

I suppose that writing tests for this edge case is not easily doable?

PR #3722, opened 2023-09-12
Avoid an infinite loop in the interactive-auth test

andybalaam:
Reading the test "should fire stateUpdated callback with error when a request fails" it looks like this would always cause an infinite loop between doRequest and statusUpdated calls. I don't know why this wasn't a problem until we updated Jest to v29.5.4, but after that point it was.

This change should fix the test failures for the Jest upgrade PR: https://github.com/matrix-org/matrix-js-sdk/pull/3670 .

---


weeman1337 (approved):
:+1:

PR #3726, opened 2023-09-14
Cleanup, separate backup bootstrap and mgmt

BillCarsonFr:
Refactor crypto test, separate 4S and backup mgmt test.

No Change in behavior.
Just split a describe in a wrapping describe with two describes, and moved around common functions.

## Checklist

---

### DIFF
--- a/spec/integ/crypto/crypto.spec.ts
+++ b/spec/integ/crypto/crypto.spec.ts
@@ -2333,14 +2333,12 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string, 
                             type: "m.secret_storage.default_key",
                             content: {
                                 key: secretStorageKey,
-                                algorithm: SECRET_STORAGE_ALGORITHM_V1_AES,
                             },
                         },
                         // Needed for secretStorage.getKey or secretStorage.hasKey
                         {
                             type: `m.secret_storage.key.${secretStorageKey}`,
                             content: {
-                                key: secretStorageKey,
                                 algorithm: SECRET_STORAGE_ALGORITHM_V1_AES,
                             },
                         },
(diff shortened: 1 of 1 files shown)


spec/integ/crypto/crypto.spec.ts
```diff
@@ -2292,44 +2292,82 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
+        function sendSyncResponse(secretStorageKey: string) {
+            syncResponder.sendOrQueueSyncResponse({
+                next_batch: 1,
+                account_data: {
+                    events: [
+                        {
+                            type: "m.secret_storage.default_key",
+                            content: {
+                                key: secretStorageKey,
+                                algorithm: SECRET_STORAGE_ALGORITHM_V1_AES,
```
weeman1337, line 2336: The `default_key` does not have the `algorithm` field according [to the spec](https://spec.matrix.org/v1.8/client-server-api/#key-storage). Should this removed here to test nearer to reality?

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2292,44 +2292,82 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
+                            },
+                        },
+                        // Needed for secretStorage.getKey or secretStorage.hasKey
+                        {
+                            type: `m.secret_storage.key.${secretStorageKey}`,
+                            content: {
+                                key: secretStorageKey,
```
weeman1337, line 2343: The `key` property doesn't exist [according to the spec](https://spec.matrix.org/v1.8/client-server-api/#msecret_storagev1aes-hmac-sha2). Should this removed here to test nearer to reality?

weeman1337 (approved):
Looks good :+1:

Two comments in moved code about tests being spec compliant. It was already like this before. But maybe we can improve here.

spec/integ/crypto/crypto.spec.ts
BillCarsonFr: good catch!

spec/integ/crypto/crypto.spec.ts
BillCarsonFr: good catch!

PR #3729, opened 2023-09-18
Delete knocked room when knock membership changes

maheichyk:
This PR suggests to delete knocked room when knock request is:
- approved
- denied
- cancelled

This changes fix the issue that knocks get 'stuck' as described here: https://github.com/matrix-org/matrix-react-sdk/pull/11573#pullrequestreview-1619853865

## Checklist

Type: enhancement

---

## ✨ Features
 * Delete knocked room when knock membership changes ([\#3729](https://github.com/matrix-org/matrix-js-sdk/pull/3729)). Contributed by @maheichyk.

### DIFF
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -30,6 +30,7 @@ import {
     IHttpOpts,
     IToDeviceEvent,
     MatrixClient,
+    MatrixEvent,
     MatrixHttpApi,
     TypedEventEmitter,
 } from "../../../src";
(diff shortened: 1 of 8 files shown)


t3chguy (changes requested):
I think this breaks the purpose of the SyncAccumulator

```
The purpose of this class is to accumulate /sync responses such that a
complete "initial" JSON response can be returned which accurately represents
the sum total of the /sync responses accumulated to date. It only handles
room data: that is, everything under the "rooms" top-level key.
```

It is meant to imitate an `initialSync` call, nothing less nothing more.

maheichyk:
@t3chguy thanks for the comment, okay. Implementation has been changed to not store anything on leave. Knocked rooms are just deleted in a similar way it is already done in some cases in `sync-accumulator.ts`. This looks like a minimum implementation to fix behavior in the linked issue.

t3chguy (approved):
Looks good to me - thanks!!

PR #3730, opened 2023-09-18
Implement isSecretStorageReady in rust

BillCarsonFr:
Fixes https://github.com/vector-im/crypto-internal/issues/152

Udpated the integration test to be able to test it propery:
- Complete mocking to correctly update account data end-points and sync account data back
- Make test work on both rust/legacy when possible.

## Checklist

---

### DIFF
--- a/spec/integ/crypto/crypto.spec.ts
+++ b/spec/integ/crypto/crypto.spec.ts
@@ -39,6 +39,7 @@ import { TestClient } from "../../TestClient";
 import { logger } from "../../../src/logger";
 import {
     Category,
+    ClientEvent,
     createClient,
     CryptoEvent,
     IClaimOTKsResult,
@@ -68,7 +69,7 @@ import {
     mockSetupCrossSigningRequests,
     mockSetupMegolmBackupRequests,
 } from "../../test-utils/mockEndpoints";
-import { AddSecretStorageKeyOpts, SECRET_STORAGE_ALGORITHM_V1_AES } from "../../../src/secret-storage";
+import { AddSecretStorageKeyOpts } from "../../../src/secret-storage";
 import { CrossSigningKey, CryptoCallbacks, KeyBackupInfo } from "../../../src/crypto-api";
 import { E2EKeyResponder } from "../../test-utils/E2EKeyResponder";
 import { DecryptionError } from "../../../src/crypto/algorithms";
@@ -2292,9 +2293,56 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string, 
             privateKey: Uint8Array.of(32, 33),
         });
 
+        /**
+         * The account data events to be returned by the sync.
+         * Will be updated when fecthMock intercepts calls to PUT `/_matrix/client/v3/user/:userId/account_data/`.
+         * Will be used by `sendSyncResponseWithUpdatedAccountData`
+         */
+        let accountDataEvents: any[] = [];
+
+        function updateAccountData(type: string, content: any) {
+            const existing = accountDataEvents.find((event) => {
+                if (event.type == type) return event;
+            });
+            if (existing) {
+                existing.content = content;
+            } else {
+                accountDataEvents.push({
+                    type,
+                    content,
+                });
+            }
+        }
+
+        function mockGetAccountData() {
+            fetchMock.get(
+                `path:/_matrix/client/v3/user/:userId/account_data/:type`,
+                (url) => {
+                    const type = url.split("/").pop();
+                    const existing = accountDataEvents.find((event) => {
+                        if (event.type == type) return event;
+                    });
+                    if (existing) {
+                        // return it
+                        return {
+                            status: 200,
+                            body: existing.content,
+                        };
+                    } else {
+                        // 404
+                        return {
+                            status: 404,
+                            body: { errcode: "M_NOT_FOUND", error: "Account data not found." },
+                        };
+                    }
+                },
+                { overwriteRoutes: true },
+            );
+        }
+
         beforeEach(async () => {
             createSecretStorageKey.mockClear();
-
+            accountDataEvents = [];
             expectAliceKeyQuery({ device_keys: { "@alice:localhost": {} }, failures: {} });
             await startClientAndAwaitFirstSync();
         });
@@ -2311,6 +2359,9 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string, 
                     `express:/_matrix/client/v3/user/:userId/account_data/m.cross_signing.${key}`,
                     (url: string, options: RequestInit) => {
                         const content = JSON.parse(options.body as string);
+                        const type = url.split("/").pop();
+                        // update account data for sync response
+                        updateAccountData(type!, content);
                         resolve(content.encrypted);
                         return {};
                     },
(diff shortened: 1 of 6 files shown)


src/rust-crypto/rust-crypto.ts
```diff
@@ -615,7 +615,31 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
      * Implementation of {@link CryptoApi#isSecretStorageReady}
      */
     public async isSecretStorageReady(): Promise<boolean> {
-        return false;
+        const secretStorageKeyTuple = await this.secretStorage.getKey();
+        if (!secretStorageKeyTuple) return false;
+
+        const [keyId] = secretStorageKeyTuple;
+
+        // Get the master signing keys stored into the secret storage
```
richvdh, line 623: This doesn't actually get the master key itself. Rather, it gets the key that is used to encrypt the master key.

```suggestion
        // Get the keys that were used to encrypt the master cross-signing key in secret storage
```

likewise the others.

richvdh (changes requested):
`bootstrapSecretStorage` uses a function called `secretStorageContainsCrossSigningKeys` which does *most* of this logic. (In particular, it doesn't check that the keys are encrypted with the default key. It also doesn't check that backup key is stored.)

It feels like we ought to be using the same logic in both places, and that we *shouldn't* be duplicating the code. In particular, the difference means that this statement on the documentation of `isSecretStorageReady` may not be true:

> That is to say, after `bootstrapSecretStorage()` completes successfully, this function should
return true.

Possibly we should update `secretStorageContainsCrossSigningKeys` to check that the keys are encrypted by the *default* 4S key?

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2324,27 +2375,18 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
-                        {
-                            type: `m.secret_storage.key.${secretStorageKey}`,
-                            content: {
-                                algorithm: SECRET_STORAGE_ALGORITHM_V1_AES,
-                            },
-                        },
-                    ],
-                },
-            });
+        function sendSyncResponseWithUpdatedAccountData() {
```
richvdh, line 2378: doc-comment on this no longer seems to be accurate .

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2324,27 +2375,18 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
+            try {
+                syncResponder.sendOrQueueSyncResponse({
+                    next_batch: 1,
+                    account_data: {
+                        events: accountDataEvents,
+                    },
+                });
+            } catch (err) {
+                // It's ok if it fails here, the sync response is cumulative and will contain
+                // the latest account data.
```
richvdh, line 2388: I am somewhat surprised that `sendOrQueueSyncResponse` fails. why is that happening?

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2422,32 +2482,41 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
+            sendSyncResponseWithUpdatedAccountData();
+
             // Finally ensure backup is working
             await aliceClient.getCrypto()!.checkKeyBackupAndEnable();
+
             await backupStatusUpdate;
         }
 
         describe("bootstrapSecretStorage", () => {
+            // Doesn't work with legacy. Legacy will try to bootstrap even without private key which is buggy.
```
richvdh, line 2503:
```suggestion
            // Doesn't work with legacy crypto, which will try to bootstrap even without private key, which is buggy.
```

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2456,52 +2525,55 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
-                    // Call again bootstrapSecretStorage
-                    await aliceClient.getCrypto()!.bootstrapSecretStorage({ createSecretStorageKey });
+                // Wait for bootstrapSecretStorage to finished
+                await bootstrapPromise;
 
-                    // createSecretStorageKey should be called only on the first run of bootstrapSecretStorage
-                    expect(createSecretStorageKey).toHaveBeenCalledTimes(1);
-                },
-            );
+                // On legacy we need to wait for ClientEvent.AccountData before calling bootstrap again.
```
richvdh, line 2557:
```suggestion
                // On legacy crypto we need to wait for ClientEvent.AccountData before calling bootstrap again.
```

richvdh (changes requested):

src/rust-crypto/secret-storage.ts
```diff
@@ -17,28 +17,49 @@ limitations under the License.
...
+
+    const records = [];
+
+    for (let index = 0; index < secretNames.length; index++) {
+        const secretName = secretNames[index];
+        const record = (await secretStorage.isStored(secretName)) || {};
+        records.push(record);
+    }
+
+    return records.every((record) => secretKeyId in record);
```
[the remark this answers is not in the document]
BillCarsonFr: false if secretNames empty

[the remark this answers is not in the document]
richvdh: huh?

spec/integ/crypto/crypto.spec.ts
BillCarsonFr: It fails with `Cannot queue more than one /sync response` with current tests.
Which is not problematic in our case.

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2292,9 +2293,56 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
             privateKey: Uint8Array.of(32, 33),
         });
 
+        /**
+         * The account data events to be returned by the sync.
+         * Will be updated when fecthMock intercepts calls to PUT `/_matrix/client/v3/user/:userId/account_data/`.
+         * Will be used by `sendSyncResponseWithUpdatedAccountData`
+         */
+        let accountDataEvents: any[] = [];
```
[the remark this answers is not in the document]
BillCarsonFr: sure

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2283,6 +2284,20 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
     describe("Secret Storage and Key Backup", () => {
+        /**
+         * The account data events to be returned by the sync.
+         * Will be updated when fecthMock intercepts calls to PUT `/_matrix/client/v3/user/:userId/account_data/`.
+         * Will be used by `sendSyncResponseWithUpdatedAccountData`
+         */
+        let accountDataEvents: Map<String, any>;
+
+        beforeEach(async () => {
+            createSecretStorageKey.mockClear();
```
richvdh, line 2295: this is a weird forward reference to `createSecretStorageKey` before it is declared. It would be normal to put the declaration *before* `beforeEach`.

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2292,12 +2307,33 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
             privateKey: Uint8Array.of(32, 33),
         });
 
-        beforeEach(async () => {
-            createSecretStorageKey.mockClear();
+        function updateAccountData(type: string, content: any) {
+            accountDataEvents.set(type, content);
+        }
```
richvdh, line 2312: inline this?

spec/integ/crypto/crypto.spec.ts
richvdh: > It fails with `Cannot queue more than one /sync response` with current tests.

hrm. could you record that in a comment?

src/rust-crypto/rust-crypto.ts
```diff
@@ -623,19 +623,21 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
     public async isSecretStorageReady(): Promise<boolean> {
+        // make sure that the cross-signing keys are stored
         const secretsToCheck = [
             "m.cross_signing.master",
             "m.cross_signing.user_signing",
             "m.cross_signing.self_signing",
         ];
 
+        // if key backup is active, we also need to check that the backup decryption key is stored
         const keyBackupEnabled = (await this.backupManager.getActiveBackupVersion()) != null;
```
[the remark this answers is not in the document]
BillCarsonFr: sure

spec/unit/rust-crypto/secret-storage.spec.ts
```diff
@@ -51,19 +56,73 @@ describe("secret-storage", () => {
...
+            const secretStorage = {
+                isStored: jest.fn().mockResolvedValue({ defaultKey: {} }),
+                getDefaultKeyId: jest.fn().mockResolvedValue("anotherCommonKey"),
+            } as unknown as ServerSideSecretStorage;
+
+            const result = await secretStorageContainsCrossSigningKeys(secretStorage);
+            expect(result).toBeFalsy();
+        });
+
+        it("Check canAccessSecrets", async () => {
```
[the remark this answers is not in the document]
BillCarsonFr: sure

spec/integ/crypto/crypto.spec.ts
BillCarsonFr: sure

spec/integ/crypto/crypto.spec.ts
BillCarsonFr: sure

richvdh (changes requested):
there are still unaddressed comments

richvdh (approved):
hrm, looks like I was confused.

LGTM now. Though CI is failing.

PR #3731, opened 2023-09-18
Implement exportCrossSigningKeysToStorage

BillCarsonFr:
Fixes https://github.com/vector-im/crypto-internal/issues/153
Fixes https://github.com/vector-im/element-web/issues/25781

`exportCrossSigningKeysToStorage` was not implemented

## Checklist

---

### DIFF
--- a/spec/unit/rust-crypto/CrossSigningIdentity.spec.ts
+++ b/spec/unit/rust-crypto/CrossSigningIdentity.spec.ts
@@ -41,6 +41,7 @@ describe("CrossSigningIdentity", () => {
             olmMachine = {
                 crossSigningStatus: jest.fn(),
                 bootstrapCrossSigning: jest.fn(),
+                exportCrossSigningKeys: jest.fn(),
                 close: jest.fn(),
             } as unknown as Mocked<RustSdkCryptoJs.OlmMachine>;
 
@@ -50,6 +51,8 @@ describe("CrossSigningIdentity", () => {
 
             secretStorage = {
                 get: jest.fn(),
+                hasKey: jest.fn(),
+                store: jest.fn(),
             } as unknown as Mocked<ServerSideSecretStorage>;
 
             crossSigning = new CrossSigningIdentity(olmMachine, outgoingRequestProcessor, secretStorage);
@@ -61,7 +64,8 @@ describe("CrossSigningIdentity", () => {
                 hasSelfSigning: true,
                 hasUserSigning: true,
             });
-            // TODO: secret storage
+            // in secret storage
+            secretStorage.get.mockResolvedValue("base64-saved-in-storage");
             await crossSigning.bootstrapCrossSigning({});
             expect(olmMachine.bootstrapCrossSigning).not.toHaveBeenCalled();
             expect(outgoingRequestProcessor.makeOutgoingRequest).not.toHaveBeenCalled();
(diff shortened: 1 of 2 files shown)


andybalaam (approved):
Looks good, thanks!

src/rust-crypto/CrossSigningIdentity.ts
```diff
@@ -108,13 +116,20 @@ export class CrossSigningIdentity {
      *   * Upload the private keys to SSSS, if it is set up
      */
     private async resetCrossSigning(authUploadDeviceSigningKeys?: UIAuthCallback<void>): Promise<void> {
+        // XXX: We must find a way to make this atomic, currently if the user do not remember is account password
+        // or 4S passphrase/key the process will fail in a bad state, with keys rotated but not uploaded or saved in 4S.
         const outgoingRequests: Array<OutgoingRequest> = await this.olmMachine.bootstrapCrossSigning(true);
 
+        // If 4s is configured we need to udpate it.
+        const hasKey = await this.secretStorage.hasKey();
+        if (hasKey) {
```
richvdh, line 125: should this be called `secretStorageIsReady` or something, rather than `hasKey`? "key" is a bit unclear: there are lots of keys involved here.

Or just inline it:

```suggestion
        if (await this.secretStorage.hasKey()) {
```

src/rust-crypto/CrossSigningIdentity.ts
```diff
@@ -123,6 +138,9 @@ export class CrossSigningIdentity {
      * (If secret storage is *not* configured, we assume that the export will happen when it is set up)
      */
     private async exportCrossSigningKeysToStorage(): Promise<void> {
-        // TODO
+        const exported: RustSdkCryptoJs.CrossSigningKeyExport = await this.olmMachine.exportCrossSigningKeys();
+        this.secretStorage.store("m.cross_signing.master", exported.masterKey!);
+        this.secretStorage.store("m.cross_signing.self_signing", exported.self_signing_key!);
+        this.secretStorage.store("m.cross_signing.user_signing", exported.userSigningKey!);
```
richvdh, line 144: how do we know that these ! assertions are safe?

spec/unit/rust-crypto/CrossSigningIdentity.spec.ts
```diff
@@ -73,6 +77,19 @@ describe("CrossSigningIdentity", () => {
             expect(olmMachine.bootstrapCrossSigning).toHaveBeenCalledWith(true);
         });
 
+        it("Shoud update 4S on reset if 4S is set", async () => {
```
richvdh, line 80:
```suggestion
        it("Shoud update 4S on reset if 4S is set up", async () => {
```

richvdh (changes requested):

src/rust-crypto/CrossSigningIdentity.ts
```diff
@@ -108,13 +116,20 @@ export class CrossSigningIdentity {
...
      */
     private async resetCrossSigning(authUploadDeviceSigningKeys?: UIAuthCallback<void>): Promise<void> {
+        // XXX: We must find a way to make this atomic, currently if the user do not remember is account password
+        // or 4S passphrase/key the process will fail in a bad state, with keys rotated but not uploaded or saved in 4S.
         const outgoingRequests: Array<OutgoingRequest> = await this.olmMachine.bootstrapCrossSigning(true);
 
+        // If 4s is configured we need to udpate it.
+        const hasKey = await this.secretStorage.hasKey();
+        if (hasKey) {
+            // Update 4S before uploading cross-signing keys
```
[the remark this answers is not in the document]
BillCarsonFr: Sure, will do.
It's to stay consistent with legacy that will first ask the 4S passphrase (to update 4S) then the account password (to upload keys). In case we have some tests on the app side

src/rust-crypto/CrossSigningIdentity.ts
BillCarsonFr: It's a private function that we call when we know we have secrets.
Do you think I should be lenient and just log if it's not stored?

BillCarsonFr: Updated, and now logging an error if secret parts unknown

src/rust-crypto/CrossSigningIdentity.ts
```diff
@@ -138,9 +139,21 @@ export class CrossSigningIdentity {
      * (If secret storage is *not* configured, we assume that the export will happen when it is set up)
      */
     private async exportCrossSigningKeysToStorage(): Promise<void> {
-        const exported: RustSdkCryptoJs.CrossSigningKeyExport = await this.olmMachine.exportCrossSigningKeys();
-        this.secretStorage.store("m.cross_signing.master", exported.masterKey!);
-        this.secretStorage.store("m.cross_signing.self_signing", exported.self_signing_key!);
-        this.secretStorage.store("m.cross_signing.user_signing", exported.userSigningKey!);
+        const exported: RustSdkCryptoJs.CrossSigningKeyExport | null = await this.olmMachine.exportCrossSigningKeys();
+        if (exported?.masterKey) {
```
richvdh, line 143: Unfortunately these aren't covered by tests, so the quality gate is failing.

You can try something like:

```suggestion
        /* istanbul ignore else (this function is only called when we know the olm machine has keys) */
        if (exported?.masterKey) {
```

richvdh (approved):
lgtm

PR #3732, opened 2023-09-18
Element-R: use the pickleKey to encrypt the crypto store

richvdh:
`pickleKey` is a passphrase set by the application for this express purpose.

Fixes https://github.com/vector-im/element-web/issues/24967

---


robintown (approved):
Makes sense

PR #3734, opened 2023-09-19
Sort imports in MatrixClient

richvdh:
My IDE keeps trying to do this, so let's just do it in its own PR

---


richvdh:
> This is just going to happen again in the future - can/should we adopt this as a code style/lint rule?

I'd say we should (generally having a consistent sorting for imports helps reduce noise in diffs) but I don't particularly want to go on a diversion to do it right now. Can we just land this as an improvement in its own right?

t3chguy (approved):
😶

PR #3743, opened 2023-09-20
Fix the warning for messages from unsigned devices

richvdh:
https://github.com/matrix-org/matrix-js-sdk/pull/3693 attempted to port the logic for event shields from react-sdk to js-sdk, but I seem to have messed up this case.

I've double checked all the logic and... it's hopefully right this time.

---

## 🐛 Bug Fixes
 * Fix the warning for messages from unsigned devices ([\#3743](https://github.com/matrix-org/matrix-js-sdk/pull/3743)).

### DIFF
--- a/src/crypto/index.ts
+++ b/src/crypto/index.ts
@@ -2738,6 +2738,8 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
         const senderId = event.getSender();
         if (!senderId || encryptionInfo.mismatchedSender) {
             // something definitely wrong is going on here
+
+            // previously: E2EState.Warning -> E2ePadlockUnverified -> Red/"Encrypted by an unverified session"
             return {
                 shieldColour: EventShieldColour.RED,
                 shieldReason: EventShieldReason.MISMATCHED_SENDER_KEY,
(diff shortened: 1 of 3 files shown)


t3chguy (approved):

src/crypto/index.ts
```diff
@@ -2738,6 +2738,8 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
         const senderId = event.getSender();
         if (!senderId || encryptionInfo.mismatchedSender) {
             // something definitely wrong is going on here
+
+            // previously: E2EState.Warning -> E2ePadlockUnverified -> Red/"Encrypted by an unverified session"
```
richvdh, line 2742: these "previously" comments relate to the old logic in `EventTile` (https://github.com/matrix-org/matrix-react-sdk/blob/v3.80.1/src/components/views/rooms/EventTile.tsx#L579-L637). I found them useful for double-checking the logic.

src/rust-crypto/rust-crypto.ts
```diff
@@ -1649,6 +1649,7 @@ function rustEncryptionInfoToJsEncryptionInfo(
     if (shieldState.message === null) {
         shieldReason = null;
     } else if (shieldState.message === "Encrypted by an unverified user.") {
+        // this case isn't actually used with lax shield semantics.
```
richvdh, line 1652: this is unrelated, but something I came across while I was figuring out what happened and it doesn't feel like worth making a separate PR for

PR #3751, opened 2023-09-25
Prevent optional arguments in existing re-emitted event handlers

davidisaaclee:
## Checklist

This is a redo of https://github.com/matrix-org/matrix-js-sdk/pull/3748

## Motivation

Re-emitting events inserts the source as an extra argument after other arguments
https://github.com/matrix-org/matrix-js-sdk/blob/9b2f22692ec35a804c6214f6f6691f04868163e6/src/ReEmitter.ts#L40-L42
Some handlers have optional arguments at the end of their signature
https://github.com/matrix-org/matrix-js-sdk/blob/6385c9c0dab8fe67bd3a8992a4777f243fdd1b68/src/models/room.ts#L283-L288
and are called with those optional arguments omitted
https://github.com/matrix-org/matrix-js-sdk/blob/6385c9c0dab8fe67bd3a8992a4777f243fdd1b68/src/models/room.ts#L2517
When a handler receives this re-emitted event, the expected `oldEventId` will instead be the source `Room`.
```ts
client.addListener(
  RoomEvent.LocalEchoUpdated, 
  (event: MatrixEvent, room: Room, oldEventId?: string) => {
    // will be true during the `emit` linked above, making types a lie
    assert(oldEventId instanceof Room); 
  })
```

## Changes
- [6f3c577](https://github.com/matrix-org/matrix-js-sdk/pull/3751/commits/6f3c577e1c900215540d41f8c997cebdf519def5) Removed optional arguments from all type signatures of event handlers for re-emitted events. (I did this by searching the codebase for `ReEmitter` and `TypedReEmitter`, and then looking through all of the event handler maps referenced in the type param.)
- [59ea6e7](https://github.com/matrix-org/matrix-js-sdk/pull/3751/commits/59ea6e7feef6e14002cfb0ed97bba53fc4602328) Updated call sites of all event handlers to match new types. (I did this by running `yarn lint:types` then fixing all the type errors caused by the previous commit.)

I think this is more comprehensive and _more_ robust than https://github.com/matrix-org/matrix-js-sdk/pull/3748, but this can easily regress if someone adds or modifies a re-emitted event handler signature to have an optional argument. I can't think of a solution for this (only thought is some lint rule, but I'm not sure if eslint would be able to know "this is an event handler signature").

---

## ✨ Features
 * Prevent optional arguments in existing re-emitted event handlers ([\#3751](https://github.com/matrix-org/matrix-js-sdk/pull/3751)). Contributed by @davidisaaclee.

### DIFF
--- a/spec/unit/ToDeviceMessageQueue.spec.ts
+++ b/spec/unit/ToDeviceMessageQueue.spec.ts
@@ -31,7 +31,7 @@ describe("onResumedSync", () => {
         onSendToDeviceSuccess = () => {};
         resumeSync = (newState, oldState) => {
             shouldFailSendToDevice = false;
-            mockClient.emit(ClientEvent.Sync, newState, oldState);
+            mockClient.emit(ClientEvent.Sync, newState, oldState, undefined);
         };
 
         store = new StubStore();
--- a/spec/unit/event-timeline-set.spec.ts
+++ b/spec/unit/event-timeline-set.spec.ts
@@ -303,8 +303,8 @@ describe("EventTimelineSet", () => {
                     messageEventIsDecryptionFailureSpy.mockReturnValue(true);
                     replyEventIsDecryptionFailureSpy.mockReturnValue(true);
 
-                    messageEvent.emit(MatrixEventEvent.Decrypted, messageEvent);
-                    replyEvent.emit(MatrixEventEvent.Decrypted, replyEvent);
+                    messageEvent.emit(MatrixEventEvent.Decrypted, messageEvent, undefined);
+                    replyEvent.emit(MatrixEventEvent.Decrypted, replyEvent, undefined);
 
                     // simulate decryption
                     messageEventIsDecryptionFailureSpy.mockReturnValue(false);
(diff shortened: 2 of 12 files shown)


davidisaaclee:
@t3chguy How would you write a lint rule for this? I could add a rule to prevent functions from having optional arguments, but that seems too restrictive for other type signatures.

I don't have experience with writing type-aware tslint rules, but the only thing I can think of is:
- add a new empty `IEventHandlerMap` interface (`interface IEventHandlerMap { /* no members */ }`)
- make all existing handler maps extend this interface (`interface ClientEventHandlerMap extends IEventHandlerMap { ... }`)
- when linting with type info: if a function definition is inside an interface that extends `IEventHandlerMap`, apply the "forbid optional parameters" rule
- Add a type parameter to `ReEmitter` / `TypedReEmitter` that the handler map type param must extend `IEventHandlerMap`
  - I think this restriction won't have any effect with an empty `IEventHandlerMap` due to TS's structural typing. To fix this, I think we could do a nominal typing hack, e.g. `interface IEventHandlerMap { __extendsIEventHandlerMap: void }` – not sure.

Drawback: afaict, type-aware linting is not currently enabled on matrix-js-sdk. Without empirical data, I think enabling type-aware linting will reduce lint performance. https://typescript-eslint.io/linting/typed-linting/#how-is-performance says "By including parserOptions.project in your config, you incur the performance penalty of asking TypeScript to do a build of your project before ESLint can do its linting."

t3chguy:
Fair

Maybe we just remove the re-emitter emitting the original emitter. It doesn't match the types so most consumers won't be able to make use of it anyway

davidisaaclee:
>Maybe we just remove the re-emitter emitting the original emitter

I'm not sure if or when I'll be able to make that change, as this PR is just a change I made in my fork that fixed the problem I was encountering. Closing this for now if this isn't the desired solution.

Issue #1156, opened 2020-01-18
Example of how to implement e2ee


menturion:
Currently, there are a few examples of how to use the SDK (e.g. VoIP).
It would be very helpful if you would also provide a detailed example for implementing e2ee.

Ishaan28malik:
I would like to work on the same, Can you suggest where am I supposed to push the detailed examples?

cedricvanrompay:
I wrote https://github.com/matrix-org/matrix-js-sdk/blob/6aa5d39357c114490076dba00482e16abe03e5ee/examples/crypto-browser/olm-device-export-import.js

@menturion did you find this file during your research ? If not maybe we should add a link to it. If you did can you tell us which parts you don't understand or what you think is missing ?

menturion:
@cedricvanrompay 

Many thanks for the info and the E2EE example. I tested it a while and I'm constantly getting the below error message ...

> Error sending event UnknownDeviceError: This room contains unknown devices which have not been verified. We strongly recommend you verify them before continuing.

... when sending e.g. a text message (`client.sendTextMessage()`) although the member devices of a given room should have been verified through that snippet ...

```
        // Marking all devices as verified
        let room = this.getRoom(roomId);
        let members = (await room.getEncryptionTargetMembers()).map(x => x["userId"])
        let memberkeys = await this.downloadKeys(members);
        for (const userId in memberkeys) {
          for (const deviceId in memberkeys[userId]) {
            await this.setDeviceVerified(userId, deviceId);
          }
        }
```

Does your code only work with certain (>=) release versions of Synapse and/or JS SDK?

DhairyaBahl:
@menturion This issue seems to be inactive from a long time. Kindly assign me this issue if its still available.

menturion:
@DhairyaBahl 

I think that this is still needed. 
IMHO, such an example should cover bootstrapping secure storage and cross-signing, DB startup, device listeners, cryptoCallbacks etc.

DhairyaBahl:
@menturion OK I will work on this issue and make a PR asap. although i don't understand the issue but its docs labelled so i will give a try on this and will take help from the other comments. 

Thanks :)

murlock1000:
In https://github.com/murlock1000/matrix-js-bootstrap I have implemented scripts on:

-  Bootstrapping Cross-signing
-  Bootstrapping Secure Storage
-  Cross-signing key recovery from Secure Storage
[...]

Issue #437, opened 2017-05-16
Docs and example for using e2e


ghost:
Being new to matrix-js-sdk, I am attempting to build a bot which supports e2e. Please excuse me if I'm asking obvious questions. 

I see that there are multiple well-explained examples for creating clients. However, I do have some issues understanding how to apply e2e in the code. 

The client works well when it comes to sending messages (having added the session store/``WebStorageSessionStore`` and device ID to the client config), but receiving messages from other users in the room ends with a ``Unable to decrypt: The sender's device has not sent us the keys for this message.``. I also see that the following is created in ``session.e2e.rooms<room>``:

    {"algorithm":"m.megolm.v1.aes-sha2"}

Also, the device list for each participating user seems to come down fine. 

What in the [implementation guide](http://matrix.org/docs/guides/e2e_implementation.html) is automatically done in the SDK, and what is required by the developer?

If anyone has a code example, that would be great for the docs I think.

ara4n:
As a quick answer, i think that https://matrix.org/docs/projects/other/matrix-recorder.html is using js-sdk with e2e, so you can probably take inspiration from that (or failing that Riot).  Most of the e2e implementation guide is already covered by the js-sdk, so relatively little should be needed in addition.

N.B. that there is a risk of hitting https://github.com/vector-im/riot-web/issues/3231 unless you persist deviceId in your app between sessions.

tommyskg:
That worked perfectly, and it's a great example btw!

Based on you last comment - it occurred to me that I obviously should not make up my own device ID either, so that is something take a note of.

Hopefully this can be helpful to the next one wondering about the same thing.

Thanks a lot!

theilgaard:
I got stuck in the exact same place as you. 

Also, the config object is missing a documentation/declaration when using `MatrixClient.setRoomEncryption(roomId, config)`. I sifted through the code to find that it had to look like `{"algorithm":"m.megolm.v1.aes-sha2"}`

nahakiole:
I'm in the same place stuck, there are no examples how to verify and upload keys from new clients. Also the whole encryption itself could be explained a bit more clear maybe graphically in the matrix docs. But with the sdk it's very unclear what we as developers have to implement to get the same interface as in elementor using emojis and qr codes to sign new keys / upload / download existing keys.

[...]

PR #3754, opened 2023-09-27
E2EE text messaging example.

RhinoDevel:
Shows how to prepare an existing room for end-to-end-encryption (E2EE) and send a text message.

Addresses issues https://github.com/matrix-org/matrix-js-sdk/issues/437 and https://github.com/matrix-org/matrix-js-sdk/issues/1156.

Signed-off-by: Marcel Timm <RhinoDevel@users.noreply.github.com>

---

## ✨ Features
 * E2EE text messaging example. ([\#3754](https://github.com/matrix-org/matrix-js-sdk/pull/3754)). Contributed by @RhinoDevel.


examples/e2ee/main.mjs
```diff
@@ -0,0 +1,126 @@
+// Description:
```
dbkr, line 1: Don't forget license header and copyright attribution (to yourself)

examples/e2ee/main.mjs
```diff
@@ -0,0 +1,126 @@
...
+//
+// - A user must exist (enter credentials below).
+// - A room must exist, where the existing user is a joined member (enter room
+//   ID, below).
+//
+// Tested with:
+//
+// - NodeJS v18.17.1
+// - matrix-js-sdk v28.2.0
+// - olm v3.1.5
```
dbkr, line 16: This probably wants to be a block comment?

examples/e2ee/main.mjs
```diff
@@ -0,0 +1,126 @@
...
+        rawUrl: 'matrix.org',
+
+        roomId: '********',
+        msg: 'Hello World!'
+    };
+
+function log(msg)
+{
+    console.log(`${sLogPrefix}${msg}`);
+};
```
dbkr, line 34: This whole log wrapper and prefix feels weird and distracting for an example, especially since the prefix is hardcoded.

examples/e2ee/main.mjs
```diff
@@ -0,0 +1,126 @@
...
+//
+// - NodeJS v18.17.1
+// - matrix-js-sdk v28.2.0
+// - olm v3.1.5
+
+import * as sdk from 'matrix-js-sdk';
+import olm from 'olm';
+
+const sLogPrefix = '*** MAIN ***: ',
+    sC = {
```
dbkr, line 22: What does `sC` mean?

examples/e2ee/main.mjs
```diff
@@ -0,0 +1,126 @@
...
+                        algorithm: 'm.megolm.v1.aes-sha2' // Seems to be OK..
+                    });
+            }
+            catch(roomEncErr)
+            {
+                log(`Error: \"${roomEncErr.message}\"!`);
+                process.exit(1);
+            }
+
+            // Marking all devices as verified:
```
dbkr, line 75: Okay, but... why?

dbkr (changes requested):
A few specific points but I think this would need a bit of work generally before being an example we could accept. Some more general points:
 * Comments, comments, comments! This is documentation so it needs to be explaining everything *what* it's doing and *why* in detail so the reader understands the process. eg. why does it need to wait for the 'PREPARED' sync state?
 * It really ought to conform to our code style. In fact it would be nice if it got checked by the linters.
 * It would be even nicer if it got run as part of CI to make sure it still worked every time
 * We don't tend to use the 'mjs' extension - is it necessary? Why?
 * You've use MemoryCryptoStore: this is great for this example but again, needs a comment to say why it's appropriate here and what the limitations are.
 * Some doc of how to run it would feel useful too.

PR #3756, opened 2023-09-28
make leaveRoomSession async

toger5:
This does not resolve the promise until the event is actually send. No network connection would make awaiting on this blocking.

## Checklist

---

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -186,7 +186,7 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
      * This will not unsubscribe from updates: remember to call unsubscribe() separately if
      * desired.
      */
-    public leaveRoomSession(): void {
+    public async leaveRoomSession(): Promise<void> {
         if (!this.isJoined()) {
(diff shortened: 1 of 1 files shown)


dbkr (commented):
The description of this PR is great and documents what the contract of this function is. Could it go in the docs for the function.? :)

Also is this definitely what we want? If you lose your net connection will you just be unable to leave?

toger5:
I am not sure. To some extend this is how it is. You cannot leave (remove the memberships) if there is no network connection.
I can imagine that its frontend logic to also provide a timeout (`await Promise.all([endCall, timeout(20sek)])`) but the end call function only returns if it finally and actually managed to end the call.

dbkr:
Yeah, this probably does make sense for the js-sdk part. It would be nice to have the function docced though.

toger5:
are u imagining: 
```
/*
* Will try to leave a rtc session and do periodic retries. If a timeout is provided this will resolve after the timeout even if the 
* leave was not sucessful. The success is indicated with the bool.
*/
leaveRoomSession(timeout: Number|undefined = undefined): Future<bool>
```

dbkr:
Yep, and ideally be explicit about when it resolved with no timeout provided, eg. "resolves once the updated member event has been sent, or if timeout is a positive number, after this time or whichever is sooner."

dbkr (approved):

PR #3758, opened 2023-09-29
Clean up integ tests for incoming user verification

richvdh:
Move the tests into `verification.spec.ts`, enable for both stacks, and other cleanups.

Each commit should stand alone.

---


richvdh:
@florianduros I think this was mostly your code to start with so you might be best placed to look at it?

florianduros (approved):
Thanks for cleaning my mess!

PR #3759, opened 2023-09-29
Remove browserify builds

t3chguy:
Notes: Remove browserify builds

---

## 🦖 Deprecations
 * Remove browserify builds ([\#3759](https://github.com/matrix-org/matrix-js-sdk/pull/3759)).


germain-gg (changes requested):
The cleanup look good.
However it seems that this will break the `examples/browser` as it symlinks the browserify build for them to work. 

We might need to update that too

germain-gg (approved):
Looks great, thanks!

t3chguy:
Force merging due to flaky CI

PR #3761, opened 2023-09-29
Make MSC3906 implementation compatible with Rust Crypto

hughns:
This is a work in progress

## Checklist

### DIFF
--- a/src/rendezvous/MSC3906Rendezvous.ts
+++ b/src/rendezvous/MSC3906Rendezvous.ts
@@ -23,6 +23,7 @@ import { DeviceInfo } from "../crypto/deviceinfo";
 import { buildFeatureSupportMap, Feature, ServerSupport } from "../feature";
 import { logger } from "../logger";
 import { sleep } from "../utils";
+import { CrossSigningKey } from "../crypto-api";
 
 enum PayloadType {
     Start = "m.login.start",
@@ -181,7 +182,8 @@ export class MSC3906Rendezvous {
     private async verifyAndCrossSignDevice(
         deviceInfo: DeviceInfo,
     ): Promise<CrossSigningInfo | DeviceInfo | ICrossSigningKey | undefined> {
-        if (!this.client.crypto) {
+        const crypto = this.client.getCrypto();
+        if (!crypto) {
             throw new Error("Crypto not available on client");
         }
 
@@ -203,9 +205,11 @@ export class MSC3906Rendezvous {
         }
         // mark the device as verified locally + cross sign
         logger.info(`Marking device ${this.newDeviceId} as verified`);
-        const info = await this.client.crypto.setDeviceVerification(userId, this.newDeviceId, true, false, true);
 
-        const masterPublicKey = this.client.crypto.crossSigningInfo.getId("master")!;
+        // FIXME: this function needs to be plumbed in
+        const info = await this.client.crypto!.setDeviceVerification(userId, this.newDeviceId, true, false, true);
+
+        const masterPublicKey = (await crypto.getCrossSigningKeyId(CrossSigningKey.Master)) ?? undefined;
 
         await this.send({
             type: PayloadType.Finish,


hughns:
@BillCarsonFr @richvdh If helpful I can progress this PR to get the simple changes that work when moving from `client.crypto` to `client.getCypto()`. Please let me know if you would like me to do so?

richvdh:
I'm not sure I have full context on this (maybe @BillCarsonFr knows more?) and I'm not quite clear what you're proposing, but any contributions that increase compatibility with Rust Crypto sound helpful!

richvdh:
Opened an issue to track this work: https://github.com/vector-im/element-web/issues/26350

BillCarsonFr:
> @BillCarsonFr @richvdh If helpful I can progress this PR to get the simple changes that work when moving from `client.crypto` to `client.getCypto()`. Please let me know if you would like me to do so?

Thanks @hughns, this work is tracked on our side and identified as required for initial release of web-r, but not a blocker to [release on develop.io](https://github.com/vector-im/element-web/pull/26291).

We will get to it after releasing to webR. As per your changes doesn't look like any API is missing? contrary to mobile side?

hughns:
> As per your changes doesn't look like any API is missing? contrary to mobile side?

@BillCarsonFr I've done some more work on the PR. I have found an API missing (or I can't find it, at least) on the Web R side which is the ability to programatically cross-sign another device.

The place in the code that this is done is [`here` (#3761)](https://github.com/matrix-org/matrix-js-sdk/pull/3761/commits/a44ababf17f5d5a49013a97d7c16b41f9b81c51c#diff-faaea844711515261c68cfb3addf060ab9bdaa8081b2bce82d3b4027f2720729L203-L206)

Is it possible to get this exposed or for the behaviour of setDeviceVerified to be backwards compatible?

richvdh:
> The place in the code that this is done is [here (#3761)](https://github.com/matrix-org/matrix-js-sdk/pull/3761/commits/a44ababf17f5d5a49013a97d7c16b41f9b81c51c#diff-faaea844711515261c68cfb3addf060ab9bdaa8081b2bce82d3b4027f2720729L203-L206)

Unfortunately this link is 404ing :(

richvdh:
> > The place in the code that this is done is [here (#3761)](https://github.com/matrix-org/matrix-js-sdk/pull/3761/commits/a44ababf17f5d5a49013a97d7c16b41f9b81c51c#diff-faaea844711515261c68cfb3addf060ab9bdaa8081b2bce82d3b4027f2720729L203-L206)
> 
> Unfortunately this link is 404ing :(

I guess it means [here](https://github.com/matrix-org/matrix-js-sdk/blob/67019a3486d2a3191df4acbb80c90aab34389d11/src/rendezvous/MSC3906Rendezvous.ts#L204-206)

richvdh:
> Is it possible to get this exposed or for the behaviour of setDeviceVerified to be backwards compatible?

I think it's probably best we distinguish clearly between the cross-signing case and the "mark as locally verified" case, not least because we can mark other users' devices as locally verified, but we cannot cross-sign them.

I can add a new method to `CryptoApi`.

richvdh:
> > Is it possible to get this exposed or for the behaviour of setDeviceVerified to be backwards compatible?
> 
> I think it's probably best we distinguish clearly between the cross-signing case and the "mark as locally verified" case, not least because we can mark other users' devices as locally verified, but we cannot cross-sign them.
> 
> I can add a new method to `CryptoApi`.

This is done in https://github.com/matrix-org/matrix-js-sdk/pull/3930

hughns:
I have successfully manually tested this PR against Element iOS 1.11.5:

- with rust SDK enabled: `feature_rust_crypto` = `true`
- without rust SDK enabled: `feature_rust_crypto` = `false`

t3chguy (approved):
Looks sane

richvdh (changes requested):
Looks broadly good, but a few suggestions for cleanups and clarifications.

src/rendezvous/MSC3906Rendezvous.ts
```diff
@@ -203,29 +202,29 @@ export class MSC3906Rendezvous {
...
             master_key: masterPublicKey,
         });
-
-        return info;
     }
 
     /**
      * Verify the device and cross-sign it.
      * @param timeout - time in milliseconds to wait for device to come online
      * @returns the new device info if the device was verified
```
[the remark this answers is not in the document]
richvdh: Taking silence as acceptance here. Nothing should be using this method outside the js-sdk anyway.

spec/unit/rendezvous/rendezvous.spec.ts
```diff
@@ -23,75 +23,142 @@ import {
...
+                return opts.userId;
+            },
+            getDeviceId() {
+                return opts.deviceId;
             },
-        },
-    } as unknown as MatrixClient;
+            getDeviceEd25519Key() {
+                return opts.deviceKey;
+            },
```
richvdh, line 135: Just noticed: I think this thing (` getDeviceEd25519Key`) is no longer used. Ideally it would be removed.

spec/unit/rendezvous/rendezvous.spec.ts
```diff
@@ -23,75 +23,142 @@ import {
...
+
+function mockDeviceMap(
+    userId: UserID,
+    deviceId: DeviceID,
+    deviceKey?: Fingerprint,
+    otherDevices: SimpleDeviceMap = {},
+): PartialDeviceMap {
+    const deviceMap: PartialDeviceMap = new Map();
+
+    const myDevices: PartialUserDevices = new Map();
```
richvdh, line 59: Now that `mockDevice` returns a `Device`, `myDevices` should now be a regular `Map<string, Device>` rather than a `PartialUserDevices`. 

Likewise, `deviceMap` should be a `Map<string, Map<string, Device>>`.

And we can update the return type of this and `makeMockClient`.

And then `PartialUserDevices` and `PartialDeviceMap` are unused.

spec/unit/rendezvous/rendezvous.spec.ts
```diff
@@ -575,13 +643,15 @@ describe("Rendezvous", function () {
             aliceRz,
             bobTransport,
             bobEcdh,
+            devices,
```
richvdh, line 646: (Just noticed: `devices` seems to be unused and it might be nice to remove it.)

richvdh (changes requested):
Sorry, there's a mix of comments here. Some of it is fairly unimportant stuff that I failed to notice on the previous review, so maybe not worth fixing now.

But it feels like you changed the return type of `mockDevice` and then failed to follow that change through, by simplifying the functions that call it. I'd like to see that finished.

Issue #2903, opened 2022-11-22
Need clarity on the browser vs non-browser version relationship


NathanC:
I'm setting up a Vite + React +  `matrix-js-sdk` project (but not using the `matrix-react-sdk`), and I'm trying to understand how this project is laid out.

To get things to work initially, I had to install the npm [events package](https://github.com/browserify/events). I loaded Olm using wasm. Then I tried to use the sdk, and had a lot of issues with it referencing a global object that did not exist, so I do this before loading the sdk:

```
global = {
     window: window,
     fetch: fetch.bind(window),
     Olm: window.Olm,
     localStorage: window.localStorage
};
```

I then followed the pattern Element is using of importing `matrix-js-sdk/src/browser-index` at the root of my project, and importing all the sdk modules from `matrix-js-sdk/src` (instead of `matrix-js-sdk` top level). This caused > 350 typescript compilation errors, as it tried to compile the `/src` directory of the sdk, and there are a lot of errors such as indexing `Partial` objects and not checking if the value is defined. I tried downgrading my typescript version, but no dice.

I ended up removing the `brower-index` import, using `matrix-js-sdk` imports instead of `matrix-js-sdk/src` imports, and actually entirely removing my importing of the browser version that I downloaded from github.

And it..somewhat works? I can log in and interact with the server, and send encrypted messages.

tl;dr; it seems to be working now even if I don't download the browser version at all or import `browser-index`. What am I missing? I'm just looking for some clarity on the reason the setup is the way it is and how I should understand/interact with it, as I haven't been able to find much documentation and have just been trying to reverse engineer Element and the react sdk. A quick overview explanation would be very helpful

t3chguy:
`browser-index` is predominantly for the `browser-matrix` browserify bundle that is offered, but is not strictly necessary when using a bundler like Vite. It used to be different before the move to fetch and webcrypto

NathanC:
I only care about supporting more modern browsers. what's a tl;dr; on what the browserify bundle offers that the standard npm package doesn't? It it just `events` and some polyfills?

I'm new to Vite (this was actually the first project I used it for), but I may try to make some contributions down the line to make the `sdk` easier to use in Vite once I understand it better. Even just having documentation on [what sort of things are expected on the global/window object](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/crypto/crypto.ts#L19-L21), or potentially falling back to `window` so there's no need to create a `global` implicit var would be useful.

[...]

PR #3763, opened 2023-09-30
Use globalThis instead of global

orbiteleven:
Switches use of `global` to `globalThis`, which is better supported when building with modern build tools like Vite.

Refs #2903

Exiting tests pass, but to actually test this one needs to include the SDK in an app built with Vite. I've tested against our own, private repo and it works, but if we really need a sample repo for an integration test I could do that.

Signed-off-by: Damon Vestervand <damon@beyondwork.ai>

## Checklist

-   [X] Tests written for new code (and old code if feasible)
-   [X] Linter and other CI checks pass
-   [X] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))

---

### DIFF
--- a/spec/olm-loader.ts
+++ b/spec/olm-loader.ts
@@ -20,7 +20,7 @@ import { logger } from "../src/logger";
 // try to load the olm library.
 try {
     // eslint-disable-next-line @typescript-eslint/no-var-requires
-    global.Olm = require("@matrix-org/olm");
+    globalThis.Olm = require("@matrix-org/olm");
     logger.log("loaded libolm");
 } catch (e) {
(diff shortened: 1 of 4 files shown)


richvdh:
Just to confirm my understanding: 
 * `global` in node.js is deprecated in favour of `globalThis` (https://nodejs.org/docs/latest-v20.x/api/globals.html#global)
 * `global` doesn't exist in the web environment, but is polyfilled for us by webpack (https://v4.webpack.js.org/api/module-variables/#global-nodejs)

richvdh (approved):
seems reasonable to me.

Issue #3145, opened 2023-02-11
"SyntaxError: Cannot use import statement outside a module" when running /examples/node


fplord:
```~/dev/matrix-js-sdk/examples/node$ npm install

> example-app@0.0.0 preinstall
> npm install ../..


up to date, audited 36 packages in 557ms

1 package is looking for funding
  run `npm fund` for details

found 0 vulnerabilities

removed 1044 packages, and audited 36 packages in 3s

1 package is looking for funding
  run `npm fund` for details

found 0 vulnerabilities
~/dev/matrix-js-sdk/examples/node$ node app
/home/bla/dev/matrix-js-sdk/examples/node/node_modules/matrix-js-sdk/src/index.ts:17
import * as matrixcs from "./matrix";
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at internalCompileFunction (node:internal/vm:73:18)
    at wrapSafe (node:internal/modules/cjs/loader:1149:20)
    at Module._compile (node:internal/modules/cjs/loader:1190:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1280:10)
    at Module.load (node:internal/modules/cjs/loader:1089:32)
    at Module._load (node:internal/modules/cjs/loader:930:12)
    at Module.require (node:internal/modules/cjs/loader:1113:19)
    at require (node:internal/modules/cjs/helpers:103:18)
    at Object.<anonymous> (/home/bla/dev/matrix-js-sdk/examples/node/app.js:3:11)
    at Module._compile (node:internal/modules/cjs/loader:1226:14)

Node.js v18.14.0
```

SAIKUMARREDDYGANUGAPENTA:
yes

vaibhavsathe02:
Getting same problem
what to do

jessebot:
#3767 should have fixed this, but the code wasn't signed off, so the PR got closed. If someone can submit another one, this can maybe be fixed?

richvdh:
"Fixed" by https://github.com/matrix-org/matrix-js-sdk/pull/4286, I believe

jessebot:
thank you!

JuliaSoriaSmith:
I am running 34.1.0, and I am still getting the error:

(node:1803592) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)

index.js:2
import * as sdk from "matrix-js-sdk";
[...]

PR #3767, opened 2023-10-02
Fix examples/node

jameshfisher:
Node.js does not understand TypeScript, so the dependency on `file:../..` won't work. It seems `build:dev` compiles matrix-js-sdk to something in lib/ that Node.js can understand. So this changes the relative package to `file:../../lib`. But lib/ is missing a `package.json`, so Node.js doesn't recognize it as a package. To fix this, I manually add a minimal `package.json`.

Should fix https://github.com/matrix-org/matrix-js-sdk/issues/3145

---


richvdh:
Thanks, and sorry this seems to have fallen down the cracks! Please could you sign off your contribution, per https://github.com/element-hq/element-web/blob/develop/CONTRIBUTING.md#sign-off ?

richvdh:
Unfortunately we can't accept this without signoff, so I'm going to close this. Let us know if you can sign it off and we can re-open the PR.

PR #3764, opened 2023-10-02
OIDC: refresh tokens

kerryarchibald:
For https://github.com/vector-im/element-web/issues/25839
Includes https://github.com/matrix-org/matrix-js-sdk/pull/3765 https://github.com/matrix-org/matrix-js-sdk/pull/3769
With https://github.com/matrix-org/matrix-react-sdk/pull/11699

- adds `refreshToken` and `tokenRefreshFunction` to HttpApi.opts and client creation
- attempts token refresh once when encountering a token error
- retries request after token refresh

## Checklist

-   [ ] Tests written for new code (and old code if feasible)
-   [ ] Linter and other CI checks pass
-   [ ] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))

---
Here's what your changelog entry will look like:

## ✨ Features
 * OIDC: refresh tokens ([\#3764](https://github.com/matrix-org/matrix-js-sdk/pull/3764)). Contributed by @kerryarchibald.

PR #3769, opened 2023-10-02
OIDC: Token refresher class

kerryarchibald:
For https://github.com/vector-im/element-web/issues/25839
See js-sdk https://github.com/matrix-org/matrix-js-sdk/pull/3764 and https://github.com/matrix-org/matrix-react-sdk/pull/11699 for use.

Adds an abstract class responsible for refreshing OIDC native tokens.
Exposes abstract function `persistTokens` to allow consumer to persist tokens in storage.
Will be used in matrix-react-sdk to create `tokenRefreshFunction` as passed to http-api

## Checklist

---


src/oidc/tokenRefresher.ts
```diff
@@ -0,0 +1,135 @@
...
+import { IDelegatedAuthConfig } from "../client";
+import { generateScope } from "./authorize";
+import { discoverAndValidateAuthenticationConfig } from "./discovery";
+import { logger } from "../logger";
+
+/**
+ * @experimental
+ * Class responsible to refreshing OIDC access tokens
+ */
+export abstract class OidcTokenRefresher {
```
richvdh, line 29: It would be nice to explain to client implementors how this should be used. Given that it is an abstract class, the docs should probably recognise that. Something like:

```suggestion
/**
 * Base class for refreshing OIDC access tokens.
 *
 * Client implementations will typically just need to provide an implementation of {@link #persistTokens}.
 * You can then use {@link #doRefreshAccessToken} in {@link ICreateClientOpts.tokenRefreshFunction}.
 */
export abstract class OidcTokenRefresher {
```

src/http-api/interface.ts
```diff
@@ -18,6 +18,13 @@ import { MatrixError } from "./errors";
 
 export type Body = Record<string, any> | BodyInit;
 
+/**
+ * @experimental
+ */
+export type TokenRefreshFunction = (refreshToken: string) => Promise<{
+    accessToken: string;
+    refreshToken?: string;
+}>;
```
richvdh, line 27: I think it would be nicer to express this as an interface (which `OidcTokenRefresher` would implement):

```suggestion
export interface TokenRefresher {
    doRefreshAccessToken(refreshToken: string): Promise<{
        accessToken: string;
        refreshToken?: string;
    }>;
}
```

You could then just pass the `tokenRefresher` instance into `ICreateClientOpts` rather than having to mess about with `bind` [here](https://github.com/matrix-org/matrix-react-sdk/pull/11699/files#diff-d3f0f3af9b2a4141351677842e0b40f673397ca63c3292c8bf98294149ff11a4R388).

richvdh: Also, I know it's `@experimental` but I still think it's worth documenting.

richvdh: Also also: I'd give the return type its own `interface` definition so you don't need to refer to it as `ReturnType<...>`, but YMMV on that

src/oidc/tokenRefresher.ts
```diff
@@ -0,0 +1,135 @@
...
+        authConfig: IDelegatedAuthConfig,
+        /**
+         * id of this client as registered with the OP
+         */
+        clientId: string,
+        /**
+         * redirectUri as registered with OP
+         */
+        redirectUri: string,
+        deviceId: string,
```
richvdh, line 47:
```suggestion
        /**
         * Device ID returned by login request.
         */
        deviceId: string,
```

(is it still a login request with OIDC?)

src/oidc/tokenRefresher.ts
```diff
@@ -0,0 +1,135 @@
...
+import { generateScope } from "./authorize";
+import { discoverAndValidateAuthenticationConfig } from "./discovery";
+import { logger } from "../logger";
+
+/**
+ * @experimental
+ * Class responsible to refreshing OIDC access tokens
+ */
+export abstract class OidcTokenRefresher {
+    public readonly oidcClientReady!: Promise<void>;
```
richvdh, line 30: I'm not entirely clear how this is meant to be used. I can only find references to it in the tests - if that's the only place then it's probably better to make it private and put some `ts-ignore` annotations in the tests?

If it's going to be used elsewhere, could you document it?

src/oidc/tokenRefresher.ts
```diff
@@ -0,0 +1,135 @@
...
+         */
+        redirectUri: string,
+        deviceId: string,
+        /**
+         * idTokenClaims as returned from authorization grant
+         * used to validate tokens
+         */
+        private readonly idTokenClaims: IdTokenClaims,
+    ) {
+        this.oidcClientReady = this.initialiseOidcClient(authConfig, clientId, deviceId, redirectUri);
```
richvdh, line 54: it looks like there's a high chance of errors getting swallowed here. Maybe a `.catch` call to log errors?

src/oidc/tokenRefresher.ts
```diff
@@ -0,0 +1,135 @@
...
+     * @returns tokens - Promise that resolves with new access and refresh tokens
+     * @throws when token refresh fails
+     */
+    public async doRefreshAccessToken(refreshToken: string): ReturnType<TokenRefreshFunction> {
+        if (!this.inflightRefreshRequest) {
+            this.inflightRefreshRequest = this.getNewToken(refreshToken);
+        }
+        try {
+            const tokens = await this.inflightRefreshRequest;
+            await this.persistTokens(tokens);
```
richvdh, line 90: should this be in `getNewToken`, so as to avoid needlessly persisting the same result multiple times?

src/oidc/tokenRefresher.ts
```diff
@@ -0,0 +1,135 @@
...
+            return tokens;
+        } catch (error) {
+            logger.error("Failed to refresh access token", error);
+            throw error;
```
richvdh, line 94: if we're going to throw the error, shouldn't it be up to the caller to log it?

spec/unit/oidc/tokenRefresher.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+limitations under the License.
+*/
+
+import fetchMock from "fetch-mock-jest";
+import { IdTokenClaims } from "oidc-client-ts";
+
+import { IDelegatedAuthConfig, OidcTokenRefresher } from "../../../src";
+import { makeDelegatedAuthConfig } from "../../test-utils/oidc";
+
+describe("TokenRefresher()", () => {
```
richvdh, line 27:
```suggestion
describe("OidcTokenRefresher", () => {
```

spec/unit/oidc/tokenRefresher.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+    const redirectUri = "https://test.org";
+    const deviceId = "abc123";
+    const idTokenClaims = {
+        exp: Date.now() / 1000 + 100000,
+        aud: clientId,
+        iss: authConfig.issuer,
+        sub: "123",
+        iat: 123,
+    };
+    const scope = `openid urn:matrix:org.matrix.msc2967.client:api:* urn:matrix:org.matrix.msc2967.client:device:${deviceId}`;
```
richvdh, line 41: a few comments would be nice here. what do all these things do?

spec/unit/oidc/tokenRefresher.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+            redirectUri: string,
+            deviceId: string,
+            idTokenClaims: IdTokenClaims,
+        ) {
+            super(authConfig, clientId, redirectUri, deviceId, idTokenClaims);
+        }
+
+        public async persistTokens(_tokens: { accessToken: string; refreshToken?: string | undefined }): Promise<void> {
+            // NOOP
+        }
```
richvdh, line 55: I kinda wonder if `OidcTokenRefresher.persistTokens` should be a noop by default. It would make this test simpler. WDYT?

spec/unit/oidc/tokenRefresher.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+    const deviceId = "abc123";
+    const idTokenClaims = {
+        exp: Date.now() / 1000 + 100000,
+        aud: clientId,
+        iss: authConfig.issuer,
+        sub: "123",
+        iat: 123,
+    };
+    const scope = `openid urn:matrix:org.matrix.msc2967.client:api:* urn:matrix:org.matrix.msc2967.client:device:${deviceId}`;
+    class TestClass extends OidcTokenRefresher {
```
richvdh, line 42: personally I'd move this to the top-level scope, but YMMV

spec/unit/oidc/tokenRefresher.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+                accessToken: "new-access-token",
+                refreshToken: "new-refresh-token",
+            });
+        });
+
+        it("should persist the new tokens", async () => {
+            const refresher = new TestClass(authConfig, clientId, redirectUri, deviceId, idTokenClaims);
+            await refresher.oidcClientReady;
+            // spy on our stub
+            jest.spyOn(refresher, "persistTokens");
```
richvdh, line 135: it would be good to add a `jest.restoreAllMocks()` in `afterEach` to make sure we don't end up leaking spies between tests

spec/unit/oidc/tokenRefresher.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+            await expect(refresher.doRefreshAccessToken("token")).rejects.toThrow(
+                "Cannot get new token before OIDC client is initialised.",
+            );
+        });
+
+        it("should refresh the tokens", async () => {
+            const refresher = new TestClass(authConfig, clientId, redirectUri, deviceId, idTokenClaims);
+            await refresher.oidcClientReady;
+
+            const refreshToken = "refresh-token";
```
richvdh, line 118: can we inline `refreshToken` (here, and in the various other tests where it's only used once)? I found myself assuming there was a reason it was factored out and wondering where else it was being used.

spec/unit/oidc/tokenRefresher.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+            const result = await refresher.doRefreshAccessToken("first-new-refresh-token");
+
+            // called token endpoint, got new tokens
+            expect(result).toEqual({
+                accessToken: "second-new-access-token",
+                refreshToken: "second-new-refresh-token",
+            });
+        });
+    });
+});
```
richvdh, line 266: nice clear tests, thanks! :+1:

spec/unit/oidc/tokenRefresher.spec.ts
```diff
@@ -0,0 +1,266 @@
+/**
+ * @jest-environment jsdom
+ */
```
richvdh, line 3: I'm a bit surprised this is necessary?

richvdh (changes requested):

spec/unit/oidc/tokenRefresher.spec.ts
kerryarchibald: OidcClient uses `window`

spec/unit/oidc/tokenRefresher.spec.ts
kerryarchibald: it's a new instance of refresher in each test, no chance of leaks

src/http-api/interface.ts
kerryarchibald: I think there is some clarity in using `ReturnType<TokenRefreshFunction>` because it makes it clear it is implementing that function type

kerryarchibald: Disagree about the interface, it needlessly complicates the refresh api within `http-api`.

src/oidc/tokenRefresher.ts
kerryarchibald: Added in `initialiseOidcClient`

src/oidc/tokenRefresher.ts
kerryarchibald: Will add the link to `ICreateClientOpts` once `tokenRefreshFunction` is added there

src/oidc/tokenRefresher.ts
```diff
@@ -16,20 +16,28 @@ limitations under the License.
...
+ *
+ * Client implementations will likely want to override {@link persistTokens} to persist tokens after successful refresh
+ *
  */
-export abstract class OidcTokenRefresher {
+export class OidcTokenRefresher {
+    /**
+     * Used to determine the OidcClient has been initialised
+     * and is ready to start refreshing tokens
+     * throws when client initialisation fails
```
richvdh, line 36: [Long sentence](https://pr3769--js-sdk-docs-previews.netlify.app/stable/classes/oidctokenrefresher#oidcClientReady) is long and not very clear. "Throws when..." should be a separate sentence.

Also, this is odd wording. It isn't really up to an API to define how that is used. You could say "*Can* be used to determine the OidcClient has been initialised", but I think it would be simpler just to say:

```suggestion
     * Promise which will complete once the OidcClient has been initialised
     * and is ready to start refreshing tokens.
     *
     * Will reject if the client initialisation fails.
```

src/oidc/tokenRefresher.ts
```diff
@@ -60,57 +71,55 @@ export abstract class OidcTokenRefresher {
...
         } finally {
             this.inflightRefreshRequest = undefined;
         }
     }
 
     /**
-     * Persist the new tokens after successfully refreshing
+     * Persist the new tokens
+     * Called after tokens are successfully refreshed
+     * This function is intended to be overriden by the consumer when persistence is necessary
```
richvdh, line 114: Again, be aware that when this is rendered, single line breaks are swallowed, so all these sentences get run together and are quite hard to parse. Please add some paragraph breaks, or at least full stops (preferably both), between sentences.

src/http-api/interface.ts
richvdh: > I think there is some clarity in using `ReturnType<TokenRefreshFunction>` because it makes it clear it is implementing that function type

well no, it says it *returns* the same thing as a `TokenRefreshFunction`, but says nothing about the parameters.

TBH if your goal is to make it clear that a function is implementing a `TokenRefreshFunction`, then IMHO the clearest way to do that is to write it in a (doc-)comment.

Thanks for pulling out the `AccessTokens` interface.

richvdh: > Disagree about the interface, it needlessly complicates the refresh api within `http-api`.

I don't really see what you mean, but fine, I don't feel that strongly: let's leave it as a function type if you prefer.

src/http-api/interface.ts
```diff
@@ -18,6 +18,17 @@ import { MatrixError } from "./errors";
...
+ * @experimental
+ */
+export type AccessTokens = {
+    accessToken: string;
+    refreshToken?: string;
+};
+/**
+ * @experimental
+ */
+export type TokenRefreshFunction = (refreshToken: string) => Promise<AccessTokens>;
```
richvdh, line 31: please document these things

src/http-api/interface.ts
```diff
@@ -18,6 +18,17 @@ import { MatrixError } from "./errors";
...
 export interface IHttpOpts {
```
richvdh, line 32: blank line would make this more readable

```suggestion

export interface IHttpOpts {
```

spec/unit/oidc/tokenRefresher.spec.ts
```diff
@@ -38,22 +38,8 @@ describe("TokenRefresher()", () => {
         sub: "123",
         iat: 123,
     };
+    // used to mock a valid token response, as consumed by OidcClient library
```
richvdh, line 41:
```suggestion
    // a mocked valid token response, as consumed by OidcClient library
```

spec/unit/oidc/tokenRefresher.spec.ts
richvdh: could do with a few comments on some of the other things here too please!

spec/unit/oidc/tokenRefresher.spec.ts
kerryarchibald: the scope is *used* in the valid token response

richvdh:
```suggestion
    // the `scope` part of a valid token response, as consumed by OidcClient library
```

then?

kerryarchibald:
@richvdh if we have agreed guidelines about active voice, tone, etc; they should be documented. I don't think several rounds of these reviews are a good use of either of our time.

richvdh:
> if we have agreed guidelines about active voice, tone, etc; they should be documented. I don't think several rounds of these reviews are a good use of either of our time.

I'm afraid I just see it as writing clear, factual documentation - something we've agreed at a team that we need to do better at.

richvdh:
@kerryarchibald I've sent you a longer response privately, but for the public record: please can you document the `AccessTokens` and `TokenRefreshFunction` types so that we can get this landed.

kerryarchibald:
@richvdh have added documentation for `AccessTokens` and `TokenRefreshFunction`

richvdh (approved):
Thank you!

PR #3770, opened 2023-10-03
Rust crypto: ensure we persist the key backup version

BillCarsonFr:
Fixes https://github.com/vector-im/element-web/issues/26259

## Checklist

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -3864,6 +3864,10 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             throw new Error("End-to-end encryption disabled");
         }
 
+        if (!backupInfo.version) {
+            throw new Error("Backup version must be defined");
+        }
+
         let totalKeyCount = 0;
         let keys: IMegolmSessionData[] = [];
 
@@ -3880,14 +3884,12 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             }
             // Cache the key, if possible.
             // This is async.
-            if (backupInfo.version != null) {
-                this.cryptoBackend
-                    .storeSessionBackupPrivateKey(privKey, backupInfo.version)
-                    .catch((e) => {
-                        logger.warn("Error caching session backup key:", e);
-                    })
-                    .then(cacheCompleteCallback);
-            }
+            this.cryptoBackend
+                .storeSessionBackupPrivateKey(privKey, backupInfo.version)
+                .catch((e) => {
+                    logger.warn("Error caching session backup key:", e);
+                })
+                .then(cacheCompleteCallback);
 
             if (progressCallback) {
                 progressCallback({
--- a/src/client.ts
+++ b/src/client.ts
@@ -3880,12 +3880,14 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             }
             // Cache the key, if possible.
             // This is async.
-            this.cryptoBackend
-                .storeSessionBackupPrivateKey(privKey, backupInfo?.version)
-                .catch((e) => {
-                    logger.warn("Error caching session backup key:", e);
-                })
-                .then(cacheCompleteCallback);
+            if (backupInfo.version != null) {
+                this.cryptoBackend
+                    .storeSessionBackupPrivateKey(privKey, backupInfo.version)
+                    .catch((e) => {
+                        logger.warn("Error caching session backup key:", e);
+                    })
+                    .then(cacheCompleteCallback);
+            }
 
             if (progressCallback) {
                 progressCallback({
--- a/src/client.ts
+++ b/src/client.ts
@@ -3881,7 +3881,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             // Cache the key, if possible.
             // This is async.
             this.cryptoBackend
-                .storeSessionBackupPrivateKey(privKey)
+                .storeSessionBackupPrivateKey(privKey, backupInfo?.version)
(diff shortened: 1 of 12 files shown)


src/crypto-api.ts
```diff
@@ -374,9 +374,22 @@ export interface CryptoApi {
...
+    /**
+     * Store the backup decryption key.
+     *
+     * This should be called if the client has received the key from another device via secret sharing (gossiping).
+     *
+     * @param key - the backup decryption key
+     * @param version - the backup version corresponding to this decryption key
+     *
+     */
+    storeSessionBackupPrivateKey(key: Uint8Array, version?: string): Promise<void>;
```
richvdh, line 391: shouldn't this be 

```suggestion
    storeSessionBackupPrivateKey(key: Uint8Array, version: string): Promise<void>;
```

?

BillCarsonFr: Not sure, I though it was the way to do overloads

richvdh: It's not required for overloads in an `interface`.

When you have overloads in a concrete class the the signature of the *implementation* must be compatible with all overloads (the "implementation signature"). But there is no implementation here.

https://www.typescriptlang.org/docs/handbook/2/functions.html#overload-signatures-and-the-implementation-signature

src/rust-crypto/rust-crypto.ts
```diff
@@ -1005,12 +1005,20 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
      * @param key - the backup decryption key
+     * @param version - the backup version for this key. If not provided will fetch the current version from server.
      */
-    public async storeSessionBackupPrivateKey(key: Uint8Array): Promise<void> {
+    public async storeSessionBackupPrivateKey(key: Uint8Array, version?: string): Promise<void> {
         const base64Key = encodeBase64(key);
 
-        // TODO get version from backupManager
-        await this.olmMachine.saveBackupDecryptionKey(RustSdkCryptoJs.BackupDecryptionKey.fromBase64(base64Key), "");
+        const activeVersion = version ?? (await this.backupManager.requestKeyBackupVersion())?.version;
```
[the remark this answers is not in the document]
richvdh: We need to keep the old one working for old crypto. It doesn't mean we have to keep the old one working with rust crypto provided it works with the react-sdk.

src/crypto-api.ts
```diff
@@ -372,10 +372,13 @@ export interface CryptoApi {
...
      *
      * This should be called if the client has received the key from another device via secret sharing (gossiping).
+     * It is the responsability of the caller to check that the decryption key is valid to the given backup version.
      *
      * @param key - the backup decryption key
+     * @param version - the backup version corresponding to this decryption key
+     *
      */
-    storeSessionBackupPrivateKey(key: Uint8Array): Promise<void>;
+    storeSessionBackupPrivateKey(key: Uint8Array, version: string): Promise<void>;
```
richvdh, line 381: this is a breaking change. We need to continue to expose the old API, even if it only works on legacy crypto.

src/crypto/index.ts
```diff
@@ -1302,7 +1302,8 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
         // make sure we have a Uint8Array, rather than a string
         if (typeof encodedKey === "string") {
             key = new Uint8Array(olmlib.decodeBase64(fixBackupKey(encodedKey) || encodedKey));
-            await this.storeSessionBackupPrivateKey(key);
+            // legacy crypto doesn't care about the version
+            await this.storeSessionBackupPrivateKey(key, "");
```
richvdh, line 1306: yes. I don't think we need to change legacy crypto's behaviour at all here? It should continue to accept callls to `storeSessionBackupPrivateKey` with `version=undefined`.

src/rust-crypto/rust-crypto.ts
```diff
@@ -1005,12 +1005,15 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
      * Implementation of {@link CryptoApi#storeSessionBackupPrivateKey}.
      *
      * @param key - the backup decryption key
+     * @param version - the backup version for this key. If not provided will fetch the current version from server.
```
richvdh, line 1008: this seems wrong

src/crypto/CrossSigning.ts
```diff
@@ -741,7 +741,7 @@ export async function requestKeysDuringVerification(
                 logger.info("Got key backup key, decoding...");
                 const decodedKey = decodeBase64(base64Key);
                 logger.info("Decoded backup key, storing...");
-                await client.crypto!.storeSessionBackupPrivateKey(Uint8Array.from(decodedKey));
+                await client.crypto!.storeSessionBackupPrivateKey(Uint8Array.from(decodedKey), "1");
```
richvdh, line 744: why "1" ???

src/client.ts
```diff
@@ -3880,12 +3880,14 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             }
             // Cache the key, if possible.
             // This is async.
-            this.cryptoBackend
-                .storeSessionBackupPrivateKey(privKey)
-                .catch((e) => {
-                    logger.warn("Error caching session backup key:", e);
-                })
-                .then(cacheCompleteCallback);
+            if (backupInfo.version != null) {
```
richvdh, line 3883: I think we can move this check much earlier.  `version` is a required parameter on `GET /_matrix/client/v3/room_keys/keys/{roomId}` etc, so can throw an error at the top if it is not set.

PR #3771, opened 2023-10-03
Remove the dist script since we no longer have a browserify build

andybalaam:
In https://github.com/matrix-org/matrix-js-sdk/pull/3759 we removed the browserify builds.

This removes the `dist` script (and handling of the `dist/` directory) from package.json.

This should persuade our `release.sh` script to make a release without failing because there are no artifacts created in the `dist/` directory.

---


t3chguy (approved):

andybalaam:
Closing since the version I merged into staging will get back-merged into develop as part of the release process.

PR #3773, opened 2023-10-03
Grab bag of Element-R cleanups

richvdh:
Nothing major here; see individual commits

---


florianduros (approved):
Seems great!

A bit concerned that the missing await wasn't catchup by some tests before

PR #3786, opened 2023-10-06
Comments attempting to explain the addEvent method

andybalaam:
This is part of my work on https://github.com/vector-im/element-web/issues/26254

---

### DIFF
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -367,6 +367,26 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
             this.timelineSet.relations?.aggregateParentEvent(event);
             this.timelineSet.relations?.aggregateChildEvent(event, this.timelineSet);
             return;
+        } else if (this.initialEventsFetched) {
+            // If initial events have not been fetched, we are OK to throw away
+            // this event, because we are about to fetch all the events for this
+            // thread from the server.
+            //
+            // If not, this looks like a bug - we should always add the event to
+            // the thread. See https://github.com/vector-im/element-web/issues/26254
+            logger.warn(
+                `Not adding event ${event.getId()} to thread timeline!
+                isNewestReply=${isNewestReply}
+                event.localTimestamp=${event.localTimestamp}
+                !!lastReply=${!!lastReply}
+                lastReply?.getId()=${lastReply?.getId()}
+                lastReply?.localTimestamp=${lastReply?.localTimestamp}
+                toStartOfTimeline=${toStartOfTimeline}
+                Thread.hasServerSideSupport=${Thread.hasServerSideSupport}
+                event.isRelation(RelationType.Annotation)=${event.isRelation(RelationType.Annotation)}
+                event.isRelation(RelationType.Replace)=${event.isRelation(RelationType.Replace)}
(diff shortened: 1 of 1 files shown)


src/models/thread.ts
```diff
@@ -90,7 +90,39 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
 
     private reEmitter: TypedReEmitter<ThreadEmittedEvents, ThreadEventHandlerMap>;
 
+    /**
+     * The last event in this thread, if we don't yet have this in the timeline.
+     *
+     * When we run processRootEvent (which I think happens during the setting-up
+     * of the thread), we set this to the event pointed to by the server in
+     * latest_event https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3440-threading-via-relations.md#event-format
+     * that came through with the thread root.
```
richvdh, line 99: probably best to refer to the stable spec rather than an MSC:

```suggestion
     * `latest_event` [1] that came through with the thread root.
     *
     * [1]: https://spec.matrix.org/v1.8/client-server-api/#server-side-aggregation-of-mthread-relationships
```

src/models/thread.ts
```diff
@@ -90,7 +90,39 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
 
     private reEmitter: TypedReEmitter<ThreadEmittedEvents, ThreadEventHandlerMap>;
 
+    /**
+     * The last event in this thread, if we don't yet have this in the timeline.
+     *
+     * When we run processRootEvent (which I think happens during the setting-up
```
richvdh, line 96: given this is a doc-comment, you could use `{@link}` to help IDEs know that it is a link to a method, which can help if (for example) the target gets renamed:

```suggestion
     * When we run {@link #processRootEvent} (which I think happens during the setting-up
```

That said, I see it makes it harder for a human to read, so ymmv here. Likewise some of the other method names in this comment.

src/models/thread.ts
```diff
@@ -324,25 +356,34 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
...
+            // When we've asked for the event to be added to the end, and we're
+            // not in the initial state, and this event belongs at the end, add it.
             this.addEventToTimeline(event, false);
             this.fetchEditsWhereNeeded(event);
         } else if (event.isRelation(RelationType.Annotation) || event.isRelation(RelationType.Replace)) {
+            // If this event is not a direct member of the thread, but is a
+            // reference to something that is, then we have two cases:
+
             if (!this.initialEventsFetched) {
+                // Either we haven't yet fetched events from the server. In this
```
richvdh, line 381: I found this hard to read: I expected to find an "or" somewhere in this paragraph.

```suggestion
                // Case 1: we haven't yet fetched events from the server. In this
```

richvdh: likewise "Case 2" below

richvdh (approved):
Yay comments! Some nits but generally this is great.

src/models/thread.ts
andybalaam: Thanks, fixed in 6a569bff4

PR #3787, opened 2023-10-06
Add events to the thread even if they appear to be out of order

andybalaam:
Fixes https://github.com/vector-im/element-web/issues/26254
(hopefully)

Reduce the number of cases where we drop events that we are trying to add to a thread.

I have lots of refactorings I would like to do to addEvent but this change isolates the functional change that may address the referenced bug.

I will be holding off on merging this until after the release candidate on 17th October 2023, so that we can try it out for a couple of weeks and evaluate whether it causes more stuck notifications.

---

### DIFF
--- a/spec/unit/models/thread.spec.ts
+++ b/spec/unit/models/thread.spec.ts
@@ -675,6 +675,69 @@ describe("Thread", () => {
             });
         });
     });
+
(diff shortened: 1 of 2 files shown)


richvdh (approved):
seems plausible to me

PR #3789, opened 2023-10-08
Define a new `Logger` interface

richvdh:
This PR replaces the existing `PrefixedLogger` interface with a more limited interface defining the parts we *actually* use. It also avoids the word "prefixed" since some implementions may choose other ways to indicate a child logger.

Ultimately, the goal here is to allow the application to provide its own `Logger` instance. This will be particularly useful in situations where there is more than one `MatrixClient` active (eg, clients which support multiple logins, or the cypress tests), but there may also be clients who want to direct log output somewhere other than the console. There is more work to do to allow that to happen, however.

Suggest reviewing this commit-by-commit.


andybalaam (approved):
Looks good, thanks!

PR #3791, opened 2023-10-09
Element-R: Don't mark QR code verification as done until it's done

richvdh:
The rust crypto sdk doesn't actually finish QR code verification until the `m.key.verification.done` is received, so make sure we don't tell the application it is done before that happens.

Fixes https://github.com/vector-im/element-web/issues/26293

---

## 🐛 Bug Fixes
 * Element-R: Don't mark QR code verification as done until it's done ([\#3791](https://github.com/matrix-org/matrix-js-sdk/pull/3791)). Fixes vector-im/element-web#26293.


florianduros (approved):
Perfect ! Thanks also for all the docs, really helpful to understand !

PR #3792, opened 2023-10-09
Allow applications to specify their own logger instance

### LINKED ISSUE #1899: Ability to have a a logger per MatrixClient

richvdh:
Allow clients to specify a logger instance when instantiating a matrix client via the `ClientOpts`.

This is is particularly useful for clients that want to support multiple matrix accounts, but I also want to use it in the cypress tests.

It builds on the work done by @clokep in #3484, and replaces that PR.

It's not yet complete - there are still *lots* of places that use the global `logger` - but it puts the framework in place and catches quite a lot of the main culprits.

Fixes https://github.com/matrix-org/matrix-js-sdk/issues/1899.

---

## ✨ Features
 * Allow applications to specify their own logger instance ([\#3792](https://github.com/matrix-org/matrix-js-sdk/pull/3792)). Fixes #1899.

### DIFF
--- a/spec/unit/http-api/fetch.spec.ts
+++ b/spec/unit/http-api/fetch.spec.ts
@@ -14,14 +14,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { mocked } from "jest-mock";
+import { Mocked } from "jest-mock";
 
 import { FetchHttpApi } from "../../../src/http-api/fetch";
 import { TypedEventEmitter } from "../../../src/models/typed-event-emitter";
 import { ClientPrefix, HttpApiEvent, HttpApiEventHandlerMap, IdentityPrefix, IHttpOpts, Method } from "../../../src";
 import { emitPromise } from "../../test-utils/test-utils";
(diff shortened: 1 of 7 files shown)


florianduros (approved):
Looks good, thanks!

richvdh:
@florianduros I put it back in draft while I fixed the tests, which I've now done - are you still happy?

I don't think it's going to be practical to meet the coverage gate on this PR since it hits a range of existing code with poor coverage.

florianduros (approved):
:shipit:

PR #3793, opened 2023-10-09
Don't use event.sender in CallMembership

dbkr:
I fell into another js-sdk trap: this is "only guaranteed to be set for events that appear in a timeline" and not state events. It does not say why. We only ever used it to get the sender user ID anyway, so just use getSender().

## Checklist

---

## 🐛 Bug Fixes
 * Don't use event.sender in CallMembership ([\#3793](https://github.com/matrix-org/matrix-js-sdk/pull/3793)).


toger5 (approved):
Thanks for the PR description. With the description its clear why we use getSender

PR #3796, opened 2023-10-12
Fix reemitter not being correctly wired on user objects created in storage classes

MidhunSureshR:
Storage classes - `memory.ts` and `indexeddb.ts` both create user objects without setting it up to emit events on matrix client.
This is sometimes preventing the memberlist from updating on incoming presence events because [this listener](https://github.com/matrix-org/matrix-react-sdk/blob/develop/src/components/views/rooms/MemberList.tsx#L152) never runs.

---

## 🐛 Bug Fixes
 * Fix reemitter not being correctly wired on user objects created in storage classes ([\#3796](https://github.com/matrix-org/matrix-js-sdk/pull/3796)). Contributed by @MidhunSureshR.

### DIFF
--- a/src/models/user.ts
+++ b/src/models/user.ts
@@ -153,6 +153,7 @@ export class User extends TypedEventEmitter<UserEvent, UserEventHandlerMap> {
     /**
      * Construct a new User. A User must have an ID and can optionally have extra information associated with it.
      * @param userId - Required. The ID of this user.
+     * @deprecated use `User.createUser`
      */
     public constructor(public readonly userId: string) {
         super();
--- a/src/models/user.ts
+++ b/src/models/user.ts
@@ -168,7 +168,7 @@ export class User extends TypedEventEmitter<UserEvent, UserEventHandlerMap> {
      * @param client - An instance of MatrixClient object
      * @returns User object with reEmitter setup on client
      */
-    public static withReEmitterOnClient(userId: string, client: MatrixClient): User {
+    public static createUser(userId: string, client: MatrixClient): User {
         const user = new User(userId);
         client.reEmitter.reEmit(user, [
             UserEvent.AvatarUrl,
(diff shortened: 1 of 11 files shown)


src/store/index.ts
```diff
@@ -62,6 +64,12 @@ export interface IStore {
      */
     storeRoom(room: Room): void;
 
+    /**
+     * Set the user creator which is used for creating User objects
+     * @param creator - A callback that accepts an user-id and returns an User object
+     */
+    setUserCreator(creator: UserCreator): void;
```
MidhunSureshR, line 71: Temporary solution until we fix https://github.com/vector-im/element-web/issues/26383

MidhunSureshR (commented):
.

src/models/user.ts
```diff
@@ -160,6 +161,25 @@ export class User extends TypedEventEmitter<UserEvent, UserEventHandlerMap> {
...
     }
 
+    /**
+     * Construct a new User whose events will also emit on MatrixClient.
+     * A User must have an ID and can optionally have extra information associated with it.
+     * @param userId - Required. The ID of this user.
+     * @param client - An instance of MatrixClient object
+     * @returns User object with reEmitter setup on client
+     */
+    public static withReEmitterOnClient(userId: string, client: MatrixClient): User {
```
germain-gg, line 171: I think we should change the naming. It should not contain `withReEmitter`, but rather describe what it's doing.
The following seems more appropriate to me, and will survive a future refactor where we decide to move away from event emitters

```suggestion
    public static createUser(userId: string, client: MatrixClient): User {
```

germain-gg, line 171: I would also suggest making the `User` constructor deprecated and commenting there that folks should use `User.createUser` instead to have all the appropriate events set up

Screenshot 2023-10-23 at 07 10 49

```
     * @deprecated use `User.withReEmitterOnClient`
```

germain-gg (changes requested):
Thank you for opening this up!
A few things I believe should be updated

germain-gg (approved):
Thank you, LGTM

PR #3799, opened 2023-10-13
Fix sending call member events on leave

dbkr:
https://github.com/matrix-org/matrix-js-sdk/pull/3756 changed the membership update function to await on the next call, but this meant it never returned and therefore never cleared `updateCallMembershipRunning`. We therefore didn't send the updated call member event when leaving, instead sending it whenever the next poll interval arrived.

This changes it to only await if we are retrying, not if we're just scheduling the next poll.

@toger5 review appreciated on this one - does this line up with what you expected your fix to do?

Fixes https://github.com/vector-im/element-call/issues/1763

## Checklist

---

## 🐛 Bug Fixes
 * Fix sending call member events on leave ([\#3799](https://github.com/matrix-org/matrix-js-sdk/pull/3799)). Fixes vector-im/element-call#1763.

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -427,7 +427,6 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
             memberships: this.makeNewMemberships(memberships, myCallMemberEvent, myPrevMembership),
         };
 
-        let resendDelay = 0;
         try {
             await this.client.sendStateEvent(
                 this.room.roomId,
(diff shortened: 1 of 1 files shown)


toger5 (approved):
The main issue is, that we checked for resendDelay before `this.triggerCallMembershipEventUpdate();`?: `   if (resendDelay) {`

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -438,13 +437,12 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
             logger.info(`Sent updated call member event.`);
 
             // check periodically to see if we need to refresh our member event
-            if (this.isJoined()) resendDelay = MEMBER_EVENT_CHECK_PERIOD;
+            if (this.isJoined()) {
+                this.memberEventTimeout = setTimeout(this.triggerCallMembershipEventUpdate, MEMBER_EVENT_CHECK_PERIOD);
+            }
         } catch (e) {
-            resendDelay = CALL_MEMBER_EVENT_RETRY_DELAY_MIN + Math.random() * 2000;
+            const resendDelay = CALL_MEMBER_EVENT_RETRY_DELAY_MIN + Math.random() * 2000;
```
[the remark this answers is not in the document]
dbkr: No, I think scopes are all the same in js (at least this one doesn't , I tested).

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -438,13 +437,12 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
             logger.warn(`Failed to send call member event: retrying in ${resendDelay}`);
-        }
-
-        if (resendDelay) {
             await new Promise((resolve) => setTimeout(resolve, resendDelay));
             await this.triggerCallMembershipEventUpdate();
         }
```
toger5, line 448: We can do the same syntax as above here now?
```suggestion
            setTimeout(this.triggerCallMembershipEventUpdate, resendDelay);
        }
```

dbkr: But I thought the point of this was that we wanted the function to block until the retries had been sent?

PR #3801, opened 2023-10-17
Deprecate `MatrixEvent.toJSON`

richvdh:
Per https://github.com/vector-im/element-web/issues/26380, this method is too easy to use accidentally, and per the comments, it doesn't even return a meaningful JSON-serialisation of the object.

---

## 🦖 Deprecations
 * Deprecate `MatrixEvent.toJSON` ([\#3801](https://github.com/matrix-org/matrix-js-sdk/pull/3801)).

### DIFF
--- a/src/models/event.ts
+++ b/src/models/event.ts
@@ -419,7 +419,7 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
      * If the event is encrypted, we attempt to mock up an event as it would have looked had the sender not encrypted it.
      * If the event is not encrypted, a copy of it is simply returned as-is.
      *
-     * @returns A copy of the event, in wire format, as it would have been had it not been encrypted.
+     * @returns A shallow copy of the event, in wire format, as it would have been had it not been encrypted.
      */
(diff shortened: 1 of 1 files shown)


src/models/event.ts
```diff
@@ -414,9 +414,12 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
...
     /**
-     * Gets the event as though it would appear unencrypted. If the event is already not
-     * encrypted, it is simply returned as-is.
-     * @returns The event in wire format.
+     * Gets the event as it would appear if it had been sent unencrypted.
+     *
+     * If the event is encrypted, we attempt to mock up an event as it would have looked had the sender not encrypted it.
+     * If the event is not encrypted, a copy of it is simply returned as-is.
+     *
+     * @returns A copy of the event, in wire format, as it would have been had it not been encrypted.
```
richvdh, line 422:
```suggestion
     * @returns A shallow copy of the event, in wire format, as it would have been had it not been encrypted.
```

andybalaam (approved):
Thank you!

PR #3807, opened 2023-10-17
Element-R: Wire up room rotation

florianduros:
## Checklist

Closes https://github.com/vector-im/element-web/issues/26318

---

## ✨ Features
 * Element-R: Wire up room rotation ([\#3807](https://github.com/matrix-org/matrix-js-sdk/pull/3807)). Fixes vector-im/element-web#26318. Contributed by @florianduros.


src/rust-crypto/RoomEncryptor.ts
```diff
@@ -103,7 +109,21 @@ export class RoomEncryptor {
         this.prefixedLogger.debug("Sessions for users are ready; now sharing room key");
 
         const rustEncryptionSettings = new EncryptionSettings();
-        /* FIXME historyVisibility, rotation, etc */
+        /* FIXME historyVisibility, etc */
+
+        // We only support megolm
+        rustEncryptionSettings.algorithm = EncryptionAlgorithm.MegolmV1AesSha2;
```
florianduros, line 115: Two solutions here:

- Set directly the Megolm algorithm since we currently only supporting megolm and we are checking at the beginning at the function that we are using Megolm.
- Use the value from  `this.encryptionSettings`, it means that we need to convert the string in the field `this.encryptionSettings.algoritm` to the rust `EncryptionAlgorithm` enum

The second solution appears to me useless since we are only using Megolm

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -103,7 +109,21 @@ export class RoomEncryptor {
...
+
+        // We need to convert the rotation period from milliseconds to microseconds
+        // See https://spec.matrix.org/v1.8/client-server-api/#mroomencryption and
+        // https://matrix-org.github.io/matrix-rust-sdk-crypto-wasm/classes/EncryptionSettings.html#rotationPeriod
+        if (typeof this.encryptionSettings.rotation_period_ms === "number") {
+            rustEncryptionSettings.rotationPeriod = BigInt(this.encryptionSettings.rotation_period_ms * 1000);
```
richvdh, line 121: it's a bit silly that the Rust `EncryptionSettings` wants a bigint rather than a `number`, but hohum

richvdh (approved):
lgtm

PR #3810, opened 2023-10-19
Element-R: log when we send to-device messages

richvdh:
The logs will look something like this:

image

The first line comes from the rust-sdk; I think it's helpful to also log just before we actually send the batch.

---

## ✨ Features
 * Element-R: log when we send to-device messages ([\#3810](https://github.com/matrix-org/matrix-js-sdk/pull/3810)).


andybalaam (approved):
Looks good, thanks!

PR #3798, opened 2023-10-13
WIP move redacted message to main

andybalaam:
Not at all ready

PR #3814, opened 2023-10-20
Don't remove thread info from a thread root when it is redacted

andybalaam:
My [recent change](https://github.com/matrix-org/matrix-js-sdk/pull/3798) moved threaded messages into the main timeline when they were redacted.

I realised we should not do this to thread roots, as they need to stay "in" both the main timeline and the thread, so we can render a thread UI from them, even though they are redacted.

Part of https://github.com/vector-im/element-web/issues/24392

---

## 🐛 Bug Fixes
 * Don't remove thread info from a thread root when it is redacted ([\#3814](https://github.com/matrix-org/matrix-js-sdk/pull/3814)). Contributed by @andybalaam.

### DIFF
--- a/spec/unit/models/event.spec.ts
+++ b/spec/unit/models/event.spec.ts
@@ -111,6 +111,29 @@ describe("MatrixEvent", () => {
             expect(mainTimelineLiveEventIds(room)).toEqual([ev.getId()]);
         });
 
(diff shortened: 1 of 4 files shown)


germain-gg (approved):
Looks good to me!

PR #3817, opened 2023-10-20
Ignore receipts pointing at missing or invalid events

andybalaam:
Helps with some of the flaking redaction tests, and some of the just-plain-failing editing tests.

We were seeing bugs where the latest receipt was not pointing at an event in this thread (or an event we had at all), so the room was considered unread, but we were unable to send a correct receipt because we already had one. This change means we effectively ignore the existing receipt because we can't use it, allowing us to send a correct one.

Part of https://github.com/vector-im/element-web/issues/24392

When I enable lots of disabled tests (in https://github.com/matrix-org/matrix-react-sdk/pull/11781 ) with this change, they pass consistently on my machine.

---

## 🐛 Bug Fixes
 * Ignore receipts pointing at missing or invalid events ([\#3817](https://github.com/matrix-org/matrix-js-sdk/pull/3817)). Contributed by @andybalaam.

### DIFF
--- a/spec/unit/room.spec.ts
+++ b/spec/unit/room.spec.ts
@@ -3255,7 +3255,7 @@ describe("Room", function () {
                     for (let i = 1; i <= 2; i++) {
                         room.getUnfilteredTimelineSet = () =>
                             ({
-                                compareEventOrdering: (event1: string) => {
+                                compareEventOrdering: (event1: string, _event2: string) => {
                                     return event1 === `eventId${i}` ? 1 : -1;
                                 },
                                 findEventById: jest.fn().mockReturnValue({} as MatrixEvent),
--- a/spec/unit/room.spec.ts
+++ b/spec/unit/room.spec.ts
@@ -1746,6 +1746,7 @@ describe("Room", function () {
             it("should acknowledge if an event has been read", function () {
                 const ts = 13787898424;
                 room.addReceipt(mkReceipt(roomId, [mkRecord(eventToAck.getId()!, "m.read", userB, ts)]));
+                room.findEventById = jest.fn().mockReturnValue({} as MatrixEvent);
                 expect(room.hasUserReadEvent(userB, eventToAck.getId()!)).toEqual(true);
             });
             it("return false for an unknown event", function () {
@@ -3151,7 +3152,7 @@ describe("Room", function () {
             room.getReadReceiptForUserId = (userId, ignore, receiptType): WrappedReceipt | null => {
                 return receiptType === ReceiptType.ReadPrivate ? ({ eventId: "eventId" } as WrappedReceipt) : null;
             };
-
+            room.findEventById = jest.fn().mockReturnValue({} as MatrixEvent);
             expect(room.getEventReadUpTo(userA)).toEqual("eventId");
         });
 
@@ -3170,10 +3171,11 @@ describe("Room", function () {
                 for (let i = 1; i <= 2; i++) {
                     room.getUnfilteredTimelineSet = () =>
                         ({
-                            compareEventOrdering: (event1, event2) => {
+                            compareEventOrdering: (event1: string) => {
                                 return event1 === `eventId${i}` ? 1 : -1;
                             },
-                        } as EventTimelineSet);
+                            findEventById: jest.fn().mockReturnValue({} as MatrixEvent),
+                        } as unknown as EventTimelineSet);
 
                     expect(room.getEventReadUpTo(userA)).toEqual(`eventId${i}`);
                 }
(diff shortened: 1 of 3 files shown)


src/models/read-receipt.ts
```diff
@@ -95,14 +96,74 @@ export abstract class ReadReceipt<
...
+        // The thread ID doesn't match up
+        logger.warn(
+            `Ignoring receipt because its thread_id (${receipt.data.thread_id}) disagrees ` +
+                `with the thread root (${event.threadRootId}) of the referenced event ` +
+                `(event ID = ${receipt.eventId})`,
+        );
+        return false;
+    }
+
+    private getLatestReceipt(userId: string, ignoreSynthesized = false): WrappedReceipt | null {
```
richvdh, line 166: default for `ignoreSynthezised` seems redundant here

src/models/read-receipt.ts
```diff
@@ -95,14 +96,74 @@ export abstract class ReadReceipt<
...
+     * Returns true if the event pointed at by this receipt exists, and its
+     * threadRootId is consistent with the thread information in the receipt.
+     */
+    private receiptPointsAtConsistentEvent(receipt: WrappedReceipt): boolean {
+        // If the receipt points at a non-existent event, it's not consistent
+        const event = this.findEventById(receipt.eventId);
+        if (!event) {
+            logger.warn(`Ignoring receipt for missing event with id ${receipt.eventId}`);
+            return false;
+        }
```
richvdh, line 130: Doesn't this automatically do the filtering-by-thread-id stuff? `Thread.findEventById` will return null if the event is in another thread?

src/models/read-receipt.ts
```diff
@@ -229,6 +290,8 @@ export abstract class ReadReceipt<
 
     public abstract setUnread(type: NotificationCountType, count: number): void;
 
+    public abstract findEventById(eventId: string): MatrixEvent | undefined;
```
richvdh, line 293: might be best to give this a comment: what exactly is its expected behaviour?

richvdh (commented):
Not entirely sure I'm grokking this

src/models/read-receipt.ts
```diff
@@ -94,15 +95,108 @@ export abstract class ReadReceipt<
...
+        if (!event) {
+            // If the receipt points at a non-existent event, we have 2
+            // possibilities:
+            //
+            // 1. We don't have the event because it's not loaded yet - probably
+            //    it's old and we're best off ignoring the receipt - we can just
+            //    send a new one when we read a new event.
+            //
+            // 2. We have a bug e.g. we misclassified this event into the wrong
+            //    thread.
```
richvdh, line 139: doesn't this also happen when:
 * `this` is a thread, and the event is moved from thread to main timeline
 * the receipt lied about which thread the event was in. (As I understand it, `getLatestReceipt` will get the latest receipt *claiming* to be in the current thread, but there is no check that the event it points to was actually in this thread.)  I suppose that would also reflect a bug, though it could be in a different client; or even (if it's a different user) malicious.

src/models/read-receipt.ts
```diff
@@ -94,15 +95,108 @@ export abstract class ReadReceipt<
...
+
+        // The receipt thread ID disagrees with the event thread ID. There are 2
+        // possibilities:
+        //
+        // 1. The event moved to a different thread after the receipt was
+        //    created. This can happen if the event was redacted because that
+        //    moves it to the main timeline.
+        //
+        // 2. There is a bug somewhere - either we put the event into the wrong
+        //    thread, or someone sent an incorrect receipt.
```
richvdh, line 187: As above: I feel these cases would be caught by `this.findEventById` most of the time. (Except possibly in the case where `this` is a room, so `findEventById` will return events in threads too.)

Again: no problem with a defense-in-depth check here, but I'd prefer it if the comment could make it clear that it's a defense-in-depth.

spec/unit/room.spec.ts
```diff
@@ -3147,106 +3148,195 @@ describe("Room", function () {
...
+                            return { eventId: "eventId2" } as WrappedReceipt;
                         }
                         return null;
                     };
 
-                    expect(room.getEventReadUpTo(userA)).toEqual(`eventId1`);
+                    for (let i = 1; i <= 2; i++) {
+                        room.getUnfilteredTimelineSet = () =>
+                            ({
+                                compareEventOrdering: (event1: string) => {
```
richvdh, line 3258: Nit: I'm unconvinced that dropping `event2` here is an improvement in clarity. It makes me want to dig out the definition of `compareEventOrdering` to see how it can possibly compare the ordering of a single event.

Better to just give it a name which reflects the fact we don't need it here:

```suggestion
                                compareEventOrdering: (event1: string, _event2: string) => {
```

PR #3825, opened 2023-10-23
Element-R: Add current version of the rust-sdk and vodozemac

florianduros:
## Checklist

Related to https://github.com/vector-im/element-web/issues/26386

---

## ✨ Features
 * Element-R: Add current version of the rust-sdk and vodozemac ([\#3825](https://github.com/matrix-org/matrix-js-sdk/pull/3825)). Contributed by @florianduros.


src/crypto/index.ts
```diff
@@ -610,6 +610,15 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
         this.backupManager.checkAndStart();
     }
 
+    /**
+     * Implementation of {@link CryptoApi#getVersion}.
+     */
+    public getVersion(): null {
+        // `Crypto.getOlmVersion()` is used for the old crypto
```
florianduros, line 617: The old crypto version is set in this way https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/client.ts#L2281 and we are using it here https://github.com/matrix-org/matrix-react-sdk/blob/develop/src/components/views/settings/tabs/user/HelpUserSettingsTab.tsx#L75

I keep for the old crypto, the `olmVersion` field and I'm adding a new way to get the version for the new crypto (used here https://github.com/matrix-org/matrix-react-sdk/pull/11785/files#diff-c612bdd14b2fc22382b3675abd80d0b85f887453389d2a4c5b0fdcf2df887bd2R88).

florianduros (commented):
I'll ask for an exception about the code coverage. I can add a dummy test to see if the old crypto returns `null` however imo, there is no real gain.

src/crypto/index.ts
richvdh: could you make this return something like `Olm ${olmVersionTuple[0]}.${olmVersionTuple[1]}.${olmVersionTuple[2]}`, so that both `Crypto` implementations have a `.getVersion` that returns a string, and just use that in the rust SDK?

florianduros: Yes good idea!

src/crypto-api.ts
```diff
@@ -39,6 +39,15 @@ export interface CryptoApi {
      */
     globalBlacklistUnverifiedDevices: boolean;
 
+    /**
+     * Return the current version of the crypto module.
+     * Rust crypto: the format is `Rust SDK ${versions.matrix_sdk_crypto}, Vodozemac ${versions.vodozemac}`
+     * Old Crypto: the format is `Olm x.x.x`
+     * XXXX: remove old crypto when we remove support for it
```
richvdh, line 46: Nit: this isn't terribly legible as it stands (see https://pr3825--js-sdk-docs-previews.netlify.app/stable/interfaces/crypto.cryptoapi#getVersion): it could do with some paragraph breaks.

Anyway, I feel like we should just give an example, rather than documenting this in a way that makes it sound like someone can rely on that exact format.

```suggestion
     * Return the current version of the crypto module.
     *
     * For example: `Rust SDK ${versions.matrix_sdk_crypto}, Vodozemac ${versions.vodozemac}`.
     *
```

spec/unit/crypto.spec.ts
```diff
@@ -115,6 +115,16 @@ describe("Crypto", function () {
         expect(Crypto.getOlmVersion()[0]).toEqual(3);
     });
 
+    it("should return the current version of the olm library", async () => {
```
richvdh, line 118:
```suggestion
    it("getVersion() should return the current version of the olm library", async () => {
```

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -143,6 +143,13 @@ describe("RustCrypto", () => {
             10000,
         );
 
+        it("should return the current version of the rust sdk and vodozemac", async () => {
```
richvdh, line 146:
```suggestion
        it("getVersion() should return the current version of the rust sdk and vodozemac", async () => {
```

richvdh (approved):
lgtm otherwise

PR #3826, opened 2023-10-23
Element-R: reduce log spam when checking server key backup

richvdh:
Fixes a lot of spam in the logs about "uncaught in promise: No room_keys found".

---


src/rust-crypto/backup.ts
```diff
@@ -460,7 +460,7 @@ export class RustBackupDecryptor implements BackupDecryptor {
         for (const [sessionId, sessionData] of Object.entries(ciphertexts)) {
             try {
                 const decrypted = JSON.parse(
-                    await this.decryptionKey.decryptV1(
+                    this.decryptionKey.decryptV1(
```
richvdh, line 463: this is slightly separate, but a cleanup while I was in the area.

`decryptionKey.decryptV1` doesn't return a promise.

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -149,64 +151,131 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("megolm-keys backup (%s)", (backe
...
+
+            const calls = fetchMock.calls("getKey");
+            expect(calls.length).toEqual(1);
+            expect(calls[0][0]).toEqual(EXPECTED_URL);
+
+            await flushBackupRequest();
+
+            // we should not have logged an error.
+            // eslint-disable-next-line no-console
+            expect(console.error).not.toHaveBeenCalled();
```
BillCarsonFr, line 250: That's a nice trick. Is it possible to check for a specific log line?

src/rust-crypto/backup.ts
BillCarsonFr: Yeah I have seen it now that I use WebStorm. I am wondering if you shouldn't update the bindings to make it async?

src/rust-crypto/rust-crypto.ts
```diff
@@ -161,43 +161,72 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
-            now - this.sessionLastCheckAttemptedTime[targetSessionId!] > KEY_BACKUP_CHECK_RATE_LIMIT
-        ) {
+        const lastCheck = this.sessionLastCheckAttemptedTime[targetSessionId];
+        if (!lastCheck || now - lastCheck > KEY_BACKUP_CHECK_RATE_LIMIT) {
             this.sessionLastCheckAttemptedTime[targetSessionId!] = now;
-
-            const path = encodeUri("/room_keys/keys/$roomId/$sessionId", {
-                $roomId: targetRoomId,
-                $sessionId: targetSessionId,
+            this.queryKeyBackup(targetRoomId, targetSessionId).catch((e) => {
```
BillCarsonFr, line 177: I think we might want to check if we have the private key before marking the check as attempted (sessionLastCheckAttemptedTime = now). Because if we attempt to get the key and you don't have the private key yet (verification not yet done) it will not retry until KEY_BACKUP_CHECK_RATE_LIMIT.  So when you verify after login, you'll have to wait?
KEY_BACKUP_CHECK_RATE_LIMIT isn't that high currently, so maybe not too important though.

BillCarsonFr (approved):
Thx for the cleaning, it was mostly copy pasted from legacy. Just a few comments

spec/integ/crypto/megolm-backup.spec.ts
richvdh: I guess you could do `expect(console.error).toHaveBeenCalledWith(expect.stringContaining("text"))`, or something. Or substitute your own `console.error` implementation.

src/rust-crypto/rust-crypto.ts
richvdh: > I think we might want to check if we have the private key before marking the check as attempted (sessionLastCheckAttemptedTime = now). 

The problem with that is that it is racy: if two calls to `startQueryKeyBackupRateLimited` happen in quick succession (eg, because there are two messages in a sync response), then both will check `sessionLastCheckAttemptedTime` and find it ok, and go on to make a backup request. We need to make sure there is no `await` between the check and the update on `sessionLastCheckAttemptedTime`.

src/rust-crypto/backup.ts
richvdh: Added an `await immediate()` in https://github.com/matrix-org/matrix-js-sdk/pull/3826/commits/131c52433043280556c1c0370be8b814da2cb4ca to make sure we don't block the event loop.

germain-gg (approved):

richvdh:
cypress failures: https://github.com/vector-im/element-web/issues/26437, https://github.com/vector-im/element-web/issues/26447, https://github.com/vector-im/element-web/issues/26452

PR #3829, opened 2023-10-24
Better before/after member comparison

dbkr:
Only trigger for when members actually join, and just generally make it a bit more understandable.

Based on https://github.com/matrix-org/matrix-js-sdk/pull/3667

## Checklist

---


dbkr:
Actually I'm just going include this change in the PR for key rotation on member leaving.

PR #3830, opened 2023-10-24
Fix members being loaded from server on initial sync (defeating lazy loading)

BillCarsonFr:
Fixes https://github.com/vector-im/element-web/issues/26418
Fixes https://github.com/vector-im/element-web/issues/26439

## Checklist

---

## 🐛 Bug Fixes
 * Fix members being loaded from server on initial sync (defeating lazy loading) ([\#3830](https://github.com/matrix-org/matrix-js-sdk/pull/3830)). Fixes vector-im/element-web#26418 and vector-im/element-web#26439.


src/rust-crypto/rust-crypto.ts
```diff
@@ -1232,10 +1232,6 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
                 config,
             );
         }
-
-        // start tracking devices for any users already known to be in this room.
-        const members = await room.getEncryptionTargetMembers();
```
BillCarsonFr, line 1237: This is what was triggering the `loadMembersIfNeeded`. Was called during initial sync when the m.room.encryption was processed.
I moved that code in the RoomEncryptor, it's now only getting the known members from existing state and not loading if neeed

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -91,7 +103,33 @@ export class RoomEncryptor {
             );
         }
 
+        // Manually call `loadMembersIfNeeded` here, because we want to know if it's the first
+        // time the room is loaded (due to lazy loading), so we can update the tracked users.
+        const fromServer = await this.room.loadMembersIfNeeded();
         const members = await this.room.getEncryptionTargetMembers();
+
+        if (fromServer) {
```
BillCarsonFr, line 111: Legacy was using another solution (I think more complex), by registering a listener on `RoomStateEvent.Update`. I looked at the code and wasn't conviced that I should do the same.

Instead I call `loadMembersIfNeeded()` to know if it was loaded from server for the first time

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -91,7 +103,33 @@ export class RoomEncryptor {
...
+            // It's the first time the room is loaded, so we need to update the tracked users
+            await this.olmMachine
+                .updateTrackedUsers(members.map((u) => new RustSdkCryptoJs.UserId(u.userId)))
+                .then(() => {
+                    this.prefixedLogger.debug(`Updated tracked users for room ${this.room.roomId}`);
+                });
+        }
+
+        // Query keys in case we don't have them for newly tracked members.
```
BillCarsonFr, line 120: It's not really related to the current problem, but I had to fix it to make the lazy loading test pass on both crypto backend. It's anyhow a real problem, and we can see several [occurence of it in rageshakes](https://rageshake-search.tools.element.io/search_logs?log=some+devices+might+be+missing). The (full) rust sdk as a [similar mecanism](https://github.com/matrix-org/matrix-rust-sdk/blob/93edf7a06462c8ee9afcc587a5725aa8956ad5ab/crates/matrix-sdk/src/room/mod.rs#L1504) to ensure the download of user keys for non tracked/dirty users.

This would need to be fixed in mobile too

spec/integ/crypto/crypto.spec.ts
```diff
@@ -670,18 +670,21 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
     });
 
     it("prepareToEncrypt", async () => {
-        expectAliceKeyQuery({ device_keys: { "@alice:localhost": {} }, failures: {} });
+        const homeserverUrl = "https://alice-server.com";
```
BillCarsonFr, line 673: I started to modernize a bit the tests by using E2EKeyResponder, because we are requesting keys a bit differently and the existing test were using fragile tactis `Alice will check Bob's device list (twice, because reasons)`.
I haven't done it all because it's a bit strange. In these test the olmTestDevice is sometime an alice device and sometimes a bob device.
If ok, I'd like to clean all this file in a following PR.

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -91,7 +103,33 @@ export class RoomEncryptor {
...
+        const fromServer = await this.room.loadMembersIfNeeded();
         const members = await this.room.getEncryptionTargetMembers();
+
+        if (fromServer) {
+            // It's the first time the room is loaded, so we need to update the tracked users
+            await this.olmMachine
+                .updateTrackedUsers(members.map((u) => new RustSdkCryptoJs.UserId(u.userId)))
+                .then(() => {
+                    this.prefixedLogger.debug(`Updated tracked users for room ${this.room.roomId}`);
+                });
```
richvdh, line 117: No need for a `.then` here:

```suggestion
            await this.olmMachine.updateTrackedUsers(members.map((u) => new RustSdkCryptoJs.UserId(u.userId)));
            this.prefixedLogger.debug(`Updated tracked users for room ${this.room.roomId}`);
```

src/rust-crypto/RoomEncryptor.ts
richvdh: it seems like, once `loadMembersIfNeeded` has returned true once, it will *always* return true, because it always returns the same promise (`membersPromise`)?

It might be better just to have a boolean in `RoomEncryptor` which is cleared once we call `updateTrackedUsers`.

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -91,7 +103,33 @@ export class RoomEncryptor {
...
+        }
+
+        // Query keys in case we don't have them for newly tracked members.
+        // This must be done before ensuring sessions. If not the devices of these users are not
+        // known yet and will not get the room key.
+        // We don't have API to only get the queries related to this member list, so we just
+        // process the pending `KeysQuery` requests from the olmMachine. (usually these are processed
+        // at the end of the sync, but we can't wait for that).
+        const request: OutgoingRequest[] = (await this.olmMachine.outgoingRequests()).filter(
```
richvdh, line 126:
```suggestion
        const requests: OutgoingRequest[] = (await this.olmMachine.outgoingRequests()).filter(
```

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -91,7 +103,33 @@ export class RoomEncryptor {
...
+            (r: OutgoingRequest) => r.type === RequestType.KeysQuery,
+        );
+        for (let i = 0; i < request.length; i++) {
+            await this.outgoingRequestProcessor.makeOutgoingRequest(request[i]);
+        }
```
richvdh, line 131:
```suggestion
        for (const request of requests) {
            await this.outgoingRequestProcessor.makeOutgoingRequest(request);
        }
```

richvdh, line 131: danger here is that we could end up making the `/keys/query` requests lots of times concurrently (once in the `outgoingRequestLoop`, once for each call to `ensureEncryptionSession`).

It might be best to do some refactoring of `outgoingRequestLoop` to allow us to wait for a single iteration of the loop (and not worry too much about limiting it to `KeysQuery` requests).

spec/integ/crypto/crypto.spec.ts
```diff
@@ -670,18 +670,21 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
 
         // tell alice she is sharing a room with bob
         syncResponder.sendOrQueueSyncResponse(getSyncResponse(["@bob:xyz"]));
         await syncPromise(aliceClient);
 
-        // we expect alice first to query bob's keys...
-        expectAliceKeyQuery(getTestKeysQueryResponse("@bob:xyz"));
-
-        // ... and then claim one of his OTKs
+        // Alice should claim one of his OTKs
```
richvdh, line 687: Nit:
```suggestion
        // Alice should claim one of Bob's OTKs
```

spec/integ/crypto/crypto.spec.ts
richvdh, line 673: For the future, maybe: consider `aliceClient.getHomeserverUrl()` rather than hardcoding the server url here.

spec/integ/crypto/crypto.spec.ts
richvdh: (likewise in various places below)

richvdh (changes requested):

src/rust-crypto/OutgoingRequestsManager.ts
```diff
@@ -0,0 +1,106 @@
...
+import { OlmMachine } from "@matrix-org/matrix-sdk-crypto-wasm";
+
+import { OutgoingRequest, OutgoingRequestProcessor } from "./OutgoingRequestProcessor";
+import { Logger } from "../logger";
+
+/**
+ * OutgoingRequestsManager: responsible for processing outgoing requests from the OlmMachine.
+ * Ensure that only one loop is going on at once, and that the requests are processed in order.
+ */
+export class OutgoingRequestsManager {
```
BillCarsonFr, line 26: Extracted the outgoing request loop management here, open to better naming if any ^^.

src/rust-crypto/RoomEncryptor.ts
BillCarsonFr: Damn, that's not intuitive. Done as you suggested

src/rust-crypto/RoomEncryptor.ts
BillCarsonFr: I have extracted the loop logic in OutgoingRequestsManager and added tests

src/rust-crypto/OutgoingRequestsManager.ts
```diff
@@ -0,0 +1,112 @@
...
+        if (this.isTaskRunning) {
+            // If the task is running, add the request to the queue and wait for completion.
+            // This ensures that the requests are processed only once at a time.
+            await new Promise<void>((resolve) => {
+                this.requestQueue.push(resolve);
+            });
+        } else {
+            await this.executeTask();
+        }
+    }
```
richvdh, line 68: I don't think we need a whole queue here; a single deferred will do.

For example, we could replace `requestQueue` with `nextLoopDeferred: undefined | IDeferred<void>`. 


```suggestion
    public doProcessOutgoingRequests(): Promise<void> {
        if (!this.isTaskRunning) {
            // The loop is not yet running. kick it off.
            await this.executeTask();
            return;
        }
     
        // The loop is already running, but we need to do another loop. Blah blah avoid race.   
        // Ensure that nextLoopDeferred is set, which tells `executeTask` to do another loop.
        if (!this.nextLoopDeferred) {
            this.nextLoopDeferred = defer();
        }
        
        await this.nextLoopDeferred.promise();
    }        
```

... and then in executeTask, replace the access to `requestQueue` with:

```typescript
let d = this.nextLoopDeferred;
this.nextLoopDeferred = undefined;
// ... loop
d.resolve();
```

src/rust-crypto/OutgoingRequestsManager.ts
```diff
@@ -0,0 +1,112 @@
...
+            this.isTaskRunning = false;
+        }
+
+        if (this.requestQueue.length > 0) {
+            if (this.stopped) {
+                this.requestQueue.forEach((resolve) => resolve());
+                return;
+            }
+            // there are a pending request that need to be executed
+            const awaitingRequests = this.requestQueue.map((resolve) => resolve);
```
richvdh, line 85: is this call to `map` just copying the array? It's not an idiom I'm familiar with: at first glance it looks like it might be trying to run the `resolve` callbacks. (Also, I suspect it's not that efficient.)

I don't think there's actually any need to copy it, since we're about to replace it with a new array anyway. But if you *did* need to copy it, I would use a spread operator and write a comment:

```suggestion
            // take a copy of the queue before we reset it.
            const awaitingRequests = [...this.requestQueue];
```

src/rust-crypto/OutgoingRequestsManager.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Process the outgoing requests from the OlmMachine.
+     *
+     * This should be called at the end of each sync, to process any requests that have been queued.
+     * In some cases if outgoing requests need to be sent immediately, this can be called directly.
+     *
+     * There is only one request running at once, and the others are queued.
+     * If a request is currently running the queued request will only trigger an additional run.
```
richvdh, line 56: the term "request" here is pretty confusing. I think you're using it to mean something different to the previous paragraph, where it refers to `OutgoingRequest`s from the OlmMachine. Could you be more explicit?

src/rust-crypto/OutgoingRequestsManager.ts
```diff
@@ -0,0 +1,112 @@
...
+                this.requestQueue.forEach((resolve) => resolve());
+                return;
+            }
+            // there are a pending request that need to be executed
+            const awaitingRequests = this.requestQueue.map((resolve) => resolve);
+            // reset the queue
+            this.requestQueue = [];
+
+            // run again and resolve all the pending requests.
+            await this.executeTask();
```
richvdh, line 90: this could potentially take a long time - indeed `executeTask` could theoretically recurse forever. Which means that the `awaitingRequests` might not be resolved for ages.

It might be easier/clearer to rewrite this as a `while` loop rather than a recursion.

src/rust-crypto/OutgoingRequestsManager.ts
```diff
@@ -0,0 +1,112 @@
...
+
+            awaitingRequests.forEach((resolve) => resolve());
+        }
+    }
+
+    private async processOutgoingRequests(): Promise<void> {
```
richvdh, line 96: could you give the private methods a doc-comment explaining what they do? I think it will help understand the code.

For example:

```suggestion
    /**
     * Make a single request to `olmMachine.outgoingRequests` and do the corresponding requests.
     */
    private async processOutgoingRequests(): Promise<void> {
```

src/rust-crypto/rust-crypto.ts
```diff
@@ -1261,7 +1256,7 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
     public onSyncCompleted(syncState: OnSyncCompletedData): void {
         // Processing the /sync may have produced new outgoing requests which need sending, so kick off the outgoing
         // request loop, if it's not already running.
-        this.outgoingRequestLoop();
+        this.outgoingRequestsManager.doProcessOutgoingRequests();
```
richvdh, line 1259: if you start an `async` method off without `await`ing, it, it's best to add a `.catch` to log any errors. (If for no other reason than it tells anyone reading the code that it is deliberately starting a background task.)

src/rust-crypto/rust-crypto.ts
```diff
@@ -1511,67 +1506,7 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
-    /** start the outgoing request loop if it is not already running */
-    private outgoingRequestLoop(): void {
-        if (this.outgoingRequestLoopRunning) {
-            // The loop is already running, but we have reason to believe that there may be new items in the queue.
-            //
-            // There is potential for a race whereby the item is added *after* `OlmMachine.outgoingRequests` checks
-            // the queue, but *before* it returns. In such a case, the item could sit there unnoticed for some time.
-            //
-            // In order to circumvent the race, we set a flag which tells the loop to go round once again even if the
-            // queue appears to be empty.
```
richvdh, line 1532: I'd like us to keep this comment, or something similar. I think it gives important context about why we do another loop.

richvdh (changes requested):
Some initial comments on the `OutgoingRequestManager` logic

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -108,27 +119,21 @@ export class RoomEncryptor {
         const fromServer = await this.room.loadMembersIfNeeded();
         const members = await this.room.getEncryptionTargetMembers();
 
-        if (fromServer) {
+        if (fromServer && !this.lazyLoadedMembersResolved) {
```
richvdh, line 122: I'm still not really sure what the return value of `loadMembersIfNeeded` is meant to mean. It seems to be undocumented, and nothing else seems to read it, which makes me doubly cautious.

Is the `fromServer` condition here necessary?

spec/unit/rust-crypto/OutgoingRequestsManager.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+import { Mocked } from "jest-mock";
+import * as RustSdkCryptoJs from "@matrix-org/matrix-sdk-crypto-wasm";
+
+import { OutgoingRequest, OutgoingRequestProcessor } from "../../../src/rust-crypto/OutgoingRequestProcessor";
+import { OutgoingRequestsManager } from "../../../src/rust-crypto/OutgoingRequestsManager";
+import { defer } from "../../../src/utils";
+import { logger } from "../../../src/logger";
+
+describe("OutgoingRequestsManager", () => {
+    /** the OutgoingRequestProcessor implementation under test */
```
richvdh, line 26: nit
```suggestion
    /** the OutgoingRequestsManager implementation under test */
```

spec/unit/rust-crypto/OutgoingRequestsManager.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+    let manager: OutgoingRequestsManager;
+
+    /** a mock  OutgoingRequestProcessor */
```
richvdh, line 29: nit

```suggestion
    /** a mock OutgoingRequestProcessor */
```

spec/unit/rust-crypto/OutgoingRequestsManager.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+    let processor: Mocked<OutgoingRequestProcessor>;
+
+    /** a mocked-up OlmMachine which processor is connected to */
```
richvdh, line 32:
```suggestion
    /** a mocked-up OlmMachine which manager is connected to */
```

spec/unit/rust-crypto/OutgoingRequestsManager.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+
+            await manager.doProcessOutgoingRequests();
+
+            expect(olmMachine.outgoingRequests).toHaveBeenCalledTimes(1);
+            expect(processor.makeOutgoingRequest).toHaveBeenCalledTimes(2);
+            expect(processor.makeOutgoingRequest).toHaveBeenCalledWith(request1);
+            expect(processor.makeOutgoingRequest).toHaveBeenCalledWith(request2);
+        });
+
+        it("Stack requests while one is already running", async () => {
```
richvdh, line 67: again I'm not really sure what we mean by "request" in this function. An `OutgoingRequest`? Or a call to `doProcessOutgoingRequests` ?

spec/unit/rust-crypto/OutgoingRequestsManager.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+            });
+
+            await manager.doProcessOutgoingRequests();
+
+            expect(olmMachine.outgoingRequests).toHaveBeenCalledTimes(1);
+        });
+    });
+
+    describe("Stop", () => {
+        it("Is stopped properly before making requests", async () => {
```
richvdh, line 122: Nit: *what* is stopped? The name of the test will be `OutgoingRequestsManager > Stop > Is stopped properly before making requests`, so that doesn't make any sense.

You could either rely on the context from the `describe` block ("Stop"), so something like `it("stops the loop properly before making requests", ... )`, or make the test name explicit: `it("The loop is stopped properly before making requests", ...)`

Likewise below.

spec/unit/rust-crypto/OutgoingRequestsManager.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+            // processed in the same loop.
+            expect(olmMachine.outgoingRequests).toHaveBeenCalledTimes(2);
+
+            expect(processor.makeOutgoingRequest).toHaveBeenCalledTimes(3);
+            expect(processor.makeOutgoingRequest).toHaveBeenCalledWith(request1);
+            expect(processor.makeOutgoingRequest).toHaveBeenCalledWith(request2);
+            expect(processor.makeOutgoingRequest).toHaveBeenCalledWith(request3);
+        });
+
+        it("Should not bubble if request is rejected", async () => {
```
richvdh, line 105:
```suggestion
        it("Should not bubble exceptions if request is rejected", async () => {
```

richvdh (commented):
a few comments on the other bits

src/rust-crypto/rust-crypto.ts
BillCarsonFr: Added in the OutgoingRequestsManager doc

src/rust-crypto/RoomEncryptor.ts
BillCarsonFr: I think that if you restart the browser ?

spec/unit/rust-crypto/OutgoingRequestsManager.spec.ts
BillCarsonFr: Updated the names

src/rust-crypto/OutgoingRequestsManager.ts
BillCarsonFr: removed in favor of deferred

src/rust-crypto/OutgoingRequestsManager.ts
BillCarsonFr: I updated the naming using calls for `doProcessOutgoingRequests`, and explicitly olmMachine OutgoingRequests for requests

src/rust-crypto/OutgoingRequestsManager.ts
BillCarsonFr: I have changed the logic here. There was a problem with the `await` here. I added a new test to ensure it's working.
It's still calling doProcessOutgoingRequests inside doProcessOutgoingRequests (but not waiting on it). It's not exactly reccursive, but moving to a non reccursive way makes the code les lisible I think

src/rust-crypto/rust-crypto.ts
BillCarsonFr: added a .catch

src/rust-crypto/RoomEncryptor.ts
richvdh: Not following you. What happens if you restart the browser?

src/rust-crypto/OutgoingRequestsManager.ts
```diff
@@ -0,0 +1,120 @@
...
+     * In some cases if OutgoingRequests need to be sent immediately, this can be called directly.
+     *
+     * Calls to doProcessOutgoingRequests() are processed synchronously, one after the other, in order.
+     * If doProcessOutgoingRequests() is called while another call is still being processed, it will be queued.
+     * Multiple calls to doProcessOutgoingRequests() when a call is already processing will be batched together.
+     */
+    public async doProcessOutgoingRequests(): Promise<void> {
+        if (this.isTaskRunning) {
+            // If the task is running, add the request to the queue and wait for completion.
+            // This ensures that the requests are processed only once at a time.
```
richvdh, line 66: there is no queue here.

```suggestion
            // If the task is already running, flag that we need another iteration, and wait for it to complete.
            // This ensures that the requests are processed only once at a time.
```

src/rust-crypto/OutgoingRequestsManager.ts
```diff
@@ -0,0 +1,120 @@
...
+                const deferred = this.nextLoopDeferred;
+                // reset the nextLoopDeferred so that any future requests are queued for another additional iteration.
+                this.nextLoopDeferred = undefined;
+
+                // Run again and resolve all the pending requests.
+                // Notice that we don't await on it, so that the current promise is resolved now.
+                // The requests that were deferred will be resolved after this new iteration.
+                this.doProcessOutgoingRequests().then(() => {
+                    deferred.resolve();
+                });
```
richvdh, line 96: we *might* be able to rely on `doProcessOutgoingRequests` not failing, but it doesn't seem safe to me.

If it does fail then `deferred` will never resolve.

Best to pass on the failure:

```suggestion
                this.doProcessOutgoingRequests().then(deferred.resolve, deferred.reject);
```

src/rust-crypto/OutgoingRequestsManager.ts
```diff
@@ -0,0 +1,120 @@
...
+     * Multiple calls to doProcessOutgoingRequests() when a call is already processing will be batched together.
+     */
+    public async doProcessOutgoingRequests(): Promise<void> {
+        if (this.isTaskRunning) {
+            // If the task is running, add the request to the queue and wait for completion.
+            // This ensures that the requests are processed only once at a time.
+            if (!this.nextLoopDeferred) {
+                this.nextLoopDeferred = defer();
+            }
+            return this.nextLoopDeferred.promise;
```
richvdh, line 70: Nit: returning a promise from an async function works, but is a bit weird. Maybe better to split  `doProcessOutgoingRequests` into a regular function and an async one.

src/rust-crypto/OutgoingRequestsManager.ts
```diff
@@ -0,0 +1,120 @@
...
+     * Process the OutgoingRequests from the OlmMachine.
+     *
+     * This should be called at the end of each sync, to process any OlmMachine OutgoingRequests created by the rust sdk.
+     * In some cases if OutgoingRequests need to be sent immediately, this can be called directly.
+     *
+     * Calls to doProcessOutgoingRequests() are processed synchronously, one after the other, in order.
+     * If doProcessOutgoingRequests() is called while another call is still being processed, it will be queued.
+     * Multiple calls to doProcessOutgoingRequests() when a call is already processing will be batched together.
+     */
+    public async doProcessOutgoingRequests(): Promise<void> {
```
richvdh, line 63: nit: I am not keen on the naming here. Normally if you have both a `foo` and a `doFoo`, the `doFoo` is the private method that actually does the thing, while `foo` is the public method that wraps it.

In other words: can we swap the names of `processOutgoingRequests` and `doProcessOutgoingRequests` ?

src/rust-crypto/OutgoingRequestsManager.ts
```diff
@@ -0,0 +1,120 @@
...
+
+                // Run again and resolve all the pending requests.
+                // Notice that we don't await on it, so that the current promise is resolved now.
+                // The requests that were deferred will be resolved after this new iteration.
+                this.doProcessOutgoingRequests().then(() => {
+                    deferred.resolve();
+                });
+            }
+        }
+    }
```
richvdh, line 99: Out of interest I tried rewriting this without the recursion.

IMHO it's clearer, but YMMV.

```suggestion
    public doProcessOutgoingRequests(): Promise<void> {
        // flag that we need at least one more iteration of the loop.
        if (!this.nextLoopDeferred) {
            this.nextLoopDeferred = defer();
        }
        
        // ... and wait for it to complete.
        let result = nextLoopDeferred.promise;
        
        // set the loop going if it is not already.
        if (!this.outgoingRequestLoopRunning) {
            this.outgoingRequestLoopInner().catch((e) => {
                // this should not happen; outgoingRequstLoopInner should return any errors via `nextLoopDeferred`.
                this.logger.error("Uncaught error in outgoing request loop", e);
            });
        }

        return result;
    } 
        
    private async outgoingRequestLoopInner(): Promise<void> {
        /* istanbul ignore if */
        if (this.outgoingRequestLoopRunning) {
            throw new Error("Cannot run two outgoing request loops");
        }        

        this.outgoingRequestLoopRunning = true;
        try {
            while (!this.stopped && this.nextLoopDeferred) {
                const deferred = this.nextLoopDeferred;
                
                // reset the nextLoopDeferred so that any future calls to doProcessOutgoingRequests are queued for another additional iteration.
                this.nextLoopDeferred = undefined;
                                
                // make the requests and feed the results back to the `nextLoopDeferred`
                await this.processOutgoingRequests().then(deferred.resolve, deferred.reject);
            }
        } finally {
             this.outgoingRequestLoopRunning = false;
         }
    }
```

richvdh (approved):
If you're keen to merge this, go ahead. I'd *prefer* it to be rewritten without the recursion but won't insist.

spec/unit/rust-crypto/OutgoingRequestsManager.spec.ts
```diff
@@ -0,0 +1,237 @@
+/*
+Copyright 2024 The Matrix.org Foundation C.I.C.
```
richvdh, line 2:
```suggestion
Copyright 2023 The Matrix.org Foundation C.I.C.
```

src/rust-crypto/OutgoingRequestsManager.ts
```diff
@@ -0,0 +1,141 @@
+/*
+Copyright 2024 The Matrix.org Foundation C.I.C.
```
richvdh, line 2:
```suggestion
Copyright 2023 The Matrix.org Foundation C.I.C.
```

PR #3831, opened 2023-10-24
Change latest node ver to '*'

dbkr:
This uses the latest cached version rather than fetching the latest released version so we don't reply on (and hammer) node's download servers for the very latest version before the actions runners get updated. We'll still stay current, just not quite so aggressively current.

## Checklist

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ env:
     ENABLE_COVERAGE: ${{ github.event_name != 'merge_group' }}
 jobs:
     jest:
-        name: "Jest [${{ matrix.specs }}] (Node ${{ matrix.node }})"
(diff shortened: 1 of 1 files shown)


t3chguy:
@dbkr could you improve the naming too?

image

t3chguy:
Also given we also have to download `18` anyway (and 20 this month when it enters LTS)

dbkr:
I don't think we can make it the actual version as I'm assuming you can't use the output of a step in the name of a job. Might be able to fix it to 'latest avail' or something?

t3chguy:
Sure something like `default` or even `latest` is better than `*` - plus `latest` means no need to change branch protections

dbkr:
We already have node 18 available though on the runner image as it is at the moment (https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#nodejs). Hopefully these will be the same, generally speaking.

t3chguy (approved):

PR #3833, opened 2023-10-25
Rotate per-participant keys when a member leaves

dbkr:
Base on https://github.com/matrix-org/matrix-js-sdk/pull/3667

With a delay both before making a new key, to try to batch up multiple
people leaving into a single key change, and a delay before actually
using the new key to allow time for it to arrive.

This increasingly feels like storing our own sender key in the same set
is suboptimal because we're starting to have to treat it more & more
specially.

This needs https://github.com/livekit/client-sdk-js/pull/908 to work reliably

## Checklist

---

## ✨ Features
 * Rotate per-participant keys when a member leaves ([\#3833](https://github.com/matrix-org/matrix-js-sdk/pull/3833)).


src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -510,20 +556,25 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
-            this.memberships
-                .filter((m) => !isMyMembership(m))
-                .map(getParticipantIdFromMembership)
-                .sort()
-                .join();
-
-        if (callMembersChanged && this.isJoined()) {
-            this.requestKeyEventSend();
+
+        if (this.isJoined() && this.makeNewKeyTimeout === undefined) {
```
weeman1337, line 560: If I got it correctly, in the current implementation everyone must leave within 3 seconds.  Should the timeout not better be extended here to wait for another amount of time? So that keys are send when there was no leave for 3 seconds?

dbkr: I didn't wan't to have a leave extend the period since I guess this could, in theory, be used to prevent a participant form ever changing their keys. This seemed simple & good as a start: it's not critical if keys have to be rotated again.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -26,13 +26,21 @@ import { MatrixError, MatrixEvent } from "../matrix";
...
 
 const MEMBERSHIP_EXPIRY_TIME = 60 * 60 * 1000;
 const MEMBER_EVENT_CHECK_PERIOD = 2 * 60 * 1000; // How often we check to see if we need to re-send our member event
 const CALL_MEMBER_EVENT_RETRY_DELAY_MIN = 3000;
 const UPDATE_ENCRYPTION_KEY_THROTTLE = 3000;
 
+// A delay after a member leaves before we create and publish a new key, because people
+// tend to leave calls at the same time
+const MAKE_KEY_DELAY = 3000;
+// The between creating and sending a new key and starting to encrypt with it. This gives others
```
robintown, line 38: I think you're missing a word here

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -462,12 +506,15 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+            // We store our own sender key in the same set along with keys from others, so it's
+            // important we don't allow our own keys to be set by one of these events (apart from
+            // the fact that we don't need it anyway because we already know our own keys).
+            logger.info("Ignoring our own keys event");
+            return;
+        }
 
+        for (const key of content.keys) {
             const encryptionKey = key.key;
             const encryptionKeyIndex = key.index;
```
robintown, line 519: These accesses are without a guard that checks whether `key` is null or undefined

toger5 (approved):

PR #3834, opened 2023-10-25
Fix up comments on `globalErrorOnUnknownDevices`

richvdh:
The current deprecation notice advises you to use a method which does something completely different.

Fixing this "properly" is slightly challenging because we don't want to support setting it to `true` in Rust Crypto; yet I don't really want to change the default for legacy crypto.

Let's just document the behaviour for now.

---


florianduros (approved):
LGTM

PR #3835, opened 2023-10-26
feat: Add autoformat and lint for ts/tsx files

Palid:
Implementation of autoformatting&autolinting on precommit hook.

Solves https://github.com/vector-im/element-web/issues/18230 (1/3)

---


.lintstagedrc
```diff
@@ -0,0 +1,3 @@
+{
+    "*.(ts|tsx)": ["eslint --fix", "prettier --write"]
```
t3chguy, line 2:
```suggestion
    "*.(ts|tsx)": ["eslint --fix", "prettier --write"]
    "*.(py|md|yaml)": ["prettier --write"]
```

t3chguy (approved):
LGTM - thanks!

PR #3839, opened 2023-10-31
Age fallback using origin_server_ts instead of 0

toger5:
## Checklist

---

## 🐛 Bug Fixes
 * Age fallback using origin_server_ts instead of 0 ([\#3839](https://github.com/matrix-org/matrix-js-sdk/pull/3839)). Contributed by @toger5.


src/matrixrtc/CallMembership.ts
```diff
@@ -91,7 +91,7 @@ export class CallMembership {
     }
 
     public isExpired(): boolean {
-        return this.getAbsoluteExpiry() < this.parentEvent.getTs() + this.parentEvent.getLocalAge();
+        return this.getMsUntilExpiry() < 0;
```
dbkr, line 94: Pedantic probably, but surely <=

dbkr (approved):

spec/unit/matrixrtc/CallMembership.spec.ts
```diff
@@ -85,7 +85,9 @@ describe("CallMembership", () => {
 
     it("considers memberships expired when local age large", () => {
         const fakeEvent = makeMockEvent(1000);
-        fakeEvent.getLocalAge = jest.fn().mockReturnValue(6000);
+        const localAge = 6000;
+        fakeEvent.getLocalAge = jest.fn().mockReturnValue(localAge);
+        fakeEvent.localTimestamp = Date.now() - localAge;
         const membership = new CallMembership(fakeEvent, membershipTemplate);
         expect(membership.isExpired()).toEqual(true);
     });
```
toger5, line 93: We now have a couple of situations in which we need to update the test to also use a custom `localTimestamp` because that is used by our `isExpired()` fn.

Is it worth refactoring the `isExpired()` so that it uses `getLocalAge` instad of localTimestamp? It would make the fn a bit odd:
```
    // gets the expiry time of the event, converted into the device's local time
        const relativeCreationTime = this.parentEvent.getTs() - this.createdTs();
        const localEventCreationTs = Date.now() - this.parentEvent.getLocalAge();
        const localMembershipCreationTs = localEventCreationTs - relativeCreationTime;

        return localMembershipCreationTs + this.data.expires;
    }

    public getMsUntilExpiry(): number {
        return this.getLocalExpiry() - Date.now();
    }

    public isExpired(): boolean {
        return this.getMsUntilExpiry() <= 0;
    }
```

Edit: above is how it would look like. I think when using this phrasing it kinda makes sense. The localEvenCreationTs is `now` - `the age`.
`const localEventCreationTs = Date.now() - this.parentEvent.getLocalAge();`
Its just odd when considering:
```
    public getLocalAge(): number {
        return Date.now() - this.localTimestamp;
    }
    ```

toger5: I think I found a better solution.

PR #3849, opened 2023-11-01
Bump matrix-sdk-crypto-wasm to 3.0.1

richvdh:
... which changes the API of `bootstrapCrossSigning` a bit.

Fixes https://github.com/vector-im/crypto-internal/issues/160

---


t3chguy (changes requested):
CI says no

richvdh:
> CI says no

And it has a point. Turns out the SDK has introduced a bug: https://github.com/matrix-org/matrix-rust-sdk/issues/2802

richvdh:
This is now ready for another look.

t3chguy (approved):

PR #3853, opened 2023-11-03
Revert "Age fallback using origin_server_ts instead of 0"

richvdh:
Reverts matrix-org/matrix-js-sdk#3839.

This turns out to have caused test failures (https://github.com/vector-im/element-web/issues/26500), and there is debate about what the correct fix is.

Fixes https://github.com/vector-im/element-web/issues/26500

---


t3chguy (approved):

toger5 (approved):
I think reverting for now makes sense.
Sorry for all the pain this caused.

Although Midhun's fix could be the better thing to merge. So we could also take the risk and see if it then works.
But we dont have to rush this one. Maybe its best unblocking everyone for sure (by reverting) and then retesting in a new pr.

PR #3854, opened 2023-11-03
Better fallback for unavailable event age

toger5:
This fixes an issue where state (particularly call member state) events were not considered expired even if they were very old. Resulted in us showing really old calls for rooms.
This applies: https://github.com/matrix-org/matrix-js-sdk/pull/3839 again after the tests were fixed downstream in the react sdk. https://github.com/matrix-org/matrix-react-sdk/pull/11827

## Checklist

---


toger5:
Is there a way to run downstream test in the PR? without adding it to the merge queue?

t3chguy:
@toger5 no, only locally

toger5:
I ran the unit tests for the matrix-react-sdk about 10 times now and they all succeeded with the fix from Midhun https://github.com/matrix-org/matrix-react-sdk/pull/11827 with this js-sdk pr.

dbkr (approved):
If you could put in the title / description that this is the same PR re-applied that would be good - took me diffing the PR against the old one to realise it was the same PR. :)

PR #3855, opened 2023-11-03
Cypress workflow: remove redundant `rust-crypto` param

richvdh:
https://github.com/matrix-org/matrix-react-sdk/pull/11828 removes the `rust-crypto` input param from the reusable cypress workflow. This gets rid of it on the calling side.


richvdh:
Actually this only makes sense when updating the version of the workflow we pull in. It's currently pinned at `v3.83.0-rc.1`. https://github.com/matrix-org/matrix-react-sdk/pull/11828 predates the branch for v3.84 so this might need to wait for a while.

andybalaam:
Force merging since the problem this fixes prevent the CI from running.

PR #3861, opened 2023-11-07
remove unhelpful logging about events without age params.

ara4n:
the conditional was unnecessary, the logline wasn't wrapped, and it creates literally tens of thousands of warn-level logspam on my account

---

### DIFF
--- a/src/models/event.ts
+++ b/src/models/event.ts
@@ -668,11 +668,6 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
      * It is supposed to only be used like this: `ev.getAge() ?? ev.fallbackAge()`
      */
     private fallbackAge(): number {
-        if (!this.getAge()) {
-            logger.warn(
-                "Age for event was not available, using `now - origin_server_ts` as a fallback. If the device clock is not correct issues might occur.",
-            );
-        }
         return Math.max(Date.now() - this.getTs(), 0);
     }
 


richvdh (commented):
I am led to wonder why you are seeing events without an `age` in the first place... It seems like the right fix here might be to change that rather than to keep calling a method that, to quote its documentation, "should never be used".

The presence of the conditional is odd though.

ara4n:
i assume they are just old events?

toger5:
The intention for adding this was that I observed numerous events without age (especially state events where we relied on the age)

I was hoping to confirm that this was just a rare issue I ran into because of some reason. So I indeed expected this log to almost never be called.

What is the rational behind the age property? Is it acceptable to not be included in events? In what conditions is that acceptable.

The fact, that this code path is always called is definitly an issue we need to investigate. 

I will look into this now. (Except if someone else already started)

toger5:
In case it is per spec to not include the age for specific events, this indeed seems to be the right fix. The codepath still makes sense since it updates the `localTimestamp` to be `now - origin_server_ts` instaed of `now` which is the better approximation.

toger5 (approved):
If not sending an age for events is acceptable and does not imply other issues removing the warning makes sense.
For call events the age is somewhat important. So another solution would be to change the condition to. `this.AGE_DEPENENT_EVENTS.includes(this.eventType())`

and make `AGE_DEPENDANT_EVENTS` a list of event types where a missing age implies potential client issues.

(`m.call.member` would be one of those events)

richvdh:
> What is the rational behind the age property? Is it acceptable to not be included in events? In what conditions is that acceptable.

As I said: typing, presence, and read-receipts do not have an `age` property. It is debatable whether any of those count as "events" but for whatever reason, the `MatrixEvent` constructor is being called for those objects.

I really don't think we should merge this as is. Fundamentally the fact that we need to rip out this logging means that we are *expecting* to call code that is currently documented as "should never be used". That is extremely misleading to anyone else attempting to read the code in future.

At the very least, we should update the documentation on the function to better describe the circumstances under which it is used. However, now that I look even closer, I have more questions about the original PR:

* Typing, presence and read-receipts also lack an `origin_server_ts`, so this is going to set `localTimestamp` to `NaN` for such events, which seems significantly worse than the current `Date.now()`.
* Although the [spec](https://spec.matrix.org/v1.8/client-server-api/#definition-clientevent) makes `age` an optional property, in practice I would strongly expect it to be present for all "real" events served over the C-S API by Synapse. If you're seeing something different, I think we should probably find out why that is.
* Maybe less important, but tbh I find the structure of the code a bit odd - in particular, why do we have two copies of `Date.now() - (this.getAge() ?? this.fallbackAge())` rather than a single function that returns the right answer?

It's also worth noting, that even where `age` is present, it is known to be [fundamentally broken](https://github.com/matrix-org/synapse/issues/8429). To be honest I have never really understood why `age` is more reliable than just using `origin_server_ts` in the first place, but the fix is pretty easy, so if you have a useful usecase for a working `age` I would recommend letting the backend team know about it!

toger5:
> Typing, presence and read-receipts also lack an origin_server_ts, so this is going to set localTimestamp to NaN for such events, which seems significantly worse than the current Date.now().

That sounds like it would be a good filter to solve the fallback.
```ts
private fallbackAge(): number {
        if (!this.getTs()) {
            // In case there is no origin_server_ts there is no way to do a fallback if getAge() is returning undefined 
            return 0;
        }
        if (!this.getAge()) {
            logger.warn(
                "`unsigned.age` for an event that has an `origing_server_ts` not available. `now - origin_server_ts` is used as a fallback. If the device clock is not correct issues might occur.",
            );
        }
        return Math.max(Date.now() - this.getTs(), 0);
    }
```

> Although the [spec](https://spec.matrix.org/v1.8/client-server-api/#definition-clientevent) makes age an optional property, in practice I would strongly expect it to be present for all "real" events served over the C-S API by Synapse. If you're seeing something different, I think we should probably find out why that is.

This sounds like it is of interest to have warnings in the console in case there is an issue. With the condition that we only even try all this if there is a `origin_server_ts` the logging should also not be all over the place anymore if I undestand you correctly.

> Maybe less important, but tbh I find the structure of the code a bit odd - in particular, why do we have two copies of Date.now() - (this.getAge() ?? this.fallbackAge()) rather than a single function that returns the right answer?

I see. The reason I was implementing it this way is to not change `this.getAge()` I was not sure if there is code where we check on/rely on it beeing `undefined`. So I wanted to explicitly add the fallback only where we used `this.getAge() ?? 0`.

I agree that: 
```ts
Date.now() - (this.getAge() ?? this.fallbackAge())
// if getAge == undefined this equals
Date.now() - Math.max(Date.now() - this.getTs(), 0);
// which in most cases is
Date.now() - Date.now() + this.getTs()
```
is kinda stupid...

It though it makes it more readable however. The alternative would be to have a fallback localTimestamp() that would just be `getTs()`. But then we would have this code:
```ts
this.localTimestamp = this.getAge() !== undefined ? Date.now() - this.getAge() : (this.getTs() !== undefined ? this.getTs() : 0);

// this could be made more readable however:
const localTimeStampFromAge = this.getAge() !== undefined ? Date.now() - this.getAge() : undefined;
const localTimeStampFromOriginServerTs = this.getTs();
this.localTimestamp = localTimeStampFromAge ?? localTimeStampFromOriginServerTs ?? Date.now();
```
```

richvdh:
closing this PR in favour of https://github.com/matrix-org/matrix-js-sdk/pull/3870. Let's debate a correct implementation of the age fallback elsewhere.

ara4n:
> To be honest I have never really understood why age is more reliable than just using origin_server_ts in the first place,

I perpetrated `age` in the first place. The reason for it (in 2014 or whever) was to have a measurement of event staleness which didn't rely on the handset's clock being correct (as otherwise it won't be able to compare origin_server_ts sensibly), and the use case was to decide when to ignore stale 1:1 VoIP signalling events, to try to stop spuriously apps ringing when they do an initial/incr sync and see old VoIP events.

PR #3862, opened 2023-11-08
Shorten TimelineWindow when an event is removed

andybalaam:
Fixes https://github.com/vector-im/element-web/issues/26498 even when https://github.com/matrix-org/matrix-js-sdk/pull/3798 is present.

Listen for events being removed in the TimelineWIndow, and shorten the window if the removal makes it longer than the available events.

---

## 🐛 Bug Fixes
 * Shorten TimelineWindow when an event is removed ([\#3862](https://github.com/matrix-org/matrix-js-sdk/pull/3862)). Fixes vector-im/element-web#26498. Contributed by @andybalaam.

### DIFF
--- a/src/timeline-window.ts
+++ b/src/timeline-window.ts
@@ -199,11 +199,11 @@ export class TimelineWindow {
         return false;
     }
 
-    private onTimelineEvent = (_event?: MatrixEvent, _room?: Room, _atStart?: boolean, removed?: boolean): void => {
+    private onTimelineEvent(_event?: MatrixEvent, _room?: Room, _atStart?: boolean, removed?: boolean): void {
         if (removed) {
             this.onEventRemoved();
         }
-    };
+    }
 
     /**
      * If an event was removed, meaning this window is longer than the timeline,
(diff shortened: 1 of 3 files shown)


src/timeline-window.ts
```diff
@@ -193,6 +193,17 @@ export class TimelineWindow {
         return false;
     }
 
+    /**
+     * Notify the window that an event was removed, so we must shrink our
+     * window, if the removed event was inside this window.
+     */
+    public removeEvent(): void {
+        const events = this.getEvents();
+        if (events.length > 0 && events.at(-1) === undefined && this.end) {
```
t3chguy, line 202: I don't think we can use `.at` - see https://github.com/vector-im/element-web/issues/24245

t3chguy (changes requested):

src/timeline-window.ts
```diff
@@ -87,6 +92,7 @@ export class TimelineWindow {
         opts: IOpts = {},
     ) {
         this.windowLimit = opts.windowLimit || 1000;
+        timelineSet.room?.on(RoomEvent.Timeline, this.onTimelineEvent.bind(this));
```
t3chguy, line 95: This bind is superfluous, you're using an arrow function already.

t3chguy (changes requested):

t3chguy (approved):

PR #3865, opened 2023-11-08
Automate checking there is no published release using the version already

t3chguy:
Fixes https://github.com/vector-im/wat-internal/issues/15

---


andybalaam (approved):

andybalaam:
All tests that caused this to be kicked out of the queue are now disabled.

andybalaam:
(Although this was re-added to the queue just before I merged the disablement, so there's a chance we'll need one more round...)

PR #3867, opened 2023-11-09
Ensure `setUserCreator` is called when a store is assigned

MidhunSureshR:
Fixes https://github.com/vector-im/element-web/issues/26520

If you're interested in know why we need a `userCreator` in the first place, see https://github.com/matrix-org/matrix-js-sdk/pull/3796#discussion_r1362060951

This PR merely adds a method that sets the store property while making sure that userCreator is also wired correctly.

---

## 🐛 Bug Fixes
 * Ensure `setUserCreator` is called when a store is assigned ([\#3867](https://github.com/matrix-org/matrix-js-sdk/pull/3867)). Fixes vector-im/element-web#26520. Contributed by @MidhunSureshR.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1497,7 +1497,6 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 
     public set store(newStore: Store) {
-        if (!newStore) throw new Error("store passed to MatrixClient.store setter is undefined!");
         this._store = newStore;
         this._store.setUserCreator((userId) => User.createUser(userId, this));
     }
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -1497,6 +1496,16 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         this._secretStorage = new ServerSideSecretStorageImpl(this, opts.cryptoCallbacks ?? {});
     }
 
+    public set store(newStore: Store) {
+        if (!newStore) throw new Error("store passed to MatrixClient.store setter is undefined!");
```
t3chguy, line 1500: Why don't we just trust TS here?

t3chguy (approved):

andybalaam:
All tests that flaked causing this to be removed from queue have now been disabled.

PR #3868, opened 2023-11-09
Update workfile to pull in changes from react-sdk

MidhunSureshR:
For fixing broken cypress tests. 

The current workfile is pinned to an earlier tag that does not contain the changes to disable sorry-cypress

---


andybalaam (approved):
Thanks!

andybalaam:
Force-merged because the merge queue build failed due to the problem this is trying to fix.

PR #3870, opened 2023-11-09
Revert "Better fallback for unavailable event age"

richvdh:
Reverts matrix-org/matrix-js-sdk#3854.

This causes significant logspam, and seems like it needs more thought for various other reasons which I've tried to set out at https://github.com/matrix-org/matrix-js-sdk/pull/3861#issuecomment-1803863086.

---


t3chguy (approved):

dbkr (approved):
Thanks for digging into this & doing the reverts.

PR #3872, opened 2023-11-10
Automatically add tech-debt issues to the right project

Johennes:
This is the equivalent of https://github.com/vector-im/element-desktop/blob/develop/.github/workflows/triage-labelled.yml.

---


t3chguy (approved):

andybalaam:
All tests that flaked causing this to be removed from queue have now been disabled.

Issue #3891, opened 2023-11-16
"Deploy documentation PR preview" builds are failing


andybalaam:
Here is an example of a failed run: https://github.com/matrix-org/matrix-js-sdk/actions/runs/6883595442

Snippet:

```
Run matrix-org/netlify-pr-preview@v2
Run bobheadxi/deployments@88ce5600046c82542f8246ac287d0a53c461bca3
targeting matrix-org/matrix-js-sdk
initializing new deployment for PR Documentation Preview @ e001e57a3483c360b8eac5bb6837e4ff9189feca
created deployment 1175042241 for PR Documentation Preview @ e001e57a3483c360b8eac5bb6837e4ff9189feca
created deployment status 2779009591 with status "in_progress"
Run matrix-org/pr-details-action@v1.3
Run actions/github-script@v6
Run nwtgck/actions-netlify@5da65c9f74c7961c5501a3ba329b8d0912f39c03
  with:
    publish-dir: docs
    deploy-message: Deploy from GitHub Actions
    enable-pull-request-comment: false
    enable-commit-comment: false
    alias: pr3886
  env:
    NETLIFY_AUTH_TOKEN: ***
    NETLIFY_SITE_ID: ***
Error: Unauthorized
```

I have created a new auth token in netlify and supplied it in the environment config here: https://github.com/matrix-org/matrix-js-sdk/settings/environments/711767567/edit but that did not affect the error.

Note that the relevant netlify account is the matrix-org one.

The workflow file is: https://github.com/matrix-org/matrix-js-sdk/blob/develop/.github/workflows/docs-pr-netlify.yaml . It is kicked off when "[Static Analysis](https://github.com/matrix-org/matrix-js-sdk/blob/develop/.github/workflows/static_analysis.yml)" finishes, which runs on pull requests, the merge queue, and pushes to develop and master, but within docs-pr-netlify.yaml we restrict it to only do anything on pull requests.

PR #3893, opened 2023-11-16
Specify the correct environment for the docs builder workflow

andybalaam:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/3891

---


t3chguy (approved):

andybalaam:
Looks like the job started passing even without this change, e.g. https://github.com/matrix-org/matrix-js-sdk/actions/runs/6889752387 . I think it passed because I deleted the NETLIFY_AUTH_KEY from the repo, leaving only my new one in the environment.

I don't understand why it used the correct environment, so I am inclined to think that this change is still good, to clarify which environment we are using.

t3chguy:
image

@andybalaam I don't think your analysis is quite right

andybalaam:
Looks good:

image

PR #3897, opened 2023-11-17
Stop logging decryptions as retries

richvdh:
Somebody seems to have decided that `isRetry` needs setting whenever we try to decrypt an event. This is nonsense, and leads to confusing logs.

---

### DIFF
--- a/src/models/MSC3089Branch.ts
+++ b/src/models/MSC3089Branch.ts
@@ -162,9 +162,8 @@ export class MSC3089Branch {
 
         if (!event) throw new Error("Failed to find event");
 
-        // Sometimes the event isn't decrypted for us, so do that. We specifically set `emit: true`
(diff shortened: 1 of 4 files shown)


src/models/MSC3089Branch.ts
```diff
@@ -162,9 +162,8 @@ export class MSC3089Branch {
 
         if (!event) throw new Error("Failed to find event");
 
-        // Sometimes the event isn't decrypted for us, so do that. We specifically set `emit: true`
-        // to ensure that the relations system in the sdk will function.
```
richvdh, line 166: `emit: true` is the default, so there's no reason to imagine the "relations system" won't work without it.

MidhunSureshR (approved):

richvdh:
force-merging because I'm not writing a test today for this

PR #3898, opened 2023-11-17
Improve logging of event encryption in `RustCrypto`

richvdh:
It's currently rather hard to see what's going on during event encryption in the rust crypto stack. One of the problems is that you can have multiple encryption operations overlapping.

The main idea here is to include an identifier for the event being encrypted on each log line. (We don't yet have an event ID, so instead use the txn ID.) In order to pass this down through the call hierarchy, I've created the idea of a `LogSpan` (drawing inspiration from [OpenTracing](https://opentracing.io/docs/overview/spans/)).

---

### DIFF
--- a/src/rust-crypto/rust-crypto.ts
+++ b/src/rust-crypto/rust-crypto.ts
@@ -375,7 +375,7 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
         const encryptor = this.roomEncryptors[room.roomId];
 
         if (encryptor) {
-            encryptor.ensureEncryptionSession(this.globalBlacklistUnverifiedDevices);
+            encryptor.prepareForEncryption(this.globalBlacklistUnverifiedDevices);
         }
     }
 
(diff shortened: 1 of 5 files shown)


richvdh:
> (drawing inspiration from [OpenTracing](https://opentracing.io/docs/overview/spans/)).

Maybe we should just be using opentracing itself. That's not a job I particularly want to take on today though.

src/rust-crypto/rust-crypto.ts
```diff
@@ -375,7 +375,7 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
         const encryptor = this.roomEncryptors[room.roomId];
 
         if (encryptor) {
-            encryptor.ensureEncryptionSession(this.globalBlacklistUnverifiedDevices);
+            encryptor.prepareForEncryption(this.globalBlacklistUnverifiedDevices);
```
andybalaam, line 378: This looks unrelated to the logging change?

andybalaam (changes requested):
One question - looks good otherwise.

src/rust-crypto/rust-crypto.ts
richvdh: No it's part of the same thing. `ensureEncryptionSession` has gained an extra param and become private. `prepareForEncryption` is a new wrapper for it.

andybalaam: Great, thanks.

andybalaam (approved):

richvdh:
tests failed due to https://github.com/vector-im/element-web/issues/26603. Retrying

PR #3899, opened 2023-11-17
move get device key API from client to crypto

BillCarsonFr:
`MatrixClient` API was exposing two methods that only worked for legacy crypto:
    - `getDeviceEd25519Key`
    - `getDeviceCurve25519Key`

=> These are used in the react-sdk for some functionality (rageshake, sentry, rendez-vous).

I have deprectaded those calls from `MatrixClient` and created a new API in `CryptoApi` (where it belongs):

`getOwnDeviceKeys(): Promise<OwnDeviceKeys>`

## Checklist

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2391,7 +2391,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @returns base64-encoded ed25519 key. Null if crypto is
      *    disabled.
      *
-     * @deprecated Use {@link CryptoAPI.getOwnDeviceKeys}
+     * @deprecated Use {@link CryptoApi.getOwnDeviceKeys}
      */
 
     public getDeviceEd25519Key(): string | null {
@@ -2404,7 +2404,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @returns base64-encoded curve25519 key. Null if crypto is
      *    disabled.
      *
-     * @deprecated Use {@link CryptoAPI.getOwnDeviceKeys}
+     * @deprecated Use {@link CryptoApi.getOwnDeviceKeys}
      */
     public getDeviceCurve25519Key(): string | null {
         return this.crypto?.getDeviceCurve25519Key() ?? null;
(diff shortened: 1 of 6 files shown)


src/rust-crypto/rust-crypto.ts
```diff
@@ -157,6 +157,21 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
+        // could be undefined if there is no such algorithm for that device.
+        if (device.curve25519Key && device.ed25519Key) {
+            return {
+                ed25519: device.ed25519Key.toBase64(),
+                curve25519: device.curve25519Key.toBase64(),
+            };
+        }
+        throw new Error("Device keys not found");
+    }
+
```
richvdh, line 174: Please could you move this underneath the `CryptoApi implementation` comment (line 375).

src/rust-crypto/rust-crypto.ts
```diff
@@ -157,6 +157,21 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
         this.checkKeyBackupAndEnable();
     }
 
+    public async getOwnDeviceKeys(): Promise<OwnDeviceKeys> {
```
richvdh, line 160:
```suggestion
    /**
     * Implementation of {@link CryptoApi#getOwnDeviceKeys}.
     */
    public async getOwnDeviceKeys(): Promise<OwnDeviceKeys> {
```

richvdh (commented):
Looks good, but quite a few nits

richvdh:
coverage check is complaining. Is it possible to test the error case in RustCrypto?

src/client.ts
```diff
@@ -2400,6 +2403,8 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      *
      * @returns base64-encoded curve25519 key. Null if crypto is
      *    disabled.
+     *
+     * @deprecated Use {@link CryptoApi.getOwnDeviceKeys}
```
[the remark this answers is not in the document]
BillCarsonFr: I hesitated but the problem is that it's not working for rust crypto. So I thought "Use" was better

BillCarsonFr:
> coverage check is complaining. Is it possible to test the error case in RustCrypto?

Don't see how to do it. Would have to simulate when the ed and curve key are not defined. It would be either that there was a crypto store error or that the current device doesn't support ed25519 and curve25519

PR #3900, opened 2023-11-17
Better fallback for the event `localTimestamp` calculation.

toger5:
State events from a couple of home servers dont return an `age` proptery in their unsigned field. (including element.io).
For an actual exampe, this room: https://matrix.to/#/#thisweekinmatrix:matrix.org can be checked (lots of member events from other homeservers are not featuring a unsigned.age field): 
Screenshot 2023-11-17 at 19 36 16

Call state events are relying on this property to determin if they are expired or not.

Before a non existent age property in the unsigned field resulted in falling back to 0. The `localTimestamp` was then `Date.now()` at the time of loading this event. Which made it always really recent -> never expired.

## Checklist

---

## ✨ Features
 * Better fallback for the event `localTimestamp` calculation. ([\#3900](https://github.com/matrix-org/matrix-js-sdk/pull/3900)). Contributed by @toger5.

### DIFF
--- a/spec/unit/matrixrtc/MatrixRTCSession.spec.ts
+++ b/spec/unit/matrixrtc/MatrixRTCSession.spec.ts
@@ -67,7 +67,7 @@ describe("MatrixRTCSession", () => {
         const expiredMembership = Object.assign({}, membershipTemplate);
         expiredMembership.expires = 1000;
         expiredMembership.device_id = "EXPIRED";
-        const mockRoom = makeMockRoom([membershipTemplate, expiredMembership], () => 10000);
+        const mockRoom = makeMockRoom([membershipTemplate, expiredMembership], 10000);
 
         sess = MatrixRTCSession.roomSessionForRoom(client, mockRoom);
         expect(sess?.memberships.length).toEqual(1);
@@ -266,7 +266,7 @@ describe("MatrixRTCSession", () => {
                 const timeElapsed = 60 * 60 * 1000 - 1000;
                 mockRoom.getLiveTimeline().getState(EventTimeline.FORWARDS)!.getStateEvents = jest
                     .fn()
-                    .mockReturnValue(mockRTCEvent(eventContent.memberships, mockRoom.roomId, () => timeElapsed));
+                    .mockReturnValue(mockRTCEvent(eventContent.memberships, mockRoom.roomId, timeElapsed));
 
                 const eventReSentPromise = new Promise<Record<string, any>>((r) => {
                     resolveFn = (_roomId: string, _type: string, val: Record<string, any>) => {
@@ -539,10 +539,8 @@ describe("MatrixRTCSession", () => {
     it("emits an event at the time a membership event expires", () => {
         jest.useFakeTimers();
         try {
-            let eventAge = 0;
-
             const membership = Object.assign({}, membershipTemplate);
-            const mockRoom = makeMockRoom([membership], () => eventAge);
+            const mockRoom = makeMockRoom([membership], 0);
 
             sess = MatrixRTCSession.roomSessionForRoom(client, mockRoom);
             const membershipObject = sess.memberships[0];
@@ -550,7 +548,6 @@ describe("MatrixRTCSession", () => {
             const onMembershipsChanged = jest.fn();
             sess.on(MatrixRTCSessionEvent.MembershipsChanged, onMembershipsChanged);
 
-            eventAge = 61 * 1000 * 1000;
             jest.advanceTimersByTime(61 * 1000 * 1000);
 
             expect(onMembershipsChanged).toHaveBeenCalledWith([membershipObject], []);
@@ -561,47 +558,49 @@ describe("MatrixRTCSession", () => {
     });
 
     it("prunes expired memberships on update", () => {
-        client.sendStateEvent = jest.fn();
+        jest.useFakeTimers();
+        try {
+            client.sendStateEvent = jest.fn();
 
-        let eventAge = 0;
-
-        const mockRoom = makeMockRoom(
-            [
+            const mockMemberships = [
                 Object.assign({}, membershipTemplate, {
                     device_id: "OTHERDEVICE",
                     expires: 1000,
                 }),
-            ],
-            () => eventAge,
-        );
-        sess = MatrixRTCSession.roomSessionForRoom(client, mockRoom);
+            ];
+            const mockRoomNoExpired = makeMockRoom(mockMemberships, 0);
 
-        // sanity check
-        expect(sess.memberships).toHaveLength(1);
-        expect(sess.memberships[0].deviceId).toEqual("OTHERDEVICE");
+            sess = MatrixRTCSession.roomSessionForRoom(client, mockRoomNoExpired);
 
-        eventAge = 10000;
+            // sanity check
+            expect(sess.memberships).toHaveLength(1);
+            expect(sess.memberships[0].deviceId).toEqual("OTHERDEVICE");
 
-        sess.joinRoomSession([mockFocus]);
+            jest.advanceTimersByTime(10000);
 
-        expect(client.sendStateEvent).toHaveBeenCalledWith(
-            mockRoom!.roomId,
-            EventType.GroupCallMemberPrefix,
-            {
-                memberships: [
-                    {
-                        application: "m.call",
-                        scope: "m.room",
-                        call_id: "",
-                        device_id: "AAAAAAA",
-                        expires: 3600000,
-                        foci_active: [mockFocus],
-                        membershipID: expect.stringMatching(".*"),
-                    },
-                ],
-            },
-            "@alice:example.org",
-        );
+            sess.joinRoomSession([mockFocus]);
+
+            expect(client.sendStateEvent).toHaveBeenCalledWith(
+                mockRoomNoExpired!.roomId,
+                EventType.GroupCallMemberPrefix,
+                {
+                    memberships: [
+                        {
+                            application: "m.call",
+                            scope: "m.room",
+                            call_id: "",
+                            device_id: "AAAAAAA",
+                            expires: 3600000,
+                            foci_active: [mockFocus],
+                            membershipID: expect.stringMatching(".*"),
+                        },
+                    ],
+                },
+                "@alice:example.org",
+            );
+        } finally {
+            jest.useRealTimers();
+        }
     });
 
     it("fills in created_ts for other memberships on update", () => {
--- a/spec/unit/matrixrtc/mocks.ts
+++ b/spec/unit/matrixrtc/mocks.ts
@@ -20,49 +20,27 @@ import { randomString } from "../../../src/randomstring";
 
 export function makeMockRoom(memberships: CallMembershipData[], localAge: number | null = null): Room {
     const roomId = randomString(8);
-    // If makeMockRoom is used instead of one of the other functions in this file,
-    // we store the localTimestamp so that subsequent calls calculate the expiration based on the localTimestamp
-    // using Date.now(). So, jest.advanceTimersByTime() can be used to simulate the passage of time.
-
-    // The actual time of the underlying mocked events will be: localTimestamp ?? localAge ?? 0
-    const localTimestamp = Date.now() - (localAge ?? 10);
+    // Caching roomState here so it does not get recreated when calling `getLiveTimeline.getState()`
+    const roomState = makeMockRoomState(memberships, roomId, localAge);
     return {
         roomId: roomId,
         getLiveTimeline: jest.fn().mockReturnValue({
-            // When setting up `makeMockRoomState` we always want to use the localTimestamp instead of localAge.
-            getState: jest.fn().mockReturnValue(makeMockRoomState(memberships, roomId, null, localTimestamp)),
+            getState: jest.fn().mockReturnValue(roomState),
         }),
     } as unknown as Room;
 }
 
-export function makeMockRoomState(
-    memberships: CallMembershipData[],
-    roomId: string,
-    /**`localAge` is ignored if localTimestamp is present.*/
-    localAge: number | null = null,
-    localTimestamp: number | null = null,
-) {
+export function makeMockRoomState(memberships: CallMembershipData[], roomId: string, localAge: number | null = null) {
+    const event = mockRTCEvent(memberships, roomId, localAge);
     return {
         getStateEvents: (_: string, stateKey: string) => {
-            // If localTimestamp is present mockRTCEvent does not recompute
-            // the localTimestamp based on Date.now() when `getStateEvents` is called.
-            const event = mockRTCEvent(memberships, roomId, localAge, localTimestamp);
-
             if (stateKey !== undefined) return event;
             return [event];
         },
     };
 }
 
-export function mockRTCEvent(
-    memberships: CallMembershipData[],
-    roomId: string,
-    /**`localAge` is ignored if localTimestamp is present.*/
-    localAge: number | null,
-    localTimestamp: number | null = null,
-): MatrixEvent {
-    const _localTimestamp = localTimestamp ?? Date.now() - (localAge ?? 10);
-
+export function mockRTCEvent(memberships: CallMembershipData[], roomId: string, localAge: number | null): MatrixEvent {
     return {
         getType: jest.fn().mockReturnValue(EventType.GroupCallMemberPrefix),
         getContent: jest.fn().mockReturnValue({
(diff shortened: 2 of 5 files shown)


src/models/event.ts
```diff
@@ -404,7 +404,8 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
         });
 
         this.txnId = event.txn_id;
-        this.localTimestamp = Date.now() - (this.getAge() ?? 0);
+        const age = this.getAge();
+        this.localTimestamp = age ? Date.now() - age : this.getTs() ?? 0;
```
richvdh, line 408: * 0 doesn't seem like a sensible default for "localTimestamp".
* Please comment this: what is it doing, and why?

src/matrixrtc/CallMembership.ts
```diff
@@ -91,7 +91,7 @@ export class CallMembership {
     }
 
     public isExpired(): boolean {
-        return this.getAbsoluteExpiry() < this.parentEvent.getTs() + this.parentEvent.getLocalAge();
+        return this.getMsUntilExpiry() <= 0;
```
richvdh, line 94: this seems unrelated to the subject of this PR "Better `localTimestamp` fallback"? It looks like its switching the expiry calculation from being based on `origin_server_ts` to being based on `age`?

Which might be a reasonable thing to do, but it's not what the PR says it does?

spec/unit/matrixrtc/mocks.ts
```diff
@@ -60,8 +60,7 @@ export function mockRTCEvent(
         }),
         getSender: jest.fn().mockReturnValue("@mock:user.example"),
         getTs: jest.fn().mockReturnValue(1000),
-        getLocalAge: getLocalAgeFn,
-        localTimestamp: Date.now(),
+        localTimestamp: Date.now() - getLocalAgeFn(),
```
richvdh, line 63: why not pass in `localAge`, rather than a function `getLocalAge` which is evaluated immediately?

richvdh (changes requested):
* Please could you update the subject of the PR to better explain what is going on. Better than what? What even is a `localTimestamp`? Basically, I shouldn't have to read the diff to get a rough idea of what a PR does.

* You say this "fixes issues with call expiration". Are these issues tracked anywhere? If so, please link them. If not, please describe them. Basically, I should know what it is you're trying to fix.

* I continue to be very surprised that you are actually seeing call events that lack an `age`. Please could you share a `/sync` result or somesuch that demonstrates this?

spec/unit/matrixrtc/mocks.ts
toger5: This would mean chaning a a couple of tests whcih use this mock fn. But this is also makes sense to me. its just a bigger refactor.

toger5: I refactored it to use a property instead of a function. It is not possible anymore to update the expiration time because the localTImestamp is now a fixed value.
So we would need to make sure that Date.now() is not called when "remocking" the rtc event which happens when`getStateEvents` is called.
Therefore the the local Timestamp is already computed in `makeMockRoom` which is only called once per test and not indirectly by the code we want to test and is passed over to `getStateEvents`.

This is much better anyways since we sometimes rely on the `localTimestamp` and sometimes on the `getLocalAge` fn (but `getLocalAge` directly relies on the `localTimestamp` so forgetting to mock both in sync would always have lead to issues. Now mocking the `localTimestamp` give is indirectly mocking all other age related methods as well.

src/matrixrtc/CallMembership.ts
toger5: It really is the exact same thing just a nicer way to write it. (so it is just a cosmetic change)
(except the `<` -> `<=` which is not really chanign anything considering the epxiration is more or less arbitrary and could very well be 35999 instead of 36000)

Let me take the time to show the equivalence:
```
getMsUntilExpiry() <= 0

using:
    public getMsUntilExpiry(): number {
        return this.getLocalExpiry() - Date.now();
    }

getLocalExpiry() - Date.now() <= 0

using:
    const relativeCreationTime = this.parentEvent.getTs() - this.createdTs();
    const localCreationTs = this.parentEvent.localTimestamp - relativeCreationTime;
    return localCreationTs + this.data.expires;
}

localCreationTs + data.expires - Date.now() <= 0
parentEvent.localTimestamp - relativeCreationTime + data.expires - Date.now() <= 0
parentEvent.localTimestamp - (parentEvent.getTs() - createdTs()) + data.expires - Date.now() <= 0
parentEvent.localTimestamp - parentEvent.getTs() + createdTs() + data.expires - Date.now() <= 0
createdTs() + data.expires <= + parentEvent.getTs() + Date.now() - parentEvent.localTimestamp

using:
    public getAbsoluteExpiry(): number {
        return this.createdTs() + this.data.expires;
    }

getAbsoluteExpiry()  <=  + parentEvent.getTs() + (Date.now() - parentEvent.localTimestamp)

using:
    public getLocalAge(): number {
        return Date.now() - this.localTimestamp;
    }

getAbsoluteExpiry() <= parentEvent.getTs() + parentEvent.getLocalAge();
```

toger5:
> * Please could you update the subject of the PR to better explain what is going on. Better than what? What even is a `localTimestamp`? Basically, I shouldn't have to read the diff to get a rough idea of what a PR does.

I searched and found a specific example.

>     * You say this "fixes issues with call expiration". Are these issues tracked anywhere? If so, please link them. If not, please describe them. Basically, I should know what it is you're trying to fix.

Not officially tracked on GH but there are reports in various matrix rooms.

>     * I continue to be very surprised that you are actually seeing call events that lack an `age`. Please could you share a `/sync` result or somesuch that demonstrates this?

This seemed to be the case for all state events whcih are received from another homeserver. (All is hard to prove but most of the federated ones I have checked dont have a `unsigned.age` field)

richvdh:
> > ```
> > * I continue to be very surprised that you are actually seeing call events that lack an `age`. Please could you share a `/sync` result or somesuch that demonstrates this?
> > ```
> 
> This seemed to be the case for all state events whcih are received from another homeserver. (All is hard to prove but most of the federated ones I have checked dont have a `unsigned.age` field)
> 
> Edit: I did some more trying and sometimes federated state events also have an age field.

Ah right, it affects *federated* events. In that case it sounds like a likely effect of https://github.com/matrix-org/synapse/issues/8429. Why don't we just fix that?

fkwp:
> > > ```
> > > * I continue to be very surprised that you are actually seeing call events that lack an `age`. Please could you share a `/sync` result or somesuch that demonstrates this?
> > > ```
> > 
> > 
> > This seemed to be the case for all state events whcih are received from another homeserver. (All is hard to prove but most of the federated ones I have checked dont have a `unsigned.age` field)
> > Edit: I did some more trying and sometimes federated state events also have an age field.
> 
> Ah right, it affects _federated_ events. In that case it sounds like a likely effect of [matrix-org/synapse#8429](https://github.com/matrix-org/synapse/issues/8429). Why don't we just fix that?

Maybe a robust design is favorable in case you are using a homeserver with that specific "bug".

toger5:
Thanks for the reply. Fixing it in synapse definitly would be the more desirable solution. I still advocat having a frontend fallback solution. Otherwise we would rely on all homeserver maintainers to update to the patched version and until then calling in EW looks broken. (It would always show ongoing calls for federated event if ppl dont properly close their call.)

Since we cannot guarantee a date until which all homeserver maintainers have patched their synapse this fallback might be helpful for a long enough time to make it worthit.

If someone can make time to fix the synapse issue as well that would be very appreciated. (I dont know if this fits as a task for me for our current goals.)

richvdh:
> > Ah right, it affects _federated_ events. In that case it sounds like a likely effect of [matrix-org/synapse#8429](https://github.com/matrix-org/synapse/issues/8429). Why don't we just fix that?
> 
> Maybe a robust design is favorable in case you are using a homeserver with that specific "bug".

Yeah I don't object to fixing both sides of it (indeed one will probably have to since it will take a while for a Synapse fix to roll out). But just fixing the client side does feel a lot like curing the symptoms rather than the disease. Having an entire stack which is full of workarounds for bugs in other parts of the stack does not lead to a stable, maintainable system.

src/matrixrtc/CallMembership.ts
richvdh: Ah yes that seems right! Thanks for taking the time to explain so clearly.

src/models/event.ts
```diff
@@ -404,8 +404,12 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
         });
 
         this.txnId = event.txn_id;
+        // The localTimestamp is calculated using the age.
+        // Some homeserver don't provide the unsigned field for state events.
```
richvdh, line 408:
```suggestion
        // Some events lack an `age` property, either because they are EDUs such as typing events,
        // or due to server-side bugs such as https://github.com/matrix-org/synapse/issues/8429.
```

src/models/event.ts
```diff
@@ -404,8 +404,12 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
...
+        // The fallback in these cases will be to use the origin_server_ts.
+        // For EDUs, the origin_server_ts also is not defined so we use Date.now().
         const age = this.getAge();
-        this.localTimestamp = age ? Date.now() - age : this.getTs() ?? 0;
+        this.localTimestamp = age ? Date.now() - age : this.getTs() ?? Date.now();
```
richvdh, line 412: we should probably do:
```suggestion
        this.localTimestamp = age !== undefined ? Date.now() - age : this.getTs() ?? Date.now();
```
 to guard against the unlikely but possible case that `age` is zero.

spec/unit/matrixrtc/mocks.ts
```diff
@@ -49,9 +53,10 @@ export function makeMockRoomState(
 export function mockRTCEvent(
     memberships: CallMembershipData[],
     roomId: string,
-    getLocalAge: (() => number) | undefined,
+    localAge: number | null,
+    localTimestamp: number | null = null,
```
richvdh, line 57: could you add a doc-comment here and to `makeMockRoomState` documenting how these two interact? (My understanding is that `localAge` is ignored if `localTimestamp` is non-null? But if you could make that explicit that would be great).

spec/unit/matrixrtc/mocks.ts
```diff
@@ -18,27 +18,31 @@ import { EventType, MatrixEvent, Room } from "../../../src";
...
-    memberships: CallMembershipData[],
-    getLocalAge: (() => number) | undefined = undefined,
-): Room {
+export function makeMockRoom(memberships: CallMembershipData[], localAge: number | null = null): Room {
     const roomId = randomString(8);
+    // If makeMockRoom is used instead of one of the other functions in this file,
+    // we store it so that subsequent calls calculate the expiration based on it using
+    // Date.now(). So, jest.advanceTimersByTime() can be used to simulate the passage of time.
+
+    // The actual time of the underlying mocked events will be: localTimestamp ?? localAge ?? 0
```
richvdh, line 27: > we store it

what is "it" in this comment? `makeMockRoom`?

I'd suggest rewording this whole comment for clarity.

```suggestion
    // Replace the `localAge` with a `localTimestamp`, so that `jest.advanceTimersByTime()` can be used
    // to simulate the passage of time and test expiry.
```

spec/unit/matrixrtc/mocks.ts
```diff
@@ -18,27 +18,31 @@ import { EventType, MatrixEvent, Room } from "../../../src";
...
+    const localTimestamp = Date.now() - (localAge ?? 10);
```
richvdh, line 28: isn't this exactly what happens in `makeMockRoomState` if you just pass in a `localAge`? Can we omit this line? (And maybe move the comment above to `mockRTCEvent`)

spec/unit/matrixrtc/MatrixRTCSession.spec.ts
```diff
@@ -561,47 +558,49 @@ describe("MatrixRTCSession", () => {
...
+                            foci_active: [mockFocus],
+                            membershipID: expect.stringMatching(".*"),
+                        },
+                    ],
+                },
+                "@alice:example.org",
+            );
+        } finally {
+            jest.useRealTimers();
+        }
```
richvdh, line 603: makes sense to do this here for consistency with the existing tests, but for future reference an `afterEach` function is probably a nicer way to do it.

spec/unit/matrixrtc/mocks.ts
richvdh: yes this is all much nicer. Thanks for taking the time to do this!

spec/unit/matrixrtc/mocks.ts
toger5: This is the core part of it. We need to cache the `localTimestamp` so that subsequent calls to `getLiveTimeline` (which then calls: `makeMockRoomState`) use this `localTimestamp` instead of the `localAge` (since `localAge` is ignored if `localTimestamp` is the present as described in the comment.)

Otherwise we would call `Date.now() - localAge` at the time where we call `getLiveTimeline` so we could of course not use the initial age but need to update it with the elapsed time. But when calling `getLiveTimeline` we cannot access the age anymore since age is not a method anymore.

By cahing the localTimestamp here we only use the timestamp when calling  `getLiveTimeline` and compute the age based on that so it is all in sync.

richvdh: > `makeMockRoomState` can be called indirectly in the tested code when the mocked `getLiveTimeline` is called. If we mock it with a `localAge` than the timestamp would be recomputed using `Date.now()` - `localAge`. So we would recompute the local timestamp "accidently" while calling `getLiveTimeline` which breaks a lot of tests (It would be a bug). Caching it here and then alsways using `localTimestamp` allows to always recompute the local age based on the timestamp. So both are correct.

I'm afraid I'm not following you. What do you mean by "if we mock it" - mock *what*?

Currently your code is:

```typescript
    const localTimestamp = Date.now() - (localAge ?? 10);
    return {
        roomId: roomId,
        getLiveTimeline: jest.fn().mockReturnValue({
            getState: jest.fn().mockReturnValue(makeMockRoomState(memberships, roomId, localAge, localTimestamp)),
        }),
    } as unknown as Room;
```

... which is exactly equivalent to:

```typescript
    const localTimestamp = Date.now() - (localAge ?? 10);
    const mockedRoomState = makeMockRoomState(memberships, roomId, localAge, localTimestamp);
    return {
        roomId: roomId,
        getLiveTimeline: jest.fn().mockReturnValue({
            getState: jest.fn().mockReturnValue(mockedRoomState),
        }),
    } as unknown as Room;
```

... which again is equivalent to

```typescript
    const mockedRoomState = makeMockRoomState(memberships, roomId, localAge,  Date.now() - (localAge ?? 10);
    return {
        roomId: roomId,
        getLiveTimeline: jest.fn().mockReturnValue({
            getState: jest.fn().mockReturnValue(mockedRoomState),
        }),
    } as unknown as Room;
```

... and I don't understand why specifying an explicit value for `makeMockRoomState`'s fourth argument is any different to letting it use its default.

toger5: > ... which is exactly equivalent to:

It is not equivalent since the initial version will call makeMockRoomState again while your proposed one wont.
This is exactly what I treid to achieve with the localTimestamp prop. Because when calling `makeMockRoomState` (like it initially did) localTimestamp gets recomputed with `Date.now() - (localAge ?? 10)` using the current `Date.now()`.

But just storing the whole mockedRoomState is muuch easier to read!
See this change: https://github.com/matrix-org/matrix-js-sdk/pull/3900/commits/54b71bb729090a7534a54bd6dbfdc3e3d738d66a

(I think this might be exactly what you tried to communicate. Thanks. I removed a couple of comments since the things the comments addressed are quiet obvious now.)

spec/unit/matrixrtc/mocks.ts
toger5: I removed the comment. With the last [commit](https://github.com/matrix-org/matrix-js-sdk/pull/3900/commits/54b71bb729090a7534a54bd6dbfdc3e3d738d66a), we store the whole return value outside the mocked method so it wont change anyways. (and its quiet clear even without comment)

spec/unit/matrixrtc/mocks.ts
richvdh: looks like this is no longer relevant because the functions in question now only have a `localAge` (rather than a `localTimestamp`. I wouldn't have minded seeing a doc-comment explaining what `localAge` does anyway, but it's not too important

richvdh (approved):
lgtm!

spec/unit/matrixrtc/mocks.ts
richvdh: For the record: it turns out that `makeMockRoomState` didn't immediately call `mockRTCEvent`, but instead delayed it until `RoomState.getStateEvents` was called. Hence, passing a `localAge` into `makeMockRoomState` would produce "unintuitive" effects. Sorry, I should have spotted this during earlier review.

`makeMockRoomState` is updated as of 42be793a5 so this is resolved.

robintown (approved):

PR #3902, opened 2023-11-20
Only await key query after lazy members resolved

BillCarsonFr:
Quick improvment on send message performance.

Previously we were waiting on all `key/query` to be processed before sharing the room keys (to ensure having the most up to date device list).

It's a bit too aggressive, and in some case was impacting time to send for no benefits.

In this change we will only wait for `key/query` if the list of member (lazy loading) was just resolved in this send cycle. Because in this case several users might be added to the tracked user list and there is a high risk that their list of devices won't be known yet. And as they might be in the room since long it would be bad that they don't get the key.

If the member list is not just resolved, we don't want to wait on all `key/query`, we just trigger a launch in background. Freshly joined users will get the key if the request is fast enough. If not they will get it for the next message.

## Checklist

---

### DIFF
--- a/src/rust-crypto/RoomEncryptor.ts
+++ b/src/rust-crypto/RoomEncryptor.ts
@@ -129,16 +129,25 @@ export class RoomEncryptor {
             await this.olmMachine.updateTrackedUsers(members.map((u) => new RustSdkCryptoJs.UserId(u.userId)));
             this.lazyLoadedMembersResolved = true;
             this.prefixedLogger.debug(`Updated tracked users for room ${this.room.roomId}`);
-        }
 
-        // Query keys in case we don't have them for newly tracked members.
-        // This must be done before ensuring sessions. If not the devices of these users are not
-        // known yet and will not get the room key.
-        // We don't have API to only get the keys queries related to this member list, so we just
-        // process the pending requests from the olmMachine. (usually these are processed
-        // at the end of the sync, but we can't wait for that).
-        // XXX future improvement process only KeysQueryRequests for the tracked users.
-        await this.outgoingRequestManager.doProcessOutgoingRequests();
+            // Query keys in case we don't have them for newly tracked members.
+            // It's important after loading members for the first time, as likely most of them won't be
+            // known yet and will be unable to decrypt messages despite being in the room for long.
+            // This must be done before ensuring sessions. If not the devices of these users are not
+            // known yet and will not get the room key.
+            // We don't have API to only get the keys queries related to this member list, so we just
+            // process the pending requests from the olmMachine. (usually these are processed
+            // at the end of the sync, but we can't wait for that).
+            // XXX future improvement process only KeysQueryRequests for the users that have never been queried.
+            await this.outgoingRequestManager.doProcessOutgoingRequests();
+        } else {
+            // if members are already loaded it's less critical to await on key queries.
+            // We might still want to trigger a processOutgoingRequests here, if the
+            // query is done timely enough, the freshly tracked users will get the room key.
+            this.outgoingRequestManager.doProcessOutgoingRequests().then(() => {
+                // nop
+            });
+        }
 
         this.prefixedLogger.debug(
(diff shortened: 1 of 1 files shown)


src/rust-crypto/RoomEncryptor.ts
```diff
@@ -142,19 +142,29 @@ export class RoomEncryptor {
...
+            logger.debug(`Processing outgoing requests`);
+            await this.outgoingRequestManager.doProcessOutgoingRequests();
+        } else {
+            // if members are already loaded it's less critical to await on key queries.
+            // We might still want to trigger a processOutgoingRequests here, if the
+            // query is done timely enough, the freshly tracked users will get the room key.
+            logger.debug(`Processing outgoing requests in background`);
+            this.outgoingRequestManager.doProcessOutgoingRequests().then(() => {
+                // nop
+            });
```
richvdh, line 166: this `then` doesn't seem to be doing anything?

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -142,19 +142,29 @@ export class RoomEncryptor {
...
+            // We don't have API to only get the keys queries related to this member list, so we just
+            // process the pending requests from the olmMachine. (usually these are processed
+            // at the end of the sync, but we can't wait for that).
+            // XXX future improvement process only KeysQueryRequests for the users that have never been queried.
+            logger.debug(`Processing outgoing requests`);
+            await this.outgoingRequestManager.doProcessOutgoingRequests();
+        } else {
+            // if members are already loaded it's less critical to await on key queries.
+            // We might still want to trigger a processOutgoingRequests here, if the
+            // query is done timely enough, the freshly tracked users will get the room key.
```
richvdh, line 162: I'm afraid I don't understand this comment?

src/rust-crypto/RoomEncryptor.ts
BillCarsonFr: Just to signal explicitly that it's async and that we don't await (also removes a warning). What would you recommand for that case?

src/rust-crypto/RoomEncryptor.ts
BillCarsonFr: tried to clarify

src/rust-crypto/RoomEncryptor.ts
richvdh: TBH I'm surprised it removes the warning, though I do see the same thing in WebStorm.

I'd probably just add a comment in this case. Another option is it add a `catch` block which logs any exceptions that get thrown.

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -142,19 +142,30 @@ export class RoomEncryptor {
...
+            // We don't have API to only get the keys queries related to this member list, so we just
+            // process the pending requests from the olmMachine. (usually these are processed
+            // at the end of the sync, but we can't wait for that).
+            // XXX future improvement process only KeysQueryRequests for the users that have never been queried.
+            logger.debug(`Processing outgoing requests`);
+            await this.outgoingRequestManager.doProcessOutgoingRequests();
+        } else {
+            // If members are already loaded it's less critical to await on key queries.
+            // We might still want to trigger a processOutgoingRequests here.
+            // The next call to ensureSessions will wait a bit on in flight key queries we are
```
richvdh, line 162:
```suggestion
            // The call to `ensureSessionsForUsers` below will wait a bit on in-flight key queries we are
```

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -142,19 +142,30 @@ export class RoomEncryptor {
...
+            // interested in. If a sync handling happens in the meantime, and some new members are added to the room
+            // or have new devices it would give us a chance to query them before sending.
```
richvdh, line 164: I'm not entirely convinced about the logic of this (if there is a `/sync` response, we'll process the outgoing requests there; and if there isn't a `/sync` response then there won't be any new members to query?) but let's just ship this anyway

richvdh (approved):
small nit but lgtm otherwise

PR #3905, opened 2023-11-22
Set up CI to lint workflows with action-validator

Johennes:
This is the equivalent of https://github.com/vector-im/element-web/pull/26621 applied to this repository

This, unfortunately, required renaming the `release-action` workflow because its name threw off the validator.

For: https://github.com/matrix-org/matrix-react-sdk/pull/11930
For: https://github.com/vector-im/element-web/pull/26641
For: https://github.com/vector-im/element-desktop/pull/1347

---

### DIFF
--- a/.github/workflows/release-make.yml
+++ b/.github/workflows/release-make.yml
@@ -1,4 +1,4 @@
-name: Release Action
+name: Release Make
 on:
     workflow_call:
         secrets:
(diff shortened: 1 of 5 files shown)


t3chguy:
> could we safely rename and do you have any suggestions for a better name?

We could but it'd have to be replicated onto the 3 repos calling this workflow

Johennes:
> We could but it'd have to be replicated onto the 3 repos calling this workflow

I assume this is safe to do outside of release days? How about renaming it into `release-executor.yml`? I'm blanking on good names.

t3chguy:
It should be but beware sod's law

t3chguy (approved):

PR #3910, opened 2023-11-28
Update Sibz/github-status-action to use node16 to silence warning

t3chguy:
---

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -17,7 +17,7 @@ jobs:
         steps:
             # We create the status here and then update it to success/failure in the `report` stage
             # This provides an easy link to this workflow_run from the PR before Cypress is done.
-            - uses: Sibz/github-status-action@faaa4d96fecf273bd762985e0e7f9f933c774918 # v1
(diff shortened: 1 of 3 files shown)


.github/workflows/sonarcloud.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         steps:
             # We create the status here and then update it to success/failure in the `report` stage
             # This provides an easy link to this workflow_run from the PR before Cypress is done.
-            - uses: Sibz/github-status-action@faaa4d96fecf273bd762985e0e7f9f933c774918 # v1
+            - uses: Sibz/github-status-action@071b5370da85afbb16637d6eed8524a06bc2053e # v1
```
MidhunSureshR, line 20: nit: since we're pinning to a commit, we could probably remove the version in comments everywhere
```suggestion
            - uses: Sibz/github-status-action@071b5370da85afbb16637d6eed8524a06bc2053e
```

t3chguy: No, this is used to drive dependabot/renovate to know which series to update within

PR #3911, opened 2023-11-28
Explicitly free some Rust-side objects

richvdh:
Finalization in Javascript is not guaranteed - sometimes the FinalizationRegistry just forgets to call the finalizers on some objects. This is problematic because it means we leak references on the Rust side, which (as well as causing memory leaks) ultimately leads to us holding open the indexeddb, which then means we can't successfully log in/out.

This isn't meant to be a comprehensive fix, but it does add explicit `free` calls in a number of places which seem particularly prone to leaking, and empirically it was enough for me to get a Cypress test which logs out and back in working again.

---

### DIFF
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -871,7 +871,7 @@ describe("RustCrypto", () => {
         });
 
         it("returns a verified UserVerificationStatus when the UserIdentity is verified", async () => {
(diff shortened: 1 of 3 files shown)


t3chguy (commented):
This seems like a really awful footgun, if we wanted the malloc life we wouldn't have been TS devs - is there any way around this on the horizon?

t3chguy:
I know almost nothing about Rust, and absolutely nothing about incorporating Rust in Javascript. Equally though if switching to rust crypto means brittle as hell code I think I'm no longer in favour

t3chguy:
No arguing that

t3chguy (approved):

PR #3912, opened 2023-11-28
Rust-crypto: fix `bootstrapCrossSigning` on second call

richvdh:
Currently, `bootstrapCrossSigning` raises an exception if it is called a second time before secret storage is set up. It is easily fixed by checking that 4S is set up before trying to export to 4S.

Also a few logging fixes while we're in the area.

Fixes https://github.com/vector-im/element-web/issues/26393

---

## 🐛 Bug Fixes
 * Rust-crypto: fix `bootstrapCrossSigning` on second call ([\#3912](https://github.com/matrix-org/matrix-js-sdk/pull/3912)). Fixes vector-im/element-web#26393.

### DIFF
--- a/spec/integ/crypto/cross-signing.spec.ts
+++ b/spec/integ/crypto/cross-signing.spec.ts
@@ -236,6 +236,19 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("cross-signing (%s)", (backend: s
                 `[${TEST_USER_ID}].[${TEST_DEVICE_ID}].signatures.[${TEST_USER_ID}].[ed25519:${SELF_CROSS_SIGNING_PUBLIC_KEY_BASE64}]`,
             );
         });
+
+        it("can bootstrapCrossSigning twice", async () => {
+            mockSetupCrossSigningRequests();
+
+            const authDict = { type: "test" };
+            await bootstrapCrossSigning(authDict);
+
+            // a second call should do nothing except GET requests
+            fetchMock.mockClear();
+            await bootstrapCrossSigning(authDict);
+            const calls = fetchMock.calls((url, opts) => opts.method != "GET");
+            expect(calls.length).toEqual(0);
+        });
     });
 
     describe("getCrossSigningStatus()", () => {
--- a/spec/integ/crypto/cross-signing.spec.ts
+++ b/spec/integ/crypto/cross-signing.spec.ts
@@ -34,6 +34,7 @@ import {
     USER_CROSS_SIGNING_PRIVATE_KEY_BASE64,
 } from "../../test-utils/test-data";
 import { E2EKeyResponder } from "../../test-utils/E2EKeyResponder";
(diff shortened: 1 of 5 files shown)


src/rust-crypto/CrossSigningIdentity.ts
```diff
@@ -66,18 +66,25 @@ export class CrossSigningIdentity {
         });
 
         if (olmDeviceHasKeys) {
-            if (!privateKeysInSecretStorage) {
+            if (!(await this.secretStorage.hasKey())) {
```
richvdh, line 69: this is the real fix in this PR: we need to check that secretStorage is set up before we try to export to it.

src/rust-crypto/CrossSigningIdentity.ts
```diff
@@ -148,17 +162,17 @@ export class CrossSigningIdentity {
...
+            await this.secretStorage.store("m.cross_signing.self_signing", exported.self_signing_key);
         } else {
             logger.error(`Cannot export SSK to secret storage, private key unknown`);
         }
         if (exported?.userSigningKey) {
-            this.secretStorage.store("m.cross_signing.user_signing", exported.userSigningKey);
+            await this.secretStorage.store("m.cross_signing.user_signing", exported.userSigningKey);
         } else {
             logger.error(`Cannot export USK to secret storage, private key unknown`);
         }
```
richvdh, line 178: these things were missing their `await` keywords, which meant that when they failed they were shouting into the void.

richvdh:
will add some more tests

spec/integ/crypto/cross-signing.spec.ts
```diff
@@ -236,6 +237,53 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("cross-signing (%s)", (backend: s
...
             );
         });
+
+        it("can bootstrapCrossSigning twice", async () => {
+            mockSetupCrossSigningRequests();
+
+            const authDict = { type: "test" };
+            await bootstrapCrossSigning(authDict);
+
+            // a second call should do nothing except GET requests
```
BillCarsonFr, line 247: What get requests?

BillCarsonFr (approved):
Good catch

spec/integ/crypto/cross-signing.spec.ts
richvdh: It does `GET /user/:user/account_data/:type` requests, because this test is running before the initialsync completes (so it has not cached any account data in memory).

Would you find this useful to include in the comment?

BillCarsonFr: No maybe not. Was trying to remember why it would do requests.
Looks like it's checking if it's in 4S or not to eventually update 4S or import. There are some comments about it in bootstrapCrossSigning. I guess the naming is not perfect, maybe it's ensureCrossSigning but also fix it :/

MidhunSureshR (approved):

PR #3913, opened 2023-11-28
Add guards against `MatrixClient.stopClient` calls

richvdh:
If we call methods on `OlmMachine` after `MatrixClient.stopClient` is called, we will end up with a "use of moved value" error. We can turn these into something more useful with judicious use of `getOlmMachineOrThrow`.

Alternatively, we can sidestep the issue by bailing out sooner.


andybalaam (approved):
Looks good thanks.

PR #3918, opened 2023-11-28
Fix upload with empty Content-Type

### LINKED ISSUE #3917: Empty Content-Type when uploading file without suffix
If you try to upload file without suffix, empty `Content-Type` header is send by Element client. Empty `Content-Type` can be blocked by some web application filters.

The reason is this line:

https://github.com/matrix-org/matrix-js-sdk/blob/c49a527e5ee24c3ba1216c464653ef3ec05812c4/src/http-api/index.ts#L53

It expects that if file content type is unknown, default MIME type `application/octet-stream` should be send. But according to specification, for unknown file, `File.type` is empty string (see https://developer.mozilla.org/en-US/docs/Web/API/Blob/type). And `??` operator consider empty string as valid value.

So instead `??` operator, `||` should be used.

JakubOnderka:
Fixes #3917

## Checklist

---

## 🐛 Bug Fixes
 * Fix upload with empty Content-Type ([\#3918](https://github.com/matrix-org/matrix-js-sdk/pull/3918)). Fixes undefined/matrix-js-sdk#3917. Contributed by @JakubOnderka.

### DIFF
--- a/src/http-api/index.ts
+++ b/src/http-api/index.ts
@@ -50,7 +50,7 @@ export class MatrixHttpApi<O extends IHttpOpts> extends FetchHttpApi<O> {
         const abortController = opts.abortController ?? new AbortController();
 
         // If the file doesn't have a mime type, use a default since the HS errors if we don't supply one.
-        const contentType = opts.type ?? (file as File).type ?? "application/octet-stream";
+        const contentType = (opts.type ?? (file as File).type) || "application/octet-stream";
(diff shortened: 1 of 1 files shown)


florianduros:
First of all, thanks for your contribution.

We need you to sign-off your contribution according to the [contributing guide](https://github.com/vector-im/element-web/blob/develop/CONTRIBUTING.md#sign-off).

florianduros (changes requested):
See previous comment

PR #3928, opened 2023-11-29
Signal key backup in cache

BillCarsonFr:
Add a new crypto event to signal when the key backup decryption key is stored in cache.
Needed for new PRs around backup (would allow to start/resume downloading from a backup)

## Checklist

---

### DIFF
--- a/spec/integ/crypto/verification.spec.ts
+++ b/spec/integ/crypto/verification.spec.ts
@@ -1259,14 +1259,15 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
 
             const requestId = await requestPromises.get("m.megolm_backup.v1");
 
+            const keyBackupIsCached = new Promise<void>((resolve) => {
+                aliceClient.on(CryptoEvent.KeyBackupPrivateKeyCached, () => {
+                    resolve();
+                });
+            });
+
             await sendBackupGossipAndExpectVersion(requestId!, BACKUP_DECRYPTION_KEY_BASE64, matchingBackupInfo);
 
-            // We are lacking a way to signal that the secret has been received, so we wait a bit..
-            jest.useRealTimers();
-            await new Promise((resolve) => {
-                setTimeout(resolve, 500);
-            });
-            jest.useFakeTimers();
+            await keyBackupIsCached;
 
             // the backup secret should be cached
             const cachedKey = await aliceClient.getCrypto()!.getSessionBackupPrivateKey();
--- a/spec/integ/crypto/verification.spec.ts
+++ b/spec/integ/crypto/verification.spec.ts
@@ -1260,8 +1260,10 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
             const requestId = await requestPromises.get("m.megolm_backup.v1");
 
             const keyBackupIsCached = new Promise<void>((resolve) => {
-                aliceClient.on(CryptoEvent.KeyBackupPrivateKeyCached, () => {
-                    resolve();
+                aliceClient.on(CryptoEvent.KeyBackupDecryptionKeyCached, (version) => {
+                    if (version === matchingBackupInfo.version) {
+                        resolve();
+                    }
                 });
             });
 
--- a/spec/integ/crypto/verification.spec.ts
+++ b/spec/integ/crypto/verification.spec.ts
@@ -1259,13 +1259,7 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
 
             const requestId = await requestPromises.get("m.megolm_backup.v1");
 
-            const keyBackupIsCached = new Promise<void>((resolve) => {
-                aliceClient.on(CryptoEvent.KeyBackupDecryptionKeyCached, (version) => {
-                    if (version === matchingBackupInfo.version) {
-                        resolve();
-                    }
-                });
-            });
+            const keyBackupIsCached = emitPromise(aliceClient, CryptoEvent.KeyBackupDecryptionKeyCached);
 
             await sendBackupGossipAndExpectVersion(requestId!, BACKUP_DECRYPTION_KEY_BASE64, matchingBackupInfo);
 
--- a/src/crypto/index.ts
+++ b/src/crypto/index.ts
@@ -231,14 +231,18 @@ export enum CryptoEvent {
     KeyBackupStatus = "crypto.keyBackupStatus",
     KeyBackupFailed = "crypto.keyBackupFailed",
     KeyBackupSessionsRemaining = "crypto.keyBackupSessionsRemaining",
+
     /**
      * Fires when a new valid backup decryption key is in cache.
+     * This will happen when a secret is received from another session, from secret storage,
+     * or when a new backup is created from this session
      *
-     * The payload is the `KeyBackupInfo`.
+     * The payload is the version of the backup for which we have the key for.
      *
      * This event is only fired by the rust crypto backend.
      */
-    KeyBackupPrivateKeyCached = "crypto.KeyBackupPrivateKeyCached",
+    KeyBackupDecryptionKeyCached = "crypto.keyBackupDecryptionKeyCached",
+
     KeySignatureUploadFailure = "crypto.keySignatureUploadFailure",
     /** @deprecated Use `VerificationRequestReceived`. */
     VerificationRequest = "crypto.verification.request",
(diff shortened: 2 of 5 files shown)


src/crypto/index.ts
```diff
@@ -231,6 +231,14 @@ export enum CryptoEvent {
...
     KeyBackupFailed = "crypto.keyBackupFailed",
     KeyBackupSessionsRemaining = "crypto.keyBackupSessionsRemaining",
+    /**
+     * Fires when a new valid backup decryption key is in cache.
+     *
+     * The payload is the `KeyBackupInfo`.
+     *
+     * This event is only fired by the rust crypto backend.
+     */
+    KeyBackupPrivateKeyCached = "crypto.KeyBackupPrivateKeyCached",
```
richvdh, line 241:
```suggestion
    KeyBackupPrivateKeyCached = "crypto.keyBackupPrivateKeyCached",
```

For consistency with the other events.

Also: don't we call this a "backup decryption key" rather than a "backup private key" elsewhere?

src/crypto/index.ts
```diff
@@ -231,6 +231,14 @@ export enum CryptoEvent {
     KeyBackupStatus = "crypto.keyBackupStatus",
     KeyBackupFailed = "crypto.keyBackupFailed",
     KeyBackupSessionsRemaining = "crypto.keyBackupSessionsRemaining",
+    /**
```
richvdh, line 234: Nit:

```suggestion

    /**
```

src/crypto/index.ts
```diff
@@ -231,6 +231,14 @@ export enum CryptoEvent {
...
+     * Fires when a new valid backup decryption key is in cache.
+     *
+     * The payload is the `KeyBackupInfo`.
+     *
+     * This event is only fired by the rust crypto backend.
+     */
+    KeyBackupPrivateKeyCached = "crypto.KeyBackupPrivateKeyCached",
     KeySignatureUploadFailure = "crypto.keySignatureUploadFailure",
```
richvdh, line 242: Nit:
```suggestion

    KeySignatureUploadFailure = "crypto.keySignatureUploadFailure",
```

src/crypto/index.ts
```diff
@@ -231,6 +231,14 @@ export enum CryptoEvent {
     KeyBackupStatus = "crypto.keyBackupStatus",
     KeyBackupFailed = "crypto.keyBackupFailed",
     KeyBackupSessionsRemaining = "crypto.keyBackupSessionsRemaining",
+    /**
+     * Fires when a new valid backup decryption key is in cache.
+     *
+     * The payload is the `KeyBackupInfo`.
```
richvdh, line 237:
```suggestion
     * The payload is the {@link Crypto.KeyBackupInfo}.
```

(or whatever it is to make that link work right)

src/crypto/index.ts
```diff
@@ -231,6 +231,14 @@ export enum CryptoEvent {
     KeyBackupStatus = "crypto.keyBackupStatus",
     KeyBackupFailed = "crypto.keyBackupFailed",
     KeyBackupSessionsRemaining = "crypto.keyBackupSessionsRemaining",
+    /**
+     * Fires when a new valid backup decryption key is in cache.
```
richvdh, line 235: It would be nice to write some words about when this happens. As I understand it, it happens when the key is read from secret storage, is received via gossiping, or when a brand new keypair is made (which happens when)?

src/rust-crypto/backup.ts
```diff
@@ -490,10 +493,15 @@ export class RustBackupDecryptor implements BackupDecryptor {
...
-    | CryptoEvent.KeyBackupFailed;
+    | CryptoEvent.KeyBackupFailed
+    | CryptoEvent.KeyBackupPrivateKeyCached;
 
 export type RustBackupCryptoEventMap = {
     [CryptoEvent.KeyBackupStatus]: (enabled: boolean) => void;
     [CryptoEvent.KeyBackupSessionsRemaining]: (remaining: number) => void;
     [CryptoEvent.KeyBackupFailed]: (errCode: string) => void;
+    /**
+     * Fired when the backup decryption key is received via secret sharing and stored in cache.
```
richvdh, line 504: aren't there other ways a decryption key can be received?

Also, we probably don't need to document again what the event means: it is sufficient to put the documentation on `CryptoEvent.KeyBackupPrivateKeyCached` IMHO.

src/rust-crypto/rust-crypto.ts
```diff
@@ -1827,4 +1828,9 @@ type RustCryptoEventMap = {
      * Fires when the trust status of a user changes.
      */
     [CryptoEvent.UserTrustStatusChanged]: (userId: string, userTrustLevel: UserVerificationStatus) => void;
+
+    /**
+     * Fires when the backup decryption key is received via secret sharing.
+     */
```
richvdh, line 1834: again, no need to repeat the documentation IMHO

src/rust-crypto/backup.ts
```diff
@@ -156,6 +156,9 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
             );
 
             await this.olmMachine.saveBackupDecryptionKey(backupDecryptionKey, backupCheck.backupInfo.version);
+            // Emit an event that we have a new backup decryption key, so that client can automatically
+            // start importing all keys from the backup.
```
richvdh, line 160: Isn't it up to the *library*, rather than the client, to start importing keys?

richvdh (changes requested):
looks sensible in principle.

We need to move the useful `CryptoEvent`s out of `crypto/index.ts` but that is probably a job for another day.

src/crypto/index.ts
BillCarsonFr: good point renamed to keyBackupDecryptionKeyCached

src/crypto/index.ts
BillCarsonFr: Update that, the payload is now just the backup version (changed following your remark on when it's fired)

src/crypto/index.ts
BillCarsonFr: Actually it was only when received via gossip, but it make more sense to make it more global. It's now when the secret is received from another device, from the 4S or when a new one is created (doc udpated)

src/rust-crypto/backup.ts
BillCarsonFr: yes, updated.
BTW, the current system where we re-emit things is a bit hard to maintain, we have to declare the events at several places. Not sure how to do better.

src/rust-crypto/backup.ts
BillCarsonFr: I used sdk, it makes it more obvious than library?

src/crypto/index.ts
```diff
@@ -231,14 +231,18 @@ export enum CryptoEvent {
     KeyBackupStatus = "crypto.keyBackupStatus",
     KeyBackupFailed = "crypto.keyBackupFailed",
     KeyBackupSessionsRemaining = "crypto.keyBackupSessionsRemaining",
+
     /**
      * Fires when a new valid backup decryption key is in cache.
+     * This will happen when a secret is received from another session, from secret storage,
+     * or when a new backup is created from this session
```
richvdh, line 238:
```suggestion
     * or when a new backup is created from this session.
```

spec/integ/crypto/verification.spec.ts
```diff
@@ -1259,14 +1259,17 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
 
             const requestId = await requestPromises.get("m.megolm_backup.v1");
 
+            const keyBackupIsCached = new Promise<void>((resolve) => {
+                aliceClient.on(CryptoEvent.KeyBackupDecryptionKeyCached, (version) => {
+                    if (version === matchingBackupInfo.version) {
+                        resolve();
+                    }
+                });
+            });
```
richvdh, line 1268: If you're only expecting `KeyBackupDecryptionKeyCached` to be emitted once, you can also spell this:

```typescript
            const keyBackupIsCached = emitPromise(aliceClient, CryptoEvent.KeyBackupDecryptionKeyCached);
            // ...
            const version = await keyBackupIsCached;
            expect(version).toEqual(matchingBackupInfo.version);
```

... which I think is clearer.

src/rust-crypto/backup.ts
```diff
@@ -156,6 +156,9 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
             );
 
             await this.olmMachine.saveBackupDecryptionKey(backupDecryptionKey, backupCheck.backupInfo.version);
+            // Emit an event that we have a new backup decryption key, so that the sdk can start
+            // importing keys from backup if needed.
+            this.emit(CryptoEvent.KeyBackupDecryptionKeyCached, backupCheck.backupInfo.version);
```
richvdh, line 161: maybe we should add a `saveDecryptionKey` method to `RustBackupManager` which calls `OlmMachine.saveBackupDecryptionKey` and emits the event, for a bit less copy-pasting?

src/crypto/index.ts
```diff
@@ -296,6 +308,13 @@ export type CryptoEventHandlerMap = {
     [CryptoEvent.KeyBackupStatus]: (enabled: boolean) => void;
     [CryptoEvent.KeyBackupFailed]: (errcode: string) => void;
     [CryptoEvent.KeyBackupSessionsRemaining]: (remaining: number) => void;
+
+    /**
+     * Fires when the backup decryption key is received and cached.
+     *
+     * @param version - The version of the backup for which we have the key for.
+     */
```
richvdh, line 316: For future reference: I wouldn't have bothered repeating the doc here, but it's fine.

richvdh (approved):
A few comments, but lgtm

src/rust-crypto/backup.ts
BillCarsonFr: Yes, thought about and got lazy. Now done

PR #3929, opened 2023-11-29
Valere/element r/backup/per session download

BillCarsonFr:
## Checklist

### DIFF
--- a/spec/unit/rust-crypto/PerSessionKeyBackupDownloader.spec.ts
+++ b/spec/unit/rust-crypto/PerSessionKeyBackupDownloader.spec.ts
@@ -1,0 +1,538 @@
+/*
+Copyright 2023 The Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { Mocked } from "jest-mock";
+import * as RustSdkCryptoJs from "@matrix-org/matrix-sdk-crypto-wasm";
+
+import {
+    KeyDownloaderEvent,
+    KeyDownloadError,
+    OnDemandBackupDelegate,
+    PerSessionKeyBackupDownloader,
+} from "../../../src/rust-crypto/PerSessionKeyBackupDownloader";
+import { logger } from "../../../src/logger";
+import { defer } from "../../../src/utils";
+import { RustBackupCryptoEventMap, RustBackupCryptoEvents, RustBackupDecryptor } from "../../../src/rust-crypto/backup";
+import * as TestData from "../../test-utils/test-data";
+import { ConnectionError, CryptoEvent, MatrixError, TypedEventEmitter } from "../../../src";
+
+describe("PerSessionKeyBackupDownloader", () => {
+    /** The downloader under test */
+    let downloader: PerSessionKeyBackupDownloader;
+
+    let delegate: Mocked<OnDemandBackupDelegate>;
+
+    const BACKOFF_TIME = 2000;
+
+    const mockEmitter = new TypedEventEmitter() as TypedEventEmitter<RustBackupCryptoEvents, RustBackupCryptoEventMap>;
+
+    async function expectConfigurationError(error: KeyDownloadError): Promise<void> {
+        return new Promise<void>((resolve) => {
+            downloader.on(KeyDownloaderEvent.QueryKeyError, (err) => {
+                if (err === error) {
+                    resolve();
+                }
+            });
+        });
+    }
+    async function expectLoopStatus(expectedLoopRunning: boolean): Promise<void> {
+        return new Promise<void>((resolve) => {
+            downloader.on(KeyDownloaderEvent.DownloadLoopStateUpdate, (loopRunning) => {
+                if (expectedLoopRunning == loopRunning) {
+                    resolve();
+                }
+            });
+        });
+    }
+
+    async function expectSessionImported(roomId: string, sessionId: string): Promise<void> {
+        return new Promise<void>((resolve) => {
+            downloader.on(KeyDownloaderEvent.KeyImported, (r, s) => {
+                if (roomId == r && sessionId == s) {
+                    resolve();
+                }
+            });
+        });
+    }
+
+    beforeEach(async () => {
+        delegate = {
+            getActiveBackupVersion: jest.fn(),
+            getBackupDecryptionKey: jest.fn(),
+            requestRoomKeyFromBackup: jest.fn(),
+            importRoomKeys: jest.fn(),
+            createBackupDecryptor: jest.fn(),
+            requestKeyBackupVersion: jest.fn(),
+            getCryptoEventEmitter: jest.fn(),
+        } as unknown as Mocked<OnDemandBackupDelegate>;
+
+        delegate.getCryptoEventEmitter.mockReturnValue(mockEmitter);
+
+        downloader = new PerSessionKeyBackupDownloader(delegate, logger, BACKOFF_TIME);
+
+        jest.useFakeTimers();
+    });
+
+    afterEach(() => {
+        downloader.stop();
+        jest.useRealTimers();
+    });
+
+    describe("Given valid backup available", () => {
+        beforeEach(async () => {
+            delegate.getActiveBackupVersion.mockResolvedValue(TestData.SIGNED_BACKUP_DATA.version!);
+            delegate.getBackupDecryptionKey.mockResolvedValue({
+                backupVersion: TestData.SIGNED_BACKUP_DATA.version!,
+                decryptionKey: RustSdkCryptoJs.BackupDecryptionKey.fromBase64(TestData.BACKUP_DECRYPTION_KEY_BASE64),
+            } as unknown as RustSdkCryptoJs.BackupKeys);
+            delegate.createBackupDecryptor.mockReturnValue({
+                decryptSessions: jest.fn().mockResolvedValue([TestData.MEGOLM_SESSION_DATA]),
+            } as unknown as RustBackupDecryptor);
+            delegate.requestKeyBackupVersion.mockResolvedValue(TestData.SIGNED_BACKUP_DATA);
+        });
+
+        it("Should download and import a missing key from backup", async () => {
+            const awaitKeyImported = defer<void>();
+
+            delegate.requestRoomKeyFromBackup.mockResolvedValue(TestData.CURVE25519_KEY_BACKUP_DATA);
+            delegate.importRoomKeys.mockImplementation(async (keys) => {
+                awaitKeyImported.resolve();
+            });
+
+            downloader.onDecryptionKeyMissingError("roomId", "sessionId");
+
+            await awaitKeyImported.promise;
+
+            expect(delegate.requestRoomKeyFromBackup).toHaveBeenCalledWith("1", "roomId", "sessionId");
+            expect(delegate.createBackupDecryptor).toHaveBeenCalledTimes(1);
+        });
+
+        it("Should not hammer the backup if the key is requested repeatedly", async () => {
+            const blockOnServerRequest = defer<void>();
+            // simulate a key not being in the backup
+            delegate.requestRoomKeyFromBackup.mockImplementation(async (version, room, session) => {
+                await blockOnServerRequest.promise;
+                return TestData.CURVE25519_KEY_BACKUP_DATA;
+            });
+
+            // Call 3 times
+            downloader.onDecryptionKeyMissingError("!roomId", "sessionId");
+            downloader.onDecryptionKeyMissingError("!roomId", "sessionId");
+            downloader.onDecryptionKeyMissingError("!roomId", "sessionId");
+
+            downloader.onDecryptionKeyMissingError("!roomId", "sessionId2");
+
+            const session2Imported = new Promise<void>((resolve) => {
+                downloader.on(KeyDownloaderEvent.KeyImported, (roomId, sessionId) => {
+                    if (sessionId === "sessionId2") {
+                        resolve();
+                    }
+                });
+            });
+            blockOnServerRequest.resolve();
+
+            await session2Imported;
+            expect(delegate.requestRoomKeyFromBackup).toHaveBeenCalledTimes(2);
+        });
+
+        it("should continue to next key if current not in backup", async () => {
+            delegate.requestRoomKeyFromBackup.mockResolvedValue(TestData.CURVE25519_KEY_BACKUP_DATA);
+
+            delegate.requestRoomKeyFromBackup.mockImplementation(async (version, room, session) => {
+                if (session == "sessionA0") {
+                    throw new MatrixError(
+                        {
+                            errcode: "M_NOT_FOUND",
+                            error: "No room_keys found",
+                        },
+                        404,
+                    );
+                } else if (session == "sessionA1") {
+                    return TestData.CURVE25519_KEY_BACKUP_DATA;
+                }
+            });
+
+            const expectImported = expectSessionImported("!roomA", "sessionA1");
+            const expectNotFound = expectConfigurationError(KeyDownloadError.MISSING_DECRYPTION_KEY);
+
+            downloader.onDecryptionKeyMissingError("!roomA", "sessionA0");
+            downloader.onDecryptionKeyMissingError("!roomA", "sessionA1");
+
+            await expectNotFound;
+            await expectImported;
+        });
+
+        it("Should not query repeatedly for a key not in backup", async () => {
+            delegate.requestRoomKeyFromBackup.mockResolvedValue(TestData.CURVE25519_KEY_BACKUP_DATA);
+
+            delegate.requestRoomKeyFromBackup.mockRejectedValue(
+                new MatrixError(
+                    {
+                        errcode: "M_NOT_FOUND",
+                        error: "No room_keys found",
+                    },
+                    404,
+                ),
+            );
+
+            const expectNotFound = expectConfigurationError(KeyDownloadError.MISSING_DECRYPTION_KEY);
+
+            downloader.onDecryptionKeyMissingError("!roomA", "sessionA0");
+
+            await expectNotFound;
+
+            const currentCallCount = delegate.requestRoomKeyFromBackup.mock.calls.length;
+
+            // Should not query again for a key not in backup
+
+            downloader.onDecryptionKeyMissingError("!roomA", "sessionA0");
+            expect(delegate.requestRoomKeyFromBackup).toHaveBeenCalledTimes(currentCallCount);
+
+            // advance time to retry
+            jest.advanceTimersByTime(BACKOFF_TIME + 10);
+
+            const expectNotFoundSecondAttempt = expectConfigurationError(KeyDownloadError.MISSING_DECRYPTION_KEY);
+            downloader.onDecryptionKeyMissingError("!roomA", "sessionA0");
+            await expectNotFoundSecondAttempt;
+        });
+
+        it("Should stop properly", async () => {
+            // Simulate a call to stop while request is in flight
+            const blockOnServerRequest = defer<void>();
+            const requestRoomKeyCalled = defer<void>();
+
+            // Mock the request to block
+            delegate.requestRoomKeyFromBackup.mockImplementation(async (version, room, session) => {
+                requestRoomKeyCalled.resolve();
+                await blockOnServerRequest.promise;
+                return TestData.CURVE25519_KEY_BACKUP_DATA;
+            });
+
+            const expectStopped = expectConfigurationError(KeyDownloadError.STOPPED);
+            const expectLoopStarted = expectLoopStatus(true);
+
+            downloader.onDecryptionKeyMissingError("!roomA", "sessionA0");
+            downloader.onDecryptionKeyMissingError("!roomA", "sessionA1");
+            downloader.onDecryptionKeyMissingError("!roomA", "sessionA2");
+            downloader.onDecryptionKeyMissingError("!roomA", "sessionA3");
+
+            await expectLoopStarted;
+            await requestRoomKeyCalled.promise;
+            downloader.stop();
+
+            blockOnServerRequest.resolve();
+            await expectStopped;
+            expect(delegate.importRoomKeys).not.toHaveBeenCalled();
+            expect(delegate.requestRoomKeyFromBackup).toHaveBeenCalledTimes(1);
+        });
+    });
+
+    describe("Given no usable backup available", () => {
+        let loopPausedPromise: Promise<void>;
+        let configurationErrorPromise: Promise<void>;
+
+        beforeEach(async () => {
+            delegate.getActiveBackupVersion.mockResolvedValue(null);
+            delegate.getBackupDecryptionKey.mockResolvedValue(null);
+
+            loopPausedPromise = expectLoopStatus(false);
+
+            configurationErrorPromise = expectConfigurationError(KeyDownloadError.CONFIGURATION_ERROR);
+        });
+
+        it("Should not query server if no backup", async () => {
+            delegate.requestKeyBackupVersion.mockResolvedValue(null);
+
+            downloader.onDecryptionKeyMissingError("!roomId", "sessionId");
+
+            await loopPausedPromise;
+            await configurationErrorPromise;
+        });
+
+        it("Should not query server if backup not active", async () => {
+            // there is a backup
+            delegate.requestKeyBackupVersion.mockResolvedValue(TestData.SIGNED_BACKUP_DATA);
+            // but it's not active
+            delegate.getActiveBackupVersion.mockResolvedValue(null);
+
+            downloader.onDecryptionKeyMissingError("!roomId", "sessionId");
+
+            await loopPausedPromise;
+            await configurationErrorPromise;
+        });
+
+        it("Should stop if backup key is not cached", async () => {
+            // there is a backup
+            delegate.requestKeyBackupVersion.mockResolvedValue(TestData.SIGNED_BACKUP_DATA);
+            // it is active
+            delegate.getActiveBackupVersion.mockResolvedValue(TestData.SIGNED_BACKUP_DATA.version!);
+            // but the key is not cached
+            delegate.getBackupDecryptionKey.mockResolvedValue(null);
+
+            downloader.onDecryptionKeyMissingError("!roomId", "sessionId");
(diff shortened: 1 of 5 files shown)


src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,576 @@
...
+/**
+ * Utility to create a delegate for the PerSessionKeyBackupDownloader that is usable by rust crypto.
+ *
+ * @param rustCrypto - The rust crypto instance.
+ * @param backupManager - The backup manager instance.
+ * @param olmMachine - The olm machine instance.
+ * @param http - The http instance.
+ */
+export function createDelegate(
+    rustCrypto: RustCrypto,
```
richvdh, line 82: I'd rather we didn't have to pass the whole of RustCrypto in here: circular relationships like this make it hard to reason about the logic flows. Could we instead:

* Call `OlmMachine.importBackedUpRoomKeys` directly instead of going via `RustCrypto.importRoomKeys`
* Expose `PerSessionKeyBackupDownloader.onBackupStatusChange` and make `RustCrypto` responsible for calling it.

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,576 @@
...
+import { ClientPrefix, IHttpOpts, MatrixError, MatrixHttpApi, Method } from "../http-api";
+import { RustBackupCryptoEventMap, RustBackupCryptoEvents, RustBackupDecryptor, RustBackupManager } from "./backup";
+import { CryptoEvent, TypedEventEmitter } from "../matrix";
+import { encodeUri, sleep } from "../utils";
+import { RustCrypto } from "./rust-crypto";
+
+/**
+ * Extract the dependence of the PerSessionKeyBackupDownloader, main reason is to make testing easier.
+ */
+export interface OnDemandBackupDelegate {
```
richvdh, line 32: I'm not convinced this "delegate" abstraction layer is helpful; rather, I think it obfuscates what's going on.

Plenty of other components (eg `BackupManager`) just take references to the `OlmMachine`, the http API, etc, and call them directly. Can't we do that here? For the tests, you need to mock out each of those dependencies, but that's not very hard?

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,576 @@
...
+}
+export type KeyDownloaderEventMap = {
+    [KeyDownloaderEvent.DownloadLoopStateUpdate]: (loopRunning: boolean) => void;
+    [KeyDownloaderEvent.DownLoopStep]: (remaining: number) => void;
+    [KeyDownloaderEvent.QueryKeyError]: (errCode: KeyDownloadError) => void;
+    [KeyDownloaderEvent.KeyImported]: (roomId: string, sessionId: string) => void;
+};
+
+/**
+ * This function is called when an 'unable to decrypt' error occurs. It attempts to download the key from the backup.
```
richvdh, line 182: > This function is called when an 'unable to decrypt' error occurs

It's not a function

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,576 @@
...
+ *
+ * The current backup API lacks pagination, which can lead to lengthy key retrieval times for large histories (several 10s of minutes).
+ * To mitigate this, keys are downloaded on demand as decryption errors occurs.
+ * While this approach may result in numerous requests, it improves user experience by reducing wait times for message decryption.
+ * 
+ * The PerSessionKeyBackupDownloader is resistant to backup configuration changes, it will automatically resume querying when
+ * the backup is configured correctly.
```
richvdh, line 189:
```suggestion
 * The PerSessionKeyBackupDownloader is resistant to backup configuration changes: it will automatically resume querying when
 * the backup is configured correctly.
```

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,576 @@
...
+    | { ok: false; error: KeyDownloadError; [key: string]: any };
+
+/** Holds the current backup decryptor and version that should be used. */
+type Configuration = {
+    backupVersion: string;
+    decryptor: RustBackupDecryptor;
+};
+
+/**
+ * Signaling for the Downloader loop.
```
richvdh, line 165:
```suggestion
 * Signalling for the downloader loop.
```

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,576 @@
...
+ * Not yet used by API, yet useful for testing.
+ */
+export enum KeyDownloaderEvent {
+    DownloadLoopStateUpdate = "download_loop_started",
+    DownLoopStep = "download_loop_step",
+    QueryKeyError = "query_key_error",
+    KeyImported = "key_imported",
```
richvdh, line 172: If these events are only used in the unit test, then I'd really like us to look hard at whether we can find other ways to write the test. It feels wrong to have to cover the code in "emit(this)" and "emit(that)" just for the benefit of testing. For example: can we infer the behaviour of the download loop from the calls to the lower-level components?

 Some of them (`DownLoopStep`?) don't even seem to be used at all.

If we keep the events:

* The names don't match the conventions we use elsewhere in the js-sdk.
* Also they need documentation as to what they indicate and what the payloads are.

robintown (commented):
Was halfway through review when I saw Rich's comments come in, so I'll defer to him

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,469 @@
...
+    CONFIGURATION_ERROR = "CONFIGURATION_ERROR",
+}
+
+/** Helper type for requested session*/
+type SessionInfo = { roomId: string; megolmSessionId: string };
+
+/** Helper type for the result of a key download. */
+type KeyDownloadResult =
+    | { ok: true; value: KeyBackupSession }
+    | { ok: false; error: KeyDownloadError; [key: string]: any };
```
richvdh, line 55: the idiomatic way to do this would be to have `queryKeyBackup` throw a [custom error type](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#custom_error_types):


```
enum KeyDownloadErrorCode {
    VERSION_MISMATCH = "VERSION_MISMATCH",
    // etc
}

class KeyDownloadError extends Error {
    constructor(private readonly code: KeyDownloadErrorCode) {
        super(code);
    }
}

// ...
try {
    let result = await queryKeyBackup(/* ... */);
} catch (e) {
    if (!e isinstance KeyDownloadError) { throw e }
    switch (e.code) { 
       case KeyDownloadErrorCode.VERSION_MISMATCH:
          // etc
    }
}
```

For `RATE_LIMITED`, you might want a second custom error type, which holds the `retryAfterMs` value.

richvdh (changes requested):
I think this is broadly sensible, though I haven't yet managed to follow all the logic through from start to finish. I've made a bunch of suggestions to make it easier to follow: please could you try them, and generally make sure that everything is consistent, and then I'll take another check on the logic here.

src/rust-crypto/backup.ts
```diff
@@ -172,6 +173,19 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
         // importing keys from backup if needed.
         this.emit(CryptoEvent.KeyBackupDecryptionKeyCached, version);
     }
+    public async importRoomKeys(keys: IMegolmSessionData[], opts?: ImportRoomKeysOpts): Promise<void> {
```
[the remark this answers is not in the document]
BillCarsonFr: Why do we have to bother with spaces, can't lint do that?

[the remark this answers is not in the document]
richvdh: I'm surprised it doesn't, tbh.

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,469 @@
...
+/**
+ * Used when an 'unable to decrypt' error occurs. It attempts to download the key from the backup.
+ *
+ * The current backup API lacks pagination, which can lead to lengthy key retrieval times for large histories (several 10s of minutes).
+ * To mitigate this, keys are downloaded on demand as decryption errors occurs.
+ * While this approach may result in numerous requests, it improves user experience by reducing wait times for message decryption.
+ *
+ * The PerSessionKeyBackupDownloader is resistant to backup configuration changes: it will automatically resume querying when
+ * the backup is configured correctly.
+ *
```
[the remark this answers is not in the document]
BillCarsonFr: Again can this be a lint thing? why do that manually as untold convention

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,469 @@
...
+        const backupDecryptor = this.backupManager.createBackupDecryptor(backupKeys.decryptionKey);
+        this.hasConfigurationProblem = false;
+        this.configuration = {
+            decryptor: backupDecryptor,
+            backupVersion: activeVersion,
+        };
+        return this.configuration;
+    }
+
+    private async decryptAndImport(sessionInfo: SessionInfo, data: KeyBackupSession): Promise<void> {
```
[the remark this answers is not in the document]
BillCarsonFr: sure why not. I have no idea of the expectations regarding method orders in fi;es

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,469 @@
...
+                    // Unfortunately the spec doesn't give us a way to differentiate between a missing key and a wrong version.
+                    // Synapse will return:
+                    //     - "error": "Unknown backup version" if the version is wrong.
+                    //     - "error": "No room_keys found" if the key is missing.
+                    // For now we check the error message, but this is not ideal.
+                    // It's useful to know if the key is missing or if the version is wrong.
+                    // As it's not spec'ed, will work only with synapse, but it's better than nothing?
+                    // Other implementations will consider this as a missing key, but soon after a backup status
+                    // change will trigger a configuration check for future keys (this one won't be retied though)
+                    if (e.data.error == "Unknown backup version") {
```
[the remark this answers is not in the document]
BillCarsonFr: ok, fallbacking to loosing that request

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,469 @@
...
+                    case KeyDownloadError.STOPPED:
+                        // If the downloader was stopped, we don't want to retry.
+                        this.pauseLoop();
+                        return;
+                }
+            }
+        }
+        this.pauseLoop();
+    }
+    /**
```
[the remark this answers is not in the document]
BillCarsonFr: this should be some lint rule if possible

src/rust-crypto/PerSessionKeyBackupDownloader.ts
BillCarsonFr: If I do that I don't see how to make tests. I can check that the spy has returned with something but I don't see how to check that it has thrown. 
I see `toHaveReturnedWith` methods, but no `toHaveThrown` or `ToHaveRejectWith`. `toThrow` is not working

richvdh: maybe you can do something with [`spiedFunction.mock.results`](https://jestjs.io/docs/mock-function-api#mockfnmockresults)

src/rust-crypto/PerSessionKeyBackupDownloader.ts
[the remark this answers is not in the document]
richvdh: > I have no idea of the expectations regarding method orders in files

I'm just looking for some sort of logic, rather than what appears to be a completely random ordering

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,480 @@
...
+            k.room_id = sessionInfo.roomId;
+        }
+        await this.backupManager.importRoomKeys(keys);
+    }
+
+    /**
+     * Gets the current backup configuration or create one if it doesn't exist.
+     *
+     * When a valid configuration is found it is cached and returned for subsequent calls.
+     * If a check is forced or a check has not yet been done, a new check is done.
```
richvdh, line 386:
```suggestion
     * Otherwise, if a check is forced or a check has not yet been done, a new check is done.
```

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,480 @@
...
+
+    /**
+     * Called when the backup status changes (CryptoEvents)
+     * This will trigger a check of the backup configuration.
+     */
+    private onBackupStatusChanged = (): void => {
+        // we want to check configuration
+        this.hasConfigurationProblem = false;
+        this.configuration = null;
+        this.getOrCreateBackupConfiguration(true).then((decryptor) => {
```
richvdh, line 183: Why do we need to set `forceCheck=true` here? As far as I can tell that only makes a difference if `hasConfigurationProblem === true`, which we know it is not.

It would be good to remove `forceCheck` altogether if possible: I find it a confusing aspect of `getOrCreateBackupConfiguration`.

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,480 @@
...
+            if (decryptor) {
```
richvdh, line 184: Nit: `getOrCreateBackupConfiguration` returns a configuration, not a decryptor:

```suggestion
```suggestion
        this.getOrCreateBackupConfiguration(true).then((configuration) => {
            if (configuration) {
```
```

File: src/rust-crypto/PerSessionKeyBackupDownloader.ts [hunk 16]
```diff
@@ -0,0 +1,480 @@
...
+    /** Returns true if the session was requested recently. */
+    private wasRequestedRecently(megolmSessionId: string): boolean {
+        const lastCheck = this.sessionLastCheckAttemptedTime.get(megolmSessionId);
+        if (!lastCheck) return false;
+        return Math.max(Date.now() - lastCheck, 0) < this.backoffDuration;
+    }
+
+    private pauseLoop(): void {
+        this.downloadLoopRunning = false;
+    }
```
richvdh, line 224: as far as I can tell, setting `downloadLoopRunning = false` doesn't actually pause a running loop: instead it just flags that the loop is no longer running. So it should be called `markLoopAsPaused` or something. Or, tbh, just inline it: I don't see that having a separate method adds much clarity.

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,480 @@
...
+                            // continue for next one
+                            this.queuedRequests.shift();
+                            break;
+                        case KeyDownloadErrorCode.NETWORK_ERROR:
+                            // We don't want to hammer if there is a problem, so wait a bit.
+                            await sleep(this.backoffDuration);
+                            break;
+                        case KeyDownloadErrorCode.STOPPED:
+                            // If the downloader was stopped, we don't want to retry.
+                            this.pauseLoop();
```
richvdh, line 300: it might be better to wrap the whole `while` loop in a `try { /*... */ } finally { this.pauseLoop(); }`, to make sure we don't forget to clear `backupLoopRunning`.

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,480 @@
...
+    /** The logger to use */
+    private readonly logger: Logger;
+
+    /** Whether the download loop is running. */
+    private downloadLoopRunning = false;
+
+    /** The list of requests that are queued. */
+    private queuedRequests: SessionInfo[] = [];
+
+    // Remembers if we have a configuration problem.
```
richvdh, line 100: Nit:

```suggestion
    /** Remembers if we have a configuration problem. */
```

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,480 @@
...
+                this.downloadKeysLoop();
+            }
+        });
+    };
+
+    /** Returns true if the megolm session is already queued for download. */
+    private isAlreadyInQueue(roomId: string, megolmSessionId: string): boolean {
+        return this.queuedRequests.some((info) => {
+            return info.roomId == roomId && info.megolmSessionId == megolmSessionId;
+        });
```
richvdh, line 195: I'm a bit concerned that this operation could get a bit slow, if the queue gets long (eg if there is a long-term problem with backup). Indeed, I'm also a bit concerned we could end up eating a lot of memory here.

Maybe we should use a different datastructure to make it easier to query, and/or remove entries after a while?

Also happy to leave this for now, and fix it later.

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,480 @@
...
+    }
+
+    /**
+     * Marks the session as not found in backup, to avoid retrying to soon for a key not in backup
+     * @param megolmSessionId - The megolm session ID that is missing.
+     * */
```
richvdh, line 201: Nit:

```suggestion
     */
```

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,480 @@
...
+    private markAsNotFoundInBackup(megolmSessionId: string): void {
+        const now = Date.now();
+        this.sessionLastCheckAttemptedTime.set(megolmSessionId, now);
+        // if too big make some cleaning to keep under control
+        if (this.sessionLastCheckAttemptedTime.size > 100) {
+            this.sessionLastCheckAttemptedTime = new Map(
+                Array.from(this.sessionLastCheckAttemptedTime).filter((sid, ts) => {
+                    return Math.max(now - ts, 0) < this.backoffDuration;
+                }),
+            );
```
richvdh, line 211: We might end up doing this for every call to markAsNotFoundInBackup (since there is no guarantee that removing the old entries will bring the size below 100), so I'm a bit worried about performance here, too.

But again, happy to leave this for now if you prefer.

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,480 @@
...
+     * Query the backup for a key.
+     *
+     * @param targetRoomId - ID of the room that the session is used in.
+     * @param targetSessionId - ID of the session for which to check backup.
+     */
+    private async queryKeyBackup(targetRoomId: string, targetSessionId: string): Promise<KeyBackupSession> {
+        const configuration = await this.getOrCreateBackupConfiguration(false);
+        if (!configuration) {
+            throw new KeyDownloadError(KeyDownloadErrorCode.CONFIGURATION_ERROR);
+        }
```
richvdh, line 327: it might be better to make the call to `getOrCreateBackupConfiguration` in `downloadKeysLoop` (and then pass the `backupVersion` into `queryKeyBackup`, and the `decryptor` into `decryptAndImport`), which would help ensure we have a matching pair of backup version and decryptor.

I don't insist though. YMMV.

spec/unit/rust-crypto/PerSessionKeyBackupDownloader.spec.ts
```diff
@@ -0,0 +1,610 @@
...
+
+describe("PerSessionKeyBackupDownloader", () => {
+    /** The downloader under test */
+    let downloader: PerSessionKeyBackupDownloader;
+
+    const mockCipherKey: Mocked<KeyBackupSession> = {} as unknown as Mocked<KeyBackupSession>;
+
+    const BACKOFF_TIME = 2000;
+
+    const mockEmitter = new TypedEventEmitter() as TypedEventEmitter<RustBackupCryptoEvents, RustBackupCryptoEventMap>;
```
richvdh, line 49: Suggest initialising this in `beforeEach`, to prevent listeners leaking between tests.

spec/unit/rust-crypto/PerSessionKeyBackupDownloader.spec.ts
```diff
@@ -0,0 +1,610 @@
...
+        } as unknown as Mocked<BackupDecryptor>;
+
+        mockBackupDecryptor.decryptSessions.mockImplementation(async (ciphertexts) => {
+            const sessionId = Object.keys(ciphertexts)[0];
+            return [mockClearSession(sessionId)];
+        });
+
+        mockRustBackupManager = {
+            getActiveBackupVersion: jest.fn(),
+            getBackupDecryptionKey: jest.fn(),
```
richvdh, line 91: RustBackupManager doesn't have a `getBackupDecryptionKey` method (I think?), so I think this is redundant?

spec/unit/rust-crypto/PerSessionKeyBackupDownloader.spec.ts
```diff
@@ -0,0 +1,610 @@
...
+    const mockEmitter = new TypedEventEmitter() as TypedEventEmitter<RustBackupCryptoEvents, RustBackupCryptoEventMap>;
+
+    let mockHttp: MatrixHttpApi<IHttpOpts & { onlyData: true }>;
+    let mockRustBackupManager: Mocked<RustBackupManager>;
+    let mockOlmMachine: Mocked<OlmMachine>;
+    let mockBackupDecryptor: Mocked<BackupDecryptor>;
+
+    let expectedSession: { [roomId: string]: { [sessionId: string]: IDeferred<void> } };
+
+    async function expectSessionImported(roomId: string, sessionId: string) {
```
richvdh, line 58: Nit: this isn't an async function: it doesn't contain any `await`s.

```suggestion
    function expectSessionImported(roomId: string, sessionId: string) {
```

spec/unit/rust-crypto/PerSessionKeyBackupDownloader.spec.ts
```diff
@@ -0,0 +1,610 @@
...
+
+            const awaitKey2Imported = defer<void>();
+
+            mockRustBackupManager.importRoomKeys.mockImplementation(async (keys) => {
+                if (keys[0].session_id === "sessionId2") {
+                    awaitKey2Imported.resolve();
+                }
+            });
+
+            // @ts-ignore
```
richvdh, line 184: it's generally considered good practice to include details about why we are overriding typescript:

```suggestion
            // @ts-ignore access to private property
```

... but I don't think it's important in unit test code.

spec/unit/rust-crypto/PerSessionKeyBackupDownloader.spec.ts
```diff
@@ -0,0 +1,610 @@
...
+            await requestRoomKeyCalled.promise;
+            downloader.stop();
+
+            blockOnServerRequest.resolve();
+
+            // let the first request complete
+            await jest.runAllTimersAsync();
+
+            expect(mockRustBackupManager.importRoomKeys).not.toHaveBeenCalled();
+            expect(callCount).toStrictEqual(1);
```
richvdh, line 294: Rather than manually maintaining a counter, I think you can probably do this with something like:

```suggestion
            expect(fetchMock.calls(`express:/_matrix/client/v3/room_keys/keys/:roomId/:sessionId`).length).toStrictEqual(1);
```

... but it's not important.

spec/unit/rust-crypto/PerSessionKeyBackupDownloader.spec.ts
```diff
@@ -0,0 +1,610 @@
...
+            mockRustBackupManager.getActiveBackupVersion.mockResolvedValue(null);
+            mockOlmMachine.getBackupKeys.mockResolvedValue(null);
+
+            // @ts-ignore
+            keyQuerySpy = jest.spyOn(downloader, "queryKeyBackup");
+        });
+
+        afterEach(async () => {
+            fetchMock.mockClear();
+        });
```
richvdh, line 311: given we have `fetchMock.mockReset` in the previous `afterEach`, I think this is redundant.

```suggestion
```

spec/unit/rust-crypto/PerSessionKeyBackupDownloader.spec.ts
```diff
@@ -0,0 +1,610 @@
...
+            await jest.runAllTimersAsync();
+
+            await expect(keyQuerySpy.mock.results[0].value).rejects.toThrow(
+                "Failed to get key from backup: CONFIGURATION_ERROR",
+            );
+        });
+    });
+
+    describe("Given Backup state update", () => {
+        it("After initial sync, when backup become trusted it should request keys for past requests", async () => {
```
richvdh, line 400: Nit:

```suggestion
        it("After initial sync, when backup becomes trusted it should request keys for past requests", async () => {
```

spec/unit/rust-crypto/PerSessionKeyBackupDownloader.spec.ts
```diff
@@ -0,0 +1,610 @@
...
+            mockOlmMachine.getBackupKeys.mockResolvedValue({
+                backupVersion: TestData.SIGNED_BACKUP_DATA.version!,
+                decryptionKey: RustSdkCryptoJs.BackupDecryptionKey.fromBase64(TestData.BACKUP_DECRYPTION_KEY_BASE64),
+            } as unknown as RustSdkCryptoJs.BackupKeys);
+
+            jest.useFakeTimers();
+        });
+
+        afterEach(() => {
+            jest.useRealTimers();
```
richvdh, line 461: Again I think these are redundant because they happen at the top level.

```suggestion
```

src/rust-crypto/PerSessionKeyBackupDownloader.ts
BillCarsonFr: I also have concerns, more about the fact that it could only grow and get out of control (particularly with the current relation handling that fetches a lot of events) rather than performance but indead valid. Let's iterate

src/rust-crypto/PerSessionKeyBackupDownloader.ts
BillCarsonFr: Yes good point, updated

src/rust-crypto/PerSessionKeyBackupDownloader.ts
```diff
@@ -0,0 +1,474 @@
...
+            while (this.queuedRequests.length > 0) {
+                // we just peek the first one without removing it, so if a new request for same key comes in while we're
+                // processing this one, it won't queue another request.
+                const request = this.queuedRequests[0];
+                try {
+                    // The backup could have changed between the time we queued the request and now, so we need to check
+                    const configuration = await this.getOrCreateBackupConfiguration();
+                    if (!configuration) {
+                        // Backup is not configured correctly, so stop the loop.
+                        this.downloadLoopRunning = false;
```
richvdh, line 264: this is redundant FWIW: the `finally` below` will handle it.

Likewise the other one on line 298.

PR #3907, opened 2023-11-23
fix uncaught exceptions in Backup Loop for rust sdk

BillCarsonFr:
## Checklist

-   [ ] Tests written for new code (and old code if feasible)
-   [x] Linter and other CI checks pass
-   [ ] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))

---
This change is marked as an *internal change* (Task), so will not be included in the changelog.

PR #3931, opened 2023-11-29
Element-R: better logging for key backup loop

richvdh:
Some minor paper-cut fixes here

---


richvdh:
seems like this is already done by #3907

PR #3932, opened 2023-11-29
Element-R: await `/keys/query` during Verification requests

richvdh:
If the acceptance of an incoming verification request races with a `/keys/query` for the other user, then subsequent operations can fail (see https://github.com/matrix-org/matrix-rust-sdk/issues/2896).

This is a slightly hacky workaround that means we wait for the devicelist to arrive for the relevant operations.

Required for a fix to https://github.com/vector-im/element-web/issues/26420.

---

### DIFF
--- a/src/rust-crypto/verification.ts
+++ b/src/rust-crypto/verification.ts
@@ -57,11 +57,13 @@ export class RustVerificationRequest
     /**
      * Construct a new RustVerificationRequest to wrap the rust-level `VerificationRequest`.
      *
-     * @param inner - VerificationRequest from the Rust SDK
-     * @param outgoingRequestProcessor - `OutgoingRequestProcessor` to use for making outgoing HTTP requests
-     * @param supportedVerificationMethods - Verification methods to use when `accept()` is called
+     * @param olmMachine - The `OlmMachine` from the underlying rust crypto sdk.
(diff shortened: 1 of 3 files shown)


src/rust-crypto/verification.ts
```diff
@@ -322,6 +333,10 @@ export class RustVerificationRequest
             throw new Error(`Unsupported verification method ${method}`);
         }
```
andybalaam, line 335: Is this the critical part that causes us to avoid the race? If so, can we have a comment here explaining that?

andybalaam (approved):
Looks sensible, but I need a little more explanation to understand it I think.

src/rust-crypto/verification.ts
```diff
@@ -322,6 +333,10 @@ export class RustVerificationRequest
...
 
+        if (!(await this.getOtherDevice())) {
```
richvdh, line 336:
```suggestion
        // make sure that we have a list of the other user's devices (workaround https://github.com/matrix-org/matrix-rust-sdk/issues/2896)
        if (!(await this.getOtherDevice())) {
```

src/rust-crypto/verification.ts
richvdh: yes, sorry. Have added a comment.

PR #3934, opened 2023-11-30
Don't back up keys that we got from backup

uhoreg:
Fixes https://github.com/vector-im/element-web/issues/26347

I also threw in a proper enum for the key source, but I can remove that

## Checklist

---

## 🐛 Bug Fixes
 * Don't back up keys that we got from backup ([\#3934](https://github.com/matrix-org/matrix-js-sdk/pull/3934)). Fixes vector-im/element-web#26347.

### DIFF
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -52,6 +52,7 @@ import * as testData from "../../test-utils/test-data";
 import { defer } from "../../../src/utils";
 import { logger } from "../../../src/logger";
 import { OutgoingRequestsManager } from "../../../src/rust-crypto/OutgoingRequestsManager";
+import { Curve25519AuthData } from "../../../src/crypto-api/keybackup";
 
 const TEST_USER = "@alice:example.com";
 const TEST_DEVICE_ID = "TEST_DEVICE";
@@ -935,16 +936,15 @@ describe("RustCrypto", () => {
 
         it("does not back up keys that came from backup", async () => {
             const rustCrypto = await makeTestRustCrypto();
-
-            const olmMachine = rustCrypto.getOlmMachineOrThrow();
+            const olmMachine: OlmMachine = rustCrypto["olmMachine"];
 
             await olmMachine.enableBackupV1(
-                testData.SIGNED_BACKUP_DATA.auth_data.public_key,
-                testData.SIGNED_BACKUP_DATA.version,
+                (testData.SIGNED_BACKUP_DATA.auth_data as Curve25519AuthData).public_key,
+                testData.SIGNED_BACKUP_DATA.version!,
             );
 
             // we import two keys: one "from backup", and one "from export"
-            const [backedUpRoomKey, exportedRoomKey, ..._] = testData.MEGOLM_SESSION_DATA_ARRAY;
+            const [backedUpRoomKey, exportedRoomKey] = testData.MEGOLM_SESSION_DATA_ARRAY;
             await rustCrypto.importRoomKeys([backedUpRoomKey], {
                 source: RoomKeySource.Backup,
             });
@@ -960,10 +960,10 @@ describe("RustCrypto", () => {
                 "rooms": {
                     [exportedRoomKey.room_id]: {
                         sessions: {
-                            [exportedRoomKey.session_id]: {}
-                        }
-                    }
-                }
+                            [exportedRoomKey.session_id]: {},
+                        },
+                    },
+                },
             });
 
             // we expect that the key "from backup" is not present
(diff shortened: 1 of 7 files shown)


src/crypto-api.ts
```diff
@@ -591,6 +591,20 @@ export interface ImportRoomKeyProgressData {
...
+ */
+export enum RoomKeySource {
+    /** from key backup */
+    Backup = "backup",
+    /** from a key export */
+    Export = "export",
+    /** via a key forward */
+    Forward = "forward",
+    /** directly from the sender */
+    Direct = "direct",
```
richvdh, line 605: Are `Forward` and `Direct` ever actually used? It seems like they are not, and since such cases are handled internally within `matrix-sdk-crypto`, it seems like they never actually will be? I don't think we should have options which will never actually be used.

src/crypto-api.ts
```diff
@@ -599,7 +613,8 @@ export interface ImportRoomKeysOpts {
     progressCallback?: (stage: ImportRoomKeyProgressData) => void;
     // TODO, the rust SDK will always such imported keys as untrusted
     untrusted?: boolean;
-    source?: String; // TODO: Enum (backup, file, ??)
+    /** Where the room key came from */
+    source?: RoomKeySource;
```
richvdh, line 617: Hrm. I don't think this should be part of the public API to the JS-SDK: the only sensible value for an external caller is `Export`.

Unfortunately, removing `source` is technically an API-breaking change, (as is changing its type, fwiw,) and we try to avoid them.

What we *can* do is something like:

* Mark `source` as `@deprecated`. Might as well do `untrusted` while we're here.
* Add a new *internal* method to `CryptoBackend` called `importRoomKeysFromBackup` or somesuch. `CryptoApi.importRoomKeys` is only for importing from export.
* Change existing codepaths in the js-sdk which use `importRoomKeys` to import from backup to use the new method.
* In the rust impl, the two methods can be quite separate. In the legacy impl, you probably want to make one call the other. Hack it together in whatever way is easiest.

WDYT?

src/crypto-api.ts
```diff
@@ -591,6 +591,20 @@ export interface ImportRoomKeyProgressData {
     total: number;
 }
 
+/**
+ * Where a room key came from
+ */
+export enum RoomKeySource {
```
BillCarsonFr, line 597: Feels a bit strange to define all these variant that are not used.
Shouldn't the react sdk use `Export` when importing from file? I don't see the Forward/Direct used also

richvdh (changes requested):

src/crypto-api.ts
uhoreg: Yup, I considered adding a separate function, but opted to make fewer changes to the API.  But I can do so, and get rid of an `if` statement.

richvdh: we can just throw an error if the `source` is set to something else.

src/crypto-api.ts
```diff
@@ -93,6 +93,15 @@ export interface CryptoApi {
...
 
+    /**
+     * Import a list of room keys restored from backup
+     *
+     * @param keys - a list of session export objects
+     * @param opts - options object
+     * @returns a promise which resolves once the keys have been imported
+     */
+    importBackedUpRoomKeys(keys: IMegolmSessionData[], opts?: ImportRoomKeysOpts): Promise<void>;
+
```
richvdh, line 104: any reason this needs to be here rather than in `CryptoBackend`?

src/rust-crypto/rust-crypto.ts
```diff
@@ -409,9 +409,32 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
     }
 
     public async importRoomKeys(keys: IMegolmSessionData[], opts?: ImportRoomKeysOpts): Promise<void> {
-        // TODO when backup support will be added we would need to expose the `from_backup` flag in the bindings
-        const jsonKeys = JSON.stringify(keys);
-        await this.olmMachine.importRoomKeys(jsonKeys, (progress: BigInt, total: BigInt) => {
+        if (opts?.source === "backup") {
+            return this.importBackedUpRoomKeys(keys, opts);
+        } else {
```
richvdh, line 414: I'd get rid of this else and outdent the rest of the function.

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -931,6 +932,50 @@ describe("RustCrypto", () => {
...
+                    [exportedRoomKey.room_id]: {
+                        sessions: {
+                            [exportedRoomKey.session_id]: {},
+                        },
+                    },
+                },
+            });
+
+            // we expect that the key "from backup" is not present
+            expect(roomKeys).not.toMatchObject({
```
richvdh, line 969: I think this is implied by the previous check?

richvdh (changes requested):

spec/unit/rust-crypto/rust-crypto.spec.ts
uhoreg: No, `toMatchObject` allows the object to contain keys that aren't present in the pattern.

src/crypto-api.ts
uhoreg: Only because I didn't know about `CryptoBackend` ;)

richvdh (approved):
lgtm!

PR #3940, opened 2023-12-05
Fix "mark_skipped" action for end-to-end tests

richvdh:
This seems to have been broken by https://github.com/matrix-org/matrix-js-sdk/pull/3914, which changed the name of the status check that is updated.

---


richvdh:
(Note that the name of the status check will still be different to the one we're currently looking for, but at least it'll be consistent with the name of the check that is updated when the tests are actually run, so we can then just configure github once.)

MidhunSureshR (approved):

PR #3947, opened 2023-12-08
Implement getting verification cancellation info in Rust crypto

uhoreg:
Fixes https://github.com/vector-im/element-web/issues/26652

## Checklist

---

## 🐛 Bug Fixes
 * Implement getting verification cancellation info in Rust crypto ([\#3947](https://github.com/matrix-org/matrix-js-sdk/pull/3947)). Fixes vector-im/element-web#26652.


uhoreg:
Don't know if I need tests for this

uhoreg:
I realized that `cancelInfo.cancelCode()` returns a matrix-rust-sdk-crypto-wasm `verification.CancelCode`, rather than a `ruma::events::key::verification::cancel::CancelCode`, which on the JavaScript side just becomes a mostly meaningless number.  I think that the best way to fix this is to make `cancelInfo.cancelCode()` return a string, which is the `.to_string()`-ed ruma `CancelCode` (since the string is all that the JS side cares about), though that of course would require incompatible changes to the wasm bindings.  I could get away without changing the wasm bindings by putting a map of numbers to the string equivalents in `rust-crypto/verification.ts`, but that seems kludgy.

richvdh:
> I realized that `cancelInfo.cancelCode()` returns a matrix-rust-sdk-crypto-wasm `verification.CancelCode`, rather than a `ruma::events::key::verification::cancel::CancelCode`, which on the JavaScript side just becomes a mostly meaningless number. I think that the best way to fix this is to make `cancelInfo.cancelCode()` return a string, which is the `.to_string()`-ed ruma `CancelCode` (since the string is all that the JS side cares about), though that of course would require incompatible changes to the wasm bindings. I could get away without changing the wasm bindings by putting a map of numbers to the string equivalents in `rust-crypto/verification.ts`, but that seems kludgy.
> 
> and if we can get rid of the `cancel_with_code` function from the Rust binding, or make it take a `String` instead (it looks like nothing in the JS SDK calls that function), then we can get rid of the wasm version of `CancelCode` completely.

Yes, all this sounds sensible to me. Don't worry about making breaking changes to the wasm bindings.

uhoreg:
wasm PR is at https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/71

The tests on this PR won't pass until it gets merged, and this PR is updated to use the new code (which I assume means just waiting for a new release of the bindings and then bumping the required version number in `package.json`)

uhoreg:
I merged in develop, which depends on the latest crypto wasm bindings.  So the tests now pass.

src/rust-crypto/verification.ts
```diff
@@ -425,7 +425,12 @@ export class RustVerificationRequest
      * this verification.
      */
     public get cancellationCode(): string | null {
-        throw new Error("not implemented");
+        const cancelInfo = this.inner.cancelInfo;
+        if (cancelInfo) {
+            return cancelInfo.cancelCode();
+        } else {
+            return null;
+        }
```
richvdh, line 433:
```suggestion
        return this.inner.cancelInfo?.cancelCode() ?? null;
```

spec/integ/crypto/verification.spec.ts
```diff
@@ -743,6 +743,8 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
             expect(toDeviceMessage.transaction_id).toEqual(transactionId);
             expect(toDeviceMessage.code).toEqual("m.user");
             expect(request.phase).toEqual(VerificationPhase.Cancelled);
+            expect(request.cancellationCode).toEqual("m.user");
+            expect(request.cancellingUserId).toEqual("@alice:localhost");
```
richvdh, line 747: I feel like there are quite a lot of codepaths here which we're not testing at all, but I won't insist for now :)

richvdh (approved):
lgtm

PR #3948, opened 2023-12-08
ElementR: Ensure there is only one call to shareRoomKeys in flight at once

BillCarsonFr:
According to rust sdk documentation there should only be one call in flight to shareRoomKeys, it was not the case.

If `prepareToEncrypt` is called in background (when starting to type a message), and the message is sent before it completes there could be 2 in flight.

## Checklist

---

### DIFF
--- a/src/rust-crypto/RoomEncryptor.ts
+++ b/src/rust-crypto/RoomEncryptor.ts
@@ -45,6 +45,9 @@ export class RoomEncryptor {
     /** whether the room members have been loaded and tracked for the first time */
     private lazyLoadedMembersResolved = false;
 
+    /** Ensures that there is only one call to shareRoomKeys at a time */
+    private currentShareRoomKeyPromise: Promise<void>;
+
     /**
      * @param olmMachine - The rust-sdk's OlmMachine
      * @param keyClaimManager - Our KeyClaimManager, which manages the queue of one-time-key claim requests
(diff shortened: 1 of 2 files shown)


src/rust-crypto/RoomEncryptor.ts
```diff
@@ -45,6 +45,9 @@ export class RoomEncryptor {
     /** whether the room members have been loaded and tracked for the first time */
     private lazyLoadedMembersResolved = false;
 
+    /** Ensures that there is only one call to shareRoomKeys at a time */
+    private currentShareRoomKeyPromise: Promise<void>;
```
florianduros, line 49:
```suggestion
    private currentShareRoomKeyPromise = Promise.resolve();
```

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -61,6 +64,8 @@ export class RoomEncryptor {
     ) {
         this.prefixedLogger = logger.getChild(`[${room.roomId} encryption]`);
 
+        this.currentShareRoomKeyPromise = Promise.resolve();
```
florianduros, line 67: We can remove it with my previous suggestion

florianduros (changes requested):
The solution seems fine but the tests are missing

florianduros (approved):

PR #3950, opened 2023-12-11
ElementR: Fix missing key check values in 4S key storage

BillCarsonFr:
Fixes https://github.com/vector-im/element-web/issues/26721

## Checklist

---

## 🚨 BREAKING CHANGES
 * ElementR: Fix missing key check values in 4S key storage ([\#3950](https://github.com/matrix-org/matrix-js-sdk/pull/3950)). Fixes vector-im/element-web#26721.

### DIFF
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -739,7 +739,7 @@ describe("RustCrypto", () => {
             // Expect the private key to be an Uint8Array with a length of 32
             expect(recoveryKey.privateKey).toBeInstanceOf(Uint8Array);
             expect(recoveryKey.privateKey.length).toBe(32);
-            // Expect passphrase info to be empty
+            // Expect passphrase info to be absent
             expect(recoveryKey.keyInfo?.passphrase).toBeUndefined();
         });
 
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -740,7 +740,7 @@ describe("RustCrypto", () => {
             expect(recoveryKey.privateKey).toBeInstanceOf(Uint8Array);
             expect(recoveryKey.privateKey.length).toBe(32);
             // Expect passphrase info to be empty
-            expect(recoveryKey.passphrase).toBeUndefined();
+            expect(recoveryKey.keyInfo?.passphrase).toBeUndefined();
         });
 
         it("should create a recovery key with password", async () => {
@@ -752,8 +752,8 @@ describe("RustCrypto", () => {
             expect(recoveryKey.privateKey).toBeInstanceOf(Uint8Array);
             expect(recoveryKey.privateKey.length).toBe(32);
             // Expect keyInfo.passphrase to be filled
-            expect(recoveryKey.passphrase?.algorithm).toBe("m.pbkdf2");
-            expect(recoveryKey.passphrase?.iterations).toBe(500000);
+            expect(recoveryKey.keyInfo?.passphrase?.algorithm).toBe("m.pbkdf2");
+            expect(recoveryKey.keyInfo?.passphrase?.iterations).toBe(500000);
         });
     });
 
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -739,8 +739,8 @@ describe("RustCrypto", () => {
             // Expect the private key to be an Uint8Array with a length of 32
             expect(recoveryKey.privateKey).toBeInstanceOf(Uint8Array);
             expect(recoveryKey.privateKey.length).toBe(32);
-            // Expect keyInfo to be empty
-            expect(Object.keys(recoveryKey.keyInfo!).length).toBe(0);
+            // Expect passphrase info to be empty
+            expect(recoveryKey.passphrase).toBeUndefined();
         });
 
         it("should create a recovery key with password", async () => {
@@ -752,8 +752,8 @@ describe("RustCrypto", () => {
             expect(recoveryKey.privateKey).toBeInstanceOf(Uint8Array);
             expect(recoveryKey.privateKey.length).toBe(32);
             // Expect keyInfo.passphrase to be filled
-            expect(recoveryKey.keyInfo?.passphrase?.algorithm).toBe("m.pbkdf2");
-            expect(recoveryKey.keyInfo?.passphrase?.iterations).toBe(500000);
+            expect(recoveryKey.passphrase?.algorithm).toBe("m.pbkdf2");
+            expect(recoveryKey.passphrase?.iterations).toBe(500000);
         });
     });
 
--- a/src/crypto-api.ts
+++ b/src/crypto-api.ts
@@ -716,8 +716,8 @@ export interface GeneratedSecretStorageKey {
     keyInfo?: {
         /** If the key was derived from a passphrase, information (algorithm, salt, etc) on that derivation. */
         passphrase?: PassphraseInfo;
-        /** The optional name of the key (human-readable), would be stored in account_data  */
-        name?: string; // maybe we don't need this?
+        /** Optional human-readable name for the key, to be stored in account_data. */
+        name?: string;
     };
     /** The raw generated private key. */
     privateKey: Uint8Array;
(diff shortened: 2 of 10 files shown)


src/rust-crypto/rust-crypto.ts
```diff
@@ -755,7 +755,8 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
 
         const secretStorageKeyObject = await this.secretStorage.addKey(
             SECRET_STORAGE_ALGORITHM_V1_AES,
-            secretStorageKey.keyInfo,
+            // add the private key in the options in order to encrypt the zero secret for key checks.
+            { ...secretStorageKey.keyInfo, key: secretStorageKey.privateKey },
```
BillCarsonFr, line 759: This is basically the fix, but I wonder if the private key should not already be in that `keyInfo`? So maybe should be done by `createSecretStorageKey`. @richvdh I think you are more familiar with that code

BillCarsonFr: The place where legacy is doing similar https://github.com/matrix-org/matrix-js-sdk/blob/13c7e0ebda88d3676c8bc403fa4c5bd88a17023a/src/crypto/index.ts#L983

richvdh: > So maybe should be done by `createSecretStorageKey`

Ugh. Yes. In an ideal world, it should. However, in practice, I think it's more complicated.

Currently we have:

```ts
export interface GeneratedSecretStorageKey {
    keyInfo?: AddSecretStorageKeyOpts;
    /** The raw generated private key. */
    privateKey: Uint8Array;
    /** The generated key, encoded for display to the user per https://spec.matrix.org/v1.7/client-server-api/#key-representation. */
    encodedPrivateKey?: string;
}
```

and

```ts
export interface AddSecretStorageKeyOpts {
    pubkey?: string;
    passphrase?: PassphraseInfo;
    name?: string;
    key?: Uint8Array;
}
```

... which means that the private key is duplicated.

I think this structure is needed because there are some flows (notably when the 4S key comes from some customisation of matrix-react-sdk) when the key does not itself need storing in 4S) - so `keyInfo` is undefined. (Of course, that doesn't work with rust crypto at the moment, because of the check on line 752, but that's the idea.)

My suggestion would be:

 * *Relax* the definition of `GeneratedSecretStorageKey` so that it doesn't include the full `AddSecretStorageKeyOpts` structure:
   ```ts
   export interface GeneratedSecretStorageKey {
       keyInfo?: {
           passphrase?: PassphraseInfo;
           name?: string;
       },
       /** The raw generated private key. */
       privateKey: Uint8Array;
       /** The generated key, encoded for display to the user per https://spec.matrix.org/v1.7/client-server-api/#key-representation. */
       encodedPrivateKey?: string;
   }
   ```

   * Make `key` a **required** property of `AddSecretStorageKeyOpts`, since there should be no reason to call `secretStorage.addKey` without setting it. I also think that `pubkey` is unused and should be removed, or at least deprecated.
   
   *  Whereever we try to build an `AddSecretStorageKeyOpts` from a `GeneratedSecretStorageKey`, copy the field across.
   
   * Make sure we *document* the behaviour of these properties. Half the reason we have this mess is that `AddSecretStorageKeyOpts.key` lacks any documentation about wtf it is.

src/secret-storage.ts
```diff
@@ -100,10 +100,12 @@ export interface PassphraseInfo {
  * Options for {@link ServerSideSecretStorageImpl#addKey}.
  */
 export interface AddSecretStorageKeyOpts {
-    pubkey?: string;
```
BillCarsonFr, line 103: This is some legacy field that was used for the deprecated `m.secret_storage.v1.curve25519-aes-sha2` algorithm.

src/crypto-api.ts
```diff
@@ -713,7 +713,8 @@ export interface CrossSigningKeyInfo {
  * Recovery key created by {@link CryptoApi#createRecoveryKeyFromPassphrase}
  */
 export interface GeneratedSecretStorageKey {
-    keyInfo?: AddSecretStorageKeyOpts;
+    /** Information to generate the key from a passphrase if any. */
```
richvdh, line 716: Nit: could be a bit clearer here

```suggestion
    /** If the key was derived from a passphrase, information (algorithm, salt, etc) on that derivation. */
```

src/secret-storage.ts
```diff
@@ -100,10 +100,12 @@ export interface PassphraseInfo {
...
+    /** Information for deriving the key from a passphrase if any. */
     passphrase?: PassphraseInfo;
+    /** Optional name of the key. */
     name?: string;
-    key?: Uint8Array;
+    /** The private key. Will be used to generate the key check values in the key info, it will not be stored on the server */
```
richvdh, line 107: Nit:

```suggestion
    /** The private key. Will be used to generate the key check values in the key info; it will not be stored on the server */
```

src/crypto-api.ts
```diff
@@ -713,7 +713,8 @@ export interface CrossSigningKeyInfo {
  * Recovery key created by {@link CryptoApi#createRecoveryKeyFromPassphrase}
```
richvdh, line 713: not a new issue, but this documentation is incomplete. This interface is also the return type of `CreateSecretStorageOpts.createSecretStorageKey`.

src/crypto-api.ts
```diff
@@ -713,7 +713,8 @@ export interface CrossSigningKeyInfo {
...
  */
 export interface GeneratedSecretStorageKey {
-    keyInfo?: AddSecretStorageKeyOpts;
+    /** Information to generate the key from a passphrase if any. */
+    passphrase?: PassphraseInfo;
```
richvdh, line 717: The problem with this change (replacing the optional `keyInfo` property with an optional `passphrase` property)   is that it means any existing implementation of `CreateSecretStorageOpts.createSecretStorageKey` (such as the one in matrix-react-sdk) will be incorrect. Worse, it will fail *silently* because new new property is optional. I think this is very dangerous.

I see two possible ways forward:

 * Retain the `keyInfo` property, but with a reduced structure. This is what I suggested before:
    ```ts
    export interface GeneratedSecretStorageKey {
        keyInfo?: {
            passphrase?: PassphraseInfo;
            name?: string;  // maybe we don't need this?
        },
        privateKey: Uint8Array;
        encodedPrivateKey?: string;
    }
    ```
 * Define a new interface to replace `GeneratedSecretStorageKey` (and deprecate the old one). Define a new property in `CreateSecretStorageOpts` to replace `createSecretStorageKey` which returns the new interface (and deprecate the old one); allow clients to provide either interface.

The second option ends up being more elegant (once we get rid of the deprecated classes/properties), but seems a lot more work.

src/crypto-api.ts
```diff
@@ -710,10 +710,16 @@ export interface CrossSigningKeyInfo {
 }
 
 /**
- * Recovery key created by {@link CryptoApi#createRecoveryKeyFromPassphrase}
+ * Recovery key created by {@link CryptoApi#createRecoveryKeyFromPassphrase}.
+ * Can also be created by {@link CreateSecretStorageOpts#createSecretStorageKey}.
```
richvdh, line 714:
```suggestion
 * Recovery key created by {@link CryptoApi#createRecoveryKeyFromPassphrase} or {@link CreateSecretStorageOpts#createSecretStorageKey}.
```

src/crypto-api.ts
```diff
@@ -710,10 +710,16 @@ export interface CrossSigningKeyInfo {
...
  */
 export interface GeneratedSecretStorageKey {
-    keyInfo?: AddSecretStorageKeyOpts;
+    keyInfo?: {
+        /** If the key was derived from a passphrase, information (algorithm, salt, etc) on that derivation. */
+        passphrase?: PassphraseInfo;
+        /** The optional name of the key (human-readable), would be stored in account_data  */
+        name?: string; // maybe we don't need this?
```
richvdh, line 721:
```suggestion
        /** Optional human-readable name for the key, to be stored in account_data. */
        name?: string;
```

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -739,8 +739,8 @@ describe("RustCrypto", () => {
             // Expect the private key to be an Uint8Array with a length of 32
             expect(recoveryKey.privateKey).toBeInstanceOf(Uint8Array);
             expect(recoveryKey.privateKey.length).toBe(32);
-            // Expect keyInfo to be empty
-            expect(Object.keys(recoveryKey.keyInfo!).length).toBe(0);
+            // Expect passphrase info to be empty
```
richvdh, line 742:
```suggestion
            // Expect passphrase info to be absent
```

richvdh (approved):
lgtm other than a few wording nits

PR #3959, opened 2023-12-12
Remove m.thread filter from relations API call

dbkr:
We used MSC3981 to pass the recurse param to the /relations endpoint so that we could get relations to events in a thread, but we kept the rel_type filter on (as m.thread) so no second-order relations would ever have been returned (a nested thread isn't a thing).

This removes the filter and does some filtering on the client side to remove any events that shouldn't live in the threaded timeline (ie. non-thread relations to the thread root event).

This should help fix stuck unreads because it will avoid the event that the receipt refers to going missing (but only on HSes that support MSC3981).

For https://github.com/vector-im/element-web/issues/26718

## Checklist

---

## 🐛 Bug Fixes
 * Remove m.thread filter from relations API call ([\#3959](https://github.com/matrix-org/matrix-js-sdk/pull/3959)).


t3chguy (approved):
Looks sane

PR #3960, opened 2023-12-14
Fetch events that read receipts point to if we don't have them

dbkr:
As the comment hopefully explains, this is a giant hack to work around lack of MSC3981 support and hopefully remove some spurious unread dots on threads. With this code, the unread dots appear briefly and then go away again once the event in question is fetched which is not fantastic but better then a perma-stuck dot.

## Checklist

### DIFF
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -140,6 +140,8 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
      */
     public replayEvents: MatrixEvent[] | null = [];
 
+    private isCheckingForMissingReceiptEvent = false;
+
     public constructor(public readonly id: string, public rootEvent: MatrixEvent | undefined, opts: IThreadOpts) {
         super();
 
@@ -624,6 +626,8 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
             }
         }
 
+        this.checkForMissingReceiptEvent();
+
         this.emit(ThreadEvent.Update, this);
     }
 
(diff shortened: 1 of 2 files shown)


src/models/thread.ts
```diff
@@ -656,6 +660,45 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
...
+     * This is mainly for situations where the read receipt is pointing to a relation event that we don't
+     * have because it's too far back in the scrollback for us to have paginated it and it doesn't come down
+     * with the thread events due to lack of MSC3981 support.
+     *
+     * If you are reading this in the future and MSC3981 is merged and widely supported then you, future
+     * space person, get the honour of building a large pyre to incinerate this piece of code and revel
+     * in the delicious Typescript fumes.
+     */
+    private async checkForMissingReceiptEvent(): Promise<void> {
+        if (this.isCheckingForMissingReceiptEvent) return;
```
t3chguy, line 675: This feels like it might want to queue/debounce rather than drop given the async action may be looking for eventId123 when the RR has moved onto eventId456 via a gappy sync during startup

src/models/thread.ts
```diff
@@ -624,6 +626,8 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
             }
         }
 
+        this.checkForMissingReceiptEvent();
```
richvdh, line 629: is it deliberate that this lacks an `await`? If so, might be good to comment it to highlight and explain why.

src/models/thread.ts
```diff
@@ -656,6 +660,45 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
...
+     * space person, get the honour of building a large pyre to incinerate this piece of code and revel
+     * in the delicious Typescript fumes.
+     */
+    private async checkForMissingReceiptEvent(): Promise<void> {
+        if (this.isCheckingForMissingReceiptEvent) return;
+        this.isCheckingForMissingReceiptEvent = true;
+
+        try {
+            const myThreadedReceipt = this.getLatestReceipt(this.client.getUserId()!, true);
+            if (myThreadedReceipt && this.timeline.length > 0 && !this.findEventById(myThreadedReceipt.eventId)) {
```
richvdh, line 680: why the condition on `timeline.length` ? A comment would help.

richvdh (commented):
I have to say I'm not enthusiastic about this hack. We already have millions of `/event/` requests to work around some other failure to land an MSC; it causes load on the server to serve all the requests, and quite a lot of load on the client in terms of making the requests, attempting to decrypt them, failing to decrypt them, checking key backup, etc etc.

I can see the need for a temporary workaround for lack of MSC3981 but my concern is that, as soon as this lands, the problem is out of sight and it's just another large nail in the coffin of Element's performance that we never get round to doing anything about.

Is there a world in which we can instead prioritise enabling MSC3981, and maybe make this hack opt-in to help test it?

src/models/thread.ts
dbkr: It wasn't, but it now has the throttle where the wrapper doesn't return a promise.

src/models/thread.ts
dbkr: Good point: added.

dbkr:
> I have to say I'm not enthusiastic about this hack. We already have millions of `/event/` requests to work around some other failure to land an MSC; it causes load on the server to serve all the requests, and quite a lot of load on the client in terms of making the requests, attempting to decrypt them, failing to decrypt them, checking key backup, etc etc.
> 
> I can see the need for a temporary workaround for lack of [MSC3981](https://github.com/matrix-org/matrix-spec-proposals/pull/3981) but my concern is that, as soon as this lands, the problem is out of sight and it's just another large nail in the coffin of Element's performance that we never get round to doing anything about.
> 
> Is there a world in which we can instead prioritise enabling [MSC3981](https://github.com/matrix-org/matrix-spec-proposals/pull/3981), and maybe make this hack opt-in to help test it?

Mm, I share your thoughts, especially having the existing set of thread related things which have gained odd workarounds for lack of server support and then presumably been forgotten. I've picked up 3981 and I don't think anything should actually stand in its way now. Synapse's impl is, afacis, correct, so it really should just be a question of FCPing, getting ticks and then turning the flag on in Synapse by default I think.

As a counterpoint, it should be fairly minimal nail as it will only trigger for specific threads that are broken, but yeah, straws and camels and all that.

If we were to make it opt-in I think it would just be for people that were desperate to unstick the unread dots: I don't really think it would help test anything as the impl is sufficiently different between this and actually having MSC3981.

PR #3967, opened 2023-12-18
Remove obsolete references to device-specific push rules

jgarplind:
## Checklist

-   [x] Tests written for new code (and old code if feasible) **Should be covered by existing tests, if any**

Been reading up on push rules, and specifically how the only supported (as per the spec) scope is `global`, e.g. here: https://github.com/matrix-org/matrix-spec/issues/637

I also found [this old PR](https://github.com/matrix-org/matrix-js-sdk/pull/1404) which did the most important parts, but then ran across some parts that unnecessarily catered for a no longer supported use case. Figured it could be a good opportunity to make a minor contribution to a much appreciated repo.

I also considered the possibility to adjust the function signatures to either not take `scope`, or to force it to be the literal `"global"`, but I'm not sure if that is desirable, with regards to API stability, etc.

Type: [task]

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -8805,8 +8805,8 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         ruleId: Exclude<string, RuleId>,
         body: Pick<IPushRule, "actions" | "conditions" | "pattern">,
     ): Promise<{}> {
-        // NB. Scope not uri encoded because devices need the '/'
-        const path = utils.encodeUri("/pushrules/" + scope + "/$kind/$ruleId", {
+        const path = utils.encodeUri("/pushrules/$scope/$kind/$ruleId", {
+            $scope: scope,
             $kind: kind,
             $ruleId: ruleId,
         });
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -8805,8 +8805,8 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         ruleId: Exclude<string, RuleId>,
         body: Pick<IPushRule, "actions" | "conditions" | "pattern">,
     ): Promise<{}> {
-        // NB. Scope not uri encoded because devices need the '/'
-        const path = utils.encodeUri("/pushrules/" + scope + "/$kind/$ruleId", {
+        const path = utils.encodeUri("/pushrules/$scope/$kind/$ruleId", {
+            $scope: scope,
```
richvdh, line 8809: I don't really see that this is a worthwhile improvement. It doesn't actually *remove* the redundant `scope` param (which would be a  concrete improvement, though hard to do without breaking backwards compatibility); rather it just url-encodes it. Since `scope` must always be `global`, this seems to make no difference at all?

richvdh (changes requested):

src/client.ts
jgarplind: I agree. It is a miniscule improvement to legibility / removing redundant comments which may confuse, and the PR description echoes your sentiment about preserving backwards compatibility.

Was hoping that a small touch-up would be appreciated, but if you prefer to keep things as they are, I'm content to cancel this PR as well.

richvdh: Yeah, let's leave it; I don't think it's worthwhile.

Maybe in future we can create `addGlobalPushRule` etc which omit the `scope` param; then we can deprecate the existing methods.

PR #3968, opened 2023-12-18
Send authenticated /versions request

dbkr:
Implements [MSC4026](https://github.com/matrix-org/matrix-spec-proposals/pull/4026).

I believe this probably is as simple as this: it will mean that the versions response can obviously change after logging in, but since the client is constructed again with an access token, this should just work (?)

A remaining question is whether this needs to be optional. Opening the PR to prompt the discussion. Apps might not expect it, but it's just the same auth that we're sending to other endpoints on the same server.

## Checklist

---

## ✨ Features
 * Send authenticated /versions request ([\#3968](https://github.com/matrix-org/matrix-js-sdk/pull/3968)).

### DIFF
--- a/spec/unit/matrix-client.spec.ts
+++ b/spec/unit/matrix-client.spec.ts
@@ -2211,8 +2211,7 @@ describe("MatrixClient", function () {
                     "org.matrix.msc3391": true,
                 },
             };
-            jest.spyOn(client.http, "request").mockResolvedValue(versionsResponse);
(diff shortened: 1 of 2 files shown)


t3chguy:
Feels like we should either make the accessToken of a client immutable or re-run this request when the token is changed, for clients which do not treat it as immutable.

dbkr:
Re-running it seems sensible - done.

richvdh (approved):

Issue #3915, opened 2023-11-28
Check for version is not allowing newer versions than the minimum version


MTRNord:
The check at: https://github.com/matrix-org/matrix-js-sdk/blob/a7496627fcc327601a2973497064ccd8fa32263d/src/autodiscovery.ts#L220 strictly requires a version to be in the response. However defining a newer version only according to https://spec.matrix.org/v1.8/client-server-api/#get_matrixclientversions is allowed to be set alone. And by nature of versions also afaik includes the older versions too.

TLDR: Shouldnt this be a greater than rather than a "includes" check? The spec doesnt mandate a server to include any prior version.

See also https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$VA5KRc9PIeXmk0Sot2AAmU7UW6BrlcRkTGm9XROR-lA?via=matrix.org&via=element.io&via=envs.net for a discussion on this in the spec room. Nheko for example seems to check 1.1-1.5 as a range check instead.

MTRNord:
https://github.com/search?q=repo%3Amatrix-org%2Fmatrix-js-sdk+%2Fv1%5C.%5Cd%2F+path%3Asrc%2F&type=code as linked by @clokep:matrix.org in the spec room suggests that js-sdk actually here wants to check for minimum support of 1.1. Not for "including and more".

t3chguy:
> However defining a newer version only according to [spec.matrix.org/v1.8/client-server-api/#get_matrixclientversions](https://spec.matrix.org/v1.8/client-server-api/#get_matrixclientversions) is allowed to be set alone.

Where does it say that?

MTRNord:
> > However defining a newer version only according to [spec.matrix.org/v1.8/client-server-api/#get_matrixclientversions](https://spec.matrix.org/v1.8/client-server-api/#get_matrixclientversions) is allowed to be set alone.
> 
> Where does it say that?

It doesnt explicitly say that. It however also doesnt require that you have to list the older versions. So this is practically open for interpretation.

With:

> Note that while v1.2 is meant to be backwards compatible with v1.1, it is not guaranteed that future versions will be fully backwards compatible with v1.1. For example, if /test were to be introduced in v1.1 and deprecated in v1.2, then it can be removed in v1.3. More information about this is described in the [deprecation policy](https://spec.matrix.org/v1.8/#deprecation-policy) below.

[...]

PR #3970, opened 2023-12-20
Bump minimum spec version to v1.5

richvdh:
We only claim to support a year of spec releases, so this should be fine.

Part of: https://github.com/matrix-org/matrix-js-sdk/issues/3915

---

## 🚨 BREAKING CHANGES
 * Bump minimum spec version to v1.5 ([\#3970](https://github.com/matrix-org/matrix-js-sdk/pull/3970)).

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 This is the [Matrix](https://matrix.org) Client-Server SDK for JavaScript and TypeScript. This SDK can be run in a
 browser or in Node.js.
 
-#### Minimum Matrix server version: v1.1
+#### Minimum Matrix server version: v1.5
 
 The Matrix specification is constantly evolving - while this SDK aims for maximum backwards compatibility, it only
 guarantees that a feature will be supported for at least 4 spec releases. For example, if a feature the js-sdk supports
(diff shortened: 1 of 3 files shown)


turt2live:
ftr, the js-sdk could probably go as far as v1.6 if it wanted to. By supporting v1.5, there's already a few servers out there which would no longer be supported - going to v1.6 doesn't appear to change that.

richvdh:
So the reason I went with 1.5 is that 1.6 wasn't released until 14 February 2023. So if we went with 1.6, we wouldn't actually be supporting the spec from a year ago. (We could bump to 1.6 as soon as 14 February 2024 arrives.)

turt2live:
Ah, the readme says "last 4", which is meant to be roughly 1 year. Considering 1.9 *just* came out though, it's hard to consider it one of the 4 indeed.

t3chguy (changes requested):
We should also drop any (now dead) code which specifically handles pre-v1.5 - hopefully that includes a lot of the threads workarounds

richvdh:
> We should also drop any (now dead) code which specifically handles pre-v1.5 - hopefully that includes a lot of the threads workarounds

Agreed, but that can be done later. It's just tech debt. There's no need to block this change on that work.

t3chguy (approved):
Sure assuming an issue to say such materialises :)

richvdh:
Issue filed at https://github.com/matrix-org/matrix-js-sdk/issues/3972

richvdh:
This will need https://github.com/matrix-org/matrix-react-sdk/pull/12081 and https://github.com/element-hq/element-web/pull/26796 to land first.

t3chguy:
And probably wants to be a major version bump

richvdh:
For the record: this got reverted in #4014.

PR #3973, opened 2023-12-22
ElementR: Ensure Encryption order per room

BillCarsonFr:
Fixes https://github.com/element-hq/element-web/issues/26684

Encryption order was not ensured with the existing code. So in some case editing an inflight message would encrypt the edit first and schedule it first. That will cause an error because it will send an edit of an unknown event.
It can also invert the message order.

There is a sheduler currently but it doesn't handle encryption, it only queues the event for sending.
The scheduler should probably handle the encryption, but for now I modified RoomEncryptor to ensure order.

Notice that there was already a ordering mechanism but only to ensure order of `olmMachine.shareRoomKey` but it was not enough.

Added a test to reproduce the problem if the fix is not present.

## Checklist

---

## 🐛 Bug Fixes
 * ElementR: Ensure Encryption order per room ([\#3973](https://github.com/matrix-org/matrix-js-sdk/pull/3973)). Fixes element-hq/element-web#26684.

### DIFF
--- a/spec/unit/rust-crypto/RoomEncryptor.spec.ts
+++ b/spec/unit/rust-crypto/RoomEncryptor.spec.ts
@@ -144,7 +144,7 @@ describe("RoomEncryptor", () => {
 
             mockOlmMachine.shareRoomKey.mockResolvedValue(undefined);
 
-            // Took this method to demonstrate the race condition
+            // Hook into this method to demonstrate the race condition
             mockRoom.getEncryptionTargetMembers
                 .mockImplementationOnce(async () => {
                     await firstTargetMembers.promise;
--- a/spec/unit/rust-crypto/RoomEncryptor.spec.ts
+++ b/spec/unit/rust-crypto/RoomEncryptor.spec.ts
@@ -186,7 +186,6 @@ describe("RoomEncryptor", () => {
             await Promise.all([firstRequest, secondRequest]);
 
             expect(firstMessageFinished).toBe("hello");
-
         });
     });
 });
--- a/spec/unit/rust-crypto/RoomEncryptor.spec.ts
+++ b/spec/unit/rust-crypto/RoomEncryptor.spec.ts
@@ -70,7 +70,6 @@ describe("RoomEncryptor", () => {
         }
 
         beforeEach(() => {
-            jest.useFakeTimers();
             mockOlmMachine = {
                 identityKeys: {
                     curve25519: {
@@ -109,10 +108,6 @@ describe("RoomEncryptor", () => {
                 mockRoom,
                 { algorithm: "m.megolm.v1.aes-sha2" },
             );
-        });
-
-        afterEach(() => {
-            jest.useRealTimers();
         });
 
         it("should ensure that there is only one shareRoomKey at a time", async () => {
@@ -178,10 +173,7 @@ describe("RoomEncryptor", () => {
 
             // suppose the second getEncryptionTargetMembers call returns first
             secondTargetMembers.resolve();
-            await jest.runAllTimersAsync();
-
             firstTargetMembers.resolve();
-            await jest.runAllTimersAsync();
 
             await Promise.all([firstRequest, secondRequest]);
 
--- a/spec/unit/rust-crypto/RoomEncryptor.spec.ts
+++ b/spec/unit/rust-crypto/RoomEncryptor.spec.ts
@@ -137,6 +137,7 @@ describe("RoomEncryptor", () => {
             expect(mockOlmMachine.shareRoomKey).toHaveBeenCalledTimes(6);
         });
 
+        // Regression test for https://github.com/element-hq/element-web/issues/26684
         it("Should maintain order of encryption requests", async () => {
             const firstTargetMembers = defer<void>();
             const secondTargetMembers = defer<void>();
--- a/spec/unit/rust-crypto/RoomEncryptor.spec.ts
+++ b/spec/unit/rust-crypto/RoomEncryptor.spec.ts
@@ -52,6 +52,11 @@ describe("RoomEncryptor", () => {
         let mockOutgoingRequestManager: Mocked<OutgoingRequestsManager>;
         let mockRoom: Mocked<Room>;
 
+        const mockRoomMember = {
+            userId: "@alice:example.org",
+            membership: "join",
+        } as unknown as Mocked<RoomMember>;
+
         function createMockEvent(text: string): Mocked<MatrixEvent> {
             return {
                 getTxnId: jest.fn().mockReturnValue(""),
@@ -65,6 +70,7 @@ describe("RoomEncryptor", () => {
         }
 
         beforeEach(() => {
+            jest.useFakeTimers();
             mockOlmMachine = {
                 identityKeys: {
                     curve25519: {
@@ -87,11 +93,6 @@ describe("RoomEncryptor", () => {
                 doProcessOutgoingRequests: jest.fn().mockResolvedValue(undefined),
             } as unknown as Mocked<OutgoingRequestsManager>;
 
-            const mockRoomMember = {
-                userId: "@alice:example.org",
-                membership: "join",
-            } as unknown as Mocked<RoomMember>;
-
             mockRoom = {
                 roomId: "!foo:example.org",
                 getJoinedMembers: jest.fn().mockReturnValue([mockRoomMember]),
@@ -108,6 +109,10 @@ describe("RoomEncryptor", () => {
                 mockRoom,
                 { algorithm: "m.megolm.v1.aes-sha2" },
             );
+        });
+
+        afterEach(() => {
+            jest.useRealTimers();
         });
 
         it("should ensure that there is only one shareRoomKey at a time", async () => {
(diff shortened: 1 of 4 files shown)


richvdh (changes requested):
as discussed elsewhere: I don't think this does the right thing

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -223,50 +257,45 @@ export class RoomEncryptor {
...
+        const prom = this.currentEncryptionPromise
+            .catch(() => {
+                // any errors in the previous claim will have been reported already, so there is nothing to do here.
+                // we just throw away the error and start anew.
+            })
+            .then(() => {
+                if (operation.event) {
+                    return this.encryptEventInner(operation.event, operation.globalBlacklistUnverifiedDevices);
+                } else {
+                    return this.prepareForEncryption(operation.globalBlacklistUnverifiedDevices);
```
richvdh, line 290: won't this recurse infinitely?

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -46,8 +46,11 @@ export class RoomEncryptor {
     /** whether the room members have been loaded and tracked for the first time */
     private lazyLoadedMembersResolved = false;
 
-    /** Ensures that there is only one call to shareRoomKeys at a time */
-    private currentShareRoomKeyPromise = Promise.resolve();
+    /**
+     * Ensures that there is only one encryption operation at a time for that room
+     * An encryption operation is either a prepareForEncryption or an encryptEvent call.
```
richvdh, line 51: Nits:

```suggestion
     * Ensures that there is only one encryption operation at a time for that room.
     *
     * An encryption operation is either a  {@link #prepareForEncryption} or an {@link #encryptEvent} call.
```

As well as improving the formatting, `@link` ensures that if the methods ever get renamed, we don't get left with a dangling reference (because the linter will catch it).

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -118,11 +121,45 @@ export class RoomEncryptor {
      * @param globalBlacklistUnverifiedDevices - When `true`, it will not send encrypted messages to unverified devices
      */
     public async prepareForEncryption(globalBlacklistUnverifiedDevices: boolean): Promise<void> {
-        const logger = new LogSpan(this.prefixedLogger, "prepareForEncryption");
+        // We consider a prepareForEncryption as an encryption promise as it will potentially share keys
+        // even if it doesn't send an event.
+        // Usually this is called when the user starts typing, so we want to make sure we have keys ready when the
+        // message is finally sent. The actual event encryption request will arrive after wait for the prepareForEncryption
+        // promise to resolve, and then do again an ensureEncryptionSession that should be no op as we already share the room key.
```
richvdh, line 128: > The actual event encryption request will arrive after wait for the prepareForEncryption promise to resolve

I don't really understand this, or what it is trying to say. "The actual event encryption request" == "a call to `encryptEvent`"?  So you're saying that the application will wait for `prepareForEncryption` to complete before it calls `encryptEvent`? I don't think that's right.

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -118,11 +121,45 @@ export class RoomEncryptor {
...
+        await this.encryptEvent(null, globalBlacklistUnverifiedDevices);
+    }
 
-        await logDuration(this.prefixedLogger, "prepareForEncryption", async () => {
-            await this.ensureEncryptionSession(logger, globalBlacklistUnverifiedDevices);
-        });
+    /**
+     * Encrypt an event for this room.
```
richvdh, line 133:
```suggestion
     * Encrypt an event for this room, or prepare for encryption.
```

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -118,11 +121,45 @@ export class RoomEncryptor {
...
+     *
+     * This will ensure that we have a megolm session for this room, share it with the devices in the room, and
+     * then encrypt the event using the session.
```
richvdh, line 136:
```suggestion
     * This will ensure that we have a megolm session for this room, share it with the devices in the room, and
     * then, if an event is provided, encrypt it using the session.
```

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -118,11 +121,45 @@ export class RoomEncryptor {
...
+     *
+     * @param event - Event to be encrypted or null if only preparing for encryption (pre share room key)
```
richvdh, line 138:
```suggestion
     * @param event - Event to be encrypted, or null if only preparing for encryption (in which case we will pre-share the room key).
```

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -118,11 +121,45 @@ export class RoomEncryptor {
...
+     * @param globalBlacklistUnverifiedDevices - When `true`, it will not send encrypted messages to unverified devices
+     */
+    public encryptEvent(event: MatrixEvent | null, globalBlacklistUnverifiedDevices: boolean): Promise<void> {
+        // Ensure order of encryption to avoid message ordering issues, as the scheduler only ensures
+        // events order after they have been encrypted.
```
richvdh, line 143: this should be *after* the call to `new LogSpan`, since it relates to the code after it.

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -118,11 +121,45 @@ export class RoomEncryptor {
...
+        const logger = new LogSpan(this.prefixedLogger, event ? event.getTxnId() ?? "" : "prepareForEncryption");
+        const prom = this.currentEncryptionPromise
+            .catch(() => {
+                // any errors in the previous claim will have been reported already, so there is nothing to do here.
```
richvdh, line 147:
```suggestion
                // Any errors in the previous call will have been reported already, so there is nothing to do here.
```

spec/unit/rust-crypto/RoomEncryptor.spec.ts
```diff
@@ -136,5 +136,48 @@ describe("RoomEncryptor", () => {
...
             expect(mockOlmMachine.shareRoomKey).toHaveBeenCalledTimes(6);
         });
+
+        it("Should maintain order of encryption requests", async () => {
+            const firstTargetMembers = defer<void>();
+            const secondTargetMembers = defer<void>();
+
+            mockOlmMachine.shareRoomKey.mockResolvedValue(undefined);
+
+            // Took this method to demonstrate the race condition
```
richvdh, line 146:
```suggestion
            // Hook into this method to demonstrate the race condition
```

spec/unit/rust-crypto/RoomEncryptor.spec.ts
```diff
@@ -136,5 +136,48 @@ describe("RoomEncryptor", () => {
             // should have been called again
             expect(mockOlmMachine.shareRoomKey).toHaveBeenCalledTimes(6);
         });
+
+        it("Should maintain order of encryption requests", async () => {
```
richvdh, line 140:
```suggestion
        // Regression test for https://github.com/element-hq/element-web/issues/26684
        it("Should maintain order of encryption requests", async () => {
```

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -223,53 +264,20 @@ export class RoomEncryptor {
...
-        return prom;
-    }
-
     /**
      * Discard any existing group session for this room
      */
     public async forceDiscardSession(): Promise<void> {
+        // XXX Clarify if also need to ensure order for that call to.
+        // It will mark the session as invalidated but not delete it, so if the current encryption promise
+        // has already check for rotation it will not be affected.
```
richvdh, line 273: Pretty sure it doesn't matter.

This is only called if someone does `/discardsession` in Element. If you do that while there are messages queued, then yes, there's a possibility that some of the messages in the queue will go to the old session. But that's ok?

richvdh (changes requested):
LGTM mostly. A few suggestions, nits, etc.

src/rust-crypto/RoomEncryptor.ts
BillCarsonFr: I think it's how it's supposed to work. Legacy is doing the same.
If you starts typing then call encrypt event, we let the prepare finish, then do the encrypt. The share room key part for the secong call should be no op as the room key is already shared

BillCarsonFr: Notice that anyhow the olmMachine.shareRoomKey() should be locked per room

richvdh: > if you starts typing then call encrypt event, we let the prepare finish, then do the encrypt.

but that's not what the comment says?

richvdh: ... or rather the comment says that we expect the application to wait, and I don't think we do. That's why we have to wait here.

BillCarsonFr: I updated the comment:
```
// If `encryptEvent` is invoked before `prepareForEncryption` has completed, the `encryptEvent` call will wait for
// `prepareForEncryption` to complete before executing.
// The part where `encryptEvent` shares the room key would be no op as it was already performed by prepareForEncryption.

File: src/rust-crypto/RoomEncryptor.ts [hunk 12]
```diff
@@ -47,8 +47,9 @@ export class RoomEncryptor {
     private lazyLoadedMembersResolved = false;
 
     /**
-     * Ensures that there is only one encryption operation at a time for that room
-     * An encryption operation is either a prepareForEncryption or an encryptEvent call.
+     * Ensures that there is only one encryption operation at a time for that room.
+     *
+     * An encryption operation is either a  {@link prepareForEncryption} or an {@link encryptEvent} call.
```
richvdh, line 52: Nit:
```suggestion
     * An encryption operation is either a {@link prepareForEncryption} or an {@link encryptEvent} call.
```

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -124,27 +125,29 @@ export class RoomEncryptor {
         // We consider a prepareForEncryption as an encryption promise as it will potentially share keys
         // even if it doesn't send an event.
         // Usually this is called when the user starts typing, so we want to make sure we have keys ready when the
-        // message is finally sent. The actual event encryption request will arrive after wait for the prepareForEncryption
-        // promise to resolve, and then do again an ensureEncryptionSession that should be no op as we already share the room key.
+        // message is finally sent.
+        // If `encryptEvent` is invoked before `prepareForEncryption` has completed, the `encryptEvent` call will wait for
+        // `prepareForEncryption` to complete before executing.
+        // The part where `encryptEvent` shares the room key would be no op as it was already performed by prepareForEncryption.
```
richvdh, line 131:
```suggestion
        // The part where `encryptEvent` shares the room key will then usually be a no-op as it was already performed by `prepareForEncryption`.
```

(It's not quite a no-op: in particular, if any servers are down during `prepareForEncryption`, we'll try to reach them again during `encryptEvent`. Or the device list could change.)

richvdh (approved):
A couple of nits, lgtm otherwise

PR #3975, opened 2023-12-24
drop 'malformed member event' log level, given all TR rooms are malformed like this.

ara4n:
---


toger5:
Is TR even compatible with this version of the JS SDK?
I guess this PR comes from trying it and realizing its very noisy.
Or is this logging one can see when openingn a TR in EW?

toger5 (approved):
A debug log makes more sense since both are not specced or offically deprecated which would justify a warning.

PR #3978, opened 2023-12-29
Support for migration from from libolm

### LINKED ISSUE #3964: Rust crypto: support migration to Rust crypto
Part of  support for [migration to Rust Crypto](https://github.com/element-hq/element-web/issues/26678). Depends on https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/78 and https://github.com/matrix-org/matrix-js-sdk/issues/3963.

 Add a new `migrateFromLegacyCrypto(legacyStore)` method to `js-sdk/src/rust-crypto`, and call it in `MatrixClient.initRustCrypto`, before calling `RustCrypto.initRustCrypto`. First, it needs to check if there might be data to migrate:
   * Call `CryptoStore.containsData` and skip out if no data.
   * Call `CryptoStore.startup`.
   * Call `CryptoStore.getMigrationState` and skip any steps below that are done. Otherwise:
   * Extract the following and send to the rust layer:
      * (Pickled) account.
      * Private cross-signing keys (via `CryptoStore.getSecretStorePrivateKey("master")` etc)
      * Encrypted rooms (via `CryptoStore.getEndToEndRooms`)
      * Backup decryption key (via `CryptoStore.getSecretStorePrivateKey("m.megolm_backup.v1")`)
      * Backup version: we'll have to fetch this from the server (and check the key matches the above).
      Afterwards, call  `CryptoStore.setMigrationState` to flag that this initial import has been done.
   *  Then, migrate the Olm and Megolm sessions, in batches.

We need to emit some sort of progress indication, so that the app can give feedback. Probably something as simple as a `crypto.dataMigrationProgress` event, whose payload includes two numbers including a total number of steps, and a number of steps so far. We can wire this into the `progress_listener` callback of the rust migration functions.

richvdh:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/3964.

---

## ✨ Features
 * Support for migration from from libolm ([\#3978](https://github.com/matrix-org/matrix-js-sdk/pull/3978)). Fixes #3964.


spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -73,16 +85,22 @@ describe("initRustCrypto", () => {
...
         } as unknown as Mocked<OlmMachine>;
     }
 
     it("passes through the store params", async () => {
+        const mockStore = { free: jest.fn() } as unknown as StoreHandle;
+        jest.spyOn(StoreHandle, "open").mockResolvedValue(mockStore);
+
         const testOlmMachine = makeTestOlmMachine();
-        jest.spyOn(OlmMachine, "initialize").mockResolvedValue(testOlmMachine);
+        jest.spyOn(OlmMachine, "init_from_store").mockResolvedValue(testOlmMachine);
```
BillCarsonFr, line 103: Surprised to see snake case. A missing jsName in wasm bindings?

BillCarsonFr (approved):
LGTM

spec/unit/rust-crypto/rust-crypto.spec.ts
richvdh: yeah :(. Unfortunately fixing it will mean a new release of the bindings. I'll leave it for now and fix it in a future version of the bindings.

src/crypto/store/base.ts
```diff
@@ -223,6 +223,11 @@ export interface ISession {
     sessionData?: InboundGroupSessionData;
 }
 
+/** Extended data on a Megolm session */
+export interface ISessionExtended extends ISession {
```
t3chguy, line 227: We don't prefix with I anymore

```suggestion
export interface SessionExtended extends ISession {
```

t3chguy (approved):
Looks sane otherwise

PR #3979, opened 2023-12-29
Use enums instead of strings for client event listeners (`client.on`, `client.once` ...) in the README to improve type inference

larskarbo:
I noticed that when using strings in the event names in the client listeners (`client.on`, `client.once`), the TypeScript types don't work. When using the enums, types work wonderfully.

This PR changes the examples in the readme to use enums (like `RoomEvent.Timeline`) instead of strings (like `"Room.timeline"`).

## Before

❌ Types in the argument do NOT work here. `event`, `room`, and `toStartOfTimeline` default to `any`

```typescript
client.on("Room.timeline", function (event, room, toStartOfTimeline) {
  ...
```

## After

✅ Types work well. `event`, `room`, and `toStartOfTimeline` have proper types.

```typescript
client.on(RoomEvent.Timeline, function (event, room, toStartOfTimeline) {
  ...
```
----

I understand the simplicity of using strings, but having proper types is important for onboarding new users.

An alternative would be to fix the type inference when using strings.

Signed-off-by: Lars Karbo <10865165+larskarbo@users.noreply.github.com>

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ await client.startClient({ initialSyncLimit: 10 });
 You can perform a call to `/sync` to get the current state of the client:
 
 ```javascript
-client.once("sync", function (state, prevState, res) {
+client.once(ClientEvent.sync, function (state, prevState, res) {
     if (state === "PREPARED") {
         console.log("prepared");
     } else {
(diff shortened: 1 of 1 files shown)


richvdh (approved):
looks good to me, thanks!

PR #3982, opened 2024-01-02
Emit events during migration from libolm

richvdh:
... so that we can display progress feedback in the UI.

part of https://github.com/element-hq/element-web/issues/26773

---

## ✨ Features
 * Emit events during migration from libolm ([\#3982](https://github.com/matrix-org/matrix-js-sdk/pull/3982)).

### DIFF
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -193,6 +193,10 @@ describe("initRustCrypto", () => {
 
             fetchMock.get("path:/_matrix/client/v3/room_keys/version", { version: "45" });
 
+            function legacyMigrationProgressListener(progress: number, total: number): void {
+                logger.log(`migrated ${progress} of ${total}`);
+            }
+
             await initRustCrypto({
                 logger,
                 http: makeMatrixHttpApi(),
@@ -204,6 +208,7 @@ describe("initRustCrypto", () => {
                 storePassphrase: "storePassphrase",
                 legacyCryptoStore: legacyStore,
                 legacyPickleKey: PICKLE_KEY,
+                legacyMigrationProgressListener,
             });
 
             // Check that the migration functions were correctly called
--- a/src/crypto/store/memory-crypto-store.ts
+++ b/src/crypto/store/memory-crypto-store.ts
@@ -533,6 +533,17 @@ export class MemoryCryptoStore implements CryptoStore {
     }
 
     /**
+     * Count the number of Megolm sessions in the database.
+     *
+     * Implementation of {@link CryptoStore.countEndToEndInboundGroupSessions}.
+     *
+     * @internal
+     */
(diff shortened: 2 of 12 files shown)


florianduros (approved):
Looks good

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -193,6 +193,10 @@ describe("initRustCrypto", () => {
 
             fetchMock.get("path:/_matrix/client/v3/room_keys/version", { version: "45" });
 
+            function legacyMigrationProgressListener(progress: number, total: number): void {
+                logger.log(`migrated ${progress} of ${total}`);
```
andybalaam, line 197: Should we be logging this in a unit test?

src/crypto/store/memory-crypto-store.ts
```diff
@@ -336,7 +336,11 @@ export class MemoryCryptoStore implements CryptoStore {
     // Olm Sessions
 
     public countEndToEndSessions(txn: unknown, func: (count: number) => void): void {
```
andybalaam, line 338: I'd love an explanation of why this takes `txn` as an argument sometime :-)

andybalaam (approved):
Looks good as far as I understand this stuff.

spec/unit/rust-crypto/rust-crypto.spec.ts
richvdh: I figured it was fairly harmless, and possibly useful when debugging the test. As far as I know there is no particular embargo against logging in tests?

src/crypto/store/memory-crypto-store.ts
richvdh: ugh. Because it wants to expose the same interface as `IndexeddbCryptoStore`. Which, for *some* methods, supports a transaction-based API.

spec/unit/rust-crypto/rust-crypto.spec.ts
andybalaam: I find unit test output hard to understand because of all the logging. My understanding was/is that well-behaved tests would avoid logging, and we have `filterConsole` in matrix-react-sdk as an attempt to document expected console messages without filling the console with spam.

Or is use of `logger` better than `console` in some way? Maybe it doesn't fill the console?

My expectation might be wrong anyway. I just don't think I've seen tests deliberately logging before and wanted to check you meant it.

richvdh: `logger` is no better than `console` as far as I know.

You may be right that the amount of spam from tests is confusing; however this is far from the only thing that gets written.

I kinda wish that jest had better facilities for handling console output from tests. For example, it might default to only showing the log output from failing tests.

andybalaam: Agreed.

PR #3987, opened 2024-01-05
fix IndexedDBStore API documentation

### LINKED ISSUE #3986: IndexedDBStore: createUser is undefined, it should be set with setUserCreator when upgrading 29.1.0 -> 30.3.0
error:

```
Failed initializing matrix:  Error: createUser is undefined, it should be set with setUserCreator()!
    startup http://localhost:8080/node_modules/.vite/deps/chunk-AUCUNUAW.js?v=96c31d21:167359
    startup http://localhost:8080/node_modules/.vite/deps/chunk-AUCUNUAW.js?v=96c31d21:167357
    promise callback*startup http://localhost:8080/node_modules/.vite/deps/chunk-AUCUNUAW.js?v=96c31d21:167355
    startClient http://localhost:8080/src/client/initMatrix.js:42
```

API call:

```js
    const indexedDBStore = new sdk.IndexedDBStore({
      indexedDB: global.indexedDB,
      localStorage: global.localStorage,
      dbName: 'web-sync-store',
    });
    await indexedDBStore.startup(); // <-- fails here

    this.matrixClient = sdk.createClient({
      baseUrl: secret.baseUrl,
      accessToken: secret.accessToken,
      userId: secret.userId,
      store: indexedDBStore,
      cryptoStore: new sdk.IndexedDBCryptoStore(global.indexedDB, 'crypto-store'),
      deviceId: secret.deviceId,
      timelineSupport: true,
      cryptoCallbacks,
      verificationMethods: [
        'm.sas.v1',
      ],
    });
```

the creation is almost exactly like what the API docs are suggesting to use, so I don't know what's going wrong here:

```js
let opts = { indexedDB: window.indexedDB, localStorage: window.localStorage };
let store = new IndexedDBStore(opts);
await store.startup(); // load from indexed db
```

WebFreak001:
Retroactively changes the changelog entry to include since when this change is needed as well, so that other people upgrading the SDK can more easily find what to change.

Fix #3986

Only documentation & error strings have changed so I have not adjusted any tests. A proper integration test would probably be nice, but I didn't have the time to look into how that is setup on this project.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ Changes in [30.0.0](https://github.com/matrix-org/matrix-js-sdk/releases/tag/v30
 
 ## 🚨 BREAKING CHANGES
  * Refactor & make base64 functions browser-safe ([\#3818](https://github.com/matrix-org/matrix-js-sdk/pull/3818)).
+ * `IndexedDBStore.startup()` must be called after using it on `sdk.createClient` now.
 
 ## 🦖 Deprecations
  * Deprecate `MatrixEvent.toJSON` ([\#3801](https://github.com/matrix-org/matrix-js-sdk/pull/3801)).
--- a/src/store/indexeddb.ts
+++ b/src/store/indexeddb.ts
@@ -90,10 +90,10 @@ export class IndexedDBStore extends MemoryStore {
      * ```
      * let opts = { indexedDB: window.indexedDB, localStorage: window.localStorage };
      * let store = new IndexedDBStore(opts);
-     * await store.startup(); // load from indexed db
      * let client = sdk.createClient({
      *     store: store,
      * });
+     * await store.startup(); // load from indexed db, must be called after createClient
      * client.startClient();
      * client.on("sync", function(state, prevState, data) {
      *     if (state === "PREPARED") {
@@ -140,7 +140,9 @@ export class IndexedDBStore extends MemoryStore {
                 logger.log(`IndexedDBStore.startup: processing presence events`);
                 userPresenceEvents.forEach(([userId, rawEvent]) => {
                     if (!this.createUser) {
-                        throw new Error("createUser is undefined, it should be set with setUserCreator()!");
+                        throw new Error(
+                            "`IndexedDBStore.startup` must be called after assigning it to the client, not before!",
+                        );
                     }
                     const u = this.createUser(userId);
                     if (rawEvent) {


CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ Changes in [30.0.0](https://github.com/matrix-org/matrix-js-sdk/releases/tag/v30
 
 ## 🚨 BREAKING CHANGES
  * Refactor & make base64 functions browser-safe ([\#3818](https://github.com/matrix-org/matrix-js-sdk/pull/3818)).
+ * * `IndexedDBStore.startup()` must be called after using it on `sdk.createClient` now.
```
richvdh, line 53:
```suggestion
 * `IndexedDBStore.startup()` must be called after using it on `sdk.createClient` now.
```

richvdh (approved):
Thank you!

WebFreak001:
um it looks like the changelog diff broke just now when I used the github edit feature - is there inconsistent line endings in the file causing the web editor diff to break?

richvdh:
> um it looks like the changelog diff broke just now when I used the github edit feature - is there inconsistent line endings in the file causing the web editor diff to break?

possibly. Not sure, I'm afraid

WebFreak001:
I just removed the changelog entry now, seems to be too much of a hassle to get the line endings right with git

richvdh:
> I just removed the changelog entry now, seems to be too much of a hassle to get the line endings right with git 

Wasn't the whole point of this PR to update the changelog?

> (also there doesn't seem to be a "unreleased" version where I would need to put the new entry for the upcoming version)

Take a look at the contributing doc. The changelog is autogenerated.

WebFreak001:
ah you are right sorry, I forgot what I actually did in this PR - I readded the changelog entry to the version where I put it earlier in the PR

richvdh (changes requested):
Thanks! A couple of final things, I'm afraid.

* It's failing CI due to a failure from the linter. It's probably complaining about line length, though I haven't checked in detail. Please either see what the matter with it is or run the changes through `prettier`.
* We require sign-off for all changes, even ones as simple as this. Sorry! Please take a look at https://github.com/element-hq/element-web/blob/develop/CONTRIBUTING.md#sign-off and if you are happy, add a "Signed-off-by" line to the PR description.

WebFreak001:
ok should be good now

richvdh:
Sorry, but as the contributing guide says, I'm afraid we don't accept contributions signed off under a pseudonym, so we'll need your real name to be able to accept this. See https://github.com/element-hq/element-web/blob/develop/CONTRIBUTING.md#private-sign-off.

richvdh (approved):
Thank you for your patience!

PR #4004, opened 2024-01-12
ElementR: fix emoji verification stalling when both ends hit start at the same time

uhoreg:
fixes https://github.com/element-hq/element-web/issues/26723

When both ends hit "start", one of the sides will "win" and the other will "lose".  The Rust crypto crate on the losing side replaces the SAS verifier with a new one.  The current code does not detect this, and continues trying to use the old one.  This PR switches the verifier to the new one when it changes.

## Checklist

---

## 🐛 Bug Fixes
 * ElementR: fix emoji verification stalling when both ends hit start at the same time ([\#4004](https://github.com/matrix-org/matrix-js-sdk/pull/4004)). Fixes element-hq/element-web#26723.


BillCarsonFr (commented):
Just a few questions

src/rust-crypto/verification.ts
```diff
@@ -658,6 +673,14 @@ export class RustSASVerifier extends BaseRustVerifer<RustSdkCryptoJs.Sas> implem
 
     /** if we can now show the callbacks, do so */
     protected onChange(): void {
+        if (this.started) {
```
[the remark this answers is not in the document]
uhoreg: Yes, if we already accepted with that Rust SAS struct, then `this.inner.accept()` will return `undefined`, so it will be a no-op.  I'll clarify in the comment.

src/rust-crypto/verification.ts
```diff
@@ -649,6 +663,7 @@ export class RustSASVerifier extends BaseRustVerifer<RustSdkCryptoJs.Sas> implem
      *    or times out.
      */
     public async verify(): Promise<void> {
+        this.started = true;
```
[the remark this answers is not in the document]
uhoreg: I moved the assignment to after the request is made, instead, since we don't really need it set right away.

BillCarsonFr (approved):

src/rust-crypto/verification.ts
```diff
@@ -75,11 +75,15 @@ export class RustVerificationRequest
         const onChange = async (): Promise<void> => {
             const verification: RustSdkCryptoJs.Qr | RustSdkCryptoJs.Sas | undefined = this.inner.getVerification();
 
-            // If we now have a `Verification` where we lacked one before, or we have transitioned from QR to SAS,
-            // wrap the new rust Verification as a js-sdk Verifier.
             if (verification instanceof RustSdkCryptoJs.Sas) {
                 if (this._verifier === undefined || this._verifier instanceof RustQrCodeVerifier) {
+                    // If we now have a `Verification` where we lacked one before, or we have transitioned from QR to SAS,
+                    // wrap the new rust Verification as a js-sdk Verifier.
```
richvdh, line 81: There are two branches to this `if` statement (`if (verification instanceof RustSdkCryptoJs.Sas)`, and `else if (verification instanceof RustSdkCryptoJs.Qr && this._verifier === undefined)`), and this comment applies to both of them. By moving it in here, it's somewhat less clear.

src/rust-crypto/verification.ts
```diff
@@ -653,11 +667,22 @@ export class RustSASVerifier extends BaseRustVerifer<RustSdkCryptoJs.Sas> implem
         if (req) {
             await this.outgoingRequestProcessor.makeOutgoingRequest(req);
         }
+        this.started = true;
```
richvdh, line 670: It might be better to call this `accepted`, since the bit we care about is calling `inner.accept()`? It's confusing either way (since it could mean sending either an `m.k.v.start` *or* an `m.k.v.accept`), but I think `accepted` would be slightly less confusing?

(also, it's not the same as `VerificationPhase.Started`)

src/rust-crypto/verification.ts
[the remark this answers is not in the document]
richvdh: can we not intercept `setInner` and just do this logic then?

src/rust-crypto/verification.ts
```diff
@@ -450,15 +454,16 @@ abstract class BaseRustVerifer<InnerType extends RustSdkCryptoJs.Qr | RustSdkCry
...
     public constructor(
-        protected readonly inner: InnerType,
+        protected inner: InnerType,
         protected readonly outgoingRequestProcessor: OutgoingRequestProcessor,
     ) {
         super();
 
         this.completionPromise = new Promise<void>((resolve, reject) => {
-            const onChange = async (): Promise<void> => {
+            const onChange = (this.onChangeCallback = async (): Promise<void> => {
```
richvdh, line 466: get rid of `onChange` and just use `this.onChangeCallback` below?

src/rust-crypto/verification.ts
```diff
@@ -450,15 +454,16 @@ abstract class BaseRustVerifer<InnerType extends RustSdkCryptoJs.Qr | RustSdkCry
 > {
     /** A promise which completes when the verification completes (or rejects when it is cancelled/fails) */
     protected readonly completionPromise: Promise<void>;
+    private onChangeCallback: (() => Promise<void>) | undefined = undefined;
```
richvdh, line 457: Given we know that `onChangeCallback` will be set in the constructor, I think there's a way to tell typescript that.

```suggestion
    private onChangeCallback!: () => Promise<void>;
```

or something to that effect, IIRC

src/rust-crypto/verification.ts
richvdh, line 670: set this *before* `await this.outgoingRequestProcessor.makeOutgoingRequest(req);`, in case the new `Verification` arrives while we're sending the `m.key.verification.start`? Not sure if that's possible, but moving the assignment would save me worrying about it.

src/rust-crypto/verification.ts
```diff
@@ -75,11 +75,15 @@ export class RustVerificationRequest
...
             if (verification instanceof RustSdkCryptoJs.Sas) {
                 if (this._verifier === undefined || this._verifier instanceof RustQrCodeVerifier) {
+                    // If we now have a `Verification` where we lacked one before, or we have transitioned from QR to SAS,
+                    // wrap the new rust Verification as a js-sdk Verifier.
                     this.setVerifier(new RustSASVerifier(verification, this, outgoingRequestProcessor));
+                } else if (this._verifier instanceof RustSASVerifier) {
+                    // We may get a new rust `Verification` if both parties try to start the verification at the same time,
+                    // but we lose the tie breaking.
+                    this._verifier.setInner(verification);
                 }
```
richvdh, line 87: Did you consider replacing the `RustSASVerifier`, rather than updating the existing one? When I designed this, I was rather hoping to avoid having to do this sort of switcheroo, and it feels like a replacement might be simpler.

If it's a case of documenting that clients aren't allowed to hold onto a handle to the verifier, that would be ok I think.

But there might be more fundamental reasons that won't work.

src/rust-crypto/verification.ts
```diff
@@ -475,13 +480,21 @@ abstract class BaseRustVerifer<InnerType extends RustSdkCryptoJs.Qr | RustSdkCry
...
                 this.emit(VerificationRequestEvent.Change);
-            };
+            });
             inner.registerChangesCallback(onChange);
         });
         // stop the runtime complaining if nobody catches a failure
         this.completionPromise.catch(() => null);
     }
 
+    public setInner(inner: InnerType): void {
```
richvdh, line 490: given this is `public` (if only in this file), would be good to document it.

richvdh (changes requested):

src/rust-crypto/verification.ts
uhoreg: I think the client can call `.verify()` on the `RustSASVerifier`, which is supposed to return once the verification succeeds, but that won't work if a new `RustSASVerifier` is used.

src/rust-crypto/verification.ts
uhoreg: done

src/rust-crypto/verification.ts
uhoreg: Moved the comment back and modified it

src/rust-crypto/verification.ts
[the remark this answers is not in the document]
uhoreg: added an `onSetInner` function that gets called when `setInner` actually changes the `inner`.

src/rust-crypto/verification.ts
uhoreg: I originally had it before, but moved it to after, in response to https://github.com/matrix-org/matrix-js-sdk/pull/4004#discussion_r1462921558  I've now moved it back to before, but wrapped the acceptance in a try-catch to address the original comment.

However, thinking about this more, I don't think we even need the ~`.started`~ `.accepted` variable.  We will only ever call `.setInner` if we started a verification, but lost the tie breaker.  So in that case, we know we definitely do want to send an acceptance.  Does that sound right?

src/rust-crypto/verification.ts
```diff
@@ -664,26 +677,28 @@ export class RustSASVerifier extends BaseRustVerifer<RustSdkCryptoJs.Sas> implem
...
+            await this.sendAccept();
+        } catch (e) {
+            this.accepted = false;
+            throw e;
+        }
+        await this.completionPromise;
+    }
+
+    /**
+     * Send the accept event, if it hasn't already been sent
```
richvdh, line 691:
```suggestion
     * Send the accept or start event, if it hasn't already been sent
```

src/rust-crypto/verification.ts
richvdh: yeah, that sounds right.

(It would be nice if we could introspect the old `RustSdkCryptoJs.Sas` to double-check that it thinks we have `accept`ed, but I that doesn't seem to be exposed in the API, so let's just assume it.)

src/rust-crypto/verification.ts
```diff
@@ -477,19 +483,39 @@ abstract class BaseRustVerifer<InnerType extends RustSdkCryptoJs.Qr | RustSdkCry
...
         this.completionPromise.catch(() => null);
     }
 
+    /**
+     * Replace the inner Rust verifier with a different one.
+     *
+     * @param inner - the new Rust verifier
+     * @internal
+     */
+    public setInner(inner: InnerType): void {
```
richvdh, line 498: this should maybe be `replaceInner` rather than `setInner`, but that is somewhat gold-plating

src/rust-crypto/verification.ts
```diff
@@ -451,15 +456,16 @@ abstract class BaseRustVerifer<InnerType extends RustSdkCryptoJs.Qr | RustSdkCry
...
     public constructor(
-        protected readonly inner: InnerType,
+        protected inner: InnerType,
         protected readonly outgoingRequestProcessor: OutgoingRequestProcessor,
     ) {
         super();
 
         this.completionPromise = new Promise<void>((resolve, reject) => {
-            const onChange = async (): Promise<void> => {
+            this.onChangeCallback = async (): Promise<void> => {
```
richvdh, line 468: I'm looking at this again and feeling like we're jumping through hoops here. Suggestion:

 * Replace `completionPromise` with a deferred (`defer()` in `js-sdk/src/utils.ts`). Critically, that means that we can hoist this code out of the closure.
 * Turn `onChangeCallback` into a regular async method. Combine it with `onChange`; the subclasses can override it and call `super.onChange()` in the overriding impl.

src/rust-crypto/verification.ts
```diff
@@ -477,19 +483,39 @@ abstract class BaseRustVerifer<InnerType extends RustSdkCryptoJs.Qr | RustSdkCry
...
      *
      * Can be overridden by subclasses to see if we can notify the application about an update.
      */
     protected onChange(): void {}
 
+    /**
+     * Hook which is called when the underlying rust class is changed by calling `setInner`.
+     */
+    protected onSetInner(): void {}
+
```
richvdh, line 518: I rather imagined that you would move ~~`setInner`~~ `replaceInner` into `RustSASVerifier`. All these callback hooks feel complicated.

src/rust-crypto/verification.ts
uhoreg: Ah, I didn't know we had a `defer`. :+1:

uhoreg: OK, I tried doing so, but tests are failing, so I must be doing something wrong.  It's giving errors like:

```
  ● verification (rust-sdk) › cancellation › can cancel in the ShowQrCodeCallbacks

    RuntimeError: unreachable

      at std::panicking::rust_panic_with_hook::h366c91ab824150d6 (wasm:/wasm/014eee0e:1:2591137)
      at core::panicking::panic_fmt::h9728c4a03fdab7ce (wasm:/wasm/014eee0e:1:3444709)
      at matrix_sdk_crypto_wasm::machine::promise_result_to_future::{{closure}}::hb5a608792e59a66e (wasm:/wasm/014eee0e:1:1744633)
      at matrix_sdk_crypto_wasm::verification::send_change_info_to_callback::{{closure}}::hae64d2c5c9c49bef (wasm:/wasm/014eee0e:1:1394630)
      at matrix_sdk_crypto_wasm::verification::Qr::register_changes_callback::{{closure}}::ha47e9b5b026dda4b (wasm:/wasm/014eee0e:1:1750252)
      at wasm_bindgen_futures::queue::Queue::new::{{closure}}::h7ef0740be65eb2e0 (wasm:/wasm/014eee0e:1:2130242)
      at wasm_bindgen::convert::closures::invoke1_mut::h638884e1ee5d047c (wasm:/wasm/014eee0e:1:3559764)
      at wasm_bindgen__convert__closures__invoke1_mut__h638884e1ee5d047c (../matrix-rust-sdk-crypto-wasm/pkg/matrix_sdk_crypto_wasm.js:275:10)
      at f (../matrix-rust-sdk-crypto-wasm/pkg/matrix_sdk_crypto_wasm.js:217:20)

  ● verification (rust-sdk) › cancellation › can cancel in the ShowQrCodeCallbacks

    thrown: "Exceeded timeout of 5000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      787 |         });
      788 |
    > 789 |         it("can cancel in the ShowQrCodeCallbacks", async () => {
          |         ^
      790 |             // have alice initiate a verification. She should send a m.key.verification.request
      791 |             const [, request] = await Promise.all([
      792 |                 expectSendToDeviceMessage("m.key.verification.request"),

      at it (spec/integ/crypto/verification.spec.ts:789:9)
      at describe (spec/integ/crypto/verification.spec.ts:716:5)
          at Array.forEach (<anonymous>)
      at Object.<anonymous> (spec/integ/crypto/verification.spec.ts:119:47)
```

The patch I currently have is:

```diff
diff --git a/src/rust-crypto/verification.ts b/src/rust-crypto/verification.ts
index a163c3a0d..18f9d0025 100644
--- a/src/rust-crypto/verification.ts
+++ b/src/rust-crypto/verification.ts
@@ -34,6 +34,7 @@ import { OutgoingRequest, OutgoingRequestProcessor } from "./OutgoingRequestProc
 import { TypedReEmitter } from "../ReEmitter";
 import { MatrixEvent } from "../models/event";
 import { EventType, MsgType } from "../@types/event";
+import { defer, IDeferred } from "../utils";
 
 /**
  * An incoming, or outgoing, request to verify a user or a device via cross-signing.
@@ -455,8 +456,7 @@ abstract class BaseRustVerifer<InnerType extends RustSdkCryptoJs.Qr | RustSdkCry
     VerifierEventHandlerMap & VerificationRequestEventHandlerMap
 > {
     /** A promise which completes when the verification completes (or rejects when it is cancelled/fails) */
-    protected readonly completionPromise: Promise<void>;
-    protected onChangeCallback!: () => Promise<void>;
+    protected readonly completionDeferred: IDeferred<void>;
 
     public constructor(
         protected inner: InnerType,
@@ -464,29 +464,29 @@ abstract class BaseRustVerifer<InnerType extends RustSdkCryptoJs.Qr | RustSdkCry
     ) {
         super();
 
-        this.completionPromise = new Promise<void>((resolve, reject) => {
-            this.onChangeCallback = async (): Promise<void> => {
-                this.onChange();
-
-                if (this.inner.isDone()) {
-                    resolve(undefined);
-                } else if (this.inner.isCancelled()) {
-                    const cancelInfo = this.inner.cancelInfo()!;
-                    reject(
-                        new Error(
-                            `Verification cancelled by ${
-                                cancelInfo.cancelledbyUs() ? "us" : "them"
-                            } with code ${cancelInfo.cancelCode()}: ${cancelInfo.reason()}`,
-                        ),
-                    );
-                }
-
-                this.emit(VerificationRequestEvent.Change);
-            };
-            inner.registerChangesCallback(this.onChangeCallback);
-        });
+        this.completionDeferred = defer();
+        inner.registerChangesCallback(() => { this.onChangeCallback() });
         // stop the runtime complaining if nobody catches a failure
-        this.completionPromise.catch(() => null);
+        this.completionDeferred.promise.catch(() => null);
+    }
+
+    protected async onChangeCallback(): Promise<void> {
+        this.onChange();
+
+        if (this.inner.isDone()) {
+            this.completionDeferred.resolve(undefined);
+        } else if (this.inner.isCancelled()) {
+            const cancelInfo = this.inner.cancelInfo()!;
+            this.completionDeferred.reject(
+                new Error(
+                    `Verification cancelled by ${
+                        cancelInfo.cancelledbyUs() ? "us" : "them"
+                     } with code ${cancelInfo.cancelCode()}: ${cancelInfo.reason()}`,
+                ),
+            );
+        }
+
+        this.emit(VerificationRequestEvent.Change);
     }
 
     /**
@@ -579,7 +579,7 @@ export class RustQrCodeVerifier extends BaseRustVerifer<RustSdkCryptoJs.Qr> impl
             this.emit(VerifierEvent.ShowReciprocateQr, this.callbacks);
         }
         // Nothing to do here but wait.
-        await this.completionPromise;
+        await this.completionDeferred.promise;
     }
 
     /**
@@ -657,7 +657,7 @@ export class RustSASVerifier extends BaseRustVerifer<RustSdkCryptoJs.Sas> implem
      */
     public async verify(): Promise<void> {
         await this.sendAccept();
-        await this.completionPromise;
+        await this.completionDeferred.promise;
     }
 
     /**
@@ -733,7 +733,7 @@ export class RustSASVerifier extends BaseRustVerifer<RustSdkCryptoJs.Sas> implem
     public replaceInner(inner: RustSdkCryptoJs.Sas): void {
         if (this.inner != inner) {
             this.inner = inner;
-            inner.registerChangesCallback(this.onChangeCallback);
+            inner.registerChangesCallback(() => { this.onChangeCallback() });
             // replaceInner will only get called if we started the verification at the same time as the other side, and we lost
             // the tie breaker.  So we need to re-accept their verification.
             this.sendAccept();
```

uhoreg: Done, except for combining with `onChange`, because I'm not sure we want to rely on remembering to call `super.onChange` when we add a new verification method.

src/rust-crypto/verification.ts
```diff
@@ -455,53 +456,39 @@ abstract class BaseRustVerifer<InnerType extends RustSdkCryptoJs.Qr | RustSdkCry
     VerifierEventHandlerMap & VerificationRequestEventHandlerMap
 > {
     /** A promise which completes when the verification completes (or rejects when it is cancelled/fails) */
```
richvdh, line 458:
```suggestion
    /** A deferred which completes when the verification completes (or rejects when it is cancelled/fails) */
```

src/rust-crypto/verification.ts
richvdh: > because I'm not sure we want to rely on remembering to call `super.onChange` when we add a new verification method.

I feel like:
 * if I were implementing a new subclass of `BaseRustVerifer`, in which I chose to override a method in that superclass... then I would remember to check if the superclass method needed calling
 * that's a pretty reasonable thing to expect of any future maintainer
 * Right now, there's nothing stopping me, as someone implementing a new verification method, from overriding `onChangeCallback` instead of `onChange` and messing things up just as badly, so any potential benefit is a bit of a placebo anyway
 * Having separate `onChangeCallback` and `onChange` methods makes the flow harder to understand

But... I'll not insist if you'd prefer to keep it this way.

richvdh (commented):
LGTM

richvdh (approved):

uhoreg:
This is ready to merge, but I'm taking this afternoon off, and I don't want to merge this and then run off right away, so I'll merge on Monday

spec/unit/rust-crypto/verification.spec.ts
```diff
@@ -152,3 +405,148 @@ function makeMockedInner(): Mocked<RustSdkCryptoJs.VerificationRequest> {
...
+    if (crossSigningSignatures[userId] && crossSigningSignatures[userId][deviceId]) {
+        for (const [keyId, signature] of Object.entries(
+            crossSigningSignatures[userId][deviceId]["signatures"][userId],
+        )) {
+            deviceKeys["signatures"][userId][keyId] = signature;
+        }
+    }
+    const crossSigningKeys = JSON.parse(uploadSigningKeysRequest.body);
+    // note: the upload signatures request and upload signing keys requests
+    // don't need to me marked as sent in the Olm machine
```
richvdh, line 444:
```suggestion
    // don't need to be marked as sent in the Olm machine
```

spec/unit/rust-crypto/verification.spec.ts
```diff
@@ -152,3 +405,148 @@ function makeMockedInner(): Mocked<RustSdkCryptoJs.VerificationRequest> {
...
+    const { uploadKeysRequest, uploadSignaturesRequest, uploadSigningKeysRequest } =
+        await olmMachine.bootstrapCrossSigning(true);
+    const deviceKeys = JSON.parse(uploadKeysRequest.body).device_keys;
+    await olmMachine.markRequestAsSent(
+        uploadKeysRequest.id,
+        uploadKeysRequest.type,
+        '{"one_time_key_counts":{"signed_curve25519":100}}',
+    );
+    const crossSigningSignatures = JSON.parse(uploadSignaturesRequest.body);
+    if (crossSigningSignatures[userId] && crossSigningSignatures[userId][deviceId]) {
```
richvdh, line 435: it looks like this should always be true? I'd be tempted to get rid of this condition.

Alternatively, a ? operator would simplify:

```suggestion
    if (crossSigningSignatures[userId]?.[deviceId]) {
```

Doesn't matter though.

spec/unit/rust-crypto/verification.spec.ts
```diff
@@ -152,3 +405,148 @@ function makeMockedInner(): Mocked<RustSdkCryptoJs.VerificationRequest> {
...
+function makeRequestLoop(
+    ourOlmMachine: RustSdkCryptoJs.OlmMachine,
+    ourDeviceKeys: IDeviceKeys,
+    ourCrossSigningKeys: CrossSigningKeys,
+    theirOlmMachine: RustSdkCryptoJs.OlmMachine,
+    theirDeviceKeys: IDeviceKeys,
+    theirCrossSigningKeys: CrossSigningKeys,
+    customHandler:
+        | ((request: OutgoingRequest | RustSdkCryptoJs.UploadSigningKeysRequest) => Promise<any>)
+        | undefined = undefined,
```
richvdh, line 467: Consider defining something like:

```ts
type CustomRequestHandler = ((request: OutgoingRequest | RustSdkCryptoJs.UploadSigningKeysRequest) => Promise<any>);
```

Also, `?` after a param is effectively the same as `| undefined`, and `= undefined` is the default.

```suggestion
    customHandler?: CustomRequestHandler,
```

spec/unit/rust-crypto/verification.spec.ts
```diff
@@ -152,3 +405,148 @@ function makeMockedInner(): Mocked<RustSdkCryptoJs.VerificationRequest> {
...
+    let stopRequestLoop = false;
+    const ourUserId = ourOlmMachine.userId.toString();
+    const ourDeviceId = ourOlmMachine.deviceId.toString();
+    const theirUserId = theirOlmMachine.userId.toString();
+    const theirDeviceId = theirOlmMachine.deviceId.toString();
+
+    async function makeOutgoingRequest(
+        request: OutgoingRequest | RustSdkCryptoJs.UploadSigningKeysRequest,
+    ): Promise<any> {
+        let resp: any = customHandler && (await customHandler(request));
```
richvdh, line 478:
```suggestion
        let resp: any = await customHandler?.(request);
```

I think?

spec/unit/rust-crypto/verification.spec.ts
```diff
@@ -152,3 +405,148 @@ function makeMockedInner(): Mocked<RustSdkCryptoJs.VerificationRequest> {
...
+    }
+
+    const promise = (async () => {
+        while (!stopRequestLoop) {
+            const requests = await ourOlmMachine.outgoingRequests();
+            for (const request of requests) {
+                await makeOutgoingRequest(request);
+            }
+        }
+    })();
```
richvdh, line 543: this is a bit hard to read.

```suggestion
    async function runLoop() {
        while (!stopRequestLoop) {
            const requests = await ourOlmMachine.outgoingRequests();
            for (const request of requests) {
                await makeOutgoingRequest(request);
            }
        }
    }
    
    const loopCompletedPromise = runLoop();
```

spec/unit/rust-crypto/verification.spec.ts
```diff
@@ -152,3 +405,148 @@ function makeMockedInner(): Mocked<RustSdkCryptoJs.VerificationRequest> {
...
+    const ourUserId = ourOlmMachine.userId.toString();
+    const ourDeviceId = ourOlmMachine.deviceId.toString();
+    const theirUserId = theirOlmMachine.userId.toString();
+    const theirDeviceId = theirOlmMachine.deviceId.toString();
+
+    async function makeOutgoingRequest(
+        request: OutgoingRequest | RustSdkCryptoJs.UploadSigningKeysRequest,
+    ): Promise<any> {
+        let resp: any = customHandler && (await customHandler(request));
+        if (resp == undefined) {
```
richvdh, line 479:
```suggestion
        if (resp === undefined) {
```

richvdh: It would be quite nice to stick this whole block into a separate `handleRequest(ourOlmMachine, theirOlmMachine)` function.

spec/unit/rust-crypto/verification.spec.ts
```diff
@@ -152,3 +405,148 @@ function makeMockedInner(): Mocked<RustSdkCryptoJs.VerificationRequest> {
...
+                    }),
+                    new RustSdkCryptoJs.RoomId(request.room_id),
+                );
+            } else if (request instanceof RustSdkCryptoJs.SignatureUploadRequest) {
+                // this only gets called at the end after the verification
+                // succeeds, so we don't actually have to do anything.
+                resp = { failures: {} };
+            }
+        }
+        if (!(request instanceof RustSdkCryptoJs.UploadSigningKeysRequest) && request.id) {
```
richvdh, line 531: I think you can spell this more succinctly as

```suggestion
        if (request.id) {
```

but maybe that upsets typescript?

richvdh (approved):
Sorry, I didn't review the tests previously. Nothing important here -- generally they look great. Just a few suggestions for cleanups which you might like if you have the time and inclination.

spec/unit/rust-crypto/verification.spec.ts
uhoreg: Indeed, TypeScript gets upset about that

PR #4006, opened 2024-01-12
ElementR: report invalid keys rather than failing to restore from backup

uhoreg:
https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/85 adds a parameter to the importBackedUpRoomKeys callback that adds a parameter giving the number of failures.  This PR passes that information along.

## Checklist

---

## 🐛 Bug Fixes
 * ElementR: report invalid keys rather than failing to restore from backup ([\#4006](https://github.com/matrix-org/matrix-js-sdk/pull/4006)).


uhoreg:
https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/85 is merged, so this can be reviewed now.

src/rust-crypto/backup.ts
```diff
@@ -207,15 +207,18 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
-            opts?.progressCallback?.(importOpt);
-        });
+        await this.olmMachine.importBackedUpRoomKeys(
+            keysByRoom,
+            (progress: BigInt, total: BigInt, failures?: BigInt): void => {
+                const importOpt: ImportRoomKeyProgressData = {
+                    total: Number(total),
+                    successes: Number(progress),
+                    stage: "load_keys",
+                    failures: failures == undefined ? 0 : Number(failures),
```
richvdh, line 217: Do we really need this conditional? Won't `failures` always be defined?

richvdh (commented):
Presumably this won't actually work until matrix-sdk-crypto-wasm is updated (https://github.com/element-hq/element-web/issues/26900)?

Also, would be nice to see a test.

src/rust-crypto/backup.ts
uhoreg: `failures` will be `undefined` if it's called from an older version of matrix-sdk-crypto-wasm.  I guess we can drop it after the `package.json` depends on the latest version.

richvdh: I think the easiest thing to do would be to do the upgrade first, *then* land this PR.

uhoreg: Fair enough.  I'll mark this as draft again until https://github.com/matrix-org/matrix-js-sdk/pull/4021 is merged, and then make the changes

uhoreg:
> Also, would be nice to see a test.

Wrote a test and found *another* place where importing could fail.

src/rust-crypto/backup.ts
```diff
@@ -205,22 +205,41 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+                // if key.room_id or key.session_id are invalid, the wasm
+                // bindings will fail when decoding keysByRoom
+                invalid++;
+                const importOpt: ImportRoomKeyProgressData = {
+                    total: keys.length,
+                    successes: 0,
+                    stage: "load_keys",
+                    failures: invalid,
+                };
+                opts?.progressCallback?.(importOpt);
```
richvdh, line 220: since this loop is synchronous, I'm not convinced we should do this callback on each iteration.

src/rust-crypto/backup.ts
```diff
@@ -205,22 +205,41 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
      */
     public async importBackedUpRoomKeys(keys: IMegolmSessionData[], opts?: ImportRoomKeysOpts): Promise<void> {
         const keysByRoom: Map<RustSdkCryptoJs.RoomId, Map<string, IMegolmSessionData>> = new Map();
+        let invalid = 0;
         for (const key of keys) {
+            if (typeof key.room_id != "string" || typeof key.session_id != "string") {
```
richvdh, line 210: hrm. Did you observe real-world backup files that were malformed in this way?

uhoreg: Oh, I just took a look at this again, and the code that calls this sets `room_id` and `session_id`, rather than it coming from unauthenticated data, so we don't need this check after all.  Will remove.

src/rust-crypto/backup.ts
```diff
@@ -212,15 +212,20 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
-                failures: 0,
-            };
-            opts?.progressCallback?.(importOpt);
-        });
+        await this.olmMachine.importBackedUpRoomKeys(
+            keysByRoom,
+            (progress: BigInt, total: BigInt, failures: BigInt): void => {
+                const importOpt: ImportRoomKeyProgressData = {
+                    // since we didn't include the invalid keys when calling
+                    // importBackedUpRoomKeys, we need to add them to the total
```
richvdh, line 220: this comment no longer makes sense.

PR #4009, opened 2024-01-16
Fix new threads not appearing.

dbkr:
We try to update the thread roots when creating a thread, but a thread can take some time to be ready after being created so we were calling it too soon. Add a listener for the Update event to update the thread roots once it's ready.

Fixes https://github.com/element-hq/element-web/issues/26799

## Checklist

---

## 🐛 Bug Fixes
 * Fix new threads not appearing. ([\#4009](https://github.com/matrix-org/matrix-js-sdk/pull/4009)). Fixes element-hq/element-web#26799.


src/models/thread.ts
```diff
@@ -363,7 +369,7 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
      * to the start (and not the end) of the timeline.
      * @param emit - whether to emit the Update event if the thread was updated or not.
      */
-    public async addEvent(event: MatrixEvent, toStartOfTimeline: boolean, emit = true): Promise<void> {
+    public addEvent(event: MatrixEvent, toStartOfTimeline: boolean, emit = true): void {
```
dbkr, line 372: This was unnecessarily async. I fixed it but that set sonarcloud off on the function complexity, so I had to pull a chunk of it out to reduce the complexity. I can pull this out to a separate PR if necessary but hopefully it's obvious enough what's going on.

robintown (approved):

src/models/thread.ts
dbkr: ah okay, good to know. I assume dit wouldn't let me do that with code complexity.

PR #4010, opened 2024-01-17
KeyBackup: Disable signalling remaining keys to upload count

BillCarsonFr:
Related to https://github.com/element-hq/element-web/issues/26783#issuecomment-1895318551

Keys count performance can be pretty bad on some configurations (big database in FF).
For now we disable signalling the remaining key count to upload on rust crypto. (used by react sdk `SecureBackupPanel.tsx` to provide some feedback).

**Note for reviewer:**
I considered caching the count by only doing it when the backup loop is started.
The problem is that at the end of each sync you might receive some new keys and that could start a backup loop, meaning that the send performances could be degraded after each sync.
So for now, it's just disabled until we improve the store performances (count shouldn't take 10s of seconds)

## Checklist

---

### DIFF
--- a/src/rust-crypto/backup.ts
+++ b/src/rust-crypto/backup.ts
@@ -342,6 +342,10 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
 
                 if (!request || this.stopped || !this.activeBackupVersion) {
                     logger.log(`Backup: Ending loop for version ${this.activeBackupVersion}.`);
+                    if (!request) {
+                        // nothing more to upload
+                        this.emit(CryptoEvent.KeyBackupSessionsRemaining, 0);
+                    }
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


src/rust-crypto/backup.ts
```diff
@@ -342,20 +342,29 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
 
                 if (!request || this.stopped || !this.activeBackupVersion) {
                     logger.log(`Backup: Ending loop for version ${this.activeBackupVersion}.`);
+                    if (!request) {
+                        // nothing more to upload
+                        this.emit(CryptoEvent.KeyBackupSessionsRemaining, 0);
```
BillCarsonFr, line 347: Some test relies on this event to be fired at the end of the backup loop.

t3chguy (approved):

src/rust-crypto/backup.ts
richvdh: this comment should probably be in the code, not the PR.

richvdh (approved):
lgtm but please move your comment into the code

BillCarsonFr:
@t3chguy @richvdh Thanks for the feedback. I created a new PR that is keeping functionality and only computing room key count if needed and once per iteration
https://github.com/matrix-org/matrix-js-sdk/pull/4015

BillCarsonFr:
Closing in favor of https://github.com/matrix-org/matrix-js-sdk/pull/4015

PR #4011, opened 2024-01-17
Use Github Artifacts v4

t3chguy:
---

### DIFF
--- a/.github/workflows/docs-pr-netlify.yaml
+++ b/.github/workflows/docs-pr-netlify.yaml
@@ -11,13 +11,10 @@ jobs:
         if: github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'pull_request'
         runs-on: ubuntu-latest
         steps:
-            # There's a 'download artifact' action, but it hasn't been updated for the workflow_run action
-            # (https://github.com/actions/download-artifact/issues/60) so instead we get this mess:
             - name: 📥 Download artifact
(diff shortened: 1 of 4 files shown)


.github/workflows/docs-pr-netlify.yaml
```diff
@@ -11,13 +11,10 @@ jobs:
         if: github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'pull_request'
         runs-on: ubuntu-latest
         steps:
-            # There's a 'download artifact' action, but it hasn't been updated for the workflow_run action
-            # (https://github.com/actions/download-artifact/issues/60) so instead we get this mess:
             - name: 📥 Download artifact
-              uses: dawidd6/action-download-artifact@e7466d1a7587ed14867642c2ca74b5bcc1e19a2d # v3
+              uses: actions/download-artifact@v4
```
dbkr, line 15: Shouldn't we still be referencing the sha so it we get a well defined version?

t3chguy: We compromise for highly trusted actions between security & convenience, given that Github can always serve us whatever they like however we specify the ref.

dbkr: Ah okay, fair enough.

dbkr (commented):
Tests seem unhappy though: is there a path with a literal asterisk in it that it's trying to upload or has it got confused?

dbkr (approved):

PR #4015, opened 2024-01-18
ElementR | backup: call expensive `roomKeyCounts` less often

BillCarsonFr:
Fixes https://github.com/element-hq/element-web/issues/26893

Avoid calling the expensive room count at each backup iteration
## Checklist

---

## 🐛 Bug Fixes
 * ElementR | backup: call expensive `roomKeyCounts` less often ([\#4015](https://github.com/matrix-org/matrix-js-sdk/pull/4015)). Fixes element-hq/element-web#26893.

### DIFF
--- a/spec/unit/rust-crypto/BackupKeyLoop.spec.ts
+++ b/spec/unit/rust-crypto/BackupKeyLoop.spec.ts
@@ -1,0 +1,156 @@
+import { Mocked } from "jest-mock";
+import fetchMock from "fetch-mock-jest";
+
+import { RustBackupManager } from "../../../src/rust-crypto/backup";
+import * as RustSdkCryptoJs from "../../../../matrix-rust-sdk-crypto-wasm";
+import { KeysBackupRequest, OlmMachine, SignatureVerification } from "../../../../matrix-rust-sdk-crypto-wasm";
+import { CryptoEvent, HttpApiEvent, HttpApiEventHandlerMap, MatrixHttpApi, TypedEventEmitter } from "../../../src";
+import { OutgoingRequestProcessor } from "../../../src/rust-crypto/OutgoingRequestProcessor";
+import * as testData from "../../test-utils/test-data";
+import * as TestData from "../../test-utils/test-data";
+import { IKeyBackup } from "../../../src/crypto/backup";
+import { IKeyBackupSession } from "../../../src/crypto/keybackup";
+import { defer } from "../../../src/utils";
+
+describe("PerSessionKeyBackupDownloader", () => {
+    /** The backup manager under test */
+    let rustBackupManager: RustBackupManager;
+
+    let mockOlmMachine: Mocked<OlmMachine>;
+
+    let outgoingRequestProcessor: Mocked<OutgoingRequestProcessor>;
+
+    const httpAPi = new MatrixHttpApi(new TypedEventEmitter<HttpApiEvent, HttpApiEventHandlerMap>(), {
+        baseUrl: "http://server/",
+        prefix: "",
+        onlyData: true,
+    });
+
+    let idGenerator = 0;
+    function mockBackupRequest(keyCount: number): KeysBackupRequest {
+        const requestBody: IKeyBackup = {
+            rooms: {
+                "!room1:server": {
+                    sessions: {},
+                },
+            },
+        };
+        for (let i = 0; i < keyCount; i++) {
+            requestBody.rooms["!room1:server"].sessions["session" + i] = {} as IKeyBackupSession;
+        }
+        return {
+            id: "id" + idGenerator++,
+            body: JSON.stringify(requestBody),
+        } as unknown as Mocked<KeysBackupRequest>;
+    }
+
+    beforeEach(async () => {
+        jest.useFakeTimers();
+        idGenerator = 0;
+
+        mockOlmMachine = {
+            getBackupKeys: jest.fn().mockResolvedValue({
+                backupVersion: TestData.SIGNED_BACKUP_DATA.version!,
+                decryptionKey: RustSdkCryptoJs.BackupDecryptionKey.fromBase64(TestData.BACKUP_DECRYPTION_KEY_BASE64),
+            } as unknown as RustSdkCryptoJs.BackupKeys),
+            backupRoomKeys: jest.fn(),
+            isBackupEnabled: jest.fn().mockResolvedValue(true),
+            enableBackupV1: jest.fn(),
+            verifyBackup: jest.fn().mockResolvedValue({
+                trusted: jest.fn().mockResolvedValue(true),
+            } as unknown as SignatureVerification),
+            roomKeyCounts: jest.fn(),
+        } as unknown as Mocked<OlmMachine>;
+
+        outgoingRequestProcessor = {
+            makeOutgoingRequest: jest.fn(),
+        } as unknown as Mocked<OutgoingRequestProcessor>;
+
+        rustBackupManager = new RustBackupManager(mockOlmMachine, httpAPi, outgoingRequestProcessor);
+
+        fetchMock.get("path:/_matrix/client/v3/room_keys/version", testData.SIGNED_BACKUP_DATA);
+    });
+
+    it("Should call expensive roomKeyCounts only once per loop", async () => {
+        const lastKeysCalled = defer();
+        const remainingEmitted: number[] = [];
+
+        const zeroRemainingWasEmitted = new Promise<void>((resolve) => {
+            rustBackupManager.on(CryptoEvent.KeyBackupSessionsRemaining, (count) => {
+                remainingEmitted.push(count);
+                if (count == 0) {
+                    resolve();
+                }
+            });
+        });
+
+        // We want several batch of keys to check that we don't call expensive room key count several times
+        mockOlmMachine.backupRoomKeys
+            .mockResolvedValueOnce(mockBackupRequest(100))
+            .mockResolvedValueOnce(mockBackupRequest(100))
+            .mockResolvedValueOnce(mockBackupRequest(100))
+            .mockResolvedValueOnce(mockBackupRequest(100))
+            .mockResolvedValueOnce(mockBackupRequest(100))
+            .mockResolvedValueOnce(mockBackupRequest(100))
+            .mockResolvedValueOnce(mockBackupRequest(2))
+            .mockImplementation(async () => {
+                lastKeysCalled.resolve();
+                return null;
+            });
+
+        mockOlmMachine.roomKeyCounts.mockResolvedValue({
+            total: 602,
+            // first iteration don't call count, it will in second after 200 keys have been saved
+            backedUp: 200,
+        });
+
+        await rustBackupManager.checkKeyBackupAndEnable(false);
+        await jest.runAllTimersAsync();
+
+        await lastKeysCalled.promise;
+        await zeroRemainingWasEmitted;
+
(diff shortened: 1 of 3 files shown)


src/rust-crypto/backup.ts
```diff
@@ -323,7 +324,14 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
         await sleep(delay);
 
         try {
-            let numFailures = 0; // number of consecutive network failures for exponential backoff
+            // number of consecutive network failures for exponential backoff
+            let numFailures = 0;
+            // `olmMachine.roomKeyCounts()` is very slow on some configurations (see more comments below),
+            // we compute it only once per `backupKeysLoop` if necessary.
```
richvdh, line 330: I feel like it's more important to document here what value the variable actually holds, rather than why it exists. 

```suggestion
            // The number of keys left to back up. (Populated lazily: see more comments below.)
```

src/rust-crypto/backup.ts
```diff
@@ -342,20 +350,37 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
                     logger.log(`Backup: Ending loop for version ${this.activeBackupVersion}.`);
+                    if (!request) {
+                        // nothing more to upload
+                        this.emit(CryptoEvent.KeyBackupSessionsRemaining, 0);
+                    }
                     return;
                 }
 
-                try {
-                    await this.outgoingRequestProcessor.makeOutgoingRequest(request);
```
richvdh, line 349: I find it a bit confusing that the call to `roomKeyCounts` now happens *before* the call to `makeOutgoingRequest`. I feel like it would be simpler the other way around:

```ts
                    await this.outgoingRequestProcessor.makeOutgoingRequest(request);
                    numFailures = 0;
                    if (this.stopped) break;

                    // ... comments about performance ...
                    if (!isFirstIteration) {
                        if(remainingToUploadCount === null) {
                            const keyCount = await this.olmMachine.roomKeyCounts();
                            remainingToUploadCount = keyCount.total - keyCount.backedUp;
                        } else {
                            const keysCountInBatch = this.keysCountInBatch(request);
                            remainingToUploadCount = Math.max(remainingToUploadCount - keysCountInBatch, 0);
                        }
                        this.emit(CryptoEvent.KeyBackupSessionsRemaining, remainingToUploadCount);
                    }
```

WDYT?

src/rust-crypto/backup.ts
```diff
@@ -388,12 +413,29 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+     * @param batch - The backup request to count the keys from.
+     *
+     * @returns The number of keys in the backup request.
+     */
+    private keysCountInBatch(batch: RustSdkCryptoJs.KeysBackupRequest): number {
+        const parsedBody: IKeyBackup = JSON.parse(batch.body);
+        let count = 0;
+        for (const entry of Object.entries(parsedBody.rooms)) {
+            count += Object.keys(entry[1].sessions).length;
+        }
```
richvdh, line 436:
```suggestion
        for (const entry of Object.values(parsedBody.rooms)) {
            count += Object.keys(entry.sessions).length;
        }
```

src/rust-crypto/backup.ts
```diff
@@ -342,20 +350,37 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
-                try {
-                    await this.outgoingRequestProcessor.makeOutgoingRequest(request);
-                    numFailures = 0;
-                    if (this.stopped) break;
+                // Keys count performance (`olmMachine.roomKeyCounts()`) can be pretty bad on some configurations (big database in FF).
+                // We detected on some M1 mac that when the object store reach a threshold, the count performances stops growing in O(n) and
+                // suddenly become very slow (40s, 60s or more). Even on other configurations, the count can take several seconds.
+                // This will block other operations on the database, like sending messages
+                // This is a workaround to avoid calling `olmMachine.roomKeyCounts()` too often, and only when necessary.
+                // We don't call it on the first loop because there could be only a few keys to upload, and we don't want to wait for the count.
```
richvdh, line 365: A few wording tweaks. "Performances" is not a word in English ;-) [*]

```suggestion
                // Key count performance (`olmMachine.roomKeyCounts()`) can be pretty bad on some configurations.
                // In particular, we detected on some M1 macs that when the object store reaches a threshold, the count
                // performance stops growing in O(n) and suddenly becomes very slow (40s, 60s or more). 
                // Even on other configurations, the count can take several seconds.
                // This will block other operations on the database, like sending messages.
                // 
                // This is a workaround to avoid calling `olmMachine.roomKeyCounts()` too often, and only when necessary.
                // We don't call it on the first loop because there could be only a few keys to upload, and we don't want to wait for the count.
```


[*]: well, it is, but only in terms of an artistic performance. "Do you go to the theatre often?" "Yes, I watched three performances last week."

richvdh (commented):
A few suggestions

src/rust-crypto/backup.ts
```diff
@@ -342,20 +350,37 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
                     }
+                }
+
+                try {
+                    await this.outgoingRequestProcessor.makeOutgoingRequest(request);
+                    numFailures = 0;
+                    if (this.stopped) break;
+                    if (remainingToUploadCount !== null) {
+                        const keysCountInBatch = this.keysCountInBatch(request);
+                        remainingToUploadCount = Math.max(remainingToUploadCount - keysCountInBatch, 0);
```
ara4n, line 381: while the backup loop runs, new megolm sessions could arrive (especially if you're backing up 400K sessions or so), so presumably memoizing the count like this could mean remainingToUploadCount gets stuck at 0 for quite a while at the end (while it backs up the additional keys that arrived).  Worst case, the user might think that the progress has got stuck at zero and cancel it somehow.  I wonder if it's worth spelling this edge case out at least as a comment, that the KeyBackupSessionsRemaining event may now lie and keep emitting 0 for a bit at the end of the backup loop?

src/rust-crypto/backup.ts
BillCarsonFr: Done here https://github.com/matrix-org/matrix-js-sdk/pull/4015/commits/418074a1961958d3a38c6b97280a91238643685b

src/rust-crypto/backup.ts
BillCarsonFr: Yes, worth explaining. Added comment here https://github.com/matrix-org/matrix-js-sdk/pull/4015/commits/8dc1ff717d10d31b63b9af7a85cc3e0d687a0252

src/rust-crypto/backup.ts
BillCarsonFr: Right, done here https://github.com/matrix-org/matrix-js-sdk/pull/4015/commits/3e7fa5639f609c85cd9791e24a8c275681830c9b

src/rust-crypto/backup.ts
```diff
@@ -342,20 +350,37 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+                    }
                     return;
                 }
 
-                try {
-                    await this.outgoingRequestProcessor.makeOutgoingRequest(request);
-                    numFailures = 0;
-                    if (this.stopped) break;
+                // Keys count performance (`olmMachine.roomKeyCounts()`) can be pretty bad on some configurations (big database in FF).
+                // We detected on some M1 mac that when the object store reach a threshold, the count performances stops growing in O(n) and
```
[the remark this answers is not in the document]
BillCarsonFr: added here https://github.com/matrix-org/matrix-js-sdk/pull/4015/commits/ac383aabf611cf5a276cad9596405ba0cbfe372d

src/rust-crypto/backup.ts
```diff
@@ -361,29 +360,38 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+                            remainingToUploadCount = keyCount.total - keyCount.backedUp;
+                        } catch (err) {
+                            logger.error("Backup: Failed to get key counts from rust crypto-sdk", err);
+                        }
+                    }
+
                     if (remainingToUploadCount !== null) {
+                        this.emit(CryptoEvent.KeyBackupSessionsRemaining, remainingToUploadCount);
                         const keysCountInBatch = this.keysCountInBatch(request);
+                        // The `remainingToUploadCount` is computed only once for the current backupKeysLoop. But new
```
richvdh, line 389: This isn't accurate; we compute `remainingToUploadCount` on each iteration.

```suggestion
                        // `OlmMachine.roomKeyCounts` is called only once for the current backupKeysLoop. But new
```

spec/unit/rust-crypto/BackupKeyLoop.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+import { CryptoEvent, HttpApiEvent, HttpApiEventHandlerMap, MatrixHttpApi, TypedEventEmitter } from "../../../src";
+import { OutgoingRequestProcessor } from "../../../src/rust-crypto/OutgoingRequestProcessor";
+import * as testData from "../../test-utils/test-data";
+import * as TestData from "../../test-utils/test-data";
+import { IKeyBackup } from "../../../src/crypto/backup";
+import { IKeyBackupSession } from "../../../src/crypto/keybackup";
+import { defer } from "../../../src/utils";
+import { RustBackupManager } from "../../../src/rust-crypto/backup";
+
+describe("PerSessionKeyBackupDownloader", () => {
```
richvdh, line 14: this... doesn't seem to be testing `PerSessionKeyBackupDownloader` ?

richvdh (changes requested):

spec/unit/rust-crypto/BackupKeyLoop.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+        mockOlmMachine.roomKeyCounts.mockResolvedValue({
+            total: 602,
+            // first iteration don't call count, it will in second after 200 keys have been saved
+            backedUp: 200,
+        });
+
+        await rustBackupManager.checkKeyBackupAndEnable(false);
+        await jest.runAllTimersAsync();
+
+        await lastKeysCalled.promise;
```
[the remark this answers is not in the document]
BillCarsonFr: Right, was usefull before I started to use Faketimers, removed

spec/unit/rust-crypto/backup.spec.ts
```diff
@@ -8,10 +8,9 @@ import * as testData from "../../test-utils/test-data";
 import * as TestData from "../../test-utils/test-data";
 import { IKeyBackup } from "../../../src/crypto/backup";
 import { IKeyBackupSession } from "../../../src/crypto/keybackup";
-import { defer } from "../../../src/utils";
 import { RustBackupManager } from "../../../src/rust-crypto/backup";
 
-describe("PerSessionKeyBackupDownloader", () => {
+describe("Import keys from backup", () => {
```
richvdh, line 13: isn't it exporting keys to backup?

richvdh (approved):
lgtm otherwise

spec/unit/rust-crypto/backup.spec.ts
BillCarsonFr: Good catch, look like I was confused by my recent work on backup import

richvdh:
@BillCarsonFr if this is failing due to flaky playwright tests please make sure that there are issues open to track the flaky tests.

PR #4017, opened 2024-01-19
fix automatic DM avatar with functional members

HarHarLinks:
needed by: https://github.com/matrix-org/matrix-react-sdk/pull/12157

## Checklist

  Signed-off-by: Kim Brose <2803622+HarHarLinks@users.noreply.github.com>


src/models/room.ts
```diff
@@ -915,7 +915,7 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
     }
 
     public getAvatarFallbackMember(): RoomMember | undefined {
-        const memberCount = this.getInvitedAndJoinedMemberCount();
+        const memberCount = this.getInvitedAndJoinedMemberCount() - this.getInvitedAndJoinedFunctionalMemberCount();
```
richvdh, line 918: don't we also need to exclude the functional members from `members` further down in this method?

richvdh (changes requested):
I'm generally uncomfortable with adding all these methods to the public API, for two reasons:
 * A lot of them are just one-line wrappers for other methods. It's a lot of clutter in the API, making it hard to find the important methods
 * "functional members" are an Element extension, and baking lots of support for them into the js-sdk seems like a bit of a conflation of concerns.

I would be inclined to limit the scope of  this change to `getAvatarFallbackMember` (possibly with a helper method), without adding new public methods.

src/models/room.ts
HarHarLinks: you're right, I only tested a "DM +  Bot", so a DM room with me (alice), bob, and frank (the functional member), which worked great. covering the additional cases was necessary to cover e.g. a DM between alice and frank ("Empty Room") which would previously have had frank's avatar.

HarHarLinks:
so [25fdc18](https://github.com/matrix-org/matrix-js-sdk/pull/4017/commits/25fdc1826cc3720fe940e5312b89df15685ed44e) ended up reverting almost all previous work. would you like me to squash it into all the previous ones? it should still be good to review (actually better to review in its entirety instead of incrementally) and may also be squashed when merging.

HarHarLinks:
(not technically in scope of this Pr since it's about the avatar only, but)

with functional members, I can e.g. have a 1:1 DM between DM between alice and frank ("Empty Room")

richvdh:
Don't worry about squashing, I'll just review it as-is, and we'll squash on merge.

src/models/room.ts
```diff
@@ -914,12 +914,38 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         return this.myUserId;
     }
 
+    // gets service members (e.g. helper bots) for exclusion
```
richvdh, line 917: For exclusion from what?

Might as well make this a proper doc-comment

```suggestion
    /**
     * Gets the "functional members" in this room.
     *
     * Returns the list of userIDs from the `io.element.functional_members` event. Does not consider the
     * current membership states of those users.
     *
     * @see https://github.com/element-hq/element-meta/blob/develop/spec/functional_members.md.
     */
```

src/models/room.ts
```diff
@@ -914,12 +914,38 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+
+    private getInvitedAndJoinedFunctionalMemberCount(functionalMembers: String[]): number {
+        return functionalMembers.reduce((count, m) => {
+            const membership = this.getMembers().find((member) => member.userId === m)?.membership;
+            if (membership && ["join", "invite"].includes(membership)) {
+                return count + 1;
+            }
+            return count;
+        }, 0);
+    }
```
richvdh, line 934: This is a really awkward interface: you have to *pass in* the list of functional members for it to filter. It should be called `getNumberOfInvitedAndJoinedMembersWhoseIDsMatchThisRandomList`

Instead: given it is only used in one place, inline it?

src/models/room.ts
```diff
@@ -914,12 +914,38 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+        const mFunctionalMembers = this.currentState.getStateEvents(UNSTABLE_ELEMENT_FUNCTIONAL_USERS.name, "");
+        if (Array.isArray(mFunctionalMembers?.getContent().service_members)) {
+            return mFunctionalMembers!.getContent().service_members;
+        }
+        return [];
+    }
+
+    private getInvitedAndJoinedFunctionalMemberCount(functionalMembers: String[]): number {
+        return functionalMembers.reduce((count, m) => {
+            const membership = this.getMembers().find((member) => member.userId === m)?.membership;
```
richvdh, line 928: calling `getMembers` and searching through it for each entry in `functionalMembers` seems rather inefficient, particularly if the list of `functionalMembers` is long. Please can you find a better way to do this? (Maybe iterate through `getMembers` looking for entries that are in `functionalMembers`, rather than the other way around?)

src/models/room.ts
```diff
@@ -3351,11 +3381,7 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         let inviteJoinCount = joinedMemberCount + invitedMemberCount - 1;
 
         // get service members (e.g. helper bots) for exclusion
-        let excludedUserIds: string[] = [];
-        const mFunctionalMembers = this.currentState.getStateEvents(UNSTABLE_ELEMENT_FUNCTIONAL_USERS.name, "");
-        if (Array.isArray(mFunctionalMembers?.getContent().service_members)) {
-            excludedUserIds = mFunctionalMembers!.getContent().service_members;
-        }
+        let excludedUserIds: string[] = this.getFunctionalMembers();
```
richvdh, line 3384: annotation is redundant here. 

```suggestion
        let excludedUserIds = this.getFunctionalMembers();
```

src/models/room.ts
```diff
@@ -914,12 +914,38 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+        }, 0);
+    }
+
     public getAvatarFallbackMember(): RoomMember | undefined {
-        const memberCount = this.getInvitedAndJoinedMemberCount();
+        const functionalMembers = this.getFunctionalMembers();
+        const invitedAndJoinedFunctionalMemberCount = this.getInvitedAndJoinedFunctionalMemberCount(functionalMembers);
+
+        // only get avatar if conversation is with a single specific other user
+        const memberCount = this.getInvitedAndJoinedMemberCount() - invitedAndJoinedFunctionalMemberCount;
```
richvdh, line 941: wouldn't it be easier to iterate through the list of members once, excluding any functional members, rather than first counting the number of functional members, then counting all members, then subtracting one from the other?

src/models/room.ts
```diff
@@ -914,12 +914,38 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
         if (memberCount > 2) {
             return;
         }
-        const hasHeroes = Array.isArray(this.summaryHeroes) && this.summaryHeroes.length;
+
+        // prefer hero as indicated by the room summary
```
richvdh, line 946: I don't really know what this comment means.

"If the room summary contains a list of heroes, prefer those users over other members"

maybe?

spec/unit/room.spec.ts
```diff
@@ -2231,6 +2231,84 @@ describe("Room", function () {
             });
             expect(room.getAvatarFallbackMember()?.userId).toBe(userD);
         });
+
+        it("should return undefined if the room is a 1:1 with a functional member", async function() {
+            const room = new Room(roomId, null!, userA);
+            // room.getMembers = jest.fn().mockReturnValue({
+            //     userB: new RoomMember(roomId, userB),
+            // });
```
richvdh, line 2239: what does this comment mean?

Likewise below

spec/unit/room.spec.ts
```diff
@@ -2231,6 +2231,84 @@ describe("Room", function () {
...
+                    skey: "",
+                    room: roomId,
+                    event: true,
+                    content: {
+                        service_members: [userB],
+                    },
+                }),
+            ]);
+            room.currentState.markOutOfBandMembersStarted();
+            room.currentState.setOutOfBandMembers([
```
richvdh, line 2287: why do we use `...OutOfBandMembers` for this test rather than `setStateEvents`? Maybe we need a separate test to check the out-of-band-members case?

richvdh (changes requested):

src/models/room.ts
HarHarLinks: I originally chose not to do this, so the parts to rip out to remove the functional members feature would be obvious if need be. Now with the refactoring it became a single char change, so I applied this feedback.

spec/unit/room.spec.ts
HarHarLinks: Because I copied it from the test above :upside_down_face: I'm not sure what the difference is in the first place. What are they, and *do* I need to do something about them?

richvdh: I wish I knew... let's ignore them for now

src/models/room.ts
```diff
@@ -914,37 +914,79 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+                m.membership &&
+                // We care most about big rooms where also possibly lots of users may have left,
+                // so we first do the O(2) membership check, as it is definitely cheap and might remove candidates before the next step.
+                ["join", "invite"].includes(m.membership) &&
+                // Then we do the O(functionalMembers.length) functionalMember check, which is probably usually still cheap.
+                !functionalMembers.includes(m.userId)
+            ) {
+                return count + 1;
+            }
+            return count;
```
richvdh, line 947: This seems a confusing way to write this. And tbh I wouldn't bother documenting the performance considerations in this much detail.

```suggestion
            if (m.membership !== 'join' && m.membership !== 'invite') return count;
            if (functionalMembers.includes(m.userId)) return count;
            return count+1;
```

src/models/room.ts
```diff
@@ -914,37 +914,79 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+        }
+        return [];
+    }
+
     public getAvatarFallbackMember(): RoomMember | undefined {
-        const memberCount = this.getInvitedAndJoinedMemberCount();
-        if (memberCount > 2) {
+        const functionalMembers = this.getFunctionalMembers();
+        // Optimizing for performance, we iterate over all members outermost, as is it the longest of the 3 lists we are comparing.
+        const nonFunctionalMemberCount = this.getMembers()!.reduce((count, m) => {
```
richvdh, line 936: I feel like a `for` loop would be easier to grok than `reduce` (functional programming is all very well, but it's not really JavaScript's thing, and it is often harder to read). YMMV though, no strong opinion.

richvdh (changes requested):
Looking much better now, nearly there!

HarHarLinks:
Thank you for taking the time to work on this with me!

src/models/room.ts
```diff
@@ -932,25 +932,15 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
-            }
-            return count;
-        }, 0);
 
         // Only generate a fallback avatar if the conversation is with a single specific other user (a "DM").
-        if (nonFunctionalMemberCount > 2) {
-            return;
-        }
+        let nonFunctionalMemberCount = 0;
+        this.getMembers()!.forEach((m) => {
```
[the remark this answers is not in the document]
HarHarLinks: I see what you mean. I'm just not that firm in all the possible JS syntaxes. Your call.

PR #4023, opened 2024-01-22
Send MSC3981 'recurse' param on `/relations` endpoint on Matrix 1.10 servers

dbkr:
So we will send the 'recurse' parameter unprefixed for servers that support 1.10 (which most should now do, since it's released).

## Checklist

---


richvdh:
For links: this is about https://github.com/matrix-org/matrix-spec-proposals/pull/3981.

richvdh (dismissed):
Seems like you're not allowed to do this yet, because it will create problems if the MSC doesn't get written up by the time spec v1.10 is released.

dbkr:
Fair enough, I'll leave it as a draft until then.

richvdh (commented):
Looks good.

Could do with a changelog entry though.

dbkr:
Changelog entry will be done by the release process from the PR title.

richvdh:
> Changelog entry will be done by the release process from the PR title.

Not if it's T-Task, I think?

Also, "Mark MSC3981 as stable in Matrix 1.10" is a crap changelog entry. What does it actually mean, for a user of matrix-js-sdk?

dbkr:
Done, hopefully this is okay information-density wise.

richvdh (approved):
thanks

PR #4024, opened 2024-01-22
Fix crypto migration for megolm sessions with no sender key

richvdh:
Fixes https://github.com/element-hq/element-web/issues/26894.

Requires https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/89 (or rather, an update to a version of matrix-rust-sdk-crypto-wasm 4.0.1 which includes it).

element-web notes: Fix potential failure in migration from legacy to Rust cryptography.

---

## 🐛 Bug Fixes
 * Fix crypto migration for megolm sessions with no sender key ([\#4024](https://github.com/matrix-org/matrix-js-sdk/pull/4024)). Fixes element-hq/element-web#26894.

### DIFF
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -161,6 +161,21 @@ describe("initRustCrypto", () => {
     });
 
     describe("libolm migration", () => {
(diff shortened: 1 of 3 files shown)


andybalaam (approved):
Looks good, thanks.

PR #4026, opened 2024-01-23
Fix type error introduced by crypto-wasm 4.0.0

richvdh:
https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/82/files#diff-cf8f469cee769b4b3555833df9943f95a94dd55a4c6c5c8b8c7419e34fc8b5d8R341 changed the return type of `Sas.decimals()` from  `Array<any> | undefined` to `Uint16Array | undefined`. This can no longer be cast to `[number, number, number]` without causing a type error (for reasons I don't *really* understand, tbh).

---

### DIFF
--- a/src/rust-crypto/verification.ts
+++ b/src/rust-crypto/verification.ts
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 import * as RustSdkCryptoJs from "@matrix-org/matrix-sdk-crypto-wasm";
-import { Emoji, QrState } from "@matrix-org/matrix-sdk-crypto-wasm";
+import { QrState } from "@matrix-org/matrix-sdk-crypto-wasm";
 
 import {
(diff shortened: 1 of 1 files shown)


src/rust-crypto/verification.ts
```diff
@@ -659,18 +660,23 @@ export class RustSASVerifier extends BaseRustVerifer<RustSdkCryptoJs.Sas> implem
     /** if we can now show the callbacks, do so */
     protected onChange(): void {
         if (this.callbacks === null) {
-            const emoji: Array<Emoji> | undefined = this.inner.emoji();
```
richvdh, line 662: might as well clean this up while we're here: the return type of `emoji` is now `Array<Emoji> | undefined`

andybalaam (approved):

PR #4031, opened 2024-01-24
Refactor `MatrixClient.encryptAndSendEvent`

richvdh:
A series of commits to try and bring this code into the 21st century. Suggest review commit-by-commit.

~~Based on https://github.com/matrix-org/matrix-js-sdk/pull/4030.~~

---

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -4756,73 +4756,68 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * encrypts the event if necessary; adds the event to the queue, or sends it; marks the event as sent/unsent
      * @returns returns a promise which resolves with the result of the send request
      */
-    protected encryptAndSendEvent(room: Room | null, event: MatrixEvent): Promise<ISendEventResponse> {
-        let cancelled = false;
-        // Add an extra Promise.resolve() to turn synchronous exceptions into promise rejections,
-        // so that we can handle synchronous and asynchronous exceptions with the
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -4759,10 +4766,10 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
                 const encryptionPromise = this.encryptEventIfNeeded(event, room ?? undefined);
                 if (!encryptionPromise) return null; // doesn't need encryption
 
-                this.pendingEventEncryption.set(event.getId()!, encryptionPromise);
+                this.eventsBeingEncrypted.add(event.getId()!);
                 this.updatePendingEventStatus(room, event, EventStatus.ENCRYPTING);
                 return encryptionPromise.then(() => {
-                    if (!this.pendingEventEncryption.has(event.getId()!)) {
+                    if (!this.eventsBeingEncrypted.delete(event.getId()!)) {
```
richvdh: has -> delete here to fix a resource leak.

dbkr (approved):
Seems good I think. Bit of a shame about having to set the event status. then set it back but not seeing an obviously better alternative.

richvdh:
> Bit of a shame about having to set the event status. then set it back but not seeing an obviously better alternative.

Well, that's not new. The root of the problem here is the caller of `encryptAndSendEvent` setting the status to SENDING when it's not yet being sent. But I'm not going to fix that today.

PR #4032, opened 2024-01-24
Add expire_ts compatibility to matrixRTC

toger5:
Currenlty we use a combination of two fields in the Call member state events (`created_ts` + `expires`)
We want to move to just use a `expire_ts`.
This needs rethinking the logic behind syncing clients clocks. 
`created_ts` was mirring the `origin_server_timestamp` on update.
But `expire_ts` can only be known when the clock sync is known.

## Checklist

---

## ✨ Features
 * Add expire_ts compatibility to matrixRTC ([\#4032](https://github.com/matrix-org/matrix-js-sdk/pull/4032)). Contributed by @toger5.

### DIFF
--- a/src/matrixrtc/CallMembership.ts
+++ b/src/matrixrtc/CallMembership.ts
@@ -101,16 +101,16 @@ export class CallMembership {
 
     // gets the expiry time of the event, converted into the device's local time
     public getLocalExpiry(): number {
-        if (this.data.expires_ts) {
-            // With expires_ts we cannot convert to local time.
-            // TODO: Check the server timestamp and compute a diff to local time.
-            return this.data.expires_ts;
-        } else {
+        if (this.data.expires) {
             const relativeCreationTime = this.parentEvent.getTs() - this.createdTs();
 
             const localCreationTs = this.parentEvent.localTimestamp - relativeCreationTime;
 
-            return localCreationTs + this.data.expires!;
+            return localCreationTs + this.data.expires;
+        } else {
+            // With expires_ts we cannot convert to local time.
+            // TODO: Check the server timestamp and compute a diff to local time.
+            return this.data.expires_ts!;
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/matrixrtc/CallMembership.ts
```diff
@@ -77,16 +91,27 @@ export class CallMembership {
...
+            return this.data.expires_ts!;
+        }
     }
 
     // gets the expiry time of the event, converted into the device's local time
     public getLocalExpiry(): number {
-        const relativeCreationTime = this.parentEvent.getTs() - this.createdTs();
+        if (this.data.expires_ts) {
+            // With expires_ts we cannot convert to local time.
+            // TODO: Check the server timestamp and compute a diff to local time.
```
fkwp, line 106: do we already have a follow-up ticket for that TODO?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -624,6 +624,9 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
         };
 
         if (prevMembership) m.created_ts = prevMembership.createdTs();
+        if (m.created_ts) m.expires_ts = m.created_ts + (m.expires ?? 0);
+        // TODO: Date.now() should be the origin_server_ts (now).
```
fkwp, line 628: do we have the follow-up ticket for that TODO already?

fkwp (approved):
lgtm

src/matrixrtc/CallMembership.ts
toger5: As discussed I refactored this to use its own class for tracking the server/client diff with unimplmeneted clock sync diffs.
(It actally computes the diff once we get an event with an age property and from that point on its using this offset)

toger5: The refactor is now on a different branch so that we can merge this "low risk"/"lightweight" version.

We will do the server-client-time sync properly here: https://github.com/matrix-org/matrix-js-sdk/pull/4038

PR #4035, opened 2024-01-26
Element-R: check persistent room list for encryption config

richvdh:
We want to make sure that servers are not sneakily hiding or changing encryption configuration settings, which we do by persisting the encryption settings to the crypto store.

---

## ✨ Features
 * Element-R: check persistent room list for encryption config ([\#4035](https://github.com/matrix-org/matrix-js-sdk/pull/4035)).


andybalaam (approved):
Looks great, thanks!

PR #4040, opened 2024-01-30
Saner releases: rebuild release automation to provide better introspection

t3chguy:
Flow

1. Draft releases are triggered by the release person
2. Draft releases are inspected & tested along with their artifacts & notes
3. Draft releases are published when ready and this kicks off any deploys e.g. npm, docs, etc

Switches to prepack hook to better dry-run release steps in drafting

---


.github/workflows/release-workflow.yml
```diff
@@ -0,0 +1,48 @@
...
+              if: inputs.final
+              run: |
+                  git checkout master
+                  git merge -X theirs staging
+                  git push origin master
+
+    update-labels:
+        name: Advance release blocker labels
+        runs-on: ubuntu-latest
+        # TODO: skip this for hotfix releases
```
t3chguy, line 35: Deferred for a later task

.github/workflows/release-drafter-workflow.yml
```diff
@@ -44,9 +47,14 @@ on:
                 description: The number of expected assets, including signatures, excluding generated zip & tarball.
                 type: number
                 required: false
+            yarn-pack:
+                description: Whether to run `yarn pack` as a dry-run
```
dbkr, line 51: Looks like it controls where it runs at all? Also, default false?

dbkr (approved):

PR #4057, opened 2024-02-06
Add "AsJson" forms of the key import/export methods

andybalaam:
Part of https://github.com/element-hq/element-web/issues/26681

When we export keys with Rust encryption, we serialise them into JSON to cross the Rust/JS boundary, then we deserialise them in `RustCrypto` to match the `CryptoApi` interface, then serialise them again in matrix-react-sdk's ExportE2eKeysDialog so that we can encrypt the information and save it to disk.

By providing an `exportRoomKeysAsJson` method, we can provide the keys as a string immediately, avoiding the need to deserialise and re-serialise them, and similarly in the import case.

---

## ✨ Features
 * Add "AsJson" forms of the key import/export methods ([\#4057](https://github.com/matrix-org/matrix-js-sdk/pull/4057)). Contributed by @andybalaam.

### DIFF
--- a/spec/integ/crypto/crypto.spec.ts
+++ b/spec/integ/crypto/crypto.spec.ts
@@ -1379,7 +1379,7 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string, 
         });
         expect(decryptedEvent.getContent().body).toEqual("42");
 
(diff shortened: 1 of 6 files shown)


MidhunSureshR (approved):

BillCarsonFr (approved):
LGTM, thx for the description

PR #4059, opened 2024-02-07
ElementR | Ensure own user and device trust are updated after migration before giving back control to the app.

BillCarsonFr:
Fixes https://github.com/element-hq/element-web/issues/26977

## Problem

During the migration process we will migrate the private device and identity keys to rust stack. 
But until an initial `key/query` is done the sdk don't have access to the public keys and won't be able to establish the current trust.
This could result in several side effects where the app would improperly see trust for a short period.
Some side effects when this race occurs:
 - Messages sent by the current device will have a red shield for a few frames before going back to normal
 - The react-sdk could show the toaster to upgrade trust.

## Proposed Solution

This PR adds a new migration step that ensures that an initial `keys/query` is done before handing back control to the app. This might slow a bit the startup, but will avoid all the mentionned side effects.
If that initial `keys/query` fails due to network error the start process is not aborted; in that case the sdk will retry later anyhow.

## Checklist

---

## 🐛 Bug Fixes
 * ElementR | Ensure own user and device trust are updated after migration before giving back control to the app. ([\#4059](https://github.com/matrix-org/matrix-js-sdk/pull/4059)). Fixes element-hq/element-web#26977.

### DIFF
--- a/spec/integ/crypto/rust-crypto.spec.ts
+++ b/spec/integ/crypto/rust-crypto.spec.ts
@@ -112,6 +112,78 @@ describe("MatrixClient.initRustCrypto", () => {
             count: 79,
         });
 
+        fetchMock.post("path:/_matrix/client/v3/keys/query", {
+            device_keys: {
+                "@vdhtest200713:matrix.org": {
+                    KMFSTJSMLB: {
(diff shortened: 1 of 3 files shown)


andybalaam (approved):
Nice explanation and clear code, thank you!

PR #4061, opened 2024-02-09
Add basic retry for rust crypto outgoing requests

BillCarsonFr:
Add basic retry for outgoing requests supported by the rust SDK.
It will retry 3 times if:
 - Http status is 429, 500, 502, 503, 504, 525 (a bit arbitrary what would you suggest?)
 - If it's a connection error (timeout, connection went down ...)

It's also looking for the `retry_after_ms` of `DEFAULT_RETRY_DELAY_MS` to wait for the appropriate time.

Fixes https://github.com/element-hq/element-web/issues/26755 as it ultimatly fails on 500 and will work if retried

## Checklist

---

## 🐛 Bug Fixes
 * Add basic retry for rust crypto outgoing requests ([\#4061](https://github.com/matrix-org/matrix-js-sdk/pull/4061)). Fixes element-hq/element-web#26755.


t3chguy (changes requested):
Retrying 504 isn't necessarily safe as the action you requested may have already been performed but the response from the backend was dropped. It can only be retried on idempotent APIs and there a resulting error may need to be caught accordingly so the caller isn't told that the API failed due to e.g. M_NOT_FOUND or whatever.

richvdh:
Virtually (?) all matrix APIs *are* idempotent, deliberately so that they can be safely retried by clients. I would argue that, for simplicity, *all* 5xx response codes should be retried.

We should maybe also think about timeouts. Should we ever retry when the request times out?

t3chguy (dismissed):
Please fix the PR title, this implies it applies to outgoing requests generally, not specifically to a part of rust crypto.

BillCarsonFr:
> Virtually (?) all matrix APIs _are_ idempotent, deliberately so that they can be safely retried by clients. I would argue that, for simplicity, _all_ 5xx response codes should be retried.
> 
> We should maybe also think about timeouts. Should we ever retry when the request times out?

I updated given that all requests are indempotent, now retrying any 5xx requests.
Also client timeouts are not currently retried (Will be AbortError and not ConnectionError). There is an option to set the local timeout on MatrixClient, but it's not used and defaults to a very very high timing, so asm you said we should probably not retry them

src/rust-crypto/OutgoingRequestProcessor.ts
```diff
@@ -201,4 +204,84 @@ export class OutgoingRequestProcessor {
...
+                return await this.rawJsonRequest(method, path, queryParams, body);
+            } catch (e) {
+                if (currentRetryCount >= maxRetryCount) {
+                    // Max number of retries reached, rethrow the error
+                    throw e;
+                }
+
+                currentRetryCount++;
+
+                const maybeRetryAfter = this.shouldWaitBeforeRetryingMillis(e);
```
uhoreg, line 229: There is already a class that implements the retry logic.  It would probably be better to use that instead of writing something else here.  https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/scheduler.ts#L54

BillCarsonFr: Yes I have seen it. Indeed I think that there should be a retry mecanism inside the `FetchHttpApi` maybe with a new flag in `IRequestOpts` `retriable` (default true as most mx request are indempotent). 
Then the event scheduler will benefit from it too (unless there needs to be special logic for events). 

The thing is that it would impact the application a bit everywhere, I'd like to have support for the web team if we want to go that way.

(on side note I noticed that the scheduler has blocked retry on 502/M_TOO_LARGE, might not be common for rust outgoing request, but might want to exclude it too)

spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
```diff
@@ -274,4 +275,279 @@ describe("OutgoingRequestProcessor", () => {
...
+                [525, { status: 525, body: { error: "SSL Handshake Failed" } }],
+            ];
+            describe.each(retryableErrors)(`When status code is %s`, (_, error) => {
+                test.each(tests)(`for request of type %ss`, async (_, RequestClass, expectedMethod, expectedPath) => {
+                    // first, mock up a request as we might expect to receive it from the Rust layer ...
+                    const testBody = '{ "foo": "bar" }';
+                    const outgoingRequest = new RequestClass("1234", testBody);
+
+                    // @ts-ignore to avoid having to do if else to switch the method (.put/.post)
+                    fetchMock[expectedMethod.toLowerCase()](expectedPath, error);
```
richvdh, line 322: can't you just do:

```suggestion
                    fetchMock.mock(expectedPath, error, {method: expectedMethod});
```

richvdh (changes requested):
Looks sensible but a few suggestions for cleanups

src/rust-crypto/OutgoingRequestProcessor.ts
richvdh: There are a two different suggestions in this thread:

 * `MatrixScheduler` and `OutgoingRequestManager`  both implement logic to decide whether to retry a request, and if so when. We could extract that logic and share it, rather than having two copies. (Presumably, we would make a new method somewhere, and have `MatrixScheduler.RETRY_BACKOFF_RATELIMIT` call it.) That seems like quite a good idea to me. @BillCarsonFr: any reason we can't do this?
 * Move the retry mechanism down to `FetchHttpApi`, so that everything (including `MatrixScheduler` and `OutgoingRequestManager`) can benefit from it. This seems like a *much* more invasive change, and one I don't think we should do right now.

BillCarsonFr: > any reason we can't do this?

The `RETRY_BACKOFF_RATELIMIT` function is a bit strange, it has an used parameter, it is trying to cast a `MatrixError` into a `ConnectionError` (this can't work? or maybe at runtime?)
It doesn't want to retry on ConnectionError (like if the client internet connection is down), and we want that.

So I think we can't use it directly as it doesn't seem to be the same logic, and touching it will be also quite invasive. The only thing we could take easily is the exponential backoff, but it's just one line.

It's still bad that the logic doesn't match, but we might want to unify that in a second step?

src/rust-crypto/OutgoingRequestProcessor.ts
```diff
@@ -201,4 +204,84 @@ export class OutgoingRequestProcessor {
 
         return await this.http.authedRequest<string>(method, path, queryParams, body, opts);
     }
+
+    private async fetchWithRetry(
```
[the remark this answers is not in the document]
BillCarsonFr: renamed

src/rust-crypto/OutgoingRequestProcessor.ts
```diff
@@ -201,4 +204,84 @@ export class OutgoingRequestProcessor {
...
+        queryParams: QueryDict,
+        body: string,
+        maxRetryCount: number = 3,
+    ): Promise<string> {
+        let currentRetryCount = 0;
+
+        // eslint-disable-next-line no-constant-condition
+        while (true) {
+            try {
+                return await this.rawJsonRequest(method, path, queryParams, body);
```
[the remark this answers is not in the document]
BillCarsonFr: changed the order

src/rust-crypto/OutgoingRequestProcessor.ts
```diff
@@ -201,4 +204,84 @@ export class OutgoingRequestProcessor {
 
         return await this.http.authedRequest<string>(method, path, queryParams, body, opts);
     }
+
+    private async fetchWithRetry(
+        method: Method,
+        path: string,
+        queryParams: QueryDict,
+        body: string,
+        maxRetryCount: number = 3,
```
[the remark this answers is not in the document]
BillCarsonFr: move it as a constant

src/rust-crypto/OutgoingRequestProcessor.ts
```diff
@@ -201,4 +204,84 @@ export class OutgoingRequestProcessor {
...
+                currentRetryCount++;
+
+                const maybeRetryAfter = this.shouldWaitBeforeRetryingMillis(e);
+                if (maybeRetryAfter) {
+                    // wait for the specified time and then retry the request
+                    await sleep(maybeRetryAfter);
+                    // continue the loop and retry the request
+                } else {
+                    throw e;
+                }
```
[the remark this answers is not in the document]
BillCarsonFr: done

src/rust-crypto/OutgoingRequestProcessor.ts
```diff
@@ -201,4 +204,84 @@ export class OutgoingRequestProcessor {
...
+     * @param httpStatus - the HTTP status code of the response
+     */
+    private canRetry(httpStatus: number): boolean {
+        // Too Many Requests
+        if (httpStatus === 429) return true;
+
+        // 5xx Errors (Bad Gateway, Service Unavailable, Internal Server Error ...)
+        // This includes gateway timeout (504) and it's ok because all the requests made here are idempotent.
+        // All keys/signatures uploads are, message and to device are because of txn_id, keys claim in worst case will claim
+        // several keys but won't cause harm.
```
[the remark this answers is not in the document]
BillCarsonFr: thx, updated

spec/unit/rust-crypto/OutgoingRequestProcessor.spec.ts
BillCarsonFr: yes it works, thx

src/rust-crypto/OutgoingRequestProcessor.ts
BillCarsonFr: Ok, so after discussion I extracted a common retry alg between the scheduler and the outgoing request processor.
2 slight changes on the previous state:
 - Outgoing request processor try 5 times in total instead of 4
 - The matrix scheduer is not retrying on client timeouts anymore (as this timeout is already > 1mn depending on browser)
 
 Here is the refactoring commit https://github.com/matrix-org/matrix-js-sdk/pull/4061/commits/c2126a6f5d3960e467360bdc1a42ed629b1a3208

richvdh (commented):
looks sensible otherwise

richvdh (approved):

src/request-retry-utils.ts
```diff
@@ -0,0 +1,63 @@
...
+ * This produces wait times of 2, 4, 8, and 16 seconds (30s total) after which we give up. If the
+ * failure was due to a rate limited request, the time specified in the error is returned.
+ *
+ * Returns -1 if the error is not retryable, or if we reach the maximum number of attempts.
+ *
+ * @param err - The error thrown by the http call
+ * @param attempts - The current number of attempts
+ * @param retryConnectionError - Whether to retry on {@link ConnectionError} (CORS, connection is down, etc.)
+ */
+export function calculateRetryBackoff(err: any, attempts: number, retryConnectionError: boolean = false): number {
```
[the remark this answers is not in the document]
BillCarsonFr: removed the defalut as not used anyhow

t3chguy:
@BillCarsonFr could you explain why we're retrying UIA requests? This just makes it such that the user has to wait 20 seconds to see the error which isn't going to change with retries. https://github.com/element-hq/element-web/issues/27863

richvdh:
> could you explain why we're retrying UIA requests? 

This was to fix a login failure which was caused by the homeserver returning a 500 error to `/keys/upload`. `/keys/upload` is a UIA request, so we have to retry it.

t3chguy:
Right but a 403 is a valid correct error code during UIA which should not be retried, yet here we're retrying it?

richvdh:
If that's the case, it's certainly not the intention of this PR: https://github.com/matrix-org/matrix-js-sdk/pull/4061/files#diff-85a675b0871ea2f4a8991c253d876126df5949b8383211936b1c74657641864cR175-R178

richvdh:
In fact there are even tests that a 401 is *not* retried?

t3chguy:
That doesn't match its description

image

But yes good spot

PR #4062, opened 2024-02-12
Add support for device dehydration v2 (Element R)

uhoreg:


src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+        this.logger.info("dehydration: device rehydrated");
+
+        let nextBatch: string | undefined = undefined;
+        let toDeviceCount = 0;
+        let roomKeyCount = 0;
+        for (;;) {
+            const path = encodeUri("/dehydrated_device/$device_id/events", {
+                $device_id: dehydratedDeviceResp.device_id,
+            });
+            const eventResp: IDehydratedDeviceEventsResp = await this.http.authedRequest<IDehydratedDeviceEventsResp>(
```
richvdh, line 180: I don't think the type annotation is needed here?

```suggestion
            const eventResp = await this.http.authedRequest<IDehydratedDeviceEventsResp>(
```

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+            }
+            throw error;
+        }
+        return true;
+    }
+
+    /** Return whether the dehydration key is stored in Secret Storage
+     */
+    public async isKeyStored(): Promise<boolean> {
+        return Boolean(await this.secretStorage.isStored(SECRET_STORAGE_NAME.name));
```
andybalaam, line 94: I think you might want to look for `SECRET_STORAGE_NAME.altName` as well as `name` here, for forward compatibility, so this code works when the stable name is in use. For example, we often use `NamespacedValue.matches` to check both. What we're doing here seems conceptually similar to `NamespacedValue.findIn` although I don't think we can actually use it here.

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+                new RustSdkCryptoJs.DeviceId(dehydratedDeviceResp.device_id),
+                JSON.stringify(dehydratedDeviceResp.device_data),
+            );
+
+        this.logger.info("dehydration: device rehydrated");
+
+        let nextBatch: string | undefined = undefined;
+        let toDeviceCount = 0;
+        let roomKeyCount = 0;
+        for (;;) {
```
richvdh, line 176: Nit: I'm not sure if this is a written convention, but we normally spell this:

```suggestion
        while (true) {
```

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+            const path = encodeUri("/dehydrated_device/$device_id/events", {
+                $device_id: dehydratedDeviceResp.device_id,
+            });
```
richvdh, line 179: pull this out of the loop?

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+    public async resetKey(): Promise<void> {
+        const key = new Uint8Array(32);
+        crypto.getRandomValues(key);
+        await this.secretStorage.store(SECRET_STORAGE_NAME.name, encodeUnpaddedBase64(key));
+        this.key = key;
+    }
+
+    /** Get and cache the encryption key from secret storage
+     *
+     * If `create` is `true`, creates a new key if no existing key is present.
```
andybalaam, line 108: Please can you document the meaning of the return value?

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+            }
+            throw err;
+        }
+
+        this.logger.info("dehydration: device found");
+
+        const rehydratedDevice = await this.olmMachine
+            .dehydratedDevices()
+            .rehydrate(
+                this.key!,
```
andybalaam, line 166: If we store the key returned by the earlier `getKey` call, we can avoid the `!` here.

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+                Method.Post,
+                path,
+                undefined,
+                nextBatch ? { next_batch: nextBatch } : {},
+                {
+                    prefix: UnstablePrefix,
+                },
+            );
+
+            if (eventResp.events.length == 0) {
```
richvdh, line 190:
```suggestion
            if (eventResp.events.length === 0) {
```

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+     *     Defaults to creating the device immediately.
+     */
+    public async scheduleDeviceDehydration(interval: number, delay?: number): Promise<void> {
+        // cancel any previously-scheduled tasks
+        this.stop();
+
+        if (delay) {
+            this.timeoutId = setTimeout(() => {
+                this.scheduleDeviceDehydration(interval);
+                this.timeoutId = undefined;
```
richvdh, line 243: Just in case `scheduleDeviceDehydration` fails:

```suggestion
                this.timeoutId = undefined;
                this.scheduleDeviceDehydration(interval);
```

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1395,6 +1398,261 @@ describe("RustCrypto", () => {
...
+                    next_batch: nextBatch + "1",
+                };
+            });
+            fetchMock.post(
+                `path:/_matrix/client/unstable/org.matrix.msc3814.v1/dehydrated_device/${encodeURIComponent(dehydratedDeviceBody.device_id)}/events`,
+                eventsResponse,
+            );
+
+            expect(await rustCrypto.rehydrateDeviceIfAvailable()).toBe(true);
+            expect(eventsResponse.mock.calls).toHaveLength(2);
```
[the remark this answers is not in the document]
andybalaam: Ah, reading [MSC3814](https://github.com/matrix-org/matrix-spec-proposals/pull/3814) I think the events are supposed to have arrived later, not been uploaded at dehydration time.

src/rust-crypto/dehydration.ts
richvdh: Also, scheduleDeviceDehydration is async. Please add an explicit `.catch` logging any error that is thrown, to make it clear that we're happy.

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+            }, delay);
+        } else {
+            await this.createAndUploadDehydratedDevice();
+            // FIXME: should we randomize the time a bit?
+            this.intervalId = setInterval(this.createAndUploadDehydratedDevice.bind(this), interval);
```
richvdh, line 248: we normally use an arrow function rather than an explicit `bind`:

```suggestion
            this.intervalId = setInterval(() => this.createAndUploadDehydratedDevice(), interval);
```

richvdh: Also, again, please add an explicit `.catch`

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1395,6 +1398,261 @@ describe("RustCrypto", () => {
...
+                return {
+                    events,
+                    next_batch: nextBatch + "1",
+                };
+            });
+            fetchMock.post(
+                `path:/_matrix/client/unstable/org.matrix.msc3814.v1/dehydrated_device/${encodeURIComponent(dehydratedDeviceBody.device_id)}/events`,
+                eventsResponse,
+            );
+
```
andybalaam, line 1487: I see no mention of `POST /dehydrated_device/claim` from [MSC2697](https://github.com/matrix-org/matrix-spec-proposals/pull/2697) - is that not included?

src/rust-crypto/rust-crypto.ts
```diff
@@ -1641,6 +1654,63 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
+    /**
+     * Reset the dehydration key
+     *
+     * Note: this does not create a new dehydrated device.  This will need to
+     * be done either by calling `createAndUploadDehydratedDevice` or
+     * `scheduleDeviceDehydration`.
+     */
+    public async resetDehydrationKey(): Promise<void> {
+        return await this.dehydrationManager.resetKey();
+    }
```
richvdh, line 1713: Could you move all these to the `CryptoApi implementation` section of the class?

src/rust-crypto/rust-crypto.ts
```diff
@@ -1641,6 +1654,63 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
     public async getOwnIdentity(): Promise<RustSdkCryptoJs.OwnUserIdentity | undefined> {
         return await this.olmMachine.getIdentity(new RustSdkCryptoJs.UserId(this.userId));
     }
+
+    /**
+     * Returns whether dehydrated devices are supported by the crypto backend
+     * and by the server.
+     */
```
richvdh, line 1661: For all of these methods: we can just write:

```suggestion
    /**
     * Implementation of {@link CryptoApi#isDehydrationSupported}.
     */
```

... which saves duplicating the documentation and reinforces the point that we are implementing the interface.

spec/unit/rust-crypto/rust-crypto.spec.ts
andybalaam: Ah, wrong MSC! This is the right link: [MSC3814](https://github.com/matrix-org/matrix-spec-proposals/pull/3814)

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+            }
+            toDeviceCount += eventResp.events.length;
+            nextBatch = eventResp.next_batch;
+            const roomKeyInfos: RustSdkCryptoJs.RoomKeyInfo[] = await rehydratedDevice.receiveEvents(
+                JSON.stringify(eventResp.events),
+            );
+            roomKeyCount += roomKeyInfos.length;
+
+            // FIXME: is this actually needed?  It looks like the OlmMachine
+            // automatically re-tries decryption
```
andybalaam, line 201: Can we resolve this?

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+        return true;
+    }
+
+    /**
+     * Creates and uploads a new dehydrated device
+     *
+     * Creates and stores a new key in secret storage if none is available.
+     */
+    public async createAndUploadDehydratedDevice(): Promise<void> {
+        await this.getKey(true);
```
andybalaam, line 215: If we store this in a variable we don't need the `!` later.

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+
+        // FIXME: should raise error if server doesn't support dehydration?
+
+        const dehydratedDevice = await this.olmMachine.dehydratedDevices().create();
+        // FIXME: should the device display name be localised? passed as a
+        // parameter?
```
andybalaam, line 221: I'm not sure this FIXME is actionable. Realistically, should we just accept the fact that we are doing this, or otherwise, log an issue about it somewhere and link from here?

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+        this.stop();
+
+        if (delay) {
+            this.timeoutId = setTimeout(() => {
+                this.scheduleDeviceDehydration(interval);
+                this.timeoutId = undefined;
+            }, delay);
+        } else {
+            await this.createAndUploadDehydratedDevice();
+            // FIXME: should we randomize the time a bit?
```
andybalaam, line 247: Again, it would be useful to have an issue or something to track this if it is needed, along with reasons why we might need it.

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1395,6 +1398,261 @@ describe("RustCrypto", () => {
...
+            };
+
+            const outgoingRequestProcessor = {
+                makeOutgoingRequest: jest.fn(),
+            } as unknown as OutgoingRequestProcessor;
+            rustCrypto["outgoingRequestProcessor"] = outgoingRequestProcessor;
+            (rustCrypto["crossSigningIdentity"] as any)["outgoingRequestProcessor"] = outgoingRequestProcessor;
+            (rustCrypto["dehydrationManager"] as any)["outgoingRequestProcessor"] = outgoingRequestProcessor;
+            const outgoingRequestsManager = new OutgoingRequestsManager(logger, olmMachine, outgoingRequestProcessor);
+            rustCrypto["outgoingRequestsManager"] = outgoingRequestsManager;
```
richvdh, line 1520: can we not do what we need with `fetch-mock`, rather than all this messing about?

andybalaam (changes requested):
Looks really good. I requested a few test changes and made some minor comments.

Have you tried this out in Element Web or some other client? I feel sure we'll miss things if not...

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1395,6 +1398,261 @@ describe("RustCrypto", () => {
...
+            const rustCrypto = await makeTestRustCrypto(
+                makeMatrixHttpApi(),
+                testData.TEST_USER_ID,
+                undefined,
+                secretStorage,
+            );
+            const olmMachine: OlmMachine = rustCrypto["olmMachine"];
+            rustCrypto["checkKeyBackupAndEnable"] = async () => {
+                return null;
+            };
```
richvdh, line 1511: again, can we use fetch-mock to make the `/room_keys/version` request return a 404 instead of this?

richvdh (changes requested):
Looks generally sensible, though I have a number of comments, ranging in severity.

I'm a bit concerned that the API is moderately complicated, and that we're expecting each application to know when to call `isDehydrationSupported`, `rehydrateDeviceIfAvailable`, `scheduleDeviceDehydration` etc. Just peeking at https://github.com/matrix-org/matrix-react-sdk/pull/12316 (which I haven't looked at in detail): couldn't the content of [src/utils/device/dehydration.ts](https://github.com/matrix-org/matrix-react-sdk/pull/12316/files#diff-0230367fdcb5f8cc3c8580f2f7032b5285e1991d75ffb792ae788e7c9fd985d5) mostly be in here, so we can reduce the API surface?

At the very least I think we need some documentation explaining to client authors when and how they are meant to use this stuff.

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+                "/dehydrated_device",
+                undefined,
+                undefined,
+                {
+                    prefix: UnstablePrefix,
+                },
+            );
+        } catch (error) {
+            const err = error as MatrixError;
+            if (err.errcode === "M_UNRECOGNIZED") {
```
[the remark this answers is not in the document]
richvdh: This got specced in MSC3743.

florianduros (approved):
Approving for dep updates

spec/unit/rust-crypto/rust-crypto.spec.ts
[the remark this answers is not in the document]
uhoreg: > Can we check here that the device we rehydrated has the same information in it as what we dehydrated?

Unfortunately, the Rust API doesn't really provide any information about the rehydrated device.

> Ah, reading https://github.com/matrix-org/matrix-spec-proposals/pull/3814 I think the events are supposed to have arrived later, not been uploaded at dehydration time.

Yes, the events get uploaded by other people who want to send events to the device.

src/rust-crypto/dehydration.ts
uhoreg: I think I tried using `while (true)` in some code a long time ago, and one of the linters complained about it.  I'll see if that's still the case.

spec/unit/rust-crypto/rust-crypto.spec.ts
uhoreg: Possibly.  I just cargo-culted from a different test.

uhoreg:
> I'm a bit concerned that the API is moderately complicated, and that we're expecting each application to know when to call `isDehydrationSupported`, `rehydrateDeviceIfAvailable`, `scheduleDeviceDehydration` etc. Just peeking at [matrix-org/matrix-react-sdk#12316](https://github.com/matrix-org/matrix-react-sdk/pull/12316) (which I haven't looked at in detail): couldn't the content of [src/utils/device/dehydration.ts](https://github.com/matrix-org/matrix-react-sdk/pull/12316/files#diff-0230367fdcb5f8cc3c8580f2f7032b5285e1991d75ffb792ae788e7c9fd985d5) mostly be in here, so we can reduce the API surface?

I think it depends on how opinionated we want to be, but I don't think that we can push much more into the js-sdk.  `isDehydrationSupported` is called in the part of the code that determines whether to enable device dehydration, which in the react SDK depends on the contents of the well-known file, but I don't think that we want to force all the clients to do that.  (In the future, `isDehydrationSupported` might also be used to allow the user the choice to enable dehydration.)

The React SDK isn't doing much with `scheduleDeviceDehydration` currently -- it never uses the `delay` parameter.  This will come in a future iteration, where the application needs to store the time when a dehydrated device was last created, so that when you re-start, it will be able to re-schedule dehydration.  I had considered making that part of the js-sdk, but the js-sdk doesn't seem to have a good place to store things like that.  (In the past, I would have thrown it in the crypto store, but we don't have access to that any more, and the rest of the SDK doesn't seem to have a good place to easily persist random bits of information.)

> At the very least I think we need some documentation explaining to client authors when and how they are meant to use this stuff.

Yes, I can try to write something.

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+            toDeviceCount += eventResp.events.length;
+            nextBatch = eventResp.next_batch;
+            const roomKeyInfos: RustSdkCryptoJs.RoomKeyInfo[] = await rehydratedDevice.receiveEvents(
+                JSON.stringify(eventResp.events),
+            );
+            roomKeyCount += roomKeyInfos.length;
+
+            // FIXME: is this actually needed?  It looks like the OlmMachine
+            // automatically re-tries decryption
+            await this.rustCrypto.onRoomKeysUpdated(roomKeyInfos);
```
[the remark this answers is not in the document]
uhoreg: If I'm reading the Rust code correctly, `onRoomKeysUpdated` will already get called automatically, so I've taken it out.

src/rust-crypto/dehydration.ts
uhoreg: If I omit the type annotation, the TypeScript checker says "error TS7022: 'eventResp' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer."

I don't understand why, because other calls to `authedRequest` don't need the type annotation.

src/rust-crypto/dehydration.ts
uhoreg: Indeed, eslint says "error  Unexpected constant condition  no-constant-condition"

uhoreg: Looks like `while (true)` is used elsewhere in the code base, and they just disable the eslint warning.

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1395,6 +1398,261 @@ describe("RustCrypto", () => {
...
+                };
+            }
+
+            // create initial secret storage
+            await rustCrypto.bootstrapCrossSigning({ setupNewCrossSigning: true });
+            await rustCrypto.bootstrapSecretStorage({
+                createSecretStorageKey,
+                setupNewSecretStorage: true,
+                setupNewKeyBackup: false,
+            });
```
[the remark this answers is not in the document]
uhoreg: I've added some comments and restructured some of the setup in response to vdh's review, so hopefully it should be clearer.  The two tests are now in separate files, so there's less advantage to factoring out the setup, and I'd prefer not to pull it out into some place far away from the test, unless you think it's still too confusing.

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1395,6 +1398,261 @@ describe("RustCrypto", () => {
             });
         });
     });
+
+    describe("dehydration", () => {
```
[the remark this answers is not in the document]
uhoreg: Made it into an integration test.

src/rust-crypto/dehydration.ts
uhoreg: removed the FIXME, will create an issue

src/rust-crypto/rust-crypto.ts
```diff
@@ -112,6 +113,8 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
 
     private readonly reemitter = new TypedReEmitter<RustCryptoEvents, RustCryptoEventMap>(this);
 
+    private readonly dehydrationManager: RustDehydrationManager;
+
```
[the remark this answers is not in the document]
uhoreg: I think that I made the changes that you wanted.

src/rust-crypto/dehydration.ts
```diff
@@ -0,0 +1,262 @@
...
+interface IDehydratedDeviceEventsResp {
+    events: IToDeviceEvent[];
+    next_batch: string;
+}
+
+export const UnstablePrefix = ClientPrefix.Unstable + "/org.matrix.msc3814.v1";
+const SECRET_STORAGE_NAME = new UnstableValue("m.dehydrated_device", "org.matrix.msc3814");
+
+/**
+ * Manages dehydrated devices
```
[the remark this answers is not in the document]
uhoreg: I added some more words.

spec/unit/rust-crypto/rust-crypto.spec.ts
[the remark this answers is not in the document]
uhoreg: Actually, I separated it out anyways.

src/rust-crypto/dehydration.ts
richvdh: somewhat amusing that the lint rule doesn't catch `for` loops.

andybalaam (approved):

src/rust-crypto/dehydration.ts
richvdh: :shrug:

richvdh (approved):
LGTM other than a few suggestions which may or may not be helpful

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1395,6 +1398,158 @@ describe("RustCrypto", () => {
...
+            await initializeSecretStorage(rustCrypto, testData.TEST_USER_ID);
+
+            fetchMock.config.overwriteRoutes = true;
+
+            // when we schedule dehydration with no delay, it should create a
+            // dehydrated device immediately
+            const firstDehydrationRequest = jest.fn(() => {
+                return {};
+            });
+            fetchMock.put(
```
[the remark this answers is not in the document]
uhoreg: I changed it to `putOnce`, but it still wants me to keep the `overwriteRoutes = true` :shrug:

richvdh:
> > I'm a bit concerned that the API is moderately complicated, and that we're expecting each application to know when to call `isDehydrationSupported`, `rehydrateDeviceIfAvailable`, `scheduleDeviceDehydration` etc.

I've been thinking about this a bit more, and it's continuing to nag at me.

In particular: we already have a bit of a mess with APIs like `isSecretStorageReady`, `bootstrapSecretStorage`, `isCrossSigningReady`, `bootstrapCrossSigning`: they have unclear semantics, and it's unclear when the application is meant to call them, and in what order. This is mostly because they were defined back in 2018 when we thought it was "obvious" what they did and when you should call them, but then they evolved and now it's really unclear.

I'm really worried that we're compounding this situation by adding a bunch more methods that the application is supposed to call at some ill-defined time during startup.

For example:
 * Is it necessary that secret storage is correctly configured before calling the dehydration methods? What is the expected behaviour of these methods if it is not configured?
 * We are likely to see every possible combination of (new/existing session), (account with/without secret storage), (device with/without cached secret storage key), (account with/without cross-signing), (device verified/not verified), (device with/without access to master cross-signing keys). I'm not sure which, if any, of those are relevant, but I do feel like we should make any constraints absolutely explicit.

> The React SDK isn't doing much with `scheduleDeviceDehydration` currently -- it never uses the `delay` parameter. This will come in a future iteration, where the application needs to store the time when a dehydrated device was last created, so that when you re-start, it will be able to re-schedule dehydration. I had considered making that part of the js-sdk, but the js-sdk doesn't seem to have a good place to store things like that. (In the past, I would have thrown it in the crypto store, but we don't have access to that any more, and the rest of the SDK doesn't seem to have a good place to easily persist random bits of information.)

A couple of things on this:
 * The MatrixClient is initialised with a `Store`. It's typically used for caching the last `/sync` response, but we could conceivably extend it to store this sort of information. The thing about `Store` is that it is a *cache*: if it is blown away, then we can resync from the server without loss of information. But that might be ok here?
 * If it's important, we can and should use the rust-side store for this sort of information. It probably means pushing more of this functionality down into the Rust layer, but that may not be a bad thing?

spec/unit/rust-crypto/rust-crypto.spec.ts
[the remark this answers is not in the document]
richvdh: :shrug:

src/crypto-api.ts
```diff
@@ -496,6 +496,76 @@ export interface CryptoApi {
...
+     * want to have more control over the dehydration process may use this
+     * function instead.
+     */
+    createAndUploadDehydratedDevice(): Promise<void>;
+
+    /**
+     * Schedule periodic creation of dehydrated devices.
+     *
+     * If the delay is omitted or 0, this function's promise will resolve after
+     * the first dehydrated device is created.
```
richvdh, line 544:
```suggestion
     * If the delay is omitted or 0, this function's promise will resolve after
     * the first dehydrated device is created. Otherwise, it resolves immediately.
```

uhoreg:
> > The React SDK isn't doing much with `scheduleDeviceDehydration` currently -- it never uses the `delay` parameter. This will come in a future iteration, where the application needs to store the time when a dehydrated device was last created, so that when you re-start, it will be able to re-schedule dehydration. I had considered making that part of the js-sdk, but the js-sdk doesn't seem to have a good place to store things like that. (In the past, I would have thrown it in the crypto store, but we don't have access to that any more, and the rest of the SDK doesn't seem to have a good place to easily persist random bits of information.)
> 
> A couple of things on this:
> 
>     * The MatrixClient is initialised with a `Store`. It's typically used for caching the last `/sync` response, but we could conceivably extend it to store this sort of information. The thing about `Store` is that it is a _cache_: if it is blown away, then we can resync from the server without loss of information. But that might be ok here?
> 
>     * If it's important, we can and should use the rust-side store for this sort of information. It probably means pushing more of this functionality down into the Rust layer, but that may not be a bad thing?

Having thought a bit more about this, I don't think it's critical that we need to store that information.  If we replace the dehydrated device early, nothing bad will happen, so on restart, we can just create a new dehydrated device right away, and continue from there with the normal scheduling.  So I can change the `schedule...` function to always create a new dehydrated device right away.  I'll also give it a default period.

Regarding the larger API concerns, I'll give it some thought, about whether I can throw things into `bootstrapCrossSigning`

uhoreg:
I've implemented a simpler API.  Now, clients just need to check if dehydration is available, and then tell it to start dehydration, and it'll do everything necessary.  Clients can re-start dehydration if it, e.g. needs to create a new dehydration key when it resets secret storage.

src/crypto-api.ts
```diff
@@ -496,6 +496,42 @@ export interface CryptoApi {
...
+    isDehydrationSupported(): Promise<boolean>;
+
+    /**
+     * Start using device dehydration.
+     *
+     * - Rehydrates a dehydrated device, if one is available.
+     * - Creates a new dehydration key, if necessary, and store it in Secret
+     *   Storage.
+     *   - If `createNewKey` is set to true, always create a new key.
+     *   - If a dehydration key is not available, create a new one.
```
richvdh, line 522: Nit:

```suggestion
     * - Creates a new dehydration key, if necessary, and stores it in Secret
     *   Storage.
     *   - If `createNewKey` is set to true, always creates a new key.
     *   - If a dehydration key is not available, creates a new one.
```

src/rust-crypto/DehydratedDeviceManager.ts
```diff
@@ -109,6 +112,35 @@ export class DehydratedDeviceManager {
...
     }
 
+    /**
+     * Start using device dehydration.
+     *
+     * - Rehydrates a dehydrated device, if one is available.
+     * - Creates a new dehydration key, if necessary, and store it in Secret
+     *   Storage.
+     *   - If `createNewKey` is set to true, always create a new key.
+     *   - If a dehydration key is not available, create a new one.
```
richvdh, line 122: Nit: again, please make the verb tenses match

src/rust-crypto/DehydratedDeviceManager.ts
```diff
@@ -250,28 +282,17 @@ export class DehydratedDeviceManager {
      * Schedule periodic creation of dehydrated devices.
      *
      * @param interval - the time to wait between creating dehydrated devices
```
richvdh, line 284: `interval` no longer exists?

src/crypto-api.ts
```diff
@@ -840,6 +806,11 @@ export interface CreateSecretStorageOpts {
      */
     setupNewSecretStorage?: boolean;
 
+    /**
+     * Create a dehydrated device if no dehydrated device is already present.
+     */
+    initialiseDeviceDehydration?: boolean;
+
```
richvdh, line 813: this doesn't seem to be implemented?

src/rust-crypto/DehydratedDeviceManager.ts
```diff
@@ -0,0 +1,309 @@
...
+     */
+    public async start(createNewKey?: boolean): Promise<void> {
+        this.stop();
+        try {
+            await this.rehydrateDeviceIfAvailable();
+        } catch (e) {
+            // If rehydration fails, there isn't much we can do about it.  Log
+            // the error, and create a new device.
+            this.logger.info("dehydration: Error rehydrating device:", e);
+        }
```
richvdh, line 137: Does this mean that if somebody accidentally calls `start` before setting up secret storage or cross-signing correctly, we'll throw away the existing dehydrated device without getting any of its encrypted goodness? Seems like a dangerous footgun.

richvdh (changes requested):
Excellent, this seems better! Still a couple of concerns.

src/crypto-api.ts
uhoreg: Right, I tried to add it, changed my mind, and missed removing this bit.

src/rust-crypto/DehydratedDeviceManager.ts
uhoreg: I've added a check (in `RustCrypto`) to make sure that secret storage and cross-signing are set up.

richvdh (approved):
LGTM!

PR #4064, opened 2024-02-13
Update MSC2965 OIDC Discovery implementation

t3chguy:
https://github.com/sandhose/matrix-doc/blob/msc/sandhose/oidc-discovery/proposals/2965-oidc-discovery.md

Switches from `m.authentication` in `.well-known/matrix/client` to `GET /auth_issuer` API as the MSC has done.

Requires https://github.com/matrix-org/matrix-js-sdk/pull/4074

---

## ✨ Features
 * Update MSC2965 OIDC Discovery implementation ([\#4064](https://github.com/matrix-org/matrix-js-sdk/pull/4064)).


richvdh (changes requested):
Help me out here. This is a chunky change, and is going to be time-consuming to review without any context.

What is the actual change in behaviour here, in more detail than "update"? I see the link to the MSC doc, but that in itself is a chunky document, and presumably some of it was previously implemented?

Is it possible to break up the change a bit, to make it easier to review (eg, it looks like one of the changes is to strip out support for `m.authentication`? I don't really understand why that's happening, given the MSC still seems to talk about it, but can you pull that out to its own commit or its own PR?)

src/client.ts
```diff
@@ -9974,6 +9963,14 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             throw err;
         }
     }
+
+    public async getAuthIssuer(): Promise<{
+        issuer: string;
+    }> {
```
richvdh, line 9969: I can't say I'm a huge fan of adding even more methods to `MatrixClient`, but I suppose that's a discussion for another day. If you're going to do it, document it and give it test coverage, please.

src/autodiscovery.ts
```diff
@@ -543,7 +419,7 @@ export class AutoDiscovery {
         }
     }
 
-    private static fetch(resource: URL | string, options?: RequestInit): ReturnType<typeof global.fetch> {
+    public static fetch(resource: URL | string, options?: RequestInit): ReturnType<typeof global.fetch> {
```
richvdh, line 422: Given that AutoDiscovery is part of the js-sdk's public API, making this public really isn't great. Can you find any other way of doing it? (Is it just to support replacing `fetchFn` in the tests? I don't really follow why we can't do what we need with fetch-mock.)

At the very least it needs documenting, but exposing this to applications feels like something we should avoid.

src/oidc/index.ts
```diff
@@ -14,9 +14,17 @@ See the License for the specific language governing permissions and
...
 export * from "./discovery";
 export * from "./error";
 export * from "./register";
 export * from "./tokenRefresher";
 export * from "./validate";
+
+export interface OidcClientConfig extends ValidatedIssuerConfig {
+    metadata: ValidatedIssuerMetadata;
+    signingKeys?: SigningKey[];
+}
```
richvdh, line 30: document this please. What does it represent?

spec/unit/autodiscovery.spec.ts
```diff
@@ -897,75 +871,4 @@ describe("AutoDiscovery", function () {
             }),
         ]);
     });
-
-    describe("m.authentication", () => {
```
richvdh, line 901: I appreciate these tests don't really apply any more, but it worries me to see tests being ripped out without new ones to replace them. Shouldn't we have similar tests for the new endpoint?

src/oidc/discovery.ts
```diff
@@ -16,8 +16,14 @@ limitations under the License.
...
+    isValidatedIssuerMetadata,
+    ValidatedIssuerConfig,
+    ValidatedIssuerMetadata,
+    validateOIDCIssuerWellKnown,
+} from "./validate";
+import { Method, timeoutSignal } from "../http-api";
+import { AutoDiscovery } from "../autodiscovery";
 
 /**
  * @experimental
```
richvdh, line 29: this comment still talks about `m.authentication`. I think it could maybe do with some other updates too.

richvdh:
> Its quite entangled so not sure how to sanely split this out without the individual commits not being atomic.

A few ideas to kick you off:
 * `OidcClientConfig` is moving to a new file. Pull out the move to its own commit.
 * There seems to be a formatting change to `spec/unit/oidc/authorize.spec.ts` ? Let's separate formatting cleanups from functional changes.
 * Make adding support for `account_management_endpoint` its own commit.
 * `validateDiscoveryAuthenticationConfig` isn't currently used anywhere afaict? Make ripping it out its own comment.

Those are just a few ideas I found quickly. I'm sure there are others; like I say, I'm finding it hard to follow at the moment so I'm not in the best place to advise. It doesn't matter if some of those commits end up being trivial; it all helps to reduce the amount of state a reviewer has to keep in their head at once.

richvdh:
I realise a lot of this is marked `@experimental`, but surely there is something using it somewhere that is going to be broken? What's the plan for managing that?

richvdh (commented):
A few drive-by comments

t3chguy:
Have split out https://github.com/matrix-org/matrix-js-sdk/pull/4071 - will rebase this PR and split into sane commits afterwards

t3chguy:
> What's the plan for managing that?

https://github.com/matrix-org/matrix-react-sdk/pull/12245

src/oidc/index.ts
```diff
@@ -14,9 +14,21 @@ See the License for the specific language governing permissions and
...
+
 export * from "./authorize";
 export * from "./discovery";
 export * from "./error";
 export * from "./register";
 export * from "./tokenRefresher";
 export * from "./validate";
+
+/**
+ * Validated config for native OIDC authentication
```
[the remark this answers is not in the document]
t3chguy: Except it isn't "as returned" - given its parsed into metadata, the jwt keys are also fetched, and things are camelCased and not snake_cased

src/oidc/register.ts
```diff
@@ -118,7 +117,7 @@ const doRegistration = async (
  * @throws when registration is not supported, on failed request or invalid response
  */
 export const registerOidcClient = async (
-    delegatedAuthConfig: IDelegatedAuthConfig & ValidatedIssuerConfig,
+    delegatedAuthConfig: OidcClientConfig,
```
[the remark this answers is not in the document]
t3chguy: `ValidatedServerConfig` is a thing in matrix-react-sdk - very odd

src/oidc/index.ts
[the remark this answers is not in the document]
richvdh: Well ok, can we find a form of wording which is both accurate and gives a bit of context to wtf this class is actually for?

"This class is the result of a call to {@link discoverAndValidateOIDCIssuerWellKnown}"

... and then mention `https://oidc-issuer.example.com/.well-known/openid-configuration` in the doc-comment for `discoverAndValidateOIDCIssuerWellKnown` ?

richvdh (approved):
Was there an answer to my question at https://github.com/matrix-org/matrix-js-sdk/pull/4064#discussion_r1491019612?

Also would like to see a mention of `.well-known/openid-configuration` somewhere in the doc-comments, per https://github.com/matrix-org/matrix-js-sdk/pull/4064#discussion_r1498923198.

LGTM otherwise

spec/unit/autodiscovery.spec.ts
t3chguy: We have a test for the new endpoint now, it landed as its own PR.

As for the rest of it, its just removing code and changing types, given tests have been updated with stubs matching the new types I'd consider that tested

PR #4065, opened 2024-02-13
Ignore memberships of users that are not in the call

toger5:

## Checklist


robintown (approved):
A question, but otherwise looks good: Are we sure that a room membership change causes the call members to be recalculated?

toger5:
> A question, but otherwise looks good: Are we sure that a room membership change causes the call members to be recalculated?

This is a very good catch. It was indeed not the case. I updated the pr

PR #4079, opened 2024-02-21
Don't fetch the root events of a thread if we already have it

dbkr:
The comment said it was checking, but it didn't, it just fetched it anyway.

## Checklist


t3chguy (commented):
The reason for this was to get the latest state (via unsigned) from the server, the comment fell out of sync with the behaviour

dbkr:
Replaced with https://github.com/matrix-org/matrix-js-sdk/pull/4080

PR #4081, opened 2024-02-22
PR template: reminder to document your stuff

richvdh:
If people are happy with this, I will replicate it to the other repos.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,6 +3,6 @@
 ## Checklist
 
 -   [ ] Tests written for new code (and old code if feasible)
--   [ ] New or updated `public`/`exported` symbols have accurate [Typedoc](https://typedoc.org/guides/overview) documentation.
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,5 +3,6 @@
 ## Checklist
 
 -   [ ] Tests written for new code (and old code if feasible)
+-   [ ] New or updated `public`/`exported` symbols have accurate [Typedoc](https://typedoc.org/guides/overview) documentation.
```
robintown, line 6: Would it be clearer to reference [TSDoc](https://tsdoc.org/) here rather than Typedoc? I don't believe we use Typedoc on the other repositories, for instance.

richvdh: errr well... js-sdk certainly uses Typedoc, so I think this is correct here.

I'm not sure we actually process the doc-comments on the other repos, so it's less well defined, but I'd argue that using a different documentation system for different repos would be a poor decision.

robintown (approved):

PR #4084, opened 2024-02-26
Add event with m.reference relation to unknown parent to the timeline.

maheichyk:
Currently event with `m.reference` relation to unknown parent event is not added to the timeline and therefore is not feed into the widget. 

`StopGapWidget.ts` checks event to be in the timeline before the marker and if event is not found and marker is there then the event is ignored.

We in the widgets rely on this setup that a room event relates to another parent room event via `m.reference` and if parent event is not loaded (that happens when timeline is relatively big and user have to scroll timeline manually to get event loaded) the event is ignored.

Therefore this PR suggest to add this event to the timeline.

## Checklist

Type: defect


maheichyk:
@dbkr, @robintown I somehow cannot add a label (adding `Type: defect` to the PR description and I think it was working before to me) but all other checks are passing, please review.

src/models/room.ts
```diff
@@ -2197,7 +2197,7 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         // Due to replies not being typical relations and being used as fallbacks for threads relations
         // If we bypass the if case above then we know we are not a thread, so if we are still a reply
         // then we know that we must be in the main timeline. Same goes if we have no associated parent event.
-        if (!parentEventId || !!event.replyEventId) {
+        if (!parentEventId || !!event.replyEventId || event.getRelation()?.rel_type === "m.reference") {
```
dbkr, line 2200: I'm not sure I really see why an `m.reference` relation needs special treatment here: doesn't the logic apply to any relation that's not a thread (or possibly reply?) relation? Seems like any event with a kind of relation we don't know about, we should assume lives in the main thread? That said, I'd appreciate a second pair of eyes on this before actually suggesting this.

Also, the comment above (apart from having half a sentence) explains both clauses in the if condition, so if you add another, please update the comment to reflect it.

dbkr (changes requested):

src/models/room.ts
dbkr: At the moment I'm more inclined to go with the other PR, which seems broadly sensible. The more I think about this, the more I think that actually there's no way we can legitimately say events live in any timeline if we don't know about their parent, so what we should actually be doing is keeping them in a holding area somewhere until we can fetch their parent. That or use https://github.com/matrix-org/matrix-spec-proposals/pull/4023

maheichyk:
The main problem in relation to the widgets was resolved by another PR to matrix-react-sdk, although would be nice to have these events in the timeline to be shown by Element and not to have a special handling for the widgets.

t3chguy:
@maheichyk a very related issue is tracked in https://github.com/element-hq/element-web/issues/27132

maheichyk:
> @maheichyk a very related issue is tracked in [element-hq/element-web#27132](https://github.com/element-hq/element-web/issues/27132)

Yes, it is the same issue to me: event (with custom relation type) is ignored because the parent event is missing (not loaded yet) in the timeline. In our widget case I could see our custom events in the timeline when they are added by the widget with devtools, but when Element is reloaded I could not see any of those. I though that it was a synapse issue, but it was not, all widgets events could be loaded fine via relation to this parent event.

t3chguy:
Its a tricky topic, the spec doesn't clearly outline how non-standard relations should interact with threads, and getting this wrong means stuck notifications due to the client & server treating those events as contained in different timelines. Ideally the client would be told which thread/main timeline the event belongs to deterministically (either top level thread_id or a value in unsigned) https://github.com/matrix-org/matrix-spec-proposals/pull/4023 would be such a solution

PR #4086, opened 2024-02-28
Make sending ContentLoaded optional for a widgetClient

toger5:
For element call we want to have control over the sendContentLoaded action. As soon as we sent this action the widget will be shown.
However we need to adjust the theme first before we show the widget to not have an awkward flash of the wrong theme during the widget loading procedure.
So this should not be sent by the js-sdk and instead be sent by element call after the theme has been setup.

## Checklist

### DIFF
--- a/src/embedded.ts
+++ b/src/embedded.ts
@@ -108,11 +108,21 @@ export class RoomWidgetClient extends MatrixClient {
     private lifecycle?: AbortController;
     private syncState: SyncState | null = null;
 
+    /**
+     *
+     * @param widgetApi - the widget api to use for the embedded "fake" client
+     * @param opts - Client related options
+     * @param sendContentLoaded - Whether to send a content loaded event.
+     *   Set to false if: the widget uses waitForIFrameLoad=true, or if the
+     *   the widget wants to send the ContentLoaded action at a later point in time
+     *   after initial setup.
+     */
     public constructor(
         private readonly widgetApi: WidgetApi,
         private readonly capabilities: ICapabilities,
         private readonly roomId: string,
         opts: IMatrixClientCreateOpts,
+        sendContentLoaded = true,
     ) {
         super(opts);
 
@@ -165,7 +175,7 @@ export class RoomWidgetClient extends MatrixClient {
         // does *not* (yes, that is the right way around) wait for this event. Let's
         // start sending this, then once this has rolled out, we can change element-web to
         // use waitForIFrameLoad=false and have a widget API that's less racy.
-        widgetApi.sendContentLoaded();
+        if (sendContentLoaded) widgetApi.sendContentLoaded();
     }
 
     public async startClient(opts: IStartClientOpts = {}): Promise<void> {
--- a/src/embedded.ts
+++ b/src/embedded.ts
@@ -121,7 +121,7 @@ export class RoomWidgetClient extends MatrixClient {
         private readonly capabilities: ICapabilities,
         private readonly roomId: string,
         opts: IMatrixClientCreateOpts,
-        sendContentLoaded = true,
+        sendContentLoaded: boolean,
     ) {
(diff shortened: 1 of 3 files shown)


src/embedded.ts
```diff
@@ -108,11 +108,21 @@ export class RoomWidgetClient extends MatrixClient {
...
     private syncState: SyncState | null = null;
 
+    /**
+     *
+     * @param widgetApi - the widget api to use for the embedded "fake" client
+     * @param opts - Client related options
+     * @param sendContentLoaded - Whether to send a content loaded event.
+     *   Set to false if: the widget uses waitForIFrameLoad=true, or if the
+     *   the widget wants to send the ContentLoaded action at a later point in time
+     *   after initial setup.
```
richvdh, line 118: Is there a difference between a "content loaded event" and a "ContentLoaded action"? If not, could you use the same name consistently? I couldn't really find much documentation about the widget API so I don't really know what they mean. Is there any documentation that we could link to? Or can we at least use the full name of a symbol in the matrix-widget-api?

Please proofread this yourself: there are two "the"s in the sentence.

Suggest more use of backticks around things that are names of javascript identifiers, such as `false` and `waitForIFrameLoad=true`

richvdh (changes requested):
There don't seem to be any tests for this - shouldn't there be?

src/embedded.ts
```diff
@@ -165,7 +174,7 @@ export class RoomWidgetClient extends MatrixClient {
         // does *not* (yes, that is the right way around) wait for this event. Let's
         // start sending this, then once this has rolled out, we can change element-web to
         // use waitForIFrameLoad=false and have a widget API that's less racy.
-        widgetApi.sendContentLoaded();
+        if (sendContentLoaded) widgetApi.sendContentLoaded();
```
dbkr, line 177: Fwiw I think the idea here is that it could just send this whether or not the other side needed it. There's no reason not to send it, as far as I can see. If the other side isn't waiting for it, it will just ignore it, so this feels like another moving part that can break.

toger5: I think the place/time where to send the content_loaded event is a "per app" decision. Each app might want to be considered done loading at a different point in time.
Most webapps have their own loading screen. If we have the widgetloading -> the app loading -> the app we have lots of screen changes. In a widget it would be much more elegant (and I think this is also one of the ideas behind the content-loaded event) if the widget loading screen would just be shown longer and the app's loading screen is skipped entirely.

dbkr: Yep, this is why I was thinking it ought to just be the widget itself that sends it (or decides when to send it). This could be a step towards that though, I assume?

toger5: Yes. Seeing it as a step towards this makes sense. Its still in the js-sdk and not the widget api repo. So a "just works without knowing how the widget api works in detail" is what we want to have here imo. Thats why having the default so that the js-sdk sends it after it has initialized all the jssdk specifc things is good I think.

My opinion in summary: the create fn optional param is a good compromise between hiding most of the widget specifics but also allowing app devs to interveine if they know what they are doing.

dbkr: This being in the js-sdk makes this even weirder IMO: the js-sdk is abstracting away a part of the widget API is... odd. I'm not really suggesting changing it now though, so this is probably fine.

dbkr (approved):

src/embedded.ts
toger5: For the embedded matroska mode this is still what we want right? The js sdk is wrapping the widget-api and makes it as simple to use as the js sdk itself.

src/embedded.ts
toger5: thanks. We should also always call it an action to make it clear if we refer to a widget message (=action) or a matrix message/event (=event).

PR #4088, opened 2024-02-29
Don't re-fetch thread root if we already have it

dbkr:
The root event of a thread used to arrive with the pagination request, but this was unspecced and so got changed to simply fetch the root event. In many (almost all) cases this shouldn't be necessary because the thread should already have its root event: re-use it if it's already there. This is only in pagination, so there's no reason to believe that the root event would have changed and needs to be re-fetched.

This removes a number of duplicate calls to the /event/ endpoint from the tests.

## Checklist


t3chguy (approved):
Looks sane

dbkr:
> > This removes a number of duplicate calls to the /event/ endpoint from the tests.
> 
> And from the actual application, right? RIGHT?

Indeed, yes, should be slightly fewer requests.

PR #4089, opened 2024-02-29
Fix race condition with sliding sync extensions

zzorba:
While debugging verification when using sliding sync, I encountered the same events being processed multiple times, causing the VerificationRequest state machine to break (some events are reentrant, but ready -> ready is a cancellable error). 

I noticed that the `onResponse` for `ExtensionToDevice` was declared `void` but overridden as `Promise<void>`, with the `nextBatch` cursor getting updated at the END of the function (after an await). In local testing, it was observed that next `onRequest` would sometimes start before the previous `onResponse` was finished, leading to the same `since` parameter being sent, leading to the same events being processed multiple times.

## Checklist

-   [X] Linter and other CI checks pass
-   [X] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md))

### DIFF
--- a/spec/integ/sliding-sync.spec.ts
+++ b/spec/integ/sliding-sync.spec.ts
@@ -107,8 +107,8 @@ describe("SlidingSync", () => {
                 onRequest: (initial) => {
                     return { initial: initial };
                 },
-                onResponse: (res) => {
-                    return {};
+                onResponse: async (res) => {
+                    return;
                 },
                 when: () => ExtensionState.PreProcess,
             };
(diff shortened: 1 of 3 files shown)


jryans:
I can confirm that device verification while using sliding sync (with legacy / non-Rust crypto) currently aborts. With this PR applied, device verification works when using sliding sync.

kegsay (approved):
Excellent catch, thank you!

t3chguy (approved):

PR #4100, opened 2024-03-06
Add `.m.rule.is_room_mention` push rule to DEFAULT_OVERRIDE_RULES

t3chguy:
To handle the client not knowing it due to the server not proactively telling us down incremental syncs

### DIFF
--- a/spec/unit/pushprocessor.spec.ts
+++ b/spec/unit/pushprocessor.spec.ts
@@ -188,6 +188,8 @@ describe("NotificationService", function () {
     });
 
     it("should add default rules in the correct order", () => {
+        // By the time we get here, we expect the PushProcessor to have merged the new .m.rule.is_room_mention rule into the existing list of rules.
+        // Check that has happened, and that it is in the right place.
         const containsDisplayNameRuleIdx = matrixClient.pushRules?.global.override?.findIndex(
             (rule) => rule.rule_id === RuleId.ContainsDisplayName,
         );
--- a/spec/unit/pushprocessor.spec.ts
+++ b/spec/unit/pushprocessor.spec.ts
@@ -29,147 +29,150 @@ describe("NotificationService", function () {
         actions: ["notify", { set_tweak: "sound", value: "default" }],
     };
 
-    // These would be better if individual rules were configured in the tests themselves.
-    const matrixClient = {
-        getRoom: function () {
-            return {
-                currentState: {
-                    getMember: function () {
-                        return {
-                            name: testDisplayName,
-                        };
-                    },
-                    getJoinedMemberCount: function () {
-                        return 0;
-                    },
-                    members: {},
-                },
-            };
-        },
-        ...mockClientMethodsUser(testUserId),
-        supportsIntentionalMentions: () => true,
-        pushRules: {
-            device: {},
-            global: {
-                content: [
-                    {
-                        actions: [
-                            "notify",
-                            {
-                                set_tweak: "sound",
-                                value: "default",
-                            },
-                            {
-                                set_tweak: "highlight",
-                            },
-                        ],
-                        enabled: true,
-                        pattern: "ali",
-                        rule_id: ".m.rule.contains_user_name",
-                    },
-                    {
-                        actions: [
-                            "notify",
-                            {
-                                set_tweak: "sound",
-                                value: "default",
-                            },
-                            {
-                                set_tweak: "highlight",
-                            },
-                        ],
-                        enabled: true,
-                        pattern: "coffee",
-                        rule_id: "coffee",
-                    },
-                    {
-                        actions: [
-                            "notify",
-                            {
-                                set_tweak: "sound",
-                                value: "default",
-                            },
-                            {
-                                set_tweak: "highlight",
-                            },
-                        ],
-                        enabled: true,
-                        pattern: "foo*bar",
-                        rule_id: "foobar",
-                    },
-                ],
-                override: [
-                    {
-                        actions: [
-                            "notify",
-                            {
-                                set_tweak: "sound",
-                                value: "default",
-                            },
-                            {
-                                set_tweak: "highlight",
-                            },
-                        ],
-                        conditions: [
-                            {
-                                kind: "contains_display_name",
-                            },
-                        ],
-                        enabled: true,
-                        rule_id: ".m.rule.contains_display_name",
-                    },
-                    {
-                        actions: [
-                            "notify",
-                            {
-                                set_tweak: "sound",
-                                value: "default",
-                            },
-                        ],
-                        conditions: [
-                            {
-                                is: "2",
-                                kind: "room_member_count",
-                            },
-                        ],
-                        enabled: true,
-                        rule_id: ".m.rule.room_one_to_one",
-                    },
-                ],
-                room: [],
-                sender: [],
-                underride: [
-                    msc3914RoomCallRule,
-                    {
-                        actions: ["dont-notify"],
-                        conditions: [
-                            {
-                                key: "content.msgtype",
-                                kind: "event_match",
-                                pattern: "m.notice",
-                            },
-                        ],
-                        enabled: true,
-                        rule_id: ".m.rule.suppress_notices",
-                    },
-                    {
-                        actions: [
-                            "notify",
-                            {
-                                set_tweak: "highlight",
-                                value: false,
-                            },
-                        ],
-                        conditions: [],
-                        enabled: true,
-                        rule_id: ".m.rule.fallback",
-                    },
-                ],
-            },
-        },
-    } as unknown as MatrixClient;
+    let matrixClient: MatrixClient;
 
     beforeEach(function () {
+        // These would be better if individual rules were configured in the tests themselves.
+        matrixClient = {
+            getRoom: function () {
+                return {
+                    currentState: {
+                        getMember: function () {
+                            return {
+                                name: testDisplayName,
+                            };
+                        },
+                        getJoinedMemberCount: function () {
+                            return 0;
+                        },
+                        members: {},
+                    },
+                };
+            },
+            ...mockClientMethodsUser(testUserId),
+            supportsIntentionalMentions: () => true,
+            pushRules: {
+                device: {},
+                global: {
+                    content: [
+                        {
+                            actions: [
+                                "notify",
+                                {
+                                    set_tweak: "sound",
+                                    value: "default",
+                                },
+                                {
+                                    set_tweak: "highlight",
+                                },
+                            ],
+                            enabled: true,
+                            pattern: "ali",
+                            rule_id: ".m.rule.contains_user_name",
+                        },
+                        {
+                            actions: [
+                                "notify",
+                                {
+                                    set_tweak: "sound",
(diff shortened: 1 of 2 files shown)


src/pushprocessor.ts
```diff
@@ -90,6 +90,73 @@ const DEFAULT_OVERRIDE_RULES: IPushRule[] = [
...
+    {
+        // Matrix v1.7
+        rule_id: ".m.rule.is_user_mention",
+        default: true,
+        enabled: true,
+        conditions: [
+            {
+                kind: ConditionKind.EventPropertyContains,
+                key: "content.m\\.mentions.user_ids",
+                value: "[the user's Matrix ID]",
```
richvdh, line 102: erm; is this actually going to work? Don't we need to substitute the user id here?

There is some magic for handling `is_user_mention` in `pushprocessor.ts`. maybe we need to extend it?

src/pushprocessor.ts
```diff
@@ -90,6 +90,73 @@ const DEFAULT_OVERRIDE_RULES: IPushRule[] = [
...
+            PushRuleActionName.Notify,
+            {
+                set_tweak: TweakName.Sound,
+                value: "default",
+            },
+            {
+                set_tweak: TweakName.Highlight,
+            },
+        ],
+    },
```
richvdh, line 136: This is not new in v1.7; the change in v1.7 is that it is *deprecated*. I don't think we need to add it here.

src/pushprocessor.ts
```diff
@@ -90,6 +90,73 @@ const DEFAULT_OVERRIDE_RULES: IPushRule[] = [
...
```
richvdh, line 115: This needs slotting in before, rather than after, a bunch of existing rules. I think we need more intelligence than just appending the new rules to the existing list.

richvdh (changes requested):

richvdh (changes requested):
this still needs to happen before existing push rules?

t3chguy:
> this still needs to happen before existing push rules?

The spec doesn't state that the defaults are ordered as far as I can tell, we can't assume what order the server actually holds the rules in. The spec says rules can specify an ordering priority within a kind and that `.m.rule.master` beats all priority, nothing else I can spot

richvdh:
> we can't assume what order the server actually holds the rules in

You absolutely can.

I'm afraid that, whatever the words of the spec might say, the ordering of the push rules is *very very important*. If that's not obvious from the spec, then clarification PRs welcome, but it doesn't change anything here.

Note that MSC3952 explicitly says where the new push rule fits in:

> The `.m.rule.is_room_mention` override push rule would appear **directly before the `.m.rule.roomnotif` push rule:**

t3chguy:
In which case the current code is already not spec compliant as it appends `.m.rule.reaction` as it should occur before `.m.rule.room.server_acl` and `.m.rule.suppress_edits` and we're not making the situation any worse yet still fixing bugs.

richvdh (changes requested):
well, `.m.rule.reaction` is safe to reorder with respect to `.m.rule.room.server_acl` at least, because they cannot both match on the same event. `.m.rule.is_room_mention` is quite a bit earlier so I don't think we can et away with the same bodge.

We just need a list of the `rule_ids` for all the override rules, so that pushprocessor.ts can put them in the right order.

t3chguy:
> well, .m.rule.reaction is safe to reorder with respect to .m.rule.room.server_acl at least, because they cannot both match on the same event. .m.rule.is_room_mention is quite a bit earlier so I don't think we can et away with the same bodge.

The client ignores `.m.rule.is_room_mention` if the event has `m.mentions` so this should be fine

richvdh (commented):
> > well, .m.rule.reaction is safe to reorder with respect to .m.rule.room.server_acl at least, because they cannot both match on the same event. .m.rule.is_room_mention is quite a bit earlier so I don't think we can et away with the same bodge.
> 
> The client ignores `.m.rule.is_room_mention` if the event has `m.mentions` so this should be fine
> 
Ok, but what about the other rules that come after `.m.rule.is_room_mention`? They might be fine but please could you add a comment explaining why it's safe for each of them to happen before `.m.rule.is_room_mention` ?

> > We just need a list of the rule_ids for all the override rules, so that pushprocessor.ts can put them in the right order.
> 
> Wouldn't this break if the spec adds a new rule inbetwixt other existing rules especially when MSC-namespaced rules are also in play, as they currently are in the hardcoded list

I don't follow, I'm afraid. I'm thinking of an algorithm that iterates through the current rules, comparing them against the expected rules, and inserting the new ones where needed.

```js
EXPECTED_OVERRIDE_RULE_IDS = [".m.rule.master", ".m.rule.suppress_notices", /* ... */ ".m.rule.suppress_edits"];

let next_expected_rule_id_index = 0;
let new_rules = [];
for (rule of current_rules) {
    if (!rule.default) {
        // got to the end of the default list
        break;
    }
    const rule_index = EXPECTED_OVERRIDE_RULE_IDS.indexOf(rule.rule_id);
    if (rule_index === -1) {
        // an unrecognised rule; copy it over
        new_rules.push(rule);
        continue;
    }
    while (rule_index > next_expected_rule_index) {
        // insert new rules
        new_rules.push(EXPECTED_OVERRIDE_RULE_IDS[next_expected_rule_index]);
        next_expected_rule_index += 1;
    }    
    // copy over the existing rule
    new_rules.push(rule);
    next_expected_rule_index += 1;
}

// TODO: now copy over any remaining EXPECTED_OVERRIDE_RULE_IDS, and finally any
// non-default rules that were in `current_rules`.
```

Now, if the spec adds a new rule which we don't know about but the server does, we'll just copy it over. Right? What's the failure mode you're thinking of?

[yes, the need to do all of this complexity is horrible. Push rules in general are horrible, and I wish we'd get rid of them rather than continually trying to patch the system with "just one more push rule". But here we are.]

t3chguy:
> Now, if the spec adds a new rule which we don't know about but the server does, we'll just copy it over. Right? What's the failure mode you're thinking of?

If the server adds Y between X & Z (X, Y, Z) but we already have a local MSC rule M in order X, M, Z. Now we don't know the total order, is it XMYZ or XYMZ, one may cause buggy behaviour

richvdh:
I mean, that's kinda on you for adopting MSC rules before they are specced. If the MSC doesn't take a view on whether it goes before or after Y, then everything implementing that MSC is basically broken because we have no idea what different implementations will do.

Presumably that shouldn't be a problem in practice, because by the time Y is specced, M has either itself been specced, or has failed to do so (in which case it seems like high time to rip it out of the codebase).

Anyway, can't we cross that bridge if/when we come to it? It's all a bit hypothetical and doesn't seem of much relevance to how we can implement *this particular change* correctly?

src/pushprocessor.ts
```diff
@@ -112,6 +151,60 @@ const DEFAULT_UNDERRIDE_RULES: IPushRule[] = [
...
+            // an unrecognised rule; copy it over
+            newRules.push(rule);
+            continue;
+        }
+        while (ruleIndex > nextExpectedRuleIdIndex) {
+            // insert new rules
+            const defaultRuleId = defaultRuleIds[nextExpectedRuleIdIndex];
+            const defaultRule = defaultRules.find((r) => r.rule_id === defaultRuleId);
+            if (defaultRule) {
+                logger.warn(`Adding default global ${defaultRule.rule_id}`);
```
richvdh, line 184: I suppose we should warn even louder if `defaultRule` is missing, as it suggests that a rule was added to `EXPECTED_DEFAULT_OVERRIDE_RULE_IDS` without an entry in `DEFAULT_OVERRIDE_RULES`?

src/pushprocessor.ts
```diff
@@ -175,53 +268,17 @@ export class PushProcessor {
...
-                override.conditions![0].value = userId;
-            } else {
-                override = originalOverride;
-            }
-
-            if (existingRule) {
-                // Copy over the actions, default, and conditions. Don't touch the user's preference.
-                existingRule.default = override.default;
-                existingRule.conditions = override.conditions;
-                existingRule.actions = override.actions;
```
richvdh, line 200: I fear we still need all this shit.

spec/unit/pushprocessor.spec.ts
```diff
@@ -184,6 +187,24 @@ describe("NotificationService", function () {
         pushProcessor = new PushProcessor(matrixClient);
     });
 
+    it("should add default rules in the correct order", () => {
```
richvdh, line 190: I was expecting to see a call to `rewriteDefaultRules` or something here, but I now realise that is happening in the `beforeEach`.

"By the time we get here, we expect the PushProcessor to have merged the new `.m.rule.is_room_mention` rule into the existing list of rules. Check that has happened, and that it is in the right place."

Is that right? If so, can you put it in a comment?

richvdh (changes requested):
Generally I feel like this is a bit under-tested, and it would lend itself well to unit tests which directly call `rewriteDefaultRules`. The special-casing of `is_user_mention`, and the fact that we update `conditions` and `actions` seems important, and I'm worried that we can rip it out without the tests caring.

t3chguy:
>  The special-casing of is_user_mention, and the fact that we update conditions and actions seems important, and I'm worried that we can rip it out without the tests caring

Sorry, what special-casing?

src/pushprocessor.ts
t3chguy: The IsUserMention stuff is not needed because we don't have a default IsUserMention rule in the map, so we don't need to update its user id condition

The copying over actions etc is questionable, as we can't determine whether we or the server have the latest and greatest conditions/actions for a given rule

src/pushprocessor.ts
```diff
@@ -112,6 +151,60 @@ const DEFAULT_UNDERRIDE_RULES: IPushRule[] = [
...
+            continue;
+        }
+        while (ruleIndex > nextExpectedRuleIdIndex) {
+            // insert new rules
+            const defaultRuleId = defaultRuleIds[nextExpectedRuleIdIndex];
+            const defaultRule = defaultRules.find((r) => r.rule_id === defaultRuleId);
+            if (defaultRule) {
+                logger.warn(`Adding default global ${defaultRule.rule_id}`);
+                newRules.push(defaultRule);
+            }
```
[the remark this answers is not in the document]
t3chguy: It is now only 3 lines and don't think it warrants an inner function

src/pushprocessor.ts
richvdh: > The IsUserMention stuff is not needed because we don't have a default IsUserMention rule in the map, so we don't need to update its user id condition

Huh. Good point. WTF was it doing there in the first place?

> The copying over actions etc is questionable, as we can't determine whether we or the server have the latest and greatest conditions/actions for a given rule

yeah, ok. I don't know if there are any instances of conditions/actions having changed on an existing specced rule, but it's certainly been a while and we'd need to have a careful think about it.

richvdh:
> Sorry, what special-casing?

The special-casing that you [just convinced me we don't need](https://github.com/matrix-org/matrix-js-sdk/pull/4100#discussion_r1517693296)

richvdh (approved):
LGTM :tada:

PR #4104, opened 2024-03-07
WIP fix for bugs first-thread-reply bugs

dbkr:
## Checklist

### DIFF
--- a/spec/unit/room.spec.ts
+++ b/spec/unit/room.spec.ts
@@ -2758,7 +2758,6 @@ describe("Room", function () {
             // XXX: If we add the relation to the thread response before the thread finishes fetching via /relations
             // then the test will fail
             await emitPromise(room, ThreadEvent.Update);
-            await emitPromise(room, ThreadEvent.Update);
             await Promise.all([emitPromise(room, ThreadEvent.Update), room.addLiveEvents([threadResponseEdit])]);
             expect(thread.replyToEvent!.getContent().body).toBe(threadResponseEdit.getContent()["m.new_content"].body);
         });
@@ -2791,7 +2790,6 @@ describe("Room", function () {
             let prom = emitPromise(room, ThreadEvent.New);
             await room.addLiveEvents([threadRoot, threadResponse1, threadResponse2]);
             const thread = await prom;
-            await emitPromise(room, ThreadEvent.Update);
 
             expect(thread).toHaveLength(2);
             expect(thread.replyToEvent.getId()).toBe(threadResponse2.getId());
@@ -2856,6 +2854,10 @@ describe("Room", function () {
                     },
                 });
 
+            room.client.fetchRelations = jest.fn().mockResolvedValue({
+                chunk: [threadResponse2Reaction.event, threadResponse2.event, threadResponse1.event],
+            });
+
             const prom = emitPromise(room, ThreadEvent.New);
             await room.addLiveEvents([threadRoot, threadResponse1, threadResponse2, threadResponse2Reaction]);
             const thread = await prom;
@@ -2896,18 +2898,20 @@ describe("Room", function () {
                     },
                 });
 
-            let prom = emitPromise(room, ThreadEvent.New);
+            const prom = emitPromise(room, ThreadEvent.New);
             await room.addLiveEvents([threadRoot, threadResponse1, threadResponse2, threadResponse2Reaction]);
             const thread = await prom;
-            await emitPromise(room, ThreadEvent.Update);
 
             expect(thread).toHaveLength(2);
             expect(thread.replyToEvent.getId()).toBe(threadResponse2.getId());
 
-            prom = emitPromise(room, ThreadEvent.Update);
             const threadRootRedaction = mkRedaction(threadRoot);
             await room.addLiveEvents([threadRootRedaction]);
-            await prom;
+
+            // We can't wait for a thread update here because there shouldn't be one (which is
+            // what we're asserting). Flush any promises to try to get more certainty that an
+            // update is not happening some time after the event is added.
+            await flushPromises();
             expect(thread).toHaveLength(2);
         });
 
@@ -2985,7 +2989,6 @@ describe("Room", function () {
 
             await emitPromise(room, ThreadEvent.Update);
             const threadResponse2Redaction = mkRedaction(threadResponse2);
-            await emitPromise(room, ThreadEvent.Update);
             await room.addLiveEvents([threadResponse2Redaction]);
             expect(thread).toHaveLength(1);
             expect(thread.replyToEvent!.getId()).toBe(threadResponse1.getId());
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -138,6 +138,7 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
      * that we've already fetched them.
      */
     public initialEventsFetched = !Thread.hasServerSideSupport;
+    private initalEventFetchProm: Promise<boolean> | undefined;
 
     /**
      * An array of events to add to the timeline once the thread has been initialised
@@ -377,15 +378,15 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
             // When there's no server-side support, just add it to the end of the timeline.
             this.addEventToTimeline(event, toStartOfTimeline);
             this.client.decryptEventIfNeeded(event);
-        } else if (!toStartOfTimeline && this.initialEventsFetched && isNewestReply) {
+        } else if (event.isRelation(RelationType.Annotation) || event.isRelation(RelationType.Replace)) {
+            this.addRelatedThreadEvent(event, toStartOfTimeline);
+            return;
+        } else if (!toStartOfTimeline && isNewestReply) {
             // When we've asked for the event to be added to the end, and we're
             // not in the initial state, and this event belongs at the end, add it.
             this.addEventToTimeline(event, false);
             this.fetchEditsWhereNeeded(event);
-        } else if (event.isRelation(RelationType.Annotation) || event.isRelation(RelationType.Replace)) {
-            this.addRelatedThreadEvent(event, toStartOfTimeline);
-            return;
-        } else if (this.initialEventsFetched) {
+        } else {
             // If initial events have not been fetched, we are OK to throw away
             // this event, because we are about to fetch all the events for this
             // thread from the server.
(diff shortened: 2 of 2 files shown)


spec/unit/room.spec.ts
```diff
@@ -2985,7 +3020,6 @@ describe("Room", function () {
 
             await emitPromise(room, ThreadEvent.Update);
             const threadResponse2Redaction = mkRedaction(threadResponse2);
-            await emitPromise(room, ThreadEvent.Update);
```
dbkr, line 2988: I can't work out why this would have been here: why would we expect simply making the event to trigger an update?

src/models/thread.ts
```diff
@@ -606,29 +607,36 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
...
+                    if (this.replyCount === 0 && this.rootEvent) {
+                        this.timelineSet.addEventsToTimeline([this.rootEvent], true, this.liveTimeline, null);
+                        this.liveTimeline.setPaginationToken(null, Direction.Backward);
+                    } else {
+                        this.initalEventFetchProm = this.client.paginateEventTimeline(this.liveTimeline, {
+                            backwards: true,
+                        });
+                        await this.initalEventFetchProm;
+                    }
+                    this.initialEventsFetched = true;
```
richvdh, line 628: I'm assuming delaying this is deliberate? But why?

src/models/thread.ts
```diff
@@ -377,15 +378,15 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
...
             // When we've asked for the event to be added to the end, and we're
             // not in the initial state, and this event belongs at the end, add it.
             this.addEventToTimeline(event, false);
             this.fetchEditsWhereNeeded(event);
-        } else if (event.isRelation(RelationType.Annotation) || event.isRelation(RelationType.Replace)) {
-            this.addRelatedThreadEvent(event, toStartOfTimeline);
-            return;
-        } else if (this.initialEventsFetched) {
+        } else {
             // If initial events have not been fetched, we are OK to throw away
```
richvdh, line 390: we're no longer checking if initial events have been fetched before hitting this code?

richvdh (changes requested):
I'm afraid I'm struggling to follow this.

> One: the notification dot on a thread would not show the appropriate notification colour for a notification on the first message of a thread: this was caused by the fact that when a thread is created, the thread would not actually contain the first reply until the async pagination had completed and the events added that way. This meant that when fixNotificationCounts came to compute the dot colour, it got it wrong because the thread was empty.

Wait, notifications come in different colours? Since when? What do the colours mean?

Generally, I'm struggling to correlate the text in the PR description to the actual changes in the PR. Could you help me out? Maybe separate commits, or comments on the diff?

src/models/thread.ts
dbkr: Correct: this is the main part of the fix. We previously checked whether initial events had been fetched and, if they hadn't threw the event we were going to add in the bin, in favour of waiting for it to arrive from the pagination. This is what caused the races, because any code that then ran to check the state of the thread would find that it was empty.

src/models/thread.ts
dbkr: Comment added in 22ff5013e & b3bbc85bf which hopefully explains things.

src/models/room.ts
```diff
@@ -2307,6 +2307,15 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
 
+        // Add the re-emitter before we start adding events to the thread so we don't miss events
+        this.reEmitter.reEmit(thread, [
+            ThreadEvent.Delete,
+            ThreadEvent.Update,
+            ThreadEvent.NewReply,
+            RoomEvent.Timeline,
+            RoomEvent.TimelineReset,
+        ]);
+
```
dbkr, line 2318: This fixes the notification bug (and is also generally much saner as far as I'm concerned): add the re-emitter straight away so it doesn't miss (node) events, rather then after we add (matrix) events.

src/models/thread.ts
```diff
@@ -377,15 +378,15 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
             // When there's no server-side support, just add it to the end of the timeline.
             this.addEventToTimeline(event, toStartOfTimeline);
             this.client.decryptEventIfNeeded(event);
-        } else if (!toStartOfTimeline && this.initialEventsFetched && isNewestReply) {
+        } else if (event.isRelation(RelationType.Annotation) || event.isRelation(RelationType.Replace)) {
+            this.addRelatedThreadEvent(event, toStartOfTimeline);
+            return;
```
dbkr, line 383: This clause has just moved up because otherwise relations would now have fallen into the other clause: they need their own special handling.

src/models/thread.ts
```diff
@@ -606,29 +607,36 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
...
-                    this.timelineSet.addEventsToTimeline([this.rootEvent], true, this.liveTimeline, null);
-                    this.liveTimeline.setPaginationToken(null, Direction.Backward);
-                } else {
-                    await this.client.paginateEventTimeline(this.liveTimeline, {
-                        backwards: true,
-                    });
-                }
-                for (const event of this.replayEvents!) {
-                    this.addEvent(event, false);
+            if (this.initalEventFetchProm) {
```
dbkr, line 610: This whole block of change is just fixing the race around `this.initialEventsFetched` which was set as soon as we started trying to fetch initial events, not once we had actually fetched them.

dbkr:
> Wait, notifications come in different colours? Since when? What do the colours mean?

Well, red & green, ie. highlight & notification respectively. 

> Generally, I'm struggling to correlate the text in the PR description to the actual changes in the PR. Could you help me out? Maybe separate commits, or comments on the diff?

Yeah, I don't blame you: this is the result of at least a week of me trying to understand what was going on here. I attempted to split the changes out, but they are quite intertwined and I failed to get the code into enough of a working state with part of the changes. I've added some diff comments which hopefully help.

src/models/thread.ts
richvdh: ok, but doesn't that mean that this comment is no longer relevant?

Also, I haven't fully grokked the change or the comment, but it feels like the block comment between lines 92 and 123 might need a re-visit?

src/models/thread.ts
```diff
@@ -377,15 +378,15 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
             // When there's no server-side support, just add it to the end of the timeline.
             this.addEventToTimeline(event, toStartOfTimeline);
             this.client.decryptEventIfNeeded(event);
-        } else if (!toStartOfTimeline && this.initialEventsFetched && isNewestReply) {
+        } else if (event.isRelation(RelationType.Annotation) || event.isRelation(RelationType.Replace)) {
+            this.addRelatedThreadEvent(event, toStartOfTimeline);
+            return;
+        } else if (!toStartOfTimeline && isNewestReply) {
             // When we've asked for the event to be added to the end, and we're
             // not in the initial state, and this event belongs at the end, add it.
```
richvdh, line 386: >  we're not in the initial state

This is no longer accurate?

src/models/thread.ts
```diff
@@ -606,29 +607,40 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
...
-                    await this.client.paginateEventTimeline(this.liveTimeline, {
-                        backwards: true,
-                    });
-                }
-                for (const event of this.replayEvents!) {
-                    this.addEvent(event, false);
+            if (this.initalEventFetchProm) {
+                await this.initalEventFetchProm;
+            } else {
+                // fetch initial event to allow proper pagination
```
richvdh, line 613: pre-existing, but:

```suggestion
                // fetch initial events to allow proper pagination
```

?

src/models/thread.ts
```diff
@@ -606,29 +607,40 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
...
+                try {
+                    // clear out any events we added previously
+                    this.timelineSet.resetLiveTimeline();
```
richvdh, line 616: Not entirely following this.  Why would we already have events, if this is the first time we are fetching events?

src/models/thread.ts
```diff
@@ -606,29 +607,40 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
...
+                        this.timelineSet.addEventsToTimeline([this.rootEvent], true, this.liveTimeline, null);
+                        this.liveTimeline.setPaginationToken(null, Direction.Backward);
+                    } else {
+                        this.initalEventFetchProm = this.client.paginateEventTimeline(this.liveTimeline, {
+                            backwards: true,
+                        });
+                        await this.initalEventFetchProm;
+                    }
+                    // We have now fetched the initial events, so set the flag. We need to do this before
+                    // we actually add the events, so it knows that it can now safely add them rather than
```
richvdh, line 629: > so it knows that it can now safely add them

what is "it" here? `this.addEvents` ?

src/models/thread.ts
```diff
@@ -606,29 +607,40 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
...
+                    // buffer them in the pending event list. The main thing is that this must remain false
+                    // while the async fetch happens, so we don't try to add events before it's ready.
```
richvdh, line 631: > before it's ready.

again, what is "it" here?

src/models/thread.ts
```diff
@@ -606,29 +607,40 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
...
+                    // if the thread has regular events, this will just load the last reply.
+                    // if the thread is newly created, this will load the root event.
+                    if (this.replyCount === 0 && this.rootEvent) {
+                        this.timelineSet.addEventsToTimeline([this.rootEvent], true, this.liveTimeline, null);
+                        this.liveTimeline.setPaginationToken(null, Direction.Backward);
+                    } else {
+                        this.initalEventFetchProm = this.client.paginateEventTimeline(this.liveTimeline, {
+                            backwards: true,
+                        });
+                        await this.initalEventFetchProm;
```
richvdh, line 626: AIUI, correct operation here relies on the bits after this being synchronous? (Otherwise next time we `await initalEventFetchProm`, the job could only be half-done.) That's ok, but might be worth highlighting. Better yet, can we not refactor this to avoid the footgun?

src/models/thread.ts
richvdh: > Also, I haven't fully grokked the change or the comment, but it feels like the block comment between lines 92 and 123 might need a re-visit?

On a more careful reading, I think the comment on `lastEvent` is still valid.

I still think that *this* comment is outdated.

richvdh (changes requested):

src/models/thread.ts
dbkr: Oh, haha, yes, I see what you mean. Fixed the comment (e2949fcde), and the block comment (059caff35).

src/models/thread.ts
dbkr: ah, good catch. Fixed.

src/models/thread.ts
dbkr: Hopefully 99267313e clarifies.

src/models/thread.ts
dbkr: Yeah, exactly. Expanded in 40b8e2083. I debated refactoring it but was quite deep in the guts already and adding this on too felt like a little too much complexity.

src/models/thread.ts
dbkr: Yep - fixed.

src/models/thread.ts
dbkr: This is more talking about the pagination request / the state of the model in general: clarified in e75e47cd8 I hope.

dbkr:
OK, think I got all the comments on this one: ready for another round - thank you!

src/models/thread.ts
```diff
@@ -377,20 +378,19 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
...
+        } else if (!toStartOfTimeline && isNewestReply) {
+            // When we've asked for the event to be added to the end,
+            // and this event belongs at the end, add it.
+            this.addEventToTimeline(event, false);
+            this.fetchEditsWhereNeeded(event);
+        } else {
             // Otherwise, we should add it, but we suspect it is out of order.
+            // This may be because we've just created the thread object and are
+            // still fetching events, in which case add it where we think is sensible
+            // and it will be replaced later once the pagination request completes.
```
richvdh, line 393: > and it will be replaced later

how will it be replaced? don't we just addEvent each of the events returned by the pagination request? We don't remove any existing events?

Otherwise what is the point of doing `resetLiveTimeline` before the pagination request?

richvdh (commented):
Seems broadly sensible. One question remaining.

src/models/thread.ts
dbkr: I'm not sure I understand. `resetLiveTimeline` removes the events that get added here before they get replaced by the result of the pagination request. Does c45def0e2 make it any clearer?

richvdh: I guess my question really amounts to: is there any way we can hit this codepath while `paginateEventTimeline` is running?

Something like:

 * we start loading the initial events, beginning with `resetLivetTimeline` then the start of a call to `paginateEventTimeline`
 * an event arrives over /sync which is added above
 * we back-paginate from the new event, resulting in new events which hit this codepath
 * the initial event load finally completes.

According to this comment, the events added in step 3 are removed, but that doesn't seem correct?

richvdh (approved):
I'm approving this because if you're happy, I'm happy

src/models/thread.ts
dbkr: Ah right yes, I think I see what you mean: really we would want to call `resetLivetTimeline` after the async bit complete but before the events from the pagination request start to be added, but `paginateEventTimeline` doesn't allow for that: we'd have to split out something separate to allow us to do that. I think I'm inclined to comment this and fix it separately: I don't think it's a trivial thing to fix as part of this PR. Wdyt?

dbkr: Fwiw I have commented this in f0dc6014a

PR #4106, opened 2024-03-11
Fix highlights from threads disappearing on new messages

dbkr:
This changes interface of Room, so this is a BREAKING CHANGE.

Correctly mirrors the logic we use for room notifications for thread notifications, ie. set only the total notifications count from the server if it's zero.

I'm not delighted with this since it ends up with function on room whose contract is to do something frankly, deeply weird and unintuitive. However, this is the hack we use for room notifications and it, empirically, works well enough. To do better, we'd need much more complex logic to overlay notification counts for decrypted messages. It could be less weird if we made the function able to zero out either type of count with an option, but since we'd only actually use it with one type, this seems worse, not better.

Fixes https://github.com/element-hq/element-web/issues/25523

## Checklist

### DIFF
--- a/spec/integ/matrix-client-syncing.spec.ts
+++ b/spec/integ/matrix-client-syncing.spec.ts
@@ -1645,6 +1645,99 @@ describe("MatrixClient syncing", () => {
             });
         });
 
+        it("should zero total notifications for threads when absent from the notifications object", async () => {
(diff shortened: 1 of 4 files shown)


florianduros (approved):

dbkr:
Sorry, this needs another look as I failed to copy the logic correctly. It now does highlights properly (making the behaviour the function even more peculiar).

florianduros (approved):

PR #4107, opened 2024-03-12
Introduce Membership TS type (take 2)

andybalaam:
This is an updated version of https://github.com/matrix-org/matrix-js-sdk/pull/2723. Original description:

> According to the [Matrix Spec](https://spec.matrix.org/v1.4/client-server-api/#room-membership) the room membership is enum with finite numbers of possible states. 

> This PR introduces a new typescript type `Membership` to be able to strictly specify what the exact membership is instead of > the generic `string` type. The scope includes:
> - New Membership type
> - Update type definitions in `Client`, `Room`, `RoomMember` and `EventContent` interfaces
> - Update existing conditions to use the new type instead of hardcoded strings
> - Update spec files with the new type
> 
> Signed-off-by: Stanislav Demydiuk [s.demydiuk@gmail.com](mailto:s.demydiuk@gmail.com)

After a discussion in our weekly meeting, we agreed this should be re-done using an enum and a union with string. This means we don't break any existing code, and we allow membership types that we are not aware of. The downside is that we can only enforce use of the enum using coding standards instead of the compiler.

Changelog:  Introduce Membership TS type ([\#4107](https://github.com/matrix-org/matrix-js-sdk/pull/4107)). Contributed by @stas-demydiuk


src/@types/partials.ts
```diff
@@ -87,3 +87,13 @@ export interface IUsageLimit {
...
+
+export enum KnownMembership {
+    Ban = "ban",
+    Invite = "invite",
+    Join = "join",
+    Knock = "knock",
+    Leave = "leave",
+}
+
+export type Membership = KnownMembership | string;
```
richvdh, line 99: could you give these things docstrings please.

spec/integ/crypto/crypto.spec.ts
```diff
@@ -1242,7 +1243,7 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
                                     content: { algorithm: "m.megolm.v1.aes-sha2" },
                                 }),
                                 testUtils.mkMembership({
-                                    mship: "join",
+                                    mship: KnownMembership.Join,
```
richvdh, line 1246: for the record, I am very ambivalent about using symbolic constants in test code. Using the literal value gives an opportunity to check that the symbolic constant is working as you think it is, and can make the test easier to understand.

(At one point we had a bug in Synapse, where  the unit tests were misusing what we *thought* was a symbolic constant, but which turned out not to be, and it ended up masking a real bug.)

But I don't have a strong feeling. Happy to stick with it now it's done.

src/@types/partials.ts
richvdh: Any idea what `@types/partials` is meant to be about, and how it differs from (say) `@types/event` ?

Also: I slightly wonder whether this should be in the new [types](https://pr4107--js-sdk-docs-previews.netlify.app/modules/types.html) entrypoint rather than adding to the clutter of the main namespace, but then apparently that is meant to be types-only which makes it no god for `enum`s. OTOH, I'm not convinced that defining the new entrypoint to be types-only was a good decision, for this reason.

WDYT?

spec/integ/crypto/crypto.spec.ts
andybalaam: Interesting, I thought I was the only person who thought that way.

I think we should leave it as I've done it, since it matches the approach we use in this codebase, but yeah in general I agree.

src/@types/partials.ts
andybalaam: > Any idea what @types/partials is meant to be about, and how it differs from (say) @types/event ?

No. The original contributor put it here, so I left it.

andybalaam: > I slightly wonder whether this should be in the new [types](https://pr4107--js-sdk-docs-previews.netlify.app/modules/types.html) entrypoint

I haven't been following this conversation closely, so I don't know and am happy to be guided. Is the types-only decision documented somewhere? Surely an enum is a type? Or does "type" here mean "stuff that disappears entirely at compile time"?

t3chguy: An enum is not a type, it maps to an Object literal in JS

andybalaam: Docstrings added in d1cc8e9b1

andybalaam:
@uhoreg @dbkr you were assigned this as code owners of different areas. Any chance you could take a quick look? Thanks!

uhoreg (approved):
Looks fairly sane

richvdh:
> @uhoreg @dbkr you were assigned this as code owners of different areas. 

Just for clarity, as a matter of process: although @uhoreg was nominated as an owner of element-crypto-web-reviewers, a review from *any* member of that group (including me!) is sufficient. We do, however, need a review from a member of element-call-reviewers such as @dbkr ...

robintown (commented):
I can't say I understand the value of this change if it doesn't make the types any stricter, and creates a new code style requirement that we're supposed to enforce manually. The Rust SDK seems to have locked down its membership types like the original PR was proposing - is there anything we can learn from its approach? (like, does it not assume that it'll be compatible with arbitrary room versions from the future?)

andybalaam:
> I can't say I understand the value of this change if it doesn't make the types any stricter, and creates a new code style requirement that we're supposed to enforce manually. The Rust SDK seems to have locked down its membership types like the original PR was proposing - is there anything we can learn from its approach? (like, does it not assume that it'll be compatible with arbitrary room versions from the future?)

I agree that this change would have more value if it strictly enforced the types, but it would also make it impossible for new membership types to be used with matrix-js-sdk. We discussed this in the weekly team meeting and chose this implementation strategy. I do think it offers some value: it makes it easy to see which membership types are available, and it prevents typos in those strings. It should be easy to follow the convention because all the surrounding code uses it.

I think the question you raised would be an excellent thing to discuss in a future meeting or in #element-dev:matrix.org but I don't think we should block this PR based on it: it's a wider question of changing our policy.

t3chguy:
You could always do something like `KnownMembership | Exclude<string, "join" | "leave" | ...>`

andybalaam:
> You could always do something like `KnownMembership | Exclude<string, "join" | "leave" | ...>`

Nice idea, but I think that would be an interface change, right? So I'd advocate for doing it separately from this change.

src/@types/partials.ts
andybalaam: Done in 485f69ccb6682713bf78792a2bf761d0c9afaa30

richvdh (approved):
looks great!

dbkr (approved):

robintown (approved):

PR #4111, opened 2024-03-14
Extend logic for local notification processing to threads

dbkr:
There's collection of logic for for processing receipts and recomputing notifications for encrypted rooms, but we didn't do the same for threads. As a reasult, when I tried pulling some of the logic over in https://github.com/matrix-org/matrix-js-sdk/pull/4106 clearing notifications on threads just broke.

This extends the logic of reprocessing local notifications when a receipt arrives to threads.

Based on https://github.com/matrix-org/matrix-js-sdk/pull/4109

## Checklist

### DIFF
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -1317,46 +1317,79 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         // This fixes https://github.com/vector-im/element-web/issues/9421
 
         // Figure out if we've read something or if it's just informational
+        // We need to work out what threads we've just recieved receipts for, so we
+        // know which ones to update. If we've received an unthreaded receipt, we'll
+        // need to update all threads.
+        let threadIds: string[] = [];
+        let hasUnthreadedReceipt = false;
+
         const content = event.getContent();
-        const isSelf =
-            Object.keys(content).filter((eid) => {
-                for (const [key, value] of Object.entries(content[eid])) {
-                    if (!utils.isSupportedReceiptType(key)) continue;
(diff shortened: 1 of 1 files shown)


dbkr:
This really needs https://github.com/matrix-org/matrix-js-sdk/pull/4104 to land before it can be sensibly tested.

src/models/thread.ts
```diff
@@ -709,7 +709,7 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
      * Return last reply to the thread, if known.
      */
     public lastReply(
-        matches: (ev: MatrixEvent) => boolean = (ev): boolean => ev.isRelation(RelationType.Thread),
+        matches: (ev: MatrixEvent) => boolean = (ev): boolean => ev.isRelation(THREAD_RELATION_TYPE.name),
```
dbkr, line 712: I think this was just broken and was hardcoded to the stable prefix, but the tests use the unstable prefix (because it's stored a s a singleton so once one test set it unstable, the rest will break unless they also use unstable).

src/@types/event.ts
```diff
@@ -136,6 +136,7 @@ export enum RelationType {
     Annotation = "m.annotation",
     Replace = "m.replace",
     Reference = "m.reference",
+    // Don't use this yet: it's only the stable version. Use THREAD_RELATION_TYPE.name.
```
t3chguy, line 139: Can we improve this comment to say *why*? Given `m.thread` is stable now.

t3chguy (approved):
Looks sane otherwise

PR #4112, opened 2024-03-15
Only check lazyLoadMembers with previous session if there was one

langleyd:
Fixes https://github.com/element-hq/element-web/issues/27195

Previously this function would return true if `lazyLoadMembers` changed or if the clientOptions did not exist on the last store and `lazyLoadMembers` was now true.

I think the only case `clientOptions` doesn't exist is the case described in the bug where the execution is halted before the options are stored. 

The options are [stored directly after this function is called](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/sync.ts#L682).

The previous session never progressed to starting the actual sync, therefore I don't think think we need to do a full reset but just allow execution to continue.

### DIFF
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -601,7 +601,7 @@ export class SyncApi {
         if (!isStoreNewlyCreated) {
             const prevClientOptions = await this.client.store.getClientOptions();
             if (prevClientOptions) {
-                let lazyLoadMembersBefore = !!prevClientOptions.lazyLoadMembers;
+                const lazyLoadMembersBefore = !!prevClientOptions.lazyLoadMembers;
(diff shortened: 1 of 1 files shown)


richvdh (approved):
> I think the only case `clientOptions` doesn't exist is the case described in the bug where the execution is halted before the options are stored.

I suspect a second case was when `clientOptions` had just been introduced, but that's long enough ago we can forget it.

The fix seems ok, but afaict most of this logic is there to handle people turning lazy-loading on/off in labs, which hasn't been supported in forever. It feels like we should be ripping out this logic rather than continuing to service the technical debt (and allow it to produce new bugs every few years). 

I'm not going to block this, but I'd be much happier if we could use the opportunity to clean up some dead code.

PR #4113, opened 2024-03-15
ElementR| Retry query backup until it works during migration to avoid spurious correption error popup

BillCarsonFr:
Fixes https://github.com/element-hq/element-web/issues/27196

## Checklist

### DIFF
--- a/src/rust-crypto/libolm_migration.ts
+++ b/src/rust-crypto/libolm_migration.ts
@@ -23,6 +23,8 @@ import { decryptAES, IEncryptedPayload } from "../crypto/aes";
 import { IHttpOpts, MatrixHttpApi } from "../http-api";
 import { requestKeyBackupVersion } from "./backup";
 import { IRoomEncryption } from "../crypto/RoomList";
+import { KeyBackupInfo } from "../crypto-api/keybackup";
+import { sleep } from "../utils";
 
 /**
  * Determine if any data needs migrating from the legacy store, and do so.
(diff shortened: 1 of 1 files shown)


src/rust-crypto/libolm_migration.ts
```diff
@@ -158,7 +160,17 @@ async function migrateBaseData(
...
-        const backupInfo = await requestKeyBackupVersion(http);
+        let backupCallDone = false;
+        let backupInfo: KeyBackupInfo | null = null;
+        while (!backupCallDone) {
+            try {
+                backupInfo = await requestKeyBackupVersion(http);
+                backupCallDone = true;
+            } catch (e) {
+                // Retry until successful, use simple constant delay
+                await sleep(2000);
```
BillCarsonFr, line 171: Figured out it was better to wait a bit, not sure we need exponential backoff though

src/rust-crypto/libolm_migration.ts
```diff
@@ -158,7 +160,17 @@ async function migrateBaseData(
     // If we have a backup recovery key, we need to try to figure out which backup version it is for.
     // All we can really do is ask the server for the most recent version.
     if (recoveryKey) {
-        const backupInfo = await requestKeyBackupVersion(http);
+        let backupCallDone = false;
+        let backupInfo: KeyBackupInfo | null = null;
+        while (!backupCallDone) {
```
andybalaam, line 165: What happens if we are offline or something like that?

src/rust-crypto/libolm_migration.ts
```diff
@@ -158,7 +160,17 @@ async function migrateBaseData(
...
+            try {
+                backupInfo = await requestKeyBackupVersion(http);
+                backupCallDone = true;
+            } catch (e) {
```
andybalaam, line 169: Should we log here? Is this a fully-expected error? I'd be more comfortable if we could see some information about `e` in the log, in case something we didn't expect happens.

src/rust-crypto/libolm_migration.ts
andybalaam: I.e. should we give up eventually?

andybalaam (commented):
Looks good - some questions

src/rust-crypto/libolm_migration.ts
BillCarsonFr: Added a log line

src/rust-crypto/libolm_migration.ts
BillCarsonFr: As discussed with @richvdh, we might as well [block the application until it works](https://github.com/matrix-org/matrix-js-sdk/pull/4090#discussion_r1517805018).
If we want to give up we would need to catch such errors in `LifeCycle#handleLoadSessionFailure` and is probably part of a bigger task for proper support of offline mode on web

andybalaam (approved):

src/rust-crypto/libolm_migration.ts
BillCarsonFr: Not planned for now, needs design and some refactoring

PR #4126, opened 2024-03-21
Clean up code for handling decryption failures

richvdh:
Various improvements, including:
 * Defining an `enum` for decryption failure reasons
 * Exposing the reason code as a property on `Event`

Review commit-by-commit.


src/crypto/algorithms/base.ts
```diff
@@ -211,12 +214,11 @@ export class DecryptionError extends Error {
     public readonly detailedString: string;
 
     public constructor(
-        public readonly code: string,
```
andybalaam, line 214: Is this a breaking change?

andybalaam (approved):
Looks good; one question. Thanks!

src/crypto/algorithms/base.ts
richvdh: Well first of all, this class (and indeed the whole file) is *supposed* to be internal, so anyone using it can keep both halves when it breaks.

In any case: It would only be a breaking change for things *constructing* a `DecryptionError`: anything reading `code` once it is created will implicitly cast the `DecryptionFailureCode` to a string. And I've not been able to find any code constructing a `DecryptionError` outside the js-sdk.

richvdh:
*However*, it turns out the removal of `MatrixEvent.badEncryptedMessage` *is* a breaking change, despite it being a private method, because there is a JS file in the react SDK which uses it. So I need to do a bit more work here.

richvdh:
Once https://github.com/matrix-org/matrix-react-sdk/pull/12364 lands, this should be good to go too.

richvdh:
going to override the coverage gate on this

PR #4128, opened 2024-03-22
Improved types for IPowerLevelsContent and hasSufficientPowerLevelFor

galash13:
while developing to use the hasSufficientPowerLevelFor method i logged the contents of a EventType.RoomPowerLevels state event for a room and found there were missing properties, which i add in this PR

### DIFF
--- a/src/models/room-state.ts
+++ b/src/models/room-state.ts
@@ -60,6 +60,8 @@ export interface IPowerLevelsContent {
     // eslint-disable-next-line camelcase
     state_default?: number;
     ban?: number;
+    historical?: number;
+    invite?: number;
     kick?: number;
     redact?: number;
 }
@@ -800,7 +802,7 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
      * @param powerLevel - The power level of the member
      * @returns true if the given power level is sufficient
      */
-    public hasSufficientPowerLevelFor(action: "ban" | "kick" | "redact", powerLevel: number): boolean {
+    public hasSufficientPowerLevelFor(action: keyof Omit<IPowerLevelsContent, "events" | "users">, powerLevel: number): boolean {
         const powerLevelsEvent = this.getStateEvents(EventType.RoomPowerLevels, "");
 
         let powerLevels: IPowerLevelsContent = {};


src/models/room-state.ts
```diff
@@ -800,7 +802,7 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
      * @param powerLevel - The power level of the member
      * @returns true if the given power level is sufficient
      */
-    public hasSufficientPowerLevelFor(action: "ban" | "kick" | "redact", powerLevel: number): boolean {
+    public hasSufficientPowerLevelFor(action: keyof Omit<IPowerLevelsContent, "events" | "users">, powerLevel: number): boolean {
```
t3chguy, line 805: Can we use a type like

```
type KeysMatching<T, V> = {[K in keyof T]-?: T[K] extends V ? K : never}[keyof T];
type Action = KeysMatching<IPowerLevelsContent, number>;

File: src/models/room-state.ts [hunk 2]
```diff
@@ -60,6 +60,8 @@ export interface IPowerLevelsContent {
     // eslint-disable-next-line camelcase
     state_default?: number;
     ban?: number;
+    historical?: number;
```
t3chguy, line 63: This does not exist in the spec so equally should not be present here https://spec.matrix.org/v1.9/client-server-api/#mroompower_levels 

`invite` is though and shocked to see this missing for so long.

galash13: you're right, and also the spec has `notifications` which is missing here, i'll add it in place of `historical`

src/models/room-state.ts
galash13: nice idea! but as they are we always get `never` as `Action` due to the interface having optional keys, i played around a bit and we'd have to adjust it in one of these ways (that i could think of):

1)  `Required`
```
type KeysMatching<T, V> = {[K in keyof T]-?: T[K] extends V ? K : never}[keyof T];
type Action = KeysMatching<Required<IPowerLevelsContent>, number>;
```

2)  `number | undefined`
```
type KeysMatching<T, V> = {[K in keyof T]-?: T[K] extends V ? K : never}[keyof T];
type Action = KeysMatching<IPowerLevelsContent, number | undefined>;
```

3) `V | undefined`
```
type KeysMatching<T, V> = {[K in keyof T]-?: T[K] extends V | undefined ? K : never}[keyof T];
type Action = KeysMatching<IPowerLevelsContent, number>;
```

i'd opt for the 1st one

src/models/room-state.ts
```diff
@@ -53,13 +53,15 @@ enum OobStatus {
 export interface IPowerLevelsContent {
     users?: Record<string, number>;
     events?: Record<string, number>;
+    notifications?: Record<string, number>;
```
t3chguy, line 56: This type would be better specifying the known fields IMO

galash13: looking at https://spec.matrix.org/v1.9/client-server-api/#mroompower_levels looks like `room` is the only field supported atm so i added that

src/models/room-state.ts
```diff
@@ -53,13 +53,15 @@ enum OobStatus {
 export interface IPowerLevelsContent {
     users?: Record<string, number>;
     events?: Record<string, number>;
+    notifications?: Record<'room', number>;
```
t3chguy, line 56: This type isn't right strictly speaking, as it would render a value of `{}` invalid.

```suggestion
    notifications?: {
        room?: number;
    };
```

t3chguy:
The only bit missing now is the sign off https://github.com/element-hq/element-web/blob/develop/CONTRIBUTING.md#sign-off

t3chguy (approved):
Thanks for your contribution

PR #4129, opened 2024-03-22
Add new enum for verification methods.

richvdh:


src/client.ts
```diff
@@ -378,8 +377,10 @@ export interface ICreateClientOpts {
      * Verification methods we should offer to the other side when performing an interactive verification.
      * If unset, we will offer all known methods. Currently these are: showing a QR code, scanning a QR code, and SAS
      * (aka "emojis").
+     *
+     * See {@link types.VerificationMethod} for a set of useful constants for this parameter.
      */
-    verificationMethods?: Array<VerificationMethod>;
+    verificationMethods?: Array<string>;
```
t3chguy, line 383: Why is the typing getting looser?

richvdh, line 383: The old `VerificationMethod` appears to amount to `"RECIPROCATE_QR_CODE" | "SAS" | string`. I think that was misleading , and anyone actually using `RECIPROCATE_QR_CODE` would discover it doesn't actually work (you have to use `m.reciprocate.v1`.

Anyway: the point is that it accepted arbitrary `string`s, so changing `VerificationMethod` -> `string` here and in `legacyDeviceVerification` is backwards-compatible.

richvdh: you raced with me: see my comment below https://github.com/matrix-org/matrix-js-sdk/pull/4129#discussion_r1535772278.

t3chguy (approved):

andybalaam (approved):

PR #4131, opened 2024-03-25
Fix progress reporting for key import

BillCarsonFr:
Fixes https://github.com/element-hq/element-web/issues/27235

## Checklist

### DIFF
--- a/spec/integ/crypto/megolm-backup.spec.ts
+++ b/spec/integ/crypto/megolm-backup.spec.ts
@@ -590,6 +590,46 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("megolm-keys backup (%s)", (backe
             expect(result.imported).toStrictEqual(20);
         });
 
+        it("Should report failures when decryption works but import fails - per room variant", async function () {
+            // @ts-ignore - mock a private method for testing purpose
+            aliceCrypto.importBackedUpRoomKeys = jest
+                .fn()
+                .mockImplementationOnce((keys: IMegolmSessionData[], version: String, opts?: ImportRoomKeysOpts) => {
+                    // report 10 failures to import
+                    opts!.progressCallback!({
+                        stage: "load_keys",
+                        successes: 20,
+                        failures: 10,
+                        total: 30,
+                    });
+                    return Promise.resolve();
+                })
+                // Ok for other chunks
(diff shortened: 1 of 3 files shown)


spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -322,6 +324,22 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("megolm-keys backup (%s)", (backe
...
+                        opts.progressCallback({
+                            stage: "load_keys",
+                            successes: keys.length,
+                            failures: 0,
+                            total: keys.length,
+                        });
+                    }
+                });
+            // @ts-ignore - mock a private method for testing purpose
+            aliceCrypto.importBackedUpRoomKeys = importMockImpl;
```
richvdh, line 342: we should definitely not do this. Doing so turns all of these tests from integration tests (where we test the whole stack, including the rust bits) into unit tests (where we are just testing the behaviour of `MatrixClient`).

To be honest, it is a problem that we have existing tests that mock out `CryptoApi.importBackedUpRoomKeys` but which claim to be integration tests, but that's a problem for another day.

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -540,6 +551,85 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("megolm-keys backup (%s)", (backe
             expect(result.imported).toStrictEqual(expectedTotal - decryptionFailureCount);
         });
 
+        it("Should report failures when decryption works but import fails", async function () {
+            // @ts-ignore - mock a private method for testing purpose
+            aliceCrypto.importBackedUpRoomKeys = jest
```
richvdh, line 556: Again: by mocking out `importBackedUpRoomKeys`, these are unit tests rather than integration tests. We have no real evidence that the actual `RustCrypto` implementation behaves the same way that these mocks do.

Could we not make up some data which causes `RustCrypto.importBackedUpRoomKeys` to report a failure, without having to mock this out?

richvdh (changes requested):
A few requests for cleanups... and hopefully a better test?

richvdh:
@BillCarsonFr I'm going to close this for now as it seems to be bitrotting.

richvdh:
ok, really closing it this time!

PR #4134, opened 2024-03-25
MSC4108 support OIDC QR code login

t3chguy:


src/rendezvous/channels/MSC4108SecureChannel.ts
```diff
@@ -0,0 +1,264 @@
...
+     * Sends a payload securely to the other device.
+     * @param payload the payload to encrypt and send
+     */
+    public async secureSend<T extends MSC4108Payload>(payload: T): Promise<void> {
+        if (!this.connected) {
+            throw new Error("Channel closed");
+        }
+
+        const stringifiedPayload = JSON.stringify(payload);
+        logger.info(`=> ${stringifiedPayload}`);
```
hughns, line 214: We should either remove this logging entirely, or only log the `type` field of the payload. Otherwise things like the secrets bundle get leaked.

```suggestion
        logger.debug(`=> {"type": ${JSON.stringify(payload.type)}, ...}`);
```

src/rendezvous/channels/MSC4108SecureChannel.ts
```diff
@@ -0,0 +1,264 @@
...
+        }
+
+        const ciphertext = await this.rendezvousSession.receive();
+        if (!ciphertext) {
+            return undefined;
+        }
+        const plaintext = await this.decrypt(ciphertext);
+        const json = JSON.parse(plaintext);
+
+        logger.info(`<= ${JSON.stringify(json)}`);
```
hughns, line 234: Same as before, we should either not log at all or only the `type`. Probably at debug log level too.

```suggestion
        logger.debug(`<= {"type": ${JSON.stringify(json.type)}, ...}`);
```

hughns (changes requested):
Only minor changes suggested to prevent secrets leakage via logs.

hughns (approved):
The latest looks good to me (ignoring the build failure).

richvdh (commented):
Sorry, I'm not going to be able to review this as it stands. Please break it up.

src/rendezvous/MSC4108SignInWithQR.ts
```diff
@@ -0,0 +1,423 @@
...
+
+interface AcceptedPayload extends MSC4108Payload {
+    type: PayloadType.ProtocolAccepted;
+}
+
+interface SecretsPayload extends MSC4108Payload, QRSecretsBundle {
+    type: PayloadType.Secrets;
+}
+
+export class MSC4108SignInWithQR {
```
dbkr, line 90: This could do with some docs as a public export, as could the others above and in other files (the reason code exports are fairly self-evident).

src/rendezvous/transports/MSC4108RendezvousSession.ts
```diff
@@ -0,0 +1,270 @@
...
+
+    /**
+     * Receives data from the rendezvous channel.
+     * @return the returned promise won't resolve until new data is acquired or the channel is closed either by the server or the other party.
+     */
+    public async receive(): Promise<string | undefined> {
+        if (!this.url) {
+            throw new Error("Rendezvous not set up");
+        }
+        // eslint-disable-next-line no-constant-condition
```
dbkr, line 184: We should probably just disable this lint rule if we're going to ignore it anyway sometimes?

t3chguy: We should probably specify `checkLoops: allExceptWhileTrue`

image

I will do that separately as it is set in https://github.com/matrix-org/eslint-plugin-matrix-org

t3chguy: Looks like we can't use `allExceptWhileTrue` until we upgrade to eslint v9 which is blocked on some of the plugins we use becoming compatible

t3chguy: https://github.com/matrix-org/eslint-plugin-matrix-org/issues/40

dbkr: ah yep, I think florian was looking at upgrading this.

dbkr (approved):

t3chguy:
> Sorry, I'm not going to be able to review this as it stands. Please break it up.

@richvdh you are only asked to review the crypto related bits to fulfil the CODEOWNERS requirement.

I have provided the links to the corresponding files for your convenience:

https://github.com/matrix-org/matrix-js-sdk/pull/4134/files#diff-1cdf58640fc7d7c465fab943126da6a0d5c9ab99301676278fdc362beea10c11
https://github.com/matrix-org/matrix-js-sdk/pull/4134/files#diff-55b0e7275b7510d86f5e201e7b377396032f2fe5f040c9681fb75d40213d2c01
https://github.com/matrix-org/matrix-js-sdk/pull/4134/files#diff-6eb6c51bc52963382d87f0e413ae91df69b9c8426285c367e27f0328db33360b
https://github.com/matrix-org/matrix-js-sdk/pull/4134/files#diff-ca8741e6064cda5d26243ec210815e74962e5617951c9ec99ef4877dad964b38
https://github.com/matrix-org/matrix-js-sdk/pull/4134/files#diff-890ab04f8e5e8b612d0069bf0c681ef35d1ef3810abba98205ee464f3c366577

PS this is already a lean version of the original PR, just with a plethora of tests and comments added which bloats it.

richvdh:
Let's land #4227 and then see where we are

PR #4135, opened 2024-03-28
Fix merging of default push rules

t3chguy:
Fixes https://github.com/element-hq/element-web/issues/27173
Regressed by https://github.com/matrix-org/matrix-js-sdk/pull/4100

### DIFF
--- a/spec/unit/pushprocessor.spec.ts
+++ b/spec/unit/pushprocessor.spec.ts
@@ -849,4 +849,158 @@ describe("rewriteDefaultRules", () => {
             ".m.rule.fallback",
         ]);
     });
+
+    it("should add missing msc3914 rule in correct place", () => {
+        const pushRules = PushProcessor.rewriteDefaultRules({
+            device: {},
+            global: {
+                // Sample push rules from a Synapse user.
+                // Note that rules 2 and 3 are backwards, this will trigger a warning in the console.
+                underride: [
+                    {
+                        conditions: [
+                            {
+                                kind: "event_match",
+                                key: "type",
+                                pattern: "m.call.invite",
+                            },
+                        ],
+                        actions: [
+                            "notify",
+                            {
+                                set_tweak: "sound",
+                                value: "ring",
+                            },
+                            {
+                                set_tweak: "highlight",
+                                value: false,
+                            },
+                        ],
+                        rule_id: ".m.rule.call",
+                        default: true,
+                        enabled: true,
+                    },
+                    {
+                        conditions: [
+                            {
+                                kind: "event_match",
+                                key: "type",
+                                pattern: "m.room.message",
+                            },
+                            {
+                                kind: "room_member_count",
+                                is: "2",
+                            },
+                        ],
+                        actions: [
+                            "notify",
+                            {
+                                set_tweak: "sound",
+                                value: "TEST1",
+                            },
+                            {
+                                set_tweak: "highlight",
+                                value: false,
+                            },
+                        ],
+                        rule_id: ".m.rule.room_one_to_one",
+                        default: true,
+                        enabled: true,
+                    },
+                    {
+                        conditions: [
+                            {
+                                kind: "event_match",
+                                key: "type",
+                                pattern: "m.room.encrypted",
+                            },
+                            {
+                                kind: "room_member_count",
+                                is: "2",
+                            },
+                        ],
+                        actions: [
+                            "notify",
+                            {
+                                set_tweak: "sound",
+                                value: "TEST2",
+                            },
+                            {
+                                set_tweak: "highlight",
+                                value: false,
+                            },
+                        ],
+                        rule_id: ".m.rule.encrypted_room_one_to_one",
+                        default: true,
+                        enabled: true,
+                    },
+                    {
+                        conditions: [
+                            {
+                                kind: "event_match",
+                                key: "type",
+                                pattern: "m.room.message",
+                            },
+                        ],
+                        actions: ["dont_notify"],
+                        rule_id: ".m.rule.message",
+                        default: true,
+                        enabled: true,
+                    },
+                    {
+                        conditions: [
+                            {
+                                kind: "event_match",
+                                key: "type",
+                                pattern: "m.room.encrypted",
+                            },
+                        ],
+                        actions: ["dont_notify"],
+                        rule_id: ".m.rule.encrypted",
+                        default: true,
+                        enabled: true,
+                    },
+                    {
+                        conditions: [
+                            {
+                                kind: "event_match",
+                                key: "type",
+                                pattern: "im.vector.modular.widgets",
+                            },
+                            {
+                                kind: "event_match",
+                                key: "content.type",
+                                pattern: "jitsi",
+                            },
+                            {
+                                kind: "event_match",
+                                key: "state_key",
+                                pattern: "*",
+                            },
+                        ],
+                        actions: [
+                            "notify",
+                            {
+                                set_tweak: "highlight",
+                                value: false,
+                            },
+                        ],
+                        rule_id: ".im.vector.jitsi",
+                        default: true,
+                        enabled: true,
+                    },
+                ] as IPushRule[],
+            },
+        });
+
+        expect(pushRules.global.underride?.map((r) => r.rule_id)).toEqual([
+            ".m.rule.call",
+            ".org.matrix.msc3914.rule.room.call",
+            ".m.rule.room_one_to_one",
+            ".m.rule.encrypted_room_one_to_one",
+            ".m.rule.message",
+            ".m.rule.encrypted",
+            ".im.vector.jitsi",
+        ]);
+    });
 });
--- a/spec/unit/pushprocessor.spec.ts
+++ b/spec/unit/pushprocessor.spec.ts
@@ -198,11 +198,12 @@ describe("NotificationService", function () {
     });
 
     it("should add default rules in the correct order", () => {
-        matrixClient.pushRules = PushProcessor.rewriteDefaultRules({
+        const pushRules = PushProcessor.rewriteDefaultRules({
             device: {},
             global: {
                 content: [],
                 override: [
+                    // Include user-defined push rules inbetween .m.rule.master and other default rules to assert they are maintained in-order.
                     {
                         rule_id: ".m.rule.master",
                         default: true,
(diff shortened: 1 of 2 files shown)


MidhunSureshR (approved):

src/pushprocessor.ts
```diff
@@ -49,6 +49,10 @@ const RULEKINDS_IN_ORDER = [
     PushRuleKind.Underride,
 ];
 
+const UserDefinedRules = Symbol("UserDefinedRules");
+
+type OrderedRules = Array<string | typeof UserDefinedRules>;
```
richvdh, line 54: suggest putting these closer to `EXPECTED_DEFAULT_OVERRIDE_RULE_IDS`

src/pushprocessor.ts
```diff
@@ -49,6 +49,10 @@ const RULEKINDS_IN_ORDER = [
     PushRuleKind.Underride,
 ];
 
+const UserDefinedRules = Symbol("UserDefinedRules");
```
richvdh, line 52:
```suggestion
// A special rule id for `EXPECTED_DEFAULT_OVERRIDE_RULE_IDS` and friends.
const UserDefinedRules = Symbol("UserDefinedRules");
```

src/pushprocessor.ts
```diff
@@ -162,24 +168,31 @@ const EXPECTED_DEFAULT_UNDERRIDE_RULE_IDS = [
...
  *
+ * @param kind - the kind of push rule set being merged.
  * @param incomingRules - the existing set of known push rules for the user.
  * @param defaultRules - a lookup table for the default definitions of push rules.
  * @param defaultRuleIds - the IDs of the expected default push rules, in order.
  *
  * @returns A copy of `incomingRules`, with any missing default rules inserted in the right place.
  */
 function mergeRulesWithDefaults(
+    kind: PushRuleKind,
```
richvdh, line 179: this is unused?

src/pushprocessor.ts
```diff
@@ -162,24 +168,31 @@ const EXPECTED_DEFAULT_UNDERRIDE_RULE_IDS = [
...
 ): IPushRule[] {
-    // Calculate the index after the last default rule in `incomingRules`
-    // to allow us to split the incomingRules into defaults and custom
-    let firstCustomRuleIndex = incomingRules.findIndex((r) => !r.default);
-    if (firstCustomRuleIndex < 0) firstCustomRuleIndex = incomingRules.length;
-
-    function insertDefaultPushRule(ruleId: string): void {
-        if (ruleId in defaultRules) {
+    // Find the indices of the edges of the user-defined rules in the incoming rules
+    const incomingRulesEnabled = incomingRules.map((rule) => rule.enabled);
```
richvdh, line 185: why are we looking at `enabled` to find user-defined rules, rather than `default`?

It's possible for default rules to be disabled, so this seems unreliable

src/pushprocessor.ts
```diff
@@ -189,7 +202,11 @@ function mergeRulesWithDefaults(
 
     let nextExpectedRuleIdIndex = 0;
     const newRules: IPushRule[] = [];
-    for (const rule of incomingRules.slice(0, firstCustomRuleIndex)) {
+    // Process the default rules by merging them with defaults
```
richvdh, line 205: merge the defaults with the defaults?

```suggestion
    // Merge our expected rules (including the incoming custom rules) into the
    // incoming default rules.

File: src/pushprocessor.ts [hunk 6]
```diff
@@ -162,24 +168,31 @@ const EXPECTED_DEFAULT_UNDERRIDE_RULE_IDS = [
...
  * @param defaultRuleIds - the IDs of the expected default push rules, in order.
  *
  * @returns A copy of `incomingRules`, with any missing default rules inserted in the right place.
  */
 function mergeRulesWithDefaults(
+    kind: PushRuleKind,
     incomingRules: IPushRule[],
     defaultRules: Record<string, IPushRule>,
-    defaultRuleIds: string[],
+    defaultRuleIds: OrderedRules,
```
richvdh, line 182: now that this is no longer just the default rules, can we rename it to `expectedRuleIds` or something?

src/pushprocessor.ts
```diff
@@ -162,24 +168,31 @@ const EXPECTED_DEFAULT_UNDERRIDE_RULE_IDS = [
 /**
  * Make sure that each of the rules listed in `defaultRuleIds` is listed in the given set of push rules.
  *
+ * @param kind - the kind of push rule set being merged.
  * @param incomingRules - the existing set of known push rules for the user.
  * @param defaultRules - a lookup table for the default definitions of push rules.
  * @param defaultRuleIds - the IDs of the expected default push rules, in order.
```
richvdh, line 174:
```suggestion
 * @param defaultRuleIds - the IDs of the expected push rules, in order.
```

src/pushprocessor.ts
```diff
@@ -162,24 +168,31 @@ const EXPECTED_DEFAULT_UNDERRIDE_RULE_IDS = [
...
-    if (firstCustomRuleIndex < 0) firstCustomRuleIndex = incomingRules.length;
-
-    function insertDefaultPushRule(ruleId: string): void {
-        if (ruleId in defaultRules) {
+    // Find the indices of the edges of the user-defined rules in the incoming rules
+    const incomingRulesEnabled = incomingRules.map((rule) => rule.enabled);
+    const userDefinedRulesRange: [number, number] = [
+        incomingRulesEnabled.indexOf(false),
+        incomingRulesEnabled.lastIndexOf(false),
+    ];
```
richvdh, line 189: we're actually doing quite a lot of iterating and copying of incomingRules. I'd instead start just by partitioning them:

```
// Split the incomingRules into defaults and custom
const incomingDefaultRules = incomingRules.filter((rule) => rule.default);
const incomingCustomRules = incomingRules.filter((rule) => !rule.default);
```

... and I think everything else can then be more easily written in terms of those, instead of `incomingRules.slice`.

Not a blocker, just a thought.

spec/unit/pushprocessor.spec.ts
```diff
@@ -12,21 +22,21 @@ describe("NotificationService", function () {
...
 
-    const msc3914RoomCallRule = {
+    const msc3914RoomCallRule: IPushRule = {
         rule_id: ".org.matrix.msc3914.rule.room.call",
         default: true,
         enabled: true,
         conditions: [
             {
-                kind: "event_match",
+                kind: ConditionKind.EventMatch,
```
richvdh, line 31: ftr I am generally deeply ambivalent about using symbolic constants in test code: using the literal gives us a chance to check that the constant is doing what we expect.

(This brought to you on the back of a Synapse bug where we had a test which used what we *thought* was a constant but turned out not to be.)

YMMV though, not a particular request to change this.

spec/unit/pushprocessor.spec.ts
```diff
@@ -188,6 +198,108 @@ describe("NotificationService", function () {
     });
 
     it("should add default rules in the correct order", () => {
+        matrixClient.pushRules = PushProcessor.rewriteDefaultRules({
```
richvdh, line 201: I'm really confused about what's going on here. Why are we overriding the push rules that we've just set up in the `beforeEach`? Doesn't calling `rewriteDefaultRules` twice invalidate the rest of the test? What is special about this set of push rules compared to what we'd have normally (write a comment)?

richvdh (changes requested):
A few thoughts

src/pushprocessor.ts
t3chguy: If it was unused the linter would throw a tantrum, it is used for the logging

src/pushprocessor.ts
t3chguy: Splitting them as per your example would mean you lose track of the location of the user-defined rules. Right now only `.m.rule.master` precedes them but the server could send other default rules as they are added in that position. This codepath is only hit once for each incoming `m.push_rules` event down /sync, those should be quite rare

spec/unit/pushprocessor.spec.ts
t3chguy: > ftr I am generally deeply ambivalent about using symbolic constants in test code: using the literal gives us a chance to check that the constant is doing what we expect.

Its preferable to `@ts-ignore`.

image

src/pushprocessor.ts
richvdh: I don't really follow. We insert the custom rules into `newRules` just before copying `RuleId.SuppressNotices` either way?

> This codepath is only hit once for each incoming m.push_rules event down /sync, those should be quite rare

Yeah, I'm arguing this rather more from the PoV of clarity than performance. Just saying: it's not like doing the `indexOf`s here rather than copying them is saving you much.

spec/unit/pushprocessor.spec.ts
richvdh: I mean I'd argue your problem here is forcing msc3914RoomCallRule to be an IPushRule, but whatever

spec/unit/pushprocessor.spec.ts
```diff
@@ -198,11 +198,12 @@ describe("NotificationService", function () {
     });
 
     it("should add default rules in the correct order", () => {
-        matrixClient.pushRules = PushProcessor.rewriteDefaultRules({
+        const pushRules = PushProcessor.rewriteDefaultRules({
```
richvdh, line 201: ok this makes things clearer, but still, if we're not going to use any of the stuff in the `beforeEach`, let's move it to a different `describe` block. It's really confusing at the moment.

src/pushprocessor.ts
richvdh: I mean, fair, but also: what logging? It's entirely possible I'm blind, but I really can't see it?

src/pushprocessor.ts
t3chguy: > I don't really follow. We insert the custom rules into newRules just before copying RuleId.SuppressNotices either way?

incomingRules is in the order of:

```
some default rules (currently just master)
user defined rules
other default rules
```

Your proposed code would merge the default rules into one list and the default orders wouldn't account for any rules in the first bucket other than `master` so the order of any additional rules specified by the server there in later spec versions go into the wrong spot


```
// Split the incomingRules into defaults and custom
const incomingDefaultRules = incomingRules.filter((rule) => rule.default);
const incomingCustomRules = incomingRules.filter((rule) => !rule.default);
```

spec/unit/pushprocessor.spec.ts
t3chguy: Even if I remove that type cast there, then I'd get the failure in the call to `rewritePushRules`

The only reason it was fine before this PR is because we passed `matrixClient.pushRules!` to `rewritePushRules` and had
image to make TS happy with the types of `pushRules` within the mock `matrixClient`

t3chguy: image

src/pushprocessor.ts
t3chguy: image

t3chguy: Huh, somehow the `${kind}` went missing in my iterations, good spot

src/pushprocessor.ts
richvdh: > Your proposed code would merge the default rules into one list and the default orders wouldn't account for any rules in the first bucket other than master so the order of any additional rules specified by the server there in later spec versions go into the wrong spot

I don't think so. In such a scenario, the new rule would end up as the second entry in `incomingDefaultRules` (after `master`, and before `suppress_notices`). Assuming we change the loop below to iterate over `incomingDefaultRules` instead of the current pair of slices of `incomingRules`:
 * We'll see `master`; it will match the expectation, and we will copy it over.
 * We'll see the hypothesized new rule; it is unrecognized, and so we will copy it over.
 * We'll see `suppress_notices`. Its index in orderedRuleIds is higher than the expectation, so first we copy over the custom rules. Then we carry on.

richvdh (approved):
I still think you should do https://github.com/matrix-org/matrix-js-sdk/pull/4135#discussion_r1542878936, but I'm not going to block this.

src/pushprocessor.ts
t3chguy: Making the proposed change the test fails as follows:

image

t3chguy: Turns out this was due to missing `".org.matrix.msc3914.rule.room.call"` in the expected defaults in the last PR

dbkr:
Force merging because of e2e flake

PR #4139, opened 2024-04-02
Crypto: use a new error code for UTDs from device-relative historical events

richvdh:
If an event was sent before our device was created, and backup is not configured, we have no hope of decrypting it. To reflect this, use a different error code for the decryption failure.

Part of https://github.com/element-hq/element-meta/issues/2313

Suggest review commit-by-commit


richvdh:
Needs some updates per https://github.com/element-hq/element-meta/issues/2313#issuecomment-2035028384

uhoreg (approved):

florianduros (approved):

PR #4140, opened 2024-04-02
Add some comments to the release drafter workflows

dbkr:
If anyone knows the answer to the question, this would be good to add too.

## Checklist


.github/workflows/release-drafter.yml
```diff
@@ -1,3 +1,5 @@
+# Generates the draft release for the js-sdk
+# XXX: When would we run this via a dispatch?
```
t3chguy, line 2: When element web builds, so element desktop has up to date release notes

dbkr: The js-sdk one? Do tghey propagate all the way through the chain again?

t3chguy: Oh sorry thought this was the ED repo. The workflow_dispatch is for manual triggering, in case of failure/workflow updates etc. The repository_dispatch on ED's is for EW triggering it

t3chguy (approved):

.github/workflows/release-drafter.yml
dbkr: ah, gotcha. Hopefully this is accurate then?

PR #4152, opened 2024-04-08
Add comment to make clear that RoomStateEvent.Events does not update related objects in the js-sdk

toger5:

## Checklist


toger5:
We use canInvite in a couple of places but changing the power levels did not trigger anything so the related buttons were still greyed out.
I was trying to use `useEventEmitterState` with `RoomStateEvent.Events` but noticed, that the objects requested by canInvite were not updated yet when the event got emitted. This comment should help mitigate this pitfall.

MidhunSureshR (approved):

PR #4153, opened 2024-04-09
Remove the currentState property from room and instead use a getter.

toger5:
## Checklist

### DIFF
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -383,20 +383,7 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
      * The room summary.
      */
     public summary: RoomSummary | null = null;
-    /**
-     * oldState The state of the room at the time of the oldest event in the live timeline.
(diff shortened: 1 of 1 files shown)


src/models/room.ts
```diff
@@ -1217,9 +1224,6 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
             // Set our new fresh timeline as the live timeline to continue syncing
             // forwards and back paginating from.
             timelineSet.setLiveTimeline(newTimeline!);
-            // Fixup `this.oldstate` so that `scrollback` has the pagination tokens
-            // available
-            this.fixUpLegacyTimelineFields();
```
t3chguy, line 1222: The comment is indeed stale but I think we still need to update the re-emitter whenever the live timeline changes

toger5: yes this is definitly required. (Tests without updating the reemitter breaks a bunch of things)

PR #4155, opened 2024-04-11
MSC 3266 compliance

AndrewFerr:
This fixes some non-compliance with [MSC 3266](https://github.com/matrix-org/matrix-spec-proposals/pull/3266), and exports the payload type.

## Checklist

### DIFF
--- a/src/matrix.ts
+++ b/src/matrix.ts
@@ -77,6 +77,7 @@ export * from "./@types/extensible_events";
 export * from "./@types/IIdentityServerProvider";
 export * from "./models/room-summary";
 export * from "./models/event-status";
+export { IRoomSummary } from "./client";
 export * as ContentHelpers from "./content-helpers";
 export * as SecretStorage from "./secret-storage";
 export type { ICryptoCallbacks } from "./crypto"; // used to be located here
--- a/src/client.ts
+++ b/src/client.ts
@@ -865,7 +865,7 @@ interface IThirdPartyUser {
     fields: object;
 }
 
-interface IRoomSummary extends Omit<IPublicRoomsChunkRoom, "canonical_alias" | "aliases"> {
+export interface IRoomSummary extends Omit<IPublicRoomsChunkRoom, "canonical_alias" | "aliases"> {
     membership?: Membership;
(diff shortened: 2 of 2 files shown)


src/matrix.ts
```diff
@@ -77,6 +77,7 @@ export * from "./@types/extensible_events";
 export * from "./@types/IIdentityServerProvider";
 export * from "./models/room-summary";
 export * from "./models/event-status";
+export { IRoomSummary } from "./client";
```
t3chguy, line 80: New exported symbols should follow the code style of not being prefixed with I

t3chguy:
Please improve the PR title given it will be used in the changelog verbatim

t3chguy (commented):
Looks sane otherwise

AndrewFerr:
> Please improve the PR title given it will be used in the changelog verbatim

Done -- let me know if the title still needs changing.

I am also willing to add an integration test for this.

t3chguy:
Yeah looks like CI won't let it in without testing

AndrewFerr:
Good thing CI is strict, because this PR requires a Synapse change too: https://github.com/element-hq/synapse/pull/17078

t3chguy:
So we'll likely need fallback code then
for older Synapse support

AndrewFerr:
> So we'll likely need fallback code then for older Synapse support

The Synapse PR has a fallback: it responds to both the old & new endpoint used by `getRoomSummary`.

Then again, `getRoomSummary` could also have a fallback: if it gets a 404 response for the new endpoint, it can retry at the old endpoint.  I'll add that now.

AndrewFerr:
Hm, even with the integration test, the code analysis check is failing.  What needs to be added to fix that?

Other than that, bd09442 adds the fallback, so once the check is fixed, this should PR be ready.

t3chguy:
> Hm, even with the integration test, the code analysis check is failing

It says you reached 33.3% coverage, out of the required 80%.

image

Things marked red are untested.

AndrewFerr:
Do I need to rebase this PR branch before it can be merged?

PR #4156, opened 2024-04-11
Element R: fix `isCrossSigningReady` not checking identity trust

BillCarsonFr:
Fix an inconsistency between legacy crypto and rust crypto when calling `isCrossSigningReady`.
Legacy is checking that cross-signing is set up and that the keys are trusted.
Rust was just checking that cross-signing was set up.

Related issue  https://github.com/element-hq/customer-retainer/issues/224

## Checklist


src/rust-crypto/rust-crypto.ts
```diff
@@ -610,15 +610,17 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
             Boolean(privateKeysCachedLocally.userSigningKey);
 
-        // The cross signing is ready if the public and private keys are available
-        return publicKeysOnDevice && (hasKeysInCache || privateKeysInSecretStorage);
+        const identity = await this.getOwnIdentity();
+
+        // The cross signing is ready if cross-signing is set up, the identity is trusted,
+        // and private keys are available somewhere.
+        return !!identity?.isVerified() && (hasKeysInCache || privateKeysInSecretStorage);
     }
```
richvdh, line 624: the behaviour of `isCrossSigningReady` is defined in `CryptoApi`.

It says:

> Checks whether cross signing:
> - is enabled on this account and **trusted by this device**
> - has private keys either cached locally or stored in secret storage

... so this sounds like fixing a *bug* in rust crypto, rather than just "fix inconsistency" ?

BillCarsonFr: Right, updating the title

src/rust-crypto/rust-crypto.ts
```diff
@@ -610,15 +610,17 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
             Boolean(privateKeysCachedLocally.masterKey) &&
             Boolean(privateKeysCachedLocally.selfSigningKey) &&
             Boolean(privateKeysCachedLocally.userSigningKey);
 
-        // The cross signing is ready if the public and private keys are available
-        return publicKeysOnDevice && (hasKeysInCache || privateKeysInSecretStorage);
+        const identity = await this.getOwnIdentity();
+
+        // The cross signing is ready if cross-signing is set up, the identity is trusted,
+        // and private keys are available somewhere.
```
richvdh, line 622: > The cross signing is ready if cross-signing is set up

This seems a bit like a tautology? What exactly does "cross-signing is set up" mean? I'm not sure it's adding any value here.

```suggestion
        // Cross signing is ready if the public identity is trusted, and the private keys 
        // are either cached, or accessible via secret-storage.
```

spec/integ/crypto/cross-signing.spec.ts
```diff
@@ -347,6 +347,81 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("cross-signing (%s)", (backend: s
 
             expect(isCrossSigningReady).toBeTruthy();
         });
+
+        it("should return false if identity is not trusted and secrets in 4S", async () => {
```
richvdh, line 351:
```suggestion
        it("should return false if identity is not trusted, even if the secrets are in 4S", async () => {
```

spec/integ/crypto/cross-signing.spec.ts
```diff
@@ -347,6 +347,81 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("cross-signing (%s)", (backend: s
...
+                                encrypted: {
+                                    lGZnUDzNKXGXR083k8JVKAtGSiBj5Dv1: {
+                                        iv: "6YJA6EYa72xbttCd981DZw==",
+                                        ciphertext: "1ZVy5fWqLPCVsJPvi2L/Pw9uS1D02MkeZ8zNwh7knA8hveOMO2PBslf8EQ==",
+                                        mac: "2EXorwWE5unhdsLx+IO7hxhZwolHyKoLV2ZKaRJoBlQ=",
+                                    },
+                                },
+                            },
+                        },
+                    ],
```
richvdh, line 415: could you write some comments about what exactly these events are and where they came from? Are they supposed to match the keys used in `test-data`, or are they deliberately different? In other words, do we expect the client to be able to decrypt them or not?

richvdh (changes requested):

src/rust-crypto/rust-crypto.ts
BillCarsonFr: Just that the public identity could be null, if you never set it up. But yes maybe not worth it

spec/integ/crypto/cross-signing.spec.ts
BillCarsonFr: Udpated

richvdh: This is much clearer now, thanks!

richvdh (approved):
LGTM!

richvdh:
@BillCarsonFr It would be good to summarise what effects this bug might have on the user experience. IIRC we have at least one bug report which we believe was caused by this? But it doesn't seem to be linked.

BillCarsonFr:
> @BillCarsonFr It would be good to summarise what effects this bug might have on the user experience. IIRC we have at least one bug report which we believe was caused by this? But it doesn't seem to be linked.

added in description

PR #4168, opened 2024-04-17
Element-R: Fix rust migration when ssss secret are stored not encryted in cache (old legacy behavior)

BillCarsonFr:
Fixes https://github.com/element-hq/element-desktop/issues/1603

Some old session might have stored the ssss cached secret in clear form (raw UInt8Array). The existing migration code will fail on such sessions with 
```
TypeError: The first argument must be one of type string, Buffer, ArrayBuffer, Array, or Array-like Object. Received type undefined
    ...
    at async getAndDecryptCachedSecretKey (vector://vector/webapp/bundles/1d2e934471ebcb41f3e7/865.js:4549:44)
```

Updated the  migration's `getAndDecryptCachedSecretKey` to support both cases similar to https://github.com/matrix-org/matrix-js-sdk/blob/4fc6ba884e166cccf7c0af18e01e5f8c7056fa6b/src/crypto/CrossSigning.ts#L659

## Checklist

### DIFF
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -304,6 +304,56 @@ describe("initRustCrypto", () => {
             }
         }, 10000);
 
+        it("migrates data from a legacy crypto store when secret are not encrypted", async () => {
+            const PICKLE_KEY = "pickle1234";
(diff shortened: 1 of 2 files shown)


andybalaam (approved):
Looks good, thanks!

PR #4169, opened 2024-04-18
Modernize window.crypto access constants

turt2live:
For https://github.com/element-hq/element-web/pull/27326

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,6 @@
         "@babel/preset-typescript": "^7.12.7",
         "@casualbot/jest-sonar-reporter": "2.2.7",
         "@matrix-org/olm": "3.2.15",
-        "@peculiar/webcrypto": "^1.4.5",
         "@types/bs58": "^4.0.1",
         "@types/content-type": "^1.1.5",
         "@types/debug": "^4.1.7",
@@ -91,7 +90,6 @@
         "@types/jest": "^29.0.0",
         "@types/node": "18",
         "@types/sdp-transform": "^2.4.5",
-        "@types/serviceworker": "^0.0.85",
         "@types/uuid": "9",
         "@typescript-eslint/eslint-plugin": "^7.0.0",
         "@typescript-eslint/parser": "^7.0.0",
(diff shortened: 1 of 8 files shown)


turt2live:
I believe the Element Web build error is failed (or lack of) branch matching to the react-sdk.

turt2live:
Sorry for the spam here. As is hopefully obvious from the commit history, I tried to use `@types/serviceworker` and that exploded in my face for the 8th time this week.

My vote is to give up for now and put a help wanted label on it.

richvdh (changes requested):
I don't really know why we're indirecting via `window` in the first place. [`crypto`](https://developer.mozilla.org/en-US/docs/Web/API/crypto_property) is defined as a global property - ie, it should be available via `globalThis` in all environments.

I suspect it's because this code predates the existence of `globalThis`, so used to say `crypto = window.crypto`, which then got (incorrectly, imho) turned into `crypto = globalThis.window.crypto`.

Likewise TextEncoder.

richvdh:
> My vote is to give up for now and put a help wanted label on it.

I don't really understand. If you're not going to keep working on the PR, could you close it please?

turt2live:
I'm not sure I understand the action requested here. Should the crypto indirection be removed?

As for closing the PR: I'm saying we give up on trying to get types to work, but the PR is required to work. I cannot commit to debugging the entire type system much further, so a help wanted issue feels like the correct approach.

richvdh:
> I'm not sure I understand the action requested here. Should the crypto indirection be removed?

I'm suggesting that, rather than doing anything special for service workers, you change lines 19-21 to be:

```
export let crypto = globalThis.crypto;
export let subtleCrypto = crypto?.subtle ?? crypto?.webkitSubtle;  // or just `crypto?.subtle`
export let TextEncoder = globalThis.TextEncoder;
```

... and be done.

> As for closing the PR: I'm saying we give up on trying to get types to work, but the PR is required to work. I cannot commit to debugging the entire type system much further, so a help wanted issue feels like the correct approach.

oh right. Well, that was far from obvious from your comment :)

richvdh (approved):
LGTM, though needs an update to the PR title.

PR #4171, opened 2024-04-19
add resignKeyBackup, updateBackupSignature, resignKeyBackup

MichaelErjemenko:
When using key backups without 4s (secure secret storage and sharing) and signing in with the web-client then no migration is performed and the session cannot backup keys without any further user interaction (resetting cross signing (and key backup)).
This seems to be related to: https://github.com/element-hq/element-web/issues/27100
ToDo: If the above is correct then a note for the Changelog will be added.

The changes here depend on the changes in [matrix-react-sdk PR 12441](https://github.com/matrix-org/matrix-react-sdk/pull/12441) .

## Checklist

Signed-off-by: Michael Schrader michael.schrader@connext.de

### DIFF
--- a/src/rust-crypto/backup.ts
+++ b/src/rust-crypto/backup.ts
@@ -545,6 +545,47 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
     }
 
     /**
+     * Use PUT /room_keys/version/{version} to change the e2e_room_keys_version database entry at the server.
+     * This version should be JUST used in rare specific cases and not in general.
+     * Currently it is used as workaround for compatibility problems between backups without 4s and required 4s at the web-client.
+     * 
+     * @param backupDecKey 
+     * @param version The version of the backup we want to update.
+     * @param signObject A method to sign the backup before it is uploaded.
+     * @returns 
+     */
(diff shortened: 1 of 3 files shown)


src/rust-crypto/backup.ts
```diff
@@ -544,6 +544,47 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+     * @param signObject A method to sign the backup before it is uploaded.
+     * @returns 
+     */
+    public async updateBackupSignature(backupDecKey: RustSdkCryptoJs.BackupDecryptionKey, version: string, signObject: (authData: AuthData) => Promise<void>): Promise<KeyBackupCreationInfo> {
+        const pubKey = backupDecKey.megolmV1PublicKey;
+        const authData = { public_key: pubKey.publicKeyBase64 };
+
+        await signObject(authData);
+
+        // An alternative implementation could be using src\crypto\EncryptionSetup.ts and EncryptionSetupOperation, similar to:
```
richvdh, line 563: that stuff is deprecated; let's not use it.

richvdh (commented):
See my comments on https://github.com/matrix-org/matrix-react-sdk/pull/12441, but also:

`resignKeyBackup` feels like a very low-level, and dangerous primitive to expose. I really don't want applications to have to decide to arbitrarily re-sign existing key backups.

If this is a thing that we need to support, maybe we can make it part of `bootstrapCrossSigning` ?

richvdh:
As with https://github.com/matrix-org/matrix-react-sdk/pull/12441: I don't really think this is the right thing to be doing.

Issue #4154, opened 2024-04-10
rust-crypto is included by default which leads to bundle size growth


ms-dosx86:
Reproduction: https://github.com/ms-dosx86/matrix-js-sdk-example

Steps to reproduce:

1. `npm i`
2. `ng build`

![image](https://github.com/matrix-org/matrix-js-sdk/assets/27197930/6f854202-cacb-4ec4-884f-116300b7f80e)

An empty project with just `createClient()` weighs 8.89mb.

You can also do `ng build --stats-json` and upload `dist/matrix-js-sdk-example/stats.json` to https://esbuild.github.io/analyze/ to see what causes such huge bundle size.

![image](https://github.com/matrix-org/matrix-js-sdk/assets/27197930/feb39bcb-a3eb-4adf-9bae-e7cf2639eb93)

This happens because of lazy rust-crypto import here

https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/client.ts#L2255

it looks fine, the source of problem is babel preset `module: 'commonjs'`

https://github.com/matrix-org/matrix-js-sdk/blob/develop/.babelrc#L10

which makes the lazy import not quite lazy.

Possible workaround for those who is facing the same problem:

1. Clone matrix-js-sdk
2. Set `module: false` and `node: 18` in `.babelrc`
3. Replace this line https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/models/MSC3089TreeSpace.ts#L38 with `import type { EncryptedFile, FileContent } from "../@types/media";`
4. `yarn build`
5. Copy `lib` and replace with it `node_modules/matrix-js-sdk/lib`.

richvdh:
I'm afraid I'm not following everything you're saying here; I'm not that familiar with all the finer details of babel and the javascript ecosystem, and I've never used `ng`.

It does seem a problem that the transpiled js-sdk forces the rust-crypto import to be resolved at load time. (We don't encounter this in element-web because we use the typescript source in `src`, rather than the transpiled code in `lib`.) And yes, it seems like we need to change the `module` setting.

A couple of questions about your solution:

> Set `module: false` and `node: 18` in `.babelrc`

I think updating `node` makes sense, but I don't entirely follow why it is *necessary* here. Please could you explain?


[...]

PR #4187, opened 2024-05-02
Preserve ESM for async imports to work correctly

ms-dosx86:
## Checklist

Fixes https://github.com/matrix-org/matrix-js-sdk/issues/4154

There were a couple of things I encountered during this PR:

1. I had to add `import type` in [`src/models/MSC3089TreeSpace.ts`](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/models/MSC3089TreeSpace.ts#L38) because `@babel/preset-typescript` refused to remove that unused import. It contains two interfaces that should be removed after `yarn build`. But I guess this https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/models/MSC3089TreeSpace.ts#L83

```ts
declare module "../@types/media" {
    interface FileContent {
        [UNSTABLE_MSC3089_LEAF.name]?: {};
    }
}
```

prevents it from being removed. `import type` fixes this with minimum amount of effort.

2. `jest` refused to run spec files with ESM enabled with `babel-jest`. It kept throwing this error
```
SyntaxError: Cannot use import statement outside a module
```

even tho I have `babel` and `babel-typescript` configured properly. I tried as described in the docs https://jestjs.io/docs/ecmascript-modules , but no success.

Then I tried `ts-jest` instead of `babel-jest`. It worked but 7 of 137 specs failed. This is one of them
```
at Object.<anonymous> (src/client.ts:235:57)
at Object.<anonymous> (src/models/thread.ts:19:1)
at Object.<anonymous> (src/models/event.ts:39:1)
at Object.<anonymous> (src/crypto/EncryptionSetup.ts:18:1)
at Object.<anonymous> (src/crypto/index.ts:36:1)
at Object.<anonymous> (spec/unit/crypto.spec.ts:7:1)
```

According to this https://github.com/kulshekhar/ts-jest/issues/1873 it might be related to circular dependencies problem since `src/crypto/index.ts` ends up importing `src/client.ts` which tries to import `./crypto` so the test fails. It works fine with `yarn build` tho. I'm not familiar with `jest` so any help here will be appreciated.

Anyway, I replaced `.babelrc` with `babel.config.js` to run tests with `modules: "commonjs"` to make `jest` work.

And finally I added an example to test async imports (`rust-crypto` import) and see that they were splitted from the main chunk. A command to build and serve the example
```
npx esbuild examples/lazy-crypto-init/index.js --bundle --splitting --outdir=examples/lazy-crypto-init --format=esm --servedir=examples/lazy-crypto-init --watch --define:global=globalThis
```

Signed-off-by: Bayyr Oorjak <the.bayyr.oorjak@gmail.com>


richvdh:
> 2\. `jest` refused to run spec files with ESM enabled with `babel-jest`. It kept throwing this error
> 
> 
> ```
> SyntaxError: Cannot use import statement outside a module
> ```

Yup. It looks like you're hitting https://github.com/jestjs/jest/issues/9860. As that issue says (eventually), you can solve the problem by adding `extensionsToTreatAsEsm: ['.ts']` to the jest config.

*However* then you discover the next problem, which is that our tests rely on [`jest.mock`](https://jestjs.io/docs/jest-object#jestmockmodulename-factory-options) to mock out entire modules, which doesn't seem to work very well either. (https://github.com/jestjs/jest/issues/10025 seems related, though it doesn't entirely match the behaviour I observe).

Anyway, I think your proposed approach of sticking to CJS for Jest is the right one, for now at least.

babel.config.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+    sourceMaps: true,
+    presets: [
+        [
+            "@babel/preset-env",
+            {
+                targets: {
+                    esmodules: true,
+                },
+                modules: process.env.NODE_ENV === "test" ? "commonjs" : false,
```
richvdh, line 10: please could you add a comment to the code to explain what this is doing, and why?

richvdh (changes requested):
Please could you also add a sign-off to your PR, as requested in https://github.com/element-hq/element-web/blob/develop/CONTRIBUTING.md#sign-off

ms-dosx86:
Hi @richvdh !

- [x] I agree that the example is not necessary and removed it.
- [x] Also added a comment as you asked. Hope you don't mind mentioning your comment there since you explained the situation very clearly.
- [x] And added a sign off in the last commit. Sorry, that was my bad. Didn't see it in the description checklist.

babel.config.js
```diff
@@ -7,6 +7,9 @@ module.exports = {
                 targets: {
                     esmodules: true,
                 },
+                // Used "commonjs" for tests due to https://github.com/matrix-org/matrix-js-sdk/pull/4187#issuecomment-2117342908
+                // and "false" to preserve ESM for the final build to make async
+                // imports work correctly.
```
richvdh, line 12:
```suggestion
                // We want to output ES modules for the final build (mostly to ensure that
                // async imports work correctly). However, junit doesn't support ES modules very
                // well yet (see https://github.com/jestjs/jest/issues/9430), so we use commonjs
                // when testing.
```

richvdh (commented):
minor wording suggestions

richvdh:
>  And added a sign off in the last commit. Sorry, that was my bad. Didn't see it in the description checklist.

Thanks. Unfortunately that doesn't help with the earlier commits. Could you edit the description of the pull request to add a `Signed-off-by` line there?

ms-dosx86:
> > And added a sign off in the last commit. Sorry, that was my bad. Didn't see it in the description checklist.
> 
> Thanks. Unfortunately that doesn't help with the earlier commits. Could you edit the description of the pull request to add a `Signed-off-by` line there?

Oh.. I see. I updated the description. Hope that it's correct this time.

babel.config.js
```diff
@@ -0,0 +1,26 @@
...
+    sourceMaps: true,
+    presets: [
+        [
+            "@babel/preset-env",
+            {
+                targets: {
+                    esmodules: true,
+                },
+                // We want to output ES modules for the final build (mostly to ensure that
+                // async imports work correctly). However, junit doesn't support ES modules very
```
richvdh, line 11: @richvdh == idiot

```suggestion
                // async imports work correctly). However, jest doesn't support ES modules very
```

richvdh (approved):
LGTM! Thanks!

PR #4195, opened 2024-05-08
Make the js-sdk compatible with MSC preferred foci and active focus.

toger5:
The call.member events are refactored in a backwards compatible way to not only have one array: `foci_active: Focus[]` but instead one array:
`foci_preferred: Focus[]`
and a single:
`focus_active: Focus`

Foci active should just be one field describing what you are currently using. In a fully cascading world this is your own foci. In the livekit scenario this describes how you connect to the correct livekit session. (using the oldest session member should be the default, but if someone is using sth else everyone should connect to the other livekit session as well. Not yet implemented)

`foci_preferred` is what `foci_active` has been. The list of possible foci this user knows about or can use.

fixes https://github.com/element-hq/element-call/issues/1737

## Checklist

### DIFF
--- a/src/@types/event.ts
+++ b/src/@types/event.ts
@@ -358,8 +358,8 @@ export interface StateEvents {
     // MSC3401
     [EventType.GroupCallPrefix]: IGroupCallRoomState;
     [EventType.GroupCallMemberPrefix]: XOR<
-        IGroupCallRoomMemberState,
-        ExperimentalGroupCallRoomMemberState | {} | SessionMembershipData
+        XOR<IGroupCallRoomMemberState, ExperimentalGroupCallRoomMemberState>,
+        XOR<SessionMembershipData, {}>
     >;
 
     // MSC3089
--- a/src/@types/event.ts
+++ b/src/@types/event.ts
@@ -57,6 +57,7 @@ import {
 } from "../webrtc/callEventTypes";
 import { EncryptionKeysEventContent, ICallNotifyContent } from "../matrixrtc/types";
 import { M_POLL_END, M_POLL_START, PollEndEventContent, PollStartEventContent } from "./polls";
+import { SessionMembershipData } from "../matrixrtc/CallMembership";
 
 export enum EventType {
     // Room state events
@@ -356,7 +357,10 @@ export interface StateEvents {
 
     // MSC3401
     [EventType.GroupCallPrefix]: IGroupCallRoomState;
-    [EventType.GroupCallMemberPrefix]: XOR<IGroupCallRoomMemberState, ExperimentalGroupCallRoomMemberState>;
+    [EventType.GroupCallMemberPrefix]: XOR<
+        IGroupCallRoomMemberState,
+        ExperimentalGroupCallRoomMemberState | {} | SessionMembershipData
+    >;
 
     // MSC3089
     [UNSTABLE_MSC3089_BRANCH.name]: MSC3089EventContent;
--- a/src/matrixrtc/CallMembership.ts
+++ b/src/matrixrtc/CallMembership.ts
@@ -29,7 +29,8 @@ export interface CallMembershipData {
     created_ts?: number;
     expires?: number;
     expires_ts?: number;
-    foci_active?: Focus[];
+    foci_active?: Focus;
+    foci_preferred?: Focus[];
     membershipID: string;
 }
 
@@ -122,7 +123,7 @@ export class CallMembership {
         return this.getMsUntilExpiry() <= 0;
     }
 
-    public getActiveFoci(): Focus[] {
-        return this.data.foci_active ?? [];
+    public getPreferredFoci(): Focus[] {
+        return this.data.foci_preferred ?? [];
     }
 }
--- a/src/matrixrtc/CallMembership.ts
+++ b/src/matrixrtc/CallMembership.ts
@@ -19,6 +19,7 @@ import { EitherAnd } from "matrix-events-sdk/lib/types";
 import { MatrixEvent } from "../matrix";
 import { deepCompare } from "../utils";
 import { Focus } from "./focus";
+import { isLivekitFocusActive } from "./LivekitFocus";
 
 type CallScope = "m.room" | "m.user";
 // Represents an entry in the memberships section of an m.call.member event as it is on the wire
@@ -32,7 +33,7 @@ export type SessionMembershipData = {
     call_id: string;
     device_id: string;
 
-    foci_active: Focus;
+    focus_active: Focus;
     foci_preferred: Focus[];
     created_ts?: number;
 
@@ -41,13 +42,14 @@ export type SessionMembershipData = {
 };
 
 export const isSessionMembershipData = (data: CallMembershipData): data is SessionMembershipData =>
-    "foci_active" in data;
+    "focus_active" in data;
+
 const checkSessionsMembershipData = (data: any, errors: string[]): data is SessionMembershipData => {
     const prefix = "Malformed session membership event: ";
     if (typeof data.device_id !== "string") errors.push(prefix + "device_id must be string");
     if (typeof data.call_id !== "string") errors.push(prefix + "call_id must be string");
     if (typeof data.application !== "string") errors.push(prefix + "application must be a string");
-    if (typeof data.foci_active?.type !== "string") errors.push(prefix + "foci_active.type must be a string");
+    if (typeof data.focus_active?.type !== "string") errors.push(prefix + "focus_active.type must be a string");
     if (!Array.isArray(data.foci_preferred)) errors.push(prefix + "foci_preferred must be an array");
     // optional parameters
     if (data.created_ts && typeof data.created_ts !== "number") errors.push(prefix + "created_ts must be number");
@@ -200,5 +202,16 @@ export class CallMembership {
 
         // MSC4143 style membership
         return this.membershipData.foci_preferred;
+    }
+
+    public getFocusSelection(): string | undefined {
+        if (isLegacyCallMembershipData(this.membershipData)) {
+            return "oldest_membership";
+        } else {
+            const focusActive = this.membershipData.focus_active;
+            if (isLivekitFocusActive(focusActive)) {
+                return focusActive.focus_selection;
+            }
+        }
     }
 }
--- a/src/matrixrtc/CallMembership.ts
+++ b/src/matrixrtc/CallMembership.ts
@@ -24,6 +24,7 @@ type CallScope = "m.room" | "m.user";
 // There are two different data interfaces. One for the Legacy types and one complient with MSC4143
 
 // MSC4143 (MatrixRTC) session membership data
+
 export interface SessionMembershipData {
     application: string;
     call_id: string;
@@ -36,6 +37,7 @@ export interface SessionMembershipData {
     // Application specific data
     scope?: CallScope;
 }
+
 export const isSessionMembershipData = (data: CallMembershipData): data is SessionMembershipData =>
     "foci_active" in data &&
     "foci_preferred" in data &&
@@ -58,6 +60,7 @@ const checkSessionsMembershipData = (data: SessionMembershipData): void => {
 };
 
 // Legacy session membership data
+
 export interface CallMembershipDataLegacy {
     application: string;
     call_id: string;
@@ -69,8 +72,10 @@ export interface CallMembershipDataLegacy {
     expires_ts?: number;
     foci_active?: Focus[];
 }
+
 export const isLegacyCallMembershipData = (data: CallMembershipData): data is CallMembershipDataLegacy =>
     "membershipID" in data && "foci_active" in data && Array.isArray(data.foci_active);
+
 const checkCallMembershipDataLegacy = (data: CallMembershipDataLegacy): void => {
     const prefix = "Malformed legacy rtc membership event: ";
     if (!(data.expires || data.expires_ts)) {
--- a/src/matrixrtc/CallMembership.ts
+++ b/src/matrixrtc/CallMembership.ts
@@ -41,24 +41,20 @@ export type SessionMembershipData = {
 };
 
 export const isSessionMembershipData = (data: CallMembershipData): data is SessionMembershipData =>
-    "foci_active" in data &&
-    "foci_preferred" in data &&
-    !Array.isArray(data.foci_active) &&
-    Array.isArray(data.foci_preferred);
-
-const checkSessionsMembershipData = (data: any): data is SessionMembershipData => {
+    "foci_active" in data;
+const checkSessionsMembershipData = (data: any, errors: string[]): data is SessionMembershipData => {
     const prefix = "Malformed session membership event: ";
-    if (typeof data.device_id !== "string") throw new Error(prefix + "device_id must be string");
-    if (typeof data.call_id !== "string") throw new Error(prefix + "call_id must be string");
-    if (typeof data.application !== "string") throw new Error(prefix + "application must be a string");
-    if (typeof data.foci_active?.type !== "string") throw new Error(prefix + "foci_active.type must be a string");
-    if (!Array.isArray(data.foci_preferred)) throw new Error(prefix + "foci_preferred must be an array");
+    if (typeof data.device_id !== "string") errors.push(prefix + "device_id must be string");
+    if (typeof data.call_id !== "string") errors.push(prefix + "call_id must be string");
+    if (typeof data.application !== "string") errors.push(prefix + "application must be a string");
+    if (typeof data.foci_active?.type !== "string") errors.push(prefix + "foci_active.type must be a string");
+    if (!Array.isArray(data.foci_preferred)) errors.push(prefix + "foci_preferred must be an array");
     // optional parameters
-    if (data.created_ts && typeof data.created_ts !== "number") throw new Error(prefix + "created_ts must be number");
+    if (data.created_ts && typeof data.created_ts !== "number") errors.push(prefix + "created_ts must be number");
 
     // application specific data (we first need to check if they exist)
-    if (data.scope && typeof data.scope !== "string") throw new Error(prefix + "scope must be string");
-    return true;
+    if (data.scope && typeof data.scope !== "string") errors.push(prefix + "scope must be string");
+    return errors.length === 0;
 };
 
 // Legacy session membership data
@@ -76,31 +72,31 @@ export type CallMembershipDataLegacy = {
 export const isLegacyCallMembershipData = (data: CallMembershipData): data is CallMembershipDataLegacy =>
     "membershipID" in data;
 
-const checkCallMembershipDataLegacy = (data: any): data is CallMembershipDataLegacy => {
+const checkCallMembershipDataLegacy = (data: any, errors: string[]): data is CallMembershipDataLegacy => {
     const prefix = "Malformed legacy rtc membership event: ";
     if (!("expires" in data || "expires_ts" in data)) {
-        throw new Error(prefix + "expires_ts or expires must be present");
+        errors.push(prefix + "expires_ts or expires must be present");
     }
     if ("expires" in data) {
         if (typeof data.expires !== "number") {
-            throw new Error(prefix + "expires must be numeric");
+            errors.push(prefix + "expires must be numeric");
         }
     }
     if ("expires_ts" in data) {
         if (typeof data.expires_ts !== "number") {
-            throw new Error(prefix + "expires_ts must be numeric");
+            errors.push(prefix + "expires_ts must be numeric");
         }
     }
 
-    if (typeof data.device_id !== "string") throw new Error(prefix + "device_id must be string");
-    if (typeof data.call_id !== "string") throw new Error(prefix + "call_id must be string");
-    if (typeof data.application !== "string") throw new Error(prefix + "application must be a string");
-    if (typeof data.membershipID !== "string") throw new Error(prefix + "membershipID must be a string");
+    if (typeof data.device_id !== "string") errors.push(prefix + "device_id must be string");
+    if (typeof data.call_id !== "string") errors.push(prefix + "call_id must be string");
+    if (typeof data.application !== "string") errors.push(prefix + "application must be a string");
+    if (typeof data.membershipID !== "string") errors.push(prefix + "membershipID must be a string");
     // optional elements
-    if (data.created_ts && typeof data.created_ts !== "number") throw new Error(prefix + "created_ts must be number");
+    if (data.created_ts && typeof data.created_ts !== "number") errors.push(prefix + "created_ts must be number");
     // application specific data (we first need to check if they exist)
-    if (data.scope && typeof data.scope !== "string") throw new Error(prefix + "scope must be string");
-    return true;
+    if (data.scope && typeof data.scope !== "string") errors.push(prefix + "scope must be string");
+    return errors.length === 0;
 };
 
 export type CallMembershipData = CallMembershipDataLegacy | SessionMembershipData;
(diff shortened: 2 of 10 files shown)


src/matrixrtc/CallMembership.ts
```diff
@@ -19,19 +19,84 @@ import { deepCompare } from "../utils";
...
+    created_ts?: number;
+
+    // Application specific data
+    scope?: CallScope;
+}
+const isSessionsMembershipData = (data: CallMembershipData): data is SessionMembershipData =>
+    "foci_active" in data &&
+    "foci_preferred" in data &&
+    "membership_id" in data &&
+    !Array.isArray(data.foci_active) &&
```
dbkr, line 43: Wouldn't "is string" be closer?

src/matrixrtc/CallMembership.ts
```diff
@@ -19,19 +19,84 @@ import { deepCompare } from "../utils";
...
+    membershipID: string;
     created_ts?: number;
     expires?: number;
     expires_ts?: number;
     foci_active?: Focus[];
-    membershipID: string;
 }
+const isLegacyCallMembershipData = (data: CallMembershipData): data is CallMembershipDataLegacy =>
+    "membershipID" in data && "foci_active" in data && Array.isArray(data.foci_active);
+const checkCallMembershipDataLegacy = (data: CallMembershipDataLegacy): void => {
```
dbkr, line 73: Nitpick in the extreme but... I like newlines between functions 😇

src/matrixrtc/CallMembership.ts
dbkr: Also, you're checking this isn't an array but the docs (and test examples) say it should be, so I'm confused.

src/matrixrtc/CallMembership.ts
```diff
@@ -114,15 +177,22 @@ export class CallMembership {
...
-    public getMsUntilExpiry(): number {
-        return this.getLocalExpiry() - Date.now();
+    public getMsUntilExpiry(): number | undefined {
+        if (isLegacyCallMembershipData(this.data)) return this.getLocalExpiry()! - Date.now();
     }
 
     public isExpired(): boolean {
-        return this.getMsUntilExpiry() <= 0;
+        if (isLegacyCallMembershipData(this.data)) return this.getMsUntilExpiry()! <= 0;
+        // MSC4143 events expire by beeing updated. So if the event exists, its not expired.
```
dbkr, line 186:
```suggestion
        // MSC4143 events expire by being updated. So if the event exists, its not expired.
```
🐝

src/matrixrtc/MatrixRTCSessionManager.ts
```diff
@@ -73,9 +73,9 @@ export class MatrixRTCSessionManager extends TypedEventEmitter<MatrixRTCSessionM
         }
         this.roomSessions.clear();
 
-        this.client.removeListener(ClientEvent.Room, this.onRoom);
-        this.client.removeListener(RoomEvent.Timeline, this.onTimeline);
-        this.client.removeListener(RoomStateEvent.Events, this.onRoomState);
+        this.client.off(ClientEvent.Room, this.onRoom);
+        this.client.off(RoomEvent.Timeline, this.onTimeline);
+        this.client.off(RoomStateEvent.Events, this.onRoomState);
```
dbkr, line 78: Well, sure, but... why?

dbkr (commented):
Sorry, I'm a bit lost here. This looks like way more than a refactor, this looks like a fundamental format change and trying to be backwards comapt. The description seems to suggest the only change is adding `foci_preferred`?

src/matrixrtc/MatrixRTCSessionManager.ts
toger5: Its what we mostly use (on, off) even though both are the same as (removeListener, addListener).
I find it easier to spot and to connect on and off then on and removeListener.

src/matrixrtc/CallMembership.ts
toger5: There are two versions of the CallMembershipData.
SessionMembershipData is the new one enforcing foci active to be a object.
`{"type":"livekit", "livekit_service_url":"https://sfu.abc"}`
Here the foci_preferred takes to role of listing all avialable foci.


And CallMembershipDataLegacy where foci_active is the array of availble foci.

src/matrixrtc/CallMembership.ts
```diff
@@ -19,19 +19,85 @@ import { deepCompare } from "../utils";
...
+    if (typeof data.call_id !== "string") throw new Error(prefix + "call_id must be string");
+    if (typeof data.application !== "string") throw new Error(prefix + "application must be a string");
+    if (typeof data.membershipID !== "string") throw new Error(prefix + "membershipID must be a string");
+    // optional elements
+    if (data.created_ts && typeof data.created_ts !== "number") throw new Error(prefix + "created_ts must be number");
+    // application specific data (we first need to check if they exist)
+    if (data.scope && typeof data.scope !== "string") throw new Error(prefix + "scope must be string");
+};
+
+export type CallMembershipData = CallMembershipDataLegacy | SessionMembershipData;
```
AndrewFerr, line 100: How about using an XOR?
```suggestion
export type CallMembershipData = XOR<CallMembershipDataLegacy, SessionMembershipData>;
```

This will let type checking do a lot more with `CallMembershipData` and may let you remove may of the runtime typechecks altogether (as long as typehints can be trusted / there aren't insidious `any`s lying around).

src/matrixrtc/CallMembership.ts
```diff
@@ -78,31 +134,39 @@ export class CallMembership {
...
+        if (!isLegacyCallMembershipData(this.data)) return undefined;
         if (this.data.expires) {
-            return this.createdTs() + this.data.expires;
+            // we know createdTs exists since we already do the isLegacyCallMembershipData check
+            return this.createdTs()! + this.data.expires;
         } else {
             // We know it exists because we checked for this in the constructor.
             return this.data.expires_ts!;
         }
     }
```
AndrewFerr, line 162: If `CallMembershipData` were an `XOR` type, the availability of fields could be deduced by static typechecking & the non-null assertion operators here could be removed.

AndrewFerr (changes requested):
Partial review

src/@types/event.ts
```diff
@@ -362,7 +363,10 @@ export interface StateEvents {
 
     // MSC3401
     [EventType.GroupCallPrefix]: IGroupCallRoomState;
-    [EventType.GroupCallMemberPrefix]: XOR<IGroupCallRoomMemberState, ExperimentalGroupCallRoomMemberState>;
+    [EventType.GroupCallMemberPrefix]: XOR<
+        IGroupCallRoomMemberState,
+        ExperimentalGroupCallRoomMemberState | {} | SessionMembershipData
+    >;
```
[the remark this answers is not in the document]
toger5: you are totally right!

src/matrixrtc/CallMembership.ts
toger5: I dont see how we could remove the non null assertion since expires_ts is an optional field on CallMembershipLegacyData and a missing field on SessionMembershipData.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -638,6 +703,20 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
 
         return m;
     }
+    /**
+     * Constructs our own membership
+     * @param prevMembership - The previous value of our call membership, if any
```
AndrewFerr, line 708:
```suggestion
```
to fix the JSDoc Checker error.

src/matrixrtc/LivekitFocus.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+export const isLivekitFocusConfig = (object: Focus): object is LivekitFocusConfig =>
+    object.type === "livekit" && "livekit_service_url" in object;
+
+export interface LivekitFocusActive extends Focus {
+    type: "livekit";
+    focus_selection: "oldest_membership";
+}
+export const isLivekitFocusActive = (object: Focus): object is LivekitFocusActive =>
+    object.type === "livekit" && "focus_selection" in object;
```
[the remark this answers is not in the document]
toger5: I dont see immediate value in the base type. But makeing the focus extend the focusConfig is good.

src/matrixrtc/CallMembership.ts
```diff
@@ -42,26 +113,16 @@ export class CallMembership {
         private parentEvent: MatrixEvent,
         private data: CallMembershipData,
```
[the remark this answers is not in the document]
AndrewFerr: Oops, not `IContent`, but `any`.

src/matrixrtc/LivekitFocus.ts
```diff
@@ -0,0 +1,40 @@
...
+export interface LivekitFocusConfig extends Focus {
+    type: "livekit";
+    livekit_service_url: string;
+}
+
+export const isLivekitFocusConfig = (object: Focus): object is LivekitFocusConfig =>
+    object.type === "livekit" && "livekit_service_url" in object;
+
+export interface LivekitFocus extends LivekitFocusConfig {
+    type: "livekit";
```
AndrewFerr, line 28: Nitpick: this can be removed as it's inherited from `LivekitFocusConfig` (even setting it to a different value is a compile-time error).

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -134,21 +159,32 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+                if (!Array.isArray(membershipContents)) {
+                    logger.warn(`Malformed member event from ${memberEvent.getSender()}: memberships is not an array`);
+                    continue;
+                }
+            } else {
+                // we have a MSC4143 event membership event
+                if (Object.keys(content).length !== 0) {
+                    membershipContents.push(content as CallMembershipData);
+                }
             }
```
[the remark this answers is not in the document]
toger5: this checks if we have an empty object. Empty is valid but means that we can ignore it. It basically just prohibts a logging `Couldn't construct call membership` when throwing here:
`                    const membership = new CallMembership(memberEvent, membershipData);
`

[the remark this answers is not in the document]
toger5: I will add a comment

src/matrixrtc/CallMembership.ts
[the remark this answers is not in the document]
toger5: IContent should also work since we only call it with IContent.

src/matrixrtc/CallMembership.ts
```diff
@@ -19,19 +19,85 @@ import { deepCompare } from "../utils";
...
+    scope?: CallScope;
+}
+const isSessionMembershipData = (data: CallMembershipData): data is SessionMembershipData =>
+    "foci_active" in data &&
+    "foci_preferred" in data &&
+    "membership_id" in data &&
+    !Array.isArray(data.foci_active) &&
+    Array.isArray(data.foci_preferred);
+
+const checkSessionsMembershipData = (data: SessionMembershipData): void => {
```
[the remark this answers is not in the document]
toger5: Mostly so I don't have to use return in one line functions.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -134,21 +159,33 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
-            if (!Array.isArray(eventMemberships)) {
-                logger.warn(`Malformed member event from ${memberEvent.getSender()}: memberships is not an array`);
+            if (membershipContents.length === 0) {
                 continue;
             }
 
-            for (const membershipData of eventMemberships) {
+            for (const membershipData of membershipContents) {
                 try {
                     const membership = new CallMembership(memberEvent, membershipData);
```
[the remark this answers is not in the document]
toger5: Yea hinting that the constructor does all the checking is a good point.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -613,30 +677,41 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+        const createdTs = prevMembership?.createdTs();
+        return {
             call_id: "",
             scope: "m.room",
             application: "m.call",
             device_id: this.client.getDeviceId()!,
             expires: this.relativeExpiry,
-            foci_active: this.activeFoci,
+            // TODO: Date.now() should be the origin_server_ts (now).
+            expires_ts: this.relativeExpiry + (createdTs ?? Date.now()),
```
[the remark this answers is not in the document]
toger5: relativeExpiry is already checked above:
```
        if (this.relativeExpiry === undefined) {
            throw new Error("Tried to create our own membership event when we're not joined!");
        }

```

I called the const createdTs because created_ts would make the linter mad.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -232,30 +270,39 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+     *                        For the livekit case this is a list of foci generated from the homeserver well-known, the current rtc session,
+     *                        or optionally other room members homeserver well known.
+     * @param joinConfig - Additional configuration for the joined session.
      */
-    public joinRoomSession(activeFoci: Focus[], manageMediaKeys?: boolean): void {
+    public async joinRoomSession(
+        fociActive: Focus,
+        fociPreferred: Focus[],
+        joinConfig?: JoinSessionConfig,
+    ): Promise<void> {
```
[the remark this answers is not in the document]
toger5: yes, this is only used in element-call -> all consumers will be updated.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -232,30 +270,39 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+     *                        For the livekit case this is a list of foci generated from the homeserver well-known, the current rtc session,
+     *                        or optionally other room members homeserver well known.
+     * @param joinConfig - Additional configuration for the joined session.
      */
-    public joinRoomSession(activeFoci: Focus[], manageMediaKeys?: boolean): void {
+    public async joinRoomSession(
+        fociPreferred: Focus[],
+        fociActive?: Focus,
+        joinConfig?: JoinSessionConfig,
+    ): Promise<void> {
```
AndrewFerr, line 283: Now that there's no more `await`, this can go back to being synchronous.
```suggestion
    public joinRoomSession(
        fociPreferred: Focus[],
        fociActive?: Focus,
        joinConfig?: JoinSessionConfig,
    ): void {
```

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -742,46 +826,54 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
-        const myPrevMembershipData = memberships.find((m) => m.device_id === localDeviceId);
-        let myPrevMembership;
-        try {
-            if (myCallMemberEvent && myPrevMembershipData && myPrevMembershipData.membershipID === this.membershipId) {
-                myPrevMembership = new CallMembership(myCallMemberEvent, myPrevMembershipData);
+        const content = myCallMemberEvent?.getContent<Record<string, unknown>>() ?? {};
+        const legacy = "memberships" in content || this.useLegacyMemberEvents;
+        let newContent: {} | ExperimentalGroupCallRoomMemberState | SessionMembershipData = {};
+        if (legacy) {
+            let myPrevMembership;
```
AndrewFerr, line 833: Nice to have:
```suggestion
            let myPrevMembership: CallMembership | undefined;
```

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -742,46 +826,54 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+                    isLegacyCallMembershipData(myPrevMembershipData) &&
+                    myPrevMembershipData.membershipID === this.membershipId
+                ) {
+                    myPrevMembership = new CallMembership(myCallMemberEvent, myPrevMembershipData);
+                }
+            } catch (e) {
+                // This would indicate a bug or something weird if our own call membership
+                // wasn't valid
+                logger.warn("Our previous call membership was invalid - this shouldn't happen.", e);
             }
```
AndrewFerr, line 852: Since what throws here is the `CallMembership` constructor, it would help to focus the try-catch block.  The constructor also has its own call to `isLegacyCallMembershipData`, so it can be removed from here.
```suggestion
            if (
                myCallMemberEvent &&
                myPrevMembershipData &&
                myPrevMembershipData.membershipID === this.membershipId
            ) {
                try {
                    myPrevMembership = new CallMembership(myCallMemberEvent, myPrevMembershipData);
                } catch (e) {
                    // This would indicate a bug or something weird if our own call membership
                    // wasn't valid
                    logger.warn("Our previous call membership was invalid - this shouldn't happen.", e);
                }
            }
```

src/matrixrtc/CallMembership.ts
```diff
@@ -14,24 +14,93 @@ See the License for the specific language governing permissions and
...
+
+    foci_active: Focus;
+    foci_preferred: Focus[];
+    created_ts?: number;
+
+    // Application specific data
+    scope?: CallScope;
+};
+
+export const isSessionMembershipData = (data: any): data is SessionMembershipData =>
```
AndrewFerr, line 43: I think it's most useful / accurate for `isSessionMembershipData` and `isLegacyCallMembershipData` to be used for distinguishing whether a `CallMembershipData` object is the legacy type or not.  That would let both functions specify a `CallMembershipData` as its parameter, and check only a single field.  This would effectively make it equivalent to an inlined typecheck (which would allow for type narrowing without a type predicate function, but would clog code / be less maintainable, as you pointed out).

The "full" checking functions could then accept an `any` and do the job of checking if an arbitrary object is one of the two types (which is how they're used in practice).

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -232,30 +270,39 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
-        this.emit(MatrixRTCSessionEvent.JoinStateChanged, true);
-        if (manageMediaKeys) {
+        if (joinConfig?.manageMediaKeys) {
             this.makeNewSenderKey();
             this.requestKeyEventSend();
         }
         // We don't wait for this, mostly because it may fail and schedule a retry, so this
         // function returning doesn't really mean anything at all.
         this.triggerCallMembershipEventUpdate();
+        this.emit(MatrixRTCSessionEvent.JoinStateChanged, true);
```
AndrewFerr, line 305: Since the event update isn't awaited, is there any need to have moved the `emit`?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -315,6 +362,21 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+            if (this.ownFocusActive.focus_selection === "oldest_membership") {
+                const oldestMembership = this.getOldestMembership();
+                return oldestMembership?.getPreferredFoci()[0];
+            }
+        }
+        if (!this.ownFocusActive) {
+            // we use the legacy call.member events so default to oldest member
+            const oldestMembership = this.getOldestMembership();
+            return oldestMembership?.getPreferredFoci()[0];
+        }
```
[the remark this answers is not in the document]
toger5: Even if the redundance is not nice to look at, I think that this way it is the easiest to follow the logic. On first sight one can see that it covers two types of membership structures. Also the two step oldest membership logic is intentional to communicate the idea that there could be other focus_selection methods.

src/matrixrtc/MatrixRTCSession.ts
toger5: It would change the logic a little bit however. removing the isLegacyCallMembershipData would allow a valid construction if an event formatted like a session Memberhsip would also work (but should not work).

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -646,15 +721,17 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
         myPrevMembershipData?: CallMembershipData,
         myPrevMembership?: CallMembership,
     ): boolean {
+        if (myPrevMembership && myPrevMembership.getMsUntilExpiry() === undefined) return false;
```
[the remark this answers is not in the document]
toger5: only one right?

src/matrixrtc/MatrixRTCSession.ts
toger5: No. It was needed because code triggered by the emit used `.isJoined` since. I changed this code however to use the isJoined parameter passed in this function so the order should not be required anymore. Its still save like this once someone decides to use `.isJoined` again somewhere in code called by the event.

src/matrixrtc/CallMembership.ts
toger5: I was trying that before but it turned out to be less readable. But you are right this is what it should be.
I still think it is a little bit less readable (in the CallMembership construcor) but it makes more sense like this.

src/matrixrtc/MatrixRTCSession.ts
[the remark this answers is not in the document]
toger5: Ah no we cannot do this.
`myPrevMembership && myPrevMembership.getMsUntilExpiry() === undefined` this is different logic: 
`(myPrevMembership?.getMsUntilExpiry() === undefined)` is true if `myPrevMembership = undefined`
while this would be false:
`myPrevMembership && myPrevMembership.getMsUntilExpiry() === undefined`

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -613,30 +677,41 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
             expires: this.relativeExpiry,
-            foci_active: this.activeFoci,
+            // TODO: Date.now() should be the origin_server_ts (now).
+            expires_ts: this.relativeExpiry + (createdTs ?? Date.now()),
+            // we use the fociPreferred since this is the list of foci.
+            // it is named wrong in the Legacy events.
+            foci_active: this.ownFociPreferred,
             membershipID: this.membershipId,
+            ...(createdTs ? { created_ts: createdTs } : {}),
+        };
```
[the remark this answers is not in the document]
AndrewFerr: Ah yes, that works even better.

src/matrixrtc/MatrixRTCSession.ts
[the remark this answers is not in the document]
AndrewFerr: Fair enough

src/matrixrtc/MatrixRTCSession.ts
AndrewFerr: If it's important enough to error-check the membership object, then it's inconsistent to say that we know its `CallMembershipDataLegacy` and assign a type to `memberships`.

Then again, the runtime checks balance out the risks of any "eager" type hints.  So I'm okay with leaving this as it is.

AndrewFerr (approved):
Some optional requests, but this looks good overall!

dbkr (commented):
I think there's a typo in the PR description where you've said `foci_active []` where you mean `focus_active`? This makes it quite a confusing read. ;)

Other than that, I think it looks plausible, although having all this to support both systems at once is heart rending. This might be more an MSC comment but what's the reason for needing to change foci_active to focus_active rather than leave it as an array?

toger5:
@dbkr I updated the description and the title. Thanks.

> focus_active rather than leave it as an array

An array forces us to specify how to deal with active elements of different types. This is either very complicated and/or has to be done on a per focus type bases. (It is not clear what use-cases this will have an how it will be used/ how we define how to interpret an array of active_foci. It could be to transition between infrastructure types or to have redundancy. In both cases all participants need to sync and map streams in both systems -> we need to expose this somehow outside the infrastructure itself... This all becomes more complicated than what we want matrixRTC to be.)
If there really is a use-case there will be specific custom active focus types that allow lists of foci and contain the necessary information how they are meant to be used. This also guarantees that everybody in the meeting is compatible with this custom focus type and knows how to deal with multiple active foci.)

> support both systems at once is heart rending

in this case it was easy enough and we are able to merge this into the js sdk without making breaking everything.
We basically want to make the transition less stressful for us so we dont have to get it working in a couple of days now since it is broken.

This way we can merge this and work on the reliable memberships with one aspect less to worry about since the js sdk will work with both formats.

dbkr (approved):

PR #4196, opened 2024-05-10
Fix state_events.ts types

t3chguy:
Regressed by https://github.com/matrix-org/matrix-js-sdk/pull/4105

### DIFF
--- a/src/@types/partials.ts
+++ b/src/@types/partials.ts
@@ -45,14 +45,6 @@ export enum RestrictedAllowType {
     RoomMembership = "m.room_membership",
 }
 
-export interface IJoinRuleEventContent {
-    join_rule: JoinRule; // eslint-disable-line camelcase
-    allow?: {
-        type: RestrictedAllowType;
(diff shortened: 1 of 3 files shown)


richvdh:
If this is a "fix", please could you link to the PR that broke it and describe what was broken, for future archaeologists?

richvdh:
The change looks right, but I'm somewhat surprised we didn't discover the problem sooner. Don't we hit this codepath in Element Web?

t3chguy:
> Don't we hit this codepath in Element Web?

Looks like it is due to explicit typecasts on the React SDK callers

image
image

richvdh:
> Looks like it is due to explicit typecasts on the React SDK callers

(we should probably remove those typecasts)

richvdh (approved):

t3chguy:
> (we should probably remove those typecasts)

Yup, have a changeset brewing locally with that and switching from the deprecated type

t3chguy:
It's now in https://github.com/matrix-org/matrix-react-sdk/pull/12513

PR #4197, opened 2024-05-10
Run complement-crypto in CI

kegsay:
Placeholder PR to get this working.


t3chguy:
@kegsay how long would you expect this run to take? If its more than ~3 mins then it should run only on the merge queue. Current longest CI here is 3 mins

kegsay:
It takes 7 minutes. Where do you want this to be placed?

t3chguy:
> Where do you want this to be placed?

Where it is placed does not matter, just add

```
if: github.event_name == 'merge_group'
```

t3chguy (approved):
🎉

PR #4200, opened 2024-05-15
Only pass id_server if we had one to begin with

t3chguy:
Fixes https://github.com/element-hq/element-web/issues/26627
Regressed by https://github.com/matrix-org/matrix-js-sdk/pull/1791


richvdh (approved):
This has been broken for *three years*? :scream:

PR #4201, opened 2024-05-15
Remove deprecated methods and fields

t3chguy:
Requires https://github.com/matrix-org/matrix-react-sdk/pull/12524

+ Removes `threepidCreds` on `EmailIdentityDict` - should use `threepid_creds`
+ Removes `Room::shouldUpgradeToVersion` - should use `Room::getRecommendedVersion` instead
+ Removes `user` field on UIA calls, uses `identifier` instead
+ Removes `address` field on Login request interface - should use `identifier` instead
+ Removes `IEncryptedFile` - should use `EncryptedFile` on types export instead
+ Removes `ICrossSigningKey` - should use `CrossSigningKeyInfo`
+ Removes `IJoinRuleEventContent` - should use `RoomJoinRulesEventContent`
+ Removes `home_server` field on RegisterResponse - should derive from `user_id` by splitting at the first colon instead


richvdh (approved):
loving it

PR #4202, opened 2024-05-15
Wire up verification cancel & mismatch for rust crypto

t3chguy:
Fixes https://github.com/element-hq/element-web/issues/27394

### DIFF
--- a/src/rust-crypto/verification.ts
+++ b/src/rust-crypto/verification.ts
@@ -710,10 +710,16 @@ export class RustSASVerifier extends BaseRustVerifer<RustSdkCryptoJs.Sas> implem
                     }
                 },
                 mismatch: (): void => {
-                    throw new Error("impl");
+                    const request = this.inner.cancelWithCode("m.mismatched_sas");
+                    if (request) {
+                        this.outgoingRequestProcessor.makeOutgoingRequest(request);
+                    }
                 },
                 cancel: (): void => {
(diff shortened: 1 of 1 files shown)


richvdh:
Thank you for picking this up, though it makes me very uneasy that we don't have a test (preferably an end-to-end one, using playwright) for this. The lack of tests is why we have all these rough edges in the first place :(.

t3chguy:
Fair, I did test it myself, can I sign a deed poll to rename myself to `Mr. Playwright` or does that not count?

richvdh (approved):
The change looks plausible

t3chguy:
I'll add a boring old automated Playwright test on https://github.com/matrix-org/matrix-react-sdk/pull/12526

richvdh:
FWIW, there's also a not-bad suite of Jest tests for verification at the js-sdk level in `spec/intec/crypto/verification.spec.ts`, which would also be nice to extend. But given a choice between that and a playwright test, I'd probably prefer the playwright one.

t3chguy:
Playwright test wired, had to disable it for legacy crypto as that doesn't show the expected mismatch modal and only shows the cancelled state as it does not set the `cancellationCode` as expected. Given its legacy and hopefully being ditched soon I don't think its worth digging into further.

andybalaam (approved):
Nice, thank you t3chguy!

PR #4210, opened 2024-05-16
`initRustCrypto`: allow app to pass in the store key directly

richvdh:
... instead of using the pickleKey. This allows us to avoid a slow PBKDF operation.

Part of https://github.com/element-hq/element-web/issues/26821.


andybalaam (approved):
Looks good, thanks!

PR #4211, opened 2024-05-17
ci: support matching branch names in complement-crypto

kegsay:
So you can make a branch `foo` and having CI run tests against the same named branch `foo` on the complement-crypto repo. This allows tests to be added during the PR process.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -77,6 +77,7 @@ jobs:
         uses: matrix-org/complement-crypto/.github/workflows/single_sdk_tests.yml@main
         with:
             use_js_sdk: "."
+            use_complement_crypto: "MATCHING_BRANCH"
 
     # we need this so the job is reported properly when run in a merge queue
     downstream-complement-crypto:


t3chguy:
Wouldn't this break if you land the js SDK pr via merge queue but don't have perms to land the complement pr, breaking All later merge queue runs.

kegsay:
> but don't have perms to land the complement pr, breaking All later merge queue runs.

For us I guess we will have perms, but for outsiders this is true. I can close this PR if you'd prefer to keep tracking `main`?

t3chguy:
I don't have merge perms on that complement crypto, nor do I think I should

kegsay:
What's the desired outcome here? If we don't want matching branch logic (because tests only run in the merge queue) then this should be closed.

t3chguy:
Either the branch matching needs to solve the problem outlined above or it'd cause more problems imo

kegsay:
Going to close this on the basis that:
 - complement crypto does not run on PR commits, only merge queue to develop
 - hence so long as the test has landed on complement crypto first, the merge queue will pass.

PR #4216, opened 2024-05-24
Bump matrix-sdk-crypto-wasm to 5.0.0

richvdh:
Slightly more involved than normal because it requires us to pass a backup version into `OlmMachine.importBackedUpRoomKeys`.

On the other hand we can now re-enable the test that was disabled in https://github.com/matrix-org/matrix-js-sdk/pull/4214 due to https://github.com/matrix-org/matrix-rust-sdk/issues/3447

Fixes: https://github.com/element-hq/element-web/issues/27165

element-web notes: Fix a bug in the Rust crypto implementation which caused repeated HTTP requests to `/keys/query` when the session was unverified.

### DIFF
--- a/src/crypto/index.ts
+++ b/src/crypto/index.ts
@@ -1888,7 +1888,11 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
     /**
      * Implementation of {@link CryptoBackend#importBackedUpRoomKeys}.
      */
-    public importBackedUpRoomKeys(keys: IMegolmSessionData[], opts: ImportRoomKeysOpts = {}): Promise<void> {
+    public importBackedUpRoomKeys(
+        keys: IMegolmSessionData[],
+        backupVersion: string,
+        opts: ImportRoomKeysOpts = {},
+    ): Promise<void> {
         opts.source = "backup";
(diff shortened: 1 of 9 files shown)


src/crypto/index.ts
```diff
@@ -1888,7 +1888,11 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
     /**
      * Implementation of {@link CryptoBackend#importBackedUpRoomKeys}.
      */
-    public importBackedUpRoomKeys(keys: IMegolmSessionData[], opts: ImportRoomKeysOpts = {}): Promise<void> {
+    public importBackedUpRoomKeys(
```
andybalaam, line 1891: Does this change a public interface and do we need to do something if so?

andybalaam (approved):
Looks good; one question.

src/crypto/index.ts
richvdh: no, CryptoBackend is an internal interface.

richvdh: (and legacy `Crypto` is documented as being an internal interface, even though we weren't very good at following that.)

richvdh:
The coverage gate is complaining about legacy crypto. I'm afraid I'm not going to fix that.

PR #4218, opened 2024-05-24
Don't run migration for Rust crypto if the legacy store is empty

andybalaam:
Fixes https://github.com/element-hq/element-web/issues/27447

### DIFF
--- a/spec/test-utils/test_indexeddb_cryptostore_dump/empty_account/index.ts
+++ b/spec/test-utils/test_indexeddb_cryptostore_dump/empty_account/index.ts
@@ -1,3 +1,19 @@
+/*
+Copyright 2024 The Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 import { DumpDataSetInfo } from "../index";
 
 /**
--- a/spec/test-utils/test_indexeddb_cryptostore_dump/empty_account/index.ts
+++ b/spec/test-utils/test_indexeddb_cryptostore_dump/empty_account/index.ts
@@ -1,0 +1,31 @@
+import { DumpDataSetInfo } from "../index";
+
+/**
+ * A key query response containing the current keys of the tested user.
+ * To be used during tests with fetchmock.
+ */
+const KEYS_QUERY_RESPONSE: any = {
+    device_keys: {},
+    master_keys: {},
+    self_signing_keys: {},
+    user_signing_keys: {},
+};
+
+/**
+ * A `/room_keys/version` response containing the current server-side backup info.
+ * To be used during tests with fetchmock.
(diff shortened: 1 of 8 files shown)


spec/test-utils/test_indexeddb_cryptostore_dump/empty_account/index.ts
t3chguy, line 1: Missing copyright

src/rust-crypto/libolm_migration.ts
```diff
@@ -83,6 +83,19 @@ export async function migrateFromLegacyCrypto(args: {
     }
 
     await legacyStore.startup();
+
+    let account;
```
t3chguy, line 87: type please

t3chguy (changes requested):

spec/test-utils/test_indexeddb_cryptostore_dump/empty_account/index.ts
andybalaam: Thanks, fixed in be1a90a1b

src/rust-crypto/libolm_migration.ts
andybalaam: Well spotted, thanks. Fixed in 8b85c6955

t3chguy (approved):

spec/test-utils/test_indexeddb_cryptostore_dump/empty_account/index.ts
```diff
@@ -0,0 +1,47 @@
...
+ * To be used during tests with fetchmock.
+ */
+const BACKUP_RESPONSE: any = {};
+
+/**
+ * A dataset containing the information for the tested user.
+ * To be used during tests.
+ */
+export const EMPTY_ACCOUNT_DATASET: DumpDataSetInfo = {
+    userId: "@emptyuser:example.com",
```
BillCarsonFr, line 41: wouldn't this be null if no account?

richvdh: No, I don't think so. `userId` is documented as "The user ID to use for the test.". It doesn't really matter what user ID we use here, since the data store is empty (so there are no pickled values to unpickle), but `@emptyuser:example.com` seems a reasonable one to me.

spec/test-utils/test_indexeddb_cryptostore_dump/empty_account/index.ts
```diff
@@ -0,0 +1,47 @@
...
+    deviceId: "EMPTYDEVIC",
+    pickleKey: "+/bcdefghijklmnopqrstu1/zyxvutsrqponmlkjih2",
+    backupResponse: BACKUP_RESPONSE,
```
richvdh, line 44: I think we can get rid of this though. We don't need a backup here.

src/rust-crypto/libolm_migration.ts
```diff
@@ -83,6 +83,19 @@ export async function migrateFromLegacyCrypto(args: {
...
+
+    let accountPickle: string | null = null;
+    await legacyStore.doTxn("readonly", [IndexedDBCryptoStore.STORE_ACCOUNT], (txn) => {
+        legacyStore.getAccount(txn, (acctPickle) => {
+            accountPickle = acctPickle;
+        });
+    });
+    if (!accountPickle) {
+        // This store is not properly set up. Nothing to migrate.
+        logger.warn("Legacy crypto store is not set up (no account found). Not migrating.");
```
richvdh, line 95: I don't think this should be a warning, since it will be logged on every client load.

richvdh (approved):
lgtm

PR #4227, opened 2024-06-03
Add crypto methods for export and import of secrets bundle

t3chguy:
Split out from https://github.com/matrix-org/matrix-js-sdk/pull/4134/

### DIFF
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -1555,6 +1555,15 @@ describe("RustCrypto", () => {
             };
             await rustCrypto.importSecretsForQrLogin(bundle);
             await expect(rustCrypto.exportSecretsForQrLogin()).resolves.toEqual(expect.objectContaining(bundle));
+            await expect(rustCrypto.getCrossSigningStatus()).resolves.toEqual({
+                privateKeysCachedLocally: {
+                    masterKey: true,
+                    selfSigningKey: true,
+                    userSigningKey: true,
+                },
+                privateKeysInSecretStorage: false,
+                publicKeysOnDevice: true,
+            });
         });
     });
 });
@@ -1576,7 +1585,10 @@ async function makeTestRustCrypto(
     http: MatrixHttpApi<IHttpOpts & { onlyData: true }> = {} as MatrixClient["http"],
     userId: string = TEST_USER,
     deviceId: string = TEST_DEVICE_ID,
-    secretStorage: ServerSideSecretStorage = {} as ServerSideSecretStorage,
+    secretStorage = {
+        isStored: jest.fn().mockResolvedValue(null),
+        getDefaultKeyId: jest.fn().mockResolvedValue("key"),
+    } as unknown as ServerSideSecretStorage,
     cryptoCallbacks: CryptoCallbacks = {} as CryptoCallbacks,
 ): Promise<RustCrypto> {
     return await initRustCrypto({
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -1516,7 +1516,7 @@ describe("RustCrypto", () => {
         });
     });
 
-    describe("exportSecretsForQrLogin", () => {
+    describe("import & export secrets bundle", () => {
         let rustCrypto: RustCrypto;
 
         beforeEach(async () => {
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -1536,7 +1536,7 @@ describe("RustCrypto", () => {
             );
         });
 
-        it("should return a JSON secrets bundle if there is something to export", async () => {
+        it("should correctly import & export a secrets bundle", async () => {
             const bundle = {
                 cross_signing: {
                     master_key: "bMnVpkHI4S2wXRxy+IpaKM5PIAUUkl6DE+n0YLIW/qs",
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -1515,6 +1515,48 @@ describe("RustCrypto", () => {
             expect(await rustCrypto.isDehydrationSupported()).toBe(true);
         });
     });
+
+    describe("exportSecretsForQrLogin", () => {
+        let rustCrypto: RustCrypto;
+
+        beforeEach(async () => {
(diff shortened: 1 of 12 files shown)


src/@types/matrix-sdk-crypto-wasm.d.ts
```diff
@@ -0,0 +1,40 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import type * as RustSdkCryptoJs from "@matrix-org/matrix-sdk-crypto-wasm";
+
+declare module "@matrix-org/matrix-sdk-crypto-wasm" {
+    interface OlmMachine {
```
BillCarsonFr, line 20: Why do we need this? isn't it already in the index.d.ts of `matrix-sdk-crypto-wasm`?

src/crypto/index.ts
```diff
@@ -576,6 +577,18 @@ export class Crypto extends TypedEventEmitter<CryptoEvent, CryptoEventHandlerMap
         }
     }
 
+    public supportsSecretsForQrLogin(): boolean {
+        return false;
+    }
+
+    public async exportSecretsForQrLogin(): Promise<QRSecretsBundle> {
+        throw new Error("Method not implemented.");
```
BillCarsonFr, line 585: More a not supported than a not implemented?

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1515,6 +1515,48 @@ describe("RustCrypto", () => {
...
+                    self_signing_key: "pfDknmP5a0fVVRE54zhkUgJfzbNmvKcNfIWEW796bQs",
+                },
+                backup: {
+                    algorithm: "m.megolm_backup.v1.curve25519-aes-sha2",
+                    key: "bYYv3aFLQ49jMNcOjuTtBY9EKDby2x1m3gfX81nIKRQ",
+                    backup_version: "9",
+                },
+            };
+            await rustCrypto.importSecretsForQrLogin(bundle);
+            await expect(rustCrypto.exportSecretsForQrLogin()).resolves.toEqual(expect.objectContaining(bundle));
```
BillCarsonFr, line 1557: Maybe we could check the getCrossSigningStatus? and see that the identity is there and trusted?

BillCarsonFr (approved):
Some minor comments

src/@types/matrix-sdk-crypto-wasm.d.ts
t3chguy: Yes but with return types of `any` - see https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/110

t3chguy: image

src/crypto/index.ts
t3chguy, line 585: We seem to use `Not implemented` already so this is consistent.

image

t3chguy: We seem to use `Not implemented` already so this is consistent.

image

spec/unit/rust-crypto/rust-crypto.spec.ts
t3chguy: Done

t3chguy: This causes the test to hang on `OlmMachine::getIdentity` and I have 0 introspection on the rust side of things to debug this.

src/@types/matrix-sdk-crypto-wasm.d.ts
richvdh: This should be fixed on the wasm side; https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/123 does so but it will need a release cycle :/

richvdh: Well, that PR doesn't help with the return type of `SecretsBundle.to_json`, to be fair. On the other hand: do you actually need to introspect the result of `SecretsBundle.to_json` outside of tests (where I would be inclined to do some `@ts-ignore`ing rather than have a separate definition to keep in step)?

richvdh (changes requested):

src/@types/matrix-sdk-crypto-wasm.d.ts
t3chguy: > Well, that PR doesn't help with the return type of SecretsBundle.to_json, to be fair. On the other hand: do you actually need to introspect the result of SecretsBundle.to_json outside of tests (where I would be inclined to do some @ts-ignoreing rather than have a separate definition to keep in step)?

It should still be a useful type otherwise you'll get junk into the import method. If it isn't meant to be introspectable then it should be a branded type so a round-trip export->import works without needing to rely on `any`

t3chguy:
> We could export the secrets bundle from one MatrixClient, and import it into another, and then check that the second client has the same cross-signing keys as the first.

We already do. I exported a bundle from a test account, baked it into the test, import it and validate that exporting yields the same bundle once again.

richvdh:
errr I seem to have pressed enter halfway through editing that comment, so sorry for the half-formed thoughts

richvdh:
> > We could export the secrets bundle from one MatrixClient, and import it into another, and then check that the second client has the same cross-signing keys as the first.
> 
> We already do. I exported a bundle from a test account & baked it into the test, import it and validate that exporting yields the same bundle once again.

So we do. Apparently I need to read better.

src/rust-crypto/rust-crypto.ts
```diff
@@ -1182,6 +1166,26 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
         this.checkKeyBackupAndEnable();
     }
 
+    /**
+     * Implementation of {@link CryptoApi#importSecretsBundle}.
+     */
+    public async importSecretsBundle(
+        secrets: Parameters<NonNullable<CryptoApi["importSecretsBundle"]>>[0],
```
richvdh, line 1173: I still think there's way too much type magic going on here. As a caller of this function, how am I meant to build one of these things?

richvdh (approved):
Thanks. Sorry this took a while to review.

richvdh:
I want to do some work in RustCrypto that will conflict with this, so merging

src/rust-crypto/rust-crypto.ts
t3chguy: > As a caller of this function, how am I meant to build one of these things?

you shouldn't construct manually, only ever exported by the matching export function

PR #4228, opened 2024-06-04
Fix incorrect assumptions about required fields in /search response

t3chguy:
Requires https://github.com/matrix-org/matrix-react-sdk/pull/12575


florianduros:
Can you give a bit of context ? thanks

t3chguy:
The spec doesn't say those fields are required yet we are assuming they are.

florianduros (approved):

PR #4240, opened 2024-06-13
Replace usages of setImmediate with setTimeout for wider compatibility

t3chguy:
https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate#browser_compatibility

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -406,7 +406,9 @@ export async function logDuration<T>(logger: BaseLogger, name: string, block: ()
 }
 
 /**
- * Promise/async version of {@link setImmediate} based on setTimeout 0 for wider compatibility.
+ * Promise/async version of {@link setImmediate}.
+ *
+ * Implementation is based on `setTimeout` for wider compatibility.
  * @deprecated Use {@link sleep} instead.
  */
 export function immediate(): Promise<void> {
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -409,7 +409,7 @@ export async function logDuration<T>(logger: BaseLogger, name: string, block: ()
  * Promise/async version of {@link setImmediate}.
  */
 export function immediate(): Promise<void> {
-    return new Promise(setImmediate);
+    return new Promise((resolve) => setTimeout(resolve));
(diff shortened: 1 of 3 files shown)


toger5 (approved):
Looks good.

Very nice to have the lint rule.

src/utils.ts
```diff
@@ -409,7 +409,7 @@ export async function logDuration<T>(logger: BaseLogger, name: string, block: ()
  * Promise/async version of {@link setImmediate}.
  */
 export function immediate(): Promise<void> {
-    return new Promise(setImmediate);
+    return new Promise((resolve) => setTimeout(resolve));
```
richvdh, line 412: Given this is only used in one place, can we just kill it off (possibly replacing it with a call to `sleep(0)`) ?

Otherwise, it feels like the documentation could do with updating to explain what it does without reference to a heavily deprecated function.

richvdh, line 412: (aside: given this doesn't work any more in modern browsers, how did we get this far without fixing it? Is there a polyfill somewhere we can disable?)

t3chguy: Not without a breaking change, given its in an export which other packages use.

t3chguy: Node supports it, EW loads a polyfill for both runtime & jest tests. See related PRs
image

richvdh: UGH. this is not something the js-sdk ought to be exporting.

Let's improve the documentation then, and/or deprecate it?

src/utils.ts
```diff
@@ -406,10 +406,11 @@ export async function logDuration<T>(logger: BaseLogger, name: string, block: ()
 }
 
 /**
- * Promise/async version of {@link setImmediate}.
+ * Promise/async version of {@link setImmediate} based on setTimeout 0 for wider compatibility.
```
richvdh, line 409:
```suggestion
 * Promise/async version of {@link setImmediate}.
 *
 * Implementation is based on `setTimeout` for wider compatibility.
```

richvdh (approved):

PR #4241, opened 2024-06-13
Fix screen sharing in recent Chrome

dbkr:
Dreadful hack to work around a bug in recent chrome/electron's WebRTC, as explained.

I'm not sure which is the least hideous out of this (ie. repeatedly calling setCodecPreferences and seeing if it crashes each time) or hardcoding the bad codec and skipping it. Opinions welcome. Edit: LiveKit work around this bug by using the receiver codecs (https://github.com/livekit/client-sdk-js/pull/1088) so that's also an option.

This will probably want backporting to the release.

Fixes https://github.com/element-hq/element-desktop/issues/1703

## Checklist

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -25,7 +25,7 @@ import { v4 as uuidv4 } from "uuid";
 import { parse as parseSdp, write as writeSdp } from "sdp-transform";
 
 import { logger } from "../logger";
-import { checkObjectHasKeys, deepCompare, isNullOrUndefined, recursivelyAssign } from "../utils";
+import { checkObjectHasKeys, isNullOrUndefined, recursivelyAssign } from "../utils";
 import { MatrixEvent } from "../models/event";
 import { EventType, TimelineEvents, ToDeviceMessageId } from "../@types/event";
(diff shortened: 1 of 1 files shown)


src/webrtc/call.ts
```diff
@@ -2381,22 +2381,42 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
...
+                }
             }
         }
 
-        const screenshareVideoTransceiver = this.transceivers.get(
-            getTransceiverKey(SDPStreamMetadataPurpose.Screenshare, "video"),
-        );
-        // setCodecPreferences isn't supported on FF (as of v113)
-        screenshareVideoTransceiver?.setCodecPreferences?.(codecs);
+        //screenshareVideoTransceiver.setCodecPreferences(codecs);
```
t3chguy, line 2419: Commented out code

t3chguy (approved):
Looks sane otherwise

src/webrtc/call.ts
dbkr: oops, thanks

fkwp (approved):
lgtm

PR #4242, opened 2024-06-13
Fix room state being updated with old (now overwritten) state and emitting for those updates.

toger5:
This is a workaround at best and mostly serves the purpose to show that this issue exist and to have a starting point to look into the underlying problem.
## Checklist

### DIFF
--- a/src/models/room-state.ts
+++ b/src/models/room-state.ts
@@ -420,6 +420,12 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
             }
 
             const lastStateEvent = this.getStateEventMatching(event);
+            // Safety measurs to not update the room (and emit the update) with older state.
+            // The sync loop really should not send old events but it does very regularly.
+            // Logging on return in those two conditions results in a large amount of logging. (on startup and when running element)
+            if ((lastStateEvent?.event.origin_server_ts ?? 0) > (event.event.origin_server_ts ?? 1)) return;
+            if (lastStateEvent?.event.unsigned?.replaces_state === event.event.event_id) return;
+
             this.setStateEvent(event);
             if (event.getType() === EventType.RoomMember) {
                 this.updateDisplayNameCache(event.getStateKey()!, event.getContent().displayname ?? "");
--- a/src/models/room-state.ts
+++ b/src/models/room-state.ts
@@ -166,7 +166,6 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
 
     public readonly beacons = new Map<BeaconIdentifier, Beacon>();
     private _liveBeaconIds: BeaconIdentifier[] = [];
-    private _isStartTimelineState = false;
 
     /**
      * Construct room state.
@@ -195,14 +194,24 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
      * As the timeline might get reset while they are loading, this state needs to be inherited
      * and shared when the room state is cloned for the new timeline.
      * This should only be passed from clone.
+     * @param isStartTimelineState - Optional. This state is marked as a start state.
+     * This is used to skip state insertions that are
+     * in the wrong order. The order is determined by the `replaces_state` id.
+     *
+     * Example:
+     * A current state events `replaces_state` value is `1`.
+     * Trying to insert a state event with `event_id` `1` in its place would fail if isStartTimelineState = false.
+     *
+     * A current state events `event_id` is `2`.
+     * Trying to insert a state event where its `replaces_state` value is `2` would fail if isStartTimelineState = true.
      */
+
     public constructor(
         public readonly roomId: string,
         private oobMemberFlags = { status: OobStatus.NotStarted },
-        isStartTimelineState: boolean = false,
+        public readonly isStartTimelineState = false,
     ) {
         super();
-        this._isStartTimelineState = isStartTimelineState;
         this.updateModifiedTime();
     }
 
@@ -340,21 +349,6 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
     }
 
     /**
-     * This state is marked as a start state. This is used to skip state insertions that are
-     * in the wrong order. The order is determined by the `replaces_state` id.
-     *
-     * Example:
-     * A current state events `replaces_state` value is `1`.
-     * Trying to insert a state event with `event_id` `1` in its place would fail if isStartTimelineState = false.
-     *
-     * A current state events `event_id` is `2`.
-     * Trying to insert a state event where its `replaces_state` value is `2` would fail if isStartTimelineState = true.
-     */
-    public get isStartTimelineState(): boolean {
-        return this._isStartTimelineState;
-    }
-
-    /**
      * Creates a copy of this room state so that mutations to either won't affect the other.
      * @returns the copy of the room state
      */
@@ -446,11 +440,11 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
             const lastId = lastStateEvent?.event.event_id;
             const newReplaceId = event.event.unsigned?.replaces_state;
             const newId = event.event.event_id;
-            if (this._isStartTimelineState) {
-                // Add an event to the start of the timeline. Its replace id not be the same as the one of the current/last start state event.
-                if (newReplaceId && newId && newReplaceId === lastId) return;
+            if (this.isStartTimelineState) {
+                // Add an event to the start of the timeline. Its replace id should not be the same as the one of the current/last start state event.
+                if (newReplaceId && lastId && newReplaceId === lastId) return;
             } else {
-                // Add an event to the end of the timeline. It should not be the same as the one replaced but the current/last end state event.
+                // Add an event to the end of the timeline. It should not be the same as the one replaced by the current/last end state event.
                 if (lastReplaceId && newId && lastReplaceId === newId) return;
             }
 
--- a/src/models/room-state.ts
+++ b/src/models/room-state.ts
@@ -43,7 +43,16 @@ export interface IMarkerFoundOptions {
      */
     timelineWasEmpty?: boolean;
 }
-
+export interface ISetStateEventsOptions {
+    /** Whether the event is inserted at the start of the timeline
+     * or at the end. This is relevant for doing a sanity check:
+     * "Is the event id of the added event the same as the replaces_state id of the current event"
+     * We need to do this because sync sometimes feeds previous state events.
+     * If set to true the sanity check is inverted
+     * "Is the event id of the current event the same as the replaces_state id of the added event"
+     */
+    toStartOfTimeline?: boolean;
+}
 // possible statuses for out-of-band member loading
 enum OobStatus {
     NotStarted,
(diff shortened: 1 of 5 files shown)


t3chguy (changes requested):

src/models/room-state.ts
```diff
@@ -420,6 +420,12 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
             }
 
             const lastStateEvent = this.getStateEventMatching(event);
+            // Safety measurs to not update the room (and emit the update) with older state.
+            // The sync loop really should not send old events but it does very regularly.
+            // Logging on return in those two conditions results in a large amount of logging. (on startup and when running element)
+            if ((lastStateEvent?.event.origin_server_ts ?? 0) > (event.event.origin_server_ts ?? 1)) return;
```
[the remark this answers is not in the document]
toger5: That is why I am labeling this as a workaround.
> This is a workaround at best and mostly serves the purpose to show that this issue exist and to have a starting point to look into the underlying problem.

There is also this chat (monologue) for context https://matrix.to/#/!jxlRxnrZCsjpjDubDX:matrix.org/$2X7ysozWciiyDHVev8iJ5yaFcZbZr57H2u8v5-7ysXw?via=matrix.org&via=envs.net&via=mozilla.org

>This is not how we do state resolution however...
(using the timestamp) This logic would need to use the replaces_state to be correct... But there is no good condition at which I can stop paginating back to see if there is really no previous state.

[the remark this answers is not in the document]
t3chguy: Sure, but its a workaround which will cause the state to be wrong in the other direction whenever someone has the time set wrong on their server, which can also be done maliciously.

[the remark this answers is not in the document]
toger5: I 100% agree that the timestamp fix is bad.
I really hope, that we dont merge any of this tbh.
It is just very clear in this PR what situation the current sync causes and where the issue lies.
Since the replces_state condition also covers the acute issue with calls being closed on join this could be a hotfix for the underlying issue that we can even reach this function with old state.

Do you have any idea why we do get some events twice and in the wrong order?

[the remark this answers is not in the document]
t3chguy: Also this code would brake for maintaining `oldState` on a room, as the state there does indeed go reverse and backwards in time

> Do you have any idea why we do get some events twice and in the wrong order?

Probably due to the above, handling the edge in two directions, at each end of the live timeline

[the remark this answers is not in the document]
toger5: But when i read the state I am using `getLiveTimeline().getState(FORWARD)` so we should always read from  the correct state.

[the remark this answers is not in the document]
t3chguy: Sure, but this method you are editing is in both Forward & Backward RoomState classes. So the timestamp comparison would at least need inverting for the Backward case.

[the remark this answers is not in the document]
t3chguy: > Where do you think one should look for a proper fix?

Would have to try trace what is causing events to be emitted out of order, I'd suggest `console.trace` on a dev build on which you try reproduce the issue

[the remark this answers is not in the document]
toger5: Since this is a very fundamental issue I am wondering if it if someone who has long experience with the sync code is a better fit to investigate this?
Putting a breakpoint at the return in this pr allows to reproduce it.

I think I will try to adjust the pr like described here:
>  I think i will change this pr to not do any timestamp logic but only use the replaces_state condition and only in the case we are adding to endState.

so it is actually a valid hotfix.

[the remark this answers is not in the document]
toger5: I updated this to respect `toStartOfTimeline` and removed the timestamp check.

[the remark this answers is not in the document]
toger5: I am also wondering why we don't do any ordering based on `toStartOfTimeline`.
`setStateEvents` should do `stateEvents.reverse().forEach` if we insert at the start of the timeline (`startState.setStateEvents`)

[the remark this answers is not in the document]
t3chguy: > Since this is a very fundamental issue I am wondering if it if someone who has long experience with the sync code is a better fit to investigate this?


I'm unfortunately not the best person for this

> so it is actually a valid duct tape fix.

Agreed

It'd be good to isolate the changes in this PR those those concerning the title of the PR, the RTC stuff is unrelated. It'd also be good to have tests to assert this does the right thing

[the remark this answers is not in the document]
toger5: Oh definitely. not exactly sure how those changes sneaked in...

toger5:
I added tests to see what this actually does.

src/models/room-state.ts
```diff
@@ -419,13 +428,26 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
                 this.setBeacon(event);
             }
 
-            const lastStateEvent = this.getStateEventMatching(event);
+            const lastStateEv = this.getStateEventMatching(event);
```
t3chguy, line 431: Why did this get renamed to something less clear?

src/models/room-state.ts
```diff
@@ -408,7 +417,7 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
      * Fires {@link RoomStateEvent.Events}
      * Fires {@link RoomStateEvent.Marker}
      */
-    public setStateEvents(stateEvents: MatrixEvent[], markerFoundOptions?: IMarkerFoundOptions): void {
+    public setStateEvents(stateEvents: MatrixEvent[], options?: IMarkerFoundOptions & ISetStateEventsOptions): void {
```
t3chguy, line 420: Why did this param name change?

src/models/room-state.ts
toger5: Otherwise, we had a lot more linewraps that made I judged as harder to read overall.

src/models/room-state.ts
toger5: Because it's now not only the marker options anymore but also includes the toStartOfTimeline prop.

src/models/room-state.ts
```diff
@@ -43,7 +43,16 @@ export interface IMarkerFoundOptions {
...
 }
-
+export interface ISetStateEventsOptions {
+    /** Whether the event is inserted at the start of the timeline
+     * or at the end. This is relevant for doing a sanity check:
+     * "Is the event id of the added event the same as the replaces_state id of the current event"
+     * We need to do this because sync sometimes feeds previous state events.
+     * If set to true the sanity check is inverted
+     * "Is the event id of the current event the same as the replaces_state id of the added event"
+     */
```
t3chguy, line 53: I don't think this makes sense, as a RoomState is only ever in one direction. This sounds like if toStartOfTimeline=false then you should always drop the state event as they're never live. It should be a property on a RoomState if required

src/models/room-state.ts
t3chguy: That's only because on L435 and L436 you're using the array syntax to set variables which is really redundant. It'd be clearer as 4 separate instantiations.

src/models/room-state.ts
toger5: That is a good observation. The state itself might be the right location to store if it's a start state or an end state.
It should never change and also not be relevant outside the state object it seems logical to belong to the state.

I have trouble understanding the exact definition of start state anyway however:

If we have a state that is only set at the beginning of the room, events from a time-line in a specific interval cannot be enough to compute the start state because the start state needs all state events BEFORE the start of the time-line but the time-line only includes events after the "start".

This is slightly off topic but i still wanted to being it up. Feel free to DM me about this or point me to where i can lookup how start state is defined.

t3chguy: as I understand it, it is the state at the start of the timeline, and it walks backwards as you backpaginate and load older events which undo the more recent state events.

toger5: That would also be how i understand start state. But this is not what we are doing.
We use the content to set the state keys. What we would need to do is set the state with the previous_content field.

| = start of timeline
|| = end of timeline
--- = time steps
XV> = state event changing the state (the content of the state event is the current state from that time on (X is an arbitrary label for the state event (f.ex. a member event) X is the type and V is the value

X0>-----|X1>----X2>----||
Here we have an example where the state is changed with the first event in the timeline. But the start imo is before that  event. So the state at the start is X0.

This becomes more clear when adding another type:
X0>--A0>--|A1>-A2>-X1>--||
At the start of the timeline we have X0 A0 but definitly not X1 which would be the first X type state event.

If we use the previous_content we have <XV as the previous content.

<X0>--<A0>--|<A1>-<A2>-<X1>--||

Then <A1 & <X1 is the same as X0> & A0> and the correct start state.

t3chguy: I suggest raising this to the wider team as the code & intent predates me

toger5: sure. Sorry for escalating this but I always found the startState to be arguably the correct name for that variable the way its implemented and this seemed to be relevant for this discussion. Somewhat relevant, you comment is just about where to store start/end.
What I do like about making it part of the setStateEvents property is that is makes it testable in a more isolated env.

toger5: To get back to the topic what do you think about the more encapsulated testability with the current approach?

t3chguy (commented):
I think your suggestions are sane

fkwp:
> I think your suggestions are sane

who else do we need to unlock this? Could you pls invite?

t3chguy:
My outstanding comments need dealing with, e.g. property of room state and random variable name change

toger5:
@t3chguy 
Just to make sure this comment:
>My outstanding comments need dealing with, e.g. property of room state and random variable name change

did not get missed,
this
> To get back to the topic what do you think about the more encapsulated testability with the current approach?

implies you still prefer making it part of the state object although it allows the encapsulated testing?

toger5:
@t3chguy I updated the variable names back to what they were. And noticed the test can be done equally simple.

src/models/room-state.ts
```diff
@@ -198,8 +199,10 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
     public constructor(
         public readonly roomId: string,
         private oobMemberFlags = { status: OobStatus.NotStarted },
+        isStartTimelineState: boolean = false,
     ) {
         super();
+        this._isStartTimelineState = isStartTimelineState;
```
t3chguy, line 205: Why not use the shorthand like the above fields?

t3chguy: `private readonly isStartTimelineState = false` then we don't need the getter either

src/models/room-state.ts
```diff
@@ -408,7 +426,7 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
      * Fires {@link RoomStateEvent.Events}
      * Fires {@link RoomStateEvent.Marker}
      */
-    public setStateEvents(stateEvents: MatrixEvent[], markerFoundOptions?: IMarkerFoundOptions): void {
+    public setStateEvents(stateEvents: MatrixEvent[], options?: IMarkerFoundOptions): void {
```
t3chguy, line 429: Not sure why this is changing anymore

src/models/room-state.ts
```diff
@@ -420,6 +438,22 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
...
+            const lastReplaceId = lastStateEvent?.event.unsigned?.replaces_state;
+            const lastId = lastStateEvent?.event.event_id;
+            const newReplaceId = event.event.unsigned?.replaces_state;
+            const newId = event.event.event_id;
+            if (this._isStartTimelineState) {
+                // Add an event to the start of the timeline. Its replace id not be the same as the one of the current/last start state event.
+                if (newReplaceId && newId && newReplaceId === lastId) return;
+            } else {
+                // Add an event to the end of the timeline. It should not be the same as the one replaced but the current/last end state event.
+                if (lastReplaceId && newId && lastReplaceId === newId) return;
```
t3chguy, line 454: I think one of these checks is wrong, they both check `newId` as a truthy but neither checks `lastId`

src/models/room-state.ts
toger5: Its changed back to not have the option to configure isStartOfTimeline when callling setStateEvents.
It is now what it was initially.

t3chguy (approved):

PR #4246, opened 2024-06-17
Fetch capabilities in the background

dbkr:
& keep them up to date.

I've elected to change the interface here and split it into two functions rather than re-use the same one. That way, the cached one doesn't need to return a promise. This does make it a BREAKING CHANGE. I've given both functions different names so it's obvious that they're different. It's a trivial code update so I don't really think its worth having a backwards compat function?

It also splits some code out of client.ts(!)

Any of the tests that use `runAllTimers()` needed replacing with something that would just run the ones they need, since this works by continually setting a timer to keep the capabilities up to date. I don't think `runAllTimers()` is a sensible thing to use unless you're testing a small piece of code where you know all the timers it's setting.

## Checklist

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2065,13 +2065,25 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 
     /**
+     * Gets the cached capabilities of the homeserver, returning cached ones if available.
+     * If there are no cached capabilities and none can be fetched, throw an exception.
+     *
+     * @returns Promise resolving with The capabilities of the homeserver
+     */
+    public async getCapabilities(): Promise<Capabilities> {
+        const caps = this.serverCapabilitiesService.getCachedCapabilities();
+        if (caps) return caps;
+        return this.serverCapabilitiesService.fetchCapabilities();
+    }
+
+    /**
      * Gets the cached capabilities of the homeserver. If none have been fetched yet,
      * return undefined.
      *
      * @returns The capabilities of the homeserver
      */
     public getCachedCapabilities(): Capabilities | undefined {
-        return this.serverCapabilitiesService.getCapabilities();
+        return this.serverCapabilitiesService.getCachedCapabilities();
     }
 
     /**
--- a/src/client.ts
+++ b/src/client.ts
@@ -226,6 +226,7 @@ import { getRelationsThreadFilter } from "./thread-utils";
 import { KnownMembership, Membership } from "./@types/membership";
 import { RoomMessageEventContent, StickerEventContent } from "./@types/events";
 import { ImageInfo } from "./@types/media";
+import { Capabilities, ServerCapabilities } from "./serverCapabilities";
 
 export type Store = IStore;
 
@@ -233,7 +234,6 @@ export type ResetTimelineCallback = (roomId: string) => boolean;
 
 const SCROLLBACK_DELAY_MS = 3000;
 export const CRYPTO_ENABLED: boolean = isCryptoAvailable();
-const CAPABILITIES_CACHE_MS = 21600000; // 6 hours - an arbitrary value
 const TURN_CHECK_INTERVAL = 10 * 60 * 1000; // poll for turn credentials every 10 minutes
 
 export const UNSTABLE_MSC3852_LAST_SEEN_UA = new UnstableValue(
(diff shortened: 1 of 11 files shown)


src/models/room.ts
```diff
@@ -611,8 +612,8 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
      * Resolves to the version the room should be upgraded to.
      */
     public async getRecommendedVersion(): Promise<IRecommendedVersion> {
-        const capabilities = await this.client.getCapabilities();
-        let versionCap = capabilities["m.room_versions"];
+        const capabilities = this.client.getCachedCapabilities();
```
t3chguy, line 615: What if `getRecommendedVersion` is called before the first capabilities response comes down?

src/rendezvous/MSC3906Rendezvous.ts
```diff
@@ -109,7 +109,7 @@ export class MSC3906Rendezvous {
         logger.info(`Connected to secure channel with checksum: ${checksum} our intent is ${this.ourIntent}`);
 
         // in stable and unstable r1 the availability is exposed as a capability
-        const capabilities = await this.client.getCapabilities();
+        const capabilities = this.client.getCachedCapabilities();
```
t3chguy, line 112: same as above

src/serverCapabilities.ts
```diff
@@ -0,0 +1,136 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { IHttpOpts, MatrixHttpApi, Method } from "./http-api";
+import { logger } from "./logger";
+
+// How often we update the server capabilities.
+// // 6 hours - an arbitrary value, but they should change very infrequently.
```
t3chguy, line 21: double comment

t3chguy (commented):
Should we maybe have an emitter if you wanted to watch for changes, like we do for well-known polling?

dbkr:
Perhaps, but I don't think it would be useful to us right now, particularly? Especially given it will only update every 6 hours, normally. I'd be inclined to keep any extra complexity out until its needed.

t3chguy (approved):

src/serverCapabilities.ts
```diff
@@ -0,0 +1,136 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { IHttpOpts, MatrixHttpApi, Method } from "./http-api";
+import { logger } from "./logger";
+
+// How often we update the server capabilities.
+// 6 hours - an arbitrary value, but they should change very infrequently.
```
BillCarsonFr, line 21: I thought it was 8 hours.
Android is doing 8 saying to match web
https://github.com/element-hq/element-android/blob/097d3923aa03b44c02888cc94a197d34e5e5f780/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/homeserver/GetHomeServerCapabilitiesTask.kt#L192

I supposed this is not spec'ed anywhere?

src/client.ts
```diff
@@ -2095,47 +2065,35 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
-            .then((r = {}) => {
-                const capabilities = r["capabilities"] || {};
-
-                // If the capabilities missed the cache, cache it for a shorter amount
-                // of time to try and refresh them later.
-                const cacheMs = Object.keys(capabilities).length ? CAPABILITIES_CACHE_MS : 60000 + Math.random() * 5000;
+    public async getCapabilities(): Promise<Capabilities> {
+        const caps = this.serverCapabilitiesService.getCachedCapabilities();
+        if (caps) return caps;
+        return this.serverCapabilitiesService.fetchCapabilities();
```
BillCarsonFr, line 2076: So it doesn't swallow errors anymore? only rendez-vous was calling it?

src/models/room.ts
```diff
@@ -636,7 +640,7 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
                     "to be supporting a newer room version we don't know about.",
             );
 
-            const caps = await this.client.getCapabilities(true);
+            const caps = await this.client.fetchCapabilities();
```
BillCarsonFr, line 643: this would throw here now? instead of returning a default value? Would impact the react-sdk?

BillCarsonFr (approved):
Just a few comments

src/serverCapabilities.ts
dbkr: Whoever wrote the js-sdk part picked 6 I guess. It's not specced and it doesn't really matter, I don't think there are any requirements on clients caching it at all.

src/client.ts
dbkr: Yeah, swallowing errors silently and returning something seems like an awful behaviour, so I've changed it favour of having the places its used catch the exception instead, and if they choose to ignore it then that's up to them. It was used in a couple of places, not that many.

src/models/room.ts
dbkr: yeah, I forgot to catch the exception here actually.

PR #4247, opened 2024-06-17
Add via parameter for MSC4156

Johennes:
This is the client-side implementation for https://github.com/matrix-org/matrix-spec-proposals/pull/4156.

## Checklist

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -4319,16 +4319,12 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             signPromise = this.http.requestOtherUrl<IThirdPartySigned>(Method.Post, url);
         }
 
-        const queryParams: QueryDict = {};
+        let queryParams: QueryDict = {};
         if (opts.viaServers) {
             queryParams.server_name = opts.viaServers;
-            switch (this.canSupport.get(Feature.MigrateServerNameToVia)) {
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -4319,9 +4319,17 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
-            queryString["server_name"] = opts.viaServers;
+            queryParams.server_name = opts.viaServers;
+            switch (this.canSupport.get(Feature.MigrateServerNameToVia)) {
+                case ServerSupport.Stable:
+                    queryParams.via = opts.viaServers;
+                    break;
+                case ServerSupport.Unstable:
+                    queryParams["org.matrix.msc4156.via"] = opts.viaServers;
+                    break;
+            }
```
Johennes, line 4332: This feels quite verbose. I wanted to use `replaceParam` but I think it doesn't work in this case because we probably don't want to send the `via` parameter if there is no server support.

t3chguy (commented):
Can we not just be lazy

> or always include both parameters

Johennes:
> Can we not just be lazy
> 
> > or always include both parameters

I thought I wouldn't pass review with that. 😅

But yes, if you're ok with this, it'll make things a bit less verbose.

t3chguy (approved):

PR #4252, opened 2024-06-18
Correctly transform base64 with multiple instances of + or /

robintown:
`String.replace` only replaces a single instance of the search pattern by default; we need a regex in `g` mode if we want to replace them all.

For https://github.com/element-hq/element-call/issues/2429

### DIFF
--- a/spec/unit/base64.spec.ts
+++ b/spec/unit/base64.spec.ts
@@ -50,17 +50,18 @@ describe.each(["browser", "node"])("Base64 encoding (%s)", (env) => {
     });
 
     it("Should encode unpadded URL-safe base64", () => {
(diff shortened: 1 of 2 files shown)


florianduros (approved):

dbkr:
🤦 Good spot

dbkr (approved):

t3chguy:
There's also `replaceAll` nowadays https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll

robintown:
I saw that, it would've required me to bump our TypeScript ES target though so I decided not to bother

PR #4259, opened 2024-06-19
Add fetching the well known in embedded mode.

toger5:
This is used to load the focus from the well known in elment-call.

## Checklist

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -7410,7 +7410,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         return this.http.authedRequest(Method.Post, path, undefined, undefined, { prefix: "" });
     }
 
-    private async fetchClientWellKnown(): Promise<void> {
+    public async fetchClientWellKnown(): Promise<void> {
         // `getRawClientConfig` does not throw or reject on network errors, instead
         // it absorbs errors and returns `{}`.
         this.clientWellKnownPromise = AutoDiscovery.getRawClientConfig(this.getDomain() ?? undefined);
--- a/src/client.ts
+++ b/src/client.ts
@@ -7410,7 +7410,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         return this.http.authedRequest(Method.Post, path, undefined, undefined, { prefix: "" });
     }
 
-    public async fetchClientWellKnown(): Promise<void> {
+    protected async fetchClientWellKnown(): Promise<void> {
         // `getRawClientConfig` does not throw or reject on network errors, instead
         // it absorbs errors and returns `{}`.
         this.clientWellKnownPromise = AutoDiscovery.getRawClientConfig(this.getDomain() ?? undefined);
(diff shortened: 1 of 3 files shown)


src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -825,7 +825,8 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
 
         const myCallMemberEvent = roomState.getStateEvents(EventType.GroupCallMemberPrefix, localUserId) ?? undefined;
         const content = myCallMemberEvent?.getContent() ?? {};
-        const legacy = "memberships" in content || this.useLegacyMemberEvents;
+        const legacy =
+            (("memberships" in content && content.membership?.length) ?? 0 > 0) || this.useLegacyMemberEvents;
```
AndrewFerr, line 829: Is the second `membership` meant to be `memberships` of a `*.call.member` event?
```suggestion
            (("memberships" in content && content.memberships?.length) ?? 0 > 0) || this.useLegacyMemberEvents;
```

toger5: of course!! thx

dbkr:
Unassigning myself for now as it looks like this is still in-progress and getting (force) pushes.

src/matrixrtc/MatrixRTCSession.ts
AndrewFerr, line 1: IMO the changes on this file should be in their own PRs, as they aren't related to well-known fetching.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -825,7 +825,8 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
 
         const myCallMemberEvent = roomState.getStateEvents(EventType.GroupCallMemberPrefix, localUserId) ?? undefined;
         const content = myCallMemberEvent?.getContent() ?? {};
-        const legacy = "memberships" in content || this.useLegacyMemberEvents;
+        const legacy =
+            (("memberships" in content && content.memberships?.length) ?? 0 > 0) || this.useLegacyMemberEvents;
```
AndrewFerr, line 829: This will fallback to the legacy format only if the _current user's_ `*.call.member` state event uses the legacy format.  Should it fallback if _any_ user's `*.call.member` state uses the legacy format?

toger5: I was thinking about this as well. But then decided that we can make the transition a bit more abrupt and only added this usecase to save time. I saw your PR however and like it more then this. So I will remove this change from this pr.

src/matrixrtc/MatrixRTCSession.ts
toger5: moved them out of here. I saw you have a PR covering those. Commented on this pr.

toger5:
> Unassigning myself for now as it looks like this is still in-progress and getting (force) pushes.

sorry for the force push.

Now it is ready to review.

We removed unrelated changes.

src/client.ts
```diff
@@ -7410,7 +7410,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         return this.http.authedRequest(Method.Post, path, undefined, undefined, { prefix: "" });
     }
 
-    private async fetchClientWellKnown(): Promise<void> {
+    public async fetchClientWellKnown(): Promise<void> {
```
AndrewFerr, line 7413: How about `protected`?  That would avoid altering the public API of `MatrixClient`.
```suggestion
    protected async fetchClientWellKnown(): Promise<void> {
```

AndrewFerr (approved):

dbkr (approved):

PR #4260, opened 2024-06-20
Use legacy call membership if anyone else is

AndrewFerr:

## Checklist

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -885,7 +885,7 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
         }
     }
 
-    private stateEventsContainOngoingLegacySession(callMemberEvents: Map<string, MatrixEvent>) {
+    private stateEventsContainOngoingLegacySession(callMemberEvents: Map<string, MatrixEvent>): boolean {
         for (const callMemberEvent of callMemberEvents.values()) {
             const content = callMemberEvent.getContent();
             if (Array.isArray(content["memberships"])) {
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -824,16 +824,9 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
         if (!localUserId || !localDeviceId) throw new Error("User ID or device ID was null!");
 
         const callMemberEvents = roomState.events.get(EventType.GroupCallMemberPrefix);
-        let legacy = !!this.useLegacyMemberEvents;
-        if (!legacy && callMemberEvents?.size) {
-            for (const callMemberEvent of callMemberEvents.values()) {
-                const content = callMemberEvent.getContent();
-                if (Array.isArray(content["memberships"]) && content["memberships"].length > 0) {
-                    legacy = true;
-                    break;
-                }
-            }
-        }
+        const legacy =
+            !!this.useLegacyMemberEvents ||
+            (callMemberEvents?.size && this.stateEventsContainOngoingLegacySession(callMemberEvents));
         let newContent: {} | ExperimentalGroupCallRoomMemberState | SessionMembershipData = {};
         if (legacy) {
             const myCallMemberEvent = callMemberEvents?.get(localUserId);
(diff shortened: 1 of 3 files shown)


src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -823,11 +823,21 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+                    break;
+                }
+            }
+        }
         let newContent: {} | ExperimentalGroupCallRoomMemberState | SessionMembershipData = {};
         if (legacy) {
+            const myCallMemberEvent = callMemberEvents?.get(localUserId);
+            const content = myCallMemberEvent?.getContent() ?? {};
             let myPrevMembership: CallMembership | undefined;
             // We know its CallMembershipDataLegacy
```
toger5, line 842: I think if we already do it properly we also need to check for the expiration of those events.

```suggestion
        const callMemberEvents = roomState.events.get(EventType.GroupCallMemberPrefix);
        const legacy =
            !!this.useLegacyMemberEvents ||
            (callMemberEvents?.size && this.stateEventsContainOngoingLegacySession(callMemberEvents));
        let newContent: {} | ExperimentalGroupCallRoomMemberState | SessionMembershipData = {};
        if (legacy) {
            const myCallMemberEvent = callMemberEvents?.get(localUserId);
            const content = myCallMemberEvent?.getContent() ?? {};
            let myPrevMembership: CallMembership | undefined;
            // We know its CallMembershipDataLegacy
```

and then add:
```
    private stateEventsContainOngoingLegacySession(callMemberEvents: Map<string, MatrixEvent>){
       return Array.from(callMemberEvents.values()).some((memberEvent)=>{
            const content = memberEvent.getContent();
            return Array.isArray(content["memberships"]) &&
            content.memberships.map((m) => new CallMembership(memberEvent, m)).filter((m)=>m.isExpired()).length > 0;
        })
    }
```

toger5:
Can you also add this change:
image

toger5:
Thanks for porting this into its own PR and making it better!

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -876,7 +876,7 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
                 this.room.roomId,
                 EventType.GroupCallMemberPrefix,
                 newContent,
-                this.useLegacyMemberEvents ? localUserId : `${localUserId}_${localDeviceId}`,
+                legacy ? localUserId : `${localUserId}_${localDeviceId}`,
```
AndrewFerr, line 872: I'll make another PR for prefixing the state key with `_`, because it's a subtle yet significant change that's unrelated to the main purpose of this PR.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -892,6 +885,20 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
         }
     }
 
+    private stateEventsContainOngoingLegacySession(callMemberEvents: Map<string, MatrixEvent>) {
+        for (const callMemberEvent of callMemberEvents.values()) {
+            const content = callMemberEvent.getContent();
+            if (Array.isArray(content["memberships"])) {
+                for (const membership of content.memberships) {
+                    if (!(new CallMembership(callMemberEvent, membership).isExpired())) {
+                        return true;
```
AndrewFerr, line 894: Changes from the suggestion:
- This iterates over `.values()` instead of `Array.from` to avoid the extra memory cost of creating a whole array out of an iterator
- This loops over `content.memberships` instead of using `.map` / `.filter` to allow early-returning as soon as a single ongoing session is found
- This negates `.isExpired()` because the goal is to find ongoing/non-expired sessions (right?)

src/matrixrtc/MatrixRTCSession.ts
toger5: perfect even better.

src/matrixrtc/MatrixRTCSession.ts
toger5: Cool. Makes sense.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -882,6 +885,20 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
         }
     }
 
+    private stateEventsContainOngoingLegacySession(callMemberEvents: Map<string, MatrixEvent>): boolean {
+        for (const callMemberEvent of callMemberEvents.values()) {
+            const content = callMemberEvent.getContent();
+            if (Array.isArray(content["memberships"])) {
+                for (const membership of content.memberships) {
+                    if (!new CallMembership(callMemberEvent, membership).isExpired()) {
```
toger5, line 893: I am currently not sure if : `!new Obj().is_sth()` works or if it is preferred to do: `!(new Obj().is_sth())`

toger5 (approved):
This looks good. Thanks.

src/matrixrtc/MatrixRTCSession.ts
AndrewFerr: I wasn't sure either, but the linter forces the parentheses to go away, so it must be fine.

PR #4261, opened 2024-06-20
Upgrade Rust Crypto SDK to 6.1.0

richvdh:
Fixes https://github.com/element-hq/element-web/issues/27590

Notes: fix a bug (https://github.com/element-hq/element-web/issues/27590) which could cause messages sent before migration to Rust Crypto are incorrectly shown as Expected UTDs.


richvdh:
> We don't want to upgrade the version in the package.json too? I guess we want at minimum the 6.1.0 of the sdk crypto

I don't think that's necessary: it's not that our code is incompatible with 6.0.0, it's just that we prefer 6.1.0 because it has bug fixes. Same as any other update of our dependencies like those that come from renovate (eg https://github.com/matrix-org/matrix-js-sdk/pull/4258): we don't normally bump the requirements in package.json.

florianduros (approved):

PR #4262, opened 2024-06-21
Prefix the user+device state key if needed

AndrewFerr:
No need to prefix it for rooms that use MSC3779.
Otherwise, prefix it to bypass the auth rule for state events with keys starting with @.

Signed-off-by: Andrew Ferrazzutti [andrewf@element.io](mailto:andrewf@element.io)

## Checklist


src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -866,7 +866,9 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
                 this.room.roomId,
                 EventType.GroupCallMemberPrefix,
                 newContent,
-                this.useLegacyMemberEvents ? localUserId : `${localUserId}_${localDeviceId}`,
+                this.useLegacyMemberEvents
```
toger5, line 869: this really should use the computed `legacy` prop instead of the config param.

computed here:
```
const legacy = "memberships" in content || this.useLegacyMemberEvents;
```

toger5 (changes requested):

src/matrixrtc/MatrixRTCSession.ts
AndrewFerr: D'oh, of course, thanks.

AndrewFerr:
Test coverage on new code is 100%, so barring review changes, this PR is ready to go.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -899,6 +899,15 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
         return false;
     }
 
+    private makeMembershipStateKey(localUserId: string, localDeviceId: string): string {
+        const stateKey = `${localUserId}_${localDeviceId}`;
+        if (/^org\.matrix\.msc3779\b/.exec(this.room.getVersion())) {
+            return stateKey;
+        } else {
+            return `_${stateKey}`;
+        }
```
toger5, line 908: Ah nice preparing for the owned state events is a smart idea.

toger5 (approved):

src/matrixrtc/MatrixRTCSession.ts
toger5: Just to give background on this: We never use the state key to actually compute the state events.
Which is somewhat broken. In theory a proper state event has to use a spec conform key.
With our implementation this is not checked (we could in theory send multiple member events for one user for the same device by using random state keys.)
This is what allows to only update the part where we compute the membership.

We probably want to check if the state key is to spec before parsing a state event.

toger5: Added an issue here to track this: https://github.com/matrix-org/matrix-js-sdk/issues/4265

PR #4263, opened 2024-06-21
Element-R: Fix resource leaks in verification logic

richvdh:
Fixes a couple of reference cycles which could cause objects to be leaked, ultimately meaning that we hold open a connection to IndexedDB, exacerbating https://github.com/vector-im/element-web/issues/25779.

Requires #4264 

Should be reviewable commit-by-commit.


richvdh:
I'm going to split this into two PRs

andybalaam (approved):
Looks good to me, nice spot.

PR #4283, opened 2024-06-26
Rename `crypto-api.ts` -> `crypto-api/index.ts`

richvdh:
I found it quite confusing having `CryptoApi` be defined so far from the `crypto-api` folder.

### DIFF
--- a/src/crypto-api/index.ts
+++ b/src/crypto-api/index.ts
@@ -15,15 +15,15 @@ limitations under the License.
 */
 
 import type { SecretsBundle } from "@matrix-org/matrix-sdk-crypto-wasm";
-import type { IMegolmSessionData } from "./@types/crypto";
-import { Room } from "./models/room";
-import { DeviceMap } from "./models/device";
-import { UIAuthCallback } from "./interactive-auth";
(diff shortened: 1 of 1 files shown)


t3chguy (approved):
LGTM but I think its a breaking change

richvdh:
> LGTM but I think its a breaking change

Shouldn't be? `import { foo } from "matrix-js-sdk/src/crypto-api";` should still work?

t3chguy:
> Shouldn't be? import { foo } from "matrix-js-sdk/src/crypto-api"; should still work?

Indeed, but `import { foo } from "matrix-js-sdk/lib/crypto-api.js;` would not. I believe ESM consumers have to import files like this

richvdh:
FML. Really?

It really feels like we need better tests for this stuff, if we're going to try and support it.

Maybe this isn't worthwhile, then.

richvdh:
It looks like ESM is borked anyway: https://github.com/matrix-org/matrix-js-sdk/issues/4287

richvdh:
... on which basis, I'm merging this without an X-Breaking-Change.

PR #4288, opened 2024-06-27
Rust crypto: Clean up handling of verification events

richvdh:
We had both an `onIncomingKeyVerificationRequest` and an `onKeyVerificationRequest` which did different, but related, things.

Improve the documentation and reduce the duplication.


kegsay (approved):
LGTM!

andybalaam (approved):

PR #4293, opened 2024-06-28
Bump node.js requirement to 20.

richvdh:
According to https://github.com/matrix-org/matrix-js-sdk?tab=readme-ov-file#supported-platforms, we *only* support the latest LTS release (which is currently 20, per https://github.com/nodejs/release#release-schedule), so this should be safe.

### DIFF
--- a/.github/workflows/release-make.yml
+++ b/.github/workflows/release-make.yml
@@ -116,6 +116,7 @@ jobs:
             - uses: actions/setup-node@v4
               with:
                   cache: "yarn"
+                  node-version-file: package.json
 
             - name: Install dependencies
               run: "yarn install --frozen-lockfile"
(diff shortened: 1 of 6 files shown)


t3chguy (approved):
I will tweak the branch protections for this

t3chguy:
@richvdh looks like this will need downstream changes first, as that CI still runs under Node 18

richvdh:
Hopefully https://github.com/matrix-org/matrix-react-sdk/pull/12726 will fix the downstream CI

richvdh:
... and https://github.com/matrix-org/complement-crypto/pull/102

richvdh:
Oh come *on*. Test flake due to https://github.com/matrix-org/complement-crypto/issues/9

PR #4294, opened 2024-06-28
Support MSC4140: Delayed events (Futures)

AndrewFerr:

## Checklist

### DIFF
--- a/spec/unit/matrix-client.spec.ts
+++ b/spec/unit/matrix-client.spec.ts
@@ -57,6 +57,7 @@ import {
     Room,
     RuleId,
     TweakName,
+    UpdateDelayedEventAction,
 } from "../../src";
 import { supportsMatrixCall } from "../../src/webrtc/call";
 import { makeBeaconEvent } from "../test-utils/beacon";
@@ -97,7 +98,7 @@ type HttpLookup = {
     method: string;
     path: string;
     prefix?: string;
-    data?: Record<string, any>;
+    data?: Record<string, any> | Record<string, any>[];
     error?: object;
     expectBody?: Record<string, any>;
     expectQueryParams?: QueryDict;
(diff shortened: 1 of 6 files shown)


src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -864,18 +866,35 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
-            if (this.isJoined() && legacy) {
-                this.memberEventTimeout = setTimeout(this.triggerCallMembershipEventUpdate, MEMBER_EVENT_CHECK_PERIOD);
+            if (this.isJoined()) {
+                if (legacy) {
+                    this.memberEventTimeout = setTimeout(
+                        this.triggerCallMembershipEventUpdate,
+                        MEMBER_EVENT_CHECK_PERIOD,
+                    );
+                } else {
+                    const res = await this.client._unstable_sendStateFuture(
```
toger5, line 882: This needs its own dedicated try catch block.
Currently if we fail to send the future (synapse not supporting it) we will retry sending the successflully sent non future join state event.

toger5:
```
            logger.warn(`Failed to send call member event: retrying in ${resendDelay}\n reason: ${e}`);
```
adding the error to the fail warning also deos not hurt i think.

src/client.ts
```diff
@@ -4608,7 +4641,38 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+        futureOpts: F,
+        txnId?: string,
+    ): Promise<ISendFutureResponse<F>>;
+    private sendCompleteEvent<F extends ISendFutureRequestOpts>(
+        roomId: string,
+        threadId: string | null,
+        eventObject: Partial<IEvent>,
+        futureOptsOrTxnId?: F | string,
+        txnIdOrVoid?: string,
+    ): Promise<ISendEventResponse | ISendFutureResponse<F>> {
```
toger5, line 4666: since this is a private member I am wondering if we can use named fields (prop obj).
```
private sendCompleteEvent(sendData: SendCompleteEventData | SendCompleteFutureEventData)
```
and then specify the the two send data interfaces.
Would also match the return Promise type.
And can be type checked easier.

But it looks like the `...Or...` naming pattern is used at other places as well. So from that point of view this should be fine.

src/client.ts
```diff
@@ -4632,6 +4696,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         }
 
         // set up re-emitter for this new event - this is normally the job of EventMapper but we don't use it here
+        // TODO: skip for futures?
```
toger5, line 4699: any particular reason we don't do this already?

src/client.ts
```diff
@@ -4835,30 +4933,37 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            let pathTemplate = `/rooms/$roomId/state${futurePathPart}/$eventType`;
             if (event.getStateKey() && event.getStateKey()!.length > 0) {
-                pathTemplate = "/rooms/$roomId/state/$eventType/$stateKey";
+                pathTemplate = `/rooms/$roomId/state${futurePathPart}/$eventType/$stateKey`;
             }
             path = utils.encodeUri(pathTemplate, pathParams);
         } else if (event.isRedaction() && event.event.redacts) {
-            const pathTemplate = `/rooms/$roomId/redact/$redactsEventId/$txnId`;
+            // TODO: support future redactions?
+            const pathTemplate = `/rooms/$roomId/redact${futurePathPart}/$redactsEventId/$txnId`;
```
toger5, line 4947: Is this actually what we want/need.
If i understand correctly this is to send a redact that does not happen now but only once the future resolves?
I think we can simply use the send endpoint and choose the event type and the content accordingly to send a redact event as a future.
Spcifying a custom future_readact endpoint maybe should be a seperate MSC since it only would be "alternative syntax"

src/embedded.ts
```diff
@@ -248,7 +248,20 @@ export class RoomWidgetClient extends MatrixClient {
...
+        event: MatrixEvent,
+        futureOpts: F,
+    ): Promise<ISendFutureResponse<F>>;
+    protected async encryptAndSendEvent<F extends ISendFutureRequestOpts>(
+        room: Room,
+        event: MatrixEvent,
+        futureOpts?: F,
+    ): Promise<ISendEventResponse | ISendFutureResponse<F>> {
+        if (futureOpts) {
+            throw new Error("Future sending via widgets is not implemented");
```
toger5, line 263: this is replaced in the widget api impl PR i guess?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -864,27 +866,62 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
-            logger.warn(`Failed to send call member event: retrying in ${resendDelay}`);
+            logger.warn(`Failed to send call member event (retrying in ${resendDelay}): ${e}`);
             await new Promise((resolve) => setTimeout(resolve, resendDelay));
             await this.triggerCallMembershipEventUpdate();
         }
     }
 
+    private scheduleRefreshFuture(refreshToken: string): void {
+        // TODO: choose a reasonable timeout
+        this.memberEventTimeout = setTimeout(() => this.refreshFuture(refreshToken), 3000);
```
toger5, line 915: So the heartbeat ping is every 3 seconds and the state event timeout 10seconds?
I think 2x should be fine. 5 and 10 seconds or 3 and 6 seconds.

Do you have any "choosing mechanism" in mind?

src/client.ts
```diff
@@ -5188,6 +5293,127 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+    }
+
+    /**
+     * Use a future token, taking the appropriate action given what the token is for.
+     *
+     * Note: This endpoint is unstable, and can throw an `Error`.
+     *   Check progress on [MSC4140](https://github.com/matrix-org/matrix-spec-proposals/pull/4140) for more details.
+     */
+    // eslint-disable-next-line
+    public async _unstable_useFutureToken(futureToken: string): Promise<{}> {
```
toger5, line 5401: `use` somehow implies the consequence of the token being invalid after use.
Which is not the case. What about `send`/`pingWith` or `updateFuture`/`updateFutureWithToken`.

src/client.ts
AndrewFerr: The `/redact/` endpoint has different semantics than `/send/` (eg. it targets a specific eventID, provides a reason for the redaction, and abstracts what data is even removed from the event), so AIUI it's different enough to warrant another future-compatible endpoint.  Though I don't think we need it right now so I'm satisfied with just throwing a "not implemented" error here.

src/client.ts
AndrewFerr: Just a lack of testing.  I'll skip it now & will check that it doesn't break anything.

src/client.ts
AndrewFerr: Yes, that naming pattern is already used elsewhere in this file & I found it to be a nice way to write overloads while getting usable type hints when calling them.

src/@types/requests.ts
```diff
@@ -76,6 +76,47 @@ export interface ISendEventResponse {
...
+export interface IFutureTimeout {
+    future_timeout: number;
+}
+
+export type ISendTimeoutFutureRequestOpts = IFutureTimeout & Partial<IFutureGroupId>;
+export type ISendActionFutureRequestOpts = IFutureGroupId;
+
+export type ISendFutureRequestOpts = ISendTimeoutFutureRequestOpts | ISendActionFutureRequestOpts;
+
+export interface IFutureTokens {
```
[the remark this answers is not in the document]
AndrewFerr: Sure, I'll change this.  I named it like this to mark it as a "base" type of sorts, but that's less useful than having types named after action/timeout types like the rest of them.

src/embedded.ts
AndrewFerr: Yes it is!

src/matrixrtc/MatrixRTCSession.ts
AndrewFerr: I don't have any particular mechanism in mind, other than not wanting to spam network requests (but to still send enough to be useful :slightly_smiling_face: )

src/client.ts
AndrewFerr: `updateFutureWithToken` is the most descriptive, so I'll go with that.

src/@types/requests.ts
```diff
@@ -76,6 +76,47 @@ export interface ISendEventResponse {
     event_id: string;
 }
 
+export interface IFutureGroupId {
```
t3chguy, line 79: See code style, we don't use `I` on new interfaces

src/client.ts
```diff
@@ -5188,6 +5293,127 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+    ): Promise<ISendFutureResponse<F>>;
+    // eslint-disable-next-line
+    public async _unstable_sendFuture<F extends ISendFutureRequestOpts>(
+        roomId: string,
+        futureOpts: F,
+        threadIdOrEventType: string | null,
+        eventTypeOrContent: string | IContent,
+        contentOrTxnId?: IContent | string,
+        txnIdOrVoid?: string,
+    ): Promise<ISendFutureResponse<F>> {
```
t3chguy, line 5327: Could we please avoid the `threadId` overload, just have the caller specify `null` for "Main" timeline

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -864,27 +866,62 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+                            },
+                            EventType.GroupCallMemberPrefix,
+                            {}, // leave event
+                            stateKey,
+                        )
+                        .then((res) => {
+                            this.cancelFutureToken = res.cancel_token;
+                            this.scheduleRefreshFuture(res.refresh_token);
+                        })
+                        .catch((e) => logger.error("Failed to send future:", e));
```
t3chguy, line 897: Can we use async-await instead of then/catch please

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -864,27 +866,62 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+                }
+            } else if (!legacy && this.cancelFutureToken !== undefined) {
+                // TODO: remove this once the Synapse MSC4140 implementation cancels timeout futures on state change
+                await this.client
+                    ._unstable_useFutureToken(this.cancelFutureToken)
+                    .catch((e) => logger.error("Failed to cancel future:", e));
```
t3chguy, line 903: Same as above

src/matrixrtc/MatrixRTCSession.ts
toger5: then lets do 5 8 as a starter. As an experiment if this already causes calls to be ended  accidentally by bad network because there is only a 3 second buffer between the timeout and the refresh?

src/client.ts
toger5: O just played around and tried sending a custom timeline events and it seems it does not work right now. there was a thread somewhere discussing this and iirc it was not far from making this work. That would help that we done need a custom redact future endpoint but instead do sth like this:
image

toger5 (approved):
Thanks for also adding the tests.
Looks good to me now.

AndrewFerr:
I finally got the code coverage of MatrixRTCSession.ts that I was after, so barring further review, I'm done with pushing more commits here.

t3chguy (approved):

spec/unit/matrix-client.spec.ts
```diff
@@ -704,6 +705,328 @@ describe("MatrixClient", function () {
...
+                roomId,
+                { parent_delay_id: timeoutDelayId },
+                threadId,
+                EventType.RoomMessage,
+                { ...content },
+                actionDelayTxnId,
+            );
+        });
+
+        it("should add thread relation if threadId is passed and the relation is missing", async () => {
```
toger5, line 838: why do we need the thread related tests?

src/@types/requests.ts
```diff
@@ -76,6 +76,56 @@ export interface ISendEventResponse {
...
+
+export type SendDelayedEventRequestOpts = SendTimeoutDelayedEventRequestOpts | SendActionDelayedEventRequestOpts;
+
+export type SendDelayedEventResponse = {
+    delay_id: string;
+};
+
+export enum UpdateDelayedEventAction {
+    Cancel = "cancel",
+    Refresh = "refresh",
```
toger5: restart
```suggestion
    Restart = "restart",
```

src/client.ts
```diff
@@ -4648,29 +4692,56 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
 
     /**
      * encrypts the event if necessary; adds the event to the queue, or sends it; marks the event as sent/unsent
      * @returns returns a promise which resolves with the result of the send request
      */
-    protected async encryptAndSendEvent(room: Room | null, event: MatrixEvent): Promise<ISendEventResponse> {
+    protected async encryptAndSendEvent(room: Room | null, event: MatrixEvent): Promise<ISendEventResponse>;
+    /**
+     * Simply sends a delayed event without encrypting it.
+     * TODO: Allow encrypted delayed events, and encrypt them properly
```
toger5, line 4727: does this mean we cannot send delayed encrypted event as of now?

toger5 (approved):
This looks very good thanks.
I think, i cannot follow why delayed events are entangled with threads (is it just because of the method overloading?

Reading through this code again I am very surpriced that this does not use a props interface param for the function paramters
```
myOverloadedFunction(props: {delay?: number, threadId?: string, eventType: EventType ...})
```

spec/unit/matrix-client.spec.ts
AndrewFerr: This is the same as the thread tests for non-delayed events & makes sure that thread handling works properly for delayed events too.

src/client.ts
AndrewFerr: Correct

AndrewFerr:
Since this PR is already reviewed, I'll squash & rebase its commits before merging, because
1) all of the sub-commits are too granular to keep in history
2) the latest `develop` fixes one of the tests that are failing on this PR right now

PR #4302, opened 2024-07-08
Element-R: detect "withheld key" UTD errors, and mark them as such

richvdh:
Partial fix to https://github.com/element-hq/element-web/issues/27653


t3chguy (approved):

src/rust-crypto/rust-crypto.ts
```diff
@@ -1789,6 +1789,13 @@ class EventDecryptor {
 
         switch (err.code) {
             case RustSdkCryptoJs.DecryptionErrorCode.MissingRoomKey:
+                if (err.maybe_withheld) {
```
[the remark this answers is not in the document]
richvdh: I'm working on updating `DecryptionFailureTracker`. But yes, it does look like we have `MatrixEvent.isEncryptedDisabledForUnverifiedDevices` which for some reason looks at the text of the error code.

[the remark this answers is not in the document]
richvdh: It looks like we actually emit the same error text as the legacy stack, so `isEncryptedDisabledForUnverifiedDevices` works fine for now. Have added a test to make sure it stays that way.

I'm not aware of anything in the UI using the message directly.

src/crypto/OlmDevice.ts
```diff
@@ -1222,7 +1222,9 @@ export class OlmDevice {
                     if (session === null || sessionData === null) {
                         if (withheld) {
                             error = new DecryptionError(
-                                DecryptionFailureCode.MEGOLM_UNKNOWN_INBOUND_SESSION_ID,
+                                withheld.code === "m.unverified"
```
uhoreg, line 1225: I think it would be clearer if the code was calculated outside the `DecryptionError(...)`, as in:

```
const failureCode = withheld.code === "m.unverified" ? ...
error = new DecryptionError(
  failureCode,
  ...
)
```

(and similarly below), but YMMV

src/rust-crypto/rust-crypto.ts
```diff
@@ -1787,6 +1787,17 @@ class EventDecryptor {
             }
         }
 
+        // If we got a withheld code, expose that.
+        if (err.maybe_withheld) {
+            // Unfortunately the Rust SDK API doesn't let us distinguish between different withheld cases, other than
```
BillCarsonFr, line 1792: Oh wow, this is because we are doing `to_string()` that uses `fmt::Display` instead of `as_str` that for ruma enum would give the code?

BillCarsonFr (approved):

src/rust-crypto/rust-crypto.ts
BillCarsonFr: So probably here https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/blob/a77375c0212a79a8650e81533c43801e559fff1e/src/error.rs#L57

Should be `as_str()` ?
I don't think it's intended that it's not the code here

src/rust-crypto/rust-crypto.ts
```diff
@@ -1787,6 +1787,17 @@ class EventDecryptor {
...
+            // by string-matching.
```
uhoreg, line 1793: :frowning_face:

uhoreg (approved):

src/rust-crypto/rust-crypto.ts
richvdh: Yes, it's because the wasm bindings do this:

```
                maybe_withheld: withheld_code
                    .as_ref()
                    .map(|code| code.to_string().to_owned().into()),
```

richvdh: >  I don't think it's intended that it's not the code here

Guess there's a lesson in testing your code here!

PR #4311, opened 2024-07-11
Support Futures via widgets

AndrewFerr:

Dependencies:
- https://github.com/matrix-org/matrix-js-sdk/pull/4294
- https://github.com/matrix-org/matrix-widget-api/pull/92

## Checklist

### DIFF
--- a/spec/integ/rendezvous/MSC4108SignInWithQR.spec.ts
+++ b/spec/integ/rendezvous/MSC4108SignInWithQR.spec.ts
@@ -268,7 +268,8 @@ describe("MSC4108SignInWithQR", () => {
         it("should abort if device doesn't come up by timeout", async () => {
             jest.spyOn(global, "setTimeout").mockImplementation((fn) => {
                 (<Function>fn)();
-                return -1;
+                // TODO: mock timers properly
+                return -1 as any;
             });
             jest.spyOn(Date, "now").mockImplementation(() => {
                 return 12345678 + mocked(setTimeout).mock.calls.length * 1000;
@@ -320,7 +321,8 @@ describe("MSC4108SignInWithQR", () => {
         it("should not send secrets if user cancels", async () => {
             jest.spyOn(global, "setTimeout").mockImplementation((fn) => {
                 (<Function>fn)();
-                return -1;
+                // TODO: mock timers properly
+                return -1 as any;
             });
 
             await Promise.all([ourLogin.negotiateProtocols(), opponentLogin.negotiateProtocols()]);
--- a/src/embedded.ts
+++ b/src/embedded.ts
@@ -26,8 +26,13 @@ import {
 } from "matrix-widget-api";
 
 import { MatrixEvent, IEvent, IContent, EventStatus } from "./models/event";
-import { ISendEventResponse, SendDelayedEventRequestOpts, SendDelayedEventResponse } from "./@types/requests";
-import { EventType } from "./@types/event";
+import {
+    ISendEventResponse,
+    SendDelayedEventRequestOpts,
+    SendDelayedEventResponse,
+    UpdateDelayedEventAction,
+} from "./@types/requests";
+import { EventType, StateEvents } from "./@types/event";
 import { logger } from "./logger";
 import {
     MatrixClient,
@@ -36,6 +41,7 @@ import {
     IStartClientOpts,
     SendToDeviceContentMap,
     IOpenIDToken,
+    UNSTABLE_MSC4140_DELAYED_EVENTS,
 } from "./client";
 import { SyncApi, SyncState } from "./sync";
 import { SlidingSyncSdk } from "./sliding-sync-sdk";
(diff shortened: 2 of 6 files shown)


src/embedded.ts
```diff
@@ -280,7 +299,80 @@ export class RoomWidgetClient extends MatrixClient {
...
+     * @experimental This currently relies on an unstable MSC (MSC4140).
+     */
+    // eslint-disable-next-line
+    public async _unstable_sendStateFuture<K extends keyof StateEvents, F extends SendFutureRequestOpts>(
+        roomId: string,
+        futureOpts: F,
+        eventType: K,
+        content: StateEvents[K],
+        stateKey: string,
+    ): Promise<SendFutureResponse<F>>;
```
t3chguy, line 317: This overload seems to not add anything?

AndrewFerr: This type signature is needed to properly override the method from the parent class, and the one following this is needed to avoid some typechecking errors, as type narrowing of return types is finicky when conditional parameter types are involved.

But I can avoid using an overload if I simply cast the return value, so I'll go with that.

AndrewFerr:
Putting this back into draft until the widget API is up-to-speed with the recent revisions to MSC4140.

src/embedded.ts
```diff
@@ -280,7 +290,48 @@ export class RoomWidgetClient extends MatrixClient {
...
+        if (!(await this.doesServerSupportUnstableFeature(UNSTABLE_MSC4140_DELAYED_EVENTS))) {
+            throw Error("Server does not support the delayed events API");
+        }
+
+        const response = await this.widgetApi.sendStateEvent(
+            eventType,
+            stateKey,
+            content,
+            roomId,
+            "delay" in delayOpts ? delayOpts.delay : undefined,
```
toger5, line 317: the linter is yelling here. Just because a depency is not updated or do we pass the wrong args here?

toger5 (approved):

src/embedded.ts
AndrewFerr: It's because of a missing dependency on the updated widget-api.  The linter passes for me locally when `yarn link`ing to the PR-equipped widget-api (other than some errors in an unrelated regtest, which I assume will go away after rebasing onto the latest `develop`...)

AndrewFerr:
I'll take this out of draft once the [dependent widget-api PR](https://github.com/matrix-org/matrix-widget-api/pull/92) is merged.

AndrewFerr:
> I'll take this out of draft once the [dependent widget-api PR](https://github.com/matrix-org/matrix-widget-api/pull/92) is merged.

Not quite -- it would be nice if the widget-api's [release workflow could be fixed](https://github.com/matrix-org/matrix-widget-api/pull/94) before cutting a new release for it, which this PR will depend on.

src/embedded.ts
AndrewFerr: This is indeed fixed now.

AndrewFerr:
After the rebasing, the effective changes since the last review start from 3c65057b.

src/embedded.ts
```diff
@@ -162,6 +182,18 @@ export class RoomWidgetClient extends MatrixClient {
         );
         capabilities.sendToDevice?.forEach((eventType) => widgetApi.requestCapabilityToSendToDevice(eventType));
         capabilities.receiveToDevice?.forEach((eventType) => widgetApi.requestCapabilityToReceiveToDevice(eventType));
+        if (
+            capabilities.sendDelayedEvents &&
+            (capabilities.sendEvent?.length ||
+                capabilities.sendMessage === true ||
+                (Array.isArray(capabilities.sendMessage) && capabilities.sendMessage.length) ||
+                capabilities.sendState?.length)
```
toger5, line 190: Good job on realizing, that requesting this is unnecassary if we dont send any state or message.

toger5 (approved):
Looks very good

src/embedded.ts
```diff
@@ -280,7 +322,56 @@ export class RoomWidgetClient extends MatrixClient {
...
+    public async _unstable_sendDelayedStateEvent<K extends keyof StateEvents>(
+        roomId: string,
+        delayOpts: SendDelayedEventRequestOpts,
+        eventType: K,
+        content: StateEvents[K],
+        stateKey = "",
+    ): Promise<SendDelayedEventResponse> {
+        if (!(await this.doesServerSupportUnstableFeature(UNSTABLE_MSC4140_DELAYED_EVENTS))) {
+            throw Error("Server does not support the delayed events API");
+        }
```
robintown, line 345: It feels a little surprising that the host client might approve the capability without having server support. (I would expect this check to be the host's responsibility, ideally)

robintown (approved):
One non-blocking comment:

spec/integ/rendezvous/MSC4108SignInWithQR.spec.ts
```diff
@@ -268,7 +268,8 @@ describe("MSC4108SignInWithQR", () => {
         it("should abort if device doesn't come up by timeout", async () => {
             jest.spyOn(global, "setTimeout").mockImplementation((fn) => {
                 (<Function>fn)();
-                return -1;
+                // TODO: mock timers properly
+                return -1 as any;
```
t3chguy, line 272: Why are these needed now but weren't before?

AndrewFerr: I'm not actually sure...it must be due to a dependency update.  Typechecks expect the return value to be a [NodeJS Timeout](https://nodejs.org/api/timers.html#class-timeout) instead of a timeout ID.

t3chguy (approved):

PR #4315, opened 2024-07-16
Remove the compare function from utils

dbkr:
and change the one use of it to just intantiate a collator and use it.

This was marked as internal module so this shouldn't be a breaking change. Of course, react-sdk was using it.

Requires: https://github.com/matrix-org/matrix-react-sdk/pull/12782

## Checklist

### DIFF
--- a/spec/unit/room.spec.ts
+++ b/spec/unit/room.spec.ts
@@ -1199,6 +1199,20 @@ describe("Room", function () {
 
                 expect(spy).toHaveBeenCalledWith(summary);
             });
+
+            it("recalculates in acceptable time without heroes", function () {
+                for (let i = 0; i < 10000; i++) {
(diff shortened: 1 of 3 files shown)


t3chguy (changes requested):
The singular collator choice was done to fix a performance issue, please prove that this does not regress that via some performance test comparisons

See https://github.com/matrix-org/matrix-js-sdk/commit/f708c473851093a149de669e49e67ead6d18e553

dbkr:
New test passes in somewhere around 80ms on my machine, same as with the old code, so instantiating another collator for each sort operation doesn't appear to materially affect things.

t3chguy:
@dbkr worth testing in browsers, given that is where the performance issue was

t3chguy (changes requested):
See comment

dbkr:
Tested on my big(ish) test account - no measurable difference between develop and this PR.

t3chguy (approved):

PR #4328, opened 2024-07-24
Be specific about what is considered a MSC4143 call member event.

toger5:
This is a solution (I am not sure it can be considered a actual fix. see below for additional thoughts on this) for EW exccessive logging for thirdroom rooms.

We do run checks on m.call.member events to check if the event can be considered a legacy event or a new session type event. This check does not consider that there is also a third content format created by thirdroom.

Hence we need to be more explicit to not consider third room call member events as session type events.

A more proper solution (that also was discussed is to rename the event type `org.matrix.msc3401.call.member`->`org.matrix.msc4143.call.member` comes to mind)
This is a much more breaking change however so we first go with this first reduce the logging step is a better first step.

## Checklist

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -169,7 +169,7 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
                     continue;
                 }
                 membershipContents = content["memberships"];
-            } else {
+            } else if ("focus_active" in content) {
                 // We have a MSC4143 event membership event
                 if (Object.keys(content).length !== 0) {
                     // We checked for empty content to not try to construct CallMembership's with {}.
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -160,6 +160,10 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
         const callMemberships: CallMembership[] = [];
         for (const memberEvent of callMemberEvents) {
             const content = memberEvent.getContent();
+
+            // Dont even bother about empty events (saves us from costly type/"key in" checks in bigger rooms)
+            if (Object.keys(content).length === 0) continue;
+
             let membershipContents: any[] = [];
             // We first decide if its a MSC4143 event (per device state key)
             if ("memberships" in content) {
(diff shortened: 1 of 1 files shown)


src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -169,7 +169,7 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
                     continue;
                 }
                 membershipContents = content["memberships"];
-            } else {
+            } else if ("focus_active" in content) {
                 // We have a MSC4143 event membership event
                 if (Object.keys(content).length !== 0) {
                     // We checked for empty content to not try to construct CallMembership's with {}.
```
robintown, line 175: This check seems unnecessary now

robintown (approved):
If CallMembership needs to validate the event content anyways, I wonder if we shouldn't just disable its `logger.warn` line. That feels appropriate given that we've built multiple mutually incompatible event formats on top of the same event type over the years.

However, using a heuristic to cut down on the amount of logging that happens in practice seems OK I guess. I won't block this if you want to keep the warning.

toger5:
The idea is that we first do a high level check to determine which type it should be. And then do a thorough check if we think we know what type it is. The thorough check should only happen if we know its a legacy member event or a session member event. And should only log if the respective event is malformatted: a session event without e.g. `application`

src/matrixrtc/MatrixRTCSession.ts
toger5: 100%

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -160,6 +160,10 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
         const callMemberships: CallMembership[] = [];
         for (const memberEvent of callMemberEvents) {
             const content = memberEvent.getContent();
+
+            // Dont even bother about empty events (saves us from costly type/"key in" checks in bigger rooms)
+            if (Object.keys(content).length === 0) continue;
+
             let membershipContents: any[] = [];
             // We first decide if its a MSC4143 event (per device state key)
```
toger5, line 168: Oh, we are really close to the perf limit here.
removing this lets the room 200ms test fail.
I am not sure what data is used in this test. If its mostly empty call member events that makes sense.
But adding this initial check is okay anyways.

We see if its empty - continue
We check if there is a membership array - add it to the list of "to parse" membership data
We check if its a session format - add it to the list of "to parse" membership data
We check if there is any "to parse" membership data

this seems like a reasonable approach.

toger5: Oh well this code is not even run. the test just seems to be super flaky.

PR #4333, opened 2024-07-30
Fix hashed ID server lookups with no Olm

dbkr:
It used the hash function from Olm (presumably to work cross-platform) but subtle crypto is available on node nowadays so we can just use that.

Refactor existing code that did this out to a common function, add tests.

## Checklist


src/crypto/digest.ts
richvdh, line 1: I'm trying to stop adding stuff to the `crypto` directory, as (a) most of it is stuff that will have to go away come the revolution (b) it's about E2EE, which this isn't (necessarily).

Would it be terrible to put this in `../base64` ?

Or anywhere, really.

src/crypto/digest.ts
```diff
@@ -0,0 +1,32 @@
...
+import { encodeUnpaddedBase64Url } from "../base64";
+
+/**
+ * @param plaintext The string to hash
+ * @returns Unpadded base64-url string representing the sha256 hash of the input
+ */
+export async function sha256Base64UrlUnpadded(plaintext: string): Promise<string> {
+    if (!globalThis.crypto.subtle) {
+        throw new Error("No WebCrypto available");
+    }
```
richvdh, line 26: this behaviour should probably be documented in the doc-comment.

src/crypto/digest.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { encodeUnpaddedBase64Url } from "../base64";
+
+/**
+ * @param plaintext The string to hash
+ * @returns Unpadded base64-url string representing the sha256 hash of the input
+ */
+export async function sha256Base64UrlUnpadded(plaintext: string): Promise<string> {
+    if (!globalThis.crypto.subtle) {
+        throw new Error("No WebCrypto available");
```
richvdh, line 25: this is misleading. WebCrypto is known to be present (it's guaranteed in all the platforms we support) -- it's subtlecrypto that may be missing, in insecure contexts.

src/crypto/digest.ts
dbkr: Ah right. base64 seems sort of confusing, but I can just move it up to the src directory?

src/crypto/digest.ts
dbkr: good point - better?

spec/unit/crypto/digest.spec.ts
richvdh, line 1: move this out of `crypto` too, please.

src/digest.ts
```diff
@@ -0,0 +1,34 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { encodeUnpaddedBase64Url } from "./base64";
+
+/**
+ * @param plaintext The string to hash
+ * @returns Unpadded base64-url string representing the sha256 hash of the input
+ * @throws If the WebCrypto API is not available, for example if the code is running
```
richvdh, line 22: this is the same thing you just fixed in the code...

src/digest.ts
```diff
@@ -0,0 +1,34 @@
...
+ *         in a web page with an insecure context (eg. served over plain HTTP).
+ */
+export async function sha256Base64UrlUnpadded(plaintext: string): Promise<string> {
```
richvdh, line 25: this name is really awkward, not least because "base64 unpadded" is not a thing, whereas "unpadded base64" really is.

Looking at it again though, would it be terrible to move the call to `encodeUnpaddedBase64Url` to the callers, so this is literally just doing the sha256 bit?

dbkr: Yeah, it was mostly because we happened to use *exactly* the same encoding of exactly the same has in another place (albeit deprecated) but I guess it's no more effort to just pass one function into the other.

src/digest.ts
dbkr: 🤦

dbkr:
I don't know why the sonarcloud results aren't coming through now :/

richvdh (approved):

src/digest.ts
```diff
@@ -0,0 +1,32 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * @param plaintext The string to hash
```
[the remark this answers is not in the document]
dbkr: I'd sort of assumed this was described completely by the @return param, but I can add a body too

t3chguy (approved):

PR #4335, opened 2024-08-01
Make code tsc es2022 compliant

t3chguy:
For https://github.com/element-hq/element-web/issues/24913

### DIFF
--- a/src/crypto/algorithms/base.ts
+++ b/src/crypto/algorithms/base.ts
@@ -66,18 +66,16 @@ export abstract class EncryptionAlgorithm {
     protected readonly crypto: Crypto;
     protected readonly olmDevice: OlmDevice;
     protected readonly baseApis: MatrixClient;
-    protected readonly roomId?: string;
 
     /**
      * @param params - parameters
      */
-    public constructor(params: IParams) {
+    public constructor(params: Omit<IParams, "roomId">) {
         this.userId = params.userId;
         this.deviceId = params.deviceId;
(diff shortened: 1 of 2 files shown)


src/crypto/algorithms/base.ts
```diff
@@ -66,18 +66,16 @@ export abstract class EncryptionAlgorithm {
     protected readonly crypto: Crypto;
     protected readonly olmDevice: OlmDevice;
     protected readonly baseApis: MatrixClient;
-    protected readonly roomId?: string;
```
richvdh, line 69: This feels like it's snuck in from a separate PR?

t3chguy: It did not, without it TSC in es2022 mode on downstream layers complains that you cannot override a field definition in a parent class without using `declare` syntax

src/crypto/algorithms/base.ts
```diff
@@ -66,18 +66,16 @@ export abstract class EncryptionAlgorithm {
...
 
     /**
      * @param params - parameters
      */
-    public constructor(params: IParams) {
+    public constructor(params: Omit<IParams, "roomId">) {
```
richvdh, line 73: feels like we should leave `params` alone even if we're removing the property on the object, since this defines the common interface for encryptors. (Likewise for `DecryptionAlgorithm`).

This code is all going away soon though, so \<shrug>

richvdh (approved):

dbkr (approved):

PR #4340, opened 2024-08-07
Don't attempt to adjust for clock skews when calculating group call membership expiry

hughns:
This change proposes that we simplify the client side logic used to calculate when a MatrixRTC call membership has expired.

The current implementation attempts to adjust for differing clocks between the local client, remote clients, the local homeserver and federated/remote homeservers.

The proposed change simplifies the logic to assume that the participants are reasonably close with the local clock being: -infinity to +30mins skewed.

## Checklist

### DIFF
--- a/spec/unit/matrixrtc/MatrixRTCSession.spec.ts
+++ b/spec/unit/matrixrtc/MatrixRTCSession.spec.ts
@@ -841,6 +841,7 @@ describe("MatrixRTCSession", () => {
 
         it("Re-sends key if a member changes created_ts", async () => {
             jest.useFakeTimers();
+            jest.setSystemTime(1000);
             try {
                 const keysSentPromise1 = new Promise((resolve) => {
                     sendEventMock.mockImplementation(resolve);


hughns:
Sorry for the spam, I meant to create this as draft in the first place.

src/matrixrtc/CallMembership.ts
```diff
@@ -184,10 +196,24 @@ export class CallMembership {
         }
     }
 
+    /**
+     * @returns The number of milliseconds until the membership expires or undefined if applicable
+     */
     public getMsUntilExpiry(): number | undefined {
-        if (isLegacyCallMembershipData(this.membershipData)) return this.getLocalExpiry()! - Date.now();
+        if (isLegacyCallMembershipData(this.membershipData)) {
+            // Assume that local clock is sufficiently in sync with other clocks in the distributed system.
```
AndrewFerr, line 204: Perhaps this is notable enough to appear in the docstring...but since it refers to & relies on internal behaviour / non-exported helper values & functions (notably `MatrixRTCSession.membershipEventNeedsUpdate`), it's more of a notice to js-sdk developers than for people using this as a dependency.  So this is fine as is.

src/matrixrtc/CallMembership.ts
```diff
@@ -184,10 +196,24 @@ export class CallMembership {
...
+            // We used to try and adjust for the local clock being skewed, but there are cases where this is not accurate.
+            // The current implementation allows for the local clock to be -infinity to +MatrixRTC.MEMBERSHIP_EXPIRY_TIME/2 (i.e. +30mins).
```
AndrewFerr, line 206:
```suggestion
            // The current implementation allows for the local clock to be -infinity to +MatrixRTCSession.MEMBERSHIP_EXPIRY_TIME/2 (i.e. +30mins).
```
Plus, maybe we should add a comment in `MatrixRTCSession.membershipEventNeedsUpdate` with a reminder to update this comment if the expiry calculation were to change & thus make this comment stale (like if `MEMBERSHIP_EXPIRY_TIME` were to change from 30 mins to some other value).

AndrewFerr (approved):
Just some documentation nitpicks

src/matrixrtc/CallMembership.ts
hughns: I've stripped the `i.e.` instead.

PR #4342, opened 2024-08-08
When state says you've left ongoing call, rejoin

AndrewFerr:
When receiving a state change that says you are no longer a member of a RTC session that you are actually still participating in, send another state event to put yourself back in the session membership.

This can happen when an administrator overwrites your call membership event (which is allowed even with MSC3757's restrictions on state), or if your delayed disconnection event (via MSC4140) timed out before your client could send a heartbeat to delay it further.

## Checklist

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -772,6 +772,14 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
         if (changed) {
             logger.info(`Memberships for call in room ${this.room.roomId} have changed: emitting`);
             this.emit(MatrixRTCSessionEvent.MembershipsChanged, oldMemberships, this.memberships);
+
+            if (this.isJoined() && !this.memberships.some(this.isMyMembership)) {
+                logger.warn("Missing own membership: force re-join");
+                // TODO: Should this be awaited? And is there anything to tell the focus?
+                this.triggerCallMembershipEventUpdate();
+                // TODO: Is this needed?
+                this.emit(MatrixRTCSessionEvent.JoinStateChanged, true);
+            }
         }
 
         if (this.manageMediaKeys && this.isJoined() && this.makeNewKeyTimeout === undefined) {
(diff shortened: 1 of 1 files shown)


src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -772,6 +772,14 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
         if (changed) {
             logger.info(`Memberships for call in room ${this.room.roomId} have changed: emitting`);
             this.emit(MatrixRTCSessionEvent.MembershipsChanged, oldMemberships, this.memberships);
+
+            if (this.isJoined() && !this.memberships.some(this.isMyMembership)) {
+                logger.warn("Missing own membership: force re-join");
+                // TODO: Should this be awaited? And is there anything to tell the focus?
+                this.triggerCallMembershipEventUpdate();
+                // TODO: Is this needed?
+                this.emit(MatrixRTCSessionEvent.JoinStateChanged, true);
```
robintown, line 781: As a consumer of this module I probably wouldn't expect this to be emitted. There is no change :D

robintown (approved):

AndrewFerr:
Marking as draft until test coverage is added.

Local testing is also hitting some strange behaviour.

AndrewFerr:
This doesn't yet fully work in widget mode + Element Web (and potentially other clients).  EW terminates the widget when it detects that there are no members left in a call.  So, if there's only one person left in a call and their membership is lost (such as by a missed heartbeat), EW will close the widget before EC has a chance to restore the membership.

A naive solution could be to make EW simply wait a few seconds for an "empty" session to become non-empty.  Ideally it could examine Livekit state (if that's even possible).

But EW _does_ need to know when a call is empty, because AFAICT that's how it clears the entry in the room preview bar.

AndrewFerr:
I've made this depend on https://github.com/matrix-org/matrix-js-sdk/pull/4494 because without it, attempts to restore membership state are likely to get rate-limited.

AndrewFerr:
Albeit not fully working for embedded mode, this is still a step in the right direction IMO.

PR #4343, opened 2024-08-09
Resend MatrixRTC encryption keys if a membership has changed

hughns:
This fixes https://github.com/element-hq/element-call/issues/2415 by having call participants proactively sending a  new `io.element.call.encryption_keys` room event with the same key if they detect that another participants has updated the `membershipID` or `created_ts` part of their call membership.

## Checklist

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -640,9 +640,8 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
         }
     };
 
-    private isMyMembership(m: CallMembership): boolean {
-        return m.sender === this.client.getUserId() && m.deviceId === this.client.getDeviceId();
-    }
+    private isMyMembership = (m: CallMembership): boolean =>
+        m.sender === this.client.getUserId() && m.deviceId === this.client.getDeviceId();
 
     /**
      * Examines the latest call memberships and handles any encryption key sending or rotation that is needed.
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -138,6 +138,10 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
     private encryptionKeys = new Map<string, Array<Uint8Array>>();
     private lastEncryptionKeyUpdateRequest?: number;
 
+    // We use this to store the last membership fingerprints we saw, so we can proactively re-send encryption keys
+    // if it looks like a membership has been updated.
+    private lastMembershipFingerprints: Set<string> | undefined;
+
     /**
      * The callId (sessionId) of the call.
      *
@@ -636,6 +640,10 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
         }
     };
 
+    private isMyMembership(m: CallMembership): boolean {
+        return m.sender === this.client.getUserId() && m.deviceId === this.client.getDeviceId();
+    }
+
     public onMembershipUpdate = (): void => {
         const oldMemberships = this.memberships;
         this.memberships = MatrixRTCSession.callMembershipsForRoom(this.room);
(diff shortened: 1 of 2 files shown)


fkwp (approved):
lgtm

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -651,32 +664,53 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+                const candidateUpdates =
+                    Array.from(oldFingerprints).some((x) => !newFingerprints.has(x)) ||
+                    Array.from(newFingerprints).some((x) => !oldFingerprints.has(x));
+                if (candidateUpdates) {
+                    logger.debug(`Member(s) have updated/reconnected: re-sending keys`);
+                    this.requestKeyEventSend();
+                }
+            } else {
+                // store the fingerprints of these latest memberships
+                this.storeLastMembershipFingerprints();
```
AndrewFerr, line 699: Is it a problem that the fingerprints set is updated only when no one has left or joined?  Without always updating it, we'd send more key updates than necessary in some cases.  If this is a cheap operation, I guess it's fine, but maybe it's better to be correct anyways.

For example, someone joins while the fingerprints set is initialized, we'd send keys but not update the fingerprints set; then if something were to trigger another membership update without anyone having joined/left, the fingerprints set would be updated & would appear to have changed -- even though the true set of fingerprints hadn't really changed since the last join -- and we'd send keys again.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -651,32 +664,53 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
-            const newMebershipIds = new Set(
-                this.memberships.filter((m) => !isMyMembership(m)).map(getParticipantIdFromMembership),
+            const newMembershipIds = new Set(
+                this.memberships.filter((m) => !this.isMyMembership(m)).map(getParticipantIdFromMembership),
             );
 
-            const anyLeft = Array.from(oldMebershipIds).some((x) => !newMebershipIds.has(x));
-            const anyJoined = Array.from(newMebershipIds).some((x) => !oldMebershipIds.has(x));
+            const anyLeft = Array.from(oldMembershipIds).some((x) => !newMembershipIds.has(x));
+            const anyJoined = Array.from(newMembershipIds).some((x) => !oldMembershipIds.has(x));
```
AndrewFerr, line 676: TIL about [`Set#symmetricDifference`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/symmetricDifference), which we can consider using once the js-sdk uses Node 22 as its minimum supported version.

src/matrixrtc/MatrixRTCSession.ts
hughns: I agree. I've [changed](https://github.com/matrix-org/matrix-js-sdk/pull/4343/commits/7a97b80861c2274026d871dc344c0c4cec59cbcf) the implementation to always recalculate the fingerprints. 👍

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -636,8 +640,17 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+        m.sender === this.client.getUserId() && m.deviceId === this.client.getDeviceId();
+
+    /**
+     * Examines the latest call memberships and handles any encryption key sending or rotation that is needed.
+     *
+     * This function should be called when the room members or call memberships might have changed.
+     */
     public onMembershipUpdate = (): void => {
         const oldMemberships = this.memberships;
+        const oldFingerprints = this.lastMembershipFingerprints;
```
AndrewFerr, line 653: Nitpick: this can be moved to immediately before `this.storeLastMembershipFingerprints()`, as it's not used outside of the `if (this.manageMediaKeys ...)` block.

AndrewFerr (approved):

src/matrixrtc/MatrixRTCSession.ts
hughns: Good point. Now [done](https://github.com/matrix-org/matrix-js-sdk/pull/4343/commits/2487ad6703380153f97df7fc58e9509fa18d9e27).

PR #4346, opened 2024-08-12
Retry event decryption failures on first failure

hughns:
## Checklist


hughns:
This is a basic fix for https://github.com/element-hq/element-call/issues/2560

fkwp (approved):
lgtm. Since we want to move to to-device based mechanism soon-ish we don't need to over eng this for now

src/matrixrtc/MatrixRTCSessionManager.ts
```diff
@@ -98,8 +98,22 @@ export class MatrixRTCSessionManager extends TypedEventEmitter<MatrixRTCSessionM
...
+            // retry after 1 second. After this we give up.
+            setTimeout(() => this.consumeCallEncryptionEvent(event, true), 1000);
+            return;
+        } else if (event.isDecryptionFailure()) {
+            logger.warn(`Decryption failed for event ${event.getId()}: ${event.decryptionFailureReason}`);
+            return;
+        } else if (isRetry) {
+            logger.info(`Decryption succeeded for event ${event.getId()} after retry`);
+        }
+
```
AndrewFerr, line 116: I think the following is semantically equivalent while avoiding some redundant branches & an extra call to `event.isDecryptionFailure()`
```suggestion
        if (event.isDecryptionFailure()) {
            if (!isRetry) {
                logger.warn(
                    `Decryption failed for event ${event.getId()}: ${event.decryptionFailureReason} will retry once only`,
                );
                // retry after 1 second. After this we give up.
                setTimeout(() => this.consumeCallEncryptionEvent(event, true), 1000);
            } else {
                logger.warn(`Decryption failed for event ${event.getId()}: ${event.decryptionFailureReason}`);
            }
            return;
        } else if (isRetry) {
            logger.info(`Decryption succeeded for event ${event.getId()} after retry`);
        }

```

AndrewFerr (approved):
I posted an optional code optimization, but LGTM overall

Issue #4347, opened 2024-08-13
Cannot use import statement outside a module (ES, CommonJS)


BLCK-B:
I tried to import matrix-js-sdk both in CommonJs and ES projects.

```
(node:22992) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)

import * as matrixcs from "./matrix";
SyntaxError: Cannot use import statement outside a module
    at wrapSafe (node:internal/modules/cjs/loader:1281:20)
    at Module._compile (node:internal/modules/cjs/loader:1321:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1416:10)
    at Module.load (node:internal/modules/cjs/loader:1208:32)
    at Module._load (node:internal/modules/cjs/loader:1024:12)
    at cjsLoader (node:internal/modules/esm/translators:348:17)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:297:7)
    at ModuleJob.run (node:internal/modules/esm/module_job:222:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:316:24)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:123:5)
```

The reported fixes, type module and .mjs, do not work.

To reproduce:
https://github.com/BLCK-B/MatrixSDKissue
- cd theCode
- npm install
- npm start  *OR alternatively* node app.mjs

Node.js v20.16.0

It reportedly does not work in tsnode either (Ho Ching):
![image](https://github.com/user-attachments/assets/735e7fe7-87e4-4d34-923f-308ca42a0a9c)

Likely related:
https://github.com/matrix-org/matrix-js-sdk/issues/4284
https://github.com/matrix-org/matrix-js-sdk/issues/3145

richvdh:
Thanks for the report.

This is actually the same thing as #4284: my "fix" in #4285 just didn't work. If you open `package.json` in https://www.npmjs.com/package/matrix-js-sdk/v/34.2.0?activeTab=code, you can see that there is no top-level `"type": "module"` setting.

I think the problem is that there are actually *two* scripts that modify `package.json` for the release process. `scripts/release/pre-release.sh` is the one that is actually used during release. `scripts/switch_package_to_release.js` is only used in CI. I have no idea why we need two of them (I don't even really know why we need one, to be honest.)

[...]

PR #4355, opened 2024-08-20
Move `type:module` declaration into package.json.

richvdh:
matrix-js-sdk is built into ECMAScript modules, and we should declare it as such. See https://nodejs.org/api/packages.html#type. Failure to do so causes problems for javascript projects attempting to build against matrix-js-sdk: see https://github.com/matrix-org/matrix-js-sdk/issues/4347.

Previously, we did this as part of the package.json switcheroo, but that is unnecessarily fragile.

matrix-react-sdk, element-web, etc are unaffected by this, because they use the typescript files directly, by importing `matrix-js-sdk/src/...`.

Before we can make the switch, we need to rename a couple of commonjs scripts.


scripts/switch_package_to_release.js
richvdh, line 1: github fails to show this, but this is a symlink

dbkr (approved):

Issue #4287, opened 2024-06-26
matrix-js-sdk no longer works in node


richvdh:
https://github.com/matrix-org/matrix-js-sdk/issues/4284 describes a problem wherein matrix-js-sdk no longer works in a CommonJS environment, but actually it's worse than that: it doesn't work in an ES modules environment either.

If we apply the fix from https://github.com/matrix-org/matrix-js-sdk/pull/4285, and run in an esm environment, I get:

```
node:internal/modules/esm/resolve:260
    throw new ERR_MODULE_NOT_FOUND(
          ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/rav/work/matrix-js-sdk/examples/node/node_modules/matrix-js-sdk/lib/matrix' imported from /home/rav/work/matrix-js-sdk/examples/node/node_modules/matrix-js-sdk/lib/index.js
    at finalizeResolution (node:internal/modules/esm/resolve:260:11)
    at moduleResolve (node:internal/modules/esm/resolve:920:10)
    at defaultResolve (node:internal/modules/esm/resolve:1119:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:542:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:511:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:241:38)
    at ModuleJob._link (node:internal/modules/esm/module_job:126:49) {
  code: 'ERR_MODULE_NOT_FOUND',
  url: 'file:///home/rav/work/matrix-js-sdk/examples/node/node_modules/matrix-js-sdk/lib/matrix'
}

Node.js v22.2.0
```

The problem appears to be that `index.js` contains:
```js
import * as matrixcs from "./matrix";
```

However, the file it actually wants is `./matrix.js`, and the nodejs ESM resolver doesn't know it needs to add the extension. Likewise for every link in the project :/

richvdh:
Apparently, the typescript way to do this is actually to

 ```ts
import {foo} from "./matrix.js"`; // note js extension
```

The `.js` is then retained in the output, but typescript is intelligent enough to use the ts file for type definitions (cf https://www.typescriptlang.org/docs/handbook/modules/reference.html#file-extension-substitution)

0ctoPusK:
I got the same error as you had

richvdh:
[...]

PR #4373, opened 2024-08-22
Attempt to fix node.js integration by tweaking imports with babel

Johennes:
Fixes: #4287

## Checklist

### DIFF
--- a/babel.config.cjs
+++ b/babel.config.cjs
@@ -1,3 +1,6 @@
+const fs = require("fs");
+const path = require("path");
+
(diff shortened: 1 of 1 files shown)


richvdh:
Hrm, interesting idea. I'm not a huge fan of adding more complexity to the build process though; I'd rather avoid this "magic" if possible.

Johennes:
Ok, fair enough.Without digging deeper, I'm note entirely sure it wouldn't have weird side effects anyway.

Issue #3304, opened 2023-04-19
Element-R: wire up `encryptAndSendToDevices`


richvdh:
This is used by element call, which could otherwise be an early adopter of the rustified crypto.

richvdh:
Looks like we'll have to expose this functionality in the rust matrix-crypto-sdk as well

richvdh:
Blocked on https://github.com/matrix-org/matrix-rust-sdk/issues/1816

richvdh:
Given that `MatrixClient.encryptAndSendToDevices` takes a `DeviceInfo`, which is part of the legacy crypto stack, we won't be able to implement this directly.

I'd suggest adding a new api `encryptToDeviceMessages` to `CryptoApi`, which takes an array of `(user_id, device_id)` pairs. The application can then call `MatrixClient.queueToDevice`on the results.

hughns:
> I'd suggest adding a new api `encryptToDeviceMessages` to `CryptoApi`, which takes an array of `(user_id, device_id)` pairs. The application can then call `MatrixClient.queueToDevice`on the results.

This sounds perfect 👍

hughns:
n.b. https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/140 is a blocker.

hughns:
@richvdh I have started an implementation at https://github.com/matrix-org/matrix-js-sdk/pull/4380. On the API shape, might you be able to take a look and see if this is going in the right direction?

The motivation behind leaving a `MatrixClient.encryptAndSendToDevices` function is that it then works nicely when using the js-sdk as a matryoshka client within the widget API.

hughns:
https://github.com/matrix-org/matrix-js-sdk/pull/4393 is now a blocker.

PR #4380, opened 2024-09-02
Add CryptoApi.encryptToDeviceMessages() and deprecate Crypto.encryptAndSendToDevices()

hughns:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/3304

- Adds CryptoApi.encryptToDeviceMessages with rust and libolm implementations.
- Deprecates Crypto.encryptAndSendToDevices.

Depends on https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/140 and https://github.com/matrix-org/matrix-js-sdk/pull/4396

## Checklist


src/crypto-api/index.ts
```diff
@@ -550,6 +551,22 @@ export interface CryptoApi {
...
+     * @param devices an array of (user ID, device ID) pairs to encrypt the payload for
+     * @param payload the payload to encrypt
+     *
+     * @returns a promise which resolves to the batch of encrypted payloads which can then be sent via {@link MatrixClient#queueToDevice}
+     */
+    encryptToDeviceMessages(
+        eventType: string,
+        devices: { userId: string; deviceId: string }[],
+        payload: ToDevicePayload,
+    ): Promise<ToDeviceBatch>;
```
richvdh, line 569: this interface is divided into sections, and you're adding this under "Import/export of secret keys". Could you find  a better place, please?

src/rust-crypto/rust-crypto.ts
```diff
@@ -1713,6 +1714,37 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
...
     }
+
+    public async encryptToDeviceMessages(
+        eventType: string,
+        devices: { userId: string; deviceId: string }[],
+        payload: ToDevicePayload,
+    ): Promise<ToDeviceBatch> {
+        const batch: ToDeviceBatch = {
+            batch: [],
+            eventType,
```
richvdh, line 1725: shouldn't this be `m.room.encrypted`?

src/rust-crypto/rust-crypto.ts
```diff
@@ -1760,6 +1761,49 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEv
     public async getOwnIdentity(): Promise<RustSdkCryptoJs.OwnUserIdentity | undefined> {
         return await this.olmMachine.getIdentity(new RustSdkCryptoJs.UserId(this.userId));
     }
+
+    public async encryptToDeviceMessages(
```
richvdh, line 1765: We normally try and give some sort of signposting doc-comment:

```suggestion
    /**
     * Implementation of {@link CryptoApi#encryptToDeviceMessages}.
     */
    public async encryptToDeviceMessages(
```

Also, please can you stick it alongside the other CryptoApi method implementations (note there are a few box comments scattered throughout the `RustCrypto` class which attempt to divide it into sections).

src/crypto-api/index.ts
```diff
@@ -364,6 +365,22 @@ export interface CryptoApi {
      */
     getEncryptionInfoForEvent(event: MatrixEvent): Promise<EventEncryptionInfo | null>;
 
+    /**
+     * Encrypts a given payload object via Olm to-device messages to a given
+     * set of devices.
+     *
+     * @param eventType the type of the event to send
```
richvdh, line 372: doc-comments in the js-sdk put a `-` between param name and description. (I actually thought we enforced this with a lint, but seems I'm wrong.)

They also end in a full-stop, preferably, though there are plenty of counterexamples of that.

```suggestion
     * @param eventType - the type of the event to send.
```

likewise elsewhere

src/crypto-api/index.ts
```diff
@@ -364,6 +365,22 @@ export interface CryptoApi {
...
+     * @param devices an array of (user ID, device ID) pairs to encrypt the payload for
```
richvdh, line 373: "pair" implies a 2-tuple to me, which isn't what these are. We can probably just gloss over it and let the type speak for itself:

```suggestion
     * @param devices - an array of devices to encrypt the payload for.
```

richvdh, line 373: I wonder if we can/should impart any advice about the maximum number of entries in `devices`?

spec/unit/rust-crypto/rust-crypto.spec.ts
richvdh, line 1: It would be nice to add an integration test in `spec/integ/crypto` that exercises the encryption without mocking out bits of OlmMachine. I'm aware that such things can be fiddly to get to work, though, so I won't insist.

richvdh:
BTW it looks like the description in the PR is a bit outdated -- could you give it a refresh?

src/crypto-api/index.ts
hughns: A cursory look suggests that the limiting factor would be the size of the claim request generated by `KeyClaimManager.ensureSessionsForUsers()`.

Might it make sense to batch that at some sensible level? Similar to the batch size of 20 that is imposed for actually sending the to-device events to the server.

As a user of the higher-level CryptoApi I think any batching would be done for me automatically (which currently isn't the case).

hughns: Agreed. I've made https://github.com/matrix-org/matrix-js-sdk/issues/4468 for this.

hughns:
@richvdh I've had a go at adding an integration test in https://github.com/matrix-org/matrix-js-sdk/pull/4380/commits/a87a29f0a27c41d592afe6e3527219b9b7cfe024.

However, whilst it passes on libolm it doesn't pass on rust SDK. It appears that `getMissingSessions()` is always returning `null`. Any idea what might be happening or missing?

richvdh:
> It appears that `getMissingSessions()` is always returning `null`. Any idea what might be happening or missing?

Well, that means that there are Olm Sessions already established for all the devices that the rust layer knows about. In this case, it probably means that the rust layer doesn't know about any of the target user's devices. 

We only fetch the device list when we are told by the homeserver that our cached device list for a given user might be outdated. And *that* only happens if we share at least one room with the offending user. (Is that likely to be a problem in practice, for this application?)

Try something like:

```ts
        const syncResponder = new SyncResponder(homeserverUrl);
        syncResponder.sendOrQueueSyncResponse(getSyncResponse([testData.BOB_TEST_USER_ID]));
```

I think that will probably do the job, though you might have to do something else to trigger the `/keys/query` request. (As long as there is a `/keys/query` request pending, `getMissingSessions` should block until it happens.)

spec/integ/crypto/to-device-messages.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+        10000,
+    );
+
+    afterEach(async () => {
+        aliceClient.stopClient();
+        fetchMock.mockReset();
+    });
+
+    describe("encryptToDeviceMessages", () => {
+        it("returns empty batch for device without key", async () => {
```
richvdh, line 94: this is probably actually testing the case that Bob doesn't have any devices, rather than when the device doesn't have any keys. Maybe add that syncResponderStuff here? Better yet, move it to `beforeEach?`

richvdh (approved):
LGTM

PR #4391, opened 2024-09-06
MSC4133 - Extended profiles

Half-Shot:
For https://github.com/matrix-org/matrix-spec-proposals/pull/4133

## Checklist

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -8807,6 +8807,128 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 
     /**
+     * Fetch a user's *extended* profile, which may include additonal keys.
+     * 
+     * @see https://github.com/tcpipuk/matrix-spec-proposals/blob/main/proposals/4133-extended-profiles.md
+     * @param userId The user ID to fetch the profile of.
+     * @returns A set of keys to property values.
+     */
+    public async getExtendedProfile(userId: string): Promise<Record<string, unknown>> {
+        if (!await this.doesServerSupportUnstableFeature("uk.tcpip.msc4133")) {
+            throw new Error('Server does not support extended profiles')
+        }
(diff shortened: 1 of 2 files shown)


src/models/profile-keys.ts
```diff
@@ -0,0 +1 @@
+export const ProfileKeyMSC4175Timezone = "us.cloke.msc4175.tz";
```
Half-Shot, line 1: This is used in https://github.com/matrix-org/matrix-react-sdk/pull/12967 and made slightly more sense to expose as a const here, than as part of Element. Thoughts?

src/client.ts
```diff
@@ -8806,6 +8808,183 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+        }
+        const userId = this.getUserId();
+
+        await this.http.authedRequest(
+            Method.Put,
+            utils.encodeUri("/profile/$userId", { $userId: userId }),
+            {},
+            profile,
+            {
+                prefix: (await this.doesServerSupportUnstableFeature("uk.tcpip.msc4133.stable"))
```
dbkr, line 8981: This feels a bit repetitive - maybe a extendedProfilePrefix or something would be useful?

src/models/profile-keys.ts
dbkr: Yeah, I'm always unsure about these things but it feels useful to have here I guess.

dbkr (commented):
otherwise this looks good I think

src/client.ts
Half-Shot: Done.

src/models/profile-keys.ts
Half-Shot: I'll add a docstring, because I think that would probably be a kind thing to do.

dbkr (approved):
Thanks!

PR #4394, opened 2024-09-09
Fix build failure in node.js example

Johennes:
Relates to: #4287
Relates to: element-hq/element-web#26922
Signed-off-by: Johannes Marbach <n0-0ne+github@mailbox.org>

## Checklist

### DIFF
--- a/examples/node/app.js
+++ b/examples/node/app.js
@@ -399,4 +399,4 @@ function fixWidth(str, len) {
     return str;
 }
 
-matrixClient.startClient({ initialSyncLimit: numMessagesToShow }); // messages for each room.
+matrixClient.startClient({ initialSyncLimit: numMessagesToShow });
--- a/examples/node/app.js
+++ b/examples/node/app.js
@@ -1,12 +1,12 @@
-var myHomeServer = "http://localhost:8008";
-var myUserId = "@example:localhost";
-var myAccessToken = "QGV4YW1wbGU6bG9jYWxob3N0.qPEvLuYfNBjxikiCjP";
-
 import clc from "cli-color";
 import fs from "fs";
 import readline from "readline";
 import sdk, { ClientEvent, EventType, MsgType, RoomEvent } from "matrix-js-sdk";
 import { KnownMembership } from "matrix-js-sdk/lib/@types/membership.js";
+
+var myHomeServer = "http://localhost:8008";
+var myUserId = "@example:localhost";
+var myAccessToken = "QGV4YW1wbGU6bG9jYWxob3N0.qPEvLuYfNBjxikiCjP";
 
 var matrixClient = sdk.createClient({
     baseUrl: myHomeServer,
--- a/examples/node/app.js
+++ b/examples/node/app.js
@@ -1,9 +1,15 @@
+var myHomeServer = "http://localhost:8008";
 var myUserId = "@example:localhost";
 var myAccessToken = "QGV4YW1wbGU6bG9jYWxob3N0.qPEvLuYfNBjxikiCjP";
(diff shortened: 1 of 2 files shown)


examples/node/package.json
```diff
@@ -5,8 +5,9 @@
     "main": "app.js",
     "author": "",
     "license": "Apache 2.0",
+    "type": "module",
     "dependencies": {
         "cli-color": "^1.0.0",
-        "matrix-js-sdk": "^32.0.0"
+        "matrix-js-sdk": "^34.5.0-rc.0"
```
Johennes, line 11: Would be nice to install the SDK from `file:../..` instead and then do something on the CI to verify it doesn't regress again. I couldn't think of a good way to do that though.

examples/node/app.js
```diff
@@ -1,9 +1,15 @@
...
 var myUserId = "@example:localhost";
 var myAccessToken = "QGV4YW1wbGU6bG9jYWxob3N0.qPEvLuYfNBjxikiCjP";
-var sdk = require("matrix-js-sdk");
-var clc = require("cli-color");
+
+import clc from "cli-color";
+import fs from "fs";
+import readline from "readline";
+import sdk, { ClientEvent, EventType, MsgType, RoomEvent } from "matrix-js-sdk";
+import { KnownMembership } from "matrix-js-sdk/lib/@types/membership.js";
```
Johennes, line 9: This is odd but the type isn't exported in `matrix.ts` even though [others are](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/matrix.ts#L62). Not sure if there's a reason for that?

richvdh, line 9: can we put the imports at the start of the file?

richvdh: Looks like an oversight to me.

examples/node/app.js
```diff
@@ -394,4 +399,4 @@ function fixWidth(str, len) {
     return str;
 }
 
-matrixClient.startClient(numMessagesToShow); // messages for each room.
+matrixClient.startClient({ initialSyncLimit: numMessagesToShow }); // messages for each room.
```
richvdh, line 402: not a new problem, but this comment is... opaque.

```suggestion
matrixClient.startClient({ initialSyncLimit: numMessagesToShow });
```

examples/node/package.json
richvdh: I'm torn on this. On the one hand, I'd like to set an example of how you might actually depend on the js-sdk in a project, which points towards an explicit version requirement here. On the other hand, I'd like to make sure that the examples continue to work as the js-sdk evolves, pointing towards `file:./..`.

Maybe we can use an explicit version, but keep it up-to-date with Renovate?

richvdh, line 11: In any case, I've prevaricated long enough that we can use 34.5 final now:

```suggestion
        "matrix-js-sdk": "^34.5.0"
```

examples/node/app.js
Johennes: Ok. I've moved this to https://github.com/matrix-org/matrix-js-sdk/pull/4405 since fixing it in this PR won't be possible if we stick to an explicit version for the JS SDK dependency.

examples/node/package.json
Johennes: Or actually, I guess we can just add this into https://docs.renovatebot.com/configuration-options/#filematch? Is there a way to test this without merging to `develop`?

richvdh: I have honestly no idea. Renovate is a mystery to me.

richvdh (approved):
LGTM for now at least

PR #4400, opened 2024-09-12
Switch sliding sync support to simplified sliding sync

dbkr:
Switches js-sdk's sliding sync support to simplified sliding sync. This is still experimental.

This does not maintain support for regular sliding sync.

## Checklist


DoM1niC:
@dbkr loading DMs names are fixed for me but Avatars don't load inital only when I click on a DM (Room) and the Timeline shows only one recent Message, like this 
image
after I reopen the Room the History will partially load....

src/common-crypto/CryptoBackend.ts
```diff
@@ -138,6 +138,15 @@ export interface SyncCryptoCallbacks {
      * @param syncState - information about the completed sync.
      */
     onSyncCompleted(syncState: OnSyncCompletedData): void;
+
+    /**
+     * Mark all tracked user's device lists as dirty.
```
richvdh, line 143:
```suggestion
     * Mark all tracked users' device lists as dirty.
```

src/common-crypto/CryptoBackend.ts
```diff
@@ -138,6 +138,15 @@ export interface SyncCryptoCallbacks {
...
+     *
+     * This method will cause additional /keys/query requests on the server, so should be used only
```
richvdh, line 145:
```suggestion
     * This method will cause additional `/keys/query` requests on the server, so should be used only
```

src/crypto-api/index.ts
```diff
@@ -418,6 +418,12 @@ export interface CryptoApi {
      */
     getEncryptionInfoForEvent(event: MatrixEvent): Promise<EventEncryptionInfo | null>;
 
+    /**
+     * All users *whose device lists we are tracking* are flagged as needing a
+     * key query. Users whose devices we are not tracking are ignored.
+     */
+    markAllTrackedUsersAsDirty(): Promise<void>;
```
richvdh, line 425: Are you sure this needs exposing to the application? If so, could you share some guidance on when an application should call it? Also, could you have a look through the whole of `CryptoApi` and see if there isn't a better place it fits in the order of methods?

src/models/room-summary.ts
```diff
@@ -14,11 +14,18 @@ See the License for the specific language governing permissions and
...
 */
 
+import type { Hero } from "./room.ts";
+
 export interface IRoomSummary {
     "m.heroes": string[];
     "m.joined_member_count"?: number;
     "m.invited_member_count"?: number;
 }
+export interface IRoomSummaryMSC4186 {
```
richvdh, line 24: interfaces shouldn't start with I

andybalaam (approved):
Looks good from a Crypto point of view.

src/models/room-summary.ts
richvdh, line 24: As usual, this needs documentation; particularly so in this case given it is part of js-sdk's public API

src/models/room-summary.ts
```diff
@@ -14,11 +14,18 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import type { Hero } from "./room.ts";
```
richvdh, line 17: feels like `Hero` should probably come and live in this file, to avoid a reference cycle.

src/models/room.ts
```diff
@@ -940,12 +955,42 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         if (nonFunctionalMemberCount > 2) return;
 
         // Prefer the list of heroes, if present. It should only include the single other user in the DM.
-        const nonFunctionalHeroes = this.summaryHeroes?.filter((h) => !functionalMembers.includes(h));
+        const nonFunctionalHeroes = this.heroes?.filter((h) => !functionalMembers.includes(h.userId));
         const hasHeroes = Array.isArray(nonFunctionalHeroes) && nonFunctionalHeroes.length;
         if (hasHeroes) {
+            // use first hero which exists
```
richvdh, line 961: exists?

src/models/room.ts
```diff
@@ -178,6 +178,20 @@ export type RoomEmittedEvents =
...
 
+// A Hero is a stripped m.room.member event which contains the key renderable fields from the event.
+// It is used in MSC4186 (Simplified Sliding Sync) as a replacement for the old 'summary' field.
+// The old form simply contained the hero's user ID, which forced clients to then look up the
+// m.room.member event in the current state. This is entirely decoupled in SSS. To ensure this
+// works in a backwards compatible way, we will A) only set displayName/avatarUrl with server-provided
+// values, B) always prefer the hero values if they are set, over calling `.getMember`. This means
+// in SSS mode we will always use the heroes if they exist, but in sync v2 mode these fields will
+// never be set and hence we will always do getMember lookups (at the right time as well).
+export type Hero = {
```
richvdh, line 189: this needs a doc-comment. The existing `//` comment might provide some inspiration, but it looks like it contains a lot of js-sdk-internal notes which don't really belong in the public API.

src/models/room.ts
```diff
@@ -178,6 +178,20 @@ export type RoomEmittedEvents =
...
     | PollEvent.New;
 
+// A Hero is a stripped m.room.member event which contains the key renderable fields from the event.
+// It is used in MSC4186 (Simplified Sliding Sync) as a replacement for the old 'summary' field.
+// The old form simply contained the hero's user ID, which forced clients to then look up the
+// m.room.member event in the current state. This is entirely decoupled in SSS. To ensure this
+// works in a backwards compatible way, we will A) only set displayName/avatarUrl with server-provided
+// values, B) always prefer the hero values if they are set, over calling `.getMember`. This means
+// in SSS mode we will always use the heroes if they exist, but in sync v2 mode these fields will
+// never be set and hence we will always do getMember lookups (at the right time as well).
```
richvdh, line 188: A bunch of this commentary actually belongs where we do the lookup in `.heroes`.

src/models/room.ts
```diff
@@ -940,12 +955,42 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+                            type: EventType.RoomMember,
+                            state_key: hero.userId,
+                            content: {
+                                displayname: hero.displayName,
+                                avatar_url: hero.avatarUrl,
+                            },
+                        }),
+                    );
+                    return heroMember;
+                }
```
richvdh, line 989: this logic feels broken.

If we have a SSSS hero who *hasn't* set a displayname or avatar, then we use `this.getMember()`. Isn't that path meant to be used only for the legacy-hero-path?

src/models/room.ts
```diff
@@ -940,12 +955,42 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
 
         // Prefer the list of heroes, if present. It should only include the single other user in the DM.
-        const nonFunctionalHeroes = this.summaryHeroes?.filter((h) => !functionalMembers.includes(h));
+        const nonFunctionalHeroes = this.heroes?.filter((h) => !functionalMembers.includes(h.userId));
         const hasHeroes = Array.isArray(nonFunctionalHeroes) && nonFunctionalHeroes.length;
         if (hasHeroes) {
+            // use first hero which exists
+            for (const hero of nonFunctionalHeroes) {
+                if (!hero.displayName && !hero.avatarUrl) {
+                    // attempt to look up renderable fields from the m.room.member event if it exists
```
richvdh, line 964: why do we check if the user is a member of the room in this case, but not in the case that `hero.displayName` or `hero.avatarUrl` have been set?

src/models/room.ts
```diff
@@ -1616,8 +1679,16 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+    public setSummary(summary: IRoomSummary | IRoomSummaryMSC4186): void {
+        // map heroes onto the MSC4186 form as that has more data
+        const heroes = summary["m.heroes"]?.map((h) => {
+            if (typeof h === "string") {
+                return {
+                    userId: h,
+                };
+            }
+            return h;
+        });
```
richvdh, line 1691: mild suggestion for clarity

```suggestion
        const heroes = summary["m.heroes"]?.map(
            (h) => (typeof h === "string") ? { userId: h } : h
        );
```

src/models/room.ts
```diff
@@ -1626,16 +1697,16 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
-                return userId !== this.myUserId;
+        if (Array.isArray(heroes) && heroes.length > 0) {
+            // filter out ourselves just in case
+            this.heroes = heroes.filter((h) => {
+                return h.userId != this.myUserId;
             });
         }
 
-        this.emit(RoomEvent.Summary, summary);
+        summary["m.heroes"] = heroes?.map((h) => h.userId);
```
richvdh, line 1707:
```suggestion
        // Convert back to the legacy heroes format for the `RoomEvent.Summary` event.
        summary["m.heroes"] = heroes?.map((h) => h.userId);
```

richvdh: can't help but feel that modifying the passed-in object in this way is bad karma that will surprise someone in future. Maybe we should document this behaviour in the doc-comment?

src/models/room.ts
```diff
@@ -3461,16 +3532,20 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
                     return;
                 }
-                const member = this.getMember(userId);
-                otherNames.push(member ? member.name : userId);
+                if (hero.displayName) {
+                    otherNames.push(hero.displayName);
+                } else {
+                    const member = this.getMember(hero.userId);
+                    otherNames.push(member ? member.name : hero.userId);
+                }
```
richvdh, line 3548: could do with a comment here: what is this for?

src/rust-crypto/rust-crypto.ts
```diff
@@ -1616,6 +1615,13 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
         });
     }
 
+    /**
+     * Implementation of {@link CryptoApi#markAllTrackedUsersAsDirty}.
+     */
```
richvdh, line 1620: note that `RustCrypto` has a "CryptoApi implementation" section, which isn't here. Though, as previously noted, I don't think this actually belongs in `CryptoApi`.

src/sliding-sync-sdk.ts
```diff
@@ -396,6 +392,7 @@ export class SlidingSyncSdk {
             room = _createAndReEmitRoom(this.client, roomId, this.opts);
         }
         await this.processRoomData(this.client, room!, roomData);
+        logger.log(`onRoomData ${roomId} took ${new Date().getTime() - start}ms`);
```
richvdh, line 395: feels like these logging changes could do with being a separate PR

src/sliding-sync.ts
```diff
@@ -82,10 +82,17 @@ export interface MSC3575SlidingSyncRequest {
     clientTimeout?: number;
 }
 
+export interface MSC4186Hero {
+    user_id: string;
+    displayname: string;
+    avatar_url: string;
+}
```
richvdh, line 89: documentation please. How does this differ from the other `Hero`?

src/sliding-sync.ts
```diff
@@ -82,10 +82,17 @@ export interface MSC3575SlidingSyncRequest {
     clientTimeout?: number;
 }
 
+export interface MSC4186Hero {
+    user_id: string;
+    displayname: string;
+    avatar_url: string;
```
richvdh, line 88: How can you be sure that these fields will be present?

richvdh (changes requested):

src/crypto-api/index.ts
dbkr: Seems like it only needed to be in `SyncCryptoCallbacks` which it already was - removed.

src/models/room.ts
dbkr: mm, have tried to clarify based on what it appears to be doing

src/models/room.ts
dbkr: Mmm, have kept the tsdoc fairly lightweight and moved the detail into room.ts

src/models/room.ts
dbkr: yeah, it kinda does - have moved the latter half there.

src/models/room.ts
dbkr: I asked Kegan and it sounds is intentional if the server doesn't populate either field, it will essentially fall back to the old behaviour:

> If we have a Hero then the server has explicitly sent data back and we should use it. But, in the case of where there is no actual data for this hero (avatar/display name) then rather than send nothing we fallback to the m.room.member event's displayname/avatar_url if they exist, as it might have some data, which is better than displaying nothing. So effectively the codepath for v2 fallback because-there-is-no-Hero and SSS-Hero-has-no-data is the same

I've tweaked the comment a bit to reflect this.

src/models/room.ts
dbkr: I asked Kegan and it sounds is intentional if the server doesn't populate either field, it will essentially fall back to the old behaviour:

> If we have a Hero then the server has explicitly sent data back and we should use it. But, in the case of where there is no actual data for this hero (avatar/display name) then rather than send nothing we fallback to the m.room.member event's displayname/avatar_url if they exist, as it might have some data, which is better than displaying nothing. So effectively the codepath for v2 fallback because-there-is-no-Hero and SSS-Hero-has-no-data is the same

I've tweaked the comment a bit to reflect this.

src/models/room.ts
dbkr: Oh, as in should it still ensure that the user is definitely in the room even if all the information comes from the hero object?

src/models/room.ts
dbkr: done in 8a26967f5

src/rust-crypto/rust-crypto.ts
dbkr: yep, it's gone.

src/sliding-sync-sdk.ts
dbkr: Mmm, they do seem a bit unrelated - gone.

src/sliding-sync.ts
dbkr: Done in 7028a88b2

dbkr: Done in 7028a88b2

dbkr: Done in 7028a88b2

src/sliding-sync.ts
dbkr: Mm, not sure why these weren't optional since the code was good about checking their existence. Fixed.

src/models/room.ts
```diff
@@ -958,8 +946,15 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
-            // use first hero which exists
+            // use first hero that has a display name or avatar url, or whose user ID
+            // can be looked up as a member of the room
             for (const hero of nonFunctionalHeroes) {
+                // The old form of heroes simply contained the hero's user ID, which forced clients to then look up the
+                // m.room.member event in the current state. This is entirely decoupled in SSS. To ensure this
+                // works in a backwards compatible way, we will A) only set displayName/avatarUrl with server-provided
+                // values, B) always prefer the hero values if they are set, over calling `.getMember`. This means
+                // in SSS mode we will use the heroes if the display name or avatar URL fields are set, but in sync v2
+                // mode these fields will never be set and hence we will always do getMember lookups (at the right time as well).
```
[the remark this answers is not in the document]
richvdh: hrm, that's not the same thing as kegan said, apparently:

> But, in the case of where there is no actual data for this hero (avatar/display name) then rather than send nothing we fallback to the m.room.member event's displayname/avatar_url if they exist, as it might have some data, which is better than displaying nothing

I don't really understand why the server would return no avatar/displayname when the hero has set them. Is this to work around a server bug?

src/models/room.ts
richvdh: I don't really understand, but let's take it to another thread.

src/models/room.ts
```diff
@@ -1679,16 +1674,16 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         return this.setUnreadNotificationCount(type, count);
     }
 
-    public setSummary(summary: IRoomSummary | IRoomSummaryMSC4186): void {
+    /**
+     * Takes either a legacy or MSC4186 room summary and updates the room with it.
+     * The summary will be converted to a legacy style summary in order to be emitted
+     * in the `RoomEvent.Summary` event.
```
richvdh, line 1680: this buries the lede

```suggestion
     * Note that the provided `summary` will be modified. (Specifically, it will be converted to a legacy-style 
     * `IRoomSummary`).
```

richvdh (changes requested):

src/models/room.ts
```diff
@@ -376,6 +362,8 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
     // read by megolm via getter; boolean value - null indicates "use global value"
     private blacklistUnverifiedDevices?: boolean;
     private selfMembership?: Membership;
+    // A Hero is a stripped m.room.member event which contains the key renderable fields from the event.
+    // It is used in MSC4186 (Simplified Sliding Sync) as a replacement for the old 'summary' field.
```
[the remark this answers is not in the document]
dbkr: Yeah, that matches my understanding.

src/models/room.ts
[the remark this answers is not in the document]
kegsay: There isn't a specific server bug here, it's primarily to avoid showing user IDs / no avatars if we can help it. The proxy would map `m.room.member` events to the `Hero` structs so they _should_ always be in sync. If they are out of sync, it will be either:
 - the Hero is blank but the member event on the client has data.
 - The member event on the client is blank and the Hero has data.

The code we're talking about is for the first point, which I think can only happen if:
 - Alice has a name/avatar set.
 - Bob goes offline for ages.
 - Alice unsets her name/avatar.
 - Bob comes back online, and is sent Alice as a Hero with no name/avatar.
 - Bob uses the old values.

..which is probably not what we want here, so happy for us to always use the Hero values and not fallback in SSS mode.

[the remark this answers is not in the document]
dbkr: I can go make it do that then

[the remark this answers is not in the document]
dbkr: OK, I've had a go at this in 7506df233, although it does mean explicitly tracking where the hero came from so I'm not 100% sure I've interpreted this correctly, but easiest to show in an actual diff than try to explain. If so, I think most of the rest of the feedback is fixed too.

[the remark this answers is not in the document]
dbkr: Oh wait no, this doesn't work at all because the same type is used in the RoomSummaryMSC4186 type

src/sliding-sync.ts
```diff
@@ -82,10 +82,14 @@ export interface MSC3575SlidingSyncRequest {
...
+ * New format of hero introduced in MSC4186 with display name and avatar URL
+ * in addition to just user_id
+ */
 export interface MSC4186Hero {
     user_id: string;
-    displayname: string;
-    avatar_url: string;
+    displayname?: string;
+    avatar_url?: string;
 }
```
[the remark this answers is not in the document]
dbkr: I've added a field to one of them too in order to track the difference. We could extend, but if we have to map then I guess we may as well have two distinct types? No strong opinion though., I have linked between the two, albeit not with proper `@link` because it doesn't work without importing the type which gets flagged as an unused import.

src/models/room.ts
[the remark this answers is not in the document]
dbkr: Right, now updated again, splitting he setSummary method in two to avoid having to fake up an object to make it look like the old format which makes one of the types unnecessary.

src/models/room-summary.ts
```diff
@@ -16,16 +16,23 @@ limitations under the License.
...
+ * This is very similar to MSC4186Hero from sliding-sync.ts but an internal format with
+ * camelCase rather than underscores.
  */
 export type Hero = {
     userId: string;
     displayName?: string;
     avatarUrl?: string;
+    /**
+     * If true, the hero is from an MSC4186 summary, in which case `displayName` and `avatarUrl` will
+     * have been set by the server if available. If false, these fields will always be undefined.
```
richvdh, line 29:
```suggestion
     * have been set by the server if available. If false, the `Hero` has been constructed from a `/v3/sync` response,
     * so these fields will always be undefined.
```

src/models/room.ts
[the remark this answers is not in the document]
richvdh: looks sensible to me now.

src/models/room.ts
```diff
@@ -1626,16 +1693,52 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
-        this.emit(RoomEvent.Summary, summary);
+        this.emit(RoomEvent.Summary, summary as IRoomSummary);
+    }
+
+    /**
+     * Takes information from the MSC4186 room summary and updates the room with it.
+     *
+     * @param heroes
+     * @param joinedCount
+     * @param invitedCount
```
richvdh, line 1711: :-1:

src/models/room.ts
```diff
@@ -3461,16 +3564,23 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
 
         // get members that are NOT ourselves and are actually in the room.
```
richvdh, line 3565: this comment looks outdated -- we no longer check if heros are in the room

src/models/room.ts
```diff
@@ -3461,16 +3564,23 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
         let otherNames: string[] = [];
-        if (this.summaryHeroes) {
+        if (this.heroes) {
             // if we have a summary, the member state events should be in the room state
```
richvdh, line 3568: this comment feels like it's got a bit lost, as well

richvdh (approved):
A few nits but generally lgtm

src/models/room.ts
```diff
@@ -1626,16 +1693,52 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+                    userId: h.user_id,
+                    displayName: h.displayname,
+                    avatarUrl: h.avatar_url,
+                    fromMSC4186: true,
+                }));
+        }
+        if (joinedCount !== undefined && Number.isInteger(joinedCount)) {
+            this.currentState.setJoinedMemberCount(joinedCount);
+        }
+        if (invitedCount !== undefined && Number.isInteger(invitedCount)) {
```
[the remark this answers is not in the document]
dbkr: Typescript complains without these checks as Number.isInteger doesn't accept `undefined` for whatever reason.

src/models/room.ts
dbkr: Mmm yeah, should be better now.

dbkr:
Force merging because the upstream will be incompatible

PR #4401, opened 2024-09-13
Move `crypto/key_passphrase.ts` to `crypto-api/key-passphrase.ts`

florianduros:
## Checklist

Changes:
- Moving `keyFromAuthData` to `common-crypto` and deprecates it. It should be deleted but it's still used.
- Move `deriveKey` into `crypto-api` and renamed into `deriveRecoveryKeyFromPassphrase`
- Deprecate `MatrixClient.keyBackupKeyFromPassword`

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -3656,7 +3656,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @param password - Passphrase
      * @param backupInfo - Backup metadata from `checkKeyBackup`
      * @returns key backup key
-     * @deprecated Use {@link keyFromAuthData} directly.
+     * @deprecated
      */
     public keyBackupKeyFromPassword(password: string, backupInfo: IKeyBackupInfo): Promise<Uint8Array> {
         return keyFromAuthData(backupInfo.auth_data, password);
--- a/src/client.ts
+++ b/src/client.ts
@@ -86,7 +86,6 @@ import {
 } from "./crypto/index.ts";
 import { DeviceInfo } from "./crypto/deviceinfo.ts";
 import { decodeRecoveryKey } from "./crypto/recoverykey.ts";
-import { keyFromAuthData } from "./crypto/key_passphrase.ts";
 import { User, UserEvent, UserEventHandlerMap } from "./models/user.ts";
 import { getHttpUriForMxc } from "./content-repo.ts";
 import { SearchResult } from "./models/search-result.ts";
@@ -223,7 +222,13 @@ import { LocalNotificationSettings } from "./@types/local_notifications.ts";
 import { buildFeatureSupportMap, Feature, ServerSupport } from "./feature.ts";
 import { BackupDecryptor, CryptoBackend } from "./common-crypto/CryptoBackend.ts";
 import { RUST_SDK_STORE_PREFIX } from "./rust-crypto/constants.ts";
-import { BootstrapCrossSigningOpts, CrossSigningKeyInfo, CryptoApi, ImportRoomKeysOpts } from "./crypto-api/index.ts";
+import {
+    BootstrapCrossSigningOpts,
+    CrossSigningKeyInfo,
+    CryptoApi,
+    ImportRoomKeysOpts,
+    keyFromAuthData,
+} from "./crypto-api/index.ts";
 import { DeviceInfoMap } from "./crypto/DeviceList.ts";
 import {
     AddSecretStorageKeyOpts,
--- a/src/client.ts
+++ b/src/client.ts
@@ -227,7 +227,6 @@ import {
     CryptoApi,
     decodeRecoveryKey,
     ImportRoomKeysOpts,
-    keyFromAuthData,
 } from "./crypto-api/index.ts";
 import { DeviceInfoMap } from "./crypto/DeviceList.ts";
 import {
@@ -244,6 +243,7 @@ import { RoomMessageEventContent, StickerEventContent } from "./@types/events.ts
 import { ImageInfo } from "./@types/media.ts";
 import { Capabilities, ServerCapabilities } from "./serverCapabilities.ts";
 import { sha256 } from "./digest.ts";
+import { keyFromAuthData } from "./common-crypto/key-passphrase.ts";
 
 export type Store = IStore;
 
--- a/src/client.ts
+++ b/src/client.ts
@@ -3650,6 +3650,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @param password - Passphrase
      * @param backupInfo - Backup metadata from `checkKeyBackup`
      * @returns key backup key
+     * @deprecated Use {@link keyFromAuthData} directly.
      */
     public keyBackupKeyFromPassword(password: string, backupInfo: IKeyBackupInfo): Promise<Uint8Array> {
         return keyFromAuthData(backupInfo.auth_data, password);
--- a/src/client.ts
+++ b/src/client.ts
@@ -85,7 +85,6 @@ import {
     isCryptoAvailable,
 } from "./crypto/index.ts";
 import { DeviceInfo } from "./crypto/deviceinfo.ts";
-import { decodeRecoveryKey } from "./crypto/recoverykey.ts";
 import { User, UserEvent, UserEventHandlerMap } from "./models/user.ts";
 import { getHttpUriForMxc } from "./content-repo.ts";
 import { SearchResult } from "./models/search-result.ts";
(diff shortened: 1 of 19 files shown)


florianduros:
Sorry, back to draft I pushed the wrong button.

florianduros:
I tried to add documentation but my knowledge around crypto is not enough. Maybe @andybalaam or @richvdh can give me a hand on it? Thanks!

src/crypto-api/key-passphrase.ts
```diff
@@ -0,0 +1,104 @@
...
+        throw new Error("Salt and/or iterations not found: " + "this backup cannot be restored with a passphrase");
+    }
+
+    return deriveKey(
+        password,
+        authData.private_key_salt,
+        authData.private_key_iterations,
+        authData.private_key_bits || DEFAULT_BITSIZE,
+    );
+}
```
richvdh, line 53: Really, this thing ought to be removed; it's about deriving a *backup* key from a *backup* passphrase. That's a thing that we used to do before 4S was a thing, but afaict it was never actually specced, and it's certainly not a normal thing to do to nowadays. There is some code in react-sdk that calls it (via MatrixClient.restoreKeyBackupWithPassword and MatrixClient.keyBackupKeyFromPassword), but I don't think it actually works. https://github.com/element-hq/element-web/issues/27455#issuecomment-2160771635).

Still, we're not really ready to rip it out just yet. Since it's not currently used directly from outside the js-sdk, I suggest that rather than exposing a new method to applications via `crypto-api`, you instead put this somewhere within `common-crypto`.

src/crypto-api/key-passphrase.ts
```diff
@@ -0,0 +1,104 @@
...
+/* eslint-enable camelcase */
+
+interface IKey {
+    key: Uint8Array;
+    salt: string;
+    iterations: number;
+}
+
+/**
+ * Derive a key from a passphrase using the salt and iterations from the auth data.
```
richvdh, line 38:
```suggestion
 * Derive a backup key from a passphrase using the salt and iterations from the auth data.
```

src/client.ts
```diff
@@ -3645,6 +3650,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @param password - Passphrase
      * @param backupInfo - Backup metadata from `checkKeyBackup`
      * @returns key backup key
+     * @deprecated Use {@link keyFromAuthData} directly.
```
richvdh, line 3653: Per my other comment: I think we just deprecate this as something we're going to stop supporting, rather than recommending an alternative.

src/crypto-api/key-passphrase.ts
```diff
@@ -0,0 +1,104 @@
...
+    return deriveKey(
+        password,
+        authData.private_key_salt,
+        authData.private_key_iterations,
+        authData.private_key_bits || DEFAULT_BITSIZE,
+    );
+}
+
+/**
+ * Derive a key from a passphrase.
```
richvdh, line 56:
```suggestion
 * Generate a new recovery key, based on a passphrase.
```

src/crypto-api/key-passphrase.ts
```diff
@@ -0,0 +1,104 @@
...
+ * @param password
+ */
+export async function keyFromPassphrase(password: string): Promise<IKey> {
+    const salt = randomString(32);
+
+    const key = await deriveKey(password, salt, DEFAULT_ITERATIONS, DEFAULT_BITSIZE);
+
+    return { key, salt, iterations: DEFAULT_ITERATIONS };
+}
```
richvdh, line 65: This isn't something that applications should be calling directly, either (instead they should call `CryptoApi.createRecoveryKeyFromPassphrase`), so again let's not add it here.

Currently, it's only called from (a) bits of legacy crypto, and (b) `RustCrypto.createRecoveryKeyFromPassphrase`.

Since it's only two lines, I'd be inclined to inline it in `RustCrypto.createRecoveryKeyFromPassphrase`, and leave the function in its old place in `crypto/key_passphrase.ts` for use by the legacy crypto stack.

src/crypto-api/key-passphrase.ts
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Derive a key from a passphrase using PBKDF2.
```
richvdh, line 68:
```suggestion
 * Derive a recovery key from a passphrase and salt using PBKDF2.
 *
 * @see https://spec.matrix.org/v1.11/client-server-api/#deriving-keys-from-passphrases
 *
```

src/crypto-api/key-passphrase.ts
```diff
@@ -0,0 +1,104 @@
...
+ * @param password
+ * @param salt
+ * @param iterations
+ * @param numBits
```
richvdh, line 72: would be good to write some words here. (There's really no value in empty `@param` lists like this.)

src/crypto-api/key-passphrase.ts
```diff
@@ -0,0 +1,104 @@
...
+ */
+export async function deriveKey(
+    password: string,
```
richvdh, line 75: consider renaming to `passphrase`, since that's the word the spec uses.

src/crypto-api/key-passphrase.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+/**
+ * Derive a key from a passphrase using PBKDF2.
+ * @param password
+ * @param salt
+ * @param iterations
+ * @param numBits
+ */
+export async function deriveKey(
```
richvdh, line 74: consider a rename:

```suggestion
export async function deriveRecoveryKeyFromPassphrase(
```

richvdh (changes requested):
These methods are all about deriving a recovery or backup *key* from a *passphrase*; the relevant spec is https://spec.matrix.org/v1.11/client-server-api/#deriving-keys-from-passphrases.

src/crypto-api/key-passphrase.ts
richvdh: btw, don't be afraid to create a second definition  of `DEFAULT_ITERATIONS` (probably with a better name) in rust-crypto. I don't think we need a second definition of `DEFAULT_BITSIZE`  -- we can just omit the parameter.

src/rust-crypto/rust-crypto.ts
```diff
@@ -100,6 +100,12 @@ interface ISignableObject {
  * @internal
  */
 export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEventMap> implements CryptoBackend {
+    /**
+     * The number of iterations to use when deriving a recovery key from a passphrase.
+     * @private
+     */
+    private readonly RECOVERY_KEY_DERIVATION_ITERATION = 500000;
```
richvdh, line 107:
```suggestion
    private readonly RECOVERY_KEY_DERIVATION_ITERATIONS = 500000;
```

src/rust-crypto/rust-crypto.ts
```diff
@@ -100,6 +100,12 @@ interface ISignableObject {
  * @internal
  */
 export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, RustCryptoEventMap> implements CryptoBackend {
+    /**
+     * The number of iterations to use when deriving a recovery key from a passphrase.
+     * @private
```
richvdh, line 105: this is redundant; we don't normally bother.

```suggestion
```

src/common-crypto/key-passphrase.ts
```diff
@@ -0,0 +1,43 @@
...
+    private_key_salt?: string;
+    private_key_iterations?: number;
+    private_key_bits?: number;
+}
+
+/**
+ * Derive a backup key from a passphrase using the salt and iterations from the auth data.
+ * @param authData - The auth data containing the salt and iterations
+ * @param passphrase - The passphrase to derive the key from
+ * @deprecated
```
richvdh, line 30:
```suggestion
 *
 * @deprecated Deriving a backup key from a passphrase is not part of the matrix spec.
 *     Instead, a random key is generated and stored/shared via 4S.
```

src/client.ts
```diff
@@ -3656,6 +3656,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @param password - Passphrase
      * @param backupInfo - Backup metadata from `checkKeyBackup`
      * @returns key backup key
+     * @deprecated
```
richvdh, line 3659:
```suggestion
     *
     * @deprecated Deriving a backup key from a passphrase is not part of the matrix spec.
     *     Instead, a random key is generated and stored/shared via 4S.
```

richvdh (approved):
LGTM other than a few nits

PR #4403, opened 2024-09-13
Update media event content types to include captions

tulir:


florianduros (approved):
Perfect, thanks for the contribution

PR #4406, opened 2024-09-13
Don't share full key history for RTC per-participant encryption

hughns:
Instead we only share what is needed.

This PR also records stats for how many keys have been sent/received and age of those received.

## Checklist

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -144,6 +144,30 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
     // if it looks like a membership has been updated.
     private lastMembershipFingerprints: Set<string> | undefined;
 
+    private currentEncryptionKeyIndex = -1;
+
+    /**
+     * The statistics for this session.
+     */
+    public statistics = {
+        counters: {
+            /**
+             * The number of times we have sent a room event containing encryption keys.
+             */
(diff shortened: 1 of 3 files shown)


src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -541,20 +573,33 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
             return;
         }
 
+        if (typeof indexToSend !== "number" && this.currentEncryptionKeyIndex === -1) {
+            logger.warn("Tried to send encryption keys event but no current key index found!");
+            return;
+        }
+
+        const keyIndexToSend = typeof indexToSend === "number" ? indexToSend : this.currentEncryptionKeyIndex;
```
robintown, line 581:
```suggestion
        const keyIndexToSend = indexToSend ?? this.currentEncryptionKeyIndex;
```

robintown (approved):
I don't fully understand what the 'total age' statistic is there for (wouldn't it be more useful to know say, the max, min, and average age?) but as long as you have a use case for it in mind, this looks fine.

hughns:
> I don't fully understand what the 'total age' statistic is there for (wouldn't it be more useful to know say, the max, min, and average age?) but as long as you have a use case for it in mind, this looks fine.

The total age is there to calculate the average age later on. I've improved the comment to explain this. 👍

Issue #4392, opened 2024-09-06
Can't resolve './rust-crypto' in '<project dir>/node_modules/matrix-js-sdk/lib' with webpack


Harm-van-Stekelenburg:
I can't get the latest release candidate to work with webpack. Now I'm not a javascript expert so this could very well be me overlooking something basic. 

I made a small project to reproduce the issue: https://github.com/Harm-van-Stekelenburg/import_webpack__matrix_js

I run into two problems with this project:

1. When running `npm run serve` I get the error:
```
ERROR in ./node_modules/matrix-js-sdk/lib/client.js 1111:29-52
Module not found: Error: Can't resolve './rust-crypto' in '<project dir>/node_modules/matrix-js-sdk/lib'
```

2. If I comment out that line in `./node_modules/matrix-js-sdk/lib/client.js` it will build but I get the following error in the browser console:
```
Uncaught ReferenceError: global is not defined
    qh matrix.js:116
    Vh matrix.js:134
    9820 index.js:2
    Webpack 3
```
I think that means the global here: https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/matrix.ts#L133

I get neither error on version 34.3.1.

I can fix the second problem with the webpack define plugin to redefine global to something like `window` but that seems a bit unnecessary.

Johennes:
Looks like this might need an additional `/index.ts` on https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/client.ts#L2252. Interestingly this didn't seem to trigger when we fixed https://github.com/matrix-org/matrix-js-sdk/issues/4287 though.

CC @richvdh.

Johennes:
> Interestingly this didn't seem to trigger when we fixed https://github.com/matrix-org/matrix-js-sdk/issues/4287 though.

Ah, small wonder. The node example doesn't use encryption. 🤦‍♂️

Johennes:
Looks like `rewriteImportExtensions` doesn't currently support dynamic imports. So this is blocked on https://github.com/babel/babel/issues/16750.

saul-jb:
@Harm-van-Stekelenburg, I have run into the same problem in the node esm environment and opened a PR to fix it: #4445

I have been adding the following command to my Dockerfile to fix it after building - perhaps you can do something similar for now.

```bash
sed -i 's|import("\./rust-crypto")|import("\./rust-crypto/index.js")|' ./node_modules/matrix-js-sdk/lib/client.js
```

[...]

PR #4409, opened 2024-09-17
Add ESLint rule to ensure .ts extensions on imports

Johennes:
When importing without an extension

```
import { MemoryCryptoStore } from "./crypto/store/memory-crypto-store";
```

this rule throws an error

```
/Users/jm/Code/matrix-js-sdk/src/matrix.ts
  19:35  error  require file extension '.ts'  node/file-extension-in-import
```

Sadly, it does _not_ throw an error when the file is imported with a `.js` extension. I'm not sure why but I suppose forgetting the extension entirely is the most common case anyway.

Relates to: https://github.com/matrix-org/matrix-js-sdk/issues/4392

## Checklist

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,5 +1,5 @@
 module.exports = {
-    plugins: ["matrix-org", "import", "jsdoc"],
+    plugins: ["matrix-org", "import", "jsdoc", "node"],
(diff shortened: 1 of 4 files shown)


florianduros (approved):
Thank you perfect

PR #4410, opened 2024-09-17
Add syntax & type check for Node.js example on CI

Johennes:
This uses `node --check` and `tsc` to perform some basic syntax and type checking on the Node.js example.

It's slightly odd to have this running on every PR when the JS SDK version is pinned in the example's `package.json`. If we find a way to update the JS SDK in the example via renovate, maybe we could switch the CI to use the local SDK instead?

## Checklist


richvdh (approved):
lgtm

.github/workflows/static_analysis.yml
```diff
@@ -43,6 +43,41 @@ jobs:
...
+              run: "npm install"
+              working-directory: "examples/node"
+
+            - name: Check Syntax
+              run: "node --check app.js"
+              working-directory: "examples/node"
+
+            - name: Typecheck
+              run: "npx tsc"
+              working-directory: "examples/node"
```
t3chguy, line 79: Does this do anything given the example contains no typescript? iirc default tsc config doesn't scan js files especially if they lack `@ts-check`

Johennes: It definitely does something.

```
$ git diff
diff --git a/examples/node/app.js b/examples/node/app.js
index 0a16e5a4d..0ca9ed8a5 100644
--- a/examples/node/app.js
+++ b/examples/node/app.js
@@ -137,7 +137,7 @@ rl.on("line", function (line) {
 // ==== END User input

 // show the room list after syncing.
-matrixClient.on(ClientEvent.Sync, function (state, prevState, data) {
+matrixClient.on("foo", function (state, prevState, data) {
     switch (state) {
         case "PREPARED":
             setRoomList();
```

```
$ npx tsc
app.js:140:17 - error TS2345: Argument of type '"foo"' is not assignable to parameter of type 'EmittedEvents | EventEmitterEvents'.

140 matrixClient.on("foo", function (state, prevState, data) {
                    ~~~~~


Found 1 error in app.js:140
```

I had to double check why and it looks like `"checkJs": true` in tsconfig achieves the same thing as `// @ts-check`: https://www.typescriptlang.org/tsconfig/#checkJs

t3chguy (approved):

PR #4423, opened 2024-09-23
Use correct return type for RTCSession encryption keys

hughns:
The functions have always possibly returned `undefined` where the key at a particular index was not known. This changes the return type to match this reality.

## Checklist

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -137,7 +137,7 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
     private manageMediaKeys = false;
     private useLegacyMemberEvents = true;
     // userId:deviceId => array of (key, timestamp)
-    private encryptionKeys = new Map<string, Array<{ key: Uint8Array; timestamp: number }>>();
+    private encryptionKeys = new Map<string, Array<{ key: Uint8Array; timestamp: number } | undefined>>();
     private lastEncryptionKeyUpdateRequest?: number;
 
     // We use this to store the last membership fingerprints we saw, so we can proactively re-send encryption keys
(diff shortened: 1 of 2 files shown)


src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -483,6 +486,7 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
             }
         }
 
+        // n.b. this will extend the array if necessary and fill in the gaps with undefined
```
AndrewFerr, line 489: It will actually make it a [sparse array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Indexed_collections#sparse_arrays) with empty slots that are skipped by iteration methods.

Which gets me thinking, maybe callers can hook into that...

AndrewFerr: > maybe callers can hook into that...

That is indeed the case; see https://github.com/element-hq/element-call/pull/2646#discussion_r1771593805 which makes the updated typehints of this PR unnecessary (though the comments & updated unit test remain useful).

hughns: Well, TypeScript doesn't seem to model sparse arrays at all. So, as far as I can see, there is no compile time hint that a sparse array can contain empty values.

I don't find this very satisfactory. /me ponders

AndrewFerr: There is [`--noUncheckedIndexedAccess`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#checked-indexed-accesses---nouncheckedindexedaccess), but that would require updating many other parts of the codebase to comply with it.

hughns:
I've made https://github.com/matrix-org/matrix-js-sdk/pull/4427 as an alternative which cleans up the API entirely. I prefer that as an approach to this.

PR #4432, opened 2024-09-25
Update OIDC registration types to match latest MSC2966 state

t3chguy:


src/oidc/register.ts
```diff
@@ -35,16 +35,16 @@ export type OidcRegistrationClientMetadata = {
 };
 
 interface OidcRegistrationRequestBody {
```
florianduros, line 37: It can be great to have a link to the MSC doc. To know which spec/msc this type refers to

florianduros (approved):

PR #4435, opened 2024-09-30
Bump matrix-rust-sdk to 9.1.0

richvdh:
Mostly, to pick up https://github.com/matrix-org/matrix-rust-sdk/pull/3985


MidhunSureshR (approved):

hughns:
@richvdh was the a reason to not bump this in the `package.json` as well as `yarn.lock`?

richvdh:
> @richvdh was the a reason to not bump this in the package.json as well as yarn.lock?

The js-sdk is still compatible with 9.0.0, so we don't *require* 9.1.0: it's just preferable.

PR #4442, opened 2024-10-04
Deprecate the crypto events which are not used by the rust-crypto

florianduros:
## Checklist

Deprecate the crypto events which are no longer used by the rust crypto.

### DIFF
--- a/src/crypto/index.ts
+++ b/src/crypto/index.ts
@@ -230,10 +230,14 @@ export interface IMegolmEncryptedContent {
 export type IEncryptedContent = IOlmEncryptedContent | IMegolmEncryptedContent;
 
 export enum CryptoEvent {
+    /** @deprecated Event not fired by the rust crypto */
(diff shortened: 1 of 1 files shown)


src/crypto/index.ts
```diff
@@ -261,6 +266,7 @@ export enum CryptoEvent {
      */
     VerificationRequestReceived = "crypto.verificationRequestReceived",
 
+    /** @deprecated Event not fired by the rust crypto */
     Warning = "crypto.warning",
     WillUpdateDevices = "crypto.willUpdateDevices",
```
richvdh, line 271: I wonder if we should deprecate this one while we're here. Even though it is fired by rust crypto, it's a bit of a bodge, and means exactly the same as "DevicesUpdated" under rust crypto.

richvdh (approved):
LGTM otherwise

PR #4377, opened 2024-08-23
Change imports for Node.js compatibility

Johennes:
Fixes: #4287

This is a variation of #4375 that uses `.ts` extensions and https://babeljs.io/docs/babel-preset-typescript#rewriteimportextensions.

## Checklist

-   [ ] Tests written for new code (and old code if feasible).
-   [ ] New or updated `public`/`exported` symbols have accurate [TSDoc](https://tsdoc.org/) documentation.
-   [x] Linter and other CI checks pass.
-   [x] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md)).

PR #4445, opened 2024-10-08
Fix the rust crypto import in esm environments.

### LINKED ISSUE #4392: Can't resolve './rust-crypto' in '<project dir>/node_modules/matrix-js-sdk/lib' with webpack

saul-jb:
Rust crypto is broken in ESM environments like nodejs and webpack, this PR simply uses babel to rewrite the import to fix it.

Fixes #4392

See also:
- https://github.com/babel/babel/issues/16750
- https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/client.ts#L2259

Signed-off-by: Saul Boyd <saul@organicdesign.co.nz>


babel.config.cjs
```diff
@@ -31,5 +31,19 @@ module.exports = {
...
+                    process.env.NODE_ENV === "test"
+                        ? []
+                        : [
+                              {
+                                  search: 'import("./rust-crypto")',
+                                  replace: 'import("./rust-crypto/index.js")',
+                              },
+                          ],
+            },
+        ],
```
Johennes, line 47: Might be worth to add a comment somewhere that this can be removed once `rewriteImportExtensions` supports dynamic imports (https://github.com/babel/babel/issues/16750).

babel.config.cjs
```diff
@@ -31,5 +31,19 @@ module.exports = {
         "@babel/plugin-transform-object-rest-spread",
         "@babel/plugin-syntax-dynamic-import",
         "@babel/plugin-transform-runtime",
+        [
+            "search-and-replace",
+            {
+                rules:
+                    process.env.NODE_ENV === "test"
```
dbkr, line 38: Why only "test"?

babel.config.cjs
```diff
@@ -31,5 +31,19 @@ module.exports = {
...
+                        ? []
+                        : [
+                              {
+                                  search: 'import("./rust-crypto")',
+                                  replace: 'import("./rust-crypto/index.js")',
```
dbkr, line 43: I'm a bit confused to be honest. So this is necessary because in nodejs environments we need one import style and elsewhere we need another, and there isn't one that works everywhere? Surely this is only evaluated at js-sdk build time so we'd just end up with one or the other depending on how the js-sdk gets babel-ed?

package.json
```diff
@@ -91,6 +91,7 @@
         "@typescript-eslint/eslint-plugin": "^7.0.0",
         "@typescript-eslint/parser": "^7.0.0",
         "babel-jest": "^29.0.0",
+        "babel-plugin-search-and-replace": "^1.1.1",
```
dbkr, line 94: I'd expect lockfile changes if we're adding a dependency.

dbkr:
Requested review from crypto too in case they have any input (although I suspect it's mostly a js-level issue)

package.json
saul-jb: Quite right, added now.

babel.config.cjs
richvdh: For the same reason as [develop/src/matrix.ts#L133](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/matrix.ts?rgh-link-date=2024-09-06T13%3A56%3A21Z#L133), I assume, though this certainly needs clarifying with a code comment.

babel.config.cjs
saul-jb: Node ESM requires that the full path be set:
https://nodejs.org/api/esm.html#esm_mandatory_file_extensions

Jest tends not to play with ESM that well, so AFAIK your tests are needing `.ts` extensions which of course will need to be re-written when compiling to JS. The babel re-write of imports don't work on dynamic imports (yet) so that is why the extension was omitted.

babel.config.cjs
richvdh: (also, I'd find it easier to grok if the condition were reversed: a list of substitutions to perform when `process.env.NODE_ENV !== "test"`)

babel.config.cjs
richvdh: This does look plausible to me, though again I think it needs code comments to explain it to future readers.

I think basically it's the same thing we ended up with in https://github.com/matrix-org/matrix-js-sdk/pull/4377: basically, whenever we're using the transpiled javascript in `lib`, we want to be using `*.js` imports: it's the only thing that keeps node.js happy. tsc is also happy with `.js` imports if it sees them: it will magically use a `.ts` instead of `.js`. The fly in the ointment is jest, which takes the original typescript source and feeds it through babel on the fly: it does not have the `.js`->`.ts` intelligence, so doesn't find the right typescript files.

@saul-jb: if you could try to explain some of this in a code comment, that would be very helpful.

babel.config.cjs
saul-jb: Alright, I've cleaned that up.

saul-jb:
I've just updated it to use the full path with the `.ts` extension and removed the eslint exception, since we are replacing it anyway. The 'blocked' comment has been removed in favor of having the documentation in the babel config.

richvdh (approved):
This looks good to me.

@saul-jb please could you add a sign-off to your PR, per https://github.com/element-hq/element-web/blob/develop/CONTRIBUTING.md#sign-off

richvdh:
(Also: many thanks for doing battle with typescript, babel, jest and nodejs to fix this!)

MidhunSureshR (approved):

PR #4447, opened 2024-10-10
Prepare delayed call leave events more reliably

AndrewFerr:
- Try sending call join after preparing delayed leave
- On leave, send delayed leave instead of a new event

## Checklist

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -1111,7 +1111,7 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
         this.memberEventTimeout = setTimeout(this.delayDisconnection, 5000);
     }
 
-    private delayDisconnection = async (): Promise<void> => {
+    private readonly delayDisconnection = async (): Promise<void> => {
         try {
             await this.client._unstable_updateDelayedEvent(this.disconnectDelayId!, UpdateDelayedEventAction.Restart);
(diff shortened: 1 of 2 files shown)


src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -1035,12 +1042,65 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
+                // TODO: Remove this once MSC4140 is stable & doesn't cancel own delayed state
+                if (this.disconnectDelayId !== undefined) {
+                    try {
+                        await this.client._unstable_updateDelayedEvent(
+                            this.disconnectDelayId,
+                            UpdateDelayedEventAction.Restart,
+                        );
+                    } catch (e) {
+                        // TODO: Retry if rate-limited
+                        if ((<MatrixError>e).errcode === "M_NOT_FOUND") {
```
AndrewFerr, line 1069: This doesn't work in embedded mode! The widget API apparently doesn't preserve the `errcode` or HTTP response code...

robintown (approved):

src/matrixrtc/MatrixRTCSession.ts
robintown: Might be worth a comment in-code then

AndrewFerr: Fixed by b070043, which also adds a comment.

robintown (approved):

PR #4460, opened 2024-10-16
Optimise MatrixRTC sender key distribution for case of single member

hughns:
Based on top of https://github.com/matrix-org/matrix-js-sdk/pull/4453.

Optimises two scenarios:

- user is only member of call, then other member(s) joins => the user can immediately rotate the key and start using it.
- there were two or more members in the call, then the other members left => the user can immediately rotate their key and start using it.

The "doesn't re-send key immediately" test is superseded by the other test cases.

## Checklist


fkwp:
> user is only member of call, then one other member joins => the user can immediately rotate the key and start using it.

wondering how that would behave in case of to-device messages. In that case you can only start sharing/sending a key with N=2 -- right?

> there were two members in the call, then the other member left => the user can immediately rotate their key and start using it.

I think rotating in that case is sane -- independent of a consumer just to be on the safe side. However, that "intermediate" key cannot be really used since there is nobody to send the key to in case of to-device messages.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -792,11 +794,21 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
...
             if (anyLeft) {
-                logger.debug(`Member(s) have left: queueing sender key rotation`);
-                this.makeNewKeyTimeout = setTimeout(this.onRotateKeyTimeout, MAKE_KEY_DELAY);
+                if (newMembershipIds.size === 1) {
+                    logger.debug(`New member(s) have left: doing immediate sender key rotation`);
+                    this.doRotateKey(false);
+                } else {
+                    logger.debug(`New member(s) have left: queueing sender key rotation`);
+                    this.makeNewKeyTimeout = setTimeout(() => this.doRotateKey(true), MAKE_KEY_DELAY);
+                }
```
robintown, line 803: It would be useful to have a comment around this area about why one would/wouldn't want there to be a delay on key rotation

robintown (approved):

src/matrixrtc/MatrixRTCSession.ts
hughns: So, I went to add the comments and I found myself reworking the logic and catering for more scenarios. It will need a re-review.

hughns:
I've converted to draft as this will collide with https://github.com/matrix-org/matrix-js-sdk/pull/4472

PR #4462, opened 2024-10-18
Add option for MatrixClient.initRustCrypto() to disable tracing.

### LINKED ISSUE #4177: Add option for MatrixClient.initRustCrypto() to disable tracing
The tracing option is by default turned on in the initRustCrypto function (rust-crypto module). This results in excessive logging information with no control over the logging levels.
```Typescript
new RustSdkCryptoJs.Tracing(RustSdkCryptoJs.LoggerLevel.Debug).turnOn(); // Line 94 in src/rust-crypto/index.ts
```

When MatrixClient.initRustCrypto() calls it, it does not have options to disable tracing or adjust its logging level. In the Rust SDK, this tracing/logging level can be set with an environment variable. In matrix-js-sdk, however, it seems to be hard-coded.

Can we either use a similar environment variable, or have it as an optional parameter in the init function?

Srayash:
## Checklist

## Fixes
fixes #4177 

## Description
This PR introduces an optional parameter, tracingEnabled, to the initRustCrypto function, allowing users to enable or disable tracing in the Rust crypto SDK.

Now, the tracing feature can be toggled:

Enabled: tracingEnabled: true - Tracing will capture debug-level logs.
Disabled: tracingEnabled: false - No tracing will occur, leading to cleaner log output and improved performance.

Signed-off-by: Srayash <srayashsingh995@gmail.com>

### DIFF
--- a/src/rust-crypto/index.ts
+++ b/src/rust-crypto/index.ts
@@ -84,22 +84,29 @@ export async function initRustCrypto(args: {
 
     /** The pickle key for `legacyCryptoStore` */
     legacyPickleKey?: string;
-
     /**
      * A callback which will receive progress updates on migration from `legacyCryptoStore`.
      *
      * Called with (-1, -1) to mark the end of migration.
      */
     legacyMigrationProgressListener?: (progress: number, total: number) => void;
+    /**
+     * Whether to enable tracing. Defaults to true.
+     */
+    tracingEnabled?: boolean;
 }): Promise<RustCrypto> {
-    const { logger } = args;
+    const { logger, tracingEnabled = true } = args;
 
     // initialise the rust matrix-sdk-crypto-wasm, if it hasn't already been done
     logger.debug("Initialising Rust crypto-sdk WASM artifact");
     await RustSdkCryptoJs.initAsync();
 
-    // enable tracing in the rust-sdk
-    new RustSdkCryptoJs.Tracing(RustSdkCryptoJs.LoggerLevel.Debug).turnOn();
+    // Enable tracing in the rust-sdk based on the parameter
+    if (tracingEnabled) {
+        new RustSdkCryptoJs.Tracing(RustSdkCryptoJs.LoggerLevel.Debug).turnOn();
+    } else {
+        new RustSdkCryptoJs.Tracing(RustSdkCryptoJs.LoggerLevel.Off).turnOn();
+    }
 
     logger.debug("Opening Rust CryptoStore");
     let storeHandle;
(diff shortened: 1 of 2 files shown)


andybalaam (approved):
The change looks ok to me, but I want to defer to the other reviewers as to whether this the right way to do this.

andybalaam:
This error from the build looks genuine though:

> src/rust-crypto/index.ts:108:65 - error TS2339: Property 'Off' does not exist on type 'typeof LoggerLevel'.
> 
> 108         new RustSdkCryptoJs.Tracing(RustSdkCryptoJs.LoggerLevel.Off).turnOn();
>                                                                     ~~~

andybalaam (changes requested):
Changes needed to fix the build

Srayash:
hey @andybalaam the error can be fixed in two ways:
- Using LoggerLevel.Error to minimize logs.
- skip .turnOn() when tracing is disabled or modify rust enum in `matrix-rust-sdk-crypto-wasm/src/tracing.rs` to include an `off` level.

Srayash:
> Changes needed to fix the build

Hi @andybalaam ,

I have made the requested changes by skipping `.turnOn()` when tracing is disabled

Please let me know if further adjustments are needed or if everything looks good now.

Thanks!

andybalaam:
> I have made the requested changes by skipping `.turnOn()` when tracing is disabled

Sounds to me like the right change.

> Please let me know if further adjustments are needed or if everything looks good now.

I just kicked off the tests but I am strongly guessing they will fail after your latest change.

andybalaam (changes requested):
Tests fail because they test for the old behaviour before the recent change.

richvdh (commented):
disabling logging altogether seems like a very bad idea. How will you diagnose any issues that occur?

src/rust-crypto/index.ts
```diff
@@ -84,22 +84,29 @@ export async function initRustCrypto(args: {
...
     await RustSdkCryptoJs.initAsync();
 
-    // enable tracing in the rust-sdk
-    new RustSdkCryptoJs.Tracing(RustSdkCryptoJs.LoggerLevel.Debug).turnOn();
+    // Enable tracing in the rust-sdk based on the parameter
+    if (tracingEnabled) {
+        new RustSdkCryptoJs.Tracing(RustSdkCryptoJs.LoggerLevel.Debug).turnOn();
+    } else {
+        // Do not initialize tracing when disabled
+    }
```
richvdh, line 109: the problem with this is that incorrectly gives the impression that tracing can be enabled or disabled at the level of an individual instance of `RustCrypto`.

For example: if I call `initRustCrypto({tracingEnabled: true})`, and then later, for a different MatrixClient instance, call `initRustCrypto({tracingEnabled: false})`, tracing will be enabled for the second instance despite `tracngEnabled` being set to false.

What we really need is to do the work in `matrix-sdk-crypto-wasm` to connect the logging from the rust tracing system to `logger`, so that it can be correctly enabled or disabled via the js-sdk's logging API.

PR #4471, opened 2024-10-24
Capture HTTP error response headers & handle Retry-After header (MSC4041)

AndrewFerr:
MSC4041 

## Checklist

### DIFF
--- a/spec/unit/http-api/utils.spec.ts
+++ b/spec/unit/http-api/utils.spec.ts
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Matrix.org Foundation C.I.C.
+Copyright 2022 - 2024 The Matrix.org Foundation C.I.C.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/spec/unit/http-api/utils.spec.ts
+++ b/spec/unit/http-api/utils.spec.ts
@@ -192,33 +192,29 @@ describe("parseErrorResponse", () => {
         headers.set("x-empty", "");
         headers.set("x-multi", "1");
         headers.append("x-multi", "2");
-    };
+    }
 
     function compareHeaders(expectedHeaders: Headers, otherHeaders: Headers | undefined) {
         expect(new Map(otherHeaders)).toEqual(new Map(expectedHeaders));
-    };
+    }
 
     it("should resolve HTTP Errors from XHR with headers", () => {
         headers.set("Content-Type", "text/plain");
         addHeaders(headers);
-        const err = parseErrorResponse(
-            {
-                ...xhrHeaderMethods,
-                status: 500,
-            } as XMLHttpRequest,
-        ) as HTTPError;
+        const err = parseErrorResponse({
+            ...xhrHeaderMethods,
+            status: 500,
+        } as XMLHttpRequest) as HTTPError;
         compareHeaders(headers, err.httpHeaders);
     });
 
     it("should resolve HTTP Errors from fetch with headers", () => {
         headers.set("Content-Type", "text/plain");
         addHeaders(headers);
-        const err = parseErrorResponse(
-            {
-                headers,
-                status: 500,
-            } as Response,
-        ) as HTTPError;
+        const err = parseErrorResponse({
+            headers,
+            status: 500,
+        } as Response) as HTTPError;
         compareHeaders(headers, err.httpHeaders);
     });
 
--- a/spec/unit/http-api/utils.spec.ts
+++ b/spec/unit/http-api/utils.spec.ts
@@ -186,62 +186,64 @@ describe("parseErrorResponse", () => {
         );
     });
 
-    function addHeaders(headers: Headers) {
-        headers.set("Age", "0");
-        headers.set("Date", "Thu, 01 Jan 1970 00:00:00 GMT"); // value contains colons
-        headers.set("x-empty", "");
-        headers.set("x-multi", "1");
-        headers.append("x-multi", "2");
-    }
+    describe("with HTTP headers", () => {
+        function addHeaders(headers: Headers) {
+            headers.set("Age", "0");
+            headers.set("Date", "Thu, 01 Jan 1970 00:00:00 GMT"); // value contains colons
+            headers.set("x-empty", "");
+            headers.set("x-multi", "1");
+            headers.append("x-multi", "2");
+        }
 
-    function compareHeaders(expectedHeaders: Headers, otherHeaders: Headers | undefined) {
-        expect(new Map(otherHeaders)).toEqual(new Map(expectedHeaders));
-    }
+        function compareHeaders(expectedHeaders: Headers, otherHeaders: Headers | undefined) {
+            expect(new Map(otherHeaders)).toEqual(new Map(expectedHeaders));
+        }
 
-    it("should resolve HTTP Errors from XHR with headers", () => {
-        headers.set("Content-Type", "text/plain");
-        addHeaders(headers);
-        const err = parseErrorResponse({
-            ...xhrHeaderMethods,
-            status: 500,
-        } as XMLHttpRequest) as HTTPError;
-        compareHeaders(headers, err.httpHeaders);
-    });
-
-    it("should resolve HTTP Errors from fetch with headers", () => {
-        headers.set("Content-Type", "text/plain");
-        addHeaders(headers);
-        const err = parseErrorResponse({
-            headers,
-            status: 500,
-        } as Response) as HTTPError;
-        compareHeaders(headers, err.httpHeaders);
-    });
-
-    it("should resolve Matrix Errors from XHR with headers", () => {
-        headers.set("Content-Type", "application/json");
-        addHeaders(headers);
-        const err = parseErrorResponse(
-            {
+        it("should resolve HTTP Errors from XHR with headers", () => {
+            headers.set("Content-Type", "text/plain");
+            addHeaders(headers);
+            const err = parseErrorResponse({
                 ...xhrHeaderMethods,
                 status: 500,
-            } as XMLHttpRequest,
-            '{"errcode": "TEST"}',
-        ) as MatrixError;
-        compareHeaders(headers, err.httpHeaders);
-    });
+            } as XMLHttpRequest) as HTTPError;
+            compareHeaders(headers, err.httpHeaders);
+        });
 
-    it("should resolve Matrix Errors from fetch with headers", () => {
-        headers.set("Content-Type", "application/json");
-        addHeaders(headers);
-        const err = parseErrorResponse(
-            {
+        it("should resolve HTTP Errors from fetch with headers", () => {
+            headers.set("Content-Type", "text/plain");
+            addHeaders(headers);
+            const err = parseErrorResponse({
                 headers,
                 status: 500,
-            } as Response,
-            '{"errcode": "TEST"}',
-        ) as MatrixError;
-        compareHeaders(headers, err.httpHeaders);
+            } as Response) as HTTPError;
+            compareHeaders(headers, err.httpHeaders);
+        });
+
+        it("should resolve Matrix Errors from XHR with headers", () => {
+            headers.set("Content-Type", "application/json");
+            addHeaders(headers);
+            const err = parseErrorResponse(
+                {
+                    ...xhrHeaderMethods,
+                    status: 500,
+                } as XMLHttpRequest,
+                '{"errcode": "TEST"}',
+            ) as MatrixError;
+            compareHeaders(headers, err.httpHeaders);
+        });
+
+        it("should resolve Matrix Errors from fetch with headers", () => {
+            headers.set("Content-Type", "application/json");
+            addHeaders(headers);
+            const err = parseErrorResponse(
+                {
+                    headers,
+                    status: 500,
+                } as Response,
+                '{"errcode": "TEST"}',
+            ) as MatrixError;
+            compareHeaders(headers, err.httpHeaders);
+        });
     });
 
     it("should set a sensible default error message on MatrixError", () => {
--- a/spec/unit/http-api/utils.spec.ts
+++ b/spec/unit/http-api/utils.spec.ts
@@ -86,16 +86,28 @@ describe("anySignal", () => {
 });
 
 describe("parseErrorResponse", () => {
+    let headers: Headers;
+    const xhrHeaderMethods = {
+        getResponseHeader: (name: string) => headers.get(name),
+        getAllResponseHeaders: () => {
+            let allHeaders = "";
+            headers.forEach((value, key) => {
+                allHeaders += `${key.toLowerCase()}: ${value}\r\n`;
+            });
+            return allHeaders;
+        },
+    };
+
+    beforeEach(() => {
+        headers = new Headers();
+    });
+
     it("should resolve Matrix Errors from XHR", () => {
+        headers.set("Content-Type", "application/json");
         expect(
             parseErrorResponse(
                 {
-                    getResponseHeader(name: string): string | null {
-                        return name === "Content-Type" ? "application/json" : null;
-                    },
-                    getAllResponseHeaders() {
-                        return "Content-Type: application/json";
-                    },
+                    ...xhrHeaderMethods,
                     status: 500,
                 } as XMLHttpRequest,
                 '{"errcode": "TEST"}',
@@ -111,14 +123,11 @@ describe("parseErrorResponse", () => {
     });
 
     it("should resolve Matrix Errors from fetch", () => {
+        headers.set("Content-Type", "application/json");
         expect(
             parseErrorResponse(
                 {
-                    headers: {
-                        get(name: string): string | null {
-                            return name === "Content-Type" ? "application/json" : null;
-                        },
-                    },
+                    headers,
                     status: 500,
                 } as Response,
                 '{"errcode": "TEST"}',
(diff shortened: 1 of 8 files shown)


robintown (approved):
Makes sense to me though the PR could use a descriptive title

src/http-api/utils.ts
```diff
@@ -122,6 +134,26 @@ function getResponseContentType(response: XMLHttpRequest | Response): ParsedMedi
...
+        const ms = Number.parseInt(retryAfter) * 1000;
+        if (!Number.isFinite(ms)) {
+            throw new Error("numeric value is too large");
+        }
+        return ms;
+    }
+    const date = new Date(retryAfter);
+    if (date.toUTCString() !== retryAfter) {
+        throw new Error("value does not match Date HTTP header syntax");
+    }
```
richvdh, line 153: somewhat wondering if this bit is necessary, but I don't have a strong opinion

src/http-api/errors.ts
```diff
@@ -63,11 +68,32 @@ export class MatrixError extends HTTPError {
...
-        super(`MatrixError: ${message}`, httpStatus);
+        super(`MatrixError: ${message}`, httpStatus, httpHeaders);
         this.errcode = errorJson.errcode;
         this.name = errorJson.errcode || "Unknown error code";
         this.data = errorJson;
     }
+
+    /**
+     * @returns the recommended delay in milliseconds to wait before retrying
+     * the request that triggered this error, or null if no delay is recommended.
```
richvdh, line 79: it feels like this method should distinguish between "This error doesn't look like it should have a retry-after value" (ie, it's not a 429 nor an M_LIMIT_EXCEEDED) and "This error looks like it should have a retry-after value, but it doesn't have one)

src/http-api/utils.ts
```diff
@@ -79,17 +79,29 @@ export function parseErrorResponse(response: XMLHttpRequest | Response, body?: s
...
-            isXhr(response) ? response.responseURL : response.url,
-        );
-    }
-    if (contentType?.type === "text/plain") {
-        return new HTTPError(`Server returned ${response.status} error: ${body}`, response.status);
-    }
-    return new HTTPError(`Server returned ${response.status} error`, response.status);
+    const httpHeaders = !isXhr(response)
+        ? response.headers
+        : new Headers(
```
richvdh, line 84: this would probably be clearer the other way around?

```
    const httpHeaders = isXhr(response)
        ? new Headers( ... )
        : response.headers;
```

src/http-api/utils.ts
```diff
@@ -79,17 +79,29 @@ export function parseErrorResponse(response: XMLHttpRequest | Response, body?: s
...
+              response
+                  .getAllResponseHeaders()
+                  ?.trim()
+                  .split(/[\r\n]+/)
+                  .map((h) => h.split(": ") as [string, string]),
```
richvdh, line 89: won't this break if there is a `: ` in one of the header values?

spec/unit/http-api/utils.spec.ts
```diff
@@ -93,6 +93,9 @@ describe("parseErrorResponse", () => {
                     getResponseHeader(name: string): string | null {
                         return name === "Content-Type" ? "application/json" : null;
                     },
+                    getAllResponseHeaders() {
+                        return "Content-Type: application/json";
```
richvdh, line 97: NB that XHR's getAllResponseHeaders is specced to lower-case the header names

```suggestion
                        return "content-type: application/json";
```

spec/unit/http-api/utils.spec.ts
```diff
@@ -93,6 +93,9 @@ describe("parseErrorResponse", () => {
...
+                    },
                     status: 500,
                 } as XMLHttpRequest,
                 '{"errcode": "TEST"}',
```
richvdh, line 101: would be good to add some more thorough testing of that header-parsing magic here. Eg: what happens if there are multiple header lines? Or if there are `: ` strings in the header values?

src/http-api/utils.ts
```diff
@@ -79,17 +79,29 @@ export function parseErrorResponse(response: XMLHttpRequest | Response, body?: s
...
-            JSON.parse(body),
-            response.status,
-            isXhr(response) ? response.responseURL : response.url,
-        );
-    }
-    if (contentType?.type === "text/plain") {
-        return new HTTPError(`Server returned ${response.status} error: ${body}`, response.status);
-    }
-    return new HTTPError(`Server returned ${response.status} error`, response.status);
+    const httpHeaders = !isXhr(response)
```
richvdh, line 82: maybe move the call to `getResponseContentType` after this, so that it can operate on the resolved Headers value rather than re-engineering it?

src/http-api/utils.ts
```diff
@@ -79,17 +79,29 @@ export function parseErrorResponse(response: XMLHttpRequest | Response, body?: s
...
+        ? response.headers
+        : new Headers(
+              response
+                  .getAllResponseHeaders()
+                  ?.trim()
+                  .split(/[\r\n]+/)
+                  .map((h) => h.split(": ") as [string, string]),
+          );
+
+    return contentType?.type === "application/json" && body
```
richvdh, line 92: I found the `if` statements much easier to folllow than the nested `?` operators, fwiw.

src/http-api/errors.ts
```diff
@@ -28,11 +29,13 @@ interface IErrorJson extends Partial<IUsageLimit> {
  * specific to HTTP responses.
  * @param msg - The error message to include.
  * @param httpStatus - The HTTP response status code.
+ * @param httpHeaders - The HTTP response headers.
  */
 export class HTTPError extends Error {
     public constructor(
         msg: string,
         public readonly httpStatus?: number,
+        public readonly httpHeaders?: Headers,
```
richvdh, line 38: afaict this is only used in a couple of places, and `httpStatus` and `httpHeaders` are set in both. Do we really need to make them optional?

richvdh (changes requested):
Looks broadly sensible, though I think it could stand to be cleaned up somewhat.

I've made quite a few comments here; feel free to push back on any you don't agree with! I think there's at least one actual bug though, so that needs a test at the very least.

> though the PR could use a descriptive title

+1 to this. And a descriptive description. Is this related to a specific change in the spec? Don't make reviewers, and future archaeologists, reverse-engineer what it is you're trying to change.

src/http-api/errors.ts
```diff
@@ -49,12 +52,14 @@ export class MatrixError extends HTTPError {
...
      * @param errorJson - The Matrix error JSON returned from the homeserver.
      * @param httpStatus - The numeric HTTP status code given
+     * @param httpHeaders - The HTTP response headers given
      */
     public constructor(
         errorJson: IErrorJson = {},
         public readonly httpStatus?: number,
         public url?: string,
         public event?: MatrixEvent,
+        public readonly httpHeaders?: Headers,
```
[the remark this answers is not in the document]
AndrewFerr: Ah true, this shadows `HttpError`'s member, which was not my intention.

`httpStatus` probably shouldn't be shadowing either, but changing it could break existing behaviour, so I'm inclined for it to be tackled it in another PR.  (I'm also not a fan of how `errcode` is redundant & separate from `data.errcode`.)

src/http-api/errors.ts
```diff
@@ -63,11 +68,32 @@ export class MatrixError extends HTTPError {
...
+        super(`MatrixError: ${message}`, httpStatus, httpHeaders);
         this.errcode = errorJson.errcode;
         this.name = errorJson.errcode || "Unknown error code";
         this.data = errorJson;
     }
+
+    /**
+     * @returns the recommended delay in milliseconds to wait before retrying
+     * the request that triggered this error, or null if no delay is recommended.
+     * @remarks Callers should first check that {@link errcode} is "M_LIMIT_EXCEEDED".
```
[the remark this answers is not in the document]
AndrewFerr: IMO the spec implies that a non-`M_LIMIT_EXCEEDED` errcode means that a retry shouldn't be attempted, as it's said that the errcode supersedes the HTTP status code (unless the errcode is `M_UNKNOWN`).

src/http-api/utils.ts
```diff
@@ -122,6 +134,26 @@ function getResponseContentType(response: XMLHttpRequest | Response): ParsedMedi
     }
 }
 
+/**
+ * Parse a Retry-After header value and convert it into a relative delay in milliseconds.
+ * @see {@link https://www.rfc-editor.org/rfc/rfc9110#section-10.2.3-2}
+ * @throws Error if the provided value is not a valid Date HTTP header value
```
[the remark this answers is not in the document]
AndrewFerr: D'oh, yes; a Date is only one of the possible formats.

src/http-api/utils.ts
AndrewFerr: I put this in lest future server developers return Retry-After with dates formatted in some other way and get surprised that they aren't supported.

src/http-api/errors.ts
AndrewFerr: Good idea, there can be another method for that.  The reason this one doesn't check the errcode is because callers tend to have an if-else chain on the errcode before checking `retry_after_ms`, so putting the check here too would be redundant.

src/http-api/errors.ts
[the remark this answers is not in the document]
richvdh: > httpStatus probably shouldn't be shadowing either, but changing it could break existing behaviour

I don't think it can?  I don't think it *actually* shadows `HttpError`'s member -- it just *looks* like it does.

Happy for you to punt that change to another PR though, in any case.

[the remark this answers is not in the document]
AndrewFerr: Yes, you're right, there's no shadowing here after all -- a parameter property named the same as a base class property ends up referring to the same property, not a distinct one.  I was stuck in a C++ mindset where base class members are distinct from subclass members.

src/http-api/errors.ts
AndrewFerr: Since this is an exported class, I left this optional so as to not break API compatibility.

On that note, it's worth considering an API version bump to make them mandatory (as well as `errcode` and `error` in `MatrixError`, which the spec says are mandatory).

richvdh: I don't think it's correct that this class is exported outside the js-sdk?

richvdh: oh, cancel that, it does seem to be. Ugh.

richvdh: I would vote for an API version bump. We don't want to have to be coding around it not being populated forever.

src/http-api/errors.ts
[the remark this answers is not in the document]
AndrewFerr: 6a55da0a2

src/http-api/errors.ts
```diff
@@ -63,11 +68,32 @@ export class MatrixError extends HTTPError {
...
+            if (retryAfter != null) {
+                return parseRetryAfterMs(retryAfter);
+            }
+        }
+        if ("retry_after_ms" in this.data) {
+            if (!Number.isInteger(this.data.retry_after_ms)) {
+                throw new Error("retry_after_ms is not an integer");
+            }
+            return this.data.retry_after_ms;
+        }
```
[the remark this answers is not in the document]
AndrewFerr: b2f1ab59e

src/http-api/errors.ts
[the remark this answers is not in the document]
AndrewFerr: This now looks up the header value regardless of the errcode & status code (though `retry_after_ms` is looked up only for `M_LIMIT_EXCEEDED`).

src/http-api/errors.ts
AndrewFerr: b31179a67

src/http-api/utils.ts
AndrewFerr: 46424f555

src/http-api/utils.ts
AndrewFerr: 03af59df5

spec/unit/http-api/utils.spec.ts
AndrewFerr: 9ba714df9

spec/unit/http-api/utils.spec.ts
AndrewFerr: The refactored tests avoid case-sensitivity issues.

src/http-api/utils.ts
AndrewFerr: 9ba714df9

src/http-api/errors.ts
AndrewFerr: For the time being I left this as it is, as optional values don't hinder much & are still used internally.  With that said, I'm in favour of a dedicated PR for making them mandatory.

richvdh:
(added a link in the description to the MSC that this is implementing)

src/http-api/errors.ts
```diff
@@ -63,11 +67,66 @@ export class MatrixError extends HTTPError {
...
+     */
+    public isRateLimitError(): boolean {
+        return this.errcode === "M_LIMIT_EXCEEDED" || (this.errcode === "M_UNKNOWN" && this.httpStatus === 429);
+    }
+
+    /**
+     * @returns the recommended delay in milliseconds to wait before retrying
+     * the request that triggered this error, or null if no delay is recommended.
+     * @throws Error if the recommended delay is an invalid value.
+     */
```
richvdh, line 87: Maybe add a link to `safeGetRetryAfterMs` here, to help people find it

src/http-api/errors.ts
```diff
@@ -63,11 +67,66 @@ export class MatrixError extends HTTPError {
...
+ * delay is recommended.
+ */
+export function safeGetRetryAfterMs(error: unknown, defaultMs: number): number {
+    if (!(error instanceof MatrixError) || !error.isRateLimitError()) {
+        return defaultMs;
+    }
+    try {
+        return error.getRetryAfterMs() ?? defaultMs;
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    } catch (_) {
```
richvdh, line 127:
```suggestion
    } catch {
```

src/http-api/errors.ts
```diff
@@ -63,11 +67,66 @@ export class MatrixError extends HTTPError {
...
+
+/**
+ * @returns the recommended delay in milliseconds to wait before retrying
+ * the request that triggered {@link error}, or {@link defaultMs} if no valid
+ * delay is recommended.
+ */
+export function safeGetRetryAfterMs(error: unknown, defaultMs: number): number {
+    if (!(error instanceof MatrixError) || !error.isRateLimitError()) {
+        return defaultMs;
+    }
```
richvdh, line 123: Why is the check on `isRateLimitError` necessary? If the error is not due to rate-limiting, then we would expect `error.getRetryAfterMs()` to return null, and therefore we fall back to `defaultMs` anyway

src/http-api/utils.ts
richvdh: > I put this in lest future server developers return Retry-After with dates formatted in some other way and get surprised that they aren't supported.

I think the surprise would be that the js-sdk *does* support it (whilst other client implementations might not)?

Anyway, being conservative about what we accept makes sense to me. :+1:

src/rust-crypto/backup.ts
```diff
@@ -460,10 +460,17 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
                             return;
                         } else if (errCode == "M_LIMIT_EXCEEDED") {
```
richvdh, line 461:
```suggestion
                        } else if (err.isRateLimitError()) {
```

spec/unit/http-api/utils.spec.ts
```diff
@@ -184,6 +186,64 @@ describe("parseErrorResponse", () => {
...
+        addHeaders(headers);
+        const err = parseErrorResponse(
+            {
+                headers,
+                status: 500,
+            } as Response,
+            '{"errcode": "TEST"}',
+        ) as MatrixError;
+        compareHeaders(headers, err.httpHeaders);
+    });
```
richvdh, line 245: suggest sticking all this in a `describe` block to keep the tests together with the helpers.

richvdh (commented):
This is looking better, thanks.

I'd suggest adding `getRetryAfterMs` and `isRateLimitError` to HttpError. Otherwise, just a few nits.

src/http-api/errors.ts
AndrewFerr: Good point.  I had done this because of the rule to give the errcode precedence over the HTTP status code, but as long as a `Retry-Header` value is there & callers want to apply _some_ delay regardless of the errcode, they may as well use the header value if one is present.

src/http-api/errors.ts
```diff
@@ -74,19 +73,36 @@ export class MatrixError extends HTTPError {
         this.data = errorJson;
     }
 
+    /**
+     * @returns whether this error is due to rate-limiting.
+     */
+    public isRateLimitError(): boolean {
+        return this.errcode === "M_LIMIT_EXCEEDED" || (this.errcode === "M_UNKNOWN" && this.httpStatus === 429);
```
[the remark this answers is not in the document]
AndrewFerr: e565d5da3 + 719fdf385

src/http-api/errors.ts
```diff
@@ -63,11 +67,66 @@ export class MatrixError extends HTTPError {
...
+    }
+
+    /**
+     * @returns the recommended delay in milliseconds to wait before retrying
+     * the request that triggered this error, or null if no delay is recommended.
+     * @throws Error if the recommended delay is an invalid value.
+     */
+    public getRetryAfterMs(): number | null {
```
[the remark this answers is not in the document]
AndrewFerr: e565d5da3

src/http-api/errors.ts
AndrewFerr: Done in e565d5da3

src/rust-crypto/backup.ts
AndrewFerr: 9c56d6dd4

spec/unit/http-api/utils.spec.ts
AndrewFerr: 91ffc9182

richvdh (approved):
Ok then. Thanks for your perseverance!

PR #4476, opened 2024-10-29
Add `restoreKeybackup` to `CryptoApi`.

florianduros:
## Checklist

The goal is to simplify the `restoreKeyBackup` api and move it out of the `MatrixClient` to the `CryptoApi`. 

- Added two methods to `CryptoApi`
   - `restoreKeyBackup`
        - if a recovery key is provided, we use it otherwise we try to getting form our local cache or from the secret storage.
        - In [`RestoreKeyBackupDialog`](https://github.com/element-hq/element-web/blob/develop/src/components/views/dialogs/security/RestoreKeyBackupDialog.tsx) we never target a specific session or room when calling the `restoreKeyBackup*` consequently I get rid of the `targetRoomId` and `targetSessionId`. 
        - Also `IKeyBackupRestoreOpts.cacheCompleteCallback` is not used in `RestoreKeyBackupDialog`, I removed also this parameter.
        - Instead of having the `keyBackupInfo` as a parameter, we directly retrieve it internally.
   - `restoreKeyBackupWithPassphrase` which is deprecated because deriving a key backup from the passphrase is not in the spec.
   - Moved the key backup specific code into the `RustBackupManager.`
- Deprecate the `restoreKeyBackup*` in `MatrixClient`

### DIFF
--- a/src/crypto-api/index.ts
+++ b/src/crypto-api/index.ts
@@ -509,7 +509,7 @@ export interface CryptoApi {
      * if they match, stores the key in the crypto store by calling {@link storeSessionBackupPrivateKey},
      * which enables automatic restore of individual keys when an Unable-to-decrypt error is encountered.
      *
-     * if we are unable to fetch the key from secret storage, there is no backup on the server, or the key
+     * If we are unable to fetch the key from secret storage, there is no backup on the server, or the key
      * does not match, throws an exception.
      */
     loadSessionBackupPrivateKeyFromSecretStorage(): Promise<void>;
--- a/src/crypto-api/index.ts
+++ b/src/crypto-api/index.ts
@@ -560,7 +560,7 @@ export interface CryptoApi {
      * @param phassphrase - The passphrase to use to restore the key backup.
      * @param opts
      *
-     * @deprecated Deriving a backup key from a passphrase is not part of the matrix spec. Instead, a random key is generated and stored/ shared via 4S.
+     * @deprecated Deriving a backup key from a passphrase is not part of the matrix spec. Instead, a random key is generated and stored/shared via 4S.
      */
     restoreKeyBackupWithPassphrase(phassphrase: string, opts?: KeyBackupRestoreOpts): Promise<KeyBackupRestoreResult>;
 
--- a/src/crypto-api/index.ts
+++ b/src/crypto-api/index.ts
@@ -471,6 +471,15 @@ export interface CryptoApi {
     ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
     /**
+     * Fetch the backup decryption key we have saved in our secret storage.
+     *
+     * This can be used for gossiping the key to other devices.
+     *
+     * @returns the key, if any, or null
+     */
+    getSecretStorageBackupPrivateKey(): Promise<Uint8Array | null>;
+
+    /**
      * Fetch the backup decryption key we have saved in our store.
      *
      * This can be used for gossiping the key to other devices.
@@ -546,14 +555,12 @@ export interface CryptoApi {
     deleteKeyBackupVersion(version: string): Promise<void>;
 
     /**
-     * Restores a key backup.
-     * If the recovery key is not provided, it will try to restore the key backup using the recovery key stored
-     * in the local cache or in the Secret Storage.
+     * Download the last key backup from the homeserver (endpoint GET /room_keys/keys/).
+     * The key backup is decrypted and imported by using the decryption key stored locally. The decryption key should be stored locally by using {@link CryptoApi#storeSessionBackupPrivateKey}.
      *
-     * @param recoveryKey - The recovery key to use to restore the key backup.
      * @param opts
      */
-    restoreKeyBackup(recoveryKey: string | undefined, opts?: KeyBackupRestoreOpts): Promise<KeyBackupRestoreResult>;
+    restoreKeyBackup(opts?: KeyBackupRestoreOpts): Promise<KeyBackupRestoreResult>;
 
     /**
      * Restores a key backup using a passphrase.
--- a/src/crypto-api/index.ts
+++ b/src/crypto-api/index.ts
@@ -22,7 +22,13 @@ import { DeviceMap } from "../models/device.ts";
 import { UIAuthCallback } from "../interactive-auth.ts";
 import { PassphraseInfo, SecretStorageCallbacks, SecretStorageKeyDescription } from "../secret-storage.ts";
 import { VerificationRequest } from "./verification.ts";
-import { BackupTrustInfo, KeyBackupCheck, KeyBackupInfo } from "./keybackup.ts";
+import {
+    BackupTrustInfo,
+    KeyBackupCheck,
+    KeyBackupInfo,
+    KeyBackupRestoreOpts,
+    KeyBackupRestoreResult,
+} from "./keybackup.ts";
 import { ISignatures } from "../@types/signed.ts";
 import { MatrixEvent } from "../models/event.ts";
 
@@ -538,6 +544,25 @@ export interface CryptoApi {
      * @param version - The backup version to delete.
      */
     deleteKeyBackupVersion(version: string): Promise<void>;
+
+    /**
+     * Restores a key backup.
+     * If the recovery key is not provided, it will try to restore the key backup using the recovery key stored
+     * in the local cache or in the Secret Storage.
+     *
+     * @param recoveryKey - The recovery key to use to restore the key backup.
+     * @param opts
+     */
+    restoreKeyBackup(recoveryKey: string | undefined, opts?: KeyBackupRestoreOpts): Promise<KeyBackupRestoreResult>;
+
+    /**
+     * Restores a key backup using a passphrase.
+     * @param phassphrase - The passphrase to use to restore the key backup.
+     * @param opts
+     *
+     * @deprecated Deriving a backup key from a passphrase is not part of the matrix spec. Instead, a random key is generated and stored/ shared via 4S.
+     */
+    restoreKeyBackupWithPassphrase(phassphrase: string, opts?: KeyBackupRestoreOpts): Promise<KeyBackupRestoreResult>;
 
     ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////
     //
--- a/src/crypto-api/index.ts
+++ b/src/crypto-api/index.ts
@@ -564,12 +564,12 @@ export interface CryptoApi {
 
     /**
      * Restores a key backup using a passphrase.
-     * @param phassphrase - The passphrase to use to restore the key backup.
+     * @param passphrase - The passphrase to use to restore the key backup.
      * @param opts
      *
      * @deprecated Deriving a backup key from a passphrase is not part of the matrix spec. Instead, a random key is generated and stored/shared via 4S.
      */
-    restoreKeyBackupWithPassphrase(phassphrase: string, opts?: KeyBackupRestoreOpts): Promise<KeyBackupRestoreResult>;
+    restoreKeyBackupWithPassphrase(passphrase: string, opts?: KeyBackupRestoreOpts): Promise<KeyBackupRestoreResult>;
 
     ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////
     //
--- a/src/crypto-api/index.ts
+++ b/src/crypto-api/index.ts
@@ -555,15 +555,18 @@ export interface CryptoApi {
     deleteKeyBackupVersion(version: string): Promise<void>;
 
     /**
-     * Download the last key backup from the homeserver (endpoint GET /room_keys/keys/).
+     * Download and restore the last key backup from the homeserver (endpoint GET /room_keys/keys/).
      * The key backup is decrypted and imported by using the decryption key stored locally. The decryption key should be stored locally by using {@link CryptoApi#storeSessionBackupPrivateKey}.
      *
+     * Warning: the full key backup may be quite large, so this operation may take several hours to complete.
+     * Use of {@link KeyBackupRestoreOpts.progressCallback} is recommended.
      * @param opts
      */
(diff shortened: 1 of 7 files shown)


richvdh:
> * Added two methods to `CryptoApi`
>   
>   * `restoreKeyBackup`
>     
>     * I went with the following logic to avoid to have to much public methods to restore a key backup: If a recovery key is provided, we use it otherwise we try to getting form our local cache or from the secret storage.

Sounds sensible.

>     * In [`RestoreKeyBackupDialog.`](https://github.com/element-hq/element-web/blob/develop/src/components/views/dialogs/security/RestoreKeyBackupDialog.tsx?rgh-link-date=2024-10-29T09%3A43%3A16Z) we never target a specific session or room when calling the `restoreKeyBackup*` consequently I get rid of the `targetRoomId` and `targetSessionId`. However since the `js-sdk` is used also by other clients, should we keep this parameters ?

No. There is no real reason for applications to be restoring keys at the room level (and certainly not at the session level). The crypto stack will fetch keys on demand. Let's drop support for those parameters.

>   * `restoreKeyBackupWithPassphrase` which is deprecated because deriving a key backup from the passphrase is not in the spec. However we are still deriving in EW so we have also to support the restoration. (But maybe we should not deprecated it ? Should we only deprecate the keybackup creation from the passphrase and not the backup restoration)

Yeah, I think that it makes sense to mark `restoreKeyBackupWithPassphrase` as deprecated.

src/rust-crypto/backup.ts
```diff
@@ -34,7 +38,7 @@ import { OutgoingRequestProcessor } from "./OutgoingRequestProcessor.ts";
 import { sleep } from "../utils.ts";
 import { BackupDecryptor } from "../common-crypto/CryptoBackend.ts";
 import { ImportRoomKeyProgressData, ImportRoomKeysOpts, CryptoEvent } from "../crypto-api/index.ts";
-import { IKeyBackupInfo } from "../crypto/keybackup.ts";
+import { IKeyBackupInfo, IKeyBackupRoomSessions } from "../crypto/keybackup.ts";
```
richvdh, line 41: this is importing stuff from the legacy crypto code?

richvdh (commented):
This looks like a very sensible start. Obviously, please make sure that any new public types and methods are clearly documented, but I think you know that.

t3chguy (commented):
Looks sane but has insufficient coverage

florianduros:
Hmm, I'll try but seems difficult since we need to support it for the old crypto until we remove it. I'm afraid that since the logic is similar between the two cryptos, we will have duplication.

florianduros (commented):
Thought: Move `RustCrypto#getBackupDecryptor` into `RustBackupManager`

src/rust-crypto/backup.ts
```diff
@@ -495,8 +497,18 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+     *
+     * @returns The total number of keys in the backup.
+     */
+    private getTotalKeyCount(keyBackup: KeyBackup): number {
         let count = 0;
-        for (const { sessions } of Object.values(parsedBody.rooms)) {
+        for (const { sessions } of Object.values(keyBackup.rooms)) {
             count += Object.keys(sessions).length;
         }
         return count;
```
richvdh, line 514: given that this is independent of `RustBackupManager`, make it a standalone utility function?

src/rust-crypto/backup.ts
```diff
@@ -591,6 +603,173 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+
+                opts?.progressCallback?.({
+                    total: totalKeyCount,
+                    successes: totalImported,
+                    stage: "load_keys",
+                    failures: totalFailures,
+                });
+            });
+        } finally {
+            backupDecryptor.free();
```
richvdh, line 694: Why is `backupDecryptor.free()` necessary?

If it *is* necessary, we'll need to document that this method (and all others that call it) destroy `backupDecryptor`.  Better, move it up to `RustCrypto.restoreKeyBackupWithKey`.

src/rust-crypto/backup.ts
```diff
@@ -591,6 +603,173 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
     public createBackupDecryptor(decryptionKey: RustSdkCryptoJs.BackupDecryptionKey): BackupDecryptor {
         return new RustBackupDecryptor(decryptionKey);
     }
+
+    /**
+     * Restore a key backup.
+     * @param backupInfoVersion - The version of the backup info.
```
richvdh, line 609: it's the version of the backup, not the version of the info.

```suggestion
     * @param backupVersion - The backup version to restore.
```

(likewise below)

src/rust-crypto/backup.ts
```diff
@@ -591,6 +603,173 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
     public createBackupDecryptor(decryptionKey: RustSdkCryptoJs.BackupDecryptionKey): BackupDecryptor {
         return new RustBackupDecryptor(decryptionKey);
     }
+
+    /**
+     * Restore a key backup.
```
richvdh, line 608: For consistency with other declarations:
```suggestion
     * Restore a key backup.
     *
```

src/rust-crypto/backup.ts
```diff
@@ -591,6 +603,173 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+                    logger.error("Error importing keys from backup", e);
+                }
+
+                opts?.progressCallback?.({
+                    total: totalKeyCount,
+                    successes: totalImported,
+                    stage: "load_keys",
+                    failures: totalFailures,
+                });
+            });
```
richvdh, line 692: Given that `handleDecryptionOfAFullBackup` is only called in one place, this callback structure feels a bit unncessary. Couldn't you just inline `handleDecryptionOfAFullBackup` ?

richvdh (changes requested):

src/rust-crypto/backup.ts
florianduros: `backupDecryptor.free()` is called in the original function. I suppose the goal is to fix a memory leak but there is no comment in the original function too. I'll add a comment but I'm not sure why we originally added it.

richvdh: oh, I didn't realise this was copied from somewhere else. Where is the original?

florianduros: https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/client.ts#L4001 you can find it here

richvdh: >  I suppose the goal is to fix a memory leak

I think more likely: backupdecryptor keeps the backup decryption key in memory; dropping it reduces the risk of another bug exposing it to an attacker.

src/rust-crypto/backup.ts
```diff
@@ -591,6 +589,148 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+    /**
+     * Restore a key backup.
+     *
+     * @param backupInfoVersion - The version of the backup to restore.
+     * @param backupDecryptor - The backup decryptor to use to decrypt the keys.
+     * @param opts - Options for the restore.
+     * @returns The total number of keys and the total imported.
+     */
+    public async restoreKeyBackup(
+        backupInfoVersion: string,
```
richvdh, line 602:
```suggestion
        backupVersion: string,
```

src/rust-crypto/backup.ts
```diff
@@ -591,6 +589,148 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
     public createBackupDecryptor(decryptionKey: RustSdkCryptoJs.BackupDecryptionKey): BackupDecryptor {
         return new RustBackupDecryptor(decryptionKey);
     }
+
+    /**
+     * Restore a key backup.
+     *
+     * @param backupInfoVersion - The version of the backup to restore.
```
richvdh, line 596:
```suggestion
     * @param backupVersion - The version of the backup to restore.
```

richvdh: likewise below.

src/rust-crypto/rust-crypto.ts
```diff
@@ -1280,6 +1292,52 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+
+        let result: KeyBackupRestoreResult;
+        try {
+            result = await this.backupManager.restoreKeyBackup(backupInfo.version, backupDecryptor, opts);
+        } finally {
+            // Free to avoid to keep in memory the decryption key stored in it. To avoid to exposing it to an attacker.
+            backupDecryptor.free();
+        }
+
+        return result;
```
richvdh, line 1338: Let's put the callback inside the try/finally, in case it throws.

Also, we don't need the temporary `result`.

```suggestion
        const backupDecryptor = await this.getBackupDecryptor(backupInfo, privateKeyFromCache);

        try {
            opts?.progressCallback?.({
                stage: "fetch",
            });

            return await this.backupManager.restoreKeyBackup(backupInfo.version, backupDecryptor, opts);
        } finally {
            // Free to avoid to keep in memory the decryption key stored in it. To avoid to exposing it to an attacker.
            backupDecryptor.free();
        }
```

richvdh (changes requested):

src/rust-crypto/rust-crypto.ts
richvdh: seriously? that works where the other doesn't? I don't understand how that is possible. Fuck, as you so succinctly put it, Javascript.

(NB it's a different case to that medium post: there is no `return` in the `finally` block. `return` in `try` blocks absolutely should be executed, whether or not there is a `finally`. They just happen *before* the finally.)

richvdh: to confirm: you did `return await this.backupManager.restoreKeyBackup`, not just `return this.backupManager.restoreKeyBackup` ?

src/crypto-api/index.ts
```diff
@@ -539,6 +554,23 @@ export interface CryptoApi {
      */
     deleteKeyBackupVersion(version: string): Promise<void>;
 
+    /**
+     * Download the last key backup from the homeserver (endpoint GET /room_keys/keys/).
```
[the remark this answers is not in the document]
richvdh: > We need the full backupInfo and not only the version to call getBackupDecryptor

Debateable: `getBackupDecryptor` only needs the full backup info to check that the key matches the backup version, and that should probably have happened when `storeSessionBackupPrivateKey` was called.

But in any case: you can get the backupInfo with a call to `GET /_matrix/client/v3/room_keys/version/{version}`

src/rust-crypto/rust-crypto.ts
florianduros: `return await this.backupManager.restoreKeyBackup` is working and not `return this.backupManager.restoreKeyBackup`. 
Both are returning a promise but in the first case, we are returning in the promise the result of `restoreKeyBackup` and in the second the call of `restoreKeyBackup` is *waiting* to be called.

So I suppose that in the first case, the finally block is called between the end of execution of `restoreKeyBackup` and the return.

richvdh: > `return await this.backupManager.restoreKeyBackup` is working and not `return this.backupManager.restoreKeyBackup`.
> Both are returning a promise but in the first case, we are returning in the promise the result of `restoreKeyBackup` and in the second the call of `restoreKeyBackup` is _waiting_ to be called.

yes. The important thing is to have an `await` within the `try` block.

src/rust-crypto/rust-crypto.ts
```diff
@@ -1280,6 +1311,54 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+     * Implementation of {@link CryptoApi#restoreKeyBackup}.
+     */
+    public async restoreKeyBackup(opts?: KeyBackupRestoreOpts): Promise<KeyBackupRestoreResult> {
+        // Get the decryption key from the cache
+        const privateKeyFromCache = await this.getSessionBackupPrivateKey();
+        if (!privateKeyFromCache) throw new Error("No decryption key found in cache");
+
+        // Get the backup version from the cache
+        const backupVersion = await this.getSessionBackupVersion();
+        if (!backupVersion) throw new Error("No backup version found in cache");
```
richvdh, line 1343: Let's avoid calling olmMachine.getBackupKeys twice (and get rid of getSessionBackupVersion).

```suggestion
        // Get the decryption key from the crypto store
        const backupKeys: RustSdkCryptoJs.BackupKeys = await this.olmMachine.getBackupKeys();
        const {decryptionKey, backupVersion} = backupKey;
        if (!decryptionKey || !backupVersion) throw new Error("No decryption key found in crypto store");
```

src/crypto-api/index.ts
```diff
@@ -555,21 +552,24 @@ export interface CryptoApi {
     deleteKeyBackupVersion(version: string): Promise<void>;
 
     /**
-     * Download the last key backup from the homeserver (endpoint GET /room_keys/keys/).
+     * Download and restore the last key backup from the homeserver (endpoint GET /room_keys/keys/).
      * The key backup is decrypted and imported by using the decryption key stored locally. The decryption key should be stored locally by using {@link CryptoApi#storeSessionBackupPrivateKey}.
```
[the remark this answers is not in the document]
florianduros: If we deprecate `storeSessionBackupPrivateKey`, how we can set the recovery key when it filled by the user?

In EW when restoring key backup, if we can't restore with cache or SSSS, we ask for the recovery key from the user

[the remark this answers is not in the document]
richvdh: I guess you mean the backup decryption key, rather than the recovery key.

I'm not sure we need to support the user entering the backup decryption key directly -- I thought most of that went away with https://github.com/element-hq/element-web/issues/27455.

But sure, let's leave this for now.

richvdh (approved):
LGTM otherwise

PR #4477, opened 2024-10-29
Make doc clearer on getCrossSigningKeyId

dbkr:
I was trying to work out why this was being used in a check. It turns out it only returns the key ID if the private part is stored locally, which seems very much non-obvious.

## Checklist

### DIFF
--- a/src/crypto-api/index.ts
+++ b/src/crypto-api/index.ts
@@ -273,7 +273,8 @@ export interface CryptoApi {
     isCrossSigningReady(): Promise<boolean>;
 
     /**
-     * Get the ID of one of the user's cross-signing keys.
+     * Get the ID of one of the user's cross-signing keys, if the private is available.
+     * If the private key is not available, this will return null.
      *
      * @param type - The type of key to get the ID of.  One of `CrossSigningKey.Master`, `CrossSigningKey.SelfSigning`,
      *     or `CrossSigningKey.UserSigning`.  Defaults to `CrossSigningKey.Master`.
--- a/src/crypto-api/index.ts
+++ b/src/crypto-api/index.ts
@@ -273,8 +273,13 @@ export interface CryptoApi {
     isCrossSigningReady(): Promise<boolean>;
 
     /**
-     * Get the ID of one of the user's cross-signing keys, if the private is available.
-     * If the private key is not available, this will return null.
+     * Get the ID of one of the user's cross-signing keys, if both private and matching
+     * public parts of that key are available (ie. cached in the local crypto store).
+     *
+     * The public part may not be available if a /keys/query request has not yet been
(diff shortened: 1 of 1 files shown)


src/crypto-api/index.ts
```diff
@@ -273,7 +273,8 @@ export interface CryptoApi {
     isCrossSigningReady(): Promise<boolean>;
 
     /**
-     * Get the ID of one of the user's cross-signing keys.
+     * Get the ID of one of the user's cross-signing keys, if the private is available.
+     * If the private key is not available, this will return null.
```
richvdh, line 277: missing "key", but it might be good to be more explicit about what "is available" means. Does "in 4S" count as "available"? (I think not?)

This is already touched on in the `@returns` tag, but I don't think "cross-signing has been initialised" is very clear either. It would be good to clarify that while you're here.

dbkr: Good question, looks like it depends what `hasMaster` etc. mean on `olmMachine.crossSigningStatus`. Do you have any idea without me delving into the rust?

richvdh: https://matrix-org.github.io/matrix-rust-sdk/matrix_sdk_crypto/olm/struct.CrossSigningStatus.html might be vaguely helpful.

It means "cached in the local crypto store"; ie the private key has been loaded from 4S or similar.

I also note that this will return null if we have the private key but not the public key (which can happen if we haven't yet done a /keys/query, or if the device that created the keys failed to publish them)

dbkr: Thanks - hopefully better now.

src/crypto-api/index.ts
```diff
@@ -273,7 +273,13 @@ export interface CryptoApi {
     isCrossSigningReady(): Promise<boolean>;
 
     /**
-     * Get the ID of one of the user's cross-signing keys.
+     * Get the ID of one of the user's cross-signing keys, if both private and matching
+     * public parts of that key are available (ie. cached in the local crypto store).
+     *
+     * The public part may not be available if a /keys/query request has not yet been
+     * performed, or if the device that created the keys failed to publish them.
```
richvdh, line 280:
```suggestion
     * The public part may not be available if a `/keys/query` request has not yet been
     * performed, or if the device that created the keys failed to publish them.
```

src/crypto-api/index.ts
```diff
@@ -273,7 +273,13 @@ export interface CryptoApi {
...
+     *
+     * If the keypair is not available, this will return null.
```
richvdh, line 282:
```suggestion
     * If either part of the keypair is not available, this will return `null`.
```

richvdh (approved):

PR #4487, opened 2024-10-30
Support MSC4222 `state_after`

dbkr:
Closes https://github.com/matrix-org/matrix-js-sdk/pull/4532
Fixes https://github.com/element-hq/element-web/issues/28536
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/4528

This implements support for `state_after` in sync responses, as per https://github.com/matrix-org/matrix-spec-proposals/pull/4222

In this, the events that arrive in the timeline should no longer be added to the state of the room. Instead, all state events arrive explicitly, duplicated, in `state_after`. Applying the events in `state_after` gives the state of the room after that sync request. This fixes problems where state events in the timeline should not actually change the state of the room, for example if old state events arrive from disconnected homeservers.

Synapse impl: https://github.com/element-hq/synapse/pull/17888

This will need aggressive testing against synapse both with and without PR 17888. It's quite an invasive change. I opted to require callers of methods that add timeline events to explicitly specify whether they want the timeline events to apply to state since defaulting to either is not really safe, and this helps us catch all the code paths where it happens.

This changes some interface in the js-sdk and so should probably be considered a breaking change (and I've removed some deprecated interfaces while I'm at it).

Changes to public functions:

- `SyncApi.injectRoomEvents()` - parameters have changed
- `Room.addLiveEvents()` - `addLiveEventOptions` parameter is now required
- `EventTimeline.addEvent()` - `toStartOfTimeline` is now required; new required option `addToState`
- `EventTimelineSet.insertEventIntoTimeline()` - new required parameter `addToState`
- `EventTimelineSet.addEventToTimeline()` - deprecated variants removed; `toStartOfTimeline` is now required; new required option `addToState`

## Checklist


src/sync.ts
```diff
@@ -1737,28 +1755,39 @@ export class SyncApi {
...
+     * @param stateAfterEventList - A list of state events. This is the state
+     * at the *END* of the timeline list if it is supplied.
      * @param timelineEventList - A list of timeline events, including threaded. Lower index
      * is earlier in time. Higher index is later.
      * @param fromCache - whether the sync response came from cache
+     *
+     * No more than one of stateEventList and stateAfterEventList must be supplied. If
+     * stateEventList is supplied, the events in timelineEventList are added to the state
+     * after stateEventList. If stateAfterEventList is supplied, the events in timelineEventList
+     * are not added to the state.
```
t3chguy, line 1775: Can we please guard against this via an overload rather than the bottom of a lengthy doc comment? Or given that it treats the list in the same way except the presence of one of the lists can we just use a param `addToState`?

src/sync-accumulator.ts
```diff
@@ -78,6 +78,7 @@ export interface ITimeline {
 export interface IJoinedRoom {
     "summary": IRoomSummary;
     "state": IState;
+    "state_after"?: IState;
```
t3chguy, line 81: Shouldn't this be unstable-prefixed in case the format changes by the time the MSC is merged and potentially corrupts the sync accumulator?

hughns: Yes, it should be. The MSC has unstable prefixes and this one slipped through. I will update it.

toger5:
This was tried in the SPA (EC with this branch directly connected to a home server)
We did run into the following issue:
 - The call member events were synced on most clients (we assume there the reliable state worked, cannot be proven of course)
 - The room members however showed incorrect bahaviours 
  - different clients had different room member counts
  - On Hughs client a room member was there and then disappeared (but still is there on other clients)
  - `room.currentState.events("m.call.member") also has those missing members (on some clients) - as does `room.currentState.getMembers()`
  - `room.currentState.events("org.matrix.msc3401.call.member") showed the same events on all clients.

toger5:
There are still issues with the local storage.
[The previous comment](https://github.com/matrix-org/matrix-js-sdk/pull/4487#issuecomment-2470339759) mentions an issue where,
> a reload on the client makes you loose all events that you have synced as part of the initial sync and came down through `state_after`

This was easily testable by reloading a client and seeing all room names to disappear.

With the fix from @t3chguy this is not the case anymore. But we still see issues after a reload (loading sync from cache).

This can be used to investigate the issue https://pr2767--element-call.netlify.app/

If you are in a call and others join and leave, you reload the client the number at the top left will not be correct. The list of `org.matrix.msc3401.call.member` also will be incorrect.

dbkr:
Thanks for continuing work on this. The changes look good as far as I can see, glad the sync accumulator wasn't too much of a big change to get working.

florianduros (approved):

spec/unit/room-state.spec.ts
```diff
@@ -1122,59 +1122,4 @@ describe("RoomState", function () {
             ).toBeFalsy();
         });
     });
-    describe("skipWrongOrderRoomStateInserts", () => {
```
toger5, line 1125: Can we make this PR close:
https://github.com/matrix-org/matrix-js-sdk/pull/4532

when merged? Since it includes the revert changes?

src/embedded.ts
```diff
@@ -284,7 +284,11 @@ export class RoomWidgetClient extends MatrixClient {
                 const rawEvents = await this.widgetApi.readStateEvents(eventType, undefined, stateKey, [this.roomId]);
                 const events = rawEvents.map((rawEvent) => new MatrixEvent(rawEvent as Partial<IEvent>));
 
-                await this.syncApi!.injectRoomEvents(this.room!, [], events);
+                if (this.syncApi instanceof SyncApi) {
+                    await this.syncApi.injectRoomEvents(this.room!, undefined, events);
+                } else {
+                    await this.syncApi!.injectRoomEvents(this.room!, events);
+                }
```
toger5, line 291: This would be nice to get a comment explaining:
What is v2 what is sliding sync?
What is the second what is the third arg?
What does `injectRoomEvents` room do in the different overloaded versions?

(This is also explained below `onEvent` but i think duplicating it here makes sense)

src/sync-accumulator.ts
```diff
@@ -650,11 +660,14 @@ export class SyncAccumulator {
...
+                    // Push to both fields to provide downgrade compatibility in the sync accumulator db
+                    // the code will prefer `state_after` if it is present
+                    roomJson["org.matrix.msc4222.state_after"]?.events.push(ev);
+                    // Roll the state back to the value at the start of the timeline if it was changed
                     if (rollBackState[evType] && rollBackState[evType][stateKey]) {
-                        // use the reverse clobbered event instead.
                         ev = rollBackState[evType][stateKey];
                     }
-                    roomJson.state.events.push(ev);
+                    roomJson.state?.events.push(ev);
```
toger5, line 670: Why do we need to make this optional? shouldnt we always be able to have a state field?
Either it being a copy of `org.matrix.msc4222.state_after` or the only place for state?

toger5 (approved):
I only have minor comments. Nice that there is also the test to the sync accumulator

richvdh (approved):
crypto changes look fine

toger5:
Is there anything else we should wait for before merging this?

t3chguy:
Yes, it will need force merging when I am done with the release process

Issue #4450, opened 2024-10-15
Error: global is not defined - when creating a matrix client


ChrisKlp:
Matrix SDK is not compatible with the latest version of Next.js (14.2.15). Creating a client causes an error in the `amendClientOpts` function, where it uses the `global` variable, which is not defined in this environment. The error started occurring with matrix-js-sdk version 34.4.0. The error appears on this line, but the global variable is used multiple times.

The code is simple. Here’s an example.

```
"use client";

import { createClient, MatrixClient } from "matrix-js-sdk";
import React, { useEffect, useRef } from "react";

export default function HomePage() {
  const matrixClient = useRef<MatrixClient | null>(null);

  useEffect(() => {
    matrixClient.current = createClient({ baseUrl: "https://matrix.org" });
  }, []);

  return <div>page</div>;
}
```

Here, I have a screenshot of the error.

![image](https://github.com/user-attachments/assets/d19e1f61-a338-4139-ae67-0e6e87668277)

The browser recognizes `globalThis` but not `global`. `global` is available only in a node.js environment. How can I run Matrix on Next.js? Thanks in advance for the help.

richvdh:
I think matrix-js-sdk currently expects the environment to provide a polyfill for `global`. For example, [Weback provides one](https://webpack.js.org/api/module-variables/#global-nodejs).

I am very surprised that you found this started with 34.4.0, because as far as I know this code has not changed in years.

As for solutions: pull-requests to replace `global` with `globalThis` would be welcome.

In the meantime, maybe you can work around it. I'm not familiar with next.js but maybe you can configure it to provide a polyfill? Alternatively, if you provide a concrete implementation of `IStore` to `createClient`, you'll avoid the problematic line:

```ts
let store = new MemoryStore({ localStorage });
let client = createClient({ store });
```

Half-Shot:
I've noticed that this started breaking somewhere in the vicinity of v33.0.0-v34.0.0. Building and running our application on v32.4.0 didn't result in an error, but v33.x.x resulted in https://github.com/matrix-org/matrix-js-sdk/issues/4284, and v34.0.0 results in the error as described above.

*Something* did change.

richvdh:
> I've noticed that this started breaking somewhere in the vicinity of v33.0.0-v34.0.0. Building and running our application on v32.4.0 didn't result in an error, but v33.x.x resulted in #4284, and v34.0.0 results in the error as described above.
> 
> _Something_ did change.

[...]

PR #4489, opened 2024-10-31
Replace usages of `global` with `globalThis`

Half-Shot:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/4450

This should solve the problem with various bundlers (Vite etc) not being happy about the usage of `global`, with `globalThis`. All modern environments support this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis, so it's a safe change.

### DIFF
--- a/spec/unit/pushprocessor.spec.ts
+++ b/spec/unit/pushprocessor.spec.ts
@@ -819,21 +819,21 @@ describe("rewriteDefaultRules", () => {
 
         // By the time we get here, we expect the PushProcessor to have merged the new .m.rule.is_room_mention rule into the existing list of rules.
         // Check that has happened, and that it is in the right place.
-        const containsDisplayNameRuleIdx = pushRules.globalThis.override?.findIndex(
+        const containsDisplayNameRuleIdx = pushRules.global.override?.findIndex(
             (rule) => rule.rule_id === RuleId.ContainsDisplayName,
         );
         expect(containsDisplayNameRuleIdx).toBeGreaterThan(-1);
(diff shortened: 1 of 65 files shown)


t3chguy:
Any chance of a linter to protect against regression here?

Half-Shot:
> Any chance of a linter to protect against regression here?

Good idea! Let's have a look.

spec/unit/pushprocessor.spec.ts
```diff
@@ -819,21 +819,21 @@ describe("rewriteDefaultRules", () => {
 
         // By the time we get here, we expect the PushProcessor to have merged the new .m.rule.is_room_mention rule into the existing list of rules.
         // Check that has happened, and that it is in the right place.
-        const containsDisplayNameRuleIdx = pushRules.global.override?.findIndex(
+        const containsDisplayNameRuleIdx = pushRules.globalThis.override?.findIndex(
```
t3chguy, line 822: 🙃

Half-Shot: Believe it or not I did actually do these replaces manually, and still fucked it up :grimacing:

t3chguy (approved):

richvdh (approved):
lgtm

fkwp (approved):
lgtm

Half-Shot:
(N.B. leaving branch active for the moment as we're depending on it to unbreak some applications, once the *next* RC comes out this can be pruned)

PR #4492, opened 2024-11-01
Send/receive error details with widgets

AndrewFerr:
Depends on https://github.com/matrix-org/matrix-widget-api/pull/100

## Checklist

### DIFF
--- a/src/embedded.ts
+++ b/src/embedded.ts
@@ -547,7 +547,9 @@ export class RoomWidgetClient extends MatrixClient {
 }
 
 function processAndThrow(error: unknown): never {
-    throw error instanceof WidgetApiResponseError && error.data.matrix_api_error
-        ? MatrixError.fromWidgetApiErrorData(error.data.matrix_api_error)
-        : error;
+    if (error instanceof WidgetApiResponseError && error.data.matrix_api_error) {
+        throw MatrixError.fromWidgetApiErrorData(error.data.matrix_api_error);
+    } else {
+        throw error;
+    }
 }
--- a/src/embedded.ts
+++ b/src/embedded.ts
@@ -17,6 +17,7 @@ limitations under the License.
 import {
     WidgetApi,
     WidgetApiToWidgetAction,
+    WidgetApiResponseError,
     MatrixCapabilities,
     IWidgetApiRequest,
     IWidgetApiAcknowledgeResponseData,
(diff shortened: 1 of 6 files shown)


src/embedded.ts
```diff
@@ -147,6 +149,26 @@ export class RoomWidgetClient extends MatrixClient {
     ) {
         super(opts);
 
+        const transportSend = this.widgetApi.transport.send.bind(this.widgetApi.transport);
```
t3chguy, line 152: Why do we need this bind?

src/embedded.ts
```diff
@@ -147,6 +149,26 @@ export class RoomWidgetClient extends MatrixClient {
...
+        // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
t3chguy, line 153: Why?

src/embedded.ts
```diff
@@ -523,3 +545,9 @@ export class RoomWidgetClient extends MatrixClient {
         }
     }
 }
+
+function processAndThrow(error: unknown): never {
+    throw error instanceof WidgetApiResponseError && error.data.matrix_api_error
+        ? MatrixError.fromWidgetApiErrorData(error.data.matrix_api_error)
+        : error;
```
t3chguy, line 552: this is unreadable, can we have an if statement with 2 throws instead please?

src/embedded.ts
AndrewFerr: This was to avoid having to copy over the type hint of the original function, but a hint might be needed after all since this is a generic function...

src/embedded.ts
```diff
@@ -147,6 +149,26 @@ export class RoomWidgetClient extends MatrixClient {
...
+        this.widgetApi.transport.send = async (action, data) => {
+            try {
+                return await transportSend(action, data);
+            } catch (error) {
+                processAndThrow(error);
+            }
+        };
+
+        const transportSendComplete = this.widgetApi.transport.sendComplete.bind(this.widgetApi.transport);
+        // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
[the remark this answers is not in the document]
AndrewFerr: See above comment

src/embedded.ts
AndrewFerr: b6f8d1412

src/embedded.ts
AndrewFerr: This is used as a way to preserve the original `send` method so it can still be used by the function that replaces it.

Alternatively this could have been an arrow function, but that would require specifying a parameter list.  Using bind gets around that.

t3chguy: This seems like a really nasty pattern, this looks like somewhere a subclass makes far more sense rather than overwriting class methods, what's the rationale for not doing it in a more conventional way?

AndrewFerr: It's because the `send` methods don't belong to `WidgetApi`, but to `ITransport`, meaning that subclassing (or even a composition-based approach) would have to work harder to get to them.

Any non-`bind` approach I can think of ends up being more complicated / requires much more copied code than this.

t3chguy (approved):

PR #4497, opened 2024-11-05
Allow using sendEvent with a custom event type.

Half-Shot:
This exposes the ability to send a custom event type outside of `TimelineEvents` for the purposes of sending events that are not part of the Matrix spec, but may be used for specific applications. This change simply exposes an overload for the existing function, and two tests to ensure the typings are valid.

## Checklist

-   [N/A] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md)).

### DIFF
--- a/spec/unit/matrix-client.spec.ts
+++ b/spec/unit/matrix-client.spec.ts
@@ -630,6 +630,45 @@ describe("MatrixClient", function () {
             await client.sendEvent(roomId, threadId, EventType.RoomMessage, { ...content }, txnId);
         });
 
+        it("overload with custom event type works", async () => {
+            const eventId = "$eventId:example.org";
+            const customEventType = "org.example.my.custom.type";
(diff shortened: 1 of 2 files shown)


t3chguy:
The overload makes the strict type moot and open the door to typos, the preferable approach is type declaration merging like EW does. https://github.com/element-hq/element-web/blob/develop/src/%40types/matrix-js-sdk.d.ts#L47-L55

Half-Shot:
Huh, TIL. That should probably be documented in some fashion, since this feels semi-common as a use case.

richvdh:
> Huh, TIL. That should probably be documented in some fashion, since this feels semi-common as a use case.

PRs to improve the documentation very much welcome :innocent:

PR #4498, opened 2024-11-05
Fix local echo in embedded mode

toger5:
This fixes remote echos not appearing in the widget api. As a consequence, an event listener in a matroska widget will never receive events sent by itself with`isSending() === false`.

This is needed in the raise hand logic for element call. We send a raise hand and need to update the ui as soon as it is sent so that pressing the button again redacts the raise hand.

The issue is caused by the following:
 - An event gets a txId assigned by the embedded client (the widget)
 - This txId is tracked, so that the event can be matched once the remote echo is received. In that case the js-sdk "ignores" the remote event and instead updates the already existing local event to reflect it is send.
  - Using the widget api the event gets a txid twice. once in the widget (but that txid does **not** get forwarded to the client through the widget api) and once while it is sent by the client.
  - The event that then gets to the widget uses a different txid. The two events cannot be matched up because of the different txid's. since we do have the same evId however we still ignore the remote echo.
  - We end up with an ignored remote echo that did not update the local event to be sucessfully sent.
  
The fix works as following:
 - We track the txId of the local event.
 - We block all remote echos until we get a response from the send action.
 - Once we have the response form the send action we know the txId - evId relation
 - We take the remote echo and update the txId to not be the one created by the cleint but by the widget matching the local event of the widget.

## Checklist

### DIFF
--- a/src/embedded.ts
+++ b/src/embedded.ts
@@ -14,8 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// eslint-disable-next-line no-restricted-imports
-import { EventEmitter } from "events";
 import {
     WidgetApi,
     WidgetApiToWidgetAction,
@@ -48,7 +46,7 @@ import {
 import { SyncApi, SyncState } from "./sync.ts";
 import { SlidingSyncSdk } from "./sliding-sync-sdk.ts";
 import { User } from "./models/user.ts";
-import { Room } from "./models/room.ts";
+import { Room, RoomEvent } from "./models/room.ts";
 import { ToDeviceBatch, ToDevicePayload } from "./models/ToDeviceMessage.ts";
 import { DeviceInfo } from "./crypto/deviceinfo.ts";
 import { IOlmDevice } from "./crypto/algorithms/megolm.ts";
@@ -131,7 +129,7 @@ export class RoomWidgetClient extends MatrixClient {
     private syncState: SyncState | null = null;
 
     private pendingSendingEventsTxId: { type: string; id: string | undefined; txId: string }[] = [];
-    private sendingEmitter = new EventEmitter();
+
     /**
      *
      * @param widgetApi - The widget api to use for communication.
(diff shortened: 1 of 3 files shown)


src/embedded.ts
```diff
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// eslint-disable-next-line no-restricted-imports
+import { EventEmitter } from "events";
```
t3chguy, line 18: Why not a TypedEventEmitter?

florianduros (commented):
What issue this PR aims to fix?

src/embedded.ts
toger5: I am wondering if `room.updatePendingEvent(event, EventStatus.SENT, response.event_id);` emits sth that can be used here. So that I do not need the emitter at all.

toger5:
@florianduros I updated the description to reflect what the PR

src/embedded.ts
toger5: @t3chguy I updated it so it uses the `RoomEvent.LocalEchoUpdated` emitted  by room.updatePendingEvent.

t3chguy:
Please avoid force pushing once you have received review as it discards any review state and means I have to review the entire change rather than just the changes.

src/embedded.ts
```diff
@@ -128,6 +133,9 @@ export class RoomWidgetClient extends MatrixClient {
     private lifecycle?: AbortController;
     private syncState: SyncState | null = null;
 
+    private pendingSendingEventsTxId: { type: string; id: string | undefined; txId: string }[] = [];
+    private pendingEventsEmitter = new TypedEventEmitter<PendingEvent.PendingEventsChanged, PendingEventHandlerMap>();
```
t3chguy, line 137: Can this just be `eventEmitter` as it is more than likely to get reused later in its life and we'd rather not have the next use spin up a 2nd.

toger5:
We do not squash + merge right?
So I guess it makes sense to squash all the changes before I add it to the queue?

t3chguy:
The merge queue only uses squash merges

toger5:
Oh we are missing an EC reviewer anyways.

fkwp (approved):

PR #4510, opened 2024-11-11
Allow configuration of MatrixRTC timers when calling joinRoomSession()

hughns:
Not all of the new config options are being tested, but the code coverage is still reached.

The use of `Math.random()` that is flagged by SonarCloud is acceptable.

## Checklist

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -42,20 +42,6 @@ import { sleep } from "../utils.ts";
 
 const logger = rootLogger.getChild("MatrixRTCSession");
 
-const MEMBERSHIP_EXPIRY_TIME = 60 * 60 * 1000;
-const MEMBER_EVENT_CHECK_PERIOD = 2 * 60 * 1000; // How often we check to see if we need to re-send our member event
-const CALL_MEMBER_EVENT_RETRY_DELAY_MIN = 3000;
-const UPDATE_ENCRYPTION_KEY_THROTTLE = 3000;
-
-// A delay after a member leaves before we create and publish a new key, because people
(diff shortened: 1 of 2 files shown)


src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -87,20 +73,80 @@ export type MatrixRTCSessionEventHandlerMap = {
...
+     * The total time between a member leaving and the call switching to new keys is therefore:
+     * makeKeyDelay + useKeyDelay
+     */
+    useKeyDelay?: number;
+
+    /**
+     * The timeout (in milliseconds) after which the server will consider the membership to have expired if it
+     * has not received a keep-alive from the client.
+     */
+    membershipServerSideExpiryTimeout?: number;
```
toger5, line 142: Do you think it is beneficial to use sth close to the `delay` event terminology since this is what one would look out for to configure the even delay timeout?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -87,20 +73,80 @@ export type MatrixRTCSessionEventHandlerMap = {
...
+
+    /**
+     * The period (in milliseconds) that the client will send membership keep-alives to the server.
+     */
+    membershipKeepAlivePeriod?: number;
```
toger5, line 147: here something with `refresh`

toger5 (approved):
Looks very good. Thanks.
Not sure what you think about the naming preference I mentioned in the review.
Merging as it is now is also fine to me.

hughns:
My thinking on the naming is that we choose things that make sense with respect to MatrixRTC concepts. But, as you say it can be argued out and refactored later if needed.

PR #4524, opened 2024-11-12
Trigger call rejoin attempt only if you were in it

AndrewFerr:
Many membership events may get synced while your own join attempt is being processed, during which you will not be in the call membership list.  So, only try to rejoin if it seems that you have _left_ the call, not when you are not (yet) in the call.

Followup of https://github.com/matrix-org/matrix-js-sdk/pull/4342

## Checklist

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -872,10 +872,11 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
             this.emit(MatrixRTCSessionEvent.MembershipsChanged, oldMemberships, this.memberships);
 
             if (
-                this.isJoined() &&
-                oldMemberships.some(this.isMyMembership) &&
-                !this.memberships.some(this.isMyMembership)
+                this.isJoined() && // we mean to be joined
+                oldMemberships.some(this.isMyMembership) && // prior state says we were joined
+                !this.memberships.some(this.isMyMembership) // current says we are not joined
             ) {
+                // State says that we left, but we still mean to be joined
                 logger.warn("Missing own membership: force re-join");
                 // TODO: Should this be awaited? And is there anything to tell the focus?
                 this.triggerCallMembershipEventUpdate();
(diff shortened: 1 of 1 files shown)


AndrewFerr:
The consequence of not having this in is calling `triggerCallMembershipEventUpdate` more often than needed -- but that's not so bad, because that function deduplicates redundant membership update attempts (as only one update can be queued at a time).  So this PR is more for efficiency than correctness.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -871,7 +871,11 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
             logger.info(`Memberships for call in room ${this.room.roomId} have changed: emitting`);
             this.emit(MatrixRTCSessionEvent.MembershipsChanged, oldMemberships, this.memberships);
 
-            if (this.isJoined() && !this.memberships.some(this.isMyMembership)) {
+            if (
+                this.isJoined() &&
+                oldMemberships.some(this.isMyMembership) &&
```
robintown, line 876: Looks good but I'd suggest a comment here, as reading this I'd probably assume this check is redundant

robintown (approved):

src/matrixrtc/MatrixRTCSession.ts
AndrewFerr: d48ee9220

AndrewFerr:
Actually, this has a gap: if the delayed leave were to fire before the join gets sent, then there will be no `/sync` response for the delayed leave (as there is no state transition for leave -> leave) and clients won't realize that their join isn't backed by a delayed leave, meaning their join can become stuck.

What would help somewhat is for `/sync` to notify clients of finalized (i.e. sent or cancelled) delayed events ([as propsed by MSC4140](https://github.com/matrix-org/matrix-spec-proposals/blob/toger5/expiring-events-keep-alive/proposals/4140-delayed-events-futures.md#on-demand)), but even then it's possible for clients to race between getting a finalized delayed leave & sending a join.

PR #4532, opened 2024-11-15
Revert "Fix room state being updated with old (now overwritten) state and emitting for those updates. (#4242)"

toger5:
This reverts commit 957329b21821c0f632de6c04fff53144f7c0e5dd.
@hughns @t3chguy 
With the resent state_after changes this seems to not be required anymore right?

## Checklist


t3chguy:
> With the resent state_after changes this seems to not be required anymore right?

Those changes haven't landed yet and this is a breaking change as a parameter is being removed, it'd need to either be bundled into the same breaking release as the state_after stuff or need a deprecation period

t3chguy:
Where even was isStartTimelineState ever true?

toger5:
> Where even was isStartTimelineState ever true?

This was discussed sync. But for the protocol i will link it here:
https://github.com/matrix-org/matrix-js-sdk/blob/325dace4377ccd8d53e26f3e0b706876fdaf4e20/src/models/event-timeline.ts#L130

@t3chguy, Is this ready to get merged?

t3chguy:
I think this should land in the same release as the MSC4222 PR, and that's being held until after the next RC due to the risk level

t3chguy (approved):

PR #4540, opened 2024-11-22
Ensure we disambiguate display names which look like MXIDs

t3chguy:


spec/unit/room-member.spec.ts
```diff
@@ -490,6 +490,42 @@ describe("RoomMember", function () {
...
+                        joinEvent,
+                    ];
+                },
+                getUserIdsWithDisplayName: function (displayName: string) {
+                    return [userA, userC];
+                },
+            } as unknown as RoomState;
+            expect(member.name).toEqual(userA); // default = user_id
+            member.setMembershipEvent(joinEvent, roomState);
+            expect(member.name).not.toEqual("Alíce"); // it should disambig.
```
florianduros, line 525: `Alíce` is the value we want to test here? I don't know member ship works but we want to the disambiguate stuff on `@clarissa\u0a83bar` name right?

florianduros (approved):

PR #4542, opened 2024-11-22
Save the key backup key to 4S during `bootstrapCrossSigning`

dbkr:
When setting up secret storage, if we have a key backup key in cache (like we do for the cross signing secrets).

## Checklist


src/rust-crypto/rust-crypto.ts
```diff
@@ -841,7 +841,19 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+        // also don't bother storing it if we're about to set up a new backup
+        if (!setupNewKeyBackup) {
+            const keyBackupKey = await this.getSessionBackupPrivateKey();
+
+            const keyBackupKeyInStorage = await secretStorageCanAccessSecrets(this.secretStorage, [
+                "m.megolm_backup.v1",
+            ]);
+
+            if (keyBackupKey && !keyBackupKeyInStorage) {
+                await this.secretStorage.store("m.megolm_backup.v1", encodeBase64(keyBackupKey));
```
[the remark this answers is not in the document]
BillCarsonFr: There should be an example in when you resetKeyBackup(). I think that is fine, maybe you could get it from the rust olmMachine directly to avoid a base64 -> byte array -> base64?

Shouldn't we check more than `keyBackupKeyInStorage` ? What if the 4S contains an old key it won't be updated? Maybe we should check that the key in cache is valid before adding it to 4S? Is the backup trusted?

Is there a link to an issue for this PR?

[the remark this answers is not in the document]
dbkr: Okay, I think that should be better. I've made it overwrite the key if different, as per the comment.

This came out as part of the key backup by default work, so it's not a huge bug currently, but would break the flow of setting up 4S after a key backup, hence no issue.

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -705,6 +705,47 @@ describe("RustCrypto", () => {
         expect(resetKeyBackup.mock.calls).toHaveLength(2);
     });
 
+    it("bootstrapSecretStorage saves megolm backup key if already cached", async () => {
```
richvdh, line 708: Can we get an integ test for this as well, please? (One which works only through the public APIs of the js-sdk, together with fetchMock)

BillCarsonFr (changes requested):
Good addition, it makes it consistent with rust-sdk when the RecoveryKey is changed.
See my comments

src/rust-crypto/rust-crypto.ts
```diff
@@ -841,11 +841,46 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
             await this.resetKeyBackup();
         }
     }
 
+    /**
+     * If we have a backup key for the current, trusted backup in cache,
+     * and we have secret storage active, save it to secret storage.
+     */
+    private async saveBackupKeyToStorage(): Promise<void> {
+        const keyBackupInfo = await this.backupManager.getServerBackupInfo();
```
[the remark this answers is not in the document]
dbkr: OK yeah, could be worth checking that the back backup is active too - I've added that. We could look at pulling out the shared logic, some of the checks certainly are the same. Doesn't seem too bad for them each to be doing the checks.

BillCarsonFr (approved):
LGTM

src/rust-crypto/rust-crypto.ts
```diff
@@ -841,11 +841,53 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
         }
 
-        if (setupNewKeyBackup) {
+        // likewise with the key backup key: if we have one, store it in secret storage (if it's not already there)
+        // also don't bother storing it if we're about to set up a new backup
+        if (!setupNewKeyBackup) {
+            await this.saveBackupKeyToStorage();
+        } else {
             await this.resetKeyBackup();
         }
```
richvdh, line 850: I don't really understand why we're changing `bootstrapSecretStorage` in a PR that claims to be about `bootstrapCrossSigning`?

src/rust-crypto/rust-crypto.ts
```diff
@@ -841,11 +841,53 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+            logger.info("Not saving backup key to secret storage: decryption key does not match backup info");
+            return;
+        }
+
+        const backupKeyFromStorage = await this.secretStorage.get("m.megolm_backup.v1");
+        const backupKeyBase64 = backupKeys.decryptionKey.toBase64();
+
+        // The backup version that the key corresponds to isn't saved in 4S so if it's different, we must assume
+        // it's stale and overwrite.
+        if (backupKeyFromStorage !== backupKeyBase64) {
```
richvdh, line 886: what is the point of doing this check, rather than just storing the secret we have in secret storage?

src/rust-crypto/rust-crypto.ts
```diff
@@ -841,11 +841,53 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+        if (!keyBackupInfo || !keyBackupInfo.version) {
+            logger.info("Not saving backup key to secret storage: no backup info");
+            return;
+        }
+
+        const activeBackupVersion = await this.backupManager.getActiveBackupVersion();
+        if (!activeBackupVersion || activeBackupVersion !== keyBackupInfo.version) {
+            logger.info("Not saving backup key to secret storage: backup keys do not match active backup version");
+            return;
+        }
```
richvdh, line 868: Are you sure it's correct that we only store the key if backup is *active* ?

src/rust-crypto/rust-crypto.ts
richvdh, line 850: Also, the list of things that `bootstrapSecretStorage` does is quite carefully listed in the documentation in `CryptoApi`. Please update to reflect.

richvdh (commented):
belated review comments from me :(

src/rust-crypto/rust-crypto.ts
dbkr: I think purely trying to save a PUT request. I had it just storing the key originally but it evolved during review.

src/rust-crypto/rust-crypto.ts
dbkr: Not 100% - originally I didn't check, although at least if it's active then you know the key is correct? Should we worry about saving a stale backup key? Obviously in the case of setting up a new 4S this is a bit moot, technically this function might one day be used for more than that though.

src/rust-crypto/rust-crypto.ts
```diff
@@ -841,11 +841,53 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+            logger.info("Not saving backup key to secret storage: no backup key");
+            return;
+        }
+
+        if (!decryptionKeyMatchesKeyBackupInfo(backupKeys.decryptionKey, keyBackupInfo)) {
+            logger.info("Not saving backup key to secret storage: decryption key does not match backup info");
+            return;
+        }
+
+        const backupKeyFromStorage = await this.secretStorage.get("m.megolm_backup.v1");
```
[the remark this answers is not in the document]
richvdh: there is `secretStorageCanAccessSecrets` which might be what you want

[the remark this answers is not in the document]
dbkr: OK, but actually we think this whole check is redundant anyway right?

PR #4562, opened 2024-11-28
Remove deprecated calls in `webrtc/call.ts`

florianduros:
## Checklist

Require https://github.com/matrix-org/matrix-js-sdk/pull/4620
`webrtc/call.ts` uses deprecated and unsupported methods by the rust cryptography. I suppose it's somewhat broken since the usage of the rust cryptography.

I removed theses calls for https://github.com/element-hq/element-web/issues/26922
We wait for the js-sdk to drop the support of the legacy crypto (https://github.com/element-hq/element-web/issues/26922) to merge it in order to keep the feature when the legacy crypto is used.

### DIFF
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -2514,7 +2514,9 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
                     return;
                 }
 
-                // Here we were calling `MatrixClient.encryptAndSendToDevices` which is not supported by the rust cryptography.
+                // TODO: Here we were sending the event to the opponent's device as a to-device message with MatrixClient.encryptAndSendToDevices.
+                // However due to the switch to Rust cryptography we need to migrate to the new encryptToDeviceMessages API.
+                throw new Error("Unimplemented");
             } else {
                 await this.client.sendToDevice(
                     eventType,
--- a/src/webrtc/call.ts
+++ b/src/webrtc/call.ts
@@ -631,19 +631,17 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
         if (!this.client.getUseE2eForGroupCall()) return;
         // It's possible to want E2EE and yet not have the means to manage E2EE
         // ourselves (for example if the client is a RoomWidgetClient)
-        if (!this.client.isCryptoEnabled()) {
+        if (!this.client.getCrypto()) {
             // All we know is the device ID
             this.opponentDeviceInfo = new DeviceInfo(this.opponentDeviceId);
             return;
         }
-        // if we've got to this point, we do want to init crypto, so throw if we can't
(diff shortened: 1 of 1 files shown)


src/webrtc/call.ts
```diff
@@ -2516,18 +2514,7 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
...
-                            userId,
-                            deviceInfo: this.opponentDeviceInfo,
-                        },
-                    ],
-                    {
-                        type: eventType,
-                        content,
-                    },
-                );
+                // Here we were calling `MatrixClient.encryptAndSendToDevices` which is not supported by the rust cryptography.
```
robintown, line 2517: Could we replace this with the recommended encryptToDeviceMessages + queueToDevice calls rather than breaking it?

src/webrtc/call.ts
```diff
@@ -631,19 +631,17 @@ export class MatrixCall extends TypedEventEmitter<CallEvent, CallEventHandlerMap
...
-        // if we've got to this point, we do want to init crypto, so throw if we can't
-        if (!this.client.crypto) throw new Error("Crypto is not initialised.");
-
         const userId = this.invitee || this.getOpponentMember()?.userId;
 
         if (!userId) throw new Error("Couldn't find opponent user ID to init crypto");
 
-        const deviceInfoMap = await this.client.crypto.deviceList.downloadKeys([userId], false);
+        // Here we were calling `MatrixClient.crypto.deviceList.downloadKeys` which is not supported by the rust cryptography.
+        const deviceInfoMap = new Map();
```
robintown, line 644: The method isn't deprecated - can we just continue using it and wait for Rust crypto to implement it?

robintown: Oh I see, it's DeviceInfo that's been replaced. The chances we want to resurrect this code are indeed low so I guess marking this as broken isn't the end of the world.

florianduros: The `crypto` property of `MatrixClient` is deprecated and has `undefined` value when the rust crypto stack is used.

src/webrtc/call.ts
robintown: Okay, well, perhaps the comment could express a more clearly that this area is a TODO if we ever tried to use it again? I'm thinking like:

```
// TODO: Here we were sending the event to the opponent's device as a to-device message with MatrixClient.encryptAndSendToDevices.
// However due to the switch to Rust cryptography we need to migrate to the new encryptToDeviceMessages API.
throw new Error("Unimplemented");
```

robintown (approved):

richvdh:
@florianduros looks like maybe this has got stuck?

florianduros:
@richvdh No, this PR can be merged when will drop the support of the legacy crypto. If we do it sooner, we will break the legacy call when using the legacy crypto.

PR #4565, opened 2024-11-28
Replace deprecate imports

florianduros:
## Checklist

- Replace some legacy crypto imports
- Deprecate `CryptoBackend.getEventEncryptionInfo`


src/common-crypto/CryptoBackend.ts
```diff
@@ -78,6 +78,7 @@ export interface CryptoBackend extends SyncCryptoCallbacks, CryptoApi {
      * Get information about the encryption of an event
      *
      * @param event - event to be checked
+     * @deprecated
```
andybalaam, line 81: Can we add a comment here? What method to use instead would be helpful.

andybalaam (approved):
One ask, looks good.

PR #4566, opened 2024-11-28
Update matrix-sdk-crypto-wasm to 11.0.0

uhoreg:
## Checklist

### DIFF
--- a/spec/integ/crypto/verification.spec.ts
+++ b/spec/integ/crypto/verification.spec.ts
@@ -82,6 +82,8 @@ import {
 } from "./olm-utils";
 import { KeyBackupInfo } from "../../../src/crypto-api";
 import { encodeBase64 } from "../../../src/base64";
+import { RustCrypto } from "../../../src/rust-crypto/rust-crypto";
+import type { IDeviceKeys } from "../../../src/@types/crypto";
 
 // The verification flows use javascript timers to set timeouts. We tell jest to use mock timer implementations
 // to ensure that we don't end up with dangling timeouts.
@@ -990,6 +992,31 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
             aliceClient.setGlobalErrorOnUnknownDevices(false);
             syncResponder.sendOrQueueSyncResponse(getSyncResponse([BOB_TEST_USER_ID]));
             await syncPromise(aliceClient);
+            const crypto = aliceClient.getCrypto()!;
+            if (crypto instanceof RustCrypto) {
+                // Rust crypto requires the sender's device keys before it accepts
+                // a verification request.
+                const bobIdentityKeys = JSON.parse(testOlmAccount.identity_keys());
+                const bobDeviceKeys: IDeviceKeys = {
+                    user_id: "@bob:xyz",
+                    device_id: "BobDevice",
+                    algorithms: ["m.olm.v1.curve25519-aes-sha2", "m.megolm.v1.aes-sha2"],
+                    keys: {
+                        "curve25519:BobDevice": bobIdentityKeys.curve25519,
+                        "ed25519:BobDevice": bobIdentityKeys.ed25519,
+                    },
+                };
+                const signature = testOlmAccount.sign(anotherjson.stringify(bobDeviceKeys));
+                bobDeviceKeys.signatures = {
+                    "@bob:xyz": {
+                        "ed25519:BobDevice": signature,
+                    },
+                };
+                e2eKeyResponder.addDeviceKeys(bobDeviceKeys);
+                await crypto.processDeviceLists({ changed: ["@bob:xyz"] });
+                crypto.onSyncCompleted({});
+                await crypto.getUserDeviceInfo(["@bob:xyz"]);
+            }
         });
 
         /**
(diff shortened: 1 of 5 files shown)


andybalaam (approved):
Thanks, tough work.

spec/integ/crypto/verification.spec.ts
```diff
@@ -990,6 +992,31 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
             aliceClient.setGlobalErrorOnUnknownDevices(false);
             syncResponder.sendOrQueueSyncResponse(getSyncResponse([BOB_TEST_USER_ID]));
             await syncPromise(aliceClient);
+            const crypto = aliceClient.getCrypto()!;
+            if (crypto instanceof RustCrypto) {
```
dbkr, line 996: I *really* don't like that we're using `instanceof` for this - apart from breaking OO paradigms, it has a tendency to break in unpredictable ways if your modules get messed up and you end up with two different `RustCrypto` classes. If the app really needs to know what it is, an `isRustCrypto()`? Or `implNeedsSenderDeviceKeys()` (although maybe that's even worse).

richvdh: The right solution here is probably a test on `backend`, as per https://github.com/uhoreg/matrix-js-sdk/blob/crypto-wasm-11/spec/integ/crypto/verification.spec.ts#L125

uhoreg: Yes, I originally tested it using `backend`, but then TypeScript would complain because I'm calling functions that are only defined for `RustCrypto` (`onSyncCompleted` and `preprocessToDeviceMessages`)

dbkr: ah, of course. I guess cast it... or test for the existence of those functions?

uhoreg: I've changed it to check using `backend`, and then cast the `crypto` object to make TypeScript happy

spec/integ/crypto/verification.spec.ts
```diff
@@ -990,6 +992,37 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
...
+                const signature = testOlmAccount.sign(anotherjson.stringify(bobDeviceKeys));
+                bobDeviceKeys.signatures = {
+                    "@bob:xyz": {
+                        "ed25519:BobDevice": signature,
+                    },
+                };
+                e2eKeyResponder.addDeviceKeys(bobDeviceKeys);
+                await rustCrypto.processDeviceLists({ changed: ["@bob:xyz"] });
+                rustCrypto.onSyncCompleted({});
+                await rustCrypto.getUserDeviceInfo(["@bob:xyz"]);
```
richvdh, line 1024: poking things directly into the private methods of the crypto stack feels like a violation of this being an "integration" test -- the premise of such tests is that they operate via the *public* api of the js-sdk.

why can't we return this information via the `syncResponder` ?

richvdh (changes requested):
^

spec/integ/crypto/verification.spec.ts
uhoreg: I had tried various things to get RustCrypto to get it to see the device, including using `syncResponder`.  The problem was that I couldn't figure out the right combination of things to `await` so that it would reliably get the device before it tried to verify.  I don't remember exactly which combinations I tried, but I tried a few different things, before concluding that it looked like I would have to hit *some* RustCrypto method.

spec/integ/crypto/verification.spec.ts
```diff
@@ -990,6 +992,37 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("verification (%s)", (backend: st
...
+                        "curve25519:BobDevice": bobIdentityKeys.curve25519,
+                        "ed25519:BobDevice": bobIdentityKeys.ed25519,
+                    },
+                };
+                const signature = testOlmAccount.sign(anotherjson.stringify(bobDeviceKeys));
+                bobDeviceKeys.signatures = {
+                    "@bob:xyz": {
+                        "ed25519:BobDevice": signature,
+                    },
+                };
```
richvdh, line 1020: There's a utility function for this, by the way:

```suggestion
                const bobDeviceKeys = getTestOlmAccountKeys(testOlmAccount, BOB_TEST_USER_ID, "BobDevice");
```

spec/integ/crypto/verification.spec.ts
uhoreg: That seems to work. I don't remember exactly what I had tried before, so I don't know why what I tried didn't work.

richvdh (approved):
LGTM!

dbkr (approved):

richvdh:
Test flake tracked at https://github.com/element-hq/element-web/issues/28634

richvdh:
Turns out it's not a flake. Sorry @uhoreg, looks like you have more work to do here

uhoreg:
Going to try adding to the merge queue again, and see if I was just really unlucky with the complement-crypto test.  The CI will still fail due to the playwright test.

uhoreg:
Complement crypto succeeded this time, so it is flaky.  https://github.com/element-hq/element-web/pull/28668 will fix the playwright test

richvdh:
Everyone (including myself): please can we *not* mark matrix-sdk-crypto-wasm upgrades as T-Task, so that they wind up in the changelog?

t3chguy:
> Everyone (including myself): please can we *not* mark matrix-sdk-crypto-wasm upgrades as T-Task, so that they wind up in the changelog?

Might want to exclude it from renovate then as that sets the label to task too

PR #4574, opened 2024-12-06
Distinguish room state and timeline events in embedded clients

robintown:
This change enables room widget clients to take advantage of the more reliable method of communicating room state over the widget API provided by [this update to MSC2762](https://github.com/matrix-org/matrix-spec-proposals/pull/4237).

Depends on https://github.com/matrix-org/matrix-widget-api/pull/116

### DIFF
--- a/src/embedded.ts
+++ b/src/embedded.ts
@@ -28,6 +28,7 @@ import {
     WidgetApiAction,
     IWidgetApiResponse,
     IWidgetApiResponseData,
+    IUpdateStateToWidgetActionRequest,
 } from "matrix-widget-api";
 
 import { MatrixEvent, IEvent, IContent, EventStatus } from "./models/event.ts";
@@ -136,6 +137,7 @@ export type EventHandlerMap = { [RoomWidgetClientEvent.PendingEventsChanged]: ()
 export class RoomWidgetClient extends MatrixClient {
     private room?: Room;
     private readonly widgetApiReady: Promise<void>;
+    private readonly roomStateSynced: Promise<void>;
     private lifecycle?: AbortController;
     private syncState: SyncState | null = null;
 
@@ -189,6 +191,11 @@ export class RoomWidgetClient extends MatrixClient {
         };
 
         this.widgetApiReady = new Promise<void>((resolve) => this.widgetApi.once("ready", resolve));
+        this.roomStateSynced = capabilities.receiveState?.length
+            ? new Promise<void>((resolve) =>
+                  this.widgetApi.once(`action:${WidgetApiToWidgetAction.UpdateState}`, resolve),
+              )
+            : Promise.resolve();
 
         // Request capabilities for the functionality this client needs to support
         if (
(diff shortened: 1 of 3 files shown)


dbkr (approved):

src/embedded.ts
```diff
@@ -574,36 +562,15 @@ export class RoomWidgetClient extends MatrixClient {
...
-                // ```
-                // if (!params.addToTimeline && !params.addToState) {
-                //     this.injectRoomEvents(this.room!, [], [event]);
-                // } else {
-                //     this.injectRoomEvents(this.room!, params.addToState ? [event] : [], params.addToTimeline ? [event] : []);
-                // }
-                // ```
-                await this.syncApi!.injectRoomEvents(this.room!, [], [event]); // Sliding Sync
+                // Sliding Sync
+                this.syncApi!.injectRoomEvents(this.room!, [], [event]);
```
toger5, line 569: is injectRoomEvents not async anymore?
why can we drop the await?

toger5 (approved):
I approve this already. Can you still explain me why we don't need the await (or add it in case we need it)

src/embedded.ts
robintown: Good catch, I was copying the commented code which lacked await.

src/embedded.ts
```diff
@@ -276,37 +284,16 @@ export class RoomWidgetClient extends MatrixClient {
...
-                    await this.syncApi.injectRoomEvents(this.room!, undefined, events);
-                } else {
-                    await this.syncApi!.injectRoomEvents(this.room!, events); // Sliding Sync
-                }
-                events.forEach((event) => {
-                    this.emit(ClientEvent.Event, event);
-                    logger.info(`Backfilled event ${event.getId()} ${event.getType()} ${event.getStateKey()}`);
-                });
-            }) ?? [],
-        );
```
toger5, line 299: we could add this code back in case the driver does not support the `update_state` (or `state_after`) version.

src/embedded.ts
```diff
@@ -574,36 +562,15 @@ export class RoomWidgetClient extends MatrixClient {
...
-                // // -> state events part of the `timelineEventList` parameter will update the state.
-                //     this.injectRoomEvents(this.room!, [], undefined, [event]);
-                // } else {
-                //     this.injectRoomEvents(this.room!, undefined, params.addToState ? [event] : [], params.addToTimeline ? [event] : []);
-                // }
-                // ```
-
-                // Passing undefined for `stateAfterEventList` allows will make `injectRoomEvents` run in legacy mode
-                // -> state events in `timelineEventList` will update the state.
-                await this.syncApi.injectRoomEvents(this.room!, [], undefined, [event]);
```
toger5, line 592: Use the `state_after` version is not supported use:
`await this.syncApi.injectRoomEvents(this.room!, [], undefined, [event]);` as before.

src/embedded.ts
```diff
@@ -574,36 +562,15 @@ export class RoomWidgetClient extends MatrixClient {
...
             } else {
-                // The code will want to be something like:
-                // ```
-                // if (!params.addToTimeline && !params.addToState) {
-                //     this.injectRoomEvents(this.room!, [], [event]);
-                // } else {
-                //     this.injectRoomEvents(this.room!, params.addToState ? [event] : [], params.addToTimeline ? [event] : []);
-                // }
-                // ```
-                await this.syncApi!.injectRoomEvents(this.room!, [], [event]); // Sliding Sync
```
toger5, line 602: also use this if the version is not supported.

toger5 (commented):
@robintown Do you think those changes could make it work without needing changes on the rust sdk?

PR #4582, opened 2024-12-11
Handle when `aud` OIDC claim is an Array

liamdiprose:
The `aud` claim of OIDC id_tokens [can be an array](https://github.com/authts/oidc-client-ts/blob/ce6d694639c58e6a1c80904efdac5eda82b82042/src/Claims.ts#L92) but the existing logic incorrectly assumes `aud` is always a string.

This PR adds the necessary check.

Signed-off-by: Liam Diprose <liam@liamdiprose.com>

## Checklist

### DIFF
--- a/spec/unit/oidc/validate.spec.ts
+++ b/spec/unit/oidc/validate.spec.ts
@@ -170,6 +170,14 @@ describe("validateIdToken()", () => {
         expect(logger.error).toHaveBeenCalledWith("Invalid ID token", new Error("Invalid audience"));
     });
 
+    it("should not throw for a list of trusted audiences", () => {
+        mocked(jwtDecode).mockReturnValue({
+            ...validDecodedIdToken,
+            aud: [clientId],
+        });
+        expect(() => validateIdToken(idToken, issuer, clientId, nonce)).not.toThrow();
+    });
+
     it("should throw when nonce does not match", () => {
         mocked(jwtDecode).mockReturnValue({
             ...validDecodedIdToken,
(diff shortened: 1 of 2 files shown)


liamdiprose:
I have a use case where the `aud` contains all the applications the token can connect to. I trust all of them but the unit tests seem to be checking that `clientId` is alone:

https://github.com/matrix-org/matrix-js-sdk/blob/8155b0acfcfdb7f1080dca0e8876783892de34ee/spec/unit/oidc/validate.spec.ts#L164-L171

Can we simply trust a token if the client is in the `aud`?

dbkr:
Yeah, I noticed the unit test above which looks wrong as far as I can see. Well... actually strictly speaking I think it's correct because I don't think a comma-separated string means multiple auds, as far as I can see... it would just be a different string, but that's not what the name of the test implies its trying to test.

So I'd be open to fixing that. If so, I'd probably suggest sanitising into an array with something like `const sanitisedAuds = typeof claims.aud === 'string' : [claims.aud] : claims.aud` and then doing the includes, because I find the logic there quite hard to reason about. I'd probably also prefer tests asserting it actually does the right thing with an array, rather than just not throwing. Wdyt?

liamdiprose:
> Well... actually strictly speaking I think it's correct because I don't think a comma-separated string means multiple auds, as far as I can see... it would just be a different string, but that's not what the name of the test implies its trying to test.

This is my understanding as well.

I've added the check for an array that doesn't contain `clientId`. I think that covers all the added code.

dbkr:
Looks sensible as far as I can see. I've asked for an additional sanity check. Remember to run prettier.

dbkr:
Sanity check say it looks good! You might need to change the source branch to something that's not develop, I think this breaks sonarcloud.

liamdiprose:
Renaming the source branch closed this PR, so I've opened a new one here: #4584

PR #4585, opened 2024-12-13
Improve documentation on various secret-storage related methods

richvdh:


src/crypto-api/index.ts
```diff
@@ -995,15 +1000,77 @@ export interface CrossSigningStatus {
...
+             * For secret retrieval, `keys` may contain several entries, and the application can return
+             * any one of the requested keys. Unless your application specifically wants to offer the
+             * user the ability to have more than one secret storage key active at a time, it is recommended
+             * to call {@link matrix.SecretStorage.getDefaultKeyId} to figure out which is the current default key, and
+             * to return `null` if the default key is not listed in `keys`.
+             */
+            keys: Record<string, SecretStorageKeyDescription>;
+        },
+        name: string,
+    ) => Promise<[string, Uint8Array] | null>;
```
richvdh, line 1059: note that this is the same API as `SecretStorageCallbacks`, but with a different doc comment. I think it's appropriate for the two to have different documentation (if an application is using `SecretStorageCallbacks` directly, they probably have a different usecase to just using it via `CryptoCallbacks`).

src/crypto-api/index.ts
```diff
@@ -330,14 +330,18 @@ export interface CryptoApi {
      * Bootstrap the secret storage by creating a new secret storage key, add it in the secret storage and
```
andybalaam, line 330: What does this mean? Do we really create a new secret storage key and then add it to secret storage?

src/crypto-api/index.ts
```diff
@@ -995,15 +1000,77 @@ export interface CrossSigningStatus {
 /**
  * Crypto callbacks provided by the application
  */
-export interface CryptoCallbacks extends SecretStorageCallbacks {
+export interface CryptoCallbacks {
```
andybalaam, line 1003: You intentionally removed the `extends SecretStorageCallbacks`? If so, maybe this should be described in the PR description, if it's significant.

andybalaam: Ah, I see you have a comment that refers to this tangentially, and also that this is a draft PR, so I may have jumped the gun.

src/crypto-api/index.ts
```diff
@@ -995,15 +1000,77 @@ export interface CrossSigningStatus {
...
+     * and the private [cross-signing keys](https://spec.matrix.org/v1.12/client-server-api/#cross-signing).
+     *
+     * The secret storage mechanism encrypts the secrets before uploading them to the server using a
+     * secret storage key. The schema supports multiple keys, but in practice only one tends to be used
+     * at once; this is the "default secret storage key" and may be known as the "recovery key" (or, sometimes,
+     * the "security key").
+     *
+     * Secret storage can be set up by calling {@link CryptoApi.bootstrapSecretStorage}. Having done so, when
+     * the crypto stack needs to access secret storage (for example, when setting up a new device, or to
+     * store newly-generated secrets), it will use this callback.
```
andybalaam, line 1020: Just for maximum explicitness, I might say "this callback (getSecretStorageKey)" or similar.

src/crypto-api/index.ts
```diff
@@ -995,15 +1000,77 @@ export interface CrossSigningStatus {
...
+     *
+     * Note that the secret storage key may be needed several times in quick succession: it is recommended
+     * that applications use a temporary cache to avoid prompting the user multiple times for the key. See
+     * also {@link cacheSecretStorageKey} which is called when a new key is created.
```
andybalaam, line 1024: Any guidance on how long to store it for?

src/crypto-api/index.ts
```diff
@@ -995,15 +1000,77 @@ export interface CrossSigningStatus {
...
+     *
+     * The helper method {@link deriveRecoveryKeyFromPassphrase} may be useful if the secret storage key
+     * was derived from a passphrase.
+     *
+     * @param opts - An options object.
+     *
+     * @param name - the name of the *secret* (NB: not the encryption key) being stored or retrieved.
+     *    This is the "event type" stored in account data.
```
andybalaam, line 1032: I don't understand this, but that may be because I don't have enough expected context.

src/crypto-api/index.ts
```diff
@@ -995,15 +1000,77 @@ export interface CrossSigningStatus {
...
+     *
+     * @returns a pair [`keyId`, `privateKey`], where `keyId` is one of the keys from the `keys` parameter,
+     *    and `privateKey` is the raw private encryption key, as appropriate for the encryption algorithm.
+     *    (For `m.secret_storage.v1.aes-hmac-sha2`, it is the input to an HKDF as defined in the
+     *    [specification](https://spec.matrix.org/v1.6/client-server-api/#msecret_storagev1aes-hmac-sha2).)
+     *
+     *    Alternatively, if none of the keys are known, may return `null` — in which case the original
+     *    storage/retrieval operation will fail with an exception.
```
andybalaam, line 1040: The "original" means the one that called this method as a callback? If so, "in which case the storage/retrieval operation that is calling this callback will fail with an exception" might help a bit. It's very awkward to talk about :-)

andybalaam (approved):
Looks good, a few comments from me, and would definitely benefit from @uhoreg 's expertise.

src/crypto-api/index.ts
richvdh: ugh, no.

richvdh: I'm minded to simplify this in favour of the more detailed list.

src/crypto-api/index.ts
richvdh:
```suggestion
     * store newly-generated secrets), it will use this callback (`getSecretStorageKey`).
```

src/crypto-api/index.ts
richvdh: ugh, no, I was trying to avoid grasping that particular nettle right now. EW caches it for the duration of operations that might reasonably require access to SS, but I don't have a good definition of that right now.

src/crypto-api/index.ts
richvdh: Indeed. This bit got cargo-culted from the definition in `SecretStorageCallbacks`, fwiw, but I'm certainly open to making minor improvements while we're copying it.

richvdh: How about:

```suggestion
     *     operation that requires access to a secret in secret storage may fail with an exception.
```

src/crypto-api/index.ts
andybalaam: Totally fair to put this off for now.

src/crypto-api/index.ts
andybalaam: Yep, sounds good

src/crypto-api/index.ts
richvdh: "Secret storage" is really just "encrypted account data" (the spec, unhelpfully, refers to "account data" as "client config": https://spec.matrix.org/v1.12/client-server-api/#client-config). So, when I store a secret in secret storage, I encrypt it with the recovery key, and stash it as an account data "event". The `name` of the secret storage secret is then the `type` of the account data item.

Is this any better?

```suggestion
     *    When the item is stored in account data, it will have this `type`.
```

src/crypto-api/index.ts
```diff
@@ -330,14 +330,18 @@ export interface CryptoApi {
      * Bootstrap the secret storage by creating a new secret storage key, add it in the secret storage and
      * store the cross signing keys in the secret storage.
```
richvdh, line 331:
```suggestion
     * Bootstrap [secret storage](https://spec.matrix.org/v1.12/client-server-api/#storage).
```

src/crypto-api/index.ts
andybalaam: Yes, that is better.

~~Maybe this is not the right place to explain it, but it's not clear to me how I would choose what argument I would need to pass in here.~~

But I've just realised that we are probably writing this documentation for someone who is *implementing* this method, not *calling* it, so this does provide the info they need. So this is good.

PR #4586, opened 2024-12-13
Avoid key prompts when resetting crypto

dbkr:
Attempting to get the backup key out of secret storage can cause the user to be prompted for their key, which is not helpful if this is being done as part of a reset. This check was redundant anyway and we can just overwrite the key with the same value.

Also fix docs and remove check for active backup.

## Checklist

### DIFF
--- a/src/rust-crypto/rust-crypto.ts
+++ b/src/rust-crypto/rust-crypto.ts
@@ -863,12 +863,6 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
             return;
         }
 
-        const activeBackupVersion = await this.backupManager.getActiveBackupVersion();
-        if (!activeBackupVersion || activeBackupVersion !== keyBackupInfo.version) {
-            logger.info("Not saving backup key to secret storage: backup keys do not match active backup version");
-            return;
-        }
-
         const backupKeys: RustSdkCryptoJs.BackupKeys = await this.olmMachine.getBackupKeys();
         if (!backupKeys.decryptionKey) {
             logger.info("Not saving backup key to secret storage: no backup key");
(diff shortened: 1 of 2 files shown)


richvdh:
AIUI this fixes a regression introduced in #4542? Could we get them combined in the changelog?

src/rust-crypto/rust-crypto.ts
```diff
@@ -880,14 +874,9 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
 
-        const backupKeyFromStorage = await this.secretStorage.get("m.megolm_backup.v1");
         const backupKeyBase64 = backupKeys.decryptionKey.toBase64();
 
-        // The backup version that the key corresponds to isn't saved in 4S so if it's different, we must assume
-        // it's stale and overwrite.
-        if (backupKeyFromStorage !== backupKeyBase64) {
-            await this.secretStorage.store("m.megolm_backup.v1", backupKeyBase64);
-        }
+        await this.secretStorage.store("m.megolm_backup.v1", backupKeyBase64);
```
richvdh, line 879: The documentation on this method says "... and we have secret storage active ..." - is that correct?

richvdh (commented):
bit alarmed by the fact we can change this stuff without any tests breaking. Is this behaviour actually tested at all?

src/rust-crypto/rust-crypto.ts
dbkr: Mmm, doesn't look like there's a way to determine if we have 4s set up, afaics, so I guess I'll just remove that bit.

dbkr:
> AIUI this fixes a regression introduced in #4542? Could we get them combined in the changelog?

I don't think we have a way of doing that automatically, but we could annotate it and flag it for does the release, although as far as what will do in the changelog, I don't believe it would be any different?

> bit alarmed by the fact we can change this stuff without any tests breaking. Is this behaviour actually tested at all?

Behaviour that specific didn't have its own tests, no.

richvdh:
> I don't think we have a way of doing that automatically, but we could annotate it and flag it for does the release, although as far as what will do in the changelog, I don't believe it would be any different?

It's a bit of a hobbyhorse of mine, but ideally a changelog would list all the PRs that contributed to making a particular change. In other words, if it takes a few PRs to land the complete feature and shake out the bugs, they all get listed and linked under one entry in the changelog, rather than dotted throughout the changelog. The Synapse changelog does quite a good job of this, for example:

image

dbkr:
Ah, I see. Unfortunately I don't think that's a thing supported by https://github.com/t3chguy/release-drafter as far as I can see.

richvdh (approved):
LGTM

src/rust-crypto/rust-crypto.ts
richvdh: There is `this.secretStorageHasAESKey()` which is what `resetKeyBackup` does, but given this is only called from `bootstrapSecretStorage`, I don't think there's any need to call it.

PR #4589, opened 2024-12-17
Update matrix-sdk-crypto-wasm to 12.0.0

hughns:
It appears to "just work"... but I might be missing something

## Checklist


t3chguy (approved):

richvdh (approved):
yup, should be good. The only breaking change was making sure that we call `initAsync`, which matrix-js-sdk does.

PR #4590, opened 2024-12-17
Use mapped types for account data content

t3chguy:
Requires https://github.com/element-hq/element-web/pull/28752

### DIFF
--- a/src/@types/common.ts
+++ b/src/@types/common.ts
@@ -20,11 +20,3 @@ export type NonEmptyArray<T> = [T, ...T[]];
 export type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
 export type XOR<T, U> = T | U extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U;
 export type Writeable<T> = { -readonly [P in keyof T]: T[P] };
-
-/**
- * Returns a union type of the keys of the input Object type whose values are assignable to the given Item type.
- * Based on https://stackoverflow.com/a/57862073
- */
-export type Assignable<Object, Item> = {
-    [Key in keyof Object]: Object[Key] extends Item ? Key : never;
-}[keyof Object];
--- a/src/@types/common.ts
+++ b/src/@types/common.ts
@@ -20,3 +20,8 @@ export type NonEmptyArray<T> = [T, ...T[]];
 export type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
 export type XOR<T, U> = T | U extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U;
 export type Writeable<T> = { -readonly [P in keyof T]: T[P] };
+
+// Based on https://stackoverflow.com/a/57862073
+export type Assignable<Obj, Item> = {
+    [Key in keyof Obj]: Obj[Key] extends Item ? Key : never;
+}[keyof Obj];
--- a/src/@types/common.ts
+++ b/src/@types/common.ts
@@ -22,9 +22,9 @@ export type XOR<T, U> = T | U extends object ? (Without<T, U> & U) | (Without<U,
 export type Writeable<T> = { -readonly [P in keyof T]: T[P] };
 
 /**
- * Returns a union type of the keys of the input object type whose values are assignable to the given type.
+ * Returns a union type of the keys of the input Object type whose values are assignable to the given Item type.
  * Based on https://stackoverflow.com/a/57862073
  */
-export type Assignable<Obj, Item> = {
-    [Key in keyof Obj]: Obj[Key] extends Item ? Key : never;
-}[keyof Obj];
+export type Assignable<Object, Item> = {
+    [Key in keyof Object]: Object[Key] extends Item ? Key : never;
+}[keyof Object];
--- a/src/@types/common.ts
+++ b/src/@types/common.ts
@@ -21,7 +21,10 @@ export type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
 export type XOR<T, U> = T | U extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U;
 export type Writeable<T> = { -readonly [P in keyof T]: T[P] };
 
-// Based on https://stackoverflow.com/a/57862073
+/**
+ * Returns a union type of the keys of the input object type whose values are assignable to the given type.
+ * Based on https://stackoverflow.com/a/57862073
+ */
 export type Assignable<Obj, Item> = {
     [Key in keyof Obj]: Obj[Key] extends Item ? Key : never;
 }[keyof Obj];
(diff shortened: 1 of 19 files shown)


richvdh:
FYI, I hate the fact that we refer to account data as "events". It's not an event by any reasonable definition of the word "event". (You could argue that a *change* in account data might count as an event when it is sent down `/sync`, but that's about as far as it goes.) https://github.com/matrix-org/matrix-spec/issues/2030 and https://github.com/matrix-org/matrix-spec/issues/897 contain some of my previous complaining about it.

I appreciate it might be scope-creeping this PR, and diverging from the current wording of the spec, in ways you don't want to take on, but if there was a way to create this PR without propagating the word "Event" further than we have to, it would make me much happier.

t3chguy:
I think for js-sdk naming consistency, given they use MatrixEvent under the hood, it should be tackled more holistically. Maybe even starting at the spec level. The spec does indeed seem to refer to Account Data events as message events in some places https://spec.matrix.org/v1.12/client-server-api/#midentity_server https://spec.matrix.org/v1.12/client-server-api/#mfully_read so definitely needs a clean up.

richvdh:
Using MatrixEvent for account data seems like the world's worst idea, but I can't disagree with your conclusion that fixing it is a job for another day. I was just wondering if we can prepare for that day in any way that would minimise future breaking changes.

Anyway, will butt out.

MidhunSureshR (approved):

src/@types/event.ts
```diff
@@ -368,3 +372,15 @@ export interface StateEvents {
     // MSC3672
     [M_BEACON_INFO.name]: MBeaconInfoEventContent;
 }
+
+export interface AccountDataEvents {
```
richvdh, line 376: documentation, please. What concept does this encapsulate?

src/@types/event.ts
```diff
@@ -368,3 +372,15 @@ export interface StateEvents {
...
+    [EventType.PushRules]: IPushRules;
+    [EventType.Direct]: { [userId: string]: string[] };
+    [EventType.IgnoredUserList]: { [userId: string]: {} };
+    "m.secret_storage.default_key": { key: string };
+    [POLICIES_ACCOUNT_EVENT_TYPE.name]: { [key: string]: any };
+    "m.identity_server": { base_url: string | null };
+    [key: `${typeof LOCAL_NOTIFICATION_SETTINGS_PREFIX.name}.${string}`]: LocalNotificationSettings;
+    [key: `m.secret_storage.key.${string}`]: SecretStorageKeyDescription;
+    "m.megolm_backup.v1": SecretInfo;
```
richvdh, line 385: should we also have `m.cross_signing.{master,self_signing,user_signing}` here?

src/@types/event.ts
```diff
@@ -368,3 +372,15 @@ export interface StateEvents {
     // MSC3672
     [M_BEACON_INFO.name]: MBeaconInfoEventContent;
 }
+
+export interface AccountDataEvents {
+    [EventType.PushRules]: IPushRules;
+    [EventType.Direct]: { [userId: string]: string[] };
```
richvdh, line 378: I feel like we should have separate types for room-level account data and global account data.

src/secret-storage.ts
```diff
@@ -200,7 +202,9 @@ export interface SecretStorageCallbacks {
     ) => Promise<[string, Uint8Array] | null>;
 }
 
-interface SecretInfo {
+export type SecretInfoKey = Assignable<AccountDataEvents, SecretInfo>;
+
+export interface SecretInfo {
```
richvdh, line 207: documentation on both these types, please.

src/@types/common.ts
```diff
@@ -20,3 +20,8 @@ export type NonEmptyArray<T> = [T, ...T[]];
 export type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
 export type XOR<T, U> = T | U extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U;
 export type Writeable<T> = { -readonly [P in keyof T]: T[P] };
+
+// Based on https://stackoverflow.com/a/57862073
```
richvdh, line 24: very good. What does it actually do? How do I use it? Please don't make me reverse-engineer typescript.

src/secret-storage.ts
```diff
@@ -200,7 +202,9 @@ export interface SecretStorageCallbacks {
     ) => Promise<[string, Uint8Array] | null>;
 }
 
-interface SecretInfo {
+export type SecretInfoKey = Assignable<AccountDataEvents, SecretInfo>;
```
richvdh, line 205: Wouldn't it be easier/clearer/make more sense generally to define `SecretInfoKey` explicitly, and derive `AccountData` from that, rather than the other way around?

richvdh, line 205: Also, I feel like this should probably be called `SecretStorageKey` or something that makes it more obviously related to, y'know, secret storage.

richvdh (changes requested):
I applaud the effort, but to make best use of it, it needs better documentation imho

src/models/invites-ignorer-types.ts
```diff
@@ -0,0 +1,58 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { UnstableValue } from "matrix-events-sdk";
+
+/// The event type storing the user's individual policies.
+///
+/// Exported for testing purposes.
```
richvdh, line 21: I don't think `///` works for ts doc?

src/client.ts
```diff
@@ -4236,7 +4237,10 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @returns Promise which resolves: an empty object
      * @returns Rejects: with an error response.
      */
-    public setAccountData(eventType: EventType | string, content: IContent): Promise<{}> {
+    public setAccountData<K extends keyof AccountDataEvents>(
```
richvdh, line 4240: there is also `setRoomAccountData` -- shouldn't we change that too?

t3chguy: That'd for a future PR to keep the diffs smaller

src/secret-storage.ts
t3chguy: That wouldn't be extensible by consumers of this library, only interfaces are extensible via declaration merging.

src/@types/event.ts
richvdh: m.direct is *not* global account data

t3chguy: Except it literally is... https://spec.matrix.org/v1.12/client-server-api/#mdirect

richvdh: ugh, sorry, you're right.

src/models/invites-ignorer-types.ts
```diff
@@ -29,6 +29,19 @@ export const IGNORE_INVITES_ACCOUNT_EVENT_KEY = new UnstableValue(
...
+    sources?: string[];
+}
+
+export interface IgnoreInvitesContent {
+    "m.ignore.invites": Policies;
+}
+
+export interface UnstableIgnoreInvitesContent {
+    "org.matrix.msc3847.ignore.invites": Policies;
+}
```
richvdh, line 43: again, please can we make a habit of writing documentation for new types that we introduce, *especially* when they are `export`ed outside the file they are defined in? These are the content of particular types of account data, correct?

richvdh (approved):
LGTM otherwise

Issue #4483, opened 2024-10-30
Device Dehydration | js-sdk: store dehydration key


BillCarsonFr:
When we create a new dehydration key or get it from SSSS, use the function from https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/160 to cache it

PR #4599, opened 2024-12-26
Device Dehydration | js-sdk: store/load dehydration key

BillCarsonFr:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/4483

Added basic support to cache created keys and read from store instead of in memory.
Then adding some basic signalling to DeviceDehydrationManager, helpful for tests and also for apps to have some feedback on device rehydration (reports progress for example)

## Checklist


src/rust-crypto/DehydratedDeviceManager.ts
```diff
@@ -166,19 +175,22 @@ export class DehydratedDeviceManager {
...
             const keyB64 = await this.secretStorage.get(SECRET_STORAGE_NAME);
             if (keyB64 === undefined) {
                 if (!create) {
                     return null;
                 }
                 await this.resetKey();
             } else {
-                this.key = decodeBase64(keyB64);
+                const bytes = decodeBase64(keyB64);
+                const key = RustSdkCryptoJs.DehydratedDeviceKey.createKeyFromArray(bytes);
```
uhoreg, line 189: `bytes` should be zeroed out after this line, to avoid keeping the key in memory

src/rust-crypto/DehydratedDeviceManager.ts
```diff
@@ -190,7 +202,7 @@ export class DehydratedDeviceManager {
      * Returns whether or not a dehydrated device was found.
      */
     public async rehydrateDeviceIfAvailable(): Promise<boolean> {
-        const key = await this.getKey(false);
+        const key = (await this.getCachedKey()) || (await this.getKey(false));
```
uhoreg, line 205: Why is `getCachedKey` called, when `getKey` already calls `getCachedKey` and returns its value if it's non-empty?

uhoreg (commented):
I only did a quick review because the the compiler errors are distracting

src/rust-crypto/DehydratedDeviceManager.ts
```diff
@@ -267,12 +283,14 @@ export class DehydratedDeviceManager {
      * Creates and stores a new key in secret storage if none is available.
      */
     public async createAndUploadDehydratedDevice(): Promise<void> {
-        const key = (await this.getKey(true))!;
+        const key = ((await this.getCachedKey()) || (await this.getKey(true)))!;
```
uhoreg, line 286:
```suggestion
        const key = (await this.getKey(true))!;
```
since `getKey` already calls `getCachedKey`

src/crypto-api/index.ts
```diff
@@ -1243,6 +1250,45 @@ export interface OwnDeviceKeys {
...
+    [DehydratedDevicesEvents.DeviceCreated]: () => void;
+    [DehydratedDevicesEvents.DeviceUploaded]: () => void;
+    [DehydratedDevicesEvents.RehydrationStarted]: () => void;
+    [DehydratedDevicesEvents.RehydrationEnded]: () => void;
+    [DehydratedDevicesEvents.RehydrationProgress]: (roomKeyCount: number, toDeviceCount: number) => void;
+    [DehydratedDevicesEvents.PickleKeyCached]: () => void;
+    [DehydratedDevicesEvents.SchedulingError]: (msg: string) => void;
+};
+
+export abstract class DehydratedDevicesAPI extends TypedEventEmitter<
```
uhoreg, line 1280: The class and methods need documentation.

spec/integ/crypto/device-dehydration.spec.ts
```diff
@@ -40,6 +41,29 @@ describe("Device dehydration", () => {
 
         await initializeSecretStorage(matrixClient, "@alice:localhost", "http://test.server");
 
+        const dehydratedDevices = matrixClient.getCrypto()!.dehydratedDevices();
```
uhoreg, line 44: It might be better to have the crypto object re-emit the dehydration events, rather than exposing the dehydration manager object.

richvdh:
> Draft because depends on rust-sdk: https://github.com/matrix-org/matrix-rust-sdk/pull/4399 and needs new wasm bindings version.

This is no longer the case, I think? Can we update the description?

andybalaam (approved):
Looks reasonable to me, but definitely needs review from people who are closer to the dehydrated devices design.

src/crypto-api/index.ts
```diff
@@ -1243,6 +1250,72 @@ export interface OwnDeviceKeys {
     curve25519: string;
 }
 
+export enum DehydratedDevicesEvents {
```
richvdh, line 1253: doc-comment here, please. What does this enum represent?

src/crypto-api/index.ts
```diff
@@ -1243,6 +1250,72 @@ export interface OwnDeviceKeys {
...
+    RehydrationEnded = "RehydrationEnded",
+    /** Emitted during rehydration, signalling the current `roomKeyCount` and `toDeviceCount` */
+    RehydrationProgress = "RehydrationProgress",
+    /** Emitted when a dehydrated device key has been cached */
+    PickleKeyCached = "PickleKeyCached",
+    /** Emitted when an error occurred during rotation of the dehydrated device */
+    SchedulingError = "SchedulingError",
+}
+
+export type DehydratedDevicesEventsMap = {
```
richvdh, line 1270: this needs a doc-comment too, if it's being exported. (Does it really need to be?)

src/crypto-api/index.ts
```diff
@@ -1243,6 +1250,72 @@ export interface OwnDeviceKeys {
...
+     *   - If `createNewKey` is set to true, always creates a new key.
+     *   - If a dehydration key is not available, creates a new one.
+     * - Creates a new dehydrated device, and schedules periodically creating
+     *   new dehydrated devices.
+     *
+     * @param createNewKey - whether to force creation of a new dehydration key.
+     *   This can be used, for example, if Secret Storage is being reset.
+     */
+    public abstract start(createNewKey?: boolean): Promise<void>;
+}
```
richvdh, line 1317: suggest putting all this stuff in a separate file within `crypto-api`, like we've done with the other stuff below.

src/crypto-api/index.ts
richvdh: Also, consider whether this really needs to be separate from `CryptoEvent`.

spec/integ/crypto/device-dehydration.spec.ts
richvdh: +1 to this. Otherwise there is redundancy in the API, between `crypto.dehydratedDevices().start()` and `crypto.startDehydration()`: which one am I supposed to use, as a user of the library?

(You'll also need to have `MatrixClient` re-emit them in `initRustCrypto`, since `CryptoApi` itself is not an event emitter, and give them better names that reflect the fact they are to do with dehydration.)

Basically, do the same thing we've done with `KeyBackupStatus` etc.

src/crypto-api/index.ts
```diff
@@ -1243,6 +1250,72 @@ export interface OwnDeviceKeys {
...
+    /** Emitted when a new dehydrated device is created locally */
+    DeviceCreated = "DeviceCreated",
+    /** Emitted when a new dehydrated device is successfully uploaded to the server */
+    DeviceUploaded = "DeviceUploaded",
+    /** Emitted when rehydration has started */
+    RehydrationStarted = "RehydrationStarted",
+    /** Emitted when rehydration has finished */
+    RehydrationEnded = "RehydrationEnded",
+    /** Emitted during rehydration, signalling the current `roomKeyCount` and `toDeviceCount` */
+    RehydrationProgress = "RehydrationProgress",
```
richvdh, line 1263: this needs better documentation on what the payload of the event is.

```suggestion
    /**
      * Emitted during rehydration, to inform the application of rehydration progress.
      *
      * The payload is a pair `[roomKeyCount: number, toDeviceCount: number]`, where
      * `roomKeyCount` is (what, exactly??) and ``toDeviceCount` is (something else??
      */
    RehydrationProgress = "RehydrationProgress",
```

richvdh (changes requested):

src/crypto-api/index.ts
uhoreg: I followed the example of key backups (src/rust-crypto/backup.ts), and moved the events to crypto-api/CryptoEvent.ts, and the event map to src/rust-crypto/DehydratedDeviceManager.ts.  But I can move them elsewhere if there's some place better.

uhoreg:
I think that I've addressed all of @richvdh 's comments.

(Of course, this can't be merged until https://github.com/matrix-org/matrix-js-sdk/issues/4634 is fixed)

florianduros (approved):

src/crypto-api/CryptoEventHandlerMap.ts
```diff
@@ -18,6 +18,7 @@ import { CryptoEvent } from "./CryptoEvent.ts";
 import { VerificationRequest } from "./verification.ts";
 import { UserVerificationStatus } from "./index.ts";
 import { RustBackupCryptoEventMap } from "../rust-crypto/backup.ts";
+import { DehydratedDevicesEventMap } from "../rust-crypto/DehydratedDeviceManager.ts";
```
richvdh, line 21: `crypto-api` really shouldn't be pulling things in from `rust-crypto`. `RustBackupCryptoEventMap` is setting a poor example here.

I think the right way to do this is to add the values from `DehydratedDevicesEventMap` to `CryptoEventHandlerMap`, and then redefine `DehydratedDevicesEventMap` as `Pick<CryptoEventHandlerMap, DehydratedDevicesEvents>` or something like that.

src/rust-crypto/DehydratedDeviceManager.ts
```diff
@@ -79,10 +79,21 @@ export class DehydratedDeviceManager {
...
+    }
+
+    private async cacheKey(key: RustSdkCryptoJs.DehydratedDeviceKey): Promise<void> {
+        await this.olmMachine.dehydratedDevices().saveDehydratedDeviceKey(key);
+        this.emit(CryptoEvent.DehydrationKeyCached);
+    }
 
     /**
-     * Return whether the server supports dehydrated devices.
+     * Implementation of {@link DehydratedDevicesAPI#isSupported}.
```
richvdh, line 96: `DehydratedDevicesAPI` no longer exists

src/rust-crypto/DehydratedDeviceManager.ts
```diff
@@ -112,18 +123,7 @@ export class DehydratedDeviceManager {
...
-     * - Creates a new dehydration key, if necessary, and stores it in Secret
-     *   Storage.
-     *   - If `createNewKey` is set to true, always creates a new key.
-     *   - If a dehydration key is not available, creates a new one.
-     * - Creates a new dehydrated device, and schedules periodically creating
-     *   new dehydrated devices.
-     *
-     * @param createNewKey - whether to force creation of a new dehydration key.
-     *   This can be used, for example, if Secret Storage is being reset.
+     * Implementation of {@link DehydratedDevicesAPI#start}.
```
richvdh, line 126: ditto

src/crypto-api/CryptoEvent.ts
```diff
@@ -90,4 +90,36 @@ export enum CryptoEvent {
...
+
+    /** Fires when a dehydrated device key has been cached */
+    DehydrationKeyCached = "dehydration.DehydrationKeyCached",
+
+    /**
+     * Fires when an error occurred during rotation of the dehydrated device.
+     *
+     * The payload is an error message as a string.
+     */
+    DehydrationSchedulingError = "dehydration.DehydrationSchedulingError",
```
richvdh, line 124: It would be good if, for each event, we could give a quick one-sentence summary of roughly when we expect them to happen, in the course of an application's lifecycle (or if there is a specific API call that causes them to fire?).

Also, if there's anything specific we might expect the application to do with the events?

src/crypto-api/CryptoEvent.ts
```diff
@@ -90,4 +90,36 @@ export enum CryptoEvent {
...
+     * Fires during rehydration, to inform the application of rehydration progress.
+     *
+     * The payload is a pair `[roomKeyCount: number, toDeviceCount: number]`,
+     * where `roomKeyCount` is the number of room keys that have been received
+     * so far, and `toDeviceCount` is the number of to-device messages received
+     * so far (including the messages containing room keys).
+     */
+    RehydrationProgress = "dehydration.RehydrationProgress",
+
+    /** Fires when a dehydrated device key has been cached */
```
richvdh, line 116: could you explain what "cached" means (stored in the local database? In memory?)

src/crypto-api/CryptoEvent.ts
richvdh, line 124: to a naive reader, "rotation" and "scheduling" sound like different things. Shouldn't this be called `DehydrationDeviceRotationError` ?

src/crypto-api/CryptoEvent.ts
```diff
@@ -90,4 +90,36 @@ export enum CryptoEvent {
...
+    DehydratedDeviceCreated = "dehydration.DehydratedDeviceCreated",
+
+    /** Fires when a new dehydrated device is successfully uploaded to the server. */
+    DehydratedDeviceUploaded = "dehydration.DehydratedDeviceUploaded",
+
+    /** Fires when rehydration has started. */
+    RehydrationStarted = "dehydration.RehydrationStarted",
+
+    /** Fires when rehydration has finished. */
+    RehydrationEnded = "dehydration.RehydrationEnded",
```
richvdh, line 104: maybe list this after `RehydrationProgress`?

Presumably it indicates that rehydration *successfully* completed? Maybe say so in the comment?

Also `RehydrationCompleted` might be a better name?

src/rust-crypto/DehydratedDeviceManager.ts
```diff
@@ -166,19 +166,26 @@ export class DehydratedDeviceManager {
      *
      * @returns the key, if available, or `null` if no key is available
      */
-    private async getKey(create: boolean): Promise<Uint8Array | null> {
-        if (this.key === undefined) {
+    private async getKey(create: boolean): Promise<RustSdkCryptoJs.DehydratedDeviceKey | null> {
+        const cachedKey = await this.getCachedKey();
+        if (!cachedKey) {
```
richvdh, line 171: suggest an early return here, to avoid two calls to `getCachedKey` and dodgy `!` operators

```suggestion
        const cachedKey = await this.getCachedKey();
        if (cachedKey) return cachedKey;
```

(can we then inline `getCachedKey`?)

src/rust-crypto/DehydratedDeviceManager.ts
```diff
@@ -304,3 +318,30 @@ export class DehydratedDeviceManager {
         }
     }
 }
+
+/**
+ * The events fired by the DehydratedDeviceManager
+ * @internal
+ */
+export type DehydratedDevicesEvents =
```
richvdh, line 326: does this need to be `exported`? It doesn't seem to be used elsewhere?

(Alternatively, maybe import it to `rust-crypto.ts` and use it to help with the re-emits)

uhoreg:
SonarCloud says I need to add more tests, but should be ready for re-review otherwise

src/crypto-api/CryptoEvent.ts
```diff
@@ -90,4 +90,62 @@ export enum CryptoEvent {
      * `progress === total === -1`.
      */
     LegacyCryptoStoreMigrationProgress = "crypto.legacyCryptoStoreMigrationProgress",
+
+    /**
+     * Fires when a new dehydrated device is created locally.
+     *
+     * After the client calls {@link CryptoApi.startDehydration}, this event
+     * will be fired every time a new dehydrated device is created.  It may fire
+     * before `startDehydration` returns. */
```
richvdh, line 99: nit
```suggestion
     * before `startDehydration` returns.
     */
```

src/rust-crypto/DehydratedDeviceManager.ts
```diff
@@ -304,3 +326,23 @@ export class DehydratedDeviceManager {
...
+    | CryptoEvent.RehydrationCompleted
+    | CryptoEvent.RehydrationError
+    | CryptoEvent.DehydrationKeyCached
+    | CryptoEvent.DehydratedDeviceRotationError;
+
+/**
+ * A map of the {@link DehydratedDeviceEvents} fired by the {@link DehydratedDeviceManager} and their payloads.
+ * @internal
+ */
+export type DehydratedDevicesEventMap = Pick<CryptoEventHandlerMap, DehydratedDevicesEvents>;
```
richvdh, line 348: nit: does this need to be `export`ed?

```suggestion
type DehydratedDevicesEventMap = Pick<CryptoEventHandlerMap, DehydratedDevicesEvents>;
```

src/rust-crypto/DehydratedDeviceManager.ts
```diff
@@ -152,11 +160,12 @@ export class DehydratedDeviceManager {
      *
      * Creates a new key and stores it in secret storage.
```
richvdh, line 161:
```suggestion
     * Creates a new key and stores it in secret storage.
     *
     * @returns The newly-generated key.
```

src/rust-crypto/DehydratedDeviceManager.ts
```diff
@@ -166,19 +175,25 @@ export class DehydratedDeviceManager {
...
+        } else {
+            const bytes = decodeBase64(keyB64);
+            try {
+                const key = RustSdkCryptoJs.DehydratedDeviceKey.createKeyFromArray(bytes);
+                await this.cacheKey(key);
+                return key;
+            } finally {
+                bytes.fill(0);
             }
         }
```
richvdh, line 196: If you're going to do early returns, might as well take benefit from them

```suggestion

        // We successfully found the key in secret storage: decode it, and cache it in
        // the rust SDK's crypto store.
        const bytes = decodeBase64(keyB64);
        try {
            const key = RustSdkCryptoJs.DehydratedDeviceKey.createKeyFromArray(bytes);
            await this.cacheKey(key);
            return key;
        } finally {
            bytes.fill(0);
        }
```

richvdh: .oO (if we're going to `.fill(0)` on `bytes`, shouldn't we also do so for `keyB64` ?)

richvdh (approved):
A few nits, but LGTM.

And yes, looks like we need a test for loading the key from 4S?

src/rust-crypto/DehydratedDeviceManager.ts
uhoreg: > .oO (if we're going to `.fill(0)` on `bytes`, shouldn't we also do so for `keyB64` ?)

Ideally we should, but keyB64 is a string and I don't think we can modify a string?

richvdh: true

PR #4611, opened 2025-01-10
Enable @typescript-eslint/consistent-type-imports lint rule

hughns:
This is a speculative PR after encountering an unexpected (and unnecessary) circular dependency in some unit tests.

Some background on this rule: https://typescript-eslint.io/blog/consistent-type-imports-and-exports-why-and-how/

We already have the equivalent of this for exports from https://github.com/matrix-org/eslint-plugin-matrix-org/blob/a1ce5fbb7d425b932c504ae2bb7f985e3c07d9fd/typescript.js#L99

## Checklist

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -139,6 +139,8 @@ module.exports = {
                     },
                 ],
                 "no-extra-boolean-cast": "error",
+                // to avoid unnecessary runtime dependencies between files
+                "@typescript-eslint/consistent-type-imports": ["error", { fixStyle: "inline-type-imports" }],
             },
         },
         {
--- a/src/@types/auth.ts
+++ b/src/@types/auth.ts
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 import { UnstableValue } from "../NamespacedValue.ts";
-import { IClientWellKnown } from "../client.ts";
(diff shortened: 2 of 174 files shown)


.eslintrc.cjs
```diff
@@ -139,6 +139,8 @@ module.exports = {
                     },
                 ],
                 "no-extra-boolean-cast": "error",
+                // to avoid unnecessary runtime dependencies between files
+                "@typescript-eslint/consistent-type-imports": ["error", { fixStyle: "inline-type-imports" }],
```
t3chguy, line 143: Should we add this to https://github.com/matrix-org/eslint-plugin-matrix-org/blob/a1ce5fbb7d425b932c504ae2bb7f985e3c07d9fd/typescript.js#L99 instead?

hughns: That would make sense to me, assuming agreeable that this is generally a Good Thing.

If so I can make the necessary PRs.

hughns: FTR, the upstream rule was added in https://github.com/matrix-org/eslint-plugin-matrix-org/pull/44

hughns:
Superseded by https://github.com/matrix-org/matrix-js-sdk/pull/4680/

PR #4614, opened 2025-01-14
Add `CryptoApi.resetEncryption`

florianduros:
## Checklist

Task https://github.com/element-hq/element-web/issues/28977
Require https://github.com/matrix-org/matrix-js-sdk/pull/4615
Add `CryptoApi.resetEncryption`.

The goal of this new api is to give an easy to reset the current encryption state of the user:
- Disable backing up room keys and delete any existing backup.
- Remove the default secret storage key from the account data (ie: the recovery key).
- Reset the cross-signing keys.
- Re-enable backing up room keys if enabled before.

### DIFF
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -1895,7 +1895,7 @@ describe("RustCrypto", () => {
             fetchMock.post("path:/_matrix/client/v3/keys/signatures/upload", {});
         });
 
-        it("key backup should stay disabled after reset", async () => {
+        it("reset should reset 4S, backup and cross-signing", async () => {
             // We don't have a key backup
             fetchMock.get("path:/_matrix/client/v3/room_keys/version", {});
 
@@ -1939,9 +1939,9 @@ describe("RustCrypto", () => {
             );
 
             // A new key backup should be created after the reset
-            let content!: KeyBackupInfo;
+            let newKeyBackupInfo!: KeyBackupInfo;
             fetchMock.post("path:/_matrix/client/v3/room_keys/version", (res, options) => {
-                content = JSON.parse(options.body as string);
+                newKeyBackupInfo = JSON.parse(options.body as string);
                 return { version: "2" };
             });
 
@@ -1949,7 +1949,7 @@ describe("RustCrypto", () => {
             await rustCrypto.resetEncryption(authUploadDeviceSigningKeys);
 
             // A new key backup should be created
-            expect(content.auth_data).toBeTruthy();
+            expect(newKeyBackupInfo.auth_data).toBeTruthy();
             // The new cross signing keys should be uploaded
             expect(authUploadDeviceSigningKeys).toHaveBeenCalledWith(expect.any(Function));
         });
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -1912,9 +1912,16 @@ describe("RustCrypto", () => {
         });
 
         it("key backup should be re-enabled after reset", async () => {
-            fetchMock.get("path:/_matrix/client/v3/room_keys/version", testData.SIGNED_BACKUP_DATA);
             // When we will delete the key backup
-            fetchMock.delete("path:/_matrix/client/v3/room_keys/version/1", {});
+            let backupIsDeleted = false;
+            fetchMock.delete("path:/_matrix/client/v3/room_keys/version/1", () => {
+                backupIsDeleted = true;
+                return {};
+            });
+            // If the backup is deleted, we will return an empty object
+            fetchMock.get("path:/_matrix/client/v3/room_keys/version", () => {
+                return backupIsDeleted ? {} : testData.SIGNED_BACKUP_DATA;
+            });
 
             // We consider the key backup as trusted
             jest.spyOn(RustBackupManager.prototype, "isKeyBackupTrusted").mockResolvedValue({
@@ -1925,18 +1932,6 @@ describe("RustCrypto", () => {
             const rustCrypto = await makeTestRustCrypto(makeMatrixHttpApi(), undefined, undefined, secretStorage);
             // We have a key backup
             expect(await rustCrypto.getActiveSessionBackupVersion()).not.toBeNull();
-
-            let nbCalls = 0;
-            fetchMock.get(
-                "path:/_matrix/client/v3/room_keys/version",
-                () => {
-                    // First call is when we check if the key backup is enabled.
-                    // Second call is when we get the next backup after we deleted the first key backup,
-                    // we return an empty object because we don't have a key backup anymore.
-                    return ++nbCalls === 1 ? testData.SIGNED_BACKUP_DATA : {};
-                },
-                { overwriteRoutes: true },
-            );
 
             // A new key backup should be created after the reset
             let newKeyBackupInfo!: KeyBackupInfo;
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -61,6 +61,7 @@ import {
     EventShieldReason,
     ImportRoomKeysOpts,
     KeyBackupCheck,
+    KeyBackupInfo,
     VerificationRequest,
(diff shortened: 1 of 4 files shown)


andybalaam (approved):
Looks like it does what it says it does!

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1879,6 +1881,79 @@ describe("RustCrypto", () => {
...
+            fetchMock.get(
+                "path:/_matrix/client/v3/room_keys/version",
+                () => {
+                    // First call is when we check if the key backup is enabled.
+                    // Second call is when we get the next backup after we deleted the first key backup,
+                    // we return an empty object because we don't have a key backup anymore.
+                    return ++nbCalls === 1 ? testData.SIGNED_BACKUP_DATA : {};
+                },
+                { overwriteRoutes: true },
+            );
```
richvdh, line 1939: It might be better to tie this into the DELETE call, so that it doesn't return the empty result until the backup is deleted.

Otherwise, you could maybe use `.once` here? 

https://www.wheresrhys.co.uk/fetch-mock/docs/API/more-routing-methods#once

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1879,6 +1881,79 @@ describe("RustCrypto", () => {
...
+
+            // We consider the key backup as trusted
+            jest.spyOn(RustBackupManager.prototype, "isKeyBackupTrusted").mockResolvedValue({
+                trusted: true,
+                matchesDecryptionKey: true,
+            });
+
+            const rustCrypto = await makeTestRustCrypto(makeMatrixHttpApi(), undefined, undefined, secretStorage);
+            // We have a key backup
+            expect(await rustCrypto.getActiveSessionBackupVersion()).not.toBeNull();
```
richvdh, line 1927: it might be good to  collect all the `fetchMock` calls together, rather than having this in the middle?

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1879,6 +1881,79 @@ describe("RustCrypto", () => {
...
+                    // First call is when we check if the key backup is enabled.
+                    // Second call is when we get the next backup after we deleted the first key backup,
+                    // we return an empty object because we don't have a key backup anymore.
+                    return ++nbCalls === 1 ? testData.SIGNED_BACKUP_DATA : {};
+                },
+                { overwriteRoutes: true },
+            );
+
+            // A new key backup should be created after the reset
+            let content!: KeyBackupInfo;
```
richvdh, line 1942: this could do with a better name

```suggestion
            let newKeyBackupInfo!: KeyBackupInfo;
```

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1879,6 +1881,79 @@ describe("RustCrypto", () => {
...
+                setDefaultKeyId: jest.fn(),
+                hasKey: jest.fn().mockResolvedValue(false),
+                getKey: jest.fn().mockResolvedValue(null),
+            } as unknown as ServerSideSecretStorage;
+
+            fetchMock.post("path:/_matrix/client/v3/keys/upload", { one_time_key_counts: {} });
+            fetchMock.post("path:/_matrix/client/v3/keys/signatures/upload", {});
+        });
+
+        it("key backup should stay disabled after reset", async () => {
```
richvdh, line 1898: seems like we're testing some other things here?

maybe "reset should reset 4S, backup and cross-signing" ?

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1879,6 +1881,79 @@ describe("RustCrypto", () => {
...
+        it("key backup should be re-enabled after reset", async () => {
+            fetchMock.get("path:/_matrix/client/v3/room_keys/version", testData.SIGNED_BACKUP_DATA);
+            // When we will delete the key backup
+            fetchMock.delete("path:/_matrix/client/v3/room_keys/version/1", {});
+
+            // We consider the key backup as trusted
+            jest.spyOn(RustBackupManager.prototype, "isKeyBackupTrusted").mockResolvedValue({
+                trusted: true,
+                matchesDecryptionKey: true,
+            });
```
richvdh, line 1923: is this used before or after the reset?

Can we not use the actual backup data in the response to `/_matrix/client/v3/room_keys/version` so we don't need to mock this?

richvdh (commented):
generally looks good, a few nits, mostly around the tests

spec/unit/rust-crypto/rust-crypto.spec.ts
florianduros: I don't get it, sorry. `rustCrypto.getActiveSessionBackupVersion()` is not a `fetchMock`

spec/unit/rust-crypto/rust-crypto.spec.ts
florianduros: It's used before the reset in order to have a trusted backup.

I suppose there is a better way to have a trusted backup returned by `/_matrix/client/v3/room_keys/version`?

spec/unit/rust-crypto/rust-crypto.spec.ts
richvdh: > `rustCrypto.getActiveSessionBackupVersion()` is not a `fetchMock`

Yes, that's my point: I wanted you to move this block of code (`jest.spyOn`, `makeTestRustCrypto`, `getActiveSessionBackupVersion`) to after the call to `fetchMock.get("path:/_matrix/client/v3/room_keys/version")`, which you've now done :+1:

richvdh (approved):
lgtm

src/rust-crypto/rust-crypto.ts
```diff
@@ -1472,6 +1473,30 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+        // Disable the recovery key and the secret storage
+        await this.secretStorage.setDefaultKeyId(null);
+
+        // Reset the cross-signing keys
+        await this.crossSigningIdentity.bootstrapCrossSigning({
+            setupNewCrossSigning: true,
+            authUploadDeviceSigningKeys,
+        });
+
+        // If key backup was enabled, we create a new backup
```
BillCarsonFr, line 1494: Why only if backup was enabled?
That means that if there was a backup that was not working (not trusted signed correctly for example), then the `resetEncryption` flow won't put me back in a working state?

Maybe what we want to check here is if there was a backup server side, or even better do it unless the opt-out flag is set (account data `m.org.matrix.custom.backup_disabled`)?

BillCarsonFr: I think that the `rust-sdk` is ultimatly looking at the opt-out flag `m.org.matrix.custom.backup_disabled`, but there is also a local setting to force disable

https://github.com/matrix-org/matrix-rust-sdk/blob/a528624274ecd557427835511eb24513f00696d3/crates/matrix-sdk/src/encryption/recovery/mod.rs#L500-L508

So I believe this code is not consistent with rust. 
Even if the rust doc is misleading https://docs.rs/matrix-sdk/latest/matrix_sdk/encryption/recovery/struct.Recovery.html#method.reset_identity
> Finally, re-enable key backups (only if they were already enabled)

It is unfortunatly not the same `enabled` meaning

richvdh: As discussed elsewhere: ugh, yes, this is not what Element X is doing.

I think we should just enable backups here, regardless of whether they were enabled before. If you're resetting, you're going to be looking at the settings dialog, so can see if you're being opted in to keybackup against your will;  and I'd rather err on the side of just making it work for everyone.

src/rust-crypto/rust-crypto.ts
```diff
@@ -1472,6 +1473,30 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+        if (backupEnabled) {
+            await this.resetKeyBackup();
+        }
```
richvdh, line 1497: while we're here: it would be good to add some logging at the start and end of `resetEncryption` to make it more obvious what's going on in rageshakes.

PR #4615, opened 2025-01-14
`defaultKeyId` of the secret storage can be set at null

florianduros:
## Checklist

Task https://github.com/element-hq/element-web/issues/28977
The goal is to be able to remove the current recovery key.

The [spec](https://spec.matrix.org/v1.13/client-server-api/#key-storage) expects a string for the key however [the rust sdk replaces the content of `m.secret_storage.default_key` width an empty object(`{}`)](https://docs.rs/matrix-sdk/latest/matrix_sdk/encryption/recovery/struct.Recovery.html#method.reset_identity). We went with the same strategy to keep it consistent.


src/@types/event.ts
```diff
@@ -373,7 +373,7 @@ export interface AccountDataEvents extends SecretStorageAccountDataEvents {
     [EventType.PushRules]: IPushRules;
     [EventType.Direct]: { [userId: string]: string[] };
     [EventType.IgnoredUserList]: { [userId: string]: {} };
-    "m.secret_storage.default_key": { key: string };
+    "m.secret_storage.default_key": { key: string } | {};
```
t3chguy, line 376: ` | {}` means accept anything. I suggest reading https://www.totaltypescript.com/the-empty-object-type-in-typescript - also not sure if this is even wanted here as `setAccountData` already has `| Record<string, never>` to allow "deleting" account data events

florianduros: Oh TIL, thanks!

The rust sdk is setting an empty object and I used the same approach for constituency between our sdks

florianduros: Yep, but we still need to modify the `AccountDataClient` interface of `secret-storage` to able to do it here. (will do)

src/secret-storage.ts
```diff
@@ -148,7 +148,10 @@ export interface AccountDataClient extends TypedEventEmitter<ClientEvent.Account
      * @param content - the content object to be set
      * @returns an empty object
      */
-    setAccountData: <K extends keyof AccountDataEvents>(eventType: K, content: AccountDataEvents[K]) => Promise<{}>;
+    setAccountData: <K extends keyof AccountDataEvents>(
+        eventType: K,
+        content: AccountDataEvents[K] | Record<string, never>,
+    ) => Promise<{}>;
```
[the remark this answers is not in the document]
t3chguy: > It seems a bit surprising to me: why does setAccountData let you set the account data to values that AccountDataEvents considers invalid?

For the same reason one can encounter timeline events which do not match TimelineEvents, the empty/redaction edge case. Account data doesn't support redaction so `{}` is used instead for the purpose.

t3chguy (approved):

src/secret-storage.ts
```diff
@@ -357,21 +363,33 @@ export class ServerSideSecretStorageImpl implements ServerSideSecretStorage {
...
-                if (ev.getType() === "m.secret_storage.default_key" && ev.getContent().key === keyId) {
+                if (ev.getType() !== "m.secret_storage.default_key") {
+                    //  Different account data item
+                    return;
+                }
+
+                const content = ev.getContent();
+                // If keyId === null, the content should be an empty object.
+                // Otherwise, keyId should be returned in the content object.
+                const isSameKey = keyId === null ? !Object.keys(content).length : content.key === keyId;
```
richvdh, line 379: Slight clarifications. Let's avoid the implicit cast from number to boolean.

```suggestion
                // If keyId === null, the content should be an empty object.
                // Otherwise, the `key` in the content object should match keyId.
                const content = ev.getContent();
                const isSameKey = keyId === null ? Object.keys(content).length === 0 : content.key === keyId;
```

richvdh (approved):
lgtm otherwise

PR #4617, opened 2025-01-15
Ensure room is in store before emitting events from it

t3chguy:
Otherwise things reacting to timeline/state events on the client may call `getRoom` get null and panic.


t3chguy:
Looks like this may need a different solution, this seems to break some edge cases

PR #4621, opened 2025-01-16
Change randomString et al to be secure

dbkr:
...and renames them, removing the special lowercase and uppercase versions and exporting the underlying function instead.

I don't believe any of these uses are speed-sensitive, so I don't really see much value in maintaining separate, fast, insecure versions. I've just changed everything to use the secure versions.

Any apps that use these will either need to take the speed hit from secure random functions and use the new ones, or write their own insecure versions. Either way, they deliberately have different names so apps will need to make a choice rather then just get moved to the secure versions.

The lowercase and uppercase versions were used exactly once each in element-web and never in js-sdk itself. The underlying function is very simple and exporting just this gives more flexibility with fewer exports.

## Checklist

### DIFF
--- a/src/randomstring.ts
+++ b/src/randomstring.ts
@@ -17,9 +17,9 @@ limitations under the License.
 
 import { encodeUnpaddedBase64Url } from "./base64.ts";
 
-const LOWERCASE = "abcdefghijklmnopqrstuvwxyz";
-const UPPERCASE = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
-const DIGITS = "0123456789";
+export const LOWERCASE = "abcdefghijklmnopqrstuvwxyz";
+export const UPPERCASE = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+export const DIGITS = "0123456789";
 
 export function secureRandomBase64Url(len: number): string {
     const key = new Uint8Array(len);
@@ -28,24 +28,29 @@ export function secureRandomBase64Url(len: number): string {
     return encodeUnpaddedBase64Url(key);
 }
 
-export function randomString(len: number): string {
-    return randomStringFrom(len, UPPERCASE + LOWERCASE + DIGITS);
+/**
+ * Generates a random string of uppercase and lowercase letters plus digits using a
+ * cryptographically secure random number generator.
+ * @param len The length of the string to generate
+ * @returns Random string of uppercase and lowercase letters plus digits of length `len`
+ */
+export function secureRandomString(len: number): string {
+    return secureRandomStringFrom(len, UPPERCASE + LOWERCASE + DIGITS);
 }
 
-export function randomLowercaseString(len: number): string {
-    return randomStringFrom(len, LOWERCASE);
-}
-
-export function randomUppercaseString(len: number): string {
-    return randomStringFrom(len, UPPERCASE);
-}
-
-function randomStringFrom(len: number, chars: string): string {
+/**
+ * Generate a cryptographically secure random string using characters given
+ * @param len The length of the string to generate
+ * @param chars The characters to use in the random string.
+ * @returns Random string of characters of length `len`
+ */
+export function secureRandomStringFrom(len: number, chars: string): string {
+    const positions = new Uint32Array(chars.length);
     let ret = "";
-
-    for (let i = 0; i < len; ++i) {
-        ret += chars.charAt(Math.floor(Math.random() * chars.length));
+    crypto.getRandomValues(positions);
+    for (let i = 0; i < len; i++) {
+        const currentCharPlace = positions[i % chars.length] % chars.length;
+        ret += chars[currentCharPlace];
     }
-
     return ret;
 }
--- a/src/randomstring.ts
+++ b/src/randomstring.ts
@@ -17,8 +17,22 @@ limitations under the License.
 
 import { encodeUnpaddedBase64Url } from "./base64.ts";
 
+/**
+ * String representing the lowercase latin alphabet for use in secureRandomStringFrom
+ * (can be combined with other such exports or other characters by appending strings)
+ */
 export const LOWERCASE = "abcdefghijklmnopqrstuvwxyz";
+
+/**
+ * String representing the uppercase latin alphabet for use in secureRandomStringFrom
+ * (can be combined with other such exports or other characters by appending strings)
+ */
 export const UPPERCASE = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+
+/**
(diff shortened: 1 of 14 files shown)


florianduros (approved):

richvdh (commented):
> Any apps that use these

Are these functions supposed to be exported to apps? They don't appear to be exported via the [public entry points to the library](https://github.com/matrix-org/matrix-js-sdk?tab=readme-ov-file#entry-points) and in consequence aren't [documented](http://matrix-org.github.io/matrix-js-sdk/index.html). I'd go so far as to say this is not an `X-Breaking-Change`: apps using internal functions within the js-sdk don't get to rely on the semver telling them if those internal functions are changing.

If you want them to be exported, I'd suggest that this would be a good time to fix that.

dbkr:
Mmm, yeah - we had been assuming that it was, but no, they should not really be exported as far as I'm concerned: I don't believe it's in the js-sdk remit to provide random string utilities (obviously element web is a special case as ever).

src/randomstring.ts
```diff
@@ -17,9 +17,9 @@ limitations under the License.
 
 import { encodeUnpaddedBase64Url } from "./base64.ts";
 
-const LOWERCASE = "abcdefghijklmnopqrstuvwxyz";
-const UPPERCASE = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
-const DIGITS = "0123456789";
+export const LOWERCASE = "abcdefghijklmnopqrstuvwxyz";
+export const UPPERCASE = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+export const DIGITS = "0123456789";
```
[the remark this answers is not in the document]
richvdh: I don't really mind. What I do think is that, if you're going to `export` things (even just internally within the library), you should document them. I think attempting to document `DIGITS` is going to feel particularly futile, hence the recommendation to move that.

Certainly moving `secureRandomStringFrom` seems like it might be reasonable, but it doesn't feel quite so silly to export it, so I don't mind.

src/randomstring.ts
```diff
@@ -28,24 +28,29 @@ export function secureRandomBase64Url(len: number): string {
...
-
-    for (let i = 0; i < len; ++i) {
-        ret += chars.charAt(Math.floor(Math.random() * chars.length));
+    crypto.getRandomValues(positions);
+    for (let i = 0; i < len; i++) {
+        const currentCharPlace = positions[i % chars.length] % chars.length;
+        ret += chars[currentCharPlace];
     }
-
     return ret;
```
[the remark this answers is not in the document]
andybalaam: Well spotted. If we do find a bug I unit test reproducing it would be awesome. You might be able to mock `crypto` to get predictable results?

andybalaam (approved):
No comments beyond what Rich said, but please do address the question about whether it works the way we expect it to.

src/randomstring.ts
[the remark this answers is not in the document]
dbkr: After some back & forth with crypto, I've updated the algorithm. The conclusion seemed to be to prioritise strict correctness so hopefully this is an acceptably comprehensible version. I have made the unit test assert a little about the behaviour but haven't gone as far as asserting specific outputs for specific random values which would really be nailing down the exact algorithm. However we could do this assert that it picks all values exactly equal number of times (for certain specific inputs).

src/randomstring.ts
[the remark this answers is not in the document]
dbkr: Now that we've gone & put the effort into secureRandomStringFrom, I've kept it here and docced the exports, hopefully in a way that feels informative rather than performative!

src/randomstring.ts
```diff
@@ -28,24 +42,60 @@ export function secureRandomBase64Url(len: number): string {
...
+    if (len < 1 || len > 32768) {
+        throw new Error("Requested random string length must be between 1 and 32768");
+    }
+
+    // We'll generate random unsigned bytes, so get the largest number less than 256 that is a multiple
+    // of the length of the character set: We'll need to discard any random values that are larger than
+    // this as we can't possibly map them onto the character set while keeping each character equally
+    // likely to be chosen (minus 1 to convert to indices in a string). (Essentially, we're using a d8
+    // to choose between 7 possibilities and re-rolling on an 8, keeping all 7 outcomes equally likely.)
+    const maxRandValue = Math.floor(255 / chars.length) * chars.length - 1;
```
dkasak, line 77:
```suggestion
    const maxRandValue = 255 - 255 % chars.length - 1;
```

We can avoid `Math.floor` using modular arithmetic.

dbkr: Fixed, and adjusted to be 256 which is what it surely should have been, and don't subtract 1 because we're already using < (rather than <=) below.

dkasak: Yes. I started writing an elaborate explanation of why the bounds were wrong, but you beat me to it by grokking it yourself beforehand. :)

src/randomstring.ts
```diff
@@ -17,9 +17,23 @@ limitations under the License.
 
 import { encodeUnpaddedBase64Url } from "./base64.ts";
 
-const LOWERCASE = "abcdefghijklmnopqrstuvwxyz";
-const UPPERCASE = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
-const DIGITS = "0123456789";
+/**
+ * String representing the lowercase latin alphabet for use in secureRandomStringFrom
```
richvdh, line 21: likewise elsewhere

```suggestion
 * String representing the lowercase latin alphabet for use in {@link secureRandomStringFrom}
```

src/randomstring.ts
```diff
@@ -28,24 +42,61 @@ export function secureRandomBase64Url(len: number): string {
...
 }
 
-export function randomLowercaseString(len: number): string {
-    return randomStringFrom(len, LOWERCASE);
-}
+/**
+ * Generate a cryptographically secure random string using characters given
+ * @param len The length of the string to generate (must be positive and less than 32768)
+ * @param chars The characters to use in the random string (between 2 and 256 characters long).
+ * @returns Random string of characters of length `len`
```
richvdh, line 59: I think our style is something like:

```suggestion
 * Generate a cryptographically secure random string using characters given.
 *
 * @param len - The length of the string to generate (must be positive and less than 32768).
 * @param chars - The characters to use in the random string (between 2 and 256 characters long).
 * @returns Random string of characters of length `len`.
```

(Note `-` after params, consistent use of full-stops, a bit of space between description and `@param`)

It would be nice to figure this out and be consistent.

richvdh (approved):
LGTM other than nitpicking about documentation

dbkr:
Force merging because it won't pass element web tests without the corresponding web PR

PR #4622, opened 2025-01-16
Remove legacy crypto support in `sync` api

florianduros:
## Checklist

### DIFF
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -59,7 +59,6 @@ import { BeaconEvent } from "./models/beacon.ts";
 import { IEventsResponse } from "./@types/requests.ts";
 import { UNREAD_THREAD_NOTIFICATIONS } from "./@types/sync.ts";
 import { Feature, ServerSupport } from "./feature.ts";
-import { Crypto } from "./crypto/index.ts";
 import { KnownMembership } from "./@types/membership.ts";
 
 const DEBUG = true;
@@ -116,13 +115,6 @@ function debuglog(...params: any[]): void {
  * Options passed into the constructor of SyncApi by MatrixClient
  */
 export interface SyncApiOptions {
-    /**
-     * Crypto manager
-     *
-     * @deprecated in favour of cryptoCallbacks
-     */
-    crypto?: Crypto;
-
     /**
      * If crypto is enabled on our client, callbacks into the crypto module
      */
@@ -642,9 +634,6 @@ export class SyncApi {
             }
             this.opts.filter.setLazyLoadMembers(true);
         }
-        if (this.opts.lazyLoadMembers) {
-            this.syncOpts.crypto?.enableLazyLoading();
-        }
     };
 
     private storeClientOptions = async (): Promise<void> => {
@@ -880,12 +869,6 @@ export class SyncApi {
                 catchingUp: this.catchingUp,
             };
 
-            if (this.syncOpts.crypto) {
-                // tell the crypto module we're about to process a sync
-                // response
-                await this.syncOpts.crypto.onSyncWillProcess(syncEventData);
-            }
-
             try {
                 await this.processSyncResponse(syncEventData, data);
             } catch (e) {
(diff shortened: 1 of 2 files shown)


src/sliding-sync-sdk.ts
t3chguy, line 1: This looks like it wants re-implementing in a rust-crypto compatible way otherwise parts of crypto won't work in sliding sync which is unexpected

florianduros: Do we want to support it? Sliding sync is kind of abandoned

t3chguy: Its in a weird middle ground where people still use it until we support SSS. Simplified Sliding Sync also needs extensions of this shape AFAIK - https://github.com/matrix-org/matrix-spec-proposals/blob/erikj/sss/proposals/4186-simplified-sliding-sync.md#extensions

The only difference AIUI between SS && SSS is the lack of sliding windows in the latter.

richvdh: I've said this elsewhere, but I don't think we need to consider adding support for rust crypto to sliding sync as a blocker for removal of the legacy crypto code.

The whole point of the `initCrypto` -> `initLegacyCrypto` switcheroo was to try to encourage people that had an important usecase for legacy crypto to come forward. Nothing happened, so we have to assume that nobody is using the E2EE support in sliding sync today.

I don't mind if anyone wants to volunteer to update this implementation, but we should be wary of getting sidetracked.

t3chguy: it'll most likely break end to end tests in Element Web which rely on sliding sync, so will be blocked from landing until many more changes happen.

dbkr: I'm a bit confused by this to be honest. Do we know what part rust crypto doesn't support? We could break crypto in the SS labs flag, its not ideal. Does it not work if you just swap the test to use rust crypto?

t3chguy: The implementation lacks support for MSC3884 in rust crypto, but has MSC3885. 

> Does it not work if you just swap the test to use rust crypto?

The sliding sync API uses the legacy crypto field for its E2EExtension support, so swapping the test over will just demonstrate the missing support, not repair it.

t3chguy: As per https://github.com/matrix-org/matrix-spec-proposals/blob/erikj/sss/proposals/4186-simplified-sliding-sync.md#extensions - both of those MSCs are required

t3chguy: Indeed it looks like a straight swap should work. It seems a lot more appropriate than just removing code & tests

florianduros: https://github.com/matrix-org/matrix-js-sdk/pull/4624 created to give it a try

src/sync.ts
```diff
@@ -642,9 +634,6 @@ export class SyncApi {
             }
             this.opts.filter.setLazyLoadMembers(true);
         }
-        if (this.opts.lazyLoadMembers) {
-            this.syncOpts.crypto?.enableLazyLoading();
-        }
```
t3chguy, line 647: Do we have documentation anywhere that confirms rust-crypto doesn't need something similar to support lazy loading being enabled?

src/sync.ts
```diff
@@ -920,15 +903,6 @@ export class SyncApi {
...
 
             if (this.client.store.wantsSave()) {
-                // We always save the device list (if it's dirty) before saving the sync data:
-                // this means we know the saved device list data is at least as fresh as the
-                // stored sync data which means we don't have to worry that we may have missed
-                // device changes. We can also skip the delay since we're not calling this very
-                // frequently (and we don't really want to delay the sync for it).
-                if (this.syncOpts.crypto) {
-                    await this.syncOpts.crypto.saveDeviceList(0);
-                }
```
t3chguy, line 930: How does rust crypto handle this? Seems like an important thing to have

florianduros: We don't have it when using EW so I suppose it's fine?

t3chguy: Or maybe its the cause of some bugs we haven't yet investigated properly?

richvdh: rust crypto persists the list of users that need a key query to indexeddb on each `/sync` request, so we don't need this extra call to flush out the data.

src/sync.ts
richvdh: I don't remember the details right now, but no this is not required.

t3chguy (approved):

PR #4626, opened 2025-01-17
Switch OIDC primarily to new `/auth_metadata` API

t3chguy:
Updates to latest variant of https://github.com/sandhose/matrix-spec-proposals/blob/msc/sandhose/oidc-discovery/proposals/2965-auth-metadata.md

Also drastically simplifies the overly-duplicated config/metadata interfaces

### DIFF
--- a/src/oidc/discovery.ts
+++ b/src/oidc/discovery.ts
@@ -16,7 +16,7 @@ limitations under the License.
 
 import { MetadataService, OidcClientSettingsStore } from "oidc-client-ts";
 
-import { isValidatedIssuerMetadata, validateOIDCIssuerWellKnown } from "./validate.ts";
+import { validateAuthMetadata } from "./validate.ts";
 import { Method, timeoutSignal } from "../http-api/index.ts";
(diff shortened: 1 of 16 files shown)


src/oidc/discovery.ts
```diff
@@ -38,23 +39,27 @@ export const discoverAndValidateOIDCIssuerWellKnown = async (issuer: string): Pr
         signal: timeoutSignal(5000),
     });
     const issuerWellKnown = await issuerWellKnownResponse.json();
-    const validatedIssuerConfig = validateOIDCIssuerWellKnown(issuerWellKnown);
+    return validateAuthMetadataAndKeys(issuerWellKnown);
+};
+
+/**
+ *
```
dbkr, line 46: Could do with some more doc

dbkr (approved):
otherwise lgtm

src/oidc/discovery.ts
t3chguy: Oops, forgot to fill, thanks for the spot

Issue #4502, opened 2024-11-06
`MatrixClient.login` persists access token but not device ID


richvdh:
The [/login](https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3login) endpoint returns, among other things, an access token and a device ID.

The access token is self-evidently important: it must be used for all future requests. `MatrixClient.login` persists the access token so that future methods on MatrixClient use it.

However, the device ID is *also* important, particularly for E2EE operations, since the device ID is used in signing operations. However, `MatrixClient.login` does not persist it. 

The following code, therefore fails, but in a subtle way:

```ts
const client = sdk.createClient({ baseUrl: "https://matrix.org" });
await client.login("m.login.password", {
    identifier: { type: "m.id.user", user: "userid" },
    password: "mysecretpassword",
});
await client.initRustCrypto();
```

The normal usage of `client.login()` is to create a new MatrixClient based on the results, but if that is the intention:
 1. The documentation should say so
 2. It should not persist the access token either.

`loginWithPassword` and `loginWithToken` are similarly affected.

PR #4627, opened 2025-01-17
`MatrixClient.login` no longer stashes the returned access token

richvdh:
Previously, `MatrixClient.login` had some very unintuitive behaviour where it
stashed the access token, but not the device id, refresh token, etc etc, which
led people to imagine that they had a functional MatrixClient when they didn't.

Ideally we would stash all the returned credentials so that the app doesn't
need to make a new MatrixClient, but that's a bit complicated (especially with
OIDC) and more than I have time for, so let's at least disable the footgun by
not saving *anything*.

Fixes: https://github.com/matrix-org/matrix-js-sdk/issues/4502

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -8244,29 +8244,23 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 
     /**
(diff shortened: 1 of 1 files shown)


richvdh:
closing in favour of #4632

PR #4628, opened 2025-01-17
Fix documentation on `CryptoEvent`

richvdh:
`CryptoApi` itself does not emit events (or at least, its public type information does not allow you to listen for events emitted by CryptoApi).


andybalaam (approved):
Great, thanks!

PR #4629, opened 2025-01-17
Add an extra consistency check in `bootstrapCrossSigning`

richvdh:
check that `importCrossSigningKeys` has actually worked

### DIFF
--- a/src/rust-crypto/CrossSigningIdentity.ts
+++ b/src/rust-crypto/CrossSigningIdentity.ts
@@ -87,7 +87,7 @@ export class CrossSigningIdentity {
                     "bootstrapCrossSigning: Cross-signing private keys not found locally, but they are available " +
                         "in secret storage, reading storage and caching locally",
                 );
-                const status: CrossSigningStatus = await this.olmMachine.importCrossSigningKeys(
+                const status = await this.olmMachine.importCrossSigningKeys(
                     masterKeyFromSecretStorage,
                     selfSigningKeyFromSecretStorage,
                     userSigningKeyFromSecretStorage,
--- a/src/rust-crypto/CrossSigningIdentity.ts
+++ b/src/rust-crypto/CrossSigningIdentity.ts
@@ -95,7 +95,7 @@ export class CrossSigningIdentity {
 
                 // Check that `importCrossSigningKeys` worked correctly (for example, it will fail silently if the
                 // public keys are not available).
-                if (!status.hasMaster || !status.hasSelfSigning || !status.hasSelfSigning) {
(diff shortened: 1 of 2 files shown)


src/rust-crypto/CrossSigningIdentity.ts
```diff
@@ -87,12 +87,18 @@ export class CrossSigningIdentity {
...
-                await this.olmMachine.importCrossSigningKeys(
+                const status: CrossSigningStatus = await this.olmMachine.importCrossSigningKeys(
                     masterKeyFromSecretStorage,
                     selfSigningKeyFromSecretStorage,
                     userSigningKeyFromSecretStorage,
                 );
 
+                // Check that `importCrossSigningKeys` worked correctly (for example, it will fail silently if the
+                // public keys are not available).
+                if (!status.hasMaster || !status.hasSelfSigning || !status.hasSelfSigning) {
```
t3chguy, line 98: This seems to check the same thing twice

t3chguy (changes requested):

src/rust-crypto/CrossSigningIdentity.ts
richvdh: gah, good spot

```suggestion
                if (!status.hasMaster || !status.hasSelfSigning || !status.hasUserSigning) {
```

src/rust-crypto/CrossSigningIdentity.ts
```diff
@@ -87,12 +87,18 @@ export class CrossSigningIdentity {
                     "bootstrapCrossSigning: Cross-signing private keys not found locally, but they are available " +
                         "in secret storage, reading storage and caching locally",
                 );
-                await this.olmMachine.importCrossSigningKeys(
+                const status: CrossSigningStatus = await this.olmMachine.importCrossSigningKeys(
```
t3chguy, line 90: Can we type this in the .d.ts file instead of needing it on every call? a la https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/%40types/matrix-sdk-crypto-wasm.d.ts#L22

t3chguy: I'm asking for the local d.ts file I linked to be updated @richvdh - not a file in another project.

t3chguy (approved):

t3chguy:
Would have been nice to have a test for this, but approved anyway

richvdh:
> Would have been nice to have a test for this, but approved anyway

Agreed; I picked this up as a quick 5-minute hack and don't really have time to make it more comprehensive.

PR #4632, opened 2025-01-20
Deprecate `MatrixClient.login` and replace with `loginRequest`

richvdh:
`MatrixClient.login` has some very unintuitive behaviour where it
stashes the access token, but not the device id, refresh token, etc etc, which
led people to imagine that they had a functional `MatrixClient` when they
didn't. In practice, you have to create a *new* `MatrixClient` given the `LoginResponse`.

As the first step for sorting this out, this deprecates the broken method and
replaces it with one that has sensible behaviour.

Related to: #4502


t3chguy (approved):
LGTM modulo coverage

Issue #4634, opened 2025-01-20
Constructor for OwnUserIdentity no longer accessible


uhoreg:
Something changed between v12.1.0 and v13 which caused `OwnUserIdentity`'s constructor to become private.  I don't see any code changes that would cause that, and as far as I can tell, `OwnUserIdentity` never should have had a constructor.  However, the test at https://github.com/matrix-org/matrix-js-sdk/blob/develop/spec/unit/rust-crypto/rust-crypto.spec.ts#L1558-L1565 tries to use it.  This means that when we bump js-sdk to depend on v13, we get an error: https://github.com/matrix-org/matrix-js-sdk/actions/runs/12764224352/job/35575962385?pr=4599

Since it seems like that test is the only place where the constructor is used, we may just want to change the test, rather try to re-add the constructor.  @richvdh since you wrote that test, do you have any opinion on this?

uhoreg:
Since I'm changing the test, I've moved the issue to js-sdk

PR #4635, opened 2025-01-21
Fix test that will break with matrix-sdk-crypto-wasm 13.0.0

uhoreg:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/4634

## Checklist

### DIFF
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -64,6 +64,8 @@ import {
     VerificationRequest,
 } from "../../../src/crypto-api";
 import * as testData from "../../test-utils/test-data";
+import { E2EKeyReceiver } from "../../test-utils/E2EKeyReceiver";
+import { E2EKeyResponder } from "../../test-utils/E2EKeyResponder";
 import { defer } from "../../../src/utils";
 import { logger } from "../../../src/logger";
 import { OutgoingRequestsManager } from "../../../src/rust-crypto/OutgoingRequestsManager";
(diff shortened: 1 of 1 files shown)


uhoreg:
The test runs correctly when I run only the `pinCurrentIdentity` tests, but when all the `spec/unit/rust-crypto/rust-crypto.spec.ts` tests are run, it times out for some reason.  So I'll have to poke at it some more.

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1532,34 +1534,47 @@ describe("RustCrypto", () => {
...
 
         it("throws an error for an unknown user", async () => {
-            await expect(rustCrypto.pinCurrentUserIdentity("@alice:example.com")).rejects.toThrow(
+            await expect(rustCrypto.pinCurrentUserIdentity("@other_user:example.com")).rejects.toThrow(
                 "Cannot pin identity of unknown user",
             );
         });
 
         it("throws an error for our own user", async () => {
-            const ownIdentity = new RustSdkCryptoJs.OwnUserIdentity();
```
BillCarsonFr, line 1559: Isn't it more simple to just do            
```
 const ownIdentity = { } as unknown as Mocked<RustSdkCryptoJs.OwnUserIdentity>;
```
Instead of bringing all the key receiver/responder framework?

I added similar tests in this PR https://github.com/matrix-org/matrix-js-sdk/pull/4646/files

BillCarsonFr (dismissed):

spec/unit/rust-crypto/rust-crypto.spec.ts
uhoreg: The problem is that `pinCurrentUserIdentity` uses [`instanceof` to check whether it's an `OwnUserIdentity`](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/rust-crypto/rust-crypto.ts#L684), so it needs a real `OwnUserIdentity` object.

andybalaam (approved):
Looks good, thanks!

PR #4652, opened 2025-01-23
Allow the embedded client to work without the `UpdateState` api version

### LINKED ISSUE #4655: Regression: Embedded mode no longer works with widget host that does not support update_state action
https://github.com/matrix-org/matrix-js-sdk/pull/4574 landed via https://github.com/matrix-org/matrix-js-sdk/commit/bdd4d82cb3587d6d6285d05bb31d07796ca3959e into the RCs for v36.1.0.

The problem is that if the widget host doesn't support the new `update_state` action then the fetching of room state will never work. For example, the matrix-rust-sdk does not support `update_state` yet.

This will impact any widget that uses the matrix-js-sdk in embedded mode including element-call.

A proposed fix is in https://github.com/matrix-org/matrix-js-sdk/pull/4652

I've marked this as a release blocker as it wasn't intended that this would be incompatible.

toger5:
Fixes #4655

The widget api was changed so state events get send using the dedicated `UpdateState` action instead of the normal timeline action: `SendEvent`.

The embedded client (used by EC) is currently only compatible with widgetDrivers that support `UpdateState`. As of now the rust sdk does not support `UpdateState` however which makes the embedded client only work on EW.

This PR changes the embedded client so that it first checks if the host client supports `UpdateState`. If it does not it uses the old widget actions to inject state events into the client sync api.

## Checklist

### DIFF
--- a/spec/unit/embedded.spec.ts
+++ b/spec/unit/embedded.spec.ts
@@ -28,7 +28,6 @@ import {
     WidgetApiToWidgetAction,
     MatrixCapabilities,
     ITurnServer,
-    IRoomEvent,
     IOpenIDCredentials,
     ISendEventFromWidgetResponseData,
     WidgetApiResponseError,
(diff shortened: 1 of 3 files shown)


hughns:
Closing as what this was based on was is now in https://github.com/matrix-org/matrix-js-sdk/pull/4652

PR #4653, opened 2025-01-23
Remove legacy crypto

florianduros:
Closes https://github.com/element-hq/element-web/issues/26922
[EW e2e tests](https://github.com/element-hq/element-web/pull/29162)

Remove the legacy crypto.  All included PRs have been reviewed.

- https://github.com/matrix-org/matrix-js-sdk/pull/4562
- https://github.com/matrix-org/matrix-js-sdk/pull/4654
- https://github.com/matrix-org/matrix-js-sdk/pull/4620
- https://github.com/matrix-org/matrix-js-sdk/pull/4622
- https://github.com/matrix-org/matrix-js-sdk/pull/4659
- https://github.com/matrix-org/matrix-js-sdk/pull/4663
- https://github.com/matrix-org/matrix-js-sdk/pull/4666
- https://github.com/matrix-org/matrix-js-sdk/pull/4668
- https://github.com/matrix-org/matrix-js-sdk/pull/4669
- https://github.com/matrix-org/matrix-js-sdk/pull/4671
- https://github.com/matrix-org/matrix-js-sdk/pull/4667
- https://github.com/matrix-org/matrix-js-sdk/pull/4672
- https://github.com/matrix-org/matrix-js-sdk/pull/4682
- https://github.com/matrix-org/matrix-js-sdk/pull/4692
- https://github.com/matrix-org/matrix-js-sdk/pull/4696

PR #4659, opened 2025-01-27
Remove deprecated methods and attributes of `MatrixClient`

florianduros:

This PR removes the deprecated methods and attributes of the the `MatrixClient`. The legacy crypto relies on using deprecated methods of the `MatrixClient` and the `MatrixClient` is also using methods from the legacy crypto.

In order to not have a gigantic PR where both the deprecated methods of the `MatrixClient` and the legacy crypto are removed, only the `MatrixClient` is changed here (with the related tests). Accordingly, the legacy crypto tests and lint are in failure.

Since the PR will be merged in `florianduros/rip-out-legacy-crypto/remove-legacy-crypto` and not in `develop`, we can accept that the lint and the legacy crypto tests are in failure in the feature branch until the legacy crypto is removed.

The unit (not related to legacy crypto) and  integ tests are updated to pass. However the cleaning of theses (remove `isNewBackendOnly`, run on each `backend` etc) tests will be done in a dedicated PR.


src/client.ts
```diff
@@ -1383,11 +1304,9 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             } else {
                 this.deviceId = opts.deviceToImport.deviceId;
                 this.credentials.userId = opts.deviceToImport.userId;
-                // will be used during async initialization of the crypto
-                this.exportedOlmDeviceToImport = opts.deviceToImport.olmDevice;
```
florianduros, line 1387: Should we ditch `opts.deviceToImport` also? Since `exportedOlmDeviceToImport` was only used in the legacy crypto, I suppose it makes no sense to only import the `userId` and the `deviceId`

dbkr (approved):
Looks sensible to me, will leave your question to the crypto experts. Looking forward to getting a lot of lines out of client.js.

spec/integ/crypto/crypto.spec.ts
```diff
@@ -1030,13 +1007,10 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("crypto (%s)", (backend: string,
...
         const room = aliceClient.getRoom(ROOM_ID);
         expect(room).toBeTruthy();
-        const p = aliceClient.prepareToEncrypt(room!);
+        aliceClient.getCrypto()?.prepareToEncrypt(room!);
 
         // we expect to get a room key message
         await expectSendRoomKey("@bob:xyz", testOlmAccount);
-
-        // the prepare request should complete successfully.
-        await p;
```
richvdh, line 1039: isn't this still useful?

richvdh (commented):
looks good to me in general

spec/integ/crypto/crypto.spec.ts
richvdh: [that looks like a bug, tbh]

richvdh (approved):
lgtm

PR #4663, opened 2025-01-28
Clean up legacy stores

florianduros:

The migration from the legacy crypto to the rust crypto uses the legacy stores. In order to be able to delete the rest of the legacy crypto, these stores shouldn't rely on the types of the legacy crypto.

This PR does:
- Keep the store methods which are needed in the migration and in the tests
- Extract the types of the legacy crypto in put then in. `src/crypto/store/base.ts`

The legacy crypto tests and linting are in failure due to the removal of the store methods. The legacy crypto will be removed in the next PR.

### DIFF
--- a/src/crypto/store/base.ts
+++ b/src/crypto/store/base.ts
@@ -25,12 +25,6 @@ import { ISignatures } from "../../@types/signed.ts";
  */
 
 export interface SecretStorePrivateKeys {
-    "dehydration": {
-        keyInfo: { [props: string]: any };
-        key: AESEncryptedSecretStoragePayload;
-        deviceDisplayName: string;
-        time: number;
-    } | null;
     "m.megolm_backup.v1": AESEncryptedSecretStoragePayload;
 }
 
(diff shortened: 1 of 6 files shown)


src/crypto/store/base.ts
```diff
@@ -14,27 +14,19 @@ See the License for the specific language governing permissions and
...
  * Internal module. Definitions for storage for the crypto module
  */
 
 export interface SecretStorePrivateKeys {
     "dehydration": {
-        keyInfo: DehydrationManager["keyInfo"];
+        keyInfo: { [props: string]: any };
         key: AESEncryptedSecretStoragePayload;
         deviceDisplayName: string;
         time: number;
```
richvdh, line 32: we should probably get rid of this "dehydration", btw. It's only used for the now-unimplemented msc2697 v1 dehydrated devices .

src/crypto/store/base.ts
```diff
@@ -81,22 +73,6 @@ export interface CryptoStore {
...
-    updateOutgoingRoomKeyRequest(
-        requestId: string,
-        expectedState: number,
-        updates: Partial<OutgoingRoomKeyRequest>,
-    ): Promise<OutgoingRoomKeyRequest | null>;
-    deleteOutgoingRoomKeyRequest(requestId: string, expectedState: number): Promise<OutgoingRoomKeyRequest | null>;
-
     // Olm Account
     getAccount(txn: unknown, func: (accountPickle: string | null) => void): void;
     storeAccount(txn: unknown, accountPickle: string): void;
```
richvdh, line 78: if these `storeXYZ` methods are only here for tests, it might be possible/good to remove them from the public `CryptoStore` interface?

richvdh (approved):
lgtm

Issue #4484, opened 2024-10-30
Device Dehydration |  js-sdk: provide functions to allow application to create dehydrated devices on restart


BillCarsonFr:
Follow up of https://github.com/matrix-org/matrix-js-sdk/issues/4483
part of https://github.com/element-hq/element-meta/issues/2603

~When the SDK starts, use the function from https://github.com/matrix-org/matrix-rust-sdk/issues/4186 to check if there is a dehydration key cached, and if so, schedule periodic device dehydration. May want to wait until a sync is done, so that we can see if the dehydration key got reset, so we don't create a dehydrated device using an old key.~

When the application starts, it should check if there is a dehydration key cached, and if so, schedule periodic device dehydration.  The SDK needs to provide functionality so that the application can do so.

PR #4664, opened 2025-01-29
Provide more options for starting dehydration

uhoreg:
Needed for starting dehydration after restart -- since we only want to do that if we already have a key cached, and we don't need to try to rehydrate a device.

Fixes https://github.com/matrix-org/matrix-js-sdk/issues/4484

## Checklist

### DIFF
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -1894,6 +1894,168 @@ describe("RustCrypto", () => {
             await rehydrationCompletedPromise;
             await rustCrypto2.stop();
         });
+
+        it("should handle dehydration start options", async () => {
+            const secretStorageCallbacks = {
+                getSecretStorageKey: async (keys: any, name: string) => {
+                    return [[...Object.keys(keys.keys)][0], new Uint8Array(32)];
+                },
+            } as SecretStorageCallbacks;
+            const secretStorage = new ServerSideSecretStorageImpl(new DummyAccountDataClient(), secretStorageCallbacks);
+
+            const e2eKeyReceiver = new E2EKeyReceiver("http://server");
+            const e2eKeyResponder = new E2EKeyResponder("http://server");
+            e2eKeyResponder.addKeyReceiver(TEST_USER, e2eKeyReceiver);
+            fetchMock.get("path:/_matrix/client/v3/room_keys/version", {
+                status: 404,
+                body: {
+                    errcode: "M_NOT_FOUND",
+                    error: "Not found",
+                },
+            });
+            fetchMock.post("path:/_matrix/client/v3/keys/device_signing/upload", {
+                status: 200,
+                body: {},
+            });
+            fetchMock.post("path:/_matrix/client/v3/keys/signatures/upload", {
+                status: 200,
+                body: {},
+            });
+            const rustCrypto = await makeTestRustCrypto(makeMatrixHttpApi(), TEST_USER, TEST_DEVICE_ID, secretStorage);
+
+            // dehydration requires secret storage and cross signing
+            async function createSecretStorageKey() {
+                return {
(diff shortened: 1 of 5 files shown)


spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1894,6 +1894,168 @@ describe("RustCrypto", () => {
...
+            );
+            fetchMock.post(/_matrix\/client\/unstable\/org.matrix.msc3814.v1\/dehydrated_device\/.*\/events/, {
+                status: 200,
+                body: {
+                    events: [],
+                    next_batch: "foo",
+                },
+            });
+
+            fetchMock.mockClear();
```
andybalaam, line 1990: I would prefer the above setup code to be in a function that takes arguments for the exact setup we need e.g. `function setup(key_cached: boolean, device_exists: boolean, ...) ...` and then separate tests for the things we want to be true:

`it("should do nothing if no key is cached and onlyIfKeyCached is true")`
`it("should create a new key and device if no key or device exist")`
`it("should rehydrate an existing device using the existing key if it exists")`
`it("should create a new device if we don't have the key for the existing device")`
`it("should create a new device even if one exists if rehdrate is false")`
`it("should create a new device every time if createNewKey is true")`

This would really help me understand what the code is supposed to do, and will help us understand what happened if future code changes affect this logic.

This is a preference, so if you have good reasons for this narrative style, you can ignore this request :-)

andybalaam (approved):

src/crypto-api/index.ts
```diff
@@ -41,6 +41,18 @@ import { MatrixEvent } from "../models/event.ts";
...
  */
 
+/**
+ * The options to start device dehydration.
+ */
+export interface StartDehydrationOpts {
+    /** Force creation of a new dehydration key. Defaults to `false`. */
+    createNewKey?: boolean;
+    /** Only start dehydration if we have a cached key. Defaults to `false`. */
+    onlyIfKeyCached?: boolean;
```
BillCarsonFr, line 51: How could I create a dehydrated device if I don't have the key cached? That would mean by creating a new key?
I am a bit confused by these options and how they work together. For example what happen if both `createNewKey` and `onlyIfKeyCached` are false?

Also why `rehydrate` optional? 

I am not sure to understand how it links to the mentionned issue (#4484).
I mean if the app restarts and device dehydration is enabled it should schedule the next dehydration, and if it cannot because it doesn't have the key, the DehydratedDeviceManager should emit/notify that to the app (like we do -should do / sooon do- when another secret is missing? or whn backup isn't working).
Then the user fix it one way or another:
- getting the key => DehydratedDeviceManager is happy and schedule
- ~reset => DehydratedDeviceManager sees that there is now a key and is happy

What am I missing?

BillCarsonFr (dismissed):

spec/unit/rust-crypto/rust-crypto.spec.ts
uhoreg: Changed.

The narrative style made sense when I was writing it, but this is probably clearer.

src/crypto-api/index.ts
uhoreg: I added more to the comments to make it clearer.  Do the options make sense now?



> I am not sure to understand how it links to the mentionned issue (#4484).
> I mean if the app restarts and device dehydration is enabled it should schedule the next dehydration, and if it cannot because it doesn't have the key, the DehydratedDeviceManager should emit/notify that to the app (like we do -should do / sooon do- when another secret is missing? or whn backup isn't working).

The idea is that if we already have the key cached, then when can go ahead and start dehydration immediately without bothering the user.  If we don't have the key cached, we'll need to prompt the user (e.g. to unlock SSSS).  4484 is just for the simple case.  The case(s) where we need to prompt the user would be covered by https://github.com/element-hq/element-web/issues/29080 (and possibly others).

BillCarsonFr (approved):

PR #4668, opened 2025-01-29
Remove usage of legacy crypto in `embedded.ts`

florianduros:

The interface of `encryptAndSendToDevices` changes because `DeviceInfo` is from the legacy crypto. In fact `encryptAndSendToDevices` only need pairs of `userId` and `deviceId`.

We need to update the interface and we can simplify it.

The CI is in failure because the root branch (which is a feature branch) has removed deprecated `MatrixClient` methods used by the legacy crypto. Legacy crypto which will be removed in this feature branch.

### DIFF
--- a/src/embedded.ts
+++ b/src/embedded.ts
@@ -54,14 +54,23 @@ import { MatrixError } from "./http-api/errors.ts";
 import { User } from "./models/user.ts";
 import { Room } from "./models/room.ts";
 import { ToDeviceBatch, ToDevicePayload } from "./models/ToDeviceMessage.ts";
-import { DeviceInfo } from "./crypto/deviceinfo.ts";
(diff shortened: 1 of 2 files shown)


t3chguy:
I think this needs review from voip as they are the codeowners for it

src/embedded.ts
```diff
@@ -128,6 +137,7 @@ export enum RoomWidgetClientEvent {
     PendingEventsChanged = "PendingEvent.pendingEventsChanged",
 }
 export type EventHandlerMap = { [RoomWidgetClientEvent.PendingEventsChanged]: () => void };
+
```
florianduros, line 140: Noisy, I'll remove it

hughns (approved):

PR #4670, opened 2025-01-29
Remove deprecated parameters and types in `CryptoApi`

florianduros:
Remove deprecated parameters in `CryptoApi` and update rust-crypto .


richvdh (changes requested):
I'm a bit nervous about this: it goes quite a long way beyond "remove legacy crypto", imho. For example `findVerificationRequestDMInProgress` and `storeSessionBackupPrivateKey` work ok in rust crypto.

Not saying we shouldn't plan to remove them, but I don't think we should do it as part of this change.

Removing the old `DecryptionFailureCode`s and `CryptoCallbacks` seems sensible.

Also, the tests are failing.

florianduros:
The legacy crypto tests are failing because of https://github.com/matrix-org/matrix-js-sdk/pull/4659.

If we plan to remove them, should be in the same release than the removal of the legacy crypto in order to avoir too many releases with breaking changes. Or do we want to do in some weeks/month?

richvdh:
> If we plan to remove them, should be in the same release than the removal of the legacy crypto in order to avoir too many releases with breaking changes. Or do we want to do in some weeks/month?

I'm not really sure yet. Right now, I'm mostly nervous that we're making sweeping changes without effective CI. 

To be honest, I think we should leave it for a release or two; maybe get a working `vNext` system going first so that we can do major-version housekeeping regularly without it being a massive headache.

Either way: not part of this PR, please?

richvdh:
@florianduros I think this got (mostly) superceded? Can we close it?

PR #4673, opened 2025-01-31
Handle empty m.room.topic

Half-Shot:
Pairs with https://github.com/matrix-org/matrix-spec/pull/2068, and an Element Web PR shortly.

The `topic` key of a topic may be undefined when it is removed, and this PR aims to support that use case. This may be considered a breaking change in that we're changing the Topic event content interfaces, but it's only to match up with reality.

NOTE: I was unsure what kind of `T-` this might be, I've gone with defect as I think it's resolving a problem where the js-sdk didn't match reality, but this could be construed as other things too.

## Checklist

### DIFF
--- a/src/@types/topic.ts
+++ b/src/@types/topic.ts
@@ -60,4 +60,4 @@ export type MTopicEvent = EitherAnd<{ [M_TOPIC.name]: MTopicContent }, { [M_TOPI
 /**
  * The event content for an m.room.topic event
  */
-export type MRoomTopicEventContent = { topic: string | null | undefined } & MTopicEvent;
+export type MRoomTopicEventContent = { topic: string | null | undefined } & (MTopicEvent | {});
--- a/src/@types/topic.ts
+++ b/src/@types/topic.ts
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { EitherAnd } from "matrix-events-sdk";
+import { EitherAnd, Optional } from "matrix-events-sdk";
 
 import { UnstableValue } from "../NamespacedValue.ts";
 import { IMessageRendering } from "./extensible_events.ts";
@@ -60,4 +60,4 @@ export type MTopicEvent = EitherAnd<{ [M_TOPIC.name]: MTopicContent }, { [M_TOPI
 /**
  * The event content for an m.room.topic event
  */
-export type MRoomTopicEventContent = { topic: string } & MTopicEvent;
+export type MRoomTopicEventContent = { topic: Optional<string> } & MTopicEvent;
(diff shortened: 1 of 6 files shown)


src/@types/topic.ts
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { EitherAnd } from "matrix-events-sdk";
+import { EitherAnd, Optional } from "matrix-events-sdk";
```
t3chguy, line 17: Can we please not use more from the matrix-events-sdk given its borderline abandoned

Half-Shot: Should it gain a `@deprecated` flag?

t3chguy (approved):

src/@types/topic.ts
richvdh: > I believe formally the spec core team owns it.

That would be very much news to me. I think you're right that it's abandoned, anyway.

(That said, it seems to me that it would make sense, as a maintenance strategy for the js-sdk, to factor out a lighter-weight library which just exposes some typescript types and maybe thin wrappers around the REST APIs: currently MatrixClient tries to be both high-level and low-level and it's confusing to use and maintain.)

t3chguy: > (That said, it seems to me that it would make sense, as a maintenance strategy for the js-sdk, to factor out a lighter-weight library which just exposes some typescript types and maybe thin wrappers around the REST APIs

Agreed, though I'd want both packages shipping out of the same repo as needing to create 3 PRs to add an endpoint & logic & consumer would be a PITA.

src/@types/topic.ts
```diff
@@ -60,4 +60,4 @@ export type MTopicEvent = EitherAnd<{ [M_TOPIC.name]: MTopicContent }, { [M_TOPI
 /**
  * The event content for an m.room.topic event
  */
-export type MRoomTopicEventContent = { topic: string } & MTopicEvent;
+export type MRoomTopicEventContent = { topic: string | null | undefined } & (MTopicEvent | {});
```
t3chguy, line 63: This allows any type to be set... https://www.totaltypescript.com/the-empty-object-type-in-typescript

I swear I left a comment but think it was on the commit so may have been lost. @Half-Shot can you please resolve this. This allows a value of `123` as the content which would error on the API for example.

t3chguy: image

Half-Shot: Sure, I'll do a followup now.

Half-Shot: http://localhost:8008/_synapse/admin/v1/users/@test_deactivated:beefy/media?order_by=created_ts&dir=b&limit=1

PR #4677, opened 2025-02-03
Enable key upload to backups where we have the decryption key

### LINKED ISSUE #4676: Keys are not uploaded to backup if the backup is not signed
Spec says https://spec.matrix.org/v1.10/client-server-api/#server-side-key-backups : trust can also be determined
> by deriving the public key from a private key that it obtained from a trusted source. Trusted sources for the private key include the user entering the key, retrieving the key stored in [secret storage](https://spec.matrix.org/v1.10/client-server-api/#secret-storage), or obtaining the key via [secret sharing](https://spec.matrix.org/v1.10/client-server-api/#sharing) from a verified device belonging to the same user.

link to problem area: https://github.com/matrix-org/matrix-js-sdk/blob/cc238c24ab8f89a25373c00570fe8292b2c3c49b/src/rust-crypto/backup.ts#L306

should be (as both are part of `BackupTrustInfo`)

```typescript
if (!trustInfo.matchesDecryptionKey && !trustInfo.trusted) {
```

alternative fix at https://github.com/matrix-org/matrix-js-sdk/blob/cc238c24ab8f89a25373c00570fe8292b2c3c49b/src/rust-crypto/backup.ts#L132
```typescript
trusted: signatureVerification.trusted() || backupMatchesSavedPrivateKey
```

related debugging: 
- https://github.com/matrix-org/matrix-js-sdk/blob/cc238c24ab8f89a25373c00570fe8292b2c3c49b/src/rust-crypto/backup.ts#L123-L124
- https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/blob/1499b237364cf5751c64dfbb0c119ebb676ccc3b/src/machine.rs#L1127
- https://github.com/matrix-org/matrix-rust-sdk/blob/57919f5480baa99474bb16ecef94525d78e23107/crates/matrix-sdk-crypto/src/backups/mod.rs#L273
- https://github.com/matrix-org/matrix-rust-sdk/blob/57919f5480baa99474bb16ecef94525d78e23107/crates/matrix-sdk-crypto/src/backups/mod.rs#L95

more details:
one of my account only has device signatures(from logged out devices) in `auth_data` and missing `master_key` signature in that case the backup is stalled for forever. Not sure we can also add `master_key` signature using https://spec.matrix.org/v1.10/client-server-api/#put_matrixclientv3room_keysversionversion without resetting version?

ajbura:
Fixes #4676 

Signed-off-by: Ajay Bura [ajbura@gmail.com](mailto:ajbura@gmail.com)

## Checklist


richvdh (changes requested):
Putting this on hold while we try to understand what bug it is trying to fix, per https://github.com/matrix-org/matrix-js-sdk/issues/4676#issuecomment-2630716430.

In any case, we'll need to see an integration test which demonstrates the problem.

richvdh (commented):
ok, the fix looks like it makes sense, but the main thing this is missing is a regression test.

Please could you take a look at `matrix-js-sdk/spec/integ/crypto/megolm-backup.spec.ts`. In there there are some existing tests for `checkKeyBackupAndEnable`: please could you take a look at them and see if you can add another which covers your usecase.

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -1077,7 +1077,46 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("megolm-keys backup (%s)", (backe
...
+
+            fetchMock.get("path:/_matrix/client/v3/room_keys/version", testData.SIGNED_BACKUP_DATA);
+
+            const result = await aliceCrypto.checkKeyBackupAndEnable();
+            expect(result).toBeTruthy();
+            expect(result!.trustInfo).toEqual({ trusted: false, matchesDecryptionKey: true });
+            expect(await aliceCrypto.getActiveSessionBackupVersion()).toEqual(testData.SIGNED_BACKUP_DATA.version);
+        });
+
+        it("does not enable a backup signed by an untrusted device and does not have decryption key", async () => {
```
richvdh, line 1119:
```suggestion
        it("does not enable a backup signed by an untrusted device when we do not have the decryption key", async () => {
```

richvdh (commented):
looks generally good to me, thanks.

A few minor requests.

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -1077,7 +1077,46 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("megolm-keys backup (%s)", (backe
...
+            await waitForDeviceList();
+            await aliceCrypto.setDeviceVerified(testData.TEST_USER_ID, testData.TEST_DEVICE_ID);
+
+            fetchMock.get("path:/_matrix/client/v3/room_keys/version", testData.SIGNED_BACKUP_DATA);
+
+            const result = await aliceCrypto.checkKeyBackupAndEnable();
+            expect(result).toBeTruthy();
+            expect(result!.trustInfo).toEqual({ trusted: true, matchesDecryptionKey: true });
+            expect(await aliceCrypto.getActiveSessionBackupVersion()).toEqual(testData.SIGNED_BACKUP_DATA.version);
+        });
```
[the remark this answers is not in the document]
ajbura: Not any particular reason, I added it as the trust is logical OR of verified signature and decryption key match, so i feel like it is a good idea to cover all 4 combination to stay safe.

spec/integ/crypto/megolm-backup.spec.ts
```diff
@@ -1077,7 +1077,46 @@ describe.each(Object.entries(CRYPTO_BACKENDS))("megolm-keys backup (%s)", (backe
...
+        it("enables a backup signed by untrusted devices but have decryption key", async () => {
+            aliceClient = await initTestClient();
+            const aliceCrypto = aliceClient.getCrypto()!;
+
+            await aliceClient.startClient();
+            await aliceCrypto.storeSessionBackupPrivateKey(
+                Buffer.from(testData.BACKUP_DECRYPTION_KEY_BASE64, "base64"),
+                testData.SIGNED_BACKUP_DATA.version!,
+            );
+            await waitForDeviceList();
```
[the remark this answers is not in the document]
ajbura: I am not sure why test is failing after flipping the statements. Can you please help me with it

src/rust-crypto/backup.ts
```diff
@@ -164,9 +165,10 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
         // We force a check to ensure to have the latest version. We also want to check that the backup is trusted
         // as we don't want to store the secret if the backup is not trusted, and eventually import megolm keys later from an untrusted backup.
```
richvdh, line 166: this comment seems to be out of date.

```suggestion
        // We force a check to ensure to have the latest version.
```

src/rust-crypto/backup.ts
```diff
@@ -164,9 +165,10 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
         const backupCheck = await this.checkKeyBackupAndEnable(true);
+        const version = backupCheck?.backupInfo.version;
 
-        if (!backupCheck?.backupInfo?.version || !backupCheck.trustInfo.trusted) {
-            // There is no server-side key backup, or the backup is not signed by a trusted cross-signing key or trusted own device.
+        if (!backupCheck || !version) {
+            // There is no server-side key backup, or the backup is not trusted.
```
richvdh, line 171: So is this one

```suggestion
            // There is no server-side key backup.
```

src/rust-crypto/backup.ts
```diff
@@ -164,9 +165,10 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
         // We force a check to ensure to have the latest version. We also want to check that the backup is trusted
         // as we don't want to store the secret if the backup is not trusted, and eventually import megolm keys later from an untrusted backup.
         const backupCheck = await this.checkKeyBackupAndEnable(true);
+        const version = backupCheck?.backupInfo.version;
```
richvdh, line 168: please could you give this a better name?

```suggestion
        const latestBackupVersion = backupCheck?.backupInfo.version;
```

src/rust-crypto/backup.ts
```diff
@@ -164,9 +165,10 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
 
-        if (!backupCheck?.backupInfo?.version || !backupCheck.trustInfo.trusted) {
-            // There is no server-side key backup, or the backup is not signed by a trusted cross-signing key or trusted own device.
+        if (!backupCheck || !version) {
```
richvdh, line 170: There is no point in testing `backupCheck` as well as `version` here. `version` cannot be set unless `backupCheck` is set.

```suggestion
        if (!version) {
```

richvdh (changes requested):

src/rust-crypto/backup.ts
ajbura: I think current comment make more sense here, as we want to early return even when server-side backup exist but program failed to trust it.
Should we just remove this comment as `checkKeyBackupAndEnable` docs explain that `KeyBackupCheck` will be return only when backup exist and is trusted?

spec/integ/crypto/verification.spec.ts
```diff
@@ -1292,7 +1292,7 @@ describe("verification", () => {
 
             const requestId = await requestPromises.get("m.megolm_backup.v1");
 
-            const infoCopy = Object.assign({}, matchingBackupInfo);
+            const infoCopy = Object.assign({}, nonMatchingBackupInfo);
```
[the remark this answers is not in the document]
ajbura: Also the above test description is a bit unclear by what it means to be *private key*

ajbura:
https://github.com/ajbura/matrix-js-sdk/blob/0c5c7a69e00d483c3d01e3aa09d221765be84b8f/src/rust-crypto/backup.ts#L179-L187

it is a bit weird that we are checking `secret` here and it is also get checked by `checkKeyBackupAndEnable` with under the hood received secret.

spec/integ/crypto/verification.spec.ts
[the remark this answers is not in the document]
richvdh: > Should i just remove this test?

Yes I think it is redundant.

ajbura:
Yeah, you are right i was confused by my own change about `@returns` doc of `checkKeyBackupAndEnable`, but it clear now. I will remove that doc.

If we only want to know the latest backupInfo should we just use the `requestKeyBackupVersion` as it feels a bit confusing to call `checkKeyBackupAndEnable` for version info.

ajbura:
from the workflow error it looks like the testing coverage has dropped from 80% mark. Do I have to cover it up?

richvdh:
> from the workflow error it looks like the testing coverage has dropped from 80% mark. Do I have to cover it up?

It's not that overall test coverage has dropped: the problem is that the coverage of *your changes* is less than 80%. Specifically, you're changing 8 lines but only covering 6 of them:

image

(lines with a grey background are those being changed; those with a red mark on the left are not covered by tests.)

So yes, it would be nice to add another test case: I think we should test the case where `/_matrix/client/v3/room_keys/version` returns an M_NOT_FOUND error. Could you add a new testcase to `verification.spec.ts`?

spec/integ/crypto/verification.spec.ts
```diff
@@ -1264,6 +1264,51 @@ describe("verification", () => {
...
+            // We are lacking a way to signal that the secret has been received, so we wait a bit..
+            jest.useRealTimers();
+            await new Promise((resolve) => {
+                setTimeout(resolve, 500);
+            });
+            jest.useFakeTimers({ doNotFake: ["queueMicrotask"] });
+
+            // the backup secret should not be cached
+            const cachedKey = await aliceClient.getCrypto()!.getSessionBackupPrivateKey();
+            expect(cachedKey).toBeNull();
```
richvdh, line 1288: It might be nice to factor the 5 copies of this code out to a helper function, but I'll not insist

spec/integ/crypto/verification.spec.ts
```diff
@@ -1338,7 +1383,7 @@ describe("verification", () => {
         async function sendBackupGossipAndExpectVersion(
             requestId: string,
             secret: string,
-            expectBackup: KeyBackupInfo,
+            expectBackup?: KeyBackupInfo | { errcode?: string; error?: string },
```
richvdh, line 1386: could you document the behavior of `expectBackup`? What does it mean if it is `undefined`, what if it has an `errcode`, etc?

spec/integ/crypto/verification.spec.ts
```diff
@@ -1264,6 +1264,51 @@ describe("verification", () => {
...
+                setTimeout(resolve, 500);
+            });
+            jest.useFakeTimers({ doNotFake: ["queueMicrotask"] });
+
+            // the backup secret should not be cached
+            const cachedKey = await aliceClient.getCrypto()!.getSessionBackupPrivateKey();
+            expect(cachedKey).toBeNull();
+        });
+
+        it("Should not accept the backup decryption key gossip when server-side key backup request error", async () => {
```
richvdh, line 1291:
```suggestion
        it("Should not accept the backup decryption key gossip when server-side key backup request errors", async () => {
```

spec/integ/crypto/verification.spec.ts
```diff
@@ -1343,26 +1348,37 @@ describe("verification", () => {
...
 
+            // the backup secret should not be cached
+            const cachedKey = await retrieveBackupPrivateKeyWithDelay();
+            expect(cachedKey).toBeNull();
+        });
+
+        /**
+         * Waits briefly to ensure the secret key has been received before retrieving the backup private key.
+         */
+        async function retrieveBackupPrivateKeyWithDelay() {
```
richvdh, line 1359:
```suggestion
        async function retrieveBackupPrivateKeyWithDelay(): Promise<Uint8Array | null> {
```

spec/integ/crypto/verification.spec.ts
```diff
@@ -1343,26 +1348,37 @@ describe("verification", () => {
...
             await sendBackupGossipAndExpectVersion(requestId!, "InvalidSecret", matchingBackupInfo);
 
+            // the backup secret should not be cached
+            const cachedKey = await retrieveBackupPrivateKeyWithDelay();
+            expect(cachedKey).toBeNull();
+        });
+
+        /**
+         * Waits briefly to ensure the secret key has been received before retrieving the backup private key.
+         */
```
richvdh, line 1358:
```suggestion
        /**
         * Waits briefly for secrets to be gossipped, then fetches the backup private key from the crypto stack.
         */
```

spec/integ/crypto/verification.spec.ts
```diff
@@ -1343,26 +1348,37 @@ describe("verification", () => {
...
-            const cachedKey = await aliceClient.getCrypto()!.getSessionBackupPrivateKey();
-            expect(cachedKey).toBeNull();
-        });
+            return aliceClient.getCrypto()!.getSessionBackupPrivateKey();
+        }
 
         /**
          * Common test setup for gossiping secrets.
          * Creates a peer to peer session, sends the secret, mockup the version API, send the secret back from sync, then await for the backup check.
+         * @param expectBackup - The expected result for the key backup request.
```
richvdh, line 1373: Despite its name, it's not really the expected result. Also by convention we leave a blank line before the first `@param:`


```suggestion
         *
         * @param expectBackup - The result to be returned from the `/room_keys/version` request.
```

spec/integ/crypto/verification.spec.ts
```diff
@@ -1264,48 +1265,59 @@ describe("verification", () => {
...
-            jest.useFakeTimers({ doNotFake: ["queueMicrotask"] });
+            await sendBackupGossipAndExpectVersion(
+                requestId!,
+                BACKUP_DECRYPTION_KEY_BASE64,
+                new MatrixError(
+                    {
+                        errcode: "M_NOT_FOUND",
+                        error: "No backup found",
+                    },
+                    401,
```
richvdh, line 1283: shouldn't this be a 404?

spec/integ/crypto/verification.spec.ts
```diff
@@ -1264,48 +1265,59 @@ describe("verification", () => {
...
+                ),
```
richvdh, line 1284: this could be more neatly written:

```
                new MatrixError({ errcode: "M_NOT_FOUND", error: "No backup found" }, 404),
```

richvdh (commented):
Nearly there I think!

spec/integ/crypto/verification.spec.ts
ajbura: improve expectBackup param description

richvdh (approved):
thank you!

:ship:

PR #4695, opened 2025-02-05
`MatrixClient.setAccountData`: await remote echo.

richvdh:
Wait for the echo to come back from the server before we assume the account data has been successfully set.

We have a bit of a problem in the dehydration setup code in that it calls `bootstrapSecretStorage` and, not unreasonably, expects `isSecretStorageReady` to return true. Currently, it doesn't, because `isSecretStorageReady` checks the in-memory account data cache, which isn't updated until the echo comes back down `/sync`.

So, the thing to do seems to be to wait for the echo to come back before proceeding. But I'm not really sure where this fits best: I considered putting it in `ServerSideSecretStorageImpl` or even `bootstrapSecretStorage`, but it really feels like this is a wider problem than just secret storage.

(A prerequisite for https://github.com/element-hq/element-web/issues/29135)

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2200,9 +2200,9 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         this.addListener(ClientEvent.AccountData, accountDataListener);
 
         try {
-            const setAccountDataPromise = retryNetworkOperation(5, () => this.setAccountDataRaw(eventType, content));
-            const res = await Promise.all([setAccountDataPromise, updatedDefer.promise]);
-            return res[0];
+            const result = await retryNetworkOperation(5, () => this.setAccountDataRaw(eventType, content));
+            await updatedDefer.promise;
+            return result;
         } finally {
             this.removeListener(ClientEvent.AccountData, accountDataListener);
         }
--- a/src/client.ts
+++ b/src/client.ts
@@ -52,7 +52,7 @@ import {
     type GroupCallEventHandlerEventHandlerMap,
 } from "./webrtc/groupCallEventHandler.ts";
 import * as utils from "./utils.ts";
-import { noUnsafeEventProps, type QueryDict, replaceParam, safeSet, sleep } from "./utils.ts";
+import { deepCompare, defer, noUnsafeEventProps, type QueryDict, replaceParam, safeSet, sleep } from "./utils.ts";
 import { Direction, EventTimeline } from "./models/event-timeline.ts";
 import { type IActionsObject, PushProcessor } from "./pushprocessor.ts";
 import { AutoDiscovery, type AutoDiscoveryAction } from "./autodiscovery.ts";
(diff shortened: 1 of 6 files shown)


src/client.ts
```diff
@@ -4240,24 +4240,69 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+    public async setAccountData<K extends keyof AccountDataEvents>(
+        eventType: K,
+        content: AccountDataEvents[K] | Record<string, never>,
+    ): Promise<{}> {
+        // If the account data is already correct, then we cannot expect an update over sync, and the operation
+        // is, in any case, a no-op.
+        //
+        // NB that we rely on this operation being synchronous to avoid a race condition: there must be no `await`
+        // between here and `this.addListener` below, in case we miss an update.
+        if (JSON.stringify(this.store.getAccountData(eventType)) === JSON.stringify(content)) return {};
```
t3chguy, line 4262: This isn't guaranteed to work, given the keys aren't being sorted. The key order is based on insertion order for string keys, so will map to JSON-order for JSON-originating content, but for the content provided by the caller that order may not match.

richvdh: ugh, true. I guess we need canonical json, or something

src/client.ts
```diff
@@ -2149,24 +2149,81 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
      */
-    public setAccountData<K extends keyof AccountDataEvents>(
+    public async setAccountData<K extends keyof AccountDataEvents>(
+        eventType: K,
+        content: AccountDataEvents[K] | Record<string, never>,
+    ): Promise<EmptyObject> {
+        // If the sync loop is not running, fall back to setAccountDataRaw.
+        if (!this.clientRunning) {
+            logger.warn(
+                "Calling `setAccountData` before the client is started: `getAccountData` may return inconsistent results.",
```
t3chguy, line 2172: `getAccountData` just wouldn't work, it'd return null or whatever its unknown return is, that is unrelated to you calling `setAccountData` so I think this warning is misleading

src/client.ts
```diff
@@ -2149,24 +2149,81 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            );
+            return await retryNetworkOperation(5, () => this.setAccountDataRaw(eventType, content));
```
t3chguy, line 2174: Why are we awaiting here? There's no error handling to necessitate it

src/client.ts
```diff
@@ -2149,24 +2149,81 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+        }
+
+        // If the account data is already correct, then we cannot expect an update over sync, and the operation
+        // is, in any case, a no-op.
+        //
+        // NB that we rely on this operation being synchronous to avoid a race condition: there must be no `await`
+        // between here and `this.addListener` below, in case we miss an update.
+        const existingData = this.store.getAccountData(eventType);
+        if (existingData && deepCompare(existingData.event.content, content)) return {};
```
t3chguy, line 2183: I thought you were against implicit boolean comparisons, if `existingData` *what*

src/client.ts
richvdh: > `getAccountData` just wouldn't work, it'd return null or whatever its unknown return is,

Well that's my point. It won't return the same thing you've just set.

src/client.ts
richvdh: Not quite sure what you're suggesting here?

src/client.ts
richvdh: I'm not against implicit boolean comparisons; I'm against having a coding style which everyone ignores. The coding style doc has changed.

I can make this more explicit if you want, but imho it's clear enough.

src/client.ts
t3chguy:
```suggestion
            return retryNetworkOperation(5, () => this.setAccountDataRaw(eventType, content));
```

src/client.ts
richvdh: yup, that sounds good.

src/client.ts
```diff
@@ -2149,24 +2149,81 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            // Anyway, what we *shouldn't* do is get stuck in a loop. I think the best we can do is check that the event
+            // type matches.
+            if (event.getType() === eventType) updatedDefer.resolve();
+        }
+        this.addListener(ClientEvent.AccountData, accountDataListener);
+
+        try {
+            const setAccountDataPromise = retryNetworkOperation(5, () => this.setAccountDataRaw(eventType, content));
+            const res = await Promise.all([setAccountDataPromise, updatedDefer.promise]);
+            return res[0];
```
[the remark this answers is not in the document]
richvdh: mmhmm fair.

```suggestion
            const result = await retryNetworkOperation(5, () => this.setAccountDataRaw(eventType, content));
            await updatedDefer.promise;
            return result;
```

src/client.ts
richvdh: oh ok.

I feel like some coding styles advise against having async functions that return a Promise. Apart from anything else, it saves you from surprises in the case where there *is* a `try/catch` (or one is added later).

I'm not a huge fan of returning Promises from from async functions, but I don't feel that strongly about it.

src/client.ts
richvdh: At the very least: we should have methods that allow you to wait for the operation to be confirmed.

My first instinct here was to add a new `setAccountDataAndAwaitRemoteEcho` or something, but thanks to the way `ServerSideSecretStorage` is implemented, using it would have meant breaking changes to the library API, so I lost enthusiasm for that.

t3chguy (approved):

andybalaam (approved):

Issue #4702, opened 2025-02-09
Misleading Type: `UIAResponse<T>`, `IAuthData` is not valid response


ajbura:
link to problem area: https://github.com/matrix-org/matrix-js-sdk/blob/30d9b0518fd51f0f591e14e57ef27391c28f0f6e/src/%40types/uia.ts#L29

the underline request the `UIAResponse<T>` is paired with never return 401 response as response:
https://github.com/matrix-org/matrix-js-sdk/blob/30d9b0518fd51f0f591e14e57ef27391c28f0f6e/src/http-api/fetch.ts#L338-L340

and instead throw it as `MatrixError` or `HTTPError`.

The type `UIAResponse<T>` should not exist.

PR #4709, opened 2025-02-13
Improve types around User Interactive Auth

t3chguy:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/4702
Requires https://github.com/element-hq/element-web/pull/29255

### DIFF
--- a/src/@types/uia.ts
+++ b/src/@types/uia.ts
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { type AuthDict, type IAuthData } from "../interactive-auth.ts";
+import { type AuthDict } from "../interactive-auth.ts";
 
 /**
  * Helper type to represent HTTP request body for a UIA enabled endpoint
@@ -22,8 +22,3 @@ import { type AuthDict, type IAuthData } from "../interactive-auth.ts";
 export type UIARequest<T> = T & {
     auth?: AuthDict;
 };
-
-/**
- * Helper type to represent HTTP response body for a UIA enabled endpoint
- */
-export type UIAResponse<T> = T | IAuthData;
--- a/src/@types/uia.ts
+++ b/src/@types/uia.ts
@@ -22,3 +22,9 @@ import { type AuthDict } from "../interactive-auth.ts";
 export type UIARequest<T> = T & {
     auth?: AuthDict;
 };
+
+/**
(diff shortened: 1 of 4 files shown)


MidhunSureshR (approved):

richvdh (commented):
Looks sensible. Could you update the PR title to be something that will make more sense in a changelog?

src/@types/uia.ts
```diff
@@ -14,16 +14,11 @@ See the License for the specific language governing permissions and
...
  * Helper type to represent HTTP request body for a UIA enabled endpoint
  */
 export type UIARequest<T> = T & {
     auth?: AuthDict;
 };
-
-/**
- * Helper type to represent HTTP response body for a UIA enabled endpoint
- */
-export type UIAResponse<T> = T | IAuthData;
```
richvdh, line 29: wondering if we should (temporarily) make this 

```
/** @deprecated */
export type UIAResponse<T> = T;
```

... to avoid having to synchronise the EW and js-sdk PRs.

No strong feelings though.

t3chguy: We shouldn't need to synchronise them, EW can land first, shouldn't need to both land at once.

richvdh: ah nice.

richvdh (approved):
LGTM

richvdh:
> Requires #4709

wrong link, I think

src/interactive-auth.ts
```diff
@@ -159,7 +158,7 @@ export class NoAuthFlowFoundError extends Error {
  *
  * The generic parameter `T` is the type of the response of the endpoint, once it is eventually successful.
  */
-export type UIAuthCallback<T> = (makeRequest: (authData: AuthDict | null) => Promise<UIAResponse<T>>) => Promise<T>;
+export type UIAuthCallback<T> = (makeRequest: (authData: AuthDict | null) => Promise<T>) => Promise<T>;
```
richvdh, line 161: It would be nice, while you're here (and since it took me a while to figure out), to clarify the documentation to say that, in the case of more auth being needed, `makeRequest` will throw a `MatrixError` whose `data` conforms to  `IAuthData`. But I don't mind if you want to leave that for another time.

src/@types/uia.ts
t3chguy: Actually no, you're right, that is a better approach and will avoid a breaking change

Issue #4570, opened 2024-12-03
SQLite store implementation effort


theobouwman:
Hi,

We are currently using matrix js sdk in our react native app. As indexeddb is not supported in RN we currently use the memory store.

The problem is that when you open a notification from a room there is a loading state because nothin is stored on device. We are using TimelineWindow for our rooms with pagination etc https://matrix-org.github.io/matrix-js-sdk/classes/matrix.TimelineWindow.html.

So I have 2 questions:
- 1) What would some kind of solution be to store the 40 recent events of a room, and load it in the TimelineWindow?
- 2) as (1) would be quick workaround I am considering contributing to the sdk to implement a SQLite store implantation. Would this mean to just implement the IStore in [https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/store/index.ts ](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/store/index.ts) methods to a SQLite implantation?

Thanks

richvdh:
@theobouwman it's worth knowing that there are *two* stores in matrix-js-sdk: the `IStore` which you have found, but the crypto stack has its own, separate, store. If you want to support end-to-end encryption, you'll also need to update the crypto stack to support a different database implementation. https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/168 has some discussion about that.

theobouwman:
@richvdh I know. I am first planning to implement the store without the end-to-end encryption.

But, by implementing the IStore (https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/store/index.ts) should be a good start right?

theobouwman:
Okay. I made a start. 
Unfortunately I am getting a `Caught /sync error [TypeError: undefined is not a function]` on sync.

```
  storedFilters set filter:@a09a125f-0346-4b98-a81c-5439d927b8bf:test.exploremomo.com:0 [{"definition": {"room": [Object]}, "filterId": "0", "roomFilter": [Object], "roomTimelineFilter": [Object], "userId": "@a09a125f-0346-4b98-a81c-5439d927b8bf:test.exploremomo.com"}]
 LOG  set filter id by name with filterId FILTER_SYNC_@a09a125f-0346-4b98-a81c-5439d927b8bf:test.exploremomo.com 0
 DEBUG  Sending initial sync request...
[...]

PR #4712, opened 2025-02-16
Mention caveat in README about IndexedDB outside browser

aspiers:
If using the SDK outside the browser with end-to-end encryption, IndexedDB usage needs to be disabled otherwise it will cause the Rust code to panic.

See also:

- https://github.com/matrix-org/matrix-js-sdk/issues/4570
- https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/168
- https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/issues/195#issuecomment-2661510343

Signed-off-by: Adam Spiers <github@adamspiers.org>

```
Notes: none
```

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,6 +321,8 @@ const matrixClient = sdk.createClient({
 await matrixClient.initRustCrypto();
 ```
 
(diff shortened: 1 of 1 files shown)


aspiers:
Looks like this needs a label setting to pass the Preview Changelog CI:

image

But I don't have access to set one.  Hopefully a team member can assist with that.

README.md
```diff
@@ -321,6 +321,8 @@ const matrixClient = sdk.createClient({
 await matrixClient.initRustCrypto();
 ```
 
+Note that by default it will attempt to use the Indexed DB provided by the browser as a crypto store. If running outside the browser, you will need to pass [an options object](https://matrix-org.github.io/matrix-js-sdk/classes/matrix.MatrixClient.html#initrustcrypto) which includes `useIndexedDB: false`, to use an ephemeral in-memory store instead.
```
richvdh, line 324: Thanks for this.  I think if we're going to suggest this, we'd also better point out the pitfalls of not using a persistent store: specifically, that you'll need to create a new device on the server side (with `MatrixClient.loginRequest`) each time your application starts.

aspiers: OK thanks for the info - I've amended the PR to include that.

PR #4713, opened 2025-02-17
MatrixRTC: MembershipManager test cases and deprecation of MatrixRTCSession.room

toger5:
Part of https://github.com/element-hq/element-call/issues/2972.

This shouldn't introduce any behaviour changes, but refactor the tests to be based on the MembershipManager abstraction. It includes some tests which would fail on the current LegacyMembershipManager implementation which will be handled by a new implementation.

The field `MatrixRTCSession.room` becomes deprecated as the `Room` should be tracked by the client outside of the `MatrixRTCSession`.

## Checklist

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -233,12 +233,14 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
     /**
      * WARN: this can in theory only be a subset of the room with the properties required by
      * this class.
-     * Outside of tests this most likely will be a full room however.
-     * @deprecated
+     * Outside of tests this most likely will be a full room, however.
+     * @deprecated Relying on a full Room object being available here is an anti-pattern. You should be tracking
+     * the room object in your own code and passing it in when needed.
      */
     public get room(): Room {
         return this.roomSubset as Room;
     }
+
     /**
      * This constructs a room session. When using MatrixRTC inside the js-sdk this is expected
      * to be used with the MatrixRTCSessionManager exclusively.
@@ -247,7 +249,7 @@ export class MatrixRTCSession extends TypedEventEmitter<MatrixRTCSessionEvent, M
      * to manage a joined MatrixRTC session.
      *
      * @param client A subset of the {@link MatrixClient} that lets the session interact with the Matrix room.
-     * @param room The room this session is attached to. A subset of a js-sdk Room that the session needs.
+     * @param roomSubset The room this session is attached to. A subset of a js-sdk Room that the session needs.
      * @param memberships The list of memberships this session currently has.
      */
     public constructor(


spec/unit/matrixrtc/MembeshipManager.spec.ts
```diff
@@ -1,3 +1,6 @@
+/**
```
hughns, line 1: I realise that this file has a spelling mistake in it. Needs fixing.

src/client.ts
```diff
@@ -502,7 +502,7 @@ export interface IStartClientOpts {
     /**
```
hughns, line 502: These are all good changes, but I suggest removing them from this PR and putting them into a separate one just to make it easier to review.

hughns (changes requested):
Just a few quick things. I haven't looked at the main test cases yet.

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,588 @@
...
+                            undefined,
+                            new Headers({ "Retry-After": "1" }),
+                        ),
+                    );
+                    await flushPromises();
+                    jest.advanceTimersByTime(1000);
+                    await flushPromises();
+                    expect(client._unstable_sendDelayedStateEvent).toHaveBeenCalledTimes(2);
+                });
+                it("abandons retry loop and sends new own membership if not present anymore !FailsForLegacy", async () => {
```
hughns, line 489:
```suggestion
                // FailsForLegacy as implementation does not re-check membership before retrying
                it("abandons retry loop and sends new own membership if not present anymore !FailsForLegacy", async () => {
```

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,588 @@
...
+                    // the membership is no longer present on the homeserver
+                    manager.onRTCSessionMemberUpdate([]);
+                    // Wait for all timers to be setup
+                    await flushPromises();
+                    jest.advanceTimersByTime(1000);
+                    // We should send a new own membership and a new delayed event after the rate limit timeout.
+                    expect(client._unstable_sendDelayedStateEvent).toHaveBeenCalledTimes(2);
+                    expect(client.sendStateEvent).toHaveBeenCalledTimes(2);
+                });
+                it("abandons retry loop if leave() was called !FailsForLegacy", async () => {
```
hughns, line 516:
```suggestion
                // FailsForLegacy as implementation does not re-check membership before retrying
                it("abandons retry loop if leave() was called !FailsForLegacy", async () => {
```

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,588 @@
...
+                    // flush promises before advancing the timers to make sure schdulers are setup
+                    await flushPromises();
+                    jest.advanceTimersByTime(10_000);
+                    expect(client._unstable_updateDelayedEvent).toHaveBeenCalledTimes(i);
+                    // TODO: check that update delayed event is called with the correct HTTP request timeout
+                    // expect(client._unstable_updateDelayedEvent).toHaveBeenLastCalledWith("id", 10_000, { localTimeoutMs: 20_000 });
+                }
+            });
+
+            it("extends `expires` when call still active !FailsForLegacy", async () => {
```
hughns, line 442: Can we add a comment to say why this FailsForLegacy? I think you mentioned that the code was removed at some point? Could we reference the commit?

hughns (changes requested):

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,588 @@
...
+                        const sendDelayedStateAttempt = new Promise<void>((resolve) => {
+                            const error = new MatrixError({ errcode: "M_LIMIT_EXCEEDED" });
+                            (client._unstable_sendDelayedStateEvent as Mock).mockImplementationOnce(() => {
+                                resolve();
+                                return Promise.reject(error);
+                            });
+                        });
+
+                        // setting the membership state should handle ratelimiting (also with a retry-after value)
+                        const sendStateEventAttempt = new Promise<void>((resolve) => {
```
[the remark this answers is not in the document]
toger5: It is part of the `sends a membership event` section right?

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,588 @@
...
+
+        describe("getsActiveFocus", () => {
+            it("gets the correct active focus with oldest_membership !FailsForLegacy", () => {
+                const getOlderstMembership = jest.fn();
+                const manager = new TestMembershipManager({}, room, client, getOlderstMembership);
+                // Before joining the active focus should be undefined (see FocusInUse on MatrixRTCSession)
+                expect(manager.getActiveFocus()).toBe(undefined);
+                manager.join([focus], focusActive);
+                // after joining we want our own focus to be the once we select
+                expect(manager.getActiveFocus()).toBe(focus);
```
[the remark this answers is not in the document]
toger5: We want the mock to represent what we actually would get from the getOldestMembership call. 
This is however not part of this logic but of the rtc session class. So we mock that the return value of the getOldestMembership call changes.

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,588 @@
...
+            it("does nothing if not joined !FailsForLegacy", () => {
+                const manager = new TestMembershipManager({}, room, client, () => undefined);
+                manager.leave();
+                expect(client._unstable_sendDelayedStateEvent).not.toHaveBeenCalled();
+                expect(client.sendStateEvent).not.toHaveBeenCalled();
+            });
+        });
+
+        describe("getsActiveFocus", () => {
+            it("gets the correct active focus with oldest_membership !FailsForLegacy", () => {
```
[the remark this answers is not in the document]
toger5: Yes there was an issue with the test. I worked on that again for the new one and remove !FailsForLegacy.

package.json
```diff
@@ -80,6 +80,7 @@
         "@babel/preset-env": "^7.12.11",
         "@babel/preset-typescript": "^7.12.7",
         "@casualbot/jest-sonar-reporter": "2.2.7",
+        "@jest/environment": "^29.7.0",
```
dbkr, line 83: What's this for? As far as I can see it's an empty package.

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,611 @@
...
+*/
+
+import { type Mock } from "jest-mock";
+import waitForExpect from "wait-for-expect";
+
+import { EventType, HTTPError, MatrixError, type Room } from "../../../src";
+import { type Focus, type LivekitFocusActive, type SessionMembershipData } from "../../../src/matrixrtc";
+import { LegacyMembershipManager } from "../../../src/matrixrtc/MembershipManager";
+import { makeMockClient, makeMockRoom, membershipTemplate, mockCallMembership, type MockClient } from "./mocks";
+import { flushPromises } from "../../test-utils/flushPromises";
```
dbkr, line 27: If this isn't marked deprecated, it should be. It's much better to find something concrete to wait for.

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,611 @@
...
+// import { MembershipManager } from "../../../src/matrixrtc/NewMembershipManager";
+
+function waitForMockCall(method: any, returnVal?: any) {
+    return new Promise<void>((resolve) => {
+        (method as Mock).mockImplementation(() => {
+            resolve();
+            return returnVal;
+        });
+    });
+}
```
dbkr, line 37: An alternative would be to just `await waitFor(() => expect(mock.toHaveBeenCalled()))` and then you don't have to interfere with the mock impl. Non blocking though.

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,611 @@
...
+
+function createAsyncHandle(method: any) {
```
dbkr, line 39: `Mock` type maybe, then you don't have to cast it?

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,611 @@
...
+            livekit_alias: "!active:active.url",
+        };
+
+        beforeEach(() => {
+            // default to fake timers
+            jest.useFakeTimers();
+            client = makeMockClient("@alice:example.org", "AAAAAAA");
+            room = makeMockRoom(membershipTemplate);
+            // provide a default mock that is like the default "non error" server behaviour.
+            (client._unstable_sendDelayedStateEvent as Mock).mockReturnValue({ delay_id: "id" });
```
dbkr, line 79: the `mocked()` function might be what you want here rather than manual casting (and in the other places)

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,611 @@
...
+            });
+        });
+
+        describe("join()", () => {
+            describe("sends a membership event", () => {
+                it("sends a membership event and schedules delayed leave when joining a call", async () => {
+                    // Spys/Mocks
+
+                    // eslint-disable-next-line camelcase
+                    const _unstable_updateDelayedEventHandle = createAsyncHandle(
```
dbkr, line 106: Not sure I understand the reason for the weird name here

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,611 @@
...
+}
+
+function createAsyncHandle(method: any) {
+    const handle: { resolve?: (...args: unknown[]) => void; reject?: (...args: any[]) => void } = {};
+    (method as Mock).mockImplementation(() => {
+        return new Promise((resolve, reject) => {
+            handle.reject = reject;
+            handle.resolve = resolve;
+        });
+    });
```
dbkr, line 46: utils/defer might also save you a line or two here. Also maybe the name should involve the word 'defer' as that's essentially what this is and I got a bit confused by 'handle'.

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,611 @@
...
+            type: "livekit",
+        };
+        const focus: Focus = {
+            type: "livekit",
+            livekit_service_url: "https://active.url",
+            livekit_alias: "!active:active.url",
+        };
+
+        beforeEach(() => {
+            // default to fake timers
```
dbkr, line 74: Warning: there's a good chance someone won't notice this is happening and it will bite them in weird ways because something that uses a short setTimeout delay won't run automatically.

spec/unit/matrixrtc/testEnvironment.ts
```diff
@@ -0,0 +1,58 @@
+/*
+Copyright 2023 The Matrix.org Foundation C.I.C.
```
dbkr, line 2: Is this code really 2 years old? ;)

package.json
```diff
@@ -124,7 +125,8 @@
         "typedoc-plugin-coverage": "^3.0.0",
         "typedoc-plugin-mdn-links": "^4.0.0",
         "typedoc-plugin-missing-exports": "^3.0.0",
-        "typescript": "^5.4.2"
+        "typescript": "^5.4.2",
+        "wait-for-expect": "^3.0.2"
```
dbkr, line 129: I think jest's `waitFor` is what you want here, this hasn't been updated in 5 years.

package.json
toger5: its for the matrixRTC test environment: `matrix-js-sdk/spec/unit/matrixrtc/testEnvironment.ts`

spec/unit/matrixrtc/MembershipManager.spec.ts
toger5: I do have to cast it when passing `client.sendStateEvent` though. Choosing any here saves me from casting it in multiple places.

toger5: But your other comment on using `mocked(client.sendStateEvent)` might be the way to go.

spec/unit/matrixrtc/MembershipManager.spec.ts
toger5: its `fn_name`+`Handle`. What do you think would be better? just `updateDelayedEventHandle`?

spec/unit/matrixrtc/testEnvironment.ts
toger5: Its not ...

package.json
toger5: this would require me to import add testing-library-dom?
I think right now we do not have `waitFor` in the js-sdk?

spec/unit/matrixrtc/MembershipManager.spec.ts
toger5: this looks nice. But needs us to depend on testing-library-dom (or just testing-library?)

spec/unit/matrixrtc/MembershipManager.spec.ts
toger5: The whole file will only use FakeTimers. So I think there are enough examples with `jest.advacenTimers...` so that will quickly resolve?

spec/unit/matrixrtc/MembershipManager.spec.ts
toger5: Changed it to MockedFunction<Any>

spec/unit/matrixrtc/memberManagerTestEnvironment.ts
```diff
@@ -0,0 +1,58 @@
...
+@jest-environment ./spec/unit/matrixrtc/memberManagerTestEnvironment.ts
+
+It is very specific to the MembershipManager.spec.ts file and introduces the following behaviour:
+ - The describe each block in the MembershipManager.spec.ts will go through describe block names `LegacyMembershipManager` and `MembershipManager`
+ - It will check all tests that are a child or indirect child of the `LegacyMembershipManager` block and skip the ones which include "!FailsForLegacy"
+   in their test name.
+*/
+
+import { TestEnvironment } from "jest-environment-jsdom";
+import { type JestEnvironmentConfig, type EnvironmentContext } from "@jest/environment";
```
hughns, line 30: Re: https://github.com/matrix-org/matrix-js-sdk/pull/4713/files#r1969592021 - remove this import statement

spec/unit/matrixrtc/memberManagerTestEnvironment.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { logger } from "../../../src/logger";
+
+class CustomEnvironment extends TestEnvironment {
+    constructor(config: JestEnvironmentConfig, context: EnvironmentContext) {
```
hughns, line 35: Re: https://github.com/matrix-org/matrix-js-sdk/pull/4713/files#r1969592021 - remove this constructor

toger5: You are right this is the same as the default impl!

spec/unit/matrixrtc/MembershipManager.spec.ts
hughns: Would something like this work using fake timers?

```
            it("does nothing if not joined", async () => {
                const manager = new TestMembershipManager({}, room, client, () => undefined);
                manager.onRTCSessionMemberUpdate([mockCallMembership(membershipTemplate, room.roomId)]);
            		await jest.advanceTimersByTimeAsync(<insert appropriate amount of time>;
                expect(client.sendStateEvent).not.toHaveBeenCalled();
                expect(client._unstable_sendDelayedStateEvent).not.toHaveBeenCalled();
                expect(client._unstable_updateDelayedEvent).not.toHaveBeenCalled();
            });
```

The amount of time could be related to the maximum timer that could feasibly have been set? (Probably the timer that would extend the membership `expiry` field... so maybe an arbitrary 24 hours would be appropriate for the test. i.e. it does nothing immediately nor for at least 24 hours)

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,625 @@
...
+                room.roomId,
+                "org.matrix.msc3401.call.member",
+                {},
+                "_@alice:example.org_AAAAAAA",
+            );
+        });
+        // FailsForLegacy because legacy implementation always sends the empty state event even though it isn't needed
+        it("does nothing if not joined !FailsForLegacy", () => {
+            const manager = new TestMembershipManager({}, room, client, () => undefined);
+            manager.leave();
```
hughns, line 347: For consistency with the other tests should this be awaited?

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -0,0 +1,625 @@
...
+        // TODO add rate limit cases.
+        it("resolves delayed leave event when leave is called", async () => {
+            const manager = new TestMembershipManager({}, room, client, () => undefined);
+            manager.join([focus], focusActive);
+            await flushPromises();
+            await manager.leave();
+            expect(client._unstable_updateDelayedEvent).toHaveBeenLastCalledWith("id", "send");
+            expect(client.sendStateEvent).toHaveBeenCalled();
+        });
+        it("send leave event when leave is called and resolving delayed leave fails", async () => {
```
hughns, line 330:
```suggestion
        it("sends leave event when leave is called and resolving delayed leave fails", async () => {
```

hughns (changes requested):
This review is for the new `leave()` test cases added by https://github.com/matrix-org/matrix-js-sdk/commit/4e60a35d54d540217eb1f406b0bcd332da64e3ee

Aside from this I have added in some comments on the open threads.

spec/unit/matrixrtc/MembershipManager.spec.ts
dbkr: I'm not convinced flushPromises is really a great way of testign this since it doesn't actually guarantee all pending async things have been run. If there's something that must necessarily happen after the event would have been emitted, you could wait for that to happen and then assert that the other thing has not. Otherwise, I think your best bet is probably just to have an explicit timeout, probably for a longer amount of time like 100ms.

hughns (approved):

hughns:
The latest commits add a deprecation, so I've updated the title, labels, description.

As such, I'm re-submitting for review from Web Team.

dbkr (approved):
Fair enough: seems reasonable

PR #4726, opened 2025-02-19
MatrixRTC: New membership manager

toger5:
Fixes: https://github.com/element-hq/voip-internal/issues/309
Part of https://github.com/element-hq/element-call/issues/2972.

This does not yet use the new manager by default. It only runs the test suite with it.

Notes on error cases encountered when making HTTP requests to the Homeserver:

Type | `MatrixClient` stack | `RoomWidgetClient` stack
-|-|-
"Local" request timeout enforced by the Matrix client | [throws](https://github.com/matrix-org/matrix-js-sdk/blob/dabb8aab22083eb30b928ed53d6493a84acf6593/src/http-api/fetch.ts#L330-L332) `Error` with `name` === `AbortError` | ?
HTTP 5xx response from homeserver | [throws](https://github.com/matrix-org/matrix-js-sdk/blob/dabb8aab22083eb30b928ed53d6493a84acf6593/src/http-api/fetch.ts#L339) [`HttpError`](https://github.com/matrix-org/matrix-js-sdk/blob/71bffb6c1bb89b5d9802667bee2626e29a4f57d5/src/http-api/utils.ts#L103-L106) | throws `MatrixError`
Network/socket error/timeout | [throws](https://github.com/matrix-org/matrix-js-sdk/blob/dabb8aab22083eb30b928ed53d6493a84acf6593/src/http-api/fetch.ts#L333) `ConnectionError` | ? 
Request timeout between widget and host Matrix client | n/a | now mapped to `ConnectionError` as part of this PR

## Checklist


hughns (commented):
Some minor tidying things for now.

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,808 @@
...
+                        }
+                    }
+                } else {
+                    this.scheduler.addAction({ ts: Date.now(), type: MembershipActionType.SendLeaveEvent });
+                }
+                break;
+        }
+    }
+
+    public async membershipLoopHandler(state: ActionSchedulerState, type: MembershipActionType): Promise<void> {
```
hughns, line 611: `public`? surely not?

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,808 @@
...
+        switch (type) {
```
hughns, line 612: I would find it slightly easier to read this if it followed the lifecycle ordering: `SendJoinEvent`, `Update`, `SendLeaveEvent`

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,808 @@
...
+                }
+                break;
+            case MembershipActionType.SendLeaveEvent:
+                // We are good already
+                if (!state.hasMemberStateEvent) return;
+
+                // This is only a fallback in case we do not have working delayed events support.
+                // first we should try to just send the scheduled leave event
+                try {
+                    this.client.sendStateEvent(
```
hughns, line 697: Missing `await`?

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,808 @@
...
+                        this.room.roomId,
+                        EventType.GroupCallMemberPrefix,
+                        {},
+                        this.makeMembershipStateKey(this.userId, this.deviceId),
```
hughns, line 701: Can this be `this.stateKey` for consistency?

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,808 @@
...
+                this.didWakeUp = false;
+            } else {
+                try {
+                    if (isDelayedLeaveActionType(nextAction.type)) {
+                        await this.manager.delayedLeaveLoopHandler(this.state, nextAction.type);
+                    } else if (isMembershipActionType(nextAction.type)) {
+                        await this.manager.membershipLoopHandler(this.state, nextAction.type);
+                    }
+                } catch (e) {
+                    throw Error("The MemberhsipManager has to shut down because of the end condition: " + e);
```
hughns, line 191: A few thoughts;

- I'm surprised we aren't doing any cleanup of the membership state
- If the intention is that this will only ever be reached when an explicit "end condition" throw was made, then should we be checking the type of error and only terminate if it was of that type
- In the current implementations of the loop handlers there aren't general purpose catch blocks, so I think any unanticipated error would hit this catch block

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,808 @@
...
+                }
+            }
+
+            if (this.resetWith) {
+                this.actions = this.resetWith;
+                this.resetWith = undefined;
+            }
+
+            this.actions = this.actions.filter((a) => a !== nextAction);
```
hughns, line 200: For some reason I had expected that you would I had expected that you would have popped the original action out of the array and then you wouldn't have to filter here.  🤷‍♂️

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,808 @@
...
+                                delay: this.membershipServerSideExpiryTimeout,
+                            },
+                            EventType.GroupCallMemberPrefix,
+                            {}, // leave event
+                            this.stateKey,
+                        );
+                        // Success we reset retires and set delayId.
+                        state.sendDelayedEventRetries = 0;
+                        state.delayId = response.delay_id;
+                        this.scheduler.addAction({ ts: Date.now(), type: MembershipActionType.SendJoinEvent });
```
hughns, line 458: I *think* this is the only place where the `SendJoinEvent` action is scheduled. So, what happens if the server doesn't support the delayed events feature? At the moment I don't *think* `SendJoinEvent` gets run.

src/matrixrtc/NewMembershipManager.ts
toger5: I will look into the future lint pr today!

src/matrixrtc/NewMembershipManager.ts
toger5: yes I really need to go through all catch blocks and think about the most reasonable "default" failure mode.
Asking the question: "Can we do the same thing in ALL other failure cases and what would that message be?"
I think that will expose a couple of other specific errors we need to handle specifically and what we need to do in the "default" case.

toger5: I went through all the catch blocks and opted for throwing in a lot of cases.
Only on the actual event sending I opted for a blind "retry" even though we do not understand the error.
But this blind retry is now also rate limited.

hughns (changes requested):
I am finding the revised catch block structure much more readable 👍 

A few things:

1. I think chunks like this are misleading as the `throw` is unreachable:

```ts
// Event sending retry (different to rate limit retries)
if (this.retryOnAnyErrorHandler(e, type)) break;

throw Error("Could not send state event because of unrecoverable error: " + e);
```

I think something like this would be clearer:

```ts
// Event sending retry (different to rate limit retries)
// throws once retry count is reached
this.retryOnAnyErrorHandler(e, type);
```

2. I saw you changed the scheduler implementation to use `pop()` and then back again. Did it not work out?

3. For some actions network and server errors (e.g. socket level timeouts, and HTTP level) will be handled by the `retryOnAnyErrorHandler()` but there are still some actions where we don't:
 - `MembershipActionType.SendMainDelayedEvent` - we throw a fatal Error
 - `MembershipActionType.RestartDelayedEvent` - we throw a fatal Error
 - `MembershipActionType.SendFirstDelayedEvent` - we silently fail (I've suggested adding some logging at a minimum)

I would like to see these handled.

4. For general debugging of this, should we be logging the errors at the start of each `catch` block?

src/matrixrtc/NewMembershipManager.ts
hughns: Wrap it like this?

```ts
class ActionScheduler {
    ...
    public constructor(
        state: ActionSchedulerState,
        private membershipLoopHandler: (state: ActionSchedulerState, type: MembershipActionType) => Promise<void>,
    ) {
        this.state = state;
    }
	...
    public async startWithActions(initialActions: Action[]): Promise<void> {
		...
                try {
                    await this.membershipLoopHandler(this.state, nextAction.type as MembershipActionType);
                } catch (e) {
                    throw Error("The MemberhsipManager has to shut down because of the end condition: " + e);
                }
		...
    }
	...

}
...
export class MembershipManager implements IMembershipManager {
	...
	// Scheduler:
    private scheduler = new ActionScheduler(
        {
            hasMemberStateEvent: false,
            running: false,
            nextRelativeExpiry: this.membershipEventExpiryTimeout,
            delayId: undefined,
            rateLimitRetries: 0,
            retries: 0,
        },
        (state, type) => this.membershipLoopHandler(state, type),
    );

    // Loop Handler:
    private async membershipLoopHandler(state: ActionSchedulerState, type: MembershipActionType): Promise<void> {
	...
}
```

src/matrixrtc/NewMembershipManager.ts
toger5: This got another update in 5a680b0e9fa43024d93fbad9f675d44b6c0585ac where I added tests for those cases

src/matrixrtc/NewMembershipManager.ts
toger5: TLDR: I require the following order: get first element, read this element from the array where i cannot be removed in anther codeblock, then it needs to be removed.
pop makes this atomic but it is required to be two steps.

There are of course workarounds like:
```
this.currentItem = this.array.pop()
getArrayWithCurrent() return [...this.array, this.currentItem]
```
And then only use getArrayWithCurrent but I did not prefer any of the alternatives to what it is now.

src/matrixrtc/NewMembershipManager.ts
toger5: There now is `scheduler.resetState()`

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -338,9 +339,11 @@ describe.each([
             );
         });
         // FailsForLegacy because legacy implementation always sends the empty state event even though it isn't needed
-        it("does nothing if not joined !FailsForLegacy", () => {
+        it("does nothing if not joined !FailsForLegacy", async () => {
             const manager = new TestMembershipManager({}, room, client, () => undefined);
-            manager.leave();
+            const errorFn = jest.fn();
+            await manager.leave().catch(errorFn);
```
hughns, line 345: I think you can do something like this instead:

```
expect(await manager.leave()).not.toThrow()
```

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -648,13 +647,34 @@ describe.each([
...
+            const unrecoverableError = jest.fn();
+            (client._unstable_sendDelayedStateEvent as Mock<any>).mockRejectedValue(new HTTPError("unknown", 501));
+            const manager = new TestMembershipManager({}, room, client, () => undefined);
+            manager.join([focus], focusActive, unrecoverableError);
+            await jest.advanceTimersByTimeAsync(1);
+
+            expect(unrecoverableError).not.toHaveBeenCalledWith();
+            expect(client.sendStateEvent).toHaveBeenCalled();
+        });
+        it("UnsupportedEndpointError does not in an unrecoverable error. We use the manager without delayed events !FailsForLegacy", async () => {
```
hughns, line 667:
```suggestion
        it("falls back to using pure state events when UnsupportedEndpointError encountered for delayed events !FailsForLegacy", async () => {
```

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -648,13 +647,34 @@ describe.each([
...
-                await flushPromises();
+            manager.join([focus], focusActive, delayEventRestartError);
+
+            for (let i = 0; i < 10; i++) {
+                await jest.advanceTimersByTimeAsync(1000);
             }
             expect(delayEventRestartError).toHaveBeenCalled();
-        }, 5000);
+        });
+        it("Errors while sending delayed events dont result in an unrecoverable error. We use the manager without delayed events !FailsForLegacy", async () => {
```
hughns, line 657:
```suggestion
        it("falls back to using pure state events when some error occurs while sending delayed events !FailsForLegacy", async () => {
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,859 @@
...
+import { HTTPError, MatrixError } from "../http-api/errors.ts";
+import { logger as rootLogger } from "../logger.ts";
+import { type Room } from "../models/room.ts";
+import { sleep } from "../utils.ts";
+import { type CallMembership, DEFAULT_EXPIRE_DURATION, type SessionMembershipData } from "./CallMembership.ts";
+import { type Focus } from "./focus.ts";
+import { isLivekitFocusActive } from "./LivekitFocus.ts";
+import { type MembershipConfig } from "./MatrixRTCSession.ts";
+
+const logger = rootLogger.getChild("MatrixRTCSessionManager");
```
hughns, line 30: My bad, I realise that this is part of MatrixRTCSession not MatrixRTCSessionManager:

```suggestion
const logger = rootLogger.getChild("MatrixRTCSession");
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,859 @@
...
+                            this.room.roomId,
+                            {
+                                delay: this.membershipServerSideExpiryTimeout,
+                            },
+                            EventType.GroupCallMemberPrefix,
+                            {}, // leave event
+                            this.stateKey,
+                        )
+                        .then((response) => {
+                            // Success we reset retires and set delayId.
```
hughns, line 469:
```suggestion
                            // Success we reset retries and set delayId.
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,859 @@
...
+                                    ts: Date.now(),
+                                    type: MembershipActionType.SendJoinEvent,
+                                });
+                                return;
+                            }
+                            return e;
+                        });
+                    // On any other error we fall back to not using delayed events and send the join state event immediately
+                    if (error) {
+                        logger.info("Not using deleayed event because: " + error);
```
hughns, line 496:
```suggestion
                        logger.info("Not using delayed event because: " + error);
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,859 @@
...
+
+    public isJoined(): boolean {
+        return this.scheduler.state.running;
+    }
+    /**
+     * Puts the MembershipManager in a state where it tries to be joined.
+     * It will send delayed events and membership events
+     * @param fociPreferred
+     * @param focusActive
+     * @param onError This will be called once the membership menager encounters an unrecoverable error.
```
hughns, line 303:
```suggestion
     * @param onError This will be called once the membership manager encounters an unrecoverable error.
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,859 @@
...
+ * It has the following tasks:
+ *  - Send the users leave delayed event before sending the memberhsip
+ *  - Sent the users membership if the state machine is started
+ *  - Check if the delayed event was canceled due to sending the membership
+ *  - update the delayed event (`restart`)
+ *  - Update the state event every ~5h = `DEFAULT_EXPIRE_DURATION` (so it does not get treated as expired)
+ *  - When the state machine is stopped:
+ *   - Disconnect the member
+ *   - Stop the timer for the delay refresh
+ *   - Stop the timer for updateint the state event
```
hughns, line 290:
```suggestion
 *   - Stop the timer for updating the state event
```

src/matrixrtc/NewMembershipManager.ts
hughns:
```suggestion
                    throw Error("The MembershipManager has to shut down because of the end condition: " + e);
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,859 @@
...
+    // state variables for a wakeup mechanism (in case we add some action externally and need to leave the current sleep)
+    private wakeupPromise?: Promise<void>;
+    private wakeup?: (value: void | PromiseLike<void>) => void;
+    private didWakeUp = false;
+
+    private actions: Action[] = [];
+    private insertions: Action[] = [];
+    private resetWith?: Action[];
+    /**
+     * This starts the main loop of the memberhsip manager that handles event sending, delayed event sending and delayed event restarting.
```
hughns, line 213:
```suggestion
     * This starts the main loop of the membership manager that handles event sending, delayed event sending and delayed event restarting.
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,859 @@
...
+     * @param initialActions The initial actions the manager will start with. It should be enough to pass: DelayedLeaveActionType.Initial
```
hughns, line 214:
```suggestion
     * @param initialActions The initial actions the manager will start with. It should be enough to pass: DelayedLeaveActionType.Initial
     * @returns Promise that resolves once all actions have run and no more are scheduled.
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,859 @@
...
+    public resetRateLimitCounter(type: MembershipActionType): void {
+        this.state.rateLimitRetries.set(type, 0);
+        this.state.retries.set(type, 0);
+    }
+}
+
+/**
+ * This class takes care of the membership management.
+ * It has the following tasks:
+ *  - Send the users leave delayed event before sending the memberhsip
```
hughns, line 282:
```suggestion
 *  - Send the users leave delayed event before sending the membership
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,859 @@
...
+                            if (this.rateLimitErrorHandler(e, "sendDelayedStateEvent", type)) return;
+                            if (this.maxDelayExceededErrorHandler(e)) {
+                                this.scheduler.addAction({
+                                    ts: Date.now(),
+                                    type: MembershipActionType.SendFirstDelayedEvent,
+                                });
+                                return;
+                            }
+                            if (this.unsupportedDelayedEndpoint(e)) {
+                                logger.info("Not using deleayed event because the endpoint is not supported");
```
hughns, line 485:
```suggestion
                                logger.info("Not using delayed event because the endpoint is not supported");
```

src/matrixrtc/NewMembershipManager.ts
toger5: Oh my response got lost here.
It was some talking about how I dont like binding to this and how using a callback constructor param is a bit of a footgun in js so I try to not use the pattern.
```
scheduler = new ActionScheduler(
        {
            hasMemberStateEvent: false,
            running: false,
            nextRelativeExpiry: this.membershipEventExpiryTimeout,
            delayId: undefined,
            rateLimitRetries: 0,
            retries: 0,
        },
       // (state, type) => this.membershipLoopHandler(state, type),
       // this is the footgun its easy to assume this should work but then the wrong this is bound...
        this.membershipLoopHandler,
    );
```

But then I wanted to try it anyways since the public `membershipLoopHandler` is also bad.

I convinced myself that this is fine, since we will not use this Constructor all over the place (e.g. this is the only place i expect it to be used) Hence noone hopefully runs into the js footgun.

Updated it to:
```
    private scheduler = new ActionScheduler(ActionScheduler.defaultState, this.membershipLoopHandler.bind(this));

```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,907 @@
...
+                error,
+            );
+        } else if (
+            error instanceof HTTPError &&
+            typeof error.httpStatus === "number" &&
+            error.httpStatus >= 500 &&
+            error.httpStatus < 600
+        ) {
+            logger.warn(
+                "Network error while sending event, retrying in " + retryDurationString + " " + retryCounterString,
```
hughns, line 879:
```suggestion
                "Server error while sending event, retrying in " + retryDurationString + " " + retryCounterString,
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,907 @@
...
+                                state.delayId = undefined;
+                                this.scheduler.addAction({ ts: Date.now(), type: MembershipActionType.SendLeaveEvent });
+                                return;
+                            }
+                            if (this.rateLimitErrorHandler(e, "updateDelayedEvent", type)) return;
+                            if (this.unsupportedDelayedEndpoint(e)) return;
+                            if (this.retryOnNetworkError(e, type)) return;
+
+                            // On any other error we fall back to SendLeaveEvent (this includes hard errors from rate limiting)
+                            this.scheduler.addAction({ ts: Date.now(), type: MembershipActionType.SendLeaveEvent });
```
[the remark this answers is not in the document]
hughns: I'm not sure if `info` or `warn` is appropriate 🤷

hughns (commented):
As per DM this is looking much saner now the network error logic is in place. I will have time to give it a proper review tomorrow.

dbkr (commented):
Phew, this is a big one. Also, nitpick: newlines between functions / tests / any individual thing that spans more than one line (like a variable declaration with a separate line comment) are quite nice.

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,896 @@
...
+
+/**
+ * This state machine tracks the state of the current membership participation
+ * and runs one central timer that wakes up a handler callback with the correct action
+ * whenever necessary.
+ *
+ * It can also be awakened whenever a new action is added which is
+ * earlier then the current "next awake".
+ * @internal
+ */
```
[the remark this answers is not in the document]
toger5: > What advantage does this have over just setting a separate timer for each thing?

With seperate timers we can always run the rist to have parallel loops. This is an easy way to track what else is schduled.

> I know this is private but it really feels like it needs its own set of tests. It's probably big enough to warrant its own file anyway.

It started as part of the class itself. I split it out to make clear they are seperate bits of logic to understand but its not thought of a generic handler.

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,896 @@
...
+    }
+
+    public addAction(action: Action): void {
+        this.insertions.push(action);
+        const nextTs = this.actions[0]?.ts;
+        if (!nextTs || nextTs > action.ts) {
+            this.didWakeUp = true;
+            this.wakeup?.();
+        }
+    }
```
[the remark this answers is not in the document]
toger5: It should be a noop if the Scheduler is not running. `startWithActions` should have resolved hende `addActions` cannot be called anymore.

Do you think a throw or a log would be appropriate?

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,896 @@
...
+    public resetActions(actions: Action[]): void {
+        this.resetWith = actions;
+        const nextTs = this.actions[0]?.ts;
+        const newestTs = actions.map((a) => a.ts).sort((a, b) => a - b)[0];
+        if (nextTs && newestTs && nextTs > newestTs) {
+            this.didWakeUp = true;
+            this.wakeup?.();
+        }
+    }
```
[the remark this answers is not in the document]
toger5: maybe calling both actions is confusing. But I think the logic checks out.

`this.actions[0]?.ts` is the ts at which we already have a timeout scheduled to continue the loop.
If the smallest ts in the new actions array (`this.resetWith`) is larger then the next ts the loop still continues at the time we have to run the next action.
But if the newest item `newestTs` (smallest ts) from `this.resetWith` is smaller (`newestTs < nextTs`) waiting for the next timeout would be too late.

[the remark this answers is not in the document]
toger5: killing and fully restarting is an intersting idea. I will explore that.

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,896 @@
...
+    private get maximumRateLimitRetryCount(): number {
+        return this.joinConfig?.maximumRateLimitRetryCount ?? 10;
+    }
+    private get maximumNetworkErrorRetryCount(): number {
+        return this.joinConfig?.maximumNetworkErrorRetryCount ?? 10;
+    }
+
+    // Scheduler:
+    private scheduler = new ActionScheduler(ActionScheduler.defaultState, this.membershipLoopHandler.bind(this));
```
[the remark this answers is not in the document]
toger5: I opted for bind since its shorter. But if the style is to not bind than we go with the arrow function

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,896 @@
...
+
+    // Loop Handler:
+    private async membershipLoopHandler(state: ActionSchedulerState, type: MembershipActionType): Promise<void> {
+        switch (type) {
+            case MembershipActionType.SendFirstDelayedEvent: {
+                // Before we start we check if we come from a state where we have a delay id.
+                if (!state.delayId) {
+                    // Normal case without any previous delayed id.
+                    await this.client
```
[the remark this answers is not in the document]
toger5: I refactored it to use then+catch because it made the main action much clearer.
`client._unstable_snedDelayedStateEvent` is more prominent now compared to inside a the first line in the try block.

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,896 @@
...
+                    retryDurationString +
+                    " " +
+                    retryCounterString,
+                error,
+            );
+        } else if (
+            error instanceof HTTPError &&
+            typeof error.httpStatus === "number" &&
+            error.httpStatus >= 500 &&
+            error.httpStatus < 600
```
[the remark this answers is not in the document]
toger5: done with this block
```
else if (error instanceof ConnectionError) {
            logger.warn(
                "Network connection error while sending event, retrying in " +
                    retryDurationString +
                    " " +
                    retryCounterString,
                error,
            );
        }
```
(this is what you mean right?)

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,896 @@
...
+            }
+            this.actions = this.actions.filter((a) => a !== nextAction);
+
+            this.actions.push(...this.insertions);
+            this.insertions = [];
+        }
+    }
+
+    public addAction(action: Action): void {
+        this.insertions.push(action);
```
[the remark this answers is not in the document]
toger5: while the `membershipLoopHandler` is running the actions array should be immutable.
I think that is not taken advantage of in the current setup. But any logic that would be based on what is scheduled next would have the guarantee, that the actions are the same during one iteration.
What is helpful however is that reset and add action do not collide and are not dependent in which order they are called.

I am sure it can also work without this but it would make it harder to predict the outcome if the order makes a difference as well.
And I think its not a lot of added code.
Wdyt?

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,918 @@
...
+     * @returns
+     * Returns true if we handled the error by rescheduling the correct next action.
+     * Returns false if it is not a network error.
+     */
+    private retryOnNetworkError(error: unknown, type: MembershipActionType): boolean {
+        // "Is a network error"-boundary
+        const retries = this.scheduler.state.networkErrorRetries.get(type) ?? 0;
+        const retryDurationString = this.callMemberEventRetryDelayMinimum / 1000 + "s";
+        const retryCounterString = "(" + retries + "/" + this.maximumNetworkErrorRetryCount + ")";
+        if (error instanceof Error && error.name === "AbortError") {
```
hughns, line 867: As per DM, I don't think this will detect network errors when EC is running as embedded widget.

I *think* we would see `MatrixError`s instead which we would need to parse.

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,918 @@
...
+     * @param error the error causing this handler check/execution
+     * @param method the method used for the throw message
+     * @param type which MembershipActionType we reschedule because of a rate limit.
+     * @throws If it is a rate limit error and the retry count got exceeded
+     * @returns Returns true if we handled the error by rescheduling the correct next action.
+     * Returns false if it is not a network error.
+     */
+    private rateLimitErrorHandler(error: unknown, method: string, type: MembershipActionType): boolean {
+        // "Is rate limit"-boundary
+        if (!(error instanceof HTTPError && error.isRateLimitError())) {
```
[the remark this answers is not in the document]
toger5: Adressed in: https://github.com/matrix-org/matrix-js-sdk/pull/4726/commits/79abd60ef8f5eb8628bcfb1adaee38e358cac3a6

src/embedded.ts
```diff
@@ -673,3 +689,12 @@ function processAndThrow(error: unknown): never {
         throw error;
     }
 }
+
+function timeoutToConnectionError(error: unknown): never {
```
hughns, line 693:
```suggestion
/**
 * This converts an "Request timed out" error from the PostmessageTransport into a ConnectionError.
 * It either throws the original error or a new ConnectionError.
 **/
function timeoutToConnectionError(error: unknown): never {
```

src/matrixrtc/NewMembershipManager.ts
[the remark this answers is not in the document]
toger5: added a warning

toger5:
I think this is ready in its first iteration.
@dbkr It does not yet address your readability/understand-ability concerns.

I created this actionable issue to work through the good comments which I have not addressed yet:
https://github.com/matrix-org/matrix-js-sdk/issues/4743

The reason those get not addressed here is, that those are better approached not one by one but with a general top to bottom discussion how to make this as easy to understand as possible. (There is a chance we can drop the `SendMainDelayedEvent` action which would greatly simplify it and allow for more radical changes that can incorporate the unaddressed feedback)

.github/workflows/sonarcloud.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         steps:
             # We create the status here and then update it to success/failure in the `report` stage
             # This provides an easy link to this workflow_run from the PR before Sonarcloud is done.
-            - uses: guibranco/github-status-action-v2@5ef6e175c333bc629f3718b083c8a2ff6e0bbfbc
+            - uses: guibranco/github-status-action-v2@7ca807c2ba3401be532d29a876b93262108099fb
```
dbkr, line 30: Possibly a merge has gone wrong here and I think you might be unintentionally reverting something?

dbkr (commented):
Thanks for all the fixes, looking much better. I think you may have had a merge gone wrong though as it looks like there's some reverts crept in here.

toger5:
> Thanks for all the fixes, looking much better. I think you may have had a merge gone wrong though as it looks like there's some reverts crept in here.
This is probably: 0cdb07544d9926cf0855a76ca5cc7dab253bdb24 which we needed to do to make CI happy. currently EC rc-1 depends on this branch. So we will merge this with develop again before we merge this.

toger5:
I created the branch in this PR https://github.com/matrix-org/matrix-js-sdk/pull/4744 to keep the the commit 0cdb07544d9926cf0855a76ca5cc7dab253bdb24 alive which we need for the EC v0.8.0-rc.1. For the actual release we will merge this PR first

src/errors.ts
```diff
@@ -52,3 +52,13 @@ export class ClientStoppedError extends Error {
         super("MatrixClient has been stopped");
     }
 }
+
+export class UnsupportedEndpointError extends Error {
```
richvdh, line 56: Flyby comment: please give this class (and any other exported symbols) some [documentation]. What does it represent? How should downstream applications use it? What do the properties within it mean?

See https://github.com/element-hq/element-web/blob/develop/code_style.md#comments

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -603,4 +616,87 @@ describe.each([
             });
         });
     });
+    describe("unrecoverable errors", () => {
+        // !FailsForLegacy because legacy does not have a retry limit and no mechanism to communicate unrecoverable errors.
+        it("throws, when reaching maximum number of retires for initial delayed event creation !FailsForLegacy", async () => {
```
andybalaam, line 621:
```suggestion
        it("throws, when reaching maximum number of retries for initial delayed event creation !FailsForLegacy", async () => {
```

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -603,4 +616,87 @@ describe.each([
...
+            const manager = new TestMembershipManager({}, room, client, () => undefined);
+            manager.join([focus], focusActive, delayEventSendError);
+
+            for (let i = 0; i < 10; i++) {
+                await jest.advanceTimersByTimeAsync(2000);
+            }
+            expect(delayEventSendError).toHaveBeenCalled();
+        });
+        // !FailsForLegacy because legacy does not have a retry limit and no mechanism to communicate unrecoverable errors.
+        it("throws, when reaching maximum number of retires !FailsForLegacy", async () => {
```
andybalaam, line 641:
```suggestion
        it("throws, when reaching maximum number of retries !FailsForLegacy", async () => {
```

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -603,4 +616,87 @@ describe.each([
...
+        it("falls back to using pure state events when UnsupportedEndpointError encountered for delayed events !FailsForLegacy", async () => {
+            const unrecoverableError = jest.fn();
+            (client._unstable_sendDelayedStateEvent as Mock<any>).mockRejectedValue(
+                new UnsupportedEndpointError("not supported", "sendDelayedStateEvent"),
+            );
+            const manager = new TestMembershipManager({}, room, client, () => undefined);
+            manager.join([focus], focusActive, unrecoverableError);
+            await jest.advanceTimersByTimeAsync(1);
+
+            expect(unrecoverableError).not.toHaveBeenCalledWith();
```
andybalaam, line 698: Should this be `not.toHaveBeenCalled()` ?

andybalaam (approved):
This is way too huge for me to have an opinion on generally, but nothing concerning from a crypto point of view.

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -172,7 +202,7 @@ export class MembershipManager implements IMembershipManager {
         this.focusActive = focusActive;
         this.leavePromiseDefer = undefined;
 
-        this.scheduler.resetState();
+        this.state = MembershipManager.defaultState;
 
         this.scheduler
```
hughns, line 207: I suspect the `.then()` needs to go before `.catch()`.

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -276,9 +306,23 @@ export class MembershipManager implements IMembershipManager {
         if (deviceId === null) throw Error("Missing deviceId in client");
         this.deviceId = deviceId;
         this.stateKey = this.makeMembershipStateKey(userId, deviceId);
+        this.state = MembershipManager.defaultState;
     }
 
-    // Membership Event parameters:
+    // MembershipManager mutable state.
+    public state: ActionSchedulerState;
```
hughns, line 313: Does this need to be public?

```suggestion
    private state: ActionSchedulerState;
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -276,9 +306,23 @@ export class MembershipManager implements IMembershipManager {
...
+    public static get defaultState(): ActionSchedulerState {
```
hughns, line 314: Ditto.

```suggestion
    private static get defaultState(): ActionSchedulerState {
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,945 @@
...
+    }
+    private get maximumNetworkErrorRetryCount(): number {
+        return this.joinConfig?.maximumNetworkErrorRetryCount ?? 10;
+    }
+
+    // Scheduler:
+    private oldStatus?: Status;
+    private scheduler = new ActionScheduler((type): Promise<ActionUpdate> => {
+        if (this.oldStatus) {
+            // we put this at the beginngin of the actsion scheduler loop handle callback since it is a loop this
```
hughns, line 371:
```suggestion
            // we put this at the beginning of the actions scheduler loop handle callback since it is a loop this
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,945 @@
...
+        this.oldStatus = this.status;
+        switch (type) {
+            case MembershipActionType.SendFirstDelayedEvent: {
+                // Before we start we check if we come from a state where we have a delay id.
+                if (!this.state.delayId) {
+                    return this.sendFirstDelayedLeaveEvent(type); // Normal case without any previous delayed id.
+                } else {
+                    // This can happen if someone else (or another client) removes our own membership event.
+                    // It will trigger `onRTCSessionMemberUpdate` queue `MembershipActionType.SendFirstDelayedEvent`.
+                    // We might still have our delyed event from the previous participation and dependent on the serve this might not
```
hughns, line 391:
```suggestion
                    // We might still have our delayed event from the previous participation and dependent on the server this might not
```

src/matrixrtc/NewMembershipManager.ts
toger5: ah es lint is what is making this a hard error?

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,945 @@
...
+                    // get automatically removed if the state changes. Hence It would remove our membership unexpectedly shortly after the rejoin.
+                    //
+                    // In this block we will try to cancel this delayed event before setting up a new one.
+
+                    return this.cancelKnownDelayIdBeforeSendFirstDelayedEvent(type, this.state.delayId);
```
hughns, line 396: The are a few places where the `type` argument is passed, but it appears to be redundant as we only ever call it with a single value `MembershipActionType.SendFirstDelayedEvent`.

Can the type argument be removed?

```suggestion
                    return this.cancelKnownDelayIdBeforeSendFirstDelayedEvent(this.state.delayId);
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,945 @@
...
+                    return createReplaceActionUpdate(MembershipActionType.SendFirstDelayedEvent);
+                }
+                if (this.isUnsupportedDelayedEndpoint(e)) {
+                    return createReplaceActionUpdate(MembershipActionType.SendJoinEvent);
+                }
+
+                // This becomes an unhandle-able error case since sth is signifciantly off if we dont hit any of the above cases
+                // when state.delayId !== undefined
+                // We do not use ignore and log this error since we would also need to reset the delayId.
+                // It is cleaner if we the frontend rejoines instead of resetting the delayId here and behaving like in the success case.
```
hughns, line 521:
```suggestion
                // It is cleaner if we, the frontend, rejoins instead of resetting the delayId here and behaving like in the success case.
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,945 @@
...
+                    // If we get a M_NOT_FOUND we know that the delayed event got already removed.
+                    // This means we are good and can set it to undefined and run this again.
+                    this.state.delayId = undefined;
+                    return createReplaceActionUpdate(MembershipActionType.SendFirstDelayedEvent);
+                }
+                if (this.isUnsupportedDelayedEndpoint(e)) {
+                    return createReplaceActionUpdate(MembershipActionType.SendJoinEvent);
+                }
+
+                // This becomes an unhandle-able error case since sth is signifciantly off if we dont hit any of the above cases
```
hughns, line 518:
```suggestion
                // This becomes an unrecoverable error case since something is significantly off if we don't hit any of the above cases
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,945 @@
...
+                // when state.delayId !== undefined
+                // We do not use ignore and log this error since we would also need to reset the delayId.
```
hughns, line 520:
```suggestion
                // We do not just ignore and log this error since we would also need to reset the delayId.
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -0,0 +1,945 @@
...
+                this.room.roomId,
+                {
+                    delay: this.membershipServerSideExpiryTimeout,
+                },
+                EventType.GroupCallMemberPrefix,
+                {}, // leave event
+                this.stateKey,
+            )
+            .then((response) => {
+                // Success we reset retries and set delayId.
```
hughns, line 456:
```suggestion
                // On success we reset retries and set delayId.
```

hughns (commented):
I like the direction this is heading 👍

src/matrixrtc/NewMembershipManager.ts
toger5: I was considering using this to simplify the useECConnection state hook in EC

hughns (commented):
Latest changes look good

dbkr (approved):
Thanks for the changes: I think splitting up the handler function has made a big difference in particular. I think I'm happy to give this a go.

src/matrixrtc/NewMembershipManager.ts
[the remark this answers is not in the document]
toger5: Solved by using a return object for the new actions.

src/matrixrtc/NewMembershipManager.ts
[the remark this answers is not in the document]
toger5: Addressed the undstandability by numerous changes.

.github/workflows/sonarcloud.yml
toger5: this has been fixed on the main branch of the PR and is only present for the v0.8.0-rc.1 branch.

hughns (approved):
Awesome. It's great to get this over the line. 👍

hughns:
@toger5 one more thing that could go in this PR, but could be a separate one, is adding some comments to say what is a workaround due to https://github.com/element-hq/synapse/pull/17810 having not landed.

This would make it easier to come back and remove the workaround later.

toger5:
> @toger5 one more thing that could go in this PR, but could be a separate one, is adding some comments to say what is a workaround due to https://github.com/element-hq/synapse/pull/17810 having not landed.

I will do this as a seperate PR.

PR #4727, opened 2025-02-21
Delete the dehydrated device when resetEncryption is called

uhoreg:
When we reset encryption (and reset cross-signing keys), delete the dehydrated device since it is (if present) signed with the old cross-signing key.

Fixes https://github.com/element-hq/element-web/issues/29131

## Checklist


richvdh (approved):
Well that was easier than I feared!

Looks like it needs more test coverage.

Also, it would be good to add/update a playwright test to check that this works at the app level.

PR #4730, opened 2025-02-27
Fix idempotency issue around token refresh

t3chguy:
Fixes https://github.com/element-hq/element-web/issues/27810

### DIFF
--- a/babel.config.cjs
+++ b/babel.config.cjs
@@ -26,6 +26,7 @@ module.exports = {
         ],
     ],
     plugins: [
+        ["@babel/plugin-proposal-decorators", { version: "2023-11" }],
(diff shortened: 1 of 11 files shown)


dbkr (approved):
Decorator seems little magic perhaps vs the more obvious wrapper function, but happy enough to go with it.

t3chguy:
> Decorator seems little magic perhaps vs the more obvious wrapper function, but happy enough to go with it.

they're supported out of the box in TS and on their way into ECMA so time to start seeing them.

We could/should ditch babel in the js-sdk given we rely on apps to bundle the matrix-js-sdk anyway and then we won't need the babel decorator proposal stuff

PR #4747, opened 2025-03-12
MatrixRTC MembershipManger: remove redundant sendDelayedEventAction and expose status

toger5:
We do already have the state `hasMemberEvent` that allows to distinguish the two cases. No need to create two dedicated actions.

This also moves code from `NewMemberhsipManager.ts` -> `types.ts`.

And we expose status through an event emitter for the membership manager.

## Checklist

### DIFF
--- a/src/matrixrtc/NewMembershipManager.ts
+++ b/src/matrixrtc/NewMembershipManager.ts
@@ -24,7 +24,12 @@ import { type Room } from "../models/room.ts";
 import { defer, type IDeferred } from "../utils.ts";
 import { type CallMembership, DEFAULT_EXPIRE_DURATION, type SessionMembershipData } from "./CallMembership.ts";
 import { type Focus } from "./focus.ts";
-import { IMembershipManager, MembershipManagerEvent, MembershipManagerEventHandlerMap, Status } from "./types.ts";
+import {
+    type IMembershipManager,
+    type MembershipManagerEventHandlerMap,
+    MembershipManagerEvent,
+    Status,
+} from "./types.ts";
 import { isLivekitFocusActive } from "./LivekitFocus.ts";
 import { type MembershipConfig } from "./MatrixRTCSession.ts";
 import { ActionScheduler, type ActionUpdate } from "./NewMembershipManagerActionScheduler.ts";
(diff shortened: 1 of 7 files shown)


src/matrixrtc/NewMembershipManager.ts
```diff
@@ -465,27 +436,44 @@ export class MembershipManager implements IMembershipManager {
...
                 this.state.delayId = response.delay_id;
-                return createInsertActionUpdate(MembershipActionType.SendJoinEvent);
+                if (this.state.hasMemberStateEvent) {
+                    // Delayed event got send because it got lost due to state event auto cancel
+                    return createInsertActionUpdate(
+                        MembershipActionType.RestartDelayedEvent,
+                        this.membershipKeepAlivePeriod,
+                    );
+                } else {
+                    // Delayed event got send because we just joined
```
hughns, line 448:
```suggestion
                    // We were scheduled because we are in the process of joining
```

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -465,27 +436,44 @@ export class MembershipManager implements IMembershipManager {
             )
             .then((response) => {
                 // On success we reset retries and set delayId.
-                this.state.rateLimitRetries.set(MembershipActionType.SendFirstDelayedEvent, 0);
-                this.state.networkErrorRetries.set(MembershipActionType.SendFirstDelayedEvent, 0);
+                this.resetRateLimitCounter(MembershipActionType.SendDelayedEvent);
                 this.state.delayId = response.delay_id;
-                return createInsertActionUpdate(MembershipActionType.SendJoinEvent);
+                if (this.state.hasMemberStateEvent) {
+                    // Delayed event got send because it got lost due to state event auto cancel
```
hughns, line 442:
```suggestion
                    // We were scheduled because the previous delayed event was cancelled
                    // due to lack of https://github.com/element-hq/synapse/pull/17810
```

hughns (commented):
I've had an initial pass over this, but I think I will need to look again with fresh eyes.

Whilst the PR could be split out into a refactor and the logic change I found it easy enough to split out whilst reviewing.

src/matrixrtc/types.ts
```diff
@@ -34,3 +37,83 @@ export interface ICallNotifyContent {
...
+     */
+    onRTCSessionMemberUpdate(memberships: CallMembership[]): Promise<void>;
+    /**
+     * The used active focus in the currently joined session.
+     * @returns the used active focus in the currently joined session or undefined if not joined.
+     */
+    getActiveFocus(): Focus | undefined;
+
+    // TypedEventEmitter methods:
+    on(event: MembershipManagerEvent.StatusChanged, listener: (oldStatus: string, newStatus: string) => void): this;
```
BillCarsonFr, line 117: why the statuses are just strings? isn't that supposed to be enum `Status`?

BillCarsonFr (approved):
Just read the code to get a better understanding of membership manager. I have question

src/matrixrtc/types.ts
toger5: Yes 100% to also be compatible/consistent with: `    [MembershipManagerEvent.StatusChanged]: (prefStatus: Status, newStatus: Status) => void;
`

hughns (approved):
With fresh eyes and the updated comments this looks good 👍

PR #4752, opened 2025-03-13
MatrixRTC: Fix running not representing what we need from `isJoined` in EC

toger5:
EC expects isJoined to represent if we should be in joined state or not. It does not correlate to what our actual state of the scheduler is. We used the scheduler running state before but on leave the running state will stay true until we successfully updated the room state.
EC expects isJoined to immediately be false.

This introduces a member variable `activated` that represents if the MemberhsipManager is trying to connect or trying to disconnect independent on the current state.

## Checklist

### DIFF
--- a/src/matrixrtc/NewMembershipManagerActionScheduler.ts
+++ b/src/matrixrtc/NewMembershipManagerActionScheduler.ts
@@ -115,12 +115,10 @@ export class ActionScheduler {
                     this._actions.push(...actionUpdate.insert);
                 }
             }
-            this.running = false;
-        } catch (e) {
-            // Set the rtc session "not running" state since we cannot recover from here and the consumer user of the
+        } finally {
+            // Set the rtc session running state since we cannot recover from here and the consumer user of the
             // MatrixRTCSession class needs to manually rejoin.
             this.running = false;
-            throw e;
         }
 
         logger.debug("Leave MembershipManager ActionScheduler loop (no more actions)");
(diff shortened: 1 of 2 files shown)


src/matrixrtc/NewMembershipManagerActionScheduler.ts
```diff
@@ -111,13 +115,13 @@ export class ActionScheduler {
                     this._actions.push(...actionUpdate.insert);
                 }
             }
+            this.running = false;
```
BillCarsonFr, line 118: Would a finally do the job?
```
        } finally {
            this.running = false;
        }
 ```

File: src/matrixrtc/NewMembershipManagerActionScheduler.ts [hunk 2]
```diff
@@ -111,13 +115,13 @@ export class ActionScheduler {
...
         } catch (e) {
             // Set the rtc session "not running" state since we cannot recover from here and the consumer user of the
             // MatrixRTCSession class needs to manually rejoin.
             this.running = false;
             throw e;
         }
```
robintown, line 124:
```suggestion
        } finally {
            this.running = false;
        }
```

Unless you want to keep the comment?

robintown (approved):

src/matrixrtc/NewMembershipManagerActionScheduler.ts
toger5: I thinks then I will not set it to false in the throw case?

src/matrixrtc/NewMembershipManagerActionScheduler.ts
toger5: Yes!

src/matrixrtc/NewMembershipManagerActionScheduler.ts
toger5: Yes you are totally right this is working and simpler.

PR #4753, opened 2025-03-17
Add reportRoom API

Half-Shot:
This arrived in Matrix 1.13 https://spec.matrix.org/latest/client-server-api/#post_matrixclientv3roomsroomidreport

## Checklist

### DIFF
--- a/spec/unit/matrix-client.spec.ts
+++ b/spec/unit/matrix-client.spec.ts
@@ -607,6 +607,26 @@ describe("MatrixClient", function () {
         });
     });
 
+    describe("reportRoom", function () {
+        const roomId = "!foo:bar";
+        const reason = "rubbish room";
+
+        it("should send an invite HTTP POST", async function () {
+            httpLookups = [
+                {
+                    method: "POST",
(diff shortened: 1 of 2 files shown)


t3chguy:
The SDK claims it doesn't support 1.13 yet: https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/version-support.ts#L26

Half-Shot:
> The SDK claims it doesn't support 1.13 yet: https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/version-support.ts#L26

There are certainly elements of 1.10+ supported in this SDK, so what's the procedure for it? Should I put a docstring in warning that it's only partially supported?

t3chguy:
Definitely worth mentioning the Matrix version requirement in the typedoc IMO, as the error from the server in such a case won't be helpful, likely a 404 M_UNRECOGNIZED

dbkr (approved):
I would say this is fine: saying that we support an certain API version is saying we support all of it, we can still support bits of newer versions. We generally have not done any kind of version checking on news APIs like this since if it's not there, it will just fail anyway.

PR #4761, opened 2025-03-18
Rust crypto: set a timeout on outgoing HTTP requests

richvdh:
In an attempt to guard against the request loop getting stuck, put a timeout on outgoing HTTP requests.

Fixes: https://github.com/element-hq/element-web/issues/29534


andybalaam (approved):
Looks good, thanks!

PR #4766, opened 2025-03-27
Add new account data for MSC4155

Half-Shot:
https://github.com/matrix-org/matrix-spec-proposals/pull/4155

## Checklist


t3chguy (changes requested):
This feels like it has no point being in the js-sdk for an MSC PoC. Why not have it just live in the application you wish to support it, presumably EW? AccountDataEvents is extensible by the host app

PR #4767, opened 2025-03-28
Stabilise MSC3765

Johennes:
https://github.com/matrix-org/matrix-spec-proposals/pull/3765 was merged which means we can switch to stable identifiers and the final state of the proposal.

## Checklist

### DIFF
--- a/src/@types/topic.ts
+++ b/src/@types/topic.ts
@@ -42,20 +42,17 @@ import { type IMessageRendering } from "./extensible_events.ts";
 /**
  * The event type for an m.topic event (in content)
  */
-export const M_TOPIC = new NamespacedValue("m.topic", "org.matrix.msc3765.topic");
+export const M_TOPIC = new NamespacedValue("m.topic");
 
 /**
  * The event content for an m.topic event (in content)
  */
 export type MTopicContent = IMessageRendering[];
 
-type MTopicStable = { [K in Exclude<typeof M_TOPIC.altName, null | undefined>]: MTopicContent };
-type MTopicUnstable = { [M_TOPIC.name]: MTopicContent };
-
 /**
  * The event definition for an m.topic event (in content)
  */
-export type MTopicEvent = (MTopicStable & MTopicUnstable) | MTopicStable | MTopicUnstable;
+export type MTopicEvent = { "m.topic": MTopicContent };
 
 /**
  * The event content for an m.room.topic event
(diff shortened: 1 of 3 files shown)


src/content-helpers.ts
```diff
@@ -189,12 +189,14 @@ export type MakeTopicContent = (topic: string | null | undefined, htmlTopic?: st
 
 export const makeTopicContent: MakeTopicContent = (topic, htmlTopic) => {
     const renderings = [];
-    if (isProvided(topic)) {
-        renderings.push({ body: topic, mimetype: "text/plain" });
-    }
+    // Put HTML first because clients will render the first type in
+    // the array that they understand
```
Johennes, line 193: This also needs to be reflected in `TopicState` so that users of the SDK actually know which rendering came first. Maybe we could just add a `preferred` flag into the type?

```
export type TopicState = {
    text?: string;
    html?: string;
    preferred?: "text" | "html"
};
```

This won't scale well but I don't realistically foresee further mimetypes being added here in the near future.

src/@types/topic.ts
```diff
@@ -42,14 +42,14 @@ import { type IMessageRendering } from "./extensible_events.ts";
...
-export const M_TOPIC = new UnstableValue("m.topic", "org.matrix.msc3765.topic");
+export const M_TOPIC = new NamespacedValue("m.topic", "org.matrix.msc3765.topic");
 
 /**
  * The event content for an m.topic event (in content)
  */
 export type MTopicContent = IMessageRendering[];
 
-type MTopicStable = { [M_TOPIC.altName]: MTopicContent };
+type MTopicStable = { [K in Exclude<typeof M_TOPIC.altName, null | undefined>]: MTopicContent };
```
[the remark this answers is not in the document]
Johennes: Looks like this doesn't play well with typedoc even when I add a doc comment to `altName`. 😕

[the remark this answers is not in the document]
Johennes: I couldn't find a good way to make this work so I decided to drop the prior unstable event format entirely.

src/@types/topic.ts
```diff
@@ -42,20 +42,17 @@ import { type IMessageRendering } from "./extensible_events.ts";
 /**
  * The event type for an m.topic event (in content)
  */
-export const M_TOPIC = new UnstableValue("m.topic", "org.matrix.msc3765.topic");
+export const M_TOPIC = new NamespacedValue("m.topic");
```
t3chguy, line 45: Shouldn't this still reference the unstable value so consumers can continue to render existing topics as they did before rather than break all existing rooms?

Johennes: I had some [trouble](https://github.com/matrix-org/matrix-js-sdk/pull/4767#discussion_r2018054292) getting it to work correctly. Maybe you have an idea how to do this properly?

t3chguy (approved):

PR #4773, opened 2025-04-01
MatrixRTC | Refactoring introduce a transport abstraction as prep work for to device transport

BillCarsonFr:
Preparation work for https://github.com/element-hq/element-call/issues/3155 part of EPIC [E2EE: Send media keys via to_device message](https://github.com/element-hq/element-call/issues/1281)

## Checklist

- [ ] Tests written for new code (and old code if feasible).
- [ ] New or updated `public`/`exported` symbols have accurate [TSDoc](https://tsdoc.org/) documentation.
- [ ] Linter and other CI checks pass.
- [ ] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md)).

PR #4785, opened 2025-04-02
MatrixRTC: ToDevice distribution for media stream keys

BillCarsonFr:
Require changes on StopGapWidget to properly pass to device events without stripping things
## Checklist

### DIFF
--- a/spec/unit/matrixrtc/MatrixRTCSession.spec.ts
+++ b/spec/unit/matrixrtc/MatrixRTCSession.spec.ts
@@ -835,7 +835,7 @@ describe("MatrixRTCSession", () => {
             it("rotates key if a member leaves", async () => {
                 jest.useFakeTimers();
                 try {
-                    const KEY_DALAY = 3000;
+                    const KEY_DELAY = 3000;
                     const member2 = Object.assign({}, membershipTemplate, {
                         device_id: "BBBBBBB",
                     });
@@ -856,7 +856,7 @@ describe("MatrixRTCSession", () => {
                         sendEventMock.mockImplementation((_roomId, _evType, payload) => resolve(payload));
                     });
 
-                    sess.joinRoomSession([mockFocus], mockFocus, { manageMediaKeys: true, makeKeyDelay: KEY_DALAY });
+                    sess.joinRoomSession([mockFocus], mockFocus, { manageMediaKeys: true, makeKeyDelay: KEY_DELAY });
                     const sendKeySpy = jest.spyOn((sess as unknown as any).encryptionManager.transport, "sendKey");
                     const firstKeysPayload = await keysSentPromise1;
                     expect(firstKeysPayload.keys).toHaveLength(1);
@@ -874,7 +874,7 @@ describe("MatrixRTCSession", () => {
                         .mockReturnValue(makeMockRoomState([membershipTemplate], mockRoom.roomId));
                     sess.onRTCSessionMemberUpdate();
 
-                    jest.advanceTimersByTime(3000);
+                    jest.advanceTimersByTime(KEY_DELAY);
                     expect(sendKeySpy).toHaveBeenCalledTimes(1);
                     // check that we send the key with index 1 even though the send gets delayed when leaving.
                     // this makes sure we do not use an index that is one too old.
--- a/spec/unit/matrixrtc/MatrixRTCSession.spec.ts
+++ b/spec/unit/matrixrtc/MatrixRTCSession.spec.ts
@@ -486,14 +486,17 @@ describe("MatrixRTCSession", () => {
             let sendStateEventMock: jest.Mock;
             let sendDelayedStateMock: jest.Mock;
             let sendEventMock: jest.Mock;
+            let sendToDeviceMock: jest.Mock;
 
             beforeEach(() => {
                 sendStateEventMock = jest.fn();
                 sendDelayedStateMock = jest.fn();
                 sendEventMock = jest.fn();
+                sendToDeviceMock = jest.fn();
                 client.sendStateEvent = sendStateEventMock;
                 client._unstable_sendDelayedStateEvent = sendDelayedStateMock;
                 client.sendEvent = sendEventMock;
+                client.encryptAndSendToDevice = sendToDeviceMock;
 
                 mockRoom = makeMockRoom([]);
                 sess = MatrixRTCSession.roomSessionForRoom(client, mockRoom);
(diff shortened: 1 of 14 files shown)


toger5 (approved):

src/matrixrtc/ToDeviceKeyTransport.ts
```diff
@@ -0,0 +1,169 @@
...
+    private receiveCallKeyEvent(fromUser: string, content: EncryptionKeysToDeviceEventContent): void {
+        // The event has already been validated at this point.
+
+        this.statistics.counters.roomEventEncryptionKeysReceived += 1;
+
+        // What is this, and why is it needed?
+        // Also to device events do not have an origin server ts
+        const now = Date.now();
+        const age = now - (typeof content.sent_ts === "number" ? content.sent_ts : now);
+        this.statistics.totals.roomEventEncryptionKeysReceivedTotalAge += age;
```
dbkr, line 103: This is all very confusing. Firstly, isn't the field `timestamp`? So why would we pass in an age? Wasn't this introduced recently (https://github.com/matrix-org/matrix-js-sdk/pull/4773)? How do we not know what it is or why its needed?

src/matrixrtc/ToDeviceKeyTransport.ts
```diff
@@ -0,0 +1,169 @@
...
+            this.prefixedLogger.warn("Malformed Event: Missing keys field");
+            return;
+        }
+
+        if (!content.member || !content.member.claimed_device_id) {
+            this.prefixedLogger.warn("Malformed Event: Missing claimed_device_id");
+            return;
+        }
+
+        // TODO session is not used so far
```
dbkr, line 162: Then this block is probably not helpful to be here since if we release this code then we'll have to work with clients that don't send 'session' for some time. We could make it validate it if present but allow it to be absent?

src/matrixrtc/ToDeviceKeyTransport.ts
```diff
@@ -0,0 +1,169 @@
...
+import { TypedEventEmitter } from "../models/typed-event-emitter.ts";
+import { type IKeyTransport, KeyTransportEvents, type KeyTransportEventsHandlerMap } from "./IKeyTransport.ts";
+import { type Logger } from "../logger.ts";
+import type { CallMembership } from "./CallMembership.ts";
+import type { EncryptionKeysToDeviceEventContent, Statistics } from "./types.ts";
+import { ClientEvent, type MatrixClient } from "../client.ts";
+import type { MatrixEvent } from "../models/event.ts";
+import { EventType } from "../@types/event.ts";
+
+export class ToDeviceKeyTransport
```
dbkr, line 26: Docs please

spec/unit/matrixrtc/MatrixRTCSession.spec.ts
```diff
@@ -832,6 +835,7 @@ describe("MatrixRTCSession", () => {
             it("rotates key if a member leaves", async () => {
                 jest.useFakeTimers();
                 try {
+                    const KEY_DALAY = 3000;
```
dbkr, line 838:
```suggestion
                    const KEY_DELAY = 3000;
```

spec/unit/matrixrtc/MatrixRTCSession.spec.ts
```diff
@@ -869,14 +874,24 @@ describe("MatrixRTCSession", () => {
                         .mockReturnValue(makeMockRoomState([membershipTemplate], mockRoom.roomId));
                     sess.onRTCSessionMemberUpdate();
 
-                    jest.advanceTimersByTime(10000);
+                    jest.advanceTimersByTime(3000);
```
dbkr, line 877: Should this also be `KEY_DELAY` or it something else that's coincidentally also 3s?

spec/unit/matrixrtc/ToDeviceKeyTransport.spec.ts
```diff
@@ -0,0 +1,259 @@
...
+
+describe("ToDeviceKeyTransport", () => {
+    const roomId = "!room:id";
+
+    let mockClient: Mocked<MatrixClient>;
+    let statistics: Statistics;
+    let mockLogger: Mocked<Logger>;
+    let transport: ToDeviceKeyTransport;
+
+    function setup() {
```
dbkr, line 36: beforeEach?

src/matrixrtc/EncryptionManager.ts
```diff
@@ -80,8 +76,11 @@ export class EncryptionManager implements IEncryptionManager {
     // if it looks like a membership has been updated.
     private lastMembershipFingerprints: Set<string> | undefined;
 
-    private currentEncryptionKeyIndex = -1;
-
+    // TODO: remove this value since I think this is not helpful to use
+    // it represents what index we actually sent over to ElementCall (which we do in a delayed manner)
+    // but it has no releavnt information for the encryption manager.
+    private mediaTrailerKeyIndexInUse = -1;
+    private latestGeneratedKeyIndex = -1;
```
dbkr, line 83: Not sure why you're adding something and then claiming in a comment that it should be removed...

Also, why the rename?

dbkr (commented):
A quick PR description saying what this is doing might have been helpful, although I think I've worked it out now apart from maybe some of the changes to EncryptionManager which don't seem all that relevant.

src/client.ts
```diff
@@ -7942,6 +7942,29 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
 
+    /**
+     * This will encrypt the payload for all devices in the list and will queue it.
+     * The type of the sent to-device message will be `m.room.encrypted`.
+     * @param eventType - The type of event to send
+     * @param devices - The list of devices to send the event to. Each device is
+     * @param payload - The payload to send. This will be encrypted.
+     * @returns Promise which resolves once queued.
+     */
+    public async encryptAndSendToDevice(
```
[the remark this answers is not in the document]
toger5: I think since it will send the event eventually calling it send it okay?
Also we are considering to allow error handling in the widget so this should not resolve once queued but once send and return the error. This is a tracked issue we will tackle as a second iteration on both rust sdk ec and js-sdk.

src/matrixrtc/EncryptionManager.ts
toger5: Mostly because of: https://github.com/matrix-org/matrix-js-sdk/pull/4785#discussion_r2035046112

But you might be right. we did the testing where we needed the logging so probably its best to just remove it.

src/matrixrtc/ToDeviceKeyTransport.ts
toger5: we definitly want to pass `now` and not `age`.
We use this later to ignore keys if they are older then the one currently in use.
Since content.sent_ts should be undefined so age = 0 in most cases. And we cannot do the ignore logic correctly.

dbkr (approved):

src/client.ts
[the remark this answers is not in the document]
toger5: I created: https://github.com/matrix-org/matrix-js-sdk/pull/4793
So we can tackle this in a better way and get rid of the TODO comment. (not yet ready to review needs to wait until next week i suspect.)

PR #4792, opened 2025-04-09
MatrixRTC: Add combined `toDeviceAndRoomKeyTransport`

toger5:
## Checklist

### DIFF
--- a/src/matrixrtc/EncryptionManager.ts
+++ b/src/matrixrtc/EncryptionManager.ts
@@ -110,6 +110,7 @@ export class EncryptionManager implements IEncryptionManager {
         this.manageMediaKeys = this.joinConfig?.manageMediaKeys ?? this.manageMediaKeys;
 
         this.transport.on(KeyTransportEvents.ReceivedKeys, this.onNewKeyReceived);
+        // Deprecate RoomKeyTransport: this can get removed.
         if (this.transport instanceof RoomAndToDeviceTransport) {
             this.transport.on(RoomAndToDeviceEvents.EnabledTransportsChanged, this.onTransportChanged);
         }
--- a/src/matrixrtc/RoomAndToDeviceKeyTransport.ts
+++ b/src/matrixrtc/RoomAndToDeviceKeyTransport.ts
@@ -21,6 +21,7 @@ import type { RoomKeyTransport } from "./RoomKeyTransport.ts";
 import type { ToDeviceKeyTransport } from "./ToDeviceKeyTransport.ts";
 import { TypedEventEmitter } from "../models/typed-event-emitter.ts";
 
+// Deprecate RoomAndToDeviceTransport: This whole class is only a stop gap until we remove RoomKeyTransport.
 export interface EnabledTransports {
     toDevice: boolean;
     room: boolean;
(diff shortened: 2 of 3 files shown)


spec/unit/matrixrtc/RoomAndToDeviceTransport.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+    });
+
+    it("should enable to device transport when starting", () => {
+        transport.start();
+        expect(transport.enabled.room).toBeFalsy();
+        expect(transport.enabled.toDevice).toBeTruthy();
+    });
+    it("only sends to device keys when sending a key", async () => {
+        transport.start();
+        const toDeviceSpy = jest.spyOn(getToDeviceTransport(transport), "sendKey");
```
BillCarsonFr, line 90: Instead of doing that, wouldn't it be better to jsut ensure which of `encryptAndSendToDevice` and `sendEvent` mock is called?

src/matrixrtc/EncryptionManager.ts
```diff
@@ -118,6 +126,9 @@ export class EncryptionManager implements IEncryptionManager {
         // as they may still be using the same ones.
         this.encryptionKeys.set(getParticipantId(this.userId, this.deviceId), []);
         this.transport.off(KeyTransportEvents.ReceivedKeys, this.onNewKeyReceived);
+        if (this.transport instanceof RoomAndToDeviceTransport) {
```
BillCarsonFr, line 129: I think we could avoid touching encryption manager by having the RoomAndToDeviceKeyTransport to cache the current senderkey in order to replay it to the new child transport.
This will be transparent to Ecryption manager

src/matrixrtc/RoomAndToDeviceKeyTransport.ts
```diff
@@ -0,0 +1,142 @@
...
+export interface EnabledTransports {
+    toDevice: boolean;
+    room: boolean;
+}
+
+export enum RoomAndToDeviceEvents {
+    EnabledTransportsChanged = "enabled_transports_changed",
+}
+export type RoomAndToDeviceEventsHandlerMap = {
+    [RoomAndToDeviceEvents.EnabledTransportsChanged]: (enabledTransports: EnabledTransports) => void;
```
BillCarsonFr, line 36: As said I think this could be avoidded by caching the current key and replay on transport change?

BillCarsonFr (commented):
I believe it could be simplified

src/matrixrtc/RoomAndToDeviceKeyTransport.ts
toger5: The reason for the emission is not only the resending but mostly that we can update the label in ec

toger5: Also this would make us loose the retry mechanisms from the Encryption manager.
SendKey for the toDevice would succeed, we cache the key, we need to switch to a new transport and we send the key over the new transport. if that key sending fails there is nothing we can do. Except we build another retrying mechanism into the transport.

toger5: discussed sync: We keep the event emissino while we use `RoomAndToDeviceTransport` as a stop gap.

spec/unit/matrixrtc/RoomAndToDeviceTransport.spec.ts
toger5: I was wondering the same. It depends how much we want to test here.
I am a fan of smaller "unit" where there are clear interfaces for the classes public api.
When using encryptAndSendToDevice we basically also test the toDeviceTransport which could be tested on its own.
This way the tests are less fragile to changes elsewhere in the codebase (not in the `RoomAndToDeviceTransport`)

BillCarsonFr (approved):
LGTM

PR #4799, opened 2025-04-11
MatrixRTC: Refactor | Introduce a new Encryption manager (used with experimental to device transport)

BillCarsonFr:
Add a new simplified EncryptionManager that will rotate the key for any membership change (join or leave or same)

## Checklist


src/matrixrtc/BasicEncryptionManager.ts
```diff
@@ -0,0 +1,237 @@
...
+    public onMembershipsUpdate(oldMemberships: CallMembership[]): void {
+        this.logger.trace(`onMembershipsUpdate`);
+
+        // This encryption manager is very basic, it will rotate the key for any membership change
+        // Request rotation of the key
+        this.ensureMediaKey();
+    }
+
+    private async rolloutOutboundKey(): Promise<void> {
+        const hasExistingKey = this.outboundSession != null;
```
toger5, line 165:
```suggestion
        const isFirstKey = this.outboundSession === null;
```
I think this makes the logic where its used easier to read

src/matrixrtc/BasicEncryptionManager.ts
```diff
@@ -0,0 +1,237 @@
...
+            if (!hasExistingKey) {
+                this.logger.trace(`Rollout immediately`);
+                // rollout immediately
+                this.onEncryptionKeysChanged(
+                    newOutboundKey.key,
+                    newOutboundKey.keyId,
+                    getParticipantId(this.userId, this.deviceId),
+                );
+            } else {
+                // Delay a bit using this key
```
toger5, line 203:
```suggestion
                // Delay a bit before using this key
```

src/matrixrtc/BasicEncryptionManager.ts
```diff
@@ -0,0 +1,237 @@
...
+                    this.logger.trace(`...Delayed rollout of index:${newOutboundKey.keyId} `);
+                    // Start encrypting with that key now that there was time to distibute it
+                    this.onEncryptionKeysChanged(
+                        newOutboundKey.key,
+                        newOutboundKey.keyId,
+                        getParticipantId(this.userId, this.deviceId),
+                    );
+                    rolledOut.resolve();
+                }, 1000);
+                return rolledOut.promise;
```
toger5, line 216:
```suggestion
                this.logger.trace(`Delay Rollout...`);
                await sleep(1000); // from "matrix-js-sdk/lib/utils"
                this.logger.trace(`...Delayed rollout of index:${newOutboundKey.keyId} `);
                // Start encrypting with that key now that there was time to distibute it
                this.onEncryptionKeysChanged(
                    newOutboundKey.key,
                    newOutboundKey.keyId,
                    getParticipantId(this.userId, this.deviceId),
                );
                // no return needed, async fn will resolve after await.
```

src/matrixrtc/BasicEncryptionManager.ts
```diff
@@ -0,0 +1,237 @@
...
+            }
+        } catch (err) {
+            this.logger.error(`Failed to rollout key`, err);
+        }
+    }
+
+    private nextKeyId(): number {
```
toger5, line 223:
```suggestion
    private nextKeyIndex(): number {
```

src/matrixrtc/BasicEncryptionManager.ts
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    private generateRandomKey(): Uint8Array {
+        const key = new Uint8Array(16);
+        globalThis.crypto.getRandomValues(key);
+        return key;
+    }
+}
+
+const getParticipantId = (userId: string, deviceId: string): ParticipantId => `${userId}:${deviceId}`;
```
toger5, line 237: maybe move this to `utils` now that we have such a file.
there are also logs where we do `${userId}:${deviceId}` which can use the utils method.

src/matrixrtc/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        const entry = this.buffer.get(participantId)!;
+        clearTimeout(entry.timeout);
+        entry.timeout = setTimeout(() => {
+            this.buffer.delete(participantId);
+        }, this.ttl);
+
+        const existing = entry.keys.get(item.keyId);
+        if (existing && existing.creationTS > item.creationTS) {
```
toger5, line 66: `creationTS`, `timestamp`, `send_ts` is all the same thing right?

Is it possible to call them the same (or maybe `sendTs` + `send_ts`)
(edit: oh its: `sent_ts`... Is that what we use in the spec.)

src/matrixrtc/utils.ts
toger5, line 1: I am not sure the time to live is the correct concept here. It would not work if the out of order receive offset time is larger than 1s.
Can we just use a "latest key buffer"? Of course we would end up with more memory use since a 100ppl call would have 100 entries where the ttl approach probably would delete them quick enough so it would never exceed a handful.

The latest key buffer would just store the most recent (by timestamp) key for the highest (considering wrap around) index.

toger5 (changes requested):

src/matrixrtc/utils.ts
BillCarsonFr: TBH I think we could do without it complitely as it is probably very edge casy. We also want to avoid keeping key in memory.
Only keeping the latest (1 key)  might not be enough. If there was 3 keys `0`, `1`, `0'` that are received in `0'` `1` , `0`, the latest would be `1` but we want `0`

src/matrixrtc/utils.ts
toger5: in the spec we should make sent_ts -> creation_ts then right? the spec should focus on the disambiguation instead of the possibility to collect reasonable stats.

src/matrixrtc/utils.ts
toger5: >  Only keeping the latest (1 key) might not be enough. If there was 3 keys 0, 1, 0' that are received in 0' 1 , 0, the latest would be 1 but we want 0

Yes covering all cases requies us to store all keys ever received.
What about we store all timestamps ever received? That is enough right. The disambigution is then just a "is it the most recent key we want to use" check.
Could be called `keyRecencyMap.isMostRecentKeyForParticipant(userId, key.creationTs)`

src/matrixrtc/BasicEncryptionManager.ts
```diff
@@ -0,0 +1,237 @@
...
+        this.transport.start();
+
+        this.ensureMediaKey();
+    }
+
+    /**
+     * Will ensure that a new key is distributed and used to encrypt our media.
+     * If this function is called repeatidly, the calls will be buffered to a single key rotation.
+     */
+    private ensureMediaKey(): void {
```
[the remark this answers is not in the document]
BillCarsonFr: The manager has changed a bit so this call is really ensuring that everyone has the keys it needs. So it will not rotate key anymore.
it will be no op if no action needed. 
I have no strong opinion on the name to use, maybe ensureOutboundKey ?

src/matrixrtc/BasicEncryptionManager.ts
```diff
@@ -0,0 +1,237 @@
...
+        const newKey: InboundEncryptionSession = {
+            key: keyBin,
+            participantId,
+            keyId: index,
+            creationTS: timestamp,
+        };
+
+        const validKey = this.keyBuffer.disambiguate(participantId, newKey);
+        if (validKey) {
+            this.onEncryptionKeysChanged(validKey.key, index, validKey.participantId);
```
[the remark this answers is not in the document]
BillCarsonFr: There are other things that can be cleaned in the constructor. I don't think we also need the `getMembership` (if we could just pass it in the join)

src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -0,0 +1,321 @@
...
+            if (this.outboundSession) {
+                this.outboundSession.sharedWith = [];
+                this.ensureMediaKey();
+            }
+        }
+    };
+
+    /**
+     * Will ensure that a new key is distributed and used to encrypt our media.
+     * If this function is called repeatidly, the calls will be buffered to a single key rotation.
```
toger5:
```suggestion
     * If this function is called repeatedly, the calls will be buffered to a single key rotation.

```

dbkr:
Putting this back into draft until it's ready as agreed in chapter sync.

src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -0,0 +1,321 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { type IEncryptionManager } from "./EncryptionManager.ts";
+import { type EncryptionConfig } from "./MatrixRTCSession.ts";
+import { type CallMembership } from "./CallMembership.ts";
+import { decodeBase64, encodeBase64 } from "../base64.ts";
+import { type IKeyTransport, type KeyTransportEventListener, KeyTransportEvents } from "./IKeyTransport.ts";
+import { logger as rootLogger, type Logger } from "../logger.ts";
```
richvdh: please don't use `logger.logger` in new code: we should be threading the correct logger through from the parent `MatrixClient`.

BillCarsonFr: https://github.com/matrix-org/matrix-js-sdk/pull/4799/commits/927de6db1e05131650349fef6849df15444cab4d

src/matrixrtc/utils.ts
BillCarsonFr: This is tracked as a separate issue https://github.com/element-hq/voip-internal/issues/374

toger5 (approved):
Did a sync (call) review session.

Open todos:
 - solve `reemission` (possible missing keys) situation
    - either by adding a stop gap key buffer to the new encryption manager
    -  or by forcing to allocate + subscribe on `joinRoomSession` (passing the matrixKeyProvider (as a keySink interface) to `joinRoomSession`

PR #4809, opened 2025-04-16
Inherit `methodFactory` extensions from the parent to the child loggers.

toger5:
This fixes an issue where a log extension (a modification to the `methodFactory`) applied to a parent logger would ONLY be executed when logging with the parent but not on any of the children created with `getChild()`.

This PR changes `getChild` to also inherit the `methodFactory` from the parent.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
         "bs58": "^6.0.0",
         "content-type": "^1.0.4",
         "jwt-decode": "^4.0.0",
-        "loglevel": "^1.7.1",
+        "loglevel": "^1.9.2",
         "matrix-events-sdk": "0.0.1",
         "matrix-widget-api": "^1.10.0",
         "oidc-client-ts": "^3.0.1",
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -131,22 +131,20 @@ interface PrefixedLogger extends loglevel.Logger, LoggerWithLogMethod {
 function getPrefixedLogger(prefix?: string): PrefixedLogger {
     const loggerName = DEFAULT_NAMESPACE + (prefix === undefined ? "" : `-${prefix}`);
     const prefixLogger = loglevel.getLogger(loggerName) as PrefixedLogger;
-    // This is a save cast since export const logger is constructed with getPrefixedLogger
-    // It is downcastet to `Logger` to minimize the public/exported api.
+
     if (prefixLogger.getChild === undefined) {
         // This is a new loglevel Logger which has not been turned into a PrefixedLogger yet.
         prefixLogger.prefix = prefix;
         prefixLogger.getChild = (childPrefix): Logger => {
-            const rootLogger = logger as PrefixedLogger;
             // create the new child logger
             const childLogger = getPrefixedLogger((prefix ?? "") + childPrefix);
             // assign the same methodFactory as the root logger.
             // this is useful if we add extensions to the root logger that modify
             // its methodFactory. (an example extension is: storing each log to a rageshake db)
-            childLogger.methodFactory = rootLogger.methodFactory;
+            childLogger.methodFactory = prefixLogger.methodFactory;
             // rebuild the child logger with the new methodFactory.
             childLogger.rebuild();
-            return childLogger as Logger;
+            return childLogger;
         };
         prefixLogger.setLevel(loglevel.levels.DEBUG, false);
     }
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -128,18 +128,25 @@ interface PrefixedLogger extends loglevel.Logger, LoggerWithLogMethod {
  *
  * @param prefix Prefix to add to each logged line. If undefined, no prefix will be added.
  */
-function getPrefixedLogger(prefix?: string): LoggerWithLogMethod {
+function getPrefixedLogger(prefix?: string): PrefixedLogger {
     const loggerName = DEFAULT_NAMESPACE + (prefix === undefined ? "" : `-${prefix}`);
     const prefixLogger = loglevel.getLogger(loggerName) as PrefixedLogger;
-
+    // This is a save cast since export const logger is constructed with getPrefixedLogger
+    // It is downcastet to `Logger` to minimize the public/exported api.
     if (prefixLogger.getChild === undefined) {
         // This is a new loglevel Logger which has not been turned into a PrefixedLogger yet.
         prefixLogger.prefix = prefix;
         prefixLogger.getChild = (childPrefix): Logger => {
-            const childLogger = getPrefixedLogger((prefix ?? "") + childPrefix) as unknown as loglevel.Logger;
-            childLogger.methodFactory = (logger as unknown as loglevel.Logger).methodFactory;
-            childLogger.setLevel(childLogger.getLevel());
-            return childLogger as unknown as Logger;
+            const rootLogger = logger as PrefixedLogger;
+            // create the new child logger
+            const childLogger = getPrefixedLogger((prefix ?? "") + childPrefix);
+            // assign the same methodFactory as the root logger.
+            // this is useful if we add extensions to the root logger that modify
+            // its methodFactory. (an example extension is: storing each log to a rageshake db)
+            childLogger.methodFactory = rootLogger.methodFactory;
+            // rebuild the child logger with the new methodFactory.
+            childLogger.rebuild();
+            return childLogger as Logger;
         };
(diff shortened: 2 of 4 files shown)


dbkr (commented):
Ummm... I think I understand the PR you referenced but I'm afraid I don't understand what's going on here, and the amount of casting through unknown and splicing of methods to other objects is terrifying. If we do this, it needs very heavy commenting!

toger5:
@dbkr I found that `rebuild` is doing the same thing as `childLogger.setLevel(childLogger.getLevel());`
Additionally it was possible to make the types a lot more easy to follow by returning `PrefixedLogger` from `getPrefixedLogger` and then chinging the type once we export logger `export const logger = getPrefixedLogger() as LoggerWithLogMethod;`

src/logger.ts
```diff
@@ -128,14 +128,26 @@ interface PrefixedLogger extends loglevel.Logger, LoggerWithLogMethod {
  *
  * @param prefix Prefix to add to each logged line. If undefined, no prefix will be added.
  */
-function getPrefixedLogger(prefix?: string): LoggerWithLogMethod {
+function getPrefixedLogger(prefix?: string): PrefixedLogger {
     const loggerName = DEFAULT_NAMESPACE + (prefix === undefined ? "" : `-${prefix}`);
     const prefixLogger = loglevel.getLogger(loggerName) as PrefixedLogger;
-
+    // This is a save cast since export const logger is constructed with getPrefixedLogger
+    // It is downcastet to `Logger` to minimize the public/exported api.
```
robintown, line 135: This comment looks like it's out of place - should it go on the `const rootLogger = logger as PrefixedLogger;` line?

src/logger.ts
```diff
@@ -128,14 +128,26 @@ interface PrefixedLogger extends loglevel.Logger, LoggerWithLogMethod {
...
+            const rootLogger = logger as PrefixedLogger;
+            // create the new child logger
+            const childLogger = getPrefixedLogger((prefix ?? "") + childPrefix);
+            // assign the same methodFactory as the root logger.
+            // this is useful if we add extensions to the root logger that modify
+            // its methodFactory. (an example extension is: storing each log to a rageshake db)
+            childLogger.methodFactory = rootLogger.methodFactory;
+            // rebuild the child logger with the new methodFactory.
+            childLogger.rebuild();
+            return childLogger as Logger;
```
robintown, line 149: Is it possible to drop this cast? Just to clarify that this is a safe downcast.

```suggestion
            return childLogger;
```

src/logger.ts
```diff
@@ -128,14 +128,26 @@ interface PrefixedLogger extends loglevel.Logger, LoggerWithLogMethod {
...
         prefixLogger.prefix = prefix;
-        prefixLogger.getChild = (childPrefix): Logger => getPrefixedLogger((prefix ?? "") + childPrefix);
+        prefixLogger.getChild = (childPrefix): Logger => {
+            const rootLogger = logger as PrefixedLogger;
+            // create the new child logger
+            const childLogger = getPrefixedLogger((prefix ?? "") + childPrefix);
+            // assign the same methodFactory as the root logger.
+            // this is useful if we add extensions to the root logger that modify
+            // its methodFactory. (an example extension is: storing each log to a rageshake db)
+            childLogger.methodFactory = rootLogger.methodFactory;
```
robintown, line 146: I'm wondering if this should instead be

```suggestion
            childLogger.methodFactory = prefixLogger.methodFactory;
```

so that if I extend a *child* of the root logger, then any *children of the child* will inherit their parent's extensions, rather than just the root extension

toger5: Yes this makes sense.
It would still give all children the same factory in the EC usecase sine we extend the root logger and all children will rebuild with their parent so it will inherit downwards.

Maybe its also important to mention that getChild and adding extensions is order dependent. Since we will rebuild only at the time where we add a child later changes to the factory will not apply even though the factory is shared between all loggers.
Because in both ways it is different if we first create all the children and then extend the root logger or if we extend the root logger and than create the children.

robintown: Sure, wouldn't hurt to mention

dbkr (approved):
OK - thanks for the comments, I think it's a bit easier to see what's happening now.

src/logger.ts
toger5: added this comment to the `getChild` docs

dbkr (approved):

BillCarsonFr (approved):
ok

robintown (approved):

richvdh:
> This is an experiment where I tried to get the same thing I am doing in: https://github.com/matrix-org/matrix-js-sdk/pull/4807

I assume it is no longer an experiment? Could you update the PR description to reflect the current state, please?

src/logger.ts
```diff
@@ -33,6 +33,10 @@ interface LoggerWithLogMethod extends Logger {
 export interface Logger extends BaseLogger {
     /**
      * Create a child logger.
+     * This child will apply the methodFactory of the parent. So any log extensions applied to the parent
```
richvdh, line 36: This linebreak does nothing. Was it your intention to start a new paragraph? (I think it would make sense to do so.) If so, remember you will need a blank line.

richvdh, line 36:
```suggestion
     * Create a child logger.
     * This child will use the `methodFactory` of the parent, so any log extensions applied to the parent
```

src/logger.ts
```diff
@@ -128,14 +132,24 @@ interface PrefixedLogger extends loglevel.Logger, LoggerWithLogMethod {
...
     if (prefixLogger.getChild === undefined) {
         // This is a new loglevel Logger which has not been turned into a PrefixedLogger yet.
         prefixLogger.prefix = prefix;
-        prefixLogger.getChild = (childPrefix): Logger => getPrefixedLogger((prefix ?? "") + childPrefix);
+        prefixLogger.getChild = (childPrefix): Logger => {
+            // create the new child logger
+            const childLogger = getPrefixedLogger((prefix ?? "") + childPrefix);
+            // assign the same methodFactory as the root logger.
+            // this is useful if we add extensions to the root logger that modify
+            // its methodFactory. (an example extension is: storing each log to a rageshake db)
```
richvdh, line 147:
```suggestion
            // its methodFactory. (An example extension is: storing each log to a rageshake db)
```

Sir, is your shift key broken? :smile:

florianduros:
@toger5 the PR breaks the https://github.com/matrix-org/complement-crypto tests. These tests are run when put in the merge queue.

toger5:
@florianduros any idea how it breaks it?

toger5:
@florianduros I think it makes sense to not use the `rebuild` method in that case especailly since we do have a workaround.

src/logger.ts
```diff
@@ -148,7 +148,10 @@ function getPrefixedLogger(prefix?: string): PrefixedLogger {
             // its methodFactory. (An example extension is: storing each log to a rageshake db)
             childLogger.methodFactory = prefixLogger.methodFactory;
             // Rebuild the child logger with the new methodFactory.
-            childLogger.rebuild();
+            // childLogger.rebuild();
+            // TODO: As a stop-gap we use the deprecated method of rebuilding the logger
+            // by setting the level. Complement Crypto uses an older version of loglevel.
```
florianduros, line 153: Should we instead update the dependencies in complement crypto?

toger5: I have done this in: https://github.com/matrix-org/complement-crypto/pull/173

florianduros: Thanks!

package.json
```diff
@@ -56,7 +56,7 @@
         "bs58": "^6.0.0",
         "content-type": "^1.0.4",
         "jwt-decode": "^4.0.0",
-        "loglevel": "^1.7.1",
+        "loglevel": "^1.9.2",
```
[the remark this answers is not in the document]
toger5: It does not hurt to include the minor patches.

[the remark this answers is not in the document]
richvdh: > It does not hurt to include the minor patches.

In this specific situation, I agree, but in general, that's not correct.

The reason it might hurt is that if a downstream application has specifically, for some reason, decided that they want to use 1.9.0 or 1.9.1. (For example, maybe it turns out that 1.9.2 introduces a regression in their application). There is no need for matrix-js-sdk to *insist* on 1.9.2, and as a general principle `package.json` should list the *earliest* versions that we are compatible with. 

Nevertheless that seems an unlikely scenario here. So I'm not going to insist you change it. But bear in mind for future :).

Issue #4818, opened 2025-04-28
Next-gen/OIDC token refresh doesn't take account of pause in execution between refresh and using a new token


hughns:
Against a homeserver running next-gen auth/OIDC with a 5 minute token timeout:

1. The SDK makes a `/sync` request with an expired `access_token`. The request fails with a `401`.
2. The SDK then automatically gets a new `access_token` using the `refresh_token` using the token endpoint.
3. The browser then doesn't give any JS execution time for >5 minutes. In my case I think my laptop was closed.
5. The SDK retries the original `/sync` request using the new `access_token`. However, this fails due to the new `access_token` having now expired.
6. The SDK does not attempt to retry further and treats as full logout.

PR #4819, opened 2025-04-28
Refactor how token refreshing works to be more resilient

t3chguy:
1. ensure we do use the new token if it is not explicitly inhibited by the caller
2. eagerly refresh token if we know it is expired
3. allow refreshing a token multiple times if e.g. on bad connection or the environment has been slept and sufficient time has passed since the last refresh attempt

Fixes https://github.com/matrix-org/matrix-js-sdk/issues/4818
Fixes https://github.com/element-hq/element-web/issues/29816


src/http-api/refresh.ts
```diff
@@ -0,0 +1,165 @@
...
+            const expiresIn = snapshot.expiry.getTime() - Date.now();
+            if (expiresIn <= REFRESH_ON_ERROR_IF_TOKEN_EXPIRES_WITHIN_MS) {
+                return TokenRefreshOutcome.Logout;
+            }
+        }
+
+        if (!snapshot || snapshot?.accessToken === this.opts.accessToken) {
+            if (attempt && attempt > 1) {
+                // Exponential backoff to ensure we don't trash the server
+                await sleep(1000 * 2 ** attempt);
```
dbkr, line 115: Should this be capped maybe?

t3chguy: the delay or the number of attempts?

dbkr: It doesn't seem too bad for it to keep retrying indefinitely, as long as the attempt are far enough apart (might be offline for a long time) so I'd say probably just the delay?

t3chguy: Done, capped at 2^5 seconds

florianduros (approved):

PR #4828, opened 2025-05-06
Write a log line when cancelling verification

richvdh:
... in an attempt to figure out what might have caused a spurious cancellation today


andybalaam (approved):
Looks good, thanks

PR #4835, opened 2025-05-13
Crypto: Support recognising a clearText vs a decrypted to-device event

BillCarsonFr:
Draft: Depends on a new wasm-bindings version https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/pull/226

Notes for reviewer:
Since the integration of the rust-sdk it was not possible to make the difference between a succesfully decrypted toDevice event and a clearText toDevice. With the new bindings it is now possible.

This PR integrates this change. The existing `MatrixEvent` end-points like `isEncrypted()`, `getClearContent()`, `isDecryptionFailure` are now expected to work properly for to-device events.

As for now the SDK is not returning and EncryptionInfo or OlmError, so some of other `MatrixEvent` will not work fully (`getSenderKey()`, `decryptionFailureReason()`).

Not sure what is really needed on all the existing code for `MatrixEvent`, or what could be removed (like all the deprecated DecryptionFailureCodes, the claimedEd25519Key, etc.., and a lots of things are megolm specific)
Maybe a dedicated MatrixToDevice event?

## Checklist

### DIFF
--- a/spec/integ/crypto/to-device-messages.spec.ts
+++ b/spec/integ/crypto/to-device-messages.spec.ts
@@ -19,11 +19,22 @@ import "fake-indexeddb/auto";
 import { IDBFactory } from "fake-indexeddb";
 
 import { getSyncResponse, syncPromise } from "../../test-utils/test-utils";
-import { createClient, type MatrixClient } from "../../../src";
+import {
+    ClientEvent,
+    createClient,
+    EventType,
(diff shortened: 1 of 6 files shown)


richvdh:
Also: I factored out some of the duplication between the two sync implementations in https://github.com/matrix-org/matrix-js-sdk/pull/4863. It would be good to land that first.

PR #4662, opened 2025-01-27
Distinguish room state and timeline events in embedded clients (#4574)

hughns:
* Distinguish room state and timeline events in embedded clients

This change enables room widget clients to take advantage of the more reliable method of communicating room state over the widget API provided by a recent update to MSC2762.

* Add missing awaits

* Upgrade matrix-widget-api

## Checklist

- [ ] Tests written for new code (and old code if feasible).
- [ ] New or updated `public`/`exported` symbols have accurate [TSDoc](https://tsdoc.org/) documentation.
- [ ] Linter and other CI checks pass.
- [ ] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md)).

PR #4849, opened 2025-05-22
Allow the embedded client to work without update_state support

robintown:
I've revived the changes from https://github.com/matrix-org/matrix-js-sdk/pull/4662 that allow RoomWidgetClient to function even if the client into which it is embedded does not support the `update_state` action.

This seems useful for easing the transition outlined at https://github.com/element-hq/voip-internal/issues/289 with our upcoming Element Call release (for which we want to test some unrelated mobile changes, where mobile currently does not support `update_state`), so I propose we add this compatibility layer.

### DIFF
--- a/spec/unit/embedded.spec.ts
+++ b/spec/unit/embedded.spec.ts
@@ -31,6 +31,9 @@ import {
     type IOpenIDCredentials,
     type ISendEventFromWidgetResponseData,
     WidgetApiResponseError,
+    UnstableApiVersion,
+    type ApiVersion,
+    type IRoomEvent,
 } from "matrix-widget-api";
 
 import { createRoomWidgetClient, MatrixError, MsgType, UpdateDelayedEventAction } from "../../src/matrix";
@@ -49,6 +52,7 @@ const testOIDCToken = {
 };
 class MockWidgetApi extends EventEmitter {
     public start = jest.fn().mockResolvedValue(undefined);
+    public getClientVersions = jest.fn();
     public requestCapability = jest.fn().mockResolvedValue(undefined);
     public requestCapabilities = jest.fn().mockResolvedValue(undefined);
     public requestCapabilityForRoomTimeline = jest.fn().mockResolvedValue(undefined);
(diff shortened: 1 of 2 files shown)


toger5 (approved):
This is surprisingly straight forward.
Thanks Robin!

spec/unit/embedded.spec.ts
```diff
@@ -668,10 +680,106 @@ describe("RoomWidgetClient", () => {
                     detail: { data: { state: [event] } },
                 }),
             );
+            // Allow the getClientVersions promise to resolve
+            await new Promise<void>((resolve) => setTimeout(resolve, 0));
```
florianduros, line 684: Can we use mock `getClientVersions` or jest fake timers instead ?   (`flushPromises` if we can't)

src/embedded.ts
```diff
@@ -259,6 +260,10 @@ export class RoomWidgetClient extends MatrixClient {
         if (sendContentLoaded) widgetApi.sendContentLoaded();
     }
 
+    public async supportUpdateState(): Promise<boolean> {
+        return (await this.widgetApi.getClientVersions())?.includes(UnstableApiVersion.MSC2762_UPDATE_STATE);
```
florianduros, line 264: `this.widgetApi.getClientVersions()` returns `Promise<ApiVersion[]>` so there is always an array. 

```suggestion
        return (await this.widgetApi.getClientVersions()).includes(UnstableApiVersion.MSC2762_UPDATE_STATE);
```

If this an error of typing in the widget api, the typing is wrong:

```suggestion
        return (await this.widgetApi.getClientVersions())?.includes(UnstableApiVersion.MSC2762_UPDATE_STATE) || false;
```

spec/unit/embedded.spec.ts
```diff
@@ -96,6 +103,11 @@ class MockWidgetApi extends EventEmitter {
         send: jest.fn(),
         sendComplete: jest.fn(),
     };
+
+    public constructor(clientVersions: ApiVersion[]) {
```
florianduros, line 107: tsdoc is welcomed :)

florianduros (approved):

spec/unit/embedded.spec.ts
toger5: added one

PR #4860, opened 2025-05-29
Include extraParams in all HTTP requests

rsb-tbg:
# Include extraParams in all HTTP requests
 
Fix a bug where queryParams from client configuration were not being attached to all HTTP requests.
 
## Problem
When creating a MatrixClient with `queryParams` in the options, these parameters were stored as `extraParams` in the HTTP API but never actually included in the request URLs. This meant global query parameters were silently ignored.
 
## Solution
Modified the `getUrl` method in `FetchHttpApi` to merge `this.opts.extraParams` with request-specific `queryParams` before encoding them into the URL.
 
## Changes
- Updated `src/http-api/fetch.ts` `getUrl` method to include global `extraParams`
- Request-specific `queryParams` take precedence over global `extraParams` when there are conflicts
- Updated ICreateClientOpts.queryParams to use QueryDict | undefined for consistency with request-specific params
- Added comprehensive unit tests in `spec/unit/http-api/fetch.spec.ts`
- Added integration tests in `spec/integ/matrix-client-opts.spec.ts`
 
## Testing
```bash
# Run unit tests
npm test -- spec/unit/http-api/fetch.spec.ts
 
# Run integration tests
npm test -- spec/integ/matrix-client-opts.spec.ts
```
 
## Verification
1. Create a MatrixClient with `queryParams: { user_id: "@rsb-tbg:localhost" }`
2. Make any API call (sync, send message, etc.)
3. Verify that `user_id=@rsb-tbg:localhost` (encoded, of course) appears in all request URLs
 
Fixes the issue where global query parameters were not being applied to requests.

### DIFF
--- a/spec/integ/matrix-client-opts.spec.ts
+++ b/spec/integ/matrix-client-opts.spec.ts
@@ -205,4 +205,109 @@ describe("MatrixClient opts", function () {
             expect(res.event_id).toEqual("foo");
         });
     });
+
+    describe("with opts.queryParams", function () {
+        let client: MatrixClient;
+        let httpBackend: HttpBackend;
+        const userId = "@rsb-tbg:localhost";
+
+        beforeEach(function () {
+            httpBackend = new HttpBackend();
+            client = new MatrixClient({
+                fetchFn: httpBackend.fetchFn as typeof globalThis.fetch,
+                store: new MemoryStore() as IStore,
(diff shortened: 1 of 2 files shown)


rsb-tbg:
Sorry, I didn't add a label and/or don't have permission but this probably qualifies as T-Defect

PR #4861, opened 2025-06-03
Fix MatrixRTC membership manager failing to rejoin in a race condition (sync vs not found response)

toger5:
This mostly introduces a test which simulates a race the memebership manager can encounter.

As described in the test:
```
            // We enforce the following scenario where we simulate that the delayed event activated and made the user leave:
            // - we wait unti the dealyed event event gets send and then mock its response to be a not found
            // - we enforce a race between the sync that informs us that our call membership state event got set to left
            //   and the not found from the delayed event: we receive the sync while we are waiting for the delayed event to be sent.
            // - While the delayed leave event gets send we inform the manager that our membership state event got set to left.
            //   (onRTCSessionMemberUpdate)
            // - only then we resolve the sending of the delayed event.
            // - we test, that the manager acknowledges the leave and sends a new membership state event.
```

## Checklist


spec/test-utils/test-run-helper.ts
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+/**
+ * Custom vscode option so that we can debug the tests in peace without running into timeouts.
+ */
+export function setVSCodeJestDebugTimeoutOverride(): void {
+    if (process.env.VSCODE_INSPECTOR_OPTIONS) {
+        jest.setTimeout(60 * 1000 * 5); // 5 minutes
+    }
+}
```
robintown: Should this instead be done globally for all tests in the setupTests.ts file?

robintown (commented):
Nice find, the fix looks sensible

spec/test-utils/test-run-helper.ts
toger5: If we want it in all tests, yes. I first wanted to see what ppl think about it in general, if there is a better way.

toger5: I think I like that

t3chguy (approved):

spec/setupTests.ts
```diff
@@ -19,3 +19,8 @@ jest.mock("../src/http-api/utils", () => ({
     // We mock timeoutSignal otherwise it causes tests to leave timers running
     timeoutSignal: () => new AbortController().signal,
 }));
+
+// Dont make test fail too soon due to timeouts while debugging.
+if (process.env.VSCODE_INSPECTOR_OPTIONS) {
```
BillCarsonFr, line 24: TIL

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -311,6 +324,44 @@ describe.each([
...
+                new MatrixError({ errcode: "M_NOT_FOUND" }),
+            );
+
+            const { resolve } = createAsyncHandle(client._unstable_sendDelayedStateEvent);
+            await jest.advanceTimersByTimeAsync(RESTART_DELAY);
+            // first simulate the sync, then resolve sending the delayed event.
+            await manager.onRTCSessionMemberUpdate([mockCallMembership(membershipTemplate, room.roomId)]);
+            resolve({ delay_id: "id" });
+            // Let the scheduler run one iteration so that the new join gets sent
+            await jest.advanceTimersByTimeAsync(0);
```
BillCarsonFr: Nit: looks like a `runOnlyPendingTimersAsync()`  ?

src/matrixrtc/NewMembershipManager.ts
```diff
@@ -229,14 +229,15 @@ export class MembershipManager
...
+            this.state.hasMemberStateEvent = false;
+
             if (this.scheduler.actions.find((a) => sendingMembershipActions.includes(a.type as MembershipActionType))) {
                 this.logger.error(
-                    "NewMembershipManger tried adding another `SendFirstDelayedEvent` actions even though we already have one in the Queue\nActionQueueOnMemberUpdate:",
+                    "NewMembershipManger tried adding another `SendDelayedEvent` actions even though we already have one in the Queue\nActionQueueOnMemberUpdate:",
                     this.scheduler.actions,
                 );
             } else {
                 // Only react to our own membership missing if we have not already scheduled sending a new membership DirectMembershipManagerAction.Join
```
BillCarsonFr, line 240: So looks like the removing `this.state.hasMemberStateEvent = false` is the important change. Is this comment still up-to-date?

BillCarsonFr (approved):
Just small nits. Thanks for getting to the bottom of it by creating a test to reproduce the race

src/matrixrtc/NewMembershipManager.ts
hughns: It's not removing the line, it is moving it a few lines up so that it is always executed.

As such, the comment is still up to date.

toger5: Yes, "react" focuses on: we initiate new actions in the action scheduler. Which we still only do if we dont have the correct actions set up yet. With the `sending a new membership DirectMembershipManagerAction.Join` bit i feel like the comment sounds like its talking about "react with an action schedule update"?

```suggestion
                // Only react to our own membership missing by scheduling new actions if we have not already scheduled sending a new membership DirectMembershipManagerAction.Join
```
Do you think it should be changed to this?

PR #4862, opened 2025-06-03
Remove `LegacyMembershipManager`

toger5:
The legacy membership manager was kept around during the transition phase. It is now in a spot where we dont use it anymore and where we are getting better results with the new implementation.

This removes all related code and deprecates fields related to it so that downstream projects can adjust. (in particular: `useNewMembershipManager`)

There was a custom test file that tested both (new and legacy implementation) this file used describe.each and now just became a simple describe. All helper code for this setup also got removed.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,8 +13,7 @@ out
 /dist
 /lib
 
-# version file and tarball created by `npm pack` / `yarn pack`
-/git-revision.txt
+# tarball created by `npm pack` / `yarn pack`
 /matrix-js-sdk-*.tgz
 
 .vscode
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+Changes in [37.9.0](https://github.com/matrix-org/matrix-js-sdk/releases/tag/v37.9.0) (2025-06-17)
+==================================================================================================
+## 🐛 Bug Fixes
+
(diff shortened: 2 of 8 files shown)


spec/unit/matrixrtc/MatrixRTCSession.spec.ts
```diff
@@ -201,58 +201,6 @@ describe("MatrixRTCSession", () => {
...
-            sess.joinRoomSession([mockFocus], mockFocus, joinSessionConfig);
-            await Promise.race([sentStateEvent, new Promise((resolve) => setTimeout(resolve, 500))]);
-
-            expect(sendStateEventMock).toHaveBeenCalledTimes(1);
-
-            await Promise.race([sentDelayedState, new Promise((resolve) => setTimeout(resolve, 500))]);
-            expect(sendDelayedStateMock).toHaveBeenCalledTimes(1);
-        }
-
-        it("sends events", async () => {
```
robintown, line 251: Are this and the tests below it being removed because they never were testing the new membership manager?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -77,6 +76,7 @@ export interface MembershipConfig {
      * Use the new Manager.
      *
      * Default: `false`.
+     * @deprecated does nothing anymore we always default to the new memberhip manager.
```
robintown, line 79: I see, so this isn't even a breaking change as we're not exporting the membership managers to begin with. Nice :+1:

PR #4863, opened 2025-06-04
Factor out common code for processing to-device events

richvdh:
`sliding-sync-sdk.ts` and `sync.ts` both have copies of this code, and it's redundant.


dbkr (approved):
Thanks!

PR #4877, opened 2025-06-17
Add prepare script

Half-Shot:
This is useful for projects which checkout the matrix-js-sdk as part of their CI process, and rely on the `matrix-js-sdk/lib` paths to exist. I believe this is relatively benign.

## Checklist

- [N/A] Tests written for new code (and old code if feasible).
- [N/A] New or updated `public`/`exported` symbols have accurate [TSDoc](https://tsdoc.org/) documentation.
- [N/A] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md)).

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,6 @@
     },
     "scripts": {
         "prepare": "yarn build",
-        "prepack": "yarn build",
         "start": "echo THIS IS FOR LEGACY PURPOSES ONLY. && babel src -w -s -d lib --verbose --extensions \".ts,.js\"",
         "clean": "rimraf lib",
         "build": "yarn build:dev",


package.json
```diff
@@ -6,6 +6,7 @@
         "node": ">=22.0.0"
     },
     "scripts": {
+        "prepare": "yarn build",
         "prepack": "yarn build",
```
t3chguy: Is prepack still needed? I'd imagine not

https://docs.npmjs.com/cli/v9/using-npm/scripts#life-cycle-scripts

Half-Shot: > Runs BEFORE the package is packed, i.e. during npm publish and npm pack

Sweet :hocho:

t3chguy (approved):

richvdh:
Doesn't this mean that we will `npm build` when `yarn install`ing a release version of matrix-js-sdk? That sounds slow and unsatisfactory.

If you're checking out matrix-js-sdk from git as part of your CI, I think you should expect to have to build it yourself.

richvdh:
Ah sorry, I think I misread the life-cycle-scripts doc. Carry on!

richvdh:
That said, could you please clarify the point of this? If I check out `matrix-js-sdk` from git, whether as part of my CI or otherwise, I have to build it. What exactly is changing as a result of this PR?

PR #4878, opened 2025-06-17
Allow customizing the IndexedDB database prefix used by Rust crypto.

clokep:
Related to #3974

I'm not 100% sure this makes sense, but I think it'll do what I want. (Which is having completely separate databases, but under the same principal/domain in IndexedDB.)

## Checklist

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1527,9 +1527,14 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     /**
      * Clear any data out of the persistent stores used by the client.
      *
+     * @param args.databasePrefix - The database name to use for indexeddb, defaults to 'matrix-js-sdk'.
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -1527,9 +1527,14 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     /**
      * Clear any data out of the persistent stores used by the client.
      *
+     * @param args.databasePrefix - The database name to use for indexeddb, defaults to 'matrix-js-sdk'.
      * @returns Promise which resolves when the stores have been cleared.
      */
-    public clearStores(): Promise<void> {
+    public clearStores(
+        args: {
+            databasePrefix?: string;
```
richvdh: this should maybe be `cryptoDatabasePrefix` ?

clokep:
Thank you!

PR #4879, opened 2025-06-18
Add account data key for enabling/disabling recovery

uhoreg:
https://github.com/element-hq/element-web/pull/30075 uses account data to remember whether recovery is enabled/disabled.  This adds the key to the account data events type


florianduros (approved):

t3chguy:
See https://github.com/element-hq/element-web/blob/2034f8b6bbbb73bbb9c275cd21c3a8a28a5682f0/src/%40types/matrix-js-sdk.d.ts#L64-L95 for an example

PR #4881, opened 2025-06-20
Remove redundant git-revision.txt file

t3chguy:


kegsay (approved):
Thanks. For future readers, this was breaking the ability to `yarn add matrix-js-sdk@https://github.com/matrix-org/matrix-js-sdk#develop` then `yarn install` because the directory created isn't a git repo so lacks `.git`. This was fine before if you were only `yarn add`ing to use `/src` directly, but since https://github.com/matrix-org/matrix-js-sdk/pull/4877 `yarn install` will now call `yarn build`, thus breaking my workflow which is used in complement-crypto - https://github.com/matrix-org/complement-crypto/actions/runs/15760042147/job/44435000767

florianduros (approved):

PR #4882, opened 2025-06-21
Update matrix-sdk-crypto-wasm to `15.0.0`

richvdh:
For js-sdk users, this includes the following:

-   Send stable identifier `sender_device_keys` for MSC4147 (Including device keys with Olm-encrypted events).
    ([#4964](https://github.com/matrix-org/matrix-rust-sdk/pull/4964))

-   Check the `sender_device_keys` field on _all_ incoming Olm-encrypted to-device messages and ignore any to-device messages which include the field but whose data is invalid (as per [MSC4147](https://github.com/matrix-org/matrix-spec-proposals/pull/4147)).
    ([#4922](https://github.com/matrix-org/matrix-rust-sdk/pull/4922))

-   Fix bug which caused room keys to be unnecessarily rotated on every send in the presence of blacklisted/withheld devices in the room.
    ([#4954](https://github.com/matrix-org/matrix-rust-sdk/pull/4954))

-   Fix [matrix-rust-sdk#2729](https://github.com/matrix-org/matrix-rust-sdk/issues/2729) which in rare cases can cause room key oversharing.
    ([#4975](https://github.com/matrix-org/matrix-rust-sdk/pull/4975))

### DIFF
--- a/src/rust-crypto/rust-crypto.ts
+++ b/src/rust-crypto/rust-crypto.ts
@@ -35,7 +35,7 @@ import { type IHttpOpts, type MatrixHttpApi, Method } from "../http-api/index.ts
 import { RoomEncryptor } from "./RoomEncryptor.ts";
 import { OutgoingRequestProcessor } from "./OutgoingRequestProcessor.ts";
 import { KeyClaimManager } from "./KeyClaimManager.ts";
-import { logDuration, MapWithDefault } from "../utils.ts";
+import { MapWithDefault } from "../utils.ts";
 import {
     type BackupTrustInfo,
     type BootstrapCrossSigningOpts,
@@ -1494,17 +1494,14 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
         unusedFallbackKeys?: Set<string>;
         devices?: RustSdkCryptoJs.DeviceLists;
     }): Promise<IToDeviceEvent[]> {
-        const result = await logDuration(logger, "receiveSyncChanges", async () => {
-            return await this.olmMachine.receiveSyncChanges(
-                events ? JSON.stringify(events) : "[]",
-                devices,
-                oneTimeKeysCounts,
-                unusedFallbackKeys,
-            );
-        });
+        const result = await this.olmMachine.receiveSyncChanges(
+            events ? JSON.stringify(events) : "[]",
+            devices,
+            oneTimeKeysCounts,
+            unusedFallbackKeys,
+        );
 
-        // receiveSyncChanges returns a JSON-encoded list of decrypted to-device messages.
-        return JSON.parse(result);
+        return result.map((processed) => JSON.parse(processed.rawEvent));
     }
 
     /** called by the sync loop to preprocess incoming to-device messages
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     ],
     "dependencies": {
         "@babel/runtime": "^7.12.5",
-        "@matrix-org/matrix-sdk-crypto-wasm": "^14.2.0",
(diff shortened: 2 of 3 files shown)


src/rust-crypto/rust-crypto.ts
```diff
@@ -1494,17 +1494,14 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
         unusedFallbackKeys?: Set<string>;
         devices?: RustSdkCryptoJs.DeviceLists;
     }): Promise<IToDeviceEvent[]> {
-        const result = await logDuration(logger, "receiveSyncChanges", async () => {
```
richvdh, line 1497: I decided to drop this spammy `logDuration` while I was in the area.

src/rust-crypto/rust-crypto.ts
```diff
@@ -1494,17 +1494,14 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+        const result = await this.olmMachine.receiveSyncChanges(
+            events ? JSON.stringify(events) : "[]",
+            devices,
+            oneTimeKeysCounts,
+            unusedFallbackKeys,
+        );
 
-        // receiveSyncChanges returns a JSON-encoded list of decrypted to-device messages.
-        return JSON.parse(result);
+        return result.map((processed) => JSON.parse(processed.rawEvent));
```
richvdh, line 1504: per https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/236 and https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/246, `receiveSyncChanges` now returns a list of `ProcessedToDeviceEvent` objects rather than just the raw JSON.

src/rust-crypto/rust-crypto.ts
```diff
@@ -1494,17 +1494,14 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
     }): Promise<IToDeviceEvent[]> {
-        const result = await logDuration(logger, "receiveSyncChanges", async () => {
-            return await this.olmMachine.receiveSyncChanges(
-                events ? JSON.stringify(events) : "[]",
-                devices,
-                oneTimeKeysCounts,
-                unusedFallbackKeys,
-            );
-        });
+        const result = await this.olmMachine.receiveSyncChanges(
```
t3chguy, line 1497: `receiveSyncChanges` returns `Promise<any>` can you please provide an improved type in `matrix-sdk-crypto-wasm.d.ts` if it is not possible for the upstream types to be improved.


`  receiveSyncChanges(to_device_events: string, changed_devices: DeviceLists, one_time_keys_counts: Map<any, any>, unused_fallback_keys?: Set<any> | null): Promise<any>;
`

richvdh: It's already addressed upstream, AFAIK? https://github.com/matrix-org/matrix-sdk-crypto-wasm/blob/e3b691b35ba4c4207b32a964a19d8c9f024041be/index.d.ts#L79

t3chguy: ah great, confusing to use declaration merging to override one's own types but if it works then great

richvdh: It was a workaround for https://github.com/rustwasm/wasm-bindgen/issues/1197. It looks like there's now a fix for that, but we haven't tried it yet.

andybalaam (approved):

richvdh:
CI failure should be fixed by https://github.com/matrix-org/complement-crypto/pull/196

richvdh:
Flaky test, filed as https://github.com/element-hq/element-web/issues/30188

PR #4891, opened 2025-06-30
crypto: Add new ClientEvent.ReceivedToDeviceMessage with proper OlmEncryptionInfo support

BillCarsonFr:
## Checklist


src/client.ts
```diff
@@ -885,7 +886,9 @@ const EVENT_ID_PREFIX = "$";
 export enum ClientEvent {
     Sync = "sync",
     Event = "event",
+    /** @deprecated Use `ReceivedToDeviceMessage`*/
```
richvdh:
```suggestion
    /** @deprecated Use {@link ReceivedToDeviceMessage}. */
```

src/client.ts
```diff
@@ -1088,6 +1091,23 @@ export type ClientEventHandlerMap = {
...
+     * @param encryptionInfo - The encryptionInfo for this message, null if sent in clear.
+     * @example
+     * ```
+     * matrixClient.on("receivedToDeviceMessage", function(message, encryptionInfo){
+     *   var claimed_sender = encryptionInfo ? encryptionInfo.sender : message.sender;
+     *   var isVerified = encryptionInfo ? encryptionInfo.verified : false;
+     *   var type = message.type;
+     * });
+     * ```
+     */
```
richvdh: For the record, this documentation is actually in the wrong place: it should be on `ClientEvent.ReceivedToDeviceMessage` (which is what the application developer will actually use) rather than on the map entry (which is just here for typescript, really). Have a look at `CryptoEvent`, which does a better job of this.

Of course, that's true of all the `ClientEvent`s entries, so no need to fix it as part of this PR. Just sharing FYI really.

src/crypto-api/index.ts
```diff
@@ -1365,6 +1365,24 @@ export interface OwnDeviceKeys {
     curve25519: string;
 }
 
+export interface OlmEncryptionInfo {
```
richvdh, line 1371: could you give this type a doc-comment explaining what it represents, please.

src/crypto-api/index.ts
```diff
@@ -1365,6 +1365,24 @@ export interface OwnDeviceKeys {
...
+    /** The user ID of the event sender, note this is untrusted data unless `isVerified` is true **/
+    sender: string;
```
richvdh, line 1373:
```suggestion
    /** The user ID of the event sender. Note this is untrusted data unless {@link isVerified} is true. **/
    sender: string;
```

src/crypto-api/index.ts
```diff
@@ -1365,6 +1365,24 @@ export interface OwnDeviceKeys {
...
+    /**
+     * The device ID of the device that sent us the event,
+     * note this is untrusted data unless `isVerified` is true.
+     * If the device ID is not known, this will be `null`.
```
richvdh:
```suggestion
     * The device ID of the device that sent us the event. Note this is untrusted data
     * unless {@link isVerified} is true. If the device ID is not known, this will be `null`.
```

src/crypto-api/index.ts
```diff
@@ -1365,6 +1365,24 @@ export interface OwnDeviceKeys {
...
+     **/
+    senderDevice?: string;
+    /** The sender device key, base64 encoded **/
```
richvdh:
```suggestion
    /** The sender device's public Curve25519 key, base64 encoded **/
```

src/crypto-api/index.ts
```diff
@@ -1365,6 +1365,24 @@ export interface OwnDeviceKeys {
...
+    senderKeyPublicBase64: string;
+    /**
+     * The verification state of the device that sent us the event, note this
```
richvdh: It's not really "the verification state", since it's just a boolean. More: whether the device was verified.

But: what does that actually mean? Does it mean that the owner has verified the device? That we have verified the owner?

src/sync-accumulator.ts
```diff
@@ -134,12 +135,36 @@ interface IAccountData {
     events: IMinimalEvent[];
 }
 
+/** A to-devive message as received from the sync. */
```
andybalaam: devive->device

src/sync-accumulator.ts
```diff
@@ -134,12 +135,36 @@ interface IAccountData {
...
 export interface IToDeviceEvent {
     content: IContent;
     sender: string;
     type: string;
 }
 
+/**
+ * Successfully processed to-device message.
+ */
+export type IToDeviceMessage = IToDeviceEvent;
```
andybalaam: What is the purpose of this alias? The compiler won't complain if you use the wrong one, right?

Also, what is the difference in meaning between `IToDeviceMessage` and `IToDeviceEvent`.

(Also, do we still use `I` prefix for interfaces?)

src/crypto-api/index.ts
```diff
@@ -1365,6 +1365,24 @@ export interface OwnDeviceKeys {
...
+    /** The user ID of the event sender, note this is untrusted data unless `isVerified` is true **/
+    sender: string;
+    /**
+     * The device ID of the device that sent us the event,
+     * note this is untrusted data unless `isVerified` is true.
+     * If the device ID is not known, this will be `null`.
+     **/
+    senderDevice?: string;
+    /** The sender device key, base64 encoded **/
+    senderKeyPublicBase64: string;
```
richvdh: This is a bit of an awkward name. At the very least, it should be "senderPublicKeyBase64", but (a) that doesn't tell us which public key it is, and (b) I don't think we particularly need to emphasise that it's the *public* key, since it's unlikely we would ever have the sender's private key. (As above, we can put that in the docs to clarify). So:

```suggestion
    senderCurve25519KeyBase64: string;
```

src/crypto-api/index.ts
```diff
@@ -1365,6 +1365,24 @@ export interface OwnDeviceKeys {
...
+    /**
+     * The verification state of the device that sent us the event, note this
+     *  is the state of the device at the time of decryption.
+     */
+    isVerified: boolean;
```
richvdh: It feels a bit odd that all the other fields have a `sender` prefix but this one does not?

```suggestion
    senderVerified: boolean;
```

src/sync.ts
```diff
@@ -1911,52 +1923,46 @@ export function _createAndReEmitRoom(client: MatrixClient, roomId: string, opts:
...
-
-            if (
-                toDeviceEvent.getType() === "m.key.verification.start" ||
-                toDeviceEvent.getType() === "m.key.verification.request"
-            ) {
-                const txnId = content["transaction_id"];
-                if (cancelledKeyVerificationTxns.includes(txnId)) {
-                    toDeviceEvent.flagCancelled();
+        .forEach(function (processedEvent) {
+            // For backwards compatibility, we emit the event as a MatrixEvent
```
andybalaam: Instead of emitting `processedEvent`? If so, it would help me if you added that to the comment.

src/client.ts
richvdh: I realise this isn't new with your change, but we no longer emit `UndecryptableToDeviceEvent`. We should deprecate that too.

src/sliding-sync-sdk.ts
```diff
@@ -150,11 +151,22 @@ class ExtensionToDevice implements Extension<ExtensionToDeviceRequest, Extension
...
-        let events = data["events"] || [];
-        if (events.length > 0 && this.cryptoCallbacks) {
-            events = await this.cryptoCallbacks.preprocessToDeviceMessages(events);
+        const events = data["events"] || [];
+        let receivedToDeviceMessages: ReceivedToDeviceMessage[];
+        if (this.cryptoCallbacks) {
+            receivedToDeviceMessages = await this.cryptoCallbacks.preprocessToDeviceMessages(events);
+        } else {
+            receivedToDeviceMessages = events
+                .filter((e) => e.type !== EventType.RoomMessageEncrypted)
```
richvdh: I wonder if we need this. We could just pass the encrypted messages up to the app?

andybalaam (approved):
I don't fully understand this, but it looks sensible. I'd like someone with more clue to take a look too.

src/sliding-sync-sdk.ts
```diff
@@ -150,11 +151,22 @@ class ExtensionToDevice implements Extension<ExtensionToDeviceRequest, Extension
...
+                .map((clearEvent) => {
+                    // Crypto is not enabled, so we just return the clear events.
+                    return {
+                        message: clearEvent,
+                        encryptionInfo: null,
+                    };
+                });
```
richvdh: A little more succinct:
```suggestion
                .map((clearEvent) => ({
                    message: clearEvent,
                    encryptionInfo: null,
                }));
```

... and put the comment at the start of the `else` clause.

src/sync-accumulator.ts
richvdh:
```suggestion
/** A (possibly decrypted) to-device message as received from the sync. */
```

src/sync-accumulator.ts
richvdh: eh, let's not have two identical types. Let's just use `IToDeviceEvent`?

src/client.ts
```diff
@@ -1088,6 +1091,23 @@ export type ClientEventHandlerMap = {
...
+    [ClientEvent.ReceivedToDeviceMessage]: (
+        message: IToDeviceMessage,
+        encryptionInfo: OlmEncryptionInfo | null,
+    ) => void;
```
richvdh: wouldn't it make more sense to use the `ReceivedToDeviceMessage` type as the payload?

src/sync.ts
```diff
@@ -1909,54 +1922,59 @@ export function _createAndReEmitRoom(client: MatrixClient, roomId: string, opts:
 /**
  * Process a list of (decrypted, where possible) received to-device events.
  *
- * Converts the events into `MatrixEvent`s, and emits appropriate {@link ClientEvent.ToDeviceEvent} events.
+ * Converts the to-device message and OlmEncryptionIngo and emits appropriate {@link ClientEvent.ReceivedToDeviceMessage} events.
```
richvdh: this sentence doesn't seem to have enough words.

src/sync.ts
```diff
@@ -1909,54 +1922,59 @@ export function _createAndReEmitRoom(client: MatrixClient, roomId: string, opts:
...
+ * Converts the events into `MatrixEvent`s, and emits the now deprecated {@link ClientEvent.ToDeviceEvent} events for compatibility.
```
richvdh:
```suggestion
 * Also converts the events into `MatrixEvent`s, and emits the now deprecated {@link ClientEvent.ToDeviceEvent} events for compatibility.
```

src/sync.ts
```diff
@@ -1909,54 +1922,59 @@ export function _createAndReEmitRoom(client: MatrixClient, roomId: string, opts:
...
-            ) {
-                const txnId = content["transaction_id"];
-                if (cancelledKeyVerificationTxns.includes(txnId)) {
-                    toDeviceEvent.flagCancelled();
+        .forEach(function (processedEvent) {
+            // For backwards compatibility, we emit the event as a MatrixEvent
+            {
+                const toDeviceEvent = processedEvent.message;
+                const content = toDeviceEvent.content;
+                const deprecatedCompatibilityEvent = new MatrixEvent(Object.assign({}, toDeviceEvent));
```
richvdh, line 1955: what is the purpose of the `Object.assign` here? might be worth adding a comment?

spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -551,11 +551,11 @@ describe("RustCrypto", () => {
             const inputs: IToDeviceEvent[] = [
                 { content: { key: "value" }, type: "org.matrix.test", sender: "@alice:example.com" },
             ];
-            const res = await rustCrypto.preprocessToDeviceMessages(inputs);
+            const res = (await rustCrypto.preprocessToDeviceMessages(inputs)).map((p) => p.message);
             expect(res).toEqual(inputs);
         });
 
-        it("should pass through bad encrypted messages", async () => {
+        it("should fail to process bad encrypted messages", async () => {
```
richvdh, line 558: > There is a change that was revealed by this [test](https://github.com/matrix-org/matrix-js-sdk/pull/4891/commits/f0a629bde24c4f1580fd36a42d9c5097aa49ef7e). Now UTD events are not returned by the preprocessToDeviceMessages. I don't really see a reason to return them?

Agreed.

src/sync-accumulator.ts
BillCarsonFr: I am trying to match a bit more the rust sdk terminlogy that uses `message` and not `event` for to-device (they don't have event_ids, and other properties compared to events).
Unfortunatly the spec uses a mix of message and event for to-device. 
So I kept `IToDeviceEvent` for the one coming down the sync, and `IToDeviceMessage` for the one succesfully processed (and decrypted if applicable)

src/crypto-api/index.ts
BillCarsonFr: > it's unlikely we would ever have the sender's private key.

On that subject I remember some questions in rooms and also issues of users concerned about seeing "keys" with a base64 values in UI and in logs. They were scared that this could be exposing private keys.
Of couse we don't do that, but I started to think that there is no reason to add `public` when it is public. But maybe that should be tracked as a separate issue

src/sliding-sync-sdk.ts
BillCarsonFr: Yes I noted that in the PR description
> There is a change that was revealed by this [test](https://github.com/matrix-org/matrix-js-sdk/pull/4891/commits/f0a629bde24c4f1580fd36a42d9c5097aa49ef7e). Now UTD events are not returned by the preprocessToDeviceMessages. I don't really see a reason to return them? But we could return them as is with a null encryption info

But I think yes we could pass them. To keep the actual behavior

src/client.ts
BillCarsonFr: I tried that first, but then I changed to use that. Because I thought it was more  natural to do like that in `javascript`? I may be just wrong. I have no strong feeling, will move to use `ReceivedToDeviceMessage`

src/sliding-sync-sdk.ts
BillCarsonFr: oh, it is a bit different. We are in the case were encryption is disabled (or config problem as there are no crypto-callbacks). So we could in that case pass the event as is. BUT when encryption is enabled, we can still discard the UTD events.

src/sync-accumulator.ts
BillCarsonFr: I removed it and keep only `IToDeviceEvent` https://github.com/matrix-org/matrix-js-sdk/pull/4891/commits/7a843ab03fed98d027a1a9e77084cca97688ca8b

src/crypto-api/index.ts
BillCarsonFr: https://github.com/matrix-org/matrix-js-sdk/pull/4891/commits/f17a5368ac954ea49b2430427cb8130d0445ef8a

src/crypto-api/index.ts
BillCarsonFr: I updated the doc:
>  If true, this message is guaranteed to be authentic as it is coming from a device belonging to a user that we have verified.

src/sliding-sync-sdk.ts
BillCarsonFr: done https://github.com/matrix-org/matrix-js-sdk/pull/4891/commits/785479c50687b0ed61767ae5139434f868f33eed

src/sync-accumulator.ts
BillCarsonFr: done

src/client.ts
BillCarsonFr: https://github.com/matrix-org/matrix-js-sdk/pull/4893

src/client.ts
BillCarsonFr: done here https://github.com/matrix-org/matrix-js-sdk/pull/4891/commits/c3ecd15297e61c41ac2689b3883005d22da9ee3c

src/sync.ts
BillCarsonFr: added. 
copy here:
```
// The message is cloned before being passed to the MatrixEvent constructor, because
// the `makeEncrypted` method will mutate the type and content properties of the original message and will interfere
// with the emitted event for `ReceivedToDeviceMessage`.

PR #4897, opened 2025-07-02
Use the client logger in more places

richvdh:
There is a `Logger` instance associated with each MatrixClient, which we should be using in preference to the global logger, to make it easier to debug systems where multiple clients are running, or where the application wants to use a `Logger` implementation other than `console.log`.

This PR goes through a number of places that are currently doing the wrong thing, and then marks the global logger as deprecated. There's more to be done here, but it's a significant improvement for simple usecases.


richvdh:
going to close this and replace it with smaller PRs

PR #4899, opened 2025-07-02
Use client logger in more places (core code)

richvdh:
There is a `Logger` instance associated with each `MatrixClient`, which we should be using in preference to the global logger, to make it easier to debug systems where multiple clients are running, or where the application wants to use a Logger implementation other than `console.log`.

This PR is one of a series that fixes up parts of the code to use the right logger. This one affects core code.

This PR also marks the global `logger` as deprecated, to try to stop people using it in new code.


toger5:
How would the client logger be accessed. It is `private readonly`?

richvdh:
> How would the client logger be accessed. It is private readonly?

Accessed where? Generally I'd expect the client to pass it into whatever you're constructing.

PR #4903, opened 2025-07-07
Tests: gate logging behind `DEBUG` env var

richvdh:
I got fed up with the amount of log spew that the tests generate: they make it almost impossible to see what's actually going on when you run the tests, in terms of which tests succeed and which fail.

This PR modifies a number of the crypto tests so that they send their logging though [debug](https://github.com/debug-js/debug) rather than just spewing it all out to the console. Hence, logging can be enabled by setting the DEBUG env var, but otherwise it is pretty much silent.

I guess we might want to do some setup so that some logs are enabled (and maybe go to a file?) under CI, but that can happen later if we need it.

Review commit-by-commit.

### DIFF
--- a/spec/test-utils/test-utils.ts
+++ b/spec/test-utils/test-utils.ts
@@ -1,6 +1,7 @@
+import mkdebug from "debug";
+
 // eslint-disable-next-line no-restricted-imports
 import type EventEmitter from "events";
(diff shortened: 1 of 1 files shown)


uhoreg (approved):
looks plausible

PR #4911, opened 2025-07-08
RTCEncryptionManager: Joiner key rotation grace period

BillCarsonFr:
Fixes https://github.com/element-hq/voip-internal/issues/371

## Checklist

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -19,7 +19,7 @@ import { TypedEventEmitter } from "../models/typed-event-emitter.ts";
 import { EventTimeline } from "../models/event-timeline.ts";
 import { type Room } from "../models/room.ts";
 import { type MatrixClient } from "../client.ts";
-import { EventType } from "../@types/event.ts";
+import { EventType, RelationType } from "../@types/event.ts";
 import { CallMembership } from "./CallMembership.ts";
 import { RoomStateEvent } from "../models/room-state.ts";
 import { type Focus } from "./focus.ts";
@@ -27,7 +27,7 @@ import { KnownMembership } from "../@types/membership.ts";
 import { MembershipManager } from "./MembershipManager.ts";
 import { EncryptionManager, type IEncryptionManager } from "./EncryptionManager.ts";
 import { logDurationSync } from "../utils.ts";
-import { type Statistics } from "./types.ts";
+import { type Statistics, type RTCNotificationType } from "./types.ts";
 import { RoomKeyTransport } from "./RoomKeyTransport.ts";
 import type { IMembershipManager } from "./IMembershipManager.ts";
 import { RTCEncryptionManager } from "./RTCEncryptionManager.ts";
@@ -65,6 +65,15 @@ export type MatrixRTCSessionEventHandlerMap = {
     ) => void;
     [MatrixRTCSessionEvent.MembershipManagerError]: (error: unknown) => void;
 };
+
+export interface SessionConfig {
+    /**
+     * What kind of notification to send when starting the session.
+     * @default `undefined` (no notification)
+     */
+    notificationType?: RTCNotificationType;
+}
+
 // The names follow these principles:
 // - we use the technical term delay if the option is related to delayed events.
 // - we use delayedLeaveEvent if the option is related to the delayed leave event.
@@ -184,7 +193,7 @@ export interface EncryptionConfig {
      */
     useKeyDelay?: number;
 }
-export type JoinSessionConfig = MembershipConfig & EncryptionConfig;
+export type JoinSessionConfig = SessionConfig & MembershipConfig & EncryptionConfig;
 
 /**
  * A MatrixRTCSession manages the membership & properties of a MatrixRTC session.
@@ -198,7 +207,10 @@ export class MatrixRTCSession extends TypedEventEmitter<
     private encryptionManager?: IEncryptionManager;
     // The session Id of the call, this is the call_id of the call Member event.
     private _callId: string | undefined;
+    private joinConfig?: SessionConfig;
     private logger: Logger;
+
+    private pendingNotificationToSend: undefined | RTCNotificationType;
     /**
      * This timeout is responsible to track any expiration. We need to know when we have to start
      * to ignore other call members. There is no callback for this. This timeout will always be configured to
@@ -463,6 +475,9 @@ export class MatrixRTCSession extends TypedEventEmitter<
             }
         }
 
+        this.joinConfig = joinConfig;
+        this.pendingNotificationToSend = this.joinConfig?.notificationType;
+
         // Join!
         this.membershipManager!.join(fociPreferred, fociActive, (e) => {
             this.logger.error("MembershipManager encountered an unrecoverable error: ", e);
@@ -564,6 +579,35 @@ export class MatrixRTCSession extends TypedEventEmitter<
     }
 
     /**
+     * Sends a notification corresponding to the configured notify type.
+     */
+    private sendCallNotify(parentEventId: string, notificationType: RTCNotificationType): void {
+        // Send legacy event:
+        this.client
+            .sendEvent(this.roomSubset.roomId, EventType.CallNotify, {
+                "application": "m.call",
+                "m.mentions": { user_ids: [], room: true },
+                "notify_type": notificationType === "notification" ? "notify" : notificationType,
+                "call_id": this.callId!,
+            })
+            .catch((e) => this.logger.error("Failed to send call notification", e));
+
+        // Send new event:
+        this.client
+            .sendEvent(this.roomSubset.roomId, EventType.RTCNotification, {
+                "m.mentions": { user_ids: [], room: true },
+                "notification_type": notificationType,
+                "m.relates_to": {
+                    event_id: parentEventId,
+                    rel_type: RelationType.unstable_RTCNotificationParent,
+                },
+                "sender_ts": Date.now(),
+                "lifetime": 30_000, // 30 seconds
+            })
+            .catch((e) => this.logger.error("Failed to send call notification", e));
+    }
+
+    /**
      * Call this when the Matrix room members have changed.
      */
     public onRoomMemberUpdate = (): void => {
@@ -603,6 +647,20 @@ export class MatrixRTCSession extends TypedEventEmitter<
             });
 
             void this.membershipManager?.onRTCSessionMemberUpdate(this.memberships);
+            // The `ownMembership` will be set when calling `onRTCSessionMemberUpdate`.
+            const ownMembership = this.membershipManager?.ownMembership;
+            if (this.pendingNotificationToSend && ownMembership && oldMemberships.length === 0) {
+                // If we're the first member in the call, we're responsible for
+                // sending the notification event
+                if (ownMembership.eventId && this.joinConfig?.notificationType) {
+                    this.sendCallNotify(ownMembership.eventId, this.joinConfig.notificationType);
+                } else {
+                    this.logger.warn("Own membership eventId is undefined, cannot send call notification");
+                }
+            }
+            // If anyone else joins the session it is no longer our responsibility to send the notification.
+            // (If we were the joiner we already did sent the notification in the block above.)
+            if (this.memberships.length > 0) this.pendingNotificationToSend = undefined;
         }
         // This also needs to be done if `changed` = false
         // A member might have updated their fingerprint (created_ts)
--- a/spec/unit/matrixrtc/MatrixRTCSession.spec.ts
+++ b/spec/unit/matrixrtc/MatrixRTCSession.spec.ts
@@ -16,11 +16,10 @@ limitations under the License.
 
 import { encodeBase64, EventType, MatrixClient, type MatrixError, type MatrixEvent, type Room } from "../../../src";
 import { KnownMembership } from "../../../src/@types/membership";
-import { type SessionMembershipData } from "../../../src/matrixrtc/CallMembership";
 import { MatrixRTCSession, MatrixRTCSessionEvent } from "../../../src/matrixrtc/MatrixRTCSession";
 import { type EncryptionKeysEventContent } from "../../../src/matrixrtc/types";
 import { secureRandomString } from "../../../src/randomstring";
-import { makeMockEvent, makeMockRoom, makeMockRoomState, membershipTemplate, makeKey } from "./mocks";
+import { makeMockEvent, makeMockRoom, membershipTemplate, makeKey, type MembershipData, mockRoomState } from "./mocks";
 import { RTCEncryptionManager } from "../../../src/matrixrtc/RTCEncryptionManager.ts";
 
 const mockFocus = { type: "mock" };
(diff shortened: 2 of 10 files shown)


spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
```diff
@@ -174,8 +176,9 @@ describe("RTCEncryptionManager", () => {
             ];
             getMembershipMock.mockReturnValue(members);
 
+            const gracePeriod = 15_000; // 15 seconds
             // initial rollout
-            encryptionManager.join(undefined);
+            encryptionManager.join({ keyRotationGracePeriodMs: gracePeriod });
```
toger5: this we can just move into the join() method?
Since we use the object config the key is written out and the reason for the value is explained

toger5:
```suggestion
            encryptionManager.join({ keyRotationGracePeriodMs: 15_000 });
```

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
```diff
@@ -214,6 +219,157 @@ describe("RTCEncryptionManager", () => {
...
+            ];
+            getMembershipMock.mockReturnValue(members);
+
+            const gracePeriod = 3_000; // 15 seconds
+            const useKeyDelay = 5_000; // 5 seconds
+            // initial rollout
+            encryptionManager.join({
+                useKeyDelay,
+                keyRotationGracePeriodMs: gracePeriod,
+            });
```
toger5:
```suggestion
            // initial rollout
            encryptionManager.join({
                useKeyDelay: 3_000, // 3 seconds
                keyRotationGracePeriodMs: 5_000, // 5 seconds
            });
```

The comment seemed wrong (3 vs 15 s)

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
```diff
@@ -214,6 +219,157 @@ describe("RTCEncryptionManager", () => {
...
+
+            onEncryptionKeysChanged.mockClear();
+            mockTransport.sendKey.mockClear();
+
+            // The existing members have been talking for 5mn
+            await jest.advanceTimersByTimeAsync(5 * 60 * 1000);
+
+            // A new member joins, that should trigger a key rotation.
+            members.push(aCallMembership("@carl:example.org", "CARLDEVICE"));
+            encryptionManager.onMembershipsUpdate([]);
```
toger5: Why can we just pass `[]` as the old memberships here?

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
```diff
@@ -214,6 +219,157 @@ describe("RTCEncryptionManager", () => {
...
+            await jest.runOnlyPendingTimersAsync();
+
+            // A new member joins, within the grace period, but under the delay period
+            members.push(aCallMembership("@david:example.org", "DAVDEVICE"));
+            await jest.advanceTimersByTimeAsync(3_000);
+            encryptionManager.onMembershipsUpdate([]);
```
toger5: As above.
Shouldn't the old memberships be the correct prev memberships, so the test more properly represents a real world scenario?

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
```diff
@@ -214,6 +219,157 @@ describe("RTCEncryptionManager", () => {
...
+            await jest.runOnlyPendingTimersAsync();
+
+            // Wait pass the delay period
+            await jest.advanceTimersByTimeAsync(5_000);
+
+            // Even though the new member joined within the grace period, the key should be rotated because the delay period has passed
```
toger5:
```suggestion
            // Even though the new member joined within the grace period, the key should be rotated because the useKeyDelay has passed
```

I stumbled over it because my brain entangled `period` with grace and useKeyDelay with `delay` so I wasnt immediatly sure what `delay period` relates to.

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
```diff
@@ -214,6 +219,157 @@ describe("RTCEncryptionManager", () => {
             expect(statistics.counters.roomEventEncryptionKeysSent).toBe(2);
         });
 
+        it("test grace period lower than delay period", async () => {
+            jest.useFakeTimers();
```
toger5: Can we get an explanation what we expect.

To me the expected behavior would be:
 - new memeber joins, we stay on the same key (k0) (for useKeyDelay) and share k1
 - another new joined within grace period, we stay on the same key (k1) (for useKeyDelay) and share the same key (k1)
 - another new joiner outside grace period but inside useKeyDelay, we stay on the (k0), we share k2 to all participants because we are outside the grace period.

```suggestion
           // similar list to the above

            jest.useFakeTimers();
```

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
```diff
@@ -214,6 +219,157 @@ describe("RTCEncryptionManager", () => {
...
+            encryptionManager.onMembershipsUpdate([]);
+            await jest.runOnlyPendingTimersAsync();
+
+            // A new member joins, within the grace period, but under the delay period
+            members.push(aCallMembership("@david:example.org", "DAVDEVICE"));
+            await jest.advanceTimersByTimeAsync(3_000);
+            encryptionManager.onMembershipsUpdate([]);
+            await jest.runOnlyPendingTimersAsync();
+
+            // Wait pass the delay period
```
toger5:
```suggestion
            // Wait past the delay period
```

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
toger5: I think I am also confused by the exact content. If the comment describes time `t0` 
 - join within the grace period -> join before 3 s (`t0` < 3)
 - "the delay period has passed" -> would only happen after 5s (`t0` > 5)

Conflicts, so I am misreading it in some way.

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
```diff
@@ -214,6 +219,157 @@ describe("RTCEncryptionManager", () => {
...
+            onEncryptionKeysChanged.mockClear();
+            mockTransport.sendKey.mockClear();
+
+            // The existing members have been talking for 5mn
+            await jest.advanceTimersByTimeAsync(5 * 60 * 1000);
+
+            // A new member joins, that should trigger a key rotation.
+            members.push(aCallMembership("@carl:example.org", "CARLDEVICE"));
+            encryptionManager.onMembershipsUpdate([]);
+            await jest.runOnlyPendingTimersAsync();
```
toger5: This will forward by 5s right? since we will have setup the useKeyDelay timer when adding `carl`?

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
```diff
@@ -214,6 +219,157 @@ describe("RTCEncryptionManager", () => {
...
+
+            // A new member joins, within the grace period, but under the delay period
+            members.push(aCallMembership("@david:example.org", "DAVDEVICE"));
+            await jest.advanceTimersByTimeAsync(3_000);
+            encryptionManager.onMembershipsUpdate([]);
+            await jest.runOnlyPendingTimersAsync();
+
+            // Wait pass the delay period
+            await jest.advanceTimersByTimeAsync(5_000);
```
toger5: This will advance by another 5s?

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
toger5: I am thinking of sth like:
```
const oldMembers = [...members];
members.push(aCallMembership("@carl:example.org", "CARLDEVICE"));
encryptionManager.onMembershipsUpdate(oldMembers);
```

Just to make sure the test env is more correct to the real world. Even if the curren impl also works with `[]`
(also makes the test easier to read if we dont know exactly what the `onMembershipsUpdate` param means)

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
```diff
@@ -214,6 +219,157 @@ describe("RTCEncryptionManager", () => {
...
+            await jest.runOnlyPendingTimersAsync();
+
+            onEncryptionKeysChanged.mockClear();
+            mockTransport.sendKey.mockClear();
+
+            const updatedMembers = [
+                aCallMembership("@bob:example.org", "BOBDEVICE"),
+                aCallMembership("@bob:example.org", "BOBDEVICE2"),
+                aCallMembership("@carl:example.org", "CARLDEVICE"),
+            ];
```
toger5: can we use the same pattern as above?
```
let members = [...]; // mut members array
encryptionManager.onMembershipsUpdate([]);
members = [...]; //update the members
encryptionManager.onMembershipsUpdate([]);
```
so we do not have to call             `getMembershipMock.mockReturnValue(updatedMembers);` twice

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
```diff
@@ -214,6 +219,157 @@ describe("RTCEncryptionManager", () => {
...
+                aCallMembership("@bob:example.org", "BOBDEVICE"),
+                aCallMembership("@bob:example.org", "BOBDEVICE2"),
+            ];
+            getMembershipMock.mockReturnValue(members);
+
+            const gracePeriod = 15_000; // 15 seconds
+            // initial rollout
+            encryptionManager.join({ keyRotationGracePeriodMs: gracePeriod });
+            encryptionManager.onMembershipsUpdate([]);
+            await jest.runOnlyPendingTimersAsync();
```
toger5: I think I do prefer an explicit 15s here as well.
If it is actually setting up the 15s timer already at this time.

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
```diff
@@ -214,6 +219,157 @@ describe("RTCEncryptionManager", () => {
...
+                    expect.objectContaining({ userId: "@carl:example.org", deviceId: "CARLDEVICE" }),
+                ],
+            );
+
+            expect(onEncryptionKeysChanged).toHaveBeenCalled();
+            await jest.advanceTimersByTimeAsync(1000);
+            expect(statistics.counters.roomEventEncryptionKeysSent).toBe(2);
+        });
+
+        it("Should rotate key when several users join within the rotation grace period", async () => {
```
toger5: Shouldn't it NOT rotate keys in this case. It should just redistribute the same key to all the members joined in the grace period?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -153,18 +153,31 @@ export interface EncryptionConfig {
...
      * e.g. if this is set to 1000, then we will send at most one key event every second.
+     * @deprecated - Not used by the new encryption manager.
      */
     updateEncryptionKeyThrottle?: number;
+
+    /**
+     * Sometimes it is necessary to rotate the encryption key after a membership update.
+     * For performance reasons we might not want to rotate the key immediately but allow future memberships to use the same key.
+     * If 5 people join in a row in less than 5 seconds, we don't want to rotate the key for each of them.
+     * If 5 people leave in a row in less than 5 seconds, we don't want to rotate the key for each of them.
```
toger5, line 173:
```suggestion
     * If 5 people leave in a row in less than 5 seconds, we don't want to rotate the key for each of them.
     * So we do share the key which was already used live for <5s to new joiners. This does result in a potential leak up to the configured time of call media.
     * This has to be considered when choosing a value for this property.
```

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -153,18 +153,31 @@ export interface EncryptionConfig {
...
-     * The delay (in milliseconds) between creating and sending a new key and starting to encrypt with it. This
-     * gives other a chance to receive the new key to minimise the chance they don't get media they can't decrypt.
-     * The total time between a member leaving and the call switching to new keys is therefore:
-     * makeKeyDelay + useKeyDelay
+     * The delay (in milliseconds) between sending a new key and starting to encrypt with it. This
+     * gives others a chance to receive the new key to minimize the chance they get media they can't decrypt.
+     *
+     * The higher this value is, the better it is for existing members as they will have a smoother experience.
+     * But it impacts new joiners: They will always have to wait `useKeyDelay` before being able to decrypt the media
+     * (as it will be encrypted with the new key after the delay only), even if the key is already arrived before the delay.
```
toger5:
```suggestion
     * (as it will be encrypted with the new key after the delay only), even if the key has already arrived before the delay.
```

src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -53,12 +53,24 @@ export class RTCEncryptionManager implements IEncryptionManager {
...
      */
     private currentKeyDistributionPromise: Promise<void> | null = null;
+
     /**
      * The time to wait before using the outbound session after it has been distributed.
      * This is to ensure that the key is delivered to all participants before it is used.
      * When creating the first key, this is set to 0, so that the key can be used immediately.
      */
-    private delayRolloutTimeMillis = 1000;
+    private delayRolloutTimeMillis = 5000;
```
toger5: is this the same as `useKeyDelay`. If so we should rename one of them.

src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -53,12 +53,24 @@ export class RTCEncryptionManager implements IEncryptionManager {
...
+
+    /**
+     * We want to avoid rolling out a new outbound key when the previous one was created less than `skipRotationGracePeriod` milliseconds ago.
+     * This is to avoid expensive key rotations when users quickly join the call in a row.
+     *
+     * This must be higher than `delayRolloutTimeMillis` to have an effect.
+     * If it is lower, the current key will always be older than the grace period.
+     * @private
+     */
+    private skipRotationGracePeriod = 10_000;
```
toger5: same here `keyRotationGracePeriodMs`

src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -102,6 +114,7 @@ export class RTCEncryptionManager implements IEncryptionManager {
     public join(joinConfig: EncryptionConfig | undefined): void {
         this.logger?.info(`Joining room`);
         this.delayRolloutTimeMillis = joinConfig?.useKeyDelay ?? 1000;
+        this.skipRotationGracePeriod = joinConfig?.keyRotationGracePeriodMs ?? 10_000;
```
toger5: This should totally be:

```suggestion
this.useKeyDelay = joinConfig?.useKeyDelay ?? 1000;
        this.keyRotationGracePeriodMs = joinConfig?.keyRotationGracePeriodMs ?? 10_000;
```

toger5: Or the other names from this class

toger5 (changes requested):
The implementation looks very simple (In terms of: done in a way that is easy to follow along) and sane. I do have questions on the tests. So I need a bit of help (with comments) to understand what is supposed to happen.

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
BillCarsonFr: I might as well deprecate it? as it is not used

toger5: ah in that case yes!

PR #4912, opened 2025-07-09
EncryptionManager:  un-deprecating EncryptionManager.getEncryptionKeys

BillCarsonFr:
Follow up on https://github.com/matrix-org/matrix-js-sdk/pull/4799

Fixes https://github.com/element-hq/voip-internal/issues/375

After discussions `IEncryptionManager#getEncryptionKeys()` cannot be deprecated.
There is a legit use case for the RTCSession beeing ready and running while the rtc app is not yet ready/started. That means that keys could be arriving already. Hence when the app is ready there should be a way to re-emit / access the keys.

## Checklist

### DIFF
--- a/spec/integ/crypto/cross-signing.spec.ts
+++ b/spec/integ/crypto/cross-signing.spec.ts
@@ -17,9 +17,10 @@ limitations under the License.
 import fetchMock from "fetch-mock-jest";
 import "fake-indexeddb/auto";
 import { IDBFactory } from "fake-indexeddb";
+import debug from "debug";
 
 import { syncPromise } from "../../test-utils/test-utils";
-import { type AuthDict, createClient, type MatrixClient } from "../../../src";
+import { type AuthDict, createClient, DebugLogger, type MatrixClient } from "../../../src";
 import { mockInitialApiRequests, mockSetupCrossSigningRequests } from "../../test-utils/mockEndpoints";
 import encryptAESSecretStorageItem from "../../../src/utils/encryptAESSecretStorageItem.ts";
 import { type CryptoCallbacks, CrossSigningKey } from "../../../src/crypto-api";
@@ -91,6 +92,7 @@ describe("cross-signing", () => {
                 accessToken: "akjgkrgjs",
                 deviceId: TEST_DEVICE_ID,
                 cryptoCallbacks: createCryptoCallbacks(),
+                logger: new DebugLogger(debug(`matrix-js-sdk:cross-signing`)),
             });
 
             syncResponder = new SyncResponder(homeserverUrl);
--- a/spec/integ/crypto/device-dehydration.spec.ts
+++ b/spec/integ/crypto/device-dehydration.spec.ts
@@ -16,8 +16,9 @@ limitations under the License.
 
 import "fake-indexeddb/auto";
 import fetchMock from "fetch-mock-jest";
+import debug from "debug";
 
-import { ClientEvent, createClient, type MatrixClient, MatrixEvent } from "../../../src";
+import { ClientEvent, createClient, DebugLogger, type MatrixClient, MatrixEvent } from "../../../src";
 import { CryptoEvent } from "../../../src/crypto-api/index";
 import { type RustCrypto } from "../../../src/rust-crypto/rust-crypto";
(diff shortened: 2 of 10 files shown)


src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -114,6 +129,7 @@ export class RTCEncryptionManager implements IEncryptionManager {
     public leave(): void {
         this.transport.off(KeyTransportEvents.ReceivedKeys, this.onNewKeyReceived);
         this.transport.stop();
+        this.participantKeyRings.clear();
```
BillCarsonFr, line 131: The legacy encryption manager was only clearing its own key here, but I don't see why?

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
```diff
@@ -25,6 +25,7 @@ import { decodeBase64, TypedEventEmitter } from "../../../src";
 import { RoomAndToDeviceTransport } from "../../../src/matrixrtc/RoomAndToDeviceKeyTransport.ts";
```
robintown, line 25: I just noticed that this file has a typo in the name: missing the P in 'encryption'

src/matrixrtc/EncryptionManager.ts
```diff
@@ -104,8 +99,16 @@ export class EncryptionManager implements IEncryptionManager {
...
+    public getEncryptionKeys(): ReadonlyMap<ParticipantId, ReadonlyArray<{ key: Uint8Array; keyIndex: number }>> {
+        const keysMap = new Map<ParticipantId, ReadonlyArray<{ key: Uint8Array; keyIndex: number }>>();
+        for (const [userId, userKeys] of this.encryptionKeys) {
+            const keys = userKeys.map((entry, index) => ({
+                key: entry.key,
+                keyIndex: index,
+            }));
+            keysMap.set(userId as ParticipantId, keys);
+        }
+        return keysMap as ReadonlyMap<ParticipantId, ReadonlyArray<{ key: Uint8Array; keyIndex: number }>>;
```
robintown:
```suggestion
        return keysMap;
```

Casting is redundant here

src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -94,9 +101,17 @@ export class RTCEncryptionManager implements IEncryptionManager {
         this.logger = parentLogger?.getChild(`[EncryptionManager]`);
     }
 
-    public getEncryptionKeys(): Map<string, Array<{ key: Uint8Array; timestamp: number }>> {
-        // This is deprecated should be ignored. Only used by tests?
-        return new Map();
+    public getEncryptionKeys(): ReadonlyMap<ParticipantId, ReadonlyArray<{ key: Uint8Array; keyIndex: number }>> {
+        const map = new Map(this.participantKeyRings);
+        return map as ReadonlyMap<string, ReadonlyArray<{ key: Uint8Array; keyIndex: number }>>;
```
robintown: Same here

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
```diff
@@ -428,6 +429,94 @@ describe("RTCEncryptionManager", () => {
...
+            ];
+            getMembershipMock.mockReturnValue(members);
+
+            // Let's join
+            encryptionManager.join(undefined);
+            encryptionManager.onMembershipsUpdate(members);
+
+            await jest.advanceTimersByTimeAsync(10);
+
+            // Simulate Carl leaving then joining back, and key received out of order
```
robintown: Need some help understanding: If this test just emits a key at index 0 and then index 5 for Carl, what part of this represents the keys being received "out of order"?

robintown (commented):
I found the title of this PR a little confusing, as the ability to re-emit keys already existed and it hasn't changed substantially. Seems this is primarily about un-deprecating `EncryptionManager.getEncryptionKeys`?

spec/unit/matrixrtc/RTCEncrytionManager.spec.ts
BillCarsonFr: copy/paste error. It's a comment from another tests

PR #4916, opened 2025-07-11
Update to matrix-sdk-crypto-wasm 15.1.0, and add new `ShieldStateCode.MismatchedSender`

richvdh:
matrix-sdk-crypto-wasm 15.1.0 adds a new `DecryptionErrorCode`, which we need to support.

### DIFF
--- a/spec/unit/rust-crypto/rust-crypto.spec.ts
+++ b/spec/unit/rust-crypto/rust-crypto.spec.ts
@@ -1114,6 +1114,7 @@ describe("RustCrypto", () => {
 
         it.each([
             [undefined, undefined, null],
+            ["Other", -1, EventShieldReason.UNKNOWN],
             [
                 "Encrypted by an unverified user.",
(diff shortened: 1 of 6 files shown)


spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -2302,6 +2309,8 @@ describe("RustCrypto", () => {
             });
 
             const rustCrypto = await makeTestRustCrypto(makeMatrixHttpApi(), undefined, undefined, secretStorage);
+            await flushPromises();
```
dbkr: Is it not possible to do this with a waitFor() or something?

andybalaam (approved):

spec/unit/rust-crypto/rust-crypto.spec.ts
richvdh: good point. I reimplemented `waitFor` and used it.

PR #4917, opened 2025-07-11
fetch api: add support for downloading raw response

richvdh:
I need to make an authenticated request to the media repo, and expect to get a binary file back. AFAICT there is no easy way to do that right now.

### DIFF
--- a/spec/unit/http-api/fetch.spec.ts
+++ b/spec/unit/http-api/fetch.spec.ts
@@ -120,7 +120,15 @@ describe("FetchHttpApi", () => {
         await expect(api.requestOtherUrl(Method.Get, "http://url")).resolves.toBe(res);
     });
 
-    it("should return text if json=false", async () => {
+    it("should set an Accept header, and parse the response as JSON, by default", async () => {
+        const result = { a: 1 };
+        const fetchFn = jest.fn().mockResolvedValue({ ok: true, json: jest.fn().mockResolvedValue(result) });
+        const api = new FetchHttpApi(new TypedEventEmitter<any, any>(), { baseUrl, prefix, fetchFn, onlyData: true });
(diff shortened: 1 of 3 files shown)


t3chguy (commented):
If I'm reading the types correctly, could the same not be achieved using onlyData=false json=false and adding a `blob` method to `TypedResponse`?

richvdh:
> If I'm reading the types correctly, could the same not be achieved using onlyData=false json=false and adding a `blob` method to `TypedResponse` given `res` already sports this method it just isn't exposed in the interface.

Yes, but `onlyData` is currently only supported at the level of the entire `FetchHttpApi`, so to do that we'd either have to instantiate a whole other `FetchHttpApi` instance with `onlyData=false` (in which case it would probably be easier just to call `fetch` directly) , or add it to `IHttpOpts` (which doesn't seem inherently better than adding `IHttpOpts.rawResponseBody`).

t3chguy (changes requested):
Okay but `ResponseType` does not consider `rawResponseBody` so the types will only be right if `onlyData=true` or undefined, which does not match your comment which says `rawResponseBody` inhibits the behaviour of `onlyData` - as with `rawResponseBody=true & onlyData=false` (or json=false frankly) you would have a return type of `TypedResponse` (or string in the json=false) case

The types for `rawResponseBody` also do not inhibit you from entering this edge case, where they could/should.

richvdh:
Worth noting that the current declaration of `ResponseType` is completely broken on two counts: first, it treats `json` as overriding `onlyData`, which it doesn't; second, it looks for `IHttpOpts.json` which doesn't exist. I've opened https://github.com/matrix-org/matrix-js-sdk/pull/4926 to fix.

PR #4920, opened 2025-07-13
Experimental support for sharing encrypted history on invite

richvdh:
Add options to `MatrixClient.invite` and `MatrixClient.joinRoom` to share and accept encrypted history on invite, per MSC4268.

There are some extensions to the test infrastructure to allow it to be properly integration tested: it may be easier to review commit-by-commit.

Based on https://github.com/matrix-org/matrix-js-sdk/pull/4919. Also requires:
 * a new version of matrix-sdk-crypto-wasm (and hence github.com/matrix-org/matrix-js-sdk/pull/4916), which includes https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/250.
 * Updates to FetchHttpApi (https://github.com/matrix-org/matrix-js-sdk/pull/4917).

Part of element-hq/element-meta#2871.


src/client.ts
```diff
@@ -2371,7 +2371,12 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      */
     public async joinRoom(roomIdOrAlias: string, opts: IJoinRoomOpts = {}): Promise<Room> {
         const room = this.getRoom(roomIdOrAlias);
-        if (room?.hasMembershipState(this.credentials.userId!, KnownMembership.Join)) return room;
+        const preJoinMembership = room?.getMember(this.getSafeUserId());
```
t3chguy: this is not a membership but a member?

src/client.ts
```diff
@@ -2371,7 +2371,12 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+        const preJoinMembershipSender = preJoinMembership?.events.member?.getSender() ?? null;
+        this.logger.debug(
+            `joinRoom[${roomIdOrAlias}]: preJoinMembership=${preJoinMembership?.membership}, preJoinMembershipSender=${preJoinMembershipSender}, opts=${JSON.stringify(opts)}`,
+        );
+        if (preJoinMembership?.membership == KnownMembership.Join) return room!;
```
t3chguy: I'm confused as to what this block is doing that is different to the old code barring the log line? Surprised to not see it using the `hasMembershipState` helper

src/client.ts
richvdh: ugh yes, this doesn't look right

src/client.ts
richvdh: Yeah, this bit is really just a refactor. The thing it does different is to define `preJoinMembershipSender`, which is used later.

We could use `hasMembershipState`, but we'd need to call it twice (once for Join, once for Invite), and we'd still need to call `getMember` yet again to get the inviter, so it seems better to (effectively) inline it.

src/client.ts
richvdh: have attempted to fix and clarify

t3chguy (approved):

spec/integ/crypto/cross-signing.spec.ts
```diff
@@ -137,9 +137,9 @@ describe("cross-signing", () => {
             const authDict = { type: "test" };
             await bootstrapCrossSigning(authDict);
 
-            // check the cross-signing keys upload
-            expect(fetchMock.called("upload-keys")).toBeTruthy();
-            const [, keysOpts] = fetchMock.lastCall("upload-keys")!;
+            // check the cross-signing keys have been uploaded
```
uhoreg:
```suggestion
            // check that the cross-signing keys have been uploaded
```

I think the other is correct, but adding "that" reads better for me

uhoreg (approved):
lgtm

PR #4896, opened 2025-07-02
MatrixRTC: MembershipManager, add a request timeout to restartDelayedEvent

fkwp:
This PR adds a rather short timeout for the http fetch request of a `restartDelayedEvent`. This is to address situations where packet loss might impact TCP connections and to improve the likelihood of getting a delayed event reset through in a given time constraint. 

## Checklist

- [x] Tests written for new code (and old code if feasible).
- [x] New or updated `public`/`exported` symbols have accurate [TSDoc](https://tsdoc.org/) documentation.
- [x] Linter and other CI checks pass.
- [x] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md)).

PR #4921, opened 2025-07-15
Custom timeout for delayed event restarts

toger5:
(alternative to: https://github.com/matrix-org/matrix-js-sdk/pull/4896)

This PR introduces a new setting `delayedEventRestartLocalTimeoutMS`.

This setting is used in the context of MatrixRTC. We need to restart the dealyed events to make sure
the HomeServer is sending the delayed rtc leave event. In bad network environments we might end up
waiting for too long for the event to arrive and we will not send another restart event until the local timeout is reached.

In those scenarios chances for success are higher if we use a lower local timeout to increase the tries we do instead of waiting
for responses on requests which are stuck.

It also sets the default value for it to: 2300 which works well.

The reason why we make this a global configuration instead of sth we pass through via the `requestOptions` as it is done in: https://github.com/matrix-org/matrix-js-sdk/pull/4896 is because of how Element Call is implemented.

In SPA mode we can control where we send `POST /_matrix/client/v1/delayed_events/{delay_id}`. But in widget mode we cannot configure the exact call to the enpoint. The widget driver will just use the raw function.

Delayed event restarts in general are a very cheap server operation. So a smaller timeout in every case seems appropriate.
## Checklist

### DIFF
--- a/spec/unit/matrix-client.spec.ts
+++ b/spec/unit/matrix-client.spec.ts
@@ -1082,6 +1082,32 @@ describe("MatrixClient", function () {
 
             await client._unstable_updateDelayedEvent(delayId, action);
         });
+
+        it("uses custom timeout for restart delayed event", async () => {
+            // make another client so we can pass creation opts
+            makeClient({ delayedEventRestartLocalTimeoutMS: 2300 });
+            const delayId = "id";
+            const action = UpdateDelayedEventAction.Restart;
+            httpLookups = [
+                {
+                    method: "POST",
+                    prefix: unstableMSC4140Prefix,
+                    path: `/delayed_events/${encodeURIComponent(delayId)}`,
+                    data: {
+                        action,
+                    },
+                },
+            ];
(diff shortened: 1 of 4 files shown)


robintown (commented):
Where should I be looking to find the default of 2300 ms?

And hmm, I'm not quite finding this solution elegant. I do like Florian's solution to cover the Element Call SPA case. For the widget case, have we considered setting 2300 ms as an opinionated default for all widget API calls to this endpoint [in Element Web](https://github.com/element-hq/element-web/blob/5b659fe2e5628ffadebbc64428a4f58d93f88526/src/stores/widgets/StopGapWidgetDriver.ts#L414)?

toger5:
> Where should I be looking to find the default of 2300 ms

It would be part of EC. The PR does not yet exist but would be a one line PR to add the config param.

toger5:
Superseeded by: https://github.com/matrix-org/matrix-js-sdk/pull/4927

PR #4927, opened 2025-07-21
Custom abort timeout logic for restarting delayed events that is compatible with the widget api

toger5:
This adds a `race` so that restart delayed events can be timed out before the js-sdk local timeout happens.

This has to be done this way because the local timeout cannot be passed to the widget driver via the widget api for a specific request.

## Checklist


robintown (approved):
Reviewed in a pairing session.

PR #4936, opened 2025-07-22
Attempt deflaking of `queueToDevice` test

richvdh:
This test randomly failed (https://github.com/matrix-org/matrix-js-sdk/actions/runs/16451549565/job/46497850366), so let's try to do some more waiting to deflake it.


richvdh:
Opened https://github.com/matrix-org/complement-crypto/issues/199 for the complement-crypto failure

PR #4937, opened 2025-07-23
Support for creator power level

dbkr:
Adds support for infinite power level specified by [MSC4289](https://github.com/matrix-org/matrix-spec-proposals/pull/4289).

This version removes powerLevelNorm which we only use in one place in Element Web (where we probably should not) and I'm not sure is generally that sensible, so I've removed it since it's a difficult concept to keep and would at best be a slight lie when one of the PLs is always infinity (eg. we'd have to say creator is also 100, probably). This makes this a BREAKING CHANGE.

I'll leave this as draft until I have a matching PR for element.

## Checklist

### DIFF
--- a/src/models/room-member.ts
+++ b/src/models/room-member.ts
@@ -241,7 +241,9 @@ export class RoomMember extends TypedEventEmitter<RoomMemberEvent, RoomMemberEve
         const oldPowerLevel = this.powerLevel;
 
         const creators = new Set<string>();
-        if (roomVersion === "org.matrix.hydra.11" || (Number.isInteger(roomVersion) && Number(roomVersion) > 12)) {
+        // This checks probably wants to be reversed once verson 12 has been around for longer
+        // (ie. assume v12 semantics except for other known versions)
+        if (roomVersion === "org.matrix.hydra.11" || roomVersion === "12") {
             const roomCreateSender = roomCreateEvent.getSender();
             if (roomCreateSender) creators.add(roomCreateSender);
             const additionalCreators = roomCreateEvent.getDirectionalContent().additional_creators;
--- a/src/models/room-member.ts
+++ b/src/models/room-member.ts
@@ -23,7 +23,6 @@ import { logger } from "../logger.ts";
 import { TypedEventEmitter } from "./typed-event-emitter.ts";
 import { EventType } from "../@types/event.ts";
 import { KnownMembership, type Membership } from "../@types/membership.ts";
-import { roomVersionIsHydra } from "src/utils/roomVersion.ts";
 
 export enum RoomMemberEvent {
     Membership = "RoomMember.membership",
@@ -222,47 +221,16 @@ export class RoomMember extends TypedEventEmitter<RoomMemberEvent, RoomMemberEve
     }
 
     /**
-     * Update this room member's power level event. May fire
-     * "RoomMember.powerLevel" if this event updates this member's power levels.
-     * @param powerLevelEvent - The `m.room.power_levels` event
-     * @param roomCreateEvent - The `m.room.create` event.
-     * @param roomVersion - The room version of the room this member is in.
+     * Update this room member's power level event. Will fire
+     * "RoomMember.powerLevel" if the new power level is different
+     * @param powerLevel - The power level of the room member.
      *
      * @remarks
      * Fires {@link RoomMemberEvent.PowerLevel}
      */
-    public setPowerLevelEvent(powerLevelEvent: MatrixEvent, roomCreateEvent: MatrixEvent, roomVersion: string): void {
-        if (powerLevelEvent.getType() !== EventType.RoomPowerLevels || powerLevelEvent.getStateKey() !== "") {
-            return;
-        }
-
-        const evContent = powerLevelEvent.getDirectionalContent();
-
-        const users: { [userId: string]: number } = evContent.users || {};
+    public setPowerLevel(powerLevel: number, powerLevelEvent: MatrixEvent): void {
         const oldPowerLevel = this.powerLevel;
-
-        const creators = new Set<string>();
-        // This checks probably wants to be reversed once verson 12 has been around for longer
-        // (ie. assume v12 semantics except for other known versions)
-        if (roomVersionIsHydra(roomVersion)) {
-            const roomCreateSender = roomCreateEvent.getSender();
-            if (roomCreateSender) creators.add(roomCreateSender);
-            const additionalCreators = roomCreateEvent.getDirectionalContent().additional_creators;
-            if (Array.isArray(additionalCreators)) additionalCreators.forEach((c) => creators.add(c));
-        }
-
-        if (creators.has(this.userId)) {
-            // As of "Hydra", If the user is a creator, they always have the highest power level
-            this.powerLevel = Infinity;
-        } else {
-            if (users[this.userId] !== undefined && Number.isInteger(users[this.userId])) {
-                this.powerLevel = users[this.userId];
-            } else if (evContent.users_default !== undefined) {
-                this.powerLevel = evContent.users_default;
-            } else {
-                this.powerLevel = 0;
-            }
-        }
+        this.powerLevel = powerLevel;
 
         if (oldPowerLevel !== this.powerLevel) {
             this.updateModifiedTime();
--- a/src/models/room-member.ts
+++ b/src/models/room-member.ts
@@ -247,11 +247,7 @@ export class RoomMember extends TypedEventEmitter<RoomMemberEvent, RoomMemberEve
             const roomCreateSender = roomCreateEvent.getSender();
             if (roomCreateSender) creators.add(roomCreateSender);
             const additionalCreators = roomCreateEvent.getDirectionalContent().additional_creators;
-            if (additionalCreators && typeof additionalCreators === "object") {
-                for (const creator of additionalCreators) {
-                    creators.add(creator);
-                }
-            }
+            if (Array.isArray(additionalCreators)) creators.push(...additionalCreators);
         }
 
         if (creators.has(this.userId)) {
--- a/src/models/room-member.ts
+++ b/src/models/room-member.ts
@@ -247,7 +247,7 @@ export class RoomMember extends TypedEventEmitter<RoomMemberEvent, RoomMemberEve
             const roomCreateSender = roomCreateEvent.getSender();
             if (roomCreateSender) creators.add(roomCreateSender);
             const additionalCreators = roomCreateEvent.getDirectionalContent().additional_creators;
-            if (Array.isArray(additionalCreators)) creators.push(...additionalCreators);
+            if (Array.isArray(additionalCreators)) additionalCreators.forEach((c) => creators.add(c));
         }
 
         if (creators.has(this.userId)) {
--- a/src/models/room-member.ts
+++ b/src/models/room-member.ts
@@ -23,6 +23,7 @@ import { logger } from "../logger.ts";
 import { TypedEventEmitter } from "./typed-event-emitter.ts";
 import { EventType } from "../@types/event.ts";
 import { KnownMembership, type Membership } from "../@types/membership.ts";
+import { roomVersionIsHydra } from "src/utils/roomVersion.ts";
 
 export enum RoomMemberEvent {
     Membership = "RoomMember.membership",
@@ -243,7 +244,7 @@ export class RoomMember extends TypedEventEmitter<RoomMemberEvent, RoomMemberEve
         const creators = new Set<string>();
         // This checks probably wants to be reversed once verson 12 has been around for longer
         // (ie. assume v12 semantics except for other known versions)
-        if (roomVersion === "org.matrix.hydra.11" || roomVersion === "12") {
+        if (roomVersionIsHydra(roomVersion)) {
             const roomCreateSender = roomCreateEvent.getSender();
             if (roomCreateSender) creators.add(roomCreateSender);
             const additionalCreators = roomCreateEvent.getDirectionalContent().additional_creators;
--- a/src/models/room-member.ts
+++ b/src/models/room-member.ts
@@ -66,7 +66,6 @@ export type RoomMemberEventHandlerMap = {
      * ```
      * matrixClient.on("RoomMember.powerLevel", function(event, member){
      *   var newPowerLevel = member.powerLevel;
-     *   var newNormPowerLevel = member.powerLevelNorm;
      * });
      * ```
      */
@@ -109,10 +108,6 @@ export class RoomMember extends TypedEventEmitter<RoomMemberEvent, RoomMemberEve
      * The power level for this room member.
      */
     public powerLevel = 0;
-    /**
-     * The normalised power level (0-100) for this room member.
-     */
-    public powerLevelNorm = 0;
     /**
      * The User object for this room member, if one exists.
      */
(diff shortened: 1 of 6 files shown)


src/models/room-member.ts
```diff
@@ -229,40 +224,48 @@ export class RoomMember extends TypedEventEmitter<RoomMemberEvent, RoomMemberEve
...
-        const oldPowerLevelNorm = this.powerLevelNorm;
 
-        if (users[this.userId] !== undefined && Number.isInteger(users[this.userId])) {
-            this.powerLevel = users[this.userId];
-        } else if (evContent.users_default !== undefined) {
-            this.powerLevel = evContent.users_default;
-        } else {
-            this.powerLevel = 0;
+        const creators = new Set<string>();
+        if (roomVersion === "org.matrix.hydra.11" || (Number.isInteger(roomVersion) && Number(roomVersion) > 12)) {
```
t3chguy: This is invalid, as per the spec https://spec.matrix.org/v1.15/rooms/

> There is no implicit ordering or hierarchy to room versions

See also https://github.com/element-hq/element-meta/issues/2921#issuecomment-3107673036

t3chguy (changes requested):

src/models/room-member.ts
dbkr: Ah, right you are. Fixed.

t3chguy:
@dbkr did you mean to request review? it is still in draft

src/models/room-member.ts
```diff
@@ -229,40 +224,50 @@ export class RoomMember extends TypedEventEmitter<RoomMemberEvent, RoomMemberEve
...
+        // (ie. assume v12 semantics except for other known versions)
+        if (roomVersion === "org.matrix.hydra.11" || roomVersion === "12") {
+            const roomCreateSender = roomCreateEvent.getSender();
+            if (roomCreateSender) creators.add(roomCreateSender);
+            const additionalCreators = roomCreateEvent.getDirectionalContent().additional_creators;
+            if (additionalCreators && typeof additionalCreators === "object") {
+                for (const creator of additionalCreators) {
+                    creators.add(creator);
+                }
+            }
```
MidhunSureshR:
```suggestion
            if (Array.isArray(additionalCreators)) creators.push(...additionalCreators);
```

src/models/room-state.ts
```diff
@@ -209,6 +212,22 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
         this.updateModifiedTime();
     }
 
+    /**
+     * Gets the version of the room
+     * @returns The version of the room, or null if it could not be determined
```
MidhunSureshR:
```suggestion
     * @returns The version of the room
```

MidhunSureshR (approved):
Looks great, just two minor things!

src/models/room-member.ts
```diff
@@ -229,40 +224,50 @@ export class RoomMember extends TypedEventEmitter<RoomMemberEvent, RoomMemberEve
...
      * "RoomMember.powerLevel" if this event updates this member's power levels.
      * @param powerLevelEvent - The `m.room.power_levels` event
+     * @param roomCreateEvent - The `m.room.create` event.
+     * @param roomVersion - The room version of the room this member is in.
      *
      * @remarks
      * Fires {@link RoomMemberEvent.PowerLevel}
      */
-    public setPowerLevelEvent(powerLevelEvent: MatrixEvent): void {
+    public setPowerLevelEvent(powerLevelEvent: MatrixEvent, roomCreateEvent: MatrixEvent, roomVersion: string): void {
```
t3chguy: Having to pass the `roomVersion` & `roomCreateEvent` feels like a really odd signature, IMO RoomMember should know its room, it should probably be a c'tor param it holds onto then it can get these itself.

src/models/room-member.ts
```diff
@@ -229,40 +224,50 @@ export class RoomMember extends TypedEventEmitter<RoomMemberEvent, RoomMemberEve
...
-        if (users[this.userId] !== undefined && Number.isInteger(users[this.userId])) {
-            this.powerLevel = users[this.userId];
-        } else if (evContent.users_default !== undefined) {
-            this.powerLevel = evContent.users_default;
-        } else {
-            this.powerLevel = 0;
+        const creators = new Set<string>();
+        // This checks probably wants to be reversed once verson 12 has been around for longer
+        // (ie. assume v12 semantics except for other known versions)
+        if (roomVersion === "org.matrix.hydra.11" || roomVersion === "12") {
```
t3chguy: Given this check is in multiple places, can we abstract it out?

src/models/room-member.ts
dbkr: This will be a pretty significant change which will also need RoomState to know about the room it's for as it has to construct RoomMembers. The thinking here was that you pass all the inputs into the function rather than it taking some from the state which hopefully makes it easier to reason about and less magic, plus just wanting to keep the class as pure as possible and not knowing anything about the context it exists in, only what it represents itself. Admittedly it does make the function signature a little odd: I was going to rename it to something that reflected it taking all the information that affected power levels but I didn't feel like it helped a great deal.

t3chguy: I feel like its a massive footgun like `setPowerLevel` used to be on MatrixClient back when you needed to pass in the existing power level event. We had a bunch of issues about people holding it wrong. I think the RoomState holding the Room object is very sane.

t3chguy: Alternatively the logic could be hoisted into RoomState where it lives in a private method and RoomMember just blindly trusts the powerLevel you give it, rather than needing the whole event. Either way its a breaking change.

src/models/room-member.ts
dbkr: Afaics it was just this one place, but sure.

t3chguy: I think I conflated initially seeing this PR 2 weeks ago as the other place

src/models/room-member.ts
dbkr: Well, this is all breaking change unfortunately: hydra sort of necessitates it. I'm not sure people holding it wrong is really a concern here? This isn't really public facing API.

I'm very reluctant to block our hydra changes on a significant refactor of a part of js-sdk's data model: i think this would add significant risk to our hydra deadline. I can look at moving it RoomState.

t3chguy: It appears as public in the API docs https://matrix-org.github.io/matrix-js-sdk/classes/matrix.RoomMember.html#setpowerlevelevent

src/models/room.ts
```diff
@@ -609,15 +608,7 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
      * @returns The version of the room, or null if it could not be determined
```
richvdh: doesn't seem to return null? Fix this while you are here?

src/utils/roomVersion.ts
```diff
@@ -0,0 +1,19 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+export function roomVersionIsHydra(roomVersion: string): boolean {
+    return roomVersion === "org.matrix.hydra.11" || roomVersion === "12";
+}
```
richvdh: I guess this is in `src/utils` because you want it to be part of the public API. In which case, it needs docs.

(I mean, all functions should have docs IMHO. But ones that are in the public API *definitely* need docs.)

dbkr: Not really, it just seemed like a logical place to put it (I will add docs for it anyway though).

src/utils/roomVersion.ts
```diff
@@ -0,0 +1,34 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * Checks if the given room version is a version that is known to use
+ * new "hydra" power level semantics (as of room version 12)
```
richvdh:
```suggestion
 * new "hydra" power level semantics (i.e. room version 12 or later)
```

src/utils/roomVersion.ts
```diff
@@ -0,0 +1,34 @@
...
+ * (see https://github.com/matrix-org/matrix-spec-proposals/pull/4289).
+ * Note that this will only return true for versions that are known to
+ * the js-sdk': any room version created subsequently will need to be added here
+ * if it uses hydra semantics.
+ *
+ * Once hydra has been rolled out in production for "long enough", it would probably
+ * makes sense to assume hydra semantiocs by default and hence check instead for room
+ * versions known to use the old semantics.
```
richvdh: These seem like implementation notes, rather than something an application developer needs to read?

(Or at least, if an app developer needs to read them, it needs rephrasing to make it clear what they are meant to take away from the text?)

Also, "semantiocs"

src/utils/roomVersion.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @param roomVersion - The version of the room to check.
+ * @returns true if the room version is known to use hydra semantics, false otherwise.
```
richvdh:
```suggestion
 * @returns `true` if the room version is known to use hydra semantics, `false` otherwise.
```

src/utils/roomVersion.ts
dbkr: My assumption was we'd write a new function, eg. `roomVersionIsNotHydra` for maximum clarity, so in that respect it will affect the caller because they'll have to switch to using a different function.

Alternatively we encapsulate this decision in this function and call it, `shouldUseHydraForRoomVersion` or something.

Have updated the docs assuming the former option anyway.

src/utils/roomVersion.ts
```diff
@@ -0,0 +1,35 @@
...
+ * new "hydra" power level semantics (ie. room version 12 or later)
+ * (see https://github.com/matrix-org/matrix-spec-proposals/pull/4289).
+ * Note that this will only return true for versions that are known to
+ * the js-sdk': any room version created subsequently will need to be added here
+ * if it uses hydra semantics.
+ *
+ * Once hydra has been rolled out in production for "long enough", it would probably
+ * makes sense to assume hydra semantics by default and hence check instead for room
+ * versions known to use the old semantics using a different function (which does not exist
+ * currently).
```
richvdh: Sorry, but this still feels like the target audience is a future js-sdk maintainer, not an application developer.

"will need to be added here"

"it would probably makes [sic] sense to"

```suggestion
 *
 * Note that this will only return `true` for versions that are known to
 * the js-sdk.
 *
 * In future, once hydra has been in production for some time, this function will likely
 * be updated to reverse the logic, and return `true` for all room versions except those
 * known to use the old semantics.
```

... or "this function will be removed in favour of another that uses the opposite semantics", or something.

richvdh: Also, isn't this a massive problem? As soon as room v13 is released, all existing clients are going to break. Frankly, given the likelihood of us going back on it, I'd have thought it would make sense to reverse the logic *now* and return true unless the room version is in the range v1 - v11.

src/utils/roomVersion.ts
```diff
@@ -0,0 +1,35 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * Checks if the given room version is a version that is known to use
+ * new "hydra" power level semantics (ie. room version 12 or later)
+ * (see https://github.com/matrix-org/matrix-spec-proposals/pull/4289).
```
richvdh:
```suggestion
 * Checks if the given room version is a version that is known to use
 * new "hydra" power level semantics (i.e. room version 12 or later),
 * in which the creator(s) of the room have a high implicit power level
 * (see https://github.com/matrix-org/matrix-spec-proposals/pull/4289).
```

src/utils/roomVersion.ts
dbkr: Possibly, yeah - this would mean we break any room versions we don't know about (eg. experimental ones) as we'll start treating them as hydra, although this will happen at some point anyway so arguably it may as well happen now. In the interest of trying to move faster, I've changed it to do this to see what folks think.

dbkr: For the record, legacy mobile and EX both use hydra semantics for ONLY the unstable hyrda version and v12, so the problem you raise will apply to those and we'd be the odd-one-out here.

dbkr:
@t3chguy @richvdh sorry to prod but we really need to get *something* merged soon.

src/utils/roomVersion.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Room versions strings that we know about and do not use hydra semantics.
+ */
+const PRE_HYDRA_ROOM_VERSIONS = ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11"];
+
+/**
+ * Checks if the given room version is one where new "hydra" power level
+ * semantics (ie. room version 12 or later) should be used
+ * (see https://github.com/matrix-org/matrix-spec-proposals/pull/4289).
+ * This will return false for versions that are known to the js-sdk' and
```
richvdh:
```suggestion
 * This will return `false` for versions that are known to the js-sdk and
```

richvdh (approved):

src/utils/roomVersion.ts
```diff
@@ -0,0 +1,35 @@
...
+ * This will return `false` for versions that are known to the js-sdk and
+ * do not use hydra: any room versions unknown to the js-sdk (experimental or
+ * otherwise) will cause the function to return true.
+ *
+ * @param roomVersion - The version of the room to check.
+ * @returns `true` if hydra semantics should be used for the room version, `false` otherwise.
+ */
+export function shouldUseHydraForRoomVersion(roomVersion: string): boolean {
+    return !PRE_HYDRA_ROOM_VERSIONS.includes(roomVersion);
+}
```
t3chguy: As per https://github.com/element-hq/element-meta/issues/2921 and prior discussions, we should be consistent between the clients. This approach also means a bunch of unstable (MSC) room versions will get wrongly tagged as hydra-enabled when they are not.

t3chguy (changes requested):

src/utils/roomVersion.ts
t3chguy: Annoyingly, this would have been handled by the room capabilities stuff which failed MSC. https://github.com/matrix-org/matrix-spec-proposals/pull/3244 - instead now both clients and servers need to hardcode a list of room versions and what they support 🙃

dbkr: Agreed, it is probably better to be consistent: have kept the function externally the same but changed the check to only consider the hydra versions.

t3chguy (approved):

src/utils/roomVersion.ts
richvdh: For the record, I disagree with this decision, but at this point it's more important to land something and make progress. I've set out my arguments at https://github.com/element-hq/element-meta/issues/2921#issuecomment-3167476763.

PR #4942, opened 2025-07-25
MatrixRTC: comply with the `manageMediaKeys` EncryptionConfig option

BillCarsonFr:
The new `RTCEncryptionManager` was not complying to the `EncryptionConfig.manageMediaKeys`.
The result is that in clear rooms, keys were sent for nothing, generating un-needed traffic.

## Checklist

### DIFF
--- a/src/matrixrtc/RTCEncryptionManager.ts
+++ b/src/matrixrtc/RTCEncryptionManager.ts
@@ -46,6 +46,8 @@ import {
  * XXX In the future we want to distribute a ratcheted key not the current one for new joiners.
  */
 export class RTCEncryptionManager implements IEncryptionManager {
+    private manageMediaKeys = false;
+
     /**
      * Store the key rings for each participant.
      * The encryption manager stores the keys because the application layer might not be ready yet to handle the keys.
@@ -126,6 +128,8 @@ export class RTCEncryptionManager implements IEncryptionManager {
     }
 
     public join(joinConfig: EncryptionConfig | undefined): void {
+        this.manageMediaKeys = joinConfig?.manageMediaKeys ?? true; // default to true
+
         this.logger?.info(`Joining room`);
         this.useKeyDelay = joinConfig?.useKeyDelay ?? 1000;
         this.keyRotationGracePeriodMs = joinConfig?.keyRotationGracePeriodMs ?? 10_000;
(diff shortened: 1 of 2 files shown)


src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -174,6 +178,7 @@ export class RTCEncryptionManager implements IEncryptionManager {
      * the calls will be coalesced to a single new distribution (that will start just after the current one has completed).
      */
     private ensureKeyDistribution(): void {
+        if (!this.manageMediaKeys) return;
```
toger5, line 187: Could we just never start the encryption manager? What is it doing if we do not use manage Media keys?
Maybe we can even not touch this file at all and just check for manageMediaKeys before calling `RTCEncryptionManager.join`

BillCarsonFr: There is no start/stop on the encryption manager.
I suppose you mean start/stop the transport?
Anyhow, I thought about not having an EncryptionManager at all when we don't want to manage keys (a `NoOpEncryptionMgt`, but the option to manage key is in the join config.

We could also stop the transport but, I figure out it woud be best to log when we receive an unexpected key?

All in all, I think it is alright for now.
Might review that when we will remove the old encryption mgr.

toger5: This function:

```

        this.manageMediaKeys = joinConfig?.manageMediaKeys ?? true; // default to true

        this.logger?.info(`Joining room`);
        this.useKeyDelay = joinConfig?.useKeyDelay ?? 1000;
        this.keyRotationGracePeriodMs = joinConfig?.keyRotationGracePeriodMs ?? 10_000;
        this.transport.on(KeyTransportEvents.ReceivedKeys, this.onNewKeyReceived);
        // Deprecate RoomKeyTransport: this can get removed.
        if (this.transport instanceof RoomAndToDeviceTransport) {
            this.transport.on(RoomAndToDeviceEvents.EnabledTransportsChanged, this.onTransportChanged);
        }

        this.transport.start();
    }
   
```
    so yes the transport start in the encryption manager join

robintown (commented):
Reviewed this and it looks fine to me, but I'm split on whether it's better to have the warnings or to disable the encryption manager entirely. So I defer to your and Timo's judgement.

robintown (approved):

PR #4944, opened 2025-07-30
Fix possible unknown state (Test that reproduces the edgecase)

toger5:
Fixes: https://github.com/element-hq/element-call/issues/3424
This PR adds a test for a edgecase where the membership manager can reach an unknown state.

The edge case is as following:
 - restarting the delayed leave event gives a 404 (not found, the delayed event got sent by the HS)
 - We have NOT yet received that our state is missing from the sync
 - We are scheduling a new SendDelayedEvent action so we can setup a new delayed state event
 - Just before we can sent out the request we receive our missing membership over sync. Because we arleady have a SendDelayedEvent scheduled we do not need to do any special logic here. No new actions are scheduled even though our membership is missing. (which is correct)
 - When finally sending the new State event we have not cleared the UpdateExpiry action -> we will insert another one (this is the behaviour that is fixed in this PR, we will never insert another one if there is one scheduled already)

The test before this change:
Screenshot 2025-07-30 at 14 47 18

## Checklist

### DIFF
--- a/spec/unit/matrixrtc/MembershipManager.spec.ts
+++ b/spec/unit/matrixrtc/MembershipManager.spec.ts
@@ -511,6 +511,32 @@ describe("MembershipManager", () => {
 
             expect(client._unstable_updateDelayedEvent).toHaveBeenCalled();
         });
+
+        it("updates the UpdateExpiry entry in the action scheduler", async () => {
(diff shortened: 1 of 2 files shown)


robintown (approved):
Makes sense to the extent that I understand the state machine here.

PR #4945, opened 2025-07-31
Allow multiple rtc sessions per room (with different sessionDescriptions)

toger5:
Before this was just hardcoded to a call session

## Checklist

### DIFF
--- a/src/matrixrtc/CallMembership.ts
+++ b/src/matrixrtc/CallMembership.ts
@@ -18,6 +18,7 @@ import { type MatrixEvent } from "../matrix.ts";
 import { deepCompare } from "../utils.ts";
 import { type Focus } from "./focus.ts";
 import { isLivekitFocusActive } from "./LivekitFocus.ts";
+import { type SessionDescription } from "./MatrixRTCSession.ts";
 
 /**
  * The default duration in milliseconds that a membership is considered valid for.
@@ -130,6 +131,7 @@ export class CallMembership {
         return this.parentEvent.getId();
     }
 
+    /// Deprecated id field. Use sessionDescription.id instead.
     public get callId(): string {
         return this.membershipData.call_id;
     }
@@ -138,6 +140,12 @@ export class CallMembership {
         return this.membershipData.device_id;
     }
 
+    public get sessionDescription(): SessionDescription {
+        return {
+            application: this.membershipData.application,
+            id: this.membershipData.call_id,
+        };
+    }
     public get application(): string | undefined {
         return this.membershipData.application;
     }
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -74,6 +74,12 @@ export interface SessionConfig {
     notificationType?: RTCNotificationType;
 }
 
+/// The session description is used to identify a session. Used in the state event.
+export interface SessionDescription {
+    id: string;
+    application: string;
+}
+
 // The names follow these principles:
 // - we use the technical term delay if the option is related to delayed events.
 // - we use delayedLeaveEvent if the option is related to the delayed leave event.
@@ -86,7 +92,7 @@ export interface MembershipConfig {
      * Use the new Manager.
      *
      * Default: `false`.
-     * @deprecated does nothing anymore we always default to the new memberhip manager.
+     * @deprecated does nothing anymore we always default to the new membership manager.
      */
     useNewMembershipManager?: boolean;
 
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -453,6 +453,7 @@ export class MatrixRTCSession extends TypedEventEmitter<
                 this.roomSubset,
                 this.client,
                 () => this.getOldestMembership(),
+                this.sessionDescription,
                 this.logger,
             );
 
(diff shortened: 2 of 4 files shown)


toger5:
> I was a bit surprised that there is one MatrixRTCSessionManager per session description (I thought there will be one global for all sorts of rtc sessions, but probably fine)

A global one was the plan initially. But this would require changing the data structure for the session map.
Since we are only interested in call session in the js-sdk this seems like the better first step.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -329,12 +347,18 @@ export class MatrixRTCSession extends TypedEventEmitter<
     }
 
     /**
-     * Return the MatrixRTC session for the room, whether there are currently active members or not
+     * Return the MatrixRTC session for the room.
+     * This returned session can be used to find out if there are currently active
```
robintown: Words are missing after 'active'

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -379,10 +403,15 @@ export class MatrixRTCSession extends TypedEventEmitter<
             "getLiveTimeline" | "roomId" | "getVersion" | "hasMembershipState" | "on" | "off"
         >,
         public memberships: CallMembership[],
+        /**
+         * The session description is used to define the exact session this object is tracking.
+         * A session is unique to another session if one of those properties differ: `roomSubset.roomId`, `sessionDescription.application`, `sessionDescription.id`.
```
robintown:
```suggestion
         * A session is distinct from another session if one of those properties differ: `roomSubset.roomId`, `sessionDescription.application`, `sessionDescription.id`.
```

src/matrixrtc/MembershipManager.ts
```diff
@@ -700,7 +701,7 @@ export class MembershipManager
 
     // HELPERS
     private makeMembershipStateKey(localUserId: string, localDeviceId: string): string {
-        const stateKey = `${localUserId}_${localDeviceId}`;
+        const stateKey = `${localUserId}_${localDeviceId}_${this.sessionDescription.application + this.sessionDescription.id}`;
```
robintown: Should we be escaping underscores in the `application` segment of the string, or is it intentional that an underscore in this string format may or may not be a separator depending on its position?

src/matrixrtc/CallMembership.ts
```diff
@@ -138,6 +142,13 @@ export class CallMembership {
         return this.membershipData.device_id;
     }
 
+    public get sessionDescription(): SessionDescription {
+        return {
+            application: this.membershipData.application,
+            id: this.membershipData.call_id,
```
robintown, line 148: Could we perhaps already support the `session.id` format here, so when we want to resolve the `// TODO: use the new format for m.rtc.member events where call_id becomes session.id` comment in MembershipManager.ts, we already have versions of matrix-js-sdk rolled out that are forward-compatible with this change?

spec/unit/matrixrtc/MatrixRTCSessionManager.spec.ts
```diff
@@ -64,4 +79,22 @@ describe("MatrixRTCSessionManager", () => {
...
+        const room1 = makeMockRoom([{ ...membershipTemplate, application: "m.other_app" }]);
+        jest.spyOn(client, "getRooms").mockReturnValue([room1]);
+        jest.spyOn(client, "getRoom").mockReturnValue(room1);
+
+        client.emit(ClientEvent.Room, room1);
+
+        mockRoomState(room1, [{ user_id: membershipTemplate.user_id }]);
+
+        const roomState = room1.getLiveTimeline().getState(EventTimeline.FORWARDS)!;
+        const membEvent = roomState.getStateEvents("")[0];
```
robintown: What's up with the empty event type here?

src/matrixrtc/MembershipManager.ts
toger5: The latter. The only relevant underscore is after the userId which does have a defined format. The first underscore after the : will be the one where the member_id starts. That the member id is `${localDeviceId}_${this.sessionDescription.application + this.sessionDescription.id}` is not really important. It could also be a hashed version of those fields.

src/matrixrtc/CallMembership.ts
toger5: I was planning to move the change with the session.id into the new namspace which uses the correct msc prefix + m.rtc.member (instead of m.call.member)

In that case it will be a compleatly new event type where we can do specific body parsing for `session.id`.

spec/unit/matrixrtc/MatrixRTCSessionManager.spec.ts
toger5: Its taken from the test above (23month old).
But let me think this through again. Now that you point it out it looks odd to me too.

src/matrixrtc/CallMembership.ts
robintown: Okay, sounds good.

src/matrixrtc/MembershipManager.ts
robintown:
```suggestion
        const stateKey = `${localUserId}_${localDeviceId}_${this.sessionDescription.application}${this.sessionDescription.id}`;
```

(nitpick)

spec/unit/matrixrtc/MatrixRTCSessionManager.spec.ts
toger5: Ah its the mock impl that allows to do this. without state key it will just return all state events. (not type filtering...)

robintown: Got it, could you comment?

PR #4962, opened 2025-08-18
Add `probablyLeft` event to the MatrixRTCSession

toger5:
Fixes: https://github.com/element-hq/element-call/issues/3454
This event is emitted when the membership manager predicts, that we have left the session. It can be in clients owning the RTCSession to stop publishing media streams until we reconnected to the server. (emit `probablyLeft = false` with value 

## Checklist

### DIFF
--- a/src/matrixrtc/IMembershipManager.ts
+++ b/src/matrixrtc/IMembershipManager.ts
@@ -17,7 +17,7 @@ limitations under the License.
 import type { CallMembership } from "./CallMembership.ts";
 import type { Focus } from "./focus.ts";
 import type { Status } from "./types.ts";
-import { type TypedEventEmitter } from "src/matrix.ts";
+import { type TypedEventEmitter } from "../models/typed-event-emitter.ts";
 
 export enum MembershipManagerEvent {
     StatusChanged = "StatusChanged",


src/matrixrtc/MembershipManager.ts
```diff
@@ -544,27 +552,46 @@ export class MembershipManager
...
-            setTimeout(() => {
-                reject(new AbortError("Restart delayed event timed out before the HS responded"));
-            }, this.delayedLeaveEventRestartLocalTimeoutMs);
+            setTimeout(
+                () => {
+                    reject(new AbortError("Restart delayed event timed out before the HS responded"));
+                },
+                Math.min(
+                    this.delayedLeaveEventRestartLocalTimeoutMs,
+                    (this.state.expectedServerDelayLeaveTs ?? Number.MAX_SAFE_INTEGER) - Date.now(),
```
robintown: I had to scan up and down this method for a bit to figure out that this timeout adjustment here is responsible for making sure that the `this.state.expectedServerDelayLeaveTs < Date.now()` condition in the `.catch` block will never be "missed" due to a previous restart request throwing an error too soon. Maybe it's worth a comment for future readers:

```suggestion
                    // Ensure that the .catch() block below runs as soon as we might want to set 'probablyLeft' to true
                    (this.state.expectedServerDelayLeaveTs ?? Number.POSITIVE_INFINITY) - Date.now(),
```

(I also found this slightly clearer to reason about with `Number.POSITIVE_INFINITY`)

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -212,8 +216,8 @@ export type JoinSessionConfig = SessionConfig & MembershipConfig & EncryptionCon
  * This class doesn't deal with media at all, just membership & properties of a session.
  */
 export class MatrixRTCSession extends TypedEventEmitter<
-    MatrixRTCSessionEvent | RoomAndToDeviceEvents,
-    MatrixRTCSessionEventHandlerMap & RoomAndToDeviceEventsHandlerMap
+    MatrixRTCSessionEvent | RoomAndToDeviceEvents | MembershipManagerEvent,
+    MatrixRTCSessionEventHandlerMap & RoomAndToDeviceEventsHandlerMap & MembershipManagerEventHandlerMap
```
robintown: I suggest we `Pick` specific entries out of the `MembershipManagerEventHandlerMap` so we don't mislead consumers into thinking that a MatrixRTCSession re-emits all the membership manager events.

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -822,6 +823,45 @@ describe("MembershipManager", () => {
...
+
+            // Let the scheduler run one iteration so that we can send the join state event
+            await waitForMockCall(client._unstable_updateDelayedEvent);
+
+            client._unstable_updateDelayedEvent = jest.fn(() => {
+                return new Promise((resolve) => {
+                    // We never resolve the delayed event so that we can test the probablyLeft event.
+                    // This simulates the case where the server does not respond to the delayed event.
+                });
+            });
```
robintown: Jest doesn't really like there to be promises that never resolve after the completion of the test - can we create a single promise and reject it at the end?

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -822,6 +823,45 @@ describe("MembershipManager", () => {
...
+            expect(probablyLeftEmit).not.toHaveBeenCalledWith(true);
+            expect(client._unstable_updateDelayedEvent).toHaveBeenCalledTimes(1);
+
+            // no emission after 10s
+            await jest.advanceTimersByTimeAsync(5000);
+            expect(client._unstable_updateDelayedEvent).toHaveBeenCalledTimes(4);
+
+            expect(probablyLeftEmit).not.toHaveBeenCalledWith(true);
+
+            // but just an instance later.
```
robintown:
```suggestion
            // but just an instant later.
```

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -822,6 +823,45 @@ describe("MembershipManager", () => {
             expect(client.sendStateEvent).toHaveBeenCalled();
         });
     });
+    describe("probablyLeft", () => {
+        it("emits probablyLeft when the membership manager could not hear back from the server for the duration of the delayed event", async () => {
```
robintown, line 848: For extra points: test that it resets to `false` eventually when connectivity returns.

src/matrixrtc/MembershipManager.ts
```diff
@@ -589,7 +599,9 @@ export class MembershipManager
                 );
             })
             .catch((e) => {
-                if (this.state.expectedServerDelayLeaveTs && this.state.expectedServerDelayLeaveTs < Date.now()) {
+                if (this.state.expectedServerDelayLeaveTs && this.state.expectedServerDelayLeaveTs <= Date.now()) {
+                    // Once we reach this point we know that the server did send the delayed leave event so we emit `probablyLeft = true`.
```
robintown:
```suggestion
                    // Once we reach this point it's likely that the server is sending the delayed leave event so we emit `probablyLeft = true`.
```

(It's not known for sure)

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -826,40 +833,48 @@ describe("MembershipManager", () => {
...
             manager.on(MembershipManagerEvent.ProbablyLeft, probablyLeftEmit);
             manager.join([focus], focusActive);
 
             // Let the scheduler run one iteration so that we can send the join state event
             await waitForMockCall(client._unstable_updateDelayedEvent);
 
-            client._unstable_updateDelayedEvent = jest.fn(() => {
-                return new Promise((resolve) => {
-                    // We never resolve the delayed event so that we can test the probablyLeft event.
-                    // This simulates the case where the server does not respond to the delayed event.
```
robintown: This comment was nice.

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -826,40 +833,48 @@ describe("MembershipManager", () => {
...
+            // Mock a sync which does not include our own membership
+            await manager.onRTCSessionMemberUpdate([]);
+            // Wait for the current ongoing delayed event sending to finish
+            await jest.advanceTimersByTimeAsync(1);
+            // We should send a new state event and an associated delayed leave event.
+            expect(client._unstable_sendDelayedStateEvent).toHaveBeenCalledTimes(2);
+            expect(client.sendStateEvent).toHaveBeenCalledTimes(2);
+            // At the same time we expect the probablyLeft event to be emitted with false so we are back operational.
+            expect(probablyLeftEmit).toHaveBeenCalledWith(false);
+            rejectStuckPromise();
```
robintown: Even better would be to put this in a `finally` block so the test doesn't necessarily wedge things if it fails either.

robintown (commented):
I still want to finish checking that this works for [the Element Call change](https://github.com/element-hq/element-call/pull/3453) before I approve, but this looks quite ready.

robintown (approved):
Working in Element Call as well

PR #4974, opened 2025-08-20
Expose the StatusChanged event through the RTCSession

toger5:

## Checklist


toger5:
I added:
```
    public get membershipManagerStatus(): Status | undefined {
        return this.membershipManager?.status;
    }
```
I feel like for completeness we should then also expose the probably left field but on the other hand the status is not just intersting in reconnecting state but in general. So maybe just exposing the status makes sense.

robintown (commented):
Yeah, technically we have no way to determine the correct value of probablyLeft at the point of subscribing either, and a getter would be nice. With that one it's easier to get away with assuming that it's always `false` at first, I guess, but who knows whether some consumer might want to start caring about the value only in the middle of a session.

One more thing on the name: could we call the getter something like `membershipStatus` or just `status`? I've always thought of the "membership manager" as being an internal implementation detail of MatrixRTCSession.

PR #4978, opened 2025-08-26
MatrixRTC: Add RTC decline event

toger5:
See: https://github.com/matrix-org/matrix-spec-proposals/pull/4310

## Checklist

### DIFF
--- a/src/matrixrtc/types.ts
+++ b/src/matrixrtc/types.ts
@@ -104,6 +104,15 @@ export interface IRTCNotificationContent extends RelationEvent {
     "lifetime": number;
 }
 
+/**
+ * MSC4310 decline event content for `org.matrix.msc4310.rtc.decline`.
+ * Sent as a standard m.reference relation to an `org.matrix.msc4075.rtc.notification` event.
+ */
+export interface IRTCDeclineContent extends RelationEvent {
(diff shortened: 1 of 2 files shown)


src/matrixrtc/types.ts
```diff
@@ -104,6 +104,15 @@ export interface IRTCNotificationContent extends RelationEvent {
     "lifetime": number;
 }
 
+/**
+ * MSC4310 decline event content for `org.matrix.msc4310.rtc.decline`.
+ * Sent as a standard m.reference relation to an `org.matrix.msc4075.rtc.notification` event.
+ */
+export interface IRTCDeclineContent extends RelationEvent {
+    /** Optional human-readable reason for the decline */
+    reason?: string;
```
dbkr: See comment on MSC (ie. the problem with "human readable" is that not all humans can read the same language).

toger5: I was considering default codes but decided we move this to another msc once we have a product idea how this would be used.

toger5: We synced about this topic in our StandUp and decided to remove the field entirely.

dbkr: cool, sounds sensible to me!

dbkr (approved):

PR #4980, opened 2025-08-26
Fix thread edit aggregation race condition

basnijholt:
## Summary

Fixes a race condition where message edits in threads fail to display correctly when they arrive before thread initialization completes. This resolves the issue where Element Web shows the original message instead of the edited version in approximately 10% of cases.

**Fixes**: https://github.com/element-hq/element-web/issues/30617

## The Problem

When edit events arrive while a thread is not yet initialized (`initialEventsFetched = false`), the aggregation system attempts to link the edit to its target message. However, since the thread isn't initialized, the target message may not be in the timeline yet, causing the aggregation to fail silently.

The bug manifests as:
- Edit events are stored in the Relations container
- BUT the `targetEvent` remains `null` because the original message wasn't findable
- Element Web's `replacingEvent()` returns null even though edits exist
- Users see the original message instead of the edited version

## The Solution

Defer aggregation for edits until after the event is in the timeline and the thread is initialized, while allowing reactions to aggregate immediately. Concretely:
1. Edits (`RelationType.Replace`): do not aggregate before initialization; queue in `replayEvents` and aggregate after adding to the timeline post-init.
2. Reactions (`RelationType.Annotation`): aggregate immediately even before initialization to keep reaction summaries available; they will be re-aggregated safely on replay (idempotent).

This prevents premature aggregation of edits that creates broken Relations with `targetEvent: null`, while preserving the existing behavior for reactions.

### Root Cause
`addRelatedThreadEvent()` aggregated relations even when `initialEventsFetched = false`. If the target message wasn’t yet present in the thread timeline, Replace aggregation created a relations container with no `targetEvent`, so the edit did not apply.

### The Fix
Defer Replace aggregation until the thread is initialized and the event has been added to the thread timeline. Keep Annotation aggregation pre‑init to preserve reaction summaries; replay is safe because aggregation is idempotent. Also removed redundant manual aggregation in the post‑init path since timeline insertion already aggregates.

### Why This Works
- Edits that arrive before initialization are queued in `replayEvents` and aggregated only after the target is present in the timeline.
- Reactions aggregate immediately so reaction summaries remain available pre-init; during replay, re-aggregation is a no-op due to deduplication.
- On replay, the thread is initialized and edit aggregation succeeds with proper target linking.

### Idempotence and Deduplication
- `Relations.addEvent` tracks relation event IDs and ignores duplicates, so re-aggregating reactions on replay does not double-count.
- `EventTimelineSet.addEventToTimeline`/`insertEventIntoTimeline` also short-circuit on already-known events, avoiding duplicate timeline entries.

## Testing

Added tests to cover both the regression and behavior guarantees:
- Edit race regression: create a thread with `initialEventsFetched = false`, add edits before the original message, then initialize and replay. Without the fix, Replace relations aggregate prematurely and fail to link; with the fix, edits apply after replay.
- Reaction idempotence: reactions aggregate pre‑init for visibility and remain deduplicated after replay (no double counting).

## Impact

- **Bug frequency**: Affected ~10% of edited messages in threads (*in my application that edits messages many times*)
- **User impact**: Messages appeared unedited when they should show edits
- **Scope**: Only affects threads with server-side support enabled
- **Risk**: Low — edit aggregation is delayed to a safe point; reaction behavior is preserved and duplicate aggregation is idempotent.

## Notes for Reviewers

The key insight is that edit aggregation must not happen until the thread is initialized and events are in the timeline. The bug was subtle because:
1. It only occurred with specific timing (edits arriving during initialization)
2. The aggregation appeared to work (Relations object created) but was broken internally
3. The 90% success rate made it seem intermittent

The fix ensures edits only aggregate when they can succeed, preventing broken Relations objects. Reactions (e.g. 👍) still aggregate immediately; re-aggregation on replay is safe due to deduplication.

## Checklist

### DIFF
--- a/spec/unit/models/thread.spec.ts
+++ b/spec/unit/models/thread.spec.ts
@@ -778,18 +778,20 @@ describe("Thread", () => {
                 // This test reproduces the race condition bug from https://github.com/element-hq/element-web/issues/30617
                 // The bug occurs when edits arrive while the thread is not initialized,
                 // causing aggregation to fail because the target event isn't in the timeline yet
-                
+
                 // Given a thread exists with server-side support enabled
                 const myUserId = "@alice:example.org";
-                const testClient = new TestClient(myUserId, "DEVICE", "ACCESS_TOKEN", undefined, { timelineSupport: false });
+                const testClient = new TestClient(myUserId, "DEVICE", "ACCESS_TOKEN", undefined, {
+                    timelineSupport: false,
+                });
                 const client = testClient.client;
                 client.supportsThreads = jest.fn().mockReturnValue(true);
-                
+
                 const room = new Room("!room:z", client, myUserId, {
                     pendingEventOrdering: PendingEventOrdering.Detached,
                 });
                 jest.spyOn(client, "getRoom").mockReturnValue(room);
-                
+
                 // Create a root event
                 const rootEvent = mkMessage({
                     room: room.roomId,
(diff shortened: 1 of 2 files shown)


basnijholt:
@dbkr, thanks for the review! I've addressed the feedback in the latest push:

- Prettier: Fixed; formatting checks pass.
- Comment: Updated to mention keeping reaction counts/summaries visible during initialization.
- Safe pre-init aggregation: For reactions this is already safe and idempotent. For edits, I chose to defer aggregation rather than make the aggregator a no-op pre-init because:
    - It avoids creating relation containers with no target and then having to re-run aggregation later.
    - Aggregation for edits now runs when it can succeed fully (after the target is in the thread timeline).
    - We rely on `EventTimelineSet`'s insertion path, which already aggregates and links targets when present.

Reactions still aggregate pre-init to keep counts visible; replay is deduped, so no double counting. I also removed redundant aggregation calls, `EventTimelineSet` already handles aggregation on insert.

If you'd prefer the alternative (making Replace aggregation a no-op pre-init in the aggregator layer), I can explore that as a follow-up, though it touches a broader surface area than the thread-local fix here.

PR #4984, opened 2025-09-01
Fix `m.topic` format

### LINKED ISSUE #4902: `m.topic` field of `m.room.topic` has wrong format
The JS SDK expects this format:

```json
{
    "topic": "All about **pizza**",
    "m.topic": [{
        "body": "All about **pizza**",
        "mimetype": "text/plain",
    }, {
        "body": "All about <b>pizza</b>",
        "mimetype": "text/html",
    }],
}
```

But the actual format defined [in the spec](https://spec.matrix.org/v1.15/client-server-api/#mroomtopic) and MSC3765 is:

```json
{
    "topic": "All about **pizza**",
    "m.topic": {
        "m.text": [{
            "body": "All about **pizza**",
            "mimetype": "text/plain",
        }, {
            "body": "All about <b>pizza</b>",
            "mimetype": "text/html",
        }],
    }
}
```

To clarify, currently the SDK expects `m.topic` to be an array of objects, but `m.topic` should be an object with an `m.text` field that is an array of objects.

tulir:
Fixes #4902

I don't think any kind of backwards compatibility makes sense. The existing topics are malformed and if rooms have such topics, it's better that they notice the formatting disappears and fix it. An automatic migration would probably also be too invasive (it'd require sending events as the user)

### DIFF
--- a/spec/unit/content-helpers.spec.ts
+++ b/spec/unit/content-helpers.spec.ts
@@ -215,11 +215,11 @@ describe("Topic content helpers", () => {
         it("creates an empty event when the topic is falsey", () => {
             expect(makeTopicContent(undefined)).toEqual({
                 topic: undefined,
-                [M_TOPIC.name]: {"m.text": []},
+                [M_TOPIC.name]: { "m.text": [] },
             });
             expect(makeTopicContent(null)).toEqual({
                 topic: null,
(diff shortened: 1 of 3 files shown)


richvdh (approved):
LGTM, thanks for fixing this.

dbkr (approved):

PR #4992, opened 2025-09-10
History sharing: do `/keys/query` before checking for key bundle

richvdh:
The next release of matrix-sdk-crypto-wasm will check that the device that sent
us the key bundle data was correctly cross-signed by its owner, which means we
need to have the owner's cross-signing keys before we check if we have the
bundle.

This replicates a change made in the Rust SDK, at https://github.com/matrix-org/matrix-rust-sdk/pull/5510/files#diff-9f89fa75c4eb3743ae674be1bb90f75169bd815a259917799c71b8a546449d51R133-R140.

See also https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/264 which adds some documentation for future unwary travellers.

### DIFF
--- a/src/rust-crypto/rust-crypto.ts
+++ b/src/rust-crypto/rust-crypto.ts
@@ -336,7 +336,8 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
         // device that sent us the bundle data was correctly cross-signed.
         //
         // TODO: it would be nice to skip this step if we have an up-to-date copy of the inviter's cross-signing keys,
-        //   but we don't have an easy way to check that.
+        //   but we don't have an easy way to check that. Possibly the rust side could trigger a key request and then
+        //   block until it happens.
         logger.info(`Checking inviter cross-signing keys`);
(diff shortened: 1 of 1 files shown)


src/rust-crypto/rust-crypto.ts
```diff
@@ -326,12 +326,22 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
 
+        // Make sure we have an up-to-date idea of the inviter's cross-signing keys, so that we can check if the
+        // device that sent us the bundle data was correctly cross-signed.
+        //
+        // TODO: it would be nice to skip this step if we have an up-to-date copy of the inviter's cross-signing keys,
+        //   but we don't have an easy way to check that. Possibly the rust side could trigger a key request and then
+        //   block until it happens.
+        logger.info(`Checking inviter cross-signing keys`);
+        const request = this.olmMachine.queryKeysForUsers([new RustSdkCryptoJs.UserId(inviter)]);
+        await this.outgoingRequestProcessor.makeOutgoingRequest(request);
```
andybalaam, line 343: Does this wait until the response is received and processed?

richvdh: yes

PR #4994, opened 2025-09-12
Implement experimental encrypted state events.

kaylendog:
Adds support for MSC3414-esque encrypted state events.

- Updates `matrix-sdk-crypto-wasm` to `15.3.0` which includes support for encrypting state events.
- Adds an `MatrixClient.enableEncryptedStateEvents` flag to configure top-level behaviour.
- Modifies `MatrixClient.sendStateEvent` to call into the WASM SDK if encryption is enabled.
- Adds `MatrixEvent.getWireStateKey`, analagous to `MatrixEvent.getWireContent`.

This can be split into multiple PRs for ease of reviewing if required!

## Checklist

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -6639,7 +6639,13 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         opts: IRequestOpts = {},
     ): Promise<ISendEventResponse> {
         const room = this.getRoom(roomId);
-        const event = new MatrixEvent({ room_id: roomId, type: eventType, state_key: stateKey, content });
+        const event = new MatrixEvent({
+            room_id: roomId,
+            type: eventType,
+            state_key: stateKey,
+            // Cast safety: StateEvents[K] is a stronger bound than IContent, which has [key: string]: any
+            content: content as IContent,
+        });
 
         await this.encryptStateEventIfNeeded(event, room ?? undefined);
 


spec/integ/crypto/crypto.spec.ts
```diff
@@ -104,107 +107,6 @@ afterEach(() => {
...
-    const encryptedMessageContent = await expectEncryptedSendMessage();
-
-    // In some of the tests, the room key is sent *after* the actual event, so we may need to wait for it now.
-    const inboundGroupSession = await inboundGroupSessionPromise;
-
-    const r: any = inboundGroupSession.decrypt(encryptedMessageContent!.ciphertext);
-    logger.log("Decrypted received megolm message", r);
-    return JSON.parse(r.plaintext);
-}
-
```
kaylendog, line 207: I have pulled these out to `olm-utils.ts` to reduce code dupe with the new `state-events.spec.ts`.

src/rust-crypto/index.ts
```diff
@@ -91,6 +91,11 @@ export async function initRustCrypto(args: {
      * Called with (-1, -1) to mark the end of migration.
      */
     legacyMigrationProgressListener?: (progress: number, total: number) => void;
+
+    /**
+     * Enable support for state encrypt
```
richvdh:
```suggestion
     * Whether to enable support for encrypting state events.
```

src/rust-crypto/libolm_migration.ts
```diff
@@ -395,6 +395,7 @@ export async function migrateRoomSettingsFromLegacyCrypto({
                 continue;
             }
             rustSettings.algorithm = RustSdkCryptoJs.EncryptionAlgorithm.MegolmV1AesSha2;
+            rustSettings.encryptStateEvents = false;
```
richvdh: isn't this the default? I'd leave it out here, rather than fill everything with experimental options.

src/models/event.ts
```diff
@@ -728,11 +729,23 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
...
     /**
-     * Get the event state_key if it has one. This will return <code>undefined
+     * Get the (decrypted, if necessary) event state_key if it has one. This will return <code>undefined
      * </code> for message events.
      * @returns The event's `state_key`.
      */
     public getStateKey(): string | undefined {
+        if (this.clearEvent) {
+            return this.clearEvent.state_key;
+        }
```
richvdh, line 740: :thinking: did we remember to check that, if we receive a regular (i.e., not-state) `m.room.encrypted` event, then the sender can't smuggle a `state_key` into its payload and make the application think it's a state event?

src/client.ts
```diff
@@ -6625,23 +6638,80 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+        if (!this.cryptoBackend) {
+            throw new Error("This room is configured to use encryption, but your client does not support encryption.");
+        }
+
+        if (!this.cryptoBackend && this.usingExternalCrypto) {
+            // The client has opted to allow sending messages to encrypted
+            // rooms even if the room is encrypted, and we haven't set up
+            // crypto. This is useful for users of matrix-org/pantalaimon
+            return;
+        }
```
richvdh, line 6677: this needs to happen before the above check

src/client.ts
```diff
@@ -6625,23 +6638,80 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+
+        // Check regular encryption conditions.
+        if (!(await this.shouldEncryptEventForRoom(event, room))) {
+            this.logger.info("Discarded - does not match regular conditions");
+            return;
+        }
+
+        // If we have a crypto impl, and *it* thinks we shouldn't encrypt, then we shouldn't.
```
richvdh: by this point we know we have a crypto impl

src/client.ts
```diff
@@ -6625,23 +6638,80 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+        if (!this.cryptoBackend && this.usingExternalCrypto) {
+            // The client has opted to allow sending messages to encrypted
+            // rooms even if the room is encrypted, and we haven't set up
+            // crypto. This is useful for users of matrix-org/pantalaimon
+            return;
+        }
+
+        // Check regular encryption conditions.
+        if (!(await this.shouldEncryptEventForRoom(event, room))) {
+            this.logger.info("Discarded - does not match regular conditions");
```
richvdh: it's not discarded. Likewise below.

src/client.ts
```diff
@@ -6625,23 +6638,80 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+                "m.room.history_visibility",
+                "m.room.guest_access",
+                "m.room.encryption",
+            ].includes(event.getType())
+        ) {
+            this.logger.info("Discarded - is a critical event");
+            return;
+        }
+
+        this.logger.info("Encrypting state event");
```
richvdh: I don't think this log line is going to be hugely helpful when it appears in the middle of a massive log file.

Either give more info about which event we are encrypting, or just get rid of it.

src/client.ts
```diff
@@ -1206,6 +1211,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
 
     private cryptoBackend?: CryptoBackend; // one of crypto or rustCrypto
     public cryptoCallbacks: CryptoCallbacks; // XXX: Intended private, used in code.
+    public enableEncryptedStateEvents: boolean;
```
richvdh: please don't add public fields to this class, there are far too many as it is

richvdh: also, can it be `readonly`?

src/sync.ts
```diff
@@ -1216,7 +1216,7 @@ export class SyncApi {
             // this helps large account to speed up faster
             // room::decryptCriticalEvent is in charge of decrypting all the events
             // required for a client to function properly
-            const timelineEvents = this.mapSyncEventsFormat(joinObj.timeline, room, false);
+            const timelineEvents = this.mapSyncEventsFormat(joinObj.timeline, room);
```
richvdh: what is with this change? At the very least the comment above needs updating.

src/sync.ts
```diff
@@ -1366,6 +1366,10 @@ export class SyncApi {
                 }
             }
 
+            for (const ev of timelineEvents.filter((ev) => ev.isState())) {
+                await this.client.decryptEventIfNeeded(ev);
+            }
```
richvdh, line 1374: please can you add a comment here? Why are we doing this, and why only for state events?

src/models/event.ts
kaylendog: This is done by `OlmMachine::verify_packed_state_key` in the [Rust SDK](https://github.com/matrix-org/matrix-rust-sdk/blob/b8cbd6c4488eb38f3580f150c52cb4c3908ef4a6/crates/matrix-sdk-crypto/src/machine/mod.rs#L2293). Might be worth adding a test for this somewhere?

src/client.ts
kaylendog: Would `Encryption cancelled` be better? I should probably also decrease the verbosity.

src/models/event.ts
richvdh: that method contains this:

```rust
        // We only need to verify state events.
        let Some(raw_state_key) = &original.state_key else { return Ok(()) };
```

... which sounds very much like we do *not* check for the scenario I described (which will have `original.state_key = None`).

Anyway yes, tests for the win :)

kaylendog: Working on a fix here: https://github.com/matrix-org/matrix-rust-sdk/pull/5666

spec/integ/crypto/state-events.spec.ts
```diff
@@ -0,0 +1,220 @@
+/*
+Copyright 2016 OpenMarket Ltd
+Copyright 2019-2023 The Matrix.org Foundation C.I.C.
```
andybalaam: If this is a new file, this should be just one `Copyright...` line:

```
Copyright 2025 The Matrix.org Foundation C.I.C.
```

spec/unit/models/event.spec.ts
```diff
@@ -335,6 +335,25 @@ describe("MatrixEvent", () => {
...
     });
 
+    describe("state key packing", () => {
+        it("should pack the state key during encryption", () => {
+            const ev = createStateEvent("$event1:server", "m.room.topic", "", { topic: "" });
+            expect(ev.getStateKey()).toStrictEqual("");
+            ev.makeEncrypted("m.room.encrypted", { ciphertext: "xyz" }, "", "");
+            expect(ev.getStateKey()).toStrictEqual("");
+            expect(ev.getWireStateKey()).toStrictEqual("m.room.topic:");
+        });
```
andybalaam, line 351: It would be good to test a state event with a non-empty state key too.

src/client.ts
andybalaam: I think no log here - this is normal e.g. if the room is not encrypted.

src/client.ts
```diff
@@ -6625,23 +6638,80 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+        // Check regular encryption conditions.
+        if (!(await this.shouldEncryptEventForRoom(event, room))) {
+            this.logger.info("Discarded - does not match regular conditions");
+            return;
+        }
+
+        // If the crypto impl thinks we shouldn't encrypt, then we shouldn't.
+        // Safety: we checked the crypto impl exists above.
+        if (!(await this.cryptoBackend!.isStateEncryptionEnabledInRoom(room.roomId))) {
+            this.logger.info("Discarded - does not match state event conditions");
```
andybalaam: This is also normal, so no log message or trace level. This is what happens when people don't apply the labs flag (right?)

src/client.ts
```diff
@@ -6625,23 +6638,80 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+                "m.room.member",
+                "m.room.join_rules",
+                "m.room.power_levels",
+                "m.room.third_party_invite",
+                "m.room.history_visibility",
+                "m.room.guest_access",
+                "m.room.encryption",
+            ].includes(event.getType())
+        ) {
+            this.logger.info("Discarded - is a critical event");
```
andybalaam: Again, very normal. Can trace if needed, or no log.

src/models/event.ts
```diff
@@ -728,11 +729,23 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
...
+        return this.event.state_key;
+    }
+
+    /**
+     * Get the (possibly encrypted) event state_key if it has one. This will return <code>undefined
+     * </code> for message events.
```
andybalaam, line 747: The state key won't be encrypted, right? It will contain the type as well as the original state key.

So "get the raw state key" would probably suffice, along with saying it returns the concatenated value if this event was originally encrypted.

src/models/event.ts
```diff
@@ -728,11 +729,23 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
     }
 
     /**
-     * Get the event state_key if it has one. This will return <code>undefined
+     * Get the (decrypted, if necessary) event state_key if it has one. This will return <code>undefined
```
andybalaam: The state key won't be decrypted (because it wasn't encrypted), but the string-packing will have been undone. Maybe you can say that more explicitly.

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -314,11 +314,22 @@ export class RoomEncryptor {
...
-            event.getType(),
-            JSON.stringify(event.getContent()),
-        );
+
+        let encryptedContent;
+        if (event.isState()) {
+            encryptedContent = await this.olmMachine.encryptStateEvent(
+                new RoomId(this.room.roomId),
+                event.getType(),
+                event.getStateKey()!,
```
andybalaam, line 328: What should happen if the state key is missing? Probably we shouldn't crash. Maybe we should log an error and return empty content.

src/rust-crypto/RoomEncryptor.ts
```diff
@@ -314,11 +314,22 @@ export class RoomEncryptor {
...
     private async encryptEventInner(logger: LogSpan, event: MatrixEvent): Promise<void> {
         logger.debug("Encrypting actual message content");
-        const encryptedContent = await this.olmMachine.encryptRoomEvent(
-            new RoomId(this.room.roomId),
-            event.getType(),
-            JSON.stringify(event.getContent()),
-        );
+
+        let encryptedContent;
+        if (event.isState()) {
```
andybalaam, line 323: I'd prefer to create the room, type and content outside the if, so the if just controls which `encrypt...` method to call, and getting state key if needed.

src/sync.ts
```diff
@@ -1366,6 +1366,12 @@ export class SyncApi {
                 }
             }
 
+            // The JS SDK does not proactively decrypt state events, but this is not applicable for state events
+            // since we should process their contents immediately.
```
andybalaam: I don't understand this, sorry! Can you try to explain further? Maybe we can chat over Matrix and figure it out.

src/rust-crypto/RoomEncryptor.ts
kaylendog: The state key can't be missing here since we assert `event.isState()` above. (Ideally, it'd be nice to use type guards here so the assertion is unnecessary, but `MatrixEvent` isn't set up like that atm.)

andybalaam: Great. Can you add a comment to point this out?

andybalaam:
Awesome, please rebase and force-push and I'll merge

PR #5010, opened 2025-09-16
Add call intent to RTC call notifications

Half-Shot:
Based on https://github.com/matrix-org/matrix-spec-proposals/pull/4075/files#r2352607204

## Checklist

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -638,12 +638,12 @@ export class MatrixRTCSession extends TypedEventEmitter<
         return undefined;
     }
 
-    public updateCallIntent(callIntent: RTCCallIntent): undefined {
+    public async updateCallIntent(callIntent: RTCCallIntent): Promise<void> {
         const myMembership = this.membershipManager?.ownMembership;
         if (!myMembership) {
             throw Error("Not connected yet");
         }
-        this.membershipManager?.updateCallIntent(callIntent);
+        await this.membershipManager?.updateCallIntent(callIntent);
     }
 
     /**
--- a/src/matrixrtc/MembershipManager.ts
+++ b/src/matrixrtc/MembershipManager.ts
@@ -231,15 +231,15 @@ export class MembershipManager
 
     private leavePromiseResolvers?: PromiseWithResolvers<boolean>;
 
-    public async onRTCSessionMemberUpdate(memberships: CallMembership[]): Promise<void> {
+    public onRTCSessionMemberUpdate(memberships: CallMembership[]): Promise<void> {
         if (!this.isActivated()) {
-            return;
+            return Promise.resolve();
         }
         const userId = this.client.getUserId();
         const deviceId = this.client.getDeviceId();
         if (!userId || !deviceId) {
             this.logger.error("MembershipManager.onRTCSessionMemberUpdate called without user or device id");
-            return;
+            return Promise.resolve();
         }
         this._ownMembership = memberships.find((m) => isMyMembership(m, userId, deviceId));
 
@@ -263,6 +263,7 @@ export class MembershipManager
                 this.scheduler.initiateJoin();
             }
         }
+        return Promise.resolve();
     }
 
     public getActiveFocus(): Focus | undefined {
@@ -286,7 +287,6 @@ export class MembershipManager
 
     public async updateCallIntent(callIntent: RTCCallIntent): Promise<void> {
         if (!this.activated || !this.ownMembership) {
-            console.log(new Error().stack);
             throw Error("You cannot update your intent before joining the call");
         }
         if (this.ownMembership.callIntent === callIntent) {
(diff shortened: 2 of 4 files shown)


src/matrixrtc/CallMembership.ts
```diff
@@ -70,36 +71,52 @@ export type SessionMembershipData = {
...
+): data is SessionMembershipData => {
     const prefix = "Malformed session membership event: ";
     if (typeof data.device_id !== "string") errors.push(prefix + "device_id must be string");
     if (typeof data.call_id !== "string") errors.push(prefix + "call_id must be string");
     if (typeof data.application !== "string") errors.push(prefix + "application must be a string");
     if (typeof data.focus_active?.type !== "string") errors.push(prefix + "focus_active.type must be a string");
     if (!Array.isArray(data.foci_preferred)) errors.push(prefix + "foci_preferred must be an array");
     // optional parameters
-    if (data.created_ts && typeof data.created_ts !== "number") errors.push(prefix + "created_ts must be number");
+    if (data.created_ts !== undefined && typeof data.created_ts !== "number") {
```
Half-Shot, line 109: This check was invalid before,

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -614,6 +620,24 @@ export class MatrixRTCSession extends TypedEventEmitter<
...
     }
 
+    /**
+     * Get the call intent for the current call, based on what members are advertising. If one or more
+     * members disagree on the current call intent, or nobody specifies one then `undefined` is returned.
+     *
+     * If all members that specify a call intent agree, that value is returned.
+     * @returns A call intent, or `undefined` if no consensus or not given.
+     */
+    public getConsensusCallIntent(): RTCCallIntent | undefined {
```
Half-Shot, line 630: This is intended for visual identification in room lists and so on, so that you can tell from a glance what the type of call is. Obviously this *can* be lies, but does require all the call users to be liars :)

src/matrixrtc/CallMembership.ts
```diff
@@ -31,37 +32,37 @@ type CallScope = "m.room" | "m.user";
...
  * MSC4143 (MatrixRTC) session membership data.
- * Represents an entry in the memberships section of an m.call.member event as it is on the wire.
+ * Represents the `session` in the memberships section of an m.call.member event as it is on the wire.
  **/
 export type SessionMembershipData = {
     /**
      * The RTC application defines the type of the RTC session.
      */
-    application: string;
+    "application": string;
```
BillCarsonFr, line 41: Quotes needed for that one? Is it in the others to avoid some warnings about camel case names

src/matrixrtc/CallMembership.ts
BillCarsonFr: because it could be 0?

src/matrixrtc/MatrixRTCSession.ts
BillCarsonFr: oh intent means audio or video. I got confused with the widget intents (startCallDm etc...)

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -667,7 +691,11 @@ export class MatrixRTCSession extends TypedEventEmitter<
     /**
      * Sends a notification corresponding to the configured notify type.
      */
-    private sendCallNotify(parentEventId: string, notificationType: RTCNotificationType): void {
+    private sendCallNotify(
```
BillCarsonFr, line 706: A bit surprising that this is not async / returns a promise. Maybe document that it is scheduling the send and will fire an event in case of success. Or it will silently fail.
I suppose there will be some retry at least for the new rtc.notification type because it has a relation and that makes it elligible for the send queue

src/matrixrtc/MembershipManager.ts
```diff
@@ -281,6 +282,16 @@ export class MembershipManager
         }
     }
 
+    /**
+     * Update the intent of a membership on the call (e.g. user is now providing a video feed)
+     * @param callIntent The new intent to set.
+     */
+    public updateCallIntent(callIntent: RTCCallIntent): void {
+        this.callIntent = callIntent;
```
BillCarsonFr, line 295: no check to see if is actually a different intent?
It is sort of new that we update the join event no? If I remember as per the spec there is something to do with the `created_ts`? so that recipient properly react to the change?

BillCarsonFr (commented):
Please see my comments

src/matrixrtc/CallMembership.ts
Half-Shot: Yep. Which, granted a nitpick, but let's be accurate :)

src/matrixrtc/CallMembership.ts
Half-Shot: This is the divine judgement of prettier :shrug: . I dare not defy it.

src/matrixrtc/MatrixRTCSession.ts
Half-Shot: I don't really know why this doesn't return a promise to be honest. @timokoesters might know the reasoning. I'll document how it works, in any case.

src/matrixrtc/MembershipManager.ts
Half-Shot: Hah, I just added that independently. I should double check the spec just allows me to update the membership though.

Half-Shot: I've added this + added tests.

BillCarsonFr (approved):
LGTM !

PR #5011, opened 2025-09-16
Stop re-adding `displayname`/`avatar` fields to leave/ban state events

gingershaped:
This PR removes a snippet of code that quietly re-added `displayname` and `avatar_url` fields to the cached version of `leave` and `ban` `m.room.member` state events stored in `RoomMember` objects. This code caused the cached member event to be inconsistent with the actual member event returned by other functions, and also caused the (potentially abusive) avatars and displaynames of banned users to appear in clients which call functions that access the cached data.

## Checklist

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -23,6 +23,7 @@ const config: Config = {
     collectCoverageFrom: ["<rootDir>/src/**/*.{js,ts}"],
     coverageReporters: ["text-summary", "lcov"],
     testResultsProcessor: "@casualbot/jest-sonar-reporter",
+    transformIgnorePatterns: ["/node_modules/(?!(uuid|p-retry|is-network-error)).+$"],
 
     // Always print out a summary if there are any failing tests. Normally
     // a summary is only printed if there are more than 20 test *suites*.
--- a/package.json
+++ b/package.json
@@ -58,10 +58,10 @@
         "matrix-events-sdk": "0.0.1",
         "matrix-widget-api": "^1.10.0",
         "oidc-client-ts": "^3.0.1",
-        "p-retry": "4",
+        "p-retry": "7",
         "sdp-transform": "^2.14.1",
         "unhomoglyph": "^1.0.6",
-        "uuid": "11"
+        "uuid": "13"
     },
     "devDependencies": {
         "@action-validator/cli": "^0.6.0",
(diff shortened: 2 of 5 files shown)


dbkr:
The comment fairly strongly suggests that it does it for a reason though… will this cause, eg. all messages from people who've left to appear with no display name / avatar? Or is this fine because that's kept in the sentinel events? Do all clients actually use them reliably?

gingershaped:
To be more specific, here's the scenario I observed:

1. troll joins and sends spam messages
2. room admin redacts the messages, bans the troll, and redacts the troll's join event
3. Cinny shows the redacted messages with the troll's abusive avatar, which it reads from the `prev_content` field of the `m.room.member` state event that banned the troll

`matrix-js-sdk`'s `getMxcAvatarUrl` function appears to be intended to return the latest avatar a user has in a room. (As far as I can tell, the SDK does not implement the algorithm suggested in [10.2.2.4](https://spec.matrix.org/v1.15/client-server-api/#displaying-membership-information-with-messages) for displaying historical display names and avatars.) When a user leaves or is banned from a room, they no longer have a latest avatar, so this function should return `undefined` (which is the behavior this PR implements). The current behavior returns the last avatar the user had _set_ in the room, which is different from 10.2.2.4's suggested algorithm because it ignores redactions to older state events.

It might be desirable in the future for `matrix-js-sdk` to include a way to return historical avatars and display names in a way that respects redactions, as described in 10.2.2.4, but I deemed such functionality to be out of the scope of this PR.

dbkr:
The js-sdk's current behaviour does seem a bit dumb here. While returning no avatar / display name for left users might be argued to be technically correct, if it's going to cause clients to suddenly start displaying left users with no display name or avatar, I don't think this is a thing we can just hit merge on. It would at least be a breaking change. If possible, fixing `getMxcAvatarUrl`'s current behaviour to respect redactions might be preferable?

gingershaped:
> The js-sdk's current behaviour does seem a bit dumb here. While returning no avatar / display name for left users might be argued to be technically correct, if it's going to cause clients to suddenly start displaying left users with no display name or avatar, I don't think this is a thing we can just hit merge on. It would at least be a breaking change. If possible, fixing `getMxcAvatarUrl`'s current behaviour to respect redactions might be preferable?

As written this is currently a breaking change, yeah. I'll investigate fixing `getMxcAvatarUrl`.

PR #5015, opened 2025-09-19
Add parseCallNotificationContent

toger5:

## Checklist

### DIFF
--- a/src/matrixrtc/types.ts
+++ b/src/matrixrtc/types.ts
@@ -13,7 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-import type { IMentions } from "../matrix.ts";
+import type { IContent, IMentions } from "../matrix.ts";
 import type { RelationEvent } from "../types.ts";
 import type { CallMembership } from "./CallMembership.ts";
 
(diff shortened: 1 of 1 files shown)


src/matrixrtc/types.ts
```diff
@@ -96,6 +96,41 @@ export interface ICallNotifyContent {
...
 export type RTCNotificationType = "ring" | "notification";
+/**
+ * This will check if the content has all the expected fields to be a valid IRTCNotificationContent.
+ * It will also cap the lifetime to 120000ms if a higher value is provided.
+ * @param content
+ * @throws if the content is invalid
+ * @returns a parsed IRTCNotificationContent
+ */
+export function parseCallNotificationContent(content: IContent): IRTCNotificationContent {
+    if (!content["m.mentions"]) {
```
BillCarsonFr: so this is expected to at least be an empty [] array? cannot be skipped? I think ruma skips it if it is not set

src/matrixrtc/types.ts
```diff
@@ -96,6 +96,41 @@ export interface ICallNotifyContent {
...
+        throw new Error("Missing or invalid notification_type");
+    }
+    if (typeof content["sender_ts"] !== "number") {
+        throw new Error("Missing or invalid sender_ts");
+    }
+    if (typeof content["lifetime"] !== "number") {
+        throw new Error("Missing or invalid lifetime");
+    }
+
+    if (content["decline_reason"] && typeof content["decline_reason"] !== "string") {
```
BillCarsonFr: This is new to me, a decline_reason in a RTCNotificationContent?

toger5: you are right this is legacy... Not there anymore. its from the time where we used the notification event as the decline as well. This should have been done in the PR that added the decline event type...

src/matrixrtc/types.ts
toger5: I think it also might already do checks on this in the notifier. Lets only check for it to be an obj here.

src/matrixrtc/types.ts
```diff
@@ -102,9 +102,45 @@ export type RTCNotificationType = "ring" | "notification";
  * May be any string, although `"audio"` and `"video"` are commonly accepted values.
  */
 export type RTCCallIntent = "audio" | "video" | string;
+
+/**
+ * This will check if the content has all the expected fields to be a valid IRTCNotificationContent.
+ * It will also cap the lifetime to 120000ms if a higher value is provided.
```
BillCarsonFr: For consistency with EX we might want to use 90s also? https://github.com/element-hq/element-x-android/blob/9dd90bcd4e278d3611f94b85cb20806c91c8ff1b/appconfig/src/main/kotlin/io/element/android/appconfig/ElementCallConfig.kt#L14
https://github.com/element-hq/element-x-android/blob/9dd90bcd4e278d3611f94b85cb20806c91c8ff1b/features/call/impl/src/main/kotlin/io/element/android/features/call/impl/utils/ActiveCallManager.kt#L127-L130

toger5: thanks for reminding me on this.

PR #5016, opened 2025-09-22
Exclude cancelled requests from in-progress lists

andybalaam:
Fixes https://github.com/element-hq/element-web/issues/29882

When we ask for the in-progress verification requests, exclude requests that have been cancelled. This means that we don't erroneously tell the user that the new request they are about to create has been cancelled.

A higher-level test for this bug is introduced in https://github.com/element-hq/element-web/pull/30835

### DIFF
--- a/src/rust-crypto/rust-crypto.ts
+++ b/src/rust-crypto/rust-crypto.ts
@@ -1040,7 +1040,7 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
             new RustSdkCryptoJs.UserId(userId),
         );
         return requests
-            .filter((request) => request.roomId === undefined)
+            .filter((request) => request.roomId === undefined && !request.isCancelled())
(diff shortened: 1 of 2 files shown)


src/rust-crypto/rust-crypto.ts
```diff
@@ -1040,7 +1040,11 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
             new RustSdkCryptoJs.UserId(userId),
         );
         return requests
-            .filter((request) => request.roomId === undefined)
+            .filter(
+                (request) =>
+                    request.roomId === undefined &&
+                    request.phase() !== RustSdkCryptoJs.VerificationRequestPhase.Cancelled,
```
uhoreg: Why are you checking the phase here, but using `isCancelled` in the other check?

andybalaam: Good point, fixed in b9d24ba4b

PR #5017, opened 2025-09-23
[MatrixRTC] Sticky Events support (MSC4354)

toger5:
Depends on #5028 

This PR adds support for sticky events in the MatrixRTCSession:
 - send sticky events instead of state events
 - update the call memberships based on sticky events instead of membership changes.
 - resend the current membership if the sticky event times out with an updated expiration ts (This is not the final solution. the sticky event timeout in iteself is enough. no expiration needed on the event content)

### DIFF
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -50,7 +50,7 @@ import {
 } from "./RoomAndToDeviceKeyTransport.ts";
 import { TypedReEmitter } from "../ReEmitter.ts";
 import { ToDeviceKeyTransport } from "./ToDeviceKeyTransport.ts";
-import { type MatrixEvent } from "src/matrix.ts";
+import { type MatrixEvent } from "../models/event.ts";
 import { RoomStickyEventsEvent, type RoomStickyEventsMap } from "../models/room-sticky-events.ts";
 
 /**
--- a/src/matrixrtc/MembershipManager.ts
+++ b/src/matrixrtc/MembershipManager.ts
@@ -21,6 +21,7 @@ import {
     type SendDelayedEventResponse,
     UpdateDelayedEventAction,
 } from "../@types/requests.ts";
+import { type EmptyObject } from "../@types/common.ts";
 import type { MatrixClient } from "../client.ts";
 import { ConnectionError, HTTPError, MatrixError } from "../http-api/errors.ts";
 import { type Logger, logger as rootLogger } from "../logger.ts";
@@ -46,7 +47,6 @@ import {
     type IMembershipManager,
     type MembershipManagerEventHandlerMap,
 } from "./IMembershipManager.ts";
-import { type EmptyObject } from "src/matrix.ts";
 
 /* MembershipActionTypes:
 On Join:  ───────────────┐   ┌───────────────(1)───────────┐


src/models/room.ts
```diff
@@ -3414,6 +3417,77 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+        }
+        this.checkStickyTimer();
+    }
+
+    private static stickyKey(event: MatrixEvent): string | undefined {
+        const k = event.getContent().sticky_key;
+        if (!k) {
+            return undefined;
+        }
+        return event.getSender() + event.getType() + k;
```
kegsay: This needs more appropriate namespacing to prevent collisions.

src/models/room.ts
```diff
@@ -3414,6 +3417,77 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+            const stickyKey = Room.stickyKey(e);
+            if (!stickyKey) {
+                logger.warn("Event from sticky sync section is missing sticky_key, ignoring", e);
+                continue;
+            }
+            if ("sticky" in e.event) {
+                logger.warn("Event from sticky sync section is missing sticky timeout", e);
+                continue;
+            }
+            this.stickyEventsMap.set(stickyKey, e);
```
kegsay: Whilst most of the time this will work since live events tend to have higher origin timestamps, this is not honouring the MSC tie-break of:

> Nothing stops users sending multiple events with the same sticky_key. To deterministically tie-break, clients which implement this behaviour MUST:
>
> - pick the one with the highest origin_server_ts,
> - tie break on the one with the highest lexicographical event ID (A < Z).

src/models/room.ts
```diff
@@ -3414,6 +3417,77 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+            this.cleanExpiredStickyEvents();
+        } else if (nextExpiry) {
+            this.stickyEventTimer = setTimeout(this.cleanExpiredStickyEvents, nextExpiry - Date.now());
+        }
+    }
+
+    private getNextStickyExpiryTs(): number | null {
+        let nextExpiry = null;
+        for (const event of this.stickyEventsMap.values()) {
+            const expiry = event.getTs() + event.event.msc4354_sticky!;
```
kegsay: This surely doesn't work? The `msc4354_sticky` field in the event is of the form:
```json
"msc4354_sticky": {
    "duration_ms": 360000
}
```

src/models/room.ts
```diff
@@ -3414,6 +3417,77 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+            if (nextExpiry === null || expiry < nextExpiry) {
+                nextExpiry = expiry;
+            }
+        }
+        return nextExpiry;
+    }
+    private cleanExpiredStickyEvents(): void {
```
kegsay: This may not be appropriate for your use case.

You are already relying on delayed events to "expire" RTC member events. I suggest you keep using that behaviour. You don't need to expire sticky events when they lose their stickiness for your use case. The purpose of sticky events is to ensure they get delivered to the client. Now they are at the client, they can persist for as long or as little as you like. In some use cases you may want to expire sticky events, but this isn't one imo.

The net result is that you are maintaining a k:v map which any user can write to, namespaced by each user, exactly the same as the intended purpose of MSC3757/MSC3779.

kegsay: Although having said this, if you see an RTC member, and then someone sends an expired RTC member but then your server disappears for >1 hour before you can get it, you won't receive the expired RTC member event when you come back online at all, so you'd need to use sticky event expiration to infer this...

toger5: This was mostly that it did not felt right to have an infinitly growing map in memory. So i would have added a "delete after 1000 entires", "delete after DURATION", "store to a db eventually" or similar anyways.

This cleanup condition was just the obvious one in this case.

src/models/room.ts
toger5: so sorting by `getTs()` is required here. thx

src/models/room.ts
toger5: the room is implicit becuase this map is stored inside the room. It cannot collide with other rooms sticky map.
Or are you thinking about sth else?

src/models/room.ts
toger5: > Although having said this, if you see an RTC member, and then someone sends an expired RTC member but then your server disappears for >1 hour before you can get it, you won't receive the expired RTC member event when you come back online at all, so you'd need to use sticky event expiration to infer this...

This is also a good point. The "opposite direction" is als a problem. We basically need to mimic the sync behavior to avoid split brains. Of course we expect delayed events to work but if they dont we have a split brain.

kegsay: By "opposite direction" you mean if a client just doesn't send the expired RTC member event right?

src/models/room.ts
kegsay: I'm thinking of sender and type clashing to allow unauthorised overwriting of user-scoped keys.

Sender | Type | key | concatenated key
-|-|-|-
@ alice:example.com|cast_to_device|baz| @ alice:example.comcast_to_devicebaz
@ alice:example.comcast[^tld]|_to_device|baz| @ alice:example.comcast_to_devicebaz


[^tld]: `.comcast` is a valid top-level domain.

src/models/room.ts
toger5: Yes instead of the server disppearing the client is "disappearing" (in terms of not doing its job in sending the hangup/leave event or forgetting/failing to send an updated sticky event) then we could have split brains based on when different clients do their sync loop.

toger5: This might be worth considering to mention in the MSC? Phrasing it a bit more abstract the following applies to all use cases that expect the local sticky map to be synchronised:

If client implementations rely on the local map to store sticky events by 4-uple to be synchronized between clients, with working network connections, sticky events have to be removed from the local db once they expire.

Otherwise the map looks different on a client that did not sync during the full sticky event lifetime. (From origin_server_ts until origin_server_ts + sticky_duration_ms). We return to the same issue we have with gappy syncs.

It is therefore recommended to run client side timers mimicking the server timers to remove sticky events with sticky_key s from the local map. Clients that did/didnt sync during the sticky duration only then end up with the map once all of them are back online.

src/models/room.ts
toger5: Of course!
Not sure what the most desired js system to use here. Probably a nested map (arrays compare by ref iirc...). I will check on this tomorrow.

toger5: Ended up adding `MultiKeyMap`

src/models/room.ts
BillCarsonFr: Agree, you need to expire locally and ignore expired events, in order to allow every clients to see the same thing (include clients getting a gap)

src/models/room.ts
Half-Shot: Having benchmarked the approach, JSONifying keys each time was adding a slight delay. I had a look at this again and doing `type`+`userId` => [event] allows for storing keys safely. In addition, it optimises for the most likely use-case of sticky events which is (application type) (user ID) => [instances of application], so we end up in a good place without losing the performance benefits of directly using the `Map` type.

To alleviate the fears from https://github.com/matrix-org/matrix-js-sdk/pull/5017#discussion_r2376146049, we are placing the format-constrained key at the end so whatever your `type` ends up containing, the UserID will always be appended. Because we no longer include `key` on the Map key, the userId is not suffixed by any values.

spec/unit/models/room.spec.ts
```diff
@@ -13,7 +13,6 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-
```
toger5, line 16: lets remove this file diff

src/models/room-sticky-events.ts
```diff
@@ -0,0 +1,212 @@
...
+     * Adds a sticky event into the local sticky event map.
+     *
+     * NOTE: This will not cause `RoomEvent.StickyEvents` to be emitted.
+     *
+     * @throws If the `event` does not contain valid sticky data.
+     * @param event The MatrixEvent that contains sticky data.
+     * @returns An object describing whether the event was added to the map,
+     *          and the previous event it may have replaced.
+     */
+    public unstableAddStickyEvent(event: MatrixEvent): { added: true; prevEvent?: MatrixEvent } | { added: false } {
```
toger5: lets use the same `_unstable_` as we do in client `_unstable_` methods.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -230,6 +230,11 @@ export interface EncryptionConfig {
 }
 export type JoinSessionConfig = SessionConfig & MembershipConfig & EncryptionConfig;
 
+interface SessionMembershipsForRoomOpts {
+    listenForStickyEvents: boolean;
+    listenForMemberStateEvents: boolean;
+}
```
toger5, line 253: Wondering if we really want to make this configurable?
It would be great if we can just always make it listen to this?
Not sure its possible however...
Do you have an idea for a case where "always on" would break?

src/matrixrtc/MatrixRTCSessionManager.ts
```diff
@@ -73,6 +73,7 @@ export class MatrixRTCSessionManager extends TypedEventEmitter<MatrixRTCSessionM
         }
 
         this.client.on(ClientEvent.Room, this.onRoom);
+        this.client.on(ClientEvent.Event, this.onEvent);
```
toger5, line 76: this in combination with:
`        this.roomSubset.on(RoomEvent.StickyEvents, this.onStickyEventUpdate);`
will result in two calls to `recalculateSessionMembers` once via `refreshRoom` -> `onRTCSessionMemberUpdate` -> `recalculateSessionMembers` And a second time through`onStickyEventUpdate`->`recalculateSessionMembers.

Or is there any check so one of them gets ignored?

A possible solution would be to check if the associated room already has a session (in `onRoom`) and if so, we do not call refresh room? So it only gets called on creating a new session?

This logic maybe should be applied to all of this file. `refreshRoom` should be: `refreshSessionStore/Map` and than we expect the session itself to listen to all the evnt updates. Listeners in the SessionManager only are responsible to find out if new sessions need to be created!

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -757,10 +811,28 @@ export class MatrixRTCSession extends TypedEventEmitter<
     /**
      * Call this when the Matrix room members have changed.
      */
-    public onRoomMemberUpdate = (): void => {
+    private onRoomMemberUpdate = (): void => {
         this.recalculateSessionMembers();
     };
 
+    /**
+     * Call this when a sticky event update has occured.
```
toger5, line 854:
```suggestion
     * Call this when a sticky event update has occurred.
```

src/matrixrtc/MatrixRTCSessionManager.ts
```diff
@@ -113,16 +115,31 @@ export class MatrixRTCSessionManager extends TypedEventEmitter<MatrixRTCSessionM
...
+        if (event.getType() !== EventType.GroupCallMemberPrefix) {
+            return;
+        }
         const room = this.client.getRoom(event.getRoomId());
         if (!room) {
-            this.logger.error(`Got room state event for unknown room ${event.getRoomId()}!`);
             return;
         }
+        this.refreshRoom(room);
+    };
```
toger5:
```suggestion
                const room = this.client.getRoom(event.getRoomId());
        if (!room || !event.unstableStickyExpiresAt || event.getType() !== EventType.GroupCallMemberPrefix) {
            return; // no room, not sticky or unrelated type = not interested.
        }
        this.refreshRoom(room);
    };
```

src/matrixrtc/MembershipManager.ts
```diff
@@ -997,14 +1017,64 @@ export class MembershipManager
...
+    ) =>
+        this.clientWithSticky._unstable_sendStickyDelayedEvent(
+            this.room.roomId,
+            STICK_DURATION_MS,
+            { delay: this.delayedLeaveEventDelayMs },
+            null,
+            EventType.GroupCallMemberPrefix,
+            { msc4354_sticky_key: this.stateKey },
+        );
+
```
toger5: This method is odd since it takes an empty object as its param and then does not use it.
 - Drop the param
 -  use the param like so

```suggestion
    protected clientSendDelayedEvent: (myMembership: EmptyObject) => Promise<SendDelayedEventResponse> = (
        myMembership,
    ) =>
        this.clientWithSticky._unstable_sendStickyDelayedEvent(
            this.room.roomId,
            STICK_DURATION_MS,
            { delay: this.delayedLeaveEventDelayMs },
            null,
            EventType.GroupCallMemberPrefix,
            {...myMembership, msc4354_sticky_key: this.stateKey },
        );

```

src/matrixrtc/MembershipManager.ts
```diff
@@ -997,14 +1017,64 @@ export class MembershipManager
...
+            MatrixClient,
+            | "getUserId"
+            | "getDeviceId"
+            | "sendStateEvent"
+            | "_unstable_sendDelayedStateEvent"
+            | "_unstable_updateDelayedEvent"
+            | "_unstable_sendStickyEvent"
+            | "_unstable_sendStickyDelayedEvent"
+        >,
+        getOldestMembership: () => CallMembership | undefined,
```
toger5: maybe worth defining a type `MembershipManagerClient`
with 
```
| "getUserId"
| "getDeviceId"
| "sendStateEvent"
| "_unstable_sendDelayedStateEvent"
| "_unstable_updateDelayedEvent"
```

and then doing:

```suggestion
        private readonly clientWithSticky: MembershipManagerClient & Pick<
            | "_unstable_sendStickyEvent"
            | "_unstable_sendStickyDelayedEvent"
        >,
        getOldestMembership: () => CallMembership | undefined,
```

src/matrixrtc/MembershipManager.ts
```diff
@@ -306,7 +325,7 @@ export class MembershipManager
      */
     public constructor(
         private joinConfig: (SessionConfig & MembershipConfig) | undefined,
-        private room: Pick<Room, "getLiveTimeline" | "roomId" | "getVersion">,
+        protected room: Pick<Room, "roomId" | "getVersion">,
         private client: Pick<
             MatrixClient,
             | "getUserId"
```
toger5: This could use the proposed: `MembershipManagerClient` (see StickyEventMembershipManager)

toger5 (commented):
Mostly a couple of ideas + I think the room-sticky-events change does not fit into this one.

toger5 (commented):
Ugh just noticed you already fixed the main part of the review.

src/models/room-sticky-events.ts
```diff
@@ -136,7 +136,6 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
         } else {
             this.unkeyedStickyEvents.add(event as StickyMatrixEvent);
         }
-
         // Recalculate the next expiry time.
         this.nextStickyEventExpiryTs = Math.min(event.unstableStickyExpiresAt, this.nextStickyEventExpiryTs);
```
toger5: lets remove this change as well to get the whole file out of the diff.

Half-Shot: sorry, this was included by mistake.

src/matrixrtc/MembershipManager.ts
Half-Shot: Great idea! done.

src/matrixrtc/MembershipManager.ts
Half-Shot: Lets just drop the param and call this `clientSendDelayedDisconnectMembership`, which makes it more clear what it does.

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -938,6 +939,67 @@ describe("MembershipManager", () => {
             expect(client.sendStateEvent).toHaveBeenCalledTimes(0);
         });
     });
+
+    describe("StickyEventMembershipManager", () => {
```
Half-Shot, line 891: This test suite needs much expanding really. This only touches the joins, we really need to test other behaviours.

src/matrixrtc/MembershipManager.ts
```diff
@@ -1022,14 +1020,68 @@ export class MembershipManager
...
+            this.room.roomId,
+            STICK_DURATION_MS,
+            null,
+            EventType.RTCMembership,
+            { ...myMembership, msc4354_sticky_key: this.memberId },
+        );
     };
+
+    protected actionUpdateFromErrors(e: unknown, t: MembershipActionType, m: string): ActionUpdate | undefined {
+        const mappedMethod = new Map([
```
Half-Shot: I'd just make this a static const, which gives you types for free and stops us creating a new Map each time we call this.

spec/unit/matrixrtc/MembershipManager.spec.ts
toger5: The only thing that this class does really is overwriting send and send delay + makeMyMembership. All this is covered in the join test.
There is also the error logging method... but that not being tested sounds fairly save.

All the rest of the behavior of the Memberhsip Manager stays as it was.

toger5: Is there anything specific you have in mind?

src/matrixrtc/MembershipManager.ts
toger5: yes that makes sense.

src/@types/event.ts
```diff
@@ -338,6 +338,7 @@ export interface TimelineEvents {
     [M_BEACON.name]: MBeaconEventContent;
     [M_POLL_START.name]: PollStartEventContent;
     [M_POLL_END.name]: PollEndEventContent;
+    [EventType.RTCMembership]: RtcMembershipData | { msc4354_sticky_key?: string };
```
BillCarsonFr: Maybe add a TODO to try to use `UnstableValue` here too?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -123,14 +125,6 @@ export function slotDescriptionToId(slotDescription: SlotDescription): string {
 // - we use a `Ms` postfix if the option is a duration to avoid using words like:
 //   `time`, `duration`, `delay`, `timeout`... that might be mistaken/confused with technical terms.
 export interface MembershipConfig {
-    /**
-     * Use the new Manager.
-     *
-     * Default: `false`.
-     * @deprecated does nothing anymore we always default to the new membership manager.
-     */
-    useNewMembershipManager?: boolean;
```
BillCarsonFr, line 132: 🎉

src/matrixrtc/MatrixRTCSession.ts
BillCarsonFr: So you can do both at the same time? Or does it fall back to state on some trigger?
If this is temporary, it might be worth a comment to kill state as soon as event/sticky is deployed enough?

src/matrixrtc/MatrixRTCSession.ts
Half-Shot: This was just deleting it since it did nothing. A bit cheeky but might as well drop it.

spec/unit/matrixrtc/MembershipManager.spec.ts
Half-Shot: I trust your judgement here, it does look like most of it is tested with the join.

src/@types/event.ts
Half-Shot: We should in a separate PR, yea. I'll open one to track it.

Half-Shot: Drafted as https://github.com/matrix-org/matrix-js-sdk/pull/5042

robintown: Is that `|` supposed to be an `&`?

src/matrixrtc/CallMembership.ts
```diff
@@ -195,6 +195,10 @@ export type SessionMembershipData = {
      * something else.
      */
     "m.call.intent"?: RTCCallIntent;
+    /**
+     * the sticky key for sticky events packed application + device_id making up the used slot + device.
```
robintown: I think the wording could be a little clearer:

```suggestion
     * The sticky key in case of a sticky event. This string encodes the application + device_id indicating the used slot + device.
```

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -786,10 +846,28 @@ export class MatrixRTCSession extends TypedEventEmitter<
...
 
+    /**
+     * Call this when a sticky event update has occured.
+     */
+    private readonly onStickyEventUpdate: RoomStickyEventsMap[RoomStickyEventsEvent.Update] = (
+        added,
+        updated,
+        removed,
+    ): void => {
+        this.logger.debug("Sticky event update", { added, updated, removed });
```
robintown: We shouldn't log these as it would leak sensitive event contents in rageshakes and is also quite a lot of data to serialize.

src/matrixrtc/MembershipManager.ts
```diff
@@ -84,6 +89,8 @@ On Leave: ─────────  STOP ALL ABOVE
 (s) Successful restart/resend
 */
 
+const STICK_DURATION_MS = 60 * 60 * 1000; // 60 minutes
```
robintown: This seems in line with the kinds of values that have been made configurable through the `MembershipConfig` in the past. I also can't imagine a great reason why a client might want to configure it, but just wanted to point that out.

Or—oh, is this value intended to represent the *maximum* allowable sticky duration?

src/@types/event.ts
Half-Shot: I think it's meant to be an `|`, as an empty membership may still contain a sticky key if we're using sticky events. Perhaps this needs a comment.

src/matrixrtc/MatrixRTCSession.ts
Half-Shot: Let's change this so we log in recalculateSessionMembers instead.

src/matrixrtc/MembershipManager.ts
Half-Shot: This is the actual time we set all memberships to be sticky for, and I've renamed and documented as such.

src/@types/event.ts
Half-Shot: (I don't think sticky key should ever be optional for a timeline event, either!)

Half-Shot:
```
$ yarn lint:types:src && yarn lint:types:module_system
$ tsc --noEmit --jsx react && tsc --noEmit --jsx react -p playwright
Error: matrix-js-sdk/src/matrixrtc/MatrixRTCSession.ts(53,34): error TS2307: Cannot find module 'src/matrix.ts' or its corresponding type declarations.
Error: matrix-js-sdk/src/matrixrtc/MembershipManager.ts(49,34): error TS2307: Cannot find module 'src/matrix.ts' or its corresponding type declarations.
```

PR #5022, opened 2025-09-30
[MatrixRTC] Multi SFU support + m.rtc.member event type support

toger5:
### What changed

- New event type:
  - Add `m.rtc.member` event type.
  - Add tests covering `m.rtc.member`.
- CallMembership:
  - `CallMembership` can now be constructed from `m.rtc.member` and exposes the same API shape as `m.call.member` to the RTC session.
  - Move “oldest member” selection logic into `CallMembership` (was previously in the membership manager). This removes unnecessary coupling: the `RTCSession` passes the context required by `CallMembership` directly.
- Membership manager (RTC session):
  - Add `useRtcMemberFormat` to control sending `m.rtc.member` instead of `m.call.member`.
    - Note: this may be merged with `useStickyEvent` in a follow-up when we consolidate the rollout.
- Naming cleanups:
  - Rename “focus” to “transport”.
  - Rename “session” to “slot”.
- Active (in use) transport resolution:
  - Remove `getFocusInUse`.
  - Rename `MatrixRTCSession.getActiveFocus` → `MatrixRTCSession.resolveActiveFocus`.
  - Use the new `CallMembership` API to compute the active transport (BREAKING).
- Joining a session:
  - Change the second argument from `activeFocus: Focus` to `multi_sfu_focus: Transport`.
  - Passing a defined `multi_sfu_focus` implies multi-SFU mode; `undefined` falls back to the oldest-membership selection.

### Breaking changes

- MatrixRTCSession:
  - Remove `getFocusInUse()`.
  - Rename `getActiveFocus()` to `resolveActiveFocus()`.
- Join API:
  - Old: `join(sessionId, activeFocus: Focus)`
  - New: `join(sessionId, multi_sfu_focus?: Transport)`
    - Defined `multi_sfu_focus` → use multi-SFU with the provided transport.
    - `undefined` → legacy behavior using the oldest membership.

Todos for consumers updating the PR (currently only known: Element Call):
- Replace calls to `getActiveFocus()` with `resolveActiveFocus()`.
- Remove any usage of `getFocusInUse()`.
- Update callers of `join(...)` to pass `multi_sfu_focus` only when opting into multi-SFU.

### Backward compatibility and rollout

- Reading: `CallMembership` can handle both `m.call.member` and `m.rtc.member`, so existing rooms continue to work.
- Emitting: gated by `useRtcMemberFormat`:
  - `false` (default): continue to emit `m.call.member`.
  - `true`: emit `m.rtc.member`.
- Multi-SFU usage is opt-in via the new `join(sessionId, multi_sfu_focus?: Transport)` signature.

### Checklist

- [x]   Tests written for new code (and old code if feasible).
- [x]  New or updated public/exported symbols have accurate [TSDoc](https://tsdoc.org/) documentation.
- [x]  Linter and other CI checks pass.

### DIFF
--- a/src/matrixrtc/CallMembership.ts
+++ b/src/matrixrtc/CallMembership.ts
@@ -203,7 +203,7 @@ const checkSessionsMembershipData = (data: IContent, errors: string[]): data is 
     if (typeof data.call_id !== "string") errors.push(prefix + "call_id must be string");
     if (typeof data.application !== "string") errors.push(prefix + "application must be a string");
     if (typeof data.focus_active?.type !== "string") errors.push(prefix + "focus_active.type must be a string");
-    if (data.focus_active !== undefined) {
+    if (data.focus_active === undefined) {
         errors.push(prefix + "focus_active has an invalid type");
     }
     if (


src/matrixrtc/IMembershipManager.ts
```diff
@@ -80,10 +79,11 @@ export interface IMembershipManager
     /**
      * Start sending all necessary events to make this user participate in the RTC session.
      * @param fociPreferred the list of preferred foci to use in the joined RTC membership event.
-     * @param fociActive the active focus to use in the joined RTC membership event.
+     * @param multiSfuFocus the active focus to use in the joined RTC membership event. Setting this implies the
+     * membership manager will use multi sfu. Use `undefined` to not use `oldest_membership` selection based sfu.
```
robintown:
```suggestion
     * membership manager will operate in a multi-SFU connection mode. If `undefined`, an `oldest_membership`
     * transport selection will be used instead.
```

src/matrixrtc/LivekitTransport.ts
robintown, line 1: We should probably call this file "LivekitTransport".

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +207,219 @@ const checkSessionsMembershipData = (
...
+        }
     }
 
-    public getPreferredFoci(): Focus[] {
-        return this.membershipData.foci_preferred;
+    /**
+     *
+     * @param oldestMembership For backwards compatibility with session membership (legacy).
+     * @returns
+     */
```
robintown, line 467: This comment needs more details

toger5: Yes this very much benefits from a verbose description to understand the whole system.

spec/unit/matrixrtc/CallMembership.spec.ts
```diff
@@ -94,11 +97,216 @@ describe("CallMembership", () => {
...
+                });
+
+                // if we are the oldest member we use our focus.
+                expect(membership.getTransport(membership)).toStrictEqual(mockFocus);
+
+                // If there is an older member we use its focus.
+                expect(membership.getTransport(oldestMembership)).toBe(membershipTemplate.foci_preferred[0]);
+            });
+
+            it("does not provide focus if the selection method is unknown", () => {
```
Half-Shot: `focus_selection=multi_sfu` is "selection method is unknown"?

spec/unit/matrixrtc/CallMembership.spec.ts
```diff
@@ -94,11 +97,216 @@ describe("CallMembership", () => {
...
+                expect(membership.applicationData).toStrictEqual({ "type": "m.call", "m.call.intent": "voice" });
+            });
+            it("returns correct scope", () => {
+                expect(membership.scope).toBe("m.room");
+            });
+            it("returns correct membershipID", () => {
+                expect(membership.membershipID).toBe("0");
+            });
+            it("returns correct unused fields", () => {
+                expect(membership.getAbsoluteExpiry()).toBe(14400000);
```
Half-Shot: ooi where does this value come from? `membershipTemplate`?

spec/unit/matrixrtc/CallMembership.spec.ts
```diff
@@ -94,11 +97,216 @@ describe("CallMembership", () => {
...
+            expect(() => {
+                new CallMembership(makeMockEvent(), {
+                    ...membershipTemplate,
+                    member: { id: "test", device_id: "test", user_id: "@test:user.id" },
+                });
+            }).not.toThrow();
+        });
+
+        it("considers memberships unexpired if local age low enough", () => {
+            // TODO link prev event
```
Half-Shot, line 304: TODOs?

spec/unit/matrixrtc/CallMembership.spec.ts
```diff
@@ -94,11 +97,216 @@ describe("CallMembership", () => {
...
+                expect(membership.getAbsoluteExpiry()).toBe(14400000);
+                expect(membership.getMsUntilExpiry()).toBe(14400000 - Date.now());
+                expect(membership.isExpired()).toBe(true);
+            });
+        });
+    });
+
+    describe("RtcMembershipData", () => {
+        beforeEach(() => {
+            jest.useFakeTimers();
```
Half-Shot: Do we use these fake timers in this test block?

src/matrixrtc/CallMembership.ts
```diff
@@ -29,6 +30,87 @@ import { type RTCCallIntent } from "./types.ts";
 export const DEFAULT_EXPIRE_DURATION = 1000 * 60 * 60 * 4;
 
 type CallScope = "m.room" | "m.user";
+type Member = { user_id: string; device_id: string; id: string };
```
Half-Shot, line 34: Are we keeping user_id?

src/matrixrtc/CallMembership.ts
```diff
@@ -29,6 +30,87 @@ import { type RTCCallIntent } from "./types.ts";
...
+        rel_type: RelationType.Reference;
+    };
+    "application": {
+        type: string;
+        // other application specific keys
+        [key: string]: any;
+    };
+    "rtc_transports": Transport[];
+    "versions": string[];
+    "sticky_key"?: string;
```
Half-Shot, line 51: should be the unstable msc field?

src/matrixrtc/CallMembership.ts
```diff
@@ -29,6 +30,87 @@ import { type RTCCallIntent } from "./types.ts";
 export const DEFAULT_EXPIRE_DURATION = 1000 * 60 * 60 * 4;
 
 type CallScope = "m.room" | "m.user";
+type Member = { user_id: string; device_id: string; id: string };
+
+export interface RtcMembershipData {
```
Half-Shot, line 36: missing leave_reason?

src/matrixrtc/CallMembership.ts
```diff
@@ -56,13 +138,13 @@ export type SessionMembershipData = {
     /**
      * The focus selection system this user/membership is using.
      */
-    "focus_active": Focus;
+    "focus_active": LivekitFocusSelection;
```
Half-Shot, line 161: ooi, should this be Livekit if it's the generic RTC spec?

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +207,233 @@ const checkSessionsMembershipData = (
     return errors.length === 0;
 };
 
+type MembershipData = { kind: "rtc"; data: RtcMembershipData } | { kind: "session"; data: SessionMembershipData };
+// TODO: Rename to RtcMembership once we removed the legacy SessionMembership from this file.
 export class CallMembership {
     public static equal(a: CallMembership, b: CallMembership): boolean {
+        if (a === undefined || b === undefined) return a === b;
```
Half-Shot: Can this be reflected in the types?

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +207,233 @@ const checkSessionsMembershipData = (
...
-    public get sessionDescription(): SessionDescription {
-        return {
-            application: this.membershipData.application,
-            id: this.membershipData.call_id,
-        };
+    /**
+     * Parsed `slot_id` (format `{application}#{id}`) into its components (application and id).
+     */
+    public get slotDescription(): SlotDescription {
+        return slotIdToDescription(this.slotId);
```
Half-Shot, line 342: I thought there was an issue with slot IDs being opaque IDs and so we can't parse them? Was that resolved?

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +207,233 @@ const checkSessionsMembershipData = (
...
     }
 
+    /** @deprecated scope is not used and will be removed in future versions. replaced by application specific types.*/
     public get scope(): CallScope | undefined {
-        return this.membershipData.scope;
+        const { kind, data } = this.membershipData;
+        switch (kind) {
+            case "rtc":
+                return undefined;
+            default: // "session":
```
Half-Shot: If default is meant to also catch session then this looks nicer to my eye:

```suggestion
            case "session":
            default:
```

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +207,233 @@ const checkSessionsMembershipData = (
...
+                return (this.createdTs() ?? "").toString();
+        }
     }
 
     public createdTs(): number {
-        return this.membershipData.created_ts ?? this.parentEvent.getTs();
+        const { kind, data } = this.membershipData;
+        switch (kind) {
+            case "rtc":
+                // TODO we need to read the referenced (relation) event if available to get the real created_ts
```
Half-Shot, line 396: Should we do this? :)

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +207,233 @@ const checkSessionsMembershipData = (
...
+     * media is published. How this selection happens depends on the `focus_active` field of the session membership.
+     * If the `focus_selection` is `oldest_membership` this will return the transport of the oldest membership
+     * in the room (based on the `created_ts` field of the session membership).
+     * If the `focus_selection` is `multi_sfu` it will return the first transport of the `foci_preferred` list.
+     * (`multi_sfu` is equivalent to how `m.rtc.member` `rtc_transports` work).
+     * @param oldestMembership For backwards compatibility with session membership (legacy). Unused in case of RTC membership.
+     * Always required to make the consumer not care if it deals with RTC or session memberships.
+     * @returns The transport this membership uses to publish media or undefined if no transport is available.
+     */
+    // TODO: make this return all transports used to publish media once this is supported.
```
Half-Shot: If this is going to eventually return multiple, is there any harm in this returning an array?

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +207,233 @@ const checkSessionsMembershipData = (
...
+        if (a === undefined || b === undefined) return a === b;
         return deepCompare(a.membershipData, b.membershipData);
     }
-    private membershipData: SessionMembershipData;
 
+    private membershipData: MembershipData;
+
+    private parentEventData: { eventId: string; sender: string };
     public constructor(
         private parentEvent: MatrixEvent,
```
Half-Shot: Can we document parentEvent? Is it:

1. The current event that contains this membership?
2. The original membership event that joined the call?

src/models/room-member.ts
```diff
@@ -388,7 +388,7 @@ export class RoomMember extends TypedEventEmitter<RoomMemberEvent, RoomMemberEve
     }
 }
 
-const MXID_PATTERN = /@.+:.+/;
+export const MXID_PATTERN = /@[^@:]+:[^@:]+/;
```
Half-Shot: Just to check, we have tests on this right? Quite a fundamental change even if eyeballing it, it looks better.

Also, ports are a valid part of the servername. The second part excludes ports after the colon.

src/matrixrtc/LivekitTransport.ts
```diff
@@ -0,0 +1,46 @@
+/*
+Copyright 2023 New Vector Ltd
```
Half-Shot: 👋 welcome to the future! Please update your copyrights as you step outside the time machine :)

src/matrixrtc/types.ts
```diff
@@ -156,3 +156,11 @@ export type Statistics = {
 
 export const isMyMembership = (m: CallMembership, userId: string, deviceId: string): boolean =>
     m.sender === userId && m.deviceId === deviceId;
+
+/**
+ *  A RTC transport is a JSON object that describes how to connect to a RTC member.
+ */
+export interface Transport {
+    type: string;
```
Half-Shot, line 164: Noting this diverges from the spec as a required field, I have asked :)

Half-Shot (changes requested):
This is a chunky beast and the spec is currently evolving rapidly. I've done my best to review, but I think this may need another pair of eyes to catch more things. Either way, plenty to think about.

spec/unit/matrixrtc/CallMembership.spec.ts
toger5: This is done in: https://github.com/matrix-org/matrix-js-sdk/pull/5031 (It will just be there until we merge the next pr)

src/matrixrtc/CallMembership.ts
toger5: yes, for the potential option of letting an admin control you membeship state.

src/matrixrtc/CallMembership.ts
toger5: in the content? I though in the content we dont use the unstable field?
> The sticky key in the content of the PDU is msc4354_sticky_key. 

It seems you are right.

src/matrixrtc/CallMembership.ts
toger5: This is for the deprecated SessionMembership type -> Will never get used generically with other rtc transports.
So the specific type is acceptable?

src/matrixrtc/CallMembership.ts
toger5: You are right !

src/matrixrtc/CallMembership.ts
toger5: This object based slot identifier (`SlotDescription`) is a relict from when there was no string packed slot_id. This was nice to some extend but does not work with the session_id (now slot id) being used as the state key for the slot moderation events.

It would be a larger change to also change the API here and the MSC guarantees, that splitting by # will only result in array of length 2 so this is save if the events are to spec.

toger5: But, I do like the object based appraoch inside the js-sdk codebase. It is used in a lot of places -> keeping diff small.

src/matrixrtc/CallMembership.ts
toger5: just skipping the break... yes that might be the way to go.

toger5: updated all places where this occurs.

src/matrixrtc/CallMembership.ts
toger5: also done in: https://github.com/matrix-org/matrix-js-sdk/pull/5031

src/matrixrtc/CallMembership.ts
toger5: I updated the doc comment to
>  Gets the **primary** transport to use for this RTC membership (m.rtc.member).

I removed the TODO since there also is: `public get transports(): Transport[] ` Which solves getting all the transports.

`getTransport` can get depreacted eventually once we do not use a selection based method anymore.
(added a comment about that as well)

src/models/room-member.ts
toger5: Thinking about it it looks too risky. Lets not change it.
The MXID check also is not super important anywayas (if there is no matching room member object it will get discarded anyways)

src/matrixrtc/types.ts
toger5: The spec makes thi optional?
> type (string, required) - this MUST be "livekit_multi_sfu"

is what i see in the spec.

src/matrixrtc/CallMembership.ts
toger5: renaming it to matrixEvent makes more sense i think.

src/matrixrtc/CallMembership.ts
toger5: I looked into this for a bit. It turns out we currently do not have any support for left memberships. Indirectly we of course update our current memberships by not including them. But we never store left memberships. see: `public static sessionMembershipsForSlot`
So the leave_reason is not part of the interface which only represents a connected membership.

src/matrixrtc/CallMembership.ts
toger5: updated

spec/unit/matrixrtc/CallMembership.spec.ts
toger5: good catch. removed before and after block.

src/models/room-member.ts
toger5: reverted this change.

spec/unit/matrixrtc/CallMembership.spec.ts
toger5: its DEFAULT_EXPIRE_DURATION.
Lets use this in the test

spec/unit/matrixrtc/CallMembership.spec.ts
toger5: good catch. the "unknown" test was missing and this test should be called:
`gets the correct active transport with multi_sfu`

toger5: added the missing test and renamed this one.

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +209,252 @@ const checkSessionsMembershipData = (
...
+
+    /** The parsed data from the Matrix event.
+     * To access checked eventId and sender from the matrixEvent.
+     * Class construction will fail if these values cannot get obtained. */
+    private matrixEventData: { eventId: string; sender: string };
     public constructor(
-        private parentEvent: MatrixEvent,
+        /** The Matrix event that this membership is based on */
+        private matrixEvent: MatrixEvent,
         data: any,
```
robintown: We should change this type to something that reflects what we actually know about `data` (probably very little, so likely `object` or `unknown`) rather than `any` which disables type checks.

src/matrixrtc/CallMembership.ts
```diff
@@ -29,6 +30,89 @@ import { type RTCCallIntent } from "./types.ts";
...
+): data is RtcMembershipData => {
+    const prefix = "Malformed rtc membership event: ";
+
+    // required fields
+    if (typeof data.slot_id !== "string") errors.push(prefix + "slot_id must be string");
+    if (typeof data.member !== "object" || data.member === null) {
+        errors.push(prefix + "member must be an object");
+    } else {
+        if (typeof data.member.user_id !== "string") errors.push(prefix + "member.user_id must be string");
+        else if (!MXID_PATTERN.test(data.member.user_id)) errors.push(prefix + "member.user_id must be a valid mxid");
```
robintown, line 71: Rather than checking against a regex, we could (and incidentally need to) check that the `user_id` matches the event's `sender`.

src/matrixrtc/CallMembership.ts
```diff
@@ -29,6 +30,89 @@ import { type RTCCallIntent } from "./types.ts";
...
+    "slot_id": string;
+    "member": Member;
+    "m.relates_to"?: {
+        event_id: string;
+        rel_type: RelationType.Reference;
+    };
+    "application": {
+        type: string;
+        // other application specific keys
+        [key: string]: any;
```
robintown:
```suggestion
        [key: string]: unknown;
```

src/matrixrtc/CallMembership.ts
```diff
@@ -29,6 +30,89 @@ import { type RTCCallIntent } from "./types.ts";
...
+    };
+    "rtc_transports": Transport[];
+    "versions": string[];
+    "msc4354_sticky_key"?: string;
+    "sticky_key"?: string;
+    /**
+     * The intent of the call from the perspective of this user. This may be an audio call, video call or
+     * something else.
+     */
+    "m.call.intent"?: RTCCallIntent;
```
robintown: Worth a comment why we're including this field in an otherwise non-application-specific type. (Convenience, I guess?)

src/matrixrtc/CallMembership.ts
```diff
@@ -29,6 +30,89 @@ import { type RTCCallIntent } from "./types.ts";
...
+        errors.push(prefix + "application must be an object");
+    } else {
+        if (typeof data.application.type !== "string") errors.push(prefix + "application.type must be a string");
+    }
+    if (data.rtc_transports === undefined || !Array.isArray(data.rtc_transports)) {
+        errors.push(prefix + "rtc_transports must be an array");
+    } else {
+        // validate that each transport has at least a string 'type'
+        for (const t of data.rtc_transports) {
+            if (typeof t !== "object" || typeof (t as any).type !== "string") {
```
robintown:
```suggestion
            if (typeof t !== "object" || t === null || typeof t.type !== "string") {
```

src/matrixrtc/CallMembership.ts
```diff
@@ -29,6 +30,89 @@ import { type RTCCallIntent } from "./types.ts";
...
+        }
+    }
+    if (data.versions === undefined || !Array.isArray(data.versions)) {
+        errors.push(prefix + "versions must be an array");
+    } else if (!data.versions.every((v) => typeof v === "string")) {
+        errors.push(prefix + "versions must be an array of strings");
+    }
+
+    // optional fields
+    const stickyKey = data.sticky_key ?? data.msc4354_sticky_key;
```
robintown: We should validate `sticky_key` and `msc4354_sticky_key` separately to avoid an object like the following passing the checks:

```typescript
{ sticky_key: "key", msc4354_sticky_key: 1 }
```

Alternatively, we could drop those two fields from the type seeing as they aren't referenced anywhere currently.

src/matrixrtc/CallMembership.ts
```diff
@@ -104,7 +188,12 @@ const checkSessionsMembershipData = (
     if (typeof data.call_id !== "string") errors.push(prefix + "call_id must be string");
     if (typeof data.application !== "string") errors.push(prefix + "application must be a string");
     if (typeof data.focus_active?.type !== "string") errors.push(prefix + "focus_active.type must be a string");
-    if (!Array.isArray(data.foci_preferred)) errors.push(prefix + "foci_preferred must be an array");
+    if (data.focus_active !== undefined && !isLivekitFocusSelection(data.focus_active)) {
+        errors.push(prefix + "focus_active has an invalid type");
+    }
+    if (data.foci_preferred !== undefined && !Array.isArray(data.foci_preferred)) {
+        errors.push(prefix + "foci_preferred must be an array");
+    }
```
robintown: Should also verify that they are objects with property `type: string`

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +209,252 @@ const checkSessionsMembershipData = (
     return errors.length === 0;
 };
 
+type MembershipData = { kind: "rtc"; data: RtcMembershipData } | { kind: "session"; data: SessionMembershipData };
+// TODO: Rename to RtcMembership once we removed the legacy SessionMembership from this file.
 export class CallMembership {
-    public static equal(a: CallMembership, b: CallMembership): boolean {
+    public static equal(a?: CallMembership, b?: CallMembership): boolean {
+        if (a === undefined || b === undefined) return a === b;
```
robintown: `deepCompare` does this for free

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +209,252 @@ const checkSessionsMembershipData = (
...
         return deepCompare(a.membershipData, b.membershipData);
     }
-    private membershipData: SessionMembershipData;
 
+    private membershipData: MembershipData;
+
+    /** The parsed data from the Matrix event.
+     * To access checked eventId and sender from the matrixEvent.
+     * Class construction will fail if these values cannot get obtained. */
+    private matrixEventData: { eventId: string; sender: string };
```
robintown:
```suggestion
    private readonly matrixEventData: { eventId: string; sender: string };
```

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +209,252 @@ const checkSessionsMembershipData = (
...
     public constructor(
-        private parentEvent: MatrixEvent,
+        /** The Matrix event that this membership is based on */
+        private matrixEvent: MatrixEvent,
```
robintown:
```suggestion
        private readonly matrixEvent: MatrixEvent,
```

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +209,252 @@ const checkSessionsMembershipData = (
...
+        if (checkSessionsMembershipData(data, sessionErrors)) {
+            this.membershipData = { kind: "session", data };
+        } else if (checkRtcMembershipData(data, rtcErrors)) {
+            this.membershipData = { kind: "rtc", data };
+        } else {
             throw Error(
-                `unknown CallMembership data. Does not match MSC4143 call.member (${sessionErrors.join(" & ")}) events this could be a legacy membership event: (${data})`,
+                `unknown CallMembership data.` +
+                    `Does not match MSC4143 call.member (${sessionErrors.join(" & ")})\n` +
+                    `Does not match MSC4143 rtc.member (${rtcErrors.join(" & ")})\n` +
```
robintown: These strings could add up in length pretty quickly and clog up the logs. Maybe it's worth having a condition based on `sessionErrors.length < rtcErrors.length` to decide what the event was more likely to be & therefore which one is more interesting to log.

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +209,252 @@ const checkSessionsMembershipData = (
...
+        const sender = matrixEvent.getSender();
+
+        if (eventId === undefined) throw new Error("parentEvent is missing eventId field");
+        if (sender === undefined) throw new Error("parentEvent is missing sender field");
+        this.matrixEventData = { eventId, sender };
     }
 
-    public get sender(): string | undefined {
-        return this.parentEvent.getSender();
+    public get sender(): string {
```
robintown, line 277: This seems like a good time to rename this to `userId` - looks like the purpose is to get the (validated) user ID of the membership, which is equal to the sender at the moment only by coincidence (we don't yet have a feature that allows mods to set your membership on your behalf).

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +209,252 @@ const checkSessionsMembershipData = (
...
 
-    public get eventId(): string | undefined {
-        return this.parentEvent.getId();
+    public get eventId(): string {
+        return this.matrixEventData.eventId;
     }
 
     /**
-     *  @deprecated Use sessionDescription.id instead.
+     * The slot id to find all member building one session `slot_id` (format `{application}#{id}`).
```
robintown:
```suggestion
     * The ID of the MatrixRTC slot that this membership belongs to (format `{application}#{id}`).
```

src/matrixrtc/CallMembership.ts
```diff
@@ -29,6 +30,89 @@ import { type RTCCallIntent } from "./types.ts";
...
+    } else {
+        if (typeof data.member.user_id !== "string") errors.push(prefix + "member.user_id must be string");
+        else if (!MXID_PATTERN.test(data.member.user_id)) errors.push(prefix + "member.user_id must be a valid mxid");
+        if (typeof data.member.device_id !== "string") errors.push(prefix + "member.device_id must be string");
+        if (typeof data.member.id !== "string") errors.push(prefix + "member.id must be string");
+    }
+    if (typeof data.application !== "object" || data.application === null) {
+        errors.push(prefix + "application must be an object");
+    } else {
+        if (typeof data.application.type !== "string") errors.push(prefix + "application.type must be a string");
```
robintown: It additionally must not contain '#'

src/matrixrtc/CallMembership.ts
robintown: Be aware that I've picked Matthew's [discussion about the slot ID grammar](https://docs.google.com/document/d/1OwMzBkEWNjqBpfqimubQ79gPzQmIXO0hCz3AbJRSFRE/edit?disco=AAABssHkYGk) back up, so there could be movement on that still.

src/matrixrtc/CallMembership.ts
```diff
@@ -120,109 +209,252 @@ const checkSessionsMembershipData = (
...
+        const { kind, data } = this.membershipData;
+        switch (kind) {
+            case "rtc":
+                return data.application.type;
+            case "session":
+            default:
+                return data.application;
+        }
+    }
+    public get applicationData(): { type: string } & Record<string, any> {
```
robintown:
```suggestion
    public get applicationData(): { type: string, [key: string]: unknown } {
```

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -103,10 +103,17 @@ export interface SessionConfig {
 /**
  * The session description is used to identify a session. Used in the state event.
  */
-export interface SessionDescription {
+export interface SlotDescription {
     id: string;
     application: string;
 }
+export function slotIdToDescription(slotId: string): SlotDescription {
+    const [application, id] = slotId.split("#");
```
robintown, line 111: String.split can result in an array of arbitrarily many elements, so I think this needs String.indexOf instead. (We want to split on just the first instance.)

```typescript
"a#b#c".split("#") // ["a", "b", "c"]
```

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -279,33 +287,53 @@ export class MatrixRTCSession extends TypedEventEmitter<
...
+     * @deprecated use `slotId` instead.
      */
     public get callId(): string | undefined {
-        return this._callId;
+        return this.slotDescription?.id;
+    }
+    /**
+     * The slotId of the call.
+     * `{application}#{appSpecificId}`
+     * It can be undefined since the slotId is only known once the first membership joins.
```
robintown, line 296: Can we do something about that? I see that the constructor of MatrixRTCSession takes a `SlotDescription` as a required argument, so it seems to me that the slot is always defined.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -335,12 +363,16 @@ export class MatrixRTCSession extends TypedEventEmitter<
...
+                    continue;
+                }
                 try {
                     const membership = new CallMembership(memberEvent, membershipData);
 
-                    if (!deepCompare(membership.sessionDescription, sessionDescription)) {
+                    if (!deepCompare(membership.slotDescription, slotDescription)) {
                         logger.info(
-                            `Ignoring membership of user ${membership.sender} for a different session:  ${JSON.stringify(membership.sessionDescription)}`,
+                            `Ignoring membership of user ${membership.sender} for a different session:  ${JSON.stringify(membership.slotDescription)}`,
```
robintown:
```suggestion
                            `Ignoring membership of user ${membership.sender} for a different slot:  ${JSON.stringify(membership.slotDescription)}`,
```

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -445,13 +480,13 @@ export class MatrixRTCSession extends TypedEventEmitter<
         public memberships: CallMembership[],
         /**
          * The session description is used to define the exact session this object is tracking.
-         * A session is distinct from another session if one of those properties differ: `roomSubset.roomId`, `sessionDescription.application`, `sessionDescription.id`.
+         * A session is distinct from another session if one of those properties differ: `roomSubset.roomId`, `slotDescription.application`, `slotDescription.id`.
          */
-        public readonly sessionDescription: SessionDescription,
+        public readonly slotDescription: SlotDescription,
```
robintown, line 484: The doc comment should be updated to no longer mention "session description." I like this language from the MSC:

> a virtual address where participants are allowed to meet

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -608,12 +642,18 @@ export class MatrixRTCSession extends TypedEventEmitter<
...
-     * Get the active focus from the current CallMemberState event
+     * Get the focus in use from a specific specified member.
+     * @param member The member for which to get the active focus. If undefined, the own membership is used.
      * @returns The focus that is currently in use to connect to this session. This is undefined
      * if the client is not connected to this session.
+     * @deprecated use `member.getTransport(session.getOldestMembership())` instead if you want to get the active transport for a specific member.
      */
-    public getActiveFocus(): Focus | undefined {
-        return this.membershipManager?.getActiveFocus();
+    public resolveActiveFocus(member?: CallMembership): Transport | undefined {
```
robintown: Since I introduced this method during the initial commits to the multi-SFU branch, it's probably fine to delete it rather than deprecating it. If anything what could get kept around and marked as deprecated are `getActiveFocus` and `getFocusInUse`, somehow.

src/matrixrtc/MatrixRTCSessionManager.ts
robintown, line 1: Definitely not a task for this PR, but I'll note that particularly with the introduction of slots, we may want to rethink the role of the MatrixRTCSessionManager going forward.

Currently "one `MatrixRTCSessionManager` is required for each MatrixRTC sessionDescription (application, session id) that the client wants to support" - but with us designing MatrixRTC such that clients of the future can add and remove slots on demand, do we really want clients to have to implement their own management of MatrixRTCSessionManager objects? :face_with_spiral_eyes: (It would be non-trivial to manage them, as each Manager object handles a single slot ID across *all rooms*, so you need to ensure *one* Manager exists for each slot ID used in 1 or more rooms, if you want to support arbitrary slots…)

A more convenient design from my perspective would be for a single MatrixRTCSessionManager object to be responsible for all rooms and all slot IDs, storing a nested map of type `Map<string, Map<string, MatrixRTCSession>>` (room ID → slot ID → session). Which brings me back to https://github.com/matrix-org/matrix-js-sdk/issues/5033 :smiley:

src/matrixrtc/MatrixRTCSessionManager.ts
```diff
@@ -56,7 +56,7 @@ export class MatrixRTCSessionManager extends TypedEventEmitter<MatrixRTCSessionM
     public constructor(
         rootLogger: Logger,
         private client: MatrixClient,
-        private readonly sessionDescription: SessionDescription = { id: "", application: "m.call" }, // Default to the Matrix Call application
+        private readonly sessionDescription: SlotDescription = { application: "m.call", id: "" }, // Default to the Matrix Call application
```
robintown:
```suggestion
        private readonly slotDescription: SlotDescription = { application: "m.call", id: "" }, // Default to the Matrix Call application
```

src/matrixrtc/MembershipManager.ts
```diff
@@ -325,7 +312,7 @@ export class MembershipManager
         if (userId === null) throw Error("Missing userId in client");
         if (deviceId === null) throw Error("Missing deviceId in client");
         this.deviceId = deviceId;
-        this.stateKey = this.makeMembershipStateKey(userId, deviceId);
+        this.memberId = this.makeMembershipStateKey(userId, deviceId);
```
robintown: Is this meant to be a state key or a member ID? It looks like this is used as both, but I was under the impression that member IDs were supposed to be unique to each period of connection to the slot.

src/matrixrtc/MembershipManager.ts
```diff
@@ -757,20 +753,42 @@ export class MembershipManager
...
-            ...(hasPreviousEvent ? { created_ts: this.ownMembership?.createdTs() } : undefined),
-        };
+    private makeMyMembership(expires: number): SessionMembershipData | RtcMembershipData {
+        const ownMembership = this.ownMembership;
+        if (this.useRtcMemberFormat) {
+            const relationObject = ownMembership?.eventId
+                ? { "m.relation": { rel_type: RelationType.Reference, event_id: ownMembership?.eventId } }
+                : {};
+            return {
+                application: { type: this.slotDescription.application, id: this.slotDescription.id },
```
robintown: I don't see this `id` field in any of the MSCs.

robintown (commented):
Finally digging in to give this a proper review, here goes:

src/matrixrtc/CallMembership.ts
toger5: I think we decided agains this for potential moderated control of rtc participants (teacher student setting)

But probably a spec discussion which can happen out of band.

src/matrixrtc/MatrixRTCSession.ts
toger5: If we do only allow memberships that obay the "no #" rule as you proposed above this should be fine. But needs a comment.

src/matrixrtc/MatrixRTCSessionManager.ts
toger5: Yes. This has been on the back of my head since quiet some time. (replacing the sessionManager with sth better suited to the js-sdk)
The nested maps also help a lot to find a good solution.

src/matrixrtc/MembershipManager.ts
toger5: yes this is not to spec. It tries to be compatible with our previous key strinc concat. (which would work if there was not the requirement that each ne connect has a unique member id.)

So we loose proper history computation. This can be acceptable for the first version. What do you think? Making this a uuid (or adding a ts at the end of the makeMembershipStateKey) should also just work.

src/matrixrtc/MembershipManager.ts
toger5: Also worth a discussion. It is the `ApplicationSpecificEnum` value it could just be part of `m.call`. For now it imght make sense to just drop it. its "" in all cases anyway.

src/matrixrtc/CallMembership.ts
toger5: also added a similar check to the slot_id

src/matrixrtc/CallMembership.ts
toger5: IContent seems to be appropriate since this is what we will pass into on construction.

src/matrixrtc/MembershipManager.ts
robintown: Yeah, and you'd have to validate that it matches the data as seen in `slot_id` if you wanted to use it for anything critical… it feels redundant.

src/matrixrtc/CallMembership.ts
toger5: But this will also be updated with the related event PR.

src/matrixrtc/MembershipManager.ts
robintown: > So we loose proper history computation. This can be acceptable for the first version. What do you think?

Sounds fine as long as there's a TODO comment

> Making this a uuid (or adding a ts at the end of the makeMembershipStateKey) should also just work.

Mhm, but the value used for the state key of state events does need to remain how it is now (the format is encoded in other places for widget capabilities)

src/matrixrtc/MatrixRTCSessionManager.ts
robintown: https://github.com/matrix-org/matrix-js-sdk/issues/5035

src/matrixrtc/CallMembership.ts
robintown: I think this does need some kind of authorization either way. So if we want a solution in which moderated control is possible, we probably need to know the auth specifics of that too - otherwise we're in a wild west of anybody being able to control anyone else's membership.

(It's worth noting that the existing implementation isn't quite in that same wild west, even without the room auth checks for user-prefixed state keys, since we make decisions based on the event `sender`. The worst that can happen is someone can disconnect you from a call, so this would be a regression.)

robintown: Resolved for now by matching `user_id` with `sender` (spec discussion still eventually needed)

src/matrixrtc/CallMembership.ts
```diff
@@ -48,40 +48,48 @@ export interface RtcMembershipData {
...
 ): data is RtcMembershipData => {
-    const prefix = "Malformed rtc membership event: ";
+    const prefix = " - ";
 
     // required fields
-    if (typeof data.slot_id !== "string") errors.push(prefix + "slot_id must be string");
+    if (typeof data.slot_id !== "string") {
+        errors.push(prefix + "slot_id must be string");
+    } else {
+        if (data.slot_id.split("#").length !== 2) errors.push(prefix + 'slot_id must include exactly one "#"');
```
robintown, line 65: I'll just note that the spec doesn't technically forbid # in the latter half of the slot ID (but we surely won't be trying to use one any time soon either).

src/matrixrtc/MembershipManager.ts
toger5: yes that is why we need to keep it like this for compatibility. I added a comment.

src/matrixrtc/CallMembership.ts
```diff
@@ -179,20 +196,21 @@ export type SessionMembershipData = {
...
     if (typeof data.focus_active?.type !== "string") errors.push(prefix + "focus_active.type must be a string");
     if (data.focus_active !== undefined && !isLivekitFocusSelection(data.focus_active)) {
         errors.push(prefix + "focus_active has an invalid type");
     }
-    if (data.foci_preferred !== undefined && !Array.isArray(data.foci_preferred)) {
-        errors.push(prefix + "foci_preferred must be an array");
+    if (
+        data.foci_preferred !== undefined &&
+        !Array.isArray(data.foci_preferred) &&
+        !data.foci_preferred.every((f: Transport) => typeof f === "object" && f !== null && typeof f.type === "string")
```
robintown:
```suggestion
        !(Array.isArray(data.foci_preferred) &&
        data.foci_preferred.every((f: Transport) => typeof f === "object" && f !== null && typeof f.type === "string"))
```

src/matrixrtc/CallMembership.ts
robintown: Oh, I misread, that was to help with accessing `.membershipData`… whatever, works either way

src/matrixrtc/MatrixRTCSession.ts
robintown: Resolved by forbidding # in the latter half of the slot ID too

src/matrixrtc/CallMembership.ts
toger5: Do you think we should forbid it in the spec?

robintown: I stand corrected, it already is forbidden in the spec:

> The # character MUST NOT be used in either application.type or ApplicationSpecificSlotEnum

src/matrixrtc/CallMembership.ts
toger5: What is this doing? its all && conditions so should not make a difference?

src/matrixrtc/CallMembership.ts
robintown: Resolving as: agree, we should get `leave_reason` support eventually but let's kick that down the road.

src/matrixrtc/CallMembership.ts
robintown: Sounds fine to me at least.

src/matrixrtc/CallMembership.ts
robintown: For now we're sticking with the grammar (which is enforced by the `checkRtcMembershipData` function). We should check again with Matthew if he's happy with that but for now it remains the spec.

src/matrixrtc/types.ts
robintown: It's required.

src/matrixrtc/CallMembership.ts
robintown: Without my change, `["foo"]` would not be caught by the check. `!a && !b` is the same as `!(a || b)`, which is a different thing from `!(a && b)`

toger5: of course... Somehow missed the !... ugh

PR #5028, opened 2025-10-01
Implement Sticky Events MSC4354

Half-Shot:
Split out from https://github.com/matrix-org/matrix-js-sdk/pull/5017 to allow for easy reviewing. Implements https://github.com/matrix-org/matrix-spec-proposals/pull/4354

As a brief overview of the MSC (though please do at least skim the MSC): This allows any member in a room to mark an event as "sticky" and ensure it gets delivered to other users *even* in the case of a gappy sync. This is useful for things like RTC where you want call member events to always reach clients, without having to bloat the room with state events that will contribute to overall DAG bloat. 

This implementation also deals with ensuring that some events will replace predecessors that share the same "sticky key", and emit the appropriate notices to users of the API.

## Checklist

### DIFF
--- a/src/@types/requests.ts
+++ b/src/@types/requests.ts
@@ -107,6 +107,9 @@ export type SendActionDelayedEventRequestOpts = ParentDelayId;
 
 export type SendDelayedEventRequestOpts = SendTimeoutDelayedEventRequestOpts | SendActionDelayedEventRequestOpts;
 
+export function isSendDelayedEventRequestOpts(opts: object): opts is SendDelayedEventRequestOpts {
+    return (opts as TimeoutDelay).delay !== undefined || (opts as ParentDelayId).parent_delay_id !== undefined;
+}
 export type SendDelayedEventResponse = {
     delay_id: string;
 };
--- a/src/@types/requests.ts
+++ b/src/@types/requests.ts
@@ -94,21 +94,19 @@ export interface ISendEventResponse {
     event_id: string;
 }
 
-export type TimeoutDelay = {
-    delay: number;
-};
-
-export type ParentDelayId = {
-    parent_delay_id: string;
-};
-
-export type SendTimeoutDelayedEventRequestOpts = TimeoutDelay & Partial<ParentDelayId>;
-export type SendActionDelayedEventRequestOpts = ParentDelayId;
-
-export type SendDelayedEventRequestOpts = SendTimeoutDelayedEventRequestOpts | SendActionDelayedEventRequestOpts;
+export type SendDelayedEventRequestOpts = { parent_delay_id: string } | { delay: number; parent_delay_id?: string };
 
 export function isSendDelayedEventRequestOpts(opts: object): opts is SendDelayedEventRequestOpts {
-    return (opts as TimeoutDelay).delay !== undefined || (opts as ParentDelayId).parent_delay_id !== undefined;
+    if ("parent_delay_id" in opts && typeof opts.parent_delay_id !== "string") {
+        // Invalid type, reject
+        return false;
+    }
+    if ("delay" in opts && typeof opts.delay !== "number") {
+        // Invalid type, reject.
+        return true;
+    }
+    // At least one of these fields must be specified.
+    return "delay" in opts || "parent_delay_id" in opts;
 }
 export type SendDelayedEventResponse = {
     delay_id: string;
--- a/src/client.ts
+++ b/src/client.ts
@@ -3419,10 +3419,11 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 
     /**
-     * Send a delayed timeline event.
+     * Send a delayed sticky timeline event.
      *
      * Note: This endpoint is unstable, and can throw an `Error`.
-     *   Check progress on [MSC4140](https://github.com/matrix-org/matrix-spec-proposals/pull/4140) for more details.
+     *   Check progress on [MSC4140](https://github.com/matrix-org/matrix-spec-proposals/pull/4140) and
+     *   [MSC4354](https://github.com/matrix-org/matrix-spec-proposals/pull/4354) for more details.
      */
     // eslint-disable-next-line
     public async _unstable_sendStickyDelayedEvent<K extends keyof TimelineEvents>(
@@ -3434,6 +3435,12 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         content: TimelineEvents[K] & { msc4354_sticky_key: string },
         txnId?: string,
     ): Promise<SendDelayedEventResponse> {
+        if (!(await this.doesServerSupportUnstableFeature(UNSTABLE_MSC4140_DELAYED_EVENTS))) {
+            throw new UnsupportedDelayedEventsEndpointError(
+                "Server does not support the delayed events API",
+                "getDelayedEvents",
+            );
+        }
         if (!(await this.doesServerSupportUnstableFeature(UNSTABLE_MSC4354_STICKY_EVENTS))) {
             throw new UnsupportedStickyEventsEndpointError(
                 "Server does not support the sticky events",
@@ -3446,7 +3453,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             roomId,
             threadId,
             eventObject: { type: eventType, content },
-            queryDict: { msc4354_stick_duration_ms: stickDuration },
+            queryDict: { "org.matrix.msc4354.sticky_duration_ms": stickDuration },
             delayOpts,
             txnId,
         });
@@ -3487,10 +3494,10 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 
     /**
-     * Send a delayed timeline event.
+     * Send a sticky timeline event.
      *
      * Note: This endpoint is unstable, and can throw an `Error`.
-     *   Check progress on [MSC4140](https://github.com/matrix-org/matrix-spec-proposals/pull/4140) for more details.
+     *   Check progress on [MSC4354](https://github.com/matrix-org/matrix-spec-proposals/pull/4354) for more details.
      */
     // eslint-disable-next-line
     public async _unstable_sendStickyEvent<K extends keyof TimelineEvents>(
@@ -3513,7 +3520,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             roomId,
             threadId,
             eventObject: { type: eventType, content },
-            queryDict: { msc4354_stick_duration_ms: stickDuration },
+            queryDict: { "org.matrix.msc4354.sticky_duration_ms": stickDuration },
             txnId,
         });
     }
--- a/src/client.ts
+++ b/src/client.ts
@@ -105,6 +105,7 @@ import {
 import { RoomMemberEvent, type RoomMemberEventHandlerMap } from "./models/room-member.ts";
 import { type IPowerLevelsContent, type RoomStateEvent, type RoomStateEventHandlerMap } from "./models/room-state.ts";
 import {
+    isSendDelayedEventRequestOpts,
     type DelayedEventInfo,
     type IAddThreePidOnlyBody,
     type IBindThreePidBody,
@@ -246,7 +247,7 @@ import {
     validateAuthMetadataAndKeys,
 } from "./oidc/index.ts";
 import { type EmptyObject } from "./@types/common.ts";
-import { UnsupportedDelayedEventsEndpointError } from "./errors.ts";
+import { UnsupportedDelayedEventsEndpointError, UnsupportedStickyEventsEndpointError } from "./errors.ts";
 
 export type Store = IStore;
 
@@ -545,6 +546,7 @@ export const UNSTABLE_MSC2666_MUTUAL_ROOMS = "uk.half-shot.msc2666.mutual_rooms"
 export const UNSTABLE_MSC2666_QUERY_MUTUAL_ROOMS = "uk.half-shot.msc2666.query_mutual_rooms";
 
 export const UNSTABLE_MSC4140_DELAYED_EVENTS = "org.matrix.msc4140";
+export const UNSTABLE_MSC4354_STICKY_EVENTS = "org.matrix.msc4354";
 
 export const UNSTABLE_MSC4133_EXTENDED_PROFILES = "uk.tcpip.msc4133";
 export const STABLE_MSC4133_EXTENDED_PROFILES = "uk.tcpip.msc4133.stable";
@@ -2672,7 +2674,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         }
 
         this.addThreadRelationIfNeeded(content, threadId, roomId);
-        return this.sendCompleteEvent(roomId, threadId, { type: eventType, content }, txnId);
+        return this.sendCompleteEvent({ roomId, threadId, eventObject: { type: eventType, content }, txnId });
     }
 
     /**
@@ -2709,12 +2711,13 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @returns Promise which resolves: to an empty object `{}`
      * @returns Rejects: with an error response.
      */
-    private sendCompleteEvent(
-        roomId: string,
-        threadId: string | null,
-        eventObject: Partial<IEvent>,
-        txnId?: string,
-    ): Promise<ISendEventResponse>;
+    private sendCompleteEvent(params: {
+        roomId: string;
+        threadId: string | null;
+        eventObject: Partial<IEvent>;
+        queryDict?: QueryDict;
+        txnId?: string;
+    }): Promise<ISendEventResponse>;
     /**
      * Sends a delayed event (MSC4140).
      * @param eventObject - An object with the partial structure of an event, to which event_id, user_id, room_id and origin_server_ts will be added.
@@ -2723,29 +2726,29 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @returns Promise which resolves: to an empty object `{}`
      * @returns Rejects: with an error response.
      */
-    private sendCompleteEvent(
-        roomId: string,
-        threadId: string | null,
-        eventObject: Partial<IEvent>,
-        delayOpts: SendDelayedEventRequestOpts,
-        txnId?: string,
-    ): Promise<SendDelayedEventResponse>;
-    private sendCompleteEvent(
-        roomId: string,
-        threadId: string | null,
-        eventObject: Partial<IEvent>,
-        delayOptsOrTxnId?: SendDelayedEventRequestOpts | string,
-        txnIdOrVoid?: string,
-    ): Promise<ISendEventResponse | SendDelayedEventResponse> {
-        let delayOpts: SendDelayedEventRequestOpts | undefined;
-        let txnId: string | undefined;
-        if (typeof delayOptsOrTxnId === "string") {
-            txnId = delayOptsOrTxnId;
-        } else {
-            delayOpts = delayOptsOrTxnId;
-            txnId = txnIdOrVoid;
-        }
-
+    private sendCompleteEvent(params: {
+        roomId: string;
+        threadId: string | null;
+        eventObject: Partial<IEvent>;
+        delayOpts: SendDelayedEventRequestOpts;
+        queryDict?: QueryDict;
+        txnId?: string;
+    }): Promise<SendDelayedEventResponse>;
+    private sendCompleteEvent({
+        roomId,
+        threadId,
+        eventObject,
+        delayOpts,
+        queryDict,
+        txnId,
+    }: {
+        roomId: string;
+        threadId: string | null;
+        eventObject: Partial<IEvent>;
+        delayOpts?: SendDelayedEventRequestOpts;
+        queryDict?: QueryDict;
+        txnId?: string;
+    }): Promise<SendDelayedEventResponse | ISendEventResponse> {
         if (!txnId) {
             txnId = this.makeTxnId();
         }
@@ -2788,7 +2791,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
 
         const type = localEvent.getType();
         this.logger.debug(
-            `sendEvent of type ${type} in ${roomId} with txnId ${txnId}${delayOpts ? " (delayed event)" : ""}`,
+            `sendEvent of type ${type} in ${roomId} with txnId ${txnId}${delayOpts ? " (delayed event)" : ""}${queryDict ? " query params: " + JSON.stringify(queryDict) : ""}`,
         );
 
         localEvent.setTxnId(txnId);
@@ -2806,9 +2809,9 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
                 return Promise.reject(new Error("Event blocked by other events not yet sent"));
             }
 
-            return this.encryptAndSendEvent(room, localEvent);
+            return this.encryptAndSendEvent(room, localEvent, queryDict);
         } else {
-            return this.encryptAndSendEvent(room, localEvent, delayOpts);
+            return this.encryptAndSendEvent(room, localEvent, delayOpts, queryDict);
         }
     }
 
@@ -2816,7 +2819,11 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * encrypts the event if necessary; adds the event to the queue, or sends it; marks the event as sent/unsent
      * @returns returns a promise which resolves with the result of the send request
      */
-    protected async encryptAndSendEvent(room: Room | null, event: MatrixEvent): Promise<ISendEventResponse>;
+    protected async encryptAndSendEvent(
+        room: Room | null,
+        event: MatrixEvent,
+        queryDict?: QueryDict,
+    ): Promise<ISendEventResponse>;
     /**
      * Simply sends a delayed event without encrypting it.
      * TODO: Allow encrypted delayed events, and encrypt them properly
@@ -2827,16 +2834,20 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         room: Room | null,
         event: MatrixEvent,
         delayOpts: SendDelayedEventRequestOpts,
-    ): Promise<SendDelayedEventResponse>;
+        queryDict?: QueryDict,
+    ): Promise<ISendEventResponse>;
     protected async encryptAndSendEvent(
         room: Room | null,
         event: MatrixEvent,
-        delayOpts?: SendDelayedEventRequestOpts,
+        delayOptsOrQuery?: SendDelayedEventRequestOpts | QueryDict,
+        queryDict?: QueryDict,
     ): Promise<ISendEventResponse | SendDelayedEventResponse> {
-        if (delayOpts) {
-            return this.sendEventHttpRequest(event, delayOpts);
+        let queryOpts = queryDict;
+        if (delayOptsOrQuery && isSendDelayedEventRequestOpts(delayOptsOrQuery)) {
+            return this.sendEventHttpRequest(event, delayOptsOrQuery, queryOpts);
+        } else if (!queryOpts) {
+            queryOpts = delayOptsOrQuery;
         }
-
         try {
             let cancelled: boolean;
             this.eventsBeingEncrypted.add(event.getId()!);
@@ -2872,7 +2883,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             }
 
             if (!promise) {
-                promise = this.sendEventHttpRequest(event);
+                promise = this.sendEventHttpRequest(event, queryOpts);
                 if (room) {
                     promise = promise.then((res) => {
                         room.updatePendingEvent(event, EventStatus.SENT, res["event_id"]);
@@ -2987,14 +2998,16 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         }
     }
 
-    private sendEventHttpRequest(event: MatrixEvent): Promise<ISendEventResponse>;
+    private sendEventHttpRequest(event: MatrixEvent, queryDict?: QueryDict): Promise<ISendEventResponse>;
     private sendEventHttpRequest(
         event: MatrixEvent,
         delayOpts: SendDelayedEventRequestOpts,
+        queryDict?: QueryDict,
     ): Promise<SendDelayedEventResponse>;
     private sendEventHttpRequest(
         event: MatrixEvent,
-        delayOpts?: SendDelayedEventRequestOpts,
+        queryOrDelayOpts?: SendDelayedEventRequestOpts | QueryDict,
+        queryDict?: QueryDict,
     ): Promise<ISendEventResponse | SendDelayedEventResponse> {
         let txnId = event.getTxnId();
         if (!txnId) {
@@ -3027,19 +3040,22 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             path = utils.encodeUri("/rooms/$roomId/send/$eventType/$txnId", pathParams);
         }
 
+        const delayOpts =
+            queryOrDelayOpts && isSendDelayedEventRequestOpts(queryOrDelayOpts) ? queryOrDelayOpts : undefined;
+        const queryOpts = !delayOpts ? queryOrDelayOpts : queryDict;
         const content = event.getWireContent();
-        if (!delayOpts) {
-            return this.http.authedRequest<ISendEventResponse>(Method.Put, path, undefined, content).then((res) => {
-                this.logger.debug(`Event sent to ${event.getRoomId()} with event id ${res.event_id}`);
-                return res;
-            });
-        } else {
+        if (delayOpts) {
             return this.http.authedRequest<SendDelayedEventResponse>(
                 Method.Put,
                 path,
-                getUnstableDelayQueryOpts(delayOpts),
+                { ...getUnstableDelayQueryOpts(delayOpts), ...queryOpts },
                 content,
             );
+        } else {
+            return this.http.authedRequest<ISendEventResponse>(Method.Put, path, queryOpts, content).then((res) => {
+                this.logger.debug(`Event sent to ${event.getRoomId()} with event id ${res.event_id}`);
+                return res;
+            });
         }
     }
 
@@ -3096,16 +3112,16 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             content[withRelTypesPropName] = opts.with_rel_types;
         }
 
-        return this.sendCompleteEvent(
+        return this.sendCompleteEvent({
             roomId,
             threadId,
-            {
+            eventObject: {
                 type: EventType.RoomRedaction,
                 content,
                 redacts: eventId,
             },
-            txnId as string,
-        );
+            txnId: txnId as string,
+        });
     }
 
     /**
@@ -3393,7 +3409,47 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         }
 
         this.addThreadRelationIfNeeded(content, threadId, roomId);
-        return this.sendCompleteEvent(roomId, threadId, { type: eventType, content }, delayOpts, txnId);
+        return this.sendCompleteEvent({
+            roomId,
+            threadId,
+            eventObject: { type: eventType, content },
+            delayOpts,
+            txnId,
+        });
+    }
+
+    /**
+     * Send a delayed timeline event.
+     *
+     * Note: This endpoint is unstable, and can throw an `Error`.
+     *   Check progress on [MSC4140](https://github.com/matrix-org/matrix-spec-proposals/pull/4140) for more details.
+     */
+    // eslint-disable-next-line
+    public async _unstable_sendStickyDelayedEvent<K extends keyof TimelineEvents>(
+        roomId: string,
+        stickDuration: number,
+        delayOpts: SendDelayedEventRequestOpts,
+        threadId: string | null,
+        eventType: K,
+        content: TimelineEvents[K] & { msc4354_sticky_key: string },
+        txnId?: string,
+    ): Promise<SendDelayedEventResponse> {
+        if (!(await this.doesServerSupportUnstableFeature(UNSTABLE_MSC4354_STICKY_EVENTS))) {
+            throw new UnsupportedStickyEventsEndpointError(
+                "Server does not support the sticky events",
+                "sendStickyEvent",
+            );
+        }
+
+        this.addThreadRelationIfNeeded(content, threadId, roomId);
+        return this.sendCompleteEvent({
+            roomId,
+            threadId,
+            eventObject: { type: eventType, content },
+            queryDict: { msc4354_stick_duration_ms: stickDuration },
+            delayOpts,
+            txnId,
+        });
     }
 
     /**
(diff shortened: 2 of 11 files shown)


src/models/event.ts
```diff
@@ -1739,6 +1757,22 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
     public setThreadId(threadId?: string): void {
         this.threadId = threadId;
     }
+
+    /**
+     * Unstable getter to try and get the sticky information for the event.
+     * If the event is not a sticky event (or not supported by the server),
+     * then this returns `undefined`.
+     */
+    public get unstableStickyContent(): { duration_ms: number; duration_ttl_ms?: number } | undefined {
```
toger5: I find the `content` naming confusing. Sinc both params are either from unsigned or top level.
`Data`?

src/models/room-sticky-events.ts
```diff
@@ -0,0 +1,234 @@
...
+     * `const addedNew = added.filter(e => !removed.includes(e));`
+     * for a list of all removed events use:
+     * `const removedOnly = removed.filter(e => !added.includes(e));`
+     * @param added - The events that were added to the map of sticky events (can be updated events for existing keys or new keys)
+     * @param removed - The events that were removed from the map of sticky events (caused by expiration or updated keys)
+     */
+    [RoomStickyEventsEvent.Update]: (added: MatrixEvent[], removed: MatrixEvent[]) => void;
+};
+
+export class RoomStickyEvents extends TypedEventEmitter<RoomStickyEventsEvent, RoomStickyEventsMap> {
```
toger5:
```suggestion
export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEvent, RoomStickyEventsMap> {
```

?

src/models/room-sticky-events.ts
```diff
@@ -0,0 +1,234 @@
...
+     *
+     * NOTE: This will not cause `RoomEvent.StickyEvents` to be emitted.
+     *
+     * @throws If the `event` does not contain valid sticky data.
+     * @param event The MatrixEvent that contains sticky data.
+     * @returns An object describing whether the event was added to the map,
+     *          and the previous event it may have replaced.
+     */
+    // eslint-disable-next-line
+    public _unstable_addStickyEvent(event: MatrixEvent): { added: true; prevEvent?: MatrixEvent } | { added: false } {
```
toger5: This signature looks unfamiliar. Can we use a trow pattern + optional MatrixEvent?

toger5: We might event want to make this private. The public api just being the `addEvents` seems to be sufficient.

src/models/room-sticky-events.ts
```diff
@@ -0,0 +1,234 @@
...
+            ]);
+        } else {
+            this.unkeyedStickyEvents.add(event);
+        }
+
+        // Recalculate the next expiry time.
+        this.nextStickyEventExpiryTs = Math.min(expiresAtTs, this.nextStickyEventExpiryTs);
+
+        // Schedule this in the background
+        setTimeout(() => this.scheduleStickyTimer(), 1);
```
toger5: why not just calling it directly?

toger5: I would expect it to emit immediatly when i add an event not 1ms later.

src/models/room-sticky-events.ts
```diff
@@ -0,0 +1,234 @@
...
+    private stickyEventTimer?: NodeJS.Timeout;
+    private nextStickyEventExpiryTs: number = Number.MAX_SAFE_INTEGER;
+    private unkeyedStickyEvents = new Set<MatrixEvent>();
+
+    public constructor() {
+        super();
+    }
+
+    // eslint-disable-next-line
+    public *_unstable_getStickyEvents(): Iterable<MatrixEvent> {
```
toger5: There should also be a key value based get method on this class i think?

src/models/room-sticky-events.ts
toger5, line 1: This file uses unstable prefixes. I am not sure this is needed. can we make the class be stored in an unstable prefix in the room? und than make this class just have regular method names?

src/models/room.ts
```diff
@@ -446,6 +461,11 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
      */
     private roomReceipts = new RoomReceipts(this);
 
+    /**
+     * Stores and tracks sticky events
+     */
+    private stickyEvents = new RoomStickyEvents();
```
toger5:
```suggestion
    private _unstable_stickyEvents = new RoomStickyEvents();
```

src/models/room.ts
```diff
@@ -3414,6 +3438,24 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         return this.accountData.get(type);
     }
 
+    /**
+     * Get an iterator of currently active sticky events.
+     */
+    // eslint-disable-next-line
+    public _unstable_getStickyEvents(): ReturnType<RoomStickyEvents["_unstable_getStickyEvents"]> {
+        return this.stickyEvents._unstable_getStickyEvents();
```
toger5:
```suggestion
        return this._unstable_stickyEvents.getStickyEvents();
```

toger5: Example when the class would use stabl method names. its still very clear that we ure interacting with unstable api.

src/models/room.ts
toger5, line 1: also expose a getStickyEventForKey method here?

toger5: maybe just for user is enough?

src/client.ts
```diff
@@ -3430,6 +3486,38 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
     }
 
+    /**
+     * Send a delayed timeline event.
+     *
+     * Note: This endpoint is unstable, and can throw an `Error`.
+     *   Check progress on [MSC4140](https://github.com/matrix-org/matrix-spec-proposals/pull/4140) for more details.
+     */
+    // eslint-disable-next-line
+    public async _unstable_sendStickyEvent<K extends keyof TimelineEvents>(
```
toger5, line 3512: comment looks wrong

src/client.ts
```diff
@@ -3393,7 +3409,47 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+    }
+
+    /**
+     * Send a delayed timeline event.
+     *
+     * Note: This endpoint is unstable, and can throw an `Error`.
+     *   Check progress on [MSC4140](https://github.com/matrix-org/matrix-spec-proposals/pull/4140) for more details.
+     */
+    // eslint-disable-next-line
+    public async _unstable_sendStickyDelayedEvent<K extends keyof TimelineEvents>(
```
toger5:
```suggestion
     * Send a delayed sticky timeline event.
     *
     * Note: This endpoint is unstable, and can throw an `Error`.
     *   Check progress on [MSC4140](https://github.com/matrix-org/matrix-spec-proposals/pull/4140) for more details.
     */
    // eslint-disable-next-line
    public async _unstable_sendStickyDelayedEvent<K extends keyof TimelineEvents>(
```

src/client.ts
toger5, line 1: This file is mostly unchanged right? Can you hint me to any new chagnes I might have missed to look at in more detail (if there are any)

src/client.ts
```diff
@@ -3393,7 +3409,47 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+                "sendStickyEvent",
+            );
+        }
+
+        this.addThreadRelationIfNeeded(content, threadId, roomId);
+        return this.sendCompleteEvent({
+            roomId,
+            threadId,
+            eventObject: { type: eventType, content },
+            queryDict: { msc4354_stick_duration_ms: stickDuration },
```
toger5: can we doulbe check that this is the final query param name we want (it is different to the delayed events one `org.matrix.msc4140.delay`)
Probably needs to be discussed in the MSC.

src/client.ts
toger5, line 1: Do you like the  `getUnstableDelayQueryOpts(delayOpts),` pattern? I am wondering if we should remove it from delayed events or add it to sticky events. `getUnstableStickyQueryOpts(stickyOpts),` (which would then add the `org.matrix.mscXXXX.` prefic (or the mscXXX_) prefix dependent on what the MSC converges to.

src/sync-accumulator.ts
```diff
@@ -540,6 +555,15 @@ export class SyncAccumulator {
...
         });
 
+        // We want this to be fast, so don't worry about clobbering events here.
+        if (data.msc4354_sticky?.events) {
+            currentData._stickyEvents = currentData._stickyEvents.concat(data.msc4354_sticky?.events);
+        }
+        // But always prune any stale events, as we don't need to keep those in storage.
+        currentData._stickyEvents = currentData._stickyEvents.filter((ev) => {
+            return Date.now() < ev.msc4354_sticky.duration_ms + ev.origin_server_ts;
+        });
```
toger5: woulnt that be a straigh up server impl mistake? do we need to run a filter (not for free) just for the case a server is broken. It will get cleard anyways once added to the sticky room map?

src/sync.ts
```diff
@@ -1082,6 +1082,8 @@ export class SyncApi {
         //              highlight_count: 0,
         //              notification_count: 0,
         //          }
+        //          "org.matrix.msc4222.state_after": { events: [] },  // only if "org.matrix.msc4222.use_state_after" is true
+        //          msc4354_sticky: { events: [] }, // only if "org.matrix.msc4354.sticky" is true
```
toger5, line 1086: here the awkwards msc prefix also becomes very apperant. I think this should be discussion in the MSC.

src/models/room-sticky-events.ts
Half-Shot: Yeah, I think your right here.

src/models/room-sticky-events.ts
Half-Shot: Yeah, valid. I was optimizing for this function to not block the sync accumulator, but on the whole I think `cleanExpiredStickyEvents` actually getting called immediately is unlikely.

src/models/room-sticky-events.ts
Half-Shot: So the intention here is that for the tests I wanted to check the rejection reasons in https://github.com/matrix-org/matrix-js-sdk/blob/dc69e8bccff30b856e4497aad165e338657a7449/spec/unit/models/room-sticky-events.spec.ts#L33-L42, but for real world use we don't care because we just reject invalid events in the background.

The alternative I suppose is we just test that events don't end up in the store, maybe that's cleaner.

src/models/event.ts
Half-Shot: Decided on info purely as from a language perspective this is informational but I wouldn't consider it data.

src/models/room.ts
Half-Shot: I think it's probably fine to keep the internal classes as "stable" naming, but the functions called by the SDK consumers should be clearly marked as unstable. Hence, I think for `room` we should keep the `public _unstable_getStickyEvents`, but the class instance and methods can be as they are.

src/client.ts
Half-Shot: Virtually none, most of the changes are within sync.

src/client.ts
Half-Shot: Ah yes, this is `org.matrix.msc4354.sticky_duration_ms` in the MSC. updating.

src/sync-accumulator.ts
Half-Shot: Perhaps I wasn't clear what this was actually doing. This is clearing out sticky events that are in our sync accumulator storage that may have accumulated from past syncs, rather than the current sync. It's easier to have this logic live here.

toger5: I see. TIL. I expected this method to run each sync. Instead it will run periodically but might do so after 1-N consecutive syncs?

src/models/room.ts
toger5: this has happened right?

src/client.ts
toger5: this one still needs to change to sticky
```suggestion
    /**
     * Send a sticky timeline event.
     *
     * Note: This endpoint is unstable, and can throw an `Error`.
     *   Check progress on [MSC4354](https://github.com/matrix-org/matrix-spec-proposals/pull/4354) for more details.
     */
    // eslint-disable-next-line
    public async _unstable_sendStickyEvent<K extends keyof TimelineEvents>(
```

src/sync.ts
toger5: This one did not change in the msc... so we just keep it i suppose...

src/sync-accumulator.ts
Half-Shot: This method runs after each sync, and works out how much of that sync should be committed to storage. This section concats in sticky events that we should keep, and discards ones that we no longer need. It's deliberately very rough because it's just used to persist some data between *but* clients actually check the `RoomStickyEventStore` for the source of truth.

src/models/room-sticky-events.ts
toger5: I cannot follow. The linked code seems to test for throw? Is this because it got updated an now the linked code also has the updated test example?

src/sync-accumulator.ts
toger5: If its per sync, than I do not understand how we could ever end up with an event matching this condition: Date.now() < ev.msc4354_sticky.duration_ms + ev.origin_server_ts in a sync?

Ah is it in case there are so few messages, that it still is in the timeline section even though it is not stikcy anymore?

toger5: But even then. This kind of events would be dropped (not added to) the `RoomStickyEventStore`

Half-Shot: Right, but it *still* accumulates in your sync store. The room sticky events store is a memory store that computes what events are sticky for your current session

The sync accumulator just calculates the accumulation of sync from first login to present, discarding any info it doesn't need along the way. You are right that the sticky store would ignore these events but they would still end up being stored on disk forever without this filter statement.

The filter statement is there to prevent endless storage growth from accumulating sticky events.

src/client.ts
Half-Shot: I kinda...don't like that we calculate the prefix to be honest. They should be opaque strings that should be changed when the implementation is stable. It's probably due a refactor if anything.

src/models/room-sticky-events.ts
Half-Shot: Yeah, I decided to stop testing for throw and just test the result of the store. It's slightly less clear but it means one less function exposed and the tests check the same thing.

src/sync-accumulator.ts
toger5: oh. Now i see. So I was misunderstanding what `currentData._stickyEvents` is.

src/sync-accumulator.ts
```diff
@@ -558,7 +558,7 @@ export class SyncAccumulator {
         // Prune out any events in our stores that have since expired, do this before we
         // insert new events.
         currentData._stickyEvents = currentData._stickyEvents.filter((ev) => {
-            return Date.now() < ev.msc4354_sticky.duration_ms + ev.origin_server_ts;
```
toger5: I think the prev logic was correct?
 - keep events where Date.now() < ev.msc4354_sticky.duration_ms + ev.origin_server_ts is true
 - keep events where Date.now() < ts the event is expired expired is true
 - keep events which expire at a time that is greate than now
 - keep events that expire int he future
 - keeep unexpired events
 ?

src/client.ts
toger5: Lets resolve this than and be okay with it being different than the delayed events.

src/models/room.ts
Half-Shot: `getStickyEvent` exists if you know the 3 parts of the tuple (and you know the room obviously)

src/client.ts
Half-Shot: Looks fixed to me?

src/sync-accumulator.ts
Half-Shot: Yeah, that looks like a mistake. Fixed.

toger5 (approved):

src/sync.ts
```diff
@@ -1402,6 +1405,14 @@ export class SyncApi {
             // we deliberately don't add accountData to the timeline
             room.addAccountData(accountDataEvents);
 
+            // events from the sticky section of the sync come first (those are the ones that would be skipped due to gappy syncs)
+            // hence we consider them as older.
+            // and we add the events from the timeline at the end (newer)
```
robintown: This makes more sense to me the other way around:

```suggestion
            // events from the sticky section of the sync are older than all other timeline events
            // (if they weren't sticky, they would be skipped due to gappy syncs)
            // hence we need to add them first.
            // then we add sticky events from the timeline section (newer)
```

src/sync.ts
```diff
@@ -1411,11 +1422,19 @@ export class SyncApi {
...
             timelineEvents.forEach(emitEvent);
             ephemeralEvents.forEach(emitEvent);
             accountDataEvents.forEach(emitEvent);
-
+            stickyEvents
+                .filter(
+                    (e) =>
+                        // Ensure we do not emit twice.
+                        !timelineEvents.some((te) => te.getId() === e.getId()),
+                )
```
robintown, line 1442: Why is this filter necessary? I thought `stickyEvents` represented just the events from the `sticky` section of the sync response.

src/sync-accumulator.ts
```diff
@@ -76,11 +76,25 @@ export interface ITimeline {
     prev_batch: string | null;
 }
 
+type StickyEventFields = {
+    msc4354_sticky: { duration_ms: number };
+    content: IRoomEvent["content"] & { msc4354_sticky_key?: string };
```
robintown:
```suggestion
    content: { msc4354_sticky_key?: string };
```

TypeScript might let you get away with this.

src/sync-accumulator.ts
```diff
@@ -540,6 +556,18 @@ export class SyncAccumulator {
             });
         });
 
+        // Prune out any events in our stores that have since expired, do this before we
+        // insert new events.
+        currentData._stickyEvents = currentData._stickyEvents.filter((ev) => {
+            return Date.now() < ev.msc4354_sticky.duration_ms + ev.origin_server_ts;
+        });
```
robintown: Do we need to be mindful of this line from the MSC when performing this check?

> The start time is `min(now, origin_server_ts)`. This ensures that malicious origin timestamps cannot specify start times in the future.

i.e. should the sync accumulator record `min(now, origin_server_ts)` for each sticky event at the time of adding it and use that here in place of `ev.origin_server_ts`?

src/models/room.ts
```diff
@@ -311,6 +313,19 @@ export type RoomEventHandlerMap = {
...
+     * For a list of all updated events use:
+     * `const updated = added.filter(e => removed.includes(e));`
+     * for a list of all new events use:
+     * `const addedNew = added.filter(e => !removed.includes(e));`
+     * for a list of all removed events use:
+     * `const removedOnly = removed.filter(e => !added.includes(e));`
+     * @param added - The events that were added to the map of sticky events (can be updated events for existing keys or new keys)
+     * @param removed - The events that were removed from the map of sticky events (caused by expiration or updated keys)
+     * @param room - The room containing the sticky events
+     */
```
robintown: I find this confusing, if the same event can belong to both `added` and `removed` at the same time then the names of those arrays seem a bit off. Presumably this means that when an event is found in the `removed` array, it might not actually represent the event that has been removed, rather it's just an event which happens to share the same sticky key?

src/models/room.ts
```diff
@@ -3414,6 +3438,54 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+    public _unstable_getKeyedStickyEvent(
+        sender: string,
+        type: string,
+        stickyKey: string,
+    ): ReturnType<RoomStickyEventsStore["getKeyedStickyEvent"]> {
+        return this.stickyEvents.getKeyedStickyEvent(sender, type, stickyKey);
+    }
+
+    /**
+     * Get an active sticky events that match the given `type` and `sender`.
```
robintown:
```suggestion
     * Get an active sticky event that matches the given `type` and `sender`.
```

src/models/room.ts
```diff
@@ -3414,6 +3438,54 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+    /**
+     * Get an iterator of currently active sticky events.
+     */
+    // eslint-disable-next-line
+    public _unstable_getStickyEvents(): ReturnType<RoomStickyEventsStore["getStickyEvents"]> {
+        return this.stickyEvents.getStickyEvents();
+    }
+
+    /**
+     * Get a sticky event that match the given `type`, `sender`, and `stickyKey`
```
robintown, line 3435:
```suggestion
     * Get a sticky event that matches the given `type`, `sender`, and `stickyKey`
```

src/@types/requests.ts
```diff
@@ -107,6 +107,9 @@ export type SendActionDelayedEventRequestOpts = ParentDelayId;
 
 export type SendDelayedEventRequestOpts = SendTimeoutDelayedEventRequestOpts | SendActionDelayedEventRequestOpts;
 
+export function isSendDelayedEventRequestOpts(opts: object): opts is SendDelayedEventRequestOpts {
+    return (opts as TimeoutDelay).delay !== undefined || (opts as ParentDelayId).parent_delay_id !== undefined;
+}
```
robintown, line 110: Nitpick, but this should probably verify that the types of `event_id`, `delay`, `parent_delay_id` match in case anyone were to use `isSendDelayedEventRequestOpts` on truly untrusted input.

robintown (commented):
I haven't yet reviewed the file for the RoomStickyEventsStore but hopefully this is a good start:

src/sync.ts
Half-Shot: Ah unfortunately not, and to be honest this might be a point of confusion in the MSC. Sticky events will end up in the `timeline` by default, and the presence of the sticky metadata makes them sticky. 

If the sync is gappy, then the sticky event instead comes down the sticky section but should be treated the same way.

This logic covers the case where they come down both, although I'm not too sure if that's *actually* likely with the implementations. @toger5 was this something we were theorising might be a problem, or actually a problem?

(Regardless, I'll document this more)

src/sync.ts
Half-Shot: I've rewritten this as:
 - I think a primer on how sticky events work was warranted at this point.
 - The ordering of adding sticky events isn't actually important here, other than perhaps for performance reasons. The RoomStickyEventStore can handle events in any order and will calculate the same response (there is an English word for this I am forgetting).

src/sync-accumulator.ts
Half-Shot: so it did!

src/sync-accumulator.ts
Half-Shot: Sure. I think really the only danger here is your store would be bloated by sticky events from the future, but since we already calculate `now` this is a free check.

Half-Shot: Of course, in doing so we do cause problems if the client or sever clock is skewed. If our time is off, we may discard events too early. This would only be a problem for persisted events across sessions, but might cause edge case problems.

src/models/room.ts
Half-Shot: Yeah, this is confusing. I wonder if we are just better off with 3 arrays for clarity.

An event can be added if it's a brand new 4-ple (roomId, type, sender, sticky_key)
An event is updated (old event removed, new event added) if the 4-ple matches.
An event is removed if it has expired, and really that's the only case of that.

So I think probably let's just be more clear with this emitter to make it easier on consumers.

Half-Shot: I'm also not sure the examples are helpful, but that's just me.

src/@types/requests.ts
Half-Shot: Agreed, if statement updated.

src/sync.ts
toger5: I think we were discussing the general issue that this section
```
 stateEvents.forEach(emitEvent);
            timelineEvents.forEach(emitEvent);
            ephemeralEvents.forEach(emitEvent);
            accountDataEvents.forEach(emitEvent);
            stickyEvents
                .filter(
                    (stickyEvent) => !timelineEvents.some((timelineEvent) => timelineEvent.getId() === stickyEvent.getId()))
                .forEach(emitEvent);
```

will emit the same things eventually. a sticky stat event might be emitted due to the state section and the timeline section or sticky section.

I think that a server puts the evnet in both does not need to be covered since it would be a straigh up server error. (it still might be nice to filter it)

I think even more important would be if the client logs that the server does not behave to spec.

src/models/event.ts
```diff
@@ -408,6 +411,14 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
 
     private readonly reEmitter: TypedReEmitter<MatrixEventEmittedEvents, MatrixEventHandlerMap>;
 
+    /**
+     * The timestamp for when this event should expire, in milliseconds.
+     * Prefers using the serve-provided value, but will fall back to local calculation.
```
robintown:
```suggestion
     * Prefers using the server-provided value, but will fall back to local calculation.
```

Also: I think it's worth documenting whether this value is considered "safe" in the sense that,

> The start time is `min(now, origin_server_ts)`. This ensures that malicious origin timestamps cannot specify start times in the future.

src/models/event.ts
```diff
@@ -1739,6 +1757,22 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
     public setThreadId(threadId?: string): void {
         this.threadId = threadId;
     }
+
+    /**
+     * Unstable getter to try and get the sticky information for the event.
+     * If the event is not a sticky event (or not supported by the server),
+     * then this returns `undefined`.
+     */
```
robintown, line 1772: This could also document that the values are safely bounded

src/models/event.ts
```diff
@@ -449,6 +460,13 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
         const age = this.getAge();
         this.localTimestamp = age !== undefined ? Date.now() - age : (this.getTs() ?? Date.now());
         this.reEmitter = new TypedReEmitter(this);
+        if (this.unstableStickyInfo) {
+            if (this.unstableStickyInfo.duration_ttl_ms) {
+                this.unstableStickyExpiresAt = Date.now() + this.unstableStickyInfo.duration_ttl_ms;
+            } else {
+                this.unstableStickyExpiresAt = this.getTs() + this.unstableStickyInfo.duration_ms;
```
robintown:
```suggestion
                this.unstableStickyExpiresAt = Math.min(this.getTs(), Date.now()) + this.unstableStickyInfo.duration_ms;
```

src/models/room-sticky-events.ts
```diff
@@ -0,0 +1,240 @@
...
+     * @param removed The events that were removed from the map due to expiry.
+     */
+    [RoomStickyEventsEvent.Update]: (
+        added: MatrixEvent[],
+        updated: { current: MatrixEvent; previous: MatrixEvent }[],
+        removed: MatrixEvent[],
+    ) => void;
+};
+
+type StickyMatrixEvent = MatrixEvent & { unstableStickyExpiresAt: number };
```
robintown: Can we export `StickyMatrixEvent` and use it for the type of events emitted in an `Update` event?

src/models/room-sticky-events.ts
```diff
@@ -0,0 +1,240 @@
...
+export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEvent, RoomStickyEventsMap> {
+    private readonly stickyEventsMap = new Map<string, Map<string, StickyMatrixEvent>>(); // (type -> stickyKey+userId) -> event
+    private readonly unkeyedStickyEvents = new Set<StickyMatrixEvent>();
+
+    private stickyEventTimer?: NodeJS.Timeout;
+    private nextStickyEventExpiryTs: number = Number.MAX_SAFE_INTEGER;
+
+    public constructor() {
+        super();
+    }
```
robintown:
```suggestion
```

src/models/room-sticky-events.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    /**
+     * Get all sticky events that are currently active.
+     * @returns An iterable set of events.
+     */
+    public *getStickyEvents(): Iterable<MatrixEvent> {
```
robintown: Could use `StickyMatrixEvent` on this and subsequent methods as well

src/models/room-sticky-events.ts
```diff
@@ -0,0 +1,240 @@
...
+     * @param sender The sender of the sticky event.
+     * @param stickyKey The sticky key used by the event. If not provided, this will search the unkeyed sticky events.
+     * @returns A matching active sticky event, or undefined.
+     */
+    public getKeyedStickyEvent(sender: string, type: string, stickyKey: string): MatrixEvent | undefined {
+        return this.stickyEventsMap.get(type)?.get(`${stickyKey}${sender}`);
+    }
+
+    /**
+     * Get an active sticky event that match the given `type` and `sender`.
```
robintown:
```suggestion
     * Get active sticky events without a sticky key that match the given `type` and `sender`.
```

src/models/room-sticky-events.ts
```diff
@@ -0,0 +1,240 @@
...
+        for (const innerMap of this.stickyEventsMap.values()) {
+            yield* innerMap.values();
+        }
+    }
+
+    /**
+     * Get an active sticky event that match the given `type`, `sender`, and `stickyKey`
+     * @param type The event `type`.
+     * @param sender The sender of the sticky event.
+     * @param stickyKey The sticky key used by the event. If not provided, this will search the unkeyed sticky events.
```
robintown: "If not provided, this will search the unkeyed sticky events" seems no longer accurate for this method

src/models/room-sticky-events.ts
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        // While we fully expect the server to always provide the correct value,
+        // this is just insurance to protect against attacks on our Map.
+        if (!sender.startsWith("@")) {
+            throw new Error("Expected sender to start with @");
+        }
+
+        let prevEvent: StickyMatrixEvent | undefined;
+        if (stickyKey) {
```
robintown: To handle a sticky key of `""` correctly:

```suggestion
        if (stickyKey !== undefined) {
```

robintown (commented):
Still working through this bit by bit as I had some thoughts I wanted to leave on the MSC as well. Here are my comments so far:

src/models/room-sticky-events.ts
```diff
@@ -0,0 +1,240 @@
...
+        updated: { current: MatrixEvent; previous: MatrixEvent }[],
+        removed: MatrixEvent[],
+    ) => void;
+};
+
+type StickyMatrixEvent = MatrixEvent & { unstableStickyExpiresAt: number };
+
+/**
+ * Tracks sticky events on behalf of one room, and fires an event
+ * whenever a sticky even is updated or replaced.
```
robintown:
```suggestion
 * whenever a sticky event is updated or replaced.
```

kathleenmotley49-maker (approved):

src/models/room-sticky-events.ts
```diff
@@ -0,0 +1,240 @@
...
+
+        // We will recalculate this as we check all events.
+        this.nextStickyEventExpiryTs = Number.MAX_SAFE_INTEGER;
+        for (const [eventType, innerEvents] of this.stickyEventsMap.entries()) {
+            for (const [innerMapKey, event] of innerEvents) {
+                // we only added items with `sticky` into this map so we can assert non-null here
+                if (now >= event.unstableStickyExpiresAt) {
+                    logger.debug("Expiring sticky event", event.getId());
+                    removedEvents.push(event);
+                    this.stickyEventsMap.get(eventType)!.delete(innerMapKey);
```
robintown, line 201: If `innerMapKey` was the last remaining entry in the inner map, it's probably good hygiene to delete the inner map itself so that we don't start leaking empty Maps.

In fact, this pattern of shuffling around nested maps is so common in the JS SDK that I feel like it deserves to be turned into a set of reusable functions. It takes a bit of type magic, but I played around with it a bit and I think something like this would make for a pretty convenient API:

```typescript
type NestedMap<K extends unknown[], V> = K extends [infer K1, ...infer K_] ? Map<K1, NestedMap<K_, V>> : V

function getNested<K extends [unknown, ...unknown[]], M extends NestedMap<K, unknown>>(map: M, ...keys: K): M extends NestedMap<K, infer V> ? V | undefined : never {
    let result: any = map
    for (const key of keys) result = result?.get(key)
    return result
}

function setNested<K extends [unknown, ...unknown[]], V, M extends NestedMap<K, V>>(map: M, value: V, ...keys: K): void {
    let m: any = map
    for (let i = 0; i < keys.length - 1; i++) {
        let inner = m.get(keys[i])
        if (inner === undefined) {
            inner = new Map()
            m.set(keys[i], inner)
        }
        m = inner
    }
    m.set(keys[keys.length - 1], value)
}

function deleteNested<K extends [unknown, ...unknown[]], M extends NestedMap<K, unknown>>(map: M, ...keys: K): boolean {
    if (keys.length === 1) {
        return map.delete(keys[0])
    } else {
        const inner = map.get(keys[0]) as Map<unknown, unknown> | undefined
        if (inner === undefined) return false
        const [key, ...keys_] = keys
        // @ts-ignore
        const result = deleteNested(inner, ...keys_)
        if (inner.size === 0) map.delete(key)
        return result
    }
}
```

Usage looks like this:

```typescript
const myMap = new Map<number, Map<string, number>>()
setNested(myMap, 999, 1, 'foo')
const y = getNested(myMap, 1, 'foo') // y = 999
deleteNested(myMap, 1, 'foo') // now myMap is empty
```

Then with nested maps of 3+ layers being less painful, this could even help us avoid the `${stickyKey}${sender}` hack found in this file.

What do you think?

src/models/room.ts
```diff
@@ -493,6 +512,10 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
         // receipts. No need to remove the listener: it's on ourself anyway.
         this.on(RoomEvent.Receipt, this.onReceipt);
 
+        this.stickyEvents.on(RoomStickyEventsEvent.Update, (...props) =>
+            this.emit(RoomEvent.StickyEvents, ...props, this),
+        );
```
robintown: It looks like the more common pattern is to use the `reEmitter`. Is there a particular reason to avoid it here?

src/models/room.ts
```diff
@@ -3414,6 +3437,54 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+    public _unstable_getKeyedStickyEvent(
+        sender: string,
+        type: string,
+        stickyKey: string,
+    ): ReturnType<RoomStickyEventsStore["getKeyedStickyEvent"]> {
+        return this.stickyEvents.getKeyedStickyEvent(sender, type, stickyKey);
+    }
+
+    /**
+     * Get an active sticky event that match the given `type` and `sender`.
```
robintown:
```suggestion
     * Get active sticky events without a sticky key that match the given `type` and `sender`.
```

src/models/room.ts
```diff
@@ -3414,6 +3437,54 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+
+    /**
+     * Add a series of sticky events, emitting `RoomEvent.StickyEvents` if any
+     * changes were made.
+     * @param events A set of new sticky events.
+     */
+    // eslint-disable-next-line
+    public _unstable_addStickyEvents(events: MatrixEvent[]): ReturnType<RoomStickyEventsStore["addStickyEvents"]> {
+        return this.stickyEvents.addStickyEvents(events);
+    }
```
robintown, line 3473: Maybe it makes sense to mark this as internal.

src/sync.ts
robintown: +1, it seemed like a bit of a prematurely defensive check. Not totally against it though if it seems reassuring to have.

src/sync-accumulator.ts
robintown: True, and I guess that's sorta what `sticky_duration_ttl_ms` is for. Maybe it's a bit overkill to invoke it here.

If we want to be serious about avoiding unbounded sync accumulator bloat, we should probably cap `duration_ms` at 1 hour here as well. And, correct me if I'm wrong, but I think the `Math.min(now, ev.origin_server_ts)` thing is only effective if the `now` in that expression is the time at which the sticky event was originally *added* to the sync accumulator. The condition `now < ev.msc4354_sticky.duration_ms + now` is always true for durations > 0.

Half-Shot: I've rewritten this statement out more clearly so I think it makes sense:

```ts
// Prune out any events in our stores that have since expired, do this before we
// insert new events.
const now = Date.now();
currentData._stickyEvents = currentData._stickyEvents.filter((ev) => {
    // If `duration_ms` exceeds the spec limit of a hour, we cap it.
    const cappedDuration = Math.min(ev.msc4354_sticky.duration_ms, MAX_STICKY_DURATION_MS);
    // If `origin_server_ts` claims to have been from the future, we still bound it to now.
    const sanitisedOriginTs = Math.min(now, ev.origin_server_ts);
    const expiresAt = cappedDuration + sanitisedOriginTs;
    return expiresAt < now;
});
```

`cappedDuration` will always be limited to a hour now.
`sanitisedOriginTs` will either be the `origin_server_ts` or the current time, whichever is less. That means it will always be the `origin_server_ts` UNLESS our local clock is behind whoever sent it. I think this is fine, as it's just to avoid accumulator bloat.

And then we just check that `sanitisedOriginTs + cappedDuration > now` which feels right?

Half-Shot: The tests seem to agree with me at least :)

src/models/event.ts
Half-Shot: I wrote some words along the same lines, but didn't include the precise details here. I think anyone more curious can check the source.

src/models/room-sticky-events.ts
Half-Shot: Partially, largely it seems like most events on Room emit the room itself, so we do need to include that. I can use slightly clever typescript incantations to reduce the duplication though :)

src/models/room-sticky-events.ts
Half-Shot: Ah yes, I split this out to a seperate func and didn't update the docs.

src/models/room-sticky-events.ts
Half-Shot: Disregard this, I've decided to just use the reemitter.

src/models/room.ts
Half-Shot: We'd have to forgo `room`, but that looks like other events work. Have updated.

src/models/room-sticky-events.ts
toger5: If we already introduce the type NestedMap it feels like a full class might make sense?
(I still like the public setNested, deleteNested) as a way to simplify working with already existing nested maps. But `myNestedMap.set(999, 1, 'foo')` seems to be the more desired api?

Half-Shot: I like the idea, although I think I'd prefer to have an instance of NestedMap than functions that operate on a Map, just makes it a bit easier to use (and my editor appreciates the method hints!).

I think based on the number of iterations this PR has already gone through, we should just keep the map structure as-is but return to this as a separate PR. We could even use this as an excuse to go round and refactor the other nested maps in the codebase, and do it in one chunk.

toger5: I created https://github.com/matrix-org/matrix-js-sdk/issues/5033 to not loose the idea and justify resolving this thread.

src/sync-accumulator.ts
robintown: Mm, the `sanitisedOriginTs` still isn't doing what I'd expect it to. Here's a failing test to demonstrate:

```typescript
it("clears stale sticky events that pretend to be from the distant future", () => {
    jest.setSystemTime(0);
    const ev = stickyEvent(999999999999);
    sa.accumulate(syncSkeleton({ msc4354_sticky: { events: [ev] } }));
    expect(sa.getJSON().roomsData[Category.Join]["!foo:bar"].msc4354_sticky?.events).toEqual([ev]);
    jest.setSystemTime(1000); // Expire the event
    sa.accumulate(syncSkeleton({}));
    expect(sa.getJSON().roomsData[Category.Join]["!foo:bar"].msc4354_sticky?.events).toBeUndefined();
});
```

PR #5030, opened 2025-10-01
Fetch the user's device info before processing a verification request

andybalaam:
If we don't have the device info for a user when we receive their verification request, we ignore it. This change gives us the best possible chance of having the right device data before we try to process the verification.
    
* Fixes [#30693](https://github.com/element-hq/element-web/issues/30693)
* Fixes [#27819](https://github.com/element-hq/element-web/issues/27819)


richvdh (approved):
LGTM

PR #5031, opened 2025-10-06
[MatrixRTC] Use relation based call membership to compute `create_ts()`

toger5:
This PR changes the construction of the `CallMembership` object.
It will add an optional `relatedEvent` parameter which will store initial connect event for this call participation.

This is then used to fetch metadata (like `application`) and the `origin_server_ts` of the initial join event.

This requires that `MatrixRTCSession.sessionMembershipsForSlot` becomes async since it now also needs to fetch related events.

This change needs a bit of test adjustments and other places of the codebase need updating.

## Checklist


kathleenmotley49-maker (approved):

toger5 (commented):
What happens if two MatrixRTCSession.sessionMembershipsForSlot are running at the same time?
The second one could finish before the other.... Is this an issue?

robintown:
@toger5 Is the diff meant to be as big as it is? I think I'm looking at a lot of lint fixes, or something.

toger5:
> Is the diff meant to be as big as it is?
Yes the diff is meant to be this big. There is a very large amount of lines which add `async` and `await`

On top of this, the construct for callMembership has a signature change which requires a change in `makeMockEvent`.

before
`new CallMembership(event, event.content, ....)
now
`new CallMembership(event,...)` content will be aquired indirectly by reading event.getContent().

spec/unit/matrixrtc/CallMembership.spec.ts
```diff
@@ -119,11 +107,13 @@ describe("CallMembership", () => {
...
-                });
+                const membership = new CallMembership(
+                    makeMockEvent(0, {
+                        ...membershipTemplate,
+                        foci_preferred: [mockFocus],
+                        focus_active: { type: "livekit", focus_selection: "multi_sfu" },
+                    }),
+                );
 
                 // if we are the oldest member we use our focus.
```
BillCarsonFr: nit on the comment? we use our focus no matter what

spec/unit/matrixrtc/CallMembership.spec.ts
```diff
@@ -41,75 +38,66 @@ describe("CallMembership", () => {
...
         it("returns preferred foci", () => {
-            const fakeEvent = makeMockEvent();
             const mockFocus = { type: "this_is_a_mock_focus" };
-            const membership = new CallMembership(fakeEvent, { ...membershipTemplate, foci_preferred: [mockFocus] });
+            const fakeEvent = makeMockEvent(0, { ...membershipTemplate, foci_preferred: [mockFocus] });
+            const membership = new CallMembership(fakeEvent);
             expect(membership.transports).toEqual([mockFocus]);
         });
 
         describe("getTransport", () => {
```
BillCarsonFr, line 91: Looks like in all the tests `foci_preferred` is always an array of 1. Can the test clarify how this works?
I think it is a non multi-sfu thing to avoid switching focus, but for multi-sfu it is always one?
=> maybe a bug in the spec? the well-known also gives a list :/ should we pick one random?

spec/unit/matrixrtc/CallMembership.spec.ts
```diff
@@ -310,11 +264,13 @@ describe("CallMembership", () => {
...
+                const oldestMembership = new CallMembership(
+                    makeMockEvent(0, {
+                        ...membershipTemplate,
+                        rtc_transports: [{ type: "oldest_transport" }],
+                    }),
+                );
+                const membership = new CallMembership(makeMockEvent(0, membershipTemplate));
 
                 // if we are the oldest member we use our focus.
                 expect(membership.getTransport(membership)).toStrictEqual({ type: "livekit" });
```
BillCarsonFr, line 372: I am a bit lost, the transport is only { type: livekit }?  shouldn't there be a `livekit_service_url` and the alias?
Maybe we can keep "realistic" test data, and then use `expect` with `expect.objectContaining({ .. })` to just test the field you want?

spec/unit/matrixrtc/CallMembership.spec.ts
```diff
@@ -324,7 +280,7 @@ describe("CallMembership", () => {
             });
         });
         describe("correct values from computed fields", () => {
```
BillCarsonFr, line 379: bonus point to have a `test.each()` that would try different configuration without reapeating code
```
test.each([
  [1, 1, 2],
  [1, 2, 3],
  [2, 1, 3],
])('add(%i, %i) -> %i', (a, b, expected) => {
  expect(a + b).toBe(expected)
})
```

spec/unit/matrixrtc/MatrixRTCSession.spec.ts
```diff
@@ -38,6 +55,8 @@ describe("MatrixRTCSession", () => {
         client.getDeviceId = jest.fn().mockReturnValue("AAAAAAA");
         client.sendEvent = jest.fn().mockResolvedValue({ event_id: "success" });
         client.decryptEventIfNeeded = jest.fn();
+        client.fetchRoomEvent = jest.fn().mockResolvedValue(undefined);
```
BillCarsonFr, line 61: Not strictly this PR, but shouldn't we use `getMockClientWithEventEmitter({})` insteaf of overriding the properties like that? 
Currently this could be doing more thing than we think and polute the test?

spec/unit/matrixrtc/MatrixRTCSessionManager.spec.ts
BillCarsonFr: this is not needed any more? there is no matching `.on`?

spec/unit/matrixrtc/MembershipManager.spec.ts
BillCarsonFr: a new way to have a mock function on a mock client, we might have to many ways of doing that ^^

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -923,18 +927,18 @@ describe("MembershipManager", () => {
...
 
     const membershipManager = new MembershipManager(undefined, room, client, callSession);
 
-    const spy = jest.spyOn(console, "error");
     // Double join
     membershipManager.join([]);
     membershipManager.join([]);
 
-    expect(spy).toHaveBeenCalled();
+    await waitFor(() => expect(spy).toHaveBeenCalled());
```
BillCarsonFr, line 964: TIL

spec/unit/matrixrtc/mocks.ts
```diff
@@ -130,23 +144,22 @@ export function makeMockEvent(
...
-    return {
-        getType: jest.fn().mockReturnValue(type),
-        getContent: jest.fn().mockReturnValue(content),
-        getSender: jest.fn().mockReturnValue(sender),
-        getTs: jest.fn().mockReturnValue(timestamp ?? Date.now()),
-        getRoomId: jest.fn().mockReturnValue(roomId),
-        getId: jest.fn().mockReturnValue(secureRandomString(8)),
-        isDecryptionFailure: jest.fn().mockReturnValue(false),
-    } as unknown as MatrixEvent;
+    return new MatrixEvent({
```
BillCarsonFr, line 161: I think there is a `makeMockEvent` utility in `mock.ts`

src/matrixrtc/CallMembership.ts
```diff
@@ -43,54 +42,46 @@ export interface RtcMembershipData {
     "application": {
         type: string;
         // other application specific keys
-        [key: string]: unknown;
+        [key: string]: any;
     };
     "rtc_transports": Transport[];
     "versions": string[];
     "msc4354_sticky_key"?: string;
```
BillCarsonFr: I wonder if we could do something similar to `UnstableValue`, I am not super familiar but worth a look?
There are things like
```
export const M_TEXT = new UnstableValue("m.text", "org.matrix.msc1767.text");
...
export type TextEventContent = EitherAnd<{ [M_TEXT.name]: string }, { [M_TEXT.altName]: string }>;
```

src/matrixrtc/CallMembership.ts
```diff
@@ -43,54 +42,46 @@ export interface RtcMembershipData {
...
     "sticky_key"?: string;
+    /**
+     * The intent of the call from the perspective of this user. This may be an audio call, video call or
+     * something else.
+     */
+    "m.call.intent"?: RTCCallIntent;
 }
 
 const checkRtcMembershipData = (
```
BillCarsonFr, line 53: that's clean

src/matrixrtc/IMembershipManager.ts
```diff
@@ -79,7 +79,6 @@ export interface IMembershipManager
     /**
      * Start sending all necessary events to make this user participate in the RTC session.
      * @param fociPreferred the list of preferred foci to use in the joined RTC membership event.
-     * If multiSfuFocus is set, this is only needed if this client wants to publish to multiple transports simultaneously.
      * @param multiSfuFocus the active focus to use in the joined RTC membership event. Setting this implies the
```
BillCarsonFr, line 83: I am still confused by that. My understanding was that there is a list of FOCI that my homeserver recommends. But if I join an existing call, there is already an active FOCI people are on, and we want to keep this one on top of my list in case I become the oldest member later, so that we don't switch focus.

So foci_active made sense to me, it was the existing Foci people are already using that I should connect on but also add on top of my list to avoid future jumps.

Now this signature do not make sense for me :/
Maybe if we change it to `fociPreferred: Transport[], oldestMembership?: MembershipData, ..`

Also I don't get why `fociPreferred` is an array

src/matrixrtc/MembershipManager.ts
```diff
@@ -422,19 +418,19 @@ export class MembershipManager
             case MembershipActionType.RestartDelayedEvent: {
                 if (!this.state.delayId) {
                     // Delay id got reset. This action was used to check if the hs canceled the delayed event when the join state got sent.
-                    return createInsertActionUpdate(MembershipActionType.SendDelayedEvent);
+                    return Promise.resolve(createInsertActionUpdate(MembershipActionType.SendDelayedEvent));
```
BillCarsonFr, line 448: is this needed? just returning the value is doing the same thing no?

BillCarsonFr (commented):
The PR looks to be doind more than what the title suggests ^^ 
Nice to see all the tests

t3chguy (changes requested):
Please fix the merge conflicts before review

spec/unit/matrixrtc/CallMembership.spec.ts
toger5: I added a comment:
```
...membershipTemplate,
                        // The list of foci_preferred provided by the homeserver. (in the test example we just provide one)
                        // The oldest member logic will use the first item in this list.
                        // The multi-sfu logic will (theoretically) also use all the items in the list at once
                        // (currently the js-sdk sets it to only one item in multi-sfu mode).
                        foci_preferred: [mockFocus],
                        focus_active: { type: "livekit", focus_selection: "oldest_membership" },
```

I hope that clarifies things.

spec/unit/matrixrtc/CallMembership.spec.ts
toger5: Update the tests to contain more realistic data.

spec/unit/matrixrtc/CallMembership.spec.ts
toger5: you mean using different `membership` objects?
I am not sure what additional stability this would result in. The code just gets fields from different branches of the object but does not do things with the values. (except the expiration which we do in a dedicated test section)

spec/unit/matrixrtc/MatrixRTCSession.spec.ts
toger5: This is definilty worth a discussion. I remember we discussed that it can also be benefitital to not use too many helpers so the tests are more accessible if devs are not familiar with the helper methods.

spec/unit/matrixrtc/MatrixRTCSessionManager.spec.ts
toger5: I think there still is. I will check for matching off/on s.

spec/unit/matrixrtc/mocks.ts
toger5: this is just updating this method?
Or do you mean test-utils.ts

src/matrixrtc/CallMembership.ts
toger5: I am not really seeing the value in this. Is there some central namespace toggle which i could use to bump everything to use the new names?

toger5: I can use the unstable type for now lets see if it is helpful in some way.

src/matrixrtc/MembershipManager.ts
toger5: the method is not async. I had issues with tests if this became an async method... It also is nto really async but just a wrapper to construct a promise.

src/matrixrtc/IMembershipManager.ts
toger5: So foci_active made sense to me, it was the existing Foci people are already using that I should connect on but also add on top of my list to avoid future jumps.

`focus_active` was just the description what system a specific user was using to decide what transport to use.

toger5: It really is only used with m.call.member events (not with m.rtc.member).

So it is basically depreacted anyways.

> Also I don't get why fociPreferred is an array

This was so multiple foci could be defined as fallbacks. All this is entirely depreacted due to multi-sfu.
In the new spec there is onlly `rtc_transports` that is the list of currently in use publishing transports.

t3chguy (changes requested):
Could we please be a little more mindful to reviewers, there's little point in reviewing something with so many CI failures given it'll need re-reviewing after fixing.

spec/unit/matrixrtc/mocks.ts
BillCarsonFr: oh right, I was expecting to see a vi.mock()

PR #5037, opened 2025-10-09
Handle redactions for sticky events (MSC4354)

Half-Shot:
## Checklist

### DIFF
--- a/src/models/room.ts
+++ b/src/models/room.ts
@@ -2634,7 +2634,12 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
             // if we know about this event, redact its contents now.
             const redactedEvent = redactId ? this.findEventById(redactId) : undefined;
             if (redactId) {
-                this.stickyEvents.handleRedaction(redactedEvent || redactId);
+                try {
+                    this.stickyEvents.handleRedaction(redactedEvent || redactId);
+                } catch (ex) {
+                    // Non-critical failure, but we should warn.
+                    logger.error("Failed to handle redaction for sticky event", ex);
+                }
             }
             if (redactedEvent) {
                 this.applyEventAsRedaction(event, redactedEvent);


src/models/room-sticky-events.ts
```diff
@@ -10,6 +10,20 @@ export enum RoomStickyEventsEvent {
 
 type StickyMatrixEvent = MatrixEvent & { unstableStickyExpiresAt: number };
 
+/**
+ * Keeps a chain of sticky events to ensure that we can roll back in case of expiry.
```
toger5:
```suggestion
 * Keeps a chain of sticky events to ensure that we can roll back in case of redactions.
```

src/models/room-sticky-events.ts
```diff
@@ -10,6 +10,20 @@ export enum RoomStickyEventsEvent {
...
+ */
+type StickyMatrixEventSet = {
```
toger5: should we just cal it a chain?
```suggestion
type StickyMatrixEventChain = {
```
Set is already a very commonly used word for HashSet s (or however they are implemented in js)

toger5: Or even better
```suggestion
type StickyMatrixEventChainItem = {
```

Half-Shot: I've totally rewritten this.

src/models/room-sticky-events.ts
```diff
@@ -29,20 +29,36 @@ export type RoomStickyEventsMap = {
  * whenever a sticky event is updated or replaced.
  */
 export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEvent, RoomStickyEventsMap> {
-    private readonly stickyEventsMap = new Map<string, Map<string, StickyMatrixEvent>>(); // (type -> stickyKey+userId) -> event
+    private readonly stickyEventsMap = new Map<string, Map<string, StickyMatrixEvent[]>>(); // (type -> stickyKey+userId) -> events (SORTED BY EXPIRY)
```
florianduros: Can we make the comment a proper tsdoc?

src/models/room-sticky-events.ts
```diff
@@ -76,6 +92,7 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
      * @param event The MatrixEvent that contains sticky data.
      * @returns An object describing whether the event was added to the map,
      *          and the previous event it may have replaced.
+     *
```
florianduros: Seems uneeded

src/models/room-sticky-events.ts
```diff
@@ -29,20 +29,36 @@ export type RoomStickyEventsMap = {
...
+     */
+    private static sortStickyEvent(eventA: StickyMatrixEvent, eventB: StickyMatrixEvent): number {
+        // sticky events are not allowed to expire sooner than their predecessor.
+        if (eventB.unstableStickyExpiresAt > eventA.unstableStickyExpiresAt) {
+            return 1;
+        } else if (eventA && eventA.getTs() === eventB.getTs() && (eventB.getId() ?? "") > (eventA.getId() ?? "")) {
+            return 1;
+        }
+        return -1;
+    }
```
florianduros, line 74: If the two events have the same expiration date, it should return `0`?

src/models/room-sticky-events.ts
```diff
@@ -29,20 +29,36 @@ export type RoomStickyEventsMap = {
...
 
+    /**
+     * Sort two sticky events by order of expiry.
+     * @returns A positive value if event A will expire sooner, or a negative value if event B will expire sooner.
+     */
+    private static sortStickyEvent(eventA: StickyMatrixEvent, eventB: StickyMatrixEvent): number {
+        // sticky events are not allowed to expire sooner than their predecessor.
+        if (eventB.unstableStickyExpiresAt > eventA.unstableStickyExpiresAt) {
+            return 1;
+        } else if (eventA && eventA.getTs() === eventB.getTs() && (eventB.getId() ?? "") > (eventA.getId() ?? "")) {
```
florianduros: why do we compare the event id?

src/models/room-sticky-events.ts
```diff
@@ -29,20 +29,36 @@ export type RoomStickyEventsMap = {
...
     /**
      * Get all sticky events that are currently active.
      * @returns An iterable set of events.
      */
     public *getStickyEvents(): Iterable<StickyMatrixEvent> {
         yield* this.unkeyedStickyEvents;
         for (const innerMap of this.stickyEventsMap.values()) {
-            yield* innerMap.values();
+            for (const events of innerMap.values()) {
+                yield events[0];
```
florianduros, line 95: why `yield` the first element?

src/models/room-sticky-events.ts
florianduros: Seems uneeded

src/models/room-sticky-events.ts
```diff
@@ -102,42 +119,39 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
...
+        // E.g. Where a malicious event type might be "rtc.member.event@foo:bar" the key becomes:
+        // "rtc.member.event.@foo:bar@bar:baz"
+        const innerMapKey = `${stickyKey}${sender}`;
+        const currentEventSet = [stickyEvent, ...(this.stickyEventsMap.get(type)?.get(innerMapKey) ?? [])].sort(
+            RoomStickyEventsStore.sortStickyEvent,
+        );
+        if (!this.stickyEventsMap.has(type)) {
+            this.stickyEventsMap.set(type, new Map());
+        }
+        this.stickyEventsMap.get(type)!.set(innerMapKey, currentEventSet);
```
florianduros:
```suggestion
        this.stickyEventsMap.get(type)?.set(innerMapKey, currentEventSet);
```

src/models/room-sticky-events.ts
```diff
@@ -229,6 +253,61 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
...
+            }
+        }
+        for (const [eventType, innerMap] of this.stickyEventsMap) {
+            for (const [key, [currentEvent, ...previousEvents]] of innerMap) {
+                if (currentEvent.getId() !== redactEventId) {
+                    continue;
+                }
+                logger.debug(`Redaction for ${redactEventId} under sticky key ${key}`);
+                // We have found the event.
+                for (const event of previousEvents) {
```
florianduros: that's a lot of for 😅

src/models/room-sticky-events.ts
```diff
@@ -229,6 +253,61 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
...
+        // but this saves us having to iterate over all events here to calculate
+        // the next expiry time.
+        for (const event of this.unkeyedStickyEvents) {
+            if (event.getId() === redactEventId) {
+                this.unkeyedStickyEvents.delete(event);
+                this.emit(RoomStickyEventsEvent.Update, [], [], [event]);
+                return;
+            }
+        }
+        for (const [eventType, innerMap] of this.stickyEventsMap) {
```
florianduros: Why do we need to all of this? A comment is welcomed :)

src/models/room-sticky-events.ts
toger5: Also why is the `eventA && eventA.getTs() === eventB.getTs() &&` bit needed. It seems `eventA` is non optional and `unstableStickyExpiresAt` also is of type `number`?
Also I am wondering about `eventA.getTs`:
 - if they do not have the same ts, they will not get compared by id? but instead just return -1?

src/models/room-sticky-events.ts
toger5:
```suggestion
```

src/models/room-sticky-events.ts
toger5, line 74: This is super tricky. I think it might not converge?

a-expiry=2, a-ts=1
b-expiry=1, b-ts=2

first sort: `[b, a]`
then the first if `eventB.unstableStickyExpiresAt > eventA.unstableStickyExpiresAt` would be false
so it would be sorted by ts

`[a, b]`

And it would toggle each sort iteration...

toger5:
```suggestion
    private static sortStickyEvent(eventA: StickyMatrixEvent, eventB: StickyMatrixEvent) {
        const order = eventB.unstableStickyExpiresAt - eventA.unstableStickyExpiresAt
        if (order === 0) return (eventB.getId() ?? "") > (eventA.getId() ?? "") ? 1 : -1;
        return order;
    }
```

I think this is correct (minus sng mistakes which the tests should immediately uncover)

I will outline why in a msc comment.

toger5: https://github.com/matrix-org/matrix-spec-proposals/pull/4354/files#r2417363680

Half-Shot: Sorry what was I smoking here. This implements https://github.com/matrix-org/matrix-spec-proposals/blob/kegan/persist-edu/proposals/4354-sticky-events.md#implementing-an-ephemeral-map:

> pick the one with the highest origin_server_ts,
> tie break on the one with the highest lexicographical event ID (A < Z).

So really this should contain two if statements. If the eventIds match, then we should throw.

src/models/room-sticky-events.ts
Half-Shot: https://github.com/matrix-org/matrix-js-sdk/pull/5037#discussion_r2420248976

src/models/room-sticky-events.ts
Half-Shot: Have hopefully improved with comments. This array is a sorted set of events where the first one is the "current" event in the sticky map, and the others are just predecessor events we don't want to expose unless the "current" one is redacted.

src/models/room-sticky-events.ts
Half-Shot: I think we still need to specify creating the map here?

src/models/room-sticky-events.ts
Half-Shot: It's *super* annoying. I've decided to use this as an excuse to refactor here. If we do store the event in the live timeline, we know the event type / sticky key / sender and can quickly get the event from the map.

src/models/room-sticky-events.ts
Half-Shot: I refactored also into a different but better shape based on the current spec (ignoring expiry time).

src/models/room-sticky-events.ts
Half-Shot: *Hopefully* much more commented now.

spec/unit/models/room-sticky-events.spec.ts
```diff
@@ -259,4 +259,163 @@ describe("RoomStickyEvents", () => {
...
+            const emitSpy = jest.fn();
+            const ev = new MatrixEvent({
+                ...stickyEvent,
+                content: {},
+                origin_server_ts: Date.now(),
+            });
+            stickyEvents.addStickyEvents([ev]);
+            stickyEvents.on(RoomStickyEventsEvent.Update, emitSpy);
+            stickyEvents.handleRedaction(stickyEvent.event_id);
+            expect(emitSpy).toHaveBeenCalledWith([], [], [ev]);
```
toger5, line 303: How do we detect the difference between a remove of a keyed (in the map) and an unkeyed (in the set of unkeyed sticky events) event.

spec/unit/models/room-sticky-events.spec.ts
```diff
@@ -259,4 +259,163 @@ describe("RoomStickyEvents", () => {
...
+            stickyEvents.on(RoomStickyEventsEvent.Update, emitSpy);
+            // Mark the middle event as redacted.
+            middleEv.setUnsigned({
+                redacted_because: {
+                    event_id: "$foo",
+                } as any,
+            });
+            // Redact the newer event
+            stickyEvents.handleRedaction(newestEv.getId()!);
+            expect(emitSpy).toHaveBeenCalledWith([], [{ current: ev, previous: newestEv }], []);
```
toger5, line 371:
```suggestion
            // expect immediate transition from newestEv -> ev and skipping middleEv
            expect(emitSpy).toHaveBeenCalledWith([], [{ current: ev, previous: newestEv }], []);
```

spec/unit/models/room-sticky-events.spec.ts
```diff
@@ -259,4 +259,163 @@ describe("RoomStickyEvents", () => {
...
+        });
+        it("should handle redaction with a cached event", () => {
```
toger5: Do I understand this test right? you can call stickyEvents.handleRedaction with a string (id) or an ev object?
```suggestion
        it("should handle redaction when using `handleRedaction` with a `MatrixEvent` parameter", () => {
```

spec/unit/models/room-sticky-events.spec.ts
toger5, line 1: This is very nice!

src/models/room-sticky-events.ts
toger5: Cool. iirc the current spec requires that the expiration time for the same key is always the same.

src/models/room-sticky-events.ts
```diff
@@ -29,20 +29,54 @@ export type RoomStickyEventsMap = {
...
+    /**
+     * Sort two sticky events by order of expiry. This assumes the sticky events have the same
+     * `type`, `sticky_key` and `sender`.
+     * @returns A positive value if event A will expire sooner, or a negative value if event B will expire sooner.
+     */
+    private static sortStickyEvent(eventA: StickyMatrixEvent, eventB: StickyMatrixEvent): number {
+        // sticky events are not allowed to expire sooner than their predecessor.
+        if (eventB.getTs() !== eventA.getTs()) {
+            return eventB.getTs() - eventA.getTs();
+        }
```
toger5:
```suggestion
        // Sticky events with the same key have to use the same expiration duration.
        // Hence, comparing via `origin_server_ts` yields the exact same result as comparing their expiration time.
        if (eventB.getTs() !== eventA.getTs()) {
            return eventB.getTs() - eventA.getTs();
        }
```

src/models/room-sticky-events.ts
```diff
@@ -54,7 +88,7 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
      * @returns A matching active sticky event, or undefined.
      */
     public getKeyedStickyEvent(sender: string, type: string, stickyKey: string): StickyMatrixEvent | undefined {
-        return this.stickyEventsMap.get(type)?.get(`${stickyKey}${sender}`);
+        return this.stickyEventsMap.get(type)?.get(`${stickyKey}${sender}`)?.[0];
```
toger5: `${stickyKey}${sender}` is used in a couple of locations. To prohibit any potential `${sender}${stickey}` or `${stickyKey} ${sender}` mistakes we could make this a `stickyMapKey(sender:string, key: string)` method.

src/models/room-sticky-events.ts
```diff
@@ -148,7 +179,10 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
     public addStickyEvents(events: MatrixEvent[]): void {
         const added: StickyMatrixEvent[] = [];
         const updated: { current: StickyMatrixEvent; previous: StickyMatrixEvent }[] = [];
-        for (const event of events) {
+        // Ensure we insert by expiry time to preserve previous chain ordering.
+        for (const event of events
+            .filter((e) => e.unstableStickyExpiresAt)
+            .sort((a, b) => a.unstableStickyExpiresAt! - b.unstableStickyExpiresAt!)) {
```
toger5: why is this not using: `sortStickyEvent`

src/models/room-sticky-events.ts
```diff
@@ -229,6 +270,118 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
...
+                    // We did not find a previous event, so just expire.
+                    innerMap.delete(key);
+                    if (innerMap.size === 0) {
+                        this.stickyEventsMap.delete(eventType);
+                    }
+                    this.emit(RoomStickyEventsEvent.Update, [], [], [currentEvent]);
+                }
+                break;
+            }
+        }
```
toger5: How much overhead would it be to just call this method recursivly once we found the event?

spec/unit/models/room-sticky-events.spec.ts
Half-Shot: For apps, they can check whether the removed event has a sticky_key or not.

spec/unit/models/room-sticky-events.spec.ts
Half-Shot: Yes, intention being it speeds up our redaction handling process. The calling code may not have an event for the redacted_id so we have to handle both cases.

src/models/room-sticky-events.ts
Half-Shot: This can be removed now. `addStickyEvent` can handle any insertion order and so we don't need to do this work.

src/models/room-sticky-events.ts
Half-Shot: Cracking idea! I've implemented it. I think this is relatively cheap since the MatrixEvent param path is cheap, and this saves us a lot of code.

spec/unit/models/room-sticky-events.spec.ts
toger5: obviously!! That sounds like a nice solution. (Did not realize its that easy)

src/models/room-sticky-events.ts
```diff
@@ -88,7 +98,7 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
      * @returns A matching active sticky event, or undefined.
      */
     public getKeyedStickyEvent(sender: string, type: string, stickyKey: string): StickyMatrixEvent | undefined {
-        return this.stickyEventsMap.get(type)?.get(`${stickyKey}${sender}`)?.[0];
+        return this.stickyEventsMap.get(type)?.get(RoomStickyEventsStore.stickyMapKey(sender, stickyKey))?.[0];
```
toger5: It seems this is using swapped params. Its annoying that TS types only go that far... If sender would be of type UserId this would be a even better solution...

src/models/room-sticky-events.ts
```diff
@@ -152,7 +162,7 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
         // constrained so that a key will always end with a @<user_id>
         // E.g. Where a malicious event type might be "rtc.member.event@foo:bar" the key becomes:
         // "rtc.member.event.@foo:bar@bar:baz"
-        const innerMapKey = `${stickyKey}${sender}`;
+        const innerMapKey = RoomStickyEventsStore.stickyMapKey(sender, stickyKey);
```
toger5: also swapped

src/models/room-sticky-events.ts
```diff
@@ -297,17 +304,18 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
...
-            const eventType = redactEvent.getType();
+            const eventType = redactedEvent.getType();
+            const sender = redactedEvent.getSender();
             const innerMap = this.stickyEventsMap.get(eventType);
-            if (!innerMap) {
+            if (!innerMap || !sender) {
                 return;
             }
-            const mapKey = `${stickyKey}${redactEvent.getSender()}`;
+            const mapKey = RoomStickyEventsStore.stickyMapKey(sender, stickyKey);
```
toger5: also swapped

src/models/room-sticky-events.ts
```diff
@@ -24,25 +24,76 @@ export type RoomStickyEventsMap = {
...
+        if (eventB.getTs() !== eventA.getTs()) {
+            return eventB.getTs() - eventA.getTs();
+        }
+
+        if ((eventB.getId() ?? "") > (eventA.getId() ?? "")) {
+            return 1;
+        }
+
+        // This should fail as we've got corruption in our sticky array.
+        throw Error("Comparing two sticky events with the same event ID is not allowed.");
```
florianduros, line 73: Are we catching the error somewhere?

Half-Shot: There was one path where we were not. Have addressed.

PR #5038, opened 2025-10-10
MSC4140: support filters on delayed event lookup 

AndrewFerr:
Support looking up scheduled/finalised delayed events, and looking up a single delayed event.

## Checklist

### DIFF
--- a/spec/unit/matrix-client.spec.ts
+++ b/spec/unit/matrix-client.spec.ts
@@ -1055,11 +1055,11 @@ describe("MatrixClient", function () {
 
         describe("lookups", () => {
             const statuses = [undefined, "scheduled" as const, "finalised" as const];
-            const delayIdLists = [[], ["d123"], ["d789", "d789"]];
+            const delayIds = [undefined, "dxyz", ["d123"], ["d456", "d789"]];
             const inputs = statuses.flatMap((status) =>
-                delayIdLists.map((delayIds) => [status, delayIds] as [(typeof statuses)[0], (typeof delayIdLists)[0]]),
+                delayIds.map((delayId) => [status, delayId] as [(typeof statuses)[0], (typeof delayIds)[0]]),
             );
-            it.each(inputs)("can look up delayed events (status = %s, delayId = %s)", async (status, delayIds) => {
+            it.each(inputs)("can look up delayed events (status = %s, delayId = %s)", async (status, delayId) => {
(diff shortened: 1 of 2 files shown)


spec/unit/matrix-client.spec.ts
```diff
@@ -1053,17 +1053,27 @@ describe("MatrixClient", function () {
...
-                    method: "GET",
-                    prefix: unstableMSC4140Prefix,
-                    path: "/delayed_events",
-                    data: [],
-                },
-            ];
+        describe("lookups", () => {
+            for (const status of [undefined, "scheduled", "finalised"] as const) {
+                for (const delayId of [undefined, "d123"]) {
+                    it(`can look up ${delayId ? "a single " : ""}${status ? status + " " : ""}delayed event${delayId ? "" : "s"}`, async () => {
```
Half-Shot: You should be able to use it.each btw, which I think is preferred :)

src/client.ts
```diff
@@ -3537,21 +3537,30 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
                 "Server does not support the delayed events API",
                 "getDelayedEvents",
             );
         }
 
-        const queryDict = fromToken ? { from: fromToken } : undefined;
+        const queryDict = {
+            from: fromToken,
+            status,
+            // TODO: support multiple delayIds, but then query parameters must not be a dict
```
Half-Shot: What should we do about this TODO? A GitHub issue, or will we fix it in this PR?

AndrewFerr: IMO a new issue is better, as it would involve changing `FetchHttpApi.authedRequest` / `.request`, which would be a far-reaching change that's better to handle/test in a dedicated PR.

t3chguy: request must already support arrays as that's how via/server_name is specified in join

t3chguy: Image

The type `QueryDict` supports arrays.

AndrewFerr: Oh, perfect!  Thanks for pointing this out.

spec/unit/matrix-client.spec.ts
AndrewFerr: a2ad5b9ca

src/client.ts
AndrewFerr: f583c6e81

PR #5039, opened 2025-10-10
[MatrixRTC] Improve livekit object validation checks

toger5:
Otherwise passing undefined throws.

## Checklist

### DIFF
--- a/src/matrixrtc/LivekitTransport.ts
+++ b/src/matrixrtc/LivekitTransport.ts
@@ -21,8 +21,9 @@ export interface LivekitTransportConfig extends Transport {
     livekit_service_url: string;
 }
 
-export const isLivekitTransportConfig = (object: any): object is LivekitTransportConfig =>
-    object &&
+export const isLivekitTransportConfig = (object: unknown): object is LivekitTransportConfig =>
+    object !== null &&
+    typeof object === "object" &&
     "type" in object &&
     object.type === "livekit" &&
     "livekit_service_url" in object &&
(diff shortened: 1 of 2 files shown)


Half-Shot (approved):
This looks like it is doing what it is supposed to do.

src/matrixrtc/LivekitTransport.ts
```diff
@@ -22,7 +22,11 @@ export interface LivekitTransportConfig extends Transport {
 }
 
 export const isLivekitTransportConfig = (object: any): object is LivekitTransportConfig =>
```
Half-Shot: Ah, this would be better as `unknown`. That way Typescript is actually type checking.

src/matrixrtc/LivekitTransport.ts
```diff
@@ -43,4 +47,8 @@ export interface LivekitFocusSelection extends Transport {
  * @deprecated see LivekitFocusSelection
  */
 export const isLivekitFocusSelection = (object: any): object is LivekitFocusSelection =>
```
Half-Shot, line 50: ditto

Half-Shot (changes requested):
Sorry, missed a bit.

Half-Shot:
(There was some feedback that the title could be better in the web chapter, so I tried to tidy it up)

src/matrixrtc/LivekitTransport.ts
Half-Shot: @toger5 this also needs to be unknown

src/matrixrtc/LivekitTransport.ts
```diff
@@ -21,14 +21,19 @@ export interface LivekitTransportConfig extends Transport {
...
+    "livekit_service_url" in object &&
+    typeof object.livekit_service_url === "string";
 
 export interface LivekitTransport extends LivekitTransportConfig {
     livekit_alias: string;
 }
 
-export const isLivekitTransport = (object: any): object is LivekitTransport =>
+export const isLivekitTransport = (object: unknown): object is LivekitTransport =>
     isLivekitTransportConfig(object) && "livekit_alias" in object;
```
robintown, line 37: I think this also needs to check for `typeof object.livekit_alias === "string"`

PR #5054, opened 2025-10-22
Add `CryptoApi.getSecretStorageStatus`

richvdh:
`isSecretStorageReady` is a bit of a blunt instrument: it's hard to see from logs *why* the secret storage isn't ready.

Add a new method which returns a bit more data.

### DIFF
--- a/src/rust-crypto/rust-crypto.ts
+++ b/src/rust-crypto/rust-crypto.ts
@@ -858,7 +858,7 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
         };
 
         for (const secretName of secretsToCheck) {
-            // check which keys this particular secret is encrypted with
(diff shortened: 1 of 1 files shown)


src/rust-crypto/rust-crypto.ts
```diff
@@ -827,20 +828,46 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+
+        const result: SecretStorageStatus = {
+            // Assume we have all secrets until proven otherwise
+            ready: true,
+            defaultKeyId,
+            secretStorageKeyValidityMap: {},
+        };
+
+        for (const secretName of secretsToCheck) {
+            // check which keys this particular secret is encrypted with
```
andybalaam:
```suggestion
            // Check which keys this particular secret is encrypted with
```

PR #5066, opened 2025-10-30
Delayed event management: split endpoints, no auth

AndrewFerr:
Add dedicated endpoints for each of the cancel/restart/send actions for updating a delayed event, and make them unauthenticated.

Also keep support for the original endpoint where the update action is in the request body, and make the split-endpoint versions fall back to it if they are unsupported by the homeserver.

## Checklist

### DIFF
--- a/spec/unit/webrtc/mediaHandler.spec.ts
+++ b/spec/unit/webrtc/mediaHandler.spec.ts
@@ -61,10 +61,10 @@ describe("Media Handler", function () {
         expect(mockMediaDevices.getUserMedia).toHaveBeenCalledWith(
             expect.objectContaining({
                 audio: expect.objectContaining({
-                    deviceId: { ideal: FAKE_AUDIO_INPUT_ID },
+                    deviceId: { exact: FAKE_AUDIO_INPUT_ID },
                 }),
                 video: expect.objectContaining({
-                    deviceId: { ideal: FAKE_VIDEO_INPUT_ID },
+                    deviceId: { exact: FAKE_VIDEO_INPUT_ID },
                 }),
             }),
         );
@@ -77,7 +77,7 @@ describe("Media Handler", function () {
         expect(mockMediaDevices.getUserMedia).toHaveBeenCalledWith(
             expect.objectContaining({
                 audio: expect.objectContaining({
-                    deviceId: { ideal: FAKE_AUDIO_INPUT_ID },
+                    deviceId: { exact: FAKE_AUDIO_INPUT_ID },
                 }),
             }),
         );
@@ -109,7 +109,7 @@ describe("Media Handler", function () {
         expect(mockMediaDevices.getUserMedia).toHaveBeenCalledWith(
             expect.objectContaining({
                 video: expect.objectContaining({
-                    deviceId: { ideal: FAKE_VIDEO_INPUT_ID },
+                    deviceId: { exact: FAKE_VIDEO_INPUT_ID },
                 }),
             }),
         );
@@ -122,10 +122,10 @@ describe("Media Handler", function () {
         expect(mockMediaDevices.getUserMedia).toHaveBeenCalledWith(
             expect.objectContaining({
                 audio: expect.objectContaining({
-                    deviceId: { ideal: FAKE_AUDIO_INPUT_ID },
+                    deviceId: { exact: FAKE_AUDIO_INPUT_ID },
                 }),
                 video: expect.objectContaining({
-                    deviceId: { ideal: FAKE_VIDEO_INPUT_ID },
+                    deviceId: { exact: FAKE_VIDEO_INPUT_ID },
                 }),
             }),
         );
@@ -330,6 +330,80 @@ describe("Media Handler", function () {
             const stream = await mediaHandler.getUserMediaStream(true, false);
 
             expect(stream.getVideoTracks().length).toEqual(0);
+        });
+
+        it("falls back to ideal deviceId when exact deviceId fails", async () => {
+            // First call with exact should fail
+            mockMediaDevices.getUserMedia
+                .mockRejectedValueOnce(new Error("OverconstrainedError"))
+                .mockImplementation((constraints: MediaStreamConstraints) => {
+                    const stream = new MockMediaStream("local_stream");
+                    if (constraints.audio) {
+                        const track = new MockMediaStreamTrack("audio_track", "audio");
+                        track.settings = { deviceId: FAKE_AUDIO_INPUT_ID };
+                        stream.addTrack(track);
+                    }
+                    return Promise.resolve(stream.typed());
+                });
+
+            const stream = await mediaHandler.getUserMediaStream(true, false);
+
+            // Should have been called twice: once with exact, once with ideal
+            expect(mockMediaDevices.getUserMedia).toHaveBeenCalledTimes(2);
+            expect(mockMediaDevices.getUserMedia).toHaveBeenNthCalledWith(
+                1,
+                expect.objectContaining({
+                    audio: expect.objectContaining({
+                        deviceId: { exact: FAKE_AUDIO_INPUT_ID },
+                    }),
+                }),
+            );
+            expect(mockMediaDevices.getUserMedia).toHaveBeenNthCalledWith(
+                2,
+                expect.objectContaining({
+                    audio: expect.objectContaining({
+                        deviceId: { ideal: FAKE_AUDIO_INPUT_ID },
+                    }),
+                }),
+            );
+            expect(stream).toBeTruthy();
+        });
+
+        it("falls back to ideal deviceId for video when exact fails", async () => {
+            // First call with exact should fail
+            mockMediaDevices.getUserMedia
+                .mockRejectedValueOnce(new Error("OverconstrainedError"))
+                .mockImplementation((constraints: MediaStreamConstraints) => {
+                    const stream = new MockMediaStream("local_stream");
+                    if (constraints.video) {
+                        const track = new MockMediaStreamTrack("video_track", "video");
+                        track.settings = { deviceId: FAKE_VIDEO_INPUT_ID };
+                        stream.addTrack(track);
+                    }
+                    return Promise.resolve(stream.typed());
+                });
+
+            const stream = await mediaHandler.getUserMediaStream(false, true);
+
+            // Should have been called twice: once with exact, once with ideal
+            expect(mockMediaDevices.getUserMedia).toHaveBeenCalledTimes(2);
+            expect(mockMediaDevices.getUserMedia).toHaveBeenNthCalledWith(
+                1,
+                expect.objectContaining({
+                    video: expect.objectContaining({
+                        deviceId: { exact: FAKE_VIDEO_INPUT_ID },
+                    }),
+                }),
+            );
+            expect(mockMediaDevices.getUserMedia).toHaveBeenNthCalledWith(
+                2,
+                expect.objectContaining({
+                    video: expect.objectContaining({
+                        deviceId: { ideal: FAKE_VIDEO_INPUT_ID },
+                    }),
+                }),
+            );
+            expect(stream).toBeTruthy();
         });
     });
 
--- a/src/webrtc/mediaHandler.ts
+++ b/src/webrtc/mediaHandler.ts
@@ -264,8 +264,19 @@ export class MediaHandler extends TypedEventEmitter<
         }
 
         if (!canReuseStream) {
-            const constraints = this.getUserMediaContraints(shouldRequestAudio, shouldRequestVideo);
-            stream = await navigator.mediaDevices.getUserMedia(constraints);
+            let constraints: MediaStreamConstraints;
+            try {
+                // Not specifying exact for deviceId means switching devices does not always work,
+                // try with exact and fallback to ideal if it fails
+                constraints = this.getUserMediaContraints(shouldRequestAudio, shouldRequestVideo, true);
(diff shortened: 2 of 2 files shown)


src/client.ts
```diff
@@ -3583,14 +3583,100 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+     * Immediately send the delayed event matching the given delayId,
+     * instead of waiting for its scheduled delivery.
+     *
+     * Note: This endpoint is unstable, and can throw an `Error`.
+     *   Check progress on [MSC4140](https://github.com/matrix-org/matrix-spec-proposals/pull/4140) for more details.
+     *
+     * @throws A M_NOT_FOUND error if no matching delayed event could be found.
+     */
+    // eslint-disable-next-line
+    public async _unstable_sendScheduledDelayedEvent(
```
AndrewFerr, line 3636: Note that the reason these new methods are named "\***Scheduled**DelayedEvent" is because this method would otherwise be named `_unstable_sendDelayedEvent`, which would clash with the existing method for the `PUT /send/{eventType}/{txnId}?delay={ms}` endpoint.  "Scheduled" was put in this new method's name avoid that clash, and the cancel/restart method names have it too for the sake of consistency.

Alternatively, the existing `_unstable_sendDelayedEvent` could be renamed to `_unstable_scheduleDelayedEvent`, so that "schedule" would mean "prepare an event for delayed delivery" and "send" would mean "immediately send a delayed event now instead of waiting for its scheduled delivery time".

AndrewFerr:
Note that this now pulls in https://github.com/matrix-org/matrix-widget-api/pull/143, so that the split/unauthed endpoints are also invoked when using the widget API (meaning they'll be used by embedded Element Call).

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -157,13 +162,13 @@ describe("MembershipManager", () => {
             it("reschedules delayed leave event if sending state cancels it", async () => {
                 const memberManager = new MembershipManager(undefined, room, client, callSession);
                 const waitForSendState = waitForMockCall(client.sendStateEvent);
-                const waitForUpdateDelaye = waitForMockCallOnce(
-                    client._unstable_updateDelayedEvent,
+                const waitForRestartScheduledDelayedEvent = waitForMockCallOnce(
```
Half-Shot, line 165: Hah, this has the advantage now that we can test *what* type of update this is 🚀

spec/unit/embedded.spec.ts
```diff
@@ -531,17 +533,46 @@ describe("RoomWidgetClient", () => {
...
+                expect(widgetApi.requestCapability).toHaveBeenCalledWith(MatrixCapabilities.MSC4157UpdateDelayedEvent);
+                await client._unstable_restartScheduledDelayedEvent("id");
+                expect(widgetApi.restartScheduledDelayedEvent).toHaveBeenCalledWith("id");
+            });
+
+            it("can send scheduled delayed events (action in request path)", async () => {
+                await makeClient({ updateDelayedEvents: true, sendEvent: ["org.matrix.rageshake_request"] });
+                expect(widgetApi.requestCapability).toHaveBeenCalledWith(MatrixCapabilities.MSC4157UpdateDelayedEvent);
+                await client._unstable_sendScheduledDelayedEvent("id");
+                expect(widgetApi.sendScheduledDelayedEvent).toHaveBeenCalledWith("id");
```
Half-Shot: Struggling to see what the difference is between the two test types path v.s body? I don't see any differences in the blocks?

spec/unit/matrix-client.spec.ts
```diff
@@ -1077,21 +1081,140 @@ describe("MatrixClient", function () {
...
+                    prefix: unstableMSC4140Prefix,
+                    path: `/delayed_events/${encodeURIComponent(delayId)}`,
+                    data: {
+                        action: UpdateDelayedEventAction.Cancel,
+                    },
+                },
+            ];
+
+            await client._unstable_cancelScheduledDelayedEvent(delayId);
+            expect(httpLookups).toHaveLength(0);
```
Half-Shot, line 1194: We don't expect this for the preceding tests?

src/client.ts
```diff
@@ -3583,14 +3583,100 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+
+    /**
+     * Cancel the scheduled delivery of the delayed event matching the provided delayId.
+     *
+     * Note: This endpoint is unstable, and can throw an `Error`.
+     *   Check progress on [MSC4140](https://github.com/matrix-org/matrix-spec-proposals/pull/4140) for more details.
+     *
+     * @throws A M_NOT_FOUND error if no matching delayed event could be found.
+     */
+    // eslint-disable-next-line
```
Half-Shot:
```suggestion
    // eslint-disable-next-line @typescript-eslint/naming-convention
```

please, and for the others

src/client.ts
```diff
@@ -3583,14 +3583,100 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+
+    /**
+     * Restart the scheduled delivery of the delayed event matching the given delayId.
+     *
+     * Note: This endpoint is unstable, and can throw an `Error`.
+     *   Check progress on [MSC4140](https://github.com/matrix-org/matrix-spec-proposals/pull/4140) for more details.
+     *
+     * @throws A M_NOT_FOUND error if no matching delayed event could be found.
+     */
+    // eslint-disable-next-line
```
Half-Shot:
```suggestion
    // eslint-disable-next-line @typescript-eslint/naming-convention
```

src/client.ts
```diff
@@ -3583,14 +3583,100 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+    /**
+     * Immediately send the delayed event matching the given delayId,
+     * instead of waiting for its scheduled delivery.
+     *
+     * Note: This endpoint is unstable, and can throw an `Error`.
+     *   Check progress on [MSC4140](https://github.com/matrix-org/matrix-spec-proposals/pull/4140) for more details.
+     *
+     * @throws A M_NOT_FOUND error if no matching delayed event could be found.
+     */
+    // eslint-disable-next-line
```
Half-Shot:
```suggestion
    // eslint-disable-next-line @typescript-eslint/naming-convention
```

src/client.ts
Half-Shot: I wonder if the word immediately needs to appear somewhere, the scheduled in this function name doesn't suggest it's an immediate action until I check the docstring.

src/embedded.ts
```diff
@@ -461,15 +460,47 @@ export class RoomWidgetClient extends MatrixClient {
      * @experimental This currently relies on an unstable MSC (MSC4140).
      */
     // eslint-disable-next-line
```
Half-Shot:
```suggestion
    // eslint-disable-next-line @typescript-eslint/naming-convention
```

src/embedded.ts
```diff
@@ -461,15 +460,47 @@ export class RoomWidgetClient extends MatrixClient {
...
 
-        await this.widgetApi.updateDelayedEvent(delayId, action).catch(timeoutToConnectionError);
+        await this.widgetApi.cancelScheduledDelayedEvent(delayId).catch(timeoutToConnectionError);
+        return {};
+    }
+
+    /**
+     * @experimental This currently relies on an unstable MSC (MSC4140).
+     */
+    // eslint-disable-next-line
```
Half-Shot:
```suggestion
    // eslint-disable-next-line @typescript-eslint/naming-convention
```

src/embedded.ts
```diff
@@ -461,15 +460,47 @@ export class RoomWidgetClient extends MatrixClient {
...
+        }
+
+        await this.widgetApi.restartScheduledDelayedEvent(delayId).catch(timeoutToConnectionError);
+        return {};
+    }
+
+    /**
+     * @experimental This currently relies on an unstable MSC (MSC4140).
+     */
+    // eslint-disable-next-line
```
Half-Shot:
```suggestion
    // eslint-disable-next-line @typescript-eslint/naming-convention
```

spec/unit/embedded.spec.ts
AndrewFerr: Oops yes; they are currently the same & shouldn't be.  This got mangled by 2ffdb5f7f2fbbe6ff97440b7eafe7b7aae5f7433.  Taking care of this now

spec/unit/matrix-client.spec.ts
AndrewFerr: No, because the preceding tests call only a single endpoint, so checking that all expected requests were made is redundant with just checking that the `client` method call succeeds.

The purpose of it in the new tests is to ensure that both the (failed) action-split endpoint & (successful) non-split endpoint are requested.  If this were not checked, it could be that the action-split endpoint succeeded instead of failed, which would give no coverage on the fallback case.

spec/unit/embedded.spec.ts
AndrewFerr: fa6a2d341

Half-Shot (approved):
Looks good!

src/embedded.ts
```diff
@@ -469,7 +473,67 @@ export class RoomWidgetClient extends MatrixClient {
...
+        }
+        await updateDelayedEvent.call(this.widgetApi, delayId).catch(timeoutToConnectionError);
+        return {};
+    }
+
+    /**
+     * @experimental This currently relies on an unstable MSC (MSC4140).
+     */
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    public async _unstable_cancelScheduledDelayedEvent(delayId: string): Promise<EmptyObject> {
```
dbkr, line 496: Is there a reasn for this to return an empty object or is it copying sendToDevice? (not sure why that does either)

src/client.ts
```diff
@@ -3570,8 +3570,13 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      *
      * Note: This endpoint is unstable, and can throw an `Error`.
      *   Check progress on [MSC4140](https://github.com/matrix-org/matrix-spec-proposals/pull/4140) for more details.
+     *
+     * @deprecated Instead use one of:
```
dbkr, line 3574: Can we not just remove an unstable endpoint?

src/client.ts
```diff
@@ -3583,14 +3588,100 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            const path = utils.encodeUri("/delayed_events/$delayId/$action", {
+                $delayId: delayId,
+                $action: action,
+            });
+            return await this.http.request(Method.Post, path, undefined, undefined, {
+                ...requestOptions,
+                prefix: `${ClientPrefix.Unstable}/${UNSTABLE_MSC4140_DELAYED_EVENTS}`,
+            });
+        } catch (e) {
+            if (e instanceof MatrixError && e.errcode === "M_UNRECOGNIZED") {
```
dbkr, line 3665: This is for homeserver backwards compat, right? If so, could use a comment.

src/client.ts
```diff
@@ -3583,14 +3588,100 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+                return await this.updateScheduledDelayedEventWithActionInBody(delayId, action, requestOptions);
+            } else {
+                throw e;
+            }
+        }
+    }
+
+    private async updateScheduledDelayedEventWithActionInBody(
```
dbkr, line 3680: Comment on this method to say it's just for HSs that support the older MSC wouldn't go amiss

src/embedded.ts
AndrewFerr: This method is for handling a CS-API endpoint, so returning an empty object is for having a minimal response payload/body.  This is done for other endpoints too (including sendToDevice)

src/client.ts
AndrewFerr: Eventually yes, but I'd like for there to be at least one release with both versions supported, so that anything that depends on this SDK can have more time to migrate to the new version.

src/client.ts
AndrewFerr: Yes it is; good idea. 91fc7c598

That also made me realise there should also be backwards compat with the non-authed endpoint: c32829ee5

src/client.ts
AndrewFerr: 4d600fc6d

src/embedded.ts
dbkr: Still not sure why it needs to return anything at all though?

AndrewFerr: These methods of the embedded client override ones from `MatrixClient`, so they have to return `{}` to fit the return signature of `Promise<EmptyObject>` from the overridden methods.  And the reason the `MatrixClient` methods have such a signature is to follow the pattern of its other methods doing so when they wrap a CS-API endpoint that the server responds to with an empty JSON object.

I suppose the reason for that pattern is to handle the case of a server adding non-standard keys to a response body that's specced as empty.

dbkr: Mmm, I find it very weird that an API would relay through the empty object to the app: letting the app dig data out of a response object that might not be empty seems fairly awful, but you're right, this does seem to be the pattern that the js-sdk uses, by and large, so I guess consistency ftw.

dbkr (approved):

fkwp (approved):

PR #5069, opened 2025-11-05
Fix media switching during legacy calls

langleyd:
Fixes https://github.com/element-hq/element-web/issues/31123

element web PR is [here](https://github.com/element-hq/element-web/pull/31182) for testing the [netlify](https://pr31182--matrix-react-sdk.netlify.app/)

A couple of cases tested manually. The main case (case 1) is easily reproducible on macOS(all browsers I believe) and was reported for Windows. There was one report it couldn't be reproduced on Linux.

Case 1: Switing camera broken(With no fix applied this is easily reproducible)
1. Start a call with Alice
2. Switch device from camera 1 to camera 2
Observed:  Local feed stays on  camera 1 and remote feed from camera 1 received by Alice
Expected: Local feed changes successfully to camera and remote feed from camera 2 received by Alice

Case 2: (When only using "exact" this is reproducible, hence the need for the try/catch)
1. Start a call with Alice
2. Switch device from camera 1 to camera 2
3. End the call
4. Disconnect Camera 2
5. Start a call
Observed: An error occurs that media cannot be connected (as we have selected "exact" and the last device saved in settings was the disconnected camera)
Expected: Camera 2 falls back to Camera 1 and this is used for local/remote feeds.

### DIFF
--- a/spec/unit/webrtc/mediaHandler.spec.ts
+++ b/spec/unit/webrtc/mediaHandler.spec.ts
@@ -331,6 +331,80 @@ describe("Media Handler", function () {
 
             expect(stream.getVideoTracks().length).toEqual(0);
         });
+
+        it("falls back to ideal deviceId when exact deviceId fails", async () => {
+            // First call with exact should fail
+            mockMediaDevices.getUserMedia
+                .mockRejectedValueOnce(new Error("OverconstrainedError"))
+                .mockImplementation((constraints: MediaStreamConstraints) => {
+                    const stream = new MockMediaStream("local_stream");
+                    if (constraints.audio) {
+                        const track = new MockMediaStreamTrack("audio_track", "audio");
+                        track.settings = { deviceId: FAKE_AUDIO_INPUT_ID };
+                        stream.addTrack(track);
+                    }
+                    return Promise.resolve(stream.typed());
+                });
+
+            const stream = await mediaHandler.getUserMediaStream(true, false);
+
+            // Should have been called twice: once with exact, once with ideal
+            expect(mockMediaDevices.getUserMedia).toHaveBeenCalledTimes(2);
+            expect(mockMediaDevices.getUserMedia).toHaveBeenNthCalledWith(
+                1,
+                expect.objectContaining({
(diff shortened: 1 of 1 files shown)


src/webrtc/mediaHandler.ts
```diff
@@ -264,8 +264,19 @@ export class MediaHandler extends TypedEventEmitter<
         }
 
         if (!canReuseStream) {
-            const constraints = this.getUserMediaContraints(shouldRequestAudio, shouldRequestVideo);
-            stream = await navigator.mediaDevices.getUserMedia(constraints);
+            let constraints: MediaStreamConstraints;
+            try {
+                // Not specifying exact for deviceId means switching devices does not always work,
+                // try with exact and fallback to ideal if it fails
+                constraints = this.getUserMediaContraints(shouldRequestAudio, shouldRequestVideo, true);
```
BillCarsonFr, line 271: I don't understand, as per doc
`ideal`: A string or an array of strings, specifying ideal values for the property. If possible, one of the listed values will be used, but if it's not possible, the user agent will use the closest possible match.

Why doing `exact` then `ideal` would work differently than just `ideal`? what am I missing?

langleyd: on macOS, Using "ideal" at the moment is not working when switching camera and providing a specific deviceId. I don't know why maybe a system bug.

By the sounds of the comments [further down in the code](https://github.com/matrix-org/matrix-js-sdk/pull/5069/files#diff-b1711ebdbd6dd34ffdb450ceb11995ceeadc2b3e39bed9f33810a772825f1eb2R464-R467), this api is not very consistent cross browser either.

toger5 (commented):
What is needed to resolve this question:

> XXX: Is this still true?

toger5:
> By the sounds of the comment, the author was not sure if it was still an issue. We'd probably need to go back and have a look at what the original issue was for that. I don't think it would affect the solution for this problem, but I can check after the fact.

Thanks for this info. I missed that it is old code.
Since this is legacy call code I would not invest too much time and just leave the XXX comment as is.

PR #5080, opened 2025-11-20
Import room key bundles received after invite.

kaylendog:
Partially fixes https://github.com/element-hq/element-web/issues/30740. This currently does not work over a session reload, which I intend to introduce in a different PR.

## Checklist

### DIFF
--- a/src/common-crypto/CryptoBackend.ts
+++ b/src/common-crypto/CryptoBackend.ts
@@ -101,7 +101,7 @@ export interface CryptoBackend extends SyncCryptoCallbacks, CryptoApi {
      * the inviter's ID.
      *
      * @param roomId - The room we were invited to, for which we did not receive a key bundle before accepting the invite.
-     * @param inviterId - The user who invited us to the room and is expected to have sent the room key bundle.
+     * @param inviterId - The user who invited us to the room and is expected to send the room key bundle.
      */
     markRoomAsPendingKeyBundle(roomId: string, inviterId: string): void;
 }
--- a/src/rust-crypto/rust-crypto.ts
+++ b/src/rust-crypto/rust-crypto.ts
@@ -130,7 +130,7 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
     private roomEncryptors: Record<string, RoomEncryptor> = {};
 
     /** mapping of room ID -> inviter ID for rooms pending MSC4268 key bundles */
-    private readonly roomsPendingKeyBundles: Record<string, string> = {};
+    private readonly roomsPendingKeyBundles: Map<string, string> = new Map();
 
     private eventDecryptor: EventDecryptor;
     private keyClaimManager: KeyClaimManager;
@@ -401,7 +401,7 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
      * Implementation of {@link CryptoBackend.markRoomAsPendingKeyBundle}.
      */
     public markRoomAsPendingKeyBundle(roomId: string, inviter: string): void {
-        this.roomsPendingKeyBundles[roomId] = inviter;
+        this.roomsPendingKeyBundles.set(roomId, inviter);
     }
 
     ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////
@@ -1725,19 +1725,22 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
                     // received a new room key bundle that we might be able to download.
                     if (
                         isRoomKeyBundleMessage(parsedMessage) &&
-                        !!this.roomsPendingKeyBundles[parsedMessage.content.room_id]
+                        this.roomsPendingKeyBundles.has(parsedMessage.content.room_id)
                     ) {
                         // No `await`-ing here, as this is called from inside the `/sync` loop.
                         this.maybeAcceptKeyBundle(
                             parsedMessage.content.room_id,
-                            this.roomsPendingKeyBundles[parsedMessage.content.room_id],
+                            this.roomsPendingKeyBundles.get(parsedMessage.content.room_id)!,
                         ).then(
                             (success) => {
                                 if (success) {
-                                    delete this.roomsPendingKeyBundles[parsedMessage.content.room_id];
+                                    this.roomsPendingKeyBundles.delete(parsedMessage.content.room_id);
                                 }
                             },
                             (err) => {
+                                this.logger.error(
+                                    `Error attempting to download key bundle for room ${parsedMessage.content.room_id}`,
+                                );
                                 this.logger.error(err);
                             },
                         );


src/client.ts
```diff
@@ -2010,6 +2010,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         this.on(ClientEvent.Event, (event) => {
             rustCrypto.onLiveEventFromSync(event);
         });
+        this.on(ClientEvent.ReceivedToDeviceMessage, (payload) => rustCrypto.onReceiveToDeviceMessage(payload));
```
richvdh: we already have logic in `RustCrypto.preprocessToDeviceMessages` for handling incoming to-device messages: I don't think we need a separate entry point.

src/rust-crypto/rust-crypto.ts
```diff
@@ -2138,6 +2151,34 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
     public async getOwnIdentity(): Promise<RustSdkCryptoJs.OwnUserIdentity | undefined> {
         return await this.olmMachine.getIdentity(new RustSdkCryptoJs.UserId(this.userId));
     }
+    /**
```
richvdh: nit: missing blank line


```suggestion

    /**
```

src/client.ts
```diff
@@ -2408,7 +2409,12 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
 
         const roomId = res.room_id;
         if (opts.acceptSharedHistory && inviter && this.cryptoBackend) {
-            await this.cryptoBackend.maybeAcceptKeyBundle(roomId, inviter);
+            // Try to accept the room key bundle specified in a `m.room_key_bundle` to-device message we (might have) already received.
+            const bundleDownloaded = await this.cryptoBackend.maybeAcceptKeyBundle(roomId, inviter);
+            // If this fails, i.e. we haven't received this message yet, we need to wait until the to-device message arrives.
+            if (!bundleDownloaded) {
+                this.cryptoBackend.markRoomAsPendingKeyBundle(roomId, inviter);
+            }
```
richvdh, line 2414: I would say that we should combine the two methods into one, to save having this logic in MatrixClient, but maybe having them separate will be better once we come to having this work over restarts?

src/rust-crypto/rust-crypto.ts
```diff
@@ -2138,6 +2151,34 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+        if (typeof roomId !== "string") {
+            return;
+        }
+
+        // Check if the room is in the map of rooms we expect to receive bundles from, otherwise discard.
+        if (this.roomsPendingKeyBundles[roomId] !== claimedSender) {
+            return;
+        }
+
+        await this.maybeAcceptKeyBundle(roomId, claimedSender);
```
richvdh: we need to remove the entry from `roomsPendingKeyBundles` if this is successful?

src/rust-crypto/rust-crypto.ts
```diff
@@ -2138,6 +2151,34 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+     * @param payload - The received to-device message payload, which includes
+     * the message content and optional encryption information.
+     */
+    public async onReceiveToDeviceMessage(payload: ReceivedToDeviceMessage): Promise<void> {
+        if (payload.message.type != "io.element.msc4268.room_key_bundle") {
+            return;
+        }
+
+        const { message, encryptionInfo } = payload;
+        const claimedSender = encryptionInfo?.sender ?? message.sender;
```
richvdh: could do with a comment here explaining why we check both fields. (as per my other comment: I don't actually think we need this at all)

src/rust-crypto/rust-crypto.ts
```diff
@@ -2138,6 +2151,34 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+
+        // Validate room ID
```
richvdh: this comment, on the other hand, seems to be stating the blindingly obvious, and is therefore redundant.

src/rust-crypto/rust-crypto.ts
richvdh: It feels a bit surprising on first reading that we don't need to use the payload of the to-device message here. The answer to that, of course, is that the rust SDK will already have stashed the details of the to-device message in the database, so this is just a trigger for it to go and look again -- but we should probably write a comment to explain that to future readers.

src/rust-crypto/rust-crypto.ts
```diff
@@ -2138,6 +2151,34 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+        const roomId = message.content.room_id;
+        if (typeof roomId !== "string") {
+            return;
+        }
+
+        // Check if the room is in the map of rooms we expect to receive bundles from, otherwise discard.
+        if (this.roomsPendingKeyBundles[roomId] !== claimedSender) {
```
richvdh: I wonder if we actually need to bother checking the sender here, or if we can leave it up to the rust SDK to check. After all, we're not using any of the payload of the message, just reminding the rust SDK to do a re-check,

spec/integ/crypto/history-sharing.spec.ts
```diff
@@ -201,6 +204,104 @@ describe("History Sharing", () => {
...
+        const room = bobClient.getRoom(ROOM_ID);
+        expect(room).toBeTruthy();
+        expect(room?.getMyMembership()).toEqual(KnownMembership.Invite);
+
+        fetchMock.postOnce(`${BOB_HOMESERVER_URL}/_matrix/client/v3/join/${encodeURIComponent(ROOM_ID)}`, {
+            room_id: ROOM_ID,
+        });
+        await bobClient.joinRoom(ROOM_ID, { acceptSharedHistory: true });
+
+        // Bob receives and attempts decrypts the megolm message, but should not be able to.
```
richvdh:
```suggestion
        // Bob receives and attempts to decrypt the megolm message, but should yet not be able to.
```

spec/integ/crypto/history-sharing.spec.ts
```diff
@@ -201,6 +204,104 @@ describe("History Sharing", () => {
...
+        );
+        bobSyncResponder.sendOrQueueSyncResponse(bobSyncResponse);
+        await syncPromise(bobClient);
+        const bobRoom = bobClient.getRoom(ROOM_ID);
+        const event = bobRoom!.getLastLiveEvent()!;
+        expect(event.getId()).toEqual("$event_id");
+        await event.getDecryptionPromise();
+        expect(event.isDecryptionFailure()).toBeTruthy();
+
+        // Now the room key bundle arrives
```
richvdh:
```suggestion
        // Now the room key bundle message arrives
```

src/client.ts
kaylendog: An alternative could be to call `markRoomAsPendingKeyBundle` from `maybeAcceptKeyBundle` - keeps the distinction between the two for when we implement the restart functionality, while moving some complexity out of this method (`joinRoom`)?

src/rust-crypto/rust-crypto.ts
kaylendog: There's a tiny cost to copy the data into the WASM land which this does avoid, but I imagine it will be so small/infrequent it's not worth worrying about.

src/rust-crypto/rust-crypto.ts
kaylendog: Addressed in c6ff37c2691929c59ac99d1ab7f6e84826dd7f32

spec/integ/crypto/history-sharing.spec.ts
kaylendog: Addressed in c6ff37c2691929c59ac99d1ab7f6e84826dd7f32

spec/integ/crypto/history-sharing.spec.ts
kaylendog: Addressed in c6ff37c2691929c59ac99d1ab7f6e84826dd7f32

src/rust-crypto/rust-crypto.ts
kaylendog: Changed in 62a2af44235ef9f45b1c9cd409e87468d48d9b43!

src/rust-crypto/rust-crypto.ts
kaylendog: I've written a nice description above the moved code which should clarify this. See 62a2af44235ef9f45b1c9cd409e87468d48d9b43

kaylendog: Fixed in 62a2af44235ef9f45b1c9cd409e87468d48d9b43

src/client.ts
kaylendog: Refactored in 62a2af44235ef9f45b1c9cd409e87468d48d9b43

src/rust-crypto/rust-crypto.ts
```diff
@@ -130,7 +130,7 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
     private roomEncryptors: Record<string, RoomEncryptor> = {};
 
     /** mapping of room ID -> inviter ID for rooms pending MSC4268 key bundles */
```
richvdh, line 132: this comment is now outdated.

src/rust-crypto/rust-crypto.ts
```diff
@@ -129,6 +129,9 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
     /** mapping of roomId → encryptor class */
     private roomEncryptors: Record<string, RoomEncryptor> = {};
 
+    /** mapping of room ID -> inviter ID for rooms pending MSC4268 key bundles */
+    private roomsPendingKeyBundles: Set<string> = new Set();
```
richvdh: as sonarcloud says: may as well make this `readonly`

src/rust-crypto/rust-crypto.ts
```diff
@@ -1721,6 +1734,21 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+        // If any of the received events are room key bundle messages, and we have previously
+        // marked the room IDs they referenced as pending key bundles, tell the Rust SDK to
+        // try and accept them, just in case they were received after invite.
+        //
+        // We don't actually need to validate the contents of the bundle message, since the
+        // Rust SDK does that for us when it process it during the above `receiveSyncChanges`.
+        for (const payload of received
+            .filter(isRoomKeyBundleMessage)
+            .filter((payload) => this.roomsPendingKeyBundles.has(payload.message.content.room_id))) {
+            const success = await this.maybeAcceptKeyBundle(payload.message.content.room_id, payload.message.sender);
```
richvdh: `payload.message.sender` is the wrong thing here, isn't it? We only want to accept key bundles from the user that actually invited us to the room: passing `payload.message.sender` means we will accept key bundles from *any* user.

I think you need to keep roomsPendingKeyBundles as a Map.

src/rust-crypto/rust-crypto.ts
```diff
@@ -329,7 +332,7 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
     /**
      * Implementation of {@link CryptoBackend.maybeAcceptKeyBundle}.
      */
-    public async maybeAcceptKeyBundle(roomId: string, inviter: string): Promise<void> {
+    public async maybeAcceptKeyBundle(roomId: string, inviter: string): Promise<boolean> {
         // TODO: retry this if it gets interrupted or it fails. (https://github.com/matrix-org/matrix-rust-sdk/issues/5112)
         // TODO: do this in the background.
         // TODO: handle the bundle message arriving after the invite (https://github.com/element-hq/element-web/issues/30740)
```
richvdh: this TODO line needs an update

src/rust-crypto/rust-crypto.ts
```diff
@@ -1721,6 +1734,21 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
 
+        // If any of the received events are room key bundle messages, and we have previously
+        // marked the room IDs they referenced as pending key bundles, tell the Rust SDK to
+        // try and accept them, just in case they were received after invite.
+        //
+        // We don't actually need to validate the contents of the bundle message, since the
+        // Rust SDK does that for us when it process it during the above `receiveSyncChanges`.
+        for (const payload of received
+            .filter(isRoomKeyBundleMessage)
+            .filter((payload) => this.roomsPendingKeyBundles.has(payload.message.content.room_id))) {
```
richvdh: I wouldn't describe this as a `payload`, since (at this point) it may not have been encrypted.

You could fix that by either (1) moving it up into `case RustSdkCryptoJs.ProcessedToDeviceEventType.Decrypted` above, or (2) by changing this and `isRoomKeyBundleMessage` to talk about `message` instead of `payload`s.

src/rust-crypto/rust-crypto.ts
```diff
@@ -1721,6 +1734,21 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
             }
         }
 
+        // If any of the received events are room key bundle messages, and we have previously
+        // marked the room IDs they referenced as pending key bundles, tell the Rust SDK to
+        // try and accept them, just in case they were received after invite.
+        //
+        // We don't actually need to validate the contents of the bundle message, since the
+        // Rust SDK does that for us when it process it during the above `receiveSyncChanges`.
+        for (const payload of received
```
richvdh: it feels a bit clunky that we have two separate loops over the incoming messages, but I don't feel that strongly

src/rust-crypto/rust-crypto.ts
```diff
@@ -2470,3 +2498,30 @@ function rustEncryptionInfoToJsEncryptionInfo(
 
 type CryptoEvents = (typeof CryptoEvent)[keyof typeof CryptoEvent];
 type RustCryptoEvents = Exclude<CryptoEvents, CryptoEvent.LegacyCryptoStoreMigrationProgress>;
```
richvdh: suggest keeping these at the end, after the utility functions like `rustEncryptionInfoToJsEncryptionInfo` and `isRoomKeyBundleMessage`

src/client.ts
richvdh: > An alternative could be to call `markRoomAsPendingKeyBundle` from `maybeAcceptKeyBundle`

I don't think we would want to call `markRoomAsPendingKeyBundle` when `maybeAcceptKeyBundle` is called from `preprocessToDeviceMessages`?

I think it's fine as-is

richvdh (changes requested):
a few more comments. Main thing is the `inviter` param to `maybeAcceptKeyBundle`.

src/rust-crypto/rust-crypto.ts
kaylendog: Addressed in 48d36217b641097737bf662fdad5b193692d3733

src/rust-crypto/rust-crypto.ts
kaylendog: Addressed in 48d36217b641097737bf662fdad5b193692d3733

src/rust-crypto/rust-crypto.ts
kaylendog: Addressed in 48d36217b641097737bf662fdad5b193692d3733

src/rust-crypto/rust-crypto.ts
kaylendog: Addressed in 48d36217b641097737bf662fdad5b193692d3733

src/rust-crypto/rust-crypto.ts
kaylendog: Ended up addressing this in 48d36217b641097737bf662fdad5b193692d3733 as part of removing the `payload` name

src/rust-crypto/rust-crypto.ts
kaylendog: Addressed in 48d36217b641097737bf662fdad5b193692d3733

src/common-crypto/CryptoBackend.ts
```diff
@@ -90,8 +90,20 @@ export interface CryptoBackend extends SyncCryptoCallbacks, CryptoApi {
...
+     */
+    maybeAcceptKeyBundle(roomId: string, inviter: string): Promise<boolean>;
+
+    /**
+     * Mark a room as pending a key bundle under MSC4268. The backend will listen for room key bundle messages, and if
+     * it sees one matching the room specified, it will automatically import it as long as the message author's ID matches
+     * the inviter's ID.
+     *
+     * @param roomId - The room we were invited to, for which we did not receive a key bundle before accepting the invite.
+     * @param inviterId - The user who invited us to the room and is expected to have sent the room key bundle.
```
richvdh:
```suggestion
     * @param inviterId - The user who invited us to the room and is expected to send the room key bundle.
```

src/rust-crypto/rust-crypto.ts
```diff
@@ -130,7 +130,7 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
     private roomEncryptors: Record<string, RoomEncryptor> = {};
 
     /** mapping of room ID -> inviter ID for rooms pending MSC4268 key bundles */
-    private roomsPendingKeyBundles: Set<string> = new Set();
+    private readonly roomsPendingKeyBundles: Record<string, string> = {};
```
richvdh: generally we try to use `Map` rather than `Record`, to avoid any danger of [prototype pollution](https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/Prototype_pollution)

src/rust-crypto/rust-crypto.ts
```diff
@@ -1716,6 +1715,34 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+                            parsedMessage.content.room_id,
+                            this.roomsPendingKeyBundles[parsedMessage.content.room_id],
+                        ).then(
+                            (success) => {
+                                if (success) {
+                                    delete this.roomsPendingKeyBundles[parsedMessage.content.room_id];
+                                }
+                            },
+                            (err) => {
+                                this.logger.error(err);
```
richvdh, line 1744: suggest giving a bit more context here, so that the logs are more useful: `error attempting to download key bundle for room ${roomId}` or so

src/common-crypto/CryptoBackend.ts
kaylendog: Addressed in f9ea67306a8ef6a0c07c7d87ee1c80f36849d3bf

src/rust-crypto/rust-crypto.ts
kaylendog: Addressed in f9ea67306a8ef6a0c07c7d87ee1c80f36849d3bf

src/rust-crypto/rust-crypto.ts
kaylendog: Addressed in f9ea67306a8ef6a0c07c7d87ee1c80f36849d3bf

PR #5103, opened 2025-12-10
[MatrixRTC] Remove three status cases that will never be set

toger5:
## Checklist


toger5:
I almost suspect that their existence was caused by first thinking what states this class might have and the list was never revised after the actual implementation of the status.

Half-Shot:
We seem to be using [Reconnecting in Element-Call](https://github.com/element-hq/element-call/blob/67d20a8f3dcb1769351249f9b58e821a4240d1a9/src/state/CallViewModel/CallViewModel.test.ts#L1259C52-L1259C64). In the future can we update Element Call to use PRs that might impact it, so CI can run through?

PR #5107, opened 2025-12-11
Implement MSC4387: M_SAFETY error

Half-Shot:
Implementation of MSC4387.

### DIFF
--- a/spec/unit/http-api/utils.spec.ts
+++ b/spec/unit/http-api/utils.spec.ts
@@ -195,7 +195,7 @@ describe("parseErrorResponse", () => {
         {
             errcode: MatrixSafetyErrorCode.name,
             error: "Spammy",
-            expiry: Date.now() + 5000,
+            expiry: 5000,
         },
         {
             errcode: MatrixSafetyErrorCode.name,


spec/integ/matrix-client-opts.spec.ts
```diff
@@ -159,7 +159,7 @@ describe("MatrixClient opts", function () {
 
             await expect(
                 Promise.all([client.sendTextMessage("!foo:bar", "a body", "txn1"), httpBackend.flush("/txn1", 1)]),
-            ).rejects.toThrow("MatrixError: [500] Unknown message");
+            ).rejects.toThrow("MatrixError: [500] Ruh roh");
```
Half-Shot, line 162: I'm surprised this wasn't caught before, but seems to me this has always been a bug.

src/client.ts
```diff
@@ -2481,8 +2481,8 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @param event - The event to resend.
      * @param room - Optional. The room the event is in. Will update the
      * timeline entry if provided.
-     * @returns Promise which resolves: to an ISendEventResponse object
-     * @returns Rejects: with an error response.
+     * @throws May throw a `MatrixSafetyError` if content is deemed unsafe.
+     * @see MatrixSafetyError
```
Half-Shot, line 2485: Reviewer, you may be wondering "Why has Half-Shot written a bunch of `@ see MatrixSafetyError` statements randomly across the file".

The intention here is to match the MSC's list of endpoints that may emit this error. I felt it was sensible to alert client developers that these endpoints may confer a specific error type if the user is trying to search or send harmful content, as oppose to usual failures.

I'm on the fence about overall usefulness, and could be persuaded they are overkill. However, it's done now so I'm going to leave unless noises are made.

dbkr (approved):

PR #5108, opened 2025-12-12
Start CallEventHandler and GroupCallEventHandler if isVoipWithNoMediaAllowed is set to true

alfredjophy:
Start CallEventHandler and GroupCallEventHandler when  `isVoipWithNoMediaAllowed` is set to true. This lets you use the SDK on the server for managing calls.


alfredjophy:
Yes thanks for the reply. I do agree the flag might not be really meant for this. Ill be using a private flag for this

PR #5109, opened 2025-12-15
Re-check outgoing requests after processing them

andybalaam:
... in case any new requests have been added during processing.

Fixes https://github.com/element-hq/element-web/issues/30988

Notes: Fix element-web #30988 by re-checking outgoing requests after processing a batch

### DIFF
--- a/src/rust-crypto/OutgoingRequestsManager.ts
+++ b/src/rust-crypto/OutgoingRequestsManager.ts
@@ -139,5 +139,29 @@ export class OutgoingRequestsManager {
                 this.logger.error(`Failed to process outgoing request ${request.type}: ${e}`);
             }
         }
+
+        // If we handled any requests this time, more may have been queued as
+        // part of that handling, so do an extra check to make sure we handle
+        // them immediately. See
+        // If we handled any requests this time, more may have been queued as
+        // part of that handling.
+        //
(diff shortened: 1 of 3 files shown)


src/rust-crypto/OutgoingRequestsManager.ts
```diff
@@ -139,5 +139,19 @@ export class OutgoingRequestsManager {
                 this.logger.error(`Failed to process outgoing request ${request.type}: ${e}`);
             }
         }
+
+        // If we handled any requests this time, more may have been queued as
+        // part of that handling, so do an extra check to make sure we handle
+        // them immediately. See
+        // https://github.com/element-hq/element-web/issues/30988
```
richvdh: It would be nice to give an example here.

```suggestion
        // If we handled any requests this time, more may have been queued as
        // part of that handling.
        //
        // For example, we may have processed a `/keys/claim` request, which
        // meant the rust side could establish an Olm session and is now ready to
        // send out an `m.secret.send` message. 
        // (See https://github.com/element-hq/element-web/issues/30988.)
        //
        // So, if we have processed any requests, flag that we need make another
        // pass around the outgoing-requests loop, to make sure we handle any
        // pending requests immediately.
```

PR #5110, opened 2025-12-15
MatrixRTC Pseudonymous livekit identities

toger5:
## Context:
 - Element Call is going to use pseudonymous identities on the SFU.
 - This is (currently) LK specific
 - We did use the `livekitParticipantIdentity` in the js-sdk (THIS SHOULD STOP)
 - The js-sdk instead now uses parts of the call membership to identify a user and leave the actual id used on the media backend up for EC.
 
### The important bits:

First we remove everything that tries to be a `participantId` and replace it with `membership: Pick<CallMembership, "userId"|"deviceId"|"memberId">`. We also deprecate `membershipID` and replace it with `memberId`.

 - **memberId** should be the UUID for a user participation (same device after reconnect -> new id). But as of now it was just a alias for the stateKey. This is now much more explicit: for the proper rtc memberships (sticky events) the memberId is part of the MembershipManager. For the old SessionMemberships the memberId is hard coded to `userId:deviceId` as we currently expect it on the livekit sfu.

CallMembership also exposes `kind` so that EC can use the sha256 version for sticky events and the string version for session events (state events).

(can be reviewed commit by commit)

## Checklist

### DIFF
--- a/src/matrixrtc/RTCEncryptionManager.ts
+++ b/src/matrixrtc/RTCEncryptionManager.ts
@@ -61,7 +61,7 @@ export class RTCEncryptionManager implements IEncryptionManager {
     private participantKeyRings = new Map<
         EncryptionKeyMapKey,
         Array<{
-            key: Uint8Array;
+            key: Uint8Array<ArrayBuffer>;
             keyIndex: number;
             membership: CallMembershipIdentityParts;
             rtcBackendIdentity: string;
@@ -130,7 +130,7 @@ export class RTCEncryptionManager implements IEncryptionManager {
         private statistics: Statistics,
         // Callback to notify the media layer of new keys
         private onEncryptionKeysChanged: (
-            keyBin: Uint8Array,
+            keyBin: Uint8Array<ArrayBuffer>,
             encryptionKeyIndex: number,
             membership: CallMembershipIdentityParts,
             rtcBackendIdentity: string,
@@ -161,7 +161,7 @@ export class RTCEncryptionManager implements IEncryptionManager {
     public getEncryptionKeys(): ReadonlyMap<
         EncryptionKeyMapKey,
         ReadonlyArray<{
-            key: Uint8Array;
+            key: Uint8Array<ArrayBuffer>;
             keyIndex: number;
             membership: CallMembershipIdentityParts;
             rtcBackendIdentity: string;
@@ -171,7 +171,7 @@ export class RTCEncryptionManager implements IEncryptionManager {
     }
 
     private keysWithoutMatchingRTCMembership: Array<{
-        key: Uint8Array;
+        key: Uint8Array<ArrayBuffer>;
         keyIndex: number;
         membership: CallMembershipIdentityParts;
     }> = [];
@@ -184,7 +184,11 @@ export class RTCEncryptionManager implements IEncryptionManager {
         });
     }
 
-    private addKeyToParticipant(key: Uint8Array, keyIndex: number, membership: CallMembershipIdentityParts): void {
+    private addKeyToParticipant(
+        key: Uint8Array<ArrayBuffer>,
+        keyIndex: number,
+        membership: CallMembershipIdentityParts,
+    ): void {
         const knownRtcMembership = this.getMemberships();
         const fullMembership = knownRtcMembership.find(
             (member) => member.userId === membership.userId && member.deviceId === membership.deviceId,
@@ -200,7 +204,7 @@ export class RTCEncryptionManager implements IEncryptionManager {
     }
 
     private addKeyToParticipantWithBackendIdentity(
-        key: Uint8Array,
+        key: Uint8Array<ArrayBuffer>,
         keyIndex: number,
         membership: CallMembershipIdentityParts,
         rtcBackendIdentity: string,
@@ -450,7 +454,7 @@ export class RTCEncryptionManager implements IEncryptionManager {
         return 0;
     }
 
-    private generateRandomKey(): Uint8Array {
+    private generateRandomKey(): Uint8Array<ArrayBuffer> {
         const key = new Uint8Array(16);
         globalThis.crypto.getRandomValues(key);
         return key;
--- a/src/@types/global.d.ts
+++ b/src/@types/global.d.ts
@@ -62,6 +62,6 @@ declare global {
 
     interface Uint8ArrayConstructor {
         // https://tc39.es/proposal-arraybuffer-base64/spec/#sec-uint8array.frombase64
-        fromBase64?(base64: string, options?: Uint8ArrayFromBase64Options): Uint8Array;
+        fromBase64?(base64: string, options?: Uint8ArrayFromBase64Options): Uint8Array<ArrayBuffer>;
     }
 }
--- a/src/base64.ts
+++ b/src/base64.ts
@@ -64,7 +64,7 @@ export function encodeUnpaddedBase64Url(uint8Array: Uint8Array): string {
     return toBase64(uint8Array, { alphabet: "base64url", omitPadding: true });
 }
 
-function fromBase64(base64: string, options: Uint8ArrayFromBase64Options): Uint8Array {
+function fromBase64(base64: string, options: Uint8ArrayFromBase64Options): Uint8Array<ArrayBuffer> {
     if (typeof Uint8Array.fromBase64 === "function") {
         // Currently this is only supported in Firefox,
         // but we match the options in the hope in the future we can rely on it for all environments.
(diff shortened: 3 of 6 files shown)


src/matrixrtc/RoomKeyTransport.ts
```diff
@@ -174,8 +177,8 @@ export class RoomKeyTransport
                 );
                 this.emit(
                     KeyTransportEvents.ReceivedKeys,
-                    userId,
-                    deviceId,
+                    // Using `${userId}:${deviceId}` makes no sense (but works). It does not matter since the RoomKeyTransport is deprecated
```
BillCarsonFr, line 180: BTW, we should remove it

toger5: I was thinking about this. In this pr? Or as a follow up?

src/matrixrtc/CallMembership.ts
```diff
@@ -21,7 +21,11 @@ import { slotDescriptionToId, slotIdToDescription, type SlotDescription } from "
 import type { RTCCallIntent, Transport } from "./types.ts";
 import { type IContent, type MatrixEvent } from "../models/event.ts";
 import { type RelationType } from "../@types/event.ts";
-import { logger } from "../logger.ts";
+import { logger as rootLogger } from "../logger.ts";
+import { sha256 } from "../digest.ts";
+import { encodeUnpaddedBase64Url } from "../base64.ts";
+
+const logger = rootLogger.getChild("[CallMembership]");
```
BillCarsonFr: I prefer when the parent looger is passed to the constructor, and the actual logger a private field.
I found it helping when I needed to specialize the prefix per instance
WDYT?

src/matrixrtc/CallMembership.ts
```diff
@@ -270,29 +274,75 @@ export class CallMembership {
...
+        /**
+         *
+         * Anonymised identity to use with the RTC backend.
+         *
+         * The rtcBackendIdentity is a hashed version of all the identity parts:
+         * `${this.userId}|${this.deviceId}|${this.memberId}`
+         *
+         * It is used to anonymize the identity of the user in the RTC backend.
+         *
+         * The computation of this value is async and should be done right after constructing the CallMembership instance.
```
BillCarsonFr: This comments looks outdated? what does it mean that it should be computed after constructing if it is already there?

src/matrixrtc/IMembershipManager.ts
```diff
@@ -26,11 +26,17 @@ export enum MembershipManagerEvent {
...
+     * Once the membershipManger has aquired the a delay id (after sending the state event)
+     * It will emit and share the delay id.
+     */
+    DelayIdChanged = "DelayIdChanged",
 }
 
 export type MembershipManagerEventHandlerMap = {
     [MembershipManagerEvent.StatusChanged]: (prefStatus: Status, newStatus: Status) => void;
     [MembershipManagerEvent.ProbablyLeft]: (probablyLeft: boolean) => void;
+    [MembershipManagerEvent.DelayIdChanged]: (delayId: string | undefined) => void;
```
BillCarsonFr, line 39: Is this related to the current PR?

src/matrixrtc/MatrixRTCSessionManager.ts
```diff
@@ -136,10 +136,10 @@ export class MatrixRTCSessionManager extends TypedEventEmitter<MatrixRTCSessionM
             return;
         }
 
-        this.refreshRoom(room);
+        void this.refreshRoom(room);
     };
 
-    private refreshRoom(room: Room): void {
+    private async refreshRoom(room: Room): Promise<void> {
```
BillCarsonFr, line 142: This looks to be always called with `void` and never waited. Could it then internally try/catch and log any error?

src/matrixrtc/RoomKeyTransport.ts
BillCarsonFr: follow up

src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -174,38 +184,39 @@ export class RTCEncryptionManager implements IEncryptionManager {
...
-        this.logger?.debug(`Received key over transport ${userId}:${deviceId} at index ${index}`);
+        this.logger?.debug(`Received key over transport ${membership.userId}:${membership.deviceId} at index ${index}`);
 
         // We received a new key, notify the video layer of this new key so that it can decrypt the frames properly.
-        const participantId = getParticipantId(userId, deviceId);
         const keyBin = decodeBase64(keyBase64Encoded);
         const candidateInboundSession: InboundEncryptionSession = {
             key: keyBin,
-            participantId,
+            membership,
```
BillCarsonFr, line 287: I don't understand why this cannot just stay the partcipant id?

src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -174,38 +184,39 @@ export class RTCEncryptionManager implements IEncryptionManager {
         }
     }
 
-    public onNewKeyReceived: KeyTransportEventListener = (userId, deviceId, keyBase64Encoded, index, timestamp) => {
+    public onNewKeyReceived: KeyTransportEventListener = (membership, keyBase64Encoded, index, timestamp) => {
```
BillCarsonFr, line 271: can it just be participant id? or is it for logging purpose?

src/matrixrtc/utils.ts
```diff
@@ -14,14 +14,15 @@ See the License for the specific language governing permissions and
...
+import { getEncryptionKeyMapKey, type CallMembershipIdentityParts } from "./EncryptionManager.ts";
+import { type InboundEncryptionSession, type EncryptionKeyMapKey } from "./types.ts";
 
 /**
  * Detects when a key for a given index is outdated.
  */
 export class OutdatedKeyFilter {
     // Map of participantId -> keyIndex -> timestamp
-    private tsBuffer: Map<ParticipantId, Map<number, number>> = new Map();
+    private tsBuffer: Map<EncryptionKeyMapKey, Map<number, number>> = new Map();
```
BillCarsonFr, line 25: same I am not sure to understand why it cannot just be the participant id?

src/matrixrtc/CallMembership.ts
toger5: yes adding the user name to the prefix also seems like a good idea!

src/matrixrtc/CallMembership.ts
toger5: indeed

src/matrixrtc/IMembershipManager.ts
toger5: Yes. This is the js-sdk PR that prepares for both element call PR's: Pseudonomous identities + delayed event delegation.

toger5: It is possible to move everything just for the delayed event delegation to another PR? wdyt?

src/matrixrtc/RTCEncryptionManager.ts
toger5: Because the participantId (rtcBackendIdentifier) is computed async and not part of toDevice message.
We have the same problem we get with the CallMembership events:
 - we get an event and have to create additional data for that event (rtcBackendIdentifier) in order to use it.
 - we can do that before letting the rtc sdk know about the event
 - we can do this just before using the event
 - OR (what we are currently doing) we can use the CallMembership computed rtcBackendIdentifier and only use the `MembershipIdetityParts` to match the key event to the CallMembership event.

src/matrixrtc/RTCEncryptionManager.ts
toger5: same as above. the async rtcBackendIdentifier is not yet known. The `membership: MembershipIdentityParts = type {userId, deviceId, memberId}` is all we know and need to match the key with the CallMembership event.

src/matrixrtc/utils.ts
toger5: Here we do not want to compute the sha256 again (already done in the membership event) and want to have an easier to debug key for the `Map`. It makes this code easier to understand in its stand-alone complexity since we do not introduce (and need to document) the specification and reason for the sha256 based backend identity string.

The prop name makes it clear, that the only purpose of this string is a unique key in this map. (at this point this really is the only purpuse it has)

src/matrixrtc/RTCEncryptionManager.ts
BillCarsonFr: That makes me wonder why it is not part of the toDevice message. If we don't do that there will be a problem no?

toger5: Yes that is definitely the alternative. We would need to change the MSC for that however. But still very valid. To some extend that is the only thing the to-device message needs to know: "what media stream do I belong to" And if we already construct a identifier for the rtcBackendIdentity we should use exactly that...

BillCarsonFr (approved):
After live peer review, fine for me.
I updated the tests for encryption manager, and I noticed that it must know our own rtcBackendIdentity early in order to send the own key as soon as possible. It cannot wait for the "remote echo" of the call membership

PR #5111, opened 2025-12-16
Mark `forwardingCurve25519KeyChain` as deprecated

richvdh:
The Rust SDK always populates this as an empty array, so we may as well get rid of it.

### DIFF
--- a/src/common-crypto/CryptoBackend.ts
+++ b/src/common-crypto/CryptoBackend.ts
@@ -200,8 +200,9 @@ export interface EventDecryptionResult {
      */
     clearEvent: IClearEvent;
     /**
-     * List of curve25519 keys involved in telling us about the senderCurve25519Key and claimedEd25519Key.
+     * No longer used.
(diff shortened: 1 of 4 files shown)


src/common-crypto/CryptoBackend.ts
```diff
@@ -200,8 +200,9 @@ export interface EventDecryptionResult {
      */
     clearEvent: IClearEvent;
     /**
-     * List of curve25519 keys involved in telling us about the senderCurve25519Key and claimedEd25519Key.
+     * No longer used.
      * See {@link MatrixEvent#getForwardingCurve25519KeyChain}.
+     * @deprecated
      */
     forwardingCurve25519KeyChain?: string[];
```
richvdh, line 207: I would have got rid of this altogether, but it turns out that EventDecryptionResult is exposed outside the js-sdk, so doing so would be a breaking change. See https://github.com/matrix-org/matrix-js-sdk/pull/5112 for a PR to deprecate its public export, at least.

PR #5113, opened 2025-12-17
Make the enableEncryptedStateEvents property on MatrixClient public

andybalaam:
This supports https://github.com/element-hq/element-web/pull/31513 which is part of https://github.com/matrix-org/matrix-rust-sdk/issues/5397

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1219,7 +1219,21 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     private cryptoBackend?: CryptoBackend; // one of crypto or rustCrypto
 
     /**
-     * Support MSC4362: Simplified Encrypted State Events
+     * Support MSC4362: Simplified Encrypted State Events.
+     *
+     * The client must be recreated for changes to this setting to take effect
+     * reliably.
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -1217,7 +1217,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     public http: MatrixHttpApi<IHttpOpts & { onlyData: true }>; // XXX: Intended private, used in code.
 
     private cryptoBackend?: CryptoBackend; // one of crypto or rustCrypto
-    private readonly enableEncryptedStateEvents: boolean;
+    public enableEncryptedStateEvents: boolean;
```
florianduros, line 1238: If it's public (also for private), should we have tsdoc?

andybalaam: Thanks, added in f24d11ad2

t3chguy: It should explain what changing it does. Will changing it apply or do you need to restart the client. Will it retroactively decrypt state events?

andybalaam: Fixed in 10936f28a

PR #5117, opened 2026-01-05
Make MatrixRTC encryption key types narrower for TS 5.9 compatibility

robintown:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-9.html#libdts-changes

TypeScript 5.9 changes some things about the `ArrayBuffer` type and makes a number of DOM types, including the subtle crypto APIs, require a narrower buffer type as their input. For example if you wanted to use `crypto.subtle.importKey` to convert a MatrixRTC encryption key buffer given by matrix-js-sdk to a `CryptoKey`, you would run into a type error with TS 5.9. Specifying the type parameter of `Uint8Array` everywhere around the MatrixRTC files fixes this breakage.

### DIFF
--- a/src/@types/global.d.ts
+++ b/src/@types/global.d.ts
@@ -62,6 +62,6 @@ declare global {
 
     interface Uint8ArrayConstructor {
         // https://tc39.es/proposal-arraybuffer-base64/spec/#sec-uint8array.frombase64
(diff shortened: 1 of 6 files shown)


t3chguy (approved):

dbkr (approved):

BillCarsonFr (approved):
ok

PR #5118, opened 2026-01-05
Remove unused property `MatrixEvent.untrusted`

richvdh:
This was never set to anything other than `false`. I think it is a hangover from pre-rust-sdk.

### DIFF
--- a/spec/unit/matrixrtc/CallMembership.spec.ts
+++ b/spec/unit/matrixrtc/CallMembership.spec.ts
@@ -28,7 +28,14 @@ function makeMockEvent(originTs = 0): MatrixEvent {
         getTs: jest.fn().mockReturnValue(originTs),
         getSender: jest.fn().mockReturnValue("@alice:example.org"),
         getId: jest.fn().mockReturnValue("$eventid"),
+        getContent: jest.fn().mockReturnValue({}),
     } as unknown as MatrixEvent;
+}
+
+function createCallMembership(ev: MatrixEvent, content: unknown): CallMembership {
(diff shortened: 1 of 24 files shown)


src/common-crypto/CryptoBackend.ts
```diff
@@ -213,11 +213,6 @@ export interface EventDecryptionResult {
      * ed25519 key claimed by the sender of this event. See {@link MatrixEvent#getClaimedEd25519Key}.
      */
     claimedEd25519Key?: string;
-    /**
-     * Whether the keys for this event have been received via an unauthenticated source (eg via key forwards, or
-     * restored from backup)
-     */
-    untrusted?: boolean;
```
t3chguy, line 220: Isn't this a breaking change?

richvdh: this is an internal interface

src/models/event.ts
```diff
@@ -1022,7 +1016,7 @@ export class MatrixEvent extends TypedEventEmitter<MatrixEventEmittedEvents, Mat
         this.clearEvent = decryptionResult.clearEvent;
         this.senderCurve25519Key = decryptionResult.senderCurve25519Key ?? null;
         this.claimedEd25519Key = decryptionResult.claimedEd25519Key ?? null;
-        this.untrusted = decryptionResult.untrusted || false;
+        this.untrusted = false;
```
richvdh:
```suggestion
```

src/models/event.ts
richvdh:
```suggestion
```

t3chguy (commented):
CI looks unhappy

PR #5131, opened 2026-01-09
Switch from Jest to Vitest

t3chguy:
Depends on https://github.com/matrix-org/matrix-js-sdk/pull/5136

Fairly simple transition:
+ `@fetch-mock/jest` -> `@fetch-mock/vitest`
+ `jest` -> `vi` mostly

Some smaller tweaks which should be obvious

Tests run much faster locally with significantly less memory pressure, in CI they go from about 3m -> 2m

### DIFF
--- a/babel.config.cjs
+++ b/babel.config.cjs
@@ -7,9 +7,15 @@ module.exports = {
                 targets: {
                     esmodules: true,
                 },
+                modules: false,
             },
         ],
-        ["@babel/preset-typescript"],
+        [
+            "@babel/preset-typescript",
+            {
+                rewriteImportExtensions: true,
+            },
+        ],
     ],
     plugins: [
         ["@babel/plugin-proposal-decorators", { version: "2023-11" }],


spec/unit/matrix-client.spec.ts
```diff
@@ -412,10 +408,9 @@ describe("MatrixClient", function () {
         // means they may call /events and then fail an expect() which will fail
         // a DIFFERENT test (pollution between tests!) - we return unresolved
         // promises to stop the client from continuing to run.
-        mocked(client.http.authedRequest).mockImplementation(function () {
+        vi.mocked(client.http.authedRequest).mockImplementation(function () {
             return new Promise(() => {});
         });
-        client.stopClient();
```
dbkr, line 418: Do we still need to do this or does vitest make it unnecessary?

t3chguy: Depends on the side of the test file, the thread gets discarded between tests but a huge test file may benefit from it. Unfortunately any unhandled exceptions (such as those from a fetch being aborted) can upset the test during cleanup.

andybalaam (approved):

spec/unit/matrix-client.spec.ts
dbkr: Fair enough.

PR #5133, opened 2026-01-09
MSC4140: add dedicated endpoint for delayed events

AndrewFerr:
Use a new, dedicated endpoint for scheduling delayed events. This should deprecate the `delay` query parameter on the `/send` and `/state` endpoints.

See:
- https://github.com/matrix-org/matrix-spec-proposals/pull/4140/commits/1618990017ed2c96ebd00b5ce6de7c61333f4bb6
- https://github.com/matrix-org/matrix-spec-proposals/pull/4140/commits/8a98ec46b501001cd859375697851456567497eb

## Checklist

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -3127,17 +3127,16 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         }
 
         if (delayOpts) {
-            return this.http.authedRequest<SendDelayedEventResponse>(
+            return await this.http.authedRequest<SendDelayedEventResponse>(
                 Method.Put,
                 path,
                 { ...getUnstableDelayQueryOpts(delayOpts), ...queryOpts },
                 content,
             );
         } else {
-            return this.http.authedRequest<ISendEventResponse>(Method.Put, path, queryOpts, content).then((res) => {
-                this.logger.debug(`Event sent to ${event.getRoomId()} with event id ${res.event_id}`);
-                return res;
-            });
+            const res = await this.http.authedRequest<ISendEventResponse>(Method.Put, path, queryOpts, content);
+            this.logger.debug(`Event sent to ${event.getRoomId()} with event id ${res.event_id}`);
+            return res;
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -3557,13 +3585,39 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            return await this.http.authedRequest(
+                Method.Put,
+                utils.encodeUri("/rooms/$roomId/state/$eventType/$stateKey", {
+                    $stateKey: stateKey,
+                    ...pathParams,
+                }),
+                getUnstableDelayQueryOpts(delayOpts),
+                content,
+                opts,
+            );
```
t3chguy, line 3619: It'd be better to adapt the MSC to communicate over `/versions` or `/capabilities` so as to not need to make 2 calls every time for the majority of servers

AndrewFerr: Using `/capabilities` is a great idea.  I wasn't aware that it allowed custom identifiers.

The MSC already defines a `/versions` key for whether the feature is enabled or not (which Synapse and the js-sdk make use of), so `/capabilities` is a good place to put more fine-grained details on the implementation of it.

AndrewFerr: Update: spec folks have informed me that this is more under the purview of `/versions` than `/capabilities`.  I'll update the MSC with appropriate new `/versions` keys before putting them in this PR.

src/client.ts
toger5, line 3064: Since we now separate the two endpoints (delay vs send) I am not sure I am at all happy with this function. 
 - We do complicated overloads which are only used to differentiate between the return types (delayId vs eventId)
 - And in the body we then completely split the impl based on the params.

In other words: It all looks to me like there should be `sendEventHttpRequest` and `sendDelayedEventHttpRequest`

toger5: I am not sure how deep we should go. I think encrypt and send event is suffering from the same overload concept. and `sendCompleteEvent` event as well. Not sure if they both should also be seperate functions?

PR #5056, opened 2025-10-28
[MatrixRTC] Implement MSC4143 slots

Half-Shot:
## Checklist

- [ ] Tests written for new code (and old code if feasible).
- [ ] New or updated `public`/`exported` symbols have accurate [TSDoc](https://tsdoc.org/) documentation.
- [ ] Linter and other CI checks pass.
- [ ] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md)).

PR #5134, opened 2026-01-09
RTC Slots: Refactoring of membership event parsing and handling

Half-Shot:
Split out from https://github.com/matrix-org/matrix-js-sdk/pull/5056

Part of a series of PRs intending to support Matrix RTC MSC4143 slot behaviours.

This one focuses on refactoring the way we handle membership from different eras of Matrix RTC, as currently we need to support both the older and newer forms of MSC4143.

This does NOT implement slots itself, but forms the basis for doing so.

Changes in this PR:

 - Splits out the Session and RTC membership types and parsers into separate files.
 - Tidies up CallMembership to use proper enums when parsing data out of the above formats.
 - Ensures sticky events are **only** supported with the newer system.
 - Checks the event type when checking the event contents, no more double parsing to work out the "best" format.

## Checklist

### DIFF
--- a/src/matrixrtc/index.ts
+++ b/src/matrixrtc/index.ts
@@ -19,5 +19,6 @@ export * from "./LivekitTransport.ts";
 export * from "./MatrixRTCSession.ts";
 export * from "./MatrixRTCSessionManager.ts";
 export type * from "./types.ts";
+export { type SessionMembershipData, type RtcMembershipData } from "./membershipData/index.ts";
 export { Status, parseCallNotificationContent, isMyMembership } from "./types.ts";
 export { MembershipManagerEvent } from "./IMembershipManager.ts";


spec/unit/matrixrtc/CallMembership.spec.ts
```diff
@@ -371,43 +394,4 @@ describe("CallMembership", () => {
...
-        });
-
-        it("calculates time until expiry", () => {
-            vi.setSystemTime(2000);
-            // should be using absolute expiry time
-            expect(membership.getMsUntilExpiry()).toEqual(DEFAULT_EXPIRE_DURATION - 1000);
-        });
-    });
-
-    it("uses unpadded base64 for RTC backend identities", async () => {
```
Half-Shot, line 425: Erp, this still needs to be tested.

src/matrixrtc/CallMembership.ts
```diff
@@ -567,45 +365,38 @@ export class CallMembership {
...
+                return undefined;
         }
-        return undefined;
     }
 
-    /**
-     * The focus_active filed of the session membership (m.call.member).
-     * @deprecated focus_active is not used and will be removed in future versions.
-     */
-    public getFocusActive(): LivekitFocusSelection | undefined {
```
Half-Shot, line 630: Entirely unused, let's ditch it.

src/matrixrtc/CallMembership.ts
```diff
@@ -32,331 +36,119 @@ import { type Logger } from "../logger.ts";
...
-        this.matrixEventData = { eventId, sender, ts };
-
-        this.logger = logger?.getChild(`[CallMembership ${sender}:${this.deviceId}]`);
-    }
-
-    /**
-     * sha256(`${this.userId}|${this.deviceId}|${this.memberId}`) for sticky events (kind = rtc)
-     * `${this.userId}:${this.deviceId}` for state events (kind = session)
-     */
-    public static async computeRtcBackendIdentity(
```
Half-Shot, line 310: Now moved to the respective type files.

src/matrixrtc/membershipData/rtc.ts
```diff
@@ -0,0 +1,148 @@
...
+    };
+    "m.relates_to"?: {
+        event_id: string;
+        rel_type: RelationType.Reference;
+    };
+    "application": RtcSlotEventContent["application"];
+    "rtc_transports": Transport[];
+    "versions": string[];
+    "msc4354_sticky_key"?: string;
+    "sticky_key"?: string;
```
Half-Shot, line 44: Note, this is missing `disconnect_reason`. I'm going to do a followup PR so not to make this PR any larger, it's already huge.

src/matrixrtc/CallMembership.ts
```diff
@@ -32,331 +36,121 @@ import { type Logger } from "../logger.ts";
...
-     * (for example caused by a homeserver crashes)
-     **/
-    "expires"?: number;
+type MembershipData =
+    | { kind: MembershipKind.RTC; data: RtcMembershipData }
+    | { kind: MembershipKind.Session; data: SessionMembershipData };
 
+type LimitedEvent = Pick<MatrixEvent, "getId" | "getSender" | "getTs" | "getType" | "getContent">;
+// TODO: Rename to RtcMembership once we removed the legacy SessionMembership is removed, to avoid confusion.
+export class CallMembership {
```
Half-Shot, line 59: N.B. There was a strong temptation to make this a abstract class and split the types into different classes. Because I think that is ultimately more invasive & we don't expect to maintain both types forever, I'm going to keep the existing `switch` case style.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -277,6 +264,11 @@ export class MatrixRTCSession extends TypedEventEmitter<
 
     public memberships: CallMembership[] = [];
 
+    /**
+     * Resolves when the session has calculated the initial membership of the session.
+     */
+    public readonly initialMembershipCalculated: Promise<void>;
```
Half-Shot, line 268: This replaces a lot of the `flushPromises` calls in the test, since there was potential for racing.

src/matrixrtc/membershipData/rtc.ts
toger5, line 1: I am wondering if the folder name is misleading. I would expect CallMembership/RtcMembership to also be in there.

Maybe call it membershipData? (or `membership_data`)

src/matrixrtc/CallMembership.ts
toger5, line 1: I really want to rename this to `RtcMembership`. Do you think this is the PR in which this should happen?

Is it as easy as that:

```
//file: src/matrixrtc/CallMembership.ts
/** @deprecated */
class CallMembership extends RtcMembership {}
// eof
```
To make the rename a non breaking change?

src/matrixrtc/CallMembership.ts
toger5: I see where you are coming from. In this specific case I have found it highly useful that that its done in a switch case style.

I had a LOT of questions like: how is this done in the different event types again?

And this way i can read just one block without jumping from file to file.

Not sure if this is helpful in any way to reduce your temptation :)

src/matrixrtc/membershipData/session.ts
toger5, line 1: Can this be called `session`?

There is another style of member events where we had one state event for all memberships per user (an array of memberships)

This is what we have called legacy. It got removed (finally) but still is used in some logs with the legacy reference.

src/matrixrtc/membership/session.ts
```diff
@@ -0,0 +1,146 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { type IContent } from "../../matrix.ts";
+import { type RTCCallIntent, type Transport } from "../types.ts";
+import { MatrixRTCMembershipParseError } from "./common.ts";
+
+/**
+ * **Legacy** (MatrixRTC) session membership data.
```
toger5: `Legacy` -> `State event based` in comments. I think that is the most clear what session events are.

src/matrixrtc/membership/session.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    // application specific data (we first need to check if they exist)
+    if (data.scope !== undefined && typeof data.scope !== "string") errors.push(prefix + "scope must be string");
+
+    if (data["m.call.intent"] !== undefined && typeof data["m.call.intent"] !== "string") {
+        errors.push(prefix + "m.call.intent must be a string");
+    }
+
+    if (errors.length) {
+        throw new MatrixRTCMembershipParseError("bar", errors);
```
toger5: "bar"? (just a stop gap message?)

src/matrixrtc/CallMembership.ts
```diff
@@ -32,331 +36,118 @@ import { type Logger } from "../logger.ts";
...
+        const evType = matrixEvent.getType();
+        const data = matrixEvent.getContent();
+        if (sender === undefined) throw new Error("matrixEvent is missing sender field");
+        try {
+            // Event types are strictly checked here.
+            if (evType === EventType.RTCMembership && checkRtcMembershipData(data, sender)) {
+                return { kind: MembershipKind.RTC, data };
+            } else if (evType === EventType.GroupCallMemberPrefix && checkSessionsMembershipData(data)) {
+                return { kind: MembershipKind.Session, data };
+            } else {
```
toger5, line 77: checking for the type first is smart! Its good we simplified the possible combinations to only allow rtc in sticky events with a new type.

src/matrixrtc/CallMembership.ts
```diff
@@ -384,76 +175,70 @@ export class CallMembership {
...
     public get applicationData(): { type: string; [key: string]: unknown } {
         const { kind, data } = this.membershipData;
         switch (kind) {
-            case "rtc":
+            case MembershipKind.RTC:
                 return data.application;
-            case "session":
+            case MembershipKind.Session:
             default:
+                // XXX: This is a hack around
```
toger5: What is the hack here?
```suggestion
                // SessionData does not have a seperated application prop. We derive it from top level prams.
```

src/matrixrtc/CallMembership.ts
```diff
@@ -478,25 +263,33 @@ export class CallMembership {
...
+                throw Error("Not possible to get memberID without knowing the membership event kind");
         }
     }
 
+    /**
+     * @deprecated renamed to `memberId`
+     */
+    public get membershipID(): string {
+        return this.memberId;
+    }
```
toger5, line 322: Do you think the next version can already drop this deprecated field?

src/matrixrtc/CallMembership.ts
toger5: Its not used in any downstream projects?
It definitly was deprecated in a prev version. So definilty fine to remove it either way.

src/matrixrtc/MembershipManager.ts
```diff
@@ -1101,7 +1098,7 @@ export class StickyEventMembershipManager extends MembershipManager {
             },
             slot_id: slotDescriptionToId(this.slotDescription),
             rtc_transports: this.rtcTransport ? [this.rtcTransport] : [],
-            member: { device_id: this.deviceId, user_id: this.userId, id: this.memberId },
+            member: { claimed_user_id: this.userId, claimed_device_id: this.deviceId, id: this.memberId },
```
toger5: This is a very big change. It really needs a shout out in the PR description!

Its a proper breaking change (we also need the breaking gh lable) since it will change the events we sent to the matrix dag.

src/matrixrtc/types.ts
```diff
@@ -199,3 +199,23 @@ export interface Transport {
...
+        // other application specific keys
+        [key: string]: unknown;
+    };
+    slot_id: string;
+}
+
+/**
+ * The session description is used to identify a session. Used in the state event.
+ */
+export interface SlotDescription {
```
toger5, line 218: I wonder if this should be called `SlotIdentifier`. I know why its not. It is easy to confuse with `slot.id`.
But it is an identifier...

Do we use the `Description` wording anywhere else?
What about `SlotIdentifierParts`
That would fit: `OwnMembershipIdentifierParts`

toger5: Maybe not worth worrying about now. If you have a very good idea/strong opinion take this comment as green light for that.

Otherwise we can just leave it as `Description`

src/matrixrtc/utils.ts
```diff
@@ -47,3 +47,18 @@ export class OutdatedKeyFilter {
         return false;
     }
 }
+
+/**
+ * Converts a slot ID into it's component application and ID portions.
+ * @param slotId e.g. `m.call#call_id`
+ */
+export function slotIdToDescription(slotId: string): SlotDescription {
```
toger5, line 56: We definitly have this pattern twice
member.id
member.userId
member.deviceId

and the hashed rtcBackendIdentity
MembershipIdentityParts


And this whole slot concept
slotId = rtcBackendIdentity
slot.id
slotDescription = MembershipIdentityParts


Seeing this makes me think this PR is the right place to unify it.

toger5 (commented):
This is really nice!

src/matrixrtc/CallMembership.ts
Half-Shot: I'm willing to bet they will be able to adapt their code. We can make the slots release a breaking release, and that can encourage them to update.

src/matrixrtc/CallMembership.ts
Half-Shot: Yeah, I think for the moment this is fine. And anyway, it's hopefully going away in the future.

src/matrixrtc/membershipData/rtc.ts
Half-Shot: Or we could just move the CallMembership/RtcMembership into the folder so it's colocated, wdyt?

src/matrixrtc/CallMembership.ts
Half-Shot: I'd *rather* wait until we drop the Session stuff, which will be a breaking change, so we can do a nice hard cutoff at the same time. I think if we rename now with little-to-none functional changes, it might be surprising if it breaks later if that makes sense?

src/matrixrtc/membershipData/session.ts
Half-Shot: Yeah, I think session makes more sense. Legacy is also what we call the webrtc 1:1 stuff sometimes. I think we should still plaster legacy over the docstrings though :)

src/matrixrtc/membership/session.ts
Half-Shot: Yep, lazy lazy lazy. Fixed.

src/matrixrtc/CallMembership.ts
Half-Shot: Oh yes definitely, let's delete this on the rename to RtcMembership.

src/matrixrtc/MembershipManager.ts
Half-Shot: Yep yep, we should call this out explicitly in the changelog.

src/matrixrtc/types.ts
Half-Shot: I'm keen on description because it's a composite of type and ID. Since it's used internally-ish, we can break it when we move to RTCMember if we decide it's rubbish.

src/matrixrtc/utils.ts
Half-Shot: I think I need to understand this a bit more to unify it? Can you give me some example code that would be unified :)

src/matrixrtc/membershipData/rtc.ts
toger5: This can then also include the rename to `claimed_...`

src/matrixrtc/CallMembership.ts
toger5: Yes the aim should be to drop the code before abstracting it!

src/matrixrtc/membershipData/rtc.ts
toger5: I prefer them on a different level. `membership/RtcMembership` is odd.
`membership/rtc` collides with the data file. So i think what we need is a seperation between the files that describe the raw event content data and the RtcMemberhsip class

src/matrixrtc/CallMembership.ts
toger5: that is why I was wondering if:
`class CallMembership extends RtcMembership {}` can still expose the CallMembership naming so its not a breaking change.
And that the breaking change might already be adjusted for users of the sdk.

src/matrixrtc/utils.ts
toger5: There are two concepts that get their unique identifier based on a set of params
`MembershipIdentityParts { id, userId, deviceId}`
And the slot description
`SlotDescription { id, application }`
Thy both have some method to convert to it to a single key (needed for matrix and livekit limitations. The slot state event cannot have multi val keys, the lk identities as well)

We call them `rtcBackendIdentity = "stringComputeExampleFunc"(MembershipIdentityParts { id, userId, deviceId})` and
`slotId = "stringComputeExampleFunc"(SlotDescription { id, application })`

They are really similar concept wise but the wording is very different

toger5: The concepts i see are:
 id parts: the set of core fields that together make the unique identifier
 identity: the combined single string representation of those fields
 and the fields `id` `userId` `application` `deviceId` that do not need a name

src/matrixrtc/membershipData/rtc.ts
toger5: So my recommendation is to make this PR a non breaking change. Its always annoying if the breaking change PRs are very large.

Than we do a super small PR adding `disconnect_reason` and rename to `claimed_...` but nothing else?

Half-Shot: Yeah, happy to do a followup :+1:

src/matrixrtc/membershipData/rtc.ts
Half-Shot: :+1:  understood.

src/matrixrtc/CallMembership.ts
Half-Shot: I just don't really see the point of doing that now. 

> And that the breaking change might already be adjusted for users of the sdk.

This actually makes things worse, because this PR introduces `RtcMembership` which at first appears to be the same as `CallMembership`, and then changes in another PR. If we rename it alongside the breaking changes, SDK developers will be more likely to pick up on the other changes (since the class has been removed).

I'm fairly high temp on this one

src/matrixrtc/MembershipManager.ts
Half-Shot: We're now going to do this in the next PR.

src/matrixrtc/CallMembership.ts
toger5: > developers will be more likely to pick up on the other changes

I can see that one.

I will remove it from the summary and resolve this thread

src/matrixrtc/utils.ts
Half-Shot: Ah gotcha, I see your point. I'm slightly cautious because the former does some fun hashing and promise returning, and the latter really just stitches together two string values into a slot ID. But yeah, I can see how that makes sense. I'm happy to go with `computeSlotId`?

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -797,15 +784,20 @@ export class MatrixRTCSession extends TypedEventEmitter<
...
-                    this.ensureRecalculateSessionMembers();
-                    this.recalculateSessionMembersDirty = false;
-                }
-            });
+                setTimeout(() => {
+                    if (this.recalculateSessionMembersDirty) {
+                        void this.ensureRecalculateSessionMembers();
+                        this.recalculateSessionMembersDirty = false;
+                    }
+                }, 1);
```
robintown: Why is this check no longer synchronous with the setting of `recalculateSessionMembersPromise` to `undefined`? It seems to me that this would introduce a new race: If the dirty flag is true, then someone calls `recalculateSessionMembersPromise` after the promise has been set to undefined but before the timeout expires, you will eventually end up with two concurrent `recalculateSessionMembers` promises.

src/matrixrtc/types.ts
```diff
@@ -199,3 +199,23 @@ export interface Transport {
...
+    id?: string;
```
robintown: What does an undefined `id` mean?

Side note, a doc comment saying that this is the "application-specific slot ID" from the MSC would help me keep this separate in my head from the "slot IDs" that you get by templating `application` and `id` together.

src/matrixrtc/utils.ts
```diff
@@ -47,3 +47,18 @@ export class OutdatedKeyFilter {
...
+    const [application, id] = slotId.split("#");
+    return { application, id };
```
robintown: Rethinking https://github.com/matrix-org/matrix-js-sdk/pull/5022#discussion_r2414527371. Can we make sure that `m.rtc.slot` state events with more than one `#` in the state key are rejected entirely by the validation code? MSC4143 says:

> The # character MUST NOT be used in either `application.type` or `application_slot_id`

and I think matrix-js-sdk ought to enforce this rule so that I can't be tricked into, for example, giving a malicious widget the capability to control the invalid `m.call#ROOM#MY_WIDGET` slot, thinking that it won't have the ability to overwrite/confuse the settings of the `m.call#ROOM` slot as computed by my client. (`"m.call#ROOM#MY_WIDGET".split("#")` = `["m.call", "ROOM", "MY_WIDGET"]`, and this code would silently ignore the `"MY_WIDGET"` element.)

src/matrixrtc/ToDeviceKeyTransport.ts
```diff
@@ -74,7 +74,7 @@ export class ToDeviceKeyTransport
             },
             room_id: this.roomId,
             member: {
-                claimed_device_id: this.membership.deviceId,
+                device_id: this.membership.deviceId,
```
robintown: Is this a backwards-compatible change?

Relatedly, I wonder when we're going to start sending & accepting `org.matrix.msc4143.rtc.encryption_key` messages. Wouldn't that be a requirement for EC's Matrix 2.0 compatibility? @toger5

src/matrixrtc/CallMembership.ts
```diff
@@ -32,331 +36,118 @@ import { type Logger } from "../logger.ts";
...
      */
-    "msc4354_sticky_key"?: string;
+    public static membershipDataFromMatrixEvent(matrixEvent: LimitedEvent): MembershipData {
+        const sender = matrixEvent.getSender();
+        const evType = matrixEvent.getType();
+        const data = matrixEvent.getContent();
+        if (sender === undefined) throw new Error("matrixEvent is missing sender field");
+        try {
+            // Event types are strictly checked here.
+            if (evType === EventType.RTCMembership && checkRtcMembershipData(data, sender)) {
```
robintown, line 73: Note: I believe the `IContent` type returned by `matrixEvent.getContent()` is a lie and we would be safer having the input type of `checkRtcMembershipData` be `object` to ensure that it really checks `m.relates_to` and stuff.

src/matrixrtc/types.ts
Half-Shot: It's very confusing that we have a slot ID, and then within that there is a {type}#{id}. I'll try some words.

src/matrixrtc/MatrixRTCSession.ts
Half-Shot: I think I missed that in the change actually.

src/matrixrtc/ToDeviceKeyTransport.ts
toger5: In the ToDeviceKeyTransport we actually want this to be `claimed_device_id`.
This porobably got "reverted" by the change that went from `claimed_device_id` -> `device_id` in the rtc_member_event.

(there it was `device_id` before this PR. And we decided to leave it at `device_id` so its not a breaking change)

toger5: see: https://github.com/matrix-org/matrix-js-sdk/pull/5134#discussion_r2712835374

src/matrixrtc/CallMembership.ts
Half-Shot: IContent appears to be entirely made up of optionals:

```ts
export interface IContent {
    [key: string]: any;
    "msgtype"?: MsgType | string;
    "membership"?: Membership;
    "avatar_url"?: string;
    "displayname"?: string;
    "m.relates_to"?: IEventRelation;

    "m.mentions"?: IMentions;
}
export interface IEventRelation {
    "rel_type"?: RelationType | string;
    "event_id"?: string;
    "is_falling_back"?: boolean;
    "m.in_reply_to"?: {
        event_id?: string;
    };
    "key"?: string;
}
```

So in practical terms it's just as safe, and `object` would remove the hints.

src/matrixrtc/utils.ts
Half-Shot: Done in [5889015](https://github.com/matrix-org/matrix-js-sdk/pull/5134/commits/5889015c285c2208a42200c865f4e9ee47d97f5d). I've opted to failing it *here* and then checking it in our validation function so that we don't accept it in any place in the code.

src/matrixrtc/types.ts
Half-Shot: Words were tried.

src/matrixrtc/MatrixRTCSession.ts
Half-Shot: I've refactored this a bit to hopefully satisfy things. I don't think we have a race since the logic is as follows:

- `membershipNeedsRecalculation` starts as false
- `ensureRecalculateSessionMembers` is called, sets `membershipNeedsRecalculation` to true and kicks off a recalculation
- `recalculateSessionMembers` immediately sets `membershipNeedsRecalculation` to false and then begins doing work
  - *any calls* made to `ensureRecalculateSessionMembers` during this work session would flip `membershipNeedsRecalculation` again
  - The first call made here would await the result of the last membership check before running another check, subsequent calls are ignored.

I've tried to document this a bit more as I go.

src/matrixrtc/ToDeviceKeyTransport.ts
Half-Shot: That does look like a mistake, well caught.

src/matrixrtc/membershipData/common.ts
```diff
@@ -0,0 +1,27 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * Thrown when an event is not valid for use with MatrixRTC.
+ */
+export class MatrixRTCMembershipParseError extends Error {
```
MidhunSureshR: Why not use an [AggregateError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError)?

src/matrixrtc/CallMembership.ts
```diff
@@ -32,331 +36,118 @@ import { type Logger } from "../logger.ts";
...
-
-    /**
-     * A list of possible foci this user knows about. One of them might be used based on the focus_active
-     * selection system.
-     */
-    "foci_preferred": Transport[];
-
+ * Describes the source event type that provided the membership data.
+ */
+enum MembershipKind {
```
MidhunSureshR, line 42: This class seems to have a lot of:
```ts
switch(kind) {
  case MembershipKind.RTC: 
      behaviour_rtc();
  case MembershipKind.Session:
      behaviour_session();
}
```

I see some potential to split this class into two: `RtcCallMembership` and `SessionClassMembership`.  But  maybe we'll remove the session related code soon?

src/matrixrtc/utils.ts
```diff
@@ -47,3 +47,24 @@ export class OutdatedKeyFilter {
...
+export function slotIdToDescription(slotId: string): SlotDescription {
+    const [application, id, ...unexpectedAdditionalValues] = slotId.split("#");
+    if (unexpectedAdditionalValues.length) {
+        throw Error(
+            "MatrixRTC Slot IDs *must* only contain two components seperated by one '#'. Additional '#' characters detected.",
+        );
+    }
+    return { application, id };
+}
+/**
```
MidhunSureshR, line 66:
```suggestion
}

/**
```

src/matrixrtc/CallMembership.ts
Half-Shot: Yep, session code is going in the bin as soon as the RTC code has been in production for a while :). We talked about splitting it out, but I suspect we're just going to delete all the legacy code in the next couple of months.

PR #5135, opened 2026-01-12
Fix reactive display name disambiguation

aditya-cherukuru:
## Problem

Display name disambiguation (showing user ID suffix like `Alice (@alice:server)`) was not updating reactively when room members changed their display names.

**Steps to reproduce:**
1. Alice, Bob, Charlie are in a room
2. Bob changes display name to "Alice"
3. Expected: Disambiguation flags should update for all members sharing the name
4. Actual: Disambiguation flags only update after a page refresh

**Root Cause:**
`updateDisplayNameCache()` in `room-state.ts` correctly updates the display name → userId mapping, but did NOT recalculate the `disambiguate` flag for other members who share (or previously shared) that display name.

## Solution

1. Track which display names are affected by a change (both old name and new name)
2. After updating the cache, call new method `recalculateDisambiguationForDisplayNames()`
3. This recalculates disambiguation for all members with those affected names
4. Emit `RoomState.Members` event so UI updates immediately

## Testing

Added 3 unit tests covering the fix:
- ✅ should disambiguate existing member when another member changes to the same name
- ✅ should un-disambiguate member when conflicting member changes to different name
- ✅ should emit RoomState.members for affected members when disambiguation changes

Run with: `yarn test -- --testPathPattern=room-state.spec.ts -t "reactive"`

## What this fixes

| Scenario | Before | After |
|----------|--------|-------|
| Bob changes name to "Alice" | Disambiguation not updated reactively ❌ | Updates immediately ✅ |
| Bob changes name from "Alice" to "Bob" | Original Alice stays disambiguated ❌ | Gets un-disambiguated ✅ |

## Remaining work (separate PR)

This PR addresses reactive updates for name changes. A separate issue exists for parted users (element-hq/element-meta#2850):
- When a user with a clashing name leaves the room, the remaining user should be un-disambiguated
- Will be addressed in a follow-up PR

---

Related to element-hq/element-web#31551

Related issues:
- Disambiguated names only dis-disambiguate after a refresh element-hq/element-web#468
- MXID disambiguation stays visible indefinitely long element-hq/element-web#4795

## Checklist
- [x] Tests written for new code
- [x] Linter and other CI checks pass

Signed-off-by: @aditya-cherukuru cherukuru.aditya01@gmail.com


dbkr:
If you believe this is a security fix, this should have been submitted privately as per https://github.com/matrix-org/matrix-js-sdk?tab=security-ov-file - had this been a real security issue, you would have just disclosed it publicly, potentially causing a major security incident (even though the original bug was public, you gave specific steps on how to exploit the bug).

However, your repro steps describe Alice (the real Alice) sending a message and her display name not showing with disambiguation. This might be a little confusing but Charlie (the imposter Alice) would still be shown with the disambiguation showing he wasn't really Alice. So, if I understand correctly I think your AI might have got a little confused.

You've also marked this as fixing 31551 but then said that it's only a partial fix and there's follow-up work coming. This special syntax will cause the bug to be closed when the PR is merged so if it doesn't completely fix it, please just say "For" or something instead of "Fixes", or create a sub-issue and mark it as fixing that.

changelog.d/5135.bugfix
```diff
@@ -0,0 +1 @@
+Fix reactive display name disambiguation to update immediately when room members change their display names.
```
dbkr: We don't use towncrier so you don't need this file.

src/models/room-state.ts
```diff
@@ -1110,6 +1110,17 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
 
     private updateDisplayNameCache(userId: string, displayName: string): void {
         const oldName = this.userIdsToDisplayNames[userId];
+
+        // Track which display names are affected by this change so we can
```
dbkr: I think doing this in a function called, `updateDisplayNameCache` is confusing as this function is now doing more than just what its name implies. It might make more sense to do it a level above, perhaps? This might also allow it be be done differently as I'm finding the `recalculateDisambiguationForDisplayNames` to be quite specific and more like just a chunk of one large function moved out to a separate function rather than a reusable unit of functionality.

src/models/room-state.ts
```diff
@@ -1131,10 +1142,47 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
...
+            const userIds = this.displayNameToUserIds.get(displayName) ?? [];
+            for (const otherUserId of userIds) {
+                if (otherUserId === excludeUserId) continue;
+
+                const member = this.members[otherUserId];
+                if (member?.events.member) {
+                    // Recalculate disambiguation by re-setting the membership event.
+                    // This will call shouldDisambiguate() with the updated room state.
+                    const oldName = member.name;
+                    member.setMembershipEvent(member.events.member, this);
```
dbkr: I don't love doing this by setting the same membership event again. Feels like if we want to update the disambiguation at other times, we should restructure to call it directly rather than faking things changing on that member.

aditya-cherukuru:
Thanks for the guidance @dbkr. Your comments really made me see the code structure in a different way, I learnt a lot from this review. I've addressed the feedback and cleaned up the architecture as suggested. Will continue to get better!

src/models/room-state.ts
```diff
@@ -514,6 +531,29 @@ export class RoomState extends TypedEventEmitter<EmittedEvents, EventHandlerMap>
...
+            for (const userId of affectedUserIds) {
+                const member = this.members[userId];
+                if (member?.events.member) {
+                    const nameChanged = member.recalculateDisambiguatedName(this);
+                    if (nameChanged) {
+                        this.emit(RoomStateEvent.Members, member.events.member, this, member);
+                    }
+                }
+            }
+        }
```
dbkr, line 562: This is looking much nicer, thanks. One thing though: your previous logic excluded the user whose name was being changed. Do we need to do that here too to avoid duplicate events?

aditya-cherukuru:
you're right! since the main loop already processes the user changing their name, my disambiguation loop would be doing redundant work for them.

I'll add a check to track which userIds were processed in the main batch and exclude them from the disambiguation pass. This ensures we only emit extra events for the *other* affected members, not duplicates.



Pushing the fix now.

PR #5137, opened 2026-01-12
Prepare for jest->vitest

t3chguy:
Split out from https://github.com/matrix-org/matrix-js-sdk/pull/5136

I suggest reviewing commits to get some context

### DIFF
--- a/spec/unit/room.spec.ts
+++ b/spec/unit/room.spec.ts
@@ -781,7 +781,9 @@ describe("Room", function () {
         );
     });
 
-    const resetTimelineTests = function (timelineSupport: boolean) {
+    describe.each(["enabled", "disabled"])("resetLiveTimeline with timeline support %s", (enabled) => {
+        const timelineSupport = enabled === "enabled";
+
         let events: MatrixEvent[];
 
         beforeEach(function () {
@@ -812,7 +814,8 @@ describe("Room", function () {
             ];
         });
 
-        it("should copy state from previous timeline", async function () {
+        // XXX: This test was previously not running and it is broken
+        it.skip("should copy state from previous timeline", async function () {
             await room.addLiveEvents([events[0], events[1]], { addToState: false });
             expect(room.getLiveTimeline().getEvents().length).toEqual(2);
             room.resetLiveTimeline("sometoken", "someothertoken");
@@ -882,13 +885,6 @@ describe("Room", function () {
             const tl = room.getTimelineForEvent(events[0].getId()!);
             expect(tl).toBe(timelineSupport ? firstLiveTimeline : null);
         });
-    };
-
-    describe("resetLiveTimeline with timeline support enabled", () => {
-        resetTimelineTests.bind(null, true);
-    });
-    describe("resetLiveTimeline with timeline support disabled", () => {
-        resetTimelineTests.bind(null, false);
     });
 
     describe("compareEventOrdering", function () {
@@ -1836,10 +1832,6 @@ describe("Room", function () {
             describe("threads enabled", () => {
                 beforeEach(() => {
                     jest.spyOn(room.client, "supportsThreads").mockReturnValue(true);
-                });
-
-                afterEach(() => {
-                    jest.restoreAllMocks();
                 });
 
                 it("returns true if there is an unthreaded receipt for a later event in a thread", async () => {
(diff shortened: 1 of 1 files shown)


spec/unit/rust-crypto/rust-crypto.spec.ts
```diff
@@ -1755,7 +1755,7 @@ describe("RustCrypto", () => {
                 fetchMock.get("path:/_matrix/client/v3/room_keys/version", testData.SIGNED_BACKUP_DATA);
 
                 const rustCrypto = await makeTestRustCrypto(makeMatrixHttpApi());
-                await expect(rustCrypto.getKeyBackupInfo()).resolves.toStrictEqual(testData.SIGNED_BACKUP_DATA);
```
BillCarsonFr, line 1758: Looks like `toStrictEqual` is asserting more tham `toMatchObject`? any reason to relax the requirement?

spec/unit/webrtc/groupCall.spec.ts
```diff
@@ -496,7 +497,6 @@ describe("Group Call", function () {
 
                     expect(groupCall.screenshareFeeds).toStrictEqual(newFeeds);
                     expect(currentFeed.dispose).toHaveBeenCalled();
-                    expect(newFeed.measureVolumeActivity).toHaveBeenCalledWith(true);
```
BillCarsonFr, line 499: Good catch. It is legacy calling? wonder if we shouldn't create an issue to track that this test was only passing due to side effect of other test

spec/integ/crypto/crypto.spec.ts
```diff
@@ -1718,7 +1718,6 @@ describe("crypto", () => {
          * @param backupVersion - The version of the created backup
          */
         async function bootstrapSecurity(backupVersion: string): Promise<void> {
-            mockSetupCrossSigningRequests();
```
BillCarsonFr, line 1721: right, looks like the `awaitCrossSigningKeyUpload` below are automatically adding the mocks (on more specific matcher, hence why no overwrite routes error)

spec/unit/models/event.spec.ts
```diff
@@ -159,7 +159,7 @@ describe("MatrixEvent", () => {
 
             // Then it disappears from the thread and appears in the main timeline
             expect(ev.threadRootId).toBeUndefined();
-            expect(mainTimelineLiveEventIds(room)).toEqual([threadRoot.getId(), ev.getId()]);
+            expect(mainTimelineLiveEventIds(room)).toEqual([threadRoot.getId(), ev.getId(), redaction.getId()]);
```
BillCarsonFr, line 162: That is strange.. Isn't it more that `redaction.getId()` is undefined?
like `expect([1,2]).toEqual([1,2, undefined]);`

Maybe these test could use things like `expect.arrayContaining([threadRoot.getId(), ev.getId()])` ?

spec/unit/room.spec.ts
```diff
@@ -812,7 +814,8 @@ describe("Room", function () {
             ];
         });
 
-        it("should copy state from previous timeline", async function () {
+        // XXX: This test was previously not running and it is broken
+        it.skip("should copy state from previous timeline", async function () {
```
BillCarsonFr, line 818: Out of curiosity, is there a task to investigate if this is important or not?

spec/unit/rust-crypto/rust-crypto.spec.ts
t3chguy: With vitest it was failing with `Serializes to same value`

spec/unit/webrtc/groupCall.spec.ts
t3chguy: Looking through the code it looks like `measureVolumeActivity` doesn't apply to screenshare streams

spec/unit/models/event.spec.ts
t3chguy: Yes, redaction before this PR had an `undefined` ID which jest allows you to skip in `toEqual` - vitest was unhappy and I thought that given real redactions have event IDs it'd make sense to make them have IDs in test too and assert them correctly

`arrayContaining` wouldn't assert order

spec/unit/room.spec.ts
t3chguy: Not from me, all the tests in this weird suite weren't running at all so this is preferable than removing them all to match the prior reality

PR #5142, opened 2026-01-16
Give RoomWidgetClient the ability to send and receive sticky events

robintown:
For https://github.com/element-hq/element-call/issues/3677

To-do:

- [x] Tests
- [x] Ensure that redactions of sticky events remove them from the sticky map

### DIFF
--- a/src/embedded.ts
+++ b/src/embedded.ts
@@ -396,6 +396,11 @@ export class RoomWidgetClient extends MatrixClient {
         if (stickyDurationMs !== undefined && typeof stickyDurationMs !== "number") {
             throw new Error("Sticky duration must be a number when defined");
         }
+        // This is save since we just checked that above
+        // We need the additional as assertion for the EW linter to be happy.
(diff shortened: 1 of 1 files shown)


src/@types/requests.ts
```diff
@@ -97,6 +97,7 @@ export interface ISendEventResponse {
     event_id: string;
 }
 
+// TODO the concept of `parent_delay_id` is not deprecated
```
robintown: @toger5 With this comment you added, do you mean that `parent_delay_id` should be marked deprecated?

toger5: It can be fully removed. But not in this PR.

t3chguy: This comment is confusing, nothing in the line below says that it is deprecated? It seems unrelated to this PR altogether

toger5: It also should say "now".
Removing it entirely also makes sense. We can have an issue tracking that.

toger5: I did remove the commented line

t3chguy (approved):

PR #5146, opened 2026-01-20
Avoid rapidly retrying failed requests

andybalaam:
After https://github.com/matrix-org/matrix-js-sdk/pull/5109 we retry failed requests in a tight loop, instead of once every sync. When requests are consistently failing, e.g. when `/keys/uploads` is failing because of a duplicate OTK, this causes us to make many requests, causing load on the server.

The fix is to reprocess the outgoing requests loop only if at least one request succeeded in the last batch.

Fixes https://github.com/element-hq/element-web/issues/31790


andybalaam:
Undrafting again so we can merge this for the RC. I am working on unit tests and will add them as soon as I can, but they probably won't make the RC.

PR #5155, opened 2026-01-20
[WIP] Temp with queryDict refactor

toger5:
## Checklist

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -103,7 +103,7 @@ import {
 import { RoomMemberEvent, type RoomMemberEventHandlerMap } from "./models/room-member.ts";
 import { type IPowerLevelsContent, type RoomStateEvent, type RoomStateEventHandlerMap } from "./models/room-state.ts";
 import {
-    isSendDelayedEventRequestOpts,
+    type SendStickyEventRequestOpts,
     UpdateDelayedEventAction,
     type DelayedEventInfo,
     type IAddThreePidOnlyBody,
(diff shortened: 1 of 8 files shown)


src/client.ts
```diff
@@ -2758,54 +2758,51 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
         }
     }
 
-    /**
-     * @param eventObject - An object with the partial structure of an event, to which event_id, user_id, room_id and origin_server_ts will be added.
-     * @param txnId - Optional.
-     * @returns Promise which resolves: to an empty object `{}`
-     * @returns Rejects: with an error response.
-     */
     private sendCompleteEvent(params: {
```
t3chguy, line 2767: Losing docs seems like a bad thing

src/client.ts
```diff
@@ -2758,54 +2758,51 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+        queryDict: QueryDict & SendDelayedEventRequestOpts;
+        txnId?: string;
+    }): Promise<SendDelayedEventResponse>;
+    private sendCompleteEvent(params: {
+        roomId: string;
+        threadId: string | null;
+        eventObject: Partial<IEvent>;
+        queryDict?: QueryDict | SendStickyEventRequestOpts;
         txnId?: string;
     }): Promise<ISendEventResponse>;
```
t3chguy, line 2783: This overload doesn't seem to add anything over the implementation signature?

toger5: This overload is only there to type the response.
I am not sure if there is a better way to do this in ts.

Based on a key existing in queryDict we either return
`SendDelayedEventResponse` or `ISendEventResponse` (`{delayId: string}` and `{eventId: string}` respectivly)

Any good idea on this would be gold!

PR #5158, opened 2026-01-22
Recalculate room name on loading members

dbkr:
Because if it's a DM room, loading members might change the room name

Also change some usernames to reuse the ones defined at the top of the file to be less confusing

## Checklist

### DIFF
--- a/spec/unit/room.spec.ts
+++ b/spec/unit/room.spec.ts
@@ -2056,7 +2056,7 @@ describe("Room", function () {
         });
     });
 
-    describe("loadMemfbersIfNeeded", function () {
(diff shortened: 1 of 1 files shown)


spec/unit/room.spec.ts
```diff
@@ -2056,7 +2056,7 @@ describe("Room", function () {
         });
     });
 
-    describe("loadMembersIfNeeded", function () {
+    describe("loadMemfbersIfNeeded", function () {
```
t3chguy:
```suggestion
    describe("loadMembersIfNeeded", function () {
```

PR #5159, opened 2026-01-23
Add support for stable OAuth2.0 aware feature from MSC3824

hughns:
## Checklist

### DIFF
--- a/tsdoc.json
+++ b/tsdoc.json
@@ -1,0 +1,11 @@
+{
+    "$schema": "https://developer.microsoft.com/en-us/json-schemas/tsdoc/v0/tsdoc.schema.json",
+    "extends": ["typedoc/tsdoc.json"],
+    "noStandardTags": false,
+    "tagDefinitions": [
+        {
+            "tagName": "@alias",
+            "syntaxKind": "modifier"
+        }
+    ]
+}


src/@types/auth.ts
```diff
@@ -48,11 +48,16 @@ export interface IPasswordFlow extends ILoginFlow {
...
 
-export const DELEGATED_OIDC_COMPATIBILITY = new UnstableValue(
-    "delegated_oidc_compatibility",
+export const OAUTH_AWARE_PREFERRED_FLOW_FIELD = new UnstableValue(
+    "oauth_aware_preferred",
     "org.matrix.msc3824.delegated_oidc_compatibility",
 );
 
+/**
+ * @deprecated use `OAUTH_AWARE_PREFERRED_FLOW_FIELD` instead.
```
hughns, line 58: The old constant name no longer makes sense so I defined a new one that matches the stable feature.

I am correct to deprecate the old name or should I have just removed it (and updated EW at the same time)?

Knip is complaining:

```
$ knip
Duplicate exports (1)
OAUTH_AWARE_PREFERRED_FLOW_FIELD|DELEGATED_OIDC_COMPATIBILITY  src/@types/auth.ts
```

t3chguy: Given it isn't marked as `@experimental` or `@internal` it should maintain semver and be deprecated as you have

PR #5163, opened 2026-01-26
widget: msc4407 push initial sticky map to widgets

BillCarsonFr:
Depends on https://github.com/matrix-org/matrix-widget-api/pull/161

Widgets have now the possibility to read sticky events. 
When a widget is loaded it only gets access to the live sticky event, so it doesn't know about pre-existing sticky events.

This a bit similar than with state, when a widget is loaded it will receive an update of the current state.
This PR does the same for sticky events, the hosting client will push the current sticky states to the widget, so we need to add support for receiving such udpates.

When `org.matrix.msc4407.initial_sticky_state` is received the sticky events are parsed and injected in the room, to that the room view from the widget is up-to-date with the room from the host application.

## Checklist

### DIFF
--- a/src/embedded.ts
+++ b/src/embedded.ts
@@ -30,6 +30,8 @@ import {
     type IWidgetApiResponseData,
     type IUpdateStateToWidgetActionRequest,
     UnstableApiVersion,
+    type IRoomEvent,
+    type IPushInitialStickyStateToWidgetActionRequest,
 } from "matrix-widget-api";
 
 import { MatrixEvent, type IEvent, type IContent, EventStatus } from "./models/event.ts";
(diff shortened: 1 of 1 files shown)


src/embedded.ts
```diff
@@ -807,6 +814,31 @@ export class RoomWidgetClient extends MatrixClient {
         await this.ack(ev);
     };
 
+    private onPushInitialStickyState = async (
+        ev: CustomEvent<IPushInitialStickyStateToWidgetActionRequest>,
+    ): Promise<void> => {
+        ev.preventDefault();
+
+        const { roomId, stickyEvents } = ev.detail.data;
```
toger5, line 822: Can this whole function be part of onEvent?
We branch out based on if there is a sticky key in the event.
We would need that anyhow for updating on new incoming sticky events. They also need to be added with `_unstable_addStickyEvents`

PR #5172, opened 2026-01-31
Fix empty string to room compatibility trick to only apply to m.call

toger5:
## Checklist

### DIFF
--- a/src/matrixrtc/CallMembership.ts
+++ b/src/matrixrtc/CallMembership.ts
@@ -373,13 +373,13 @@ export class CallMembership {
     public get slotId(): string {
         const { kind, data } = this.membershipData;
         if (data.application === "m.call") {
-            this.logger?.info("use slotId compat hack emptyString -> ROOM");
             switch (kind) {
                 case "rtc":
                     return data.slot_id;
                 case "session":
                 default: {
                     const [application, id] = [this.application, data.call_id];
+
                     // INFO_SLOT_ID_LEGACY_CASE  (search for all occurances of this INFO to get the full picture)
                     // The spec got changed to use `"ROOM"` instead of `""` empyt string for the implicit default call.
                     // State events still are sent with `""` however. To find other events that should end up in the same call,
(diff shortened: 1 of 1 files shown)


src/matrixrtc/CallMembership.ts
```diff
@@ -372,28 +372,46 @@ export class CallMembership {
      */
     public get slotId(): string {
         const { kind, data } = this.membershipData;
+        if (data.application === "m.call") {
+            this.logger?.info("use slotId compat hack emptyString -> ROOM");
```
robintown: I'd expect to see this log only in case `id === ""` actually matches

spec/unit/matrixrtc/MembershipManager.spec.ts
```diff
@@ -160,6 +162,45 @@ describe("MembershipManager", () => {
...
+                        application: "m.call",
+                        call_id: "custom",
+                        device_id: "AAAAAAA",
+                        expires: 14400000,
+                        foci_preferred: [focus],
+                        membershipID: "@alice:example.org:AAAAAAA",
+                        focus_active: focusActive,
+                        scope: "m.room",
+                    },
+                    "_@alice:example.org_AAAAAAA_m.callcustom",
```
robintown, line 191: I was wondering at FOSDEM, do we really expect this state key to read `m.callcustom` instead of `m.call#custom` or something? (Not at all important since this is a legacy case, but if it's wrong maybe it's a quick fix.)

toger5: I think we cannot fix it without a breaking change. We would need to make a custom rule that if the call_id is "" we do NOT add the `#`

src/matrixrtc/CallMembership.ts
toger5: Agreed.
c626b0f5b59b3f5b9a12ac9c23d785c1b50a3cfb

PR #5188, opened 2026-02-12
Disable complement crypto tests temporarily

dbkr:
As per comment

## Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -96,8 +96,8 @@ jobs:
         name: Downstream Complement Crypto tests
         runs-on: ubuntu-24.04
         if: always()
-    #    needs:
-    #        - complement-crypto
+        #    needs:
+        #        - complement-crypto
         steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -91,15 +91,15 @@ jobs:
...
-    downstream-complement-crypto:
-        name: Downstream Complement Crypto tests
-        runs-on: ubuntu-24.04
-        if: always()
-        needs:
-            - complement-crypto
-        steps:
-            - if: needs.complement-crypto.result != 'skipped' && needs.complement-crypto.result != 'success'
-              run: exit 1
+    #downstream-complement-crypto:
```
t3chguy: I think this task wants to stay otherwise you also need to update the branch protections, cleaner to just stub the task out

dbkr: oh, I already remove the branch protection but I can do that instead

PR #5191, opened 2026-02-16
Update getUrlPreview to use /_matrix/client/v1/media/preview_url

Half-Shot:
This has been the case since v1.11 https://spec.matrix.org/v1.12/changelog/v1.11/

If we care about backwards compat further than v1.11 (so June 2024) then we can also do a switch for the older endpoint.

## Checklist

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -52,35 +52,35 @@ export interface BaseLogger {
      *
      * @param msg - Data to log.
      */
-    trace(...msg: any[]): void;
+    trace(this: void, ...msg: any[]): void;
 
     /**
      * Output debug message to the logger.
      *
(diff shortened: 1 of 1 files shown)


t3chguy:
> If we care about backwards compat further than v1.11 (so June 2024) then we can also do a switch for the older endpoint.

We do, unless we want to remove <v1.11 support from https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/version-support.ts#L27-L36 AIUI

Half-Shot:
oh well, it's like two lines.

Half-Shot:
(I entirely forgot T-Tasks don't appear in the changelog. Going with Enhancement)

PR #5214, opened 2026-03-09
History sharing: resume key-bundle import on restart

richvdh:
Part of https://github.com/matrix-org/matrix-rust-sdk/issues/5112

(Review commit-by-commit)

### DIFF
--- a/src/rust-crypto/rust-crypto.ts
+++ b/src/rust-crypto/rust-crypto.ts
@@ -395,15 +395,14 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
         } catch (err) {
             logger.warn(`Error receiving encrypted bundle:`, err);
 
-            // Even though we were unable to import the bundle, we still clear the flag that indicates that we
+            throw err;
+        } finally {
+            // Even if we were unable to import the bundle, we still clear the flag that indicates that we
             // are waiting for the bundle to be received. The only reason this can happen is that the bundle was
             // malformed somehow, so we don't want to keep retrying it.
             await this.olmMachine.clearRoomPendingKeyBundle(new RustSdkCryptoJs.RoomId(roomId));
-
-            throw err;
         }
 
-        await this.olmMachine.clearRoomPendingKeyBundle(new RustSdkCryptoJs.RoomId(roomId));
         return true;
     }
 


src/rust-crypto/rust-crypto.ts
```diff
@@ -380,20 +380,26 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+            // Even though we were unable to import the bundle, we still clear the flag that indicates that we
+            // are waiting for the bundle to be received. The only reason this can happen is that the bundle was
+            // malformed somehow, so we don't want to keep retrying it.
+            await this.olmMachine.clearRoomPendingKeyBundle(new RustSdkCryptoJs.RoomId(roomId));
+
             throw err;
         }
 
-        // TODO: also clear this if the bundle was malformed.
         await this.olmMachine.clearRoomPendingKeyBundle(new RustSdkCryptoJs.RoomId(roomId));
```
kaylendog: This could perhaps be simplified using the elusive `finally` block?

src/rust-crypto/rust-crypto.ts
```diff
@@ -1740,7 +1742,20 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+                            this.logger.debug(
+                                `Not yet accepting key bundle for room where we are not awaiting a bundle: ${roomId}`,
+                            );
+                        } else if (
+                            Date.now() - pendingDetails.inviteAcceptedAtMillis >
+                            MAX_INVITE_ACCEPTANCE_MS_FOR_KEY_BUNDLE
+                        ) {
+                            this.logger.info(
+                                `Ignoring key bundle for room we joined too long ago: ${roomId}, joining time: ${new Date(pendingDetails.inviteAcceptedAtMillis).toISOString()}`,
+                            );
```
kaylendog, line 1755: I think we should discard details that are deemed too old so we avoid accumulating data forever.

src/rust-crypto/index.ts
```diff
@@ -247,5 +247,16 @@ async function initOlmMachine(
...
 
+    // If we have any recently-joined rooms, see if we have a pending key bundle for them.
+    for (const pendingDetails of await olmMachine.getAllRoomsPendingKeyBundles()) {
+        if (Date.now() - pendingDetails.inviteAcceptedAtMillis <= MAX_INVITE_ACCEPTANCE_MS_FOR_KEY_BUNDLE) {
+            const roomId = pendingDetails.roomId.toString();
+            logger.info(
+                `Checking for pending key bundle for recently-joined room ${roomId} (joined ${new Date(pendingDetails.inviteAcceptedAtMillis).toISOString()})`,
+            );
+            await rustCrypto.maybeAcceptKeyBundle(roomId, pendingDetails.inviterId.toString());
+        }
```
kaylendog, line 263: Ah, per my earlier comment, perhaps we can discard all bundles that don't meet this criteria here?

src/rust-crypto/rust-crypto.ts
richvdh: good spot, thank you.

src/rust-crypto/index.ts
richvdh: yup, done.

src/rust-crypto/rust-crypto.ts
richvdh: I don't think we need to do it here as well as on restart.

PR #5231, opened 2026-03-18
Rotate the current room key when we see a member leave

kaylendog:
This PR adds logic to the `RustCrypto.onRoomMembership` event handler that discards (and hence rotates) the current session whenever we see an event `leave` membership.

Part of https://github.com/element-hq/element-meta/issues/3078

## Checklist

### DIFF
--- a/spec/integ/crypto/history-sharing.spec.ts
+++ b/spec/integ/crypto/history-sharing.spec.ts
@@ -68,6 +68,7 @@ afterEach(() => {
 const ROOM_ID = "!room:example.com";
 const ALICE_HOMESERVER_URL = "https://alice-server.com";
 const BOB_HOMESERVER_URL = "https://bob-server.com";
+const CHARLIE_HOMESERVER_URL = "https://charlie-server.com";
 
 async function createAndInitClient(homeserverUrl: string, userId: string, setupNewCrossSigning = true) {
     mockInitialApiRequests(homeserverUrl, userId);
@@ -91,6 +92,8 @@ describe("History Sharing", () => {
     let aliceSyncResponder: SyncResponder;
     let bobClient: MatrixClient;
     let bobSyncResponder: SyncResponder;
+    let charlieClient: MatrixClient;
+    let charlieSyncResponder: SyncResponder;
 
     beforeEach(async () => {
         // anything that we don't have a specific matcher for silently returns a 404
@@ -100,6 +103,7 @@ describe("History Sharing", () => {
 
         const aliceId = TEST_USER_ID;
         const bobId = "@bob:xyz";
+        const charlieId = "@charlie:zyx";
 
         const aliceKeyReceiver = new E2EKeyReceiver(ALICE_HOMESERVER_URL, "alice-");
         const aliceKeyResponder = new E2EKeyResponder(ALICE_HOMESERVER_URL);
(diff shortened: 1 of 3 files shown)


src/rust-crypto/rust-crypto.ts
```diff
@@ -1947,6 +1947,12 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
             });
         }
 
+        // If it's not our membership, and it was a leave event, rotate the room key.
+        if (member.userId !== this.olmMachine.userId.toString() && member.membership === KnownMembership.Leave) {
+            this.logger.info(`Rotating session for room ${roomId} due to member leaving the room`);
+            this.forceDiscardSession(member.roomId);
+        }
+
         const enc = this.roomEncryptors[roomId];
```
kaylendog, line 1951: I think it's sufficient to check this here, rather than a dedicated `RoomStateEvent.Events` handler - I followed the code up from this handler and I believe the membership event is fired as a child of the method firing the top-level `Events` event.

src/rust-crypto/rust-crypto.ts
```diff
@@ -1947,6 +1947,28 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+         * leaves, resulting in M2 being encrypted with the same session as M1.
+         * This would allow Charlie to decrypt M2 if he ever gains access to
+         * the event.
+         *
+         * This conditional discards the current room key if the membership event
+         * is a `leave` event.
+         */
+        if (member.userId !== this.olmMachine.userId.toString() && member.membership === KnownMembership.Leave) {
+            this.logger.info(`Rotating session for room ${roomId} due to member leaving the room`);
+            this.forceDiscardSession(member.roomId);
```
andybalaam: In the normal case, will we discard the session again when we send a message?

andybalaam (commented):
Looks good to me - one question, and I would definitely like Rich to have a look too.

src/rust-crypto/rust-crypto.ts
kaylendog: I believe so? I don't think this should be much of an issue since session rotation is fairly cheap and we only send keys out when we actually send a message, but would mean we have empty sessions in the store.

andybalaam: Ah, I guess if this key is not actually sent out this might be fine. Would be interested in @richvdh 's input.

kaylendog: Now that I think about it, I'm not sure we will discard, since we'll need to create a new session when we next send a message, at which point we should compute who has the current session from scratch?

andybalaam: Rich says this is fine because we create sessions lazily when we need them, so discarding twice has no effect.

andybalaam (approved):
Yep, looks great. I feel like this would be easier to test with a real homeserver. Maybe this is one for complement-crypto, or maybe we can do that somewhere within rust-sdk?

src/rust-crypto/rust-crypto.ts
richvdh: I think a membership event is only fired if the client sees a *change* in membership. In the case of a gappy sync, we might see a user going from `leave` to `leave`, but we still need to rotate the session in that case.

spec/integ/crypto/history-sharing.spec.ts
```diff
@@ -108,23 +112,39 @@ describe("History Sharing", () => {
...
+        bobKeyClaimResponder.addKeyReceiver(aliceId, aliceClient.deviceId!, aliceKeyReceiver);
 
         aliceSyncResponder.sendOrQueueSyncResponse({});
         await syncPromise(aliceClient);
 
         bobSyncResponder.sendOrQueueSyncResponse({});
         await syncPromise(bobClient);
+
+        charlieSyncResponder.sendOrQueueSyncResponse({});
+        await syncPromise(charlieClient);
```
richvdh, line 147: not thrilled about adding a "charlie" for all the tests: it adds to log spam and slows tests down.

Somewhat wondering if we should move the client intialisation out to a separate function which takes an argument for the number of clients to create.

richvdh (changes requested):

spec/integ/crypto/history-sharing.spec.ts
kaylendog: Sounds reasonable, I'll have a go at that.

src/rust-crypto/rust-crypto.ts
kaylendog: Addressed in 943c0dbda3558b56e3d7f3640ab37f92a0d5d175

src/rust-crypto/rust-crypto.ts
```diff
@@ -1955,6 +1956,38 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+         * 1. Alice sends a message M1 in room X;
+         * 2. Bob invites Charlie, who joins and immediately leaves the room;
+         * 3. Alice sends another message M2 in room X.
+         *
+         * Under the old logic, Alice would not rotate her key after Charlie
+         * leaves, resulting in M2 being encrypted with the same session as M1.
+         * This would allow Charlie to decrypt M2 if he ever gains access to
+         * the event.
+         *
+         * This conditional discards the current room key if the membership event
```
t3chguy:
```suggestion
         * This conditionally discards the current room key if the membership event
```

src/rust-crypto/rust-crypto.ts
```diff
@@ -1955,6 +1956,38 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+        /**
+         * Previously, it was sufficient to check if we need to rotate the room key
+         * prior to sending a message. However, the history sharing feature
+         * (MSC4268) breaks this logic:
+         *
+         * 1. Alice sends a message M1 in room X;
+         * 2. Bob invites Charlie, who joins and immediately leaves the room;
+         * 3. Alice sends another message M2 in room X.
+         *
+         * Under the old logic, Alice would not rotate her key after Charlie
```
t3chguy: Could we consolidate the old logic with this? Is there any benefit to only rotating prior to sending a message in some cases and on membership change on others? The complexity feels risky

t3chguy (approved):

src/rust-crypto/rust-crypto.ts
richvdh: I think we need both: they guard against different cases. We still need the old logic to handle users removing *devices*.

spec/integ/crypto/history-sharing.spec.ts
kaylendog: I've pulled out these changes to a separate branch that I think is worth reviewing independently of this one. I'll open a PR once this is merged.

src/rust-crypto/rust-crypto.ts
```diff
@@ -1961,20 +1976,16 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
          * This would allow Charlie to decrypt M2 if he ever gains access to
          * the event.
          *
-         * This conditional discards the current room key if the membership event
+         * This conditionally discards the current room key if the membership event
          * is a `leave` event.
```
richvdh: we should probably try to explain why we need to do this in `onRoomStateEvent` rather than `onRoomMembership`.

```suggestion
         * To counter this, we proactively discard any active outgoing Megolm session when we see a `leave`
         * event. Note that we have to do this in `onRoomStateEvent` rather than `onRoomMembership`,
         * because `onRoomMembership` is only called when we see a *change* in membership. In the case of
         * a gappy sync, we might miss Charlie's invite and join, and only see the final `leave` event (so
         * his membership goes from `leave` to `leave`).
```

richvdh (approved):
lgtm otherwise

spec/integ/crypto/history-sharing.spec.ts
```diff
@@ -696,257 +696,284 @@ describe("History Sharing", () => {
...
+                );
+                syncResponse.rooms.join[ROOM_ID].timeline.events.push(
+                    mkEventCustom({
+                        content: { membership: KnownMembership.Leave },
+                        type: EventType.RoomMember,
+                        sender: charlieClient.getSafeUserId(),
+                        state_key: charlieClient.getSafeUserId(),
+                    }),
+                );
+            }
```
richvdh: I don't think we really want to be calling `getSyncResponse` in either case here. It doesn't make sense for Bob to receive more copies of all the events he already received back at line 707.

Just build the incremental sync from scratch: for the non-gappy case, add a join and a leave to the timeline for charlie; for the gappy case, just add the leave and set limited:true.

kaylendog: Addressed in https://github.com/matrix-org/matrix-js-sdk/pull/5231/commits/905a2379a137bbd45a3270be5585148f530e851c

PR #5242, opened 2026-03-26
Disable Complement Crypto tests

t3chguy:
Disabled Complement Crypto tests due to flakiness.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -88,7 +88,7 @@ jobs:
 
     complement-crypto:
         name: "Run Complement Crypto tests"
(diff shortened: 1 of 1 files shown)


richvdh:
Per https://github.com/matrix-org/complement-crypto/issues/158, I'll try and look at the failures when I get a chance

MidhunSureshR (approved):

richvdh:
Turns out that most of the failures currently are due to https://github.com/matrix-org/complement-crypto/issues/234. We'll update chromium.

richvdh:
This got reverted by #5257.

PR #5246, opened 2026-03-31
Implement MSC4429: Profile Updates for Legacy Sync

Half-Shot:
https://github.com/matrix-org/matrix-spec-proposals/pull/4429

This implements half functional support for profile updates over legacy sync. <del>It does not yet fully support storing them over sessions, but does handle updates.</del>

Now stores updates in indexeddb with write-through caching on the get API so they'll eventually all be stored in the database even if you turn the feature flag on after your initial sync.

Closes https://github.com/element-hq/wat-internal/issues/464

## Checklist

### DIFF
--- a/spec/unit/matrix-client.spec.ts
+++ b/spec/unit/matrix-client.spec.ts
@@ -375,6 +375,7 @@ describe("MatrixClient", function () {
                 "startup",
                 "deleteAllData",
                 "setUserCreator",
+                "storeUserProfiles",
             ] as const
         ).reduce((r, k) => {
             r[k] = vi.fn();
--- a/spec/unit/matrix-client.spec.ts
+++ b/spec/unit/matrix-client.spec.ts
@@ -19,7 +19,7 @@ limitations under the License.
  */
 
 import fetchMock from "@fetch-mock/vitest";
-import { type Mocked } from "vitest";
+import { type MockedObject, type Mocked } from "vitest";
 
 import { logger } from "../../src/logger";
 import {
@@ -89,6 +89,7 @@ import { type CryptoBackend } from "../../src/common-crypto/CryptoBackend";
 import { SyncResponder } from "../test-utils/SyncResponder.ts";
 import { mockInitialApiRequests } from "../test-utils/mockEndpoints.ts";
 import { type Transport } from "../../src/matrixrtc/index.ts";
+import { type IStore } from "../../src/store/index.ts";
 
 vi.useFakeTimers();
 
@@ -385,6 +386,7 @@ describe("MatrixClient", function () {
         store.getClientOptions = vi.fn().mockReturnValue(Promise.resolve(null));
         store.storeClientOptions = vi.fn().mockReturnValue(Promise.resolve(null));
         store.isNewlyCreated = vi.fn().mockReturnValue(Promise.resolve(true));
+        store.getUserProfile = vi.fn().mockReturnValue(undefined);
 
         // set unstableFeatures to a defined state before each test
         unstableFeatures = {
@@ -1323,6 +1325,18 @@ describe("MatrixClient", function () {
             ];
             await expect(client.getExtendedProfileProperty(userId, "test_key")).resolves.toEqual("foo");
             expect(httpLookups).toHaveLength(0);
+        });
+
+        it("can fetch a property from a extended user profile with a cached profile", async () => {
+            const testProfile = {
+                test_key: "foo",
+            };
+            (client.store as MockedObject<IStore>).getUserProfile.mockImplementation((requestedUserId) => {
+                expect(requestedUserId).toEqual(userId);
+                expect("test_key").toEqual("test_key");
+                return testProfile;
+            });
+            await expect(client.getExtendedProfileProperty(userId, "test_key")).resolves.toEqual("foo");
         });
 
         it("can set a property in our extended profile", async () => {
--- a/spec/unit/matrix-client.spec.ts
+++ b/spec/unit/matrix-client.spec.ts
@@ -1331,7 +1331,7 @@ describe("MatrixClient", function () {
             const testProfile = {
                 test_key: "foo",
             };
-            (client.store as MockedObject<IStore>).getUserProfile.mockImplementation((requestedUserId) => {
+            (client.store as MockedObject<IStore>).getUserProfile.mockImplementation(async (requestedUserId) => {
                 expect(requestedUserId).toEqual(userId);
                 expect("test_key").toEqual("test_key");
                 return testProfile;
--- a/spec/unit/matrix-client.spec.ts
+++ b/spec/unit/matrix-client.spec.ts
@@ -1327,13 +1327,12 @@ describe("MatrixClient", function () {
             expect(httpLookups).toHaveLength(0);
         });
 
-        it("can fetch a property from a extended user profile with a cached profile", async () => {
+        it("can fetch a property from an extended user profile with a cached profile", async () => {
             const testProfile = {
                 test_key: "foo",
             };
             (client.store as MockedObject<IStore>).getUserProfile.mockImplementation(async (requestedUserId) => {
                 expect(requestedUserId).toEqual(userId);
-                expect("test_key").toEqual("test_key");
                 return testProfile;
             });
             await expect(client.getExtendedProfileProperty(userId, "test_key")).resolves.toEqual("foo");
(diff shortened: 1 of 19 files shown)


spec/unit/matrix-client.spec.ts
```diff
@@ -1325,6 +1327,18 @@ describe("MatrixClient", function () {
             expect(httpLookups).toHaveLength(0);
         });
 
+        it("can fetch a property from a extended user profile with a cached profile", async () => {
+            const testProfile = {
+                test_key: "foo",
+            };
+            (client.store as MockedObject<IStore>).getUserProfile.mockImplementation(async (requestedUserId) => {
+                expect(requestedUserId).toEqual(userId);
+                expect("test_key").toEqual("test_key");
```
dbkr: eh?

src/sync.ts
```diff
@@ -634,6 +634,17 @@ export class SyncApi {
             }
             this.opts.filter.setLazyLoadMembers(true);
         }
+        // I feel like this is the wrong home for this.
```
dbkr: yeah, not exactly sure what this function claims to be doing but it's not lazy-loading related

src/sync.ts
```diff
@@ -1138,6 +1149,28 @@ export class SyncApi {
...
+                }
+            }
+            if (usersToRemove.length) {
+                await client.store.removeUserProfiles(usersToRemove);
+            }
+            if (profilesToAmend) {
+                await client.store.storeUserProfiles(profilesToAmend);
+            }
+        }
+
```
dbkr, line 1184: this could use a comment or maybe even being moved out to its own function

spec/unit/matrix-client.spec.ts
Half-Shot: :laughing: ah, yes that's clearly not it.

src/models/user.ts
```diff
@@ -26,6 +26,8 @@ export enum UserEvent {
     LastPresenceTs = "User.lastPresenceTs",
 }
 
+export type SyncUserProfile = Record<string, unknown>;
```
t3chguy: Comment pls

src/store/index.ts
```diff
@@ -254,6 +254,25 @@ export interface IStore {
      */
     removeToDeviceBatch(id: number): Promise<void>;
 
+    /**
+     * Store user profile details from a sync. Existing profiles will be overwritten.
+     * @param userProfileTuples - A set of userIds to profiles.
+     */
+    storeUserProfiles(userProfileTuples: Array<[string, SyncUserProfile]>): Promise<void>;
```
t3chguy: ooi why is it an array of tuples vs a record/similar?

src/sync.ts
```diff
@@ -1138,6 +1159,29 @@ export class SyncApi {
             });
         }
 
+        // handle user profile updates (MSC4429)
+        const userUpdate = data["users"] ?? data["org.matrix.msc4429.users"];
+        if (typeof userUpdate === "object" && userUpdate !== null) {
+            const usersToRemove = [];
```
t3chguy: type pls

src/sync-accumulator.ts
```diff
@@ -249,6 +260,7 @@ function isTaggedEvent(event: IRoomEvent): event is TaggedEvent {
  */
 export class SyncAccumulator {
     private accountData: Record<string, IMinimalEvent> = {}; // $event_type: Object
+
```
t3chguy: seems unrelated

src/sync-accumulator.ts
```diff
@@ -183,6 +187,12 @@ export interface IDeviceLists {
     left?: string[];
 }
 
+export interface IUserUpdate {
```
t3chguy: Comment pls

t3chguy: Also no I prefix anymore as per code style

src/filter.ts
```diff
@@ -67,6 +72,11 @@ interface IRoomFilter {
     timeline?: IRoomEventFilter;
     account_data?: IRoomEventFilter;
 }
+
+interface IProfileFieldsFilter {
```
t3chguy: No I pls

src/filter.ts
```diff
@@ -242,4 +252,16 @@ export class Filter {
     public setIncludeLeaveRooms(includeLeave: boolean): void {
         setProp(this.definition, "room.include_leave", includeLeave);
     }
+
+    /**
+     * @param ids The field IDs to sync.
```
t3chguy, line 261: the first line of a jsdoc should describe what the method does

src/filter.ts
```diff
@@ -35,11 +38,13 @@ function setProp(obj: Record<string, any>, keyNesting: string, val: any): void {
...
-    presence?: IFilterComponent;
-    account_data?: IFilterComponent;
-    room?: IRoomFilter;
+    "event_fields"?: string[];
+    "event_format"?: "client" | "federation";
+    "presence"?: IFilterComponent;
+    "account_data"?: IFilterComponent;
+    "room"?: IRoomFilter;
+    "profile_fields"?: IProfileFieldsFilter;
+    "org.matrix.msc4429.profile_fields"?: IProfileFieldsFilter;
```
t3chguy: Shouldn't these refer to `profileFieldsFilterName`

dbkr: Ideally but I can't get typescript to be happy about it:

 > A computed property name must be of type 'string', 'number', 'symbol', or 'any'.

src/store/index.ts
dbkr: seems like there was no particular reason - I've gone with a map because non-static keys, although it has to be said that when called with a single update this is now the same with the addition of a Map constructor.

src/store/index.ts
```diff
@@ -254,6 +254,25 @@ export interface IStore {
      */
     removeToDeviceBatch(id: number): Promise<void>;
 
+    /**
+     * Store user profile details from a sync. Existing profiles will be overwritten.
+     * @param userProfiles - A map of userIds to profiles.
+     */
+    storeUserProfiles(userProfiles: Map<string, SyncUserProfile>): Promise<void>;
```
t3chguy, line 261: This type referencing `unknown` makes it very weak. Why are we using unknown in storage? Using `unknown` for "down sync" as the jsdoc says is fine, if it is then validated and upgraded to a stronger type

src/models/user.ts
```diff
@@ -26,6 +26,11 @@ export enum UserEvent {
     LastPresenceTs = "User.lastPresenceTs",
 }
 
+/**
+ * An object of extended profile attributes for a user as it arrives down the sync stream.
+ */
+export type SyncUserProfile = Record<string, unknown>;
```
t3chguy: Presumably this isn't truly `unknown`, given its JSON it can be a much more specific type

dbkr: We could move (or copy) https://github.com/element-hq/element-web/blob/develop/apps/web/src/%40types/json.ts to js-sdk which would make it a bit more specific but I think that's about it. For storage, we don't really know much more other than that it's not null or undefined I think.

t3chguy: Well presumably we do know a lot more, for it to even be storable, given we presumably can't store complex cyclic objects

t3chguy: The spec is broad https://spec.matrix.org/v1.17/client-server-api/#profiles allowing any JSON type - so yeah possibly we should copy the types from EW

dbkr: Done. should be a little less vague now.

src/@types/json.ts
```diff
@@ -0,0 +1,16 @@
+/*
+Copyright 2024 New Vector Ltd.
```
t3chguy, line 2: Inheriting the copyright might be fine but I think we need to give up the license, least because it refers to LICENSE files which do not exist in this repo

src/@types/json.ts
```diff
@@ -0,0 +1,16 @@
...
+
+// Types for JSON and JSON objects, copied from element-web (left in both places as I don't think we
+// want this as a part of js-sdk's interface and I don't think it's worth it being its own package)
+
+export type JsonValue = null | string | number | boolean;
+export type JsonArray = Array<JsonValue | JsonObject | JsonArray>;
+export interface JsonObject {
+    [key: string]: JsonObject | JsonArray | JsonValue;
+}
+export type Json = JsonArray | JsonObject;
```
t3chguy, line 16: this should probably be `JsonDocument`

t3chguy (approved):

PR #5256, opened 2026-04-01
Expand the comment on CryptoApi.getUserDeviceInfo saying we request info from the server

andybalaam:
## Checklist

### DIFF
--- a/src/crypto-api/index.ts
+++ b/src/crypto-api/index.ts
@@ -207,8 +207,8 @@ export interface CryptoApi {
      * For any users whose device lists are cached (due to sharing an encrypted room with the user), the
      * cached device data is returned, unless it is stale.
      *
-     * If there are users with stale cached entries, a `/keys/query` request is made to the server to
-     * refresh their device info.
+     * If there are users with stale cached entries, wait (with some timeout) for any in-progress
+     * `/keys/query` request to complete.
      *
      * If there are uncached users, and the `downloadUncached` parameter is set to `true`,
      * a `/keys/query` request is made to the server to retrieve these devices.
--- a/src/crypto-api/index.ts
+++ b/src/crypto-api/index.ts
@@ -205,7 +205,10 @@ export interface CryptoApi {
      * Get the device information for the given list of users.
      *
      * For any users whose device lists are cached (due to sharing an encrypted room with the user), the
-     * cached device data is returned.
+     * cached device data is returned, unless it is stale.
+     *
+     * If there are users with stale cached entries, a `/keys/query` request is made to the server to
+     * refresh their device info.
      *
      * If there are uncached users, and the `downloadUncached` parameter is set to `true`,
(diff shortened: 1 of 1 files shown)


src/crypto-api/index.ts
```diff
@@ -205,7 +205,10 @@ export interface CryptoApi {
      * Get the device information for the given list of users.
      *
      * For any users whose device lists are cached (due to sharing an encrypted room with the user), the
-     * cached device data is returned.
+     * cached device data is returned, unless it is stale.
+     *
+     * If there are users with stale cached entries, a `/keys/query` request is made to the server to
+     * refresh their device info.
```
richvdh: Is that correct? Or is it rather that we wait for any pending `/keys/query` requests to happen (via the background sync loop)?

I guess there's not much difference from the caller's PoV, but I think being more precise might be less confusing.

richvdh: Also, we probably ought to update `userHasCrossSigningKeys` too, which I think works the same way.

andybalaam: We do appear to actually make the request: https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/rust-crypto/rust-crypto.ts#L579

andybalaam: `userHasCrossSigningKeys` already has:

>      * - If the user is not tracked locally and downloadUncached is set to true,
>     *   a `/keys/query` request is made to the server to retrieve the cross signing keys.
>     * - Otherwise, return false

So I think it's OK as-is?

richvdh: I think that's untracked users (which we mention in the next paragraph), rather than stale users (who must be tracked, otherwise we wouldn't know they were stale)

richvdh: it doesn't say anything about tracked users with stale cached entries though?

andybalaam: I don't think [userHasCrossSigningKeys](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/rust-crypto/rust-crypto.ts#L503) does the same thing. It appears to block on querying if we are asking about ourself, but otherwise not. Did I miss something?

andybalaam: Ah right, yes. Done in 2a7feb0ab

richvdh: Well, it calls [`olmMachine.getIdentity(userId)`](https://github.com/matrix-org/matrix-sdk-crypto-wasm/blob/37ad1c02a68768e8d559b1d7f21f9f52b4486ba2/src/machine.rs#L847-L868), which has the same one second timeout as the call to `olmMachine.getUserDevices` in `RustCrypto.getUserDevices`.

andybalaam: Right, addressed in 22e063a5c

richvdh (approved):

PR #5258, opened 2026-04-02
Support the stable prefix for MSC4287 (key backup preference)

andybalaam:
## Checklist

### DIFF
--- a/src/crypto-api/index.ts
+++ b/src/crypto-api/index.ts
@@ -626,7 +626,7 @@ export interface CryptoApi {
      *  * Disables 4S, deleting the info for the default key, the default key pointer itself and any
      *    known 4S data (cross-signing keys and the megolm key backup key).
      *  * Deletes any dehydrated devices.
-     *  * Sets the "m.org.matrix.custom.backup_disabled" account data flag to indicate that the user has disabled backups.
+     *  * Sets the "m.key_backup" account data flag to indicate that the user has disabled backups.
      */
     disableKeyStorage(): Promise<void>;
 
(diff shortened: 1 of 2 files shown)


t3chguy (approved):

src/crypto-api/index.ts
```diff
@@ -626,7 +626,7 @@ export interface CryptoApi {
      *  * Disables 4S, deleting the info for the default key, the default key pointer itself and any
      *    known 4S data (cross-signing keys and the megolm key backup key).
      *  * Deletes any dehydrated devices.
-     *  * Sets the "m.org.matrix.custom.backup_disabled" account data flag to indicate that the user has disabled backups.
+     *  * Sets the "m.key_backup" account data flag to indicate that the user has disabled backups.
```
uhoreg: I'm not seeing any evidence that the implementation of this function actually does this.  And it looks like it's being done in `element-web` (at least for the old name).  Is this comment a lie, or am I missing something?

andybalaam: This is a comment on the interface, which is implemented in element-web. Or did I get confused?

uhoreg: Isn't this the `CryptoApi` interface, so implemented by js-sdk's `RustCrypto` class?

andybalaam: Sorry, yes I was totally confused! Fixed in 85da90b15

uhoreg (approved):

andybalaam:
Spec PR is https://github.com/matrix-org/matrix-spec/pull/2354

PR #5268, opened 2026-04-14
Adapt LiveKit Identity hash calculation to latest MSC4195 update

fkwp:
This PR is doing two things:
- Changing the serialization / construction of the LiveKit participant id hash
- Fixing a regression regarding base64 URL vs Normal encoding

For reference the MatrixRTC Authorization Service matching change is here https://github.com/element-hq/lk-jwt-service/pull/172/changes

Note: No need for synchronized release because this is under a dev flag.
But worth saying that Element Call "compatibility" and "Matrix 2.0" mode won't work until the JWT and JS-SDK are both released.

## Checklist


spec/unit/matrixrtc/CallMembership.spec.ts
```diff
@@ -427,7 +427,7 @@ describe("CallMembership", () => {
         });
         it("uses unpadded base64 for RTC backend identities", async () => {
             const membership = await CallMembership.parseFromEvent(makeMockEvent(0, { ...membershipTemplate }));
-            expect(membership.rtcBackendIdentity).toBe("j9N1u04ZbvI9qKf3cxrf2NauD-fIGJ4uAcYkfI9V7SY");
```
toger5, line 430: I am surprised this did not catch the change from `encodeUnpaddedBase64` to `encodeUnpaddedBase64Url`

spec/unit/matrixrtc/MembershipData.spec.ts
t3chguy, line 1: copyright

t3chguy (changes requested):
Something odd with this PR

Image

spec/unit/matrixrtc/CallMembership.spec.ts
BillCarsonFr: I think the test did catch the problem ! but the test was updated :/ https://github.com/matrix-org/matrix-js-sdk/pull/5134/changes

Probably we should have give the test a more explicit name than `uses unpadded base64 for RTC backend identities`?

This title makes you think the focus is on `unpadded` but it was more on the none URL encoded thing..
Maybe this:
`.toBe("2+h2ELE1XY/NsuveToZOekORCoyQMO6V0W7XZUWk5Q4")`
and
`not.toBe("2-h2ELE1XY_NsuveToZOekORCoyQMO6V0W7XZUWk5Q4")`

spec/unit/matrixrtc/MembershipData.spec.ts
fkwp: https://github.com/matrix-org/matrix-js-sdk/pull/5268/commits/ab15718b577a299732a1fef8ae10dfa07df9436e

spec/unit/matrixrtc/CallMembership.spec.ts
fkwp: we now have the test vector as part of the MSC -- that should also do the trick?

fkwp:
> Something odd with this PR
> 
> Image

Indeed very odd (messed up rebase?) ; now fixed

BillCarsonFr (approved):
Approved. This has regressed here https://github.com/matrix-org/matrix-js-sdk/pull/5134 , there used to be a test but it was updated, it was not explicit enough of what it was testing. The test vector should correct that.

Apparently this was not breaking EC integration tests because there is some randomness in the tests (userIds) and the encoded values are not always differents. This was fixed by test retries?

BillCarsonFr:
@fkwp Could you update the description of the PR to explain what are the **updates**?
For our future self.
This PR is doing to things:
- Fixing a regression regarding base64 URL vs Normal encoding
- Changing the serialization / construction of the id to hash

For reference the auth service matching change is here https://github.com/element-hq/lk-jwt-service/pull/172/changes

No need for synchronised release because this is under a dev flag.
But worth saying that compatibility and matrix 2.0 won't work until the JWT and JS-SDK are both released

t3chguy (approved):

PR #5281, opened 2026-04-16
Fix key backup cached event ordering

gumadeiras:
## Summary

Fixes a Rust crypto key-backup event ordering bug where creating a new backup could emit `CryptoEvent.KeyBackupDecryptionKeyCached` before local backup state was coherent for listeners.

Before this change, `setupKeyBackup()` stored the generated backup key through `saveBackupDecryptionKey()`, which immediately emitted the cached-key event. `PerSessionKeyBackupDownloader` listens to that event and may immediately query backup configuration, seeing stale or undefined backup info and logging `Unsupported algorithm undefined`.

This PR keeps generated-backup setup quiet, then has `RustCrypto.resetKeyBackup()` store the recovery key and use the successful backup creation response to cache server backup info, enable the matching local backup, and only then emit the cached-key event. Received backup secrets now use the same coherent activation path after validating that the secret matches the current server-side backup.

## Testing

```sh
pnpm test spec/unit/rust-crypto/backup.spec.ts spec/unit/rust-crypto/PerSessionKeyBackupDownloader.spec.ts -- --runInBand
pnpm test spec/unit/rust-crypto/rust-crypto.spec.ts -- --runInBand
pnpm test spec/integ/crypto/verification.spec.ts -t "Should accept the backup decryption key gossip if valid" -- --runInBand
pnpm test spec/integ/crypto/megolm-backup.spec.ts -t "storeSessionBackupPrivateKey|loadSessionBackupPrivateKeyFromSecretStorage|checkKeyBackupAndEnable" -- --runInBand
pnpm lint:types
pnpm exec eslint src/rust-crypto/backup.ts src/rust-crypto/rust-crypto.ts spec/unit/rust-crypto/backup.spec.ts spec/unit/rust-crypto/rust-crypto.spec.ts
pnpm exec prettier --check src/rust-crypto/backup.ts src/rust-crypto/rust-crypto.ts spec/unit/rust-crypto/backup.spec.ts spec/unit/rust-crypto/rust-crypto.spec.ts
```

## Notes

Notes: Fix key-backup cached event timing so listeners do not observe stale or undefined backup info immediately after creating a new backup.

Maintainer label suggestion: `T-Defect`. The Preview Changelog workflow currently requires one release-type label, and this PR fixes an observable key-backup event ordering defect.

Signed-off-by: Gustavo Madeira Santana <fromweb1@gumadeiras.com>

## Checklist

- [x] Sign-off given on the changes.


andybalaam:
Hi @gumadeiras thank you for the PR! I'm having a little trouble understanding the description - did you write it yourself or use AI? If AI, please could you explain in your own words? (I'm struggling to understand e.g. "Received backup secrets now use the same coherent activation path".)

These questions might help clarify:

* What problem is this fixing? How would a user experience this problem?
* What code paths have been changed? What user actions would trigger those code paths?

(Please don't copy and paste answers from an AI btw.)

gumadeiras:
> Hi @gumadeiras thank you for the PR! I'm having a little trouble understanding the description - did you write it yourself or use AI? If AI, please could you explain in your own words? (I'm struggling to understand e.g. "Received backup secrets now use the same coherent activation path".)
> 
> These questions might help clarify:
> 
> * What problem is this fixing? How would a user experience this problem?
> * What code paths have been changed? What user actions would trigger those code paths?
> 
> (Please don't copy and paste answers from an AI btw.)

hi Andy, sorry for the lack of clarity!

the problem is a race condition when creating a new encrypted-message backup
the SDK currently does two things: (1) saves a new backup key and (2) updates its local state so it knows which backup version on the server is active

when (1) finishes it immeadiately triggers "backup key is cached", before (2) finishes
so when we are setting up or doing a backup reset we were getting warnings from the SDK (from `PerSessionKeyBackupDownloader`)
logs were showing:
"Unsupported algorithm undefined"
"Got current backup version from server: undefined"

the fix i'm proposing is just delaying the "backup key cached" message to the end of the cycle to avoid the race condition

and this is really just fixing a warning that happens even if everything is successful

just adding code refs for what i said (how it works today):
(1) creating a new backup goes through saveBackupDecryptionKey() from setupKeyBackup() and triggers CryptoEvent.KeyBackupDecryptionKeyCached (this is all in backup.ts)
then PerSessionKeyBackupDownloader gets triggered because it listens for KeyBackupDecryptionKeyCached (these are in PerSessionKeyBackupDownloader.ts)


the code that changes here:

resetKeyBackup(): user sets up or resets backup. code path: CryptoApi.resetKeyBackup → RustCrypto.resetKeyBackup() → enableKeyBackupFromCreation()

setupKeyBackup(): internal part of creating the backup version. It now only stores the generated key at creation time (this is in backup.ts)

handleBackupSecretReceived(): user verifies/signs in another device and receives the backup key via secret sharing. API docs says this is one way to do backup restore (crypto-api/index.ts), and the changed code path for this part is handleBackupSecretReceived(); this path is also now fixed for that same race condition

let me know if i can provide any other details!

andybalaam:
@gumadeiras thank you for taking the time to help me understand. I will take another look and also hand over to @richvdh who said he had some thoughts about the organisation of this code.

src/rust-crypto/rust-crypto.ts
```diff
@@ -1382,8 +1382,7 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
             await this.secretStorage.store("m.megolm_backup.v1", backupInfo.decryptionKey.toBase64());
         }
 
-        // we can check and start async
-        this.checkKeyBackupAndEnable();
```
andybalaam: I'm not sure why it's OK to remove this call to `checkKeyBackupAndEnable`. I don't think I see equivalent code in `enableKeyBackupFromCreation`. The old code ends up in `doCheckKeyBackup` which does a bunch of stuff that presumably is useful?

andybalaam (commented):
I'm going to ask @richvdh to take over this review since he is more familiar with this code, but I'm still struggling to understand. Perhaps this change does multiple things at once, or maybe I'm just having a bad day...?

I understand that there's a race where we are emitting an event that results in other places checking our internal state before we've updated it, but I'm not understanding how that description fits the code yet.

src/rust-crypto/rust-crypto.ts
gumadeiras: this removes it from the just created backup path in `resetKeyBackup()` only; it is still called through other paths

i dropped it because `resetKeyBackup()` has just created the backup itself so my assumption is that it doesn't need to be rechecked since it was just created? if the server accepted the create request and returned the new version, we can use that creation result as the current backup state instead of immediately doing a second check

the new helper still does the important local enable work: cache `serverBackupInfo`, mark `checkedForBackup`, call the existing `enableKeyBackup(...)`, set `activeBackupVersion`, emit `KeyBackupStatus`, and start `backupKeysLoop()`

so the intent is not to remove backup enabling; only to avoid the extra check before triggering `KeyBackupDecryptionKeyCached`

if `checkKeyBackupAndEnable()` has another required function which i may have totally missed, then yeah i should rework this back in; let me know!

paths that still call RustCrypto.checkKeyBackupAndEnable:
startup calls RustCrypto.checkKeyBackupAndEnable()
public API calls RustCrypto.checkKeyBackupAndEnable()
own-user trust changes call RustCrypto.checkKeyBackupAndEnable()
getServerBackupInfo() calls RustBackupManager.checkKeyBackupAndEnable(false)
upload recovery calls RustBackupManager.checkKeyBackupAndEnable(true)

gumadeiras:
> I'm going to ask @richvdh to take over this review since he is more familiar with this code, but I'm still struggling to understand. Perhaps this change does multiple things at once, or maybe I'm just having a bad day...?
> 
> I understand that there's a race where we are emitting an event that results in other places checking our internal state before we've updated it, but I'm not understanding how that description fits the code yet.

hopefully this helps a bit but let me know if not

just to recap; the race i am trying to fix is:

1. `setupKeyBackup()` creates a new backup
2. it saves the new backup key
3. saving the key immediately emits `KeyBackupDecryptionKeyCached`
4. `PerSessionKeyBackupDownloader` hears that event and checks backup state
5. but `resetKeyBackup()` has not yet finished enabling/caching the newly-created backup locally

so the listener can see `serverBackupInfo` / `activeBackupVersion` before they match the backup that was just created. that is where the `Unsupported algorithm undefined` log came from

the fix i was trying to make is: for a backup we just created, first update local backup state from the creation result, then emit `KeyBackupDecryptionKeyCached`

file changes

`src/rust-crypto/rust-crypto.ts`
changes `resetKeyBackup()` to call `enableKeyBackupFromCreation(backupInfo)` instead of kicking off `checkKeyBackupAndEnable()` in the background
why: the backup was just created, so this uses the create response to enable/cache it immediately instead of doing a second async discovery check

`src/rust-crypto/backup.ts`: separates “store backup key” from “emit cached-key event”; makes new-backup creation store the key without emitting; adds the helper that caches/enables a known usable backup before emitting; applies the same ordering to received backup secrets. reason: both new backup creation and received backup secrets can emit `KeyBackupDecryptionKeyCached`, so both need backup state first

adds `enableKeyBackupFromCreation()` and `enableKeyBackupFromInfo()`.
why: this is the replacement for the just-created-backup path. it sets `serverBackupInfo`, marks `checkedForBackup`, calls existing `enableKeyBackup()`, then emits `KeyBackupDecryptionKeyCached`

changes `handleBackupSecretReceived()` to use the same ordered path
why: received backup secrets emit the same cached-key event, so they should also only emit it after matching backup state is coherent (`serverBackupInfo`, `activeBackupVersion`, and the emitted event version all refer to the same backup)

`spec/unit/rust-crypto/rust-crypto.spec.ts`
adds a regression test for `resetKeyBackup()` after stale “no backup” state
why: proves the cached-key event sees active version, event version, and server version all matching

`spec/unit/rust-crypto/backup.spec.ts`: proves `setupKeyBackup()` does not emit early; proves received backup secrets emit after backup state is cached/enabled; verifies local backup enabling happens. reason: covers the event-ordering behavior changed in `backup.ts`


`spec/unit/rust-crypto/backup.spec.ts`
adds tests that `setupKeyBackup()` does not emit the cached-key event early, and that received backup secrets emit only after backup state is cached/enabled
why: covers the two event-ordering paths changed in `backup.ts`

edit: also if you believe this is the wrong direction or want anything reworked; please let me know i'd be happy to iterate as needed. i can also test any other solution in our production code to test other approaches

src/rust-crypto/backup.ts
```diff
@@ -214,16 +214,88 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
         return false;
     }
 
-    public async saveBackupDecryptionKey(
+    private async storeBackupDecryptionKey(
         backupDecryptionKey: RustSdkCryptoJs.BackupDecryptionKey,
         version: string,
     ): Promise<void> {
         await this.olmMachine.saveBackupDecryptionKey(backupDecryptionKey, version);
+    }
```
richvdh: I don't think there's any point in creating a dedicated function for this single line of code.

src/rust-crypto/backup.ts
```diff
@@ -214,16 +214,88 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+    ): Promise<void> {
+        await this.storeBackupDecryptionKey(backupDecryptionKey, version);
         // Emit an event that we have a new backup decryption key, so that the sdk can start
         // importing keys from backup if needed.
+        this.emitBackupDecryptionKeyCached(version);
+    }
+
+    private emitBackupDecryptionKeyCached(version: string): void {
         this.emit(CryptoEvent.KeyBackupDecryptionKeyCached, version);
     }
```
richvdh: likewise: `this.emit(CryptoEvent.KeyBackupDecryptionKeyCached, version);` really isn't very much longer than `this.emitBackupDecryptionKeyCached(version);`

src/rust-crypto/backup.ts
```diff
@@ -568,7 +640,7 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
             },
         );
 
-        await this.saveBackupDecryptionKey(randomKey, res.version);
+        await this.storeBackupDecryptionKey(randomKey, res.version);
```
richvdh: Note to self: this change is the main point of the PR. It avoids emitting a `KeyBackupDecryptionKeyCached` event, meaning that `PerSessionKeyBackupDownloader` will not immediately query for the new backup state.

src/rust-crypto/backup.ts
```diff
@@ -205,7 +205,7 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
             this.logger.info(
                 `handleBackupSecretReceived: Valid decryption key for the current server-side backup version (${latestBackupInfo.version}) received`,
             );
-            await this.saveBackupDecryptionKey(backupDecryptionKey, latestBackupInfo.version);
+            await this.saveBackupDecryptionKeyForCurrentBackup(backupDecryptionKey, latestBackupInfo);
```
richvdh: I'm not really following why this needs to change. Your description only mentions a problem happening during `setupKeyBackup`, but this is unrelated to `setupKeyBackup`. Does a similar problem also affect receiving the backup decryption key from other devices? I'm not sure why it would, because we don't update `RustBackupManager.serverBackupInfo` here (as far as I can tell), so there is nothing to race against?

src/rust-crypto/backup.ts
```diff
@@ -214,16 +214,88 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+            this.serverBackupInfo = previousServerBackupInfo;
+            this.checkedForBackup = previousCheckedForBackup;
+            throw e;
+        }
+    }
+
+    private async saveBackupDecryptionKeyForCurrentBackup(
+        backupDecryptionKey: RustSdkCryptoJs.BackupDecryptionKey,
+        backupInfo: KeyBackupInfo,
+    ): Promise<void> {
```
richvdh: This could do with a doc comment. How does it differ from `saveBackupDecryptionKey`, and why would I want one rather than the other?

src/rust-crypto/backup.ts
```diff
@@ -214,16 +214,88 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+     * against eventual consistency.
+     *
+     * @param backupInfo - The newly-created backup details.
+     */
+    public async enableKeyBackupFromCreation(backupInfo: KeyBackupCreationInfo): Promise<void> {
+        await this.enableKeyBackupFromInfo({
+            algorithm: backupInfo.algorithm,
+            auth_data: backupInfo.authData,
+            version: backupInfo.version,
+        });
```
richvdh: why not just `this.enableKeyBackupFromInfo(backupInfo)` ?

src/rust-crypto/backup.ts
```diff
@@ -214,16 +214,88 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+
+        this.emitBackupDecryptionKeyCached(backupInfo.version);
+    }
+
+    private async enableKeyBackupFromInfo(backupInfo: KeyBackupInfo): Promise<void> {
```
richvdh: this could also do with a doc comment. How does it differ from `enableKeyBackup` (which also takes a `backupInfo`)?

src/rust-crypto/backup.ts
```diff
@@ -214,16 +214,88 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+        this.checkedForBackup = true;
+
+        try {
+            const activeVersion = await this.getActiveBackupVersion();
+            if (activeVersion === null) {
+                await this.enableKeyBackup(backupInfo);
+            } else if (activeVersion !== version) {
+                await this.disableKeyBackup();
+                await this.enableKeyBackup(backupInfo);
+            }
```
richvdh: This is (a subset of) the same logic as in `doCheckKeyBackup`, right? Can we factor it out and reuse it?

src/rust-crypto/backup.ts
richvdh: Since it seems to be a helper for `handleBackupSecretReceived`, I'd inline it, which means you can get rid of the check that `backupInfo.version` is defined.

src/rust-crypto/backup.ts
```diff
@@ -214,16 +214,88 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+        });
+
+        this.emitBackupDecryptionKeyCached(backupInfo.version);
+    }
+
+    private async enableKeyBackupFromInfo(backupInfo: KeyBackupInfo): Promise<void> {
+        const version = backupInfo.version;
+        if (!version) {
+            throw new Error("Cannot enable key backup without version");
+        }
```
richvdh: I think you can enforce this via the type system: declare the `backupInfo` parameter to have the type `KeyBackupInfo & { version: string }`

src/rust-crypto/backup.ts
```diff
@@ -214,16 +214,88 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+        try {
+            const activeVersion = await this.getActiveBackupVersion();
+            if (activeVersion === null) {
+                await this.enableKeyBackup(backupInfo);
+            } else if (activeVersion !== version) {
+                await this.disableKeyBackup();
+                await this.enableKeyBackup(backupInfo);
+            }
+        } catch (e) {
+            this.serverBackupInfo = previousServerBackupInfo;
```
richvdh: I somewhat wonder if this is necessary, given we know that these are in fact the details of the backup on the server.

src/rust-crypto/backup.ts
richvdh: What will happen if this races with an ongoing call to `doCheckKeyBackup`, or another call to `doCheckKeyBackup` happens while this is in progress? You might have to wait for `keyBackupCheckInProgress` to complete, or something.

richvdh (changes requested):
Hi @gumadeiras sorry for taking a while to get back to you on this. 

Firstly, thanks very much for the contribution. I agree that the current code looks wrong, though I'm surprised we haven't heard more reports about it. Did you see any symptoms of it other than the log lines? Which application are you using (Element-Web, or something else?)

Indeed, it might be helpful if you could open an issue (against matrix-js-sdk or the application) describing exactly what symptoms you are seeing, without going into details about the underlying cause or potential fixes, so that we can be clear what we're trying to fix here. Including logs might also be helpful.

Now, regarding your fix. I'm rather worried that it makes an area of the code that is already a bit complicated even more complicated. It leads to weird scenarios where we cache the backup decryption key without emitting the `KeyBackupDecryptionKeyCached` event. It also increases coupling between `RustCrypto` and `RustBackupManager`, in that we now rely on `RustBackupManager.enableKeyBackupFromCreation` being called after `RustBackupManager.setupKeyBackup` to ensure that the `KeyBackupDecryptionKeyCached` event is emitted. 

My main suggestion is, rather than having `resetKeyBackup` make separate calls to `setupKeyBackup` and `enableKeyBackupFromCreation`, we just combine the two -- in other words, move the call to`enableKeyBackupFromCreation` into `setupKeyBackup`. That would mean we could update `serverBackupInfo` *before* calling `saveBackupDecryptionKey`, and thus sidestep much of the issue.

Some other points:
 - One of the existing integration tests seems to be failing. Please take a look at it and see if you can figure out why.
 - It would be nice to see an integration test which demonstrates that the issue has been fixed.

I've also made a number of comments on the code, some of which might become somewhat irrelevant if you refactor it a bit.

richvdh:
@gumadeiras some other quick bits of feedback on your PR:

1. I found [this comment](https://github.com/matrix-org/matrix-js-sdk/pull/5281#issuecomment-4270430507) really hard to read. Some capital letters and full stops/periods would go a long way, as would using `backticks` for identifiers. The random line breaks don't help either.

2. As a general rule, it's not the case that more words == better. I appreciate your efforts in trying to help us understand the change, but I think you could do with being a bit more succinct in your answers: I found I spent quite a long time reading your descriptions!

3. On which note... I share @andybalaam's suspicion that a lot of this text, and code, has been generated by AI. I suspect that the reason you've given these long answers is because most of them have been generated by AI? Using AI isn't necessarily a problem, but you as the human need to take responsibility for your submission, and I'd much rather read a short explanation **in your own words** about the reason you feel a change is necessary, rather than the AI's explanation for its own change.

4. In [this comment](https://github.com/matrix-org/matrix-js-sdk/pull/5281#issuecomment-4280967160) you go into detail about different parts of the change. Unfortunately it's rather hard for me to correlate with the code. If you feel that sort of context is helpful, I'd recommend adding it as comments on the relevant parts of the diff (https://github.com/matrix-org/matrix-js-sdk/pull/5281/changes) rather than as a big block of text.

gumadeiras:
> Hi @gumadeiras sorry for taking a while to get back to you on this.
> 
> Firstly, thanks very much for the contribution. I agree that the current code looks wrong, though I'm surprised we haven't heard more reports about it. Did you see any symptoms of it other than the log lines? Which application are you using (Element-Web, or something else?)

no worries; thanks for the feedback! I'll go through it as soon as possible

I primarily use Element Desktop and ElementX on iOS, and the matrix channel for [openclaw](https://github.com/openclaw/openclaw) (I'm the maintainer for Matrix)

i haven't seen any symptoms other than the logs; what i originally thought was a problem turned out to just be the noisy logs

gumadeiras:
> @gumadeiras some other quick bits of feedback on your PR:
> 
> 1. I found [this comment](https://github.com/matrix-org/matrix-js-sdk/pull/5281#issuecomment-4270430507) really hard to read. Some capital letters and full stops/periods would go a long way, as would using `backticks` for identifiers. The random line breaks don't help either.
> 2. As a general rule, it's not the case that more words == better. I appreciate your efforts in trying to help us understand the change, but I think you could do with being a bit more succinct in your answers: I found I spent quite a long time reading your descriptions!
> 3. On which note... I share @andybalaam's suspicion that a lot of this text, and code, has been generated by AI. I suspect that the reason you've given these long answers is because most of them have been generated by AI? Using AI isn't necessarily a problem, but you as the human need to take responsibility for your submission, and I'd much rather read a short explanation **in your own words** about the reason you feel a change is necessary, rather than the AI's explanation for its own change.
> 4. In [this comment](https://github.com/matrix-org/matrix-js-sdk/pull/5281#issuecomment-4280967160) you go into detail about different parts of the change. Unfortunately it's rather hard for me to correlate with the code. If you feel that sort of context is helpful, I'd recommend adding it as comments on the relevant parts of the diff (https://github.com/matrix-org/matrix-js-sdk/pull/5281/changes) rather than as a big block of text.

The original PR body was generated with AI and I reviewed/edited. All follow up messages were fully written by me. They weren't long because of AI (I wrote them), I was just trying to be more thorough with code refs to help understand the changes I made, since @andybalaam wanted to better understand why I made the changes I made. Similarly for the code changes, some parts were AI assisted but I fully reviewed/edited everything. I'll leave the next round of comments directly in the code to help with clarity

src/rust-crypto/backup.ts
```diff
@@ -582,6 +601,18 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+        const backupInfo: KeyBackupInfoWithVersion = {
+            algorithm: pubKey.algorithm,
+            auth_data: authData,
+            version: res.version,
+        };
+        // saveBackupDecryptionKey emits KeyBackupDecryptionKeyCached. Cache and
+        // enable the created backup first so listeners observe the new version.
+        // This backup was just created and signed locally, so use the creation
+        // response instead of doing another discovery/trust check before the event.
+        this.serverBackupInfo = backupInfo;
```
gumadeiras, line 631: This is the main ordering change.

`saveBackupDecryptionKey()` emits `KeyBackupDecryptionKeyCached`, so `setupKeyBackup()` now updates `serverBackupInfo` and enables/switches to the created backup before saving the key.

That means listeners handling `KeyBackupDecryptionKeyCached` should see the new backup version immediately.

src/rust-crypto/rust-crypto.ts
gumadeiras: Just reiterating what I said before for this part:

`resetKeyBackup()` no longer calls `checkKeyBackupAndEnable()` after `setupKeyBackup()`.

The enable step moved into `setupKeyBackup()` itself, using the version returned by the successful backup creation request. That avoids doing a second discovery check before emitting `KeyBackupDecryptionKeyCached`.

spec/integ/crypto/crypto.spec.ts
```diff
@@ -2021,11 +2021,33 @@ describe("crypto", () => {
...
+                            serverVersion: serverInfo?.version,
+                        });
+                    });
+                });
 
                 await aliceClient.getCrypto()!.resetKeyBackup();
                 await awaitDeleteCalled;
                 await newBackupStatusUpdate;
                 await newBackupUploadPromise;
+                await expect(keyBackupCachedState).resolves.toEqual({
```
gumadeiras, line 2053: This is the regression check for the event ordering: when `KeyBackupDecryptionKeyCached` fires, the event version, active backup version, and server backup version should all match the newly-created backup.

gumadeiras:
Hi all, sorry this took a while; life got chaotic. I reworked this to make the change narrower following your feedback.

The new version keeps the ordering fix inside `setupKeyBackup()`:

1. store the new backup version in local state,
2. enable or switch to that backup locally,
3. then save the backup key.

Step 3 still emits `KeyBackupDecryptionKeyCached`, but listeners should now see the new backup version when they handle the event.

I also made `setupKeyBackup()` wait for any existing `keyBackupCheckInProgress` before deleting/creating/enabling the new backup. After the server returns the new version, the code uses that creation response as the local backup state before emitting `KeyBackupDecryptionKeyCached`. I think that is the right scope because this backup was just created and signed locally.

This change only fixes the create-backup path where the cached-key event could fire before the newly-created backup was active locally. I noticed `Unsupported algorithm undefined` happens in other pathways, outside of this fix. One place I saw it logged was when an old backup is being deleted/disabled during reset, when there really is no current backup info, the unsupported algorithm warning will be logged. 

Let me know if this is more aligned with how you were thinking!

src/rust-crypto/backup.ts
```diff
@@ -345,34 +353,41 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+            await this.enableOrSwitchKeyBackup(backupInfo, activeVersion);
         }
         return { backupInfo, trustInfo };
     }
 
-    private async enableKeyBackup(backupInfo: KeyBackupInfo): Promise<void> {
+    private async enableOrSwitchKeyBackup(
+        backupInfo: KeyBackupInfoWithVersion,
+        activeVersion?: string | null,
+    ): Promise<void> {
```
richvdh: Please could you give this a doc-comment, saying what it does and what the parameters do.

In particular, please note the different behaviours between `null`, `undefined`, and a stringy value for `activeVersion`.

src/rust-crypto/backup.ts
```diff
@@ -559,6 +574,10 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
      * @returns a KeyBackupCreationInfo - All information related to the backup.
      */
     public async setupKeyBackup(signObject: (authData: AuthData) => Promise<void>): Promise<KeyBackupCreationInfo> {
```
richvdh, line 588: It would be good to update the documentation on this method to record that it also enables key backup upload and download using the new backup version, and saves the decryption key in the rust-sdk's cryptostore.

src/rust-crypto/backup.ts
```diff
@@ -345,34 +353,41 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
...
+            // This will remove any pending backup request, remove the backup key and reset the backup state of each room key we have.
+            await this.disableKeyBackup();
+            // Enabling will now trigger re-upload of all the keys
+            await this.enableKeyBackup(backupInfo);
+        } else {
+            this.logger.debug(`Backup version ${backupInfo.version} still current`);
+        }
+    }
+
+    private async enableKeyBackup(backupInfo: KeyBackupInfoWithVersion): Promise<void> {
```
richvdh: Again, a docstring would help:

```suggestion
    /**
     * Helper for {@link enableOrSwitchKeyBackup}.
     * 
     * Enables key backup upload and download for the given backup version. Also emits
     * a {@link CryptoEvent.KeyBackupStatus} event.
     */
    private async enableKeyBackup(backupInfo: KeyBackupInfoWithVersion): Promise<void> {
```

spec/unit/rust-crypto/backup.spec.ts
```diff
@@ -140,4 +141,46 @@ describe("Upload keys to backup", () => {
         expect(outgoingRequestProcessor.makeOutgoingRequest).toHaveBeenCalledTimes(1);
         expect(mockOlmMachine.roomKeyCounts).toHaveBeenCalledTimes(0);
     });
+
+    it("Should emit key cached after enabling a newly-created backup", async () => {
+        fetchMock.hardReset();
+        fetchMock.mockGlobal();
```
richvdh: Are these necessary? I find them a bit scary, as this sort of thing often implies that there are interactions between the test cases that we are trying to work around.

Does it work if you instead add an `{ overwriteRoutes: true }` options object to the calls to `fetchMock.get` and `fetchMock.post` ?

src/rust-crypto/backup.ts
```diff
@@ -559,6 +574,10 @@ export class RustBackupManager extends TypedEventEmitter<RustBackupCryptoEvents,
      * @returns a KeyBackupCreationInfo - All information related to the backup.
      */
     public async setupKeyBackup(signObject: (authData: AuthData) => Promise<void>): Promise<KeyBackupCreationInfo> {
+        if (this.keyBackupCheckInProgress) {
```
richvdh, line 590: Let's add a comment:

```suggestion
        // Wait for any active call to `checkKeyBackupAndEnable` to complete, to avoid racing with it
        if (this.keyBackupCheckInProgress) {
```

richvdh, line 590: I think there is still a potential problem with a call to `checkKeyBackupAndEnable` starting *after* the call to `setupKeyBackup` begins (but before it finishes), but let's leave that problem for another time.

richvdh (commented):
> Hi all, sorry this took a while; life got chaotic

No problem at all. Likewise, sorry for taking a while to review it.

This looks much closer to what I was imagining, thank you!

I've made a few minor comments, if you wouldn't mind.

richvdh:
@gumadeiras are you still interested in working on this?

Half-Shot:
Thanks for the PR, please reopen if you plan to work on it in the future.

richvdh:
I'll take a look at this; it's a real bug that should be fixed.

spec/unit/rust-crypto/backup.spec.ts
richvdh: cf7b8e9891f9a9dc39ae5c8eafdf4f9f1199841

src/rust-crypto/backup.ts
richvdh: ec0c239feb43d21791ad8b033b57209af6a98a7b

src/rust-crypto/backup.ts
richvdh: 80f6de33a

richvdh (approved):

richvdh:
The CI is failing for somewhat nonsense reasons around type checks on KeyBackupInfo. I've created https://github.com/matrix-org/matrix-js-sdk/pull/5419 to clean up these types; this will then be easier to land.

PR #5297, opened 2026-04-21
Remove dependency on `uuid`

t3chguy:


florianduros (approved):

toger5 (approved):
This seems to only be used for the legacy calling.
So I am not deeply familiar with this code/spec.
It does look very sane however.

PR #5307, opened 2026-04-30
fix(MatrixRTCSession): replace no-arg .finally() with .then(fn, fn) to fix crash on React Native / Hermes

### LINKED ISSUE #5308: MatrixRTCSession: .finally() without argument crashes on React Native / Hermes — TypeError: undefined is not a function
### Description

Using matrix-js-sdk `v41.4.0` in a React Native application with the Hermes JS engine, **every call to `MatrixRTCSession.ensureRecalculateSessionMembers()` throws an unhandled `TypeError: undefined is not a function`** at startup.

The error is silent in normal Metro output (it appears only as a sequence of opaque unhandled promise rejection IDs) and leaves the entire `MatrixRTCSession` layer in a permanently broken state,every room's `initialMembershipCalculated` promise is rejected, so RTC session membership is never computed.

---

### Steps to reproduce

1. Create a React Native app (Expo SDK 55 / bare RN 0.83.6) that uses matrix-js-sdk `41.4.0`.
2. Authenticate and start a Matrix client with at least one joined room.
3. Call `client.matrixRTC.start()` (or rely on automatic start during sync).
4. Observe Metro / logcat output.

**Expected:** RTC sessions initialise silently.
**Actual:** One `TypeError: undefined is not a function` per room appears in the Metro log, all originating from `promise/setimmediate/finally.js` (React Native's internal Promise polyfill) via `InternalBytecode.js`.

---

### Minimal reproduction snippet

```ts
// Minimal probe,run this anywhere after React Native initialises:
Promise.resolve()
  .finally()         // ← no argument
  .then(() => console.log("OK"))
  .catch((e) => console.error("CRASH:", e.message)); // prints "CRASH: undefined is not a function"
```

---

### Root cause

`MatrixRTCSession.ensureRecalculateSessionMembers()` in
`src/matrixrtc/MatrixRTCSession.ts` contains:

```js
this.recalculateSessionMembersPromise =
  this.recalculateSessionMembersPromise
    .finally()   // ← called with no argument (undefined)
    .then(() => this.recalculateSessionMembers());
```

In browsers and Node.js, `Promise.prototype.finally(undefined)` is a spec-compliant no-op that passes the settled value through. However, the Promise polyfill shipped with React Native (`promise/setimmediate/finally.js`) performs a strict `typeof onFinally !== 'function'` check and **throws** rather than treating `undefined` as a pass-through.

Spec reference,[MDN: `Promise.prototype.finally`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally):

> If `onFinally` is not a function, it is internally replaced with a function that simply returns the received value. This means `.finally()` (no argument) is equivalent to `.then(x => x, x => { throw x; })` and must **not** throw.

React Native's polyfill diverges from this requirement.

---

### Affected versions

| Component | Version |
|---|---|
| **matrix-js-sdk** | **41.4.0** (regression introduced with MatrixRTC refactor) |
| React Native | 0.83.6 |
| Hermes | bundled with RN 0.83.6 |
| Expo SDK | 55 |
| `promise` npm package (RN built-in polyfill) | ships with RN, `promise/setimmediate/finally.js` |
| Node.js (dev host) | 24.15.0 LTS |
| iOS / Android | both affected (Hermes is default on both) |

---

### Impact

- **All React Native clients** using matrix-js-sdk ≥ v41 are affected whenever at least one joined room exists.
- `MatrixRTCSession.initialMembershipCalculated` is rejected for every room, silently disabling membership tracking for the entire RTC layer.
- No console error is produced by default,the failure is visible only via Metro's opaque rejection IDs.

---

### Suggested fix

Replace `.finally().then(fn)` with `.then(onFulfilled, onRejected)`,semantically identical, compatible with all runtimes:

```diff
- this.recalculateSessionMembersPromise = this.recalculateSessionMembersPromise
-   .finally()
-   .then(() => this.recalculateSessionMembers());
+ this.recalculateSessionMembersPromise = this.recalculateSessionMembersPromise.then(
+   () => this.recalculateSessionMembers(),
+   () => this.recalculateSessionMembers(),
+ );
```

A PR with this fix is available: https://github.com/matrix-org/matrix-js-sdk/pull/5307

---

### Workaround (for app developers)

Until an upstream release is published, patch `node_modules/matrix-js-sdk` with
[`patch-package`](https://github.com/ds300/patch-package):

```diff
# patches/matrix-js-sdk+41.4.0.patch
-    this.recalculateSessionMembersPromise = this.recalculateSessionMembersPromise.finally().then(() => this.recalculateSessionMembers());
+    this.recalculateSessionMembersPromise = this.recalculateSessionMembersPromise.then(
+      () => this.recalculateSessionMembers(),
+      () => this.recalculateSessionMembers(),
+    );
```

Then add to `package.json`:
```json
"scripts": {
  "postinstall": "patch-package"
}
```

JeanLuX:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/5308

### What

`ensureRecalculateSessionMembers()` called `.finally()` with no argument to swallow a previous rejection before chaining the next recalculation.

On React Native / Hermes, the RN Promise polyfill (`promise/setimmediate/finally.js`) requires `onFinally` to be a callable — passing `undefined` throws immediately:

```
TypeError: undefined is not a function
```

This rejects `initialMembershipCalculated` for every room at startup, silently breaking the entire RTC session membership layer on all React Native clients.

### Fix

Replace `.finally().then(fn)` with `.then(fn, fn)` — semantically equivalent on all runtimes:

```diff
-    this.recalculateSessionMembersPromise = this.recalculateSessionMembersPromise.finally().then(() => this.recalculateSessionMembers());
+    this.recalculateSessionMembersPromise = this.recalculateSessionMembersPromise.then(
+      () => this.recalculateSessionMembers(),
+      () => this.recalculateSessionMembers(),
+    );
```

**File:** `src/matrixrtc/MatrixRTCSession.ts` — `ensureRecalculateSessionMembers()`

### Verified on

| Component | Version |
|---|---|
| matrix-js-sdk | 41.4.0 |
| React Native | 0.83.6 |
| Hermes | bundled with RN 0.83.6 |
| Expo SDK | 55 |

### Checklist

- [x] Semantics preserved — recalculation runs on both resolve and reject
- [x] No behaviour change on browser / Node.js runtimes
- [x] Unit tests pass locally (`yarn test`)
- [x] Signed-off-by / CLA (if required)

### DIFF
--- a/spec/unit/matrixrtc/MatrixRTCSession.spec.ts
+++ b/spec/unit/matrixrtc/MatrixRTCSession.spec.ts
@@ -911,6 +911,31 @@ describe("MatrixRTCSession", () => {
             expect(sess!.membershipStatus).toBe(Status.Connecting);
         });
     });
+    it("ensureRecalculateSessionMembers still runs after a rejected promise (React Native / Hermes regression)", async () => {
+        // Regression test for: .finally() without an argument throws a TypeError on the
+        // React Native Promise polyfill used by Hermes, permanently breaking the
+        // recalculation chain. The fix uses .then(onFulfilled, onRejected) instead.
+        const mockRoom = makeMockRoom([sessionMembershipTemplate]);
+        sess = MatrixRTCSession.sessionForSlot(client, mockRoom, callSession);
+        await sess.initialMembershipCalculated;
+
+        const privateSession = sess as unknown as {
+            recalculateSessionMembersPromise: Promise<void>;
+            membershipNeedsRecalculation: boolean;
+            ensureRecalculateSessionMembers: () => Promise<void>;
+        };
+
+        // Inject a rejected promise to simulate what the Hermes TypeError would produce.
+        privateSession.recalculateSessionMembersPromise = Promise.reject(new Error("simulated rejection"));
+        // Reset the deduplication flag so a new recalculation is scheduled.
+        privateSession.membershipNeedsRecalculation = false;
+
+        // Must resolve (not reject) — recalculation runs regardless of prior rejection.
+        await expect(privateSession.ensureRecalculateSessionMembers()).resolves.toBeUndefined();
+        // Membership list is still usable after the recovery.
+        expect(sess.memberships).toBeDefined();
+    });
+
     it("reemits membershipManager events", () => {
         sess = MatrixRTCSession.sessionForSlot(client, makeMockRoom([rtcMembershipTemplate]), callSession);
         const delayIdChanged = vi.fn();
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -755,9 +755,10 @@ export class MatrixRTCSession extends TypedEventEmitter<
         }
         this.membershipNeedsRecalculation = true;
         // Chain the recalculation.
-        this.recalculateSessionMembersPromise = this.recalculateSessionMembersPromise
-            .finally()
-            .then(() => this.recalculateSessionMembers());
+        this.recalculateSessionMembersPromise = this.recalculateSessionMembersPromise.then(
+            () => this.recalculateSessionMembers(),
+            () => this.recalculateSessionMembers(),
+        );
         return this.recalculateSessionMembersPromise;
     }
 


toger5 (approved):

toger5:
Thanks for finding this issue!

PR #5340, opened 2026-05-22
MSC4140: use auth for delayed event management

AndrewFerr:
Do this to let ratelimiting apply per user instead of per source IP address, if the server supports that.

See https://github.com/element-hq/synapse/pull/19794

## Checklist


dbkr (changes requested):
Can you add a comment to say that's why it's doing an authed request please?

AndrewFerr:
> Can you add a comment to say that's why it's doing an authed request please?

0f2c39cc1 adds a comment adapted from the commit message of 7889b99.

dbkr (approved):
Thanks

PR #5349, opened 2026-06-01
Fix duplicate sticky events for encrypted events without a sticky key

### LINKED ISSUE #5205: Bug | Sticky event maps register several events for the same sticky key
I was building a devtool utility for sticky events, and I noticed that the `_unstable_getStickyEvents` returns several events per unique `msc4354_sticky_key`

Image

This is unexpected, there should be only one active sticky event per sticky key.

The problem is that the sticky key is in the event content and is encrypted.
But the `unstableStickyExpiresAt` is at the event root.

## What happens

1.  The SDK consider an encrypted event as sticky (because there is a `unstableStickyExpiresAt` in the event)
2. Then it tries to get the `msc4354_sticky_key` in the content. But it fails because the event is not yet decrypted (decryptionPromise is pending)
3. So the event is added in the list of `unkeyed` events [unkeyedStickyEvents](https://github.com/matrix-org/matrix-js-sdk/blob/5739b59faaf33fa55457dcaa76ffebe0506d1466/src/models/room-sticky-events.ts#L51), that is events that have no sticky key. 

later on the event is decrypted (and mutated in place), but it wrongly stays in the `unkeyedStickyEvents` list!

## What should happen

The `RoomStickyEventsStore` should wait for event to be decrypted before processing it

AurimasG12:
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/5205

## Changes

- `_unstable_addStickyEvents` in room.ts is now `async` and runs each sticky through `decryptStickyEvent` before calling the store, so encrypted stickies are decrypted (and get their `msc4354_sticky_key`) before insertion. `RoomStickyEventsStore` rejects unkeyed m.room.encrypted events so ciphertext cannot sit in `unkeyedStickyEvents`. sync.ts and embedded.ts now await that async add path.
- Tests were added in room-sticky-events.spec.ts and room.spec.ts for the store guard and the decrypt-then-add flow.

## Cause
Encrypted sticky events arrive as `m.room.encrypted` with no `msc4354_sticky_key` in the ciphertext, so the SDK treated them as unkeyed stickies and stored them immediately. After decryption, the same logical event was added again (now with a sticky key or as a plain event), so the sticky store held two entries for one sticky.

## Fix
`_unstable_addStickyEvents` now decrypts sticky events first via `decryptStickyEvent`, and only passes through events that are already keyed, plain unkeyed stickies, or successfully decrypted. As a backstop, `RoomStickyEventsStore` refuses to add unkeyed `m.room.encrypted` events, and sync/widget paths `await` the async add so decryption finishes before the store is updated.

## Checklist


src/models/room-sticky-events.ts
```diff
@@ -156,6 +157,8 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
         const stickyEvent = event as StickyMatrixEvent;
 
         if (stickyKey === undefined) {
+            if (event.isEncrypted() || type === EventType.RoomMessageEncrypted) return { added: false };
+
             this.unkeyedStickyEvents.add(stickyEvent);
```
toger5, line 182: This seems to make more sense to be done further up the top?
We dont need to read: `sender` `assertIsUserId` ... if its encrypted.

src/models/room-sticky-events.ts
toger5, line 157: I think a check just at the beginning should be cleaner?
```suggestion
        const type = event.getType();
        // Early excit if we are trying to add an encrypted event (skip it, the js-sdk-will call it again with the unencrypted event.
        if (event.isEncrypted()) {
            logger.warn("addStickyEvent called with encrypted event. Events are expected to be decrypted before adding!");
            return { added: false }
        };
        const stickyKey = event.getContent().msc4354_sticky_key;
```

src/models/room.ts
```diff
@@ -3475,8 +3475,48 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+
+        try {
+            await this.client.decryptEventIfNeeded(event);
+            if (event.isBeingDecrypted()) await event.getDecryptionPromise();
+        } catch (e) {
+            logger.warn(`Failed to decrypt sticky event ${event.getId()}, skipping`, e);
+            return null;
+        }
+
+        if (typeof event.getContent().msc4354_sticky_key === "string") return event;
```
toger5: I am not sure this is correct. The MSC does not mandate the sticky_key to be present.
It is also checked later to determine if it ends up in the unkeyed events list or the keyed list. So here we can skip the check.

See:
```
        if (typeof stickyKey !== "string" && stickyKey !== undefined) {
            throw new Error(`${event.getId()} is missing msc4354_sticky_key`);
        }
```

src/models/room-sticky-events.ts
toger5: Those two are also 100% equivalent right? `event.isEncrypted()` vs `type === EventType.RoomMessageEncrypted`

src/models/room.ts
```diff
@@ -3475,8 +3475,48 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
...
+
+    /**
+     * Resolves a sticky event to its decrypted form before it is inserted into the store.
+     *
+     * Returns `null` for events that should not be stored: those with missing sticky metadata,
+     * or encrypted events that still lack a `msc4354_sticky_key` after a decryption attempt.
+     * Encryption failures are logged and treated as skipped events so that a single
+     * undecryptable sticky cannot abort the surrounding sync processing.
+     */
+    private async decryptStickyEvent(event: MatrixEvent): Promise<MatrixEvent | null> {
```
toger5: This function seems to do a couple of checks that are redundant.

src/models/room.ts
```diff
@@ -3475,8 +3475,48 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
      * @internal
      */
     // eslint-disable-next-line
-    public _unstable_addStickyEvents(events: MatrixEvent[]): ReturnType<RoomStickyEventsStore["addStickyEvents"]> {
-        return this.stickyEvents.addStickyEvents(events);
+    public async _unstable_addStickyEvents(
+        events: MatrixEvent[],
+    ): Promise<ReturnType<RoomStickyEventsStore["addStickyEvents"]>> {
```
toger5: where is this coming from:
`Promise<ReturnType<RoomStickyEventsStore["addStickyEvents"]>>`
this: `ReturnType<RoomStickyEventsStore["addStickyEvents"]>` is `void` right?

toger5 (commented):
I added a couple of comments and looked into the code.

The more I look the more I see, that the JS sdk is already doing a good amount of checks.

especially `decryptEventIfNeeded` is rather powerful. It checks `shouldAttemptDecryption` and `isBeingDecrypted`.


So my current guess is that the cleanest solution would be to fix the encryption situation just in the addStickyEvent() method (not in `addStickyEvents`)
Since we do not have the encryption info in the type signature, I prefer changing `addStickyEvent` to work with both. encrypted and unencrypteed events.
so adding `await decryptEventIfNeeded` as the first line in `addStickyEvent` and making `addStickyEvent` async.
Some of the events (all from the sticky section should already be getting decrypted due to `mapSyncEventsFormat` in sync.ts. And the one from the timeline will be decrypted before checking the content.

I find the current `decryptStickyEvent` method confusing as it implies a lot of things need custom checks/solutions that the js-sdk already does/abstracts.

Combined this should:
 - eliminate the need for: `decryptStickyEvent`
 - change `_unstable_addStickyEvents` to:
```
 public async _unstable_addStickyEvents(
        events: MatrixEvent[],
    ): void {
        return void this.stickyEvents.addStickyEvents(decrypted);
    }
```

 - and change `addStickyEvents` to:
```
/**
     * Add a series of sticky events, emitting `RoomEvent.StickyEvents` if any
     * changes were made.
     * @param events A set of new sticky events.
     */
    public async addStickyEvents(events: MatrixEvent[]): Promise<void> {
        const added: StickyMatrixEvent[] = [];
        const updated: { current: StickyMatrixEvent; previous: StickyMatrixEvent }[] = [];
        Promise.allSettled(
            events.map(async (event) => {
                try {
                    // We use allSettled (instead of a for loop) as we do not care about the order.
                    // addStickyEvent will ignore updates if it already has a newer event. (if they get processed in the "wrong" order)
                    const result = await this.addStickyEvent(event);
                    if (result.added) {
                        if (result.prevEvent) {
                            // e is validated as a StickyMatrixEvent by virtue of `addStickyEvent` returning added: true.
                            updated.push({ current: event as StickyMatrixEvent, previous: result.prevEvent });
                        } else {
                            added.push(event as StickyMatrixEvent);
                        }
                    }
                } catch (ex) {
                    logger.warn("ignored invalid sticky event", ex);
                }
            }),
        );
        if (added.length || updated.length) this.emit(RoomStickyEventsEvent.Update, added, updated, []);
        this.scheduleStickyTimer();
    }
```

This is possible due to addStickyEvents NOT returning the added event but relying on an emitter based approach.
So we can `void` their promises and make them async without needing to make _unstable_addStickyEvents async.
Does this sound sane?

src/sync.ts
```diff
@@ -1470,7 +1470,7 @@ export class SyncApi {
                 );
                 // Note: We calculate sticky events before emitting `.Room` as it's nice to have
                 // sticky events calculated and ready to go.
-                room._unstable_addStickyEvents(stickyEventsAndStickyEventsFromTheTimeline);
+                await room._unstable_addStickyEvents(stickyEventsAndStickyEventsFromTheTimeline);
```
toger5: I am not sure its the right thing to block the sync loop on this?
In this case we are lucky, that the sticky api is event based. So we do not need to await.

toger5: Or is there a particular concern with this from the initial review:
> change _unstable_addStickyEvents to:
```
public async _unstable_addStickyEvents(
        events: MatrixEvent[],
    ): void {
        return void this.stickyEvents.addStickyEvents(events);
    }
```

src/models/room-sticky-events.ts
```diff
@@ -129,7 +134,26 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
      * @returns An object describing whether the event was added to the map,
      *          and the previous event it may have replaced.
      */
-    private addStickyEvent(event: MatrixEvent): { added: true; prevEvent?: StickyMatrixEvent } | { added: false } {
+    private async addStickyEvent(
+        event: MatrixEvent,
+    ): Promise<{ added: true; prevEvent?: StickyMatrixEvent } | { added: false }> {
+        if (this.decryptEventIfNeeded) {
+            try {
+                await this.decryptEventIfNeeded(event);
```
toger5, line 142:
```suggestion
                await this.decryptEventIfNeeded(event);
                if(event.getType() == EventType.RoomMessageEncrypted) throw Error("still encrypted after decryption attempt");
```

src/models/room-sticky-events.ts
```diff
@@ -129,7 +134,26 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
...
+            }
+        }
+
+        const type = event.getType();
+        if (type === EventType.RoomMessageEncrypted) {
+            logger.warn(
+                "addStickyEvent called with encrypted event. Events are expected to be decrypted before adding!",
+            );
+            return { added: false };
+        }
```
toger5, line 155:
```suggestion
```

toger5: This case should never occur. It still feels saver to do the additional check however. But we should just add it after the await above.
Otherwise its a bit noisy

src/models/room-sticky-events.ts
toger5: or similar error msg

src/models/room.ts
```diff
@@ -453,7 +453,7 @@ export class Room extends ReadReceipt<RoomEmittedEvents, RoomEventHandlerMap> {
     /**
      * Stores and tracks sticky events
      */
-    private stickyEvents = new RoomStickyEventsStore();
+    private stickyEvents = new RoomStickyEventsStore((event) => this.client.decryptEventIfNeeded(event));
```
toger5, line 456: I am not following this change?
Was it there before?
What changed so the constructor can take a callback?

toger5 (changes requested):
```
public _unstable_addStickyEvents(events: MatrixEvent[]): ReturnType<RoomStickyEventsStore["addStickyEvents"]> {
        return this.stickyEvents.addStickyEvents(events);
    }
```
This should get an explicit `void` so we can see the promise gets dropped.

toger5:
This test also needs to be investigated: `should NOT filter sticky events before the retention period`
Looks like it actually might point to a regression, but could also just be a test that needs adaption.

Half-Shot:
@AurimasG12 Are you still interested in working on this?

dbkr:
Moving this to closed as there's been no response. We can re-open if someone would like to pick it up.

PR #5350, opened 2026-06-01
Update user profile data stores before emit

dbkr:
So the listeners get the new data

## Checklist


florianduros (commented):
What the PR is trying to fix?

dbkr:
Specifically it led to a bug in element web where the user status icon would update to the one before the latest because it got the event but the data hadn't been updated yet.

florianduros:
Hmm maybe it can fix https://github.com/element-hq/element-web/issues/33652?

florianduros (approved):

dbkr:
I think that bug is actually talking about presence.

PR #5355, opened 2026-06-03
SECURITY: Update matrix-sdk-crypto-wasm to v18.3.1

dbkr:
Updates to a new version of matrix-sdk-crypto-wasm to resolve [GHSA-wfq4-36m3-9g42](https://github.com/matrix-org/matrix-rust-sdk/security/advisories/GHSA-wfq4-36m3-9g42) / [CVE-2026-45056](https://www.cve.org/CVERecord?id=CVE-2026-45056).

## Checklist

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -10,3 +10,6 @@ overrides:
 
 allowBuilds:
     unrs-resolver: false
+
+minimumReleaseAgeExclude:
+    - "@matrix-org/matrix-sdk-crypto-wasm@18.3.1"
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -15,4 +15,4 @@ minimumReleaseAgeExclude:
     # Bypassing for this version, obviously this is only necessary
     # temporarily so if you're reading this, you can almost certainly
     # remove it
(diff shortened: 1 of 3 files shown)


pnpm-workspace.yaml
```diff
@@ -10,3 +10,9 @@ overrides:
 
 allowBuilds:
     unrs-resolver: false
+
+minimumReleaseAgeExclude:
+    # Bypassing for this version, obviously this is only necessary
+    # temporarily so if you're reading this, you can almost certainly
+    # remove it
+    - "@matrix-org/matrix-sdk-crypto-wasm@18.3.1"
```
t3chguy: Why not just allowlist the matrix-org scope? https://github.com/element-hq/element-web/blob/develop/pnpm-workspace.yaml#L116-L120

dbkr: I did wonder since we trust ourselves, although possibly if were ever to release a pwned package it might help contain the fallout a little?

t3chguy: Not really as this config isn't read by npm consumers, so would just prevent us (renovate) updating to it, which would be caught in review anyway

dbkr: ah ok, yeah

t3chguy (approved):

PR #5189, opened 2026-02-13
Handle secret pushing for key backups

uhoreg:
- push backup key when we reset backup
- handle pushed backup key

fixes https://github.com/matrix-org/matrix-js-sdk/issues/5143

depends on https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/295 (and a new release of the wasm bindings)

## Checklist

- [x] Tests written for new code (and old code if feasible).
- [x] New or updated `public`/`exported` symbols have accurate [TSDoc](https://tsdoc.org/) documentation.
- [x] Linter and other CI checks pass.
- [x] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md)).

PR #5357, opened 2026-06-03
Add integ test that key backup is enabled after receiving key

richvdh:
Adds a regression test for https://github.com/element-hq/element-web/issues/32894.

It turns out that this was fixed already, by https://github.com/matrix-org/matrix-js-sdk/pull/5189 (and this test fails on the code before that PR landed)


t3chguy (approved):

andybalaam (approved):
Looks good, thanks!

PR #5371, opened 2026-06-15
Switch from prettier to oxfmt

t3chguy:
For https://github.com/element-hq/wat-internal/issues/497


.oxfmtrc.jsonc
MidhunSureshR, line 1: This could be a `.jsonc`

MidhunSureshR (approved):

PR #5376, opened 2026-06-17
Switch from eslint to oxlint

t3chguy:
Some rules are missing, e.g. camelCase
Some rules are differently opinionated, some even a tiny bit buggy.

It respects the old eslint comments so didn't remove those.

But we don't need to strive for perfection when it is this much faster, taking 1.7s to lint every file in the repo. IDE lint hints are instant now.

It identified a handful of what looked like legitimate bugs in usages of unbound methods in event handlers, around thread & receipts, which are likely to cause real runtime bugs.


florianduros (approved):

src/embedded.ts
```diff
@@ -527,19 +527,17 @@ export class RoomWidgetClient extends MatrixClient {
...
 
-        let updateDelayedEvent: (delayId: string) => Promise<unknown>;
         switch (action) {
             case UpdateDelayedEventAction.Cancel:
-                updateDelayedEvent = this.widgetApi.cancelScheduledDelayedEvent;
+                await this.widgetApi.cancelScheduledDelayedEvent(delayId).catch(timeoutToConnectionError);
                 break;
             case UpdateDelayedEventAction.Restart:
-                updateDelayedEvent = this.widgetApi.cancelScheduledDelayedEvent;
+                await this.widgetApi.cancelScheduledDelayedEvent(delayId).catch(timeoutToConnectionError);
```
robintown, line 535: The invoked method should read `restartScheduledDelayedEvent` here. Looks like that's always been a typo.

robintown (approved):
I have reviewed the webrtc/\*, matrixrtc/\*, and embedded.ts changes on behalf of VoIP

toger5 (approved):
I am shocked by the 3k loc reduction of pnpm-lock. (positively shocked)

src/embedded.ts
t3chguy: @robintown are you suggesting I change it here or will you take it in another PR?

andybalaam (approved):
Nice!

src/embedded.ts
robintown: Don't bother, I will take care of it.

.oxlintrc.jsonc
```diff
@@ -0,0 +1,179 @@
...
+    "plugins": ["eslint", "typescript", "unicorn", "import", "jsdoc", "node", "promise", "vitest"],
+    "categories": {
+        "correctness": "error",
+        "perf": "error",
+    },
+    "options": {
+        "typeAware": true,
+        "reportUnusedDisableDirectives": "off",
+        "maxWarnings": 0,
+        "denyWarnings": true,
```
Johennes: Should one of these be enough?

```
options.denyWarnings

type: boolean

Ensure warnings produce a non-zero exit code.

Equivalent to passing --deny-warnings on the CLI.
options.maxWarnings

type: integer

Specify a warning threshold. Exits with an error status if warnings exceed this value.

Equivalent to passing --max-warnings on the CLI.
```

src/embedded.ts
robintown: https://github.com/matrix-org/matrix-js-sdk/pull/5380

PR #5389, opened 2026-06-26
Add method `CryptoApi.getUserCrossSigningKeys`

richvdh:
I want to expose user identities in the dev tools, so we need to expose it from the js-sdk.

### DIFF
--- a/src/crypto-api/index.ts
+++ b/src/crypto-api/index.ts
@@ -1222,10 +1222,15 @@ export interface CreateSecretStorageOpts {
 
 /** Types of cross-signing key */
 export enum CrossSigningKey {
-    Master = "master",
-    SelfSigning = "self_signing",
-    UserSigning = "user_signing",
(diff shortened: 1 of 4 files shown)


t3chguy (approved):

src/crypto-api/index.ts
```diff
@@ -1222,11 +1222,16 @@ export interface CreateSecretStorageOpts {
 
 /** Types of cross-signing key */
 export enum CrossSigningKey {
```
andybalaam, line 1224: I might argue that this should be called `CrossSigningKeyType` but I don't think it's worth breaking compatibility for the change.

andybalaam (approved):
All good, thanks.

src/crypto-api/index.ts
richvdh: indeed.

PR #5390, opened 2026-06-29
Properly support Matrix v1.18 OAuth2 APIs

### LINKED ISSUE #5304: matrix-js-sdk requires `.well-known/openid-configuration` endpoint to log in with OAuth
While working on implementing support for OAuth login in Continuwuity, I found that the matrix-js-sdk sends a request to `https://<issuer>/.well-known/openid-configuration` before opening the `authorization_endpoint`, performed by [this call to `createSigninRequest`](https://github.com/matrix-org/matrix-js-sdk/blob/69985ee350a33ba75f1ad11f96468344f0c92a8d/src/oidc/authorize.ts#L180) which calls [`getAuthorizationEndpoint`](https://github.com/authts/oidc-client-ts/blob/9a30d6e5a97bbc4d5e10e650458c398d4e86b332/src/OidcClient.ts#L126) in oidc-client-ts. This behavior is described in [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414#section-3) but is not required by the Matrix specification, which has [a dedicated endpoint](https://spec.matrix.org/v1.18/client-server-api/#server-metadata-discovery) for fetching this configuration. (MAS responds to this well-known URL with the same response as the Matrix-specific discovery endpoint.) I would expect the SDK to only use the Matrix-specific discovery endpoint instead of also sending requests to the RFC 8414 endpoint.

### LINKED ISSUE #5305: matrix-js-sdk sends `openid` scope when performing authorization code grant
While working on implementing support for OAuth login in Continuwuity, I found that the matrix-js-sdk includes the `openid` scope when performing the authorization code grant, which is not one of the [allocated scopes](https://spec.matrix.org/v1.18/client-server-api/#allocated-scope-tokens) listed in the Matrix specification. I believe that this behavior is caused by the oidc-client-ts library injecting the scope, since the SDK itself [does not request it](https://github.com/matrix-org/matrix-js-sdk/blob/69985ee350a33ba75f1ad11f96468344f0c92a8d/src/oidc/authorize.ts#L65), but I don't know where in the library the scope is added. Since this scope is only relevant for OIDC, not OAuth, I would expect the SDK not to provide it when signing in.

### LINKED ISSUE #32617: OAuth 2.0-based login fail if the device clock is not in sync
### Steps to reproduce

1. Have your system clock a few minutes ahead of the time
2. Try to login on a OAuth 2.0-enabled server
3. The callback to Element Web will fail

#### What did you expect?

The login to work/go through

#### What happened instead?

The login failed, because Element Web asks for the `openid` scope, which makes MAS issue an `id_token`.
The `openid` scope isn't necessary (and arguably a spec-violation, since EW won't work on servers that don't support the `openid` scope), and EW *verifies* the `id_token` claims. In those claims, there is an expiration timestamp, that will make the verification fail if it is too much in the past. There is also a 'not before'/'issued at' timestamp that will fail if it is too much in the future.

### Operating system

_No response_

### Browser information

_No response_

### URL for webapp

develop.element.io

### Application version

Element Nightly version: 2026022301 Crypto version: Rust SDK 0.16.0 (82e97d4), Vodozemac 0.9.0

### Homeserver

matrix.org

### Will you send logs?

Yes

t3chguy:
Drop support for `/auth_issuer` fallback in `getAuthMetadata`

Update Matrix auth APIs to support what is specced in v1.18 rather than an early variant of the originating MSC

Dropping `oidc-client-ts` & `jwt-decode` in the process

Fixes https://github.com/matrix-org/matrix-js-sdk/issues/5305
Fixes https://github.com/matrix-org/matrix-js-sdk/issues/5304
Fixes https://github.com/element-hq/element-web/issues/32617
Requires https://github.com/element-hq/element-web/pull/34026


spec/unit/oauth/authorize.spec.ts
```diff
@@ -0,0 +1,365 @@
+/**
+ * @vitest-environment happy-dom
+ */
+
+/*
+Copyright 2023 The Matrix.org Foundation C.I.C.
```
dbkr, line 6: year?

spec/unit/oauth/discover.spec.ts
```diff
@@ -0,0 +1,66 @@
+/*
+Copyright 2023 The Matrix.org Foundation C.I.C.
```
dbkr, line 2: year here too

src/@types/type-guards.ts
```diff
@@ -0,0 +1,91 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { logger } from "../logger.ts";
+
+export const isRecord = (value: unknown): value is Record<string, unknown> =>
```
dbkr, line 23: I feel like, while these are mostly obvious, some tsdoc wouldn't hurt

src/oauth/authorize.ts
```diff
@@ -0,0 +1,274 @@
+/*
+Copyright 2023 The Matrix.org Foundation C.I.C.
```
dbkr, line 2: year

src/oauth/discover.ts
```diff
@@ -0,0 +1,102 @@
+/*
+Copyright 2023 The Matrix.org Foundation C.I.C.
```
dbkr, line 2: year

dbkr (commented):
Looks generally sensible!

andybalaam (approved):

spec/unit/oauth/discover.spec.ts
t3chguy: The stuff here was moved from `validate.spec.ts` - nothing new was written

src/oauth/authorize.ts
t3chguy: The stuff here was moved from oidc/authorize.ts, almost no new code was written, mostly removed

src/oauth/discover.ts
t3chguy: The stuff here was moved from oidc/discovery.ts, almost no new code was written, mostly removed

spec/unit/oauth/authorize.spec.ts
t3chguy: Github isn't seeing it as a move but it is https://github.com/matrix-org/matrix-js-sdk/blob/develop/spec/unit/oidc/authorize.spec.ts

spec/unit/oauth/discover.spec.ts
dbkr: ok, figured that may be the case

dbkr (approved):

PR #5404, opened 2026-07-08
Ignore Content-Type parameters when polling MSC4108 rendezvous channel

krithin:
MSC4108RendezvousSession.receive() compared the Content-Type header against "text/plain" with strict string equality. If a reverse proxy in front of the rendezvous server appends parameters to the header (e.g. nginx's `charset` directive rewrites it to `text/plain; charset=utf-8`), every incoming payload was treated as "no new message" and silently discarded, deadlocking QR code login until the session expired.

Parse out the media type per RFC 9110 (parameters follow the type/subtype after a semicolon; type names are case-insensitive) before comparing.

## Checklist

### DIFF
--- a/spec/unit/rendezvous/MSC4108RendezvousSession.spec.ts
+++ b/spec/unit/rendezvous/MSC4108RendezvousSession.spec.ts
@@ -210,6 +210,34 @@ describe("MSC4108RendezvousSession", () => {
         }
     });
 
+    it("GET with content-type parameters, e.g. added by an intermediary proxy", async function () {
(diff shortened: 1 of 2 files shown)


t3chguy:
Unfortunately the spec dictates `text/plain` and does not mention dealing with any parameters https://github.com/matrix-org/matrix-spec-proposals/blob/87f8317a902cd7bc5c2d2d225f71021b3a509e2d/proposals/4108-oidc-qr-login.md

This is the v2024 edition of MSC4108 which is the only variant currently supported. The v2025 (latest) edition of the MSC does not have this issue seemingly.

PR #5412, opened 2026-07-13
docs: clarify QR-code verification flow and generateQRCode preconditions

### LINKED ISSUE #5092: Error: Unsupported verification method
Hi, I'm getting an error when trying to verify a user via QR. I've tried using the SAS method, and it worked, but for some reason when I try to use any of the QR method it does not work. (Also, on the documentation is not very clear on how should I use this methods).

When the `VerificationRequest` instance is on phase `Ready` , if I try to call this functions, I always get the same error:

```ts
import { VerificationMethod } from 'matrix-js-sdk/lib/types';


request.startVerification(VerificationMethod.Reciprocate);
request.startVerification(VerificationMethod.ScanQrCode);
request.startVerification(VerificationMethod.ShowQrCode);
```

`Error: Unsupported verification method: m.reciprocate.v1`
`Error: Unsupported verification method: m.qr_code.show.v1`
`Error: Unsupported verification method: m.qr_code.scan.v1`

I also tried explicitly adding all those methods to the `verificationMethods` property on the initialization params, but it doesn't work.

If I try to do something like 
```ts
const data = await request.generateQRCode()
```
I always get undefined. Even if the current phase is `Ready`. The method Docstring says:

>Generate the data for a QR code allowing the other device to verify this one, if it supports it.
Only returns data once phase is [VerificationPhase.Ready](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) and the other party can scan a QR code; otherwise returns undefined.

Honestly, I don't know what I'm doing wrong. Is it a bug?

The error seems to come from a web assembly file, so its impossible for me to debug this.

Image

**If I click on `matrix_sdk_crypto_wasm_bg.wasm`** I see this:

Image

mvanhorn:
## Summary

Clarifies the TSDoc for the QR-code verification flow on `VerificationRequest` in
`src/crypto-api/verification.ts`. No runtime or behavioural change; comments only.

- `startVerification()` now states that it is for SAS (emoji) verification (`m.sas.v1`) and will reject
  QR-code methods with an "Unsupported verification method" error, and points to `generateQRCode()` /
  `scanQRCode()` for the QR flow.
- `generateQRCode()` now enumerates the conditions under which it returns data versus `undefined`:
  `phase` is `Ready`, the other party supports scanning (`m.qr_code.scan.v1`), and the local device has
  its cross-signing keys available. The missing cross-signing keys case is the common cause of an
  unexpected `undefined` (see the test at
  `spec/integ/crypto/verification.spec.ts:438`: "QRCode fails if we don't yet have the cross-signing
  keys").
- `scanQRCode()` gains a one-line usage note that it is called after scanning the other device's QR code.

## Why this matters

Reported in #5092: it is not obvious how to do QR-code verification with the crypto API. Callers reach
for `startVerification(VerificationMethod.ScanQrCode | ShowQrCode)` and hit "Unsupported verification
method", then find `generateQRCode()` returns `undefined` even when `phase === Ready` and
`otherPartySupportsMethod(...)` is true for the QR methods, with no documented reason. As @richvdh noted
on the issue, QR is supported but "it's not obvious how to use it", and contributions to the
documentation to make it clearer would be much appreciated. These docstrings capture the SAS-vs-QR split
and the cross-signing precondition so the flow is discoverable from the API reference.

Closes #5092

## Checklist

- [x] Tests written for new code (and old code if feasible). — N/A: documentation-only change, no code modified. Preconditions documented are grounded in the existing test `spec/integ/crypto/verification.spec.ts:438`.
- [x] Linter and other CI checks pass. — `tsc --noEmit` passes; `oxlint` and `oxfmt --check` clean on the changed file.

### DIFF
--- a/src/crypto-api/verification.ts
+++ b/src/crypto-api/verification.ts
@@ -119,7 +119,15 @@ export interface VerificationRequest extends TypedEventEmitter<
     /**
      * Send an `m.key.verification.start` event to start verification via a particular method.
      *
-     * This is normally used when starting a verification via emojis (ie, `method` is set to `m.sas.v1`).
+     * This is used for SAS (emoji) verification: `method` should be set to `m.sas.v1`. It does not start
+     * QR-code verification, and passing a QR-code method (such as `m.reciprocate.v1`, `m.qr_code.scan.v1`,
+     * or `m.qr_code.show.v1`) will be rejected with an "Unsupported verification method" error.
+     *
+     * For QR-code verification, use {@link VerificationRequest#generateQRCode} to display a QR code to the
+     * other device, or {@link VerificationRequest#scanQRCode} to consume a QR code scanned from it. This is
+     * only possible once the `phase` is {@link VerificationPhase.Ready}; when it is, a client can typically
+     * offer the user any of three options: show a QR code, scan the other party's QR code, or fall back to
+     * emoji (SAS) verification via this method.
      *
      * @param method - the name of the verification method to use.
      *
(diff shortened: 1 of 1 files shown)


andybalaam (approved):
Great, thanks!

PR #5419, opened 2026-07-14
Improve types for `/room_keys/version` requests

richvdh:
Currently we have a single interface which serves both as the request body for `POST /_matrix/client/v3/room_keys/version`, and the response body for `GET /_matrix/client/v3/room_keys/version`. Although the two are related, the latter has a bunch of mandatory fields, so the single interface serves neither purpose well.

Let's create a new type for the POST request body.


uhoreg (approved):
looks sensible

t3chguy (approved):

richvdh:
This needs https://github.com/element-hq/element-web/pull/34273 to land first

PR #5290, opened 2026-04-20
Support stable identifiers for MSC4268

andybalaam:
## Checklist

- [x] Tests written for new code (and old code if feasible).
- [x] New or updated `public`/`exported` symbols have accurate [TSDoc](https://tsdoc.org/) documentation.
- [ ] Linter and other CI checks pass.
- [x] Sign-off given on the changes (see [CONTRIBUTING.md](https://github.com/matrix-org/matrix-js-sdk/blob/develop/CONTRIBUTING.md)).

PR #5421, opened 2026-07-15
Bring generate-test-data into line with the file it generates

andybalaam:
It turns out that in #5202 I updated the test data without updating the script that generates it, and conversely in #5290 I updated the script without re-running it. This PR brings them back into sync.

## Checklist


spec/test-utils/test-data/generate-test-data.py
```diff
@@ -270,9 +271,12 @@ def build_test_data(user_data, prefix = "") -> str:
 /** The encrypted CLEAR_EVENT by MEGOLM_SESSION_DATA */
 export const {prefix}ENCRYPTED_EVENT: Partial<IEvent> = {json.dumps(encrypted_event, indent=4)};
 
-/** base64-encoded backup decryption (private) key */
+/** base64-encoded backup decryption (private) key that matches the public key in {prefix}CURVE25519_KEY_BACKUP_DATA */
 export const {prefix}BACKUP_DECRYPTION_KEY_BASE64 = "{ user_data['B64_BACKUP_DECRYPTION_KEY'] }";
 
+/** base64-encoded backup decryption (private) key that does not match the public key in {prefix}CURVE25519_KEY_BACKUP_DATA */
+export const {prefix}BACKUP_DECRYPTION_KEY_BASE64_ALT = "dh4fP2LITyJusgnb0dEq/SQK253WGObvLxXF5FEX6qc";
```
richvdh, line 278: could we write a comment about how this string was created?

richvdh (approved):

richvdh:
Sidebar: I wonder if these files should have CODEOWNERS be the crypto team, since it is basically crypto data?

florianduros (approved):

spec/test-utils/test-data/generate-test-data.py
andybalaam: Done in 5df843c9b

andybalaam:
> Sidebar: I wonder if these files should have CODEOWNERS be the crypto team, since it is basically crypto data?

Proposed in https://github.com/matrix-org/matrix-js-sdk/pull/5422

PR #5422, opened 2026-07-15
Make test-data owned by Crypto team

andybalaam:
Since this directory is for generating crypto keys etc. for tests, so is usually Crypto team business.

(Feel free to push back though - this is just a suggestion from me.)

## Checklist


.github/CODEOWNERS
```diff
@@ -1,17 +1,18 @@
...
-/spec/unit/crypto         @matrix-org/element-crypto-web-reviewers
-/spec/unit/rust-crypto    @matrix-org/element-crypto-web-reviewers
+/src/crypto-api            @matrix-org/element-crypto-web-reviewers
+/src/crypto                @matrix-org/element-crypto-web-reviewers
+/src/rust-crypto           @matrix-org/element-crypto-web-reviewers
+/spec/integ/crypto         @matrix-org/element-crypto-web-reviewers
+/spec/unit/crypto.spec.ts  @matrix-org/element-crypto-web-reviewers
+/spec/unit/crypto          @matrix-org/element-crypto-web-reviewers
+/spec/unit/rust-crypto     @matrix-org/element-crypto-web-reviewers
+/spec/test-utils/test-data @matrix-org/element-crypto-web-reviewers
```
t3chguy: I think this should be better scoped, e.g. by creating a crypto dir within.

richvdh: The current contents of that directory are all about crypto data (per https://github.com/matrix-org/matrix-js-sdk/blob/develop/spec/test-utils/test-data/generate-test-data.py#L18). You may have a point that the directory should be named to reflect that, though.

richvdh: I think the rest of the file is approximately alphabetically sorted, so it would be good to maintain that.

andybalaam: Renamed the dir to `crypto-test-data` in 1cb24fd2a

andybalaam: Sorted and separated into sections in 0211da5d6

t3chguy (approved):

dbkr (approved):

dbkr:
seems reasonable

richvdh (approved):

PR #5423, opened 2026-07-15
Add can_subscribe on RTC member events and move rtc_transports to transports.published

Johennes:
This adds `transports.can_subscribe` from https://github.com/matrix-org/matrix-spec-proposals/pull/4143. It's always hardcoded to `["livekit"]` for now as we have no other transports yet.

Fixes: element-hq/voip-internal#645

## Checklist


BillCarsonFr:
Given this is updating to the new format, why not remove the deprecated `rtc_transports`?
Not sure we need to do any sort of transition, this is behind a lab flag can break

Johennes:
> Given this is updating to the new format, why not remove the deprecated `rtc_transports`? Not sure we need to do any sort of transition, this is behind a lab flag can break

Good point. Have added the change in the last commit.

BillCarsonFr (approved):
👍

PR #5425, opened 2026-07-16
Implement MSC4506 knock push notifications

ara4n:
This lets js-sdk know about push notifications for knock rules as per MSC4506, as needed for https://github.com/element-hq/element-web/pull/34268

### DIFF
--- a/src/pushprocessor.ts
+++ b/src/pushprocessor.ts
@@ -515,7 +515,7 @@ export class PushProcessor {
     }
 
     /**
-     * MSCxxxx `recipient_permission` condition: matches if the user these push
+     * MSC4506 `recipient_permission` condition: matches if the user these push
      * rules are being evaluated for (i.e. us) has a power level at least that
      * required to perform the `m.room.power_levels` action named by `key`
      * (e.g. "invite"), in the room the event is in.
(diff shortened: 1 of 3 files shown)


src/client.ts
t3chguy, line 1: This seems unrelated to MSC4506

ara4n: hm, yup - this is autojoin code for MSC4509.

src/pushprocessor.ts
```diff
@@ -510,6 +514,41 @@ export class PushProcessor {
         return room.currentState.mayTriggerNotifOfType(notifLevelKey, ev.getSender()!);
     }
 
+    /**
+     * MSC4506 `recipient_permission` condition: matches if the user these push
+     * rules are being evaluated for (i.e. us) has a power level at least that
+     * required to perform the `m.room.power_levels` action named by `key`
+     * (e.g. "invite"), in the room the event is in.
+     */
+    private eventFulfillsRecipientPermCondition(cond: IRecipientPermissionCondition, ev: MatrixEvent): boolean {
```
t3chguy, line 523: This should be gated behind an opt-in as it is relying on an MSC.

PR #5426, opened 2026-07-16
Defer key bundle downloads on server-initiated joins (MSC4509)

ara4n:
A server-initiated join (e.g. an auto-accepted knock) lands on all of our user's devices in the same sync instant, so each would eagerly download the MSC4268 room key bundle. Instead, only download it when the server designates this device as the claimer (a new org.matrix.msc4509.key_bundle_claim to-device message, only trusted from our own user), or lazily upon the first undecryptable event in the room. Client-initiated joins via joinRoom keep the eager behaviour.

The MSC4509 behaviour is unstable, so it is opt-in via the new `unstableMSC4509KeyBundleClaim` client option: when unset (the default), key_bundle_claim messages are ignored and server-initiated joins keep the eager per-device download. Element Web enables it behind a `share_history_on_autojoin` feature flag (PR to follow).

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1310,6 +1310,11 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     // A manager for determining which invites should be ignored.
     public readonly ignoredInvites: IgnoredInvites;
 
+    /** Map from room ID to the user who invited us, for rooms with a pending invite.
+     * Used to do the key-bundle bookkeeping if the server joins us to the room itself
+     * (e.g. auto-accepting a knock, synapse#16307) rather than via `joinRoom`. */
+    private readonly pendingKeyBundleInviters = new Map<string, string>();
+
     public readonly matrixRTC: MatrixRTCSessionManager;
 
     private serverCapabilitiesService: ServerCapabilities;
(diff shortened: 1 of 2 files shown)


andybalaam:
@ara4n where can we read about this `org.matrix.mscxxxx.key_bundle_claim`? Is this something you are proposing merging now, or e.g. to support an in-progress MSC?

richvdh:
msc4509 it appears.

Please can such unspecced features be gated behind some sort of opt-in from the application (which in turn should gate it behind a labs flag)

src/client.ts
```diff
@@ -1310,6 +1325,24 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+    private readonly pendingKeyBundleInviters = new Map<string, string>();
+
+    /** Map from room ID to inviter, for rooms the server joined us to whose key bundle
+     * we are deferring (MSC4509): the bundle is only downloaded if the server designates
+     * us as the claimer, or lazily on the first undecryptable event in the room. */
+    private readonly lazyKeyBundleRooms = new Map<string, string>();
+
+    /** Room IDs for which an `org.matrix.msc4509.key_bundle_claim` designation arrived
+     * before we processed the corresponding join. */
+    private readonly earlyKeyBundleClaims = new Set<string>();
```
richvdh, line 1340: MatrixClient is already well over 7000 lines long and has far too many properties. Please can we move all the logic and these fields out to a different class, which can be gated as whole behind `unstableMSC4509KeyBundleClaim`?

src/client.ts
```diff
@@ -1430,10 +1464,110 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+                this.pendingKeyBundleInviters.delete(member.roomId);
+                this.lazyKeyBundleRooms.delete(member.roomId);
+                this.earlyKeyBundleClaims.delete(member.roomId);
+            }
+        });
+
+        // MSC4509: the server may designate this device as the one which should eagerly
+        // download a room key bundle after a server-initiated join (see above). Only our
+        // own homeserver can deliver a to-device message with our own user as sender.
+        this.on(ClientEvent.ToDeviceEvent, (event) => {
```
richvdh, line 1527: As Sonar points out, ToDeviceEvent is deprecated.

src/client.ts
```diff
@@ -1430,10 +1464,110 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+                        // Flag that we are waiting for a key bundle, as `joinRoom` would have.
+                        await cryptoBackend.markRoomAsPendingKeyBundle(member.roomId, finalInviter);
+                        // Unlike a client-initiated join, this join happened on all of our
+                        // user's devices at once, so don't eagerly download the bundle from
+                        // every one of them (MSC4509): only claim it now if the server has
+                        // designated us as the claimer; otherwise wait to be designated, or
+                        // for the first undecryptable event in the room.
+                        if (!this.msc4509KeyBundleClaimEnabled) {
+                            // MSC4509 disabled: download eagerly, as `joinRoom` would.
+                            await cryptoBackend.maybeAcceptKeyBundle(member.roomId, finalInviter);
```
richvdh, line 1507: Let's not sneak this change of behaviour in as part of implementing an unstable MSC.

src/client.ts
```diff
@@ -1430,10 +1464,110 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
     }
 
+    /**
+     * Download and import a deferred room key bundle (MSC4509) for a room the server
+     * joined us to, if we haven't already.
+     *
+     * If the room's join hasn't been processed yet (a designation racing ahead of the
+     * join in the same sync), remember the designation for the join handler instead.
+     */
+    private claimLazyKeyBundle(roomId: string, reason: string): void {
```
richvdh, line 1556: This is a somewhat misleading name, given it won't actually "claim" the bundle if we haven't been invited yet.

src/client.ts
```diff
@@ -1430,10 +1464,110 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+            this.earlyKeyBundleClaims.add(roomId);
+            return;
+        }
+        this.lazyKeyBundleRooms.delete(roomId);
+        const cryptoBackend = this.cryptoBackend;
+        if (!cryptoBackend) return;
+        this.logger.info(`Claiming deferred key bundle for ${roomId}: ${reason}`);
+        cryptoBackend.maybeAcceptKeyBundle(roomId, inviter).catch((e) => {
+            this.logger.error(`Error accepting deferred key bundle for room ${roomId}:`, e);
+        });
```
richvdh, line 1568: Presumably this means that if the attempt to download the bundle fails (eg the server gives a 500), we'll never actually do the download? Is that a known problem?

src/client.ts
```diff
@@ -1430,10 +1464,110 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
+        // have done. Watch our own membership: remember who invited us, and on an
+        // invite -> join (or knock -> join: an accepted knock can be auto-joined straight past
+        // the invited state, the invite coalesced away) transition run the same
+        // mark-pending/maybe-accept dance as `joinRoom` (idempotent when the join did come from
+        // `joinRoom`).
+        this.on(RoomMemberEvent.Membership, (event, member, oldMembership) => {
+            if (member.userId !== this.getUserId()) return;
+            if (member.membership === KnownMembership.Invite) {
+                const inviter = event.getSender();
+                if (inviter) this.pendingKeyBundleInviters.set(member.roomId, inviter);
```
richvdh, line 1479: This seems like it will fail in the case of a gappy sync, since we might not see the invite event?

richvdh (changes requested):
Some thoughts

richvdh:
Putting this back to draft for now.

Issue #5206, opened 2026-03-03
After a refresh all expired sticky events get resurrected?


BillCarsonFr:
See capture for reproduction.

All sticky events where expired, then I refreshed the browser, and I can see a bunch of sticky event "resurrected" to their initial TTL

![Image](https://github.com/user-attachments/assets/cc4dcb93-9bc8-4444-a1b6-38e1e220c09a)

BillCarsonFr:
Hello @ibharathboga , I just created the issue but haven't looked at the reason. @Half-Shot touched this code more recently

PR #5432, opened 2026-07-17
Fix for sticky event duration resets on page refresh

crime-think:
Fix for #5206 by using absolute time alongside duration


t3chguy:
Please update your title to make sense as a changelog entry

t3chguy (commented):
This could do with a test of some description

dbkr (commented):
Where does `msc4354_sticky_ttl_start_ms` come from? The MSC doesn't mention it.

Also, yes, test please.

Half-Shot:
(Drafting until I can get around to this)

PR #5437, opened 2026-07-21
MatrixRTC: Support the optional leave_reason in rtc membership content

BillCarsonFr:
Adds support for `leave_reason` as per [MSC4143](https://github.com/matrix-org/matrix-spec-proposals/pull/4143)

Breaking change:
- Leave event has now a `leave_reason` optional key instead of enpty
- Previously for leaving the membership manager was just firing the scheduled leave, now we are cancelling it, then we send a proper leave event with the reason 

## Checklist

### DIFF
--- a/src/@types/event.ts
+++ b/src/@types/event.ts
@@ -58,6 +58,8 @@ import {
     type IRTCDeclineContent,
     type EncryptionKeysEventContent,
     type ICallNotifyContent,
+    type LeaveMembershipEventContent,
+    type LeaveReason,
 } from "../matrixrtc/types.ts";
 import { type M_POLL_END, type M_POLL_START, type PollEndEventContent, type PollStartEventContent } from "./polls.ts";
 import { type RtcMembershipData, type SessionMembershipData } from "../matrixrtc/membershipData/index.ts";
@@ -359,7 +361,7 @@ export interface TimelineEvents {
     [M_BEACON.name]: MBeaconEventContent;
     [M_POLL_START.name]: PollStartEventContent;
     [M_POLL_END.name]: PollEndEventContent;
-    [EventType.RTCMembership]: RtcMembershipData | { msc4354_sticky_key: string }; // An object containing just the sticky key is empty.
+    [EventType.RTCMembership]: RtcMembershipData | { msc4354_sticky_key: string; leave_reason?: LeaveReason }; // An object containing just the sticky key is empty.
 }
 
 /**
@@ -394,7 +396,11 @@ export interface StateEvents {
 
     // MSC3401
     [EventType.GroupCallPrefix]: IGroupCallRoomState;
-    [EventType.GroupCallMemberPrefix]: IGroupCallRoomMemberState | SessionMembershipData | EmptyObject;
+    [EventType.GroupCallMemberPrefix]:
+        | IGroupCallRoomMemberState
+        | SessionMembershipData
+        | LeaveMembershipEventContent
+        | EmptyObject;
     [EventType.RTCMembership]: RtcMembershipData | EmptyObject;
     // MSC3089
     [UNSTABLE_MSC3089_BRANCH.name]: MSC3089EventContent;
--- a/src/@types/event.ts
+++ b/src/@types/event.ts
@@ -361,7 +361,7 @@ export interface TimelineEvents {
     [M_BEACON.name]: MBeaconEventContent;
     [M_POLL_START.name]: PollStartEventContent;
     [M_POLL_END.name]: PollEndEventContent;
-    [EventType.RTCMembership]: RtcMembershipData | { msc4354_sticky_key: string; leave_reason?: LeaveReason }; // An object containing just the sticky key is empty.
+    [EventType.RTCMembership]: RtcMembershipData | { msc4354_sticky_key: string; leave_reason?: LeaveReason };
 }
 
 /**
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -35,6 +35,7 @@ import type {
     RTCCallIntent,
     Transport,
     SlotDescription,
+    LeaveReason,
 } from "./types.ts";
 import {
     MembershipManagerEvent,
@@ -566,7 +567,10 @@ export class MatrixRTCSession extends TypedEventEmitter<
      * A timeout can be provided so that there is a guarantee for the promise to resolve.
      * @returns Whether the membership update was attempted and did not time out.
      */
-    public async leaveRoomSession(timeout: number | undefined = undefined): Promise<boolean> {
+    public async leaveRoomSession(
+        timeout: number | undefined = undefined,
+        leaveReason: LeaveReason | undefined = undefined,
+    ): Promise<boolean> {
         if (!this.isJoined()) {
             this.logger.info(`Not joined to session in room ${this.roomSubset.roomId}: ignoring leave call`);
             return false;
@@ -576,7 +580,7 @@ export class MatrixRTCSession extends TypedEventEmitter<
 
         this.encryptionManager!.leave();
 
-        const leavePromise = this.membershipManager!.leave(timeout);
+        const leavePromise = this.membershipManager!.leave(timeout, leaveReason);
         this.emit(MatrixRTCSessionEvent.JoinStateChanged, false);
 
         return await leavePromise;
@@ -857,7 +861,7 @@ function quickFilterNonRelevantContents(content: IContent, logger: Logger): bool
     // Ignore sticky keys for the count
     const eventKeysCount = Object.keys(content).filter((k) => k !== "msc4354_sticky_key").length;
     // Don't even bother about empty events (saves us from costly type/"key in" checks in bigger rooms)
-    if (eventKeysCount === 0) return false;
+    if (eventKeysCount === 0 || (eventKeysCount === 1 && "leave_reason" in content)) return false;
 
     // We first decide if it's a MSC4143 event (per device state key)
     if (eventKeysCount > 1 && "application" in content) {
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -565,6 +565,9 @@ export class MatrixRTCSession extends TypedEventEmitter<
      * The membership update required to leave the session will retry if it fails.
      * Without network connection the promise will never resolve.
      * A timeout can be provided so that there is a guarantee for the promise to resolve.
+     *
+     * @param timeout
+     * @param leaveReason - The reason for the leave
      * @returns Whether the membership update was attempted and did not time out.
      */
     public async leaveRoomSession(
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -588,7 +588,7 @@ export class MatrixRTCSession extends TypedEventEmitter<
      * Without network connection the promise will never resolve.
      * A timeout can be provided so that there is a guarantee for the promise to resolve.
      *
-     * @param timeout
+     * @param timeout - Optional timeout in milliseconds that fires if the leave takes too long to complete.
      * @param leaveReason - The reason for the leave
      * @returns Whether the membership update was attempted and did not time out.
      */
--- a/src/matrixrtc/MembershipManager.ts
+++ b/src/matrixrtc/MembershipManager.ts
@@ -17,20 +17,28 @@ import { AbortError } from "p-retry";
 
 import { EventType, RelationType } from "../@types/event.ts";
 import { type ISendEventResponse, type SendDelayedEventResponse } from "../@types/requests.ts";
-import { type EmptyObject } from "../@types/common.ts";
 import type { MatrixClient } from "../client.ts";
 import { ConnectionError, HTTPError, MatrixError } from "../http-api/errors.ts";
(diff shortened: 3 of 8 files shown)


src/matrixrtc/MembershipManager.ts
Johennes: I think this is obsolete now?

src/matrixrtc/MembershipManager.ts
```diff
@@ -467,7 +500,7 @@ export class MembershipManager
                 }
                 // This is only a fallback in case we do not have working delayed events support.
                 // first we should try to just send the scheduled leave event
```
Johennes: I think this comment also needs updating.

src/matrixrtc/types.ts
```diff
@@ -225,3 +225,31 @@ export interface SlotDescription {
...
+export const LEAVE_REASON_DELAYED: LeaveReason = {
+    code: "delayed_leaved",
+    reason: "The user was removed due to inactivity (no heartbeat received).",
+};
+
+/** The member left intentionally (e.g. by hanging up a call) */
+export const LEAVE_REASON_HANGUP: LeaveReason = {
+    code: "user_leave",
+    reason: "The user intentionally left the rtc session.",
+};
```
Johennes, line 272: The MSC currently has these:

>     - `leave`: The member left intentionally (e.g. by hanging up a call).
>     - `delayed_leave`: The member left through a scheduled delayed leave event (see the
>       [lifecycle] section below).
>     - `slot_closed`: The member left because the slot was closed midway through the session.

src/matrixrtc/MembershipManagerActionScheduler.ts
```diff
@@ -132,7 +144,19 @@ export class ActionScheduler {
...
+                    ts: Date.now(),
+                    type: MembershipActionType.CancelledScheduledDelayedLeaveEvent,
+                },
+                {
+                    ts: Date.now(),
+                    type: MembershipActionType.SendLeaveEvent,
+                    data: leaveReason ? { leaveReason } : undefined,
+                },
+            ],
+        });
```
Johennes, line 157: I wonder if we should do it the other way around. Send the leave first, cancel the delayed event afterwards. That way, if your client dies inbetween the two requests, you don't leave a lingering joined member event.

Johennes (dismissed):
Overall looks good. Just some details.

src/matrixrtc/MembershipManagerActionScheduler.ts
BillCarsonFr: right, that would be more symetrical to the join process
send delayed, send join --- send leave, cancel delayed?

src/matrixrtc/types.ts
```diff
@@ -242,3 +242,37 @@ export interface SlotDescription {
...
+export const LEAVE_REASON_HANGUP: LeaveReason = {
+    code: "leave",
+    reason: "The user intentionally left the rtc session.",
+};
+
+/** The member left because the call slot was closed */
+export const LEAVE_REASON_SLOT_CLOSED: LeaveReason = {
+    code: "slot_closed",
+    reason: "The member left because the slot was closed midway through the session",
+};
```
dbkr, line 278: I appreciate I'm really here for web-level code review but these reasons don't really seem like they match the codes? `delayed_leave` sounds more like, "timeout" or something: I guess the 'delayed' part refers to the delayed event that got sent. The human readable text talks about timeouts though, is it actually user facing? If so, how is it translated? If it's really user-facing I'd expect something more like, "Connection to the user was lost".

The leave reason of "leave" seems tatutological. The const says hangup, so shouldn't the code be, "hangup"?

And, a much more petty quibble, but I'd just say, "the slot was closed while they were still joined" (isn't the slot closing the end of the session by definition?)

src/matrixrtc/MembershipManagerActionScheduler.ts
```diff
@@ -132,7 +144,19 @@ export class ActionScheduler {
     public initiateJoin(): void {
         this.wakeup?.({ replace: [{ ts: Date.now(), type: MembershipActionType.SendDelayedEvent }] });
     }
-    public initiateLeave(): void {
-        this.wakeup?.({ replace: [{ ts: Date.now(), type: MembershipActionType.SendScheduledDelayedLeaveEvent }] });
+    public initiateLeave(leaveReason?: LeaveReason): void {
+        this.wakeup?.({
```
dbkr: The ? implies that wakeup might not be there in which case this will fail silently which seems nonideal.

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -586,9 +587,15 @@ export class MatrixRTCSession extends TypedEventEmitter<
      * The membership update required to leave the session will retry if it fails.
      * Without network connection the promise will never resolve.
      * A timeout can be provided so that there is a guarantee for the promise to resolve.
+     *
+     * @param timeout
```
dbkr: I appreciate 'timeout' is fairly self-eveident but if it could get a token piece of doc just so it's not empty that would be good (after all, it's not a given that's in milliseconds).

dbkr (commented):
Mostly looks fine but you've said 'feat' in the PR title but 'task' in the label. It sounds like an enhancement (we use the labels instead of conventional commits because our PR titles go in the changelog).

src/matrixrtc/types.ts
BillCarsonFr: The code are taken directly from the MSC https://github.com/matrix-org/matrix-spec-proposals/blob/toger5/matrixRTC/proposals/4143-matrix-rtc.md 

I'll update and use the text in the MSC for the human readable.
The value is not made to be user facing nor translated, it is just human readable; like other human-readable reason in the spec it is to be used as a fallback if the client does not understand the code.

src/matrixrtc/MatrixRTCSession.ts
BillCarsonFr: I added something:
> Optional timeout in milliseconds that fires if the leave takes too long to complete.

src/matrixrtc/MembershipManagerActionScheduler.ts
BillCarsonFr: Not sure why there is a `?`, didn't want to change existing behavior. But I removed it has it seems not needed

dbkr (approved):

src/matrixrtc/types.ts
```diff
@@ -242,3 +242,37 @@ export interface SlotDescription {
...
+/** The member left through a scheduled delayed leave event */
+export const LEAVE_REASON_DELAYED: LeaveReason = {
+    code: "delayed_leave",
+    reason: "The member left through a scheduled delayed leave event.",
+};
+
+/** The member left intentionally (e.g. by hanging up a call) */
+export const LEAVE_REASON_HANGUP: LeaveReason = {
+    code: "leave",
+    reason: "The member left intentionally (e.g. by hanging up a call).",
```
robintown, line 271: It feels a little funny to have a user-facing string that's nonspecific about the type of application. I would rather omit the `reason`s in this file and (ideally) instead have them passed to the membership manager by the application's code.

src/@types/event.ts
```diff
@@ -361,7 +363,7 @@ export interface TimelineEvents {
     [M_BEACON.name]: MBeaconEventContent;
     [M_POLL_START.name]: PollStartEventContent;
     [M_POLL_END.name]: PollEndEventContent;
-    [EventType.RTCMembership]: RtcMembershipData | { msc4354_sticky_key: string }; // An object containing just the sticky key is empty.
+    [EventType.RTCMembership]: RtcMembershipData | { msc4354_sticky_key: string; leave_reason?: LeaveReason };
```
robintown, line 366: Shouldn't there be more properties in this variant, like `slot_id` and `member`?

src/@types/event.ts
```diff
@@ -396,7 +398,11 @@ export interface StateEvents {
 
     // MSC3401
     [EventType.GroupCallPrefix]: IGroupCallRoomState;
-    [EventType.GroupCallMemberPrefix]: IGroupCallRoomMemberState | SessionMembershipData | EmptyObject;
+    [EventType.GroupCallMemberPrefix]:
+        | IGroupCallRoomMemberState
+        | SessionMembershipData
+        | LeaveMembershipEventContent
+        | EmptyObject;
     [EventType.RTCMembership]: RtcMembershipData | EmptyObject;
```
robintown, line 406: By the way, should this state event type still exist? (If so, should it also get a `LeaveMembershipEventContent` variant?)

src/@types/event.ts
```diff
@@ -396,7 +398,11 @@ export interface StateEvents {
 
     // MSC3401
     [EventType.GroupCallPrefix]: IGroupCallRoomState;
-    [EventType.GroupCallMemberPrefix]: IGroupCallRoomMemberState | SessionMembershipData | EmptyObject;
+    [EventType.GroupCallMemberPrefix]:
+        | IGroupCallRoomMemberState
+        | SessionMembershipData
+        | LeaveMembershipEventContent
+        | EmptyObject;
```
robintown, line 405: The `EmptyObject` variant is redundant to the `LeaveMembershipEventContent` variant, since that interface also admits an empty object

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -887,7 +894,7 @@ function quickFilterNonRelevantContents(content: IContent, logger: Logger): bool
     // Ignore sticky keys for the count
     const eventKeysCount = Object.keys(content).filter((k) => k !== "msc4354_sticky_key").length;
     // Don't even bother about empty events (saves us from costly type/"key in" checks in bigger rooms)
-    if (eventKeysCount === 0) return false;
+    if (eventKeysCount === 0 || (eventKeysCount === 1 && "leave_reason" in content)) return false;
```
robintown, line 897:
```suggestion
    if (eventKeysCount === 0 || "leave_reason" in content) return false;
```

To account for the other keys that can be part of a `leave` membership (`slot_id` etc.)

src/matrixrtc/MembershipManagerActionScheduler.ts
```diff
@@ -14,6 +15,11 @@ export interface Action {
      * can also be thought of as the type of the action
      */
     type: MembershipActionType;
+
+    /**
+     * Additional parameters of the action
+     */
+    data?: MembershipActionData[MembershipActionType];
```
robintown, line 22:
```suggestion
    data: MembershipActionData[MembershipActionType];
```

Seems reasonable to require the data when it's specified for an action type

src/matrixrtc/MembershipManagerActionScheduler.ts
```diff
@@ -130,9 +139,21 @@ export class ActionScheduler {
...
+                {
+                    ts: Date.now(),
+                    type: MembershipActionType.SendLeaveEvent,
+                    data: leaveReason ? { leaveReason } : undefined,
+                },
+                {
+                    ts: Date.now(),
+                    type: MembershipActionType.CancelledScheduledDelayedLeaveEvent,
+                },
+            ],
```
robintown, line 156: I am glad to see that this is sending the leave event before canceling, since that's presumably a better way of dealing with the lack of atomicity in this operation. But ideally, the membership manager would only cancel the delayed event in step 2 if it actually *succeeds* at sending the leave event in step 1. Do you see any way to make the machine do that?

src/matrixrtc/MembershipManager.ts
```diff
@@ -1064,18 +1105,25 @@ export class StickyEventMembershipManager extends MembershipManager {
         super(joinConfig, room, clientWithSticky, sessionDescription, parentLogger);
     }
 
-    protected clientSendDelayedDisconnectMembership: () => Promise<SendDelayedEventResponse> = () =>
-        this.clientWithSticky._unstable_sendStickyDelayedEvent(
+    protected clientSendDelayedDisconnectMembership: () => Promise<SendDelayedEventResponse> = () => {
+        this.logger.debug(
+            `StickyMembershipManager send delayed disconnect membership event memberId: ${this.memberId}`,
```
robintown, line 1110:
```suggestion
            `send delayed disconnect membership event memberId: ${this.memberId}`,
```

(It's already prefixed)

src/@types/event.ts
Johennes: I think we've settled this in https://github.com/matrix-org/matrix-js-sdk/pull/5544 (which adds `slot_id`). `member` will have to be added by a future PR.

src/matrixrtc/types.ts
Johennes: Yeah, agreed. Have wrapped it around in c31cdc16f4f7be9421b3c4414e096272b3b76b20.

src/@types/event.ts
Johennes: Hm, this one is odd. It has the Matrix 2.0 type but it's a state event. We shouldn't have that combination at all and it doesn't appear to be used AFAICT. Have removed it in 4224775394fcbebf6530658349f5c919e9181406.

src/@types/event.ts
Johennes: `LeaveMembershipEventContent` no longer includes `EmptyObject` now after the fix from https://github.com/matrix-org/matrix-js-sdk/pull/5437#discussion_r3702488084.

src/matrixrtc/MembershipManagerActionScheduler.ts
Johennes: I _think_ f9f2fc2a581eae2b2c0b259bd320fc83475bb065 should achieve it.

PR #5440, opened 2026-07-21
Fix pre-initialization thread edits when target is known

basnijholt:
[MindRoom](https://github.com/mindroom-ai/mindroom) uses Matrix edits as a streaming transport for AI replies. A reply first appears as a short placeholder such as `Thinking…`, then gets edited repeatedly as the answer is generated.

The original version of this problem was reported in [element-hq/element-web#30617](https://github.com/element-hq/element-web/issues/30617) and led to #4980. That fix correctly defers edits when their target has not been loaded yet, so they can be replayed after thread initialization.

We later hit a narrower remaining race in the MindRoom Chat room overview ([downstream context](https://github.com/mindroom-ai/mindroom-chat/pull/193)). The placeholder event was already available to the UI, but initial thread pagination was still pending. Deferring every edit in that state meant the room preview could remain on `Thinking…` for the whole streamed response even though the SDK already had both the target and its edits.

This keeps the #4980 behavior when the target is unknown. When the target is already known, edits are aggregated immediately and the thread emits an update so previews can render the new content. Initial pagination can also replace the cached target with a fresh `MatrixEvent` instance for the same event ID, so relation collections now rebind to that new instance instead of keeping only the first object they saw.

The regression test holds pagination open, sends two edits to a known placeholder, verifies that the placeholder and thread update before pagination completes, then completes pagination with a fresh target object and verifies that the final edit is rebound to it. The existing target-not-yet-loaded test still covers the deferred replay path from #4980.

Validated with:

- `pnpm test`
- `pnpm build`
- `pnpm lint`

Notes: Fix thread previews that can remain on stale pre-initialization content after edits arrive.

## Checklist

### DIFF
--- a/spec/unit/models/thread.spec.ts
+++ b/spec/unit/models/thread.spec.ts
@@ -899,6 +899,86 @@ describe("Thread", () => {
                 expect(replaceIds[2]).toBe(edit3.getId());
             });
 
+            it("aggregates edits while initialization is pending when their target is already known", async () => {
+                const previousSupport = Thread.hasServerSideSupport;
+                Thread.setServerSideSupport(FeatureSupport.Stable);
+
+                const myUserId = "@alice:example.org";
+                const testClient = new TestClient(myUserId, "DEVICE", "ACCESS_TOKEN", undefined, {
+                    timelineSupport: false,
+                });
+                const client = testClient.client;
+                client.supportsThreads = vi.fn().mockReturnValue(true);
+
+                const room = new Room("!room:z", client, myUserId, {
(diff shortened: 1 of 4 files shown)


basnijholt:
The `Preview Changelog` failure is the required-label gate: this PR needs the maintainer-only `T-Defect` label. The fork `Tests`, `Static Analysis`, and Element Web E2E workflows also still need maintainer approval. Could a maintainer apply the label and approve those runs? Local `pnpm test`, `pnpm build`, and `pnpm lint` are green.

t3chguy (approved):
Looks sane, thanks

PR #5443, opened 2026-07-22
Allow passing a CA cert PEM when setting up Rust crypto

andybalaam:
Depends on https://github.com/matrix-org/matrix-sdk-crypto-wasm/pull/333/commits

## Checklist

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -215,8 +215,8 @@ importers:
         specifier: ^8.0.0
         version: 8.0.0
       '@matrix-org/matrix-sdk-crypto-wasm':
-        specifier: ^18.3.1
-        version: 18.3.1
+        specifier: ^18.4.0
+        version: 18.4.0
       another-json:
         specifier: ^0.2.0
         version: 0.2.0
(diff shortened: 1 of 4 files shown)


src/rust-crypto/index.ts
```diff
@@ -134,6 +142,7 @@ export async function initRustCrypto(args: {
         storeHandle,
         args.legacyCryptoStore,
         args.enableEncryptedStateEvents,
+        args.caCertsPem,
```
uhoreg, line 145: This function is getting a lot of parameters.  Given that most of them are just from `args` unchanged, I wonder if it would be better to just pass `args` itself to `initOlmMachine`, and let it extract the information that it needs.

uhoreg (approved):
LGTM other than a small suggestion

pnpm-lock.yaml
t3chguy, line 1: Something has gone wrong in your lockfile generation, you seem to have unlocked the dependency on the manager itself and rolled back a bunch of dependencies.

pnpm-workspace.yaml
t3chguy, line 1: Why are comments removed from this file and the big re-org?

t3chguy (changes requested):

src/rust-crypto/index.ts
andybalaam: Agree - I will do it as a separate PR.

pnpm-lock.yaml
andybalaam: Fixed and force-pushed as discussed over Matrix.

pnpm-workspace.yaml
andybalaam: Fixed and force-pushed as discussed over Matrix.

t3chguy (approved):

src/rust-crypto/index.ts
andybalaam: PR is https://github.com/matrix-org/matrix-js-sdk/pull/5450

PR #5446, opened 2026-07-23
Remove support for MSC3391 & MSC3852

t3chguy:
As they were both rejected & closed.

Synapse already removed support for MSC3852 which means it already did nothing: https://github.com/element-hq/synapse/pull/19430

Requires https://github.com/element-hq/element-web/pull/34400

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2351,7 +2351,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 
     /**
-     * Sets the users thamsc3391t the current user should ignore.
+     * Sets the users that the current user should ignore.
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -2373,7 +2351,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     }
 
     /**
-     * Sets the users that the current user should ignore.
+     * Sets the users thamsc3391t the current user should ignore.
```
florianduros: typo?

florianduros (approved):

src/client.ts
t3chguy:
```suggestion
     * Sets the users that the current user should ignore.
```

PR #5451, opened 2026-07-28
Allow providing the CA certs to MatrixClient.initRustCrypto

andybalaam:
~~Depends on https://github.com/matrix-org/matrix-js-sdk/pull/5450~~

## Checklist

- [x] New or updated `public`/`exported` symbols have accurate

### DIFF
--- a/spec/integ/crypto/rust-crypto.spec.ts
+++ b/spec/integ/crypto/rust-crypto.spec.ts
@@ -17,6 +17,7 @@ limitations under the License.
 import "fake-indexeddb/auto";
 import { IDBFactory } from "fake-indexeddb";
 import fetchMock from "@fetch-mock/vitest";
+import { OlmMachine } from "@matrix-org/matrix-sdk-crypto-wasm";
 
 import { createClient, IndexedDBCryptoStore } from "../../../src";
 import { populateStore } from "../../test-utils/test_indexeddb_cryptostore_dump";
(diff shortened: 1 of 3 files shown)


dbkr (approved):

spec/integ/crypto/rust-crypto.spec.ts
```diff
@@ -121,6 +122,62 @@ describe("MatrixClient.initRustCrypto", () => {
...
+        expect(initFromStore).toHaveBeenCalledWith(
+            expect.anything(),
+            expect.anything(),
+            expect.anything(),
+            expect.anything(),
+            CA_PEM,
+        );
+    });
+
+    const CA_PEM = `-----BEGIN CERTIFICATE-----
```
MidhunSureshR, line 125: Can we put this above the test where this is used?

MidhunSureshR (approved):

spec/integ/crypto/rust-crypto.spec.ts
andybalaam: Done in 95828c780

src/client.ts
```diff
@@ -1969,6 +1969,9 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
      * @param args.storagePassword - An alternative to `storageKey`. A password which will be used to derive a key to
      *    encrypt the store with. Deriving a key from a password is (deliberately) a slow operation, so prefer
      *    to pass a `storageKey` directly where possible.
+     * @param args.caCertsPem - Optional PEM-formatted string that provides CA certificates. These will be used to check
+     *    X.509 signatures on user identities. Any user identity that has a valid signature according to the supplied
+     *    CAs will be considered verified, without any manual verification taking place.
```
richvdh, line 1974: Since this is unspecced, we should warn clients about using it.

```suggestion
     * @param args.caCertsPem - Optional PEM-formatted string that provides CA certificates. These will be used to check
     *    X.509 signatures on user identities. Any user identity that has a valid signature according to the supplied
     *    CAs will be considered verified, without any manual verification taking place.
     *
     *    NOTE: this is an unspecified extension to Matrix. Applications should exercise caution when using it.
```

richvdh (approved):

PR #5460, opened 2026-08-03
Fix package failing to load in Node ESM due to directory import

### LINKED ISSUE #5459: 42.0.0 does not load in Node ESM: directory import in lib/oauth
Upgrading a plain Node ESM service from 41.9.0 to 42.0.0 broke it at import time, before any of my own code runs. I get the same crash on Node 22.18.0 and 24.18.1.

```sh
mkdir t && cd t && npm init -y && npm pkg set type=module
npm i matrix-js-sdk@42.0.0
echo 'import { createClient } from "matrix-js-sdk"; console.log(typeof createClient)' > i.mjs
node i.mjs
```

```
Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import '.../node_modules/matrix-js-sdk/lib/http-api' is not supported resolving ES modules imported from .../node_modules/matrix-js-sdk/lib/oauth/index.js
```

It comes down to two lines in the published package. `lib/oauth/index.js:23` has `import { HTTPError, Method } from "../http-api";` and `lib/oauth/authorize.js:21` has `import { Method } from "../http-api";`. Both point at a directory with no extension, which Node ESM will not resolve. The other 21 imports of that module elsewhere in `lib` use `../http-api/index.js` or `../http-api/index.ts` and are fine. Because `lib/matrix.js:43` does `export * from "./oauth/index.js"`, those two lines take the entire package down with them.

I hoped 42.1.0-rc.0 would have fixed it, but it fails the same way, and `src/oauth/index.ts:37` on develop still has the extensionless import. As far as I can tell it arrived with #5390, which added the oauth module.

I have pinned to 41.9.0 for now.

syphernl:
Fixes #5459.

`src/oauth/index.ts` and `src/oauth/authorize.ts` import `../http-api` without an extension. That emits a directory import into `lib`, and Node ESM does not resolve directories, so it throws `ERR_UNSUPPORTED_DIR_IMPORT`. Because `lib/matrix.js` re-exports `./oauth`, this takes down any plain `import { createClient } from "matrix-js-sdk"` in a Node ESM project on 42.0.0 and 42.1.0-rc.0.

Changed both to `../http-api/index.ts`, matching the other 13 http-api imports in `src`.

I verified the fix against the published 42.0.0 by rewriting those two lines in `lib` to the emitted form (`../http-api/index.js`), after which the package imports cleanly on Node 22 and 24.

Notes: Fix matrix-js-sdk failing to import in Node ESM environments

### DIFF
--- a/.oxlintrc.jsonc
+++ b/.oxlintrc.jsonc
@@ -165,6 +165,9 @@
                 "no-console": "error",
                 "no-extra-boolean-cast": "error",
                 "unicorn/prefer-node-protocol": "off",
+
+                // Require extensions on imports to maintain Node compatibility and avoid issues with ESM resolution
+                "import/extensions": ["error", "ignorePackages"],
             },
         },
         {
--- a/src/oauth/authorize.ts
+++ b/src/oauth/authorize.ts
@@ -24,7 +24,7 @@ import {
     hasRequiredStringProperty,
     isRecord,
 } from "../@types/type-guards.ts";
-import { Method } from "../http-api";
+import { Method } from "../http-api/index.ts";
 import { OAuthGrantType } from "./register.ts";
 import { sleep } from "../utils.ts";
 
--- a/src/oauth/index.ts
+++ b/src/oauth/index.ts
@@ -34,7 +34,7 @@ import {
 } from "./register.ts";
 import { encodeUnpaddedBase64Url } from "../base64.ts";
 import { sha256 } from "../digest.ts";
-import { HTTPError, Method } from "../http-api";
+import { HTTPError, Method } from "../http-api/index.ts";
 import { logger } from "../logger.ts";
 import { OAuth2Error } from "./error.ts";
 import { secureRandomString } from "../randomstring.ts";


t3chguy:
If this is something we need to worry about then it ought to have a lint rule

t3chguy:
@syphernl could you add this patch to your PR, I tested locally and it raises both of the issues which you are fixing (and no more).

```patch
Index: .oxlintrc.jsonc
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/.oxlintrc.jsonc b/.oxlintrc.jsonc
--- a/.oxlintrc.jsonc	(revision Staged)
+++ b/.oxlintrc.jsonc	(date 1785752553830)
@@ -165,6 +165,9 @@
                 "no-console": "error",
                 "no-extra-boolean-cast": "error",
                 "unicorn/prefer-node-protocol": "off",
+
+                // Require extensions on imports to maintain Node compatibility and avoid issues with ESM resolution
+                "import/extensions": ["error", "ignorePackages"],
             },
         },
         {

syphernl:
@t3chguy The linting rule has been added.

t3chguy (approved):
Thanks!

PR #5470, opened 2026-08-05
rtc: Cache RTC discovery in the MatrixClient

BillCarsonFr:
Allow the matrix-js-sdk to cache the `rtc/transport` discovery result as it is doing with similar other server side calls (well-known or capabilities)

This would allow the widget driver (in Element-web) to use this cached value instead of fetching every time https://github.com/element-hq/element-web/pull/34393

EW has added it's own caching in CallStore, but I think this should be done in the js-sdk.

This is best reviewed commit by commit?
- 510824ad19c30720c3f1b0cd5622e0ed7974cf18 introduces the cachedValue generic mecanism and use it for transport discovery
- b96554584bc48cd8e3069695c72b509f85033d19 Refactors the client to use the same mecanism for well-known (there is a behavior changes, it is getting retry on transient/quota errors)

## Checklist

### DIFF
--- a/src/clientCachedValue.ts
+++ b/src/clientCachedValue.ts
@@ -1,0 +1,188 @@
+/*
+Copyright 2026 The Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { type MatrixClient } from "./client.ts";
+import { calculateRetryBackoff } from "./http-api/index.ts";
+import { type Logger } from "./logger.ts";
+import { sleep } from "./utils.ts";
+
+/**
+ * Defines a generic mechanism to fetch and cache values for the client.
+ */
+export abstract class ClientCachedValue<ValueType> {
+    protected cached?: ValueType;
+    protected fetchPromise?: Promise<ValueType>;
+    protected ttlTimeoutHandle?: ReturnType<typeof setTimeout>;
+
+    private isStopped = false;
+    private ttlMillis?: number;
+
+    private readonly logger: Logger;
+    /**
+     * Build a generic mechanism allowing to fetch and cache a given value from the homeserver.
+     * @param name - The name of the cached value (for tracing/logs).
+     * @param client - The matrix client
+     * @param ttlMillis - The time-to-live before the value get refreshed. If undefined use refresh() to clear cache.
+     * @param rootLogger
+     */
+    public constructor(
+        protected readonly name: string,
+        protected readonly client: MatrixClient,
+        ttlMillis: number | undefined,
+        rootLogger: Logger,
+    ) {
+        this.ttlMillis = ttlMillis;
+        this.logger = rootLogger.getChild(`ClientCachedValue<${name}>`);
+    }
+
+    /**
+     * Gets the cached value if any.
+     */
+    public get(): ValueType | undefined {
+        return this.cached;
+    }
+
+    /**
+     * Ensures that the value was fetched once before getting the actual cached value.
+     */
+    public async wait(): Promise<ValueType | undefined> {
+        try {
+            await this.doFetch();
+        } catch (error) {
+            // Ignore errors, as wait should not throw
+        }
+        return this.cached;
+    }
+
+    /**
+     * Call this as early as possible and when the client is already capable of fetching the value.
+     */
+    public start(): void {
+        this.isStopped = false;
+        // Request a fetch as soon as possible
+        void this.doFetch();
+    }
+
+    /**
+     * Set the time-to-live for the cached value.
+     * @param TTL The time-to-live in milliseconds.
+     */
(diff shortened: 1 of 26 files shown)


src/clientCachedValue.ts
```diff
@@ -0,0 +1,191 @@
...
+
+import { type MatrixClient } from "./client.ts";
+import { calculateRetryBackoff } from "./http-api/index.ts";
+import { type Logger } from "./logger.ts";
+import { sleep } from "./utils.ts";
+
+/**
+ * Defines a generic mechanism to fetch and cache values for the client.
+ */
+export abstract class ClientCachedValue<ValueType> {
```
dbkr: This seems like more than just caching, it's actively re-fetching on a timer.

Also, I think we are trying to move away from inheritance as a way of getting functionality which I think is considered a bit of an anti-pattern in favour of aggregation, ie. instantiating an object that does the thing.

src/clientCachedValue.ts
```diff
@@ -0,0 +1,191 @@
...
+            this.logger.debug("Cached value fetch on start did fail", err);
+        });
+    }
+
+    /**
+     * Set the time-to-live for the cached value.
+     * @param TTL The time-to-live in milliseconds.
+     */
+    public setTTLMillis(TTL: number): void {
+        this.ttlMillis = TTL;
```
dbkr: wouldn't we expect this to update the timeout?

src/pollingCachedValue.ts
```diff
@@ -0,0 +1,191 @@
...
+     * Determines whether a non-transient error should be cached.
+     * @param error The error to evaluate.
+     * @protected
+     */
+    protected shouldCacheError(error: any): boolean {
+        // By default retry to fetch the value on error
+        return false;
+    }
+
+    private async doFetch(force: boolean = false): Promise<void> {
```
dbkr, line 122: docs please, it's not clear what 'force' means here

src/clientCachedValue.ts
```diff
@@ -0,0 +1,191 @@
...
+    private async doFetch(force: boolean = false): Promise<void> {
+        if (this.fetchPromise && !force) {
+            await this.fetchPromise;
+            return;
+        }
+
+        this.fetchPromise = this.fetchWithRetryPolicy();
+
+        try {
+            const value = await this.fetchPromise;
```
dbkr: why use the value variable rather than just assigning to this.cached?

src/pollingCachedValue.ts
```diff
@@ -0,0 +1,191 @@
...
+            if (this.isStopped) {
+                this.fetchPromise = undefined;
+                return;
+            }
+            this.valueCached(value);
+            // The value is cached only for a given time.
+            if (this.ttlMillis) {
+                this.ttlTimeoutHandle = setTimeout(() => {
+                    this.fetchPromise = undefined;
+                    void this.doFetch();
```
dbkr, line 142: This in particular feels like a weird hybrid between periodic updating where you would probably try to refresh the value in advance of it expiring and a cache that expires but is then just repopulated the next time it's needed.

src/clientCachedValue.ts
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /**
+     * Ensures that the value was fetched once before getting the actual cached value.
+     */
+    public async wait(): Promise<ValueType | undefined> {
+        try {
+            await this.doFetch();
+        } catch (error) {
+            // Ignore errors, as wait should not throw
```
dbkr: shouldn't it? what if the value can't be fetched? This feels like something for the function's tsdoc.

src/pollingCachedValue.ts
BillCarsonFr: Agree, I tried to keep it mostly to accomodate the pre-existing well-known behavior that was doing that, so I can refactor both value to use the new mecanism

src/clientCachedValue.ts
BillCarsonFr: True, I renamed to `ClientPollingCachedValue`, it is better but naming is hard.

I also did the refactoring to avoid inheritence, by having a concrete class doing the mecanism and having a factory with configuration option.
This is easier to review commit by commit I think, I am adding the commits for convinience:
68443535f2f3555b7aba4f2fb91afb0e0883f2b8 Do the renaming
e08796b87837af63786b997f5391fe7dda09b27a The factory refactoring

src/clientCachedValue.ts
BillCarsonFr: That's true, but given this was always set once and before the start, I moved it as a parameter of the start to avoid this caveat 0e6487c8cd3132dabc24e9342bae4193c9d7bd7c

src/pollingCachedValue.ts
BillCarsonFr: updated f70d8cab2dca52e10711e9b471b86a5a0d89e5c8

src/clientCachedValue.ts
BillCarsonFr: done

src/clientCachedValue.ts
BillCarsonFr: You are right, I changed it. f20518b95a273859431754abad19f5e307460058

dbkr (approved):
Thanks, yeah - it's the awkward situation of two things that share very similar logic but not the same.

BillCarsonFr:
The upstream fails because of a test I fixed in this PR https://github.com/element-hq/element-web/pull/34581/changes

The test was accessing a non-public API?
` client["clientWellKnown"] = {`

Replaced by a spyOn `vi.spyOn(client, "getClientWellKnown").mockResolvedValue(wn);`

FWIW created an upstream PR resolve the problem https://github.com/element-hq/element-web/pull/34618

PR #5472, opened 2026-08-06
Validate that an error from OAuth token refresh looks like an OAuth error before treating as logout

hughns:
The spec says for handling errors when making a (https://spec.matrix.org/v1.19/client-server-api/#refresh-token-grant request:

> The client MUST handle access token refresh failures as follows:
>
> - If the refresh fails due to network issues or a 5xx HTTP status code from the server, the client should retry the request with the old refresh token later.
> - If the refresh fails due to a 4xx HTTP status code from the server, the client should consider the session logged out.

This is in the context that the token endpoint is specified to return well-formed RFC 6749 responses.

This PR validates that the 4xx response looks like an OAuth error prior to treating it as the a session logout which reverts the logic to be closer to what it was prior to https://github.com/matrix-org/matrix-js-sdk/pull/5390.

[For interest, the [equivalent logic in matrix-rust-sdk](https://github.com/matrix-org/matrix-rust-sdk/blob/85873619dabaa21cc40444ef58053704ced695f9/crates/matrix-sdk/src/client/futures.rs#L128-L149) has an even stricter check and only treats as a logout if the OAuth error code is `invalid_grant`. MSC4526 proposes to standardise on similar behaviour]

## Checklist

### DIFF
--- a/src/oauth/tokenRefresher.ts
+++ b/src/oauth/tokenRefresher.ts
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 import { type AccessTokens, HTTPError, type TokenRefreshFunction, TokenRefreshLogoutError } from "../http-api/index.ts";
+import { OAuth2HTTPError } from "./error.ts";
 import { type OAuth2 } from "./index.ts";
 
 /**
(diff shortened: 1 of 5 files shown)


dbkr:
@t3chguy might want a look too given he worked on the original PR, but looks okay to me.

src/oauth/tokenRefresher.ts
```diff
@@ -54,8 +55,14 @@ export class TokenRefresher {
     };
 
     private shouldLogoutOnError(error: HTTPError): boolean {
-        // As per https://spec.matrix.org/v1.18/client-server-api/#refresh-token-grant
-        return typeof error.httpStatus === "number" && error.httpStatus < 500 && error.httpStatus >= 400;
+        // Treat as logout as per https://spec.matrix.org/v1.18/client-server-api/#refresh-token-grant
+        // after making sure it is an RFC 6749 section 5.2 error response
+        return (
+            error instanceof OAuth2HTTPError &&
```
t3chguy, line 61: instanceof can be a bit of a footgun in environments where it is possible to end up with 2 instances of a given library

t3chguy (approved):

PR #5473, opened 2026-08-07
Don't treat Matrix-shaped errors as OAuth token refresh grant logouts

hughns:
https://github.com/matrix-org/matrix-spec/pull/2432 confirmed that the expected structure of an error from the token endpoint is OAuth 2.0 and not Matrix.

This PR builds on https://github.com/matrix-org/matrix-js-sdk/pull/5472 by checking if an error response looks like a Matrix error (using `errcode`) prior to checking if it is an OAuth 2.0 error response. If it looks like a Matrix error then we don't treat it as an invalid refresh token.

Notes:

- A future PR could apply this error parsing logic to other OAuth requests.
- MSC4526 proposes a revised approach for handling token refresh errors that would be more robust (and align with this PR).

## Checklist

### DIFF
--- a/.github/actions/sign-release-tarball/action.yml
+++ b/.github/actions/sign-release-tarball/action.yml
@@ -22,7 +22,7 @@ runs:
 
         - name: Upload tarball signature
           if: ${{ inputs.upload-url }}
-          uses: shogo82148/actions-upload-release-asset@394b3c11c3cfc038b5396ad265c074065cf875c3 # v1
+          uses: shogo82148/actions-upload-release-asset@d0cb37bb764791e72e64e643cd7056f74f1e77d7 # v1
           with:
               upload_url: ${{ inputs.upload-url }}
               asset_path: ${{ env.VERSION }}.tar.gz.asc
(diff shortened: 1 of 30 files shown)


t3chguy (approved):

richvdh:
@hughns is there a github issue (in either matrix-js-sdk or element-web) that describes the problem this is seeking to resolve, along with some of the history of how it got introduced and the user-visible symptoms?

If not, please could there be? :innocent:

richvdh:
AIUI:

* After https://github.com/element-hq/element-web/pull/34026 but before #5472: any 4xx response from the server or proxy in response to a token refresh will be treated as a logout.
* After #5472 but before this PR: any 4xx token-refresh response with an `error` field in the body will be treated as a logout.
* After this PR: a 4xx token-refresh response with an `error` field in the body will be treated as a logout *unless* it also has an `errcode` in the body.

hughns:
> @hughns is there a github issue (in either matrix-js-sdk or element-web) that describes the problem this is seeking to resolve, along with some of the history of how it got introduced and the user-visible symptoms?
> 
> If not, please could there be? 😇

I've added https://github.com/matrix-org/matrix-js-sdk/issues/5494

PR #832, opened 2019-02-01
Cross Signing Support

dbkr:
As per https://github.com/matrix-org/matrix-doc/pull/1756

There are also some things in here which will need speccing in addition to the MSC above (as it currently stands), most notably, putting the SSK & USK in the backup auth_data.

PR #5485, opened 2026-08-14
make sure `setDefaultKeyId` resolves even if no change is made

### LINKED ISSUE #5474: RustCrypto.resetEncryption() never resolves when secret storage account data is already empty
`CryptoApi.resetEncryption()` (rust crypto) hangs forever, without any error, when the account's secret storage account data has already been wiped to empty objects. This is exactly the state that a previously interrupted `resetEncryption()` call leaves behind, so one failed reset attempt permanently bricks all subsequent reset attempts for the account. The user-visible consequence is Element's "Reset your cryptographic identity" dialog spinning on "Reset in progress..." indefinitely (a companion issue in element-web will follow).

Line numbers below are for v42.1.0 (the version bundled in Element Web/Desktop 1.12.25); the relevant logic is unchanged on current `develop`.

### Root cause
1. `resetEncryption()` awaits `deleteSecretStorage()` (rust-crypto.ts L1601), which ends with `secretStorage.setDefaultKeyId(null)` (rust-crypto.ts L1630).
2. `ServerSideSecretStorageImpl.setDefaultKeyId()` (secret-storage.ts L373-L402) returns a manually constructed Promise that is resolved only from a `ClientEvent.AccountData` listener, once the `m.secret_storage.default_key` echo arrives over `/sync`. The fulfillment of the inner `setAccountData()` call is ignored; only `.catch()` is wired (secret-storage.ts L397-L400).
3. Since #4695, `MatrixClient.setAccountData()` short-circuits with no HTTP request and no `ClientEvent.AccountData` emission when the new content deep-equals the locally synced content (client.ts L2249-L2250).
4. When `m.secret_storage.default_key` already holds `{}`, the inner call in step 2 no-ops, no echo ever arrives, the listener never fires, and the Promise never settles. `resetEncryption()` awaits it forever; `bootstrapCrossSigning()` and the new key backup (rust-crypto.ts L1604-L1610) are never reached.

The four `store(name, null)` calls in `deleteSecretStorage()` (rust-crypto.ts L1621-L1624) dedup-skip the same way in this state, because `store(name, null)` writes `{}` via `setAccountData(name, {})` (secret-storage.ts L513-L518). That is why an affected client sends no requests at all after the (harmless, caught) dehydrated-device DELETE and `room_keys/version` GET.

There is a second flavor of the same hang: if the local sync store disagrees with the server (server already has `{}`, but the local store lacks the event), the PUT is issued, but the server sends no `/sync` account-data update for a no-op PUT, so `setAccountData()` itself blocks forever at `await updatedResolvers.promise` (client.ts L2271).

### How accounts reach this state
`resetEncryption()` performs the destructive account-data wipes before the cross-signing key upload. If the flow is interrupted at the upload step (auth or MAS approval cancelled, the return flow failing as in element-hq/matrix-authentication-service#5450, or the app being closed), the account is left with secret storage account data wiped to `{}`, no key backup, and the old cross-signing keys. Every later `resetEncryption()` call then deadlocks as above.

Verified live on a matrix.org account with Element Web and Element Desktop 1.12.25 (js-sdk 42.1.0); both reproduce identically.

### Steps to reproduce
1. On an account whose `m.cross_signing.*`, `m.megolm_backup.v1` and `m.secret_storage.default_key` account data are all `{}` (starting a reset and cancelling it at the auth prompt produces this state), call `client.getCrypto().resetEncryption(cb)`.
2. Observe `DELETE .../dehydrated_device` (404), `GET .../room_keys/version` (404), then no further requests. The returned Promise never settles and `cb` is never invoked.

### Expected
`resetEncryption()` either completes (idempotently) or rejects. A reset should be re-runnable after interruption, since interruption at the auth step is common.

### Suggested fix
`setDefaultKeyId()` could resolve on fulfillment of the inner `setAccountData()` call instead of relying exclusively on the listener; since #4695, `setAccountData()` already awaits the remote echo itself, so the listener duplicates that logic and misbehaves exactly in the dedup case. Alternatively, short-circuit `setDefaultKeyId(null)` when the current content already matches.

### Workaround for affected users
From the client, write non-empty content into `m.secret_storage.default_key` and re-run the reset:

    client.setAccountData("m.secret_storage.default_key", { key: "dummy" })

### Code references (v42.1.0)
- resetEncryption: https://github.com/matrix-org/matrix-js-sdk/blob/v42.1.0/src/rust-crypto/rust-crypto.ts#L1591-L1611
- deleteSecretStorage: https://github.com/matrix-org/matrix-js-sdk/blob/v42.1.0/src/rust-crypto/rust-crypto.ts#L1619-L1631
- ServerSideSecretStorageImpl.setDefaultKeyId: https://github.com/matrix-org/matrix-js-sdk/blob/v42.1.0/src/secret-storage.ts#L373-L402
- SecretStorage.store(name, null) writing {}: https://github.com/matrix-org/matrix-js-sdk/blob/v42.1.0/src/secret-storage.ts#L513-L518
- MatrixClient.setAccountData dedup short-circuit: https://github.com/matrix-org/matrix-js-sdk/blob/v42.1.0/src/client.ts#L2249-L2250
- MatrixClient.setAccountData echo wait: https://github.com/matrix-org/matrix-js-sdk/blob/v42.1.0/src/client.ts#L2271

uhoreg:
fixes https://github.com/matrix-org/matrix-js-sdk/issues/5474

Make `setDefaultKeyId` just `setAccountData` and return, rather than listening for the value to come back down the sync.  `setAccountData` already does that now, when it actually tries to change the value, and does not make an HTTP call if we try to set the account data value to the current value, which means that `setDefaultKeyId` would never get notified of the value being changed (since it doesn't get changed), and so would never return.

## Checklist

### DIFF
--- a/spec/unit/secret-storage.spec.ts
+++ b/spec/unit/secret-storage.spec.ts
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import fetchMock from "@fetch-mock/vitest";
 import { type Mocked } from "vitest";
 
 import {
@@ -29,7 +30,9 @@ import {
 } from "../../src/secret-storage";
 import { secureRandomString } from "../../src/randomstring";
 import { type SecretInfo } from "../../src/secret-storage.ts";
-import { type AccountDataEvents, ClientEvent, MatrixEvent, TypedEventEmitter } from "../../src";
+import { type AccountDataEvents, createClient, TypedEventEmitter } from "../../src";
+import { SyncResponder } from "../test-utils/SyncResponder.ts";
+import { mockInitialApiRequests } from "../test-utils/mockEndpoints.ts";
 
 declare module "../../src/@types/event" {
     interface SecretStorageAccountDataEvents {
@@ -287,12 +290,9 @@ describe("ServerSideSecretStorageImpl", function () {
     describe("setDefaultKeyId", function () {
         let secretStorage: ServerSideSecretStorage;
         let accountDataAdapter: Mocked<AccountDataClient>;
-        let accountDataPromise: PromiseWithResolvers<void>;
         beforeEach(() => {
             accountDataAdapter = mockAccountDataClient();
-            accountDataPromise = Promise.withResolvers();
             accountDataAdapter.setAccountData.mockImplementation(() => {
-                accountDataPromise.resolve();
                 return Promise.resolve({});
             });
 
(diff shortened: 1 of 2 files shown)


florianduros (approved):

richvdh:
For the archaeological record:

`setDefaultKeyId` has had a listener on `ClientEvent.AccountData` to wait for the remote echo ever since 4S was introduced in https://github.com/matrix-org/matrix-js-sdk/pull/832. That was fine at that point, because `MatrixClient.setAccountData` didn't have any logic to filter out no-op changes, so there would reliably be a remote echo.

Later, https://github.com/matrix-org/matrix-js-sdk/pull/4695 changed `MatrixClient.setAccountData` so that (a) it did nothing if the change was a no-op; (b) it had its own listener on `ClientEvent.AccountData` to wait for the remote echo. [Aside: the comments added in that change also assert that there will be no remote echo in the case of a no-op change of account data; it is unclear to me now if that is accurate.] That change made the listener in `setDefaultKeyId` both redundant and unreliable (since in the case of changes that `setAccountData` deemed a no-op, there would be no remote echo). 

Nevertheless, `setDefaultKeyId` continued to work ok in the specific case of *removing* the default key ID (i.e. setting the account data to `{}`), because, until https://github.com/matrix-org/matrix-js-sdk/pull/5446/changes#diff-7237d115811d917dbc234402ea3e85ced2a568c07005f45344d3a513f30cf68cL295-L301, an account data value of `{}` was special-cased, so repeated calls to `MatrixClient.setAccountData` with a value of `{}` would make an API hit each time.

In other words, between #4695 and #5446, it was working ok, but for all the wrong reasons. 

This PR completes the work that should have been done in #4695, by removing the now-redundant listener from `setDefaultKeyId`.

PR #5486, opened 2026-08-15
Add support for key rotation participant limit

toger5:
Part of the load test solutions (scale to 200users)

This needs to be used in element call. see: https://github.com/element-hq/element-call/pull/4172

The idea:
Currently we rotate the key each time someone joins or leaves:
"rotate the key": Generate a new key -> send it to everyone else in the call (this is done by each device as we do per sender keys)
This results in a ~ N^2 amount of keys being distributed.
This PR allows configuring a `key_rotation_participant_limit` at which clients will stop doing rotations. Instead
 - a new joiner will get the current in use key from all participants. (N keys per join, 0 keys per leave)
 
Security:
 - This is a tradeoff trading post compromising and forward secrecy for perfromance. This is why it is configurable in the js-sdk. Each call might have different requirements.
 - Up to 30 participants there is no perceived difference between the two approaches.
 - In larger calls it is common for most participants of a room to join. Things that ppl say target a wider audience. The full awareness who you are addressing (who is still in the call) is less present for users publishing media. In small calls chances are high that one says things just addressing the 3-10 users in the room. It is more imporatnat that leakead media cannot be decrypted.
 - This still is superior to a shared secret as we still only share the keys with the users that have been joined at least once. (a shared secret also would give access to the key for all room members independent of them joining the call)
 - **Edit** this PR was update to now [default to always rotate keys](https://github.com/matrix-org/matrix-js-sdk/pull/5486/commits/1d6dc4b753ad815d1362323c6ab4b9d04eaa8460) (also in large calls). This makes the key rotation limit an opt in property client publishers can activate by setting the introduced config value.
 
It is configurable via the join config.
To be discussed: should be configurable via slot? No. Not in the initial impl.
## Checklist

### DIFF
--- a/src/matrixrtc/EncryptionManager.ts
+++ b/src/matrixrtc/EncryptionManager.ts
@@ -16,6 +16,12 @@ export function getEncryptionKeyMapKey(membership: CallMembershipIdentityParts):
  */
 export interface IEncryptionManager {
     /**
+     * Whether the session currently has too many participants for the media key to be rotated.
+     * @see EncryptionConfig.keyRotationParticipantLimit
+     */
+    readonly isKeyRotationSuppressed: boolean;
+
+    /**
      * Joins the encryption manager with the provided configuration.
      *
      * @param joinConfig - The configuration for joining encryption, or undefined
--- a/src/matrixrtc/EncryptionManager.ts
+++ b/src/matrixrtc/EncryptionManager.ts
@@ -16,7 +16,7 @@ export function getEncryptionKeyMapKey(membership: CallMembershipIdentityParts):
  */
 export interface IEncryptionManager {
     /**
-     * Whether the session currently has too many participants for the media key to be rotated.
+     * Whether the key rotation is currently halted.
      * @see EncryptionConfig.keyRotationParticipantLimit
      */
     readonly isKeyRotationSuppressed: boolean;
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -835,6 +835,8 @@ export class MatrixRTCSession extends TypedEventEmitter<
         // Clear the flag.
         this.membershipNeedsRecalculation = false;
         const oldMemberships = this.memberships;
+        // Needs to be computed before `this.memberships` is updated below, since it is derived from it.
+        const wasKeyRotationSuppressed = this.isKeyRotationSuppressed;
 
         this.memberships = await MatrixRTCSession.sessionMembershipsForSlot(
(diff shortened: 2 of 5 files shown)


src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -34,6 +34,17 @@ import {
 import { OutdatedKeyFilter } from "./utils.ts";
 import { computeRtcIdentityRaw } from "./membershipData/rtc.ts";
 
+/**
+ * Default for {@link EncryptionConfig.keyRotationParticipantLimit}.
+ *
+ * Setting this to undefined implies that we do not have a limit and full rotations are always done.
+ * This is the most secure and least performant option.
+ * It is highly recommended to set this to < 50 for acutal client deployments.
```
Johennes:
```suggestion
 * It is highly recommended to set this to < 50 for actual client deployments.
```

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -204,6 +207,17 @@ export interface EncryptionConfig {
...
     keyRotationGracePeriodMs?: number;
 
+    /**
+     * The number of participants in the session at which the media key will no longer be rotated.
+     *
+     * Rotating a key requires sending it to every participant device, so in large sessions the cost of rotating
+     * on every join/leave becomes prohibitive. At this limit the current key is kept and
+     * distributed to new joiners. No new keys are generated for joiners/leavers.
+     *
+     * Defaults to 30.
```
Johennes: It looks like `DEFAULT_KEY_ROTATION_PARTICIPANT_LIMIT` defaults to `undefined`?

src/matrixrtc/EncryptionManager.ts
```diff
@@ -15,6 +15,12 @@ export function getEncryptionKeyMapKey(membership: CallMembershipIdentityParts):
  * @internal
  */
 export interface IEncryptionManager {
+    /**
+     * Whether the session currently has too many participants for the media key to be rotated.
+     * @see EncryptionConfig.keyRotationParticipantLimit
+     */
```
Johennes, line 21: Nit: Several comments in the diff are worded like this and it seems a little duplicative to me. I think it might be more helpful if we documented what something is (e.g. "Whether key rotation is currently halted or not") rather than why it might have been set ("the session currently has too many participants"). Then we can put the documentation for why the property is enabled in just one place (e.g. when the property is set in the `isKeyRotationSuppressed` getter).

Johennes (approved):
This looks sane to me. Just some nits.

PR #5489, opened 2026-08-19
OAuth refresh token logout handling as per MSC4526

hughns:
To-do:

- [ ] Is backwards compatibility important for this? (i.e. does the behaviour need to be opt-in/out?)

## Checklist


t3chguy:
> (i.e. does the behaviour need to be opt-in/out?)

I believe it ought to be opt-in given it is an unstable MSC implementation, which also means it'd need a labs flag to be enabled in EW

PR #5502, opened 2026-08-26
Fix one-time keys being re-uploaded on every to-device/device-list sync with matrix-sdk-crypto-wasm ≥ 18.5.0

### LINKED ISSUE #5501: With crypto-wasm ≥ 18.5.0, RustCrypto asserts a one-time-key count of zero on every to-device / device-list sync, republishing 50 OTKs per sync and breaking new Olm sessions
Hello everyone. Me and my teammates are working on the development of a Matrix client, and we were experiencing a serious bug these last days. After investigating it appears to be a regression caused by the upgrade of `matrix-sdk-crypto-wasm` in version `18.5.0` (but the fix should be here). Here is a bug report generated with AI:

## Bug report

`RustCrypto` fans a single `/sync` response out into up to three `OlmMachine.receiveSyncChanges()` calls, and two of them pass an **empty** one-time-key count map because the caller has no count to give. Since `@matrix-org/matrix-sdk-crypto-wasm` 18.5.0, an empty map is (correctly, per the sync v2 spec) interpreted as "zero one-time keys on the server", so the account resets its uploaded-key count and generates a fresh batch of 50 keys on almost every sync. The server-side stock grows without bound (we measured 20,040 keys on one device), vodozemac evicts the private halves of older batches, and peers who claim one of those older keys can never establish an Olm session with the device.

`matrix-js-sdk` 42.x declares `^18.4.0`, so a fresh install or a routine lockfile refresh resolves to 18.5.0/18.6.0 and picks this up. Element Web's lockfile currently pins 18.4.0, which is the only reason it is not affected today.

Line numbers below are for v42.2.0; the code is unchanged on current `develop`.

### Steps to reproduce

1. `matrix-js-sdk` 42.2.0 with `@matrix-org/matrix-sdk-crypto-wasm` 18.5.0 or 18.6.0.
2. Sign in and let the client sync, doing nothing.
3. Watch `POST /_matrix/client/v3/keys/upload` and the server's `device_one_time_keys_count`.

**Actual:** a batch of exactly 50 one-time keys is uploaded on (almost) every sync that carries `to_device` events or `device_lists` changes. Fresh key ids each time, so uploads succeed silently. Measured over 20 ordinary syncs each carrying to-device events and device-list changes, then a true count of ~130k:

| crypto-wasm | `/keys/upload` calls | keys published |
|---|---|---|
| 18.4.0 | 0 | 0 |
| 18.5.0 | 20 | 1000 |
| 18.6.0 | 20 | 1000 |

**Expected:** one batch at startup, then top-ups only as keys are claimed.

### Root cause

1. `receiveSyncChanges` defaults the count map to empty
   ([rust-crypto.ts#L1701-L1703](https://github.com/matrix-org/matrix-js-sdk/blob/v42.2.0/src/rust-crypto/rust-crypto.ts#L1701-L1703)):
   ```ts
   private async receiveSyncChanges({
       events,
       oneTimeKeysCounts = new Map<string, number>(),
   ```
2. Two of its three callers rely on that default:
   `preprocessToDeviceMessages` ([#L1725-L1728](https://github.com/matrix-org/matrix-js-sdk/blob/v42.2.0/src/rust-crypto/rust-crypto.ts#L1725-L1728) — the comment says *"We have no info on device-list changes, one-time-keys, or fallback keys, so just pass empty data"*) and `processDeviceLists` ([#L1836-L1841](https://github.com/matrix-org/matrix-js-sdk/blob/v42.2.0/src/rust-crypto/rust-crypto.ts#L1836-L1841)). Only `processKeyCounts` ([#L1816-L1825](https://github.com/matrix-org/matrix-js-sdk/blob/v42.2.0/src/rust-crypto/rust-crypto.ts#L1816-L1825)) passes the real value.
3. Within one sync, the real value arrives last: `sync.ts` calls `preprocessToDeviceMessages` ([#L1195](https://github.com/matrix-org/matrix-js-sdk/blob/v42.2.0/src/sync.ts#L1195)), then `processDeviceLists` ([#L1574](https://github.com/matrix-org/matrix-js-sdk/blob/v42.2.0/src/sync.ts#L1574)), then `processKeyCounts` ([#L1583](https://github.com/matrix-org/matrix-js-sdk/blob/v42.2.0/src/sync.ts#L1583)). Sliding sync has the same shape via the to-device extension ([sliding-sync-sdk.ts#L157](https://github.com/matrix-org/matrix-js-sdk/blob/v42.2.0/src/sliding-sync-sdk.ts#L157)).
4. [matrix-rust-sdk#6780](https://github.com/matrix-org/matrix-rust-sdk/pull/6780) (shipped in crypto-wasm 18.5.0, see [CHANGELOG](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/blob/main/CHANGELOG.md)) made `receive_sync_changes()` treat a missing `signed_curve25519` count as **0** — the sync v2 semantics — and added a separate `receive_sync_changes_msc4186()` that keeps the old "missing ⇒ unchanged" behaviour. The wasm binding only exposes the strict variant, and its `one_time_keys_counts` parameter is a required `Map` ([machine.rs#L380-L384](https://github.com/matrix-org/matrix-rust-sdk-crypto-wasm/blob/v18.5.0/src/machine.rs#L380-L384)), so there is currently no way to say "I don't know".
5. Each empty-map call therefore sets `uploaded_key_count = 0`, vodozemac generates 50 keys, and `onSyncCompleted` publishes them. The truthful count that arrives moments later does not discard the batch already generated.

So the wasm change is not a bug — it fixed an under-upload case (a sync v2 response that omits `device_one_time_keys_count` genuinely means zero). The bug is that the SDK fabricates a count on paths where it does not have one.

### Impact

vodozemac keeps ~100 one-time keys per account and evicts the oldest as new ones are generated, so the private halves of all earlier batches are gone while the server still hands out their public halves. Any peer who claims one of those keys sends a pre-key message the device cannot open (`MissingOneTimeKey`), the Megolm session key never arrives, and the recipient sees `MEGOLM_UNKNOWN_INBOUND_SESSION_ID` (withheld code `None`) — permanently, since rust-crypto has no key-forwarding fallback and no backup holds a key nobody could decrypt. Existing Olm sessions keep working, so it surfaces as "new DMs / new rooms with this user never decrypt". Server-side, `e2e_one_time_keys_json` grows to tens of thousands of rows per device.

### Suggested fix

- **In the SDK (implementable today):** have `RustCrypto` remember the last `oneTimeKeysCounts` / `unusedFallbackKeys` received via `processKeyCounts` and pass them on the to-device and device-list paths instead of an empty map. Additionally, reorder `processSyncResponse` so `processKeyCounts` runs *before* `preprocessToDeviceMessages` / `processDeviceLists`, so the cache is warm even for the first sync after a restart. Note that reordering **alone** does not help — the last `receiveSyncChanges` call in a sync wins, so an empty map after the real count still resets it.
- **In the binding (longer term):** make `one_time_keys_counts` optional in `receiveSyncChanges` so "unknown" is expressible, and/or expose `receiveSyncChangesMsc4186` for the sliding-sync path, whose `e2ee` extension is specified with "missing ⇒ unchanged" semantics.
- A unit test spying on `olmMachine.receiveSyncChanges` and asserting the count argument on all three paths would have caught this; current tests either mock `receiveSyncChanges` or only call `preprocessToDeviceMessages([])`.

### Workaround

Pin the binding (no SDK downgrade needed, the crypto store survives, no re-login):

```json
"overrides": { "@matrix-org/matrix-sdk-crypto-wasm": "18.4.0" }
```

This stops the accumulation but heals nothing already published: affected devices need their server-side stock trimmed to the newest ~50 keys before peers can start sessions with them again.

### Versions

- `matrix-js-sdk` 42.2.0 (code path identical in 42.0.0/42.1.0 and `develop`)
- `@matrix-org/matrix-sdk-crypto-wasm` 18.5.0 and 18.6.0 affected; 18.4.0 not
- Synapse 1.155.0; Chrome 150 / Linux; third-party web client

### Related

- matrix-org/matrix-rust-sdk#6780 — the semantic change
- matrix-org/vodozemac#318 — `MAX_ONE_TIME_KEYS` hard-coded, which is what evicts the older private halves
- matrix-org/matrix-rust-sdk#4222, #5200, #6520 — likely downstream symptoms of a diverged OTK id space
- SableClient/Sable#1075 / #1078 — the same count plumbing failing the opposite way in another js-sdk client (count lost rather than asserted zero); both incidents stem from the API being unable to express "unknown"
- Separate issue to follow: when `/keys/upload` fails with a non-retryable 4xx (e.g. `M_UNKNOWN: One time key … already exists` after a store reset with a kept `device_id`), `OutgoingRequestProcessor` never calls `markRequestAsSent`, so the identical request is re-issued after every sync for the life of the session with no signal to the application.

LpZ-squall:
Hello everyone.

This PR fixes https://github.com/matrix-org/matrix-js-sdk/issues/5501. The code was written with Claude Code (Claude Fable 5); I have tested it successfully in the Matrix client I am developing.

As explained in the issue, this is a pretty bad bug: an affected device keeps uploading a new batch of 50 one-time keys on almost every sync. The server accumulates tens of thousands of keys, and since vodozemac only keeps the private half of the most recent ~100, peers who claim an older key can no longer open an Olm session with the device — new conversations become permanently undecryptable, while existing sessions keep working, which makes it hard to diagnose.

Sable's unreleased dev branch already resolves crypto-wasm 18.5.0, so its next release would be affected, and other matrix-js-sdk clients will be as their lockfiles refresh. Element Web is not affected yet only because its lockfile still pins 18.4.0 — and, as the tests in this PR show, bumping it is currently blocked by this bug.

### Why

`RustCrypto` calls `OlmMachine.receiveSyncChanges` up to three times per `/sync` response, and the to-device and device-list calls passed an empty one-time-key count because they have none of their own. Since matrix-sdk-crypto-wasm 18.5.0 ([matrix-rust-sdk#6780](https://github.com/matrix-org/matrix-rust-sdk/pull/6780)) an empty count means "zero keys on the server", so each such sync made the `OlmMachine` generate and upload 50 new one-time keys. Details and measurements in the issue.

### What

`RustCrypto` now remembers the count the server last reported and passes it on those calls. The `OlmMachine` learns that count from two places, so both are mirrored:

- `processKeyCounts` stores the `/sync` count, and `SyncApi.processSyncResponse` now calls it *before* processing to-device events and device lists, so the stored count is the one from the same response.
- `OutgoingRequestProcessor` reports the `one_time_key_counts` from each `/keys/upload` response (new optional callback), covering the time before `/sync` has reported a count.

The `SyncCryptoCallbacks.processKeyCounts` TSDoc documents the new ordering.

### Testing

- Unit tests (`spec/unit/rust-crypto/rust-crypto.spec.ts`, `OutgoingRequestProcessor.spec.ts`, new `spec/unit/sync.spec.ts`) assert the count handed to the `OlmMachine` on each path, the `/keys/upload` callback, and the call order in `processSyncResponse`.
- End-to-end test in `spec/integ/crypto/crypto.spec.ts` with the real `OlmMachine`: a sync carrying only to-device events and device-list changes must not trigger a `/keys/upload`.
- To see the failure, bump the binding: `pnpm add -D @matrix-org/matrix-sdk-crypto-wasm@18.6.0 --ignore-scripts` and run `spec/integ/crypto`. Without this change the new test fails (`expected 2 to deeply equal 1`) and 18 existing tests in `crypto.spec.ts` / `verification.spec.ts` time out, because the first to-device sync triggers a second key upload that `E2EKeyReceiver` deliberately never answers. With it, all 215 pass on 18.6.0, and the full suite passes on the lockfile's 18.4.0. So this also unblocks bumping the binding past 18.4.0.
- Verified in a production third-party client against 18.5.0: one-time-key uploads stop after the initial batch.

### Notes for reviewers

- CI runs on 18.4.0, where an empty count is ignored, so the new end-to-end test passes either way today; it starts guarding the behaviour as soon as the binding is bumped.
- The changelog gate needs the maintainer-only `T-Defect` label.
- The code was written with the help of Claude Code; I reviewed and tested it.

## Checklist

### DIFF
--- a/src/rust-crypto/rust-crypto.ts
+++ b/src/rust-crypto/rust-crypto.ts
@@ -175,7 +175,9 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
         private readonly enableEncryptedStateEvents: boolean = false,
     ) {
         super();
-        this.outgoingRequestProcessor = new OutgoingRequestProcessor(logger, olmMachine, http);
+        this.outgoingRequestProcessor = new OutgoingRequestProcessor(logger, olmMachine, http, (oneTimeKeyCounts) =>
+            this.onKeysUploadResponse(oneTimeKeyCounts),
+        );
         this.outgoingRequestsManager = new OutgoingRequestsManager(
             this.logger,
             olmMachine,
@@ -1691,16 +1693,42 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
     ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
     /**
+     * The most recent one-time key counts reported by the server, either in a `/sync` response (see
+     * {@link processKeyCounts}) or in the response to a `/keys/upload` request (see {@link onKeysUploadResponse}).
+     *
+     * `OlmMachine.receiveSyncChanges` requires a one-time key count on every call and (since
+     * matrix-sdk-crypto-wasm 18.5.0) treats a missing `signed_curve25519` entry as "there are no one-time keys
+     * on the server", which makes the `OlmMachine` generate a fresh batch of keys for upload. To avoid that
+     * happening on the calls made from {@link preprocessToDeviceMessages} and {@link processDeviceLists}, which
+     * have no count of their own, we remember the last count the server actually reported and pass that instead.
+     * See https://github.com/matrix-org/matrix-js-sdk/issues/5501.
+     *
+     * `undefined` until the server has reported a count.
+     */
+    private lastOneTimeKeysCounts?: Map<string, number>;
+
+    /**
+     * Called with the `one_time_key_counts` from the response to each `/keys/upload` request.
+     *
+     * The `OlmMachine` updates its own idea of the number of uploaded keys from this response, so we do the same
+     * with {@link lastOneTimeKeysCounts}, to keep the two in step.
+     */
+    private onKeysUploadResponse(oneTimeKeyCounts: Record<string, number>): void {
+        this.lastOneTimeKeysCounts = new Map<string, number>(Object.entries(oneTimeKeyCounts));
+    }
+
+    /**
      * Apply sync changes to the olm machine
      * @param events - the received to-device messages
-     * @param oneTimeKeysCounts - the received one time key counts
+     * @param oneTimeKeysCounts - the received one time key counts. If omitted, the most recent counts reported by
+     *    the server are used (see {@link lastOneTimeKeysCounts}), or an empty map if there are none yet.
      * @param unusedFallbackKeys - the received unused fallback keys
      * @param devices - the received device list updates
      * @returns A list of processed to-device messages.
      */
     private async receiveSyncChanges({
         events,
-        oneTimeKeysCounts = new Map<string, number>(),
+        oneTimeKeysCounts = this.lastOneTimeKeysCounts ?? new Map<string, number>(),
         unusedFallbackKeys,
         devices = new RustSdkCryptoJs.DeviceLists(),
     }: {
@@ -1723,8 +1751,9 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
      * @returns A list of preprocessed to-device messages.
      */
     public async preprocessToDeviceMessages(events: IToDeviceEvent[]): Promise<ReceivedToDeviceMessage[]> {
-        // send the received to-device messages into receiveSyncChanges. We have no info on device-list changes,
-        // one-time-keys, or fallback keys, so just pass empty data.
+        // send the received to-device messages into receiveSyncChanges. We have no info on device-list changes or
+        // fallback keys, so pass empty data for those. The one-time key count defaults to the last count reported by
+        // the server (see `receiveSyncChanges`).
         const processed = await this.receiveSyncChanges({ events });
 
         const received: ReceivedToDeviceMessage[] = [];
@@ -1819,6 +1848,11 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
     ): Promise<void> {
         const mapOneTimeKeysCount = oneTimeKeysCounts && new Map<string, number>(Object.entries(oneTimeKeysCounts));
         const setUnusedFallbackKeys = unusedFallbackKeys && new Set<string>(unusedFallbackKeys);
+
+        if (mapOneTimeKeysCount !== undefined) {
+            // Remember the count, for use on subsequent `receiveSyncChanges` calls which have no count of their own.
+            this.lastOneTimeKeysCounts = mapOneTimeKeysCount;
+        }
 
         if (mapOneTimeKeysCount !== undefined || setUnusedFallbackKeys !== undefined) {
             await this.receiveSyncChanges({
--- a/spec/integ/crypto/crypto.spec.ts
+++ b/spec/integ/crypto/crypto.spec.ts
@@ -1452,6 +1452,43 @@ describe("crypto", () => {
             expect(res.keysCount).toBeGreaterThan(0);
             expect(res.fallbackKeysCount).toBeGreaterThan(0);
         });
+
+        it("should not upload one-time keys when a sync only carries to-device messages or device list changes", async () => {
+            // Regression test for https://github.com/matrix-org/matrix-js-sdk/issues/5501: the crypto layer has to
+            // give the OlmMachine a one-time key count when it processes to-device messages and device list changes,
+            // and if it passes an empty count, the OlmMachine (since matrix-sdk-crypto-wasm 18.5.0) takes that to
+            // mean that the server has no one-time keys, and generates a new batch on every such sync.
+            expectAliceKeyQuery({ device_keys: { "@alice:localhost": {} }, failures: {} });
+            await startClientAndAwaitFirstSync();
+
+            // Wait for the initial batch of one-time keys to be uploaded. The E2EKeyReceiver replies with the number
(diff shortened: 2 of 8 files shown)


src/rust-crypto/rust-crypto.ts
```diff
@@ -1690,17 +1692,43 @@ export class RustCrypto extends TypedEventEmitter<RustCryptoEvents, CryptoEventH
...
+     * @param oneTimeKeysCounts - the received one time key counts. If omitted, the most recent counts reported by
+     *    the server are used (see {@link lastOneTimeKeysCounts}), or an empty map if there are none yet.
      * @param unusedFallbackKeys - the received unused fallback keys
      * @param devices - the received device list updates
      * @returns A list of processed to-device messages.
      */
     private async receiveSyncChanges({
         events,
-        oneTimeKeysCounts = new Map<string, number>(),
+        oneTimeKeysCounts = this.lastOneTimeKeysCounts ?? new Map<string, number>(),
```
kaylendog, line 1731: I don't think this is quite right. For sync v2, `device_one_time_keys_count` is allowed to be absent if that particular algorithm has no OTKs. However, this pulls in the previous count, so there's no way for the client to ever realise it has run out of OTKs.

E.g. I upload 50 OTKs to the server, the count of which I cache locally. Over time, these get used up until the last key is claimed, at which point the server omits the `device_one_time_keys_count` field. However, this parameter defaults to the last count (i.e. `1`), so we assume we have one remaining OTK for the rest of time.

kaylendog:
I have been working on a similar (although more complex) fix here: https://github.com/matrix-org/matrix-js-sdk/pull/5503

richvdh:
@LpZ-squall thanks for the contribution. The team were previously working on this so I'm going to close this in favour of #5503.

LpZ-squall:
Thanks for your reply and for the other fix :+1:.

PR #5505, opened 2026-08-27
Encryption-Key rotation slow down

toger5:
This changes the implementation to include the following "slow downs" + jitter delays.

## In text:
 - if there has not been a rotation in a longer time and a rotation trigger event occurs we dont rotate all users at the same time. instead we introduce random jitter to distribute the rotations in time.
 - If a client has already done a rotation recently, we postpone the next rotation based on the number of participants.
 
This is configurable via a to-device message contingent that is shared over the whole call.

```
 /**
     * The amount of keys that can be shared per minute.
     * This is a shared contingent so it scales quadratically with the amount of users.
     * Each additional user increases
     *  - the amount of to-device messages per rotation
     *  - the amount of all rotations (whole call) per leave/join event
     *
     * The client can only share sharedPerMinuteKeyRotationContingent/N keys per minute where N is the number of call participants.
     * As each rotation sends N-1 to-device messages, the client can only do sharedPerMinuteKeyRotationContingent/(N*(N-1)) = clientRotationsPerMinute.
     * This implies that each key share has to be at least 60s / clientRotationsPerMinute apart.
     *
     * This implies the following rotation interval:
     *  - 2000: 50 users: 1.2min, 100 users: 4.9min, 200: users: 19.9min
     *  - 3000: 50 users: 0.8min, 100 users: 3.3min, 200: users: 13.2min
     *  - 5000: 50 users: 0.5min, 100 users: 1.9min, 200: users: 7.9min
     */
    sharedPerMinuteToDeviceContingent?: number;
```

## In pictures
Screenshot 2026-08-28 at 13 48 12
There are two approaches to implement this system. This picture shows a time-window based approach.
Each participant (with an initial jitter) starts a time window. There is just one timer that triggers a callback at the end of each time window.
During the time window we record any joins/leaves and if there has been one, we distribute the keys.
If the number of participants change we update the duration of the next time window.
The picture visualises two clients. The "global"-time offset is the jitter so that the end of the timewindow are not colliding.

This PR implements the same idea in a different way. (to minimise the diff to the previous implementation)

Screenshot 2026-08-28 at 13 58 13
There is a jitter and a grace period. If keys rotate quickly we wait until the grace period before doing the next rotation.
If keys have not changed in some time we use the jitter to spread the otherwise occuring burst rotation.
## Checklist


src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -94,15 +94,54 @@ export class RTCEncryptionManager implements IEncryptionManager {
...
+     * The client can only share sharedPerMinuteKeyRotationContingent/N keys per minute where N is the number of call participants.
+     * As each rotation sends N-1 to-device messages, the client can only do sharedPerMinuteKeyRotationContingent/(N*(N-1)) = clientRotationsPerMinute.
+     * This implies that each key share has to be at least 60s / clientRotationsPerMinute apart.
+     *
+     * This implies the following rotation interval:
+     *  - 2000: 50 users: 1.2min, 100 users: 4.9min, 200: users: 19.9min
+     *  - 3000: 50 users: 0.8min, 100 users: 3.3min, 200: users: 13.2min
+     *  - 5000: 50 users: 0.5min, 100 users: 1.9min, 200: users: 7.9min
+     */
+    private sharedPerMinuteToDeviceContingent = 6000;
```
BillCarsonFr: This concept looks a bit complex to me, and the implication of that are not so clear.
Basically it is saying that I have a dynamic cool-down period after a rotation, and this can be 20mn.

So if an event happens that requires a rotation that I cannot do, what are the consequence?
- There is a new joiner but I cannot rotate for 20mn, so this new joiner will have to wait 20mn before beeing able to decrypt me? Or I have to give away the current key I am using thus cancelling the rotation that we said was required, meaning we can be leaking hours of call history to the joiner?

- If there is a new leaver but I cannot rotate for 20mn, then I have to leak 20mn of future call history.

The leaver case is bounded (dynamic but bounded), the key will be rotated after cooldown.

It is a bit harder to see with a joiner. If we are in a stable call (no changes since long, long standing key already distributed), then I will not be in cool-down, so I can rotate.
The case in which I cannot rotate means that the key was rotated not so long ago (<= cool-down duration). So two possibilities:
- I don't share the current key to the joiner => The joiner is in waiting list to get my stream for the remaining cool-down
- I share the current key, so I leak a bit of history but bounded by the cool-down period.

So it looks like that the joiner case is also bounded?

Also who controls this flag?

src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -94,15 +94,54 @@ export class RTCEncryptionManager implements IEncryptionManager {
...
+     * 15 minutes force key rotation.
+     *
+     * This is a stop gap until we have ratcheting.
+     * On join we want/need to share the current key with any participatns in the call.
+     * This is particularly important in larger calls, where the jitter can span multiple minutes.
+     * Hence we can end up with media delays on join of multiple minutes.
+     *
+     * As a workaround we share immediatlye but make sure we never use a key that is older than 15min + jitter.
+     */
+    private maxKeyTTL: number | undefined = 60 * 15 * 1000;
```
BillCarsonFr: Ok so I see that this was added as en effort to add a maximum boundary to what could be leaked when we decide to share an existing key when we cannot rotate but still want the new comer to participate in the call instead of waiting for next distribution.

But as per my analysis on `sharedPerMinuteToDeviceContingent` it appears that the cool-down acts as a boundary already?

Having a TTL is a good thing and as merits by itself standalone. For example if one of the joiner is compromised temporarly only TTL duration of the call will be compromised.

src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -94,15 +94,54 @@ export class RTCEncryptionManager implements IEncryptionManager {
...
+     * That is where the quadratic scaling comes from: more to-device messages per rotation for one user and more user who need to rotate.
+     * The call as a whole can only afford a limit amount per time.
+     * See {@link RTCEncryptionManager.sharedPerMinuteToDeviceContingent} for more details.
      *
      * This must be higher than `useKeyDelay` to have an effect.
      * If it is lower, the current key will always be older than the grace period.
      * @private
      */
-    private keyRotationGracePeriodMs = 10_000;
+    private get keyRotationGracePeriodMs(): number {
```
BillCarsonFr: I know this is an attempt to not modify the existing code, but in the current implementation this is really a grace period, that applies to new-joiner that allows to share an existing young key to them instead of rotating. So it is really a grace period.

It is now also acting as rotation cool-down that delays the needed rotation to later, to ensure that a leaver with a young key will only have a limited/bounded access to future call history.

So I think now the naming is bad

src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -117,7 +156,7 @@ export class RTCEncryptionManager implements IEncryptionManager {
      * This will ensure that a new round is started after the current one.
      * @private
      */
-    private needToEnsureKeyAgain = false;
+    private needToEnsureKeyAgain: { scheduled: boolean; fakeMemberChange: boolean } | undefined = undefined;
```
BillCarsonFr, line 174: What is that `fakeMemberChange`?

src/matrixrtc/RTCEncryptionManager.ts
```diff
@@ -277,29 +317,30 @@ export class RTCEncryptionManager implements IEncryptionManager {
      * If this function is called repeatedly while a distribution is in progress,
      * the calls will be coalesced to a single new distribution (that will start just after the current one has completed).
      */
-    private ensureKeyDistribution(): void {
+    private ensureKeyDistribution(scheduled = false, fakeMemberChange = false): void {
```
BillCarsonFr, line 344: any docs on what scheduled mean? and fakeMemberChange?

Currently `ensureKeyDistribution` is called after membership changes or as a followup when a rollout did finish and ensure was call meanwhile.

It basically woks based on snapshots of participant. It can, only based on the previous snapshot and the current snapshot. 
Now the snapshot is not only membership but also the scheduled and fakeMemberChange thing?

I find this hard to understand, the naming should be reviewed, or we should change the "key distribution promise pattern" pattern to match your new vision.

spec/unit/matrixrtc/RTCEncryptionManager.spec.ts
```diff
@@ -638,6 +766,672 @@ describe("RTCEncryptionManager", () => {
...
+         *
+         * On top of the participants that stay for the whole simulation there is one extra participant that
+         * joins and leaves {@link CHANGES_PER_INTERVAL} times per rotation interval for
+         * {@link TOGGLING_INTERVALS} intervals, and is then gone for {@link QUIET_INTERVALS} of them. That
+         * flapping participant has no manager of its own, we only care about how the others react to it.
+         *
+         * Only what happens after everybody has joined is recorded: filling up the call is not something the
+         * contingent is meant to limit.
+         */
+        async function simulateCall(participantCount: number): Promise<SimulationResult> {
```
BillCarsonFr, line 1192: this is great 🤩

BillCarsonFr (changes requested):
I am sorry but I have to push back on this.
I know that it attemps to change the current behavior without touching to much the encryption manager pattern, but it looks to me that the behavior change is too profound for that.

I am not so much confident that the current change is not affecting small calls but only big calls. It is adding a good layer of complexity, as a proof we got it wrong when we thought the joiner could get unbounded history.

My gutt feeling is that we should go to simpler:

- 1️⃣ Let's stop basing key rotation on a sub group of the room. And instead base it on the room membership. That means we never rotate keys when people jump in and out of the call. Instead we rotate when people join in and out of the room. This would be a simpler varation of the current encryption manager

- 2️⃣ We do time based rotation. Rotate every 15mn, unrelated to particiapant changes.

Eventually we could have a dynamic pickup of the encryption manager:
- if room size < 50, keep the current sub-group manager
- else use the room membership based one

PR #5506, opened 2026-08-28
Log OAuth requests and responses

hughns:
The OAuth 2.0 flows in [src/oauth/](src/oauth/) called `fetch` directly, so none of the requests to the identity provider's registration, device authorization, token or revocation endpoints showed up in the logs. That makes debugging login and token refresh problems noticeably harder than debugging Client-Server API calls, which `FetchHttpApi` logs.

This adds a `fetchWithLogging` wrapper that emits the same `-->` / `<--` debug lines as `FetchHttpApi`, and uses it for every OAuth 2.0 request:

```
OAuth2: --> POST https://auth.org/token
OAuth2: <-- POST https://auth.org/token [42ms 200]
```

As in `FetchHttpApi`, neither the request nor the response body is logged, and query parameter values are redacted — these requests routinely carry credentials. `sanitizeUrlForLogs` has been lifted out of `FetchHttpApi` into `src/http-api/logging.ts` so both can share it, with no change in behaviour.

`OAuth2`, `OAuth2.registerClient`, `startDeviceAuthorization` and `waitForDeviceAuthorization` now accept an optional `Logger` so callers can route these lines to the logger of their choice (e.g. a `MatrixClient`'s logger). All of them default to the js-sdk root logger, so this is backwards compatible.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code) (with review by human)

### DIFF
--- a/src/http-api/logging.ts
+++ b/src/http-api/logging.ts
@@ -18,6 +18,7 @@ limitations under the License.
  * Produce a version of the given URL which is safe to write to logs, by redacting the values of any query parameters,
  * as they may contain secrets.
  *
+ * @internal
  * @param url - the URL to sanitize.
  * @returns the sanitized URL, or `"??"` if the URL could not be parsed.
  */
--- a/src/http-api/logging.ts
+++ b/src/http-api/logging.ts
@@ -1,0 +1,45 @@
+/*
+Copyright 2026 The Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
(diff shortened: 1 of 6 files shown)


hughns:
Please let me know if there is a preferred way to wire up the logger.

src/oauth/fetch.ts
```diff
@@ -0,0 +1,53 @@
...
+ * `FetchHttpApi` does for Client-Server API requests.
+ *
+ * Neither the request nor the response body is logged, as they routinely contain credentials.
+ * Query parameter values are redacted for the same reason.
+ *
+ * @param logger - the logger to write the request and response lines to.
+ * @param resource - the URL to request.
+ * @param options - the options to pass to `fetch`.
+ * @returns the `Response`, whatever its status code.
+ * @throws rethrows whatever `fetch` threw, having logged it.
```
t3chguy, line 33: should probably be `@internal` so we don't have to consider the API surface sacred

t3chguy (approved):

src/http-api/logging.ts
```diff
@@ -0,0 +1,45 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * Produce a version of the given URL which is safe to write to logs, by redacting the values of any query parameters,
+ * as they may contain secrets.
+ *
+ * @param url - the URL to sanitize.
+ * @returns the sanitized URL, or `"??"` if the URL could not be parsed.
```
t3chguy, line 23: Ditto `@internal` here - as it was previously `private`

richvdh:
Context for this PR: it might help debug https://github.com/element-hq/element-web/issues/30826

PR #5513, opened 2026-09-01
Support the stable version of mutual rooms

CatCraftYT:
Adds support for the latest stable version of [mutual rooms](https://spec.matrix.org/latest/client-server-api/#mutual-rooms) in the spec (v1.19 currently).

I made sure to maintain support for the older unstable versions, since I'm not sure what the policy is on those. I tested it on a real server as well, seems to work.

## Checklist

I signed off with my GitHub username. I hope that's still okay :)

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -556,6 +556,7 @@ export const GET_LOGIN_TOKEN_CAPABILITY = new NamespacedValue(
 export const UNSTABLE_MSC2666_SHARED_ROOMS = "uk.half-shot.msc2666";
 export const UNSTABLE_MSC2666_MUTUAL_ROOMS = "uk.half-shot.msc2666.mutual_rooms";
 export const UNSTABLE_MSC2666_QUERY_MUTUAL_ROOMS = "uk.half-shot.msc2666.query_mutual_rooms";
+export const STABLE_MSC2666_QUERY_MUTUAL_ROOMS = "uk.half-shot.msc2666.query_mutual_rooms.stable";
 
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -6095,72 +6096,113 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
...
-        // Newer variant that renamed shared rooms to mutual rooms
-        const mutualRoomsSupport = await this.doesServerSupportUnstableFeature(UNSTABLE_MSC2666_MUTUAL_ROOMS);
-
-        // Latest variant that changed from path elements to query elements
-        const queryMutualRoomsSupport = await this.doesServerSupportUnstableFeature(
-            UNSTABLE_MSC2666_QUERY_MUTUAL_ROOMS,
-        );
-
-        if (!sharedRoomsSupport && !mutualRoomsSupport && !queryMutualRoomsSupport) {
+    public async getMutualRooms(userId: string): Promise<string[]> {
```
Half-Shot: Can you split this function into two private functions for stable/unstable, which will make this easier to read.

Half-Shot (changes requested):
Looks good, but can you split up that one function. You can keep `getMutualRooms` the public function but having two private functions to handle the different endpoints would make this an easier parse.

Half-Shot:
We're having fun with the code duplication check again, I'll just fix this...

Half-Shot (approved):

PR #5516, opened 2026-09-02
Set hasCurrentUserParticipated when the current user sends a thread reply

### LINKED ISSUE #5515: Thread.hasCurrentUserParticipated stays false after the current user replies
### Description

`Thread.hasCurrentUserParticipated` is only ever set from the homeserver's bundled `current_user_participated` field on the thread root's aggregated `m.thread` relation. It is not updated when the current user actually sends a reply into the thread.

As a result, immediately after the local user replies in a thread they had not previously participated in, `hasCurrentUserParticipated` remains `false` until the server re-sends the root event's bundled relation (which, outside of a gappy/`limited` sync, may not happen for a long time).

### Where the flag is (and isn't) written

All references in `src/models/thread.ts` (as of `develop` @ 28c07759e):

- Set from the server bundle only — [`processRootEvent()`, L548](https://github.com/matrix-org/matrix-js-sdk/blob/28c07759e/src/models/thread.ts#L548):
  ```ts
  this._currentUserParticipated = !!bundledRelationship.current_user_participated;
  ```
- Reset to `false` on full thread redaction — [`onRedaction()`, L250](https://github.com/matrix-org/matrix-js-sdk/blob/28c07759e/src/models/thread.ts#L250):
  ```ts
  this._currentUserParticipated = false;
  ```
- The root-bundle refresh is guarded to run **once** — [`updateThreadMetadata()`, L647](https://github.com/matrix-org/matrix-js-sdk/blob/28c07759e/src/models/thread.ts#L647):
  ```ts
  if (!this.processRootEventPromise) {
      this.processRootEventPromise = this.updateThreadFromRootEvent();
  }
  ```

`onEcho()` (which fires on the user's own reply) calls `updateThreadMetadata()`, but because `processRootEventPromise` is already set no re-fetch happens, and neither `onEcho()` nor `addEvent()` writes `_currentUserParticipated`. So a just-sent reply does not flip the flag.

### Steps to reproduce

1. Alice starts a thread in a room (root message + one reply).
2. Bob (the current user) opens that thread and posts a reply.
3. Inspect `thread.hasCurrentUserParticipated` for that thread on Bob's client.

**Expected:** `true` — Bob has participated (he replied).
**Actual:** `false`, until the server later re-sends the root event's bundled relation.

### Impact

Any consumer that reads `hasCurrentUserParticipated` to decide "is this thread mine?" gets the wrong answer for freshly-replied threads. Two concrete cases in element-web:

- **Threads Activity Centre** categorises unread threads into *My threads* / *Other threads* using this flag; a thread the user just replied in wrongly appears under *Other threads*. See this comment in PR https://github.com/element-hq/element-web/pull/32851#issuecomment-5437348090
- **Per-room Threads panel "My threads" filter** — the incremental live update only adds a thread root to the participated timeline set when the flag is set ([`Room.updateThreadRootEvents`, room.ts L2500](https://github.com/matrix-org/matrix-js-sdk/blob/28c07759e/src/models/room.ts#L2500)):
  ```ts
  this.updateThreadRootEvent(this.threadsTimelineSets?.[0], thread, ...);   // All: always
  if (thread.hasCurrentUserParticipated) {                                  // My: only if flag set
      this.updateThreadRootEvent(this.threadsTimelineSets?.[1], thread, ...);
  }
  ```
  This is usually masked because the participated set is seeded from the authoritative server endpoint (`GET /_matrix/client/v1/rooms/{roomId}/threads?include=participated`, MSC3856) on open and re-fetched on reopen — but the live path shares the same staleness.

### Suggested fix

When an `m.thread` reply authored by the current user is added to the thread, latch `_currentUserParticipated = true` locally (in addition to the server bundle). Concretely, in `addEvent()` right after `this.setEventMetadata(event)`:

```ts
if (event.getSender() === this.client.getUserId() && event.isRelation(THREAD_RELATION_TYPE.name)) {
    this._currentUserParticipated = true;
}
```

Gating on `THREAD_RELATION_TYPE` (rather than any relation) matches the server's `current_user_participated` semantics, so the user's own reactions and edits do not count as participation. `addEvent()` is the common path for both local echo and remote sync events, so this covers "I just replied" as well as "my reply arrived from another device". The existing redaction path still clears the flag, and the server bundle remains the authoritative source afterwards.

A PR will follow, I hope this approach looks right.

### Environment

- matrix-js-sdk: `develop` (verified @ 28c07759e)
- Server-side thread support (MSC3440 / MSC3856) enabled.

### Downstream workaround

element-web currently works around this locally in `useUnreadThreadRooms.ts` (`hasCurrentUserSentInThread`, which inspects the local timeline for a sent `m.thread` reply) - see [element-hq/element-web#32851](https://github.com/element-hq/element-web/pull/32851 and the commit https://github.com/element-hq/element-web/pull/32851/commits/11c06e75b16a746d20e39134b9f5b23231c342f4). Fixing the root cause here would let that be dropped.

nathanael-h:
Fixes #5515

### Problem

`Thread.hasCurrentUserParticipated` is only populated from the homeserver's
bundled `current_user_participated` field on the thread root (`processRootEvent()`),
and that refresh is guarded to run only once. Nothing sets the flag when the
current user replies, so it stays `false` after a local reply until the server
re-sends the root's bundled relation.

Since `Room.updateThreadRootEvents` only adds a root to the participated timeline
set when the flag is set, a just-replied thread is missing from that set.

### Fix

In `addEvent()`, latch the flag when the current user authors an `m.thread` reply

## Checklist

### DIFF
--- a/spec/unit/models/thread.spec.ts
+++ b/spec/unit/models/thread.spec.ts
@@ -714,6 +714,60 @@ describe("Thread", () => {
         });
     });
 
+    describe("hasCurrentUserParticipated", () => {
+        const myUserId = "@bob:example.org";
+        const otherUserId = "@alice:example.org";
+        const roomId = "!room:example.org";
+
+        let previousThreadHasServerSideSupport: FeatureSupport;
+
+        beforeAll(() => {
+            previousThreadHasServerSideSupport = Thread.hasServerSideSupport;
+            Thread.setServerSideSupport(FeatureSupport.Stable);
+        });
+
+        afterAll(() => {
+            Thread.setServerSideSupport(previousThreadHasServerSideSupport);
+        });
+
+        /**
+         * Create a client whose user is `myUserId`, plus a thread rooted by someone else which
+         * the current user has not participated in (the root carries no bundled relation, so the
+         * server's `current_user_participated` flag is absent).
+         */
+        async function createThreadWithoutMe(): Promise<{ client: MatrixClient; thread: Thread }> {
+            const client = createClient();
+            vi.spyOn(client, "getUserId").mockReturnValue(myUserId);
+            const thread = await createThread(client, otherUserId, roomId);
+            expect(thread.hasCurrentUserParticipated).toBe(false);
+            return { client, thread };
+        }
+
+        it("is set when the current user replies in the thread", async () => {
+            const { thread } = await createThreadWithoutMe();
+
+            thread.addEvent(createThreadMessage(thread.id, myUserId, roomId, "my reply"), false);
+
+            expect(thread.hasCurrentUserParticipated).toBe(true);
+        });
+
+        it("is not set when another user replies in the thread", async () => {
+            const { thread } = await createThreadWithoutMe();
+
+            thread.addEvent(createThreadMessage(thread.id, otherUserId, roomId, "their reply"), false);
+
+            expect(thread.hasCurrentUserParticipated).toBe(false);
+        });
+
+        it("is not set by the current user's reaction to a thread event", async () => {
+            const { client, thread } = await createThreadWithoutMe();
+
+            thread.addEvent(mkReaction(thread.rootEvent!, client, myUserId, roomId), false);
+
+            expect(thread.hasCurrentUserParticipated).toBe(false);
+        });
+    });
+
     describe("addEvent", () => {
         describe("Given server support for threads", () => {
             let previousThreadHasServerSideSupport: FeatureSupport;
--- a/src/models/thread.ts
+++ b/src/models/thread.ts
@@ -369,6 +369,15 @@ export class Thread extends ReadReceipt<ThreadEmittedEvents, ThreadEventHandlerM
         // Modify this event to point at our room's state, and mark its thread
         // as this.
         this.setEventMetadata(event);
+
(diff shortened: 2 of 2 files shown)


florianduros (changes requested):
Code looks good thanks!

The PR description is too exhaustive and mentions EW where the issue is purely on the js-sdk side.

Can you refine the PR description to keep it short and concise about the current issue?

nathanael-h:
You're welcome, thanks for your review. I shortened the description of the PR.

florianduros (approved):

florianduros:
@nathanael-h last step, sonar is unhappy:

image

nathanael-h:
Fixed — the tests are unchanged, I only moved them.

The cause turned out to be placement rather than duplicated test code. `thread.spec.ts` already has a 21-line duplication on `develop`, between `describe("Editing events") > describe("Given server support for threads")` and the identical setup in `describe("addEvent")`. My new `describe` was inserted immediately before the second of those, so its last two lines (the closing `});` and the blank line) fell inside that duplicated block and were counted as duplicated *new* code: 2 lines out of 63, i.e. 3.2%.

Moving the block to the end of the file's top-level `describe` takes it away from that boundary, so no new line intersects a duplicated block any more. Left as a separate commit so the move is easy to check.

PR #5519, opened 2026-09-03
Threads: cap concurrent initial-events fetches at 3

ara4n:
This is the root cause of folks repeatedly getting kicked out of MatrixRTC calls on the element.io homeserver; it turns out that a single user opening the ThreadPanel in EW is enough to stall the server enough to cause delayed events to be starved and all users to timeout of calls.

Opening a thread list (`Room.fetchRoomThreads`) creates a `Thread` for every root in the room, and each one immediately paginates its timeline to load its initial events, which is a recursive `/relations` request. In a room with 30 threads that is 30 heavy requests fired at once; in element-hq/element-call-rageshakes#17271-17276 one user opening the Threads panel did exactly this, the requests held every database connection of the homeserver's only client-reader worker for ~55s, and everyone else on that server (including every MatrixRTC delayed-leave keep-alive) stalled behind them. Run at most 3 initial-events fetches at a time; the rest queue in creation order and nothing else about thread loading changes.

(generated by fable).

### DIFF
--- a/spec/unit/models/thread.spec.ts
+++ b/spec/unit/models/thread.spec.ts
@@ -714,6 +714,69 @@ describe("Thread", () => {
         });
     });
 
+    describe("initial events fetch", () => {
+        it("fetches the initial events of at most 3 threads at a time", async () => {
+            const previousSupport = Thread.hasServerSideSupport;
+            Thread.setServerSideSupport(FeatureSupport.Stable);
+            try {
+                const myUserId = "@alice:example.org";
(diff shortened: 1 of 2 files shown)


dbkr (approved):
Discussed out-of-band on whether this is a good idea or whether this should be a server side fix. I would say the server definitely ought to be resilient to this and should probably prioritise time-sensitive delayed events over paginations, but I think this is probably also a useful thing to have on the client side: three seems like a reasonable number of requests in flight, it's good to not hammer the server with loads all at once.

PR #5520, opened 2026-09-03
Bring OAuth token refresh into the js-sdk

dbkr:
Changes the API so that refreshing tokens is now js-sdk's job rather than the client, so the js-sdk fetches the auth metadata. This allows it to construct the OAuth object lazily which means it can retry if, for example, there was no internet when the client started (otherwise the app would have to wait until it could fetch auth metadata before starting the client).

Also removes support for the unstable prefixed MSC2965: this has been stable for about a year now. Supporting stable/unstable meant we had to check /versions before requesting auth metadata, which of course we had to do without auth, else get infinite loops, which would have added a buncvh of complexity to manage fetching and caching authed/ unauthed /versions requests.

## Checklist

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -73,7 +73,8 @@ import {
     MediaPrefix,
     Method,
     retryNetworkOperation,
-    type TokenRefreshFunction,
+    type onTokenRefreshCallback,
+    type OAuth2ClientConfig,
     type Upload,
     type UploadOpts,
     type UploadResponse,
@@ -318,11 +319,24 @@ export interface ICreateClientOpts {
     refreshToken?: string;
 
     /**
-     * Function used to attempt refreshing access and refresh tokens
-     * Called by http-api when a possibly expired token is encountered
-     * and a refreshToken is found
+     * Called when the access tokens are refreshed.
+     * The client must replace the tokens it had stored previously which will no
+     * longer be valid.
+     *
+     * Required if a {@link IHttpOpts.onTokenRefresh} is provided.
+     *
+     * @param newAccessToken The new access token.
+     * @param newRefreshToken The new refresh token.
      */
-    tokenRefreshFunction?: TokenRefreshFunction;
+    onTokenRefresh?: onTokenRefreshCallback;
+
+    /**
+     * If this is an OAuth2-native session (as per MSC2965/MSC3861), the client ID and redirect URI this
+     * application is registered with the delegated auth server under. Required for the SDK to be able to
+     * refresh (given `refreshToken`) and revoke (e.g. on logout) tokens for the session; the auth server's
+     * metadata will be discovered lazily, and the OAuth2 client constructed, the first time this is needed.
+     */
+    oauth2ClientConfig?: Pick<OAuth2ClientConfig, "clientId" | "redirectUri">;
 
     /**
      * Identity server provider to retrieve the user's access token when accessing
@@ -1349,7 +1363,12 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
             idBaseUrl: opts.idBaseUrl,
             accessToken: opts.accessToken,
             refreshToken: opts.refreshToken,
-            tokenRefreshFunction: opts.tokenRefreshFunction,
+            onTokenRefresh: opts.onTokenRefresh,
+            oauth2ClientConfig: opts.oauth2ClientConfig && {
+                ...opts.oauth2ClientConfig,
+                deviceId: this.deviceId ?? undefined,
+                getAuthMetadata: (): Promise<ValidatedAuthMetadata> => this.getAuthMetadata(),
+            },
             prefix: ClientPrefix.V3,
             onlyData: true,
             extraParams: opts.queryParams,
@@ -6199,6 +6218,7 @@ export class MatrixClient extends TypedEventEmitter<EmittedEvents, ClientEventHa
     /**
      * Get the API versions supported by the server, along with any
      * unstable APIs it supports
+     *
      * @returns The server /versions response
      */
     public async getVersions(): Promise<IServerVersions> {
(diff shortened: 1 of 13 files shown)


src/client.ts
```diff
@@ -323,11 +319,24 @@ export interface ICreateClientOpts {
...
+     *
+     * Required if a {@link IHttpOpts.onTokenRefresh} is provided.
+     *
+     * @param newAccessToken The new access token.
+     * @param newRefreshToken The new refresh token.
+     */
+    onTokenRefresh?: onTokenRefreshCallback;
+
+    /**
+     * If this is an OAuth2-native session (as per MSC2965/MSC3861), the client ID and redirect URI this
```
t3chguy: The redirect URI might change during the runtime, e.g. as the user navigates around the app with hash-routing. Having to instantiate a new MatrixClient each time a user routes feels bad

src/oauth/tokenRefresher.ts
```diff
@@ -26,16 +27,15 @@ export class TokenRefresher {
...
     ) {}
 
     /**
      * Attempt token refresh using given refresh token
      * @param refreshToken - refresh token to use in request with token issuer
-     * @returns tokens - Promise that resolves with new access and refresh tokens
      * @throws when token refresh fails
      */
-    public tokenRefreshFunction: TokenRefreshFunction = async (refreshToken: string): Promise<AccessTokens> => {
+    public async doRefresh(refreshToken: string): Promise<AccessTokens> {
```
t3chguy: not a great name for a public method imo

src/client.ts
```diff
@@ -323,11 +318,22 @@ export interface ICreateClientOpts {
...
+     * @param newRefreshToken The new refresh token.
+     */
+    onTokenRefresh?: onTokenRefreshCallback;
+
+    /**
+     * If this is an OAuth2-native session (as per MSC2965/MSC3861), the OAuth client ID this
+     * application is registered with the delegated auth server under.
      */
-    tokenRefreshFunction?: TokenRefreshFunction;
+    oauthClientId?: string;
```
t3chguy, line 347: This feels like a chicken and egg problem which necessitates the use of 3 client objects to perform auth.

1. client with just hsUrl so you can getAuthMetadata in order to perform dynamic client registration to get `oauthClientId`
2. client with oauthClientId so you can auth
3. authed client

dbkr: Should we just move all of that into js-sdk do you think? (Is that possible?)

t3chguy: It would be possible to an extent, it'd mean js-sdk has to grow new storage fields. It'd also likely get entangled with things which the js-sdk shouldn't do i.e handling the environment like opening new tabs, looking at `window.location` and such to read params, so a line should be drawn somewhere imo

dbkr: Yeah, feels like more changes than I'd want to pile into this refactor at least. We could allow the client ID to be set after instantiation?

dbkr: Or separate out getAuthMetadata() to something you could use separately from the client, although it would still need a baseUrl.

src/client.ts
dbkr: Have moved this to be a param of the two methods that use it.

src/oauth/tokenRefresher.ts
dbkr: yeah, fair - `refreshTokens` is better?

src/client.ts
```diff
@@ -323,11 +323,22 @@ export interface ICreateClientOpts {
...
 
     /**
-     * Function used to attempt refreshing access and refresh tokens
-     * Called by http-api when a possibly expired token is encountered
-     * and a refreshToken is found
+     * Called when the access tokens are refreshed.
+     * The client must replace the tokens it had stored previously which will no
+     * longer be valid.
+     *
+     * Required if a {@link IHttpOpts.onTokenRefresh} is provided.
```
t3chguy: This doesn't make much sense? X is required if X is provided on a struct the caller doesn't control?

src/http-api/refresh.ts
```diff
@@ -28,25 +30,22 @@ export const enum TokenRefreshOutcome {
...
-type Opts = Pick<IHttpOpts, "tokenRefreshFunction" | "logger" | "refreshToken" | "accessToken">;
+type Opts = Pick<IHttpOpts, "onTokenRefresh" | "logger" | "refreshToken" | "accessToken" | "oauth2ClientConfig">;
 
 /**
  * This class is responsible for managing the access token and refresh token for authenticated requests.
- * It will automatically refresh the access token when it is about to expire, and will handle unknown token errors.
+ * It will automatically refresh the tokens when the access token is about to expire and can handle
+ * Unknown Token errors (via @{link handleUnknownToken})
+ *
+ * It will update the @{link opts} object with new tokens as they are refreshed, and also call the onTokenRefresh callback if provided.
```
t3chguy: `{@link ...` maybe?

src/client.ts
dbkr: Ah, that's because I meant refreshToken. Have docced those parameters too while I'm there.

src/http-api/interface.ts
```diff
@@ -39,12 +40,31 @@ export type AccessTokens = {
 };
 
 /**
- * Function that performs token refresh using the given refreshToken.
- * Returns a promise that resolves to the refreshed access and (optional) refresh tokens.
+ * A callback function called when tokens are refreshed such that stored tokens can be updated with
+ * the refreshed ones.
+ */
+export type onTokenRefreshCallback = (newTokens: AccessTokens) => void;
```
t3chguy: I thought our types should be PascalCase rather than camelCase

src/http-api/interface.ts
```diff
@@ -39,12 +40,31 @@ export type AccessTokens = {
...
+    /** The OAuth2 client ID this application is registered with the delegated auth server under. */
+    clientId: string;
+    /** The device ID of the current session, used to scope refreshed/revoked tokens to this session. */
+    deviceId?: string;
+    /**
+     * Discovers and validates the delegated auth server's metadata for the current homeserver.
+     * @returns validated authentication metadata
+     * @throws when delegated auth config is invalid or unreachable
+     */
+    getAuthMetadata: () => Promise<ValidatedAuthMetadata>;
```
t3chguy: feels odd for a config interface to take a getter callback? Is this intended to be an internal interface maybe?

t3chguy (approved):

src/http-api/interface.ts
dbkr: Oh yes, good shout, in fact it should probably lose the 'on' altogether

src/http-api/interface.ts
dbkr: Yeah, this was probably a bit strange. Maybe better separated out into its own param? ab43783bf does this.

dbkr:
Merging along with https://github.com/element-hq/element-web/pull/34918 as merge CI won't pass on either until both are merged due to API change.

PR #5522, opened 2026-09-05
MatrixRTC: restart the delayed leave event before extending a membership's expiry

ara4n:
Fixes a way for a MatrixRTC membership to end up on the server with no delayed leave event behind it, so that nothing removes it when the client goes away.

**Scenario** (from [element-call-rageshakes#17311](https://github.com/element-hq/element-call-rageshakes/issues/17311), confirmed against the Synapse logs): a laptop sleeps mid-call. No restarts reach the server, so it sends the delayed leave and the membership is gone. On laptop wake, there's a race between the /sync response which tells the client that their membership has expired, and the client setting up a delayed event: the `MembershipManager`'s overdue `UpdateExpiry` action fires before the sync carrying the leave is processed and re-sends the full membership as a plain state event, with `created_ts` preserved. That resurrects the membership with no delayed leave event scheduled for it. The tab then had its browser quit, and the membership stayed for the full `expires` (50 hours with Element Call's config).

**Fix**: `updateExpiryOnJoinedEvent` now restarts the delayed leave event before sending the update. A success is the server's guarantee that the leave outlives the membership update, so the ordering is safe regardless of local clocks. A 404 means the leave has been sent, and we go through the normal re-join path (new delayed event first, then the membership) rather than re-sending blind. Other errors are handled as they are for restarts. Two tests cover the ordering and the re-join.

Generated by Fable.

## Checklist

### DIFF
--- a/spec/unit/matrixrtc/MembershipManager.spec.ts
+++ b/spec/unit/matrixrtc/MembershipManager.spec.ts
@@ -709,6 +709,77 @@ describe("MembershipManager", () => {
             expect((vi.mocked(client.sendStateEvent).mock.calls[0][2] as SessionMembershipData).expires).toBe(10_000);
             expect(manager.status).toBe(Status.Connected);
         });
+
+        it("retries extending `expires` when restarting the delayed leave event is rate limited", async () => {
+            const manager = new MembershipManager(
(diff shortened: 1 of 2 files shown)


Johennes (approved):
Made a bunch of fixes to improve upon the initial proposal.

PR #5523, opened 2026-09-05
MatrixRTC: don't let a superseded membership recalculation overwrite a newer one

ara4n:
Fixes ghost call members lingering after a sync that carries several membership state events at once.

**Scenario** (from [element-call-rageshakes#17311](https://github.com/element-hq/element-call-rageshakes/issues/17311)): Element Call is reopened after a few hours away. The first incremental sync's `state_after` holds a handful of `m.call.member` changes, which `RoomState.setStateEvents` emits one by one. `MatrixRTCSessionManager` calls `session._onRTCSessionMemberUpdate()` for each, so several `recalculateSessionMembers` runs start back to back. Each reads the room state at its start, then awaits `CallMembership.parseFromEvent` per member, so a run that started earlier (seeing more members still present) finishes later and overwrites the result of the run that saw the final state. The client log shows exactly this: five recalculations emitting 1, 1, 2, 2, 3 members in that order, ending with two users the server had removed hours earlier, who then vanished on the next sync response. The sync data itself was correct.

**Fix**: each recalculation takes a generation number; one that has been overtaken discards its result and awaits the newest run, so callers (the session manager's started/ended detection, joining) still observe the final state. Serialising through `ensureRecalculateSessionMembers` instead changed the timing that the join/notification path relies on, so this keeps the immediate recalculation and just drops stale results.

The new test drives two updates back to back with the state changing in between and fails on develop (the stale one-member result wins).

Generated by Fable.

## Checklist

### DIFF
--- a/spec/unit/matrixrtc/MatrixRTCSession.spec.ts
+++ b/spec/unit/matrixrtc/MatrixRTCSession.spec.ts
@@ -833,6 +833,27 @@ describe("MatrixRTCSession", () => {
             expect(onMembershipsChanged).toHaveBeenCalled();
         });
 
+        it("does not let an older recalculation overwrite a newer one", async () => {
+            const bob = { ...sessionMembershipTemplate, user_id: "@bob:example.org", device_id: "BBBBBBB" };
+            const mockRoom = makeMockRoom([sessionMembershipTemplate, bob]);
+            sess = MatrixRTCSession.sessionForSlot(client, mockRoom, callSession);
+            await sess.initialMembershipCalculated;
+            expect(sess.memberships).toHaveLength(2);
+
+            // Two membership state events land in one sync batch, so two updates are
+            // requested back to back before either recalculation has finished: Bob leaves,
+            // then Alice leaves. Parsing memberships is async and takes longer the more
+            // there are, so the stale (larger) result would otherwise finish last and win.
+            mockRoomState(mockRoom, [sessionMembershipTemplate]);
+            const first = sess._onRTCSessionMemberUpdate();
+            mockRoomState(mockRoom, []);
+            const second = sess._onRTCSessionMemberUpdate();
+            await Promise.all([first, second]);
+            await flushPromises();
+
+            expect(sess.memberships).toHaveLength(0);
+        });
+
         // TODO: re-enable this test when expiry is implemented
         it.skip("emits an event at the time a membership event expires", () => {
             vi.useFakeTimers();
--- a/src/matrixrtc/MatrixRTCSession.ts
+++ b/src/matrixrtc/MatrixRTCSession.ts
@@ -801,6 +801,9 @@ export class MatrixRTCSession extends TypedEventEmitter<
 
     // helper variables to make sure we do not have parallel running recalculations.
     private recalculateSessionMembersPromise: Promise<void> = Promise.resolve();
+    // Incremented by each recalculation, so one that has been overtaken can tell.
+    private recalculationGeneration = 0;
+    private latestRecalculation: Promise<void> = Promise.resolve();
 
     /**
      * Ensures that membership is recalculated when the state of the session may have changed.
@@ -828,18 +831,37 @@ export class MatrixRTCSession extends TypedEventEmitter<
      *
      * This function should be called when the room members or call memberships might have changed.
      */
-    private readonly recalculateSessionMembers = async (): Promise<void> => {
+    private readonly recalculateSessionMembers = (): Promise<void> => {
+        const recalculation = this.doRecalculateSessionMembers();
+        this.latestRecalculation = recalculation;
+        return recalculation;
+    };
+
+    private readonly doRecalculateSessionMembers = async (): Promise<void> => {
         // Clear the flag.
         this.membershipNeedsRecalculation = false;
+        const generation = ++this.recalculationGeneration;
         const oldMemberships = this.memberships;
         // Needs to be computed before `this.memberships` is updated below, since it is derived from it.
         const wasKeyRotationSuppressed = this.isKeyRotationSuppressed;
 
-        this.memberships = await MatrixRTCSession.sessionMembershipsForSlot(
+        const memberships = await MatrixRTCSession.sessionMembershipsForSlot(
             this.room,
             this.slotDescription,
             this.calculateMembershipsOpts,
         );
+        // Recalculations are async and can overlap: `_onRTCSessionMemberUpdate` is called once per
+        // membership state event, so a batch of them starts several at once. Each reads the room
+        // state when it starts, and parsing takes longer the more memberships there are, so an
+        // older one can finish last. Its result is stale: drop it and let the newest one win,
+        // otherwise members who have since left linger until the next state change.
+        if (generation !== this.recalculationGeneration) {
+            this.logger.debug("Discarding superseded membership recalculation");
+            // Still let our caller see the final state before continuing.
+            await this.latestRecalculation;
+            return;
+        }
+        this.memberships = memberships;
 
         const changed =
             oldMemberships.length != this.memberships.length ||


ara4n:
I asked fable to clarify this - given naively it looks like the problem here is that we had stale members in indexeddb which flashed up before the server expired them.  However, the actual problem was that the server had expired them, and told us so in /sync *before we joined the call*. The bug here though meant that the membership updates overlapped and raced, so we only actually got accurate state when the next member event changed... which was the one we set for ourself as we (re)joined the call.  So, this PR fixes the "flash of old ghosts" problem by ensuring the ghosts get flushed out before we even join the call.

Arguably we could also address the flash by ignoring cached rtc members and load afresh from the server each time, although that's not obviously correct: if we just did a client refresh, it's probably a good thing to have cached the current members for the room so we can recover rapidly.  So in fact this PR is the right fix.

---

Fable says:

Some more detail on how this presents in practice, since it looks like a brief flash rather than a lingering ghost, and the reason is worth spelling out.

The race needs the client to already hold members who are about to be removed, and then to apply a batch of membership state events in one go. In the rageshake case the "already held" state came from the IndexedDB sync snapshot saved hours earlier, but any live in-memory state from earlier syncs works the same way: a single incremental sync in which two members leave is enough. Each recalculation reads the state when it starts and awaits `parseFromEvent` once per member it sees, so the run that saw more members reliably finishes later. Shrinking state (leaves) is exactly the direction where the older, larger, stale result overtakes the newer, correct one; growing state comes out right for the same reason, which is why joins never showed this.

Timeline at the 21:50 launch from the rageshake, all times UTC:

1. Cached snapshot loads: plum, aqua, egret and the user's old membership are in the session.
2. 21:50:31.988: the first network sync applies its (correct) `state_after`, one event at a time. Five recalculations start; they emit 1, 1, 2, 2, 3 members in that order. The user's own stale membership was dropped, since it was applied first and no larger stale run finished after it, but plum and aqua survive.
3. The user sits in the lobby for ~3s. Nothing renders tiles, nothing changes, the ghosts sit in the session state unseen.
4. 21:50:34.6: Join is clicked. The call view is built from the session as it stands, so blank tiles appear for plum, aqua and egret.
5. 21:50:35.87: the user's join state event is sent; 21:50:36.06 it echoes back over sync. That's a single state event, so one uncontested recalculation runs and reads the state the first sync had already made correct. Plum and aqua vanish. Egret stays, because that membership genuinely was still in room state (a separate issue, #5522).

So the ghost does hang around until the next membership state event, but in the SPA that event is always the user's own join, which lands a second or two after tiles first appear. That is why it looks like a flash. A client that reaches the in-call view without sending a state event, or a batch of leaves arriving mid-call, would keep the ghost until something else changed. The local Playwright reproduction (user A's cache holds B, B churns in and out while A is away, A returns) showed the same: a tile for B for about a second, until A's own join echoed, with the captured `state_after` already marking B as gone.

Johennes:
This looked plausible to me upon review. However, asking Claude to self-review brought up a plethora of potential issues:

```json
{
  "level": "high",
  "findings": [
    {
      "file": "src/matrixrtc/MatrixRTCSession.ts",
      "line": 861,
      "category": "correctness",
      "short_summary": "Newest run rejecting discards every concurrent result",
      "summary": "If the newest recalculation rejects, every concurrent recalculation's result is thrown away, `membershipNeedsRecalculation` has already been cleared so nothing retries, and the rejection is propagated to callers that had a perfectly good result.",
      "failure_scenario": "Verified with an instrumented run: two overlapping `_onRTCSessionMemberUpdate()` calls; run A (gen N) resolves with the correct `[]`, run B (gen N+1) rejects (e.g. `collectMembersEvents` throws, or `callMemberships.sort` hits a `createdTs()` that throws). A hits the `generation !== this.recalculationGeneration` branch, discards its valid `[]`, then `await this.latestRecalculation` re-throws B's error. `sess.memberships` stays at the pre-update value (1 stale member) and both `refreshRoom` callers log `Error updating RTC session members`. Before this PR, A would have applied `[]`. Because line 842 already set `membershipNeedsRecalculation = false`, no recalculation is scheduled, so the ghost member persists until the next unrelated state change — the exact symptom the PR set out to fix. The same coupling means a recalculation that never settles now stalls every earlier in-flight run.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/matrixrtc/MatrixRTCSession.ts",
      "line": 836,
      "category": "correctness",
      "short_summary": "Generation bump and latestRecalculation set in two places",
      "summary": "`recalculationGeneration` is bumped inside `doRecalculateSessionMembers` (line 843) but `latestRecalculation` is assigned by the caller one statement later, so the two are only consistent as long as the synchronous prefix of `doRecalculateSessionMembers` never yields.",
      "failure_scenario": "Correctness of the discard branch rests on an unstated invariant: `latestRecalculation` must already point at the run that owns the current generation. Today that holds only because lines 842-848 (`collectMembersEvents` → `room._unstable_getStickyEvents()`, `isSlotClosed`, `roomState.getStateEvents`) run synchronously. Add any `await` before the generation bump, or move the bump past the first `await`, and a superseded run will `await` the *previous* (already-settled) run's promise, return before the real winner writes `this.memberships`, and hand its caller the ghost-member state again — with no test catching it, since the new test only exercises the happy ordering. Fix: bump the generation and record the promise in one place (e.g. store a deferred created in `recalculateSessionMembers` before invoking the worker).",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/matrixrtc/MatrixRTCSession.ts",
      "line": 834,
      "category": "simplification",
      "short_summary": "Third concurrency mechanism added instead of fixing bypass",
      "summary": "Root cause not addressed: `_onRTCSessionMemberUpdate` bypasses the existing `ensureRecalculateSessionMembers` coalescing, and the PR adds a third concurrency mechanism on top of the two that already exist instead of removing the bypass.",
      "failure_scenario": "`MatrixRTCSession` now carries four overlapping pieces of concurrency state for one operation: `membershipNeedsRecalculation`, `recalculateSessionMembersPromise`, `recalculationGeneration`, `latestRecalculation`. `ensureRecalculateSessionMembers` already does the right thing — a request arriving while a run is queued-but-unstarted is coalesced into it (that run reads state at start, so it sees the newer state), and a request arriving while a run is in flight is queued behind it. The only reason the bug exists is that `_onRTCSessionMemberUpdate` — whose own comment at line 797 says \"this function should only be used for testing purposes\" — is the production hot path (`MatrixRTCSessionManager.refreshRoom:175`) and calls `recalculateSessionMembers` directly. Routing it through `ensureRecalculateSessionMembers` removes the race, the wasted work, and two of the four state variables. The PR body dismisses this because it \"changed the timing that the join/notification path relies on\" — that timing dependency is itself the defect worth fixing rather than working around.",
      "verdict": "PLAUSIBLE"
    },
    {
      "file": "src/matrixrtc/MatrixRTCSessionManager.ts",
      "line": 181,
      "category": "correctness",
      "short_summary": "Duplicate SessionStarted/SessionEnded per sync batch event",
      "summary": "Because all concurrent recalculations now deterministically resolve to the same final state, every in-flight `refreshRoom` for a sync batch emits its own `SessionStarted`/`SessionEnded`, duplicating the event for each membership event in the batch.",
      "failure_scenario": "Verified by replaying `refreshRoom`'s started/ended logic against a real session: two back-to-back updates that take the call from 2 members to 0 emit `[\"SessionEnded\", \"SessionEnded\"]`. Each `refreshRoom` captured `wasActiveAndKnown = true` before its await; with the fix both observe `nowActive === false` afterwards, so both emit. The PR body claims the session manager's started/ended detection \"still observe[s] the final state\" without noting that making all callers agree also makes the duplicate emission systematic — a five-event batch emits five `SessionEnded`s. `refreshRoom` has no dedup, so consumers (Element Call ringing/teardown) see N identical lifecycle events.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "spec/unit/matrixrtc/MatrixRTCSession.spec.ts",
      "line": 843,
      "category": "test-coverage",
      "short_summary": "Regression test relies on microtask-tick timing accident",
      "summary": "The regression test forces the stale run to finish last via an implementation detail (parsing 0 memberships takes fewer microtask ticks than parsing 1) rather than controlling resolution order, so it can silently stop testing anything.",
      "failure_scenario": "The test's whole premise is the comment's \"parsing memberships is async and takes longer the more there are\". If `sessionMembershipsForSlot` ever gains an unconditional `await` before the per-member loop (or `computeBackendIdentityAndVerifyMemberEvents` is batched/parallelised), both runs settle in start order, the stale run lands first, and the test passes against an implementation with the generation check removed — a false negative on the exact bug it guards. I confirmed the test does currently fail on `develop` (`expected [...] to have a length of +0 but got 1`), so it works today; it just isn't robust. Stubbing `MatrixRTCSession.sessionMembershipsForSlot` with two deferreds and resolving the newer one first makes the ordering explicit and also lets the test cover the winner-finishes-first ordering, which is currently untested.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "spec/unit/matrixrtc/MatrixRTCSession.spec.ts",
      "line": 854,
      "category": "test-coverage",
      "short_summary": "Test never asserts MembershipsChanged emission",
      "summary": "The test asserts only the final `memberships` length and never checks the emitted `MembershipsChanged` events, so it does not cover the user-visible symptom (a spurious emission carrying ghost members).",
      "failure_scenario": "The reported failure was Element Call being told the call had members who had left — i.e. a `MembershipsChanged` emission with the stale list, not just a stale field read afterwards. The test would still pass against an implementation that discarded the stale *assignment* but still emitted `MembershipsChanged`/called `membershipManager.onRTCSessionMemberUpdate` with the stale list. Adding a `MatrixRTCSessionEvent.MembershipsChanged` spy and asserting it fired exactly once, with `[]` as the new list, covers the actual regression. (The sibling test three lines above already uses this spy pattern.)",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/matrixrtc/MatrixRTCSession.ts",
      "line": 848,
      "category": "efficiency",
      "short_summary": "Discarded runs still do full O(N*M) membership parse",
      "summary": "Discarded recalculations still perform the full membership parse, so a sync batch of N membership events in a call of M members does O(N*M) parses and per-member async identity hashes and throws N-1 of the results away.",
      "failure_scenario": "The rageshake in the PR description shows five recalculations for one sync batch. Each one runs `collectMembersEvents` plus `CallMembership.parseFromEvent` for every member, and for RTC-kind memberships `parseFromEvent` awaits `computeRtcIdentityRaw` (a crypto digest) per member. With this fix all five still run to completion and four are dropped at line 858 — the work is wasted after the fact rather than avoided. In a large call re-joined after a long absence (many `m.call.member` changes in `state_after`) this is the main CPU cost on the first incremental sync. Coalescing through `ensureRecalculateSessionMembers` would collapse the batch to one or two runs.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/matrixrtc/MatrixRTCSession.ts",
      "line": 802,
      "category": "correctness",
      "short_summary": "Field comment contradicts new parallel design",
      "summary": "The comment heading the concurrency fields now states the opposite of the design: it claims recalculations never run in parallel, which is precisely the situation this PR exists to handle.",
      "failure_scenario": "`// helper variables to make sure we do not have parallel running recalculations.` now heads all three fields including the two new ones, whose entire purpose is to cope with recalculations that *do* run in parallel. `ensureRecalculateSessionMembers`'s TSDoc (line 809) likewise still says \"Also ensures that only one recalculation is made at a time\". The next reader debugging a membership race will read these, conclude recalculations are serialised, and look elsewhere — which is roughly how this bug survived. The comments should say that recalculations may overlap and that the newest generation wins.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/matrixrtc/MatrixRTCSession.ts",
      "line": 859,
      "category": "observability",
      "short_summary": "Discard logged at debug with no generations or count",
      "summary": "The only diagnostic for a discarded recalculation is a context-free `debug` log, omitting the two numbers needed to diagnose the next occurrence.",
      "failure_scenario": "This bug was found from a rageshake whose value came from the `info`-level \"Memberships ... have changed: emitting (N members)\" lines. `debug` is commonly filtered out of rageshakes, and the message carries neither the generation numbers nor how many memberships were discarded, so a future recurrence (e.g. the winner-rejects path above) leaves no trace correlating with the surviving `info` lines. Log the generation, the current generation and `memberships.length`, and consider `info` to match the adjacent emission log.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/matrixrtc/MatrixRTCSession.ts",
      "line": 840,
      "category": "simplification",
      "short_summary": "TSDoc left on wrapper; needless arrow class field",
      "summary": "`doRecalculateSessionMembers` is an undocumented private arrow class field that is never used as a callback, while the descriptive TSDoc block above now documents the three-line wrapper instead of the function it describes.",
      "failure_scenario": "The block comment \"Call this when anything that could impact rtc memberships has changed... Examines the latest call memberships and handles any encryption key sending or rotation that is needed\" now sits on `recalculateSessionMembers`, which does none of that, while the 60-line function that does has no doc at all. The arrow-field form also allocates a closure per `MatrixRTCSession` instance for no benefit — it is only ever called as `this.doRecalculateSessionMembers()`, so a plain `private async doRecalculateSessionMembers(): Promise<void>` method is equivalent and cheaper. Move the existing TSDoc down onto it and give the wrapper a one-liner explaining that it records the newest run.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "spec/unit/matrixrtc/MatrixRTCSession.spec.ts",
      "line": 852,
      "category": "simplification",
      "short_summary": "flushPromises() after Promise.all is dead code",
      "summary": "`await flushPromises()` after `await Promise.all([first, second])` is dead code.",
      "failure_scenario": "`first` is the superseded run; it only resolves after `await this.latestRecalculation`, i.e. after `second` has finished writing `this.memberships` and running every side effect. By the time `Promise.all` resolves there is nothing left to flush, so the extra tick neither stabilises nor guards anything — it just implies to the next reader that the assertion needs extra settling, inviting cargo-culting into future tests.",
      "verdict": "CONFIRMED"
    }
  ]
}
```

Johennes (approved):
I think Claude missed the root cause of the issue. We're calling `_onRTCSessionMemberUpdate` in the production code but that method was only ever meant to be used for testing purposes.

```typescript
// We allow this name schema since this function should only be used for testing purposes.
public _onRTCSessionMemberUpdate = async (): Promise<void> => {
```


In fact, we already have a safe method that enforces serialization: `_onRTCSessionMemberUpdate`

```typescript
/**
 * Ensures that membership is recalculated when the state of the session may have changed.
 * Also ensures that only one recalculation is made at a time.
 * @returns A promise resolving when the state has been recalculated.
 */
private ensureRecalculateSessionMembers(): Promise<void> {
```

I have changed the code to use that method instead and extended the tests.

PR #5525, opened 2026-09-08
Treat an event as read if the user sent a later one

### LINKED ISSUE #34904: doesTimelineHaveUnreadMessages: thread stays unread when a non-counting event lands after your own reply
### Steps to reproduce

1. Alice starts a thread in a room and posts a message.
2. Bob (you) reads it and replies in the thread.
3. Alice — or anyone other than Bob — then sends an event into that thread that does **not** trigger
   an unread count: a reaction, an edit, a redacted message, an `m.room.member` change, a beacon, or
   any event with no renderer.
4. Look at the thread's unread state on Bob's client (`doesTimelineHaveUnreadMessages`, e.g. via the
   Threads Activity Centre or the room list unread dot).

### Outcome

**What did you expect?**

The thread reads as read. Bob replied *after* the only incoming message, so he has demonstrably seen
it, and nothing unread-triggering has arrived since.

**What happened instead?**

The thread reads as **unread**, and stays that way until Bob gets a threaded read receipt covering
Alice's message.

### Analysis

`doesTimelineHaveUnreadMessages` ([`Unread.ts:67`](https://github.com/element-hq/element-web/blob/develop/apps/web/src/Unread.ts#L67))
does:

```ts
const latestImportantEventId = findLatestImportantEvent(room.client, timeline)?.getId();
if (latestImportantEventId) {
    return !room.hasUserReadEvent(myUserId, latestImportantEventId);
}
```

Two things combine:

1. `findLatestImportantEvent` filters through `eventTriggersUnreadCount`, which returns `false` for
   our own events ([`Unread.ts:24-27`](https://github.com/element-hq/element-web/blob/develop/apps/web/src/Unread.ts#L24-L27)).
   So the "latest important event" is the newest message **from someone else** — Alice's message in
   the repro, never Bob's reply.

2. `room.hasUserReadEvent()` resolves in matrix-js-sdk `RoomReceipts.hasUserReadEvent()`. With no
   receipt covering Alice's message, the only thing that can save us is this shortcut:

   ```ts
   // TODO: what if they sent the second-last event in the thread?
   if (this.userSentLatestEventInThread(threadId, userId)) {
       return true;
   }
   ```

   and `userSentLatestEventInThread` only checks the *very last* timeline entry:

   ```ts
   return !!(timeline && timeline.length > 0 && timeline[timeline.length - 1].getSender() === userId);
   ```

   (`src/models/room-receipts.ts:161` and `:185` on `develop` — note the pre-existing TODO on the
   line above the shortcut; this exact case is already flagged there.)

So as soon as *any* event lands after Bob's reply, the "I sent the last message, it's all read"
shortcut stops applying, while the event Bob is being judged against is still Alice's older message.
The thread flips back to unread.

The reply itself doesn't need to be the trigger — any non-counting event after our latest reply
reproduces it, which makes reactions on threads a fairly easy way to hit this.

### Possible fix

Rather than "did the user send the literal last event", the question should be "is the user's latest
event at or after the latest incoming event". That could be fixed in either layer:

- in `doesTimelineHaveUnreadMessages` / `findLatestImportantEvent` in element-web, or
- in the js-sdk shortcut, which is where the existing TODO lives.

### Context

Raised in review of [#32851](https://github.com/element-hq/element-web/pull/32851) (Threads Activity
Centre cross-room threads panel), where the TAC currently carries a local guard
(`hasUnreadAfterMyLatestReply`) to suppress the false positive. Filing this so the guard can be
removed once the underlying behaviour is fixed.

### Environment

- Application version: `develop` (analysis against `develop` at time of drafting)
- Homeserver: any — reproduces with local timeline inspection, independent of server notification
  counts.

nathanael-h:
Moved here from element-hq/element-web#34905 at @Half-Shot's request: the fix belongs in the js-sdk rather than being patched around in Element Web.

Fixes element-hq/element-web#34904

### Problem

When no receipt covers an event, `RoomReceipts.hasUserReadEvent` falls back to a shortcut: if the user sent the *literal last* event in that event's timeline, treat everything in the timeline as read.

```ts
// TODO: what if they sent the second-last event in the thread?
if (this.userSentLatestEventInThread(threadId, userId)) { ... }
```

```ts
return !!(timeline && timeline.length > 0 && timeline[timeline.length - 1].getSender() === userId);
```

That leaves the TODO's own question unanswered, plus two gaps:

1. **Anything landing after our own message takes the timeline back to unread.** A reaction, an edit, a redaction or a membership change arrives after our reply and the shortcut stops applying — judged against a message we had already read before replying.
2. **A thread root is never covered by it.** Our reply lives in the thread's timeline, the root lives in the main one (`inMainTimelineForReceipt` returns true for thread roots), so no receipt of ours — synthetic or real — can point at the root, and the shortcut looks at the main timeline, where the root itself is typically the last event.

Gap 2 is what Element Web hits in element-hq/element-web#34904: `doesTimelineHaveUnreadMessages` asks whether we have read the latest *important* event of a thread, our own events are never "important", so the question becomes "have we read this thread root?" — and the answer was no, however many times we had replied in it.

### Change

Ask the right questions instead:

- did the user send an event at or after this one, in this event's timeline? (`userSentEventAfterOrSame`, using the existing `isAfterOrSame` ordering rather than "is it the last one");
- and, for a thread root, did the user reply in the thread it roots? (`userRepliedInThread` — you can't reply to a thread without seeing its root.)

Events that are still on their way to the server, or failed to get there, are ignored: they have no place in the timeline yet, and a failed send tells us nothing about what the user has seen.

### Tests

Five tests in `spec/unit/models/room-receipts.spec.ts`, three of which fail on `develop`:

| test | on `develop` |
| --- | --- |
| read if we sent a later event in the thread, even if we didn't send the last | ❌ fails |
| unread for a thread event that arrived after the last one we sent | ✅ passes (guard against over-eagerness) |
| events still being sent are ignored | ❌ fails |
| a thread root is read if we replied in its thread | ❌ fails |
| a thread root is unread if we did not reply in its thread | ✅ passes (guard) |

The first three need a thread whose events came in via pagination rather than live, so that no synthetic receipts exist for their senders and the fallback is actually exercised — hence the new `setupPaginatedThread` helper.

`vitest run spec/` shows no new failures (the 16 pre-existing failures in this working copy are all in `spec/unit/crypto/**` and `spec/integ/crypto/**`, untouched here), and `tsc --noEmit` and `oxfmt --check` are clean for the changed files.

### Follow-up

element-hq/element-web#34905 becomes unnecessary with this: `doesTimelineHaveUnreadMessages` gets the right answer from the js-sdk with no change of its own. I'll close it once this lands, along with the local guard that element-hq/element-web#32851 currently carries in `useUnreadThreadRooms.ts`.

### DIFF
--- a/spec/unit/models/room-receipts.spec.ts
+++ b/spec/unit/models/room-receipts.spec.ts
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 import {
+    EventStatus,
     FeatureSupport,
     type MatrixClient,
     MatrixEvent,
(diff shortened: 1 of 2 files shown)


florianduros (approved):
Thanks, this looks great!

PR #5526, opened 2026-09-09
Fix type definitions on `import|exportSecretsBundle`

richvdh:
Currently these are defined in terms of the return type of `SecretsBundle.to_json`. The problem is that `SecretsBundle.to_json` is only typed via a `declare module` doodad in matrix-js-sdk, which means it is not visible to downstream projects. Instead, they end up thinking that `exportSecretsBundle` returns an `unknown`.

Fix this by factoring out the definition to a new type, and returning that explicitly.

### DIFF
--- a/src/@types/matrix-sdk-crypto-wasm.d.ts
+++ b/src/@types/matrix-sdk-crypto-wasm.d.ts
@@ -16,20 +16,23 @@ limitations under the License.
 
 import type * as RustSdkCryptoJs from "@matrix-org/matrix-sdk-crypto-wasm";
 
+/** Return type of `SecretsBundle.to_json` */
+type SecretsBundleJson = {
+    cross_signing: {
+        master_key: string;
+        self_signing_key: string;
+        user_signing_key: string;
+    };
+    backup?: {
+        algorithm: string;
+        key: string;
+        backup_version: string;
+    };
+};
+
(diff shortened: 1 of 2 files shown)


andybalaam (approved):
Great, thank you!

t3chguy:
@richvdh what exact error/issue/symptom is this fixing in terms of visible developer impact on the consumer/app?

richvdh:
> @richvdh what exact error/issue/symptom is this fixing in terms of visible developer impact on the consumer/app?

In element-web (and other downstream users of matrix-js-sdk, I assume), if you just `pnpm install`, rather than doing something custom for matrix-js-sdk, then you get compilation errors of this nature:

```
../../node_modules/.pnpm/matrix-js-sdk@file+matrix-js-sdk+matrix-js-sdk.tgz/node_modules/matrix-js-sdk/src/rendezvous/MSC4108SignInWithQR.ts:109:50 - error TS2312: An interface can only extend an object type or intersection of object types with statically known members.

109 interface SecretsPayload extends MSC4108Payload, Awaited<ReturnType<NonNullable<CryptoApi["exportSecretsBundle"]>>> {
```

t3chguy (approved):

richvdh:
Tests failing due to https://github.com/element-hq/element-web/issues/33663 and https://github.com/matrix-org/matrix-js-sdk/issues/5534

PR #5541, opened 2026-09-13
Preserve read receipts of different types during sync accumulation

### LINKED ISSUE #5540: ReceiptAccumulator drops receipt types for the same user and thread
## Problem

`ReceiptAccumulator` drops either the public or private read receipt when both belong to the same user and thread.

The map is keyed by user, but it must also be keyed by receipt type. Today, `m.read.private` can overwrite `m.read`, or the reverse.

This affects both unthreaded and threaded receipts in versions `41.3.0-rc.0` and `42.3.0`.

## Impact

The client can save the incomplete receipts together with a newer sync token. After restart, the missing receipt may not be sent again.

This can leave unread state stale across clients.

## Minimal reproduction

```ts
const accumulator = new ReceiptAccumulator();
accumulator.consumeEphemeralEvents([
  newReceipt("$public", ReceiptType.Read, "@me:example.org", 200),
  newReceipt("$private", ReceiptType.ReadPrivate, "@me:example.org", 100),
]);

accumulator.buildAccumulatedReceiptEvent("!room:example.org");
```

**Actual:** only the private receipt remains.

**Expected:** both receipt types remain.

## Fix

Store the latest receipt per:

- unthreaded: user + receipt type;
- threaded: thread + user + receipt type.

A newer receipt of the same type must still replace the older one.

rubinsh:
Fixes #5540.

## Problem

`ReceiptAccumulator` stores one receipt per user and thread. Public and private read receipts therefore overwrite each other.

A saved sync can keep the newer sync token but lose one receipt. After restart, the client may never receive that receipt again.

## Fix

Store receipts by type as well as user and thread.

This preserves `m.read` and `m.read.private` independently. A newer receipt still replaces an older receipt of the same type.

## Testing

```bash
pnpm exec vitest run spec/unit/receipt-accumulator.spec.ts
pnpm test
pnpm lint
pnpm build
pnpm lint:knip
```

The parameterized regression covers both unthreaded and threaded public/private receipts. Existing tests cover replacement within the same type.

Local results:

- focused tests: 9/9;
- full suite: 147 files, 3,065 passed, 10 skipped, 13 todo;
- types, lint, formatting, workflow validation, build, and knip: passed.

Notes: Preserve public and private read receipts independently in saved sync state.

Maintainer: please add the `T-Defect` label.

## Checklist

- [x] Tests added.
- [x] No public API added.
- [x] Local checks pass.
- [x] Commit is signed off.


florianduros (approved):
Seems good, thanks

PR #5542, opened 2026-09-14
Fix sync.ts to emit state events when msc4222 state_after is enabled

maheichyk:
When MSC4222 is enabled, updates to the state are coming via `state_after`. In this case a `state` is omitted and `timeline` may be limited and not necessary to contain all the state.

Currently, `MatrixClient` doesn't use `state_after` (or `stateAfterEvents` is the code) to emit the events.

This PR fixes the issue and changes `MatrixClient` to emit the state events when `state_after` is enabled.

Edit: PR is updated to consider "duplicate" state events in the timeline section that could be re-emitted. Added tests to cover different syc responses with and without of `state_after`.

## Checklist


src/sync.ts
```diff
@@ -1487,7 +1487,11 @@ export class SyncApi {
                 // should this get a sync section as an additional event emission param (e, syncSection))?
                 stateEvents.forEach(emitEvent);
                 stateAfterEvents.forEach(emitEvent);
-                timelineEvents.forEach(emitEvent);
+                const filteredTimelineEvents =
```
dbkr, line 1493: Could this just get a comment please? If I didn't have the context of this PR this would be fairly non-obvious to me why this is doing this.

maheichyk: sure, have added a comment to the timeline filtering

dbkr (approved):
Thanks. Please remember not to force push, it completely breaks github's PR interface.

PR #5543, opened 2026-09-14
Fix cached sync waiting for missing relation parents

basnijholt:
We ran into this in [MindRoom Chat](https://github.com/mindroom-ai/mindroom-chat), our Matrix client for [MindRoom](https://github.com/mindroom-ai/mindroom).
On iOS, reopening the app could leave it sitting on the loading screen even though chats were already cached.
While investigating, we found that replaying a saved sync can wait for the server when a cached edit or reaction refers to a message that has fallen out of the saved timeline.

With thread support enabled, `Room.addLiveEvents` waits for `fetchRoomEvent` to retrieve the missing parent, even when `fromCache` is true.
Cache replay waits for every room before emitting `PREPARED`, so one slow or unreachable server request can keep the loading screen up despite having chats available locally.
We reproduced this with the published 42.3.0 package: with the parent request held open, replay stalled and the next cached message never reached the timeline; releasing the request let it continue.
The same lookup is still present on `develop`.

The change skips the parent request during cache replay.
The relation stays in the room's existing relation collection and attaches when its parent is loaded later.
Live events still fetch missing parents as before.
Our downstream patch and persisted IndexedDB regression tests are in [mindroom-ai/mindroom-chat#245](https://github.com/mindroom-ai/mindroom-chat/pull/245).
This addresses the network wait we reproduced; we haven't measured the overall launch-time improvement on a physical iPhone yet.

The new tests cover a parent request that never returns, edits and reactions attaching when the parent arrives later, and the existing live-event lookup.
To try them locally:

1. Run `pnpm install --frozen-lockfile`.
2. Run `pnpm test spec/unit/room.spec.ts`.
3. Remove `!fromCache &&` from the new condition and rerun the tests: all three cache regressions fail, while the live-event test still passes.

All 3,070 tests pass with `pnpm test --maxWorkers=8`.
`pnpm build`, `pnpm lint:types`, and `pnpm lint:js` also pass.

The upstream workflows are waiting for maintainer approval.
Please approve those runs and add `T-Defect` for the changelog check.

## Checklist

- [x] New or updated `public`/`exported` symbols have accurate TSDoc documentation (no public API changes).
- [x] Sign-off given on the changes.


florianduros (approved):
Thanks for the fix, looks good

PR #5544, opened 2026-09-14
Update RTC notification and decline events as per MSC4075

Johennes:
This updates the code around RTC notifications and declines to the latest state of [MSC4075](https://github.com/matrix-org/matrix-spec-proposals/pull/4075/changes). This is fairly large but can be reviewed by commit.

## Checklist


src/@types/event.ts
```diff
@@ -361,7 +361,7 @@ export interface TimelineEvents {
     [M_BEACON.name]: MBeaconEventContent;
     [M_POLL_START.name]: PollStartEventContent;
     [M_POLL_END.name]: PollEndEventContent;
-    [EventType.RTCMembership]: RtcMembershipData | { msc4354_sticky_key: string }; // An object containing just the sticky key is empty.
+    [EventType.RTCMembership]: RtcMembershipData | { slot_id?: string; msc4354_sticky_key: string };
```
robintown: Can we make `slot_id` required on this type?

I'm getting deja vu asking this, and it's not necessarily a concern for this PR, but I also wonder why there are no `member` and `leave_reason` properties on the leave variant.

spec/unit/matrixrtc/MatrixRTCSession.spec.ts
```diff
@@ -436,6 +436,25 @@ describe("MatrixRTCSession", () => {
             expect(sess?.memberships[0].isExpired()).toEqual(false);
             expect(sess?.slotDescription.id).toEqual("ROOM");
         });
+        it("ignores left sticky memberships", async () => {
+            const mockRoom = makeMockRoom([]);
+            mockRoom._unstable_getStickyEvents.mockImplementation(() => {
+                // Left memberships only contain the slot ID and the sticky key.
+                const ev = mockRTCEvent(
+                    { user_id: "@left:user.example", slot_id: "m.call#ROOM", msc4354_sticky_key: "MEMBER" },
```
robintown, line 451: Also a `member` object and optional `leave_reason`

src/matrixrtc/MatrixRTCSession.ts
```diff
@@ -265,7 +267,7 @@ export class MatrixRTCSession extends TypedEventEmitter<
 > {
     private membershipManager?: IMembershipManager;
     private encryptionManager?: IEncryptionManager;
-    private joinConfig?: SessionConfig;
+    private joinConfig?: JoinSessionConfig;
```
robintown: I believe this having type `SessionConfig` was intentional, the idea being that the join method hands off the membership and encryption configs to their respective modules and then stores only the properties relevant to the session as a whole, to promote separation of concerns.

spec/unit/matrixrtc/MatrixRTCSession.spec.ts
```diff
@@ -736,10 +750,74 @@ describe("MatrixRTCSession", () => {
...
+            sendStickyEventMock.mockRejectedValueOnce(
+                new UnsupportedStickyEventsEndpointError("nope", "sendStickyEvent"),
+            );
+            sendEventMock.mockResolvedValueOnce({ event_id: "regular-evt" });
+            const didSendNotification = new Promise((resolve) => {
+                sess!.once(MatrixRTCSessionEvent.DidSendCallNotification, resolve);
+            });
+
+            sess!["joinConfig"] = { notificationType: "ring", notificationLifetimeMs: 30_000 };
+            sess!["sendCallNotify"]("$membership", "ring");
```
robintown: I was looking for where this test calls `joinRTCSession` and didn't spot this line at first, it feels like either these tests should simulate a full join or the `sendCallNotify` method deserves to be a standalone function.

src/matrixrtc/types.ts
```diff
@@ -173,7 +181,16 @@ export interface IRTCNotificationContent extends RelationEvent {
...
-export interface IRTCDeclineContent extends RelationEvent {}
+export interface IRTCDeclineContent extends RelationEvent {
+    /**
+     * The sticky key as per MSC4354. Must be equal to the event ID of the notification being declined.
+     *
+     * Optional because servers without MSC4354 support can't deliver the decline as a sticky event. Such a
+     * decline still counts, but only for devices that receive it directly, since it isn't re-delivered after
+     * a gappy or initial sync.
+     */
+    msc4354_sticky_key?: string;
```
robintown, line 486: From reading the client.ts code, it looks like we set this field even when the event does not truly have a sticky duration? So I believe it could be required, same as `IRTCNotificationContent["msc4354_sticky_key"]`.

src/matrixrtc/membershipData/common.ts
```diff
@@ -14,6 +14,19 @@ See the License for the specific language governing permissions and
...
+/**
+ * Whether the content of an `m.rtc.member` or `m.call.member` event represents a left membership.
+ *
+ * Leaving a slot is expressed by emptying the event's content, so nothing is left beyond the slot the
+ * membership applied to and the MSC4354 sticky key. Events from before `slot_id` was added carry neither,
+ * and so can't be attributed to a slot at all.
+ */
+export function isLeftMembershipContent(content: IContent): boolean {
+    return Object.keys(content).every((key) => key === "slot_id" || key === "msc4354_sticky_key");
+}
```
robintown, line 27: ^ Another place where we would need to account for `member` and `leave_reason`.

Though hm, I'm not sure if exhaustively checking the keys is the right choice, as it breaks if we ever extend membership `content` with a new property. Shouldn't we check `member.membership`? It seems like matrix-js-sdk missed the MSC update that introduced `member.membership` in general.

robintown (commented):
My comments so far from reviewing commits 1-5:

src/@types/event.ts
Johennes: > Can we make `slot_id` required on this type?

Hm, I'm not sure. It was missing from the old type for left events?

> I'm getting deja vu asking this, and it's not necessarily a concern for this PR, but I also wonder why there are no `member` and `leave_reason` properties on the leave variant.

`leave_reason` sits idle in https://github.com/matrix-org/matrix-js-sdk/pull/5437. `member` we just documented as a known implementation gap on the MSC since the ID is just forwarded from the joined event and `membership` is always constant.

Johennes: (Needless to say, the implementation differences are a _huge_ mess that doesn't help at all in implementing further changes.)

spec/unit/matrixrtc/MatrixRTCSession.spec.ts
Johennes: Those are intentionally missing currently as explained in https://github.com/matrix-org/matrix-js-sdk/pull/5544#discussion_r4015540534.

src/matrixrtc/membershipData/common.ts
Johennes: Yeah, it does (see https://github.com/matrix-org/matrix-js-sdk/pull/5544#discussion_r4015540534). I think you have a point with not exhaustively checking the properties. Though maybe we should make that change together with introducing `member`? This PR basically just adds `slot_id` to the existing check.

src/matrixrtc/types.ts
Johennes: Yes, we still set it. I'm not sure if we can make it required here, though. Would that not create a problem in parsing incoming events from older clients then?

src/matrixrtc/MatrixRTCSession.ts
Johennes: Oh, I see. Turns out this wasn't even needed anymore and just a leftover from an earlier iteration. Reverted it in 7b3a8833be3cedaf1ee2fd4d39f2d4b8fecf4989.

spec/unit/matrixrtc/MatrixRTCSession.spec.ts
Johennes: Yeah, good point. Have re-arranged the tests in 39d69f42699c1e943487af8362bfcb0f78658114.

src/@types/event.ts
robintown: > Hm, I'm not sure. It was missing from the old type for left events?

I think that's okay, as this is the Matrix 2.0 event type which nobody is using in production so far

> `leave_reason` sits idle in #5437. `member` we just documented as a known implementation gap on the MSC since the ID is just forwarded from the joined event and `membership` is always constant.

Ah, thanks for the reminder. I will continue the review with the gaps in mind.

src/matrixrtc/membershipData/common.ts
robintown: Agreed, let's do that later

src/matrixrtc/types.ts
robintown: True, and I see now that `IRTCNotificationContent["msc4354_sticky_key"]` becomes optional in a later commit as well. I take it you're trying to make it possible to adopt `parseCallNotificationContent` as soon as we want in element-web and drop its custom parsing, which makes a lot of sense to me.

src/@types/event.ts
Johennes: > I think that's okay, as this is the Matrix 2.0 event type which nobody is using in production so far

Oh right. This is the membership event, not the invite. 🤦‍♂️ Have made it required with 4de76475374d9de701829555450a2f240b1bd2e5.

src/matrixrtc/types.ts
```diff
@@ -123,35 +129,314 @@ export const RTC_NOTIFICATION_MAX_LIFETIME_MS = 2 * 60 * 1000; // 2 minutes
...
+function isNonNegativeInteger(value: unknown): value is number {
+    return Number.isInteger(value) && (value as number) >= 0;
+}
+
+/**
+ * Looks up a membership event by ID in the room's timeline and sticky event map.
+ *
+ * @returns The event, or `undefined` if it isn't available locally.
+ */
+function findMembershipEvent(room: RTCNotificationValidationContext["room"], eventId: string): MatrixEvent | undefined {
```
robintown: Might be nice to call this `findTimelineOrStickyEvent` to reflect that it doesn't check the event type

robintown: Should this check the room state map as well, so it can find `org.matrix.msc3401.call.member` state events as reliably as sticky events?

(I actually wonder if we *only* need the state events check. If a client is sending sticky `org.matrix.msc4143.rtc.member` events, we can probably assume that they have an up to date invites implementation and would have included a `slot_id`.)

src/matrixrtc/types.ts
```diff
@@ -123,35 +129,314 @@ export const RTC_NOTIFICATION_MAX_LIFETIME_MS = 2 * 60 * 1000; // 2 minutes
...
+            throw new Error("Notification is not the current sticky event for its sender and slot");
+        }
     }
 
-    const cappedLifetime = content["lifetime"] >= 90000 ? 90000 : content["lifetime"];
-    return { ...content, lifetime: cappedLifetime } as IRTCNotificationContent;
+    // `isSlotClosed` returns undefined when no slot event exists at all, which is just as disqualifying.
+    if (isSlotClosed(room, slotIdToDescription(slotId)) !== false) {
+        throw new Error(`Slot ${slotId} does not exist or is not open`);
+    }
```
robintown, line 346: So this will break notifications for rooms without a slot, right? I think we need to postpone this check until the slots migration has happened, or make it conditional on the membership event type or something.

src/matrixrtc/types.ts
```diff
@@ -123,35 +129,314 @@ export const RTC_NOTIFICATION_MAX_LIFETIME_MS = 2 * 60 * 1000; // 2 minutes
...
+    // expired one would suppress notifications for hours.
+    const expiry = membership.getAbsoluteExpiry();
+    return expiry === undefined || expiry > now;
+}
+
+/**
+ * Whether the receiving user has declined the given notification.
+ *
+ * Declines are looked up through their `m.reference` relation rather than the sticky event map, so that a
+ * decline still counts when it couldn't be sent as a sticky event.
```
robintown: ~~Checking the sticky map as well would be nice for completeness of the MSC implementation, if not too annoying~~ Ah, but I guess that the event being sticky actually guarantees that it will end up in the set of relations known to the client, so no need.

src/models/room-sticky-events.ts
```diff
@@ -189,6 +196,27 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
...
+     */
+    private addStickyEventOnceDecrypted(event: MatrixEvent): void {
+        const onEventDecrypted = (): void => {
+            if (event.isDecryptionFailure()) {
+                // The event may still be decrypted later, e.g. once the keys arrive. Keep listening.
+                event.once(MatrixEventEvent.Decrypted, onEventDecrypted);
+                return;
+            }
+            // The event has expired while we were waiting for it, so there is nothing to add.
+            if (event.unstableStickyExpiresAt !== undefined && event.unstableStickyExpiresAt <= Date.now()) return;
```
robintown: Looks like `addStickyEvents` already performs this check: https://github.com/matrix-org/matrix-js-sdk/blob/15ed5fe089394f54e52eb6a0aa21e1ac791230af/src/models/room-sticky-events.ts#L146

src/models/room-sticky-events.ts
```diff
@@ -131,6 +131,13 @@ export class RoomStickyEventsStore extends TypedEventEmitter<RoomStickyEventsEve
      *          and the previous event it may have replaced.
      */
     private addStickyEvent(event: MatrixEvent): { added: true; prevEvent?: StickyMatrixEvent } | { added: false } {
+        // The map is keyed on the event's type and sticky key, both of which are encrypted. Wait for the
+        // event to be decrypted, otherwise it would be filed under `m.room.encrypted` with no sticky key.
+        if (event.isBeingDecrypted() || event.shouldAttemptDecryption()) {
+            this.addStickyEventOnceDecrypted(event);
+            return { added: false };
+        }
```
robintown, line 139: I was going to say I see the potential for races here, but upon further inspection `addStickyEvent` always sorts events by timestamp when updating a keyed entry. May or may not be worth a comment

src/embedded.ts
```diff
@@ -550,6 +550,12 @@ export class RoomWidgetClient extends MatrixClient {
      * A widget can't probe the homeserver for MSC4354 support and doesn't need to: the host either granted it
      * the capability to send sticky events or it didn't. A missing capability is reported the same way a
      * missing server feature is, so that callers fall back to a regular event in both cases alike.
+     *
+     * The host may still refuse the request, typically because its own homeserver turns out not to support
+     * sticky events. Such a refusal reaches the widget without Matrix API error details, as those only accompany
+     * errors from requests the host actually made to the homeserver. It is therefore reported as unsupported too,
+     * with the host's error as `cause`. A refusal that does carry Matrix API error details came from the
+     * homeserver itself and is rethrown as is.
```
robintown, line 558: This makes me wonder whether widget hosts should proactively deny capabilities that they know their homeserver to not support, though that discussion is probably best left for another time.

robintown (commented):
OK, have looked fairly thoroughly through the rest of the changes.

src/matrixrtc/types.ts
Johennes: Good point. Changed in 1599dbe1584b4d4961da14c5265f885b124aa8e3.

src/matrixrtc/types.ts
Johennes: Yes, though I _think_ this should be fine. Element Web doesn't actually call this function at the moment: https://github.com/element-hq/element-web/blob/52c6d24dbb756118ec07fd405b3e681b7e96cc6c/apps/web/src/Notifier.ts#L611

My thought was to add another path in Element Web that calls this functions only when the slots lab feature is turned on.

src/embedded.ts
Johennes: Yeah, they probably should, agree.

src/matrixrtc/types.ts
Johennes: Oh. Yes, you're right on both. Have fixed this in ef6df0a7aede10d16be69644062dc24289f43e6d.

src/models/room-sticky-events.ts
Johennes: Good catch. Removed in 24cf29d7e7d66e859812d53b8fd17b00eed8a626.

src/models/room-sticky-events.ts
Johennes: Have added a comment in 4e09985999e56dfd3ebb19e654a290c2b1e3e558.

src/matrixrtc/types.ts
robintown: > My thought was to add another path in Element Web that calls this functions only when the slots lab feature is turned on.

OK, sounds reasonable

robintown (approved):

src/matrixrtc/types.ts
Johennes: Just pushed my draft EW PR here: https://github.com/element-hq/element-web/pull/35089 (only last commit is the actual change)

Johennes:
Sorry, I noticed that I forgot to add the push rule IDs. Have added them in 49e3e62bdd2b26e368db1516ddd702d7d3628671.

src/matrixrtc/types.ts
```diff
@@ -106,57 +112,379 @@ export type RTCNotificationType = "ring" | "notification";
...
+     *
+     * Optional because servers without MSC4354 support can't deliver the decline as a sticky event. Such a
+     * decline still counts, but only for devices that receive it directly, since it isn't re-delivered after
+     * a gappy or initial sync.
+     */
+    msc4354_sticky_key?: string;
+}
 
 export enum Status {
     Disconnected = "Disconnected",
```
dbkr, line 490: Probably not one for this PR which is complex enough already, but 'types/ts/ seems to bed a lot more than just types so feels like it needs a rename / refactor.

src/matrixrtc/utils.ts
```diff
@@ -67,3 +70,42 @@ export function slotIdToDescription(slotId: string): SlotDescription {
...
+    return slotEvent.getContent<RtcSlotEventContent>();
+}
+
+/**
+ * Whether the given slot is closed.
+ *
+ * @returns `true` if the slot is closed, `false` if the slot is open or `undefined`
+ * if no slot exists.
+ */
+export function isSlotClosed(
```
dbkr, line 99: This being 'isClosed' is causing double negatives that make my head hurt, especially since when used it's compared to false - isSlotOpen would feel more intuitive unless there's a specific reason for it to be like this?

dbkr (approved):
otherwise lgtm I think

src/matrixrtc/types.ts
Johennes: Hm, yeah good point.

src/matrixrtc/utils.ts
Johennes: Good call. Unfortunately, this one is already public and used in EW. I have added a matching `isSlotOpen` in 56ebf9217b4eaab204cd56e7b33f321534751b93 though which makes the condition read a little easier.

PR #5546, opened 2026-09-15
Tweak npm publishing to wait for package being ready

t3chguy:


dbkr (approved):
ah, so it's actually become slower

PR #5550, opened 2026-09-15
Let applications publish their own data in MatrixRTC memberships

ara4n:
The `application` object of an `m.rtc.member` event may carry keys specific to the application, and `CallMembership.applicationData` already reads them back, but there was no way to send any: the membership manager builds the content itself.

- `SessionConfig.applicationData` is merged into the `application` object of the membership (or, for legacy `m.call.member` events, spread at the top level beside the session fields, which is where `applicationData` now reads it from too; the session's own fields always win).
- `MatrixRTCSession.updateApplicationData()` replaces it mid-call, re-sending the membership as `updateCallIntent` does, and is a no-op when it hasn't changed.

Element Call needs this to publish a participant's consent to being recorded in their membership, where the recorder can find it

## Checklist


src/matrixrtc/CallMembership.ts
```diff
@@ -261,10 +275,14 @@ export class CallMembership {
...
-                // SessionData does not have application data as such. We return specific
-                // properties in use by other getters in this class, for compatibility.
-                return { "type": data.application, "m.call.intent": data["m.call.intent"] };
+            default: {
+                // SessionData has no application object: the application's own
+                // data sits at the top level beside the session fields, so
+                // everything but those is the application's
+                const { application, ...rest } = data;
+                for (const key of sessionFields) delete rest[key as keyof typeof rest];
+                return { ...rest, "type": application, "m.call.intent": data["m.call.intent"] };
```
Johennes, line 284: This feels like it's going to be brittle but I guess it's acceptable as presumably we will knife the legacy membership path entirely in the medium term.

Johennes (approved):

